PR #22358, opened 2025-10-20
[NFC] Merge common type constraints

krzysz00:
We've got some replication of the TensorOrVector and TensorOrMemref definitions across dialects, so I've gone and moved them into a common utility file.


hanhanW (commented):
Thanks for the cleanup. LGTM if you fix the mismatching error.

hanhanW (approved):

PR #22339, opened 2025-10-16
[DT][NFCI] Implement getOffsetsSizesStrides for GPU padding resolver.

hanhanW:
The revision implement the interface method with the existing logic; it allows us to remove padding specific load/store patterns, because they use interfaces.

No new tests are added since it aims to be a refactor PR and tests should be covered by [materialize_encoding_into_padding.mlir](https://github.com/iree-org/iree/blob/3b1730e5fa15a520505000b3f3232bd2a230793b/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_into_padding.mlir)

It is a step towards https://github.com/iree-org/iree/issues/20160

PR #22359, opened 2025-10-20
[DT] Mark partial slices unsupported in padding encoding resolver.

hanhanW:
The implementation was wrong; the tests are missing. It addresses a post review comment: https://github.com/iree-org/iree/pull/22339/files#r2438738373

### DIFF
--- a/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
+++ b/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
@@ -586,6 +586,14 @@ struct GPUPadEncodingLayoutMaterializerAttr final
     if (!boundType || !boundType.getEncoding()) {
       return failure();
     }
+    // Only handle cases where the slice spans the whole
+    // `!iree_tensor_ext.dispatch.tensor` type.
+    // TODO(hanchung): Enable partial slices. It was copied from pattern's
+    // implementaion, i.e., the users, and it can be dropped after we move the
+    // checks to the interface implementations.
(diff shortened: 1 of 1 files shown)


kuhar (approved):
Looks good but could we add a test?

hanhanW:
> Looks good but could we add a test?

We can only add negative test, but we don't test all corner cases. This PR should be treated as a post-fix of the other PR.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
```diff
@@ -586,6 +586,14 @@ struct GPUPadEncodingLayoutMaterializerAttr final
     if (!boundType || !boundType.getEncoding()) {
       return failure();
     }
+    // Only handle cases where the slice spans the whole
+    // `!iree_tensor_ext.dispatch.tensor` type.
+    // TODO(hanchung): Enable partial slices. It was copied from pattern's
+    // implementaion, i.e., the users, and it can be dropped after we move the
+    // checks to the interface implementations.
```
jtuyls, line 593: I am unsure about in which case this check can be dropped in the future? Do you want to support partial slices here?

hanhanW: Yes, I plan to do it in a follow-up. I think it is very similar to https://github.com/iree-org/iree/pull/22360. The main difference between the two resolvers is that the padding resolver gets padded DispatchTensorType while the identity resolver gets the original DispatchTensorType. The offsets/strides/sizes and RankedTensorType types used in compute ops are the same.

jtuyls (approved):
LGTM

PR #22360, opened 2025-10-20
[DT] Support partial load/store for identity encoding resolver.

hanhanW:
The revision adds the support of partial load/store lowering for identity encoding resolver; it removes the checks from `MaterializeTensorExtDispatchTensorLoadOp` and `MaterializeTensorExtDispatchTensorStoreOp` because they belong to
encoding resolver implementation details.

The data-tiling encoding resolvers all have the check:

https://github.com/iree-org/iree/blob/fcae3fcd1f5032a24ca00d913a6f026cb37edcf1/compiler/src/iree/compiler/Codegen/ExternalInterfaces/Utils.h#L136-L141

The check for padding resolver:

https://github.com/iree-org/iree/blob/4127b869cc72b230b56c331e53db7ca71de067b1/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp#L596-L603

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
@@ -229,13 +229,6 @@ struct MaterializeTensorExtDispatchTensorLoadOp
   matchAndRewrite(IREE::TensorExt::DispatchTensorLoadOp loadOp,
                   OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    // Only handle operations where the load covers the entire
-    // `!iree_tensor_ext.dispatch.tensor` type.
-    // TODO(ravishankarm): Relax this for partial loads.
-    if (!loadOp.isLoadOfWholeSource()) {
-      return rewriter.notifyMatchFailure(loadOp, "unhandled partial loads");
-    }
-
     auto sourceType = loadOp.getSourceType();
     auto boundTensorType = cast<RankedTensorType>(sourceType.getBoundType());
     auto *typeConverter = static_cast<const MaterializeEncodingTypeConverter *>(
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
```diff
@@ -547,13 +547,6 @@ struct MaterializeTensorExtDispatchTensorLoadOp
   matchAndRewrite(IREE::TensorExt::DispatchTensorLoadOp loadOp,
                   OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    // Only handle operations where the load covers the entire
-    // `!iree_tensor_ext.dispatch.tensor` type.
-    // TODO(ravishankarm): Relax this for partial loads.
-    if (!loadOp.isLoadOfWholeSource()) {
```
jtuyls, line 235: It doesn't matter too much if this PR and https://github.com/iree-org/iree/pull/22359 go in close together, but to be completely clean, we should merge https://github.com/iree-org/iree/pull/22359 before this PR to make sure that the user checks are there until we add back the implementation check.

jtuyls (approved):
LGTM

compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
hanhanW: Agree, and I plan to do that. Sorry that I forgot to mention it in a comment. :)

hanhanW: > TBD: point it to codebase instead of PR.

I think I can only point it to codebase after I land the padding PR. I'll be more explicit next time.

PR #22368, opened 2025-10-21
[LinalgExt][NFC] Split the op definition between pure ops and LinalgExt ops

### LINKED ISSUE #20862: [LinalgExt] Split the op definition between pure ops and LinalgExt ops
Similar to upstream LinAlg dialect, we should decouple the op definition between pure ops (e.g., linalg_ext.yield) and LinalgExtOp (e.g., scatter/gather/sort/etc). This way, we don't need to attach interface implementation when we create a new LinalgExt op. It can be done by including the generated `.inc.h` or `.inc.cpp` files. E.g.,

```cpp
void mlir::linalg::registerTilingInterfaceExternalModels(
    DialectRegistry &registry) {
  registry.addExtension(+[](MLIRContext *ctx, linalg::LinalgDialect *dialect) {
    registerOne<linalg::GenericOp>(ctx);
    linalg::PackOp::attachInterface<PackOpTiling>(*ctx);
    linalg::UnPackOp::attachInterface<UnPackOpTiling>(*ctx);
    registerAll<
#include "mlir/Dialect/Linalg/IR/LinalgStructuredOps.cpp.inc"
        >(ctx);
  });
}
```

Currently, it can benefit at least four places in the codebase:

https://github.com/iree-org/iree/blob/7c55b5245a00212e54bdb653a77930a26774795c/compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp#L479-L506

https://github.com/iree-org/iree/blob/7c55b5245a00212e54bdb653a77930a26774795c/compiler/src/iree/compiler/Codegen/Interfaces/PartitionableLoopsInterface.cpp#L241-L270

https://github.com/iree-org/iree/blob/7c55b5245a00212e54bdb653a77930a26774795c/compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp#L689-L715

https://github.com/iree-org/iree/blob/7c55b5245a00212e54bdb653a77930a26774795c/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp#L2216-L2238

sakupan102:
Split LinalgExtOp into PureOp and LinalgExtOp. This enables automatically attaching the right interfaces by simply including the generated .cpp.inc files, preventing missed interface attachments.

Closes https://github.com/iree-org/iree/issues/20862

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
+++ b/compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
@@ -331,6 +331,15 @@ struct LinalgExtOpInterface
   }
 };
 
+template <typename... Ops>
+struct LinalgExtOpInterfaceHelper {
+  static void registerOpInterface(MLIRContext *context) {
+    (void)std::initializer_list<int>{
+        0, (Ops::template attachInterface<LinalgExtOpInterface<Ops>>(*context),
+            0)...};
+  }
+};
+
 /// Returns the buffers of the source and destination for pack and unpack ops.
 /// Returns a failure if the buffers can not be found.
 template <typename OpTy>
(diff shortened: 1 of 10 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -21,12 +21,8 @@ include "mlir/Interfaces/ViewLikeInterface.td"
 // Base class.
 //===----------------------------------------------------------------------===//
 
-class IREELinalgExt_PureOp<string mnemonic, list<Trait> traits = []> :
-    Op<IREELinalgExt_Dialect, mnemonic, traits> {
-}
-
 class IREELinalgExt_Op<string mnemonic, list<Trait> traits = []> :
-    IREELinalgExt_PureOp<mnemonic, !listconcat(traits,
+    Op<IREELinalgExt_Dialect, mnemonic, !listconcat(traits,
```
hanhanW: Can't this file include `LinalgExtPureOps.td`?

hanhanW (commented):
Thanks for the patch!

I think MapScatterOp and attention ops should belong to `LinalgExtOps.td`. (I don't know why they dont inherit LinalgExtOp for now, but it is a separate issue.)

My mental model is that we structure it like linalg ops. There are pure ops and structure ops (that I call compute ops in IREE). The MapScatterOp and attention ops describe the computation, so they belong to `LinalgExtOps.td`.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
sakupan102: If this file directly includes `LinalgExtPureOps.td`, it redefines pureOps, which causes an error.
It is better to move the base class to `LinalgExtBase.td` and include it from there.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -318,6 +270,8 @@ def IREELinalgExt_GatherOp : IREELinalgExt_Op<"gather",
   let hasCanonicalizer = 1;
 }
 
+// map_scatter op does not have enough operands to use AttrSizedOperandSegments trait
+// so we inherit from IREELinalgExt_PureOp for now.
```
hanhanW: The comment is not accurate to me. The reason of using AttrSizedOperandSegments for an op if it has at least two variadic operands. I think you can drop the comments for this case and other cases. I'll do a cleanup for this in a follow-up. Thanks.

https://mlir.llvm.org/docs/DefiningDialects/Operations/#variadic-operands

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -686,37 +695,17 @@ void registerBufferizationInterfaces(DialectRegistry &registry) {
...
+  registry.addExtension(
+      +[](MLIRContext *ctx, IREE::LinalgExt::IREELinalgExtDialect *dialect) {
+        LinalgExtOpInterfaceHelper<
+#define GET_OP_LIST
+#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp.inc"
+            >::registerOpInterface(ctx);
+        IREE::LinalgExt::AttentionOp::attachInterface<
+            LinalgExtOpInterface<IREE::LinalgExt::AttentionOp>>(*ctx);
+        IREE::LinalgExt::MapScatterOp::attachInterface<
+            LinalgExtOpInterface<IREE::LinalgExt::MapScatterOp>>(*ctx);
```
hanhanW: We no longer need them, right? Because they should be registered with the above include.

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -446,31 +446,12 @@ void registerUtilExternalModels(DialectRegistry &registry) {
...
-        LinalgOpTiedOpInterface<IREE::LinalgExt::WinogradOutputTransformOp>>(
-        *context);
-    IREE::LinalgExt::Im2colOp::attachInterface<
-        LinalgOpTiedOpInterface<IREE::LinalgExt::Im2colOp>>(*context);
+    LinalgOpTiedOpInterfaceHelper<
+#define GET_OP_LIST
+#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp.inc"
+        >::registerOpInterface(context);
     IREE::LinalgExt::AttentionOp::attachInterface<
         LinalgOpTiedOpInterface<IREE::LinalgExt::AttentionOp>>(*context);
```
hanhanW: This is no longer needed.

hanhanW (commented):
overall looks good, I think we miss few removal; I think we can drop the comments for three ops.

hanhanW (approved):

MaheshRavishankar (commented):
Hanhan approved it, but missing context as to why we need this? Its a bit of an artificial split. I have no strong opinions, but curious why we did this to start with (especially since we only moved a single op into a different file).

hanhanW:
> Hanhan approved it, but missing context as to why we need this? Its a bit of an artificial split. I have no strong opinions, but curious why we did this to start with (especially since we only moved a single op into a different file).

The main reason is described here: https://github.com/iree-org/iree/issues/20862#issue-3075122063

It will allow us to have less burden when adding a new LinalgExt op. The interface registration (e.g., memory side effect, bufferization, etc) will be done automatically, so we no longer need commit like https://github.com/iree-org/iree/commit/dce563a9a9253168e8829e56852864e88ebc8f0f

hanhanW:
@sakupan102 can you fix the bazel and update PR description? https://google.github.io/eng-practices/review/developer/cl-descriptions.html has the guidance about what a good PR title and description can be.

Groverkss:
> I think MapScatterOp and attention ops should belong to `LinalgExtOps.td`. (I don't know why they dont inherit LinalgExtOp for now, but it is a separate issue.)

IIRC it's because most of linalg ext ops are trying to act like linalg.generic syntax with ins()/outs() which the interface requires. Attention/MapScatter have a fixed number of arguements and don't need that.

Most ops in linalg_ext don't need variadic arguements and but we never changed it.

Groverkss:
LGTM, we need to revamp LinalgExtInterface to not have variadic arguements in future though, most ops don't need it in linalg_ext

hanhanW:
> LGTM, we need to revamp LinalgExtInterface to not have variadic arguements in future though, most ops don't need it in linalg_ext

This is what I'm going to do in a follow-up. Ops should add the interface when they need to; it should not belong to the LinalgExtOps traits today. (It's mostly legacy.)

MaheshRavishankar:
> > LGTM, we need to revamp LinalgExtInterface to not have variadic arguements in future though, most ops don't need it in linalg_ext
> 
> This is what I'm going to do in a follow-up. Ops should add the interface when they need to; it should not belong to the LinalgExtOps traits today. (It's mostly legacy.)

Yeah my bad! It wrote that at a time when I didn't know better

PR #22284, opened 2025-10-12
Remove `moveCrossThreadOutermost`

bjacob:
Fixes #22144.

E2E matmul tests passed as-is, confirming that this wasn't needed anymore. The diff is adjusting lit-tests.

PR #22369, opened 2025-10-21
Give `inner_tiled` a strict verifier and explicit semantics with boolean parameters `distributed` and `opaque`

### LINKED ISSUE #22336: `inner_tiled` op needs explicit semantics and a strict verifier
The `inner_tiled` op rides successive lowerings. Some of these are semantically straightforward, e.g. bufferization. But some other lowerings cause nontrivial changes to its semantics and the shapes of its operands:
1. `ExpandUndistributedInnerTilesPass` expands the 2D tiles into higher-dimensional tiles with expand_shape and transpose ("swizzle").
2. `DistributeInnerTiledToLanesPass` changes from workgroup to thread semantics, slicing the tiles.

This means we have 3 different "semantics" for this op:
1. Not yet expanded or distributed.
2. Expanded but not yet distributed.
3. Distributed to threads.

This is currently implied by the shapes of operands.

The verifier, accordingly, is very lax: it only requires the tiles element counts to match.

As a result, the verifier has false negatives and confusing diagnostics.

The false negatives are concerning because the ill-formed IR that is allowed past the verifier is known to cause assertion failures (in assert-enabled build) and out-of-bounds accesses (when asserts are disabled) in the compiler, see https://github.com/iree-org/iree/pull/22284#discussion_r2429588669 . That is why this Issue is tagged 'Bug' not just 'Enhancement'.

Instead, `inner_tiled` should have an additional attribute parameter selecting its semantics among the above 3. The verifier should then strictly enforce the specified semantics.

bjacob:
Fixes #22336.

Discussed on Discord around https://discord.com/channels/689900678990135345/1254843174111678555/1430237660781215794


MaheshRavishankar (commented):
Ill let other people provide more detailed feedback, but IMO this implies that this op is not designed properly, and is being used in contexts where it shouldnt be. Adding extra attributes on top seems like a band-aid to me. 

Btw, the new attributes added should at least be spelled out in the op description.

qedawkins:
Instead of adding these attributes on the operation itself, we could instead add it to the `kind` descriptor. When I first introduced the precursor to this op I left verification somewhat sparse because an interface method for verification smelled wrong. Perhaps the real mistake though was `ExpandUndistributedInnerTiledDims` and `DistributeInnerTiledToX` passes not updating the `kind`. The semantics of the op itself are really quite simple and would be good to keep as such:
 - Iterate over the outer dimensions of the input operands
 - Extract an inner tile from each operand based on the data type
 - Invoke the functionality described by the `kind` attribute on said inner tile.
 
 The reason for an attribute descriptor instead of a body region is that the kinds of transformations we needed to do (expansion/distribution) are substantially harder when dealing with regions. The price of that is verbose attributes.

The reason for a single op over an interface is that patterns that transform the outer dims would necessarily have shared implementations meaning the interface would likely bloat quickly (not that the `InnerTiledDescAttr` isn't somewhat bloated, but that's kind of unavoidable).

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
```diff
@@ -194,23 +197,53 @@ InnerTiledOp::inferReturnTypes(MLIRContext *, std::optional<Location>,
...
+      return tiledOp->emitOpError(llvm::formatv(
+          "operand type {}, implying tile type {}, is incompatible with the "
+          "target MMA tile type {}, given that the op has distributed={} and "
+          "opaque={}",
+          opType, VectorType::get(operandTileShape, operandElemType),
+          mmaVectorType, tiledOp.getDistributed(), tiledOp.getOpaque()));
+    }
   }
-  return result;
+  return llvm::success();
```
kuhar:
```suggestion
  return success();
```

bjacob:
@MaheshRavishankar / @qedawkins , some high-level replies:

> this implies that this op is not designed properly

I think it was designed initially just right for the problem space that we were entering then, given that there was no way that we could see the destination so this would have to evolve anyway.  Evolve it did -- and along the way it accrued technical debt in that some of the evolutions came at a cost in terms of clarity and definedness of semantics. You could say that "supporting distribution" was part of the initial scope so that shouldn't have compromised semantics, but that "initial scope" was a really big chunk in itself, and maybe playing fast and loose was the reasonable thing to do, particularly as in exchange for the lax semantics we got more flexibility for the subsequent evolutions. One thing that in hindsight is impressive is that even though the initial non-DT flavors relied on opaque shapes, when I did DT I was able to do strict expanded shapes *without changing the framework* - so the new/better/stricter form was able to evolve within the existing lax form, thanks to the laxness.

Key point: This PR is not having strong opinions about anything.  It's just the smallest incremental change to this op that recovers clarity and definedness of semantics.

The concern about the lax verifier and loose semantics is not just an abstract concern. The lax verifier meant that we had false positives allowing ill-formed IR to pass and later cause out-of-bounds accesses in the compiler (https://github.com/iree-org/iree/pull/22284#discussion_r2429588669) as mentioned in #22336. The verifier code was also convoluted. These were not incidental details but were forced by the loose semantics.

> Btw, the new attributes added should at least be spelled out in the op description.

Yup, omission, doing that.

> Instead of adding these attributes on the operation itself, we could instead add it to the kind descriptor.

We _could_ do that and I'm not opposed to it, but:
1. It's not clear to me that this is better aligned with the semantics. "distributed" in particular is more about the tensor operands than it is about the mmaKind. "opaque" is equal parts about both".
2. It will result in a bit more C++ code, as rewrites of InnerTiledOp's can't anymore just forward `getKind()`.
3. It will be a bit more work here, and I was hoping to merge quickly as this PR touches many tests.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
bjacob: SG, will fold that into the next push.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -322,29 +322,43 @@ def IREECodegen_InnerTiledOp : Op<IREECodegen_Dialect, "inner_tiled", [
...
       : vector<4xf16>, vector<4xf16> into vector<4xf32>
     ```
 
     This operation can represent an intrinsic both in undistributed
-    (workgroup/subgroup/warp) and distributed (thread) level. The descriptor
-    attribute specifies the inner tile sizes for both the undistributed form
-    (where the operands represent the data to be processed by an entire group of
-    parallel workers) and the distributed form (where the operands represent the
-    data processed by a single workitem/thread).
+    (workgroup/subgroup/warp) and distributed (thread) level. This is indicated
```
qedawkins, line 335: Mentioning `subgroup/warp/thread` here indicates this is in the wrong dialect. The codegen dialect shouldn't include GPU concepts.

bjacob: I didn't type those words, just rearranged them. Preexisting issues, big discussions - can we keep that for a separate Issue (want me to file one?)

bjacob: Or are you saying that's proof that `distributed` belongs in the Kind attr not in the op itself so you really want this change in this PR?

bjacob: But above I didn't exhaust all the reasons why putting `distributed` in Kind is not trivial - the existing design of MMAAttr-like attributes, with side-by-side functions to query distributed/undistributed tiles, is built around an assumption that becomes obsolete if `distributed` moves in.

bjacob: I can see two ways that this issue (polluting codegen dialect with GPU specifics) could be resolved:
1. Either we move `distributed` and `opaque` to the Kind attr as you suggested above, but that sets off a chain reaction of design adjustments that I would really like to keep as followups.
2. Or we go back to the linked Discord discussion where we debated whether we wanted that boolean `distributed`, or rather an enum `scope`. The good thing with the scope enum is that there could be a common `thread` scope that by definition applies everywhere (GPU or CPU) and then we could have some GPU-specific scope enum value like `subgroup` but it would not be too bad to have a GPU-specific enum value within a non-GPU-specific enum type.  Similar to how portable scoped atomics have scope values that can be specific to some targets.

I think that 2. is small enough to fit in this PR if you want, but 1. is either too big (if done right) or not worth it.

qedawkins:
> We _could_ do that and I'm not opposed to it, but:
> 
>     1. It's not clear to me that this is better aligned with the semantics. "distributed" in particular is more about the tensor operands than it is about the mmaKind. "opaque" is equal parts about both".
> 
>     2. It will result in a bit more C++ code, as rewrites of InnerTiledOp's can't anymore just forward `getKind()`.
> 
>     3. It will be a bit more work here, and I was hoping to merge quickly as this PR touches many tests.

At this point I'm not any more an owner of this op than you/others given how much it has evolved, and if this is the direction we have to go right now I don't want to cause needless churn and won't block, but I do want to share my concerns.

> "distributed" in particular is more about the tensor operands than it is about the mmaKind

I don't agree. From the perspective of the `inner_tiled` op, before and after distribution the iteration over outer dimensions remains the same. What changes is how each tile is processed (before distribution, each lane must load the subvector they are responsible for). So in my mind the thing that had loosely defined semantics wasn't the `inner_tiled` op, but the `kind` describing its body.

2) and 3) are both true, but what I am concerned about is an ever growing set of "semantic modifiers" for the `kind` that we add to the `inner_tiled` op and are hard to interpret in a vacuum.

bjacob:
> I don't agree. From the perspective of the inner_tiled op, before and after distribution the iteration over outer dimensions remains the same. What changes is how each tile is processed (before distribution, each lane must load the subvector they are responsible for). So in my mind the thing that had loosely defined semantics wasn't the inner_tiled op, but the kind describing its body.

The other thing that changes with distribution is the *shape* of the tensor operands, in the inner dimensions.

I can see the mental model you are laying out here, where distributedness is a property of the Kind attributes.  But if we look at the current MMAAttr-like attributes and even the underlying interfaces like `InnerTileDescAttrInterface`, they don't know about distributedness and their interface methods reflect that, having side-by-side methods for distributed and undistributed cases, which is exactly what they wouldn't have if these interfaces were meant to cover distributedness (see below).

So I think I agree with you on the destination, but I think that getting there in a way that is really worth doing (doesn't just shift a semantics gap one step down) will take more changes than should be fitted into this PR. I think that this PR as-is is a necessary first step towards that, as by tightening the existing semantics it takes care of the hard part (climbing out of a hole of loose semantics) and leaves only a more reasonable follow-up to do (evolve design within clear constraints covered by verifiers).

Do you want me to file the Issue?

```
  DeclareAttrInterfaceMethods<IREECodegen_InnerTileDescAttrInterface, [
    "getExpectedNumInputs",
    "getExpectedNumOutputs",
    "verifyIndexingMaps",
    "getUndistributedTileTypes",
    "getDistributedTileTypes",
    "getUndistributedTileDimExpansion",
    "populateOperandOffsetsSizesStrides",
    "getDistributionMappingKind",
    "getDistributionWorkerCount",
    "buildUnderlyingOperations",
  ]>
```

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
qedawkins: 2) sounds ok to me if that's the easiest way forward. I also don't mind discussing in a call.

qedawkins:
> But if we look at the current MMAAttr-like attributes and even the underlying interfaces like InnerTileDescAttrInterface, they don't know about distributedness and their interface methods reflect that, having side-by-side methods for distributed and undistributed cases, which is exactly what they wouldn't have if these interfaces were meant to cover distributedness (see below).

Yes, I think you hit the nail on the head here and I should have been clearer when stating my concerns. The `InnerTileDescAttrInterface` is as much to blame for the current need for additional semantic modifiers as the inner_tiled op itself.

> I think that this PR as-is is a necessary first step towards that, as by tightening the existing semantics it takes care of the hard part (climbing out of a hole of loose semantics) and leaves only a more reasonable follow-up to do (evolve design within clear constraints covered by verifiers).

Part of why I said I wouldn't block was because the changes I'm asking for amounts to a substantial rewrite, and if we are on the same page then this SGTM.

MaheshRavishankar:
Just looking on from fairly far off, this operation has been rewritten once already. Going by the rule of three (you need three tries to get something right), its OK to have substantial changes and staging them, but what's missing is someone effectively owning the state and making sure that we are heading towards what is (eventually) stable.

bjacob:
@qedawkins , @krzysz00 , what I just pushed here should have all what we discussed.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
```diff
@@ -9,6 +9,55 @@
...
+  let cppNamespace = "::mlir::iree_compiler::IREE::Codegen";
+  let description = [{
+    Attribute interface used to describe aspects of tile shape semantics
+    that are orthogonal to InnerTileDescAttrInterface and that may change during
+    lowerings, whereas the InnerTileDescAttrInterface describes aspects that are
+    invariant during lowerings. This includes:
+    1. Expanded vs opaque tile shapes.
+    2. On GPU: distributed vs undistributed tile shapes.
+    Having this split into a separate interface allows mutating those
+    attribtutes separately.
```
qedawkins, line 23:
```suggestion
    attributes separately.
```

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
```diff
@@ -9,6 +9,55 @@
...
+           /*retTy=*/"bool",
+           /*methodName=*/"getOpaque",
+           /*args=*/(ins)>];
+}
+
+def IREECodegen_AnyInnerTiledSemanticscAttr
+    : Attr<Or<[CPred<"::llvm::isa<::mlir::iree_compiler::IREE::Codegen::"
+                     "InnerTiledSemanticsAttrInterface>($_self)">,
+]>,
+           "buffer-like constant attribute values"> {
```
qedawkins: Short description here looks copy-paste (which itself might have been copy-paste)

Groverkss (commented):
Looks fine to me, I don't have much opinion on this

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
```diff
@@ -194,23 +196,70 @@ InnerTiledOp::inferReturnTypes(MLIRContext *, std::optional<Location>,
...
+          llvm::formatv("operand element type {} does not match expected MMA "
+                        "tile element type {}",
+                        operandElemType, mmaElemType));
+    }
+    ArrayRef<int64_t> operandShape = cast<ShapedType>(opType).getShape();
+    ArrayRef<int64_t> operandTileShape(
+        operandShape.drop_front(map.getNumResults()));
+    ArrayRef<int64_t> mmaTileShape = mmaVectorType.getShape();
+    SmallVector<int64_t> permutedMmaTileShape(mmaTileShape);
+    auto permutations = tiledOp.getPermutations();
```
qedawkins: nit: spell out type

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
```diff
@@ -288,50 +337,28 @@ LogicalResult InnerTiledOp::verify() {
...
-      return emitOpError("permutations require undistributed semantics");
-    }
-  }
-
   if (getPermutations()) {
-    for (auto permAttr : getPermutations()->getAsRange<DenseI64ArrayAttr>()) {
-      if (!isPermutationVector(permAttr.asArrayRef()))
-        return emitOpError("invalid permutation");
+    for (auto [index, permAttr] :
+         llvm::enumerate(getPermutations()->getAsRange<DenseI64ArrayAttr>())) {
```
qedawkins: I believe this is missing verification that the number of permutations equals the operand count (needed to keep later indexing in bounds).

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp
```diff
@@ -1394,10 +1407,15 @@ distributeInnerTiledOp(RewriterBase &rewriter,
...
 
+  auto distributedSemantics = IREE::GPU::InnerTiledSemanticsAttr::get(
+      context, /*distributed=*/true, semantics.getOpaque());
+
   // Step 3. Create the new inner_tiled op.
   auto newTiledOp = IREE::Codegen::InnerTiledOp::create(
       rewriter, loc, inputSlices, initSlices, tiledOp.getIndexingMaps(),
-      tiledOp.getIteratorTypes(), tiledOp.getKind());
+      tiledOp.getIteratorTypes(), tiledOp.getKind(),
+      cast<IREE::Codegen::InnerTiledSemanticsAttrInterface>(
```
qedawkins: Is this cast required? It looks like `InnerTiledSemanticsAttr` implements the interface directly.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
```diff
@@ -163,6 +163,33 @@ def IREEGPU_DotProductOpsAttr : EnumAttr<
...
+]> {
+  let mnemonic = "mma_semantics";
+  let cppNamespace = "::mlir::iree_compiler::IREE::GPU";
+
+  let description = [{
+    Attribute describing aspects of MMA semantics that are orthogonal to the
+    aspects described in the mma_layout attrs, and that may change during
+    lowerings while the mma_layout's typically stay constant. This includes:
+    1. Expanded vs opaque tile shapes.
+    2. Distributed vs undistributed tile shapes.
```
qedawkins, line 185: nit: clarify that distributed request non opaque semantics.

qedawkins (approved):

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
bjacob: Good catch: added several more test cases here.

PR #22372, opened 2025-10-21
[DT] Support partial load/store for GPU padding encoding resolver.

hanhanW:
I don't see an actual use case, and the revision is mostly for completeness. By definition, the padding resolver behaves like an identity resolver. The main difference is that the DispatchTensorType types get padded. The offsets, sizes and strides do not get transformed in this case.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
+++ b/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
@@ -593,17 +593,9 @@ struct GPUPadEncodingLayoutMaterializerAttr final
     if (!boundType || !boundType.getEncoding()) {
       return failure();
     }
-    // Only handle cases where the slice spans the whole
-    // `!iree_tensor_ext.dispatch.tensor` type.
-    // TODO(hanchung): Enable partial slices. It was copied from pattern's
-    // implementaion, i.e., the users, and it can be dropped after we move the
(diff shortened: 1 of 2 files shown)


hanhanW:
@kuhar this is the test case that I was not able to add in https://github.com/iree-org/iree/pull/22359#pullrequestreview-3358357037, FYI

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
```diff
@@ -593,17 +593,9 @@ struct GPUPadEncodingLayoutMaterializerAttr final
...
-    // TODO(hanchung): Enable partial slices. It was copied from pattern's
-    // implementaion, i.e., the users, and it can be dropped after we move the
-    // checks to the interface implementations.
-    if (!type.doesSliceSpanWholeTensor(dynamicDims, offsets, sizes, strides)) {
-      return failure();
-    }
-    newSizes = getMixedValues(boundType.getShape(), dynamicDims, builder);
-    newOffsets.resize(newSizes.size(), builder.getIndexAttr(0));
-    newStrides.resize(newSizes.size(), builder.getIndexAttr(1));
+    newSizes.assign(sizes.begin(), sizes.end());
```
kuhar, line 596: We should add an `.assign` overload that accepts `ArrayRef`s...
(obviously non-blocking for this PR)

kuhar (approved):

PR #22354, opened 2025-10-20
Bump LLVM to llvm/llvm-project@c8cf393

Muzammiluddin-Syed-ECE:
New Revert:
- https://github.com/llvm/llvm-project/commit/8c05b5cca8784814cf11ac9d85c4ab59d2952ca5 causes numerical errors see (https://github.com/iree-org/iree/pull/22354#issuecomment-3424225175)

Still carrying these reverts:
- https://github.com/llvm/llvm-project/pull/160615 (See https://github.com/iree-org/iree/issues/22171)

Fixes:
- API change to linalg::rewriteAsPaddedOp (https://github.com/llvm/llvm-project/commit/1508a8ec8d62ab1e9bdc8b7e0dbaaec9075b631f)
- API change to populateMathToROCDLConversionPatterns (https://github.com/llvm/llvm-project/commit/fbbffc11690c0f47bdc74939b290696df4c1c061)
- Changes to lit tests

Issue #22171, opened 2025-10-01
Remove revert of https://github.com/llvm/llvm-project/pull/160615 on LLVM integrate


amd-eochoalo:
* **What needs to be done?** Drop revert for [fcf79e5](https://github.com/llvm/llvm-project/commit/fcf79e5276299c351b924f7a08f6c3c6a6c6a1ee)
* **Why was the revert added?** [fcf79e5](https://github.com/llvm/llvm-project/commit/fcf79e5276299c351b924f7a08f6c3c6a6c6a1ee) makes sure that folding takes place until a fixed point is reached. However, folding also enables a loop unrolling optimization which may be semantically incorrect in the context of GPUs. The test in question is [tile_and_distribute_scatter.mlir](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/SPIRV/test/tile_and_distribute_scatter.mlir)

Click [here](https://gist.github.com/amd-eochoalo/db848c7f1ee717374a6b08125a3837ca) to see the complete output before and after the commit has been added.

Looking into the diff of the output

```diff
15d14
<           %c1 = arith.constant 1 : index
35,45c34,40
<               %thread_id_y = gpu.thread_id  y
<               %block_dim_y = gpu.block_dim  y
<               scf.for %arg2 = %thread_id_y to %c1 step %block_dim_y {
<                 scf.for %arg3 = %thread_id_x to %5 step %block_dim_x {
<                   %subview_2 = memref.subview %subview[%arg2, %arg3] [1, 1] [1, 1] : memref<1x?xi32, #map2> to memref<1x1xi32, strided<[500, 1], offset: ?>>
<                   %subview_3 = memref.subview %subview_0[%arg2, 0] [1, 1] [1, 1] : memref<1x1xi32, #map3> to memref<1x1xi32, strided<[1, 1], offset: ?>>
<                   %subview_4 = memref.subview %subview_1[0, %arg3] [100, 1] [1, 1] : memref<100x?xi32, #map2> to memref<100x1xi32, strided<[500, 1], offset: ?>>
[...]

PR #22374, opened 2025-10-22
Bump LLVM to llvm/llvm-project@9abbec6

Muzammiluddin-Syed-ECE:
Still carrying these reverts:
- https://github.com/llvm/llvm-project/pull/160615 (See https://github.com/iree-org/iree/issues/22171)
- https://github.com/llvm/llvm-project/pull/163440 (See https://github.com/iree-org/iree/pull/22354#issuecomment-3424225175)

Fixes: 
- Remove UnsafeFPmath option from llvm target options (https://github.com/llvm/llvm-project/commit/76f15ead0100e1a846f3e8d055d91ac669e297be)

ci-extra: test_torch

### DIFF
--- a/compiler/plugins/target/ROCM/ROCMTarget.cpp
+++ b/compiler/plugins/target/ROCM/ROCMTarget.cpp
@@ -638,7 +638,6 @@ public:
         }
         llvm::TargetOptions opt;
         opt.AllowFPOpFusion = llvm::FPOpFusion::Fast;
-        opt.UnsafeFPMath = false;
         opt.NoInfsFPMath = false;
         opt.NoNaNsFPMath = true;
         // Be extra cautious while this is less tested, and prevent unknown


MaheshRavishankar (commented):
oh, why did we drop the unsafefpMath = false flag?

Muzammiluddin-Syed-ECE:
> oh, why did we drop the unsafefpMath = false flag?

It seems its part of a larger discussion on dealing with fast-math pragma's, see related [RFC](https://discourse.llvm.org/t/rfc-honor-pragmas-with-ffp-contract-fast/80797/10)

Muzammiluddin-Syed-ECE:
> > oh, why did we drop the unsafefpMath = false flag?
> 
> It seems its part of a larger discussion on dealing with fast-math pragma's, see related [RFC](https://discourse.llvm.org/t/rfc-honor-pragmas-with-ffp-contract-fast/80797/10)

Iiuc its basically that the 'ffast-math' option enables the same behavior as 'enable-unsafe-math' making this option redundant

Groverkss (approved):
Added torch models ci, let's see if there are any regressions, but should be fine since you mention this flag is just redundant now.

PR #22378, opened 2025-10-22
[CI][DO NOT REVIEW] Add performance tracking graphs

Yu-Zhewen:
ci-extra: test_torch


Yu-Zhewen:
Can't test from a fork branch as secret token is required. Move to #22382 instead.

PR #22379, opened 2025-10-22
[CI][Torch] Add llama 8b fp16 quality tests

Groverkss:
Adds a llama 8b fp16 quality test. The quality test actually runs a greedy scheduler that takes in text tokens and outputs text tokens. These tokens can be checked using a tokenizer. We check for exact equality in ci, but if something changes due to reassociations, we can check what the output tokens mean after decoding them using the tokenizer script (in comments in the scheduler ir) and update them.

Note that this also updates the fp16 ir because the old IR was wrong for accuracy. It was probably exported with a wrong causal masking scheme. It changes the number of dispatches, so something to look for there.,

Scheduler IR: https://sharkpublic.blob.core.windows.net/sharkpublic/iree-test-suites/torch-models/llama_8b_fp16/greedy_scheduler.mlir

skip-ci: Manually run CI


MaheshRavishankar (changes requested):
Why opt-level=O0?

Groverkss:
> Why opt-level=O0?

I switched it to O1, O2/O3 fail for argmax compilation with shared memory problems

Groverkss:
This should fail TOM, but I have verified it works before the transfer_write guard commit.

tests/external/iree-test-suites/torch_models/llama_8b_fp16/modules/scheduler_gfx942.json
```diff
@@ -0,0 +1,11 @@
+{
+  "mlir": "https://sharkpublic.blob.core.windows.net/sharkpublic/iree-test-suites/torch-models/llama_8b_fp16/greedy_scheduler.mlir",
+  "compiler_flags": [
+      "--iree-hal-target-device=hip",
+      "--iree-hip-target=gfx942",
```
kuhar, line 5: This scheduler is target-specific -- should we also give it a _gfx942 suffix?

MaheshRavishankar (approved):
We should be using O3 here. Maybe an issue to file and follow up on later on.

@bangtianliu  I know you are busy with some other things, but something you can look into?

bangtianliu:
> @bangtianliu I know you are busy with some other things, but something you can look into?

Sure, I will look at the argmax compilation issue.

Groverkss:
CI run: https://github.com/iree-org/iree/actions/runs/18772199972

tests/external/iree-test-suites/torch_models/llama_8b_fp16/modules/scheduler_gfx942.json
Groverkss: done

Groverkss:
Landing, accuracy CI should pass after the revert on TOM

Issue #16168, opened 2024-01-19
Exported I/O buffer views improperly alias with tied ops.


benvanik:
This is likely a missing case in the MaterializeCopyOnWrite pass. An input program that has a single tied operation (in my case from a linalg.matmul) will not have a clone on the input inserted and will mutate the original input in-place when that's not what the user indicated. We should only do this when iree.abi.output is assigned (the hal.tensor.export has a tied buffer attached).

The current workaround is to pass in copies of the arguments but that's pretty terrible.

```mlir
#config = #iree_codegen.lowering_config<tile_sizes = [[32, 128], [4, 4], [0, 0, 32]]>
#executable_target_vulkan_spirv_fb = #hal.executable.target<"vulkan", "vulkan-spirv-fb", {spirv.target_env = #spirv.target_env<#spirv.vce<v1.3, [Shader, GroupNonUniform], [SPV_KHR_storage_buffer_storage_class, SPV_KHR_variable_pointers]>, api=Vulkan, #spirv.resource_limits<max_compute_workgroup_size = [128, 128, 64], subgroup_size = 64, cooperative_matrix_properties_khr = []>>}>
#translation = #iree_codegen.translation_info<SPIRVBaseVectorize pipeline_depth = 3>
#compilation = #iree_codegen.compilation_info<lowering_config = #config, translation_info = #translation, workgroup_size = [32, 8, 1]>
#device_target_vulkan = #hal.device.target<"vulkan", {executable_targets = [#executable_target_vulkan_spirv_fb], legacy_sync}>
module attributes {hal.device.targets = [#device_target_vulkan]} {
  func.func @matmul_accumulate_256x128xi8_times_128x512xi8_into_256x512xi32_for_SPIRVVectorizeNVIDIA_32_128_4_4_0_0_32_32_8_1(%arg0: tensor<256x128xi8>, %arg1: tensor<128x512xi8>, %arg2: tensor<256x512xi32>) -> tensor<256x512xi32> {
    %0 = linalg.matmul {compilation_info = #compilation} ins(%arg0, %arg1 : tensor<256x128xi8>, tensor<128x512xi8>) outs(%arg2 : tensor<256x512xi32>) -> tensor<256x512xi32>
    return %0 : tensor<256x512xi32>
  }
}
```

```mlir
// -----// IR Dump After FormDispatchWorkgroups (iree-flow-form-dispatch-workgroups) //----- //
[...]

PR #22380, opened 2025-10-22
Adding markAllAnalysesPreserved to verification passes.

benvanik:
Verification passes are read-only and don't modify the IR so they should preserve all analyses to avoid unnecessary recomputation in the pass pipeline. We don't use too many analyses today we expect to survive across them but will start to in the future and it's good hygiene.

Part of #16168 PR sequence (1/6).


MaheshRavishankar (approved):

hanhanW (approved):
I did not know `markAllAnalysesPreserved`, and it looks good.

PR #22383, opened 2025-10-22
Adding some Stream canonicalizations and RefineUsagePass improvements.

benvanik:
This finally allows us to turn transfers into clones when possible (same source/target) and remove some special casing around it. The bad behavior for things like returning constants from public functions and other internal<->external interop would cause mapping to resource lifetime to go pear-shaped and either produce incorrect programs (well, "underspecified" programs) or additional copies. We may still be underspecified but now are at least fast and consistent. Since we don't have coverage elsewhere apparently (outside of some sharktank stuff) I've added tests for the common cases of constant I/O.

There's a growing need to change up `!stream.resource<...>` a bit to better model the high-fidelity usage information we capture so we can better map it to HAL buffers but that is left for future cleanup for now.

Part of #16168 PR sequence (3/6).


benvanik:
(this was too hard to separate from the subsequent scf improvements due to some tests we have that actually use control flow - I'm going to merge this in with 16168-4 and continue from there)

PR #22387, opened 2025-10-23
[CI][TorchModels] Add data-tiling for Llama 8B Fp8 on gfx942

Abhishek-Varma:
-- This commit adds data-tiling benchmark tests for Llama 8B on gfx942.

Summary:

No. of dispatches for NON-DATA-TILED: 1485 
No. of dispatches for DATA-TILED: 1711

`DECODE`
```
       NON-DATA TILED   |      DATA-TILED
128  |    8.26          |        53
2048 |    11.14         |        55
```
`PREFILL`
```
       NON-DATA TILED   |      DATA-TILED
128  |    27            |        28.5
2048 |    169.6         |        201.5
```

ci-extra: test_torch

### DIFF
--- a/tests/external/iree-test-suites/torch_models/llama_8b_fp16/prefill_benchmark_seq2048_mi325.json
+++ b/tests/external/iree-test-suites/torch_models/llama_8b_fp16/prefill_benchmark_seq2048_mi325.json
@@ -36,5 +36,5 @@
         "value": "34x2097152xf16"
       }
     ],
-    "golden_time_ms": 250.5
+    "golden_time_ms": 275.55
 }
--- a/tests/external/iree-test-suites/torch_models/llama_8b_fp8/compstat_gfx942_data_tiling.json
+++ b/tests/external/iree-test-suites/torch_models/llama_8b_fp8/compstat_gfx942_data_tiling.json
@@ -1,0 +1,10 @@
+{
+  "type": "compstat",
+  "markers": [
+    "hip",
+    "gfx942"
+  ],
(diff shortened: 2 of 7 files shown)


Groverkss (commented):
Can you also add a compstat test?

MaheshRavishankar (approved):
Looks good. 

@hanhanW  is the extra number of dispatches with data tiling expected?

hanhanW:
> Looks good.
> 
> @hanhanW is the extra number of dispatches with data tiling expected?

It is set to 1711 in the PR, but I'd expect it to be as the same as non-data-tiling which is [1485](https://github.com/iree-org/iree/blob/8867197f8f751edc82eaba951c84d078b2f77bc9/tests/external/iree-test-suites/torch_models/llama_8b_fp8/compstat_gfx942.json#L8).

This is a new model for data-tiling, so some encoding fusion may be different. We can start with what it is and burn down the dispatch numbers. I'll file an issue and assign to @Abhishek-Varma for now, since he has been looking at fp8 models for data-tiling.

hanhanW:
@Abhishek-Varma FYI that we need `ci-extra: test_torch` to trigger torch jobs. They are not run in presubmit by default. I added an empty commit to trigger the job. We can land it if the job is green.

hanhanW (changes requested):
There are failures in https://github.com/iree-org/iree/actions/runs/19512698757/job/55859120039?pr=22387

@Abhishek-Varma I recently learned that we should set it to `val*1.1`. 10% is the usual max delta across chips from the same SKU.

hanhanW (approved):
Thanks, test_torch is passing!

PR #18630, opened 2024-09-28
[python] Add an iree.build package with API/tooling for program building.

stellaraccident:
This is the first step of providing a unified user-oriented build tool for IREE export and compilation. This initial version supports the CLI environment, network/fetch actions, and ONNX import+upgrade. A follow-on step will build out the compiler invocation rules, specifically focusing on getting option, device, and parameter handling and manifest/metadata propagated properly.

The intent is that downstream users of the IREE/Turbine tools can have one stop modules to efficiently export and compile arbitrarily complicated pipelines without reinventing the wheel or needing to do the advanced compiler juggling that comes with some of the more complicated (and performant) workflows.

As a by-product this will standardize a directory layout and manifest/metadata to accompany it for pipeline construction. The intent is that this will simplify serving tooling, which currently has to get a number of out of band parameters and files put together in the right way and matching how they were compiled.

PR #22391, opened 2025-10-23
Fix `iree.build` source directory being gitignore'd

rkayaith:
The gitignore path is currently wrong (it should be `iree/build`, not `iree/compiler/build`). This switches to selectively ignoring `build` directories instead of selectively un-ignoring, since that seems safer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,10 @@
 /build/
 /build-*/
 Testing/
+/compiler/build/
+/runtime/build/
+/integrations/tensorflow/python_projects/iree_tf/build/
+/integrations/tensorflow/python_projects/iree_tflite/build/
 
 # Bazel artifacts
 **/bazel-*
--- a/.gitignore
+++ b/.gitignore
@@ -22,11 +22,9 @@
 .DS_Store
 
 # CMake artifacts
-build/
-build-*/
+/build/
+/build-*/
 Testing/
-# Include iree.build package
-!compiler/bindings/python/iree/compiler/build/
 
 # Bazel artifacts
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -19,11 +19,9 @@
 .DS_Store
 
 # CMake artifacts
-build/
-build-*/
+/build/
+/build-*/
 Testing/
-# Include iree.build package
-!compiler/bindings/python/iree/compiler/build/
```
ScottTodd, line 29: This is more nuanced, see https://github.com/iree-org/iree/pull/18630#discussion_r1781325856

ScottTodd (changes requested):

.gitignore
rkayaith: I don't have full context on those `compiler/build/` and `runtime/build/` directories, but if they're just temporary outputs of the packaging process that can be cleaned up after, I don't think they should be git-ignored - that just makes them harder to notice. But I also think almost none of the stuff we have ignored currently should actually be there (people should use a local gitignore config instead).

I'm happy to defer to you guys though, let me know if you'd like those paths added and I can do that.

ScottTodd: They are build directories from building Python packages using https://github.com/iree-org/iree/blob/main/compiler/setup.py and https://github.com/iree-org/iree/blob/main/runtime/setup.py. Standard stuff, hundreds/thousands of generated files. Could be moved to subfolder .gitignore files or the pattern could be changed here, but they need to be excluded for all developers.

ScottTodd (approved):

PR #22396, opened 2025-10-23
[LinalgExt] Don't force MxK layout for im2col output

Max191:
Introduces a new field, `output_perm`, to the `iree_linalg_ext.im2col` op, and allows the dims in the im2col result to assume any permutation. The `output_perm` represents the permutation from the canonical `B0xB1x...xM0xM1x...xK0xK1x...` layout to the actual result layout.

Without the result dim order constraint, the im2col op can now maintain the inner contiguous dims from the input in the output tensor. This allows for more flexibility in the kinds of transformations we do when copying from global to shared memory.

This PR also updates the `ConvertConvToIm2ColOpPass` so that the im2col op produces an output in the same dim order as the input tensor, which makes the codegen path more similar to what we have for regular GEMM. With this change, the global to shared memory copy will behave the same for convolution as it does for GEMM.

### Benchmark results ###

The following are the geomeans of the speedup with (new) vs. without (old) this PR, computed as `benchmark_time_old / benchmark_time_new` (`> 1` indicates performance improvement). The benchmark times were collected on an MI300X chip, and are subject to noisy measurements, but they give a good indication of the performance impact.

Speedup geomean for all convs: `1.04`
Speedup geomean for forward convs: `1.003`
Speedup geomean for backwards data convs: `0.996`
Speedup geomean for backwards weight convs: `1.13`

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/DerivedConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/DerivedConfigUtils.cpp
@@ -46,11 +46,12 @@ static SmallVector<int64_t> getVectorSizeTileSizes(int64_t rank,
 /// would give tile sizes of [1, 1, 4] no matter what because we won't be able
 /// to collapse the vector.transfer_read that results from this choice of tile
 /// size.
-static SmallVector<int64_t> getVectorTileSizesFromLoopRanges(
-    ArrayRef<int64_t> loopRanges, int64_t numThreads, int64_t vectorSize,
-    bool allowMultiDimCollapse = true, bool vectorizeOutermost = false) {
+static SmallVector<int64_t>
+getVectorTileSizesFromLoopRanges(ArrayRef<int64_t> loopRanges,
+                                 int64_t numThreads, int64_t vectorSize,
+                                 bool allowMultiDimCollapse = true) {
   int64_t rank = loopRanges.size();
-  int64_t targetDim = vectorizeOutermost ? 0 : rank - 1;
+  int64_t targetDim = rank - 1;
   int64_t targetRange = loopRanges[targetDim];
 
   // If any loop ranges are dynamic, default to a simple vector size based
@@ -80,12 +81,11 @@ static SmallVector<int64_t> getVectorTileSizesFromLoopRanges(
   }
 
   // Let the tile size for the target dim be the smaller of the vector size and
-  // the target loop range. Return here, if `allowMultiDimCollapse` is false, or
-  // `vectorizeOutermost` is true, because we don't expect consecutive
-  // dimensions to be vectorizable contiguously for these cases.
+  // the target loop range. Return here, if `allowMultiDimCollapse` is false,
+  // because we don't expect consecutive dimensions to be vectorizable
+  // contiguously for these cases.
   tileSizes[targetDim] = std::min(targetRange, maxVectorSize);
-  if (targetRange >= maxVectorSize || !allowMultiDimCollapse ||
-      vectorizeOutermost) {
+  if (targetRange >= maxVectorSize || !allowMultiDimCollapse) {
     return tileSizes;
   }
 
@@ -124,7 +124,7 @@ SmallVector<int64_t> deriveLinalgOpThreadTileSizes(linalg::LinalgOp linalgOp,
   return tileSizes;
 }
 
-SmallVector<int64_t>
+static SmallVector<int64_t>
 deriveIm2colOpThreadTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
                               int64_t numThreads) {
   if (!im2colOp.hasPureTensorSemantics()) {
@@ -135,20 +135,10 @@ deriveIm2colOpThreadTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
                        getElementTypeOrSelf(im2colOp->getResultTypes()[0])
                            .getIntOrFloatBitWidth();
 
-  // If the im2col input tensor has the batch dim at last, im2col output tensor
-  // has an implicit transpose to move the batch dim in front, and tiling should
-  // be along the batch dim. Currently only a single batch dim is supported for
-  // tiling along the batch dim.
-  unsigned innerDim = im2colOp.getInputRank() - 1;
-  bool singleBatchDimInnermost = im2colOp.getBatchPos().size() == 1 &&
-                                 im2colOp.getBatchPos().back() == innerDim;
-  bool vectorizeOutermost = singleBatchDimInnermost ? true : false;
-
   // Im2col cannot coalesce past the inner/outer most dim so always default to
   // only the inner/outer most tile size being the vector size (or smaller).
   return getVectorTileSizesFromLoopRanges(loopRanges, numThreads, vectorSize,
-                                          /*allowMultiDimCollapse=*/false,
-                                          vectorizeOutermost);
+                                          /*allowMultiDimCollapse=*/false);
 }
 
 SmallVector<int64_t> deriveThreadTileSizes(Operation *op) {
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/DerivedConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/DerivedConfigUtils.cpp
@@ -143,8 +143,7 @@ deriveIm2colOpThreadTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
   SmallVector<int64_t> vectorizableOutputDims = vectorizationMap[innerInputDim];
   std::optional<int64_t> vectorizableDim = std::nullopt;
   if (!vectorizableOutputDims.empty()) {
-    vectorizableDim = *std::max_element(vectorizableOutputDims.begin(),
-                                        vectorizableOutputDims.end());
+    vectorizableDim = *llvm::max_element(vectorizableOutputDims);
   }
 
   // Im2col cannot coalesce past the inner/outer most dim so always default to
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/DerivedConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/DerivedConfigUtils.cpp
@@ -46,12 +46,12 @@ static SmallVector<int64_t> getVectorSizeTileSizes(int64_t rank,
 /// would give tile sizes of [1, 1, 4] no matter what because we won't be able
 /// to collapse the vector.transfer_read that results from this choice of tile
 /// size.
-static SmallVector<int64_t>
-getVectorTileSizesFromLoopRanges(ArrayRef<int64_t> loopRanges,
-                                 int64_t numThreads, int64_t vectorSize,
-                                 bool allowMultiDimCollapse = true) {
+static SmallVector<int64_t> getVectorTileSizesFromLoopRanges(
+    ArrayRef<int64_t> loopRanges, int64_t numThreads, int64_t vectorSize,
+    bool allowMultiDimCollapse = true,
+    std::optional<int64_t> vectorizableDim = std::nullopt) {
   int64_t rank = loopRanges.size();
-  int64_t targetDim = rank - 1;
+  int64_t targetDim = vectorizableDim.has_value() ? *vectorizableDim : rank - 1;
   int64_t targetRange = loopRanges[targetDim];
 
   // If any loop ranges are dynamic, default to a simple vector size based
@@ -135,10 +135,23 @@ deriveIm2colOpThreadTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
                        getElementTypeOrSelf(im2colOp->getResultTypes()[0])
                            .getIntOrFloatBitWidth();
 
+  // Vectorize the innermost output dim that is vectorizable with the innermost
+  // input dim.
+  int64_t innerInputDim = im2colOp.getInputRank() - 1;
+  SmallVector<SmallVector<int64_t>> vectorizationMap =
+      im2colOp.getInputToOutputDimVectorizationMap();
+  SmallVector<int64_t> vectorizableOutputDims = vectorizationMap[innerInputDim];
+  std::optional<int64_t> vectorizableDim = std::nullopt;
+  if (!vectorizableOutputDims.empty()) {
+    vectorizableDim = *std::max_element(vectorizableOutputDims.begin(),
+                                        vectorizableOutputDims.end());
+  }
+
   // Im2col cannot coalesce past the inner/outer most dim so always default to
   // only the inner/outer most tile size being the vector size (or smaller).
   return getVectorTileSizesFromLoopRanges(loopRanges, numThreads, vectorSize,
-                                          /*allowMultiDimCollapse=*/false);
+                                          /*allowMultiDimCollapse=*/false,
+                                          vectorizableDim);
 }
 
 SmallVector<int64_t> deriveThreadTileSizes(Operation *op) {
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
@@ -627,10 +627,7 @@ chooseDimToVectorize(OpBuilder &b, Location loc, Im2colOp im2colOp,
   if (vectorizableOutputDims.empty()) {
     return std::nullopt;
   }
-  SmallVector<int64_t> kOutputDims = im2colOp.getKOutputDims();
-  SetVector<int64_t> kDimSet(kOutputDims.begin(), kOutputDims.end());
-  SetVector<int64_t> kPosSet(im2colOp.getKPos().begin(),
-                             im2colOp.getKPos().end());
+  SetVector<int64_t> kDimSet(llvm::from_range, im2colOp.getKOutputDims());
   // There may be multiple output dims that we can vectorize, so prioritize the
   // innermost dims first.
   std::sort(vectorizableOutputDims.begin(), vectorizableOutputDims.end());
@@ -638,12 +635,6 @@ chooseDimToVectorize(OpBuilder &b, Location loc, Im2colOp im2colOp,
   // one that is vectorizable.
   while (!vectorizableOutputDims.empty()) {
     int64_t outputDimToVectorize = vectorizableOutputDims.pop_back_val();
-    OpFoldResult outputDimSize = iterationDomain[outputDimToVectorize].size;
-    auto constTileSize = getConstantIntValue(outputDimSize);
-    if (constTileSize) {
-      outputDimSize = b.getIndexAttr(constTileSize.value());
-    }
-
     // If a K dim is being vectorized, then it is contiguous along either the
     // input channel dimension, or the filter kernel window. If it is contiguous
     // along the kernel window, then the actual inner slice size is equal to the
@@ -670,6 +661,7 @@ chooseDimToVectorize(OpBuilder &b, Location loc, Im2colOp im2colOp,
       // TODO(Max191): Support vectorization along the M dimension.
       continue;
     }
+    OpFoldResult outputDimSize = iterationDomain[outputDimToVectorize].size;
     if (!willBeContiguousSlice(innerSliceSize, outputDimSize, offset)) {
       continue;
     }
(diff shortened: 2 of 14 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_im2col.mlir
```diff
@@ -379,8 +379,8 @@ module {
 //   CHECK-DAG:       %[[hIDX:.+]] = affine.apply #[[$MAP2]](%[[mParts]]#0, %[[kParts]]#1)
 //   CHECK-DAG:       %[[wIDX:.+]] = affine.apply #[[$MAP2]](%[[mParts]]#1, %[[kParts]]#2)
 //       CHECK:       %[[IN_SLICE:.+]] = tensor.extract_slice %[[ARG0]][%[[kParts]]#0, %[[hIDX]], %[[wIDX]], 0] [1, 1, 1, 4] [1, 1, 1, 1] : tensor<16x26x18x4xf32> to tensor<1x1x1x4xf32>
-//       CHECK:       %[[EMPTY:.+]] = tensor.empty() : tensor<4x1x1x1xf32>
-//       CHECK:       %[[TRANS:.+]] = linalg.transpose ins(%[[IN_SLICE]] : tensor<1x1x1x4xf32>) outs(%[[EMPTY]] : tensor<4x1x1x1xf32>) permutation = [3, 1, 2, 0]
+//       CHECK:       %[[INIT:.+]] = tensor.extract_slice %[[OUT2]][0, %[[M0]], %[[M1]], %[[K]]] [4, 1, 1, 1] [1, 1, 1, 1] : tensor<4x2x2x2xf32> to tensor<4x1x1x1xf32>
+//       CHECK:       %[[TRANS:.+]] = linalg.transpose ins(%[[IN_SLICE]] : tensor<1x1x1x4xf32>) outs(%[[INIT]] : tensor<4x1x1x1xf32>) permutation = [3, 1, 2, 0]
```
Max191, line 383: This is the only real change to the decomposition. We would normally just wait for the empty op to be eliminated, which will produce a slice of the destination, so this just generates the slice directly now.

yzhang93: Nice! It would be good to add another example to show the case where `output_perm` is permutated.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -620,6 +620,45 @@ getIGEMMGenericConvDetails(linalg::LinalgOp linalgOp) {
   auto filterMapGEMM =
       AffineMap::get(numParallelDims + numKDims, 0, filterDims, ctx);
 
+  // Compute the output permutation for the im2col tensor to match the
```
yzhang93, line 664: I think it would be cleaner to move this logic into a separate utility function.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -620,6 +620,45 @@ getIGEMMGenericConvDetails(linalg::LinalgOp linalgOp) {
...
+      }
+      capturedDims.insert(convDim);
+      auto iGEMMDim = cast<AffineDimExpr>(convToIgemmDimMap[convDim]);
+      int64_t colTensorDim = inputMapGEMM.getResultPosition(iGEMMDim).value();
+      colTensorDimsInConvInputOrder.push_back(colTensorDim);
+    }
+  }
+  llvm::SetVector<int64_t> capturedOutputPermDims;
+  SmallVector<int64_t> im2colOutputPerm;
+  for (int64_t dim : llvm::reverse(colTensorDimsInConvInputOrder)) {
```
yzhang93: why a reverse is needed here?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -614,6 +615,69 @@ OnlineAttentionOp::decomposeOperation(OpBuilder &b) {
...
+  std::sort(vectorizableOutputDims.begin(), vectorizableOutputDims.end());
+  // Check each dim in order from innermost to outermost, and return the first
+  // one that is vectorizable.
+  while (!vectorizableOutputDims.empty()) {
+    int64_t outputDimToVectorize = vectorizableOutputDims.pop_back_val();
+    OpFoldResult outputDimSize = iterationDomain[outputDimToVectorize].size;
+    auto constTileSize = getConstantIntValue(outputDimSize);
+    if (constTileSize) {
+      outputDimSize = b.getIndexAttr(constTileSize.value());
+    }
```
yzhang93: This seems to be redundant?

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -620,6 +620,45 @@ getIGEMMGenericConvDetails(linalg::LinalgOp linalgOp) {
   auto filterMapGEMM =
       AffineMap::get(numParallelDims + numKDims, 0, filterDims, ctx);
 
+  // Compute the output permutation for the im2col tensor to match the
+  // dimension order of the input tensor.
```
jerryyin, line 665: I still can't wrap my head around this: Why a layout consistent with input(global) will be better for performance? 

I understand that we need to pay the price of a bad layout, previously, we always write as [B, M, K] layout which is more friendly for MFMAs because there's no need to do any re-layout after ds_read. ds_write, on the contrary, is worse because a transpose need to happen before writing to LDS.

With the new approach we seem to be just doing the opposite, making ds_write easier but ds_read later worse. What is the essence of the improvement here? Do we reduce the bank conflict here?

jerryyin (commented):
I'm unfamiliar with a large chunk of the code base this PR touches. I may not be qualified to review this PR until I've touched those personally. From what I can understand, the output permutation looks like a very reasonable infra.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
Max191: The idea is to keep the im2col op consistent with the layout of the input. It is not necessarily better to move the inefficient accesses from the write stage to the read stage (as is happening here), but keeping the layout consistent allows us to compose other transformations (like a better way of transposing in register/crossbar) that may be better than simply writing in the transposed layout. Otherwise, we are forcing the transpose to happen during the im2col op, when we may want it to happen some other way.

jerryyin: Cool thanks for the explanation. I've read through your changes and they look good. I didn't follow the vectorization or tiling related changes very well. (Maybe a hint for you to offload some of those future implementations to me so I have a  good enough understanding to review :-p ). If you get blocked by the reviews let me know and maybe we can do a short walk-through and approve.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/DerivedConfigUtils.cpp
```diff
@@ -135,20 +135,23 @@ deriveIm2colOpThreadTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
...
+  // Vectorize the innermost output dim that is vectorizable with the innermost
+  // input dim.
+  int64_t innerInputDim = im2colOp.getInputRank() - 1;
+  SmallVector<SmallVector<int64_t>> vectorizationMap =
+      im2colOp.getInputToOutputDimVectorizationMap();
+  SmallVector<int64_t> vectorizableOutputDims = vectorizationMap[innerInputDim];
+  std::optional<int64_t> vectorizableDim = std::nullopt;
+  if (!vectorizableOutputDims.empty()) {
+    vectorizableDim = *std::max_element(vectorizableOutputDims.begin(),
+                                        vectorizableOutputDims.end());
```
kuhar:
```suggestion
    vectorizableDim = *llvm::max_element(vectorizableOutputDims);
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -2352,32 +2352,96 @@ void Im2colOp::setMixedMStrides(SmallVector<OpFoldResult> mStrides) {
...
+  SmallVector<int64_t> inverseOutPerm =
+      invertPermutationVector(getOutputPerm());
+  return llvm::map_to_vector(llvm::seq<int64_t>(begin, end),
+                             [&](int64_t dim) { return inverseOutPerm[dim]; });
+}
+
+SmallVector<SmallVector<int64_t>>
+Im2colOp::getInputToOutputDimVectorizationMap() {
+  SetVector<int64_t> batchInputDims;
+  batchInputDims.insert(getBatchPos().begin(), getBatchPos().end());
```
kuhar:
```suggestion
  SetVector<int64_t> batchInputDims(llvm::from_range, getBatchPos());
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -2352,32 +2352,96 @@ void Im2colOp::setMixedMStrides(SmallVector<OpFoldResult> mStrides) {
...
+  SetVector<int64_t> mInputDims;
+  mInputDims.insert(getMPos().begin(), getMPos().end());
+  SetVector<int64_t> kInputDims;
+  kInputDims.insert(getKPos().begin(), getKPos().end());
```
kuhar: also here

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -620,6 +620,45 @@ getIGEMMGenericConvDetails(linalg::LinalgOp linalgOp) {
...
+  auto inputMap = indexingMaps[0];
```
kuhar: nit: spell out the type

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
Max191: I will add a comment about it. The reason is that we want to keep the innermost occurrences of dimensions, because the innermost one has impact on contiguity of memory access patterns.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
Max191: Good catch!

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -614,6 +615,61 @@ OnlineAttentionOp::decomposeOperation(OpBuilder &b) {
...
+  std::sort(vectorizableOutputDims.begin(), vectorizableOutputDims.end());
+  // Check each dim in order from innermost to outermost, and return the first
+  // one that is vectorizable.
+  while (!vectorizableOutputDims.empty()) {
+    int64_t outputDimToVectorize = vectorizableOutputDims.pop_back_val();
+    // If a K dim is being vectorized, then it is contiguous along either the
+    // input channel dimension, or the filter kernel window. If it is contiguous
+    // along the kernel window, then the actual inner slice size is equal to the
+    // size of the corresponding kernel window dimension. Otherwise, the inner
+    // slice size is just the size of the input tensor's inner dimension.
```
yzhang93, line 642: I'm a bit confused. If it is contiguous along the kernel window, isn't the kernel window dim also inner dimension? What makes this different from the other case?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -751,6 +771,10 @@ FailureOr<SmallVector<Value>> Im2colOp::decomposeOperation(OpBuilder &b) {
   for (scf::ForOp loop : loopNest.loops) {
     ivs.push_back(loop.getInductionVar());
   }
+  if (maybeOutputDimToVectorize.has_value()) {
+    Value zero = arith::ConstantIndexOp::create(b, loc, 0);
+    ivs.insert(ivs.begin() + maybeOutputDimToVectorize.value(), zero);
```
yzhang93, line 779: Can you add some comments of why need to insert zero?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -614,6 +615,61 @@ OnlineAttentionOp::decomposeOperation(OpBuilder &b) {
...
+  SmallVector<SmallVector<int64_t>> vectorizationMap =
+      im2colOp.getInputToOutputDimVectorizationMap();
+  SmallVector<int64_t> vectorizableOutputDims = vectorizationMap[innerInputDim];
+  if (vectorizableOutputDims.empty()) {
+    return std::nullopt;
+  }
+  SetVector<int64_t> kDimSet(llvm::from_range, im2colOp.getKOutputDims());
+  // There may be multiple output dims that we can vectorize, so prioritize the
+  // innermost dims first.
+  std::sort(vectorizableOutputDims.begin(), vectorizableOutputDims.end());
```
kuhar:
```suggestion
  llvm::sort(vectorizableOutputDims);
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -614,6 +615,61 @@ OnlineAttentionOp::decomposeOperation(OpBuilder &b) {
...
+        if (mPos == innerInputDim) {
+          innerSliceSize = kernelSize;
+        }
+      }
+    }
+
+    // If the input slice is contiguous along the innermost dimension, then it
+    // is vectorizable. If it is not, then move on to the next innermost dim.
+    SmallVector<int64_t> mOutputDims = im2colOp.getMOutputDims();
+    SetVector<int64_t> mDimSet(mOutputDims.begin(), mOutputDims.end());
```
kuhar: also here

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -880,22 +896,16 @@ FailureOr<SmallVector<Value>> Im2colOp::decomposeOperation(OpBuilder &b) {
...
   sliceOffsets[kPos] = inputKOffset.front();
-  if (!contiguousAlongB) {
-    int ivIdx = 0;
-    for (auto bPos : getBatchPos()) {
-      sliceOffsets[bPos] = ivs[ivIdx++];
-    }
+  int ivIdx = 0;
+  SmallVector<int64_t> inverseOutputPerm =
+      invertPermutationVector(getOutputPerm());
+  for (auto bPos : getBatchPos()) {
```
kuhar: Use enumerate

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -910,77 +920,61 @@ FailureOr<SmallVector<Value>> Im2colOp::decomposeOperation(OpBuilder &b) {
...
-  } else {
-    int64_t minRank = std::min<int64_t>(inputRank, outputRank);
-    for (int i = 0; i < minRank - 1; i++) {
-      inputTileSizes.push_back(b.getIndexAttr(1));
+  // For now, only extract a 1D slice when the vectorized dim is not innermost
+  // in the output, and the input and output ranks are different. Otherwise,
+  // try to preserve the original rank to avoid rank reducing slices.
+  int64_t sliceRank = std::min<int64_t>(inputRank, outputRank);
+  SmallVector<int64_t> inputToOutputSlicePerm =
+      llvm::to_vector(llvm::seq<int64_t>(0, sliceRank));
```
kuhar:
```suggestion
  auto inputToOutputSlicePerm =
      llvm::to_vector(llvm::seq<int64_t>(0, sliceRank));
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -910,77 +920,61 @@ FailureOr<SmallVector<Value>> Im2colOp::decomposeOperation(OpBuilder &b) {
...
-  if (contiguousAlongB && inputRank != outputRank) {
-    inputTileSizes.push_back(innerInputTileSize);
-  } else {
-    int64_t minRank = std::min<int64_t>(inputRank, outputRank);
-    for (int i = 0; i < minRank - 1; i++) {
-      inputTileSizes.push_back(b.getIndexAttr(1));
+  // For now, only extract a 1D slice when the vectorized dim is not innermost
+  // in the output, and the input and output ranks are different. Otherwise,
+  // try to preserve the original rank to avoid rank reducing slices.
+  int64_t sliceRank = std::min<int64_t>(inputRank, outputRank);
```
kuhar: It's generally discouraged to pass template arguments to `min`/`max` -- cast the arguments instead

yzhang93 (approved):
LGTM. Thanks for refactoring the codes! I just have some minor comments/questions.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
Max191: It's because when the kernel window is innermost, this is like the NCHW layout. The access pattern will be like:
```
+-+-+-+-+-...
|0|1|2| |
+-+-+-+-+-
|3|4|5| |
+-+-+-+-+-
|6|7|8| |
+-+-+-+-+-
| | | | |
+-+-+-+-+-
| | | | |
.
.
.
```

So the corresponding dimension in the input will be `W`, but the size of the tile is only the kernel window size, which is why we use the kernel window size as the `innerSliceSize`.

Max191:
I added some benchmark results that I collected to the PR description, but repeating them here:


### Benchmark results ###

>The following are the geomeans of the speedup with (new) vs. without (old) this PR, computed as `benchmark_time_old / benchmark_time_new` (`> 1` indicates performance improvement). The benchmark times were collected on an MI300X chip, and are subject to noisy measurements, but they give a good indication of the performance impact.
>
>Speedup geomean for all convs: `1.04`
>Speedup geomean for forward convs: `1.003`
>Speedup geomean for backwards data convs: `0.996`
>Speedup geomean for backwards weight convs: `1.13`

Overall, this as an improvement for backwards weight convs, and doesn't affect the other convolution layouts.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h
```diff
@@ -180,6 +180,10 @@ struct IGEMMGenericConvDetails {
   SmallVector<ReassociationIndices> filterReassocIndices;
   /// The iterator type list for a convolution with IGEMM indexing. .
   SmallVector<utils::IteratorType> igemmLoopIterators;
+  /// The output permutation for the im2col tensor with respect to a layout of
+  /// BxMxK. The result of the permutation should match the order that the
+  /// output dims are represented in the input tensor.
+  SmallVector<int64_t> im2colOutputPerm;
```
nirvedhmeshram, line 186: I would like to understand, why we need to express our op with reference to a BxMxK layout, instead if we made an implicit assumption that the op always preserves the same ordering as the input could we have ended up with simplifying some of this logic?

nirvedhmeshram (approved):
Great work relaxing this constraint on the im2col op!

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h
Max191: The input order may not always be enough to get a fully non-ambiguous permutation because the dimensions of the input do not always map 1 to 1 with the output. The two shapes need to be fully defined on their own, and then we have to do what we can to make things line up well for vectorization.

Also, having the permutation referencing something simple like BxMxK makes transformations easier IMO, because we can write logic that uses the simple constant layout, and then just apply permutations to it.

PR #22398, opened 2025-10-23
[DispatchCreation] Add clean up pattern for fusing pad into split reduction dispatch

yzhang93:
When fusing pad into `scf.forall` within split reduction dispatch, we do not want to generate `zeroSliceGuards`. This PR added a clean up pattern to avoid generating such codes.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
@@ -86,12 +86,9 @@ tileOpAndWrapInDispatch(RewriterBase &rewriter, TilingInterface op,
   scf::SCFTileAndFuseOptions tileAndFuseOptions;
   // Only fuse along the dest operand.
   scf::SCFTileAndFuseOptions::ControlFnTy fusionControlFn =
-      [fusePad](tensor::ExtractSliceOp, OpResult result, bool isDestArg)
+      [](tensor::ExtractSliceOp, OpResult result, bool isDestArg)
       -> std::optional<scf::SCFTileAndFuseOptions::ControlFnResult> {
     if (isDestArg) {
-      return scf::SCFTileAndFuseOptions::ControlFnResult{false};
-    }
-    if (fusePad && isa<tensor::PadOp>(result.getOwner())) {
       return scf::SCFTileAndFuseOptions::ControlFnResult{false};
     }
     return std::nullopt;
@@ -102,6 +99,7 @@ tileOpAndWrapInDispatch(RewriterBase &rewriter, TilingInterface op,
   if (fusePad) {
     MLIRContext *context = rewriter.getContext();
     RewritePatternSet cleanupPatterns(context);
+    // When fusing pads we do not want to generate zeroSliceGuards.
     cleanupPatterns.insert<linalg::ExtractSliceOfPadTensorSwapPattern>(
         context,
         [](tensor::ExtractSliceOp) { return /*zeroSliceGuard=*/false; });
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
```diff
@@ -99,6 +99,16 @@ tileOpAndWrapInDispatch(RewriterBase &rewriter, TilingInterface op,
   tileAndFuseOptions.setFusionControlFn(fusionControlFn);
   tileAndFuseOptions.setTilingOptions(std::move(options));
 
+  if (fusePad) {
```
MaheshRavishankar, line 99: I think we need to need to change the control function above to not fuse pads too right?

yzhang93: I don't understand. Do you mean we need this pattern even without enable pad fusion?

Max191: This pattern will replace the regular fusion for pad. So, like Mahesh is suggesting, we need to avoid fusing pad through the regular tiling interface by updating the control function.

Max191 (commented):
Can you also add a test for this?

compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
Groverkss: Rename this to allowZeroSlice like this for consistency: https://github.com/iree-org/iree/blob/3a3249552e9bb10539708a0db8ff4fbb14e76537/compiler/src/iree/compiler/Codegen/Common/TileAndFuseUtils.cpp#L388C9-L388C23 please

Groverkss: cleanup patterns always run before looking for what to fuse in, so you don't really need to change the control function. This is correctly implemented imo. You either allow zero slices (in which case this pattern runs) or you don't (in which case pad fuses and you get the zero slice check)

Max191: I may be wrong, but I'm pretty sure we don't even want to fuse this if we can't fuse it without the zero slice guard, so probably better to remove it from the control function. The `fusePad` boolean is not about whether or not to generate the zero slice guard IIUC, it's just about whether to fuse pads during this pass or not. The strategy for fusing pads should always be to not generate the zero slice guard.

MaheshRavishankar: Yeah what Max said. Its not a clean up pattern thing. Its a user assertion that "I dont want the guard". So dont fuse with pad during tile and fuse, but do it later on by just swapping the extract slice is what I was thinking as well

MaheshRavishankar (approved):

yzhang93:
> Can you also add a test for this?

I modified on the original test by adding necessary checks. https://github.com/iree-org/iree/pull/22398/files#diff-35138a29a1a088b233e39e6acb8ce372a02f74a4c8b9c0852caa17c8f6224923R163

This test include both enable/disable fusePad. Let me know if more tests are needed.

compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
yzhang93: Thanks for all your explanations. I've updated the control function accordingly.

Max191 (approved):

PR #22406, opened 2025-10-24
[NFC] Replace all uses of OpBuilder.create<OpTy> with OpTy::create

Muzammiluddin-Syed-ECE:
In preparation for this upstream commit: https://github.com/llvm/llvm-project/commit/c0b42ec05344707d94805ec795a7bc8d33a09594


kuhar (approved):
Looks good but it seems like there are some test failures?

PR #22407, opened 2025-10-24
Add final dispatch name to AMDGPU Register spill warning

sebvince:
Warnings from --iree-hip-enable-register-spill-warning are difficult to interpret when a module is split into many dispatches. Adding the dispatch name to the warning would make debugging easier

### DIFF
--- a/compiler/plugins/target/ROCM/ROCMTarget.cpp
+++ b/compiler/plugins/target/ROCM/ROCMTarget.cpp
@@ -280,12 +280,17 @@ static void checkRegisterSpilling(IREE::HAL::ExecutableVariantOp &variantOp,
 
   if (!llvm::offloading::amdgpu::getAMDGPUMetaDataFromImage(
           llvm::MemoryBufferRef(obj, ""), infoMap, abiVersion)) {
-    for (const auto &[_, metaData] : infoMap) {
+    for (llvm::StringMapEntry<llvm::offloading::amdgpu::AMDGPUKernelMetaData>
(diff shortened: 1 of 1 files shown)


compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -280,12 +280,17 @@ static void checkRegisterSpilling(IREE::HAL::ExecutableVariantOp &variantOp,
 
   if (!llvm::offloading::amdgpu::getAMDGPUMetaDataFromImage(
           llvm::MemoryBufferRef(obj, ""), infoMap, abiVersion)) {
-    for (const auto &[_, metaData] : infoMap) {
+    for (llvm::StringMapEntry<llvm::offloading::amdgpu::AMDGPUKernelMetaData>
+             &entry : infoMap) {
+      StringRef dispatchName = entry.getKey();
+      llvm::offloading::amdgpu::AMDGPUKernelMetaData &metaData =
+          entry.getValue();
```
kuhar: This is quite verbose, I think the previous impl with structured bindings was more readable...

kuhar (approved):
It might be worth considering moving this to optimization remarks: https://mlir.llvm.org/docs/Remarks/

sebvince:
> It might be worth considering moving this to optimization remarks: https://mlir.llvm.org/docs/Remarks/

Yes, I'll do this in a separate PR and add more info at the same time (such as the occupancy).

Issue #20673, opened 2025-04-29
[tuner] `--iree-codegen-link-tuning-specs` fails to deduplicate symbol names


rkayaith:
### What happened?

Input IR: https://gist.github.com/rkayaith/e7fec15c3b6a37de945bee737996065e
Error:
```
$ iree-opt /tmp/tmpikj732uu/tmp_input.mlir --iree-codegen-link-tuning-specs
/tmp/tmpikj732uu/tmp_input.mlir:56:5: error: redefinition of symbol named 'm0_apply_op_config'
    transform.named_sequence @m0_apply_op_config(%arg0: !transform.any_op {transform.readonly}, %arg1: !transform.any_param {transform.readonly}) {
    ^
/tmp/tmpikj732uu/tmp_input.mlir:56:5: note: see current operation: 
"transform.named_sequence"() <{arg_attrs = [{transform.readonly}, {transform.readonly}], function_type = (!transform.any_op, !transform.any_param) -> (), sym_name = "m0_apply_op_config"}> ({
^bb0(%arg0: !transform.any_op, %arg1: !transform.any_param):
  "transform.annotate"(%arg0, %arg1) <{name = "compilation_info"}> : (!transform.any_op, !transform.any_param) -> ()
  "transform.yield"() : () -> ()
}) : () -> ()
/tmp/tmpikj732uu/tmp_input.mlir:37:5: note: see existing symbol definition here
    transform.named_sequence @apply_op_config(%arg0: !transform.any_op {transform.readonly}, %arg1: !transform.any_param {transform.readonly}) {
    ^
```

### What component(s) does this issue relate to?

Compiler

PR #20127, opened 2025-02-28
[Codegen][Tuner] merge the default td specs

bangtianliu:
This PR adds the implementation of merging all the default td specs (if all inner modules are) in the LinkTuningSpecsPass. It mainly merges multiple transform.foreach_match operations found across different inner modules into a single consolidated operation inside a newly created top-level `__kernel_config` NamedSequenceOp. 

Issue: https://github.com/nod-ai/shark-ai/issues/810

PR #22409, opened 2025-10-24
[Codegen][Tuner] solve name conflicts for merging td specs

bangtianliu:
**Follow-up fix for:** https://github.com/iree-org/iree/pull/20127  
**Fixes:** https://github.com/iree-org/iree/issues/20673, https://github.com/nod-ai/shark-ai/issues/2440

This issue was discovered while tuning BOO, where the same naming conflict pattern caused the pass to fail.

## Problem

When a module contains both a conflicting name and a suffixed variant of that name, the pass could create duplicate names during conflict resolution.

### Example
```mlir
module {  // First module
  @apply_op_config
}
module {  // Second module
  @apply_op_config        // conflicts → would rename to @apply_op_config_0
  @apply_op_config_0      // but this already exists! → BUG
}
```

The second module's `@apply_op_config` would be renamed to `@apply_op_config_0`, but that name already exists in the same module, creating a duplicate.

## Solution

This PR checks the newly generated name against `seenNames` and increments the numeric suffix until finding a unique name:
- `@apply_op_config` → `@apply_op_config_1` (skipping `_0` since it exists)

The fix uses a simple numeric suffix scheme (e.g., `_0`, `_1`, `_2`) instead of module prefixes for cleaner and more maintainable code.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/LinkTuningSpecsPass.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/LinkTuningSpecsPass.cpp
@@ -143,51 +143,24 @@ static TuningSpecsToMerge collectTuningSpecsToMerge(ModuleOp module) {
 }
 
 // Renames a `NamedSequenceOp` to resolve name conflicts caused by merging
-// tuning specs.
-// The name conflict resolution strategy follows below rules:
-//   1. If the `NamedSequenceOp` is inside a module with a valid symbol name,
-//      its new name is prefixed with its containing module's symbol name.
-//   2. If the module has no symbol name, an incrementing counter is used
-//      to generate a unique prefix (e.g., `m0_`, `m1_`, etc.).
-//   3. If the prefixed name still conflicts with an existing name, a numeric
-//      suffix is appended (e.g., `m0_apply_op_config_0`) until a unique name
-//      is found.
+// tuning specs by appending a numeric suffix until a unique name is found.
 static void updateNamedSequenceOp(
     NamedSequenceOp op, OpBuilder &builder,
     llvm::DenseMap<NamedSequenceOp, ForeachMatchOp> &namedSequenceToUser,
-    llvm::DenseMap<ModuleOp, std::string> &unnamedModuleNames,
-    unsigned &unnamedModuleCounter, llvm::DenseSet<StringRef> &seenNames) {
+    llvm::DenseSet<StringRef> &seenNames) {
   StringRef specName = op.getSymName();
-  ModuleOp parentModule = op->getParentOfType<ModuleOp>();
-  assert(parentModule);
-  StringAttr parentSymbol = parentModule.getSymNameAttr();
-  std::string moduleName;
-  if (parentSymbol) {
-    moduleName = parentSymbol.getValue().str();
-  } else {
-    if (unnamedModuleNames.contains(parentModule)) {
-      moduleName = unnamedModuleNames[parentModule];
-    } else {
-      std::string newModuleName =
-          llvm::formatv("m{}", unnamedModuleCounter).str();
-      ++unnamedModuleCounter;
-      unnamedModuleNames[parentModule] = newModuleName;
-      moduleName = newModuleName;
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Common/LinkTuningSpecsPass.cpp
```diff
@@ -174,7 +177,17 @@ static void updateNamedSequenceOp(
...
+
+  // Ensure the new name is unique by appending a counter if needed.
+  std::string uniqueNewSpecName = newSpecName;
+  unsigned suffix = 0;
+  while (seenNames.contains(uniqueNewSpecName)) {
+    uniqueNewSpecName = llvm::formatv("{}_{}", newSpecName, suffix).str();
+    ++suffix;
+  }
+
+  seenNames.insert(builder.getStringAttr(uniqueNewSpecName).getValue());
```
kuhar: There's no need to build a string attribute here

compiler/src/iree/compiler/Codegen/Common/LinkTuningSpecsPass.cpp
```diff
@@ -144,16 +144,19 @@ static TuningSpecsToMerge collectTuningSpecsToMerge(ModuleOp module) {
...
 // Renames a `NamedSequenceOp` to resolve name conflicts caused by merging
 // tuning specs.
-// The name conflict resolution strategy follows below two rules:
+// The name conflict resolution strategy follows below rules:
 //   1. If the `NamedSequenceOp` is inside a module with a valid symbol name,
 //      its new name is prefixed with its containing module's symbol name.
 //   2. If the module has no symbol name, an incrementing counter is used
 //      to generate a unique prefix (e.g., `m0_`, `m1_`, etc.).
+//   3. If the prefixed name still conflicts with an existing name, a numeric
+//      suffix is appended (e.g., `m0_apply_op_config_0`) until a unique name
```
kuhar: Could we make it simpler by appending unique suffix only? Not sure how much value there is in having these prefixes since usually the names we append come from tiny modules with a single matcher only.

kuhar (changes requested):
Typo in the PR title: s/conflictions/conflicts/

Max191 (approved):

compiler/src/iree/compiler/Codegen/Common/LinkTuningSpecsPass.cpp
```diff
@@ -143,38 +143,24 @@ static TuningSpecsToMerge collectTuningSpecsToMerge(ModuleOp module) {
...
-      moduleName = newModuleName;
-    }
+
+  // Ensure the name is unique by appending a numeric suffix if needed.
+  std::string uniqueNewSpecName = specName.str();
+  unsigned suffix = 0;
+  while (seenNames.contains(uniqueNewSpecName)) {
+    uniqueNewSpecName = llvm::formatv("{}_{}", specName, suffix).str();
+    ++suffix;
   }
```
kuhar: nit: You can turn this into a for loop since `suffix` is not used anywhere else

```suggestion
  for (unsigned suffix = 0; seenNames.contains(uniqueNewSpecName); ++suffix) {
    uniqueNewSpecName = llvm::formatv("{}_{}", specName, suffix).str();
  }
```

compiler/src/iree/compiler/Codegen/Common/LinkTuningSpecsPass.cpp
```diff
@@ -237,11 +223,8 @@ static LogicalResult resolveAndMoveNamedSequenceOps(
 
   // Update conflicted named sequence ops.
   if (!nameConflictOps.empty()) {
-    llvm::DenseMap<ModuleOp, std::string> unnamedModuleNames;
-    unsigned unnamedModuleCounter = 0;
     for (NamedSequenceOp op : nameConflictOps) {
```
kuhar: We don't need the `if` statement above

kuhar (approved):

PR #22413, opened 2025-10-24
[CI][TorchModels] Update flags for CLIP test.

MaheshRavishankar:
Dropping some flags that are superfluous and we dont want to really support these.

- `--iree-hip-waves-per-eu=2` is not a flag we really want to support, and looks like compiling with thihs does not actually work as expected since I get a warning saying "expected 2 waves but can schedule only 1"

- `--iree-dispatch-creation-enable-fuse-horizontal-contractions=true` does not seem to have any effect.

- `--iree-opt-const-eval=false`, to re-enable const-eval.

ci-extra: test_torch

### DIFF
--- a/tests/external/iree-test-suites/torch_models/sdxl/modules/clip_gfx942.json
+++ b/tests/external/iree-test-suites/torch_models/sdxl/modules/clip_gfx942.json
@@ -5,10 +5,7 @@
         "--iree-hip-target=gfx942",
         "--iree-opt-level=O3",
         "--iree-opt-generalize-matmul=false",
-        "--iree-opt-const-eval=false",
-        "--iree-hip-waves-per-eu=2",
         "--iree-llvmgpu-enable-prefetch",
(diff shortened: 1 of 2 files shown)


tests/external/iree-test-suites/torch_models/sdxl/modules/clip_gfx942.json
```diff
@@ -5,10 +5,8 @@
         "--iree-hip-target=gfx942",
         "--iree-opt-level=O3",
         "--iree-opt-generalize-matmul=false",
-        "--iree-opt-const-eval=false",
-        "--iree-hip-waves-per-eu=2",
+	"--iree-dispatch-creation-enable-fuse-horizontal-contractions=true",
```
Groverkss: Did you want to remove this or keep this? Either way, can you fix the indentation if you wanted to keep this?

Groverkss (approved):

PR #22416, opened 2025-10-26
[GlobalOptimization]fixed constexpr to avoid hoisting dequant operations

VimalWill:
The PR includes a fix to prevent hoisting of the bit extension (i.e., the dequantization operation), which was previously missing in ConstExpr.cpp. The absence of this logic caused the dequantization operation to be incorrectly hoisted in HoistIntoGlobalsPass.cpp.

Regards,
Vimal William

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.cpp
+++ b/compiler/src/iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.cpp
@@ -6,6 +6,8 @@
 
 #include "iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.h"
 
+#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtDialect.h"
+#include "iree/compiler/Dialect/LinalgExt/Utils/Utils.h"
 #include "iree/compiler/Dialect/Util/Analysis/Constant/OpOracle.h"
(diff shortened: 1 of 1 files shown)


compiler/src/iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.cpp
Groverkss, line 469: Wouldn't this already be handled by this?

Groverkss (changes requested):

compiler/src/iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.cpp
VimalWill: In some cases, the size-based heuristic may fail to detect dequantization hoisting (e.g., for tensors like tensor<3x3x3x32xi8>), resulting in inflated global constants. Introducing the semantic-based condition ensures such bit-extend (i8→f32) operations are excluded from hoisting, preventing unnecessary runtime memory overhead and preserving quantized model compactness.

PS: I found it's hosting dequantization (3x3x3x32xI8 -> 3x3x3x32xf32)

VimalWill: @Groverkss, what's your idea on this ?

PR #22417, opened 2025-10-26
[Codegen] Cleanup VectorLayoutAnalysis testing

Groverkss:
- Moves VectorLayoutAnalysis to a pass instead of a transform op. This cleans up a lot of boiler plate over the transform extension testing which required adding it to each test. Also makes it easier to just run the pass and look at the result when debugging.
- Merge VectorLayoutAnalysis build target with Common target. Common already depends on VectorExt, this doesn't add more dependencies, and was an outlier.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "iree/compiler/Codegen/Common/VectorLayoutAnalysis.h"
+#include "iree/compiler/Codegen/Common/Passes.h"
 
 #include <cassert>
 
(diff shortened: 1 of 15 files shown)


krzysz00 (approved):

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -1410,3 +1411,39 @@ void VectorLayoutAnalysis::dump() {
...
+
+      for (OpResult result : op->getOpResults()) {
+        if (auto layout = analysis.getLayout<Attribute>(result)) {
+          // Print layout attr to a string.
+          std::string layoutStr;
+          llvm::raw_string_ostream s(layoutStr);
+          s << layout;
+          // Emit remark.
+          op->emitRemark("layout of result #" +
+                         Twine(result.getResultNumber()) + " is " + s.str());
```
kuhar, line 1443: This could be:

```suggestion
          op->emitRemark("layout of result #")
                         << result.getResultNumber() << " is " << layout;
```

PR #22418, opened 2025-10-26
[Codegen] Cleanup VectorLayoutAnalysis details

Groverkss:
This patch moves the implementation details of VectorLayoutAnalysis to it's implementation file. The class needed to be exposed in an initial implementation, but exposing these details isn't required anymore, and a simple function call is enough.

PR #22420, opened 2025-10-27
[Codegen] Rewrite VectorLayoutAnalysis to a simpler implementation

Groverkss:
- Cleansup VectorLayoutAnalysis implementation to stop using upstream DataFlowAnalysis framework and instead use a simple worklist to track changes and propagate layout. This works very similar to the previous implementation, but is much less convoluted and much smaller.
- Also fixes a bug in VectorLayoutAnalysis after the upstream DataFlowAnalysis framework moved to ProgramPoint. ProgramPoint tracks the Block iterator but we add operations during the propagation, which meant all ProgramPoint's were invalidated and the analysis was working out of pure luck. It just happened that we added almost everything to the worklist, making it run to fixpoint eventually.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/Transforms.h
+++ b/compiler/src/iree/compiler/Codegen/Common/Transforms.h
@@ -81,7 +81,7 @@ namespace IREE::VectorExt {
 class VectorLayoutInterface;
 } // namespace IREE::VectorExt
 
-/// Analyzes the root op and it's nested ops to propagate vector layouts
+/// Analyzes the root op and its nested ops to propagate vector layouts
 /// originating from to_vector operations. Example:
 ///
 ///    %root = vector.transfer_read
@@ -95,8 +95,8 @@ class VectorLayoutInterface;
 ///          |
 ///          --> %root2, %b and %e must have the same layout
 ///
-/// Here, the user provided an anchor point for %root, fixing it's layout to L.
-/// The layout then uses it's inference rules to find the layout of other
+/// Here, the user provided an anchor point for %root, fixing its layout to L.
+/// The layout then uses its inference rules to find the layout of other
 /// values:
 ///
 ///    %root = vector.transfer_read
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -301,16 +301,18 @@ void LayoutInfo::setLayoutOrClone(OpOperand *val,
   }
   // Always clone constant like ops and set the layout on them.
   OpBuilder b(val->getOwner());
-  Operation *defOp = val->get().getDefiningOp();
-  bool isConstantLike = defOp->hasTrait<OpTrait::ConstantLike>();
-  if (isConstantLike ||
-      llvm::isa_and_nonnull<vector::StepOp, vector::CreateMaskOp,
-                            vector::ConstantMaskOp>(defOp)) {
-    b.setInsertionPoint(defOp);
-    Operation *cloned = b.clone(*defOp);
-    val->set(cloned->getResult(0));
-    layouts[cloned->getResult(0)] = layout;
-    return;
+  if (Operation *defOp = val->get().getDefiningOp()) {
+    bool isConstantLike = defOp->hasTrait<OpTrait::ConstantLike>();
+    bool isDuplicatable =
+        isa<vector::StepOp, vector::CreateMaskOp, vector::ConstantMaskOp>(
+            defOp);
+    if (isConstantLike || isDuplicatable) {
+      b.setInsertionPoint(defOp);
+      Operation *cloned = b.clone(*defOp);
+      val->set(cloned->getResult(0));
+      layouts[cloned->getResult(0)] = layout;
+      return;
+    }
   }
 
   if (!hasLayout(val->get())) {
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -5,1390 +5,348 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Common/Transforms.h"
 #include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
-#include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.h"
 
 #include <cassert>
 
 #include "llvm/ADT/STLExtras.h"
-#include "llvm/Support/Debug.h"
+#include "llvm/Support/DebugLog.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
-#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/Diagnostics.h"
 #include "mlir/IR/Location.h"
 
-#define DEBUG_TYPE "iree-vector-layout-analysis"
-
-using namespace mlir;
-using namespace mlir::iree_compiler;
-using namespace mlir::iree_compiler::IREE::VectorExt;
-
-// Forward declarations.
-class DistributionLayout;
-class PropagateLayout;
-class EnforceLayout;
-
-class DistributionLayout : public AnalysisState {
-public:
-  explicit DistributionLayout(Value val) : AnalysisState(val) {}
-
-  TypedValue<VectorType> getValue() const {
-    LatticeAnchor anchor = getAnchor();
-    Value val = cast<Value>(anchor);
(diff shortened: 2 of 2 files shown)


Groverkss:
Depends on : https://github.com/iree-org/iree/pull/22418

compiler/src/iree/compiler/Codegen/Common/Transforms.h
```diff
@@ -77,6 +77,76 @@ FailureOr<IREETilingResult>
 tileDispatchUsingSCFForOp(RewriterBase &rewriter, TilingInterface op,
                           linalg::LinalgTilingOptions options);
 
+namespace IREE::VectorExt {
+class VectorLayoutInterface;
+} // namespace IREE::VectorExt
+
+/// Analyzes the root op and it's nested ops to propagate vector layouts
```
kuhar:
```suggestion
/// Analyzes the root op and its nested ops to propagate vector layouts
```

compiler/src/iree/compiler/Codegen/Common/Transforms.h
```diff
@@ -77,6 +77,76 @@ FailureOr<IREETilingResult>
...
+///      --> anchored to layout L (using a to_layout op)
+///    %root2 = vector.transfer_read
+///    %c = arith.mulf %root, %b
+///          |
+///          --> %root, %b and %c must have the same layout
+///    %e = arith.divf %b, %root2
+///          |
+///          --> %root2, %b and %e must have the same layout
+///
+/// Here, the user provided an anchor point for %root, fixing it's layout to L.
```
kuhar:
```suggestion
/// Here, the user provided an anchor point for %root, fixing its layout to L.
```

compiler/src/iree/compiler/Codegen/Common/Transforms.h
```diff
@@ -77,6 +77,76 @@ FailureOr<IREETilingResult>
...
+/// The layout then uses it's inference rules to find the layout of other
```
kuhar:
```suggestion
/// The layout then uses its inference rules to find the layout of other
```

kuhar (approved):

kuhar:
There may be corner cases where simple worklist analyzes don't work like IR with unreachable blocks, but I don't think we care about those here.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -4,1440 +4,376 @@
...
-#include "llvm/Support/Debug.h"
+#include "llvm/Support/DebugLog.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
-#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/Diagnostics.h"
 #include "mlir/IR/Location.h"
```
newling:
```suggestion

```

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -4,1440 +4,376 @@
...
+      return;
+    }
+    if (hasLayout(val)) {
+      return;
+    }
+    layouts[val] = layout;
+    forward.push(val);
+    backward.push(val);
+  }
+  void setLayoutOrClone(OpOperand *val, VectorLayoutInterface layout) {
```
newling: Move out of class definition like propagateLayout[Forward|Backward]

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -4,1440 +4,376 @@
...
+    if (!isa<ShapedType>(val->get().getType())) {
+      // Don't set layouts on non-shaped types. This would anyway be an empty
+      // layout.
+      return;
+    }
+    // Always clone constant like ops and set the layout on them.
+    OpBuilder b(val->getOwner());
+    Operation *defOp = val->get().getDefiningOp();
+    if (llvm::isa_and_nonnull<arith::ConstantOp, vector::StepOp,
```
newling: You could use ConstantLike (poison will match) something like 

```c++
    bool isConstantLike = defOp->hasTrait<OpTrait::ConstantLike>();
    if (isConstantLike ||
        llvm::isa<vector::StepOp, vector::CreateMaskOp, vector::ConstantMaskOp>(
            defOp)) {
 ...
```

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -4,1440 +4,376 @@
...
-    vectorLayout = layout;
+    // Otherwise, create a to_layout op to change the layout.
+    Value v = val->get();
+    Value layourtedV = ToLayoutOp::create(b, v.getLoc(), v, layout);
+    val->set(layourtedV);
+    layouts[layourtedV] = layout;
+    return;
   }
+  VectorLayoutInterface getLayout(Value val) { return layouts.lookup(val); }
+  bool hasLayout(Value val) { return layouts.contains(val); }
```
newling:
```suggestion
  bool hasLayout(Value val) const { return layouts.contains(val); }
```

Also, I guess you could make functions `setLayoutOrClone` etc private, although I suppose this struct is more just an internal detail needed to share the graph state (forward/backward) across functions.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -4,1440 +4,376 @@
...
-               .succeeded());
-    vectorLayout = layout;
+    // Otherwise, create a to_layout op to change the layout.
+    Value v = val->get();
+    Value layourtedV = ToLayoutOp::create(b, v.getLoc(), v, layout);
+    val->set(layourtedV);
+    layouts[layourtedV] = layout;
+    return;
   }
+  VectorLayoutInterface getLayout(Value val) { return layouts.lookup(val); }
```
newling:
```suggestion
  VectorLayoutInterface getLayout(Value val) const { return layouts.lookup(val); }
```

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -4,1440 +4,376 @@
...
+      }
+      if (auto ifOp = dyn_cast<scf::IfOp>(parentOp)) {
+        Value thenArg = ifOp.getThenRegion().getArgument(operandIdx);
+        Value elseArg = ifOp.getElseRegion().getArgument(operandIdx);
+        Value result = ifOp->getResult(operandIdx);
+        setLayoutIfUnset(thenArg, layout);
+        setLayoutIfUnset(elseArg, layout);
+        setLayoutIfUnset(result, layout);
+        continue;
+      }
```
newling, line 92:
```suggestion
      }
      assert(false && "must add logic for this yield case");
```

Or some other way of making it easier to debug if we get an scf.while or something one day. I guess we want to evenutally better encapsulate the op specific logic in an interface or something.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -4,1440 +4,376 @@
...
+    info.setLayoutIfUnset(toLayout.getResult(), toLayout.getLayout());
+  });
+  // Propgate all layout information until fixpoint. Give priority to
+  // forward propagation and only do backward propagation when there is no
+  // forward propagation work left.
+  while (!info.forward.empty() || !info.backward.empty()) {
+    SmallVector<Value> changed;
+    if (!info.forward.empty()) {
+      Value val = info.forward.front();
+      info.forward.pop();
```
newling, line 348: Is there a reason to use breadth-first (first-in first-out) here? You could use a SmallVector instead of a std::queue and do depth-first (first-in last-out, push to and pop from back).

newling (commented):
LGTM, definitely easier to follow the logic now, thanks! 

My comments are again all pretty minor and optional.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: It's an internal class anyay, there is really no need for these things to be private

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: The analysis is "best case" I don't think adding an assert there is a good idea.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: I'm mainly trying to replicate what the old analysis did to start. We can improve this over time.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
newling: Point taken, but in general being internal is definitely not a reason to not use `private`. Using `private` makes it immediately clear to the user how objects are used/called.

PR #14718, opened 2023-08-17
[Flow] Teach RaiseSpecialOps to raise tensor.extract and views

Groverkss:
This patch teaches RaiseSpecialOps pass to raise tensor.extract to an input arguement to a linalg.generic. It also teaches it to raise linalg.generic to tensor.expand_shape.

Fixes #14742

PR #22434, opened 2025-10-27
[Global Opt] Raise tensor.extract to input

### LINKED ISSUE #2594: [Fusilli] ConvDGrad should only produce a single dispatch
The sample from https://github.com/nod-ai/shark-ai/pull/2582 should only require a single dispatch. I will add a link to the sample here once it has been merged.

IanWood1:
The PR that added `raiseTensorExtractToInput` (https://github.com/iree-org/iree/pull/14718) only used the `linalg.generic` as an intermediate step when trying to convert to a `tensor` view-like op. However, we should be replace the generic op with the `tensor.extract`-less generic op when it can't be converted to a view-like op.

For example, the linalg op in the test case `test_extract_to_transpose` cannot be converted into a view but should be converted to a simple transpose.


Groverkss (approved):
LGTM, I don't know why I implemented it like that.

PR #22435, opened 2025-10-27
Fixes value bounds interface for ExpandShapeOp

jtuyls:
This was causing an assertion failure in llama 8b fp8 reported by @pdhirajkumarprasad. I will need to fix this as well in upstream and will add a testcase there. Then this will be removed here once upstream is integrated.

### DIFF
--- a/compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
+++ b/compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
@@ -408,7 +408,7 @@ struct ExpandShapeOpValueBoundsInterface
                                        ValueBoundsConstraintSet &cstr) const {
     auto expandOp = cast<memref::ExpandShapeOp>(op);
     assert(value == expandOp.getResult() && "invalid value");
-    cstr.bound(value)[dim] == expandOp.getOutputShape()[dim];
+    cstr.bound(value)[dim] == expandOp.getMixedOutputShape()[dim];
   }
 };
 


pdhirajkumarprasad:
I tried original design with the above change and it's working fine

Yu-Zhewen (approved):

benvanik (changes requested):
You need a test.

(I'm not liking that a memref thing is in the Util dialect specific file - if you're planning to remove this that's good, but in the future don't do that please)

jtuyls:
> You need a test.
> 
> (I'm not liking that a memref thing is in the Util dialect specific file - if you're planning to remove this that's good, but in the future don't do that please)

@benvanik This is not that easy to test in IREE itself because we don't include `test.reify_bounds`. I think it belongs in upstream, as mentioned in the original PR and in the description here as well and it will be removed soon. Here is the upstream PR that fixes it with test: https://github.com/llvm/llvm-project/pull/165333.

Edit: the upstream PR got merged already.

> (I'm not liking that a memref thing is in the Util dialect specific file - if you're planning to remove this that's good, but in the future don't do that please)

This is not in a util dialect specific file, it contains all kinds of interface implementations for upstream operations.

jtuyls:
@benvanik Also, considering this is a hotfix for our e2e shark tests and benchmarks and it will be tested in upstream and removed here very soon, could we just land this as is?

MaheshRavishankar (approved):
This looks fine to me if this is going to be landed upstream..
Or @jtuyls for hot fix like this, just add what you need to the iree fork of LLVM and you can commit this upstream and drop the commit in IREE submodule.

jtuyls:
Closing in favor of https://github.com/iree-org/iree/pull/22460

PR #22437, opened 2025-10-27
Adding HAL virtual memory APIs.

benvanik:
These are not currently implemented anywhere and support queries will return false for all targets.

### DIFF
--- a/runtime/src/iree/hal/allocator.h
+++ b/runtime/src/iree/hal/allocator.h
@@ -255,6 +255,58 @@ typedef struct iree_hal_external_buffer_t {
 } iree_hal_external_buffer_t;
 
 //===----------------------------------------------------------------------===//
+// Virtual Memory Management
+//===----------------------------------------------------------------------===//
+
+// Opaque handle to a physical memory allocation.
+// Physical memory represents actual device memory that is not accessible until
+// mapped to a virtual address space.
+typedef struct iree_hal_physical_memory_t iree_hal_physical_memory_t;
+
+// Memory protection flags for controlling access to virtual address ranges.
+// Maps to mprotect/VirtualProtect (POSIX/Windows), cuMemSetAccess (CUDA), etc.
+enum iree_hal_memory_protection_bits_t {
+  // No access. Removes existing access permissions.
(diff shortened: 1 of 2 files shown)


runtime/src/iree/hal/allocator.h
```diff
@@ -254,6 +254,58 @@ typedef struct iree_hal_external_buffer_t {
...
+  IREE_HAL_MEMORY_PROTECTION_NONE = 0ull,
+
+  // Read-only access.
+  // Platform: PROT_READ (POSIX), PAGE_READONLY (Windows), PROT_READ (CUDA).
+  IREE_HAL_MEMORY_PROTECTION_READ = 1ull << 0,
+
+  // Write-only access (uncommon, typically combined with READ).
+  // Platform: Not directly supported on CPU (needs READ), PROT_WRITE (CUDA).
+  IREE_HAL_MEMORY_PROTECTION_WRITE = 1ull << 1,
+
```
AWoloszyn, line 280: No execute bit? :D

runtime/src/iree/hal/allocator.h
```diff
@@ -422,6 +474,167 @@ IREE_API_EXPORT iree_status_t iree_hal_allocator_export_buffer(
...
+// The reserved range has no access permissions and cannot be accessed until
+// physical memory is mapped via iree_hal_allocator_virtual_memory_map and
+// permissions are granted via iree_hal_allocator_virtual_memory_protect.
+//
+// The returned |out_virtual_buffer| is a buffer representing the reserved VA
+// range. It must be released by the caller with
+// iree_hal_allocator_virtual_memory_release.
+//
+// Returns IREE_STATUS_UNAVAILABLE if virtual memory is not supported.
+IREE_API_EXPORT iree_status_t iree_hal_allocator_virtual_memory_reserve(
```
AWoloszyn, line 523: Are we allowed to map part of this range, and leave other parts unmapped? So long as we only access the mapped parts.

benvanik: yep!

runtime/src/iree/hal/allocator.h
benvanik: None in HSA/CUDA/HIP, so we're good for now :)

AWoloszyn (approved):

PR #22441, opened 2025-10-28
[LinalgExt] Initial support for aten::flex_attention and rewrite to linalg_ext.attention 

keshavvinayak01:
This PR adds comprehensive support for converting PyTorch's `torch::hop_flex_attention` operation to IREE's `linalg_ext.attention` operation. The implementation includes proper handling of score modification functions, mask functions, and correct softmax computation using base-e exponential (exp) instead of base-2 exponential (exp2).

## Changes

- Added `FlexAttentionOpConversion` pattern to convert `torch.hop_flex_attention` to `linalg_ext.attention`
- Modified `AttentionOp` to pass iteration indices as block arguments to the region
- Enables score modification functions to access batch, head, query sequence, and key sequence indices
- Added lit tests for both LSE (log-sum-exp) and non-LSE cases
- Simplified region handling by using block arguments instead

## Fixes:
- PyTorch's flex attention already supplies the correct scale for base-e softmax. This commit fixes the computation to use `exp` instead of `exp2`:
- The use_exp2 flag is mostly unused in dialect conversions and passes, I presume it's used as a KernelOption. The changes here will not modify the default behavior.

## Testing:
- I ran the entire flex_attention_hop implementation with randomised input tensors, (Also see [torch-mlir](https://github.com/llvm/torch-mlir/pull/4366)) through `aot.export` and compared against eager mode, and I noticed no accuracy losses (On CPU)
- Command: `iree-compile --iree-stream-partitioning-favor=min-peak-memory --iree-hal-target-device=local --iree-hal-local-target-device-backends=llvm-cpu --iree-llvmcpu-target-triple=x86_64-pc-linux-elf --iree-llvmcpu-debug-symbols=false ../torch-mlir/exported_ir.mlir -o ./flex_attention_cpu.vmfb`

### DIFF
--- a/compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
@@ -9,6 +9,7 @@
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
 #include "iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Arith/Utils/Utils.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Math/IR/Math.h"
@@ -210,10 +211,10 @@ struct FlexAttentionOpConversion
     // Check enable_gqa flag
     bool enableGqa = false;
     if (!matchPattern(enableGqaValue,
-                      torch::Torch::m_TorchConstantBool(&enableGqa)))
+                      torch::Torch::m_TorchConstantBool(&enableGqa))) {
       return rewriter.notifyMatchFailure(op,
                                          "expected constant enable_gqa value");
-
+    }
     // TODO: GQA (Grouped Query Attention) support is not yet implemented
     if (enableGqa) {
       return rewriter.notifyMatchFailure(
@@ -345,9 +346,9 @@ struct FlexAttentionOpConversion
             }
 
             // Call mask_mod_fn(b, h, q_idx, kv_idx)
-            auto callOp =
-                b.create<func::CallOp>(loc, TypeRange{torchBoolScalarType},
-                                       maskModSymbol, ValueRange(torchIndices));
+            auto callOp = b.create<func::CallOp>(
+                loc, TypeRange{torchBoolScalarType}, *maskModSymbol,
+                ValueRange(torchIndices));
             Value torchMaskResult = callOp.getResult(0);
 
             // Convert result back to builtin tensor
@@ -382,8 +383,18 @@ struct FlexAttentionOpConversion
       }
     }
 
-    Value outputTensor = rewriter.create<tensor::EmptyOp>(
+    Value outputTensorEmpty = rewriter.create<tensor::EmptyOp>(
         loc, outputShape, floatType, outputDynSizes);
+
+    // Initialize output tensor with identity value (0.0 for addition)
+    Value outputInit = arith::getIdentityValue(arith::AtomicRMWKind::addf,
+                                               floatType, rewriter, loc,
+                                               /*useOnlyFiniteValue=*/true);
+    Value outputTensor =
+        rewriter
+            .create<linalg::FillOp>(loc, ValueRange{outputInit},
+                                    ValueRange{outputTensorEmpty})
+            .getResult(0);
 
     // Create max and sum tensors [B, H, M]
     SmallVector<int64_t> maxSumShape = {batch, numHeads, seqLenQ};
@@ -396,10 +407,28 @@ struct FlexAttentionOpConversion
       }
     }
 
-    Value maxTensor = rewriter.create<tensor::EmptyOp>(
+    Value maxTensorEmpty = rewriter.create<tensor::EmptyOp>(
         loc, maxSumShape, floatType, maxSumDynSizes);
-    Value sumTensor = rewriter.create<tensor::EmptyOp>(
+    Value sumTensorEmpty = rewriter.create<tensor::EmptyOp>(
         loc, maxSumShape, floatType, maxSumDynSizes);
+
+    // Initialize max and sum tensors with identity values for reductions
+    // Max reduction identity: -inf
+    Value maxInit = arith::getIdentityValue(arith::AtomicRMWKind::maximumf,
+                                            floatType, rewriter, loc,
+                                            /*useOnlyFiniteValue=*/true);
+    // Sum reduction identity: 0.0
+    Value sumInit = arith::getIdentityValue(arith::AtomicRMWKind::addf,
+                                            floatType, rewriter, loc);
+
+    Value maxTensor = rewriter
+                          .create<linalg::FillOp>(loc, ValueRange{maxInit},
+                                                  ValueRange{maxTensorEmpty})
+                          .getResult(0);
+    Value sumTensor = rewriter
+                          .create<linalg::FillOp>(loc, ValueRange{sumInit},
+                                                  ValueRange{sumTensorEmpty})
+                          .getResult(0);
 
     // Build indexing maps for online_attention
     // Standard maps: Q, K, V, scale, [mask], output, max, sum
@@ -480,7 +509,7 @@ struct FlexAttentionOpConversion
 
         // Call score_mod_fn(score, b, h, q_idx, kv_idx)
         auto callOp = rewriter.create<func::CallOp>(
-            loc, TypeRange{torchScalarType}, scoreModeSymbol,
+            loc, TypeRange{torchScalarType}, *scoreModeSymbol,
             ValueRange(callArgs));
         Value torchResult = callOp.getResult(0);
 
--- a/compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
@@ -175,6 +175,19 @@ struct FftRfftOpConversion
 struct FlexAttentionOpConversion
     : public OpRewritePattern<torch::Torch::AtenFlexAttentionOp> {
   using OpRewritePattern::OpRewritePattern;
+
+  // Makes it convenient to pass around commonly used types.
+  struct TypeInfo {
+    Type i32Type;
+    Type si32Type;
+    RankedTensorType scalarTensorType;
+    RankedTensorType i32ScalarTensorType;
+    RankedTensorType boolScalarTensorType;
+    torch::Torch::ValueTensorType torchScalarType;
+    torch::Torch::ValueTensorType torchI32ScalarType;
+    torch::Torch::ValueTensorType torchBoolScalarType;
+  } mutable typeInfo;
+
   LogicalResult matchAndRewrite(torch::Torch::AtenFlexAttentionOp op,
                                 PatternRewriter &rewriter) const override {
     Location loc = op.getLoc();
@@ -183,6 +196,8 @@ struct FlexAttentionOpConversion
     Value key = op.getKey();
     Value value = op.getValue();
     Value scaleValue = op.getScale();
+    auto scoreModSymbol = op.getScoreModFn();
+    auto maskModSymbol = op.getMaskModFn();
 
     bool returnLseValue;
     if (!matchPattern(op.getReturnLse(),
@@ -236,120 +251,29 @@ struct FlexAttentionOpConversion
     Value scale = arith::ConstantOp::create(
         rewriter, loc, floatType, rewriter.getFloatAttr(floatType, scaleVal));
 
-    auto scoreModeSymbol = op.getScoreModFn();
-    auto maskModSymbol = op.getMaskModFn();
-
-    auto toBuiltinTensor = [&](Value torchTensor) -> Value {
-      auto torchType =
-          cast<torch::Torch::ValueTensorType>(torchTensor.getType());
-      return torch::TorchConversion::ToBuiltinTensorOp::create(
-          rewriter, loc, torchType.toBuiltinTensor(), torchTensor);
-    };
-
-    Value builtinQuery = toBuiltinTensor(query);
-    Value builtinKey = toBuiltinTensor(key);
-    Value builtinValue = toBuiltinTensor(value);
+    Value builtinQuery = convertToBuiltinTensor(rewriter, loc, query);
+    Value builtinKey = convertToBuiltinTensor(rewriter, loc, key);
+    Value builtinValue = convertToBuiltinTensor(rewriter, loc, value);
 
     // Declare common types for mask and score modification regions.
-    Type i32Type = rewriter.getI32Type();
-    Type si32Type =
-        IntegerType::get(rewriter.getContext(), 32, IntegerType::Signed);
-    auto scalarTensorType = RankedTensorType::get({}, floatType);
-    auto i32ScalarTensorType = RankedTensorType::get({}, i32Type);
-    auto boolScalarTensorType = RankedTensorType::get({}, rewriter.getI1Type());
-    auto torchScalarType = rewriter.getType<torch::Torch::ValueTensorType>(
-        ArrayRef<int64_t>{}, floatType);
-    auto torchI32ScalarType = rewriter.getType<torch::Torch::ValueTensorType>(
-        ArrayRef<int64_t>{}, si32Type);
-    auto torchBoolScalarType = rewriter.getType<torch::Torch::ValueTensorType>(
-        ArrayRef<int64_t>{}, rewriter.getI1Type());
+    setTypeInfo(rewriter, floatType);
     Value zero = arith::ConstantFloatOp::create(
         rewriter, loc, floatType,
         llvm::APFloat::getZero(floatType.getFloatSemantics()));
     Value mask;
     if (maskModSymbol) {
-      // Create mask tensor [B, H, M, N] with values 0.0 (attend) or -inf
-      // (mask).
-      SmallVector<int64_t> maskShape = {batch, numHeads, seqLenQ, seqLenKV};
-      SmallVector<Value> maskDynSizes;
-
-      // Handling dynamic dimensions
-      for (int i = 0; i < 4; ++i) {
-        if (maskShape[i] == torch::Torch::kUnknownSize) {
-          Value idx =
-              arith::ConstantIndexOp::create(rewriter, loc, std::min(i, 2));
-          Value dim = tensor::DimOp::create(
-              rewriter, loc, i <= 2 ? builtinQuery : builtinKey, idx);
-          maskDynSizes.push_back(dim);
-        }
-      }
-
-      Value maskTensor = tensor::EmptyOp::create(rewriter, loc, maskShape,
-                                                 floatType, maskDynSizes);
-      // Create linalg.generic to materialize mask
-      SmallVector<AffineMap> maskMaps;
-      maskMaps.push_back(AffineMap::getMultiDimIdentityMap(4, ctx));
-
-      SmallVector<utils::IteratorType> iteratorTypes(
-          4, utils::IteratorType::parallel);
-
-      Value negInf = arith::ConstantFloatOp::create(
-          rewriter, loc, floatType,
-          llvm::APFloat::getInf(floatType.getFloatSemantics(),
-                                /*Negative=*/true));
-
-      auto maskGeneric = linalg::GenericOp::create(
-          rewriter, loc, TypeRange{maskTensor.getType()}, ValueRange{},
-          ValueRange{maskTensor}, maskMaps, iteratorTypes,
-          [&](OpBuilder &b, Location loc, ValueRange args) {
-            // Get indices and convert to torch tensors.
-            SmallVector<Value> torchIndices;
-            for (unsigned i = 0; i < 4; ++i) {
-              Value idx = linalg::IndexOp::create(b, loc, i);
-              Value idxI32 = arith::IndexCastOp::create(b, loc, i32Type, idx);
-              Value idxTensor = tensor::FromElementsOp::create(
-                  b, loc, i32ScalarTensorType, ValueRange{idxI32});
-              Value torchIdx =
-                  torch::TorchConversion::FromBuiltinTensorOp::create(
-                      b, loc, torchI32ScalarType, idxTensor);
-              torchIndices.push_back(torchIdx);
-            }
-
-            // Call mask_mod_fn(b, h, q_idx, kv_idx).
-            auto callOp =
-                func::CallOp::create(b, loc, TypeRange{torchBoolScalarType},
-                                     *maskModSymbol, ValueRange(torchIndices));
-            Value torchMaskResult = callOp.getResult(0);
-
-            Value maskResult =
-                torch::TorchConversion::ToBuiltinTensorOp::create(
-                    b, loc, boolScalarTensorType, torchMaskResult);
-
-            Value maskBool =
-                tensor::ExtractOp::create(b, loc, maskResult, ValueRange{});
-
-            // Select: mask ? 0.0 : -inf.
-            Value maskValue =
-                arith::SelectOp::create(b, loc, maskBool, zero, negInf);
-
-            linalg::YieldOp::create(b, loc, maskValue);
-          });
-
-      mask = maskGeneric.getResult(0);
+      FlatSymbolRefAttr maskModRef =
+          FlatSymbolRefAttr::get(ctx, *maskModSymbol);
+      mask = createModifiedMask(rewriter, loc, ctx, maskModRef, batch, numHeads,
+                                seqLenQ, seqLenKV, floatType, builtinQuery,
+                                builtinKey, zero);
     }
 
-    // Create output tensor for attention
+    // Create output tensor for attention.
     SmallVector<Value> outputDynSizes;
     SmallVector<int64_t> outputShape = {batch, numHeads, seqLenQ, valueDim};
-    // Handle dynamic dimensions
-    for (int i = 0; i < 4; ++i) {
-      if (outputShape[i] == torch::Torch::kUnknownSize) {
-        Value idx = arith::ConstantIndexOp::create(rewriter, loc, i);
-        Value dim = tensor::DimOp::create(
-            rewriter, loc, i < 3 ? builtinQuery : builtinValue, idx);
-        outputDynSizes.push_back(dim);
-      }
-    }
+    computeDynamicSizes(rewriter, loc, outputShape, outputDynSizes,
+                        builtinQuery, builtinValue);
 
     // Initialize output tensor with identity value (0.0 for addition).
     Value outputInit = arith::getIdentityValue(arith::AtomicRMWKind::addf,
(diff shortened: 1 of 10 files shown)


hanhanW:
I can review after you fix all build failures. It looks not ready to me when CI is so red.

compiler/plugins/input/Torch/InputConversion/test/unstructured_linalg_ext.mlir
```diff
@@ -99,3 +99,150 @@ func.func @fft_rfft.last(%arg0: !torch.vtensor<[3,8,16],f32>) -> !torch.vtensor<
 // CHECK:             %[[VAR12:.*]] = torch.aten.cat %[[VAR11]], %[[INTM1]] : !torch.list<vtensor<[3,8,9,1],f32>>, !torch.int -> !torch.vtensor<[3,8,9,2],f32>
 // CHECK:             %[[VAR13:.*]] = torch.aten.view_as_complex %[[VAR12]] : !torch.vtensor<[3,8,9,2],f32> -> !torch.vtensor<[3,8,9],complex<f32>>
 // CHECK:             return %[[VAR13]] : !torch.vtensor<[3,8,9],complex<f32>>
+
```
hanhanW, line 103: Missing a separator (i.e., `// -----`) here.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+    // Get tensor types
+    auto queryType = cast<torch::Torch::ValueTensorType>(query.getType());
+    auto keyType = cast<torch::Torch::ValueTensorType>(key.getType());
+    auto valueType = cast<torch::Torch::ValueTensorType>(value.getType());
+
+    // TODO: See if this check is necessary (Op assertations)
+    if (!queryType.hasSizes() || !keyType.hasSizes() || !valueType.hasSizes()) {
+      return rewriter.notifyMatchFailure(op,
+                                         "expected tensors with known sizes");
+    }
```
hanhanW: It looks like a debug code? So is it necessary? What does `Op assertations`? Can you expand the context a bit?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+
+    ArrayRef<int64_t> queryShape = queryType.getSizes();
+    ArrayRef<int64_t> valueShape = valueType.getSizes();
+
+    // TODO: See if this check is necessary (Op assertations)
+    // Query shape: [B, H, M, E]
+    if (queryShape.size() != 4) {
+      return rewriter.notifyMatchFailure(op, "expected 4D query tensor");
+    }
```
hanhanW: ditto

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
 
+struct FlexAttentionOpConversion
+    : public OpRewritePattern<torch::Torch::AtenFlexAttentionOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(torch::Torch::AtenFlexAttentionOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    MLIRContext *ctx = getContext();
+    // Extract operands (new order: query, key, value, 8 block_mask tensors,
+    // scale, return_lse)
```
hanhanW: I don't see why we need the comment. It is adding noise to me. We already have getters, so the order does not matter at all?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+    Value scale = arith::ConstantOp::create(
+        rewriter, loc, floatType, rewriter.getFloatAttr(floatType, scaleVal));
+
+    // TODO: See if this check is necessary (Op assertations)
+    // Check return_lse flag
+    bool returnLse = false;
+    if (!matchPattern(returnLseValue,
+                      torch::Torch::m_TorchConstantBool(&returnLse)))
+      return rewriter.notifyMatchFailure(op,
+                                         "expected constant return_lse value");
```
hanhanW: ditto: wrap braces

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+    }
+    FloatType floatType = cast<FloatType>(elementType);
+
+    // Handle scale parameter
+    // Default scale: 1.0 / sqrt(head_dim)
+    double scaleVal;
+    if (!matchPattern(scaleValue,
+                      torch::Torch::m_TorchConstantFloat(&scaleVal)))
+      // Fallback to default scale
+      scaleVal = 1.0 / std::sqrt(static_cast<double>(headDim));
```
hanhanW: IREE style nit: wrap braces. Please check other code as well.

note: you also need braces even you are following LLVM style: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+    if (headDim == torch::Torch::kUnknownSize) {
+      return rewriter.notifyMatchFailure(op, "NYI: dynamic head dimension");
+    }
+
+    // Get element type
+    Type elementType = queryType.getOptionalDtype();
+    if (!elementType || !isa<FloatType>(elementType)) {
+      return rewriter.notifyMatchFailure(op, "expected float element type");
+    }
+    FloatType floatType = cast<FloatType>(elementType);
```
hanhanW: Why not just `auto floatType = dyn_cast<queryType.getOptionalDtype()` and check floatType directly?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+      for (int i = 0; i < 4; ++i) {
+        if (maskShape[i] == torch::Torch::kUnknownSize) {
+          Value idx = arith::ConstantIndexOp::create(rewriter, loc, i);
+          Value dim =
+              (i < 2) ? tensor::DimOp::create(rewriter, loc, builtinQuery, idx)
+              : (i == 2)
+                  ? tensor::DimOp::create(rewriter, loc, builtinQuery, idx)
+                  : tensor::DimOp::create(
+                        rewriter, loc, builtinKey,
+                        arith::ConstantIndexOp::create(rewriter, loc, 2));
```
hanhanW: Please break it into several if-else statements. This is not easy to follow because there are 7 lines.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+          rewriter, loc, floatType,
+          rewriter.getFloatAttr(floatType,
+                                -std::numeric_limits<double>::infinity()));
+
+      auto maskGeneric = linalg::GenericOp::create(
+          rewriter, loc, TypeRange{maskTensor.getType()}, ValueRange{},
+          ValueRange{maskTensor}, maskMaps, iteratorTypes,
+          [&](OpBuilder &b, Location loc, ValueRange args) {
+            // Get indices and convert to torch tensors
+
```
hanhanW: remove the blank line. The above comment belongs to the below implementation, right?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+      SmallVector<utils::IteratorType> iteratorTypes(
+          4, utils::IteratorType::parallel);
+
+      Value zero = arith::ConstantOp::create(
+          rewriter, loc, floatType, rewriter.getFloatAttr(floatType, 0.0));
+      Value negInf = arith::ConstantOp::create(
+          rewriter, loc, floatType,
+          rewriter.getFloatAttr(floatType,
+                                -std::numeric_limits<double>::infinity()));
```
hanhanW: For zero, you can use `arith::ConstantFloatOp::create(..., floatType, 0)`.

For negInf, you should try this method: https://github.com/llvm/llvm-project/blob/3a2de951a40877e3d17aa87abb35565da1040612/llvm/include/llvm/ADT/APFloat.h#L1080-L1084

`double` != floatType. The precision can mismatch.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +171,349 @@ struct FftRfftOpConversion
...
+
+    indexingMaps.push_back(oMap);
+
+    // Create decomposition config with use_exp2 flag
+    // PyTorch's compiled kernels use exp2 for performance, so we match that
+    SmallVector<NamedAttribute> decompositionConfigAttrs;
+    decompositionConfigAttrs.push_back(
+        rewriter.getNamedAttr("use_exp2", rewriter.getBoolAttr(false)));
+    DictionaryAttr decompositionConfig =
+        rewriter.getDictionaryAttr(decompositionConfigAttrs);
```
hanhanW: This is a red flag. We should not pre-set any config in lowering. Can we identify such case and set the config in codegen?

E.g.,


> PyTorch's compiled kernels use exp2 for performance, so we match that

Is it true for all the backends in pytorch? You are leaking other project's config to IREE.

compiler/plugins/input/Torch/InputConversion/test/unstructured_linalg_ext.mlir
```diff
@@ -99,3 +99,150 @@ func.func @fft_rfft.last(%arg0: !torch.vtensor<[3,8,16],f32>) -> !torch.vtensor<
...
+  %3 = torch.aten.add.Tensor %arg0, %2, %int1_0 : !torch.vtensor<[],f32>, !torch.vtensor<[],f32>, !torch.int -> !torch.vtensor<[],f32>
+  %int1_1 = torch.constant.int 1
+  %4 = torch.aten.add.Tensor %3, %1, %int1_1 : !torch.vtensor<[],f32>, !torch.vtensor<[],f32>, !torch.int -> !torch.vtensor<[],f32>
+  %5 = torch.aten.tanh %4 : !torch.vtensor<[],f32> -> !torch.vtensor<[],f32>
+  return %5 : !torch.vtensor<[],f32>
+}
+func.func private @sdpa_mask0(%arg0: !torch.vtensor<[],si32>, %arg1: !torch.vtensor<[],si32>, %arg2: !torch.vtensor<[],si32>, %arg3: !torch.vtensor<[],si32>) -> !torch.vtensor<[],i1> {
+  %0 = torch.aten.ge.Tensor %arg2, %arg3 : !torch.vtensor<[],si32>, !torch.vtensor<[],si32> -> !torch.vtensor<[],i1>
+  return %0 : !torch.vtensor<[],i1>
+}
```
hanhanW: This is op level lowering test. Do you really need all of them? Can you simplify it to a single attention op?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -171,13 +522,15 @@ class ConvertTorchUnstructuredToLinalgExtPass final
...
                     linalg::LinalgDialect, arith::ArithDialect,
+                    math::MathDialect, func::FuncDialect,
                     torch::TorchConversion::TorchConversionDialect>();
   }
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     RewritePatternSet patterns(context);
 
     patterns.add<FftRfftOpConversion>(context);
+    patterns.add<FlexAttentionOpConversion>(context);
```
hanhanW: Merge them.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
hanhanW, line 1: Instead of having ~300 lines function, can you try to break them into several reasonable small functions?

There is a magic number (i.e., 4) in many places. Please declare a meaningful variable for it, if applicable. Magic number is very bad.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -192,18 +191,54 @@ static Value applyPostQKMatmulElementwise(OpBuilder &builder, Location loc,
...
+        // Clone the region body inline
+        IRMapping mapping;
+        for (auto [arg, regionArg] :
+             llvm::zip_equal(regionArgs, region.front().getArguments())) {
+          mapping.map(regionArg, arg);
+        }
+
+        for (Operation &op : region.front().without_terminator()) {
+          b.clone(op, mapping);
+        }
```
hanhanW, line 230: remove the blank line as they belong to the same comments.

hanhanW (changes requested):
High level comment:
- Please add period at the end of all your comments.
- The best code is self-documenting. Please delete redundant comments: https://google.github.io/styleguide/cppguide.html#Comments E.g., `// Yield modified score` is redundant to me.

I did not point out all the code that has the issue, please do self-review before you request the next review. Thanks!

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: Yeah, it's a small check to verify the input types has sizes, required by the rewrite. I'll remove the comments.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: Thanks for pointing this out, this is a massive assumption, and the computation _might_ vary backend-to-backend. Maybe I can pass a flag in the `iree-linalg-ext-decompose-attention` pass? I verified we need to use `exp2` there, at least on the CPU for correct computations. While testing for GPU, it'd give me the liberty to turn that flag or not, assuming that pass is even part of the pipeline (Most likely not).

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -798,6 +789,12 @@ def IREELinalgExt_AttentionOp : IREELinalgExt_Op<"attention",
     If an additional mask argument M is included, the result of the first matmul is modified according to:
 
     Q @ K.T += M
+
+    Region:
+      The region body can receive either 1 or 5 block arguments:
+        - 1 argument (legacy): score (element type of output)
+        - 5 arguments: score, b (batch index), h (head index), m (query seq index), n (key/value seq index)
```
IanWood1, line 796: Is the idea that there will be a follow-up to remove support for the "legacy" 1 argument version?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
hanhanW: > Maybe I can pass a flag in the iree-linalg-ext-decompose-attention pass?

Starting it with this is better than the setting it in the lowering.

> I verified we need to use exp2 there, at least on the CPU for correct computations.

I haven't reviewed the details yet, but it sounds like numeric difference from fp re-association. The golden values from the framework might be off, or we need to adjust tolerance for the op test.

hanhanW:
Please click `Re-request` item or ping me on discord when it is ready for the review. Thanks! :)

image

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
keshavvinayak01: I guess eventually, but that's not my immediate plan. A follow up PR would be to add tiling support for indexed regions in attention ops, that'd be more important.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: I didn't try an op test, I just compared the output of eager-mode vs our exported path. From [flex_attention.py](https://github.com/pytorch/pytorch/blob/0fabc3ba44823f257e70ce397d989c8de5e362c1/torch/nn/attention/flex_attention.py#L1647C1-L1663C35) : 
```
ln2 = math.log(2.0)
return_lse = return_lse or return_aux is not None and return_aux.lse
return_max = return_aux is not None and return_aux.max_scores


lse_scaled = lse * ln2 if (return_lse and lse.numel() > 0) else None
max_scaled = (
    max_scores * ln2 if (return_max and max_scores.numel() > 0) else None
)


if return_aux is not None:
    return out, AuxOutput(
        lse=lse_scaled,
        max_scores=max_scaled,
    )


if return_lse:
    return out, lse_scaled
```

The computation uses exp in the C++ template implementation. The scaling by ln2 converts the base-2 LSE to base-e for the return value. To match PyTorch's numerical behavior, we should use exp (not exp2) in the computation, even though the internal LSE is stored in base-2. 

Anyways, this is my empirical conclusion, let me know if it makes sense.

compiler/plugins/input/Torch/InputConversion/test/unstructured_linalg_ext.mlir
keshavvinayak01: I could rewrite a simpler score function, but yes, I need an op lowering test that demonstrates we can properly rewrite functional references.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: It is likely the fp re-association that causes this numerical variance, but It's not within a tolerable range. I've observed differences of upto ~50% in element wise values.

compiler/plugins/input/Torch/InputConversion/test/unstructured_linalg_ext.mlir
hanhanW: I see, it looks like `score_mod_fn, mask_mod_fn` are two new optional attributes from the new PR: https://github.com/llvm/torch-mlir/pull/4366

To have better test coverage, you should add three more tests for different configurations.

keshavvinayak01: Done.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +172,353 @@ struct FftRfftOpConversion
...
+    : public OpRewritePattern<torch::Torch::AtenFlexAttentionOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  // Attention tensors are 4D: [batch, head, query_seq, key_seq].
+  static constexpr int kAttentionRank = 4;
+  // Modification functions receive 4 index arguments: (b, h, m, n).
+  static constexpr int kNumModificationIndices = 4;
+
+  // Makes it convenient to pass around commonly used types.
+  struct TypeInfo {
```
MaheshRavishankar, line 185: I dont think we want to do this. This is an unnecessary indirection, and potentially this is creating new types in the global context without needing it.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +172,353 @@ struct FftRfftOpConversion
...
+    int64_t seqLenKV = keyType.getSizes()[2];
+    int64_t valueDim = valueShape[3];
+
+    if (headDim == torch::Torch::kUnknownSize) {
+      return rewriter.notifyMatchFailure(op, "NYI: dynamic head dimension");
+    }
+
+    // Check if the element type is a float.
+    Type elementType = queryType.getOptionalDtype();
+    auto floatType = dyn_cast<FloatType>(elementType);
```
MaheshRavishankar, line 244: Just a note. I dont know the torch operator here, but if this operation is expected to have a float type only you can add that check to the verifier of the torch op, and avoid checking it here. Just makes the code less verbose. If needed you can assert that it is float type here.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +172,353 @@ struct FftRfftOpConversion
...
+    SmallVector<Value> outputDynSizes;
+    SmallVector<int64_t> outputShape = {batch, numHeads, seqLenQ, valueDim};
+    computeDynamicSizes(rewriter, loc, outputShape, outputDynSizes,
+                        builtinQuery, builtinValue);
+
+    // Initialize output tensor with identity value (0.0 for addition).
+    Value outputInit = arith::getIdentityValue(arith::AtomicRMWKind::addf,
+                                               floatType, rewriter, loc,
+                                               /*useOnlyFiniteValue=*/true);
+    Value outputTensor = tensor::SplatOp::create(rewriter, loc, outputInit,
```
MaheshRavishankar, line 287: I am not sure we create splat operations.  If this is for the output tensor maybe create a `linalg.fill` operation? Thats what the compiler handles best. In any case as long as when you compile the splat gets converted to a fill later on, its a moot point.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +172,353 @@ struct FftRfftOpConversion
...
+  }
+
+  Value convertToBuiltinTensor(PatternRewriter &rewriter, Location loc,
+                               Value torchTensor) const {
+    auto torchType = cast<torch::Torch::ValueTensorType>(torchTensor.getType());
+    return torch::TorchConversion::ToBuiltinTensorOp::create(
+        rewriter, loc, torchType.toBuiltinTensor(), torchTensor);
+  }
+
+  void setTypeInfo(PatternRewriter &rewriter, FloatType floatType) const {
```
MaheshRavishankar, line 360: Nit: I am usually not a proponent of member functions. This seems to be just a helper method. You can move this and others out of the class as utility functions.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +172,353 @@ struct FftRfftOpConversion
...
+                                                        rewriter.getI1Type());
+  }
+
+  void computeDynamicSizes(PatternRewriter &rewriter, Location loc,
+                           const SmallVector<int64_t> &shape,
+                           SmallVector<Value> &dynSizes, Value first,
+                           Value second) const {
+    for (int i = 0; i < kAttentionRank; ++i) {
+      if (shape[i] == torch::Torch::kUnknownSize) {
+        Value idx =
```
MaheshRavishankar, line 384: I am not sure I understand what is happening in this method.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -163,6 +172,353 @@ struct FftRfftOpConversion
...
+                           Value zero) const {
+    // Create mask tensor [B, H, M, N] with values 0.0 (attend) or -inf
+    // (mask).
+    SmallVector<int64_t> maskShape = {batch, numHeads, seqLenQ, seqLenKV};
+    SmallVector<Value> maskDynSizes;
+
+    computeDynamicSizes(rewriter, loc, maskShape, maskDynSizes, builtinQuery,
+                        builtinKey);
+
+    Value maskTensor = tensor::EmptyOp::create(rewriter, loc, maskShape,
```
MaheshRavishankar, line 408: Why is the mask float type? Thats a lot of storage for just a boolean?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -923,6 +920,15 @@ def IREELinalgExt_OnlineAttentionOp : IREELinalgExt_Op<"online_attention",
     it over the entire softmax reduction dimension by:
       x, _, sum : results
       x = (1 / sum) * x
+
+    Region:
+      The region body receives the following block arguments:
+        - score: the computed score value from Q @ K.T (element type of output)
```
MaheshRavishankar, line 926: Nit: Can you make this such that the index operands come first and then the score....

Not Nit: The way the attention operation is setup it can increase dimensionality of the operation. So there isnt necessarily one batch dimension, there could be multiple batch dimensions. Same for head/m/n etc. That probably needs to be accounted for in the change. Just having a single batch dimension is not going to work.

MaheshRavishankar (changes requested):
I think this PR needs to be split-up. There is a signficant change to the ext.attention operation that this is trying to do with the change to the block arguments of the region. I think that is a worthwhile change, but we need to make sure this done in a way that meshes well with all the transformations surrounding the operation. As it is done today this isnt done. Lets split this PR into
1. Changes to the ext.attention operation
2. Lowering of flex attention operation to ext.attention operation.

keshavvinayak01:
@MaheshRavishankar Thanks for the review, I'll incorporate some of the suggested changes. 
As for the PR bifurcation, It's a bit difficult to isolate the lowering and the required changes. The lowering relies on block arguments being supported which is a change to the AttentionOp interface itself.
Could you be more specific about how you see this split being done?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: Makes sense, I'll move it to the `::verify` of the torch op itself. I think a lot of checks here can be moved to the verifier.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: I used this helper in two places:

1. For the mask tensor. Shape = (B, Hq, L, S). Any of these may be dynamic, so we extract B/Hq/L from the query tensor and S from the key tensor. The resulting dynamic sizes are passed to `tensor.empty` when materialising the mask.

2. For the output tensor. Shape = (B, Hq, L, Ev). Since Ev is statically known, only B/Hq/L may be dynamic. The helper again generates the needed `tensor.dim` ops from the query/value tensors so that `tensor.splat`/`tensor.empty` gets the correct dynamic extents.

Assuming the standard 4D layout:
```
  Query: (B, Hq, L, E)
  Key:   (B, Hkv, S, E)
  Value: (B, Hkv, S, Ev)
```
When constructing new tensors (mask/output), we need dynamic sizes for dimensions that come from the input shapes.

- For dims (B, H, L), the runtime sizes always come from the query tensor.
- For dim 3, the required runtime size depends on what we are building:
    - For the mask (shape = B×H×L×S), the 3rd axis is S, which lives at
      index 2 of the Key tensor. 
    - For the output (shape = B×H×L×Ev), Ev is statically known, so we never need a dynamic dimension for `i = 3`. 

The helper works fine, albeit it's a bit complicated. I'll add a verbose comment on top of it so it's easier to digest.

MaheshRavishankar:
> @MaheshRavishankar Thanks for the review, I'll incorporate some of the suggested changes. As for the PR bifurcation, It's a bit difficult to isolate the lowering and the required changes. The lowering relies on block arguments being supported which is a change to the AttentionOp interface itself. Could you be more specific about how you see this split being done?

Yeah, first you should make a change to the `AttentionOp` itself which will get the block arguments support. Then you can make the changes to the lowering? The part adding block arguments here needs to be flushed out a bit more. Do we really need the change to get block arguments in? One option is to not change block arguments but use `linalg_ext.index` (similar to `linalg.index). That might be easier to land without too many breakages.

keshavvinayak01:
> > @MaheshRavishankar Thanks for the review, I'll incorporate some of the suggested changes. As for the PR bifurcation, It's a bit difficult to isolate the lowering and the required changes. The lowering relies on block arguments being supported which is a change to the AttentionOp interface itself. Could you be more specific about how you see this split being done?
> 
> Yeah, first you should make a change to the `AttentionOp` itself which will get the block arguments support. Then you can make the changes to the lowering? The part adding block arguments here needs to be flushed out a bit more. Do we really need the change to get block arguments in? One option is to not change block arguments but use `linalg_ext.index` (similar to `linalg.index). That might be easier to land without too many breakages.

Previously, that's how the implementation looked like, I added support for `linalg_ext.index` ops inside the attention region. That worked, but bloated up the IR, so I thought this was a simpler solution.

 In any case, It's easier to support index ops inside the attention region (With small modifications to the `::verify` method), so I'll switch back to it, because it's less likely to break things in the future.

Regardless, I'll have to add tiling support for these index ops in the attention region, which will be a follow up PR.

MaheshRavishankar:
Why would you need to add support for index ops while tiling. You shouldn't need to

keshavvinayak01:
I had discussed this with @Groverkss earlier, the goal was to add support for iree_linalg_ext.index in the attention op region to enable index-based logic (e.g., masking for RPE). At the time, he pointed out that using the `mask` operand in `AttentionOp` was a hack.

While most tiled ops don’t require index in the region, this op does. It relies on index-dependent semantics like relative position encoding, which can’t be expressed correctly without index after tiling.

PR #22446, opened 2025-10-28
[ukernels] Add missing specializations on gfx942/gfx950 and associated e2e tests 

sebvince:
The primary purpose of this PR is to add missing e2e tests to cover all combinations of ukernels. We have different ukernels for static and dynamic shapes, so this PR introduces an option to generate_e2e_matmul_tests.py that allows specifying the dynamicity of m, n, and k.

Since ukernels require specialization to be enabled ([see PR #22425](https://github.com/iree-org/iree/pull/22425)), this PR also adds the following missing specializations :
- bf16 on gf942 
- f16, bf16, f8 on gfx950.

This will positively affect performance on dynamic shape matmuls without compile-time bounds information. For example :

```
  !A_size = tensor<?x4096xbf16>
  !B_size = tensor<4096x4096xbf16>
  !C_size = tensor<?x4096xf32>
  
  func.func @matmul(
  %A : !A_size, %B : !B_size) -> !C_size {
  %m = tensor.dim %A, %c0 : tensor<?x4096xbf16>
  %empty = tensor.empty(%m) : !C_size
  %C = linalg.fill ins(%cst : f32) outs(%empty : !C_size) -> !C_size
  %0 = linalg.matmul 
     indexing_maps = [affine_map<(m, n, k) -> (m, k)>, 
                     affine_map<(m, n, k) -> (n, k)>,// transpose
                     affine_map<(m, n, k) -> (m, n)>]
                     ins(%A, %B : !A_size, !B_size)
                     outs(%C : !C_size) -> !C_size
  return %0 : !C_size
  }
  ```
Before PR: Time (ms): 30.831274
After PR: Time (ms): 0.284123

### DIFF
--- a/tests/e2e/matmul/generate_e2e_matmul_tests.py
+++ b/tests/e2e/matmul/generate_e2e_matmul_tests.py
@@ -246,9 +246,9 @@ def parse_arguments():
         required=False,
     )
     parser.add_argument(
-        "--is_dynamic",
+        "--mnk_dynamicities",
         type=str,
-        help="Custom dynamicity mask for m,n,k. Format: bool,bool,bool (e.g., is_dynamic=True,False,False)",
+        help="Custom dynamicity mask for m,n,k. Format: dynamic|static,dynamic|static,dynamic|static (e.g., --mnk_dynamicities=dynamic,static,static)",
         required=False,
     )
     return parser.parse_args()
@@ -301,7 +301,7 @@ def main(args):
     # Parse custom MNK values if provided
     shapes_id = ShapesId(args.shapes)
     ShapesId.set_custom_mnk(shapes_id, args.mnk)
-    ShapesId.set_dynamicity(shapes_id, args.is_dynamic)
+    ShapesId.set_dynamicity(shapes_id, args.mnk_dynamicities)
 
     (functions, calls) = generate(
         lhs_rhs_type=MatrixElemTypeId(args.lhs_rhs_type),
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: tests/e2e/matmul/generate_e2e_matmul_tests.py [hunk 1]
```diff
@@ -245,6 +245,12 @@ def parse_arguments():
         help="Custom MNK values for CUSTOM_MNK shape. Format: m,n,k (e.g., --mnk=64,128,256)",
         required=False,
     )
+    parser.add_argument(
+        "--is_dynamic",
```
bjacob: Rename this to a flag name that contains the substrict `mnk` so that, just like with the `--mnk` flag, it is clear in what order the flags are passed.  Maybed `--mnk_dynamicities` would also make it clear that this is meant to be used in conjunction with `--mnk` (which the help string should also mention).

bjacob (changes requested):
There is a discrepancy between the PR title which suggests it's a test-only PR, and the PR contents that add new specialization patterns.  I'm not competent to review those, so please get another reviewer for that.

jtuyls (approved):
LGTM. The specialization patterns make sense to me as they are the same as the existing ones, just for different types. If you have any, it would be useful to list experimental result with and without these specialization patterns for the new gfx950 patterns in the PR description.

sebvince:
> LGTM. The specialization patterns make sense to me as they are the same as the existing ones, just for different types. If you have any, it would be useful to list experimental result with and without these specialization patterns for the new gfx950 patterns in the PR description.

After some testing, I found that some f16 & bf16 configurations of pingpong are slower that the default path (independent from this PR). This is due to the fact that current pingpong uses MFMA_F32_16x16x16_F16 instead of MFMA_F32_16x16x32_F16. To keep in mind if we want to enable ukernels by default. 
I've created a separate task to fix this : https://github.com/iree-org/iree/issues/22457

bjacob (approved):

Issue #22448, opened 2025-10-28
[Codegen][ROCm] Non-deterministic crash in `config_igemm_tile_and_fuse.mlir`


kuhar:
This test started crashing recently, the first report comes from windows after the integrate: https://github.com/iree-org/iree/pull/22427#issuecomment-3455717608

I am able to reproduce it on ubuntu 24.04 w/ clang 20:

Input
```mlir
// RUN: iree-opt --mlir-print-local-scope --split-input-file --iree-gpu-test-target=gfx942 \
// RUN: --iree-codegen-llvmgpu-use-igemm=true --iree-codegen-llvmgpu-igemm-pad-convolution=false --pass-pipeline="builtin.module(iree-llvmgpu-select-lowering-strategy)" %s | FileCheck %s --check-prefixes=CHECK,GFX942

// RUN: iree-opt --mlir-print-local-scope --split-input-file --iree-gpu-test-target=mi300x@hip \
// RUN: --iree-codegen-llvmgpu-use-igemm=true --iree-codegen-llvmgpu-igemm-pad-convolution=false --pass-pipeline="builtin.module(iree-llvmgpu-select-lowering-strategy)" %s | FileCheck %s --check-prefixes=CHECK,MI300X

// RUN: iree-opt --mlir-print-local-scope --split-input-file --iree-gpu-test-target=gfx942 \
// RUN: --iree-codegen-llvmgpu-use-igemm=true --iree-codegen-llvmgpu-igemm-pad-convolution=true --pass-pipeline="builtin.module(iree-llvmgpu-select-lowering-strategy)" %s | FileCheck %s --check-prefix=PAD-CONV-GFX942


#map = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1 + d4, d2, d5)>
#map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d3, d4, d5)>
#map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d2, d3)>
module {
  func.func @conv_nhwc_filter_5x1_unaligned(%arg0: tensor<16x42x19x64xbf16>, %arg1: tensor<64x5x64xbf16>, %arg2: tensor<16x38x19x64xf32>) -> tensor<16x38x19x64xf32> {
[...]

PR #22366, opened 2025-10-21
Bump LLVM to llvm/llvm-project@683e2bf

Muzammiluddin-Syed-ECE:
Still carrying these reverts:
- https://github.com/llvm/llvm-project/pull/160615 (See https://github.com/iree-org/iree/issues/22171)
- https://github.com/llvm/llvm-project/pull/163440 (See https://github.com/iree-org/iree/pull/22354#issuecomment-3424225175)

Fixes: 
- Remove deprecated references to `.CaseLowers`
- Added dependencies required due to new added dependency to LLVMIR https://github.com/llvm/llvm-project/pull/163881

PR #22450, opened 2025-10-28
Bump LLVM to llvm/llvm-project@466c526

Yu-Zhewen:
Drop revert on https://github.com/llvm/llvm-project/pull/163994, as the failure on windows_x64_msvc was caused by a different issue (See https://github.com/iree-org/iree/issues/22448)

Still carrying two remaining reverts as https://github.com/iree-org/iree/pull/22366

ci-extra: test_torch


kuhar (approved):

Groverkss (approved):
pkgci failures are from ossci machine problems which are being tracked

PR #22454, opened 2025-10-28
[LLVMGPU] Fix coding standards / style issues in config utils

kuhar:
Honestly, too many issues to list in the PR description.

Not an NFC because some things like the wrong accumulator type in `std::reduce` might be actual bugs that we haven't hit yet.

ci-extra: windows_x64_msvc

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -63,7 +63,7 @@ LogicalResult setDataTiledMmaInnerTiledLoweringConfig(
   // single subgroup.
   const int64_t targetSubgroupSize = dataTiledMmaAttr.getSubgroupSize();
   int64_t flatWorkgroupSize = dataTiledMmaAttr.getFlatWorkgroupSize();
-  std::array<int64_t, 3> workgroupSize{flatWorkgroupSize, 1, 1};
+  std::array<int64_t, 3> workgroupSize = {flatWorkgroupSize, 1, 1};
 
   // Set all workgroup and reduction tile sizes to 1, since the data tiled
   // kernel has the scope of an entire workgroup, and the reduction tiling is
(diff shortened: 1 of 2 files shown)


hanhanW:
I don't see any change related to `std::accumulate`. Is it expected?

I think we hit an issue about accumulator type before. E.g., https://github.com/iree-org/iree/commit/2034585639738d4dbec6d12ac82fa9712dda9b4d

hanhanW:
> I don't see any change related to `std::accumulate`. Is it expected?
> 
> I think we hit an issue about accumulator type before. E.g., [2034585](https://github.com/iree-org/iree/commit/2034585639738d4dbec6d12ac82fa9712dda9b4d)

https://en.cppreference.com/w/cpp/algorithm/accumulate.html shows that it is implementation details; we need to use proper type for init value.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -158,13 +156,13 @@ static std::optional<ComputeBitwidths> getComputeBitwidthForType(Type type) {
               return std::nullopt;
             }
           })
-      .Default([](Type) { return std::nullopt; });
```
Groverkss, line 161: Returning std::nullopt is fine right?

Groverkss (approved):
LGTM

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
kuhar: Doesn't compile

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1360,15 +1359,13 @@ LogicalResult setTileAndFuseLoweringConfig(IREE::GPU::TargetAttr target,
 
     // Threads for distribution. Use `minPreferredNumThreads` at least, but no
     // more than 4 subgroups.
-    int64_t minPreferredNumThreads = std::reduce(
-        workgroupTileSizeMultiples.begin(), workgroupTileSizeMultiples.end(), 1,
```
kuhar, line 1364: @hanhanW here the init value is wrong: was supposed to be int64 but we used plain int. The init value type decides the accumulator type.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Groverkss: oh makes sense

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -604,24 +602,25 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   // or not. Also when this is a part of the heuristic it will be easier
   // to take into account the element type instead of the constants
   // 128 and 32 that were derived for bf16 type with f32 accumulation.
-  auto getDimBounds = [&](SmallVector<int64_t> dims,
-                          bool PaddingCanBeExpensive) -> SmallVector<int64_t> {
+  auto getDimBounds = [&](ArrayRef<int64_t> dims,
```
kuhar, line 605: FYI, @nirvedhmeshram

Issue #8469, opened 2022-03-05
Decompose cpuinfo usage and plumb through new information.


benvanik:
There are two things cpuinfo is used for: architectural features and topology. #5417 adds the executable processor info support but we still need something we can use for the task system topology logic.

The problem is that cpuinfo information is very spotty across platforms - though it has an enticing API the information is unreliably available (in some cases not linking, in others just returning 0's at runtime). This makes it hard to author logic that uses this information (#4654 related) and liters our code with [workarounds](https://github.com/google/iree/blob/dd02da552a858f1aa0563a8473504c047a4c61d9/iree/task/topology_cpuinfo.c#L35), carveouts for emscripten, bare-metal special casing, etc. The flaw here is trying to normalize all possible information from all platforms and all architectures under a single API. This isn't the problem we have though: we need a few bits of information to make some basic decisions and can that logic needs to be robust to the platforms/architectures it's designed for.

We can distill what topology information we need into a two major categories: node/core/processor hierarchy and cache information for each so that we can create task system workers and measure the amount of work we put into them. [iree/task/topology.h](https://github.com/google/iree/blob/dd02da552a858f1aa0563a8473504c047a4c61d9/iree/task/topology.h) already has the representation we need for the hierarchy within a node and should be extended to include the cache information in distilled form.

Ruy does a good job of explaining and condensing the required information to the same set: https://sourcegraph.com/github.com/google/ruy/-/blob/ruy/cpu_cache_params.h?L72-79
Basically, last-level cache (LLC) for the shared cache immediately prior to going off-chip and the exclusive cache for the cache unique to each core (though possibly shared with multiple logical processors) without interference from other cores.

We do want to support NUMA and heterogenous processor complexes - this means that we still need to associate cache information with the topology we query. How this is represented and chosen varies wildly across platforms and environments - often not being available at all - and instead of trying to normalize the API it'll probably make sense to just add `iree/task/topology_win32.c` + `iree/task/topology_darwin.c` + etc as we have in other places. This lets us keep the API as fat as possible based around the `iree_task_topology_t` and hide the details of how things are decided to the place where that information is (sometimes) present.

Major tasks:
* [ ] Add exclusive cache information to `iree_task_topology_group_t` (pull from cpuinfo to start)
* [ ] Add last-level cache information to `iree_task_topology_t` (pull from cpuinfo to start)
* [ ] Plumb cache information to `iree_hal_processor_info_v0_t` (#5417) so executables can use it
[...]

PR #22455, opened 2025-10-28
Adding sysfs topology detection logic and switching to it by default.

benvanik:
This avoids using cpuinfo just to get this standardized information and will let us drop that as a dependency. We'll let this soak for a bit to see if anyone notices before finally removing cpuinfo entirely. For now `-DIREE_ENABLE_CPUINFO=ON` can be used to override and cpuinfo will be fetched (or the system dep will be used).

This is not meant to be the end-all be-all of topology information but is more featureful than cpuinfo for what we needed and can be easily updated. Testing is going to be its own thing so I have a way to manually do it for now. Real testing will require a corpus of sysfs info in json/yaml that we plug in to our file I/O as a virtual filesystem - otherwise these sysfs trees grow to 10-20MB. Since we're using standard sysfs information (no tables/etc builtin) we should be fairly robust outside of kernel versioning (which are hopefully handled well enough 🤞) and our own logic, and if we go heavy on CPU workloads again we can invest in tuning that (everything today is best-guess only).

Progress on #8469.


compiler/setup.py
```diff
@@ -20,7 +20,7 @@
 # This can be set with the IREE_COMPILER_API_CMAKE_BUILD_DIR env var.
 #
 # Select CMake options are available from environment variables:
-#   IREE_ENABLE_CPUINFO
+#   IREE_ENABLE_LLD
```
bjacob: I can see a number of environment variable queries in this file. Why single this one out here?

runtime/src/iree/task/testdata/sysfs/README.md
```diff
@@ -0,0 +1,131 @@
...
+   - ANY mode includes all cores
+   - HIGH mode filters to high-capacity cores only
+   - LOW mode filters to low-capacity cores only
+5. **Graceful degradation:** Missing files handled without errors
+
+## Adding New Test Cases
+
+Good candidates for test data to capture locally (but only check in small/unusual ones):
+
+- **NUMA systems:** Multi-socket x86 servers (Xeon, EPYC) - capture locally, don't check in
```
bjacob, line 121: It's uncomfortable that the configurations we care the most about are also the ones we can't check in. Is there a way to trim those configurations to a subset that is small enough to be checked in but still contains the relevant information?  I mean, do we actually make use of the whole 10 MB of text ?

runtime/src/iree/task/api.c
```diff
@@ -107,6 +107,23 @@ IREE_FLAG(string, task_topology_performance_level, "any",
...
+IREE_FLAG(
+    string, task_topology_distribution, "scatter",
+    "Strategy for distributing cores across cache domains (CCXs) within the\n"
+    "selected NUMA node(s) (use --task_topology_nodes to control NUMA "
+    "locality):\n"
+    "  `compact` - Fill cache domains sequentially (better for compute).\n"
+    "  `scatter` - Round-robin across domains (better for memory bandwidth).");
+
+IREE_FLAG(
+    string, task_topology_favor, "",
```
bjacob, line 119: I had difficulty parsing this because as `favor` is a suffix here, I tried parsing it as a noun instead of a verb. Then I mentally parsed it as `flavor`.

Replace `favor` by another term that is unambiguously a verb ? E.g. `prefer` or `optimize` ?

runtime/src/iree/task/api.c
```diff
@@ -107,6 +107,23 @@ IREE_FLAG(string, task_topology_performance_level, "any",
...
+    "High-level preset for common deployment scenarios (overrides "
+    "distribution\n"
+    "and performance_level if specified):\n"
+    "  `latency` - Minimize single-request latency (compact + high-perf).\n"
+    "  `throughput` - Maximize batch throughput (scatter + any-perf).\n"
+    "  `efficiency` - Minimize power consumption (compact + low-perf).");
```
bjacob: Difficulty interpreting "perf" here.  IIUC this refers to sets of cores: either high-power or low-power cores. Clarify?  Every core is a "performant" core for some performance metric.

runtime/src/iree/task/topology_sysfs.c
```diff
@@ -0,0 +1,981 @@
...
+  uint64_t scale = 1;
+  if (iree_string_view_consume_suffix(&text, IREE_SV("K")) ||
+      iree_string_view_consume_suffix(&text, IREE_SV("k"))) {
+    scale = 1024;
+  } else if (iree_string_view_consume_suffix(&text, IREE_SV("M")) ||
+             iree_string_view_consume_suffix(&text, IREE_SV("m"))) {
+    scale = 1024 * 1024;
+  } else if (iree_string_view_consume_suffix(&text, IREE_SV("G")) ||
+             iree_string_view_consume_suffix(&text, IREE_SV("g"))) {
+    scale = 1024ULL * 1024ULL * 1024ULL;
```
bjacob: At some point, bit shifts are your friends :-)

runtime/src/iree/task/topology_sysfs.c
```diff
@@ -0,0 +1,981 @@
...
+  IREE_ASSERT_ARGUMENT(out_size);
+  *out_size = 0;
+
+  text = iree_string_view_trim(text);
+  if (iree_string_view_is_empty(text)) {
+    return iree_make_status(IREE_STATUS_INVALID_ARGUMENT,
+                            "size string is empty");
+  }
+
+  // Check for optional suffix (K/M/G, case-insensitive).
```
bjacob: I haven't seen another suffix than `K` ?

runtime/src/iree/task/topology_sysfs.c
```diff
@@ -0,0 +1,981 @@
...
+// Returns IREE_STATUS_NOT_FOUND if the cache index doesn't exist.
+static iree_status_t iree_sysfs_query_cache_level(
+    uint32_t processor, uint32_t cache_index,
+    iree_sysfs_cache_info_t* out_cache) {
+  // Read cache type (Data, Instruction, or Unified).
+  // If this fails the cache index doesn't exist.
+  char path[256];
+  snprintf(path, sizeof(path), "%s/cpu/cpu%u/cache/index%u/type",
+           iree_sysfs_get_root_path(), processor, cache_index);
+  char buffer[64];
```
bjacob, line 409: Eeergh stack buffers, particularly the second one for file I/O... If it were me, I would allocate `buffer` on the heap. In case there's ever a bug in a future implementation of `iree_sysfs_read_small_file` or even in the Linux kernel driver for `/sys` causing a buffer overrun, I'd rather have it be a heap buffer overrun.  Regarding `path`, it's easier to live with (and a common idiom that I've done too) because you're only trusting `snprintf`.

runtime/src/iree/task/testdata/sysfs/README.md
benvanik: Yeah, that's how sysfs works unfortunately when trying to do this file-based emulation approach: lots of tiny 4 byte files that expand to full pages, and when you have 192 cores you end up with a lot of little files. As noted in the PR description the right way is to not check in the files (gzip'ed or bare) and instead have a virtual sysfs that we can feed with more compact JSON/YAML. This is just a smoke test for now until that's built out, where we can have it capture into JSON/YAML instead.

compiler/setup.py
benvanik: 🤷 was weird to ever have this section to start with, given that these things always get out of sync - just removed it.

runtime/src/iree/task/topology_sysfs.c
benvanik: Good catch - sysfs is a mess with this (some nodes are bytes, some are suffixed KB, others are unsuffixed KB, etc) - I was being defensive, but scanning the kernel code it's only ever those 3 so I simplified here and added a note.

runtime/src/iree/task/topology_sysfs.c
benvanik: fread with buffer_size - 1 should never have a bug - if it does, we (and linux/libc) have bigger issues :)
Since what we read are several byte values the risk of introducing bugs juggling heap memory are higher than our one place where we call fread. Added some comments/asserts to clarify that the read function is correct.

bjacob: Here you're trusting that future IREE contributors preserve the invariant that iree_sysfs_read_small_file  does fread with buffer_size - 1 .

bjacob: And re Linux -- agree on the *contractual* behavior, but the *effective* behavior of fread is entirely up to each filesystem driver, meaning that one filesystem could have a bug without "linux/libc having bigger issues".

The fs driver writes *directly* to the user buffer.

runtime/src/iree/task/api.c
benvanik: Intel/Apple use "performance" and "efficiency" core naming (Qualcomm also has "prime" cores): P-cores, E-cores. It's arm big.LITTLE but with a less silly name.

As for the comment, it's describing what enum it is setting:
"IREE_TASK_TOPOLOGY_PERFORMANCE_LEVEL_ANY" = "any-perf"
"IREE_TASK_TOPOLOGY_PERFORMANCE_LEVEL_LOW" = "low-perf"
"IREE_TASK_TOPOLOGY_PERFORMANCE_LEVEL_HIGH" = "high-perf"

These map to the `any`/`low`/`high` values for `--task_topology_performance_level=`.

```
// Selects what core types in a heterogeneous core cluster are used.
// This maps to x86 efficiency/performance cores and ARM big.LITTLE cores.
//
// Hosting applications can decide whether they want low power consumption/less
// contention on high performance cores by forcing only low performance cores
// or predictable(ish) low latency by forcing only high performance cores. On
// homogeneous core clusters, where wall-time is the primary metric, or where
// contention is unlikely selecting all cores can usually result in the lowest
// latency. Each application with each set of programs will need to evaluate for
// themselves what to use based on their duty cycle, concurrently issued work,
// and user experience.
typedef enum iree_task_topology_performance_level_e {
  // Selects all cores.
  IREE_TASK_TOPOLOGY_PERFORMANCE_LEVEL_ANY = 0,
  // Selects "E(fficiency)" cores that favor lower power/thermal load.
  IREE_TASK_TOPOLOGY_PERFORMANCE_LEVEL_LOW,
  // Selects "P(erformance)" cores that favor higher power/thermal load.
  IREE_TASK_TOPOLOGY_PERFORMANCE_LEVEL_HIGH,
} iree_task_topology_performance_level_t;
```

Note that there _is_ a relative metric for core performance, and that's what we use on ARM to detect big.LITTLE :)
https://docs.kernel.org/6.17/scheduler/sched-capacity.html
`/cpu_capacity` gives us that, and more performant cores have higher capacities. It's what lets us avoid needing to cpuid and have big tables like cpuinfo for this purpose. Intel doesn't use that yet, unfortunately, and has a different CPU core tree (cpu_core/ vs cpu_atom/) that's a pain to deal with so I ignore them for now on Linux (on Windows we handle ARM/X86 the same as Windows has a standard API for them, and on Mac we sysctl `hw.nperflevels`).

runtime/src/iree/task/topology_sysfs.c
bjacob: >  risk of introducing bugs juggling heap memory are higher than our one place where we call fread.

This thread a really good sales pitch for C++ over C :-)

runtime/src/iree/task/api.c
benvanik: 👍 "Favor" is a term I use in the codebase (`--iree-stream-partitioning-favor=` / `IREE::Stream::FavorAttr`) so I'll stick with that. `--task_topology_*` is the flag namespace so should be (mentally) parsed separately. The flag is intended to be read as a whole: '--task_topology_favor=latency` = "task topology selection should favor latency-focused workloads", '--task_topology_favor=throughput` = "task topology selection should favor throughput-focused workloads", etc.

runtime/src/iree/task/topology_sysfs.c
benvanik: If anyone ever tries to contribute a meaningful change to one of these files I'd be elated ;)
It's about risk: the single compilation-unit local fread helper function in this file used exclusively for reading sysfs contents for use in the file is low risk. It's not a public API, or even an internal one.

bjacob (approved):
All good!

PR #22456, opened 2025-10-29
Migrate custom LDBG macro to LLVM’s built-in debug logging

Yu-Zhewen:
This change removes IREE’s local LDBG macro definition in favor of the LLVM-provided version introduced in https://github.com/llvm/llvm-project/pull/143704.

This update is also in preparation for https://github.com/llvm/llvm-project/pull/165429, which would otherwise cause a macro redefinition error due to the existing local LDBG definition.


Groverkss (approved):

bjacob (approved):

hanhanW (approved):

kuhar (approved):
Thanks

Max191 (approved):

PR #22458, opened 2025-10-29
[ci] Switch to new RDNA3 runners

Groverkss:
The old w7900 runner has lots of problems and has been down consistently for a while now. This PR switches to a newly built 7900XTX. Both machines have the same architecture (gfx1100), but have different skus.


hanhanW (approved):

kuhar (changes requested):
Turns out we may have some tests that require 2 gpus: `iree-test-suites/sharktank_models/llama3.1/test_llama.py::test_decode[hip_tp2]`, but shark55-ci only has a single GPU.

MaheshRavishankar (approved):

PR #22459, opened 2025-10-29
[Codegen] Fix more coding style / standards issues

kuhar:


compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
```diff
@@ -174,9 +172,7 @@ static TileSwizzle getIntrinsicSwizzle(MMAIntrinsicTy intrinsic,
   }
   // Next come `layout.thread` dims.
   int64_t subgroupSize = getIntrinsicSubgroupSize(intrinsic);
-  int64_t numThreadsInLayout =
-      std::reduce(layout.thread.begin(), layout.thread.end(), 1LL,
-                  std::multiplies<int64_t>());
+  int64_t numThreadsInLayout = llvm::product_of(layout.thread);
```
Max191, line 175: Wow, I had no idea `llvm::product_of` existed, nice!

Max191 (approved):
Thanks for doing all these cleanups!

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
hanhanW: It was added recently, IIRC.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -416,12 +416,11 @@ getMinTilingSizesForEachDim(mlir::FunctionOpInterface entryPointFn,
       if (minTileSizes[unrollDim] <= 1) {
         continue;
       }
-      int64_t factor = seen ? 1LL : maxUnrollFactor;
+      int64_t factor = seen ? 1 : maxUnrollFactor;
```
hanhanW, line 419: Why is 1 better than 1LL in this case?

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
```diff
@@ -23,7 +21,7 @@ using Kind = TileSwizzle::Dim::Kind;
 
 SmallVector<int64_t>
 sliceSwizzledShape(const TileSwizzle &swizzle,
-                   const std::function<bool(TileSwizzle::Dim)> &predicate) {
+                   llvm::function_ref<bool(TileSwizzle::Dim)> predicate) {
```
hanhanW, line 24: Why do we prefer llvm::function_ref?

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
kuhar: it doesn't matter because standard common type promotion rules apply and LL is not necessarily even int64_t

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
kuhar: Because it's guaranteed not to allocate. For function refs that are only called but never stored, we don't have to worry about lifetimes.

kuhar: Similar to StringRef vs. std::string

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
kuhar: Almost every use of `std::accumulate` and `std::reduce` is incorrect because the accumulator type is determined by the initial value passed. `llvm::product_of` and `llvm::sum_of` don't have this issue.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
hanhanW: I see, thanks!

hanhanW (approved):

PR #22460, opened 2025-10-29
Remove value bounds interface for ExpandShapeOp

jtuyls:
This is included in upstream now, so can be removed: https://github.com/iree-org/llvm-project/blob/90a67589b3e15f4375df85dea64cf0d3940a4d5f/mlir/lib/Dialect/MemRef/IR/ValueBoundsOpInterfaceImpl.cpp#L62


MaheshRavishankar (approved):

hanhanW (approved):
We no longer need https://github.com/iree-org/iree/pull/22435, right?

PR #22461, opened 2025-10-29
[Dispatch Creation] Don't add unfusable consumers to fusion group

IanWood1:
There are cases where two ops can't be fused as it would result in use-def violation (see the diagram below). Currently, they are still placed in the same fusion group. This means that the consumer gets marked by the analysis as being in a fusion group causing it to potentially miss out on fusion opportunities. This change moves this check so that it occurs during the analysis and modifies the old check to error out.

```
    A (in fusion group)
    |  \
    |   \
    |    v
    |    B (unfusable consumer of A)
    |   /
    |  /
    v v
    C (trying to fuse with A)
```

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
@@ -480,6 +480,9 @@ getFusableUses(MLIRContext *context, Operation *op,
     if (isa<tensor::DimOp>(user)) {
       continue;
     }
+    if (op->getBlock() != user->getBlock()) {
+      continue;
+    }
     fusableUses.insert(&use);
   }
 
--- a/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
@@ -20,6 +20,7 @@
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Debug.h"
+#include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Analysis/TopologicalSortUtils.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
@@ -43,6 +44,7 @@
 #include "mlir/Pass/Pass.h"
 #include "mlir/Support/LLVM.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/RegionUtils.h"
 
 #define DEBUG_TYPE "iree-dispatch-creation-form-dispatch-regions"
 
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
```diff
@@ -1391,15 +1391,16 @@ util.func public @avoid_illegal_consumer_fusion(%arg0: tensor<75600x5120xbf16>)
...
 //       CHECK:     %[[GENERIC0:.+]] = linalg.generic
 //       CHECK:     %[[GENERIC1:.+]] = linalg.generic
 //  CHECK-SAME:         ins(%[[GENERIC0]] :
 //       CHECK:     flow.return %[[GENERIC1]], %[[GENERIC0]]
-//       CHECK:   %[[EXPAND_SHAPE:.+]] = tensor.expand_shape %[[DISPATCH]]#1
+//       CHECK:   %[[EXPAND_SHAPE:.+]] = tensor.expand_shape %[[DISPATCH0]]#1
+//       CHECK:   %[[DISPATCH1:.+]] = flow.dispatch.region
 //       CHECK:   %[[GENERIC2:.+]] = linalg.generic
-//  CHECK-SAME:       ins(%[[EXPAND_SHAPE]], %[[DISPATCH]]#0 :
-//       CHECK:   util.return %[[GENERIC2]]
```
IanWood1, line 1402: Note for reviewers: this checks that `hasTransitiveDependencyOnFusionGroup` prevents the fusion since the old check will now error out.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -145,6 +147,49 @@ class FusionGroup {
...
   void insert(Operation *op);
 
+  /// Returns true if `consumerOp` has a transitive dependency on the fusion
+  /// group. This means that some transitive dependency of `consumerOp` (not in
+  /// the fusion group) itself uses an operation in the fusion group. This is
+  /// required for fusion because it must be legal to take a program slice that
+  /// contains only the ops in the fusion group.
+  bool hasTransitiveDependencyOnFusionGroup(Operation *consumerOp) const {
+    BackwardSliceOptions options;
+    DominanceInfo dominance(consumerOp);
```
MaheshRavishankar: Does the Dominator need to be put in here? I prefer to pass the dominator in as an operand.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -974,7 +1028,7 @@ createFusionGroups(TensorDimTrackingRewriter &rewriter,
       auto newRegionOp = IREE::Flow::moveFollowingOpIntoDispatchRegion(
           rewriter, consumer, regionOp);
       if (failed(newRegionOp)) {
-        continue;
+        return consumer->emitOpError("failed to move consumer into region");
```
MaheshRavishankar, line 1034: Does this need to error out? Can it just continue as before to not break cases that might slip through above?

MaheshRavishankar (changes requested):

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
IanWood1: I wanted this to error out because with these changes it should never error out. I'm trying to enforce the invariant that all ops in `FusionGroup` can be legally fused.


Maybe a warning would be better here?

MaheshRavishankar: Yeah, warning is fine. Error will stop the world when we dont really need to. You can make it a warning + assert as well.

MaheshRavishankar (approved):
Looks good after removing the emitError in favor of warning/assert.

PR #22464, opened 2025-10-29
Fix e2e matmul mxfp4 tests on gfx950 post #22446

bjacob:
https://github.com/iree-org/iree/pull/22446 regressed building those tests, which are gfx950-only.

https://discord.com/channels/689900678990135345/689957613152239638/1433170968620175411


jtuyls (approved):
Thanks! Apologies for not catching it either. We should really enable gfx950 tests.

sebvince (approved):

PR #22465, opened 2025-10-29
Harmonize `*ScaledMMAAttr` operand order and drop `MMAFragment`

bjacob:
This PR does multiple things that were easier done all at once:
1. Harmonize `*ScaledMMAAttr` operand order:
   * The operand order of `ScaledMMAAttr` was `lhs, lhs_scale, rhs, rhs_scale`, while the operand order of `DataTiledScaledMMAAttr` was `lhs, rhs, lhs_scale, rhs_scale`.
   * This PR changes `ScaledMMAAttr` to match the `DataTiledScaledMMAAttr` convention. This propagates to a change of operand order in the enclosing `inner_tiled` ops.
2. Drop `MMAFragment`:
   * There used to be a TableGen enum `MMAFragment`, that had unclear semantics: the enum values were sometimes used as opaque symbolic enums to refer to operand by "role", e.g. "Lhs", and sometimes used as the underlying integer values as operand indices, e.g. "Rhs == 1". This was originally OK as all MMA-like ops had the same 3 operands Lhs, Rhs, Acc. But when ScaledMMAAttr was introduced, that didn't... scale: now the preexisting enum value "Rhs == 1" didn't equal anymore the corresponding operand index under the `lhs, lhs_scale, rhs, rhs_scale` convention (1 != 2) and even regardless of convention, the enum value "Acc ==2" never corresponded to operand index anymore (2 != 4).
   * This PR drops MMAFragment and generalizes the use of plain integer `operandIndex`.
   * This is made reasonable to implement by the harmonization of operand orders (above 1.). If we tried doing this without 1., then we would feel more of a need for opaque "role" enums to replace MMAFragment without having to fix up operand indices between two conventions.
3. Drop some logic that only existed to make up for the discrepancy between convensions and the fuzzy semantics of MMAFragment.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
@@ -101,39 +101,13 @@ static void interleave(TileSwizzle &swizzle, size_t srcIdx, int expandedIdx) {
 template <typename MMAIntrinsicTy>
 static TileSwizzle getIntrinsicSwizzle(MMAIntrinsicTy intrinsic,
                                        unsigned operandIdx) {
-  IREE::GPU::MMASingleSubgroupLayout layout;
+  IREE::GPU::MMASingleSubgroupLayout layout =
+      IREE::GPU::getSingleSubgroupLayout(intrinsic, operandIdx);
   const bool isScaled =
       std::is_same<MMAIntrinsicTy, IREE::GPU::ScaledMMAIntrinsic>::value;
-  const unsigned lhsIdx = 0;
-  const unsigned rhsIdx = 1;
-  const unsigned lhsScalesIdx = 2;
-  const unsigned rhsScalesIdx = 3;
-  const bool isLHSorRHS = operandIdx == lhsIdx || operandIdx == rhsIdx;
-  if (isScaled) {
-    // The operand mapping for `getSingleSubgroupLayout` follows a different
-    // operand order than is used for TileSwizzle, so we need to remap the
-    // operandIdx to get the right layout. The layouts for TileSwizzle vs.
-    // `getSingleSubgroupLayout` are shown below:
-    //             | TileSwizzle | getSingleSubgroupLayout
-    //         LHS | 0           | 0
-    //         RHS | 1           | 2
-    //  LHS Scales | 2           | 1
-    //  RHS Scales | 3           | 3
-    //         ACC | 4           | 4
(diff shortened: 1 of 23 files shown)


kuhar:
Very harmonious, this PR restores tranquility.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
```diff
@@ -233,16 +207,14 @@ static size_t getInnermostNonInternalDimIdx(
...
-  const bool isRhsScales = isScaled && operandIdx == rhsScalesIdx;
-  const bool isLhsScales = isScaled && operandIdx == lhsScalesIdx;
-  if (operandIdx == lhsIdx || isLhsScales) {
+  using MMAIntrinsicTy = decltype(mma.getIntrinsic());
+  const bool isScaled = std::is_same<MMAIntrinsicTy, ScaledMMAIntrinsic>::value;
+  const bool isLhs = isIntrinsicLhs<MMAIntrinsicTy>(operandIdx);
+  const bool isRhs = isIntrinsicRhs<MMAIntrinsicTy>(operandIdx);
+  const bool isAcc = isIntrinsicAcc<MMAIntrinsicTy>(operandIdx);
+  const bool isLhsScale = isIntrinsicLhsScale<MMAIntrinsicTy>(operandIdx);
+  const bool isRhsScale = isIntrinsicRhsScale<MMAIntrinsicTy>(operandIdx);
```
hanhanW, line 216: It looks much better than magic numbers to me.

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp
```diff
@@ -1096,27 +1098,16 @@ FailureOr<IREE::Codegen::InnerTiledOp> convertScaledContractionToInnerTiledMma(
...
-          eltTypes[2] ||
-      cast<RankedTensorType>(inputs[4].getType()).getElementType() !=
-          eltTypes[4]) {
-    return failure();
+  for (int i :
+       {kScaledMMAOperandLhs, kScaledMMAOperandRhs, kScaledMMAOperandAcc}) {
+    if (cast<RankedTensorType>(inputs[i].getType()).getElementType() !=
+        eltTypes[i]) {
+      return failure();
+    }
```
hanhanW, line 1110: It looks much better to me!

hanhanW (approved):
I did not participate in the discussion about what the order should be, as there are reordering changes in lit tests. The change itself looks good to me.

krzysz00 (approved):
LGTM

PR #22466, opened 2025-10-29
[DispatchCreation] Set split reduction size for ArgCompare

bangtianliu:

Debugging the arg_compare op compilation issue in llama 8b fp16 quality tests from PR #22379. 

### Problem:

The issue is with this `arg_compare` operation: 
```mlir
    %29:2 = iree_linalg_ext.arg_compare dimension(2) ins(%28 : tensor<4x1x128256xf16>) outs(%11, %12 : tensor<4x1xf16>, tensor<4x1xi32>) {
    ^bb0(%arg5: f16, %arg6: f16):
      %31 = arith.cmpf ogt, %arg5, %arg6 : f16
      iree_linalg_ext.yield %31 : i1
    } -> tensor<4x1xf16>, tensor<4x1xi32>
```
Without split-reduction support, the entire reduction along dimension 2 (128,256 elements) is processed along `LLVMGPUDistribute` pipeline , which attempts to allocate the entire reduction dimension to shared memory. This exceeds the 64KB shared memory limit and causes compilation failures.

### Solution
This PR extends the `SetSplitReductionSizesPass` to support splitting large reduction dimensions in `iree_linalg_ext.arg_compare` operations.

Another solution is to support arg_compare op along VectorDistribute pipeline (on my plate), which will takes longer time to implement. Therefore, let's solve the problem from dispatch creation stage first as an immediate solution.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -476,22 +476,17 @@ private:
     ArrayRef<int64_t> inputShape = inputType.getShape();
     int64_t reductionDim = argCompareOp.getDimension();
     int64_t reductionSize = inputShape[reductionDim];
-
-    if (reductionSize == ShapedType::kDynamic) {
+    if (ShapedType::isDynamic(reductionSize)) {
       return std::nullopt;
     }
-
     if (reductionSize < splitReductionTargetSize) {
       return std::nullopt;
     }
-
     int64_t tileSize = findSmallestFactorWithLowerBound(
                            reductionSize, splitReductionTargetSize)
                            .value_or(reductionSize);
-
     LDBG() << "arg_compare split: dim=" << reductionDim
            << " size=" << reductionSize << " tile=" << tileSize;
-
     return SmallVector<int64_t>{tileSize};
   }
 };
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -6,6 +6,7 @@
 
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtDialect.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtInterfaces.h"
+#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
 #include "iree/compiler/DispatchCreation/Passes.h"
 #include "llvm/Support/DebugLog.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
@@ -107,6 +108,12 @@ struct SetSplitReductionSizesPass final
 
       // --- Case 3: Matmul-like operations ---
       if (auto tileSizes = getMatmulLikeReductionSizes(tilingOp)) {
+        IREE::LinalgExt::setSplitReductionAttribute(tilingOp, *tileSizes);
+        return;
+      }
+
+      // --- Case 4: arg_compare operations ---
+      if (auto tileSizes = getArgCompareReductionSizes(tilingOp)) {
         IREE::LinalgExt::setSplitReductionAttribute(tilingOp, *tileSizes);
         return;
       }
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/test/set_split_reduction_sizes_outer_reduction.mlir
```diff
@@ -179,3 +179,24 @@ util.func public @arg_compare_negative_outer_dynamic_reduction(
 
   util.return %res_val, %res_idx : tensor<64xf32>, tensor<64xindex>
 }
+
+// -----
+
+// CHECK-LABEL: @arg_compare_large_inner_reduction
+util.func public @arg_compare_large_inner_reduction(%arg0: tensor<4x1x128256xf16>)
+    -> tensor<4x1xi32> {
+  // CHECK: iree_linalg_ext.split_reduction = [1336 : index]
```
kuhar, line 188: Maybe it would be also worth adding a test case for when split reduction is not applied because the reduction dimension is below the threshold?

bangtianliu: Sure, will do!

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -455,6 +462,41 @@ struct SetSplitReductionSizesPass final
...
+    int64_t reductionDim = argCompareOp.getDimension();
+    int64_t reductionSize = inputShape[reductionDim];
+
+    if (reductionSize == ShapedType::kDynamic) {
+      return std::nullopt;
+    }
+
+    // This threshold is set heuristically without extensive empirical data.
+    // It may be refined in the future as more practical cases are encountered.
+    const int64_t minSizeToSplit = 1024;
```
hanhanW: I don't follow split reduction work for a while, but why not make it configurable from CLI?

The other question, that may be resolved if we make it configurable, is that these methods look more like static methods to me. I don't know why we put them within the class.

bangtianliu: Thanks for the feedback!
1. On making minSizeToSplit configurable: I just noticed we can simply reuse the existing `splitReductionTargetSize` option for this case. 
2. On refactoring to static methods: I agree with you. I can address this in a separate PR to keep the scope of this change focused.

hanhanW: You may want different option for different ops, but we can go with this for now. It is easy to extend.

> On refactoring to static methods: I agree with you. I can address this in a separate PR to keep the scope of this change focused.

My guess is that the original author wanted to use splitReductionTargetSize in the method, but we can pass the value to method as well. I'm +1 on moving them to static local methods. And yeah, let's address it in a separate PR.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -455,6 +462,38 @@ struct SetSplitReductionSizesPass final
...
+    }
+
+    ShapedType inputType = argCompareOp.getInputType();
+    ArrayRef<int64_t> inputShape = inputType.getShape();
+    int64_t reductionDim = argCompareOp.getDimension();
+    int64_t reductionSize = inputShape[reductionDim];
+
+    if (reductionSize == ShapedType::kDynamic) {
+      return std::nullopt;
+    }
```
hanhanW: Prefer `ShapedType::isDynamic(reductionSize)`. We also need a test for dynamic case.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -455,6 +462,38 @@ struct SetSplitReductionSizesPass final
...
+
+    int64_t tileSize = findSmallestFactorWithLowerBound(
+                           reductionSize, splitReductionTargetSize)
+                           .value_or(reductionSize);
+
+    LDBG() << "arg_compare split: dim=" << reductionDim
+           << " size=" << reductionSize << " tile=" << tileSize;
+
+    return SmallVector<int64_t>{tileSize};
+  }
```
hanhanW, line 491: optional: I'd drop blank lines because this is not a large function and they belong to the same block to me.

>
> Do not add blank lines where indentation already provides clear delineation, such as at the start or end of a code block. Do use blank lines to separate code into closely related chunks, analogous to paragraph breaks in prose. Within a statement or declaration, usually only insert line breaks to stay within the [line length limit](https://google.github.io/styleguide/cppguide.html#Line_Length), or to attach a comment to only part of the contents.

hanhanW (changes requested):
looks good, just few nits.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
bangtianliu: Sure, thanks for the link.

hanhanW (approved):
thanks!

kuhar (approved):

MaheshRavishankar (commented):
Thanks @bangtianliu . Is this turned on by default?

bangtianliu:
> Thanks @bangtianliu . Is this turned on by default?

No, I think it still depends on `--iree-dispatch-creation-enable-split-reduction`.
(Please correct me if I’m wrong, CC @yzhang93.)

My last PR addressing the race condition is here: https://github.com/iree-org/iree/pull/22694.
Once all related PRs land, I believe we can enable it by default.

yzhang93:
> No, I think it still depends on `--iree-dispatch-creation-enable-split-reduction`. (Please correct me if I’m wrong, CC @yzhang93.)
> 
> My last PR addressing the race condition is here: #22694. Once all related PRs land, I believe we can enable it by default.

@bangtianliu You are right. It still needs `--iree-dispatch-creation-enable-split-reduction` to turn on. I had a draft to turn split reduction on by default but paused the work after our meeting figuring out we can still keep the flag in BOO/fusilli.

PR #22472, opened 2025-10-30
[DT] Collapse MaterializeEncodingIntoPaddingPass into the generic pass.

### LINKED ISSUE #20160: Collapse MaterializeEncodingIntoPadding into the generic pass
@kuhar has been developing the GPU padding approach with the data-tiling flow. We branched a [MaterializeEncodingIntoPadding](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingIntoPadding.cpp) pass that cooks the padding logic separately.

Now we observe that the divergence is about how we lower flow.dispatch.tensor load/store ops, see [patterns](https://github.com/iree-org/iree/blob/3692165533ecb7ac6e87449a7c480f6a39da7d9c/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingIntoPadding.cpp#L231-L232) for details. I think we should be able to collapse the patterns if we move pad/pack/swizzle logic to [attribute interface methods](https://github.com/iree-org/iree/blob/3692165533ecb7ac6e87449a7c480f6a39da7d9c/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td#L119-L157). Because the whole point is about getting correct shape type and dispatch the index/offset/strides.

@pashu123 can you pick this up and expand the subtasks in the issue?

hanhanW:
This PR consolidates encoding materialization by merging the separate padding encoding pass into the main `MaterializeDeviceEncodingPass`. The changes eliminate the specialized `MaterializeEncodingIntoPaddingPass` and handle both data tiling and padding encoding resolution through a unified pass. The pass still prioritizes the pre-configured encoding resolver. If it is not present, it tries testing encoding resolver.

Key changes:
- Removed the separate `MaterializeEncodingIntoPaddingPass` and its associated `addEncodingToPaddingPasses` function.
- Extended `MaterializeDeviceEncodingPass` with a new `test-gpu-encoding-resolver` option that supports both `gpu_data_tiling` and `gpu_padding` resolver kinds for testing purpose.
- Updated test files to use the unified pass with appropriate resolver configuration.

Fixes https://github.com/iree-org/iree/issues/20160


hanhanW:
I need to update few more comments and docs.

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
```diff
@@ -59,42 +61,43 @@ updateFuncSignature(FunctionOpInterface funcOp,
...
 }
 
-static LogicalResult
-materializeFuncOpEncodings(FunctionOpInterface funcOp,
-                           IREE::HAL::ExecutableTargetAttr targetAttr,
-                           bool testCLGPUTarget = false) {
+static LogicalResult materializeFuncOpEncodings(
+    FunctionOpInterface funcOp, IREE::HAL::ExecutableTargetAttr targetAttr,
+    ::mlir::iree_compiler::detail::TestingResolverKind resolverSource =
+        ::mlir::iree_compiler::detail::TestingResolverKind::kNone) {
```
kuhar: do we need these `::mlir::` prefixes?

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
```diff
@@ -59,42 +61,43 @@ updateFuncSignature(FunctionOpInterface funcOp,
...
+      SmallVector<NamedAttribute> configItems;
+      IREE::GPU::TargetAttr targetAttr =
+          getGPUTargetAttr(ctx, /*target=*/nullptr);
+      if (targetAttr) {
         addConfigGPUTarget(ctx, targetAttr, configItems);
-        return cast<IREE::Encoding::LayoutMaterializerAttr>(
-            IREE::GPU::GPUEncodingResolverAttr::get(
-                ctx, DictionaryAttr::get(ctx, configItems)));
+        switch (resolverSource) {
+        case ::mlir::iree_compiler::detail::TestingResolverKind::
```
kuhar: also here and below

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
```diff
@@ -177,11 +181,11 @@ materializeFuncOpEncodings(FunctionOpInterface funcOp,
...
-          // If we have a pack/unpack consumer and a producer that has multiple
-          // uses, this _probably_ means the producer won't get dce'd. If that
-          // is the case, by folding the consumer pack/unpack, we break the
-          // producer consumer chain between them and inhibit fusion later in
-          // the pipeline.
+          // If we have a pack/unpack consumer and a producer that has
+          // multiple uses, this _probably_ means the producer won't get
+          // dce'd. If that is the case, by folding the consumer pack/unpack,
+          // we break the producer consumer chain between them and inhibit
+          // fusion later in the pipeline.
```
kuhar: why change the comments?

kuhar (approved):
Thanks for cleaning this up

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
hanhanW: There are other `detail` namespaces in either `mlir` or `iree_compiler` which makes it confusing.

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
hanhanW: hmm, not sure what happened. I'll revert it.

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
kuhar: I'd think that `iree_compiler::detail` would suffice, no?

compiler/src/iree/compiler/Codegen/Common/Passes.h
```diff
@@ -13,6 +13,7 @@
 #define IREE_COMPILER_CODEGEN_COMMON_PASSES_H_
 
 #include "iree/compiler/Codegen/Common/CombineLayoutTransformation.h"
+#include "iree/compiler/Codegen/Common/EncodingUtils.h"
```
jtuyls, line 16: This shouldn't be needed?

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -616,8 +616,18 @@ def MaterializeDeviceEncodingPass :
...
+    Option<"testGPUEncodingResolver", "test-gpu-encoding-resolver", "::mlir::iree_compiler::detail::TestingResolverKind",
+           /*default=*/"detail::TestingResolverKind::kNone",
+           "Flag used for lit-testing GPU target only. Not for general usage",
+           [{llvm::cl::values(
+             clEnumValN(::mlir::iree_compiler::detail::TestingResolverKind::kGPUDataTiling, "gpu_data_tiling",
+                        "Use iree_gpu.gpu_encoding_resolver<>."),
+             clEnumValN(::mlir::iree_compiler::detail::TestingResolverKind::kGPUPadding, "gpu_padding",
+                        "Use iree_gpu.gpu_padding_resolver<>."),
+             clEnumValN(::mlir::iree_compiler::detail::TestingResolverKind::kNone, "none",
+                        "Ignore testing resolvers.")
```
jtuyls:
```suggestion
                        "Ignore testing resolvers. Fall back to identity resolver.")
```
Nit: Might be useful to mention this here as well.

jtuyls (approved):
LGTM. Very nice!

compiler/src/iree/compiler/Codegen/Common/Passes.h
hanhanW: This is needed because we declare enum class and use it in `Passes.td`. To be more specific, it is required by this include:

https://github.com/iree-org/iree/blob/b973deabd2e4872c3127702b5c5b4b2b9c6d2992/compiler/src/iree/compiler/Codegen/Common/Passes.h#L181-L182

compiler/src/iree/compiler/Codegen/Common/Passes.td
hanhanW: Thanks, I missed it!

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
hanhanW: > And what’s worse: unqualified search starts at the bottom of that list and stops as soon as there is a namespace match.

Sorry, I was wrong on my previous response. And it is why I was confused about the above sentence. The root cause is that we are using namespace and the tablegen generates `detail` namespace in `IREE::Encoding` dialect. I removed the `using namespace` and it looks happy now.

https://github.com/iree-org/iree/blob/b973deabd2e4872c3127702b5c5b4b2b9c6d2992/compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp#L39

hanhanW: > I'd think that iree_compiler::detail would suffice, no?

This is correct. I misremembered the guidance from absl. I read it again this morning.

https://abseil.io/tips/130

https://abseil.io/tips/119

PR #22474, opened 2025-10-30
[DT][NFCI] Use no-rollback driver for MaterializeEncoding passes.

### LINKED ISSUE #22452: Switch encoding materialization pass (codegen) to use no-rollback dialect conversion driver
It is faster, more straightforward, and easier to debug. I had few bad debugging experiences and the new driver may improve it, as the logic is simple: we only look at a single op and the types to generate a few new ops.

hanhanW:
No-rollback driver is usually more efficient and debugging friendly, as the state is cleaner. The author of the revision already uses it in some debugging, and it is helpful in terms of debugging messages.

See https://discourse.llvm.org/t/rfc-a-new-one-shot-dialect-conversion-driver for more details.

Fixes https://github.com/iree-org/iree/issues/22452


jtuyls (approved):
LGTM, nice!

PR #22478, opened 2025-10-30
CMake: When `rocminfo` is present, ask users to explicitly enable or disable ROCm testing.

bjacob:
This is a leading cause of surprises and caveats for new users - it's something that is not naturally fitting in generic CMake instructions, but is important for people setting up for ROCm development.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,39 +248,6 @@ else()
 endif()
 
 #-------------------------------------------------------------------------------
-# HIP Default Target Configuration.
-#
-# HIP does not have a stable instruction set like NVIDIA PTX; it requires
-# binaries specific to a target chip. We have tests that generate and run
-# deployable code which need to specify the proper target chip.
-#-------------------------------------------------------------------------------
-
-# When ROCm is detected, force the user to make a conscious decision to enable
-# or disable ROCm testing. It's hard to enable automatically for a given target
-# chip because that would rely too much on guessing what the target chip is.
-# But silently continuing with ROCm testing disabled by default has also been
-# surprising.
-if (NOT DEFINED CACHE{IREE_HIP_TEST_TARGET_CHIP})
-  find_program(_ROCMINFO_COMMAND rocminfo)
-  if (_ROCMINFO_COMMAND)
-    execute_process(COMMAND ${_ROCMINFO_COMMAND} OUTPUT_VARIABLE _ROCMINFO_OUTPUT)
-    string(REGEX MATCH "Name:[ ]*gfx[0-9]+" _ROCMINFO_OUTPUT_NAME_LINE "${_ROCMINFO_OUTPUT}")
-    string(REGEX MATCH "gfx[0-9]+" _ROCMINFO_GFX_NAME "${_ROCMINFO_OUTPUT_NAME_LINE}")
-    if (_ROCMINFO_GFX_NAME)
-      # This needs to be FATAL_ERROR not SEND_ERROR, else the set(CACHE) below
-      # will prematurely hide this error in case cmake is run again without
-      # addressing the issue.
-      message(FATAL_ERROR "ROCm device detected: ${_ROCMINFO_GFX_NAME}. Please explicitly either enable or disable ROCm testing.\n"
-                      "To enable ROCm testing:\n    cmake -DIREE_HIP_TEST_TARGET_CHIP=${_ROCMINFO_GFX_NAME}\n"
-                      "To disable ROCm testing:\n    cmake -DIREE_HIP_TEST_TARGET_CHIP=\n" )
-    endif()
-  endif()
-endif()
-set(IREE_HIP_TEST_TARGET_CHIP "" CACHE STRING
-  "Target chip for HIP tests that need to compile device code. \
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -255,6 +255,27 @@ endif()
 # deployable code which need to specify the proper target chip.
 #-------------------------------------------------------------------------------
 
+# When ROCm is detected, force the user to make a conscious decision to enable
+# or disable ROCm testing. It's hard to enable automatically for a given target
+# chip because that would rely too much on guessing what the target chip is.
+# But silently continuing with ROCm testing disabled by default has also been
+# surprising.
+if (NOT DEFINED CACHE{IREE_HIP_TEST_TARGET_CHIP})
```
kuhar: Shouldn't we either make this conditional on the hip backend being enabled or also warn when the backend is not being built?

CMakeLists.txt
```diff
@@ -255,6 +255,27 @@ endif()
...
+  find_program(_ROCMINFO_COMMAND rocminfo)
+  if (_ROCMINFO_COMMAND)
+    execute_process(COMMAND ${_ROCMINFO_COMMAND} OUTPUT_VARIABLE _ROCMINFO_OUTPUT)
+    string(REGEX MATCH "Name:[ ]*gfx[0-9]+" _ROCMINFO_OUTPUT_NAME_LINE "${_ROCMINFO_OUTPUT}")
+    string(REGEX MATCH "gfx[0-9]+" _ROCMINFO_GFX_NAME "${_ROCMINFO_OUTPUT_NAME_LINE}")
+    if (_ROCMINFO_GFX_NAME)
```
kuhar: This is not guaranteed to pick the right default on machines with multiple gpus. The rocminfo enumeration order may be different than the iree device list order, so you may suggest an architecture that will fail without setting `HIP_VISIBLE_DEVICES=n`

CMakeLists.txt
bjacob: Good idea, thanks!

CMakeLists.txt
bjacob: Good catch, thanks!

CMakeLists.txt
```diff
@@ -418,6 +406,44 @@ if(IREE_HAL_EXECUTABLE_PLUGIN_SYSTEM_LIBRARY)
...
+#-------------------------------------------------------------------------------
+
+# When ROCm is detected on the build host and a corresponding HAL driver is enabled,
+# force the user to make a conscious decision to enable or disable ROCm testing.
+# It's hard to enable automatically for a given target
+# chip because that would rely too much on guessing what the target chip is.
+# But silently continuing with ROCm testing disabled by default has also been
+# surprising.
+if ((IREE_HAL_DRIVER_HIP OR IREE_HAL_DRIVER_AMDGPU) AND NOT DEFINED CACHE{IREE_HIP_TEST_TARGET_CHIP})
+  find_program(_ROCMINFO_COMMAND rocminfo)
```
ScottTodd, line 424: You could try using `offload-arch` (formerly `amdgpu-arch`) instead, as that is more generic, available on Windows too, and produces more machine-friendly output

kuhar (approved):
Looks fine to me but I'm not an expert and haven't tried running this. It would be good to get at least one more approval.

CMakeLists.txt
kuhar: I don't have `offload-arch` in my installation (rocm 7.0), and `amdgpu-arch` have llvm version suffixes, which would make it a bit tricky to select...

bjacob: Wait this is great but the only place I can see `offload-arch` is in my LLVM directory from a tarball. AFAICS, if I hadn't switched to a llvm tarball as my toolchain (e.g. if I were still using the system toolchain) I wouldn't have this program at all?

bjacob: There is a `amdgpu-arch` under /opt/rocm-7.0.x but not in PATH - it's in some LLVM toolchain subdir.

bjacob: Normally I'd be thrilled to switch to building `offload-arch` ourselves from the LLVM tree but here I need this information early in configure.

bjacob: > Available on Windows too

Is rocminfo not available on Windows?

ScottTodd: Both have offload-arch / amdgpu-arch.

Windows has hipinfo.

Linux has rocminfo and rocm_agent_enumerator.

I'm still figuring out how to eventually turn that into something sane.

bjacob: No urgency on merging this .... this PR is just cherries on cakes.

One safe but kinda suboptimal resolution is to just build the offload-arch target from our llvm-project tree. Then give up on producing a configure-time diagnostic and instead produce a ctest-time diagnostic, which also removes the assumption that build host == test host.

bjacob:
@kuhar and I decided to just merge this as-is because however imperfect, it's better than the current state. When `rocminfo` is not in PATH (e.g. on Windows) it just doesn't do anything, as `find_program` fails.

Anyway, can't merge at the moment due to CI breakage.

ScottTodd (approved):

PR #22479, opened 2025-10-30
[GPU] Add serial tiling level

Groverkss:
This PR adds a serial tiling level, meant to tile parallel iterator types with serial loops (as opposed to ones mapped to workgroup/thread/subgroup/lane/etc.). Previously, VectorDistribute was using "reduction" tiling level to map parallel iterators which cannot be tiled to workgroups to "reduction" tiling level and hacking around it.

The serial tiling level also uses a scf.forall without a mapping, which also hints that all writes are independent.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUApplyTilingLevel.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUApplyTilingLevel.cpp
@@ -65,7 +65,8 @@ void GPUApplyTilingLevelPass::runOnOperation() {
   if (tilingLevel != IREE::GPU::TilingLevel::Reduction &&
       tilingLevel != IREE::GPU::TilingLevel::Thread &&
       tilingLevel != IREE::GPU::TilingLevel::Subgroup &&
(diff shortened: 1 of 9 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUApplyTilingLevel.cpp
```diff
@@ -65,7 +65,8 @@ void GPUApplyTilingLevelPass::runOnOperation() {
   if (tilingLevel != IREE::GPU::TilingLevel::Reduction &&
       tilingLevel != IREE::GPU::TilingLevel::Thread &&
       tilingLevel != IREE::GPU::TilingLevel::Subgroup &&
-      tilingLevel != IREE::GPU::TilingLevel::PartialReduction) {
+      tilingLevel != IREE::GPU::TilingLevel::PartialReduction &&
+      tilingLevel != IREE::GPU::TilingLevel::Serial) {
```
kuhar: use `llvm::is_contained`

qedawkins (approved):
LGTM % Jakub's nit

efric (approved):
Thanks, this is more clear :smile:

PR #22486, opened 2025-10-30
[Codegen][GPU] Allow intentional padding for non-K-major matmul layouts

jerryyin:
Only disallow intentional padding for K-major layouts ([K,M] × [K,N]) where transposedLhs=true and transposedRhs=false. Experiments have shown that allowing padding for other layout variants (NN, NT, TT) provides decent performance improvements.

This has improved 2.69% of GEMMs from 322 GEMMs we track. In particular, the biggest improvements comes from the cases that would have been excluded because we used to only allow [M, K] [N, K] layouts:

> hipblaslt-bench   --api_method c -m 16384 -n 150000 -k 4096 --lda 16384 --ldb 4096 --ldc 16384   --ldd 16384  --stride_a 0 --stride_b 0   --stride_c 0 --stride_d 0  --alpha   1.000000 --beta 0.000000 --transA N --transB N --batch_count 1  --a_type bf16_r --b_type bf16_r --c_type   bf16_r --d_type bf16_r --scale_type f32_r --bias_type f32_r   --compute_type f32_r

Improved by 79%, from 235860 to 50129 us

> hipblaslt-bench --api_method c   -m 8192 -n 150000 -k 2048 --lda 8192 --ldb 2048 --ldc 8192 --ldd 8192  --stride_a 0 --stride_b 0 --stride_c 0   --stride_d 0  --alpha 1.000000 --beta   0.000000 --transA N --transB N --batch_count 1  --a_type bf16_r --b_type bf16_r --c_type   bf16_r --d_type bf16_r --scale_type f32_r --bias_type f32_r   --compute_type f32_r

Improved by 77%, from 54707 to 12549 us

> hipblaslt-bench --api_method c   -m 4096 -n 150000 -k 16384 --lda 4096 --ldb 16384 --ldc 4096 --ldd 4096  --stride_a 0 --stride_b 0 --stride_c 0   --stride_d 0  --alpha 1.000000 --beta   0.000000 --transA N --transB N --batch_count 1  --a_type bf16_r --b_type bf16_r --c_type   bf16_r --d_type bf16_r --scale_type f32_r --bias_type f32_r   --compute_type f32_r

Improved by 77% from 229361 to 52979 us

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -646,16 +646,11 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   Type lhsElemType = getElementTypeOrSelf(lhs);
   Type rhsElemType = getElementTypeOrSelf(rhs);
   Type initElemType = getElementTypeOrSelf(init);
-  // For GEMM: Only disallow padded configurations for K-major layouts
-  // ([K,M] × [K,N]) where transposedLhs=true and transposedRhs=false.
-  // Experiments have shown that allowing padding for other layout variants
-  // (NN, NT, TT) provides decent performance improvements.
-  //
-  // For convolutions: Keep conservative padding policy to avoid overhead.
-  bool disallowPadding = isGemm ? (transposedLhs && !transposedRhs)
-                                : (transposedLhs || !transposedRhs);
-  GPUMatmulShapeType problem{getDimBounds(mDims, disallowPadding),
-                             getDimBounds(nDims, disallowPadding),
+  // Intentionally padded GEMM proved to be beneficial for performance for
+  // the following layouts: 1) [M, K] x [K, N] 2) [M, K] x [N, K]
+  // Therefore we disallow padding only when LHS is transposed.
+  GPUMatmulShapeType problem{getDimBounds(mDims, transposedLhs),
+                             getDimBounds(nDims, transposedLhs),
                              getDimBoundsNoPad(kDims),
                              getDimBoundsNoPad(batchDims),
                              lhsElemType,
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -646,19 +646,21 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   Type lhsElemType = getElementTypeOrSelf(lhs);
   Type rhsElemType = getElementTypeOrSelf(rhs);
   Type initElemType = getElementTypeOrSelf(init);
-  // We only disallow padded configurations for K-major layouts ([K,M] × [K,N])
-  // where transposedLhs=true and transposedRhs=false. Experiments have shown
-  // that allowing padding for other layout variants (NN, NT, TT) provides
-  // decent performance improvements. Once we fix the root causes for padding
-  // overhead in K-major layouts, we can allow padding for K-major layouts.
-  GPUMatmulShapeType problem{
-      getDimBounds(mDims, transposedLhs && !transposedRhs),
-      getDimBounds(nDims, transposedLhs && !transposedRhs),
-      getDimBoundsNoPad(kDims),
-      getDimBoundsNoPad(batchDims),
-      lhsElemType,
-      rhsElemType,
-      initElemType};
+  // For GEMM: Only disallow padded configurations for K-major layouts
+  // ([K,M] × [K,N]) where transposedLhs=true and transposedRhs=false.
+  // Experiments have shown that allowing padding for other layout variants
+  // (NN, NT, TT) provides decent performance improvements.
+  //
+  // For convolutions: Keep conservative padding policy to avoid overhead.
+  bool disallowPadding = isGemm ? (transposedLhs && !transposedRhs)
+                                : (transposedLhs || !transposedRhs);
+  GPUMatmulShapeType problem{getDimBounds(mDims, disallowPadding),
+                             getDimBounds(nDims, disallowPadding),
+                             getDimBoundsNoPad(kDims),
+                             getDimBoundsNoPad(batchDims),
+                             lhsElemType,
+                             rhsElemType,
+                             initElemType};
 
   bool mustBeAligned = true;
   std::optional<GPUMMASchedule> schedule = getMmaScheduleFromProblemAndTarget(
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -646,13 +646,14 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   Type lhsElemType = getElementTypeOrSelf(lhs);
   Type rhsElemType = getElementTypeOrSelf(rhs);
   Type initElemType = getElementTypeOrSelf(init);
-  // TODO (nirvedhmeshram): We only voluntarily allow padded configurations
-  // for tranpose_b layouts as that's where we currently don't have any overhead
-  // for padding. Other layouts still can have overhead and once we fix the root
-  // causes for that we can relax this condition.
+  // We only disallow padded configurations for K-major layouts ([K,M] × [K,N])
+  // where transposedLhs=true and transposedRhs=false. Experiments have shown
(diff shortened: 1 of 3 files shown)


Max191 (commented):
> Only disallow intentional padding for K-major layouts ([K,M] × [K,N]) where transposedLhs=true and transposedRhs=false.

Did you test this layout and find negative/no performance impact?

nirvedhmeshram:
> > Only disallow intentional padding for K-major layouts ([K,M] × [K,N]) where transposedLhs=true and transposedRhs=false.
> 
> Did you test this layout and find negative/no performance impact?

I have done some experiments which showed bad results for that layout see column K in N variable tab of this spreadsheet
https://docs.google.com/spreadsheets/d/1igKccxmLCwPMHKEPCMoXe6F-LBCkGHhkogYVCWpxtiU/edit?gid=2003089531#gid=2003089531

Additionally, for the matmul layout which this PR is enabling the intentional padding for the results were mostly positive but occasionally had regressions hence I didnt enable it in the original PR (column H).  The reason for the mixed results is that while we benefit from the better tiles, the N dimension that is being padded is an inner dimension so we will end up with the if/else masked load situation. However, I think it does make sense to make this trade-off for the matmul case and we should instead focus on improving unaligned perf for these layouts instead of keeping the hueristic being sub-optimal to favor aligned configurations.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -645,13 +645,14 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
-  // causes for that we can relax this condition.
+  // We only disallow padded configurations for K-major layouts ([K,M] × [K,N])
+  // where transposedLhs=true and transposedRhs=false. Experiments have shown
+  // that allowing padding for other layout variants (NN, NT, TT) provides
+  // decent performance improvements. Once we fix the root causes for padding
+  // overhead in K-major layouts, we can allow padding for K-major layouts.
   GPUMatmulShapeType problem{
-      getDimBounds(mDims, transposedLhs || !transposedRhs),
-      getDimBounds(nDims, transposedLhs || !transposedRhs),
+      getDimBounds(mDims, transposedLhs && !transposedRhs),
```
nirvedhmeshram: Is this condition strong enough, it will allow something like ([K,M] X [NxK]) which is not a layout we do testing for, so we shouldnt allow it I think?
I think what we should do is `getDimBounds(mDims, transposedLhs)` so basically we allow both matmul and matmul_tranpose_b but not any other layouts for now.

jerryyin: I think it is easier to get my reply in a thread. I did give a second read at the excel sheets you have but have trouble correlating some of the experimental values with baseline, and it'd be good to clarify with you here:

> but occasionally had regressions hence I didnt enable it in the original PR (column H)

Is column H [M, K] x [K, N] layout? Also, is column H's baseline in column F? If answer to both of them are yes, then I can see your point. The regression seems to be particularly common in smaller convolutions

>  the N dimension that is being padded is an inner dimension so we will end up with the if/else masked load situation. 

Nice analysis!

> However, I think it does make sense to make this trade-off for the matmul case

Yes, I can take your feedback and make this change for matmul only. Right now I'm only staring at matmul and can confirm this change is very positive for matmul cases we track.

jerryyin:
@Max191 Throughout the entire BOO GEMM sheet we are tracking, this change is very positive and reduced a lot of model overhead. Also, yes, it does cause some regressions in smaller shaped GEMMs but the improvement for those trans A/B large gemm shapes far outweigh the damage it has caused. We can potentially walk through the baseline/experimental result next time so I can demonstrate that to you.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
nirvedhmeshram: > Is column H [M, K] x [K, N] layout? Also, is column H's baseline in column F? If answer to both of them are yes, then I can see your point. The regression seems to be particularly common in smaller convolutions

Yup that is correct
> Yes, I can take your feedback and make this change for matmul only. Right now I'm only staring at matmul and can confirm this change is very positive for matmul cases we track.

Sounds good!

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -646,18 +646,21 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
-      lhsElemType,
-      rhsElemType,
-      initElemType};
+  // For GEMM: Only disallow padded configurations for K-major layouts
+  // ([K,M] × [K,N]) where transposedLhs=true and transposedRhs=false.
+  // Experiments have shown that allowing padding for other layout variants
+  // (NN, NT, TT) provides decent performance improvements.
+  //
+  // For convolutions: Keep conservative padding policy to avoid overhead.
+  bool disallowPadding = isGemm ? (transposedLhs && !transposedRhs)
```
nirvedhmeshram: Do we really need this special handling for convs vs GEMM?, secondly this condtion ` (transposedLhs && !transposedRhs)` seems bad for GEMM as well as per my previous point that some untested layouts will sneak through,  could we just have `transposedLhs` as  universal condition ideally or just for GEMM if it must be that way due to some severe regressions that are conv specific?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -646,18 +646,21 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
-      getDimBounds(nDims, transposedLhs || !transposedRhs),
-      getDimBoundsNoPad(kDims),
-      getDimBoundsNoPad(batchDims),
-      lhsElemType,
-      rhsElemType,
-      initElemType};
+  // For GEMM: Only disallow padded configurations for K-major layouts
+  // ([K,M] × [K,N]) where transposedLhs=true and transposedRhs=false.
+  // Experiments have shown that allowing padding for other layout variants
+  // (NN, NT, TT) provides decent performance improvements.
```
nirvedhmeshram: I dont follow the terminology of `(NN, NT, TT)` can we use MNK to describe the layout?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
jerryyin: Upon double checking the data I've collected, I think the major uplift come from [M, K] [K, N] layout. So at the very minimal we should allow this layout to be padded. Together with default enabled layout [M, K] [N, K] that indeed left us to be transposedLhs. 

Are you saying that we should just unify GEMM and conv with a similar `transposedLhs` condition here? Upon double checking your excel sheet conv's regular matmul case actually does have a general uplife.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
jerryyin: We can but the two letter description is one universal approach in describing the GEMM layout everywhere else and quite succinct. I think we should be able to allow those to appear in the comments.

It's actually quite easy to process mentally too:
 - NN: [M, K] x [K, N]
 - Any other layout that transpose a particular operand has a T in that letter. For example NT means [M, K] x [N, K] because the second operand has a transpose from the default NN layout.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
nirvedhmeshram: My spreadsheet is for GEMM as well, but I think conv would behave similarly will have to run the shapes to make sure. And yes my point is since this condition is to disallowPadding you will end up setting it to false and hence allow padding for some untested layouts along with matmul and matmul_transpose_b.

jerryyin: Yep, agreed, this is done in latest commit.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
nirvedhmeshram: I see, thanks for explaining!

nirvedhmeshram (approved):
LGTM

PR #22490, opened 2025-10-30
[Codegen][GPU] Generalize linalg.reduce operations

bangtianliu:

Debugging the arg_compare op compilation issue in llama 8b fp16 quality tests from PR https://github.com/iree-org/iree/pull/22379.

https://github.com/iree-org/iree/actions/runs/19219575275/job/54935278831?pr=22394#step:7:122

### Problem
When enabling split reduction for arg_compare operations (`--iree-dispatch-creation-enable-split-reduction`), the TypePropagationPass fails on linalg.reduce operations with illegal i1 types:

```mlir
%reduced = linalg.reduce ins(%3 : tensor<1x1xi1>) outs(%5 : tensor<i1>) dimensions = [0, 1] 
  (%in: i1, %init: i1) {
    %7 = arith.andi %in, %init : i1
    linalg.yield %7 : i1
  }
```

### Solution 

This PR enables GPU generalization of linalg.reduce operations by adding support in GPUGeneralizeNamedOps pass, which converts named `linalg.reduce` ops to `linalg.generic` form.


compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -429,6 +429,91 @@ struct IREELinalgExtSortTypePropagation
   }
 };
 
+/// Pattern to legalize `linalg.reduce` operations.
+struct ReduceOpTypePropagation
+    : public TypePropagationPattern<linalg::ReduceOp> {
```
kuhar:
```suggestion
struct ReduceOpTypePropagation final
    : TypePropagationPattern<linalg::ReduceOp> {
```

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -429,6 +429,91 @@ struct IREELinalgExtSortTypePropagation
...
+  using TypePropagationPattern<linalg::ReduceOp>::TypePropagationPattern;
```
kuhar:
```suggestion
  using TypePropagationPattern::TypePropagationPattern;
```

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -429,6 +429,91 @@ struct IREELinalgExtSortTypePropagation
...
+
+  LogicalResult
+  matchAndRewrite(linalg::ReduceOp reduceOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    bool needsConversion = false;
+    for (auto &operand : reduceOp->getOpOperands()) {
```
kuhar: spell out types that are not obvious based on the immediate context: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -429,6 +429,91 @@ struct IREELinalgExtSortTypePropagation
...
+      Type operandType = operand.get().getType();
+      Type legalizedType = this->getTypeConverter()->convertType(operandType);
```
kuhar: what if type conversion fails?

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -429,6 +429,91 @@ struct IREELinalgExtSortTypePropagation
...
+    if (!needsConversion) {
+      return rewriter.notifyMatchFailure(
+          reduceOp, "all types already legal within conversion pattern");
+    }
+
+    SmallVector<Type> resultTypes;
+    for (auto init : reduceOp.getInits()) {
+      Type legalizedType =
+          this->getTypeConverter()->convertType(init.getType());
+      resultTypes.push_back(legalizedType);
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -429,6 +429,91 @@ struct IREELinalgExtSortTypePropagation
...
+    rewriter.setInsertionPoint(yieldOp);
+
+    SmallVector<Value> convertedYieldValues;
+    bool needsUpdate = false;
+    for (Value yieldValue : yieldOp->getOperands()) {
+      Type yieldType = yieldValue.getType();
+      std::optional<Type> legalizedYieldType =
+          legalizeStorageElementType(yieldType);
+
+      Value valueToYield = yieldValue; // Default to original value
```
kuhar:
```suggestion
      Value valueToYield = yieldValue; // Default to original value.
```
https://llvm.org/docs/CodingStandards.html#commenting

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -429,6 +429,97 @@ struct IREELinalgExtSortTypePropagation
...
+      if (!legalizedType) {
+        return reduceOp.emitOpError("failed to get legalized type for operand");
+      }
+      if (operandType != legalizedType) {
+        needsConversion = true;
+        break;
+      }
+    }
+
+    if (!needsConversion) {
```
MaheshRavishankar: Hmm... This should actually already be checked by the conversion pattern driver. If you get here. it pretty much means that the conversion is needed.

MaheshRavishankar (changes requested):
Why dont we just generalize the ops before we even get here... Do we care about keeping the `linalg.reduce` as a `linalg.reduce` ?

I think just adding `linalg::ReduceOp` here : https://github.com/iree-org/iree/blob/2294bb85e56c126084a61fd9dee95f50bad4eaaa/compiler/src/iree/compiler/Codegen/Common/GPU/GPUGeneralizeNamedOps.cpp#L61 should also fix the problem you are hitting.

bangtianliu:
> Why dont we just generalize the ops before we even get here... Do we care about keeping the `linalg.reduce` as a `linalg.reduce` ?
> 
> I think just adding `linalg::ReduceOp` here :
> 
> https://github.com/iree-org/iree/blob/2294bb85e56c126084a61fd9dee95f50bad4eaaa/compiler/src/iree/compiler/Codegen/Common/GPU/GPUGeneralizeNamedOps.cpp#L61
> 
> should also fix the problem you are hitting.

Yes, I tried it, and generalizing linalg.reduction also solves the problem. If that’s the case, I’ll update this PR to generalize the linalg.reduction op instead.

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
bangtianliu: Yes, correct.

MaheshRavishankar (approved):

kuhar (approved):

PR #22357, opened 2025-10-20
[DispatchCreation] Set split reduction size for GEMM with large k dim

yzhang93:
This PR adds basic support for setting split reduction size for matmul-like ops with large K dim. Note that the constant thresholds are empirically chosen based on limited data (1x1 filter weight backward convs) and may not generalize to all cases. It's challenging to find a single threshold to apply for all shapes. The bottom line is to improve the performance for extremely large K cases while not to degrade many smaller shapes.

PR #22275, opened 2025-10-10
[DispatchCreation] Add split reduction for weight backward convs

yzhang93:
Weight backward convolutions have a special CHWN layout, where the filter sizes (corresponding to output image sizes in forward convolutions) are typically large, while the output spatial dimensions are small. This makes the split reduction strategy particularly effective. This PR adds support to split these convs along the input channel dimension.

Some experimental thresholds are applied to filter out cases that won't benefit from splitting reduction. Particular checks include:

- When the batch and output channel sizes are large, the workload tends to distributed across many workgroups, making split reduction little to no effect.
- When the input spatial sizes are small while the batch and output channel sizes are relatively larger (medium size), split reduction often has no effect or even degrades performance.

PR #22491, opened 2025-10-30
[DispatchCreation] Enable splitting multiple reduction dimensions for weight backward convs

yzhang93:
This PR is a follow-up for https://github.com/iree-org/iree/pull/22275. It removes the constraint that only splitting input channel dimension, and added support to split across multiple dimensions. The heuristics for setting multi-dimension tile sizes is similar to what is for GEMM https://github.com/iree-org/iree/pull/22357. More than half of the tracked weight backward shapes are benefiting from this change.

Example runtime comparison for
`convbfp16 -n 16 -c 16 -H 225 -W 225 -k 64 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC --iter 100`

- Without split reduction: 19352.8 ms
- Split only the input channel dimension: 1445.1 ms
- Split multiple reduction dimensions: 371.7 ms

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -176,8 +176,10 @@ private:
       return std::nullopt;
     }
 
-    if (convDims->inputChannel.empty() || convDims->outputChannel.empty() ||
-        convDims->batch.empty() || convDims->filterLoop.empty()) {
+    // Require non-empty filter, output channel, and batch dimensions to derive
+    // heuristics. The input channel dimension may be empty.
+    if (convDims->outputChannel.empty() || convDims->batch.empty() ||
+        convDims->filterLoop.empty()) {
       LDBG() << "skipping op; missing convolution dimensions";
       return std::nullopt;
     }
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -208,27 +210,43 @@ struct SetSplitReductionSizesPass final
...
 
-    std::optional<int64_t> batchLastDim = outputMap.getResultPosition(
-        getAffineDimExpr(convDims->batch.back(), outputMap.getContext()));
-    if (!batchLastDim || batchLastDim.value() != outputShape.size() - 1) {
-      LDBG() << "skipping op; not batch last layout";
+    auto getDimPositions = [&](ArrayRef<unsigned> dims, const AffineMap &map) {
+      SmallVector<int64_t> positions;
+      for (auto dim : dims) {
+        for (auto [idx, e] : llvm::enumerate(map.getResults())) {
+          if (e.isFunctionOfDim(dim)) {
```
MaheshRavishankar: I dont know if you want to do `isFunctionOfDim` here? Looking at how you are using it, you are only looking for `AffineDimExpr` usage of the dimension. I think the `getResultPosition` is closer to what I think would be better to use.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -239,6 +257,20 @@ struct SetSplitReductionSizesPass final
       return std::nullopt;
     }
 
+    // Compute the product of the specified dimensions. If any dimension list is
+    // empty, return 1.
+    auto getSizeAt = [](ArrayRef<int64_t> shape, ArrayRef<int64_t> pos) {
+      int64_t totalSize = 1;
+      for (unsigned i : pos)
+        totalSize *= shape[i];
```
MaheshRavishankar, line 263: Please add an assert here that the shape[i] is not dynamic. This is relying on that condition being checked above, but if you drop that condition there it will mess up this logic. Also Nit, IREE norm is to always uses `{` `}` even if there is just one statement

MaheshRavishankar (approved):
Few nits. Approving, but please address before landing.

PR #22499, opened 2025-10-31
[Stream] Fixing update order and improving the cache for ReplicateGlobalsPerAffinity pass.

hanhanW:
The update order matters, because the source could be using a new global that is not cached. The revision fixes operand update ordering issue by collecting updates first, then applying them atomically.

The revision also fixes a bug in the `ReplicateGlobalsPerAffinity` pass where operations with multiple results were not properly cached, causing redundant operation creation.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/ReplicateGlobalsPerAffinity.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/ReplicateGlobalsPerAffinity.cpp
@@ -164,6 +164,7 @@ ValuePerAffinityHelper::getOrCreateGlobalForAffinity(
   // The cache was already populated in the constructor.
   Operation *insertionPoint =
       cachedInsertionPointForGlobal.lookup(globalOp.getOperation());
+  assert(insertionPoint && "failed to find insertion point for global");
 
   // Create the new global and initializer after the insertion point.
   Location loc = globalOp.getLoc();
@@ -223,8 +224,13 @@ Value ValuePerAffinityHelper::getOrCreateAffinityOpForAffinity(
       clone(builder, affinityOp, affinityOp->getResultTypes(), newOperands);
   newAffinityOp->setDiscardableAttrs(
       affinityOp->getDiscardableAttrDictionary());
-  cachedValuePerAffinity[key] =
-      newAffinityOp->getResult(opResult.getResultNumber());
+  // Cache all the results, so it won't create new operations when querying for
+  // other results that have the same affinity.
+  for (auto [oldResult, newResult] :
+       llvm::zip_equal(affinityOp->getResults(), newAffinityOp->getResults())) {
+    ValueAffinityPair resultKey = {oldResult, affinityAttr};
+    cachedValuePerAffinity[resultKey] = newResult;
+  }
   return cachedValuePerAffinity[key];
 }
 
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Dialect/Stream/Transforms/ReplicateGlobalsPerAffinity.cpp
```diff
@@ -516,6 +522,16 @@ void ReplicateGlobalsPerAffinityPass::runOnOperation() {
...
+
+    // Order matters here because it is wrong if an operand is already updated
+    // with a new value. E.g.,
+    //   %0 = flow.dispatch @dispatch0(%global)
+    //   %1 = flow.reshape %0 ...
+    // If we update %0 first, then when updating %1 we will see a new global
+    // from %0 which uses the new global, that is not cached in
+    // `globalPerAffinityHelper`.
+    // Thus, we collect all the updates first, then apply them.
+    SmallVector<std::pair<OpOperand *, Value>> updateList;
```
sommerlukas, line 534: Is there a specific reason to use `std::pair` here? I noticed that in other places in this file, `std::tuple` is used, even with just two elements.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/replicate_globals_per_affinity.mlir
```diff
@@ -148,6 +148,43 @@ util.func private @ambiguous_indirect_global_load(%arg0: tensor<f32>, %arg1: ten
...
+// CHECK: util.global private @[[$GLOBAL_A:.+]] : tensor<10xf32>
+// CHECK: util.initializer {
+// CHECK:   %[[LOAD:.+]] = util.global.load @[[$GLOBAL]]
+// CHECK:   %[[TRANSFER_A:.+]] = flow.tensor.transfer %[[LOAD]] {{.+}} to #hal.device.affinity<@[[$DEVICE_A]]>
+// CHECK:   util.global.store %[[TRANSFER_A]], @[[$GLOBAL_A]]
+// CHECK: }
+util.global private @global : tensor<10xf32>
+
+// CHECK-LABEL: @ambiguous_indirect_global_load_and_multi_result(
+// CHECK-SAME:    %[[ARG0:[a-zA-Z0-9]+]]
```
sommerlukas, line 164: What is the reason to be explicit with the allowed characters here compared to the `.+` used in other places? Is it to avoid matching `,` and including it in the substitution block?

sommerlukas (commented):
Just two minor questions from my side for now.

compiler/src/iree/compiler/Dialect/Stream/Transforms/ReplicateGlobalsPerAffinity.cpp
hanhanW: I think they should be pair if we exactly need two elements. I may use tuple for convenience when I prototyped the pass, and it results in the current status.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/replicate_globals_per_affinity.mlir
hanhanW: This way you don't have to match types when you have multiple `CHECK-SAME`.

hanhanW: oh, even for single `CHECK-SAME`, you don't have to match `: ` and the rest of the type. Otherwise, it captures all the rest chars.

benvanik (approved):

Issue #20011, opened 2025-02-17
[CPU] Incorrect result with CPU for onnx ReverseSequence


IanWood1:
### What happened?

I added a folding pattern in https://github.com/iree-org/iree/pull/20006 which caused a onnx test to fail. I was able to create a minimal reproducer that shows the difference between correct and incorrect results by placing the `tensor.extract_slice` in a dispatch so that it must be copied. It seems that in the bad accessing the tensor along the innermost dimension instead of the outermost. The "incorrect" example produces correct results when run on rocm.

### Steps to reproduce your issue

1. Save the reproducers. They are functionally equivalent but produce different results. 

### Incorrect Results
```mlir
util.func public @main(%0 : tensor<4x4xf32>) -> (tensor<?xf32>, index) {
  %13 = arith.constant 4 : index
  %extracted_slice_0 = tensor.extract_slice %0[0, 0] [%13, 2] [1, 1] : tensor<4x4xf32> to tensor<?x2xf32>
  %c1 = arith.constant 1 : index
  %c0 = arith.constant 0 : index
  %14 = arith.subi %13, %c1 : index
  %15 = tensor.empty(%13) : tensor<?xf32>
  %16 = flow.dispatch.region -> (tensor<?xf32>{%13}) {
    %61 = linalg.generic {indexing_maps = [affine_map<(d0) -> (d0)>], iterator_types = ["parallel"]} outs(%15 : tensor<?xf32>) {
    ^bb0(%out: f32):
      %62 = linalg.index 0 : index
      %63 = arith.subi %14, %62 : index
      %extracted_20 = tensor.extract %extracted_slice_0[%63, %c0] : tensor<?x2xf32>
      linalg.yield %extracted_20 : f32
    } -> tensor<?xf32>
    flow.return %61 : tensor<?xf32>
  }
  util.return %16, %13 : tensor<?xf32>, index
}
```

### Correct Results
```mlir
util.func public @main(%0 : tensor<4x4xf32>) -> (tensor<?xf32>, index) {
  %13 = arith.constant 4 : index
  %reg = flow.dispatch.region -> (tensor<?x2xf32>{%13}) {
   %extracted_slice_0 = tensor.extract_slice %0[0, 0] [%13, 2] [1, 1] : tensor<4x4xf32> to tensor<?x2xf32>
   flow.return %extracted_slice_0 : tensor<?x2xf32>
  }

  %c1 = arith.constant 1 : index
  %c0 = arith.constant 0 : index
[...]

PR #22503, opened 2025-10-31
[Dispatch Creation] Drop unit dims from tensor.extract ops

IanWood1:
Adds pattern for `tensor.extract` to fold unit dimensions. Without this pattern there will be reshapes left in the program which may re-introduce the unit dims when propagated.

The added xfail already has an issue https://github.com/iree-org/iree/issues/20011.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/FoldUnitExtentDims.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FoldUnitExtentDims.cpp
@@ -26,6 +26,7 @@
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Tensor/Transforms/Transforms.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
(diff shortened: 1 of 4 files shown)


Groverkss (commented):
No opinion on this, but I wonder if a better way to handle this is actually to have a collapse/expand -> extract folding pattern just like we do for FoldMemrefAliasOps. It's effectively the same thing. It would handle things more generally and would ensure we always remove reshapes for tensor.extract.

IanWood1:
> No opinion on this, but I wonder if a better way to handle this is actually to have a collapse/expand -> extract folding pattern just like we do for FoldMemrefAliasOps. It's effectively the same thing. It would handle things more generally and would ensure we always remove reshapes for tensor.extract.

I don't have a strong opinion here either but here are some thoughts:

First, this pattern is more "active". It doesn't require an directly adjacent reshape to remove the unit dims from the op. If every op only folded with collapse/expand ops, we wouldn’t be able to eliminate unit dims in cases where the reshape isn’t the immediate producer. That’s partly why the existing patterns in this pass are structured as they are. `tensor.extract` might have different constraints, but my patch directly addresses the immediate issue I ran into. I’m not opposed to adding both patterns, though.

Second, having a `collapse/expand -> extract` folding pattern could be harmful in some cases. For example, take a `tensor.expand` that expands from a 1D tensor to 4D. Folding this into the extract means you have to linearize the indices, which feels less canonical and might obscure the original intent (although it might help with some fusion). If it's just unit dims then it doesn't matter.

Groverkss:
> > No opinion on this, but I wonder if a better way to handle this is actually to have a collapse/expand -> extract folding pattern just like we do for FoldMemrefAliasOps. It's effectively the same thing. It would handle things more generally and would ensure we always remove reshapes for tensor.extract.
> 
> I don't have a strong opinion here either but here are some thoughts:
> 
> First, this pattern is more "active". It doesn't require an directly adjacent reshape to remove the unit dims from the op. If every op only folded with collapse/expand ops, we wouldn’t be able to eliminate unit dims in cases where the reshape isn’t the immediate producer. That’s partly why the existing patterns in this pass are structured as they are. `tensor.extract` might have different constraints, but my patch directly addresses the immediate issue I ran into. I’m not opposed to adding both patterns, though.
> 
> Second, having a `collapse/expand -> extract` folding pattern could be harmful in some cases. For example, take a `tensor.expand` that expands from a 1D tensor to 4D. Folding this into the extract means you have to linearize the indices, which feels less canonical and might obscure the original intent (although it might help with some fusion). If it's just unit dims then it doesn't matter.

When I say folding pattern I didn't mean a "canonicalization folder", sorry for mis communicating that. I just meant a pattern that you can run. Either is okay with me, I'm mainly offering another solution. Feel free to land what you think is better.

MaheshRavishankar (approved):

MaheshRavishankar (commented):
Does this work for rank-reducing extract slices? The base logic should cause you are using just the source type and offsets/sizes. Just need to check the new operation created is valid.

IanWood1:
> Does this work for rank-reducing extract slices? The base logic should cause you are using just the source type and offsets/sizes. Just need to check the new operation created is valid.

Are you referring to adding a different pattern with similar logic as this one but for extract_slices? I think that would work with some small changes.

MaheshRavishankar (commented):
I didnt realize this is `tensor.extract`. That is strange that we have a tensor.extract outside of a `linalg` op. Whats the source of this error.

IanWood1:
I didn't put the `tensor.extract` ops inside of a `linalg.generic` for the tests since it didn't seem necessary for what I was testing for this pattern. But in the example I was looking at from SDXL, the extract ops were in "gather"-like linalg generics. Either way, this still causes unit dims to fail to completely fold away.

I can add a test with the `tensor.extract` nested in a `linalg.generic` if that seems needed.

MaheshRavishankar:
> I didn't put the `tensor.extract` ops inside of a `linalg.generic` for the tests since it didn't seem necessary for what I was testing for this pattern. But in the example I was looking at from SDXL, the extract ops were in "gather"-like linalg generics. Either way, this still causes unit dims to fail to completely fold away.
> 
> I can add a test with the `tensor.extract` nested in a `linalg.generic` if that seems needed.

I was just curious. I already approved the PR.

PR #22511, opened 2025-10-31
[CI] Run w7900 tests on any runner with two w7900 gpus

kuhar:


saienduri:
Can you update all the other references as well?

saienduri (approved):

kuhar:
We can relax some of these later -- many tests don't care if they run on 7900xtx or w7900.

PR #22519, opened 2025-11-02
Implement `subgroups_k` in data-tiled MMA layouts

### LINKED ISSUE #22206: DataTiledMMAAttr should have subgroups_k alongside subgroups_{m,n}
https://discord.com/channels/689900678990135345/1254843174111678555/1423757726416244817

bjacob:
Fixes #22206.

The motivation is that subgroups_k is something that some non-data-tiled ukernels already do, so this should unblock having data-tiled kernels doing the same:
https://github.com/qedawkins/iree/blob/01d0ff8f6123300af453cc83ac13021e89a3f3f0/compiler/plugins/target/ROCM/builtins/ukernel/mlir/iree_mlir_ukernel_mmt_8x64_f4f4f32_gfx950.mlir#L302

Codegen is not currently able to lower this correctly, so this is ukernel-only for the time being.  It would be nontrivial (though feasible) to teach codegen to do this, so it's not clear if we even want to.


Max191 (approved):
Nice, LGTM!

PR #22523, opened 2025-11-03
[Codegen][GPU] Introduce scf::pipelineForLoop function from upstream for prefetchSharedMemory pass

jerryyin:
This PR refactors the loop prefetch pipelining implementation to improve maintainability and separation of concerns. Please feel free to review by commit, it is going to be much easier than review the entire change al-together. Below is the summarized per commit change. Also, this is the minimal change possible to make the prefetch shared memory pass continue to work as-is while keeping the (exact) same functionality. I still plan to refactor a fair amount of the prefetcher code but figured that this is already a good amount of code changes to review.

1. Introduced scf::pipelineForLoop utility from upstream
2. Removed unused code: Cleaned up unused emission functions from LoopPrefetcher, narrowing its focus to stage analysis
3. Introduced explicit operation clustering (read/compute/write groups) with debug logging for visibility into scheduling decisions
4. Fixed barrier handling: Remove original barriers before multi-stage pipelining (they're unpipelined-specific) and insert new barriers at correct pipelined locations
5. Improved scheduling logic: Separated opToCluster map population from finalSchedule construction—operations are now sorted 

Impact: No functional changes; all existing tests pass. The refactoring makes the pipelining logic clearer and easier to extend.

Next major steps includes completely replacing LoopPrefetcher class with more robust pattern matching logic; Plumb the numStage variable as pass option and introduce 3 stage pipeline; Implement double buffering.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/LLVMGPUUtils.h
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/LLVMGPUUtils.h
@@ -55,6 +55,12 @@ void packSharedMemoryAlloc(mlir::FunctionOpInterface funcOp);
 
 // Prefetches data written to shared memory for the next iteration. Returns the
 // new loop on success or failure when the `forOp` is not supported.
+//
+// numStages: Controls the depth of prefetching/pipelining.
+//   - 1: No pipelining (operations remain in original order)
+//   - 2: Two-stage pipeline (default) - prefetches 1 iteration ahead
+//   - 3+: Multi-stage pipeline - prefetches (numStages-1) iterations ahead
+// Stream pipeliner always prefetches one less than the number of stages.
 FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
                                                scf::ForOp forOp,
                                                unsigned numStages = 2);
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/LLVMGPUUtils.h
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/LLVMGPUUtils.h
@@ -56,7 +56,8 @@ void packSharedMemoryAlloc(mlir::FunctionOpInterface funcOp);
 // Prefetches data written to shared memory for the next iteration. Returns the
 // new loop on success or failure when the `forOp` is not supported.
 FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
-                                               scf::ForOp forOp);
+                                               scf::ForOp forOp,
+                                               unsigned numStages = 2);
 
 /// Insert barriers and wait operations if there are allocs of a different alias
 /// group before the given alloc.
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -293,43 +293,67 @@ populateOpToStageMap(LoopPrefetcher &prefetcher, scf::ForOp forOp,
   }
 }
 
-// Populates cluster IDs for each operation in the schedule.
-// Each operation gets a unique cluster ID based on its position in the
-// schedule.
-static void populateOpToClusterMap(
-    LoopPrefetcher &prefetcher,
-    const llvm::DenseMap<Operation *, unsigned> &opToStage,
-    std::vector<std::pair<Operation *, unsigned>> &finalSchedule,
-    llvm::DenseMap<Operation *, unsigned> &opToCluster) {
+// Populates cluster IDs for each operation based on stage groupings.
+// Operations are assigned cluster IDs in execution order: read -> compute ->
+// write. Each stage group gets its own cluster ID.
+static void
+populateOpToClusterMap(LoopPrefetcher &prefetcher,
+                       llvm::DenseMap<Operation *, unsigned> &opToCluster) {
 
   unsigned clusterID = 0;
 
-  // Build the schedule in desired execution order: read -> compute -> write
-  // Each operation forms its own "cluster" (a position in the schedule)
+  // Assign cluster ID 0 to all read stage operations
   for (Operation *op : prefetcher.readStage) {
-    if (opToStage.count(op)) {
-      finalSchedule.push_back({op, opToStage.lookup(op)});
-      opToCluster[op] = clusterID;
-    }
+    opToCluster[op] = clusterID;
   }
   ++clusterID;
 
+  // Assign cluster ID 1 to all compute stage operations
   for (Operation *op : prefetcher.computeStage) {
-    if (opToStage.count(op)) {
-      finalSchedule.push_back({op, opToStage.lookup(op)});
-      opToCluster[op] = clusterID;
-    }
+    opToCluster[op] = clusterID;
   }
   ++clusterID;
 
+  // Assign cluster ID 2 to all write stage operations
   for (Operation *op : prefetcher.writeStage) {
+    opToCluster[op] = clusterID;
+  }
+
+  LDBG() << "Built opToCluster map with " << clusterID + 1 << " clusters";
+}
+
+// Builds the final schedule using opToCluster and opToStage mappings.
+// Sorts operations by cluster ID, maintaining original order within each
+// cluster.
+static void buildFinalSchedule(
+    LoopPrefetcher &prefetcher,
+    const llvm::DenseMap<Operation *, unsigned> &opToStage,
+    const llvm::DenseMap<Operation *, unsigned> &opToCluster,
+    std::vector<std::pair<Operation *, unsigned>> &finalSchedule) {
+
+  // Collect all operations from all stages with their cluster IDs
+  SmallVector<Operation *> allOps;
+  allOps.append(prefetcher.readStage.begin(), prefetcher.readStage.end());
+  allOps.append(prefetcher.computeStage.begin(), prefetcher.computeStage.end());
+  allOps.append(prefetcher.writeStage.begin(), prefetcher.writeStage.end());
+
+  // Sort by cluster ID, maintaining original order within each cluster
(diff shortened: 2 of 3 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -26,6 +26,8 @@
 #include "mlir/IR/Visitors.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
```
qedawkins, line 28: nit:
```suggestion
```

compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir
```diff
@@ -18,7 +18,7 @@ func.func @prefetch_add(%arg0: memref<128xf32>) {
   // CHECK: vector.transfer_write %[[PRO_READ]], %[[SHARED]]
   // CHECK: %[[OUT:.*]] = scf.for %[[IV:.*]] = %[[C0]] to %[[C127]] step %[[C1]] iter_args(%[[ARG:.*]] = %[[CST]])
   %0 = scf.for %arg1 = %c0 to %c128 step %c1 iter_args(%arg2 = %cst) -> (vector<1xf32>) {
-    // CHECK-DAG: %[[IVPLUS1:.*]] = arith.addi %[[IV]], %[[C1]]  overflow<nsw> : index
+    // CHECK-DAG: %[[IVPLUS1:.*]] = arith.addi %[[IV]], %[[C1]] : index
```
qedawkins, line 21: Why did we drop this? We still want `overflow<nsw>` here.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -301,105 +234,292 @@ class LoopPrefetcher {
...
+      return;
+    opToStage[op] = stage;
+  };
 
-    for (Operation *op : writeStage) {
-      rewriter.clone(*op, mapping);
+  if (numStages == 1) {
+    // Single-stage pipelining: all ops in stage 0.
+    for (Operation &op : forOp.getBody()->without_terminator()) {
+      assignOp(&op, /*stage=*/0);
```
qedawkins: What is the point of everything getting stage 0? Isn't that a no-op?

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/LLVMGPUUtils.h
```diff
@@ -56,7 +56,8 @@ void packSharedMemoryAlloc(mlir::FunctionOpInterface funcOp);
 // Prefetches data written to shared memory for the next iteration. Returns the
 // new loop on success or failure when the `forOp` is not supported.
 FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
-                                               scf::ForOp forOp);
+                                               scf::ForOp forOp,
+                                               unsigned numStages = 2);
```
qedawkins, line 66: I don't follow what numStages mean in the context of prefetching (even after reading the code below). At a minimum document the meaning and valid values for it here.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -301,105 +234,292 @@ class LoopPrefetcher {
...
+  std::vector<std::pair<Operation *, unsigned>> finalSchedule;
+  finalSchedule.reserve(opToStage.size());
+  buildFinalSchedule(prefetcher, opToStage, opToCluster, finalSchedule);
+
+  if (finalSchedule.empty()) {
+    return failure();
+  }
+
+  // Dump the planned schedule before pipelining (while ops are still valid)
+  dumpSchedule(finalSchedule, opToCluster);
```
qedawkins: nit: Wrap in LLVM_DEBUG macro for clarity.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -301,105 +234,292 @@ class LoopPrefetcher {
...
+
+  scf::PipeliningOption options = buildPipeliningOption(finalSchedule);
+
+  FailureOr<scf::ForOp> newForOpOr = invokePipelineForLoop(forOp, options);
+  if (failed(newForOpOr))
```
qedawkins: nit: IREE style always adds braces even around one-liners.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -301,105 +234,292 @@ class LoopPrefetcher {
...
+      if (hasSharedMemory(readOp.getBase())) {
+        found = true;
+        return WalkResult::interrupt();
       }
+      return WalkResult::advance();
+    });
+    return found;
+  };
+
+  // Helper to check if operation or its nested ops have shared memory writes
```
qedawkins:
```suggestion
  // Helper to check if operation or its nested ops have shared memory writes.
```

qedawkins: (also above)

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -301,105 +234,292 @@ class LoopPrefetcher {
...
+    const std::vector<std::pair<Operation *, unsigned>> &finalSchedule) {
+  scf::PipeliningOption options;
+  options.getScheduleFn =
+      [finalSchedule](scf::ForOp loop,
+                      std::vector<std::pair<Operation *, unsigned>> &outSched) {
+        outSched = finalSchedule;
+      };
+
+  options.peelEpilogue = true;
+  options.supportDynamicLoops = false;
```
qedawkins, line 368: TODO for dynamic trip count support. We'll probably want to run the in range analysis to restrict to cases where we know K is large.

qedawkins (commented):
Cool! A few nits and confusion over the `numStages` option.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir
jerryyin: Do you know what nsw bring to the table? I'm always confused by that. I was assuming this is a "good to have" but not actually making any difference to the prefetched K loop.

BTW, his index addition was previous manually populated as in 

https://github.com/iree-org/iree/blob/6816a399f169db9ab0d30bc3cf5e743bc46fb033/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp#L100-L101

Now since we use upstream scf::pipelineForLoop(), we don't have direct control over this index addition since the upstream function take over this. If we want to change the index addition, and assuming this nsw flag indeed matters, I can submit an upstream PR for that.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Yes, this is more meant for completeness so we can generate the non-pipelined code if user gives a numStage = 0 to this pass. 

You are also right that plumb the `numStages == 1` case isn't that useful. I didn't pay very close attention to numStages since I was planning to use a future PR to just allow this to be configurable. 

For this PR, I can check for numStages earlier and bail from this pass when `numStages <= 1`.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/LLVMGPUUtils.h
jerryyin: `numStages` gives user the control over depth of prefetching/pipelining. This option is there because upstream pipeliner doesn't give us that option. The upstream pipeliner behavior is such that it will always prefetch one less stage amount of stages: So if we define 2 stages, it'd prefetch 1 stage; If we define 3 stages, it'd prefetch 2 stage. With the numStages option, we'd be defining the actual number of stages according to this user input and this allows it to be configurable - either we want shallow prefetch for large gemm or deep prefetch for smaller gemm.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Yah, I'll add a comment referencing your TODO.

We can potentially expose all those pipeline options as the pass options. I'd leave that to future work. It just made me realize how configurable pipelining should actually be versus our existing implementation.

kuhar:
I will review tomorrow

qedawkins (approved):
LGTM % the one nit I added on discord (fail on numStages > 2 if it's currently unimplemented).

compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir
krzysz00: `nsw` is the promise that an addition shall not overflow in the signed sense - that is, that `infinite_sext(x + y) == infinite_sext(x) + infinite_sext(y)` .

It's an optimization hint and makes it easier to generate good code, but shouldn't affect correctness.

PR #22526, opened 2025-11-03
[Torch] Disable deprecation declaration warnings when building torch-mlir-dialects

hanhanW:


benvanik (changes requested):
You'll need to guard this like this:
```
    if(MSVC)
      # MSVC: C4996 is the deprecation warning.
      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4996")
    else()
      # GCC/Clang: Use standard deprecation warning flag.
      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-declarations")
    endif()
```

hanhanW:
> You'll need to guard this like this:
> 
> ```
>     if(MSVC)
>       # MSVC: C4996 is the deprecation warning.
>       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4996")
>     else()
>       # GCC/Clang: Use standard deprecation warning flag.
>       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-declarations")
>     endif()
> ```

Ah thanks, I see why we have the branch in the other file now. I added the comments and applied the fix.

benvanik (approved):

PR #22531, opened 2025-11-03
[GPU] Clean up misc issues in IREEGPUAttrs. NFC.

kuhar:


compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -272,7 +268,6 @@ MMASingleSubgroupLayout getSingleSubgroupLayout(MMAIntrinsic intrinsic,
   };
   const MMASingleSubgroupLayout gfx12WmmaAcc16x16 = gfx12WmmaKx16(16);
 
-  // clang-format off
```
kuhar, line 275: FYI, there's no matching `clang-format on` so this doesn't do anything

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -35,9 +34,6 @@
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/Interfaces/LoopLikeInterface.h"
 
-#define DEBUG_TYPE "iree-gpu-attrs"
```
hanhanW, line 38: I thought it is required by LDBG print. It is no longer the case?

https://github.com/iree-org/iree/blob/497642b19668cc33a2a3377c3d7fa48c21d343d0/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp#L945

kuhar: Good catch, I removed one more line than I wanted, thanks!

hanhanW (approved):
LGTM, thanks!

PR #22533, opened 2025-11-03
[Codegen] Add WorkgroupCountHintOp to defer populating the workgroup count

qedawkins:
Currently ReconcileTranslationInfoPass is responsible for:
1. Comparing translation_info for consistency
2. Converting `scf.forall` with workgroup mapping to explicit uses of mapping ids.
3. Populating workgroup count regions based on the trip count of the resolved `scf.forall` op(s).

(3) requires the pass to nest at the variant level and precludes any other code paths from being involved with workgroup count resolution. This PR adds an operation + resolver pass for hinting at the workgroup count. In a follow up ReconcileTranslationInfoPass will be updated to generate these ops directly.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CMakeLists.txt
+++ b/compiler/src/iree/compiler/Codegen/Common/CMakeLists.txt
@@ -140,6 +140,7 @@ iree_cc_library(
     "ReplaceSlowMinMaxOps.cpp"
     "ReshapePatterns.cpp"
     "ResolveSwizzleHints.cpp"
+    "ResolveWorkgroupCountHints.cpp"
     "SpecializeExports.cpp"
     "SplitFullPartialTransferPass.cpp"
     "StripCompilationInfoPass.cpp"
--- a/compiler/src/iree/compiler/Codegen/Common/Passes.td
+++ b/compiler/src/iree/compiler/Codegen/Common/Passes.td
@@ -307,6 +307,11 @@ def ReconcileTranslationInfoPass
   ];
 }
 
+def ResolveWorkgroupCountHintsPass
+    : Pass<"iree-codegen-resolve-workgroup-count-hints", "IREE::HAL::ExecutableVariantOp"> {
+  let summary = "Materialize workgroup count hints in the workgroup count region per export.";
+}
+
 def ReplaceSlowMinMaxOpsPass
     : InterfacePass<"iree-codegen-replace-slow-min-max-ops", "mlir::FunctionOpInterface"> {
   let summary =
--- a/compiler/src/iree/compiler/Codegen/Common/Passes.td
+++ b/compiler/src/iree/compiler/Codegen/Common/Passes.td
@@ -310,6 +310,87 @@ def ReconcileTranslationInfoPass
 def ResolveWorkgroupCountHintsPass
     : Pass<"iree-codegen-resolve-workgroup-count-hints", "IREE::HAL::ExecutableVariantOp"> {
   let summary = "Materialize workgroup count hints in the workgroup count region per export.";
+  let description = [{
+    Resolves `iree_codegen.workgroup_count_hint` ops by materializing a globally
+    agreeable workgroup count per export. This pass performs a depth-first walk
+    of the target variant's call graph constructing the necessary information to
+    materialize workgroup count hints for that function.
+
+    #### Supported IR Structure
+
+    The pass expects executable variants containing exports with associated
+    entry point functions. Each entry point function (or functions it calls)
+    may contain `iree_codegen.workgroup_count_hint` ops:
+
+    ```mlir
+    hal.executable.export @entry_point
+        count(%device: !hal.device, %workload: index) -> (index, index, index) {
+      %x, %y, %z = iree_tensor_ext.dispatch.workgroup_count_from_slice %workload
+      hal.return %x, %y, %z : index, index, index
+    }
+    module {
+      func.func @entry_point(%device: !hal.device) {
+        %workload = iree_tensor_ext.dispatch.workload.ordinal %val, 0 : index
+        iree_codegen.workgroup_count_hint sizes(%workload)
+      }
+    }
+    ```
+
+    #### Supported Control Flow
+
+    The pass handles the following control flow patterns:
+
+    1. **Direct hints**: Workgroup count hints directly in entry point functions.
+
+    2. **Transitive function calls**: Hints in functions called (directly or
+       transitively) from the entry point. The backward slice of operands
+       required to compute the hint is tracked across call boundaries.
+
+    3. **Conditional hints (`scf.if`)**: When hints are wrapped in `scf.if`
+       operations, the conditions are preserved in the workgroup count region
+       if possible.
+       ```mlir
+       %o0 = iree_tensor_ext.dispatch.workload.ordinal ...
+       %o1 = iree_tensor_ext.dispatch.workload.ordinal ...
+       %cond = arith.cmpi ..., %o0, %o1
+       scf.if %cond {
+         iree_codegen.workgroup_count_hint sizes(%x, %y, %z)
+       } else {
+         iree_codegen.workgroup_count_hint sizes(%a, %b, %c)
+       }
+       ```
+       This generates an `scf.if` in the workgroup count region rather than
+       pessimistically taking the maximum.
+
+    #### Resolution of Multiple Hints
+
+    When multiple `workgroup_count_hint` ops contribute to the same export
+    (either directly or through function calls), the final workgroup count
+    is computed as the **elementwise maximum** across all hints:
+
+    ```mlir
+    // Given:
+    iree_codegen.workgroup_count_hint sizes(%x, %y, %z)
+    iree_codegen.workgroup_count_hint sizes(%a, %b, %c)
+
+    // Resolves to:
+    %wx = arith.maxsi %x, %a
+    %wy = arith.maxsi %y, %b
+    %wz = arith.maxsi %z, %c
+    hal.return %wx, %wy, %wz
+    ```
+
+    #### Backward Slice Requirements
+
+    The backward slice of hint operands must consist only of:
+    - `iree_tensor_ext.dispatch.workload.ordinal` ops (leaf values)
+    - Pure/side-effect-free operations
+    - Function arguments (for transitive call support)
+
+    Operations with memory effects or that depend on device-local values
+    (e.g., `hal.interface.*` ops, workgroup/thread IDs) result in a compiler
+    error for hint operands and omit conditionals for `scf.if` operands.
+  }];
 }
 
 def ReplaceSlowMinMaxOpsPass
--- a/compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
@@ -1,0 +1,1043 @@
+// Copyright 2025 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//=== ResolveWorkgroupCountHints.cpp -------------------------------------===//
+//
+// Resolves `iree_codegen.workgroup_count_hint` ops by materializing a globally
+// agreeable workgroup count per export. This pass performs a depth first walk
+// of the target variant's call graph constructing the necessary information to
+// materialize workgroup count hints for that function.
+//
+// Workgroup counts are materialized by again walking the callgraph starting
+// from each exported function and mapping the required operands to construct
+// the hints for a function using the operands of the callsite we're traversing
+// from.
+//
+// Additionally any conditional statements (scf.if ops) transitively wrapping
+// a workgroup count hint are tracked and materialized in the workgroup count
+// region as well. This avoids pessimizing the workgroup count for situations
+// where we are switching implementations on uniform host provided parameters
+// (workloads/device queries).
+//
+// TODO: Implement device query support.
+//
+//===---------------------------------------------------------------------===//
+
+#include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "iree/compiler/Dialect/HAL/IR/HALTraits.h"
+#include "iree/compiler/Dialect/TensorExt/IR/TensorExtOps.h"
+#include "mlir/Analysis/SliceAnalysis.h"
+#include "mlir/Analysis/TopologicalSortUtils.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Interfaces/FunctionInterfaces.h"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_RESOLVEWORKGROUPCOUNTHINTSPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+
+class ResolveWorkgroupCountHintsPass final
+    : public impl::ResolveWorkgroupCountHintsPassBase<
+          ResolveWorkgroupCountHintsPass> {
+public:
+  using Base::Base;
+  void runOnOperation() override;
+};
+
+//===---------------------------------------------------------------------===//
+// Utilities
+//===---------------------------------------------------------------------===//
+
+static Value negateValue(RewriterBase &rewriter, Location loc, Value v) {
+  auto one =
+      arith::ConstantIntOp::create(rewriter, loc, rewriter.getI1Type(), 1);
+  return arith::XOrIOp::create(rewriter, loc, v, one);
+}
+
+static Value andValues(RewriterBase &rewriter, Location loc, Value l, Value r) {
+  return arith::AndIOp::create(rewriter, loc, l, r);
+}
+
+static SmallVector<Value, 3> maxValueVectors(RewriterBase &rewriter,
+                                             Location loc, ArrayRef<Value> l,
+                                             ArrayRef<Value> r) {
+  SmallVector<Value, 3> max;
+  for (auto [lv, rv] : llvm::zip_equal(l, r)) {
+    max.push_back(arith::MaxSIOp::create(rewriter, loc, lv, rv));
+  }
+  return max;
+}
+
+using OperandWalkFn = std::function<LogicalResult(OpOperand &)>;
+static LogicalResult walkOperandsAndCapturesImpl(Operation *root,
+                                                 Operation *curr,
+                                                 DominanceInfo &dominance,
+                                                 OperandWalkFn callback) {
+  for (OpOperand &operand : curr->getOpOperands()) {
+    if (dominance.properlyDominates(operand.get(), root)) {
+      if (failed(callback(operand))) {
+        return failure();
+      }
+    }
+  }
+
+  for (Region &region : root->getRegions()) {
+    for (Block &body : region.getBlocks()) {
+      for (Operation &op : body.getOperations()) {
+        if (failed(
+                walkOperandsAndCapturesImpl(root, &op, dominance, callback))) {
+          return failure();
+        }
+      }
+    }
+  }
+  return success();
+}
+
+// Recursive walk of all operands and implicit captures of |root|. This
+// allows fetching all values transitively required by |root|. If
+// |skipCaptures| is set, this will skip lookup of implicit captures for
+// |root| only.
+static LogicalResult walkOperandsAndCaptures(Operation *root,
+                                             OperandWalkFn callback,
+                                             bool skipCaptures) {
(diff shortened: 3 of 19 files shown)


qedawkins:
Still adding a few more negative tests but implementation is ready for review.

compiler/src/iree/compiler/Codegen/Common/CMakeLists.txt
```diff
@@ -140,6 +140,7 @@ iree_cc_library(
     "ReplaceSlowMinMaxOps.cpp"
     "ReshapePatterns.cpp"
     "ResolveSwizzleHints.cpp"
+    "ResolveWorkgroupCountHints.cpp"
```
krzysz00, line 143: Tentative question - why not just do this in PropagateDispatchSizeBounds?

compiler/src/iree/compiler/Utils/ParseUtils.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2025 The IREE Authors
```
krzysz00: Unused?

krzysz00 (commented):
Very quick notes from a late skim

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
```diff
@@ -0,0 +1,891 @@
...
+    sizes.push_back(arith::ConstantIndexOp::create(rewriter, loc, 1));
+  }
+
+  if (placeholder) {
+    Operation *parent = placeholder->getParentOp();
+    SmallVector<Value, 3> newSizes = parent->getResults();
+    rewriter.replaceOp(placeholder, sizes);
+    sizes = newSizes;
+    rewriter.setInsertionPointAfter(parent);
+  }
```
Max191, line 721: Can you add a short comment to describe what the placeholder is here? There is a lot of function nesting in this implementation, so comments at each nesting level are helpful to understand the implementation top down without having to jump around too much.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
```diff
@@ -0,0 +1,891 @@
...
+    }
+    map.map(ordinalOp.getResult(), workloadVals[ordinal]);
+  }
+  for (auto op : slice) {
+    // TODO(#13038) This is a WAR for ops ending up in workgroup count
+    // computation. They should not. Some pre-processing at MaterializeEncoding
+    // time might make these go away.
+    if (isa<IREE::Codegen::QueryTileSizesOp>(op)) {
+      Value constVal =
+          arith::ConstantIndexOp::create(rewriter, op->getLoc(), 16);
```
Max191: Should it just return failure here? Why do we replace the ops with `16`?

Max191 (commented):
I'll come back to review this further later, but I think more comments would help the review process.

Please add comments to all the `process*` functions and any other static functions with non-trivial implementations. The logic takes some time to parse, and there is a lot of jumping around between function calls, so comments are important here.

compiler/src/iree/compiler/Utils/ParseUtils.h
qedawkins: oh oops this was a part of a previously combined change, good catch.

compiler/src/iree/compiler/Codegen/Common/CMakeLists.txt
qedawkins: Pass nest level is different. This has to be run at the variant level since it updates export ops. Also AFAICT PropagateDispatchSizeBounds is an optimization pass that can run many times. This pass must run only once.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
qedawkins: This matches the prior functionality (tied to the same TODO found in ReconcileTranslationInfo). I don't plan on fixing or changing that here.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
qedawkins: `placeholder` is explained here: https://github.com/iree-org/iree/pull/22533/files#diff-584eb0039c459e0e6b9f56e9744377d1596f8909ede161543d41cc8c8b002a8eR566-R569

I can add something here but generally overly repetitive comments hurt code readability IMO.

qedawkins:
Sorry about the extra ping, I screwed up my worktree :(

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -307,6 +307,11 @@ def ReconcileTranslationInfoPass
   ];
 }
 
+def ResolveWorkgroupCountHintsPass
+    : Pass<"iree-codegen-resolve-workgroup-count-hints", "IREE::HAL::ExecutableVariantOp"> {
+  let summary = "Materialize workgroup count hints in the workgroup count region per export.";
```
MaheshRavishankar, line 312: This needs a bit more description maybe to make clear what this pass does.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -493,4 +494,59 @@ def IREECodegen_InnerTiledOp : Op<IREECodegen_Dialect, "inner_tiled", [
...
+//===----------------------------------------------------------------------===//
+// HintWorkgroupCountOp
+//===----------------------------------------------------------------------===//
+
+def IREECodegen_WorkgroupCountHintOp :
+    Op<IREECodegen_Dialect, "workgroup_count_hint", []> {
+  let summary = [{Hints at the workgroup count to set}];
+
+  let description = [{
+    Captures a set of values to use as the workgroup count. The backward slice
```
MaheshRavishankar, line 507: I am leaving a comment as I read through it. It seems like there might be some impedence mismatch between these operations and the operations in `ha.executable.export`. The operations like `iree_tensor_ext.dispatch.workgroup_count_from_slice` effectively represent a "function" that takes some values as arguments and return the workgroup count. I'd expect these to be similar in vein so that the replacement of that operation in the export is more straight-forward. Maybe thats what it does, im still digesting the PR. Leaving this comment more for me to return later and adapt if needed.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -493,4 +494,59 @@ def IREECodegen_InnerTiledOp : Op<IREECodegen_Dialect, "inner_tiled", [
...
+        iree_codegen.workgroup_count_hint sizes(%c, %d, %e) maxDim(1)
+      }
+    }
+    ```
+
+    resolves to
+
+    ```
+    hal.executable.export @entry_point {
+      return max(%a * %b, %c), %d, %e
```
MaheshRavishankar: I am not immediately sure we want to take multiple hints, and "collapse" them. Also, I think it would be better to just force all the maxDims to be the "least" and then see how to resolve multiple hints. This resolution seems pretty non-intuitive.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -493,4 +494,59 @@ def IREECodegen_InnerTiledOp : Op<IREECodegen_Dialect, "inner_tiled", [
...
+
+    `maxDim` must be ∈ {0, 1, 2} = {x, y, z}.
+  }];
+
+  let arguments = (ins
+    Variadic<Index>:$sizes,
+    DenseI64ArrayAttr:$static_sizes
+  );
+  let results = (outs);
+  let hasCustomAssemblyFormat = 1;
```
MaheshRavishankar, line 557: You should be able to parse this using assembly format. The parse and print methods for dynamic index lists should help here : https://github.com/iree-org/llvm-project/blob/33d3ba67dd2145f9f4d6b78d91c7a9211153819e/mlir/include/mlir/Interfaces/ViewLikeInterface.h#L164

MaheshRavishankar (commented):
Pushing some comments. I still need to review the big pass added here, but some comments from other things,

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
```diff
@@ -0,0 +1,1043 @@
...
+    // Something went wrong with error catching during analysis or
+    // materialization.
+    return failure();
+  }
+  return results;
+}
+
+/// Replaces the given `workgroup_count_from_slice` op with the program slice
+/// described by the FunctionSlice |root|.
+static LogicalResult replaceFromSliceOpWithFunctionSlice(
```
MaheshRavishankar, line 893: Seems strange to me that this is taking the entire `sliceMap` as an argument. Also you probably want to pass that as reference.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
```diff
@@ -0,0 +1,1043 @@
...
+/// } else {
+///   iree_codegen.workgroup_count_hint(4, 5, 6)
+/// }
+/// ```
+///
+/// This will generate (effectively):
+///
+/// ```
+/// hal.executable.export count(%cond: i1)
+///   %x, %y, %z = scf.if %cond {
```
MaheshRavishankar, line 258: Can we actually generate such `scf.if` on the host side? This code effectively gets transplanted into the host-side computation. It might then affect how the command buffer is created.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
```diff
@@ -0,0 +1,1043 @@
...
+/// func.func @constant_hint() {
+///   iree_codegen.workgroup_count_hint 1, 2, 3
+/// }
+/// ```
+///
+/// This would resolve to:
+///
+/// ```
+/// hal.executable.export @entry_point count(%x, %y, %z, %a, %b, %c) {
+///   %wx = max(%x, %a, 1)
```
MaheshRavishankar, line 842: I am not sure this makes sense to me. Having max along each individual dimension seems very strange and very hard to reason about. The only logical thing I can think of is to linearize and then take the max. Everything else seems to be hard to navigate.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
qedawkins: I really wanted to use that but dynamic index lists don't populate the operand types. It lets you take a vector for types when parsing/printing, but that results in something like `(%0: index, %1: index, ...)` and I don't want the types inline. I still don't understand why setting the operand as `Variadic<Index>` doesn't automatically populate all operand types with index. Probably something I don't understand about custom parser/printers.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
qedawkins: My mistake, I need to update the description here. I removed `maxDim` from the op and made it the responsibility of whoever introduces the hints. So the above
```
iree_codegen.workgroup_count_hint sizes(%a, %b)
iree_codegen.workgroup_count_hint sizes(%c, %d, %e)
```
instead materializes to
```
return max(%a, %c), max(%b, %d), max(%e, 1)
```
i.e. per dimensions max of all the hints. Most of the time we'll just set linearized hints so it'll look something like
```
return max(%linearized0, %linearized1, ..., %linearizedn), %c1, %c1
```
which is exactly what happens in ReconcileTranslationInfo today.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
qedawkins: This is effectively what happens. The resolver pass implements this per exported function:
1. Look up all `iree_tensor_ext.dispatch.workload.ordinal` ops to tie to the inputs to the `iree_tensor_ext.dispatch.workgroup_count_from_slice` op.
2. Trace out the forward program slice of those ordinals and remove any IR that is not a part of computing an `iree_codegen.workgroup_count_hint`.
3. Clone that forward slice in place of the the `workgroup_count_from_slice` and take the maximum over all reachable `workgroup_count_hint` ops.

The implementation looks fairly different than this description (it captures backwards slices starting from the `workgroup_count_hint`) because it's much more efficient to implement the code that way (finding a forward slice and culling non-workgroup_count_hint stuff is the same as getting the backward slice).

I can update the description of this op + the resolver pass with this 1)2)3 if that helps.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
```diff
@@ -0,0 +1,1043 @@
...
+                                                        ValueRange workloadVals,
+                                                        HintSlice &hintSlice) {
+  Location loc = hintSlice.hintOp.getLoc();
+  // If an scf.if wrapping the cloned hint producers is generated, a placeholder
+  // `unrealized_conversion_cast` op is generated as a placeholder for the
+  // values returned by the cloned backwards slice described in |hintSlice|.
+  // The IR would look something like:
+  //
+  // scf.if %cond {
+  //   %x, %y, %z = builtin.unrealized_conversion_cast // no operands
```
MaheshRavishankar, line 677: Where do we get this from? Do we generate the unrealize_conversion_cast?

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
```diff
@@ -0,0 +1,1043 @@
...
+                      ValueRange workloadVals,
+                      std::vector<ConditionSlice> &conditions) {
+  Value acc;
+  for (ConditionSlice &condition : conditions) {
+    if (!llvm::all_of(condition.funcArgLeaves,
+                      [&](BlockArgument b) { return map.contains(b); })) {
+      continue;
+    }
+
+    if (failed(materializeSlice(rewriter, map, workloadVals, condition.slice,
```
MaheshRavishankar, line 618: Why reimplement this? There is already a method https://github.com/iree-org/iree/blob/2538af386c9400187899127584d475de65c16cf8/compiler/src/iree/compiler/Codegen/Transforms/Transforms.cpp#L425 that is supposed to do this. Why not use that?

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
MaheshRavishankar: I think taking individual max this way is ill-defined. I think it is just better off to error out, and only deal with max, when things are linearized.

compiler/src/iree/compiler/Codegen/Common/test/resolve_workgroup_count_hints.mlir
```diff
@@ -0,0 +1,454 @@
...
+        iree_codegen.workgroup_count_hint(4, 2, 6)
+        return
+      }
+    }
+  }
+}
+// CHECK-LABEL: hal.executable private @resolve_multiple_calls
+//       CHECK:   hal.executable.export public @entry_point
+//  CHECK-SAME:     layout({{.+}}) count(%{{.+}}: !hal.device) -> (index, index, index)
+//       CHECK:     hal.return %c4, %c5, %c6 : index, index, index
```
MaheshRavishankar, line 33: Echoing an earlier comment. This is taking max along each individual dimension, but that does not seem well-defined. I'd just error out in this case.

MaheshRavishankar (changes requested):
Ok, I left a few more comments, and I had to resort to using Claude to explain the code to me. I understand now what you are trying to do here. Kudos for trying to bite of the whole generality at once. Its great that you worked through some complex scenarios here.

At the same time though I am not sure some of the handling of the if-conditions and combining hints work out in general. I find it hard to work in abstract, and prefer to think of real use cases. I cant give you a counter-example, but some of the ways of handling conditions/combining different hints seem to be choices, and I would rather have waited for some use case to make that choice. Left one specific comment where the current heuristic does not seem as well-defined to me. For the rest of the stuff, what you have now is good to land and iterate on specific use cases if need be.

Broader question though
1. The existing ReconcileTranslationInfo pass supports generating code for different linearizations. Would that be handled? I am assuming that you would just change those to generate the hints and allow it to be that way.
2. I want to try to reuse as much as possible. If we can try to either use existing functionality or change them accodingly that would help. I understand the advantage of redoing, but if there are two place where the same thing is done, then if something is broken/fixed on one path, it will not port over. I left a specific comment about how the workgroup count region is materialized. We should either reuse the path we have now, ir if what you have is more general, just replace it. Its also OK to say that you will eventually replace it, but just need it to be clear that this is the plan.

Still marking as request changes, but really after that round this should be good to land! Overall, very nice work!

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
qedawkins: Right, so I left the linearization up to whoever introduces the workgroup count hint ops. That simplifies the logic in this pass a bit as well as leaves open the possibility of not linearizing if we want (without needing to rework this pass).

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
qedawkins: Yep, scf.if on the host side for non-device side stuff should work fine. This also gets transplanted after the command buffer is formed so it can't cause any issues.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
qedawkins: Transiently, it's just here to make the replacement logic easier because the builder for `scf.if` infers the result types from the values yielded from its body, so I need to put something there when I construct the if, but that happens before I clone over the actual workgroup count logic. The unrealized conversion cast acts as a placeholder and allows the IR to verify along the way which is nice.

compiler/src/iree/compiler/Codegen/Common/test/resolve_workgroup_count_hints.mlir
qedawkins: See my response in the earlier comment.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
qedawkins: Good catch on the reference. I could move everything to an analysis class and make the sliceMap a member, but then I'm passing around a `*this` pointer instead of the map pointer, and since that's the only variable I'd need on the class it doesn't seem worth it. If you think that structure would be better for readability I can do that though.

compiler/src/iree/compiler/Codegen/Common/ResolveWorkgroupCountHints.cpp
qedawkins: Made a shared utility for both. There wasn't much to share, but at least the TODO is in one place.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
qedawkins: Updated the op description.

qedawkins:
In re:

> I want to try to reuse as much as possible. If we can try to either use existing functionality or change them accodingly that would help. I understand the advantage of redoing, but if there are two place where the same thing is done, then if something is broken/fixed on one path, it will not port over. I left a specific comment about how the workgroup count region is materialized. We should either reuse the path we have now, ir if what you have is more general, just replace it. Its also OK to say that you will eventually replace it, but just need it to be clear that this is the plan.

I shared a utility on the one part you commented on and the rest of the ReconcileTranslationInfo overlap is replaced in the follow up: #22549

It's easier to do it as a follow up than to try to replace everything in one large PR so that's why I structured it as two changes.

> The existing ReconcileTranslationInfo pass supports generating code for different linearizations. Would that be handled? I am assuming that you would just change those to generate the hints and allow it to be that way.

Yep! The follow up just replaces all the logic that resolves workgroup counts directly from the loops to generate (linearized) workgroup count hints and it ends up working out pretty cleanly.

MaheshRavishankar (approved):
Ok, lets land and iterate on this. I think this pass is pretty complex (and its annoying that we have to roll out our own backward slice implementation, but I see why it is done).

PR #22535, opened 2025-11-03
Example of using HalModuleDebugSink to find numerical divergence 

newling:
Describes with tested code examples, the approach that I
used to find the dispatch in Llama 8b fp16 that introduced a 
numerical regression.


samples/sink_callback/example_1.py
```diff
@@ -0,0 +1,68 @@
...
+  %zero = arith.constant 0.0 : f32
+  %init = linalg.fill ins(%zero : f32) outs(%empty : tensor<?x16xf32>)
+               -> tensor<?x16xf32>
+  %mat = linalg.matmul ins(%arg0, %arg1 : tensor<?x16xf32>, tensor<16x16xf32>)
+                      outs(%init : tensor<?x16xf32>) -> tensor<?x16xf32>
+  %out = linalg.softmax dimension(1) ins(%mat : tensor<?x16xf32>)
+                        outs(%empty : tensor<?x16xf32>) -> tensor<?x16xf32>
+  return %out : tensor<?x16xf32>
+}
+"""
```
Groverkss: Can we move these to .mlir files outside of this python file?

samples/sink_callback/example_1.py
```diff
@@ -0,0 +1,68 @@
...
+        output_file="simple.vmfb",
+        extra_args=[
+            "--iree-hal-target-device=hip[0]",
+            "--iree-hip-target=gfx942",
+            "--iree-flow-trace-dispatch-tensors",  # <-- crucial to emit traces
+        ],
+    )
+
+    # Set up configuration + debug sink
+    config = rt.Config("hip")
```
Groverkss: Can we use cpu instead?

samples/sink_callback/example_1.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    # Load modules
+    with open("simple.vmfb", "rb") as f:
+        module = rt.VmModule.from_buffer(config.vm_instance, f.read())
+    vm_modules = rt.load_vm_modules(hal_module, module, config=config)
+
+    # Run function
+    A = np.random.rand(4, 16).astype(np.float32)
+    B = np.random.rand(16, 16).astype(np.float32)
```
Groverkss: Use seed to keep this deterministic

samples/sink_callback/example_1.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+
+
+def callback(key: str, buffer_views: list[rt.HalBufferView]):
+    print(f"--- Debug callback for key={key}  ---")
+    for i, bv in enumerate(buffer_views):
+        arr = bv.map().asarray(
+            bv.shape, rt.HalElementType.map_to_dtype(bv.element_type)
+        )
+        print(f"  Tensor {i}: shape={arr.shape}, mean={arr.mean():.6f}")
```
Groverkss: Instead of printing, maybe add asserts to what some expected values should be. We need to run this in ci to test if the samples work.

samples/sink_callback/example_2.py
```diff
@@ -0,0 +1,224 @@
...
+    model_b_mlir = model_a_mlir.replace("3.1415", "2.7182")
+
+    driver_mlir = r"""
+module @driver {
+  util.func private @foo.foo(%arg0: tensor<?x18xf32>, %arg1: tensor<18x18xf32>) -> tensor<?x18xf32>
+  util.func public @main(%arg0: tensor<?x18xf32>, %arg1: tensor<18x18xf32>) {
+    %0 = util.call @foo.foo(%arg0, %arg1) : (tensor<?x18xf32>, tensor<18x18xf32>) -> tensor<?x18xf32>
+    util.return
+  }
+}"""
```
Groverkss: Move these both to mlir files outside of this python file.

samples/sink_callback/example_1.py
```diff
@@ -0,0 +1,68 @@
...
+    # TODO(newling) better way to specify target generically.
+    compiler.compile_str(
+        MLIR_MODEL,
+        output_file="simple.vmfb",
+        extra_args=[
+            "--iree-hal-target-device=hip[0]",
+            "--iree-hip-target=gfx942",
+            "--iree-flow-trace-dispatch-tensors",  # <-- crucial to emit traces
+        ],
+    )
```
Groverkss: Use tempfile library to create a path for the file to ensure you are not messing up with other files

samples/sink_callback/example_2.py
```diff
@@ -0,0 +1,224 @@
...
+
+    artifacts_dir = Path(__file__).parent / "example_2_artifacts"
+    artifacts_dir.mkdir(exist_ok=True)
+
+    irpa_path = artifacts_dir / "archive.irpa"
+    a_vmfb_path = artifacts_dir / "model_a.vmfb"
+    b_vmfb_path = artifacts_dir / "model_b.vmfb"
+    driver_vmfb_path = artifacts_dir / "driver.vmfb"
+    output_a_log = artifacts_dir / "debug_output_a.txt"
+    output_b_log = artifacts_dir / "debug_output_b.txt"
```
Groverkss: Use templfile to get a temporary directory instead of this

Groverkss (changes requested):
Can we add some CMakeLists.txt and run this in CI just like other samples?

newling:
> Can we add some CMakeLists.txt and run this in CI just like other samples?

I completely missed the CMakeLists.txt file! Yes, will try to add to CI.

samples/sink_callback/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# Copyright 2022 The IREE Authors
```
kuhar:
```suggestion
# Copyright 2025 The IREE Authors
```

samples/sink_callback/model_for_async.mlir
```diff
@@ -0,0 +1,23 @@
+
+// softmax(matmul(arg0,weights)) + pi
```
kuhar:
```suggestion
// softmax(matmul(arg0,weights)) + pi
```

samples/sink_callback/README.md
```diff
@@ -0,0 +1,38 @@
+This directory contains examples to get developers started
+with using IREE's HalModuleDebugSink feature with callbacks
+to bisect layers of multi-dispatch model, to locate numerical
+divergences. It contains 2 examples:
+
+## Simple callback example: `sync_callback_log.py`
+
+A simple multi-dispatch example showing how to log statistics of
+all dispatch inputs and outputs. When a model is compiled with
+--iree-flow-trace-dispatch-tensors, IREE inserts trace points
```
kuhar:
```suggestion
`--iree-flow-trace-dispatch-tensors`, IREE inserts trace points
```

samples/sink_callback/sink_callback_log.py
```diff
@@ -0,0 +1,97 @@
...
+    # We assume that softmax(matmul(A, B)) compiled to 2 dispatches.
+    assert (
+        len(callback_results) == 5
+    ), "2 dispatches. mm:2->1, sm:1->1. Expected 5 tensors."
+
+    # A
+    # Example:
+    # ['main_dispatch_0::main_dispatch_0_matmul_Dx16x16_f32 inputs',
+    #                                                             0,
+    #                                                          16.0]
```
kuhar: The formatting is a bit weird here

kuhar (approved):

samples/sink_callback/sink_callback_log_async.py
```diff
@@ -0,0 +1,149 @@
...
+    # --- Compile both models (Model B differs only by a constant) ---
+    model_a_path = os.path.join(base_dir, "model_for_async.mlir")
+    vmfb_a = compiler.compile_file(
+        model_a_path,
+        target_backends=["vmvx"],
+        extra_args=["--iree-flow-trace-dispatch-tensors"],
+    )
+
+    with open(model_a_path) as f:
+        model_b_mlir = f.read().replace("3.1415", "2.7182")
```
Groverkss: Have this as a seperate file please, this can be hard to debug if some ir changes

samples/sink_callback/sink_callback_log_async.py
```diff
@@ -0,0 +1,149 @@
...
+
+def main():
+    config = rt.Config("local-sync")
+    base_dir = os.path.dirname(__file__)
+
+    # --- Compile both models (Model B differs only by a constant) ---
+    model_a_path = os.path.join(base_dir, "model_for_async.mlir")
+    vmfb_a = compiler.compile_file(
+        model_a_path,
+        target_backends=["vmvx"],
```
Groverkss: Why vmvx and not llvm-cpu?

Groverkss (approved):

samples/sink_callback/sink_callback_log_async.py
newling: Just replacing vmvx with llvm-cpu results in error:

```
Defaulting to targeting a generic CPU for the target architecture will result in poor performance. Please specify a target CPU and/or a target CPU feature set. If it is intended to target a generic CPU, specify "generic" as the CPU.
```

so it's a bit more complicated. Use of vmvx copied from https://github.com/iree-org/iree/blob/6816a399f169db9ab0d30bc3cf5e743bc46fb033/samples/py_custom_module/decode_secret_message.py#L100

Groverkss: If we can't use llvm-cpu then something is wrong. The erorr mentioned is a warning and you can fix it by adding `--iree-llvmcpu-target-cpu=host`.

I think we should get it working with llvm-cpu instead of vmvx.

Groverkss (changes requested):
Can we switch the tests to llvm-cpu instead of vmvx?

samples/sink_callback/sink_callback_log_async.py
newling: > I think we should get it working with llvm-cpu instead of vmvx.

Ok will do, can you say why? 

> The erorr mentioned is a warning and you can fix it by adding --iree-llvmcpu-target-cpu=host.

Yeah that's sort of what I meant by "it's a bit more complicated"

MaheshRavishankar: That is just a warning. I dont think it is an error. Without specifying the `--iree-llvmcpu-target-cpu=host` it does not use the right vector size, available ukernels on CPU, etc. It should still be correct as it will use a generic CPU backend. I think adding that flag is not a big deal.

newling: Ok, done!

Groverkss (approved):
Great work!

PR #22342, opened 2025-10-17
[GPU][Codegen] Expand iteration space based on new `expand_dims` attribute

efric:
This patch introduces iteration space expansion for reductions in the VectorDistribute path.

Specifically, we:

1. Add a new attribute, `expand_dims`, for reductions.
2. Introduce a new pass, `GPUExpandDimensions`, which uses `expand_dims` to expand the iteration space of relevant dimensions.
3. Refactor common functionality shared between `GPUExpandDimensions` and `BlockDynamicDimensions` into reusable utilities.
4. Refactor encoding helpers from `EncodingAttrs.cpp` into reusable utilities. 

This change also enables [chain FMA](https://github.com/iree-org/iree/pull/21855) in matvec codegen as we iterate along the K reduction dimension.  

---
**Performance Summary**

**IREE benchmark module**

* Only expansion: ~4% improvement
* Expansion + chain FMA: ~11% improvement

**rocprof**

* Only expansion: ~13% worse
* Expansion + chain FMA: ~9% better

**Register usage**
* 10% reduction (60 → 54 registers for matvec dispatches)

**Instruction latency (post-reduction loop epilogue)**
* 3.5% improvement (340 → 328 total mean latency)

---

**Notes**

* As a follow-up, we can explore applying iteration space expansion to the reduction in attention
* Right now, we only expand one dimension into two although the implementation supports expansion to N dimensions.
* Please note this PR changes the reduction order, some expect some minor changes to the numerics
* This is does not improve performance by itself/can cause regression without chain FMA https://github.com/iree-org/iree/pull/21855 

Traces for matvec dispatches are attached for all variations (original, only expansion, and expansion + chain FMA).

[115_expansion_and_chain.tar.gz](https://github.com/user-attachments/files/23268046/115_expansion_and_chain.tar.gz)
[115_nothing.tar.gz](https://github.com/user-attachments/files/23268047/115_nothing.tar.gz)
[115_only_expansion.tar.gz](https://github.com/user-attachments/files/23268048/115_only_expansion.tar.gz)

Fixes: #22153

ci-extra: test_torch

PR #22544, opened 2025-11-04
[NFC][Codegen] Extract dimension expansion helpers into common utilities

efric:
This patch extracts common functionality for dimension expansion into reusable utility functions in Transforms.cpp/h and updates the existing BlockDynamicDimensions pass to use them. This is in preparation for splitting up the PR in static dimension expansion (https://github.com/iree-org/iree/pull/22342).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
@@ -7,9 +7,16 @@
 #include "iree/compiler/Codegen/Common/Transforms.h"
 #include "iree/compiler/Codegen/Common/CombineLayoutTransformation.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
+#include "iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h"
+#include "iree/compiler/Dialect/Util/IR/UtilOps.h"
 #include "llvm/ADT/ScopeExit.h"
 #include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/Arith/Utils/Utils.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -728,4 +735,99 @@ void populateSwapExtractWithCollapsePattern(RewritePatternSet &patterns) {
   patterns.add<SwapCollapseShapeWithSlicePattern>(patterns.getContext());
 }
 
+namespace {
+
+struct RemoveOptimizationBarrier final
+    : public OpRewritePattern<IREE::Util::OptimizationBarrierOp> {
```
kuhar, line 741:
```suggestion
struct RemoveOptimizationBarrier final
    : OpRewritePattern<IREE::Util::OptimizationBarrierOp> {
```

kuhar (approved):
LGTM but wait for @Groverkss / @qedawkins before submitting

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -728,4 +735,99 @@ void populateSwapExtractWithCollapsePattern(RewritePatternSet &patterns) {
   patterns.add<SwapCollapseShapeWithSlicePattern>(patterns.getContext());
 }
 
+namespace {
+
+struct RemoveOptimizationBarrier final
```
qedawkins, line 740: Ouch, this is a bad pattern. We should look into removing it from that BlockDynamicDimensions pass in the future.

qedawkins (approved):

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
qedawkins: UnrealizedConversionCast would be a much better choice for how this is being used. That's exactly the intended use case of unrealized casts.

hanhanW:
What happens to this PR? Is it stale? Do we close it or land it?

efric:
It's stale; the relevant parts of this were merged as part of a larger PR #23076

PR #21855, opened 2025-09-05
[LLVMGPU][Codegen] Emit packed chain FMA from select multi_reductions and contracts

efric:
This patch teaches the vector lowering pipeline to:

1. Rewrite `vector.multi_reduction<add>` whose input is `arith.mulf` into a `vector.contract` via `vector::populateVectorReductionToContractPatterns`
2. Lower a restricted set of `vector.contract` into packed FMA chains.

Previously lowering `vector.multi_reduction` of the same form produced elementwise pack-muls per K-slice and then reduced them with a left-associated, serial chain of `v_add_f{16, 32}`

`(mul(a0 ,b0) + (mul(a1, b1) + … + acc`

The new lowering emits a single nested FMA chain and folds the accumulation into the `math.fma` c-operand

`fma(a0 ,b0, fma(a1, b1, fma(a2, b2, fma(a3, b3, acc))))`

To do this, we first permute the reduction and parallel dimensions of the `LHS` and `RHS` to the order of `[reduction, ..., parallel, ...]`. The `LHS` and `RHS` are then collapsed to a 2D shape of `{Π reduction dimensions, Π parallel dimensions}`. Then we form the FMA chain by iterating backwards, seeded by the accumulator. 

Not all forms of `vector.contract` are suitable in the current approach. For example, when an operand drops a parallel iterator as in matmul-like contracts. We require both sides to share the same 2D tuple. Unsupported cases fall back to the existing lowering. 

Fixes: #21483 (variant of original issue; for [issue #21513](https://github.com/iree-org/iree/issues/21513)).

PR #22545, opened 2025-11-04
[GPU][Codegen] Static iteration space expansion with `expand_dims` lowering config 

efric:
This is the meat of #22342, but with the NFC refactor parts removed. This is just for convenience in reviewing. 

---
This patch introduces iteration space expansion for reductions in the VectorDistribute path.

Specifically, we:

1. Add a new lowering config attribute, `expand_dims`, for reductions.
2. Introduce a new pass, `GPUExpandDimensions`, which uses `expand_dims` to expand the iteration space of relevant dimensions.
3. Refactor common functionality shared between `GPUExpandDimensions` and `BlockDynamicDimensions` into reusable utilities.

This change also enables [chain FMA](https://github.com/iree-org/iree/pull/21855) in matvec codegen as we iterate along the K reduction dimension.  

---
**Performance Summary**

**IREE benchmark module**

* Only expansion: ~4% improvement
* Expansion + chain FMA: ~11% improvement

**rocprof**

* Only expansion: ~13% worse
* Expansion + chain FMA: ~9% better

**Register usage**
* 10% reduction (60 → 54 registers for matvec dispatches)

**Instruction latency (post-reduction loop epilogue)**
* 3.5% improvement (340 → 328 total mean latency)

---

**Notes**

* As a follow-up, we can explore applying iteration space expansion to the reduction in attention
* Right now, we only expand one dimension into two. It is possible to extend the implementation to support expanding to N dimensions if this is useful. I abstained from doing this for now to keep things simple. 

Traces for matvec dispatches are attached for all variations (original, only expansion, and expansion + chain FMA).

[115_expansion_and_chain.tar.gz](https://github.com/user-attachments/files/23268046/115_expansion_and_chain.tar.gz)
[115_nothing.tar.gz](https://github.com/user-attachments/files/23268047/115_nothing.tar.gz)
[115_only_expansion.tar.gz](https://github.com/user-attachments/files/23268048/115_only_expansion.tar.gz)


efric:
Since we have been iterating on the feedback in the big PR https://github.com/iree-org/iree/pull/22342. Closing this one to avoid confusion.

PR #22549, opened 2025-11-04
[Codegen] Use workgroup_count_hint for most code paths

qedawkins:
This switches ReconcileTranslationInfo to generate
`iree_codegen.workgroup_count_hint` ops instead of directly updating the
workgroup count region.

Additionally this change exposed a significant number of stale tests
that had semantically unclear usage of `workgroup_count_from_dag_root`.
This was resulting in such tests silently failing to materialize the
workgroup count and falling back to an empty workgroup count region
(ultimately == {1, 1, 1}). In practice we always chose {1, 1, 1} in the
code for those tests because the shapes were so small, but it was pure
chance that the workgroup counts were lining up.

All affected tests have now been updated to use `from_slice`.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Transforms/Transforms.cpp
+++ b/compiler/src/iree/compiler/Codegen/Transforms/Transforms.cpp
@@ -512,6 +512,36 @@ LogicalResult lowerWorkgroupCountFromSliceOp(
                                         maxWorkgroupParallelDims);
 }
 
+LogicalResult createWorkgroupCountHint(RewriterBase &rewriter, Location loc,
+                                       ArrayRef<OpFoldResult> workgroupCount,
+                                       int maxWorkgroupParallelDims,
+                                       bool reverse) {
+  SmallVector<OpFoldResult> results(reverse ? llvm::reverse(workgroupCount)
+                                            : workgroupCount);
+  if (results.size() > maxWorkgroupParallelDims) {
+    MutableArrayRef<OpFoldResult> resultsRef =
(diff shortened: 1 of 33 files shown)


qedawkins:
Depends on #22533

compiler/src/iree/compiler/Codegen/Transforms/Transforms.cpp
```diff
@@ -512,6 +512,36 @@ LogicalResult lowerWorkgroupCountFromSliceOp(
                                         maxWorkgroupParallelDims);
 }
 
+LogicalResult createWorkgroupCountHint(RewriterBase &rewriter, Location loc,
+                                       ArrayRef<OpFoldResult> workgroupCount,
+                                       int maxWorkgroupParallelDims,
+                                       bool reverse) {
+  SmallVector<OpFoldResult> results(reverse ? llvm::reverse(workgroupCount)
+                                            : workgroupCount);
```
kuhar, line 520: fyi, there's a new helper for this: https://github.com/llvm/llvm-project/pull/171040 `reverse_conditionally` we can switch to that after the next integrate

kuhar (approved):

compiler/src/iree/compiler/Codegen/Transforms/Transforms.cpp
qedawkins: oh neat, this code pattern is awkward to write right now.

krzysz00 (approved):
No objection

PR #22552, opened 2025-11-04
[Codegen] Allow iree_codegen.swizzle_hint to operate on tensors

krzysz00:
This change will allow us to introduce swizzle hints during promotion when needed for performance (such as on MXFP4).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
@@ -3126,6 +3126,51 @@ func.func @transfer_gather(%source : tensor<?x64xf16>, %indices: vector<8xindex>
 
 // -----
 
+#pipeline_layout = #hal.pipeline.layout<bindings = [
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>
+]>
+func.func @swizzle_hint() {
+  %c0 = arith.constant 0 : index
+  %c4 = arith.constant 4 : index
+  %c64 = arith.constant 63 : index
+  %c512 = arith.constant 512 : index
+  %arg0 = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) alignment(64) offset(%c0)
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<1024xf32>>
+  %arg1 = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) alignment(64) offset(%c0)
+    : !iree_tensor_ext.dispatch.tensor<writeonly:tensor<1024xf32>>
+  %0 = iree_tensor_ext.dispatch.tensor.load %arg0, offsets=[0], sizes=[1024], strides=[1]
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<1024xf32>> -> tensor<1024xf32>
+  %temp = bufferization.alloc_tensor() : tensor<512xf32>
+  %swizzled.temp = iree_codegen.swizzle_hint %temp[#iree_codegen.rotate_rows<256, 4>] : tensor<512xf32>
+  %1 = scf.forall (%arg2) in (64) shared_outs(%arg3 = %swizzled.temp) -> tensor<512xf32> {
+    %off = arith.muli %arg2, %c4 : index
+    %slice = tensor.extract_slice %0[%off] [4] [1] : tensor<1024xf32> to tensor<4xf32>
+    scf.forall.in_parallel {
+      tensor.parallel_insert_slice %slice into %arg3[%off] [4] [1] : tensor<4xf32> into tensor<512xf32>
+    }
+  } {mapping = [#gpu.thread<linear_dim_0>]}
+  // Somewhat convoluted swap-things-around "kernel" to provide a minimal example
+  // showing that swizzle hints persist on a tensor.
+  scf.forall (%arg2) in (64) {
+    %arg2.reversed = arith.subi %c64, %arg2 : index
+    %off.rev = arith.muli %arg2.reversed, %c4 : index
+    %off = arith.muli %arg2, %c4 : index
+    %slice = tensor.extract_slice %1[%off.rev] [4] [1] : tensor<512xf32> to tensor<4xf32>
+    iree_tensor_ext.dispatch.tensor.store %slice, %arg1, offsets=[%off], sizes=[4], strides=[1]
+      : tensor<4xf32> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<1024xf32>>
+  } {mapping = [#gpu.thread<linear_dim_0>]}
+
+  return
+}
+
+// CHECK-LABEL: func.func @swizzle_hint
+// CHECK: %[[SHARED:.+]] = memref.alloc() : memref<512xf32>
+// CHECK: %[[SWIZZLED:.+]] = iree_codegen.swizzle_hint %[[SHARED]][{{.*}}] : memref<512xf32>
+// CHECK-COUNT-2: memref.subview %[[SWIZZLED]]
+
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -119,13 +119,13 @@ def IREECodegen_SwizzleHintOp : Op<IREECodegen_Dialect, "swizzle_hint", [
     SameOperandsAndResultType, Pure]> {
   let summary = [{Hint to swizzle accesses according to an access pattern.}];
   let description = [{
-    Optimization hint to swizzle all accesses to the memref this takes a view
-    of. This only affects reads/writes immediately consuming this operation and
-    is best effort. If the desired swizzling is not apparently possible, this
+    Optimization hint to swizzle all accesses to a memref or tensor this takes a
+    view of. This only affects reads/writes immediately consuming this operation
+    and is best effort. If the desired swizzling is not apparently possible, this
```
Max191:
```suggestion
    Optimization hint to swizzle all accesses to the memref or tensor that this takes a
    view of. This only affects reads/writes that immediately consume this operation
    and is best effort. If the desired swizzling is not apparently possible, this
```

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -119,13 +119,13 @@ def IREECodegen_SwizzleHintOp : Op<IREECodegen_Dialect, "swizzle_hint", [
...
     op will no-op. As a result, it should not be relied on for correctness.
 
-    Any subviews on this operation will cause swizzling to fail. The expectation
-    is for all view like operations to fold into the accessing ops
+    Any subviews on this operation will cause swizzle application to fail. The
+    expectation is for all view like operations to fold into the accessing ops
```
Max191:
```suggestion
    Any subviews on this operation will cause the swizzle application to fail. The
    expectation is for all view like operations to fold into the accessing ops
```

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -236,6 +234,46 @@ struct StoreToBufferOpInterface
...
+                          bufferization::BufferizationState &state) const {
+    auto hintOp = cast<IREE::Codegen::SwizzleHintOp>(op);
+    FailureOr<Value> maybeBuffer =
+        getBuffer(rewriter, hintOp.getOperand(), options, state);
+    if (failed(maybeBuffer))
+      return failure();
+    Value operandMemref = *maybeBuffer;
+
+    replaceOpWithNewBufferizedOp<IREE::Codegen::SwizzleHintOp>(
+        rewriter, op, operandMemref, hintOp.getSwizzle());
```
Max191:
```suggestion
    replaceOpWithNewBufferizedOp<IREE::Codegen::SwizzleHintOp>(
        rewriter, op, *maybeBuffer, hintOp.getSwizzle());
```

Max191 (approved):
LGTM, maybe wait to give a chance for @qedawkins or @kuhar to review before landing.

Also, will there ever be cases where we need to implement TilingInterface for this op (just wondering)?

krzysz00:
@Max191 Re tiling interface - I have *no* idea, I just know that this is a prerequisite change to whatever the future looks like, so I might as well make it its own PR

compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
```diff
@@ -3126,6 +3126,51 @@ func.func @transfer_gather(%source : tensor<?x64xf16>, %indices: vector<8xindex>
...
+  %c4 = arith.constant 4 : index
+  %c64 = arith.constant 63 : index
+  %c512 = arith.constant 512 : index
+  %arg0 = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) alignment(64) offset(%c0)
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<1024xf32>>
+  %arg1 = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) alignment(64) offset(%c0)
+    : !iree_tensor_ext.dispatch.tensor<writeonly:tensor<1024xf32>>
+  %0 = iree_tensor_ext.dispatch.tensor.load %arg0, offsets=[0], sizes=[1024], strides=[1]
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<1024xf32>> -> tensor<1024xf32>
+  %temp = bufferization.alloc_tensor() : tensor<512xf32>
```
kuhar: Could we make these function arguments?

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -236,6 +234,44 @@ struct StoreToBufferOpInterface
     return success();
   }
 };
+
+struct SwizzleHintOpInterface
+    : public BufferizableOpInterface::ExternalModel<
```
kuhar:
```suggestion
struct SwizzleHintOpInterface final
    : BufferizableOpInterface::ExternalModel<
```

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -236,6 +234,44 @@ struct StoreToBufferOpInterface
...
+          SwizzleHintOpInterface, IREE::Codegen::SwizzleHintOp> {
+  bool bufferizesToMemoryRead(Operation *op, OpOperand &opOperand,
+                              const AnalysisState &state) const {
+    return false;
+  }
```
kuhar:
```suggestion
  bool bufferizesToMemoryRead(Operation *, OpOperand &,
                              const AnalysisState &) const {
    return false;
  }
```
to prevent unused variable warnings. Also below

compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
krzysz00: Local style for these tests seems to be to use bindings everywhere?

kuhar: the test case above uses function arguments

krzysz00: Oh, yeah, I'll simplify it then

kuhar (approved):

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -236,6 +234,43 @@ struct StoreToBufferOpInterface
...
+    return {{op->getResult(0), BufferRelation::Equivalent, /*definite=*/true}};
+  }
+
+  LogicalResult bufferize(Operation *op, RewriterBase &rewriter,
+                          const BufferizationOptions &options,
+                          bufferization::BufferizationState &state) const {
+    auto hintOp = cast<IREE::Codegen::SwizzleHintOp>(op);
+    FailureOr<Value> maybeBuffer =
+        getBuffer(rewriter, hintOp.getOperand(), options, state);
+    if (failed(maybeBuffer))
```
qedawkins: nit: IREE style adds braces around one-liners.

qedawkins (approved):
nice, lgtm

qedawkins:
If you are in a situation where you're trying to tile this op you're likely on a bad path. I know why you want to do it and I don't have a solution yet but we can brainstorm something.

PR #22554, opened 2025-11-04
Using our own tablegen with depfile support.

benvanik:
Experimenting with this. Currently we are incorrect at head: changes to .td included by a target .td will not trigger rebuilds and build with old cached copies. The bazel side handles this by explicitly specifying deps but that doesn't quite work when lowered to cmake as LLVM's rules do a bunch of silly stuff to prevent it. The proper solution is available anyway, though: depfiles. Tablegen can produce them and they give us perfect deps. So here's an attempt at using them.


build_tools/cmake/iree_tablegen_doc.cmake
```diff
@@ -70,9 +77,30 @@ function(iree_tablegen_doc)
...
+    set(_OUTPUT_FILE ${CMAKE_CURRENT_BINARY_DIR}/${_OUTPUT})
+
+    # Create add_custom_command with depfile support for Ninja.
+    if(CMAKE_GENERATOR MATCHES "Ninja")
+      add_custom_command(
+        OUTPUT ${_OUTPUT_FILE}
+        COMMAND ${_TBLGEN_EXE} ${_COMMAND} ${_DIALECT} ${_INCLUDE_DIRS} ${_FLAGS} ${_TD_FILE_ABS}
+                --write-if-changed -o ${_OUTPUT_FILE} -d ${_OUTPUT_FILE}.d
+        DEPENDS ${_TBLGEN_TARGET} ${_TBLGEN_EXE} ${_TD_FILE_ABS}
+        DEPFILE ${_OUTPUT_FILE}.d
```
qedawkins, line 89: Is DEPFILE ninja only? I'm a complete noob at this build file stuff and don't know why this would be ninja only (or if we care about other generators).

qedawkins (approved):

build_tools/cmake/iree_tablegen_doc.cmake
benvanik: Mostly just because that's what LLVM does upstream - Makefiles should support DEPFILE, but I haven't tested them.

Issue #22555, opened 2025-11-05
Make IREE ubsan-warning-free


kuhar:
Recently we hit a few issues with nondeterminism caused by undefined behavior in the compiler (e.g., branching on uninitialized data members). These can often be caught by the undefined behavior sanitizer, but to make it useful, we should first make sure our baseline builds are ubsan-warning-free. Ideally, we should have a ubsan bulid in the CI.

I compiled IREE with asan + ubsan enabled (clang-20) and collected logs from `ninja iree-test-deps`. Because we have something like 1.5k inputs, this printed a whole lot of warnings that had to be summarized to make the logs actionable.

```
================================================================================
UBSAN Error Summary (Compact)
================================================================================
Total unique (file, error) combinations: 41
Total error occurrences: 151,902
================================================================================

Note: Addresses, function names, and type aliases normalized.
      Lines grouped by error type within each file.


================================================================================
/home/jakub/iree/iree/runtime/src/iree/base/internal/synchronization.c
================================================================================

  [74x] /home/jakub/iree/iree/runtime/src/iree/base/internal/synchronization.c
          call to function <name> through pointer to incorrect function type 'bool (*)(void *)'
          Lines: 823, 837


================================================================================
/home/jakub/iree/iree/runtime/src/iree/base/internal/threading_pthreads.c
================================================================================

  [37x] /home/jakub/iree/iree/runtime/src/iree/base/internal/threading_pthreads.c
          call to function <name> through pointer to incorrect function type 'int (*)(void *)'
          Lines: 119


================================================================================
/home/jakub/iree/iree/runtime/src/iree/hal/resource.h
================================================================================

  [37x] /home/jakub/iree/iree/runtime/src/iree/hal/resource.h
          call to function <name> through pointer to incorrect function type 'void (*)(struct iree_hal_resource_t *)'
          Lines: 90


================================================================================
/home/jakub/iree/iree/runtime/src/iree/hal/utils/deferred_command_buffer.c
================================================================================

  [37x] /home/jakub/iree/iree/runtime/src/iree/hal/utils/deferred_command_buffer.c
          call to function <name> through pointer to incorrect function type 'struct iree_status_handle_t *(*)(struct iree_hal_command_buffer_t *, struct iree_hal_buffer_binding_table_t, struct iree_hal_cmd_header_t *)'
          Lines: 809


================================================================================
/home/jakub/iree/iree/runtime/src/iree/vm/bytecode/dispatch.c
[...]

PR #22557, opened 2025-11-05
[runtime] Require aligned memory accesses by default

kuhar:
The C/C++ standards require all memory accesses to be aligned, even if the target hardware allows for unaligned accesses. Change the default to require alignment, so that we get closer to being UB-warning free when running with the UBSan enabled.

Advanced users can choose to disable this via the build system.

Also simplify the implementation of unaligned loads/stores and rely on `memcpy` being a compiler intrinsic.

Issue: https://github.com/iree-org/iree/issues/22555

ci-extra: test_torch, windows_x64_msvc

### DIFF
--- a/runtime/src/iree/base/alignment.h
+++ b/runtime/src/iree/base/alignment.h
@@ -309,19 +309,6 @@ static inline void iree_page_align_range(void* base_address,
 // Alignment-safe memory accesses
 //===----------------------------------------------------------------------===//
 
-// Map little-endian byte indices in memory to the host memory order indices.
-#if defined(IREE_ENDIANNESS_LITTLE)
-#define IREE_LE_IDX_1(i) (i)
-#define IREE_LE_IDX_2(i) (i)
-#define IREE_LE_IDX_4(i) (i)
-#define IREE_LE_IDX_8(i) (i)
-#else
-#define IREE_LE_IDX_1(i) (i)
-#define IREE_LE_IDX_2(i) (1 - (i))
-#define IREE_LE_IDX_4(i) (3 - (i))
-#define IREE_LE_IDX_8(i) (7 - (i))
-#endif  // IREE_ENDIANNESS_*
-
 #if IREE_MEMORY_ACCESS_ALIGNMENT_REQUIRED_8
 
 static inline uint8_t iree_unaligned_load_le_u8(const uint8_t* ptr) {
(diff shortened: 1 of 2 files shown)


bjacob (approved):

bjacob:
@benvanik , looking at the current unaligned_load implementations:

```c
static inline uint32_t iree_unaligned_load_le_u32(const uint32_t* ptr) {
  const uint8_t* p = (const uint8_t*)ptr;
  return ((uint32_t)p[IREE_LE_IDX_4(0)]) |
         ((uint32_t)p[IREE_LE_IDX_4(1)] << 8) |
         ((uint32_t)p[IREE_LE_IDX_4(2)] << 16) |
         ((uint32_t)p[IREE_LE_IDX_4(3)] << 24);
}
static inline float iree_unaligned_load_le_f32(const float* ptr) {
  uint32_t uint_value = iree_unaligned_load_le_u32((const uint32_t*)ptr);
  float value;
  memcpy(&value, &uint_value, sizeof(value));
  return value;
}
```

Since we don't support a BE target architecture anyway, we can just do this instead:

```c
static inline uint32_t simple_unaligned_load_u32(const void* ptr) {
  uint32_t value;
  memcpy(&value, ptr, sizeof value);
  return value;
}
```

An added benefit of `simple_unaligned_load_u32` is that it is easy for the compiler to compile to a single load instruction when the target load instruction supports an unaligned address. This should allow reconciling C++-correctness and performance.

amd-eochoalo (approved):

kuhar:
@bjacob done, I simplified the code a bunch. Let's see if the performance remains the same.

Groverkss (approved):

bjacob:
@kuhar , I was proposing this more for a follow-up - I'm personally OK with that change since I proposed it, but @benvanik might have thoughts. At least, the function names containing `_le_` need to be updated.  The `_le_` corresponded to the fact that these functions implemented LE loads regardless of native endianness.  The new implementation is native-endianness, which happens to be LE everywhere that IREE goes.  The difference is thus mostly just conceptual but aligning the function name with implementation is still important - as the `_le_` suggests that the implementation is more like what it was before this PR.

kuhar:
There's a todo and an `#error` pragna in the code to handle the BE code paths -- to me it seems like it was never supported, but it it becomes, I'd expect it to share the implementation and update the names.

bjacob:
I expect the opposite resolution: BE will never come back and we will just drop the _le_ from the name.

kuhar:
> I expect the opposite resolution: BE will never come back and we will just drop the _le_ from the name.

I looked around and I don't think we want to do that -- these functions are also used to load externally serialized data, e.g., in  https://github.com/iree-org/iree/blob/main/runtime/src/iree/io/formats/safetensors/safetensors_parser.c. If we support more file formats in the future, we may want to implement `_be_` variants even if all we support is LE hosts.

bjacob:
OK I'll admit that's a theoretical possibility (BE file formats).

Still, the `_le_` in the name is now un-descriptive as what the function's code actually does is native endianness, which just happens to be LE.  There is nothing about that function that is endianness-specific in anyway, so the `_le_` is unnecessary.

kuhar:
It loads/store LE data to/from the native endianness

PR #22562, opened 2025-11-05
[GPU] Add barrier before prologue when prefetching

### LINKED ISSUE #22468: Non-deterministic results for convolution with strided input slice
The following function gives non-deterministic results on CDNA3 (MI300x):

```mlir
#map = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d4, d1 + d5 * 2, d2 + d6 * 2, d3)>
#map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d4, d5, d6, d0)>
#map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d1, d2, d3)>
#map3 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
module @module {
  util.func public @conv_2d_bfloat16_weight_backward_12x118x182x448_nhwc_896x1x1x448_fhwc_nhwf_2x2s_0x0p_1x1d_1g(%0: tensor<12x59x91x896xbf16>, %1: tensor<12x118x182x448xbf16>) -> tensor<896x1x1x448xbf16> {
    %2 = tensor.empty() : tensor<896x1x1x448xf32>
    %3 = linalg.fill ins(%cst : f32) outs(%2 : tensor<896x1x1x448xf32>) -> tensor<896x1x1x448xf32>
    %4 = linalg.generic {indexing_maps = [#map, #map1, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel", "reduction", "reduction", "reduction"]} ins(%1, %0 : tensor<12x118x182x448xbf16>, tensor<12x59x91x896xbf16>) outs(%3 : tensor<896x1x1x448xf32>) {
    ^bb0(%in: bf16, %in_0: bf16, %out: f32):
      %9 = arith.extf %in : bf16 to f32
      %10 = arith.extf %in_0 : bf16 to f32
      %11 = arith.mulf %9, %10 : f32
      %12 = arith.addf %out, %11 : f32
      linalg.yield %12 : f32
    } -> tensor<896x1x1x448xf32>
    %5 = tensor.empty() : tensor<896x1x1x448xbf16>
    %6 = linalg.generic {indexing_maps = [#map3, #map3], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%4 : tensor<896x1x1x448xf32>) outs(%5 : tensor<896x1x1x448xbf16>) {
    ^bb0(%in: f32, %out: bf16):
      %9 = arith.truncf %in : f32 to bf16
      linalg.yield %9 : bf16
    } -> tensor<896x1x1x448xbf16>
    util.return %6 : tensor<896x1x1x448xbf16>
  }
}
```

### Compile Command ###

```
iree-compile --iree-hal-target-backends=rocm --iree-hip-target=mi300x --iree-opt-level=O3 repro.mlir -o repro.vmfb
```
### Benchmark Command ###

```
iree-run-module --device=hip://0 --module=repro.vmfb --function=conv_2d_bfloat16_weight_backward_12x118x182x448_nhwc_896x1x1x448_fhwc_nhwf_2x2s_0x0p_1x1d_1g --input=@in0.npy --input=@in1.npy --output=@out.bin
```

`in0.npy` and `in1.npy` can be generated with:

```py
from ml_dtypes import bfloat16
import numpy as np
in0 = np.array(np.random.random((12, 59, 91, 896)), dtype=bfloat16)
in1 = np.array(np.random.random((12, 118, 182, 448)), dtype=bfloat16)
np.save("in0.npy", in0)
np.save("in1.npy", in1)
```

Running the benchmark command multiple times yields differing outputs with each run. The number of differing elements is very small (roughly between 0-20 positions in the output tensor), but also non-deterministic.

nirvedhmeshram:
This is needed if the prefetched loop is a nested loop. Since the barrier is outside the main loop there should not be any noticeable performance impact on the more common unnested case.
Fixes : https://github.com/iree-org/iree/issues/22468

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -64,6 +64,9 @@ public:
   void emitPrologue(RewriterBase &rewriter) {
     Location loc = forOp.getLoc();
     Value zero = arith::ConstantIndexOp::create(rewriter, loc, lb);
+    // Barrier before starting the prologue in case the prefetching is happening
+    // inside a nested loop.
+    emitBarrier(loc, rewriter);
     // Directly write in the prologue and use the shared memory to communicate
     // data instead of the loop carried values. Read (0)
     emitRead(mapping[0], rewriter, zero);
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir
@@ -14,6 +14,7 @@ func.func @prefetch_add(%arg0: memref<128xf32>) {
   %c0 = arith.constant 0 : index
   // CHECK-DAG: %[[SHARED:.*]] = memref.alloc() : memref<1xf32, #gpu.address_space<workgroup>>
   %alloc = memref.alloc() : memref<1xf32, #gpu.address_space<workgroup>>
+  // CHECK-DAG: gpu.barrier
   // CHECK-DAG: %[[PRO_READ:.*]] = vector.transfer_read %[[GLOBAL]]
   // CHECK: vector.transfer_write %[[PRO_READ]], %[[SHARED]]
   // CHECK: %[[OUT:.*]] = scf.for %[[IV:.*]] = %[[C0]] to %[[C127]] step %[[C1]] iter_args(%[[ARG:.*]] = %[[CST]])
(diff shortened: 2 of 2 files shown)


jerryyin:
FYI I'm in middle of completely re-writing the prefetch pass: https://github.com/iree-org/iree/pull/22523. I think we should make this work in the rewritten code if possible. Sorry for the late notice.

The new branch use analysis to insert barriers and it seems to be making the right decisions from IR I've checked. Will your case work in that PR? In particular please check for function `insertPipelineBarriers()` and see if IR dumped before/after make sense to you.

nirvedhmeshram:
> FYI I'm in middle of completely re-writing the prefetch pass: #22523. I think we should make this work in the rewritten code if possible. Sorry for the late notice.
> 
> The new branch use analysis to insert barriers and it seems to be making the right decisions from IR I've checked. Will your case work in that PR? In particular please check for function `insertPipelineBarriers()` and see if IR dumped before/after make sense to you.

I tried your PR, and it doesnt seem to fix the race condition in this case, please see [this](https://gist.github.com/nirvedhmeshram/e11eae084664fa9cf117f6bd86287d62) and you will notice that we need a barrier before the prologue which I dont think gets added?

jerryyin:
I should be able to address it real quick. Then I may need your help in testing it.

I think the only problem is that this may insert redundant barriers that may not be necessary. So the right approach is probably to move the barrier insertion logic to a independent pass - This way the pass will have global view of when shared memory read and write happens. This will allow the barrier to be inserted correctly and precisely where they are needed.

nirvedhmeshram:
Sounds good, usually if the barriers are not in the main loop there is no noticable performance impact. But yeah refactoring the barrier insertion further makes sense as a follow up.

jerryyin:
Closing the PR as fix has landed.

PR #22563, opened 2025-11-05
[Integrate] Drop the revert of llvm/llvm-project@fcf79e5

### LINKED ISSUE #22171: Remove revert of https://github.com/llvm/llvm-project/pull/160615 on LLVM integrate

hanhanW:
It fixes `tileDispatchUsingSCFForOp` tiling decisions. When deciding skip trivial tiling dimensions, the function should                                                                                                consider the cases where workgroup size is larger than tile size.                                                                                                      In such cases, we should not simply fold the tiling as it is wrong.                                                                                                    

`tileDispatchUsingSCFForOp` does not have the way to get the workgroup size, so the fix is to plumb the skip decision into the function.

Carrying reverts:
- https://github.com/iree-org/llvm-project/commit/33d3ba67dd2145f9f4d6b78d91c7a9211153819e, asking questions in original PR https://github.com/llvm/llvm-project/pull/157330#issuecomment-3487738664

Closes https://github.com/iree-org/iree/issues/22171

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/TileDispatchUsingInterface.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/TileDispatchUsingInterface.cpp
@@ -11,6 +11,7 @@
 #include "iree/compiler/Codegen/Transforms/Transforms.h"
 #include "iree/compiler/Codegen/Utils/Utils.h"
 #include "iree/compiler/Dialect/TensorExt/IR/TensorExtOps.h"
+#include "llvm/ADT/SmallBitVector.h"
 #include "llvm/Support/Debug.h"
 #include "mlir/Analysis/TopologicalSortUtils.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
@@ -265,9 +266,10 @@ static LogicalResult replaceAllStoresWithTiledVersion(
   return success();
 }
 
-FailureOr<IREETilingResult>
-tileDispatchUsingSCFForOp(RewriterBase &rewriter, TilingInterface op,
-                          linalg::LinalgTilingOptions options) {
+FailureOr<IREETilingResult> tileDispatchUsingSCFForOp(
+    RewriterBase &rewriter, TilingInterface op,
+    linalg::LinalgTilingOptions options,
+    const std::optional<llvm::SmallBitVector> &skipTiledLoops) {
   OpBuilder::InsertionGuard guard(rewriter);
   rewriter.setInsertionPointAfter(op);
 
@@ -312,7 +314,10 @@ tileDispatchUsingSCFForOp(RewriterBase &rewriter, TilingInterface op,
         rewriter, loc, numTilesExprs,
         {range.offset, range.size, range.stride, tileSize});
     if (isOneInteger(numTiles)) {
-      continue;
+      if (!skipTiledLoops || skipTiledLoops->empty() ||
+          !skipTiledLoops->test(index)) {
+        continue;
+      }
     }
 
     tilingResult.tiledLoops.set(index);
--- a/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVTileAndDistribute.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVTileAndDistribute.cpp
@@ -46,22 +46,21 @@ namespace mlir::iree_compiler {
 // Invocation tiling utils
 //===----------------------------------------------------------------------===//
 
-/// Return the workgroup sizes from the export op.
+/// Returns the workgroup sizes from the export op. Returns empty list if it can
+/// not be found.
 static SmallVector<int64_t>
 getWorkgroupSizes(mlir::FunctionOpInterface funcOp) {
-  SmallVector<int64_t> workgroupSize;
   std::optional<IREE::HAL::ExecutableExportOp> exportOp = getEntryPoint(funcOp);
-  if (!exportOp)
-    return workgroupSize;
-
-  std::optional<ArrayAttr> workgroupSizeAttr = exportOp->getWorkgroupSize();
-  if (!workgroupSizeAttr)
-    return workgroupSize;
-
-  for (auto attr : workgroupSizeAttr.value()) {
-    workgroupSize.push_back(cast<IntegerAttr>(attr).getInt());
+  if (!exportOp) {
+    return {};
   }
-  return workgroupSize;
+  std::optional<ArrayAttr> workgroupSizeAttr = exportOp->getWorkgroupSize();
+  if (!workgroupSizeAttr) {
+    return {};
+  }
+  return llvm::map_to_vector(workgroupSizeAttr.value(), [](Attribute attr) {
+    return cast<IntegerAttr>(attr).getInt();
+  });
 }
 
 static llvm::SmallBitVector
@@ -69,15 +68,15 @@ computeSkipTileLoops(TilingInterface op, ArrayRef<int64_t> configTileSizes,
                      ArrayRef<int64_t> workgroupSize) {
   size_t numLoops = op.getLoopIteratorTypes().size();
   llvm::SmallBitVector notTiledLoops(numLoops, false);
-
   for (size_t index = 0; index < numLoops; ++index) {
-    if (index >= workgroupSize.size())
+    if (index >= workgroupSize.size()) {
       continue;
+    }
     int64_t tileSizeVal =
         index < configTileSizes.size() ? configTileSizes[index] : 0;
     if (tileSizeVal == 0)
       continue;
-    // workgroupSize and tileSizes are in opposite order.
+    // `workgroupSize` and `tileSizeVal` are in opposite order.
     if (tileSizeVal < workgroupSize[workgroupSize.size() - index - 1])
       notTiledLoops.set(index);
   }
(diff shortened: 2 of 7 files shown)


compiler/src/iree/compiler/Codegen/Common/TileDispatchUsingInterface.cpp
```diff
@@ -265,9 +266,10 @@ static LogicalResult replaceAllStoresWithTiledVersion(
   return success();
 }
 
-FailureOr<IREETilingResult>
-tileDispatchUsingSCFForOp(RewriterBase &rewriter, TilingInterface op,
-                          linalg::LinalgTilingOptions options) {
+FailureOr<IREETilingResult> tileDispatchUsingSCFForOp(
+    RewriterBase &rewriter, TilingInterface op,
+    linalg::LinalgTilingOptions options,
+    const std::optional<llvm::SmallBitVector> &skipTiledLoops) {
```
hanhanW, line 272: https://abseil.io/tips/163

@lialan I think this may copy the value, if someone passes `std::optional<...>`. I think what we want is `absl::Nullable<const Foo*>`, but I dont find similar things in LLVM or std. @kuhar maybe you know?

lialan: @hanhanW I quickly asked chatGPT: https://chatgpt.com/share/690b9a93-3230-800c-a6a4-e7f88c606716

I think it is somewhat correct, but is okay to do so.

krzysz00: Yeah, `std::optional<SmallBitVector&>` probably makes more sense

hanhanW: we can revisit. I think the main failure was lit tests, so we can land it without waiting all the jobs. I don't see any regression on the jobs that are already finished.

hanhanW: > Yeah, `std::optional<SmallBitVector&>` probably makes more sense

The article states that it doesn't compile.

hanhanW:
Converting to draft because I'm reimplementing the logic. I think we can get rid of the API changes.

compiler/src/iree/compiler/Codegen/Common/TileDispatchUsingInterface.cpp
kuhar: You can use an optional pointer

kuhar: or pass a function reference to control this instead of leaking the exact data type

compiler/src/iree/compiler/Codegen/SPIRV/SPIRVTileAndDistribute.cpp
```diff
@@ -43,9 +46,47 @@ namespace mlir::iree_compiler {
...
+    if (index >= workgroupSize.size()) {
+      continue;
+    }
+    int64_t tileSizeVal =
+        index < configTileSizes.size() ? configTileSizes[index] : 0;
+    if (tileSizeVal == 0)
+      continue;
+    // `workgroupSize` and `tileSizeVal` are in opposite order.
+    if (tileSizeVal < workgroupSize[workgroupSize.size() - index - 1])
+      notTiledLoops.set(index);
```
hanhanW, line 81: @lialan Is it correct? It looks like the workgroup_size is always the first level of tiling sizes, and they should always be greater than second level tiling sizes?

https://github.com/iree-org/iree/blob/45c05ca5e4a1977984e4e6a54530afc446fc8593/compiler/src/iree/compiler/Codegen/SPIRV/test/tile_and_distribute.mlir#L79-L89

I'm trying to set tile sizes to zero instead of passing `skipTileLoops`, and it turns out that I need to update few lit tests. Then I'm not confident about this change.

hanhanW (commented):
The change itself looks okay to me. I was thinking if we can move the `skip` logic to the compute function (e.g., set some tile sizes to zeros), but it

hanhanW:
I'll rebase this because I'm not sure if we can land it this way or not. Starting a new integrate in https://github.com/iree-org/iree/pull/22565 first.

compiler/src/iree/compiler/Codegen/SPIRV/SPIRVTileAndDistribute.cpp
lialan: >  It looks like the workgroup_size is always the first level of tiling sizes, and they should always be greater than second level tiling sizes?

Honestly, I don't know about this constraint. @hanhanW 

but I tried to use different workgroup sizes to invalidate the tests, and there are no complains.

compiler/src/iree/compiler/Codegen/Common/TileDispatchUsingInterface.cpp
MaheshRavishankar: I dont think having a `const std::optional<BitVector> &` is a good practice. Just make this `const BitVector &` (and see comment below).

PR #22566, opened 2025-11-05
Add passes to insert and remove barriers 2/2

IanWood1:
This change adds 2 passes to insert and remove tensor barrier ops.

- **InsertTensorBarriersPass**: Inserts barrier.start and barrier.end operations at the boundaries of computation regions (tensor/linalg/linalgext ops).  This identifies values flowing into and out of compute operations and
  wraps them with barrier operations.
- **RemoveTensorBarriersPass**: Removes barrier operations by replacing their results with their operands, since barriers are identity operations.

Depends on https://github.com/iree-org/iree/pull/22577


hanhanW (commented):
I think this PR should be split into several PRs or commits:

- Introduce new ops and add roundtrip tests. (can be together or one for each op)
- Implement the passes (can be in a single PR as they are all related.)

I can review the change, and it is not hard to me. However, one of my points is that it makes future contributor hard to track down the actual changes. They need to filter out the new ops first to read the change.

IanWood1:
@hanhanW I have split out the changes to add the ops and put it in https://github.com/iree-org/iree/pull/22577. I will now use this PR for just adding the passes. Let me know if you want me to make further changes the structure of the PRs.

compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
```diff
@@ -0,0 +1,102 @@
...
+        op->getOperand(0).getDefiningOp<IREE::TensorExt::BarrierStartOp>();
+    if (!barrierStartOp) {
+      return failure();
+    }
+
+    // Update reshape's operand to use barrier.start's input
+    rewriter.modifyOpInPlace(
+        op, [&]() { op->getOpOperand(0).set(barrierStartOp.getValue()); });
+
+    rewriter.setInsertionPointAfter(op);
```
MaheshRavishankar: Its better to use an insertion guard when changing the insertion point.

compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
```diff
@@ -0,0 +1,102 @@
...
+    Operation *reshapeOp = barrierEndOp.getValue().getDefiningOp();
+    if (!reshapeOp ||
+        !isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp>(reshapeOp)) {
+      return failure();
+    }
+    Value reshapeSrc = reshapeOp->getOperand(0);
+
+    // Create a new barrier.end before the reshape with the reshape's source
+    // type
+    rewriter.setInsertionPoint(reshapeOp);
```
MaheshRavishankar: Same here. Use a insertion guard.

compiler/src/iree/compiler/DispatchCreation/Passes.cpp
```diff
@@ -131,6 +131,7 @@ static void addDispatchRegionCreationPreprocessingPasses(
       // 2. Bubble up expand_shape ops (or sink collapse_shape ops) to get
       //    elementwise operation into higher dimensions for more fusion
       //    opportunities.
+      .addPass(DispatchCreation::createFoldReshapesIntoTensorBarriersPass)
```
MaheshRavishankar: Do we want to do this before? I think we want to slow walk adding the passes to the pipeline. Lets follow this
1) Add the passes to insert/and remove the barriers. These should be no-ops.
2) We can then insert the passes to propagate reshapes across the barriers.

IanWood1: Do you want me to remove the files related to folding reshapes from this PR and land it separately? Or just remove all uses from the pipeline?

IanWood1: I'm going to remove the reshape-related changes for now. This will also be useful for BOO's transpose issues, which don't require the reshape folding. I have backed up the branch to https://github.com/iree-org/iree/compare/main...IanWood1:iree:compute_boundary_ops_backup?expand=1

compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
IanWood1: I will fix this when landing this in a different PR.

MaheshRavishankar (approved):
Loks good. I should have pointed it out earlier, but maybe rename the ops to `compute_barrier` to signify that this isnt a real barrier.

compiler/src/iree/compiler/DispatchCreation/test/insert_tensor_barriers.mlir
```diff
@@ -0,0 +1,173 @@
...
+                     affine_map<(d0, d1) -> (d0, d1)>],
+    iterator_types = ["parallel", "parallel"]
+  } ins(%arg0 : tensor<4x8xf32>) outs(%empty : tensor<4x8xf32>) {
+  ^bb0(%in: f32, %out: f32):
+    linalg.yield %in : f32
+  } -> tensor<4x8xf32>
+  util.return %0 : tensor<4x8xf32>
+}
+// CHECK-LABEL: util.func public @simple_linalg
+//  CHECK-SAME:     %[[ARG0:[a-zA-Z0-9]+]]: tensor<4x8xf32>
```
hanhanW, line 16: nit: the benefits of using regexp is that we can avoid matching the types. The types do not matter in the test, so I suggest removing the checks.

IanWood1:
> Loks good. I should have pointed it out earlier, but maybe rename the ops to `compute_barrier` to signify that this isnt a real barrier.

Sounds good, that's a better name. I'll open a follow-up PR to change the name.

compiler/src/iree/compiler/DispatchCreation/InsertTensorBarriers.cpp
```diff
@@ -0,0 +1,144 @@
...
+    });
+  }
+}
+
+struct InsertTensorBarriersPass final
+    : public impl::InsertTensorBarriersPassBase<InsertTensorBarriersPass> {
+  using Base::Base;
+
+  void runOnOperation() override {
+    auto funcOp = getOperation();
```
hanhanW, line 86: style nit: spell out the type.

compiler/src/iree/compiler/DispatchCreation/InsertTensorBarriers.cpp
```diff
@@ -0,0 +1,144 @@
...
+      }
+      builder.setInsertionPointAfterValue(val);
+      auto startOp =
+          IREE::TensorExt::BarrierStartOp::create(builder, val.getLoc(), val);
+
+      val.replaceUsesWithIf(startOp.getResult(), [&](OpOperand &use) {
+        return isComputeOp(use.getOwner()) &&
+               !isa<tensor::DimOp>(use.getOwner());
+      });
+    }
```
hanhanW, line 108: nit: I'd remove the blank line because they belong to the same code block to me. I.e., I don't see benefits from having a blank line.

hanhanW (commented):
just few nits

IanWood1:
@hanhanW looks like I merged this while you were mid-review. I'll address your comments alongside the renaming of these ops.

hanhanW:
> @hanhanW looks like I merged this while you were mid-review. I'll address your comments alongside the renaming of these ops.

sgtm thanks!

PR #22567, opened 2025-11-05
[CI] Change numprocesses to 1 for amdgpu_vulkan_O0

hanhanW:
There is a kernel bug about using multiple threads. Using multiple threads can hit some kernel mode bug. Thus, we limit it to `1` for CI stability.

Failing sample, where the runner is `shark55-ci`: https://github.com/iree-org/iree/actions/runs/19114203732/job/54619458769


ScottTodd:
Oof, for Vulkan now too? Please track that down internally too, running two Vulkan (or ROCm) unit tests in parallel really shouldn't destabilize a machine.

hanhanW:
> Oof, for Vulkan now too? Please track that down internally too, running two Vulkan (or ROCm) unit tests in parallel really shouldn't destabilize a machine.

tag you on the internal thread.

kuhar (approved):

PR #22008, opened 2025-09-16
[Dispatch Creation] Fuse bcast with attention instead of producer

IanWood1:
This change waits to fuse broadcasts until after reshape propagation allowing broadcasts to fuse with consumers. This is important because it allows codegen to better infer the dims of the attention op. More generally, fusing a broadcast with consumer is better since it reduces the chance of materializing the broadcasted tensor.

Closes https://github.com/iree-org/iree/issues/22005

PR #22568, opened 2025-11-06
[DispatchCreation] Fuse residual broadcast producers after consumer fusion

### LINKED ISSUE #22500: [Dispatch Creation] Producers aren't fused due to consumer broadcast
Due to https://github.com/iree-org/iree/pull/22008 we don't fuse elementwise ops if the consumer is broadcasting the producer. This leads to the example below having 2 dispatches when it should only require 1.

```mlir
util.func public @test(%293: tensor<4xi64>, %__hoisted_tensor_64xf32_0: tensor<64xf32>) -> tensor<4x64xf32> {
  %cst_4 = arith.constant 4.000000e+00 : f32
  %306 = tensor.empty() : tensor<4xf32>
  %324 = tensor.empty() : tensor<4x64xf32>
  %325 = linalg.generic {indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>], iterator_types = ["parallel"]} ins(%293 : tensor<4xi64>) outs(%306 : tensor<4xf32>) {
  ^bb0(%in: i64, %out: f32):
    %1656 = arith.sitofp %in : i64 to f32
    linalg.yield %1656 : f32
  } -> tensor<4xf32>
  %326 = linalg.generic {indexing_maps = [affine_map<(d0, d1) -> (d1)>, affine_map<(d0, d1) -> (d0)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = ["parallel", "parallel"]} ins(%__hoisted_tensor_64xf32_0, %325 : tensor<64xf32>, tensor<4xf32>) outs(%324 : tensor<4x64xf32>) {
  ^bb0(%in: f32, %in_456: f32, %out: f32):
    %1656 = arith.mulf %in, %in_456 : f32
    %1657 = arith.addf %1656, %cst_4 : f32
    linalg.yield %1657 : f32
  } -> tensor<4x64xf32>
  util.return %326 : tensor<4x64xf32>
}
```

IREE @ b4ba1e8acfc062c7d7d2e4d518bffb517682227c

sakupan102:
After https://github.com/iree-org/iree/pull/22008 we don't fuse elementwise ops if the consumer is broadcasting the producer, which left some broadcast ops isolated. The new change detects those residual broadcast producers after the consumer fusion and folds them into the same dispatch.

Closes https://github.com/iree-org/iree/issues/22500

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/Passes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/Passes.cpp
@@ -139,10 +139,20 @@ static void addDispatchRegionCreationPreprocessingPasses(
                 /*fuseTruncateOps=*/clEnableEarlyTruncFusion,
                 /*fuseBroadcastOps=*/false});
       })
+      // 4. Perform elementwise operation fusion again. This is to fuse
+      // broadcast consumer ops that could not be fused earlier.
+      .addPass([]() {
+        return DispatchCreation::createElementwiseOpFusionPass(
+            ElementwiseOpFusionPassOptions{
+                /*intraDispatch=*/false,
+                /*fuseMultiReduction=*/clEnableElementWiseFuseMultiReduction,
+                /*fuseTruncateOps=*/clEnableEarlyTruncFusion,
+                /*fuseBroadcastOps=*/true});
+      })
       .addPass(IREE::Flow::createCanonicalizePass)
       .addPass(mlir::createCSEPass)
 
-      // 4. After elementwise operation fusion sink reshapes that block
+      // 5. After elementwise operation fusion sink reshapes that block
       //    producer-consumer fusion.
       .addPass(DispatchCreation::createSinkReshapesPass)
       .addPass(IREE::Flow::createCanonicalizePass)
--- a/compiler/src/iree/compiler/DispatchCreation/test/pipeline_tests.mlir
+++ b/compiler/src/iree/compiler/DispatchCreation/test/pipeline_tests.mlir
@@ -468,3 +468,44 @@ util.func @split_reduction_2d_by_tiling(%arg0 : tensor<?x2048x128xf32>) -> tenso
 //       CHECK:     iree_tensor_ext.dispatch.tensor.store %[[FORALL]]
 //       CHECK:   %[[DISPATCH1:.+]] = flow.dispatch.workgroups[%[[D0]]](%[[DISPATCH0]], %[[D0]])
 //       CHECK:   return %[[DISPATCH1]]
+
+// -----
+
+util.func @broadcast_consumer(%arg0 : tensor<4xf32>)
+    -> tensor<4x64xf32> {
+  %init_vec = tensor.empty() : tensor<4xf32>
+  %vec4 = linalg.generic
+      { indexing_maps = [affine_map<(d0) -> (d0)>,
+                          affine_map<(d0) -> (d0)>],
+        iterator_types = ["parallel"] }
+      ins(%arg0 : tensor<4xf32>) outs(%init_vec : tensor<4xf32>) {
+    ^bb0(%in: f32, %out: f32):
+      %cst2 = arith.constant 2.0 : f32
+      %cst05 = arith.constant 5.000000e-01 : f32
+      %mul = arith.mulf %in, %cst2 : f32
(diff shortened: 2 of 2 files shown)


MaheshRavishankar (commented):
Running the same pass twice is strange... @IanWood1 do you want to weigh in here?

hanhanW (commented):
I don't have much context, but running two passes back-to-back with slightly different config looks like an issue to me. Shouldn't we fix it in the pass?

compiler/src/iree/compiler/DispatchCreation/Passes.cpp
```diff
@@ -145,10 +139,20 @@ static void addDispatchRegionCreationPreprocessingPasses(
                 /*fuseTruncateOps=*/clEnableEarlyTruncFusion,
                 /*fuseBroadcastOps=*/false});
```
IanWood1, line 140: This might work too
```suggestion
                /*fuseBroadcastOps=*/true});
```

compiler/src/iree/compiler/DispatchCreation/test/pipeline_tests.mlir
```diff
@@ -468,3 +468,44 @@ util.func @split_reduction_2d_by_tiling(%arg0 : tensor<?x2048x128xf32>) -> tenso
...
+  } -> tensor<4xf32>
+
+  %init_mat = tensor.empty() : tensor<4x64xf32>
+  %broadcast = linalg.generic
+      { indexing_maps = [affine_map<(d0, d1) -> (d0)>,
+                          affine_map<(d0, d1) -> (d0, d1)>],
+        iterator_types = ["parallel", "parallel"] }
+      ins(%vec4 : tensor<4xf32>) outs(%init_mat : tensor<4x64xf32>) {
+    ^bb0(%in: f32, %out: f32):
+      linalg.yield %in : f32
```
IanWood1, line 497: It might be worth adding an operand to this `linalg.generic` and doing some sort of computation similar to https://github.com/iree-org/iree/issues/22500. Linalg ops with no computation might be treated differently.

IanWood1 (commented):
`ElementwiseOpFusionPassOptions`'s was initially intended to clean up the IR by fusing multiple simple linalg ops together. @MaheshRavishankar has noted in the past that this pass should not effect how dispatches are formed. Instead, it just makes the IR easier to read and reduces the number of ops the compiler has to deal with.

However, this pass now seems to be fairly load-bearing. The fusion decisions directly influence how dispatches are formed and which ops end up in which dispatch regions. We likely want to make this pass a bit smarter and have heuristics to determine if we want to fuse an op with its producer or consumer. This might be a more difficult task and will probably require some reworking of the pass. For example, here we would want to fuse the `%1` with `%0` and not `%2`:

```
%0 = producer
%1 = transpose %0
%2 = matmul_transpose_b %arg, %1
```

This is similar to the issue https://github.com/iree-org/iree/pull/22008 was trying to resolve where we want to prefer fusing broadcasts with attention ops (I don't think that PR had the best solution).

Also, note that this pass runs before propagating expand shapes. At this stage in the pipeline, the reshapes will obscure potential fusion opportunities, so should be conservative about which ops we fuse since we won't know what is optimal.



@MaheshRavishankar @hanhanW any thoughts on this?

hanhanW:
I'll leave the decision to you two. My point is that adding a pass to a pipeline to fix a problem is easy, but removing a pass from a pipeline is extremely hard because it may introduce some regressions. We should at least understand what it enables, but not just add a pass saying that it can fuse something cool.

compiler/src/iree/compiler/DispatchCreation/Passes.cpp
sakupan102: Isn’t this going to fuse the bcast with its producer rather than with linalg_ext.attention?

MaheshRavishankar:
I think we need to have separation of concerns. This pass cant do everything. It will need to have a fixed function. Looking back at the original PR, maybe a better solution would have been to move the patterns that fuses transposes with attention out of that pass and run that before elementwise op fusion. Then we would not need to do any changes to elementwise op fusion itself.

sakupan102:
I'll colse this PR and open another issue to separate functionality of `ElementwiseOpFusionPassOptions`

PR #22569, opened 2025-11-06
Adding new .td metadata classes and making our defs consistent.

benvanik:
These new metadata classes will be used by an iree-tblgen feature that allows for exporting rich IR metadata to JSON (where it can be queried, turned into markdown, or used by tooling).

All our dialects kind of scattered around their base classes, inconsistently used base classes at all (often replicating cppNamespace and dialect references), and were otherwise hard to consistently modify. This brings them all into a somewhat standard base form and then fixes the stream dialect to match util's layout to keep files smaller and makes it easier to find things (no soup of Base.td everything). I updated most of the main dialects except iree_linalg_ext, as it's doing something odd.


AWoloszyn (approved):

ScottTodd (approved):
If we customize `iree-tblgen` further we could also rework or remove https://github.com/iree-org/iree/blob/main/docs/website/postprocess_dialect_docs.py

PR #22571, opened 2025-11-06
[runtime] Fix incorrect alignment assumptions

kuhar:
Add assertions to check if the specified assumptions actually hold.

Issue: https://github.com/iree-org/iree/issues/22555

ci-extra: windows_x64_msvc


amd-eochoalo (approved):

runtime/src/iree/vm/stack.c
```diff
@@ -12,6 +12,7 @@
 #include <string.h>
 
 #include "iree/base/api.h"
+#include "iree/base/assert.h"
```
benvanik: Do not include iree/base/assert.h outside of iree/base/ - iree/base/api.h includes it already

runtime/src/iree/vm/bytecode/dispatch.c
```diff
@@ -126,6 +126,7 @@ static iree_status_t iree_vm_bytecode_function_enter(
       iree_host_align(ref_register_count * sizeof(iree_vm_ref_t), 16);
   iree_host_size_t frame_size =
       header_size + i32_register_size + ref_register_size;
+  IREE_ASSERT_ALIGNED(frame_size, 16);
```
benvanik: We should use `iree_max_align_t` instead of hardcoding 16 everywhere (I see some existing 16's, it'd be nice to fix those - if not here, then at least let's make the new ones that are easy to spot in this review right).

benvanik (changes requested):
Thanks for digging into this!

runtime/src/iree/vm/bytecode/dispatch.c
kuhar: `iree_max_align_t` is defined as `long double` -- that's less than 16 so it wouldn't catch the underalignment in IREE_BUILTIN_ASSUME_ALIGNED.

Are you saying that we should reduce the assumed alignment to iree_max_align_t?

kuhar: On windows it's 16

kuhar: If that's the case, the logic wouldn't need any changes -- we would only update the IREE_BUILTIN_ASSUME_ALIGNMENT checks

amd-eochoalo (approved):
I think it looks good to me but it looks like the tests for [Test ONNX / test_onnx_ops :: amdgpu_hip_rdna3_O3](https://github.com/iree-org/iree/actions/runs/19319159545/job/55257242205?pr=22571#logs) need to also be updated.

bjacob (approved):

benvanik (approved):

PR #22574, opened 2025-11-06
[build] Use CMake targets from upstream `flatcc` build

AaronStGeorge:
This PR changes `flatcc` CMake build to use upstream targets rather than creating new targets around source files inside the flatcc sub-module. Using upstream targets ensures that we could potentially use a system installed version of `flatcc` with `find_package`. The upstream `flatcc` targets are slightly larger than those built by bazel, but LTO should remove any unused code.

### DIFF
--- a/build_tools/third_party/flatcc/CMakeLists.txt
+++ b/build_tools/third_party/flatcc/CMakeLists.txt
@@ -4,138 +4,67 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-set(FLATCC_ROOT "${IREE_ROOT_DIR}/third_party/flatcc/")
+# ----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


benvanik (changes requested):
Supportive if we can do verification on what we build and install. This was very carefully crafted to reduce bloat/install garbage/work on bare metal systems. I don't implicitly trust that this has the same behavior and we'd need to inspect the outputs to ensure it does - or keep both and make this a user option to opt into to start (less ideal).

ScottTodd:
We discussed the integration points for IREE, fusilli, fusilli-plugin, and hipDNN today. Do you want to keep this PR open or abandon/re-work it? (Just checking on my review queue for what is still needing attention)

AaronStGeorge:
Given that we're going to just fetch the `flatcc` submodule as part of `TheRock` build, this change should not be needed. I'll close the PR.

PR #22575, opened 2025-11-06
[Codegen][GPU] Update heuristic to consider distribution from split reduction

yzhang93:
This PR updates the heuristic for deducing the MMA schedule to account for workload distributions introduced by split reduction. It should avoid decreasing the seed of `bestMNTileCountPerSubgroup` when sufficient parallelism is already introduced by split-reduction loops.

With this change, some shapes that previously showed performance regressions now achieve improved performance. For example, 

`convbfp16 -n 128 -c 384 -H 24 -W 48 -k 512 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC --iter 100`

- Without split reduction: 814us
- After split reduction: 1073us
- Split reduction with heuristic update: 398us

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -613,17 +613,17 @@ static int64_t adjustSeedsForWgpCount(const GPUMatmulShapeType &problem,
         bestMNTileCountPerSubgroup * intrinsic.mSizes[0] * intrinsic.nSizes[0];
     int64_t workgroupSize =
         mnTileSizePerSubgroup * bestSubgroupCountPerWorkgroup;
-    return mSize * nSize / workgroupSize;
+    int64_t numWorkgroups = mSize * nSize / workgroupSize;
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -618,6 +619,12 @@ static int64_t adjustSeedsForWgpCount(const GPUMatmulShapeType &problem,
   LDBG() << "Estimated number of workgroups: " << numWorkgroups
          << ", WGP count: " << wgpCount;
 
+  if (splitReductionTripCnt != 0) {
```
nirvedhmeshram: could you add a comment here on why we do this that is there in the PR description
something along the lines of.
```
It should avoid decreasing the seed of bestMNTileCountPerSubgroup when sufficient parallelism is already introduced by split-reduction loops.
```

nirvedhmeshram (approved):
LGTM

PR #22577, opened 2025-11-06
[TensorExt] Add barrier ops and roundtrip tests 1/2

IanWood1:
Add `iree_tensor_ext.barrier.start` and `iree_tensor_ext.barrier.end` operations to control where reshape operations can be propagated during compilation. These barriers prevent undesirable reshape movement, which is useful for controlling reshape propagation at the edges of the program where reshapes exist to convert between function ABI shapes and program dimensionality.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
@@ -483,13 +483,6 @@ def IREETensorExt_BarrierStartOp : IREETensorExt_PureOp<"barrier.start", [
     attr-dict-with-keyword
   }];
 
-  let builders = [
-    OpBuilder<(ins "Value":$value), [{
-      build($_builder, $_state, value.getType(), value,
-            IREE::Util::buildDynamicDimsForValue($_state.location, value, $_builder));
-    }]>,
-  ];
-
   let extraClassDeclaration = [{
     ValueRange getOperandDynamicDims(unsigned idx) { return getValueDims(); }
     ValueRange getResultDynamicDims(unsigned idx) { return getValueDims(); }
@@ -521,13 +514,6 @@ def IREETensorExt_BarrierEndOp : IREETensorExt_PureOp<"barrier.end", [
     type($result)
     attr-dict-with-keyword
   }];
-
-  let builders = [
-    OpBuilder<(ins "Value":$value), [{
-      build($_builder, $_state, value.getType(), value,
-            IREE::Util::buildDynamicDimsForValue($_state.location, value, $_builder));
-    }]>,
-  ];
 
   let extraClassDeclaration = [{
     ValueRange getOperandDynamicDims(unsigned idx) { return getValueDims(); }
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
@@ -455,4 +455,86 @@ def IREETensorExt_DispatchWorkloadOrdinalOp :
   let hasFolder = 1;
 }
 
+//===---------------------------------------------------------------------===//
+// Barrier Ops
+//===---------------------------------------------------------------------===//
+
+def IREETensorExt_BarrierStartOp : IREETensorExt_PureOp<"barrier.start", [
+  Util_ShapeAwareOp,
+]> {
+  let summary = "Tensor barrier start operation";
+  let description = [{
+    This barrier prevents certain transformations (like reshape propagation)
+    from moving operations down across this boundary.
+    The operand is returned as the result with identical type.
+  }];
+
+  let arguments = (ins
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -455,4 +455,86 @@ def IREETensorExt_DispatchWorkloadOrdinalOp :
...
+    OpBuilder<(ins "Value":$value), [{
+      build($_builder, $_state, value.getType(), value,
+            IREE::Util::buildDynamicDimsForValue($_state.location, value, $_builder));
+    }]>,
+  ];
+
+  let extraClassDeclaration = [{
+    ValueRange getOperandDynamicDims(unsigned idx) { return getValueDims(); }
+    ValueRange getResultDynamicDims(unsigned idx) { return getValueDims(); }
+  }];
```
hanhanW, line 521: unused?

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -455,4 +455,86 @@ def IREETensorExt_DispatchWorkloadOrdinalOp :
...
```
hanhanW, line 489: unused? I don't find it in this PR and the other PR.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -455,4 +455,86 @@ def IREETensorExt_DispatchWorkloadOrdinalOp :
...
+    type($result)
+    attr-dict-with-keyword
+  }];
+
+  let builders = [
+    OpBuilder<(ins "Value":$value), [{
+      build($_builder, $_state, value.getType(), value,
+            IREE::Util::buildDynamicDimsForValue($_state.location, value, $_builder));
+    }]>,
+  ];
```
hanhanW: I think it is better to add the builder with the other PR, which makes sure that they are not dead code.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -455,4 +455,86 @@ def IREETensorExt_DispatchWorkloadOrdinalOp :
...
```
hanhanW: ditto

hanhanW (commented):
let's add a few invalid tests to `invalid.mlir`: https://github.com/iree-org/iree/tree/main/compiler/src/iree/compiler/Dialect/TensorExt/IR/test

(you have to create the file)

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
IanWood1: It's used for the `Util_ShapeAwareOp` interface. I think this is important for canonicalizing tensor.dim ops for dynamic programs.

hanhanW: I see, thanks

hanhanW (commented):
Thanks. The PR itself looks good to me. Let me review the other one, just in case if we need to update the semantics.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -455,4 +455,72 @@ def IREETensorExt_DispatchWorkloadOrdinalOp :
...
+]> {
+  let summary = "Tensor barrier start operation";
+  let description = [{
+    This barrier prevents certain transformations (like reshape propagation)
+    from moving operations down across this boundary.
+    The operand is returned as the result with identical type.
+  }];
+
+  let arguments = (ins
+    AnyRankedTensor:$value,
```
MaheshRavishankar, line 473: To make the IR a bit more succinct did you want to take multiple operands and return multiple results such that every operand maps to the corresponding result? I know the dynamic dims will get complicated. Just a suggestion.

MaheshRavishankar (approved):

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.td
IanWood1: Is it possible to do this with tablegen's "assemblyFormat"? I don't see a way to split `value_dims` so that each argument has its own set of dynamic dims.

hanhanW: you'll need custom parser and printer in this case I think.

IanWood1: I'm going to keep this as is. I don't think the hassle is worth supporting multiple operands/results.

PR #22578, opened 2025-11-06
Revert "[CI] Change numprocesses to 1 for amdgpu_vulkan_O0"

hanhanW:
Reverts iree-org/iree#22567

I made a mistake when I'm debugging `Test ONNX / test_onnx_models :: amdgpu_vulkan` issue. The original change was for `Test OONX / test_onnx_ops :: amdgpu_vulkan_O0`. It reverts the change.

This is no longer needed, because:

- It updated wrong config.
- We just identify that the job passes on other runners, and the issue may be driver.

### DIFF
--- a/.github/workflows/pkgci_test_onnx.yml
+++ b/.github/workflows/pkgci_test_onnx.yml
@@ -48,7 +48,7 @@ jobs:
             config-file: onnx_ops_gpu_hip_rdna3_O3.json
             runs-on: [Linux, X64, gfx1100]
           - name: amdgpu_vulkan_O0
-            numprocesses: 1
+            numprocesses: 4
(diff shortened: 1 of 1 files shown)


kuhar (changes requested):
there are a bunch of known kernel driver bugs that trigger when running tests in parallel -- I think it's safer to keep as-is, even if it takes a bit longer to run

hanhanW:
> there are a bunch of known kernel driver bugs that trigger when running tests in parallel -- I think it's safer to keep as-is, even if it takes a bit longer to run

We can keep it as-is, but I'd like to point out that I don't see any issue on this job in the past.

kuhar:
it won't show in one run but will break faster over time until it won't detect the gpu

PR #22580, opened 2025-11-06
[CI] Force amdgpu_vulkan runner be shark10-ci

hanhanW:
Updated the runner configuration for amdgpu_vulkan to include the shark10-ci label due to Vulkan driver issues. This forces the job to run on shark10-ci only.


hanhanW:
I checked that only shark10-ci has the label: https://github.com/iree-org/iree/actions/runners?tab=self-hosted

MaheshRavishankar (approved):

kuhar (approved):

amd-eochoalo:
I just saw this, I made the same change in the torch CI here: https://github.com/iree-org/iree/pull/22588

PR #22581, opened 2025-11-06
[LinalgExt] Add iree_linalg_ext.unmask

Groverkss:
This patch adds iree_linalg_ext.unmask operation to represent a "masked copy" to a destination tensor. This operation is very similar to a tensor.extract_slice, but explicitly gives vectorization/bufferization a hint that this operation should always be thought of as a copy.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
@@ -1385,6 +1385,36 @@ def IREELinalgExt_Im2colOp : IREELinalgExt_Op<"im2col",
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// Unmask
+//===----------------------------------------------------------------------===//
+
+def IREELinalgExt_UnMaskOp : IREELinalgExt_PureOp<"unmask", [
+  AllTypesMatch<["dest", "result"]>,
+  Pure,
+  DestinationStyleOpInterface
+  ]> {
+  let summary = "";
+  let description = [{}];
+  let arguments = (ins AnyRankedTensor:$src,
+                       AnyRankedTensor:$dest);
+  let results = (outs AnyShaped:$result);
+  let assemblyFormat = [{
+    attr-dict
+    $src `into` $dest `:` type($src) `->` type($result)
+  }];
+  let hasVerifier = 1;
+  let hasFolder = 1;
+
+  let extraClassDeclaration = [{
+    // Method to implement for specifying output range for
+    // DestinationStyleOpInterface
+    MutableOperandRange getDpsInitsMutable() {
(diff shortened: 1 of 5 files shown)


Groverkss:
More discussion: https://discord.com/channels/689900678990135345/1436113909151957082

Max191 (commented):
Small naming nit: If this is essentially an inverse of tensor.pad (plus some hints for vectorization/bufferization), then can we call it `unpad`?

Groverkss:
> Small naming nit: If this is essentially an inverse of tensor.pad (plus some hints for vectorization/bufferization), then can we call it `unpad`?

I'm not sure sure about "unpad", "pad" pads a tensor with a value. masking is essentially padding with poison. The correct name for ignoring some part of the computation with be "unmasking" the tensor, because you are treating the other values as poison. (Also I would have to do a bit of rebasing, I can do it, but unless this is a strong concern, I'd rather not.)

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/roundtrip.mlir
```diff
@@ -2186,3 +2186,12 @@ func.func @split_reduction_mapping(%arg0 : index,
 //  CHECK-SAME:       #iree_linalg_ext.split_reduction_mapping<1>,
 //  CHECK-SAME:       #iree_linalg_ext.split_reduction_mapping<0>,
 //  CHECK-SAME:       #iree_linalg_ext.split_reduction_mapping<2>]
+
+// -----
```
IanWood1, line 2190: Could you add a fully static case?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -1385,6 +1385,36 @@ def IREELinalgExt_Im2colOp : IREELinalgExt_Op<"im2col",
...
+// Unmask
+//===----------------------------------------------------------------------===//
+
+def IREELinalgExt_UnMaskOp : IREELinalgExt_PureOp<"unmask", [
+  AllTypesMatch<["dest", "result"]>,
+  Pure,
+  DestinationStyleOpInterface
+  ]> {
+  let summary = "";
+  let description = [{}];
```
IanWood1, line 1398: Could you add a description of which elements are dropped? From what I can tell, this is similar to an extract slice with zero offsets and sizes equal to `dest`.

MaheshRavishankar (commented):
I think it is useful to capture the dynamic dimensions of the source and destination as well in the operation. That has been generally useful.

Groverkss:
> I think it is useful to capture the dynamic dimensions of the source and destination as well in the operation. That has been generally useful.

I think it does capture it. This is a DPS operation, so the destination dimensions are captured as an input.

Issue #22485, opened 2025-10-30
RFC: Implement the pass that unifies the encodings for each global.


hanhanW:
# UnifyEncodingForGlobals Pass - Implementation Plan

## Overview
This pass unifies multiple encoded versions of the same immutable global to reduce memory footprint in IREE's Stream dialect. When a single source global (e.g., a model weight) is encoded multiple times with different encodings for different uses, this pass selects a unified encoding and updates all references.

Initial Prototype: https://github.com/hanhanW/iree/blob/hanhan-prototype-globals-with-multi-encoding-snapshot-20251022/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp

# UnifyEncodingForGlobals Pass - Design Rationale

Below explains the design rationale for the UnifyEncodingForGlobals pass and why it meets IREE's bar as a retargetable compiler infrastructure.

---

## 1. Problem Statement

When the same source data (e.g., model weights) is used in multiple dispatch operations with different encoding requirements, IREE currently creates multiple encoded copies of the same data. For example, in LLaMA models:

```mlir
// Same weight encoded twice with different iteration_sizes
%enc1 = stream.tensor.encode %weight -> tensor<4096x4096xf32, #encoding<iteration_sizes=[?, 4096, 4096]>>
%enc2 = stream.tensor.encode %weight -> tensor<4096x4096xf32, #encoding<iteration_sizes=[4, 4096, 4096]>>
```

This wastes memory because it may produce 2x encoded globals during execution,
if they result in different layouts.

**Goal**: Unify multiple encodings of the same immutable source into a single encoding, reducing memory footprint.

---

## 2. Why Immutability Enables This Optimization

The foundation of this optimization rests on **immutable globals**:

1. **Immutability Guarantee**: IREE's `VerifyInitializationOrderPass` ensures immutable globals are initialized exactly once, only in initializers or initializer-only functions.

2. **Deterministic Data**: Since immutable globals never change after initialization, any computation derived purely from immutable globals produces deterministic results.

3. **Safe Unification**: If two `stream.tensor.encode` ops consume the same immutable source, we can safely use a unified encoding for both - the underlying data is guaranteed identical, which avoid memory footprint bloat.

This is why the pass strictly requires both source AND encoded globals to be immutable.

---

### 3.1 The Resolver Architecture

IREE uses an **encoding resolver** infrastructure that is central to retargetability:

```
[...]

PR #22582, opened 2025-11-07
[DT] Add support for layout transfer in MaterializeEncoding pass.

### LINKED ISSUE #19896: Data-Tiling: Missing layout transfer support in materialization patterns
https://github.com/iree-org/iree/commit/561ae02174b36cdd1a815d1685d62469a6a29886 teaches the materialization pass to lower binding ops with resolved layouts, if any. However, the layout transfer is not done yet. There could be a mismatch between "resolved layout" and "execution layout" in load/store ops. The resolved layout indicates the layout that is resolved during encoding specialization. The execution layout indicates the actual layout we prefer on the executable target.

E.g.,

```mlir
#encoding = #iree_encoding.encoding<operand_index = 2 : index, op_type =  matmul, element_types = [f32, f32, f32],
  layouts = [#iree_cpu.vmvx_encoding_layout<configuration = {encoding_info = {innerDimsPos = [0, 1], innerTileSizes = [A, B], outerDimsPerm = [0, 1]}}>]>
#executable_target = #hal.executable.target<"llvm-cpu", "xyz", {cpu_features = "+avx512f", encoding = #iree_cpu.cpu_encoding_layout<>, target_triple = "x86_64-xyz-xyz"}>
#map = affine_map<(d0, d1, d2) -> (d0, d2)>
#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
#pipeline_layout = #hal.pipeline.layout<bindings = [#hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, Indirect>], flags = Indirect>
#encoding1 = #iree_encoding.encoding<operand_index = 2 : index, op_type =  matmul, element_types = [f32, f32, f32], user_indexing_maps = [#map, #map1, #map2], round_dims_to = array<i64: 1, 32, 32>>
func.func @unset_encoding_RES_with_layout() attributes {
  hal.executable.target = #executable_target
} {
  %0 = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) alignment(64) offset(%c0) flags("ReadOnly|Indirect") : !flow.dispatch.tensor<readonly:tensor<1x10xf32, #encoding>>
  %1 = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) alignment(64) offset(%c0) flags(Indirect) : !flow.dispatch.tensor<writeonly:tensor<1x10xf32>>
  %2 = flow.dispatch.tensor.load %0, offsets = [0, 0], sizes = [1, 10], strides = [1, 1] : !flow.dispatch.tensor<readonly:tensor<1x10xf32, #encoding>> -> tensor<1x10xf32, #encoding1>
  %3 = iree_encoding.unset_encoding %2 : tensor<1x10xf32, #encoding1> -> tensor<1x10xf32>
  flow.dispatch.tensor.store %3, %1, offsets = [0, 0], sizes = [1, 10], strides = [1, 1] : tensor<1x10xf32> -> !flow.dispatch.tensor<writeonly:tensor<1x10xf32>>
  return
}
```

Looking at the `%2` flow.dispatch.tensor.load operation. It loads a `tensor<1x10xf32, #encoding1>` from `!flow.dispatch.tensor<readonly:tensor<1x10xf32, #encoding>>` type. The encoding information of `tensor<1x10xf32, #encoding1>` is derived from the executable target (i.e., llvm-cpu in the example). It could be resolved into `{innerDimsPos = [0, 1], innerTileSizes = [1, 16], outerDimsPerm = [0, 1]}` layout. However, the layout can mismatch with the incoming layout (which has innerTileSizes = [A, B]). The truth is that the attribute attached in the encoding knows the details. We should be able to implement interface methods (e.g., bringToGlobalLayout()) from the attached attribute, and implement the other interface methods (e.g., bringToTiledLayout()) from the encoding attribute within the executable target. This is how we materialize the layout transfer in codegen. Codegen might not be able to handle the case or it could be very inefficient, but this is a requirement for functionality.

cc @benvanik  @MaheshRavishankar @bjacob @Max191

hanhanW:
The revision adds another phase before materialization that breaks layout mismatched load/store ops into the matched load/store op plus encoding ops.

The new phase is needed because the no-rollback conversion driver requires the intermediate IR being legal. We can update the code to generate the final result, but it makes codebase complicated. Running an additional phase that uses cheap `walkAndApplyPatterns` driver is better.

Fixes https://github.com/iree-org/iree/issues/19896

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
@@ -282,6 +282,147 @@ struct MaterializeTensorExtDispatchTensorStoreOp
 };
 
 //===---------------------------------------------------------------------===//
+// Patterns for layout transfers. They decompse load/store ops into
+// set_encoding/unset_encoding + load/store, if the converted types mismatch.
+//===---------------------------------------------------------------------===//
+
+/// Returns the value that brings `src` to `destType` by inserting the necessary
+/// encoding ops.
+static Value generateEncodingTransferOps(RewriterBase &rewriter, Value src,
+                                         ArrayRef<Value> dynamicDims,
+                                         RankedTensorType destType) {
+  auto srcType = cast<RankedTensorType>(src.getType());
+  if (srcType == destType) {
+    return src;
+  }
+  Value value = src;
+  if (srcType.getEncoding()) {
+    value = IREE::Encoding::UnsetEncodingOp::create(
+        rewriter, src.getLoc(), srcType.dropEncoding(), value, dynamicDims);
+  }
+  if (destType.getEncoding()) {
+    value = IREE::Encoding::SetEncodingOp::create(rewriter, src.getLoc(),
+                                                  destType, value);
+  }
+  return value;
+}
+
+/// Pattern to convert `iree_tensor_ext.dispatch.tensor.load` operation when
+/// materializing the encoding.
+struct DecomposeMismatchEncodingTensorLoadOp
+    : public OpRewritePattern<IREE::TensorExt::DispatchTensorLoadOp> {
+  using OpRewritePattern<
+      IREE::TensorExt::DispatchTensorLoadOp>::OpRewritePattern;
+
+  DecomposeMismatchEncodingTensorLoadOp(
+      MaterializeEncodingTypeConverter &converter, MLIRContext *ctx,
+      PatternBenefit benefit = 0)
+      : OpRewritePattern<IREE::TensorExt::DispatchTensorLoadOp>(ctx, benefit),
+        typeConverter(converter) {}
+
+  LogicalResult matchAndRewrite(IREE::TensorExt::DispatchTensorLoadOp loadOp,
+                                PatternRewriter &rewriter) const override {
+    if (!loadOp.isLoadOfWholeSource()) {
+      return rewriter.notifyMatchFailure(loadOp, "unhandled partial loads");
+    }
+
+    // We have to check the bound type from converted DispatchTensorType because
+    // it is what we'll see in encoding materialization. E.g.,
+    // GPUPaddingResolver converts RankedTensorType into the same type, but it
+    // creates different IREE::TensorExt::DispatchTensorType that may have
+    // larger tensor shape for bound type.
+    IREE::TensorExt::DispatchTensorType srcType = loadOp.getSourceType();
+    auto convertedSrcType =
+        typeConverter.convertType<IREE::TensorExt::DispatchTensorType>(srcType);
+    RankedTensorType destType = loadOp.getResult().getType();
+    if (typeConverter.convertType(convertedSrcType.getBoundType()) ==
+        typeConverter.convertType(destType)) {
+      return rewriter.notifyMatchFailure(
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
```diff
@@ -281,6 +281,147 @@ struct MaterializeTensorExtDispatchTensorStoreOp
...
+        rewriter, loc, boundTensorType, loadOp.getSource(),
+        loadOp.getSourceDims(), loadOp.getMixedOffsets(),
+        loadOp.getMixedSizes(), loadOp.getMixedStrides());
+    SmallVector<Value> dynamicDims;
+    for (OpFoldResult value : loadOp.getMixedSizes()) {
+      if (isa<Attribute>(value)) {
+        continue;
+      }
+      dynamicDims.push_back(cast<Value>(value));
+    }
```
Max191:
```suggestion
    ArrayRef<Value> dynamicDims = loadOp.getSizes();
```

compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
```diff
@@ -281,6 +281,147 @@ struct MaterializeTensorExtDispatchTensorStoreOp
...
+
+    Location loc = storeOp.getLoc();
+    Value valueToStore = storeOp.getValue();
+    SmallVector<Value> dynamicDims;
+    for (OpFoldResult value : storeOp.getMixedSizes()) {
+      if (isa<Attribute>(value)) {
+        continue;
+      }
+      dynamicDims.push_back(cast<Value>(value));
+    }
```
Max191:
```suggestion
    ArrayRef<Value> dynamicDims = storeOp.getSizes();
```

Max191 (approved):
LGTM % one small request: Can you add a warning (or at least debug) print in each of the patterns on success, since having layout mismatches is likely going to be a performance pitfall? It would be good to be able to have this information dumped easily.

(Or maybe it would be better somewhere like SpecializeEncodings, which would also be fine, but having the print somewhere seems useful to me)

hanhanW:
> LGTM % one small request: Can you add a warning (or at least debug) print in each of the patterns on success, since having layout mismatches is likely going to be a performance pitfall? It would be good to be able to have this information dumped easily.
> 
> (Or maybe it would be better somewhere like SpecializeEncodings, which would also be fine, but having the print somewhere seems useful to me)

I was thinking to print it in my new pass (issue: https://github.com/iree-org/iree/issues/22485), then I realized it may not have layout transfer request. It only knows that the encodings are changed. So maybe having a debug print is okay here.

compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
hanhanW: Thanks, I did not know the method. btw, we should use `SmallVector`. Otherwise, there is a dangling pointer/reference.

```
Object backing the pointer will be destroyed at the end of the full-expression
```

PR #22583, opened 2025-11-07
[NFC] Refresh golden values for benchmarks.

hanhanW:
The numbers are updated based on https://github.com/iree-org/iree/actions/runs/19124329086


hanhanW:
We may not want to update mi250 numbers because I'm seeing the tolerance is 1.3 which is higher than others (i.e. 1.1).

I'm not sure how tolerance works in torch_models, but I still update the numbers based on the log.

MaheshRavishankar (commented):
How much tolerance are there in these numbers?

hanhanW:
> How much tolerance are there in these numbers?

All of them are 1.1, except mi250 which is 1.3

hanhanW:
Let's hold a bit. I want to compare with other CI jobs tomorrow. There are many jobs in queue now..

hanhanW:
Actually there is another sample, let me check if some numbers are off a lot or not.

https://github.com/iree-org/iree/actions/runs/19148522429

tests/external/iree-test-suites/sharktank_models/benchmarks/sdxl/e2e_rocm.json
```diff
@@ -34,7 +34,7 @@
         "mi308": 1.1
     },
     "golden_time_ms": {
-        "mi250": 990.0,
+        "mi250": 900.0,
```
hanhanW, line 37: It's 912 in the other log.

tests/external/iree-test-suites/torch_models/llama_8b_fp16/decode_benchmark_seq2048_mi325.json
```diff
@@ -39,5 +39,5 @@
         "value": "34x2097152xf16"
       }
     ],
-    "golden_time_ms": 7.5
+    "golden_time_ms": 6.8
```
hanhanW, line 42: It's 6.93 in the other sample

tests/external/iree-test-suites/torch_models/llama_8b_fp16/prefill_benchmark_seq2048_mi325.json
```diff
@@ -36,5 +36,5 @@
         "value": "34x2097152xf16"
       }
     ],
-    "golden_time_ms": 260.0
+    "golden_time_ms": 242.0
```
hanhanW, line 39: It's 244 in the other sample.

tests/external/iree-test-suites/torch_models/llama_8b_fp8/prefill_benchmark_seq2048_mi325.json
```diff
@@ -36,5 +36,5 @@
         "value": "34x2097152xf8E4M3FNUZ"
       }
     ],
-    "golden_time_ms": 180.0
+    "golden_time_ms": 163.0
```
hanhanW, line 39: It's 164.7 in the other sample

hanhanW (commented):
I pointed out the cases that differ a bit between the logs. Others are very similar.

Groverkss (approved):
LGTM, although I don't see a reason why we have threshold multipliers for things in the shark_models ci, another reason to move away from it.

MaheshRavishankar (approved):

hanhanW:
> LGTM, although I don't see a reason why we have threshold multipliers for things in the shark_models ci, another reason to move away from it.

I guess it is because sometimes GPUs are not stable for latency?

hanhanW:
I'll land the PR tomorrow, just in case if my change breaks CI. :/

Groverkss:
> > LGTM, although I don't see a reason why we have threshold multipliers for things in the shark_models ci, another reason to move away from it.
> 
> I guess it is because sometimes GPUs are not stable for latency?

The multiplier is not machine specific. The threshold essentially multiplies whatever check we have by the value. So if we were checking for max 300.0ms with a multiplier of 1.1x, we might as well just check for 330.0ms, it's the same thing. It's just an extra number with no meaning.

hanhanW:
It looks like I need to relax the numbers a bit for torch_models: https://github.com/iree-org/iree/actions/runs/19175420303/job/54819614759

PR #22595, opened 2025-11-07
[CI] Refresh golden values for failing benchmarks: min(val*1.1, val+5ms)

hanhanW:
Based on the discussion on discord, we set the golden values to `std::min(val * 1.1, val + 5ms)`: https://discord.com/channels/689900678990135345/689957613152239638/1436476349400486031

10% is the usual max delta across chips from the same SKU

ci-extra: test_torch

### DIFF
--- a/tests/external/iree-test-suites/torch_models/llama_8b_fp16/decode_benchmark_seq128_mi325.json
+++ b/tests/external/iree-test-suites/torch_models/llama_8b_fp16/decode_benchmark_seq128_mi325.json
@@ -39,5 +39,5 @@
         "value": "34x2097152xf16"
       }
     ],
-    "golden_time_ms": 5.5
+    "golden_time_ms": 6.05
 }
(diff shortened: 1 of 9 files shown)


hanhanW:
Let's wait for the full CI result before landing it.

hanhanW:
Restarting the CI because I missed `ci-extra: test_torch` in PR description. test_torch is not run by default.

kuhar (approved):
this makes sense because 10% is the usual max delta across chips from the same SKU

hanhanW:
> 10% is the usual max delta across chips from the same SKU

thanks, good to know! let me add it to PR description.

hanhanW:
The numbers are updated based on

image

hanhanW:
The failing job is because it is using shark10-ci. See https://discord.com/channels/689900678990135345/689957613152239638/1436492741151293570 for more details. It is a known issue. Landing the PR now.

PR #22596, opened 2025-11-07
Update split reduction cutoff conditions

yzhang93:
The main change of this PR is to slightly simplify the cutoff condition for split reduction. Instead of setting cutoff values for each individual dimensions, it now compares the ratio between the reduction size and the output size. The ratio threshold is adjusted to exclude small-shape edge cases where split reduction caused performance regressions.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -206,7 +206,6 @@ private:
       return std::nullopt;
     }
 
-    AffineMap inputMap = linalgOp.getMatchingIndexingMap(input);
     AffineMap filterMap = linalgOp.getMatchingIndexingMap(filter);
     AffineMap outputMap = linalgOp.getMatchingIndexingMap(output);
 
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -271,36 +270,34 @@ struct SetSplitReductionSizesPass final
...
-            outputChannelSize > mediumDimSize && batchSize > mediumDimSize) {
-          LDBG() << "skipping op; small input spatial size";
-          return std::nullopt;
-        }
-      }
+    // When the reduction size is small relative to the output sizes, split
+    // reduction often has no effect or even degrades performance.
+    SmallVector<int64_t> tileSizes = std::move(*maybeSizes);
+    int64_t reductionSize = std::accumulate(tileSizes.begin(), tileSizes.end(),
+                                            1, std::multiplies<>());
```
hanhanW: This usually has bugs, because the implementation can use int type for accumulator. The prefer way is `llvm::product_of(tileSizes)`.

hanhanW (commented):
btw, this is not NFC. The behavior changes, right?

hanhanW:
> btw, this is not NFC. The behavior changes, right?

I may be wrong, because I did not review the details.

yzhang93:
> btw, this is not NFC. The behavior changes, right?

Yeah, this may not be strictly NFC. The cutoff thresholds are updated but the outcome results are similar.

MaheshRavishankar (approved):
It would be good if we had some tests to back the heuristics being added here.

compiler/src/iree/compiler/DispatchCreation/test/set_split_reduction_sizes_conv.mlir
```diff
@@ -77,14 +77,14 @@ util.func public @no_split_large_N_F_sizes(%arg0: tensor<16x98x50x1024xf32>, %ar
 #map = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d4, d1 + d5, d2 + d6, d3)>
 #map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d4, d5, d6, d0)>
 #map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d1, d2, d3)>
-util.func public @no_split_small_H_W_sizes(%arg0: tensor<16x26x18x288xf32>, %arg1: tensor<16x24x16x288xf32>, %arg2: tensor<288x3x3x288xf32>) -> tensor<288x3x3x288xf32> {
-  %0 = linalg.generic {indexing_maps = [#map, #map1, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel", "reduction", "reduction", "reduction"]} ins(%arg0, %arg1 : tensor<16x26x18x288xf32>, tensor<16x24x16x288xf32>) outs(%arg2 : tensor<288x3x3x288xf32>) {
+util.func public @no_split_small_H_W_sizes(%arg0: tensor<16x26x18x96xf32>, %arg1: tensor<16x24x16x96xf32>, %arg2: tensor<96x3x3x96xf32>) -> tensor<96x3x3x96xf32> {
```
yzhang93, line 80: Update the shapes of the test to reflect the effect of this PR. Split reduction would be applied on this shape before the change. After the change this case is excluded from applying split reduction.

PR #22598, opened 2025-11-08
[Codegen][Tuner] expose python binding for getIGEMMGenericConvDetails

bangtianliu:
This PR introduces a new Python binding `get_igemm_generic_conv_details` that analyzes convolution operations and returns their IGEMM (implicit GEMM) structure, including loop bounds and problem dimensions. 

This is the first step toward enabling the tuner to support convolutions with additional layouts beyond `nhwc_hwcf`.


compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
```diff
@@ -91,6 +91,21 @@ ireeCodegenGetAttentionOpDetail(MlirAffineMap qMap, MlirAffineMap kMap,
...
+struct ireeCodegenIGEMMGenericConvDetails {
+  MlirAttribute igemmLoopBounds;
+  MlirAttribute convDimsBatch;
+  MlirAttribute convDimsOutputImage;
+  MlirAttribute convDimsOutputChannel;
+  MlirAttribute convDimsFilterLoop;
+  MlirAttribute convDimsInputChannel;
+  MlirAttribute convDimsDepth;
+  bool isOutputChannelFirst;
+  bool isValid;
```
kuhar: What does this mean? Would it make sense to have a separate helper function for checking if IGEMM details can be queried for a given op in the first place?

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -244,3 +246,71 @@ bool ireeCodegenMlirOperationIsACodegenAttentionOp(MlirOperation op) {
...
       unwrap(op));
 }
+
+ireeCodegenIGEMMGenericConvDetails
+ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op) {
+  mlir::Operation *operation = unwrap(op);
+  auto linalgOp = llvm::dyn_cast<mlir::linalg::LinalgOp>(operation);
+
+  if (!linalgOp) {
+    return ireeCodegenIGEMMGenericConvDetails{
```
kuhar: Yeah, I think this would be better handled inside python bindings, and to assert here instead.

kuhar (changes requested):

compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
kuhar: The rationale is that we generally prefer to invalid states not to be representable: https://geeklaunch.io/blog/make-invalid-states-unrepresentable/

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
bangtianliu: I can do it in this way as shown below:
```C
  ireeCodegenIGEMMGenericConvDetails result{};
  auto linalgOp = llvm::dyn_cast<mlir::linalg::LinalgOp>(unwrap(op));
  if (!linalgOp) {
    return result;
  }

  llvm::FailureOr<mlir::iree_compiler::IREE::LinalgExt::IGEMMGenericConvDetails>
      maybeDetails =
          mlir::iree_compiler::IREE::LinalgExt::getIGEMMGenericConvDetails(
              linalgOp);
  if (failed(maybeDetails)) {
    return result;
  }
```
Similar to : https://github.com/llvm/llvm-project/pull/135253/files#diff-77af7f1bca93ed6c9c25c48a91a73d279a6fac6f9b26c06ad5937258a475a96dR88-R96

compiler/bindings/python/IREECompilerDialectsModule.cpp
```diff
@@ -29,6 +29,9 @@ namespace py = nanobind;
 using namespace nanobind::literals;
 using namespace mlir::python::nanobind_adaptors;
 
+using ireeCodegenIGEMMGenericConvDetails =
+    struct ireeCodegenIGEMMGenericConvDetails;
```
kuhar: What does this do?

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -149,3 +149,182 @@ def test_isa_attention_op():
...
+    details = iree_codegen.get_igemm_generic_conv_details(root_op_list[0])
+    assert (
+        details is not None
+    ), "IGEMM details should be valid for generic 1D conv weight backward"
+    assert details.igemm_loop_bounds == [
+        96,
+        3,
+        96,
+        98304,
+    ], f"Expected [96,3,96,98304], got {details.igemm_loop_bounds}"
```
kuhar: I think we should either switch these to pytest and get these prints for free or repeating the expected values in error messages

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -244,3 +246,49 @@ bool ireeCodegenMlirOperationIsACodegenAttentionOp(MlirOperation op) {
...
+    return result;
+  }
+
+  llvm::FailureOr<mlir::iree_compiler::IREE::LinalgExt::IGEMMGenericConvDetails>
+      maybeDetails =
+          mlir::iree_compiler::IREE::LinalgExt::getIGEMMGenericConvDetails(
+              linalgOp);
+  if (failed(maybeDetails)) {
+    return result;
+  }
```
kuhar: This doesn't remove the invalid state, it just changes the representation.

My suggestion is to have a new public C API function that checks if something has igemm details, and separately a get function that asserts that the igemm details can be queried.

kuhar (changes requested):

compiler/bindings/python/IREECompilerDialectsModule.cpp
bangtianliu: mainly for the convenience of below C++ function:

```C++
static std::optional<ireeCodegenIGEMMGenericConvDetails>
GetIGEMMGenericConvDetails(MlirOperation op)
```

kuhar: I don't understand what it does

bangtianliu: just allow C++ code to use the C struct without `struct` keyword. But nvm, it will be fixed in the next commit.

compiler/bindings/python/test/api/tuner_api_test.py
bangtianliu: Changed it to 
```
    assert details.igemm_loop_bounds == [
        96,
        3,
        96,
        98304,
    ], f"got {details.igemm_loop_bounds}"
```

The expected values are already visible in the assertion itself.

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -244,3 +246,59 @@ bool ireeCodegenMlirOperationIsACodegenAttentionOp(MlirOperation op) {
...
+  auto linalgOp = llvm::dyn_cast<mlir::linalg::LinalgOp>(unwrap(op));
+  if (!linalgOp) {
+    return false;
+  }
+
+  llvm::FailureOr<mlir::iree_compiler::IREE::LinalgExt::IGEMMGenericConvDetails>
+      maybeDetails =
+          mlir::iree_compiler::IREE::LinalgExt::getIGEMMGenericConvDetails(
+              linalgOp);
+  return succeeded(maybeDetails);
```
kuhar:
```suggestion
  return succeeded(mlir::iree_compiler::IREE::LinalgExt::getIGEMMGenericConvDetails(
              linalgOp));
```

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -244,3 +246,59 @@ bool ireeCodegenMlirOperationIsACodegenAttentionOp(MlirOperation op) {
...
+
+  const mlir::iree_compiler::IREE::LinalgExt::IGEMMGenericConvDetails &details =
+      *maybeDetails;
+
+  mlir::Builder builder(linalgOp.getContext());
+
+  // Helper to convert unsigned to int64_t.
+  auto toInt64 = [](const llvm::SmallVector<unsigned, 2> &vec) {
+    return llvm::map_to_vector(
+        vec, [](unsigned val) { return static_cast<int64_t>(val); });
```
kuhar:
```suggestion
    return llvm::map_to_vector(
        vec, llvm::StaticCastTo<int64_t>);
```

bangtianliu: Thanks! good to know that we can write in this way!

kuhar: It landed in llvm a week ago

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -149,3 +149,174 @@ def test_isa_attention_op():
...
+    ], f"got {details.conv_dims_output_channel}"
+    assert details.conv_dims_filter_loop == [
+        4,
+        5,
+    ], f"got {details.conv_dims_filter_loop}"
+    assert details.conv_dims_input_channel == [
+        6
+    ], f"got {details.conv_dims_input_channel}"
+    assert (
+        details.is_output_channel_first == False
```
kuhar:
```suggestion
        not details.is_output_channel_first
```

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -244,3 +246,56 @@ bool ireeCodegenMlirOperationIsACodegenAttentionOp(MlirOperation op) {
...
+         "Failed to get IGEMM details; must check with "
+         "ireeCodegenHasIGEMMGenericConvDetails first");
+
+  const mlir::iree_compiler::IREE::LinalgExt::IGEMMGenericConvDetails &details =
+      *maybeDetails;
+
+  mlir::Builder builder(linalgOp.getContext());
+
+  // Helper to convert unsigned to int64_t.
+  auto toInt64 = [](const llvm::SmallVector<unsigned, 2> &vec) {
```
kuhar:
```suggestion
  auto toInt64 = [](ArrayRef<unsigned> vec) {
```
to avoid over-relying on the exact small vector size

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -149,3 +149,174 @@ def test_isa_attention_op():
...
+    ], f"got {details.conv_dims_output_image}"
+    assert details.conv_dims_filter_loop == [4], f"got {details.conv_dims_filter_loop}"
+
+    # Test with a non-conv operation.
+    module_str = """
+        module {
+            func.func @matmul(%arg0: tensor<4x4xf32>, %arg1: tensor<4x4xf32>) -> tensor<4x4xf32> {
+                %cst = arith.constant 0.000000e+00 : f32
+                %0 = tensor.empty() : tensor<4x4xf32>
+                %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<4x4xf32>) -> tensor<4x4xf32>
```
kuhar: You can make %1 a third function argument to make this test case more concise

bangtianliu:
I am not fully confident about exposing everything inside this struct to python binding: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h#L159-L195.  Some of the fields look redundant (like `igemmOperands`). 

I’ll need to finish understanding how convolution is handled inside `setIGEMMConvolutionLoweringConfig` before deciding what actually needs to be exposed.

compiler/bindings/python/IREECompilerDialectsModule.cpp
```diff
@@ -693,4 +694,80 @@ NB_MODULE(_ireeCompilerDialects, m) {
...
+      .def_prop_ro(
+          "filter_reassoc_indices",
+          [](const ireeCodegenIGEMMGenericConvDetails &self)
+              -> std::vector<std::vector<int64_t>> {
+            std::vector<std::vector<int64_t>> result;
+            MlirAttribute attr = self.filterReassocIndices;
+            assert(!mlirAttributeIsNull(attr) && mlirAttributeIsAArray(attr) &&
+                   "filterReassocIndices should be a valid ArrayAttr");
+            size_t n = mlirArrayAttrGetNumElements(attr);
+            result.reserve(n);
```
kuhar:
```suggestion
            MlirAttribute attr = self.filterReassocIndices;
            assert(!mlirAttributeIsNull(attr) && mlirAttributeIsAArray(attr) &&
                   "filterReassocIndices should be a valid ArrayAttr");
            size_t n = mlirArrayAttrGetNumElements(attr);
            std::vector<std::vector<int64_t>> result;
            result.reserve(n);
```
try to keep the definition close to the first use

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -244,3 +245,76 @@ bool ireeCodegenMlirOperationIsACodegenAttentionOp(MlirOperation op) {
...
+    iteratorAttrs.push_back(
+        builder.getStringAttr(mlir::utils::stringifyIteratorType(iterType)));
+  }
+  result.igemmLoopIterators = wrap(builder.getArrayAttr(iteratorAttrs));
+
+  result.im2colOutputPerm =
+      wrap(builder.getI64ArrayAttr(details.im2colOutputPerm));
+
+  llvm::SmallVector<mlir::Attribute> reassocAttrs;
+  for (const auto &indices : details.filterReassocIndices) {
```
kuhar: can you spell out this type? https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

kuhar (approved):

Max191 (approved):

PR #22605, opened 2025-11-10
[Codegen][GPU] Replace prefetchLoop with stage-based backward slicing

jerryyin:
This PR rewrites the `LoopPrefetcher` class (and leaves the pipelining step https://github.com/iree-org/iree/pull/22523 untouched). 

Instead of relying on a one-shot recursive approach, it now classify loop operations into read/write/compute stages using backward slice from identified root operations. This PR now separate the pattern matching logic into 3 steps: 1) stage identification 2) slice computation 3) classification. This makes the prefetcher easier to maintain and experiment with different pipeline flavors.

Note: This PR has no functional impact, all unit test continue to pass.


nirvedhmeshram (approved):
LGTM

jerryyin:
Note I'm merging for now to be able to submit successive PRs. I admit multiple stake holders are on vacation and want to remind that post-PR reviews are always welcomed!

PR #22606, opened 2025-11-10
Improving support for iree_codegen.extract_strided_metadata.

benvanik:
The existing handling was inconsistently converting the better iree_codegen op into the memref op which dropped information - now instead we convert upstream into iree_codegen and use that for our patterns before at the end going back to memref. This also adds support for walking over metadata ops on the way to the binding (like assume_alignment) - it's still not great (won't support functions/control flow/etc) but generally we don't want bindings used outside of the entry point function anyway.

This uncovered we could finally kill some TODOs in our samples _and_ it exposed most were broken unless they all had unique buffers and zero offsets and no alignments. I discovered this while adding a better buffer reuse path that caused anything using the extract_strided_metadata op to start failing due to the round-tripping through upstream. There's probably better approaches or different ways we could do this, but this was required for correctness and felt like the minimal change that is still in-line with what the code seems to expect. And it lets us shrink those samples a lot!

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/IREEExpandStridedMetadata.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/IREEExpandStridedMetadata.cpp
@@ -45,6 +45,27 @@ static AffineMap getMulMap(MLIRContext *context) {
   return AffineMap::get(0, 2, s0 * s1);
 }
 
+/// Walks up the def-use chain to find the HAL interface binding subspan
+/// that a memref value derives from, passing through transparent wrapper ops.
+static std::optional<IREE::HAL::InterfaceBindingSubspanOp>
+getSourceInterfaceBinding(Value memrefValue) {
+  Value source = memrefValue;
+  // Walk through transparent wrapper operations.
+  while (source) {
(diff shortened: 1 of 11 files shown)


compiler/src/iree/compiler/Codegen/Common/IREEExpandStridedMetadata.cpp
```diff
@@ -45,6 +45,27 @@ static AffineMap getMulMap(MLIRContext *context) {
   return AffineMap::get(0, 2, s0 * s1);
 }
 
+/// Walks up the def-use chain to find the HAL interface binding subspan
+/// that a memref value derives from, passing through transparent wrapper ops.
+static IREE::HAL::InterfaceBindingSubspanOp
```
MaheshRavishankar: Nit: Could you make this `std::optional<IREE::HAL::InterfaceBindingSubspanOp>` ?

MaheshRavishankar (approved):
Nice! Thanks! Further proof that upstream memref.extract_strided_metadata is broken if anyone actually pays attention to issues.

PR #22609, opened 2025-11-10
Adding some Stream canonicalizations and RefineUsage improvements.

benvanik:
This finally allows us to turn transfers into clones when possible (same source/target) and remove some special casing around it. The load-bearing behavior was things like returning constants from public functions and other internal<->external interop that would cause mapping to resource lifetime to go pear-shaped and either produce incorrect programs (well, "underspecified" programs) or additional copies. We may still be underspecified but now are at least fast and consistent.

The new GlobalStorage bit during usage analysis lets us fix correctness issues with constant returns from public functions and internal/external interop.

There's a growing need to change up `!stream.resource<...>` a bit to better model the high-fidelity usage information we capture so we can better map it to HAL buffers but that is left for future cleanup for now.

Part of #16168 PR sequence (3/6).


benvanik:
(dependent on #22606)

PR #22481, opened 2025-10-30
[ROCM] Update bounds for large f16 data-tiling ukernel

jtuyls:
As part of trying to enable the tensor ukernels flag by default (https://github.com/iree-org/iree/pull/22318), it was discovered that some ukernels perform worse on some matmul benchmarks, resulting in regressions: https://github.com/iree-org/iree/actions/runs/18651161302/job/53169844154?pr=22318.

Through experimentation I found that the large f16 data-tiling ukernel only starts performing well from M==512, N==+-32832, K==+-512.

**Benchmark Results (K=4096, N=16384)**

| M         | No UKernels   | All UKernels  |
|-----------|---------------|---------------|
| 4         | 0.739 ms      | 0.832 ms      |
| 8         | 0.752 ms      | 0.782 ms      |
| 16        | 0.700 ms      | 0.731 ms      |
| 32        | 0.700 ms      | 0.737 ms      |
| 64        | 0.718 ms      | 0.823 ms      |
| 128       | 0.870 ms      | 0.765 ms      |
| 256       | 0.701 ms      | 0.853 ms      |
| 512       | 0.754 ms      | 1.02 ms       |
| 1024      | 0.988 ms      | 0.949 ms      |
| 2048      | 1.07 ms       | 1.5 ms        |
| 4096      | 2.07 ms       | 2.58 ms       |
| 8192      | 2.85 ms       | 3.73 ms       |

**Benchmark Results (K=4096, N=32768)**

| M         | No UKernels   | All UKernels  |
|-----------|---------------|---------------|
| 4         | 0.784 ms      | 0.74 ms       |
| 8         | 0.724 ms      | 0.841 ms      |
| 16        | 0.699 ms      | 0.822 ms      |
| 32        | 0.784 ms      | 0.806 ms      |
| 64        | 0.755 ms      | 0.784 ms      |
| 128       | 0.938 ms      | 0.874 ms      |
| 256       | 0.821 ms      | 0.806 ms      |
| 512       | 1.02 ms       | 1.04 ms       |
| 1024      | 1.33 ms       | 1.57 ms       |
| 2048      | 2.18 ms       | 2.47 ms       |
| 4096      | 3.37 ms       | 3.64 ms       |
| 8192      | 5.65 ms       | 6.88 ms       |

**Benchmark Results (K=4096, N=65536)**

| M         | No UKernels   | All UKernels  |
|-----------|---------------|---------------|
| 4         | 0.831 ms      | 0.931 ms      |
[...]

PR #22620, opened 2025-11-11
[e2e] Use remarks to verify ukernel match

### LINKED ISSUE #22443: Add a test-only flag to verify that a specific ukernel is actually used
See https://github.com/iree-org/iree/pull/22350#issuecomment-3422274181. 

For e2e tests, we need a mechanism to compare the ukernels that are required with those that are actually used, and to check whether any ukernel is used at all.

Yu-Zhewen:
Certain e2e matmul tests expect that a particular ukernel is actually selected (so that numerical correctness for that ukernel can be verified). However, we currently only check ukernel selection through lit tests.

This PR integrates the use of MLIR [RemarkEngine](https://mlir.llvm.org/docs/Remarks/) by adding two new flags to `iree-compile`:
- `remarks-filter`: emit all remarks whose category matches the given regex pattern. If empty, no remarks are emitted.
- `remarks-output-file`: output file for remarks in YAML format.

Then in e2e tests, a post-build verification checks if the expected string (currently the ukernel name) appears in the remarks YAML file.

Closes: #22443

ci-extra: windows_x64_msvc

### DIFF
--- a/build_tools/cmake/iree_e2e_generated_runner_test.cmake
+++ b/build_tools/cmake/iree_e2e_generated_runner_test.cmake
@@ -119,7 +119,10 @@ function(iree_e2e_runner_test)
     add_custom_command(
       TARGET "${_NAME}${_RULE_VARIANT_NAME}"
       POST_BUILD
-      COMMAND bash -c "if [ -f '${_REMARKS_FILE}' ] && grep -q '${_RULE_CHECK_REMARKS_PATTERN}' '${_REMARKS_FILE}'; then echo '✓ Found expected ukernel pattern: ${_RULE_CHECK_REMARKS_PATTERN}'; else echo 'ERROR: Expected pattern ${_RULE_CHECK_REMARKS_PATTERN} not found in remarks' >&2; exit 1; fi"
+      COMMAND "${Python3_EXECUTABLE}"
+              "${PROJECT_SOURCE_DIR}/build_tools/scripts/check_remarks_pattern.py"
+              "${_REMARKS_FILE}"
+              "${_RULE_CHECK_REMARKS_PATTERN}"
       VERBATIM
     )
   endif()
--- a/build_tools/cmake/iree_e2e_generated_runner_test.cmake
+++ b/build_tools/cmake/iree_e2e_generated_runner_test.cmake
@@ -30,6 +30,9 @@
 #   TEST_DEFINED: Whether to define a test target.
 #   TEST_DISABLED: The test target will be skipped and its status will be
 #       'Not Run'.
+#   REMARKS_FILTER: Emit all remarks whose category matches the given regex pattern. If empty, no remarks are emitted.
+#       Only applied to the test module, not the calls module.
+#   CHECK_REMARKS_PATTERN: Pattern to search for in the emitted remarks.
 function(iree_e2e_runner_test)
   if(NOT IREE_BUILD_TESTS)
     return()
@@ -43,7 +46,7 @@ function(iree_e2e_runner_test)
   cmake_parse_arguments(
     _RULE
     ""
-    "NAME;TEST_TYPE;VARIANT_NAME;TESTS_SRC;TESTS_VMFB;CALLS_SRC;CALLS_VMFB;TRACE;TARGET_BACKEND;DRIVER;TEST_RUNNER;TEST_DEFINED;TEST_DISABLED"
+    "NAME;TEST_TYPE;VARIANT_NAME;TESTS_SRC;TESTS_VMFB;CALLS_SRC;CALLS_VMFB;TRACE;TARGET_BACKEND;DRIVER;TEST_RUNNER;TEST_DEFINED;TEST_DISABLED;REMARKS_FILTER;CHECK_REMARKS_PATTERN"
     "COMPILER_FLAGS;RUNNER_ARGS;LABELS"
     ${ARGN}
   )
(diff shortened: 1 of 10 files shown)


tests/e2e/matmul/CMakeLists.txt
```diff
@@ -1815,7 +1843,7 @@ iree_generated_e2e_runner_test(
     "--lhs_rhs_type=f16"
     "--acc_type=f32"
     "--shapes=custom_mnk"
-    "--mnk=1024,1024,1024"
+    "--mnk=1024,32832,1024"
```
Yu-Zhewen, line 1846: This test is updated due to https://github.com/iree-org/iree/pull/22481

Yu-Zhewen:
I was adding remark checks for all e2e tests that specify `--iree-hip-enable-tensor-ukernels`, but I wasn’t sure how to handle the tests introduced in #21383. They’re labeled with `requires-gpu-cdna3`, but since they’re placed under
 https://github.com/iree-org/iree/blob/827552306dc3d12a0a9613513e7da398ebc4217e/tests/e2e/matmul/CMakeLists.txt#L2159
they’ll never actually run?
 
Also, `--transpose_rhs` and `--mnk_dynamicities` are missing from `GENERATOR_ARGS`, so these tests won’t match any existing ukernel.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -842,6 +851,27 @@ bool Invocation::initializeInvocation() {
...
+                                        /*failed=*/""};
+    // Always use REMARK_POLICY_ALL for now.
+    auto createPolicy = []()
+        -> std::unique_ptr<mlir::remark::detail::RemarkEmittingPolicyBase> {
+      return std::make_unique<mlir::remark::RemarkEmittingPolicyAll>();
+    };
+    // Always use YAML streamer for now.
+    if (failed(mlir::remark::enableOptimizationRemarksWithLLVMStreamer(
+            session.context, remarksOutputFile, llvm::remarks::Format::YAML,
+            createPolicy(), cats))) {
```
kuhar:
```suggestion
    // Always use REMARK_POLICY_ALL for now.
    // Always use YAML streamer for now.
    if (failed(mlir::remark::enableOptimizationRemarksWithLLVMStreamer(
            session.context, remarksOutputFile, llvm::remarks::Format::YAML,
            std::make_unique<mlir::remark::RemarkEmittingPolicyAll>(), cats))) {
```

compiler/plugins/target/ROCM/Dialect/ROCM/Transforms/ApplyBuiltinPDLPatterns.cpp
```diff
@@ -396,6 +397,11 @@ class ApplyBuiltinPDLPatternsDriverPass final
       if (!builtinName || !ukernelDesc) {
         return WalkResult::advance();
       }
+      // Emit remark using structured API.
+      remark::analysis(op->getLoc(),
+                       remark::RemarkOpts::name("UKernel").category(
+                           "ApplyBuiltinPDLPatternsDriverPass"))
```
kuhar: Is there no pass function that returns its name?

kuhar (commented):
Nice! I wanted to give remarks a try for a while now.

Should we also check in a test that shows how to add remarks to your usual `iree-opt` invocation? This is obfuscated a bit currently through e2e tests.

kuhar (approved):
Looks great

build_tools/cmake/iree_e2e_generated_runner_test.cmake
```diff
@@ -98,6 +114,16 @@ function(iree_e2e_runner_test)
...
 
+  # Add POST_BUILD command to check remarks if CHECK_REMARKS_PATTERN is specified
+  if(_RULE_CHECK_REMARKS_PATTERN)
+    add_custom_command(
+      TARGET "${_NAME}${_RULE_VARIANT_NAME}"
+      POST_BUILD
+      COMMAND bash -c "if [ -f '${_REMARKS_FILE}' ] && grep -q '${_RULE_CHECK_REMARKS_PATTERN}' '${_REMARKS_FILE}'; then echo '✓ Found expected ukernel pattern: ${_RULE_CHECK_REMARKS_PATTERN}'; else echo 'ERROR: Expected pattern ${_RULE_CHECK_REMARKS_PATTERN} not found in remarks' >&2; exit 1; fi"
+      VERBATIM
+    )
+  endif()
```
ScottTodd, line 128: This is _probably_ fine for most developers, but in general try to a void Bash for better portability across operating systems. Seems like this could either be a Python script or it could be part of the test process itself instead of a custom command running at `POST_BUILD`.

kuhar: would switching to posix sh help here?

bjacob: CMake provides its own mini shell-like commands as `cmake -E <command>`.  When writing custom commands in CMake, try that before invoking an actual shell.

In this case though, the logic in this command feels like a better fit for a small Python program like Scott says.

Yu-Zhewen: I am replacing it with a python script, as we may want more "complex" post-processing in the future.

ScottTodd: Great, thank you!

PR #22622, opened 2025-11-12
[DispatchCreation] Enable split reduction by default

yzhang93:
Always enables split reduction for convolutions and GEMMs with large reduction dimensions. Added a pass option to disable split on outer reduction ops such as batch norm or arg_compare ops. Set the option as true to develop and test on these ops.


hanhanW:
I'm trying to clear review queue. What happens to this PR? Is it stale?

yzhang93:
Yeah, I think this needs to be tested on more models. Currently, enabling it by flag is enough.

Issue #22081, opened 2025-09-24
Multi-device execution breaking up command buffers and incorrect device placement


Alex-Vasile:
### What happened?

Compiling a subset of an fp4 quantized model with an assembly kernel appears to cause IREE incorrectly place certain executions on the wrong device (device 0 instead of device 1) and to break up execution into multiple smaller command buffers.

I have included mlirs and the resulting decode function traces. `same_device` and `same_device_2` are for two trimmed down transformer blocks that occur on the same device (blocks 61&62 on device 0 for `same_device` and blocks 63&64 for device 1 for `same_device_2`). `diff_device` also contains 2 blocks but on different devices: block 62 on device 0 and block 63 on device 1.

Both `same_device` and `same_device_2` contain a single command buffer on their respective devices (ignore the small execution on deivce 0 for `same_device_2`, it's the token embedding and needed to be included`). `diff_device` meanwhile has multiple command buffers on *both* devices as well as incorrectly mixing executions on devices 0 and 1 (at t=2.522s there is a command buffer with an f4gemm on deivce 0 that breaks up the execution on device 1).

input MLIRS:
[diff_device.mlir.txt](https://github.com/user-attachments/files/22514628/diff_device.mlir.txt)
[same_device.mlir.txt](https://github.com/user-attachments/files/22514629/same_device.mlir.txt)
[same_device_2.mlir.txt](https://github.com/user-attachments/files/22514856/same_device_2.mlir.txt)

Traces of the two different decode runs:
[diff_device.tracy.txt](https://github.com/user-attachments/files/22514619/diff_device.tracy.txt)
[same_device.tracy.txt](https://github.com/user-attachments/files/22514618/same_device.tracy.txt)
[same_device_2.tracy.txt](https://github.com/user-attachments/files/22514854/same_device_2.tracy.txt)

### Steps to reproduce your issue

Compile command:
```
iree-compile $MLIR_PATH \
    -o=$VMFB_PATH \
    --iree-hal-target-device="hip[0]" \
    --iree-hal-target-device="hip[1]" \
    --iree-hal-target-device="hip[2]" \
    --iree-hal-target-device="hip[3]" \
    --iree-hal-target-device="hip[4]" \
    --iree-hal-target-device="hip[5]" \
    --iree-hal-target-device="hip[6]" \
    --iree-hal-target-device="hip[7]" \
    --iree-hip-target=gfx950 \
    --iree-opt-level=O3 \
[...]

PR #22623, opened 2025-11-12
[Flow] Transfer globals per affinity instead of replicating

### LINKED ISSUE #22564: Add a second mode to ReplicateGlobalsPerAffinity that create minimal globals
I discussed the pass with @MaheshRavishankar few weeks ago, and we think that we can have a second mode for the pass. The current implementation focuses on performance, but it makes IREE specialize a global multiple times instead of having unspecialized and a copy each use.

There are cases that copy can be eliminated if all the devices can directly access the memory. Thus, the goal is having a mode that inserts `flow.tensor.transfer` ops. It allows us to experiment different modes in the first place.

Pass implementation: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/Stream/Transforms/ReplicateGlobalsPerAffinity.cpp

sommerlukas:
Extends the existing pass to replicate globals per affinity with a second mode: Instead of replicating globals, which is beneficial for performance, but increases the memory footprint, this new mode transfers the globals to each affinity where they are needed. Other passes could eliminate the transfers/copies where possible.

Similar to the replication, this is currently guarded by an experimental flag.

Fixes #22564.


compiler/src/iree/compiler/Dialect/Stream/Transforms/test/transfer_globals_per_affinity.mlir
hanhanW, line 1: Are the test cases the same? I think we should add a run with different check prefix in the same file.

E.g.,

https://github.com/iree-org/iree/blob/5bdb8d286d4b815fca265fe7fd951317aa5126a1/compiler/src/iree/compiler/Codegen/Common/test/decompose_pack_unpack_ops.mlir#L1-L4

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
```diff
@@ -106,9 +112,15 @@ void buildStreamTensorPassPipeline(OpPassManager &passManager,
...
 
+  assert((!clReplicateGlobalsPerAffinity || !clTransferGlobalsPerAffinity) &&
+         "Cannot replicate and transfer at the same time.");
   if (clReplicateGlobalsPerAffinity) {
     passManager.addPass(IREE::Stream::createReplicateGlobalsPerAffinityPass());
   }
+  if (clTransferGlobalsPerAffinity) {
+    passManager.addPass(IREE::Stream::createReplicateGlobalsPerAffinityPass(
+        IREE::Stream::ReplicateGlobalsPerAffinityPassOptions{true}));
+  }
```
hanhanW: You can pass `clTransferGlobalsPerAffinity` to `createReplicateGlobalsPerAffinityPass`. I think they can be both true. The difference is just about which mode it is picking up. E.g.,

```
if (clReplicateGlobalsPerAffinity) {
passManager.addPass(IREE::Stream::createReplicateGlobalsPerAffinityPass(
        IREE::Stream::ReplicateGlobalsPerAffinityPassOptions{clTransferGlobalsPerAffinity}));
}
```

hanhanW (changes requested):

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ iree_lit_test_suite(
     "specialize_dispatches.mlir"
     "specialize_encodings.mlir"
     "sync_initializers.mlir"
+    "transfer_globals_per_affinity.mlir"
```
benvanik: keep test files prefixed with the name of the pass

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/transfer_globals_per_affinity.mlir
sommerlukas: Yes, the input is the same as the test for replicate, just with a different expected output. I'll merge them to one file with prefix 👍

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
sommerlukas: Should we rename the new flag in that case, e.g., to something like `iree-stream-experimental-transfer-to-replicate-globals`?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/transfer_globals_per_affinity.mlir
```diff
@@ -0,0 +1,332 @@
...
+// CHECK-NOT: util.global
+util.global private @global : tensor<10xf32>
+
+// CHECK-LABEL: @unknown_global_device(
+// CHECK-SAME:    %[[ARG0:[a-zA-Z0-9]+]]
+util.func private @unknown_global_device(%arg0: tensor<10xf32>) -> (tensor<10xf32>, tensor<10xf32>) {
+  // CHECK: %[[OPERAND_A:.+]] = flow.tensor.transfer %[[ARG0]] {{.+}} to #hal.device.affinity<@[[$DEVICE_A]]>
+  %0 = flow.tensor.transfer %arg0 : tensor<10xf32> to #hal.device.affinity<@device_a>
+
+  // CHECK-DAG: util.global.load immutable @[[$GLOBAL]]
```
sommerlukas: After running the pass, the original load of the global is still around (but unused), so we need to match it here. An alternative would be to `canonicalize` the output of the pass before `FileCheck`. What would you prefer @hanhanW?

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
```diff
@@ -106,9 +112,15 @@ void buildStreamTensorPassPipeline(OpPassManager &passManager,
     passManager.addPass(IREE::Stream::createAnnotateAffinitiesPass());
   }
 
+  assert((!clReplicateGlobalsPerAffinity || !clTransferGlobalsPerAffinity) &&
```
benvanik: Do not use asserts for error handling - handle errors properly by returning error messages indicating what the problem is. Asserts are disabled in most builds, and hitting an assert for most users with them enabled is a "the compiler crashed issue." Asserts should _never_ guard user input, only internal compiler invariants.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/transfer_globals_per_affinity.mlir
```diff
@@ -0,0 +1,332 @@
...
+  // CHECK-DAG: %[[TRANSFER_A:.+]] = flow.tensor.transfer %[[LOAD_A]] {{.+}} to #hal.device.affinity<@[[$DEVICE_A]]>
+  %global = util.global.load immutable @global : tensor<10xf32>
+
+  // CHECK: flow.dispatch @dispatch(%[[OPERAND_A]], %[[TRANSFER_A]])
+  %1 = flow.dispatch @dispatch(%0, %global) : (tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
+
+  // CHECK: %[[OPERAND_B:.+]] = flow.tensor.transfer %[[ARG0]] {{.+}} to #hal.device.affinity<@[[$DEVICE_B]]>
+  %2 = flow.tensor.transfer %arg0 : tensor<10xf32> to #hal.device.affinity<@device_b>
+
+  // CHECK-DAG: %[[LOAD_B:.+]] = util.global.load immutable @[[$GLOBAL]]
```
sommerlukas: Currently, we load before every transfer. Alternatively, we could extend the caching to load the global only once. There is a tradeoff involved with regard to lifetime of the value vs. repeated loading, but I'm not 100% sure what passes later in the pipeline would do. Do you think it's OK to load before every transfer @hanhanW?

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
```diff
@@ -28,6 +28,12 @@ static llvm::cl::opt<bool> clReplicateGlobalsPerAffinity(
         "Replicates globals for each unique affinity they are used with."),
     llvm::cl::init(false));
 
+static llvm::cl::opt<bool> clTransferGlobalsPerAffinity(
+    "iree-stream-experimental-transfer-globals-per-affinity",
```
benvanik: This should be a mode flag or enum - one flag for the feature, either it's off (not set) or one of N enum values passed as an option to the pass.

benvanik (changes requested):
/cc @hanhanW We need to get this out of stream - it's 100% a flow pass and shouldn't be in the stream dialect (flow lowers to stream, stream should only interact with flow via conversion or interfaces, not direct op references - e.g., if you have to include FlowOps.h then it's a layering violation).

hanhanW:
Let's try to move it to the end of flow pipeline before moving forward. We can add it right after `Cleanup executable contents`.

https://github.com/iree-org/iree/blob/82216e4400d16ed7ed51fb97c9d3d5613ffaad72/compiler/src/iree/compiler/Dialect/Flow/Transforms/Passes.cpp#L230C7-L247

Please help verify if it still fixes the issue. The verification step is:

- Download the MLIR file from the [original issue](https://github.com/iree-org/iree/issues/22081): https://gist.githubusercontent.com/MaheshRavishankar/6045e9ac1d921c55785eb7c5639c80ea/raw/27022b77962078f0ef5677693413d414a474f59d/llama_8b_fp16.mlir
- Compile the program to Stream.

```bash
iree-compile ~/llama_8b_fp16.mlir -o ~/fixed.stream.mlir \
    --compile-to=stream \
    --iree-hip-target=gfx950 \
    --iree-hal-target-device="hip[0]" \
    --iree-hal-target-device="hip[1]" \
    --iree-opt-level=O3 \
    --iree-dispatch-creation-propagate-collapse-across-expands=true \
    --iree-stream-affinity-solver-max-iterations=1024 \
    --iree-hal-indirect-command-buffers=true \
    --iree-stream-resource-memory-model=discrete \
    --iree-hip-specialize-dispatches \
    --iree-hal-memoization=true \
    --iree-codegen-enable-default-tuning-specs=true --iree-stream-experimental-replicate-globals-per-affinity
```

Track the SSA uses of `%arg4` in `@prefill_bs4$async`. With the flag, you should not see the copy ops within `stream.cmd.execute. E.g.,

```mlir
%13 = stream.tensor.import on(#hal.device.affinity<@__device_1>) %arg4 : !hal.buffer_view -> tensor<?x1048576xf8E4M3FN>{%11} in !stream.resource<external>{%12}

%297 = stream.cmd.execute on(#hal.device.affinity<@__device_1>) await(%198) => ... %13 as %arg12
// no copy on %arg12
```

Without the pass, we are seeing the copy.

```mlir
%13 = stream.tensor.import on(#hal.device.affinity<@__device_1>) %arg4 : !hal.buffer_view -> tensor<?x1048576xf8E4M3FN>{%11} in !stream.resource<external>{%12}
%42 = stream.tensor.import on(#hal.device.affinity<@__device_1>) %arg4 : !hal.buffer_view -> tensor<?x1048576xf8E4M3FN>{%17} in !stream.resource<external>{%31}

%313 = stream.cmd.execute ... %13 as %arg10
  stream.cmd.copy %arg10[%c0], %arg12[%70], %12 : !stream.resource<external>{%12} -> !stream.resource<transient>{%71}
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/transfer_globals_per_affinity.mlir
sommerlukas: I've merged them to one file with corresponding prefixes now.

sommerlukas:
I've moved the pass over to Flow in a separate PR: https://github.com/iree-org/iree/pull/22634

compiler/src/iree/compiler/Dialect/Flow/Transforms/ReplicateGlobalsPerAffinity.cpp
```diff
@@ -318,6 +338,10 @@ class OpOperandAffinityStateMap {
 struct ReplicateGlobalsPerAffinityPass
     : public impl::ReplicateGlobalsPerAffinityPassBase<
           ReplicateGlobalsPerAffinityPass> {
+
+  using impl::ReplicateGlobalsPerAffinityPassBase<
+      ReplicateGlobalsPerAffinityPass>::ReplicateGlobalsPerAffinityPassBase;
+
```
hanhanW: `using Base::Base;` and remove blank lines.

hanhanW (approved):
LGTM, just one nit.

compiler/src/iree/compiler/Dialect/Flow/Transforms/ReplicateGlobalsPerAffinity.cpp
sommerlukas: Thanks, done.

hanhanW:
@benvanik do you want to take a second look before we land this?

benvanik (approved):

PR #22625, opened 2025-11-12
Implementing initial end-to-end support for external transients.

### LINKED ISSUE #22385: Add support for external transient memory
Tracking for the project to add the `hal.tensor.transients` and `stream.resource.transients` ops, analysis, and passes for converting transient resource allocations into subviews of a user-provided storage buffer (which could be a function argument, a manual allocation, result of an external call, etc).

Design doc/prompt for Claude: https://gist.github.com/benvanik/9f6ca5fdc17e0e50938bf34a2437cac2

## Overview
Enable users to provide their own buffers for transient memory allocation, with automatically generated query functions to calculate required sizes. This supports kernel JIT use cases where applications need control over transient allocations to achieve zero allocations in steady state.

## Motivation
When building a kernel JIT on top of IREE, users provide IR of their linalg ops which we compile into dispatches. The host code needs to schedule execution with transient allocation, but users need to control and pre-allocate transient memory ahead of time. This feature provides size query functions and allows passing storage buffers to functions.

## Approach
1. Add `iree.abi.transients` attribute to mark function arguments as transient storage providers
2. Generate size query functions (with `iree.abi.reflection` annotation) that compute required transient buffer sizes
3. Transform the compilation pipeline to use user-provided buffers instead of internal allocation

## Example Input
```mlir
util.func public @my_kernel(
    %arg0: tensor<?xf32>,
    %arg1: index,
    %storage: !hal.buffer {iree.abi.transients}
) -> tensor<?xf32> {
  // Computation that requires transient allocations.
  %result = ...
  util.return %result : tensor<?xf32>
}
```

## Example Output

**Generated Size Query Function:**
```mlir
// Automatically generated with constant size annotation.
// If the size was calculated based on shape information or other
// user arguments they'd be here along with all the arithmetic.
util.func public @my_kernel__transients_size(%arg0: !hal.buffer_view) -> index attributes {
  iree.abi.reflection = {size = 4096 : index}
} {
  %c4096 = arith.constant 4096 : index
  util.return %c4096 : index
}
```

**Transformed Function IR:**
```mlir
// Original function now uses provided storage for transients.
util.func public @my_kernel(...) attributes {
  iree.abi.reflection = {iree.abi.transients.size = "my_kernel__transients_size"}
} {
  // stream.resource.alloca replaced with stream.resource.subview from %storage.
  %transient = stream.resource.subview %storage[%offset] : !stream.resource<transient>
  // ... rest of computation uses subviewed storage.
}
```

## Implementation Plan
See detailed design doc: https://gist.github.com/benvanik/9f6ca5fdc17e0e50938bf34a2437cac2

**Phased approach:**
- Phase 0-1: ABI/HAL/Stream foundation (IR design validation)
- Phase 2: Stubbed end-to-end implementation (simple cases)
- Phase 3: Timeline liveness analysis infrastructure
- Phase 4-5: Production implementation with call graph propagation

## Key Technical Details
- **Timeline-based liveness analysis**: Tracks async timepoint use-def chains for accurate liveness ranges across asynchronous execution
- **Size computation hoisting**: Propagates size calculations up call graphs, enabling nested transient packing
- **Zero-allocation steady state**: Users call size query once, then reuse pre-allocated buffers across invocations

benvanik:
Experimental early version that supports quite a few cases but no calls and no data-dependent values (control flow is _mostly_ supported, which will be critical with dispatch specialization).

This was cherry picked from a deep work branch that had many fixes - there may be spurious copies occasionally until that branch lands, but in the simple cases tested here we're good. Of particular note is the cherry picked RefineUsage.cpp changes which work around an infinite pattern application bug (fixed for real in future changes).

Fixes #22385.


compiler/src/iree/compiler/Bindings/Native/Transforms/WrapEntryPoints.cpp
```diff
@@ -674,6 +696,23 @@ createExportWrapperFunc(IREE::ABI::InvocationModel invocationModel,
     asyncResults[resultIndex] = cast<OpResult>(aliasOp.getResult());
   }
 
+  // Annotate transient storage if provided.
+  // Create one transients op per tensor result.
+  if (transientStorage) {
```
MaheshRavishankar, line 702: I am not sure I understand the use case for transient storage for results.

compiler/src/iree/compiler/Dialect/HAL/IR/test/tensor_folding.mlir
```diff
@@ -40,6 +40,45 @@ util.func public @foldTensorExportImport(%arg0: tensor<5xi32>) -> tensor<5xi32>
 
 // -----
 
+// CHECK-LABEL: @FoldConsecutiveTransientsSameStorage
+// CHECK-SAME: (%[[ARG0:.+]]: tensor<4xf32>, %[[STORAGE:.+]]: !hal.buffer)
+util.func public @FoldConsecutiveTransientsSameStorage(%arg0: tensor<4xf32>, %storage: !hal.buffer) -> tensor<4xf32> {
+  // CHECK-NOT: hal.tensor.transients
+  %0 = hal.tensor.transients %arg0 : tensor<4xf32> from %storage : !hal.buffer
```
MaheshRavishankar, line 47: Im curious why you think this will ever happen. I expect this to be generated by front end and I am not really sure I even understand what it means for a `hal.tensor.transient` to have it source be another `hal.tensor.transient`. Shouldnt that just be an error?

compiler/src/iree/compiler/Dialect/Stream/Conversion/HALToStream/test/abi_ops.mlir
```diff
@@ -112,3 +112,138 @@ util.func public @tensorBarrier(%tensor0: tensor<3xf32>, %tensor1: tensor<?xf32>
...
+util.func public @importBufferViewCrossDevice(%view: !hal.buffer_view) -> tensor<4xf32> attributes {
+  stream.affinity = #hal.device.promise<@dev_b>
+} {
+  //  CHECK-DAG: %[[SIZE:.+]] = stream.tensor.sizeof on(#hal.device.promise<@dev_a>) tensor<4xf32>
+  //      CHECK: %[[RESOURCE:.+]] = stream.tensor.import on(#hal.device.promise<@dev_a>) %[[VIEW]] : !hal.buffer_view ->
+  // CHECK-SAME:     tensor<4xf32> in !stream.resource<external>{%[[SIZE]]}
+  // CHECK-NEXT: %[[TRANSFER:.+]] = stream.async.transfer
+  // CHECK-SAME:     %[[RESOURCE]] : !stream.resource<external>{%[[SIZE]]} from(#hal.device.promise<@dev_a>) ->
+  // CHECK-SAME:     to(#hal.device.promise<@dev_a>) !stream.resource<*>{%[[SIZE]]}
+  %0 = hal.tensor.import on(#hal.device.promise<@dev_a>) %view : !hal.buffer_view -> tensor<4xf32>
```
MaheshRavishankar, line 133: Is this related to the change for transients? Im just curious really.

compiler/src/iree/compiler/Dialect/Stream/Transforms/AnnotateConstantTransientSize.cpp
```diff
@@ -0,0 +1,153 @@
...
+  // Use pattern matching to detect constant return value.
+  IntegerAttr constantValue;
+  if (matchPattern(returnOp.getOperand(0), m_Constant(&constantValue))) {
+    return constantValue;
+  }
+
+  return std::nullopt;
+}
+
+struct AnnotateConstantTransientSizePass
```
MaheshRavishankar, line 66: I dont see this pass being run anywhere? Is this supposed to be run by the caller somehow?

tools/test/iree-run-module-in-place.mlir
```diff
@@ -0,0 +1,69 @@
...
+// RUN: (iree-compile %s \
+// RUN:    --iree-hal-target-device=local \
+// RUN:    --iree-hal-local-target-device-backends=vmvx | \
+// RUN:  iree-run-module \
+// RUN:    --device=local-task \
+// RUN:    --module=- \
+// RUN:    --function=in_place_computation \
+// RUN:    --input="64xf32=1.0" \
+// RUN:    --input="64xf32=0" \
+// RUN:    --input="&512xi8=0") | \
```
IanWood1: What does `&` do?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/emplace_transients_scf.mlir
```diff
@@ -0,0 +1,860 @@
...
+  // CHECK-DAG: %[[C0_I32:.+]] = arith.constant 0 : i32
+  %c0_i32 = arith.constant 0 : i32
+  %c128 = arith.constant 128 : index
+
+  // Loop with transient allocation inside, first iter awaits input timepoint.
+  // CHECK: scf.for %{{.+}} = %[[LB]] to %[[UB]] step %[[STEP]] iter_args(%[[ITER_TP:.+]] = %[[INPUT_TIMEPOINT]]) -> (!stream.timepoint) {
+  %final_tp = scf.for %iv = %lb to %ub step %step iter_args(%iter_tp = %input_timepoint) -> !stream.timepoint {
+    // Pack with single slice for loop-reused transient (now inside loop).
+    // CHECK: %[[PACK:.+]]:2 = stream.resource.pack slices({
+    // CHECK-NEXT:   [0, 0] = %c128{{(_[0-9]+)?}}
```
MaheshRavishankar: Probably want to capture `%[[C128:.+]] = arith.constant 128` and use it here?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/emplace_transients_scf.mlir
```diff
@@ -0,0 +1,860 @@
...
+  //
+  // CHECK: %[[STORAGE_VIEW:.+]] = stream.resource.subview %[[STORAGE]][%c0{{(_[0-9]+)?}}] : !stream.resource<transient>{%[[STORAGE_SIZE]]} -> !stream.resource<transient>{%[[PACK]]#0}
+
+  // Allocate transient with size from scf.if, awaiting input timepoint.
+  // CHECK: %[[TRANSIENT_SUBVIEW:.+]] = stream.resource.subview %[[STORAGE_VIEW]][%[[PACK]]#1]
+  %transient, %alloca_timepoint = stream.resource.alloca uninitialized await(%input_timepoint) =>
+      !stream.resource<transient>{%adjusted_size} => !stream.timepoint
+  // Subview from alloca replacement flows into execute.
+  // CHECK: %[[EXEC_TP:.+]] = stream.cmd.execute await(%[[INPUT_TIMEPOINT]]) => with(%[[TRANSIENT_SUBVIEW]] as %[[CAPTURE:.+]]: !stream.resource<transient>{%[[ADJUSTED_SIZE]]}) {
+  // CHECK:   stream.cmd.fill %c0_i32, %[[CAPTURE]][%c0 for %[[ADJUSTED_SIZE]]] : i32 -> !stream.resource<transient>{%[[ADJUSTED_SIZE]]}
```
MaheshRavishankar: Here and other places. I see the SSA name directly being used in tests, I find that a bit of an anti-pattern to rely on SSA value names. Id rather capture it with a CHECK* and use the capture name here, or just ignore the SSA value if it is not important,

compiler/src/iree/compiler/Bindings/Native/Transforms/test/wrap_entry_points.mlir
```diff
@@ -229,3 +229,88 @@ util.func private @importEncodingsCaller(%arg0: tensor<?x2xi32>) -> tensor<2x?xi
...
+//  CHECK-NEXT:   %[[ARG0_DIM:.+]] = hal.buffer_view.dim<%[[ARG0]] : !hal.buffer_view>[0]
+//  CHECK-NEXT:   %[[ARG0_TENSOR:.+]] = hal.tensor.import %[[ARG0]] "input0" : !hal.buffer_view -> tensor<?xf32>{%[[ARG0_DIM]]}
+//  CHECK-NEXT:   %[[RET_TENSOR:.+]] = util.call @_transientsEntry(%[[ARG0_TENSOR]], %[[STORAGE]])
+//       CHECK:   %[[RET_DIM:.+]] = tensor.dim %[[RET_TENSOR]]
+//       CHECK:   %[[ANNOTATED:.+]] = hal.tensor.transients %[[RET_TENSOR]] : tensor<?xf32>{%[[RET_DIM]]} from %[[STORAGE]] : !hal.buffer
+//  CHECK-NEXT:   %[[RET_VIEW:.+]] = hal.tensor.export %[[ANNOTATED]] "output0" : tensor<?xf32>{%[[RET_DIM]]} -> !hal.buffer_view
+//  CHECK-NEXT:   util.return %[[RET_VIEW]] : !hal.buffer_view
+//  CHECK-NEXT: }
+
+// CHECK-LABEL: util.func private @_transientsEntry
```
MaheshRavishankar, line 248: I am not sure what this generated function is from. I thought initially it is the `_transient_size`, but it is not.

MaheshRavishankar (commented):
Ok, I looked through this as much as I could follow. Most of looks fine to me. I am still unclear how to use it though. Here is my mental model

The "front-end" generates an extra argument that is annotated with `{iree.abi.transients}` . This is of type !hal.buffer.
Once emplace transients runs we have the size of the transient. The materialize transient size function generates a new function  with the name "<fn_name>_transient_size" that has the same arguments that tells you now much the transient size should be. 
There is another pass "AnnotateConstantTransientSize" that annotates the original function with the constant size if it is contant (which is a very useful feature) :) .

But as a front-end when you compile a program with IREE, what is the expected way the front end should use either the generated "_transient_size" function or the constant size annotation. That part isnt clear to me as a user of iree-compile.
Also what happends in the final generated vmfb to the "_transient_size" function. Is it just  a callable function that executes on the IREE VM?

compiler/src/iree/compiler/Bindings/Native/Transforms/WrapEntryPoints.cpp
benvanik: The transient op is metadata saying "any transient memory used in producing this value should use this storage" - the results are what we produce. Clarified comment.

compiler/src/iree/compiler/Dialect/HAL/IR/test/tensor_folding.mlir
benvanik: Defense-in-depth. Frontends are trash and produce this kind of thing all the time: we may have started with multiple functions that got inlined, or the frontend may tag it on every single value, or values may deduplicate even if they appeared the same in the frontend. It's legal for it to happen and so we handle it.

compiler/src/iree/compiler/Dialect/Stream/Conversion/HALToStream/test/abi_ops.mlir
benvanik: Missing test coverage found while adding the tests below it (we had nothing testing these ops at all).

compiler/src/iree/compiler/Dialect/Stream/Transforms/AnnotateConstantTransientSize.cpp
benvanik: good catch! cherry pick bug 🤦

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/emplace_transients_scf.mlir
benvanik: Oof thought I had caught all of these - fixed comprehensively.

compiler/src/iree/compiler/Bindings/Native/Transforms/test/wrap_entry_points.mlir
benvanik: This is how the WrapEntryPointsPass works - it creates a new function with the original contents (`@_transientsEntry`) and then the public function gets all the ABI goo (above). The name is just the name of the public function with a leading underscore. There's no size function here. This preserves the original function signature so that if there are other calls inside the module to that public function they still work (just renamed to use the new name) as the public function has its signature change.

benvanik:
Yep! The frontends can either use our `iree.abi.*` helpers that `WrapEntryPointsPass` handles or directly emit the same IR the pass does (pytorch for example goes straight to the hal ops for much better control over aliasing/etc). Depends on how the user is producing IR to feed into the compiler.

Yeah, these size functions are just functions in the program. Whatever is running the program can call the size functions or sniff the metadata to see if they have a constant with the normal VM API. If you think of an IREE module as a shared library with some exported functions then how those are called by the user code loading the library is up to it (is it one shot, does it care about sniffing constants or is it fine always calling for simplicity, does it know its workload ahead of time or is it caching every call, etc are all dependent on the user). For querying the name of the size generation function (don't assume it's always just a suffix) you can use `iree_vm_function_lookup_attr_by_name` with `iree.abi.transients.size` to get the size function name to call, or `iree.abi.transients.size.constant` to see if it has a constant directly on it in which case you can skip calling it. This should exist in the python layer if that's how the user uses them, and the functions are called in the same way - they just don't take fences (because they aren't async) and return a size instead of tensor results.

Our tflite bindings show how this can be done as it does something similar: the tflite `WrapEntryPointsPass` emits shape query functions and the interpreter.c `_TfLiteInterpreterRefreshInputShapes` calls them. There it's factored/optimized because every invocation needs to do this so it reuses the VM stack frame to avoid needing to set it up for every single input, but it's just an iree_vm_invoke with the right args just like calling into any function. The size functions here should probably take shape arrays instead of buffer views but I wanted to keep it simple to start. If trying to call these functions when you only have shapes you can just make a buffer view with a allocated dummy HAL buffer (we don't have an iree_hal_buffer_create_empty or something helper to give you back a 0 byte buffer, but it's easy enough to do and all buffer views could share the same dummy buffer).

benvanik:
Added a sample showing how to use this. Here's what the IR looks like, if you want to see what the functions/reflection/etc info is:
```mlir
module {
  hal.executable private @in_place_computation_linked_vmvx {
    hal.executable.binary public @vmvx_bytecode_fb attributes {data = dense<"0x504B03042D000000000000002100097E9F87FFFFFFFFFFFFFFFF090019006D6F64756C652E6662CAFE010000010010003C090000000000003C090000000000003C090000080000004952454508F7FFFF2400000001000000540400006804000074040000A406000000000F00DC060000180000000D0000006C696E6B65645F6D6F64756C6500000056F7FFFF0C0000001C0000005C00000010000000000202010302010201020103020102011000000000000000AC0300009803000088030000700300005C0300004C030000680200005802000044020000340200001C02000030010000200100000C010000FC00000003000000500200002801000004000000DCF7FFFF080000003C00000032000000696E5F706C6163655F636F6D7075746174696F6E5F64697370617463685F325F656C656D656E74776973655F36345F663332000014000000010000000D00000008000000000000000F000000000000001A00000000000000250000000000000028000000000000002B000000080000003600000002000000410000000D0000004C0000000F0000005D0000000F000000640000000F000000750000000F0000007C0000000F000000840000000F0000008C0000000F000000940000000F0000009C0000000F000000A30000000F000000AB0000000D000000C4F8FFFF0E00000002000000C2F8FFFF78020000270000000E000000E0F8FFFF0C00000002000000DEF8FFFF5C020000260000000D000000FCF8FFFF080000003C00000032000000696E5F706C6163655F636F6D7075746174696F6E5F64697370617463685F315F656C656D656E74776973655F36345F6633320000120000000100000000000000080000000000000013000000000000001E00000000000000260000000000000029000000000000002C000000060000003700000008000000420000000B0000004B0000000B000000520000000B000000630000000B0000006B0000000A000000730000000B0000007A0000000B000000820000000B000000890000000B0000009100000008000000BCF9FFFF04000000020000000A00000006000000E8F9FFFF0900000002000000E6F9FFFF54010000210000000F00000004FAFFFF070000000200000002FAFFFF380100001F0000000C00000020FAFFFF080000003C00000032000000696E5F706C6163655F636F6D7075746174696F6E5F64697370617463685F305F656C656D656E74776973655F36345F6633320000110000000100000000000000080000000000000013000000040000001E00000000000000260000000000000029000000000000002C00000002000000370000000600000042000000040000004B00000004000000520000000400000063000000040000006B0000000300000073000000040000007B0000000400000082000000040000008B00000006000000F0FAFFFF0500000002000000EEFAFFFF4C000000180000000C000000F4FAFFFF0400000002000000030000000200000020FBFFFF01000000020000001EFBFFFF1C00000012000000010000002EFBFFFF0C0000001A0000000F0000003500000073616D706C65732F65787465726E616C5F7472616E7369656E74732F696E5F706C6163655F636F6D7075746174696F6E2E6D6C697200000005000000900000007C000000680000004C000000240000000300000010010000C80000008000000003000000E00100008C01000038010000B8FBFFFF040000001400000021766D2E6C6973743C21766D2E6275666665723E00000000DCFBFFFF040000000A00000021766D2E6275666665720000F4FBFFFF04000000030000006936340004FCFFFF04000000030000006933320014FCFFFF0400000003000000663332002CFCFFFF080000000200000032000000696E5F706C6163655F636F6D7075746174696F6E5F64697370617463685F325F656C656D656E74776973655F36345F663332000070FCFFFF080000000100000032000000696E5F706C6163655F636F6D7075746174696F6E5F64697370617463685F315F656C656D656E74776973655F36345F6633320000ACFCFFFF0400000032000000696E5F706C6163655F636F6D7075746174696F6E5F64697370617463685F305F656C656D656E74776973655F36345F6633320000F4FCFFFF08000000180000000F000000307272726969696969696969695F760001000000040000001CFDFFFF08000000180000000D00000062696E64696E675F636F756E74000000010000003300000044FDFFFF08000000180000000F000000307272726969696969696969695F760001000000040000006CFDFFFF08000000180000000D00000062696E64696E675F636F756E74000000010000003200000094FDFFFF08000000180000000F000000307272726969696969696969695F76000100000004000000BCFDFFFF08000000180000000D00000062696E64696E675F636F756E74000000010000003200000003000000000000008E000000010000000000040009000300900000009400000001000000000004000A00030028010000AE000000010000000000040009000300D8010000790D0100000000000F010000000000000002000F40000000000000000400E0050000803F01000E06000C08001A028008000300000000801A02C00000030000000180564A0000000000007950060004000000570000620000000000008A000000000079E033008006000000E00A000001000000E0340180060000002A060002000600564A000000000000795A00000000790D0100000000000F010000000000000002000F40000000000000000400E0050000004001000E06000C08001A028008000300000000801A02C00000030000000180564A00000000000079500600040000005700006200000000000090000000000079E033008006000000E00C0000010000002A060004000800E0340180080000002A060002000600564A0000000000795A000000000000790D0200000000000D0100000001000F010000000000000002000F400000000000000004000E06000C08001A028008000300000000801A028001000300000001801A02C00000030000000280565C0000000002000600000007000100795000000400060057060074000000000000AA0000000000792A000004000600E033008006000600E033018000000700E00A060007000600E0340280000006002A000002000000565C0000000000795A0000000008000C00040008000600080004000A001000040008000C0008000800000004002400280004000000080000000C00000000001000140000000000000018001C0020002400000000000000000000000000000000000000000000000000504B01021E032D000000000000002100097E9F87FFFFFFFFFFFFFFFF09001C0000000000000000000000FFFFFFFF6D6F64756C652E6662010018003C090000000000003C090000000000000000000000000000504B06062C000000000000002D002D0000000000000000000100000000000000010000000000000053000000000000008009000000000000504B060700000000D30900000000000001000000504B0506000000000100010053000000FFFFFFFF0000"> : vector<2613xi8>, format = "vmvx-bytecode-fb", mime_type = "application/x-flatbuffers"}
  }
  util.global private @__device_0 : !hal.device
  util.initializer {
    %c18_i32 = arith.constant 18 : i32
    %false = arith.constant false
    %0 = util.null : !hal.device
    %device_count = hal.devices.count : index
    cf.br ^bb1(%c0, %c0, %0 : index, index, !hal.device)
  ^bb1(%1: index, %2: index, %3: !hal.device):  // 2 preds: ^bb0, ^bb4
    %4 = util.cmp.eq %3, %0 : !hal.device
    %5 = arith.cmpi slt, %1, %device_count : index
    %6 = arith.andi %4, %5 : i1
    cf.cond_br %6, ^bb2, ^bb5
  ^bb2:  // pred: ^bb1
    %device_n = hal.devices.get %1 : !hal.device
    %ok, %value = hal.device.query<%device_n : !hal.device> key("hal.device.id" :: "local*") : i1, i1 = false
    cf.cond_br %value, ^bb3, ^bb4(%false : i1)
  ^bb3:  // pred: ^bb2
    %ok_0, %value_1 = hal.device.query<%device_n : !hal.device> key("hal.executable.format" :: "vmvx-bytecode-fb") : i1, i1 = false
    cf.br ^bb4(%value_1 : i1)
  ^bb4(%7: i1):  // 2 preds: ^bb2, ^bb3
    %8 = arith.cmpi eq, %2, %c0 : index
    %9 = arith.select %7, %c1, %c0 : index
    %10 = arith.addi %2, %9 : index
    %11 = arith.andi %7, %8 : i1
    %12 = arith.select %11, %device_n, %0 : !hal.device
    %13 = arith.addi %1, %c1 : index
    cf.br ^bb1(%13, %10, %12 : index, index, !hal.device)
  ^bb5:  // pred: ^bb1
    cf.cond_br %4, ^bb6, ^bb7
  ^bb6:  // pred: ^bb5
    util.status.check_ok %c18_i32, "HAL device `__device_0` not found or unavailable: #hal.device.target<\22local\22, [#hal.executable.target<\22vmvx\22, \22vmvx-bytecode-fb\22, {iree.encoding.resolver = #iree_cpu.vmvx_encoding_resolver<>, ukernels = \22none\22}>]>"
    cf.br ^bb7
  ^bb7:  // 2 preds: ^bb5, ^bb6
    util.global.store %3, @__device_0 : !hal.device
    util.return
  }
  util.global private @__device_0_query_0_hal_executable_format_vmvx_bytecode_fb : i1
  util.initializer {
    %__device_0 = util.global.load @__device_0 : !hal.device
    %ok, %value = hal.device.query<%__device_0 : !hal.device> key("hal.executable.format" :: "vmvx-bytecode-fb") : i1, i1 = false
    util.global.store %value, @__device_0_query_0_hal_executable_format_vmvx_bytecode_fb : i1
    util.return
  }
  util.global private @__device_0_executable_0_in_place_computation_linked_vmvx : !hal.executable
  util.initializer {
    %0 = util.null : !hal.executable
    %c14_i32 = arith.constant 14 : i32
    %c-1 = arith.constant -1 : index
    %c-1_i64 = arith.constant -1 : i64
    %__device_0_query_0_hal_executable_format_vmvx_bytecode_fb = util.global.load @__device_0_query_0_hal_executable_format_vmvx_bytecode_fb : i1
    %__device_0 = util.global.load @__device_0 : !hal.device
    %1 = arith.select %__device_0_query_0_hal_executable_format_vmvx_bytecode_fb, %c0, %c-1 : index
    %2 = arith.cmpi eq, %1, %c0 : index
    cf.cond_br %2, ^bb1, ^bb2
  ^bb1:  // pred: ^bb0
    %executable = hal.executable.create device(%__device_0 : !hal.device) affinity(%c-1_i64) target(@in_place_computation_linked_vmvx::@vmvx_bytecode_fb) : !hal.executable
    cf.br ^bb3(%executable : !hal.executable)
  ^bb2:  // pred: ^bb0
    util.status.check_ok %c14_i32, "HAL device `__device_0` does not support any variant of executable `in_place_computation_linked_vmvx`; available formats: [vmvx-bytecode-fb]"
    cf.br ^bb3(%0 : !hal.executable)
  ^bb3(%3: !hal.executable):  // 2 preds: ^bb1, ^bb2
    util.global.store %3, @__device_0_executable_0_in_place_computation_linked_vmvx : !hal.executable
    util.return
  }
  util.func public @in_place_computation(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer) -> !hal.buffer_view attributes {iree.abi.stub, iree.reflection = {iree.abi.declaration = "sync func @in_place_computation(%input0: tensor<64xf32>, %input1: tensor<64xf32> {iree.abi.output = 0 : index}, %input2: !hal.buffer {iree.abi.transients = unit}) -> (%output0: tensor<64xf32>)", iree.abi.transients.size = @in_place_computation_transients_size, iree.abi.transients.size.constant = 512 : index}} {
    %c64 = arith.constant 64 : index
    %c256 = arith.constant 256 : index
    %c-1_i64 = arith.constant -1 : i64
    %0 = util.null : !hal.fence
    %c-1_i32 = arith.constant -1 : i32
    %__device_0 = util.global.load immutable @__device_0 : !hal.device
    %__device_0_executable_0_in_place_computation_linked_vmvx = util.global.load immutable @__device_0_executable_0_in_place_computation_linked_vmvx : !hal.executable
    %element_type_f32 = hal.element_type<f32> : i32
    %dense_row_major = hal.encoding_type<dense_row_major> : i32
    hal.buffer_view.assert<%arg0 : !hal.buffer_view> message("input0") shape([%c64]) type(%element_type_f32) encoding(%dense_row_major)
    %buffer = hal.buffer_view.buffer<%arg0 : !hal.buffer_view> : !hal.buffer
    %allocator = hal.device.allocator<%__device_0 : !hal.device> : !hal.allocator
    hal.buffer.assert<%buffer : !hal.buffer> message("tensor") allocator(%allocator : !hal.allocator) minimum_length(%c256) type(DeviceVisible) usage("TransferSource|TransferTarget|Transfer|DispatchStorageRead|DispatchStorageWrite|DispatchStorage")
    %buffer_0 = hal.buffer_view.buffer<%arg1 : !hal.buffer_view> : !hal.buffer
    hal.buffer.assert<%buffer_0 : !hal.buffer> message("tensor") allocator(%allocator : !hal.allocator) minimum_length(%c256) type(DeviceVisible) usage("TransferSource|TransferTarget|Transfer|DispatchStorageRead|DispatchStorageWrite|DispatchStorage")
    %length = hal.buffer.length<%arg2 : !hal.buffer> : index
    hal.buffer.assert<%arg2 : !hal.buffer> message("tensor") allocator(%allocator : !hal.allocator) minimum_length(%length) type("DeviceVisible|DeviceLocal") usage("TransferSource|TransferTarget|Transfer|DispatchStorageRead|DispatchStorageWrite|DispatchStorage")
    %cmd = hal.command_buffer.create device(%__device_0 : !hal.device) mode("OneShot|AllowInlineExecution") categories("Transfer|Dispatch") affinity(%c-1_i64) : !hal.command_buffer
    hal.command_buffer.dispatch<%cmd : !hal.command_buffer> target(%__device_0_executable_0_in_place_computation_linked_vmvx : !hal.executable)[%c0] workgroups([%c1, %c1, %c1]) bindings([
      (%buffer : !hal.buffer)[%c0, %c256], 
      (%arg2 : !hal.buffer)[%c0, %length]
    ]) flags("None")
    hal.command_buffer.execution_barrier<%cmd : !hal.command_buffer> source("Dispatch|Transfer|CommandRetire") target("CommandIssue|Dispatch|Transfer") flags("None")
    hal.command_buffer.dispatch<%cmd : !hal.command_buffer> target(%__device_0_executable_0_in_place_computation_linked_vmvx : !hal.executable)[%c1] workgroups([%c1, %c1, %c1]) bindings([
      (%arg2 : !hal.buffer)[%c0, %length], 
      (%arg2 : !hal.buffer)[%c0, %length]
    ]) flags("None")
    hal.command_buffer.execution_barrier<%cmd : !hal.command_buffer> source("Dispatch|Transfer|CommandRetire") target("CommandIssue|Dispatch|Transfer") flags("None")
    hal.command_buffer.dispatch<%cmd : !hal.command_buffer> target(%__device_0_executable_0_in_place_computation_linked_vmvx : !hal.executable)[%c2] workgroups([%c1, %c1, %c1]) bindings([
      (%arg2 : !hal.buffer)[%c0, %length], 
      (%buffer : !hal.buffer)[%c0, %c256], 
      (%buffer_0 : !hal.buffer)[%c0, %c256]
    ]) flags("None")
    hal.command_buffer.execution_barrier<%cmd : !hal.command_buffer> source("Dispatch|Transfer|CommandRetire") target("CommandIssue|Dispatch|Transfer") flags("None")
    hal.command_buffer.finalize<%cmd : !hal.command_buffer>
    %fence = hal.fence.create device(%__device_0 : !hal.device) flags("None") : !hal.fence
    hal.device.queue.execute<%__device_0 : !hal.device> affinity(%c-1_i64) wait(%0) signal(%fence) commands(%cmd) flags("None")
    %status = hal.fence.await until([%fence]) timeout_millis(%c-1_i32) flags("None") : i32
    util.status.check_ok %status, "failed to wait on timepoint"
    %view = hal.buffer_view.create buffer(%buffer_0 : !hal.buffer)[%c0, %c256] shape([%c64]) type(%element_type_f32) encoding(%dense_row_major) : !hal.buffer_view
    util.return %view : !hal.buffer_view
  }
  util.func public @in_place_computation_transients_size(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer) -> index {
    %c512 = arith.constant 512 : index
    util.return %c512 : index
  }
}
```

MaheshRavishankar (approved):
Ok, I looked through the sample, and this is a good example of how to do this. 

@sjain-stanford  and @rkayaith  I think this is something you both want to look at. Look at the sample that was added here, specifically the samples/external_transients/main.c, the query_device_size method. This should be something that should be usable within Fusilli/BOO driver right? Please let me/us know if you see any issues.

PR #22626, opened 2025-11-12
[DT] Allow to enable/disable interleaving separately for M/N/K dimensions, for each operand

### LINKED ISSUE #22071: `DataTiledMMAAttr`: uniformly support interleaving and non-interleaving `intrinsics` unrolling parameters.
Currently, `DataTiledMMAAttr` has `intrinsics_{m,n,k}` parameter with a discrepancy in their semantics, as the `k` one is interleaving layouts, not the others:
https://github.com/iree-org/iree/blob/4e986f8cb180a17129877b0ce4e58580e5ad38a9/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td#L270-L274

For MXFP4 data-tiling, it seems that we have a use case to at least experiment with interleaving on the M and N dimensions, for contiguous access of 4xi32 scale words. This is a good occasion to finally resolve that discrepancy.

bjacob:
Fixes #22071.

Along the way, major simplifications to `getSwizzleImpl`. The fact that these changes are legal and even NFC on existing cases is nontrivial so let us explain why that is. Consider the existing code: https://github.com/iree-org/iree/blob/82216e4400d16ed7ed51fb97c9d3d5613ffaad72/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp#L217-L231

As the comment in the above code explains, the condition `if (!isScaled || isLhsScale) {` ensured that in a scaled matmul, only the LHS/RHS scales operands get interleaving but not the LHS/RHS elements themselves, while of course in a non-scaled matmul, it is the LHS/RHS elements that get interleaving.

That was a bit of magic that was reasonable "V1" stuff for scaled matmuls, but was technical debt to carry forward. This PR removes this magic by moving the decision of which operands to interleave, explicitly into the IR. Thus, scaled matmuls get `operands_interleaving_intrinsics_k = [2, 3]` meaning the scale operands while non-scaled matmuls get `operands_interleaving_intrinsics_k = [0, 1]` meaning the LHS/RHS element operands.

The other simplification in the above code is to rewrite
```c++
int interleavingIdx =getInnermostNonInternalDimIdx(swizzle.expandShape[1]);
if (isLhsScale) {
  interleavingIdx = swizzle.expandShape[1].size() - 1;
}
```

into just

```c++
getInnermostNonInternalDimIdx(swizzle.expandShape[1]);
```

That is, just dropping the `if(isLhsScale)` branch.  Why is that correct? That branch is by definition only taken on LHS scale operands of scaled matmuls. The values would only differ if such operands had a tile layout involving an "internal" dimension, which in the terminology of SingleSubgroupLayout is an `element` dimension. But no such case exists at present. On the contrary, existing mfma_scale instructions show how the scaling dimension in the LHS element matrix is an internal dimension, and is exactly the dimension that doesn't arise in the LHS scale matrix.
https://github.com/iree-org/iree/blob/82216e4400d16ed7ed51fb97c9d3d5613ffaad72/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp#L1353-L1366


compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
Yu-Zhewen: The description here also needs to be refreshed.

Yu-Zhewen (approved):
Really nice improvement! This makes tuning much easier.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
Yu-Zhewen: Same here

hanhanW (approved):

PR #22632, opened 2025-11-13
[LLVMCPU] Support tile-and-fuse anchoring on producer ops

### LINKED ISSUE #22484: [CPU] Improve TileAndFuseProducerConsumer pass to handle producers that have more dimensions
Improve LLVMCPUTileLastOpAndFuseProducerConsumerPass to tile and fuse operations before the root operation, particularly handling broadcast and dequant ops in linalg.generic form.

Mainly need to clean this TODO:

https://github.com/iree-org/iree/blob/731e5cae1a7d7da9a2a947a9708e6a3c85984366/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUTileAndFuseProducerConsumer.cpp#L241-L245

Goal: land https://github.com/hanhanW/iree/commit/87a2182946cf04be4b3844a4c5fe9c8a1fa0a9bf (without changes in KernelDispatch.cpp) and add proper tests.

hanhanW:
This patch improves the CPU pipeline by enabling tile-and-fuse to anchor on producer operations (ops that come before the root op in the compute sequence), extending the existing support for consumer-based anchoring.

Key changes:

1. Enhanced getLastAnchorOp to support anchoring before and after the root op:
   - getLastAnchorOpAfterRootOp
   - getLastAnchorOpBeforeRootOp

2. Updated runOnOperation to handle multiple anchor points:
   - Find both consumer anchors (after root) and producer anchors (before root)
   - Process all anchor ops for tile-and-fuse transformations
   - Enables fusion of producers at vector_inner_parallel tiling level

3. Added VectorInnerParallelTiles level to Mmt4dTilingExpert pipeline:
   - Models tiling/fusion for dimensions not captured in root op config
   - Uses last operation with the tiling level as anchor
   - Runs after VectorReductionTiles and before ForallToFor conversion

4. Rename test file from anchoring_last_op to anchoring_non_root_op

Fixes https://github.com/iree-org/iree/issues/22484

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUTileAndFuseProducerConsumer.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUTileAndFuseProducerConsumer.cpp
@@ -52,31 +52,44 @@ static Operation *getRootOp(ArrayRef<Operation *> computeOps,
   return rootOp;
 }
 
-/// Returns the last operation that has `level` tiling level in lowering config.
-/// When `startBeforeRootOp` is true, only considers operations that appear
-/// before the root op (or ukernel ops) in the compute sequence, enabling
-/// anchoring on producer operations. When false, considers all operations
-/// including those after the root, enabling anchoring on consumer operations.
-/// Returns nullptr if no suitable anchor operation is found.
-static Operation *getLastAnchorOp(ArrayRef<Operation *> computeOps,
-                                  IREE::CPU::TilingLevel level,
-                                  bool startBeforeRootOp) {
-  bool meet = true;
-  if (startBeforeRootOp) {
-    meet = false;
-  }
+/// Returns the last operation that has `level` tiling level in lowering config
+/// after the root op (or ukernel ops) in the compute sequence.
+static Operation *getLastAnchorOpAfterRootOp(ArrayRef<Operation *> computeOps,
+                                             IREE::CPU::TilingLevel level) {
   for (Operation *op : llvm::reverse(computeOps)) {
     IREE::Codegen::LoweringConfigAttrInterface loweringConfig =
         getLoweringConfig(op);
-    if (meet && loweringConfig &&
+    if ((loweringConfig && loweringConfig.hasWorkgroupTilingLevel()) ||
+        isa<IREE::Codegen::UKernelGenericOp>(op)) {
+      break;
+    }
+    if (loweringConfig &&
         loweringConfig.hasTilingLevel(llvm::to_underlying(level))) {
       return op;
     }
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUTileAndFuseProducerConsumer.cpp
```diff
@@ -57,16 +53,31 @@ static Operation *getRootOp(ArrayRef<Operation *> computeOps,
...
-/// Returns nullptr if the operation does not exist.
+/// When `startBeforeRootOp` is true, only considers operations that appear
+/// before the root op (or ukernel ops) in the compute sequence, enabling
+/// anchoring on producer operations. When false, considers all operations
+/// including those after the root, enabling anchoring on consumer operations.
+/// Returns nullptr if no suitable anchor operation is found.
 static Operation *getLastAnchorOp(ArrayRef<Operation *> computeOps,
-                                  IREE::CPU::TilingLevel level) {
+                                  IREE::CPU::TilingLevel level,
+                                  bool startBeforeRootOp) {
```
Yu-Zhewen: Instead of having a flag here, I fell it might be clearer to separate this into two functions: `getLastAnchorOpBeforeRoot` and `getLastAnchorOp`.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUTileAndFuseProducerConsumer.cpp
```diff
@@ -234,37 +245,39 @@ void LLVMCPUTileAndFuseProducerConsumer::runOnOperation() {
   IRRewriter rewriter(funcOp);
 
   SmallVector<Operation *> computeOps = getComputeOps(funcOp);
-  Operation *anchorOp;
+  SmallVector<Operation *> anchorOps;
   if (anchorOnRootOp) {
-    anchorOp = getRootOp(computeOps, tilingLevel);
+    Operation *anchorOp = getRootOp(computeOps, tilingLevel);
```
Yu-Zhewen, line 263: I know this function already exists, but renaming it to something like `getRootAnchorOp` would be more descriptive (a root op && has the given tiling level).

hanhanW: sgtm!

hanhanW: After thinking a bit, I'd like to keep it `getRootOp`. In codegen context, there is a single root op; the definition of the root op is that it has distribution level of tiling config.

Yu-Zhewen (approved):
Thanks!

PR #22634, opened 2025-11-13
[Flow] Move ReplicateGlobalsPerAffinity pass to Flow

sommerlukas:
The experimental pass to replicate globals per unique device affinity lived in Stream so far, but operates on Flow dialect before conversion to Stream.

Move the pass to Flow and place it right after executable cleanup in the flow pipeline if enabled through the experimental flag.


sommerlukas:
The pass still uses the Stream Affinity interfaces and attributes for device affinity.

sommerlukas:
I verified that the pass still has the desired effect with the motivating example from https://github.com/iree-org/iree/issues/22081 as described by @hanhanW [here](https://github.com/iree-org/iree/issues/22081#issuecomment-3524038619) and [here](https://github.com/iree-org/iree/pull/22623#issuecomment-3524036667).

hanhanW (approved):
LGTM, thanks

PR #22636, opened 2025-11-13
Update split reduction heuristic for extreme large GEMMs

yzhang93:
This PR added two heuristic improvements for extreme large GEMMs.

-  If M/N/K sizes are all very large, e.g., M:N:K as 4096:16384:150000, we don't need to split the reduction dimension as the workload tends to distributed across many workgroups and degrades the performance.
- For extreme large K and relatively small M/N size, e.g., 128:134:16800000, we want to increase the number of splits for more parallels on reduction.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -59,7 +59,7 @@ findSmallestFactorWithLowerBound(int64_t x, int64_t lowerBound) {
   assert(lowerBound > 0);
   // We expect all numbers here to be relatively small, so just do trial
   // division (with a limit just to be safe).
-  static constexpr int64_t kMaxIterations = 1 << 22;
+  static constexpr int64_t kMaxIterations = 1 << 15;
   int64_t upperBound = std::min(x, kMaxIterations);
   for (int64_t i = lowerBound; i <= upperBound; i++) {
     if (x % i == 0) {
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -59,7 +59,7 @@ findSmallestFactorWithLowerBound(int64_t x, int64_t lowerBound) {
   assert(lowerBound > 0);
   // We expect all numbers here to be relatively small, so just do trial
   // division (with a limit just to be safe).
-  static constexpr int64_t kMaxIterations = 1 << 15;
+  static constexpr int64_t kMaxIterations = 1 << 22;
   int64_t upperBound = std::min(x, kMaxIterations);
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -398,6 +398,17 @@ struct SetSplitReductionSizesPass final
     // The constants below are determined based on empirical data.
     const int64_t ratioThreshold = 384;
     const int64_t largeKSize = 24576;
+    const int64_t largeMNSize = 1024;
```
jerryyin, line 401: This constant looks fairly aggressive to me: each MFMA only can handle mxn of 16x16 on mi300x. By setting this limit to 1024, it will allow 64 MFMAs to consume the entire M/N. I think we should limit it further to 1 workgroup tile, say 256. This allows 1 tile per parallel dimension to still not go to split K (with decent tile size). 

BTW, this is purely based on my intuition. So if you find a counter example, let me know.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -59,7 +59,7 @@ findSmallestFactorWithLowerBound(int64_t x, int64_t lowerBound) {
   assert(lowerBound > 0);
   // We expect all numbers here to be relatively small, so just do trial
   // division (with a limit just to be safe).
-  static constexpr int64_t kMaxIterations = 1 << 15;
+  static constexpr int64_t kMaxIterations = 1 << 22;
```
jerryyin: Will we stuck in this loop for a long time if x isn't divisible by any i? I'm a bit worried that upperBound can be very large and in the mean while the loop below only increment one per iteration.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
yzhang93: There are fairly number of configs with M/N close to 1024 that benefit from split reduction. For example `aten::mm '[[1024, 150000], [150000, 1024]]'`, the runtime **without split** is 5408us and **with split** is 1356us.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
yzhang93: The concern makes sense. Actually I checked it again, and there's no need to increase this value anymore. Initially, I modified it because the `limitParallelLoops` was 16, and when K size is extremely large, e.g., 16,800,000, the tile size lower bound would be `16800000/16 = 1050000` which is larger than the upper bound `1>>15`, so it didn't find the solution. Now as I increase the `limitParallelLoops` to 2048, the lower bound would be smaller than the upper bound. I'll change the value back.

jerryyin (approved):
LGTM

PR #22639, opened 2025-11-14
[CI] Optimize and clean up asan and tsan build scripts

kuhar:
Make the build_and_test script for sanitizers use good defaults, both in CI and locally.

Previously, the asan script would take 19m 55s to run on my local machine. With this PR, it takes 8m 5s.
In CI, it would previously take 29m 58s and with this PR it takes 17m 12s.

* `sccache` is really slow in a local setup -- don't unconditionally enable it. The proper cmake flags are already exported by `setup_sccache.sh` invoked by the workflow.
* Enable LLD in asan builds.
* Enable split dwarf and thin archives for better handling of debug info.
* Export compile commands for better IDE/lsp integration (why not).
* Account for the new build error when `rocminfo` is detected but no hip target is set.


kuhar:
In CI asan takes 17m 12s now (full rebuild because of cmake flags) https://github.com/iree-org/iree/actions/runs/19352760651/job/55367879854?pr=22639 vs. 29m 58s after the most recent integrate (~clean cache): https://github.com/iree-org/iree/actions/runs/19305542530/job/55211932126.

MaheshRavishankar (approved):

amd-eochoalo (approved):

build_tools/cmake/build_and_test_tsan.sh
```diff
@@ -24,16 +24,18 @@ IREE_HAL_DRIVER_CUDA="${IREE_HAL_DRIVER_CUDA:-${OFF_IF_DARWIN}}"
 IREE_HAL_DRIVER_HIP="${IREE_HAL_DRIVER_HIP:-${OFF_IF_DARWIN}}"
 IREE_TARGET_BACKEND_CUDA="${IREE_TARGET_BACKEND_CUDA:-${OFF_IF_DARWIN}}"
 IREE_TARGET_BACKEND_ROCM="${IREE_TARGET_BACKEND_ROCM:-${OFF_IF_DARWIN}}"
+IREE_HIP_TEST_TARGET_CHIP="${IREE_HIP_TEST_TARGET_CHIP:-}"
```
bjacob, line 27: Do you anticipate TSan becoming available on AMDGPU ?

build_tools/cmake/build_and_test_tsan.sh
```diff
@@ -47,30 +49,33 @@ CMAKE_ARGS=(
...
+echo "::group::Building test deps"
 "$CMAKE_BIN" --build "${BUILD_DIR}" --target iree-test-deps -- -k 0
+echo "::endgroup::"
 
 # Disable actually running GPU tests. This tends to yield TSan reports that are
 # specific to one's particular GPU driver and therefore hard to reproduce across
-# machines and often un-actionable anyway.
-# See e.g. https://github.com/iree-org/iree/issues/9393
+# machines and often non-actionable anyway.
+# See e.g., https://github.com/iree-org/iree/issues/9393.
```
bjacob: I think you can have either no commas at all, or a comma on each side of e.g., but maybe not just this?

build_tools/cmake/build_and_test_tsan.sh
```diff
@@ -47,30 +49,33 @@ CMAKE_ARGS=(
...
   "-DRUN_HAVE_STD_REGEX=0"
   "-DRUN_HAVE_POSIX_REGEX=0"
   "-DCOMPILE_HAVE_GNU_POSIX_REGEX=0"
-  "-DCMAKE_CXX_COMPILER_LAUNCHER=sccache"
-  "-DCMAKE_C_COMPILER_LAUNCHER=sccache"
+
+  "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
 )
 
+echo "::group::Configuring CMake"
```
bjacob, line 63: What does ::group:: do?  Is it an escape sequence interpreted by some terminal emulator?

bjacob (commented):
Another substantial build latency gain that we have been leaving on the table for years is where we build `all` first, then `iree-test-deps`:

```
ninja
...
ninja iree-test-deps
```

As opposed to just building all at once:

```
ninja all iree-test-deps
```

Which allows test-deps to start building as soon as iree-compile is ready, while some late 'all' targets finish building.

(I don't know if `all` is redundant/implied by `iree-test-deps` or not).

CC @ScottTodd who disagreed in the previous iteration of the conversation, with the argument that keeping the two steps separate helps interpreting build failures.

build_tools/cmake/build_and_test_tsan.sh
kuhar: We can build those e2e tests but not necessarily execute them. If someone feels adventurous, they can try running them locally.

build_tools/cmake/build_and_test_tsan.sh
kuhar: I don't know if it has any function, it's just to match the style of the other scripts

build_tools/cmake/build_and_test_tsan.sh
bjacob: Ah, that is a good point.

kuhar:
> Another substantial build latency gain that we have been leaving on the table for years is where we build `all` first, then `iree-test-deps`:
> 
> ```
> ninja
> ...
> ninja iree-test-deps
> ```
> 
> As opposed to just building all at once:
> 
> ```
> ninja all iree-test-deps
> ```
> 
> Which allows test-deps to start building as soon as iree-compile is ready, while some late 'all' targets finish building.

I will check this, but separately from this PR.

build_tools/cmake/build_and_test_tsan.sh
bjacob: I'm guessing, but it's just a guess, that it is parsed by the GitHub actions web view and is what drives the foldable sections in the logs :
https://github.com/iree-org/iree/actions/runs/19369474312/job/55421252106
{A7D0C06B-2C84-4C23-AEC5-9C120E102E75}

So yes, it's reasonable to follow what other scripts do.

kuhar:
@bjacob I checked and it only saves 4s of out of 8 minutes of build time on my system (which is well within the noise), so IMO not worth changing

PR #22642, opened 2025-11-14
[Dispatch Creation] Add pass to fold reshapes into barriers

IanWood1:
Adds a new pass, `FoldReshapesIntoTensorBarriersPass`, that moves `tensor.expand_shape` and `tensor.collapse_shape` operations through compute barrier operations. Specifically, it moves the reshape ops above `iree_tensor_ext.compute_barrier.start` ops and below `iree_tensor_ext.compute_barrier.end, effectively containing them to the edges of the program.

**Note:** This pass is not currently integrated into the main pipeline, as some users depend on reshapes to block transpose fusion for performance reasons. This can be added after determining that it doesn't negatively impact these cases.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
@@ -7,6 +7,7 @@
 #include "iree/compiler/Dialect/TensorExt/IR/TensorExtOps.h"
 #include "iree/compiler/DispatchCreation/Passes.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/Dominance.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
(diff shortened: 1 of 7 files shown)


compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // Update reshape's operand to use compute_barrier.start's input
+    rewriter.modifyOpInPlace(
+        op, [&]() { op->getOpOperand(0).set(barrierStartOp.getValue()); });
+
+    IRRewriter::InsertionGuard guard(rewriter);
+    rewriter.setInsertionPointAfter(op);
+    Value reshapeResult = op->getResult(0);
+    auto newBarrier = IREE::TensorExt::ComputeBarrierStartOp::create(
+        rewriter, barrierStartOp.getLoc(), reshapeResult);
+    reshapeResult.replaceUsesWithIf(
```
MaheshRavishankar: I think its better to go through the rewriter to do this replace. It sometimes makes the pattern rewriter behave strangely when you dont do that.

compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
```diff
@@ -0,0 +1,106 @@
...
+        newBarrier.getResult(), [&](OpOperand &use) {
+          return use.getOwner() != newBarrier &&
```
MaheshRavishankar: Do you need to check that the use is dominated by the barrier?

MaheshRavishankar (approved):
Left a few comments. Trust they will be addressed before landing, but otherwise looks good.

MaheshRavishankar:
Actually while reviewing https://github.com/iree-org/iree/pull/22679 I was wondering if we want to start with just limiting the reshape propagation across barriers for just unit-dims.

The reasoning is that really we want to fold the unit-dimensions into the ABI ops, but that has issues. So we are using these operations to move the reshapes of unit-dims "close" to the ABI ops and adding a barrier to stop them from propagating back down.

compiler/src/iree/compiler/DispatchCreation/FoldReshapesIntoTensorBarriers.cpp
IanWood1: Good point, that is better than checking for dim ops specifically.

IanWood1:
> Actually while reviewing #22679 I was wondering if we want to start with just limiting the reshape propagation across barriers for just unit-dims.
> 
> The reasoning is that really we want to fold the unit-dimensions into the ABI ops, but that has issues. So we are using these operations to move the reshapes of unit-dims "close" to the ABI ops and adding a barrier to stop them from propagating back down.

I think there were some cases in fusilli where reshapes were causing problems even with all non-unit dims. But, that might be a more reasonable place to start.

MaheshRavishankar (approved):

PR #22651, opened 2025-11-15
[CI] Optimize cmake flags for debug info builds

kuhar:
Enable split dwarf and thin archives.

Do not reduce test parallelism in the clang job.

### DIFF
--- a/.github/workflows/ci_linux_x64_clang.yml
+++ b/.github/workflows/ci_linux_x64_clang.yml
@@ -42,6 +42,8 @@ jobs:
             -DIREE_BUILD_PYTHON_BINDINGS=ON \
             -DIREE_ENABLE_LLD=ON \
             -DIREE_ENABLE_ASSERTIONS=ON \
+            -DIREE_ENABLE_SPLIT_DWARF=ON \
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_linux_x64_clang.yml
```diff
@@ -54,8 +56,6 @@ jobs:
           sccache --show-stats
       - name: Run CTest
         run: ./build_tools/cmake/ctest_all.sh "${BUILD_DIR}"
-        env:
-          CTEST_PARALLEL_LEVEL: 32
```
amd-eochoalo, line 58: Have you run done a test run for this one and see noticed the differences? I know this is just an anecdote but sometimes increasing the parallelism made the process slower due to competition for resources. But it highly depends on the machine.

amd-eochoalo (approved):

.github/workflows/ci_linux_x64_clang.yml
kuhar: nproc is better for lit tests only, but I don't have exact numbers

PR #22662, opened 2025-11-15
Drop unnecessary namespaces from cast functions in dialect vm vmvx etc. NFC. 10/10

kuhar:
This removes llvm:: and mlir:: namespace prefixes from casting functions (isa, cast, dyn_cast, cast_or_null, dyn_cast_or_null, isa_and_nonnull, dyn_cast_if_present, cast_if_present, isa_and_present) where they are unnecessary due to 'using' declarations in mlir/Support/LLVM.h.

These functions are brought into scope by headers like mlir/Support/LLVM.h which is included (directly or transitively) by most MLIR-based code.


bangtianliu (approved):
Thanks!

hanhanW (approved):

PR #22663, opened 2025-11-16
[RISCV] Clean up toolchain CMake configuration

HanKuanChen:
- Remove redundant CMAKE_SYSTEM_PROCESSOR setting
- Remove RISCV_HOST_TAG variable and hardcode "linux" path
- Add quotes around CMAKE_FIND_ROOT_PATH
- Remove empty variable initializations (RISCV_COMPILER_FLAGS, RISCV_COMPILER_FLAGS_CXX, RISCV_LINKER_FLAGS, RISCV_LINKER_FLAGS_EXE)
- Remove redundant RISCV_COMPILER_FLAGS since CMAKE_BUILD_TYPE already sets optimization flag


build_tools/cmake/riscv.toolchain.cmake
```diff
@@ -16,15 +16,11 @@ if(RISCV_TOOLCHAIN_INCLUDED)
 endif(RISCV_TOOLCHAIN_INCLUDED)
 set(RISCV_TOOLCHAIN_INCLUDED true)
 
-set(CMAKE_SYSTEM_PROCESSOR riscv)
```
mshockwave, line 19: is there any specific reason we're getting rid of this? According to cmake's document, we have to set it to target architecture when cross-compiling: https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html

build_tools/cmake/build_riscv.sh
```diff
@@ -23,7 +23,6 @@ BUILD_DIR="${1:-${IREE_TARGET_BUILD_DIR:-build-riscv}}"
 BUILD_TYPE="${IREE_BUILD_TYPE:-RelWithDebInfo}"
 RISCV_PLATFORM="${IREE_TARGET_PLATFORM:-linux}"
 RISCV_ARCH="${IREE_TARGET_ARCH:-riscv_64}"
-RISCV_COMPILER_FLAGS="${RISCV_COMPILER_FLAGS:--O3}"
```
mshockwave, line 26: just want to make sure I understand correctly: the reason you want to remove this flag is because ideally, the optimization level should be controlled by `CMAKE_BUILD_TYPE`? (And users could customize the compiler flags with command line flag like `-DCMAKE_CXX_FLAGS` anyway?)

build_tools/cmake/riscv.toolchain.cmake
HanKuanChen: The value is set again in line 61 - 65.
```
if(RISCV_CPU MATCHES "riscv_64")
  set(CMAKE_SYSTEM_PROCESSOR riscv64)
elseif(RISCV_CPU MATCHES "riscv_32")
  set(CMAKE_SYSTEM_PROCESSOR riscv32)
endif()
```
Here is redundant.

build_tools/cmake/build_riscv.sh
HanKuanChen: Yes. Because `CMAKE_BUILD_TYPE` is already used. `CMAKE_BUILD_TYPE` will set optimization level and debug option. If we set `O3` again, there will have two optimization level in the command, which confuses users.

mshockwave (approved):
LGTM

HanKuanChen:
@ScottTodd @hanhanW It seems like we need your approval. Only you have write permission.

build_tools/cmake/build_riscv.sh
hanhanW: Can you add this information to PR description? It seems helpful, thanks.

HanKuanChen: done

hanhanW (approved):

PR #22672, opened 2025-11-17
[Codegen] Fold bitcast into bufferized tensor load

### LINKED ISSUE #22601: Support folding bitcast into LoadFromBufferOp
We already support folding bitcast into DispatchTensorLoadOp:
https://github.com/iree-org/iree/blob/3a1a6f49ef3b01a6ee8208b04714d028394e0bc0/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp#L830

However, when combine layout transformation is enabled,
https://github.com/iree-org/iree/blob/3a1a6f49ef3b01a6ee8208b04714d028394e0bc0/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L1129-L1130

converting DispatchTensorLoadOp to LoadFromBufferOp.

After the conversion, we will need a new pattern to fuse bitcast ops into LoadFromBufferOp, similar to the existing handling for DispatchTensorLoadOp.

This fusion is required by the FP4 ukernel, which relies on bitcast ops to convert data to i8 in order to use `amdgpu.gather_to_lds`.

Yu-Zhewen:
We already support folding bitcast operations into `iree_tensor_ext.dispatch.tensor.load`. However, when `--iree-llvmgpu-combine-layout-transformation` is enabled, these tensor loads are bufferized before the folding pattern can be applied. Therefore, this PR extends the folding support.

One use case is the FP4 pingpong ukernel, which requires bitcasting FP4 data to `i8` to use `amdgpu.gather_to_lds` operations. Without this folding, `ComprehensiveBufferize` would fail.

Closes: #22601

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
@@ -1298,15 +1298,15 @@ void populateReshapeToInterfaceTensorPatterns(RewritePatternSet &patterns) {
                   FoldExpandShapeIntoInterfaceTensorLoad,
                   FoldExpandShapeIntoInterfaceTensorStore,
                   FoldInnerBitcastIntoInterfaceTensorLoad,
-                  FoldInnerBitcastIntoInterfaceTensorStore,
-                  FoldInnerBitcastIntoLoadFromBuffer>(patterns.getContext());
+                  FoldInnerBitcastIntoInterfaceTensorStore>(
+      patterns.getContext());
 }
 
 void populateFoldTensorReshapeIntoBufferPatterns(RewritePatternSet &patterns) {
   patterns.insert<
       FoldCollapseShapeIntoLoadFromBuffer, FoldExpandShapeIntoLoadFromBuffer,
-      FoldCollapseShapeIntoStoreToBuffer, FoldExpandShapeIntoStoreToBuffer>(
-      patterns.getContext());
+      FoldCollapseShapeIntoStoreToBuffer, FoldExpandShapeIntoStoreToBuffer,
+      FoldInnerBitcastIntoLoadFromBuffer>(patterns.getContext());
 }
 
 } // namespace mlir::iree_compiler
--- a/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
@@ -1003,6 +1003,9 @@ struct FoldInnerBitcastIntoLoadFromBuffer
                                 PatternRewriter &rewriter) const override {
     Value bitcastSrc = bitcastOp.getSource();
 
+    // TODO(#22712): This pattern matching is fragile. Simplify this by using
+    // a memref.bitcast or similar operation once available.
+
     // Step 1: Check for iree_codegen.load_from_buffer.
     auto loadOp = bitcastSrc.getDefiningOp<IREE::Codegen::LoadFromBufferOp>();
     if (!loadOp) {
--- a/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
@@ -10,8 +10,10 @@
 #include "iree/compiler/Dialect/HAL/IR/HALOps.h"
 #include "iree/compiler/Dialect/TensorExt/IR/TensorExtOps.h"
 #include "llvm/Support/FormatVariadic.h"
+#include "mlir/Dialect/AMDGPU/IR/AMDGPUDialect.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/Transforms/Transforms.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -968,6 +991,105 @@ struct FoldInnerBitcastIntoInterfaceTensorStore
...
+    auto bitcastSrcType = cast<RankedTensorType>(bitcastSrc.getType());
+    auto bitcastResType = cast<RankedTensorType>(bitcastOp.getType());
+
+    // Verify that only the inner most dimension is changed by the bitcast by
+    // comparing dynamic and static sizes for equality.
+    if (bitcastOp.getSourceDims() != bitcastOp.getResultDims() ||
+        bitcastSrcType.getShape().drop_back() !=
+            bitcastResType.getShape().drop_back() ||
+        ShapedType::isDynamic(bitcastSrcType.getShape().back())) {
+      return failure();
```
hanhanW: I think we can move the comment to error message, which makes debugging easier.

```
return rewriter.notify...(op, "expected that only the innermost dimension is changed by the bitcast op");
```

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -968,6 +991,105 @@ struct FoldInnerBitcastIntoInterfaceTensorStore
...
+    }
+
+    // Fail if the inner most dim doesn't match between subspan and tensor.
+    if (subspanType.getShape().back() != bitcastSrcType.getShape().back()) {
+      return failure();
+    }
```
hanhanW: ditto, move the comment to failure message.

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -968,6 +991,105 @@ struct FoldInnerBitcastIntoInterfaceTensorStore
...
+    int64_t newInnerSize = bitcastResType.getShape().back();
+    int64_t oldInnerSize = bitcastSrcType.getShape().back();
+
+    // Adjust the innermost dimension size.
+    MemRefLayoutAttrInterface newSubspanLayout;
+    SmallVector<int64_t> newSubspanShape(subspanType.getShape());
+    newSubspanShape.back() = newInnerSize;
+    // Scale offset and strides by (newInnerSize / oldInnerSize).
+    if (auto stridedLayout =
+            dyn_cast_or_null<StridedLayoutAttr>(subspanType.getLayout())) {
```
hanhanW: nit: we prefer `dyn_cast_if_present`.

`dyn_cast_or_null` has been marked that it will be deprecated for three years. I just did a cleanup for IREE today: https://github.com/iree-org/iree/commit/33db935a33ddab6a8358418c0ae6dc263e56b0bb

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -968,6 +991,105 @@ struct FoldInnerBitcastIntoInterfaceTensorStore
...
+      int64_t newOffset = stridedLayout.getOffset();
+      if (!ShapedType::isDynamic(newOffset)) {
+        newOffset = (newOffset * newInnerSize) / oldInnerSize;
+      }
```
hanhanW: It can be `ShapedType::isStatic` which avoids double negations. Same for the below use

hanhanW (changes requested):

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -968,6 +991,107 @@ struct FoldInnerBitcastIntoInterfaceTensorStore
...
+      return rewriter.notifyMatchFailure(bitcastOp, "no fat_raw_buffer_cast");
+    }
+    Value bufferCastSource = bufferCastOp.getSource();
+
+    // Step 3: Check for hal.interface.binding.subspan.
+    auto subspanOp =
+        bufferCastSource.getDefiningOp<IREE::HAL::InterfaceBindingSubspanOp>();
+    if (!subspanOp) {
+      return rewriter.notifyMatchFailure(bitcastOp, "no binding subspan");
+    }
```
Max191, line 1028: The pattern matching is a bit fragile here. I think for the tensor version, we need to fold the bitcast all the way into the binding so that bufferization doesn't choke on it, but for the bufferized cases I think we can get away with something simpler. It is one of the advantages of the early bufferization path.

I think you could just match for the load_from_buffer op, and then just insert a `memref.reinterpret_cast` op on the source of the load_from_buffer to do the bitcast, so you have `iree_codegen.load_from_buffer(memref.reinterpret_cast(%src))`.

I think there shouldn't be any issues with it, but you can try that and see if it works for your case. Then we can skip the matching on the long op chain.

Max191 (changes requested):

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
Yu-Zhewen: I tried using `memref.reinterpret_cast`, but then I realized it doesn't support casting element types...

https://github.com/llvm/llvm-project/blob/6665642ce40c70b65624a5aa67566725c5a87da5/mlir/lib/Dialect/MemRef/IR/MemRefOps.cpp#L1943-L1945

Max191: Ah, I see. I guess what we really need is a memref.bitcast op. I think the only op that would change the element type is a memref.view op, but that requires it to be contiguous. This is fine for now then, but could you open up a github issue and add a TODO comment referencing it to do this using some sort of bitcast once there is an op we can use for it?

Yu-Zhewen: Thanks! Done.

hanhanW (approved):

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -1171,8 +1298,8 @@ void populateReshapeToInterfaceTensorPatterns(RewritePatternSet &patterns) {
                   FoldExpandShapeIntoInterfaceTensorLoad,
                   FoldExpandShapeIntoInterfaceTensorStore,
                   FoldInnerBitcastIntoInterfaceTensorLoad,
-                  FoldInnerBitcastIntoInterfaceTensorStore>(
-      patterns.getContext());
+                  FoldInnerBitcastIntoInterfaceTensorStore,
+                  FoldInnerBitcastIntoLoadFromBuffer>(patterns.getContext());
```
Max191: nit: We should probably add this to the `populateFoldTensorReshapeIntoBufferPatterns` below, since these are affecting the load_from_buffer ops.

Max191 (approved):
LGTM, just one comment. Please address it before landing.

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
Yu-Zhewen: Ah, right. Thanks for pointing that out!

Issue #17769, opened 2024-06-28
ASAN issues on hip / rocm runtime


suryajasper:
### What happened?

I am seeing memory leaks due to the ROCm & HIP runtime backends in IREE. These issues can be reproduced simply by initializing and releasing a hip / rocm HAL device:
```
==1685580==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 792 byte(s) in 11 object(s) allocated from:
    #0 0x7eff0965c1e7 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x7efefb74d1ab  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x611ab)

Direct leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7eff0965c1e7 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x7efefb738b5a  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x4cb5a)
    #2 0x7efefb72f3c5  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x433c5)
    #3 0x7efefb744cee  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x58cee)
    #4 0x7eff0456e555  (/opt/rocm-6.1.0/lib/libamdhip64.so+0x33c555)

Direct leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7eff0965c1e7 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x7efefb738b5a  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x4cb5a)
    #2 0x7efefb72f3c5  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x433c5)
    #3 0x7efefb72a76f  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x3e76f)
    #4 0x7efefb744cee  (/opt/rocm-6.1.0/lib/libhsa-runtime64.so.1+0x58cee)
    #5 0x7eff0456e555  (/opt/rocm-6.1.0/lib/libamdhip64.so+0x33c555)

Indirect leak of 2464 byte(s) in 7 object(s) allocated from:
[...]

PR #22675, opened 2025-11-17
Suppress ROCm lsan errors in HIP driver tests

qedawkins:
There are known leaks in certain rocm libraries that can pollute test output. Add a suppressions file to ignore such warnings.

Related to https://github.com/iree-org/iree/issues/17769


kuhar (approved):
We should also enable this in CI once we have more self-hosted runners

PR #22679, opened 2025-11-17
[Preprocessing] Add compute_barrier in ConvertConvFilterToChannelsLast pass

yzhang93:
This PR adds `iree_tensor_ext.compute_barrier` around transpose ops to avoid reshape operations being propagated and thus block undesirable fusion.

### DIFF
--- a/compiler/src/iree/compiler/Preprocessing/Common/ConvertConvFilterToChannelsLast.cpp
+++ b/compiler/src/iree/compiler/Preprocessing/Common/ConvertConvFilterToChannelsLast.cpp
@@ -213,28 +213,20 @@ struct ConvertGenericChwfToFhwc : public OpRewritePattern<linalg::GenericOp> {
 
     Location loc = linalgOp.getLoc();
 
-    // Insert compute_barrier.start before creating transpose op to avoid
-    // movements of reshape ops and undesirable fusion.
-    auto barrierStartOp = IREE::TensorExt::ComputeBarrierStartOp::create(
-        rewriter, loc, filterVal);
-
     AffineMap transposedFilterMap = applyPermutationToResults(filterMap, perm);
-    Value transposedFilter =
-        createTransposeOp(rewriter, loc, barrierStartOp.getResult(), perm);
+    Value transposedFilter = createTransposeOp(rewriter, loc, filterVal, perm);
 
-    // Insert compute_barrier.end after the tranpose op and replace its uses.
-    auto barrierEndOp = IREE::TensorExt::ComputeBarrierEndOp::create(
+    // Insert compute_barrier.start to avoid propagation of reshape ops and
+    // undesirable fusion.
+    auto barrierStartOp = IREE::TensorExt::ComputeBarrierStartOp::create(
         rewriter, loc, transposedFilter);
-    transposedFilter.replaceUsesWithIf(
-        barrierEndOp.getResult(),
-        [&](OpOperand &use) { return use.getOwner() == linalgOp; });
 
     SmallVector<utils::IteratorType> iterators =
         linalgOp.getIteratorTypesArray();
 
     auto genericOp = linalg::GenericOp::create(
         rewriter, loc, outputVal.getType(),
-        ValueRange{inputVal, barrierEndOp.getResult()}, outputVal,
+        ValueRange{inputVal, barrierStartOp.getResult()}, outputVal,
         ArrayRef<AffineMap>{inputMap, transposedFilterMap, outputMap},
         iterators);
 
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Preprocessing/Common/ConvertConvFilterToChannelsLast.cpp
```diff
@@ -212,15 +213,28 @@ struct ConvertGenericChwfToFhwc : public OpRewritePattern<linalg::GenericOp> {
...
+        rewriter, loc, filterVal);
+
     AffineMap transposedFilterMap = applyPermutationToResults(filterMap, perm);
-    Value transposedFilter = createTransposeOp(rewriter, loc, filterVal, perm);
+    Value transposedFilter =
+        createTransposeOp(rewriter, loc, barrierStartOp.getResult(), perm);
+
+    // Insert compute_barrier.end after the tranpose op and replace its uses.
+    auto barrierEndOp = IREE::TensorExt::ComputeBarrierEndOp::create(
+        rewriter, loc, transposedFilter);
```
IanWood1: Not blocking, this just raised some questions:

I think this op needs better-defined semantics. When I was adding these two ops, the idea was that `iree_tensor_ext.compute_barrier.start` would demarcate the beginning of the program and `iree_tensor_ext.compute_barrier.end` would demarcate the end. These ops act as a one-way barrier for "metadata" ops (e.g. `tensor.expand_shape` and `tensor.collapse_shape`). [Add pass to fold reshapes into barriers](https://github.com/iree-org/iree/pull/22642) uses this to move reshapes up through "start" barriers and down through "end" barriers. This removes the reshapes from the main region of the program.


This PR will create MLIR similar to

```
%0 = iree_tensor_ext.compute_barrier.start ...
%1 = linalg.transpose %0
%2 = iree_tensor_ext.compute_barrier.end %1
// conv op...
```

I'm not sure what this means. The barriers are being used around a single `linalg.transpose` operation in the middle of the program, rather than demarcating the boundaries of a compute region. Clearly, we can't fuse the transpose with anything. But, should we be moving reshapes out of the transpose "region"? I think this use case shows it would be useful to give this op a clearer definition so that each part of the pipeline can treat it similarly.

For BOO, we know that the main compute of the program is just a single conv. We could do this:

```
%0 = linalg.transpose ...
%1 = iree_tensor_ext.compute_barrier.start %0
// conv op...
```

yzhang93: Thanks for the clarification. So if I understand correctly, it seems the intended behavior should be similar to `InsertTensorBarriersPass`, with the difference that `linalg.transpose` should live outside the `iree_tensor_ext.compute_barrier` region. Would this create any conflicts or issues when `InsertTensorBarriersPass` later runs in the pipeline?

MaheshRavishankar: So I was thinking about this, actually adding multiple barriers actually does not do anything than limit fusion, which is what we are trying to do. The `InsertBarriers` pass adds a barrier at the start and the end for managing the default heuristics, but add other barrier start/end will have no downside apart from blocking fusion of this op with producer/consumer.

@IanWood1  regarding your pass to fold reshapes across barriers, a question did come to mind as to how do we control what we move across the barrier. Reshapes are special. They eventually just get absorbed. 

I think we can start with this just adding a barrier_start after the transpose op and revisit as we learn more.

compiler/src/iree/compiler/Preprocessing/Common/ConvertConvFilterToChannelsLast.cpp
```diff
@@ -212,15 +213,28 @@ struct ConvertGenericChwfToFhwc : public OpRewritePattern<linalg::GenericOp> {
...
+    transposedFilter.replaceUsesWithIf(
```
MaheshRavishankar: Nit here : I think we want to use the rewriter to the replacement. Without that sometimes the rewriter mis-tracks things and goes off the rails that are hard to reason about.

compiler/src/iree/compiler/Preprocessing/Common/ConvertConvFilterToChannelsLast.cpp
```diff
@@ -244,6 +250,17 @@ struct ConvertGenericChwfToFhwc : public OpRewritePattern<linalg::GenericOp> {
...
+    // Insert compute_barrier.end after the compute op and replace its uses.
+    Value genericVal = genericOp.getResult(0);
+    auto barrierEndOp =
+        IREE::TensorExt::ComputeBarrierEndOp::create(rewriter, loc, genericVal);
+    rewriter.replaceUsesWithIf(genericVal, barrierEndOp.getResult(),
+                               [&](OpOperand &use) {
+                                 return use.getOwner() != genericOp &&
+                                        use.getOwner() != barrierEndOp &&
+                                        !isa<tensor::DimOp>(use.getOwner());
+                               });
```
IanWood1: I don't think we need the ComputeBarrierEndOp here right? We aren't trying to block it from fusing with any consumers.

IanWood1 (approved):
Looks good. Maybe the ComputeBarrierEndOp isn't needed?

compiler/src/iree/compiler/Preprocessing/Common/ConvertConvFilterToChannelsLast.cpp
yzhang93: Yes, the `ComputeBarrierEndOp` is not needed in this case. I'll remove it if `barrier.start` and `barrier.end` do not have to be in pair.

MaheshRavishankar: Yeah, please do remove.

PR #22682, opened 2025-11-18
Add support for debug prefix maps in IREE's CMake build

krzysz00:
This support roughtly follows LLVM's support for the same feature, which allows multiple builds of the same code to share ccache entries by remapping absolute paths in debug info and `__FILE__` to relative ones.

The current build already dose this remapping for `__FILE__`, but now we have the option to be more expansive and even more path-independent. We also avoid including the source directory name so that multiple worktrees can share cache entries.


docs/website/docs/developers/building/cmake-with-ccache.md
```diff
@@ -69,6 +69,20 @@ Notes:
   based on wrapping the compiler in a script that prepends `ccache`. See this
   [article](https://crascit.com/2016/04/09/using-ccache-with-cmake/).
 
+## Using relative paths in your build
+
+If you will build IREE from multiple directories in similar locations, such as if
+you are using git worktries, you should set
+`IREE_USE_RELATIVE_PATHS_IN_FILES=ON`. This, along with `ccache`'s
+`base_dir` option, will caues the compiler to rewrite absolute paths in
```
benvanik:
```suggestion
`base_dir` option, will cause the compiler to rewrite absolute paths in
```

benvanik (approved):

krzysz00:
Closing as not quite what was wanted and I think I can do better than this

Issue #21185, opened 2025-06-25
[Encoding] Missing interface implementations for PackedStorageAttr


hanhanW:
Today, we have few interfaces for describing an encoding type attribute. The packed_storage attribute was introduced before we have the abstractions. Thus, there are special logics spread in the codebase. I think we can implement the interface methods and use the interface methods to plumb through the encoding. E.g., we may want to use `convertType` or something similar to achieve the same thing.

https://github.com/iree-org/iree/blob/d267eb17fdfc421a01d1d21134f093941bb1e311/compiler/src/iree/compiler/Dialect/Stream/Conversion/HALToStream/Patterns.cpp#L104-L109

Encoding Doc:

https://github.com/iree-org/iree/blob/d267eb17fdfc421a01d1d21134f093941bb1e311/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingBase.td#L22-L73

sommerlukas:
Summary of recent discussion of this issue and https://github.com/iree-org/iree/issues/21184 on Discord: 

**End goal:** IREE should be able to support both encodings with correct behavior; then we can evaluate and do the flip for i1 cases.

**Current status:** for plain tensors, all the subtypes are packed by default except for `i1` types. We need to align the behavior or have better documentation. 

**Next steps:**
- Add the `SerializableAttr` interface to the existing `packed_storage` attribute by implementing `calculateStorageSizeInBytes` and `isSerialized` methods for the encoding.
- Drop all the hardcoded logic for the `packed_storage` encoding and use the encoding interfaces instead. The `packed_storage` attribute should be handled by the standard logic [here](https://github.com/iree-org/iree/blob/f5fc7fd07bdc1c1a88a354c0e315db8abca69a16/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L93-L97) after implementing the interface.
- Adapt [`attention_i1_mask.mlir`](https://github.com/iree-org/iree/blob/f5fc7fd07bdc1c1a88a354c0e315db8abca69a16/tests/e2e/linalg_ext_ops/attention_i1_mask.mlir) to start with the encodings.

[...]

PR #22688, opened 2025-11-18
[Encoding] Add SerializableAttr interface to packed_storage

sommerlukas:
Implement the `SerializableAttr` interface for the `packed_storage` encoding attribute and remove special case treatment to calculate storage size in favor of interface usage.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
@@ -7,6 +7,7 @@
 #include "iree/compiler/Dialect/Encoding/IR/EncodingTypes.h"
 
 #include "iree/compiler/Dialect/Encoding/IR/EncodingDialect.h"
+#include "iree/compiler/Dialect/Util/IR/UtilTypes.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/Debug.h"
@@ -91,6 +92,68 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
     res = arith::MaxUIOp::create(builder, loc, res, requestedSize);
   }
   return res;
+}
+
+//===---------------------------------------------------------------------===//
+// iree_encoding.packed_storage
+//===---------------------------------------------------------------------===//
+
+bool PackedStorageAttr::isSerialized() const { return true; }
+
+static Type getExtendedType(Type ty) {
+  auto intTy = dyn_cast<IntegerType>(ty);
+  if (!intTy) {
+    return ty;
+  }
+  unsigned bitWidth = intTy.getWidth();
+  if (bitWidth < 8 || llvm::isPowerOf2_32(bitWidth)) {
+    return ty;
+  }
+  unsigned alignedBitWidth = IREE::Util::getRoundedElementByteWidth(intTy) * 8;
+  if (alignedBitWidth == bitWidth) {
+    return ty;
+  }
+  // For integer types that are bigger then 8 bits and are not byte-aligned, we
+  // need to extend them.
+  return IntegerType::get(ty.getContext(), alignedBitWidth,
+                          intTy.getSignedness());
+}
+
+Value PackedStorageAttr::calculateStorageSizeInBytes(
+    Location loc, OpBuilder &builder, RankedTensorType type,
+    ValueRange dynamicDims) const {
+  Type alignedElementType = getExtendedType(type.getElementType());
+  int64_t staticCount =
+      IREE::Util::getRoundedElementByteWidth(alignedElementType);
+
+  // Calculate static dimensions if there are any
+  for (unsigned i = 0; i < type.getRank(); ++i) {
+    if (!type.isDynamicDim(i)) {
+      staticCount *= type.getDimSize(i);
+    }
+  }
+
+  // Emit computation of dynamic dimensions.
+  auto value =
+      arith::ConstantIndexOp::create(builder, loc, staticCount).getResult();
+  for (auto dim : dynamicDims) {
+    value = builder.createOrFold<arith::MulIOp>(loc, value, dim);
+  }
+
+  // For sub-byte element types we need to divide by the number of elements that
+  // can be packed into one byte.
+  unsigned elementBits = IREE::Util::getTypeBitWidth(alignedElementType);
+  if (elementBits < 8) {
+    if (!llvm::isPowerOf2_32(elementBits)) {
+      // Sub-byte types that are not a power of two are currently not supported.
+      return nullptr;
+    }
+    unsigned elementsPerByte = 8 / elementBits;
+    auto divisor =
+        arith::ConstantIndexOp::create(builder, loc, elementsPerByte);
+    value = builder.createOrFold<arith::CeilDivUIOp>(loc, value, divisor);
+  }
+  return value;
 }
 
 //===---------------------------------------------------------------------===//
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
@@ -137,7 +137,7 @@ Value PackedStorageAttr::calculateStorageSizeInBytes(
   assert(llvm::isPowerOf2_32(elementBits) &&
          "packed_storage only allowed for power-of-two types");
 
-  // Calculate static dimensions if there are any
+  // Calculate static dimensions if there are any.
   int64_t staticCount = 1;
   for (unsigned i = 0; i < type.getRank(); ++i) {
     if (!type.isDynamicDim(i)) {
@@ -148,7 +148,7 @@ Value PackedStorageAttr::calculateStorageSizeInBytes(
   // Emit computation of dynamic dimensions.
   auto value =
       arith::ConstantIndexOp::create(builder, loc, staticCount).getResult();
-  for (auto dim : dynamicDims) {
+  for (Value dim : dynamicDims) {
     value = builder.createOrFold<arith::MulIOp>(loc, value, dim);
   }
 
@@ -165,15 +165,15 @@ LogicalResult PackedStorageAttr::verifyEncoding(
     llvm::function_ref<mlir::InFlightDiagnostic()> emitError) const {
   unsigned elementBitWidth = getTypeBitWidth(elementType);
   if (elementBitWidth > 7) {
-    return emitError() << "Bit-width of the element type is " << elementBitWidth
+    return emitError() << "bit-width of the element type is " << elementBitWidth
                        << " but packed_storage is currently only supported for "
                        << "sub-byte types";
   }
   if (!llvm::isPowerOf2_32(elementBitWidth)) {
     return emitError()
-           << "Bit-width of the element type is " << elementBitWidth
+           << "bit-width of the element type is " << elementBitWidth
            << " but packed_storage currently only supports types with "
-           << "power-of-two bitwidth.";
+           << "power-of-two bitwidth";
   }
   return success();
 }
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
@@ -149,7 +149,8 @@ Value PackedStorageAttr::calculateStorageSizeInBytes(
   auto value =
       arith::ConstantIndexOp::create(builder, loc, staticCount).getResult();
   for (Value dim : dynamicDims) {
-    value = builder.createOrFold<arith::MulIOp>(loc, value, dim);
+    value = builder.createOrFold<arith::MulIOp>(
+        loc, value, dim, arith::IntegerOverflowFlags::nsw);
   }
 
   // For sub-byte element types we need to divide by the number of elements that
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
@@ -129,32 +129,16 @@ static int32_t getRoundedElementByteWidth(Type type) {
   return llvm::PowerOf2Ceil(byteAligned);
 }
 
-static Type getExtendedType(Type ty) {
-  auto intTy = dyn_cast<IntegerType>(ty);
-  if (!intTy) {
-    return ty;
-  }
-  unsigned bitWidth = intTy.getWidth();
-  if (bitWidth < 8 || llvm::isPowerOf2_32(bitWidth)) {
-    return ty;
-  }
-  unsigned alignedBitWidth = getRoundedElementByteWidth(intTy) * 8;
-  if (alignedBitWidth == bitWidth) {
-    return ty;
-  }
-  // For integer types that are bigger then 8 bits and are not byte-aligned, we
-  // need to extend them.
-  return IntegerType::get(ty.getContext(), alignedBitWidth,
-                          intTy.getSignedness());
-}
-
 Value PackedStorageAttr::calculateStorageSizeInBytes(
     Location loc, OpBuilder &builder, RankedTensorType type,
     ValueRange dynamicDims) const {
-  Type alignedElementType = getExtendedType(type.getElementType());
-  int64_t staticCount = getRoundedElementByteWidth(alignedElementType);
+  unsigned elementBits = getTypeBitWidth(type.getElementType());
+  assert(elementBits <= 8 && "packed_storage only allowed for sub-byte types");
+  assert(llvm::isPowerOf2_32(elementBits) &&
+         "packed_storage only allowed for power-of-two types");
 
   // Calculate static dimensions if there are any
+  int64_t staticCount = 1;
   for (unsigned i = 0; i < type.getRank(); ++i) {
     if (!type.isDynamicDim(i)) {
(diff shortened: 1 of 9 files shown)


sommerlukas:
This implements the first steps outlined in #21185.

The special case treatment of tensors with `packed_storage` encoding is mainly removed in `ElementPackingUtils` - `calculateStorageElementCountInBytes`. There are other cases where special case handling is happening, e.g., [here](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/Stream/Conversion/HALToStream/Patterns.cpp#L107-L109) where we need to decide what we want to do.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
hanhanW, line 1: We duplicated two methods from Util to Encoding dialect because we do not want the dependency (i.e., encoding -> util):
  - getTypeBitWidth
  - getRoundedElementByteWidth

Can you try to use the methods in this file?

https://github.com/iree-org/iree/commit/c618134f5d077a19b97add87f268e722a86b78df

hanhanW (commented):
Below can be deleted as well because now we implement the interface for the encoding.

https://github.com/iree-org/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/Stream/Transforms/EncodeTensors.cpp#L50-L54

hanhanW (commented):
I think below can also be deleted:

https://github.com/iree-org/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L99-L100


The below one is tricky because we don't have `calculateStorageElementOffsetInBytes` interface methods yet. IIRC, @benvanik asked for it in our early design discussion before. I ended up with not adding the interface method because I did not see a need for data-tiling use case. A follow-up would be adding such interface method to [SerializableAttr](https://github.com/iree-org/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td#L76) like what we've done for `calculateStorageSizeInBytes`. It will need a few implementations for other encodings, though. Can you help file an issue for this?

https://github.com/iree-org/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L142-L143

sommerlukas:
> I think below can also be deleted:
> 
> https://github.com/iree-org/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L99-L100

If I'm not mistaken, I've already removed the check for the `PackedStorageAttr` in this PR: https://github.com/iree-org/iree/blob/1e1ed7d523c0ac67ed79ae3673f3b33e06dd63a0/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L98-L101


> The below one is tricky because we don't have `calculateStorageElementOffsetInBytes` interface methods yet. IIRC, @benvanik asked for it in our early design discussion before. I ended up with not adding the interface method because I did not see a need for data-tiling use case. A follow-up would be adding such interface method to [SerializableAttr](https://github.com/iree-org/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td#L76) like what we've done for `calculateStorageSizeInBytes`. It will need a few implementations for other encodings, though. Can you help file an issue for this?
> 
> https://github.com/iree-org/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L142-L143

Yes, adding that method to the interface also crossed my mind yesterday, I've filed https://github.com/iree-org/iree/issues/22699 to implement that.

I was also wondering if we should add `needToPackSubByteElements` to the `SerializableAttr` interface, as it has a similar special casing for `PackedStorageAttr`:

https://github.com/iree-org/iree/blob/8867197f8f751edc82eaba951c84d078b2f77bc9/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L52-L60

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
sommerlukas: Sorry, I missed that we already had those utilities in the file. I've removed the new build dependency again.

hanhanW: no worries, it is not documented anywhere. The fork methods are only mentioned in the commit message.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
```diff
@@ -93,6 +93,97 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
...
+  // need to extend them.
+  return IntegerType::get(ty.getContext(), alignedBitWidth,
+                          intTy.getSignedness());
+}
+
+Value PackedStorageAttr::calculateStorageSizeInBytes(
+    Location loc, OpBuilder &builder, RankedTensorType type,
+    ValueRange dynamicDims) const {
+  Type alignedElementType = getExtendedType(type.getElementType());
+  int64_t staticCount = getRoundedElementByteWidth(alignedElementType);
```
hanhanW: I thought it can just be `product_of(shape) * elementType.bitwidth ceildiv 8 * 8`? Why do we need `getExtendedType`?

(We'll need more tests for non-i1 cases if we decide to go with the current approach.)

sommerlukas: This follows the logic from `legalizeStorageElementTypeImpl` here: https://github.com/iree-org/iree/blob/098828a403d018b7401e88812fe7501225632b8a/compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp#L62-L81

For types with bit-width greater than 8, `needToPackSubByteElementBitWidthImpl` returns `false` and `legalizeStorageElementTypeImpl` will return a new type that is extended to the next full byte. 

I tried to stay close to the existing logic, but we can of course deviate more. In the existing logic, for types with more than eight bits, every element would be extended to the next full byte and only then we calculate the storage based on the tensor shape.

This is visible in the test case with `i12` that I added: Without this logic, we would fit `3xi12` into 36 bit or 5 bytes, with the existing logic, we end up with 6 bytes, because every `i12` is extended to `i16`.

I think the rationale behind this is alignment: Without this logic, the second `i12` in our example would start in the middle of the second byte and extend into the third byte. With the logic, types are either aligned or fit entirely into one byte in all cases (because we don't allow non-power-of-two, so all types smaller than one byte always fit into the byte they start in).

hanhanW: Thanks for digging into this. I think they may be legacy code. Let's restrict the semantic of the encoding for now; we can invent `aligned_packed_storage` or extend the encoding semantic in the future when there is a need.

Given that the work was started to support i1, let's not invent things without concrete use cases. So I'm +1 on dropping those logic and make things simple for now. It will be a basic encoding for IREE like `iree_encoding.identity`.

sommerlukas: Ok, so do we want to apply the following type restrictions on `packed_storage` then?
- No element types bigger than one byte
- No element types that are not power-of-two (already forbidden ATM).

Or do you still want to support types bigger than one byte, but with potentially unaligned access?

hanhanW: I want to support types bigger than one byte, but with potentially unaligned access.

Reason: I think encoding is just a hint about how data is stored in memory. The way I read this encoding is that it packs data back-to-back for all the cases. We can't guess what others want for alignment, and they may differ. It is easier to start with treating it as a naive encoding.

RE future support: I think we may have `group_packed_storage` which is not serialized until the resolver gives it the value that controls how to pack the elements. This is definitely a different topic, and it is how my future picture about such support. (It may be a little different when I see a concrete problem.)

sommerlukas: The current situation in IREE is that `i1` is packed if indicated by the `packed_storage` encoding or the experimental option, otherwise it's extended to byte alignment. Other sub-byte types (`i2`, `i4`) are packed by default, sub-byte types that are not power-of-two are not supported. All types bigger than one byte are always extended to the next full byte and therefore aligned.

I've looked into how we could support packed, unaligned storage for types bigger than one byte (super-byte) with `packed_storage` and I'm starting to think that we should not support that at the moment, or at least not in this PR, for the following reasons:
* It is effectively a new feature not available in IREE at the moment, and we should probably have a use case to add it.
* Corresponding to `needToPackSubByteElements`, we would also need a `needToPackSuperByteElements` (or put both in one method with a more generic name) and plumb that through all the places that currently use `needToPackSubByteElements`.
* We would need to extend `calculateStorageElementOffsetInBytes`. For the case of super-byte types that are not a multiple of 8 (e.g., `i12`, `i14`), we would need to emit the following computation to compute the element offset at runtime: `floorDiv((elementLinearIndex * bitWidth), 8)`, where the `floorDiv` by 8 could be implemented as shift. As the `bitWidth` divided by 8 can be non-integer values, this needs to be computed at runtime and not at compile time.
* Type legalization in the `TypePropagationPass` is potentially also affected and needs updating.
* We would need to treat the experimental option `iree-experimental-packed-i1-storage` different from the `packed_storage` encoding. While the `packed_storage` encoding would sign up for packing of all types, I think based on the name of the option, users would reasonably expect super-byte types to keep their default, unpacked, aligned encoding when using the option.
* We'd need additional testing for packing of super-byte types across the different layers.

IMHO, we should only implement this feature if we have a use case. For now, I'd restrict `packed_storage` to only be applied to sub-byte element types or, as a middle ground, types with bit-width <8 bit or multiple of 8 bits (e.g., allow `i24`).

If we decide to implement the feature, we should probably file an issue and implement it in a separate PR(s), given the number of changes necessary.

hanhanW: Thanks for digging into the details. TLDR is that let's scope the work for i1 types and just pack them back-to-back.

> Corresponding to needToPackSubByteElements, we would also need a needToPackSuperByteElements (or put both in one method with a more generic name) and plumb that through all the places that currently use needToPackSubByteElements.

It may be more than you thought. Packing super byte may need a different encoding that allows encoding resolver to propagate their preference. It really depends on how an encoding is introduced. The information may have to be hardcoded if they are from frontend because only they can understand what the format is.

> We would need to extend calculateStorageElementOffsetInBytes. For the case of super-byte types that are not a multiple of 8 (e.g., i12, i14), we would need to emit the following computation to compute the element offset at runtime: floorDiv((elementLinearIndex * bitWidth), 8), where the floorDiv by 8 could be implemented as shift. As the bitWidth divided by 8 can be non-integer values, this needs to be computed at runtime and not at compile time.
> Type legalization in the TypePropagationPass is potentially also affected and needs updating.

We do have narrow type emulation, although there are some limits. IIRC, the upstream implementation is able to emulate unaligned cases. We'll need to avoid it in practice, though. Plumbing is also very involved, though.

sommerlukas: Ok, thanks for the feedback!

I've restricted `packed_storage` to sub-byte, power-of-two types now, enforced by the tensor encoding verification. Based on that, I've simplified the logic to compute the storage size.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
```diff
@@ -93,6 +93,91 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
...
+
+Value PackedStorageAttr::calculateStorageSizeInBytes(
+    Location loc, OpBuilder &builder, RankedTensorType type,
+    ValueRange dynamicDims) const {
+  unsigned elementBits = getTypeBitWidth(type.getElementType());
+  assert(elementBits <= 8 && "packed_storage only allowed for sub-byte types");
+  assert(llvm::isPowerOf2_32(elementBits) &&
+         "packed_storage only allowed for power-of-two types");
+
+  // Calculate static dimensions if there are any
```
hanhanW:
```suggestion
  // Calculate static dimensions if there are any.
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
```diff
@@ -93,6 +93,91 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
...
+}
+
+LogicalResult PackedStorageAttr::verifyEncoding(
+    llvm::ArrayRef<int64_t> shape, mlir::Type elementType,
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError) const {
+  unsigned elementBitWidth = getTypeBitWidth(elementType);
+  if (elementBitWidth > 7) {
+    return emitError() << "Bit-width of the element type is " << elementBitWidth
+                       << " but packed_storage is currently only supported for "
+                       << "sub-byte types";
```
hanhanW, line 171: > Also, to match error message styles commonly produced by other tools, start the first sentence with a lowercase letter, and finish the last sentence without a period, if it would end in one otherwise.

https://llvm.org/docs/CodingStandards.html#error-and-warning-messages

```suggestion
    return emitError() << "bit-width of the element type is " << elementBitWidth
                       << " but packed_storage is currently only supported for "
                       << "sub-byte types";
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
```diff
@@ -93,6 +93,91 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
...
+  }
+  if (!llvm::isPowerOf2_32(elementBitWidth)) {
+    return emitError()
+           << "Bit-width of the element type is " << elementBitWidth
+           << " but packed_storage currently only supports types with "
+           << "power-of-two bitwidth.";
```
hanhanW: ditto

```suggestion
    return emitError()
           << "bit-width of the element type is " << elementBitWidth
           << " but packed_storage currently only supports types with "
           << "power-of-two bitwidth.";
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
```diff
@@ -93,6 +93,91 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
...
+      staticCount *= type.getDimSize(i);
+    }
+  }
+
+  // Emit computation of dynamic dimensions.
+  auto value =
+      arith::ConstantIndexOp::create(builder, loc, staticCount).getResult();
+  for (auto dim : dynamicDims) {
+    value = builder.createOrFold<arith::MulIOp>(loc, value, dim);
+  }
```
hanhanW: style nit: spell out the type when it is not obvious. (I also forgot it sometimes)

https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

The other suggestion is using `arith::IntegerOverflowFlags::nsw`. I learned that it enables folding for such operation. Some old implementation does not do this because it didn't exist when they were implemented.

We already do this for padding attribute, fyi:

https://github.com/iree-org/iree/blob/5f5da9dad9dcc8045d77f6a5436b5981bb780e70/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp#L457C46-L457C78

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/encode_host_tensors_encoding.mlir
hanhanW, line 1: I probably missed the test coverage before, can we add a test case for dynamic case?

hanhanW (commented):
Just some nits, nice!

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/encode_host_tensors_encoding.mlir
sommerlukas: I have one test for the dynamic case in the next PR [here](https://github.com/iree-org/iree/pull/22722/files#diff-af498e8abf54fb8b8c378791936cd53d9e69923cf3f694071f22ce503d89cd9d). Are you OK to defer to that PR or should I bring the test over to this PR?

hanhanW: Why not copy it to this PR instead? I prefer having the test in this PR, so the state of the codebase is cleaner. We can do it in a separate if it is tricky, but I'm not seeing any issue. Let me know if it is tricky!

sommerlukas: You're right, makes sense to include it hre. I've moved it over to this PR.

hanhanW (approved):
thanks, LGTM!

PR #21903, opened 2025-09-09
[DispatchCreation]: Add FormSplitReductionDispatchesPass support for ArgCompare op

bangtianliu:
This PR adds the `PartialReductionOpInterface` to ArgCompareOp, enabling split-k reduction for this op via the `FormSplitReductionDispatchesPass`. This pass leverages upstream’s `scf::tileUsingSCF` with the `ReductionTilingStrategy::PartialReductionOuterParallel` strategy.

For reference, the original idea was proposed by Mahesh and can be found here: https://hackmd.io/-iRV4xydSIKzsup9fJ5xbg.

PR #22694, opened 2025-11-18
[DispatchCreation] Add FoldExtractSliceOfBroadcast Pattern

bangtianliu:

Debugging the arg_compare op compilation issue in llama 8b fp16 quality tests from PR https://github.com/iree-org/iree/pull/22379.

CI Error: https://github.com/iree-org/iree/actions/runs/19219575275/job/54935278831?pr=22394#step:7:122

**Context**: This pattern is specific to split reduction for argcompare ops. The issue arises from using `linalg.broadcast` 
https://github.com/iree-org/iree/pull/21903#discussion_r2334813511 to create init values and indices for splitting argcompare along the reduction dimension (since argcompare has no fixed neutral element—the identity depends on the comparator region).  When `tileAndFuseProducerOfSlice` moves the `linalg.broadcast` inside `scf.forall`  during split reduction formation, the broadcast writes to aliased shared memory across parallel workgroups, causing a race condition. 

This PR adds `FoldExtractSliceOfBroadcast` pattern to `FormSplitReductionDispatches` pass that folds `extract_slice(broadcast)` into the `linalg.broadcast` input when the `tensor.extract_slice` is rank-reducing along broadcasted dimensions.

This fold is critical for correctness in parallel contexts. When a broadcast's outs operand comes from extracting a slice of a shared_outs tensor in `scf.forall`, the broadcast operation writes to that extracted slice, which 
aliases back to the shared tensor. Since the extracted slice can span parallel dimensions (e.g., `[4, 1, 1]` where dim 0 is the batch dimension), multiple parallel workgroups would all write to the same aliased memory location, **causing a race condition**.

By folding to the broadcast input (which doesn't alias the shared tensor), subsequent passes can properly tile and distribute the initialization per workgroup, avoiding the race.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
@@ -27,23 +27,28 @@ namespace mlir::iree_compiler::DispatchCreation {
 
 namespace {
 
-/// Pattern to fold extract_slice(broadcast) into the broadcast input when
-/// the extract_slice is rank-reducing and only extracts along the broadcasted
-/// dimensions (with size 1), leaving all non-broadcasted dimensions intact.
-/// This is valid because the broadcast only duplicates data along the
-/// broadcasted dimension, and extracting a single slice from that dimension
-/// gives back the original input.
+/// WAR (Workaround): Pattern to fold extract_slice(broadcast) into the
+/// broadcast input when the extract_slice is rank-reducing and only extracts
+/// along the broadcasted dimensions (with size 1), leaving all non-broadcasted
+/// dimensions intact. This is valid because the broadcast only duplicates data
+/// along the broadcasted dimension, and extracting a single slice from that
+/// dimension gives back the original input.
 ///
-/// This fold is critical for correctness in parallel contexts: when the
-/// broadcast's outs operand comes from extracting a slice of a shared_outs
-/// tensor in scf.forall, the broadcast operation writes to that extracted
-/// slice, which aliases back to the shared tensor. Since the extracted slice
-/// spans parallel dimensions (e.g., [4, 1, 1] where dim 0 is the batch
-/// dimension), multiple parallel workgroups would all write to the same
-/// aliased memory location, causing a race condition. By folding to the
-/// broadcast input (which doesn't alias the shared tensor), subsequent passes
-/// can properly tile and distribute the initialization per workgroup, avoiding
-/// the race.
+/// TODO(Bangtian): Investigate and fix the root cause of the race condition in
+/// the lower levels of the stack to eliminate the need for this fold pattern
+/// workaround.
+///
+/// This fold masks an underlying race condition in the lower levels of the
+/// stack. When the broadcast's outs operand comes from extracting
+/// a slice of a shared_outs tensor in scf.forall, the broadcast operation
+/// writes to that extracted slice, which aliases back to the shared tensor.
+/// Since the extracted slice spans parallel dimensions (e.g., [4, 1, 1] where
+/// dim 0 is the batch dimension), multiple parallel workgroups would all write
+/// to the same aliased memory location, creating a race condition. By folding
+/// to the broadcast input (which doesn't alias the shared tensor), this pattern
+/// works around the race, allowing subsequent passes to properly tile and
+/// distribute the initialization per workgroup. However, the root cause of the
+/// race condition exists deeper in the stack and is not addressed by this fold.
 ///
 /// Example:
 ///   %broadcast = linalg.broadcast ins(%in : tensor<4x1xf16>)
@@ -63,11 +68,13 @@ struct FoldExtractSliceOfBroadcast final
     auto broadcastOp =
         extractOp.getSource().getDefiningOp<linalg::BroadcastOp>();
     if (!broadcastOp) {
-      return failure();
+      return rewriter.notifyMatchFailure(
+          extractOp, "source is not a linalg.broadcast operation");
     }
 
     if (!extractOp.hasUnitStride()) {
-      return failure();
+      return rewriter.notifyMatchFailure(extractOp,
+                                         "extract_slice has non-unit stride");
     }
 
     auto inputType =
@@ -77,12 +84,14 @@ struct FoldExtractSliceOfBroadcast final
     auto extractResultType =
         dyn_cast<RankedTensorType>(extractOp.getResult().getType());
     if (!inputType || !broadcastOutputType || !extractResultType) {
-      return failure();
+      return rewriter.notifyMatchFailure(
+          extractOp, "operand or result types are not RankedTensorType");
     }
 
     // Extract result type must match broadcast input type.
     if (inputType != extractResultType) {
-      return failure();
+      return rewriter.notifyMatchFailure(
+          extractOp, "extract result type does not match broadcast input type");
     }
 
     // Verify that we're extracting from offset [0, 0, ..., 0] with the same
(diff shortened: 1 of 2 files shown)


IanWood1 (commented):
I'm trying to understand this part of the PR description:

> Since the extracted slice can span parallel dimensions (e.g., [4, 1, 1] where dim 0 is the batch dimension), multiple parallel workgroups would all write to the same aliased memory location, causing a race condition.

Here's the IR after this pass, but without the pattern you added: https://gist.github.com/IanWood1/d2cd56b7981864c6ad5a6caf372c4dd3. From what I can tell, it seems like each `linalg.broadcast` is operating on a unique slice of the shared outs.


Also, is it possible to get a `linalg.broadcast` that isn't foldable?

bangtianliu:
> I'm trying to understand this part of the PR description:
> 
> > Since the extracted slice can span parallel dimensions (e.g., [4, 1, 1] where dim 0 is the batch dimension), multiple parallel workgroups would all write to the same aliased memory location, causing a race condition.

Here is the link to gist with out the pattern I added after TileAndDistributeToWorkgroupsUsingForallOpPass  : https://gist.github.com/bangtianliu/3501089ff3e43ba69fae0e515bbd2817

```
// Line 21: Outer loop with split_reduction_mapping
%11:2 = scf.forall (%arg0) = (0) to (128256) step (1336) 
        shared_outs(%arg1 = %6, %arg2 = %5) -> ... {
  ...
  // Lines 29-30: The problematic sequence
  %extracted_slice = tensor.extract_slice %arg1[0, 0, %17] [4, 1, 1] [1, 1, 1]
  %broadcasted = linalg.broadcast ins(%9 : tensor<4x1xf16>) 
                                  outs(%extracted_slice : tensor<4x1x1xf16>)
  ...
  // Line 36: Nested loop with workgroup mapping on BATCH dimension
  %20:2 = scf.forall (%arg3, %arg4) = (0, 0) to (4, 1) step (1, 128)
          shared_outs(%arg5 = %extracted_slice_0, %arg6 = %extracted_slice_3)
  ...
} {mapping = [#iree_linalg_ext.split_reduction_mapping<0>]}
```
The nested scf.forall at line 51 has mapping = [#iree_codegen.workgroup_mapping<y>, ...] which distributes the batch dimension across multiple parallel workgroups. When these workgroups (4) execute in parallel within the same outer iteration, they all read/write to memory that was just written by the broadcast.

> Also, is it possible to get a `linalg.broadcast` that isn't foldable?

My answer is highly chance no. But if it happens, then it requires handling it in subsequent distribution pass. Below is detailed explanation for my answer.

The pattern arises specifically from split reduction support for the arg_compare op. Here's why it's always foldable in this context:
During split reduction transformation, the initialization values are broadcast to match the number of partitions (creating the extra split dimension). When this broadcast is fused inside the scf.forall loop, it operates on a unit-sized slice in the split dimension. The subsequent rank-reducing extract_slice removes that unit dimension to satisfy arg_compare's input requirements, which expects rank-2 tensors (in above example shared in your gist). Since this sequence is mechanically generated by the split reduction transformation and always follows the same pattern (broadcast along a unit dimension → rank-reducing extract), it should always be foldable in the context of split reduction for arg_compare.

MaheshRavishankar:
Ok, the main question I am trying to answer is that do we need the `linalg.broadcast` to be fused into the `scf.forall` ? The way I undertand it, the first dispatch needs an initial value for the result buffer. I think we sshouldnt be folding the `linalg.broadcast` into the scf.forall at all? This might be something I should disable in tile and fuse directly I think.

bangtianliu:
> Ok, the main question I am trying to answer is that do we need the `linalg.broadcast` to be fused into the `scf.forall` ? The way I undertand it, the first dispatch needs an initial value for the result buffer. I think we sshouldnt be folding the `linalg.broadcast` into the scf.forall at all? This might be something I should disable in tile and fuse directly I think.

No, we donot need. 

You're correct to question this. The broadcast should not be fused into the scf.forall in this context.
The root issue is that the tiling interface method tileAndFuseProducerOfSlice automatically fuses producer operations (including the linalg.broadcast) into the scf.forall when it sees that the broadcast produces a slice being consumed inside the loop.

compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
```diff
@@ -27,6 +27,116 @@ namespace mlir::iree_compiler::DispatchCreation {
...
 namespace {
 
+/// Pattern to fold extract_slice(broadcast) into the broadcast input when
+/// the extract_slice is rank-reducing and only extracts along the broadcasted
+/// dimensions (with size 1), leaving all non-broadcasted dimensions intact.
+/// This is valid because the broadcast only duplicates data along the
+/// broadcasted dimension, and extracting a single slice from that dimension
+/// gives back the original input.
+///
+/// This fold is critical for correctness in parallel contexts: when the
```
MaheshRavishankar: Please update this based on our discussion that this is really hiding the race condition and there is a race condition somewhere in the lower levels of the stack. The tiling itself seems ok.

compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
```diff
@@ -27,6 +27,116 @@ namespace mlir::iree_compiler::DispatchCreation {
...
+struct FoldExtractSliceOfBroadcast final
+    : OpRewritePattern<tensor::ExtractSliceOp> {
+  using Base::Base;
+
+  LogicalResult matchAndRewrite(tensor::ExtractSliceOp extractOp,
+                                PatternRewriter &rewriter) const override {
+    auto broadcastOp =
+        extractOp.getSource().getDefiningOp<linalg::BroadcastOp>();
+    if (!broadcastOp) {
+      return failure();
```
MaheshRavishankar: Nit: It is useful to add a `rewriter.notifyMatchFailure()` here with a line that says why the match failed. That will be printed out in the debug log.

compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
```diff
@@ -27,6 +27,116 @@ namespace mlir::iree_compiler::DispatchCreation {
...
+
+    // Sizes should match input dimensions (accounting for broadcast dims).
+    ArrayRef<int64_t> broadcastDims = broadcastOp.getDimensions();
+    int64_t broadcastRank = broadcastOutputType.getRank();
+
+    // Verify that for broadcast dimensions, the size is 1.
+    if (llvm::any_of(broadcastDims, [&](int64_t broadcastDim) {
+          std::optional<int64_t> size =
+              getConstantIntValue(sizes[broadcastDim]);
+          return !size || *size != 1;
```
MaheshRavishankar: You can just use `isOneInteger` here.

compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
```diff
@@ -27,6 +27,116 @@ namespace mlir::iree_compiler::DispatchCreation {
...
+    // broadcasted (i.e., dimensions that existed in the original input).
+    auto nonBroadcastDims = llvm::to_vector(llvm::make_filter_range(
+        llvm::seq<int64_t>(0, broadcastRank),
+        [&](int64_t i) { return !llvm::is_contained(broadcastDims, i); }));
+
+    // Verify that for non-broadcast dimensions, sizes match input shape.
+    if (llvm::any_of(llvm::enumerate(nonBroadcastDims), [&](auto pair) {
+          auto [idx, inputDim] = pair;
+          std::optional<int64_t> extractSize =
+              getConstantIntValue(sizes[inputDim]);
```
MaheshRavishankar: You can use `isConstantIntValue` here.

compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp
```diff
@@ -27,6 +27,116 @@ namespace mlir::iree_compiler::DispatchCreation {
...
+    if (inputType != extractResultType) {
+      return failure();
+    }
+
+    // Verify that we're extracting from offset [0, 0, ..., 0] with the same
+    // shape as the input (essentially undoing the broadcast).
+    SmallVector<OpFoldResult> offsets = extractOp.getMixedOffsets();
+    SmallVector<OpFoldResult> sizes = extractOp.getMixedSizes();
+    if (llvm::any_of(offsets, [](OpFoldResult offset) {
+          std::optional<int64_t> constOffset = getConstantIntValue(offset);
```
MaheshRavishankar: You can use `isConstantIntValue` here.

MaheshRavishankar (approved):
Please make explicit that this is a WAR and that this is masking an underlying race condition.

PR #22702, opened 2025-11-19
[HAL] fix IREE_HAL_MAX_QUEUES to be number of bits in queue affinity type

schuermans-roofline:
The computation of the number of bits in the queue affinity type iree_hal_queue_affinity_t needs to use a multiplication by 8 to get from the number of bytes to the number of bits. Before, it incorrectly used a divison by 8. This resulted in IREE_HAL_MAX_QUEUES having an incorrect value of 1 instead of 64.


AWoloszyn (approved):
Thanks!!

PR #22703, opened 2025-11-19
[DispatchCreation] Move RemoveTensorBarriers to end of pipeline

IanWood1:
`RemoveTensorBarriersPass` was being run too early in the pipeline, which prevented the barriers from blocking fusion. Barriers should be removed at the very end of the pipeline.


IanWood1:
cc @rkayaith

MaheshRavishankar (approved):

Issue #22502, opened 2025-10-31
[Dispatch Creation] Grouped conv dgrad results in two dispatches


IanWood1:
This MLIR should only require a single dispatch https://gist.github.com/IanWood1/6476b80b0bc19309ef200a6b79d960ac

Compiled with:

```
iree-compile iree-compile-input.mlir   -o /dev/null --compile-to=dispatch-creation --iree-opt-level=O3
```

IREE @ 16e8d36a54bbca3c4fa046159ea341827d97753b

~Update:~
~I think I'm going to have to land the changes to "sink" reshapes through reduction ops (https://github.com/iree-org/iree/pull/22341). There were a few CI regressions that led to me closing the PR but I'll try to resolve them. I will try to land fixes for the different problems separately:~
~1. https://github.com/iree-org/iree/pull/22503~
~2. Try to fix llama regression where transpose now wants to fuse with the producer attention instead of the consumer reduction. I attempted this https://github.com/iree-org/iree/pull/22341/commits/69c439b674371ab08ee474f0bb993a0a6a22eddc but it had problems.~
~3. Check for more regressions...~
~4. Actually land the "sink" change.~

Update v2:
I moved the sinking issue to its own issue https://github.com/iree-org/iree/issues/22631. This should now be solved by adding the compute_barrier ops to prevent reshape movement.

IanWood1:
This only effects grouped convs because the torch-mlir lowering uses an expand_shape to expand out the group dimension. This ends up blocking the transpose from fusing.

a-sidorova:
It should be fixed by the PR https://github.com/iree-org/iree/pull/22904. The `aten.convolution_backward` will be directly lowered to `linalg` dialect. It will allow IREE to handle assigning reduction/parallel dims directly - no need to leave several transposes/expand/collapse in the separate dispatch regions.

```
iree-compile iree-compile-input.mlir --compile-to=dispatch-creation --iree-opt-level=O3
module @module {
  util.func public @main$async(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer_view, %arg3: !hal.fence, %arg4: !hal.fence) attributes {inlining_policy = #util.inline.never, iree.abi.model = "coarse-fences", iree.abi.stub} {
    %0 = hal.tensor.import wait(%arg3) => %arg1 : !hal.buffer_view -> tensor<16x64x32x256xf32>
[...]

PR #22341, opened 2025-10-17
[Dispatch Creation] Sink reshapes through reduction ops

IanWood1:
Testing for CI regressions

List of blockers:
- This causes some missed cases for horizontal fusion [`a379176` (#22341)](https://github.com/iree-org/iree/pull/22341/commits/a379176f49a2dfc6c933b285d0e2e925c143c377)
- Bug in dispatch creation [`b6ca77a` (#22341)](https://github.com/iree-org/iree/pull/22341/commits/b6ca77aef64e99f08702cdb233b3f8b18286e6a1)
- UNRESOLVED: A regression when a transpose gets fused as the consumer of a matmul for the test_torch tests suite on mi325: [Before PR](https://gist.github.com/IanWood1/922590ab01ad78a839f8e382ac10c2c2) and [After PR](https://gist.github.com/IanWood1/b9aaa1591a1e47e587e6309194bc6b09). Possibly related to the flow bitcasts.

ci-extra: test_torch

PR #22707, opened 2025-11-19
[Dispatch Creation] Add aggressive reshape movement flag

IanWood1:
Adds new compiler flag `--dispatch-creation-enable-aggressive-reshape-movement` to enable moving reshapes through reduction operations. The top-level flag controls the pass option `enableReshapeMovementAcrossReductions` in BubbleUpExpandShapes.

I have been working to enable this by default on https://github.com/iree-org/iree/pull/22341; there are a few issues/regressions to fix. For now, I'm adding this under a flag because this is needed in Fusilli https://github.com/iree-org/fusilli/issues/18 https://github.com/iree-org/iree/issues/22502

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
@@ -464,7 +464,7 @@ void BubbleUpExpandShapesPass::runOnOperation() {
           // If producer generic op is elementwise op, bubble up the expand
           // shape past this operation.
           // If bubbling across reduction ops is enabled, allow all generic ops.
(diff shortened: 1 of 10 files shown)


MaheshRavishankar (approved):
Ok! Do we need to options to the pass? Can we just make it "aggressive reshape movement" that just moves the expand shapes up and collapse shapes down "aggressively". I assume this is all transient, so this is fine for me as is as well. Just curious!

IanWood1:
The other option is used by the make single dispatch pipeline https://github.com/iree-org/iree/blob/22d6055c459d9c572f932954a698ad2ee259ca09/compiler/src/iree/compiler/Preprocessing/Passes.cpp#L155-L158

Changing this to be a single pass option may change functionality slightly there. But I don't think it would break anything. I'll try it.

Issue #22631, opened 2025-11-13
[Dispatch Creation] Enable moving reshapes through reduction ops


IanWood1:
It should be possible to remove the checks:
https://github.com/iree-org/iree/blob/f5fc7fd07bdc1c1a88a354c0e315db8abca69a16/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp#L467-L469
https://github.com/iree-org/iree/blob/f5fc7fd07bdc1c1a88a354c0e315db8abca69a16/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp#L481-L483 

But this causes a few performance regressions in models tracked on CI:
1. https://github.com/iree-org/iree/pull/22503
2. ~There is a llama regression where transpose ops now get fused with attention ops (producer) instead of the consumer reduction. This makes the reduction non-contiguous. I attempted this at https://github.com/iree-org/iree/pull/22341/commits/69c439b674371ab08ee474f0bb993a0a6a22eddc, but it had problems.~
3. https://github.com/iree-org/iree/pull/22708
4. https://github.com/iree-org/iree/pull/22711
5. A regression when a transpose gets fused as the consumer of a matmul for the test_torch tests suite on mi325: [Before PR](https://gist.github.com/IanWood1/922590ab01ad78a839f8e382ac10c2c2) and [After PR](https://gist.github.com/IanWood1/b9aaa1591a1e47e587e6309194bc6b09). Possibly related to the flow bitcasts.
6. Check for more regressions...

IanWood1:
Addig this under a flag https://github.com/iree-org/iree/pull/22707 until regressions are fully addressed

PR #22708, opened 2025-11-20
[Dispatch Creation] Don't fuse uses from above

IanWood1:
When fusing with a consumer operation with producers, the consumer may also use values from the producer in its body (e.g., `tensor.extract`). This change prevents fusing in those cases because we only support fusion through operands.

This was discovered because `moveFollowingOpIntoDispatchRegion` was generating illegal IR. When moving a consumer into the region, it wasn't updating the consumer's uses of the region if they were nested in the consumer's region (only operands). This led to illegal IR. This PR also fixes `moveFollowingOpIntoDispatchRegion` to update all uses by the target op.

Progress on https://github.com/iree-org/iree/issues/22631

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
+++ b/compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
@@ -543,20 +543,14 @@ moveFollowingOpIntoDispatchRegion(RewriterBase &rewriter, Operation *target,
   rewriter.setInsertionPoint(body.getTerminator());
   Operation *clonedTarget = rewriter.clone(*target);
 
-  // Replace any operands returned by the `regionOp` with the results yielded
-  // inside of the `regionOp`.
-  for (OpOperand &operand : clonedTarget->getOpOperands()) {
-    if (operand.get().getDefiningOp() != regionOp) {
-      continue;
-    }
-    auto returnOp =
-        cast<IREE::Flow::ReturnOp>(regionOp.getBody().front().getTerminator());
-    auto opResult = cast<OpResult>(operand.get());
-    Value yieldedValue = returnOp->getOperand(opResult.getResultNumber());
-    rewriter.modifyOpInPlace(clonedTarget, [&]() {
-      clonedTarget->setOperand(operand.getOperandNumber(), yieldedValue);
-    });
-  }
+  // Replace all of `clonedTarget` uses of `regionOp` with the values yielded
+  // from inside the region.
+  auto returnOp =
+      cast<IREE::Flow::ReturnOp>(regionOp.getBody().front().getTerminator());
+  rewriter.replaceOpUsesWithIf(
+      regionOp, returnOp.getOperands(), [&](OpOperand &operand) {
+        return clonedTarget->isAncestor(operand.getOwner());
+      });
 
   // Gather all uses of `target`.
   for (auto [index, result] : llvm::enumerate(target->getResults())) {
--- a/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
@@ -127,6 +127,19 @@ public:
   FailureOr<AffineMap> getRootParallelLoopToOpMap(Operation *op) const;
 
   bool isFusable(Operation *op) const {
+    // We only handle fusion across operation's operands. Don't fuse if the
+    // operation is using values in the fusion group in it's body.
+    bool hasUseFromAbove = false;
+    mlir::visitUsedValuesDefinedAbove(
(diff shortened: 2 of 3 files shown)


compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
```diff
@@ -543,20 +543,14 @@ moveFollowingOpIntoDispatchRegion(RewriterBase &rewriter, Operation *target,
...
-      clonedTarget->setOperand(operand.getOperandNumber(), yieldedValue);
-    });
-  }
+  // Replace all of `clonedTarget` uses of `regionOp` with the values yielded
+  // from inside the region.
+  auto returnOp =
+      cast<IREE::Flow::ReturnOp>(regionOp.getBody().front().getTerminator());
+  rewriter.replaceOpUsesWithIf(
+      regionOp, returnOp.getOperands(), [&](OpOperand &operand) {
+        return clonedTarget->isAncestor(operand.getOwner());
```
MaheshRavishankar, line 552: So just to confirm, `isAncestor` includes the operation as well.

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
```diff
@@ -1837,3 +1837,41 @@ util.func public @no_fusion_across_blocks(%arg0: tensor<3x2xf32>) -> tensor<f32>
...
+  %matmul = linalg.matmul ins(%arg0, %arg1 : tensor<?x?xf32>, tensor<?x?xf32>)
+      outs(%empty : tensor<?x?xf32>) -> tensor<?x?xf32>
+  %empty2 = tensor.empty(%d0, %d1) : tensor<?x?xf32>
+  %consumer = linalg.generic { indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = ["parallel", "parallel"]}
+      ins(%matmul : tensor<?x?xf32>)
+      outs(%empty2 : tensor<?x?xf32>) {
+  ^bb0(%in: f32, %out: f32):
+    %c0_idx = arith.constant 0 : index
+    %c1_idx = arith.constant 1 : index
+    %extracted = tensor.extract %matmul[%c0_idx, %c1_idx] : tensor<?x?xf32>
```
MaheshRavishankar, line 1859: Interesting example.

First, I dont know why we are lowering it this way. This seems awfully strange lowering. Are we sure we know whats happening with the input lowering of torch-mlir. I dont think we want to end up with this lowering.

It does so happen that the SSA violation is being used to prevent this fusion, but we should prevent this fusion "structurally" as well, i.e. when we decide if this is a fusable consumer, we should mark it as not fusable. due to the `tensor.extract` usage in the consumer. That would also avoid the issue and seems like the more correct handling of this in my view.

MaheshRavishankar: EDIT: I see that you have checked that exactly below. Nice!

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -127,6 +127,19 @@ class FusionGroup {
   FailureOr<AffineMap> getRootParallelLoopToOpMap(Operation *op) const;
 
   bool isFusable(Operation *op) const {
+    // We only handle fusion across operation's operands. Don't fuse if the
+    // operation is using values in the fusion group in it's body.
+    bool hasUseFromAbove = false;
+    mlir::visitUsedValuesDefinedAbove(
+        op->getRegions(), [&](OpOperand *operand) {
+          if (loopMaps.contains(operand->get().getDefiningOp())) {
```
MaheshRavishankar, line 135: What is `loopMaps` ?

MaheshRavishankar (approved):
Nicely done!

compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
IanWood1: Yes!

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
IanWood1: It's a map that contains all of the operations currently in the "fusion group".

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
IanWood1: This test might be a bit over-simplified, making it look stranger than it is. I've lost the exact IR from CI that was causing this issue but I should be able to re-create it on https://github.com/iree-org/iree/pull/22341

PR #22714, opened 2025-11-20
[docs] Clarify code review process

kuhar:
This is largely inspired by https://llvm.org/docs/CodeReview.html although I decided not to reference it directly, since our practices are not identical.

### DIFF
--- a/docs/website/docs/developers/general/contributing.md
+++ b/docs/website/docs/developers/general/contributing.md
@@ -234,8 +234,8 @@ information on using pull requests.
   comments should be addressed as well, either through follow-up discussion
   or code changes. Small issues may be fixed forward, but reviewers may also
   ask for your PR to be reverted. This is not a sign that your was faulty -- we
-  prefer to err on the side of caution and reverting allow more time for design
-  discussion.
+  prefer to err on the side of caution and reverting allows more time for
+  design discussion.
 
 ### :material-check-all: GitHub Actions workflows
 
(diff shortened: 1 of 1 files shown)


docs/website/docs/developers/general/contributing.md
ScottTodd: How about
1. keeping the "Please keep PRs small" bullet point at the top of the list. I would sequence these steps in terms of the stages of a code review - start with small PRs, get reviews, address comments, get approval, merge.
2. Adding a level of indentation for the "You may see trusted core contributors bending this rule", OR rewording it to cover "these rules" not just "this rule" now that there are multiple rules in this list

docs/website/docs/developers/general/contributing.md
```diff
@@ -218,6 +218,18 @@ use GitHub pull requests (PRs) for this purpose. Consult
 [GitHub Help](https://help.github.com/articles/about-pull-requests/) for more
 information on using pull requests.
 
+* We require review comments to be addressed before merging, even when there is
+  an approval from another reviewer. Unless stated otherwise, we treat request
```
ScottTodd, line 228: FYI there is a github setting that can enforce this: 

Image

I prefer the ability to leave conversations "unresolved" though, in case there is useful discussion there that I don't want hidden.

ScottTodd (approved):

MaheshRavishankar (approved):

docs/website/docs/developers/general/contributing.md
kuhar: yeah sometimes conversations are left unresolved because it's a free-form discussion instead of request for change

krzysz00 (commented):
lgtm

docs/website/docs/developers/general/contributing.md
hanhanW: I usually leave the conversion open when there are useful information for others.

PR #22719, opened 2025-11-20
[Input] Register IREETensorExtDialect for Torch plugin

IanWood1:
Registers `IREETensorExtDialect` to fix unregistered dialect error from bitcast ops created by BitCastTensorPass. Strangely, this error only occurs when printing IR and debug info, which is why it's not getting caught by CI. I think this was missed when `BitCastOp` was moved out of Flow.


compiler/plugins/input/Torch/PluginRegistration.cpp
```diff
@@ -66,6 +67,7 @@ struct TorchSession
     registry.insert<IREE::Flow::FlowDialect>();
```
IanWood1, line 67: I think Flow could be removed here, but I'm not 100% sure.

hanhanW: I'm not sure if we need it for something like `flow.tensor.transfer` hints or not.

hanhanW (approved):

PR #22722, opened 2025-11-21
[Stream] Encode packed_storage device and host tensors

sommerlukas:
Extends the Stream passes `EncodeDeviceTensors` and `EncodeHostTensors` to correctly handle tensors with the `packed_storage` encoding attached.

The two passes now encode tensors with the `packed_storage` attribute the same way that `i1` tensors were encoded with the experimental `iree-experimental-packed-i1-storage` option active.


sommerlukas:
This implements a part of the changes necessary for the third step outlined in #21185. More work is necessary further down in the pipeline (codegen).

PR is currently marked as draft, as it's based on #22688 and must be rebased after that was merged.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
```diff
@@ -99,6 +99,17 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
...
+bool PackedStorageAttr::isCompatibleWith(
+    IREE::Encoding::SerializableAttr other) const {
+  // The packed_storage encodings is compatible with empty encodings, because it
+  // doesn't impact tensor shape.
+  if (!other) {
+    return true;
+  }
+  // Otherwise, packed_storage is only compatible with itself.
+  return isa<IREE::Encoding::PackedStorageAttr>(other);
+}
```
hanhanW: This is outside my expectation. Can you elaborate why we need changes for encoding compatibility?

The interface method was created for padding encoding support in [6ba9cce27](https://github.com/iree-org/iree/commit/6ba9cce27f741c984c9d3914b679b2fb4fa798c3), and the `stream.tensor.clone` op did not allow different types. Do we hit such case in the work? Sharing some IR dump and context would be helpful here.

On the other hand, this belongs to a separate PR; you can maybe add a test like the reference commit.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/BUILD.bazel
```diff
@@ -32,6 +32,7 @@ iree_lit_test_suite(
             "emplace_transients.mlir",
             "emplace_transients_scf.mlir",
             "encode_device_tensors.mlir",
+            "encode_device_tensors_packed_storage.mlir",
```
hanhanW: I think naming it as `encode_device_tensors_encoding.mlir` looks better, which is more consistent with the host test (i.e., `encode_host_tensors_encoding.mlir`).

compiler/src/iree/compiler/Dialect/Stream/Transforms/EncodeTensors.cpp
```diff
@@ -365,7 +368,7 @@ struct EncodeTensorSplatOp
     }
 
     // Canonicalize the fill pattern into an integer type [i8, i16, i32, i64].
-    auto pattern = canonicalizeFillPattern(op.getValue(), rewriter);
+    auto pattern = canonicalizeFillPattern(op.getValue(), resultType, rewriter);
```
hanhanW, line 371: We should add tests for `stream.tensor.splat` to reflect the change.

compiler/src/iree/compiler/Dialect/Stream/Transforms/EncodeTensors.cpp
```diff
@@ -460,7 +463,7 @@ struct EncodeTensorFillOp
     }
 
     // Canonicalize the fill pattern into an integer type [i8, i16, i32, i64].
-    auto pattern = canonicalizeFillPattern(op.getValue(), rewriter);
+    auto pattern = canonicalizeFillPattern(op.getValue(), targetType, rewriter);
```
hanhanW, line 466: We should add tests for `stream.tensor.fill` to reflect the change.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/encode_device_tensors_encoding.mlir
hanhanW, line 1: IIUC, it tests that the element type is not extended to i8 when packed_storage encoding is used?

compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp
```diff
@@ -86,6 +86,13 @@ Type legalizeStorageElementType(Type elementType) {
                                         /*isPackedStorage=*/clEnableI1Support);
 }
 
+Type legalizeStorageElementType(RankedTensorType shapedType) {
+  bool isPackedStorage =
+      clEnableI1Support || IREE::Encoding::hasPackedStorageAttr(shapedType);
+  return legalizeStorageElementTypeImpl(shapedType.getElementType(),
+                                        isPackedStorage);
+}
```
hanhanW, line 94: We really need to kill the `clEncodingI1Support` flag and drop `*Impl` wrappers. The logic is hard to follow.

hanhanW (changes requested):
Thanks @sommerlukas, it will take me some time to review all the details because it's been a while since I studied the code.

How far are we to switch i1 masked attention test to use the encoding? I'd love to see a cleanup that removes those `*Impl` wrappers.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
sommerlukas: The intent of this change is to make `packed_storage` encoding compatible with no encoding at all (`nullptr`). 

The rationale for this change is the `flow.tensor.bitcast` operation that we use for the `i1` mask: 
```mlir
%mask = flow.tensor.bitcast %i8mask : tensor<2xi8> -> tensor<4x4xi1, #iree_encoding.packed_storage>
```
The `flow.tensor.bitcast` is lowered to `stream.tensor.clone` [here](https://github.com/iree-org/iree/blob/1a6681936a78138f7ae2b0fd9309d28dfe01ac5b/compiler/src/iree/compiler/Dialect/Stream/Conversion/FlowToStream/Patterns.cpp#L135-L160).

`stream.tensor.clone` however verifies that the encodings are compatible [here](https://github.com/iree-org/iree/blob/1a6681936a78138f7ae2b0fd9309d28dfe01ac5b/compiler/src/iree/compiler/Dialect/Stream/IR/StreamOps.cpp#L1928-L1949).

To allow a `flow.tensor.bitcast` from no encoding to `packed_storage` to be lowered, we need to add this compatibility.

Do you want to handle this scenario differently?

I can definitely move that change to a separate PR and add a test, too, if we want to stick to this approach.

compiler/src/iree/compiler/Dialect/Stream/Transforms/EncodeTensors.cpp
sommerlukas: I've added a test for `i1` `splat` with `packed_storage` in `encode_host_tensors_packing.mlir`

compiler/src/iree/compiler/Dialect/Stream/Transforms/EncodeTensors.cpp
sommerlukas: I've added a test for `i1` `fill` with `packed_storage` in `encode_host_tensors_packing.mlir`

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/encode_device_tensors_encoding.mlir
sommerlukas: Yes, exactly.

compiler/src/iree/compiler/Dialect/Encoding/Utils/ElementPackingUtils.cpp
sommerlukas: Yes, I agree, it can make things difficult to follow.

sommerlukas:
> How far are we to switch i1 masked attention test to use the encoding? I'd love to see a cleanup that removes those `*Impl` wrappers.

As described on DIscord, the current status is:
- With the changes from this PR, we can compile an attention example with `i1` mask and `packed_storage` attribute to Stream.
- With the changes from #22740 building on top of this PR, we can run that example through the entire compilation pipeline. I still need to verify that the computed result is correct.

The caveat is that the input must not have any unit extent dimensions, e.g., tensor shapes must be `4x4xi1` rather than `1x4x4xi1`. Otherwise, the `FoldUnitExtentDims` will produce invalid attention ops, because it doesn't handle `packed_storage` nicely yet.

I also don't yet have a good overview if there are any frontends that rely on the experimental option so that we could remove the `*Impl` wrappers.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
hanhanW: Okay, thanks for the reference code. This is exactly what I want; I'm happy that the theory compatible interface works well! Our design looks more promising to me now. (I was thinking to revert the change, but it looks really reasonable in theory to me.)

Anyway, what I'm looking for test coverage. We can't only have e2e working. So here is the new requests
- Let's have separate PR for the implementation.
- Let's add the test to flow->stream conversion, but not tensor_ops.

hanhanW:
> The caveat is that the input must not have any unit extent dimensions, e.g., tensor shapes must be 4x4xi1 rather than 1x4x4xi1. Otherwise, the FoldUnitExtentDims will produce invalid attention ops, because it doesn't handle packed_storage nicely yet.

After we land these PRs, let's add the test first.

> I also don't yet have a good overview if there are any frontends that rely on the experimental option so that we could remove the *Impl wrappers.

It's experimental path and we moved to other approaches, so you don't need to worry about the frontend issues. I just checked the search result on Github, and I dont find any other users.

https://github.com/search?q=--iree-experimental-packed-i1-storage&type=code

hanhanW (approved):
LGTM, many thanks

hanhanW:
Merging the PR because it's been out for a while. It is a step towards better world, and we will clean up the tech debts. I think @sommerlukas already has a branch that shows i1 working e2e with the encoding.

PR #22723, opened 2025-11-21
[Util] Implement InferIntDivisibilityOpInterface for affine ops

Max191:
This PR implements the InferIntDivisibilityOpInterface for affine.apply, affine.min, and affine.max operations. Affine apply gets the divisibility of its result expression, and affine.min/max gets the GCD of all result expression divisibilities. The implementation supports the following divisibilities and any compositions of them:
- Multiplication: product of operand divisibilities
- Addition/Subtraction: GCD (union) of operand divisibilities
- Division (floor/ceil): quotient when evenly divisible, else 1
- Modulo: falls back to minimum divisibility (1,1)

This PR also adds the TestIntegerDivisibilityAnalysis pass to more directly test divisibility analysis without relying on IR optimizations. The pass probes values consumed by `"iree_unregistered.test_int_divisibility"` ops and annotates them with computed divisibility attributes.

There is a small change to the arith.divui divisibility implementation to fallback to minimum divisibility when there is a remainder division, because we can't infer the divisibility when there is a remainder.


compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
```diff
@@ -0,0 +1,157 @@
+// RUN: iree-opt --split-input-file --iree-util-test-integer-divisibility-analysis --allow-unregistered-dialect %s | FileCheck %s
```
jerryyin, line 1: I searched for divisibility related test out of curiosity and find a `integer_divisibility.mlir`. That pass is already invoking `--iree-util-optimize-int-arithmetic`. Why do you need to invent this new test flag instead of relying on that one, but to add more affine.apply test cases there?

Max191: The optimize-int-arithmetic pass converts affine ops to arith ops before running the analysis, so it does not really test the analysis itself. We needed a pass to actually check what the analysis is setting, as opposed just verifying that some folding applies.

compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
```diff
@@ -0,0 +1,157 @@
...
+  util.return %2 : index
+}
+
+// -----
+
+// CHECK-LABEL: @affine_apply_add_gcd
+util.func @affine_apply_add_gcd(%arg0 : index, %arg1 : index) -> index {
+  %0 = util.assume.int %arg0<udiv = 16> : index
+  %1 = util.assume.int %arg1<udiv = 24> : index
+  %2 = affine.apply affine_map<(d0, d1) -> (d0 + d1)>(%0, %1)
```
nirvedhmeshram: would it be possible to support something like d0-d1, this happens in tiling of unaligned shape code. e.g
```
%39 = affine.min affine_map<(d0, d1) -> (8, d0 - d1)>(%26, %38)
```
and i have recently realized having divisibility analysis for this might be necessary to minimize creating scalar global loads.

nirvedhmeshram (approved):
LGTM, I do have a non-blocking but related question

MaheshRavishankar:
So, For affine ops we have traditionally just converted them to arith ops and done that. Not sure how tricky it is to add support for all affine expressions

compiler/src/iree/compiler/Dialect/Util/Transforms/TestIntegerDivisibilityAnalysis.cpp
```diff
@@ -0,0 +1,67 @@
...
+      if (op->getName().getStringRef() ==
+              "iree_unregistered.test_int_divisibility" &&
+          op->getNumOperands() == 1) {
+        queryOps.emplace_back(op, op->getOperand(0));
+      }
+    });
+
+    DataFlowSolver solver;
+    solver.load<dataflow::DeadCodeAnalysis>();
+    solver.load<dataflow::SparseConstantPropagation>();
```
hanhanW: Do we need them? If so, is it worth to add a documentation?

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -46,6 +48,229 @@ getDivisibilityOfOperand(Value v,
...
+class AffineExprDivisibilityFinder
+    : public AffineExprVisitor<AffineExprDivisibilityFinder,
+                               IREE::Util::ConstantIntDivisibility> {
+public:
+  using ExprDivisibilityMap =
+      llvm::DenseMap<AffineExpr, IREE::Util::ConstantIntDivisibility>;
+  AffineExprDivisibilityFinder(ExprDivisibilityMap &divisibilityMap)
+      : divisibilityMap(divisibilityMap) {}
+
+  /// Constant expressions are trivial, since they are always static.
```
hanhanW: This looks more like implementation details, so let's move it into the function?

Same for `visitDimExpr` and `visitSymbolExpr`.

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -46,6 +48,229 @@ getDivisibilityOfOperand(Value v,
...
+    }
+    IREE::Util::ConstantIntDivisibility lhsDiv = visit(expr.getLHS());
+    auto constRhs = dyn_cast<AffineConstantExpr>(expr.getRHS());
+    if (!constRhs) {
+      return IREE::Util::ConstantIntDivisibility(1, 1);
+    }
+    int64_t constValue = constRhs.getValue();
+    uint64_t divUDiv = lhsDiv.udiv() % static_cast<uint64_t>(constValue) == 0
+                           ? lhsDiv.udiv() / static_cast<uint64_t>(constValue)
+                           : 1;
```
hanhanW: `static_cast<uint64_t>` on negative number creates a very large positive number. E.g., `-4` becomes `18446744073709551612`.

I'd initialize the value to `1` and create few if statement to handle the case.

compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
```diff
@@ -0,0 +1,157 @@
...
+// -----
+
+// CHECK-LABEL: @affine_max_uniform_divisibility
+util.func @affine_max_uniform_divisibility(%arg0 : index) -> index {
+  %0 = util.assume.int %arg0<udiv = 32> : index
+  %1 = affine.max affine_map<(d0) -> (d0, d0 - 64)>(%0)
+  // CHECK: divisibility = "udiv = 32, sdiv = 32"
+  %2 = "iree_unregistered.test_int_divisibility"(%1) : (index) -> index
+  util.return %2 : index
+}
```
hanhanW, line 134: I honestly don't know how MLIR handles negative constants. This may not cover the test case. I suspect this one is `DimExpr subtract PositiveConstExpr`.

Maybe we can add a test like `(d0) -> (d0 * -4))` to exercise negative constants directly.

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -46,6 +48,229 @@ getDivisibilityOfOperand(Value v,
...
+  visitInvalidExpr(AffineBinaryOpExpr expr) {
+    return IREE::Util::IntegerDivisibility::getMinDivisibility().getValue();
+  }
+
+  /// Helper shared by ceildiv and floordiv implementations. The divisibility of
+  /// a division is simply the quotient of its constituents' divisibilities as
+  /// long as the division has no remainder. If there is a remainder, then the
+  /// divisibility cannot be easily inferred, so we fallback to the minimum
+  /// divisibility.
+  IREE::Util::ConstantIntDivisibility visitDivExpr(AffineBinaryOpExpr expr) {
```
hanhanW, line 150: Do we return `{1, 1}` for zero divisibility? The input is not valid, but we should not crash right? I think falling back to `{1, 1}` seems reasonable in this case.

If you agree with this, please also add a test.

compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
Max191: This will work already, since subtraction is just handled as addition with a negative coefficient.

Max191:
> So, For affine ops we have traditionally just converted them to arith ops and done that. Not sure how tricky it is to add support for all affine expressions

This PR should support arbitrary affine expressions. It implements the basic binary operators, and then the full divisibility is just a composition of them.

I'm not sure if lowering to arith ops is a great solution, since this analysis will be used early on in compilation (e.g., sometimes before all tiling and distribution is done), and we do a lot of affine apply/min/max composition during these compilation phases. The affine divisibility implementation turned out to not be too bad, since I mainly just had to implement the core operators.

compiler/src/iree/compiler/Dialect/Util/Transforms/TestIntegerDivisibilityAnalysis.cpp
Max191: I think this is actually more needed for analyzing through some control flow ops, so I think it's not needed for this PR. I'll remove it here and then add it in the next PR about scf induction variable args.

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
Max191: Yes, thanks for the catch! I need to handle the zero case.

compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
Max191: Divisibilities should all just be the absolute values, so we just take the absolute value for negative constants. I added a test for it though.

compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
```diff
@@ -0,0 +1,169 @@
...
+
+// -----
+
+// CHECK-LABEL: @affine_apply_with_symbol
+util.func @affine_apply_with_symbol(%arg0 : index, %arg1 : index) -> index {
+  %0:2 = util.assume.int %arg0<udiv = 16>,
+                         %arg1<udiv = 16> : index, index
+  %1 = affine.apply affine_map<(d0)[s0] -> (d0 + s0)>(%0#0)[%0#1]
+  // CHECK: divisibility = "udiv = 16, sdiv = 16"
```
MaheshRavishankar, line 97: Why doesnt this become 32?

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -46,6 +48,233 @@ getDivisibilityOfOperand(Value v,
...
+public:
+  using ExprDivisibilityMap =
+      llvm::DenseMap<AffineExpr, IREE::Util::ConstantIntDivisibility>;
+  AffineExprDivisibilityFinder(ExprDivisibilityMap &divisibilityMap)
+      : divisibilityMap(divisibilityMap) {}
+
+  IREE::Util::ConstantIntDivisibility
+  visitConstantExpr(AffineConstantExpr expr) {
+    // Constant expressions are trivial, since they are always static.
+    uint64_t constValue = std::abs(expr.getValue());
```
MaheshRavishankar, line 70: So this is going to make `sdiv` positive. I dont fully understand how this was set up, but just wanted to see if you had a rationale for this, or (like me) this is not fully reasoned about.

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -46,6 +48,233 @@ getDivisibilityOfOperand(Value v,
...
+      return divisibilityMap[expr];
+    }
+    auto constRhs = dyn_cast<AffineConstantExpr>(expr.getRHS());
+    // Division by zero is undefined, so return the minimum divisibility.
+    if (!constRhs || constRhs.getValue() == 0) {
+      return IREE::Util::ConstantIntDivisibility(1, 1);
+    }
+    int64_t constValue = std::abs(constRhs.getValue());
+    IREE::Util::ConstantIntDivisibility lhsDiv = visit(expr.getLHS());
+    uint64_t divUDiv = lhsDiv.udiv() % static_cast<uint64_t>(constValue) == 0
```
MaheshRavishankar: Nit: do cast once for readability?

MaheshRavishankar (approved):
Nice! Looks good to me.

Maybe @bjacob has more nuanced feedback.

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
bjacob: Browsing the code (which is new to me), the `ConstantIntDivisibility` constructor at the next line surprisingly takes `(uint64_t udiv, uint64_t sdiv)`:
https://github.com/iree-org/iree/blob/b9afdb9e963620ff3dc78329465dad28a5bfdcb6/compiler/src/iree/compiler/Dialect/Util/IR/UtilTypes.h#L178

I guess that this motivates `uint64_t constValue` above, as it is going to be casted to `uint64_t` anyway at the next line. 

I wonder if that constructor is a typo or intentional?  Noticing that it is at least consistent with this method a few lines down:

https://github.com/iree-org/iree/blob/b9afdb9e963620ff3dc78329465dad28a5bfdcb6/compiler/src/iree/compiler/Dialect/Util/IR/UtilTypes.h#L186

To me, not knowing anything about the specifics here, this looks suspisciously close to a violation of "don't use unsigned to denote non-negative values" rule,
>  In particular, do not use unsigned types to say a number will never be negative. 

Happy to dive into this (I still don't know anything about this stuff)  BUT this looks like in order to be useful, we need to dive into the first principles here -- what are the exact semantics that we want here, why are unsigned types involved, should they be involved etc.

MaheshRavishankar: I am not entirely sure. Maybe @stellaraccident  has more context on those, but I would propose first landing this and then iterating on the "signed" vs "unsigned" parts here. +1 on nailing this down a bit more though. This seems like a good thing to address soon before it gets out of hand.

compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
Max191: These are just divisibilities, so you have to take the GCD. `%arg0` and `%arg1` can more generally be described as `x * 16` and `y * 16`, where `x` and `y` are any integers. The sum of them is `(x + y) * 16`, which still has a divisibility of 16.

(e.g., `%arg0` could be 16, and `%arg1` could be 32, and then the sum is `48`, which is not divisible by 32)

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
Max191: Dropping a summary of what Benoit and I discussed here so it is not lost: https://discord.com/channels/689900678990135345/1442894583624106036/1442903483014578317

Full thread starts here: https://discord.com/channels/689900678990135345/1442894583624106036

IanWood1:
@Max191 It looks like this is breaking compilation of torch_models for llama 8b fp8 with data tiling. This is post submit CI run [before](https://github.com/iree-org/iree/actions/runs/20032789150/job/57449215649) and [after](https://github.com/iree-org/iree/actions/runs/20037515165/job/57463661192) this PR was merged. You can ignore the dispatch count failure. I'm fixing it here https://github.com/iree-org/iree/pull/22855

PR #22728, opened 2025-11-21
Initial tools/utils/ iree-* developer utilities.

benvanik:
Developer-facing utilities  for IREE contributors, organized as:

**Lit Test Tools** (`tools/utils/bin/iree-lit-*`):
- `iree-lit-list`: List test cases in files
- `iree-lit-extract`: Extract individual test cases
- `iree-lit-replace`: Replace test case content atomically
- `iree-lit-test`: Run tests in isolation with debug capabilities
- `iree-lit-lint`: Lint tests against STYLE_GUIDE.md

**CI Triage Tools** (`tools/utils/bin/iree-ci-*`):
- `iree-ci-triage`: Analyze CI failures from GitHub Actions
- `iree-ci-garden`: Manage failure corpus for pattern development

Use `--help` on any tool for detailed usage.
See `tools/utils/README.md` for full documentation.

**Setup**:
```bash
pip install -e tools/utils
export PATH="$PATH:$PWD/tools/utils/bin"
# or, if you use direnv:
cp .envrc.example .envrc
direnv allow
```

**Claude Code Integration**:
- Skill: iree-lit-tools for MLIR test authoring guidance
- Commands: /iree-ci-triage, /iree-lit-test, /iree-lit-lint

Skills and commands should be detected automatically.
The `auto-approve-iree-tools.sh` is a useful pre-bash hook that approves all iree-* tools. Somewhat #yolo so it's opt in, but it does allow for much less naggy triage sessions and is required to get smooth heredoc support. `cp .claude/settings.json.template .claude/settings.json` to enable.

### iree-lit-list
List test cases in a file.
```bash
iree-lit-list test.mlir              # Show all cases with metadata
iree-lit-list test.mlir --count      # Just the count
iree-lit-list test.mlir --names      # Space-separated names
iree-lit-list test.mlir --json       # Machine-readable
```

### iree-lit-extract
Extract individual test cases.
```bash
iree-lit-extract test.mlir --case 3        # Extract case #3
iree-lit-extract test.mlir --line 123      # Extract case containing line 123
iree-lit-extract test.mlir --name "foo"    # Extract case named "foo"
iree-lit-extract test.mlir -c 1,3,5        # Multiple cases
```

### iree-lit-replace
Replace a test case atomically (reads new content from stdin).
```bash
iree-lit-replace test.mlir --case 3 < fixed_case.mlir

# Heredoc for inline replacement (no temp files needed):
iree-lit-replace test.mlir --case 3 <<'EOF'
// RUN: iree-opt --my-pass %s | FileCheck %s
// CHECK-LABEL: @my_test
func.func @my_test() {
  return
}
EOF
```

### iree-lit-test
Run tests in isolation with debug capabilities.
```bash
iree-lit-test test.mlir                    # Run all cases
iree-lit-test test.mlir --case 2           # Run only case #2
iree-lit-test test.mlir -c 1-3             # Run cases 1, 2, 3
iree-lit-test test.mlir --name "foo"       # Run case named "foo"
iree-lit-test test.mlir --verbose          # Show full output
iree-lit-test test.mlir --extra-flags "--debug"  # Inject flags
iree-lit-test test.mlir --dry-run          # Show commands without running

# Heredoc for rapid testing (no temp files needed):
iree-lit-test --run 'iree-opt --my-pass %s | FileCheck %s' <<'EOF'
// CHECK-LABEL: @test_fusion
// CHECK: my.fused_op
func.func @test_fusion() {
  %0 = my.op1
  %1 = my.op2 %0
  return
}
EOF
```

### iree-lit-lint
Lint tests against the style guide.
```bash
iree-lit-lint test.mlir                    # Lint all cases
iree-lit-lint test.mlir --case 2           # Lint only case #2
iree-lit-lint test.mlir --errors-only      # Only show errors
iree-lit-lint test.mlir --help-style-guide # Show full style guide
```

(most of this was authored with claude and dogfooded on significant test refactoring/authoring, YMMV and it needs refinement but it's already been a massive help)

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -1,0 +1,67 @@
+# Rules for .claude/ Directory
+
+**Every token costs.** Anything checked into `.claude/` is paid for on every invocation by every developer. Guard this directory jealously.
+
+## CLAUDE.md
+
+**Bar: Benefits ALL developers on EVERY use.**
+
+We do not yet have project-wide instructions that meet this bar. Keep your instructions in:
+- `~/.claude/CLAUDE.md` (personal global)
+- `CLAUDE.local.md` (personal per-project, gitignored)
+
+`/init` is quite decent at generating a `CLAUDE.md` and you can write your own prompt based on the kind of work you do to guide it. There's some [good example init-like prompts](https://github.com/kaushikgopal/dotfiles/blob/master/.ai/commands/initialize.md) out there to use as reference.
+
+Bad example: [pytorch/CLAUDE.md](https://github.com/pytorch/pytorch/blob/2dd529df0092799f68ee7afcf52338276906706a/CLAUDE.md) - includes testing syntax that benefits a fraction of uses.
+
+## Skills
+
+**Bar: Benefits nearly all developers. Durable (useful 1 year from now). Graduated from well-honed workflows.**
+
+Skills trigger automatically and their descriptions appear in context regardless of whether they are used (context bloat). A skill is ready for inclusion in the main repository when you've explained the same approach to Claude multiple times across many sessions and it has stabilized into a reliable workflow _and_ that workflow is one adopted by several other contributors. Otherwise, keep it local or in shared plugins.
+
+**Requirements:**
+- Broadly applicable (not for a single pass, single refactoring, or single API)
+- Concise (reference style guides, don't embed them)
+- Durable (core workflow aid, not a temporary migration helper)
+
+**Anti-patterns:**
+- Triggers on common actions (docstrings, formatting) - see [pytorch docstring skill](https://github.com/pytorch/pytorch/blob/43c30f607eeca0d3e9a26911d9c2131fc250eadd/.claude/skills/docstring/SKILL.md)
+- Narrow scope (single API conversion) - see [pytorch AT_DISPATCH_V2 skill](https://github.com/pytorch/pytorch/blob/3ca216ae172e35adde34a319a1a01faaf218e7c5/.claude/skills/add-uint-support/SKILL.md)
+- Embeds reference material that belongs in `--help` or external docs
+
+**Prefer instead:**
+- `~/.claude/skills/` for personal skills
+- [Plugins](https://code.claude.com/docs/en/plugins) for sharing with small groups
+
+## Commands
+
+**Bar: Daily workflow automation. Graduated from manual processes.**
+
+Commands should encode workflows that engineers do manually today. They must be:
+- Used frequently (not one-off investigations)
+- Stable (the underlying tools/processes are mature)
+- Broadly applicable (useful to most engineers, not specialists)
+
+**Prefer instead:**
+- `~/.claude/commands/` for personal commands
+- [Plugins](https://code.claude.com/docs/en/plugins) for sharing with small groups
+
+## Hooks
+
+Hooks auto-approve or modify Claude's behavior. They must be:
+- Security-conscious (document threat model)
+- Minimal (approve specific tools, not broad categories)
+- Self-documenting (embed rationale in the script)
+
+## Summary
+
+| Location | Bar | Examples |
+|----------|-----|----------|
+| `CLAUDE.md` (checked in) | All devs, every use | None yet |
+| `.claude/skills/` | Nearly all devs, durable | `iree-lit-tools` |
+| `.claude/commands/` | Daily workflows | `/iree-ci-triage`, `/iree-lit-test` |
+| `~/.claude/` | Personal | Your preferences |
+| Plugins | Small groups | Team-specific workflows |
+
+When in doubt, keep it local. Graduate to checked-in only after proving value across the team.
--- a/.claude/commands/iree-ci-triage.md
+++ b/.claude/commands/iree-ci-triage.md
@@ -1,0 +1,31 @@
+---
+allowed-tools: Bash(iree-ci-triage:*), Bash(gh:*), Read, Grep
+description: Triage CI failures for a PR or commit
+---
+
+## Context
+
+**Target:**
+!`if [ -n "$ARGUMENTS" ]; then echo "PR/run: $ARGUMENTS"; else echo "No PR specified - showing recent failing PRs"; fi`
+
+**CI Triage Results:**
+!`if [ -n "$ARGUMENTS" ]; then PYTHONPATH=tools/utils python3 -m ci.iree_ci_triage --pr "$ARGUMENTS" --checklist 2>&1 | head -200; else gh pr list --state open --limit 10 --json number,title,headRefName --template '{{range .}}#{{.number}} {{.title}} ({{.headRefName}}){{"\n"}}{{end}}' 2>&1; fi`
+
+## Your Task
+
+Analyze the CI triage results above and help fix the failures:
+
+1. **If no PR was specified**: Ask which PR to triage from the list shown
+
+2. **For each failure category**:
+   - **Compile errors**: Read the failing file and fix the issue
+   - **Test failures**: Run `iree-lit-test` to debug, then fix CHECK patterns or code
+   - **Sanitizer issues**: Analyze the stack trace and fix memory/thread issues
+   - **Infrastructure flakes**: Note these are transient, suggest re-running
+
+3. **Priority order**: Fix errors before warnings, compile errors before test failures
+
+4. **After fixing**: Verify the fix locally before suggesting it's ready
+
+Use `iree-ci-triage --pr N --json` for machine-readable output if needed.
+Use `iree-lit-lint` and `iree-lit-test` for test-related failures and reference the `iree-lit-tools` skill for guidance on style and common workflows.
--- a/.claude/commands/iree-lit-test.md
+++ b/.claude/commands/iree-lit-test.md
@@ -1,0 +1,37 @@
+---
+allowed-tools: Bash(iree-lit-*), Bash(iree-opt:*), Read, Edit
+description: Run and debug a lit test
+---
+
+## Context
+
+**Target file(s):**
+!`if [ -n "$ARGUMENTS" ]; then echo "$ARGUMENTS"; else git diff --name-only HEAD 2>/dev/null | grep '\.mlir$' | head -5 || echo "No .mlir files in diff"; fi`
+
+**Test structure:**
+!`file="$ARGUMENTS"; if [ -z "$file" ]; then file=$(git diff --name-only HEAD 2>/dev/null | grep '\.mlir$' | head -1); fi; if [ -n "$file" ] && [ -f "$file" ]; then PYTHONPATH=tools/utils python3 -m lit_tools.iree_lit_list "$file" 2>&1; else echo "No test file found"; fi`
+
+## Your Task
+
+Run and debug the lit test(s):
+
+1. **If no file specified**: Work with the modified `.mlir` files shown above
+
+2. **Run the test**:
+   ```bash
+   iree-lit-test <file> --case N          # Run specific case
+   iree-lit-test <file> --verbose         # See full output
+   iree-lit-test <file> --dry-run         # See commands without running
+   ```
+
+3. **If a test fails**:
+   - Use `--verbose` to see the actual vs expected output
+   - Extract the case: `iree-lit-extract <file> --case N`
+   - Run the pipeline manually to see actual IR
+   - Update CHECK patterns to match actual output
+
+4. **Before finishing**:
+   - Run `iree-lit-lint <file>` to check for style issues
+   - Verify all cases pass: `iree-lit-test <file>`
+
+Use the `iree-lit-tools` skill for guidance on style and common workflows.
--- a/.claude/commands/iree-pre-commit.md
+++ b/.claude/commands/iree-pre-commit.md
@@ -1,0 +1,32 @@
+---
+allowed-tools: Bash(pre-commit run:*), Bash(git add:*), Bash(git status:*), Bash(git diff:*), Bash(bash -c:*)
+description: Run pre-commit on modified files and fix issues
+---
+
+## Context
+
+**Modified files (staged or unstaged):**
+!`git diff --name-only HEAD`
+
+**Pre-commit results on modified files:**
+!`bash -c 'files=$(git diff --name-only HEAD); if [ -n "$files" ]; then pre-commit run --files $files --show-diff-on-failure 2>&1; else echo "No modified files to check"; fi'`
+
+**Current git status:**
+!`git status --short`
+
+## Your Task
+
+Analyze the pre-commit results above and fix all issues that were **not** auto-fixed:
+
+1. **Auto-fixed issues (IGNORE these)**: When you see "files were modified by this hook", those files were already fixed automatically (e.g., Black formatting, clang-format). No action needed.
+
+2. **Manual fixes needed (FIX these)**: When you see detailed error messages with line numbers and specific issues (e.g., Ruff linting errors, type annotation issues), these require manual fixes.
+
+For each issue requiring manual fixes:
+- Read the affected files
+- Apply the fixes following the error messages
+- Verify the fix with `pre-commit run --files <file>` after editing
+
+After fixing all issues, run `pre-commit run` again to verify everything passes.
+
+**Important**: Only fix files that have real errors. Don't touch files that were auto-formatted unless they still have errors after auto-formatting.
--- a/.envrc.example
+++ b/.envrc.example
@@ -1,0 +1,10 @@
+# direnv configuration for IREE tools/utils
+# Copy to .envrc and run `direnv allow`
+#
+# See: https://direnv.net/
+
+# Add tools to PATH
+PATH_add tools/utils/bin
+
+# Optional: Set build directory for tools that need iree-opt/FileCheck
+# export IREE_BUILD_DIR="../builds/main"
--- a/build_tools/cmake/build_and_test_asan.sh
+++ b/build_tools/cmake/build_and_test_asan.sh
@@ -48,17 +48,9 @@ CMAKE_ARGS=(
   "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
 )
 
-echo "::group::Configuring CMake"
 "${CMAKE_BIN?}" "${CMAKE_ARGS[@]?}"
-echo "::endgroup::"
-
-echo "::group::Building all"
 "${CMAKE_BIN?}" --build "${BUILD_DIR?}" -- -k 0
-echo "::endgroup::"
-
-echo "::group::Building test deps"
 "${CMAKE_BIN?}" --build "${BUILD_DIR?}" --target iree-test-deps -- -k 0
-echo "::endgroup::"
 
 # Respect the user setting, but default to as many jobs as we have cores.
 export CTEST_PARALLEL_LEVEL=${CTEST_PARALLEL_LEVEL:-$(nproc)}
@@ -116,16 +108,12 @@ label_exclude_regex="($(IFS="|" ; echo "${label_exclude_args[*]?}"))"
 
 pushd "${BUILD_DIR?}"
 
-echo "::group::Running main project ctests"
 ctest \
   --timeout 900 \
   --output-on-failure \
   --no-tests=error \
   --label-exclude "${label_exclude_regex}"
-echo "::endgroup::"
 
-echo "::group::Running llvm-external-projects tests"
 cmake --build . --target check-iree-dialects -- -k 0
-echo "::endgroup::"
 
 popd
--- a/tools/utils/CLAUDE.md
+++ b/tools/utils/CLAUDE.md
@@ -1,0 +1,64 @@
+# CLAUDE.md - Agent Rules for tools/utils/
+
+Rules for AI agents contributing to IREE developer utilities.
+
+## Critical Rules
+
+1. **Never `print()` directly** - Use `console.error/warn/note/success` for messages
+2. **Never `sys.exit(N)`** - Use `return exit_codes.SUCCESS/ERROR/NOT_FOUND`
+3. **Never raw file writes** - Use `fs.safe_write_text()` for atomic UTF-8 writes
+4. **Always add common flags** - Call `cli.add_common_output_flags(parser)`
+5. **JSON purity** - Never mix `print()` with JSON output; use `console.print_json()`
+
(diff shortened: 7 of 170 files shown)


qedawkins (commented):
Nice! I'm not going to actually go read all of this, so some pointers on where to look for review would help while I go explore.

I have a couple high level concerns also:

1. How do we plan to prevent personal preference from causing drift/tension in checked in context? Most of what I see (i.e. the commands) look fine, but I don't really know how to approach reviewing changes to those files.
2. If I am making a change that breaks one of these tools somehow, am I expected to maintain/fix it? For example, imagine I want to do a CI restructure and end up invalidating the CI triaging tools. I don't want additional maintenance burden on developers for convenience tools.

benvanik:
Yeah, I don't expect anyone to read this :P If you wanted to give it a shot that'd be useful, though! (I've tried it on two machines, but they're pretty similar)

RE 1: Good question! I'm not going to allow any context to land in the repo :) See https://github.com/iree-org/iree/blob/14c86589bb46f031da51aebc94f247f6fd7f0300/.claude/README.md (I wanted to call it HUMANS.md, but then github wouldn't render it by default). I am an extreme context miser so the bar is very high. Instead of people dropping pytorch-quality crap we're going to probably just setup an iree-org/claude-plugins or something and then can add as many random things we want there ala https://github.com/anthropics/claude-code/tree/main/plugins. The skills checked in are all-developers/nearly-all-of-the-time, and there's not very many of those. There's also not very many commands that meet that bar. Basically, local is always best for stuff, and if you want to share in more than a gist a plugins repo is the best way.

RE 2: I'd expect claude to fix them if there was any issue, the code is really well setup for that. As for brittleness: I tried to make everything generic so it's just github API calls to query jobs, their status/check reports, and fetch their logs. So workflow changes won't impact anything (though I do have our "summary" workflows hidden by default), which is nice (it'll work in forks with other workflows, etc). The error extraction from the logs is all CI-agnostic and it can run on local builds to triage issues as well, and they're all best-effort (we miss some today, but I've got quite a few of the major things like cmake/bazel/onnx/lit/mlir). The intent is that someone (me) runs iree-ci-garden to grab all runs, does a classification to see if there's any new failures we couldn't identify, and then have claude go figure them out using the documentation/guides/etc. I'd love to make it a nightly github action at some point so it could send PRs for novel failures, but things are quite slow moving in infra land and most of the errors are stable (as stable as textual errors are). Worst case if nothing works you still get a list of failed jobs and their logs downloaded so pointing an agent at them is stupid easy. Same with /iree-ci-triage PR# - worst case you at least have the logs local and can start working on them.

qedawkins:
> RE 1: Good question! I'm not going to allow any context to land in the repo :)

Great, that works for me. I find myself reorganizing the kind of context available for hyper specific/agents/tasks a lot and having to deal with an extra file inserting itself into my workflow would be annoying. Just not checking any in sounds great.

> Instead of people dropping pytorch-quality crap we're going to probably just setup an iree-org/claude-plugins or something and then can add as many random things we want there

This was exactly what I was going to suggest if we hit the point of needing it. SGTM

> RE 2

Sounds good. As long as it stays as generic as possible and non-blocking then no complaints from me.

> Worst case if nothing works you still get a list of failed jobs and their logs downloaded so pointing an agent at them is stupid easy. Same with /iree-ci-triage PR# - worst case you at least have the logs local and can start working on them.

Yeah this is what I've been doing and it'll always work.

It's late for me but I'll finish a review tomorrow. I still want to read the contents of `.claude` on review and as a human I'm still a little apprehensive of checking in so much code from bots :P. The tools were great when I tried them before though so I'll try to review soon.

.claude/README.md
```diff
@@ -0,0 +1,67 @@
+# Rules for .claude/ Directory
```
krzysz00, line 1: Shouldn't this not be in this directory then?

.claude/commands/iree-ci-triage.md
```diff
@@ -0,0 +1,31 @@
+---
+allowed-tools: Bash(iree-ci-triage:*), Bash(gh:*), Read, Grep
+description: Triage CI failures for a PR or commit
+---
+
+## Context
+
+**Target:**
+!`if [ -n "$ARGUMENTS" ]; then echo "PR/run: $ARGUMENTS"; else echo "No PR specified - showing recent failing PRs"; fi`
```
krzysz00, line 9: Can we shellcheck these? For one thing, I'm thinning these should be `[[`

.claude/commands/iree-pre-commit.md
```diff
@@ -0,0 +1,32 @@
+---
+allowed-tools: Bash(pre-commit run:*), Bash(git add:*), Bash(git status:*), Bash(git diff:*), Bash(bash -c:*)
```
krzysz00, line 2: Isn't allowing bash -c equivalent to allowing anything?

tools/utils/CONTRIBUTING.md
```diff
@@ -0,0 +1,265 @@
...
+**Skip when:**
+- Tool only parses/lists/extracts text
+- Tool is a pure utility (formatting, file manipulation)
+
+### Search order
+
+1. `IREE_BUILD_DIR` environment variable (override)
+2. `./build/` (in-tree build)
+3. `../<worktree>-build/` (worktree pattern)
+4. `../iree-build/` (main repo build)
```
krzysz00, line 232: So  I'd like to call for my workflow - which I want to share out, where the source is in iree/<worktree>/src and the build is in iree/<worktree>/build - to be added here 

Though I can also set the environment variable thanks to direnv 

(I should present on my new setup at some point ...)

krzysz00 (commented):
Will review properly later, initial comments because I looked

benvanik:
> as a human I'm still a little apprehensive of checking in so much code from bots :P

Me too! I went pretty hard on the testing, style enforcement (ruff + custom rules), and documentation (style guides, developer docs, claude docs, and usage) and spent a decent amount of time refining it over the last few weeks. Even though there are certainly issues I do stand by it enough to submit it under my name, since I made it better than I could have written solo and it's now our most robust piece of infrastructure :P My tolerance for AI slop is quite low so honestly my biggest fear with landing something like this has been that someone will think this is tacit approval to throw slop in the repo when it most definitely isn't.

krzysz00:
I'm going to call for splitting the PR on principle

benvanik:
Nah, not worth it.

```
● tools/utils/ distribution (47,051 total lines)
  ═══════════════════════════════════════════════

  Category              Lines    Files    %
  ────────────────────────────────────────────
  Code (*.py)          21,767      66    46%
  Tests (test_*.py)    19,658      46    42%
  Documentation (*.md)  3,894      12     8%
  Fixtures              1,595      24     3%
  Bin wrappers            137       7    <1%

  Roughly 46% code, 45% tests+fixtures, 8% docs.
```

As an admin, I'm more than happy with that. Since I don't expect anyone to be reading these PRs (unless you're volunteering to spend the next few weeks reviewing them ;) there's no value. These are no-ops to the project (besides pre-commit) as they have no dependencies, so there's no real risk.

krzysz00:
I'm going to put a tentative soft block on this one until at least one non-Ben human has read the code and has some understanding of what it's all doing. I'm willing to be talked out of this over the next few days, though.

(I could take a the few days needed to be said human, if it came down to it, And I'd just like to state my position that, if this PR wasn't by someone I trust to have taken this process seriously, I would be closing it as unreviewable right now)

.envrc.example
```diff
@@ -0,0 +1,10 @@
+# direnv configuration for IREE tools/utils
+# Copy to .envrc and run `direnv allow`
+#
+# See: https://direnv.net/
+
+# Add tools to PATH
+PATH_add tools/utils/bin
+
+# Optional: Set build directory for tools that need iree-opt/FileCheck
+# export IREE_BUILD_DIR="../builds/main"
```
krzysz00, line 10: There're some more things I'd think to add in here, namely adding the build's main directory to PATH

(Heck, I've got a use_iree that I need to tweak a bit but could pass on)

krzysz00 (commented):
So, having slept on this, there is actually a graph of separable PRs here and I do think we should land this as such (both because of the precedent and because it makes it easier to understand the history of this stuff).

For instance,
a) Adding guidelines for AI context 
b) Some chsnges to CMake utility scripts 
c)
c.1) Definingthe tools/utils directory 
c.2) Adding tools for working with lit rests 
c.3) Adding tools for diagnosing CI failures (might be separable into the library core, the configuration, and the scripts?)

There might be more, but those are the ones that come to mind 

(I'd maybe make the argument that the code and the instructions to AI tools on how to use it could be separate PRs ... but realistically said instructions are approximately documentation so it's fine)

And speaking of documentation ... I think a bunch of this Markdown should be in our developer documentation on the website. At the very least, there should be references to tools/utils and your options for setting it up there

benvanik:
Noted, and post-submission patches/reviews are very welcome as others start to use it.

.claude/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+**Bar: Benefits ALL developers on EVERY use.**
+
+We do not yet have project-wide instructions that meet this bar. Keep your instructions in:
+- `~/.claude/CLAUDE.md` (personal global)
+- `CLAUDE.local.md` (personal per-project, gitignored)
+
+`/init` is quite decent at generating a `CLAUDE.md` and you can write your own prompt based on the kind of work you do to guide it. There's some [good example init-like prompts](https://github.com/kaushikgopal/dotfiles/blob/master/.ai/commands/initialize.md) out there to use as reference.
+
+Bad example: [pytorch/CLAUDE.md](https://github.com/pytorch/pytorch/blob/2dd529df0092799f68ee7afcf52338276906706a/CLAUDE.md) - includes testing syntax that benefits a fraction of uses.
```
qedawkins, line 15: The linked CLAUDE.md is pretty small. Can we just import it or something like it here to avoid throwing shade? It looks a little targeted...

.claude/README.md
```diff
@@ -0,0 +1,67 @@
...
+Skills trigger automatically and their descriptions appear in context regardless of whether they are used (context bloat). A skill is ready for inclusion in the main repository when you've explained the same approach to Claude multiple times across many sessions and it has stabilized into a reliable workflow _and_ that workflow is one adopted by several other contributors. Otherwise, keep it local or in shared plugins.
+
+**Requirements:**
+- Broadly applicable (not for a single pass, single refactoring, or single API)
+- Concise (reference style guides, don't embed them)
+- Durable (core workflow aid, not a temporary migration helper)
+
+**Anti-patterns:**
+- Triggers on common actions (docstrings, formatting) - see [pytorch docstring skill](https://github.com/pytorch/pytorch/blob/43c30f607eeca0d3e9a26911d9c2131fc250eadd/.claude/skills/docstring/SKILL.md)
+- Narrow scope (single API conversion) - see [pytorch AT_DISPATCH_V2 skill](https://github.com/pytorch/pytorch/blob/3ca216ae172e35adde34a319a1a01faaf218e7c5/.claude/skills/add-uint-support/SKILL.md)
```
qedawkins, line 30: Same here. Also I question whether sending agents on a web search really saves more context than an abridged inline example (because we both know that even if this were called HUMANS.md, bots are gonna be the ones reading this the most :P)

.claude/commands/iree-lit-test.md
```diff
@@ -0,0 +1,37 @@
...
+   ```bash
+   iree-lit-test <file> --case N          # Run specific case
+   iree-lit-test <file> --verbose         # See full output
+   iree-lit-test <file> --dry-run         # See commands without running
+   ```
+
+3. **If a test fails**:
+   - Use `--verbose` to see the actual vs expected output
+   - Extract the case: `iree-lit-extract <file> --case N`
+   - Run the pipeline manually to see actual IR
```
qedawkins, line 30: Does not have to be added here, but a flag or tool to extract the command to reproduce the failing test case could be worth it. I found that unless I gave it the command explicitly, the reproducer command it came up with kept morphing into different tools (llvm-lit, python's lit, pytest, all sorts of random stuff). I'd say 80-20 that it actually found the correct command from a line like this.

.envrc.example
qedawkins, line 10: Is this supposed to be an actual path or a reference based on your setup? I think we should keep setup/directory specific paths out of these files.

build_tools/cmake/build_and_test_asan.sh
```diff
@@ -48,17 +48,9 @@ CMAKE_ARGS=(
   "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
```
qedawkins, line 48: Why are we dropping the groupings here and in subsequent tests?

tools/utils/CLAUDE.md
```diff
@@ -0,0 +1,64 @@
...
+4. **Always add common flags** - Call `cli.add_common_output_flags(parser)`
+5. **JSON purity** - Never mix `print()` with JSON output; use `console.print_json()`
+
+## Test Command
+
+```bash
+python -m unittest discover -s tools/utils -p "test_*.py" -v
+```
+
+**Expected output**: `Ran 356 tests ... OK`
```
qedawkins, line 19: I've found exact numbers to throw off the bot sometimes, can we do something like this?
```suggestion
**Expected output**: `Ran {{[0-9]+}} tests ... OK`
```

tools/utils/CONTRIBUTING.md
```diff
@@ -0,0 +1,265 @@
...
+```bash
+# From IREE repo root (primary workflow):
+python -m unittest discover -s tools/utils -p "test_*.py" -v
+
+# From tools/utils directory:
+cd tools/utils
+python -m unittest discover -s . -p "test_*.py" -v
+```
+
+**Never declare "tests pass" until seeing: `Ran 356 tests ... OK`**
```
qedawkins, line 106: Same here

tools/utils/CONTRIBUTING.md
qedawkins, line 232: Same comment here. I'm not a fan of the idea of setup specific paths making their way into files like this. They can cause artifacts for people who don't follow it exactly.

tools/utils/lit_tools/CONTRIBUTING.md
```diff
@@ -0,0 +1,336 @@
...
+python3 -m unittest discover -s tools/utils/test/lit_tests -v
+
+# All utils tests
+python3 -m unittest discover -s tools/utils -p "test_*.py" -v
+```
+
+## Tool Implementation Template
+
+```python
+# Copyright 2025 The IREE Authors
```
qedawkins, line 61: nit: Wildcard year in templates (we'll be glad in a few weeks :P)

tools/utils/lit_tools/STYLE_GUIDE.md
```diff
@@ -0,0 +1,967 @@
...
+✅ GOOD - Attribute propagation pass verifies attributes:
+// CHECK: util.global @tensor {{.*}} {noinline}
+// (Verifies attribute was propagated)
+```
+
+**Domain-Specific Examples**:
+- **Stream timepoint passes**: Always capture timepoint SSA values
+- **HAL memory passes**: Always capture buffer descriptors and memory types
+- **SCF canonicalization**: Always capture control flow results and branch targets
+- **Attribute propagation**: Always verify attribute presence/changes
```
qedawkins, line 95: We may want to move some of these examples into dialect specific style guides and keep the examples here using the most ubiquitous stuff (arith/util/func/builtin) that spans many compiler phases.

tools/utils/test/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Test Fixtures
+
+Located in `test/lit_tests/fixtures/`:
+- `simple_test.mlir` - Single test case
+- `split_test.mlir` - Three test cases with `// -----` separators
+- `ten_cases_test.mlir` - Ten test cases for multi-case selection testing
+- Additional specialized fixtures
+
+All fixtures use valid MLIR syntax and can be processed by iree-opt.
```
qedawkins, line 213: This section looks spurious, or at least the bullets?

tools/utils/test/lit_tests/fixtures/README.md
```diff
@@ -0,0 +1,213 @@
...
+## Missing Coverage (Potential Future Fixtures)
+
+**Suggested additions**:
+- `unnamed_cases_test.mlir` - Test cases without CHECK-LABEL (unnamed)
+- `large_file_test.mlir` - File with 10+ test cases (stress test)
+- `empty_case_test.mlir` - Test case with no content between delimiters
+- `nested_labels_test.mlir` - Multiple CHECK-LABELs in one case
+- `no_checks_test.mlir` - Test case with no CHECK patterns at all
+
+These are not currently needed but could be added if new edge cases are discovered.
```
qedawkins, line 213: Same here. I don't think we want a Missing Coverage section checked in like this (or just add the tests now).

qedawkins (approved):
Ok I read through the markdown and left a few comments (and also very quickly learned it's not for me). It's mostly minor comments so approving, but I'm not intending to circumvent the rest of the discussion with the approval.

.envrc.example
krzysz00: I'm also going to suggest that a relative path isn't good practice in an `envrc`

.envrc.example
```diff
@@ -0,0 +1,10 @@
+# direnv configuration for IREE tools/utils
+# Copy to .envrc and run `direnv allow`
+#
+# See: https://direnv.net/
+
+# Add tools to PATH
+PATH_add tools/utils/bin
```
krzysz00, line 7:
```suggestion
PATH_add "$PWD/tools/utils/bin"
```

tools/utils/CONTRIBUTING.md
```diff
@@ -0,0 +1,265 @@
...
+## Authoring Guidelines (Required)
+
+- Always add common output flags with `cli.add_common_output_flags(parser)`.
+- Use `console.error/warn/note/success` for all human messages; never `print()` directly.
+- Use `console.print_json(...)` for JSON; do not mix JSON and human text on stdout.
+- Honor `--quiet`: suppress non-essential text automatically via `console.*`.
+- Return `exit_codes.SUCCESS/ERROR/NOT_FOUND` instead of magic numbers.
+- Write files via `fs.safe_write_text(...)` (UTF-8, atomic rename, normalized newlines).
+- Shebang policy: only `bin/` wrappers have shebangs.
+- **CRITICAL**: Never disable lint checks (ruff, black) with `# noqa` without explicit approval.
```
krzysz00, line 38: As a human - approval from who? Where? When (during review, before?) Under what circumstances is an override appropriate?

build_tools/cmake/build_and_test_asan.sh
```diff
@@ -116,16 +108,12 @@ label_exclude_regex="($(IFS="|" ; echo "${label_exclude_args[*]?}"))"
 
 pushd "${BUILD_DIR?}"
 
-echo "::group::Running main project ctests"
```
krzysz00, line 119: Isn't this used for making the logs pretty or something?

tools/utils/ci/PATTERN_GUIDE.md
```diff
@@ -0,0 +1,668 @@
...
+
+✅ **Good**: Matches all variants
+```yaml
+rocm_error:
+  pattern: 'ROCm error|rocm.*error'
+```
+
+### 5. **Escape Special Regex Characters**
+
+Special characters need escaping:
```
krzysz00, line 126: ... This could be much shorted to "[fields] are a regular expression in Python syntax, you need to escape the usual special characters (`.*+?()[]{}`)"

tools/utils/ci/README.md
```diff
@@ -0,0 +1,150 @@
...
+gh auth status
+
+# Login if needed
+gh auth login
+```
+
+## Environment Variables
+
+- `IREE_CI_CORPUS_DIR` - CI garden corpus location (default: `~/.iree-ci-corpus`)
+- `GH_TOKEN` - GitHub authentication token (usually managed by `gh auth`)
```
krzysz00, line 138: I don't think this is needed?

krzysz00 (commented):
Minor comments on the Markdown

.claude/README.md
ScottTodd, line 1: It may be worth adding a new page to https://iree.dev/developers/ for how to use these utilities too

build_tools/cmake/build_and_test_asan.sh
ScottTodd: It was yeah, though it sorta broke scrolling in unexpected ways.

The changes here could also be moved to their own PR.

tools/utils/ci/PATTERN_GUIDE.md
```diff
@@ -0,0 +1,668 @@
...
+error_pattern:
+  pattern: 'error'  # Matches "error", "errors", "error-free", etc.
+```
+
+✅ **Good**: Matches specific error signatures
+```yaml
+compile_error:
+  pattern: 'error: .*|compilation terminated|fatal error:'
+  # Matches C++ compiler errors specifically
+```
```
ScottTodd, line 81: nit: I also had Claude generate some style guide patterns and we didn't really like starting with "Bad" and following with "Good". We settled on starting with "Preferred" and following with "Avoid", like so: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guide.md#pin-action-uses-versions-to-commit-shas

I also see that `tools/utils/lit_tools/STYLE_GUIDE.md` starts with "Good" and follows with "Bad" (inverted order from this guide)

tools/utils/common/cli.py
```diff
@@ -0,0 +1,190 @@
...
+        action="store_true",
+        default=sys.stderr.isatty(),
+        help="Human-friendly formatting (default: auto-detect terminal)",
+    )
+    parser.add_argument(
+        "--no-pretty",
+        action="store_false",
+        dest="pretty",
+        help="Disable colored output",
+    )
```
ScottTodd, line 48: This could probably use https://docs.python.org/3/library/argparse.html#argparse.BooleanOptionalAction instead of defining two separate arguments for `--pretty` and `--no-pretty` (which might conflict if both set)

ScottTodd (commented):
Scanned through and left a few comments. I'm generally supportive of this if there is enough shared value, especially since the utils are in their own subdirectory. I'll let others who have been more active in the repository lately do a more detailed review if they want.

PR #22734, opened 2025-11-23
[Runtime][HIP] Correct O(log n) bound search logic and eliminate O(n) loop(#22733)

### LINKED ISSUE #22733: [Runtime][Hip] Inefficient O(n) and Potentially Incorrect Implementation of lower_bound and upper_bound in `runtime/src/iree/hal/drivers/hip/util/tree.c`
### What happened?

The implementations of `iree_hal_hip_util_tree_lower_bound` and `iree_hal_hip_util_tree_upper_bound` in `runtime/src/iree/hal/drivers/hip/util/tree.c` have issues with correctness and performance. I think these need to be fixed.

#### Problem 1: `upper_bound` Returns Incorrect Results

The current implementation violates the definition of `upper_bound` : [L518-539](https://github.com/iree-org/iree/blob/main/runtime/src/iree/hal/drivers/hip/util/tree.c#L518-L539)
```c
iree_hal_hip_util_tree_node_t* iree_hal_hip_util_tree_upper_bound(
    const iree_hal_hip_util_tree_t* tree, iree_host_size_t key) {
    // ...
    if (key == node->key) {
      return node; // this should be return NEXT node, not current node
    // ...
```

we can check definition in `tree.h` : [L131](https://github.com/iree-org/iree/blob/main/runtime/src/iree/hal/drivers/hip/util/tree.h#L131-L133)
```c
// Returns the first node in the tree that has a key that is > |key| or NULL;
iree_hal_hip_util_tree_node_t* iree_hal_hip_util_tree_upper_bound(
    const iree_hal_hip_util_tree_t* tree, iree_host_size_t key);
```

Expected : `upper_bound(key)` should return the first element **strictly greater than** `key`.
Actual : When `key` exists in the tree, it returns that node instead of the next one.

For example, 
when tree has `{10, 20, 30}`.
lower_bound(20) should be 20.
upper_bound(20) should be 30. (but current returns  20)

####  Problem 2: `upper_bound` Has O(n) Worst-Case 
After the initial BST traversal, `upper_bound` uses a while loop:
```c 
while (last && last->key <= key) {
  last = iree_hal_hip_util_tree_node_next(last);
}
```
**Worst-case scenario**:
```
Tree: {1, 2, 3, ..., 1000}
Query: upper_bound(999)

If initial search ends at node with key=1:
- Iteration 1: next(1) → 2, continue while 2 ≤ 999
- Iteration 2: next(2) → 3, continue while 3 ≤ 999
- ...
- Iteration 998: next(998) → 999, continue while 999 ≤ 999
- Iteration 999: next(999) → 1000, exit

Total: ~999 iterations × O(log n) per next() = O(n log n)
```
This defeats the purpose of using a balanced tree structure.

#### Unnecessary `next()` Calls in Both Functions (refactor)
 Both functions rely on `iree_hal_hip_util_tree_node_next()` calls after BST traversal, instead of tracking the answer **during** traversal.
```c
// lower_bound (O(log n) but still unnecessary)
if (!last || last->key > key) {
  return last;
}
return iree_hal_hip_util_tree_node_next(last);  // we can remove that.

// upper_bound - O(n)
while (last && last->key <= key) {
  last = iree_hal_hip_util_tree_node_next(last);  // related problem 2
}
```

In my opinion, we can eliminate all `next()` calls by tracking the answer during traversal.

_No response_

### Version information

b98c1b92cb630bd696992f47df591bb2f247a8d7

### Additional context

_No response_

swote-git:
Refactors `iree_hal_hip_util_tree_lower_bound` and `iree_hal_hip_util_tree_upper_bound` to use the standard O(log n) bound search algorithm.

This resolves two issues:
1. **Misbehavior:** Ensures `upper_bound` adheres to the strict 'greater than' definition.
2. **Performance:** Eliminates the previous O(n) post-traversal loop which relied on repeated, inefficient calls to `iree_hal_hip_util_tree_node_next`.

fixes : #22733

### DIFF
--- a/runtime/src/iree/hal/drivers/hip/util/tree.c
+++ b/runtime/src/iree/hal/drivers/hip/util/tree.c
@@ -498,44 +498,31 @@ iree_hal_hip_util_tree_node_t* iree_hal_hip_util_tree_get(
 iree_hal_hip_util_tree_node_t* iree_hal_hip_util_tree_lower_bound(
     const iree_hal_hip_util_tree_t* tree, iree_host_size_t key) {
   iree_hal_hip_util_tree_node_t* node = tree->root;
-  iree_hal_hip_util_tree_node_t* last = NULL;
+  iree_hal_hip_util_tree_node_t* result = NULL;
   while (node->is_sentinel == false) {
-    last = node;
-    if (key == node->key) {
-      return node;
-    } else if (key < node->key) {
+    if (key <= node->key) {
+      result = node;
       node = node->left;
     } else {
       node = node->right;
     }
   }
-  if (!last || last->key > key) {
-    return last;
-  }
-  return iree_hal_hip_util_tree_node_next(last);
+  return result;
 }
 
 iree_hal_hip_util_tree_node_t* iree_hal_hip_util_tree_upper_bound(
     const iree_hal_hip_util_tree_t* tree, iree_host_size_t key) {
   iree_hal_hip_util_tree_node_t* node = tree->root;
-  iree_hal_hip_util_tree_node_t* last = NULL;
+  iree_hal_hip_util_tree_node_t* result = NULL;
   while (node->is_sentinel == false) {
-    last = node;
-    if (key == node->key) {
-      return node;
-    } else if (key < node->key) {
+    if (key < node->key) {
+      result = node;
       node = node->left;
     } else {
       node = node->right;
     }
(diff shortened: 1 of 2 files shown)


AWoloszyn (approved):

AWoloszyn:
Thanks!

PR #22737, opened 2025-11-24
[GPU] Fix alignment check for scaled matmul

### LINKED ISSUE #22559: llama 405b fp4 prefill with direct codegen hangs during execution
Following the steps here to compile 405b with direct codegen causes the benchmark to hang.
https://github.com/nod-ai/iree-model-benchmark/tree/main/llama3

This issue only occurs with direct codegen, and compiling with the asm kernel works fine.

Need to triage and identify which commit introduced the regression, as this setup used to work correctly.

CC: @MaheshRavishankar

Yu-Zhewen:
## Problem

The current alignment check in `GPUHeuristics.cpp` is incorrect for any intrinsic that has multiple M, N, and K dimensions. The root cause is that the product of intrinsic sizes is passed to `GPUMMASchedule` instead of passing the individual dimension sizes as a vector.

## Example

https://github.com/iree-org/iree/blob/b98c1b92cb630bd696992f47df591bb2f247a8d7/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp#L516-L525

Consider the scaled MFMA where `intrinsic.kSizes = [K, KB] = [4, 32]`. Instead of passing the vector `[4, 32]`, the value `128` (product: 4 × 32) is passed to `GPUMMASchedule`.

https://github.com/iree-org/iree/blob/b98c1b92cb630bd696992f47df591bb2f247a8d7/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp#L98-L108

Assume tile size = `[4, 1]`. The returned schedule sizes become `[4, 128]` instead of the correct `[16, 32]`. As a result, the last dimension `128` always makes the alignment check fail, since the problem size of KB is `32` and `32 % 128 != 0`.

When the alignment check fails, no intrinsic is selected and the operation falls back to complete serialization. This leads to extremely slow execution for workloads like Llama 405B FP4 prefill with direct codegen.

## Solution

This PR passes all intrinsic sizes as vectors to `GPUMMASchedule`.

## Performance

**Llama 405B FP4 prefill direct codegen with shark-ai:**
- Before: 11 minutes
- After: 234 ms

Closes: #22559

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -175,7 +175,7 @@ static FailureOr<GPUMMASchedule> fitScheduleInSharedMemory(
            << schedule << "\nShrinking schedule...";
 
     auto decrementIfPossible =
-        [](SmallVectorImpl<int64_t> &sizes) -> LogicalResult {
+        [](MutableArrayRef<int64_t> sizes) -> LogicalResult {
       for (int64_t &size : sizes) {
         if (size <= 1)
           continue;
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -38,9 +38,9 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
 llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
                               const GPUMMASchedule &schedule) {
   os << "mmaKind " << schedule.mmaKind << ", ";
-  os << "mSizes: " << schedule.mSize << ", ";
-  os << "nSizes: " << schedule.nSize << ", ";
-  os << "kSizes: " << schedule.kSize << ", ";
+  os << "mSizes: " << schedule.mSizes << ", ";
+  os << "nSizes: " << schedule.nSizes << ", ";
+  os << "kSizes: " << schedule.kSizes << ", ";
   os << "mTileSizes: " << schedule.mTileSizes << ", ";
   os << "nTileSizes: " << schedule.nTileSizes << ", ";
   os << "kTileSizes: " << schedule.kTileSizes << ", ";
@@ -52,11 +52,14 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
 static int64_t calculateOperandsSharedMemoryUsedInBytes(
     const GPUMMASchedule &schedule, int64_t lhsBitwidth, int64_t rhsBitwidth,
     int64_t numRhs = 1) {
-  int64_t tileM = schedule.mSize * llvm::product_of(schedule.mTileSizes) *
+  int64_t tileM = llvm::product_of(schedule.mSizes) *
+                  llvm::product_of(schedule.mTileSizes) *
                   llvm::product_of(schedule.mSubgroupCounts);
-  int64_t tileN = schedule.nSize * llvm::product_of(schedule.nTileSizes) *
+  int64_t tileN = llvm::product_of(schedule.nSizes) *
+                  llvm::product_of(schedule.nTileSizes) *
                   llvm::product_of(schedule.nSubgroupCounts);
-  int64_t tileK = schedule.kSize * llvm::product_of(schedule.kTileSizes);
+  int64_t tileK =
+      llvm::product_of(schedule.kSizes) * llvm::product_of(schedule.kTileSizes);
   return (tileM * tileK * lhsBitwidth + numRhs * tileN * tileK * rhsBitwidth) /
          8;
 }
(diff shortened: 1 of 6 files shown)


Yu-Zhewen:
I also checked this PR has no impact on the perplexity of direct codegen version (mean_perplexity: 5.345724). 

I did not include the performance number for the ASM kernel in the PR description (356.886 ms). I found that the ASM version currently produces very poor perplexity results (mean perplexity: 787626.703125), so its timing may be meaningless. I will raise a new issue on this (https://github.com/nod-ai/amd-shark-ai/issues/2691)

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -68,12 +68,10 @@ struct GPUMMAHeuristicSeeds {
...
-  // Native MMA intrinsic size along M dimension for a subgroup.
-  int64_t mSize = 0;
-  // Native MMA intrinsic size along N dimension for a subgroup.
-  int64_t nSize = 0;
-  // Native MMA intrinsic size along K dimension for a subgroup.
-  int64_t kSize = 0;
+  // Native MMA intrinsic sizes along M, N and K dimensions for a subgroup.
+  SmallVector<int64_t> mSizes;
+  SmallVector<int64_t> nSizes;
+  SmallVector<int64_t> kSizes;
```
kuhar: Should we set the small size to something realistic like 2? I'm worried that this struct will end up consuming a lot of stack space with all the small vectors.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -761,7 +761,7 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     // Multiply by the intrinsic shape for the inner most dim as we distribute
     // to workgroups before packing to intrinsic.
     if (i == mDims.size() - 1) {
-      workgroupTileSizes[mDim] *= schedule->mSize;
+      workgroupTileSizes[mDim] *= llvm::product_of(schedule->mSizes);
```
kuhar: What do you think about defining methods to query the products? For example `schedule->getTotalMSize()`, and similar for other vectors.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -368,12 +368,15 @@ setConvolutionVectorDistributionConfig(IREE::GPU::TargetAttr target,
...
+  workgroupTileSizes[mDim] = schedule->mSubgroupCounts[0] *
+                             schedule->mTileSizes[0] *
+                             llvm::product_of(schedule->mSizes);
+  workgroupTileSizes[nDim] = schedule->nSubgroupCounts[0] *
+                             schedule->nTileSizes[0] *
+                             llvm::product_of(schedule->nSizes);
 
-  reductionTileSizes[kDim] = schedule->kTileSizes[0] * schedule->kSize;
+  reductionTileSizes[kDim] =
+      schedule->kTileSizes[0] * llvm::product_of(schedule->kSizes);
```
Muzammiluddin-Syed-ECE: Perhaps we should add assert statements in these other configs that don't expect the length of K tile sizes to be greater than 1.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
Yu-Zhewen: Yeah, make sense. Changed to `SmallVector<int64_t, 2>` for these variables and also those below.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Yu-Zhewen: Done

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
Yu-Zhewen: Assertion added and I also refresh the Todos a bit

https://github.com/iree-org/iree/blob/63d62415ae1de8264a65c9cc854db4d5bb41f63d/compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp#L286-L290

Yu-Zhewen:
> I also checked this PR has no impact on the perplexity of direct codegen version (mean_perplexity: 5.345724).
> 
> I did not include the performance number for the ASM kernel in the PR description (356.886 ms). I found that the ASM version currently produces very poor perplexity results (mean perplexity: 787626.703125), so its timing may be meaningless. I will raise a new issue on this ([nod-ai/amd-shark-ai#2691](https://github.com/nod-ai/amd-shark-ai/issues/2691))

The ASM perplexity is actually fine. I forgot to use the preshuffled weights when running it....

CC: @bjacob

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -65,50 +65,48 @@ static int64_t
...
+                               : tileCount[i];
+        });
+    // Multiply by intrinsic sizes, applying to the inner dimensions, as
+    // the outer dimensions are unrolling factors. For example, if tileCount
+    // = [a, b, c, d] and intrinsicSizes = [x, y], the result is [a, b, c*x,
+    // d*y].
+    assert(intrinsicSizes.size() <= sizes.size() &&
+           "intrinsic sizes should not exceed tile count sizes");
+    for (auto &&[intrinsicSize, size] :
+         llvm::zip(llvm::reverse(intrinsicSizes), llvm::reverse(sizes))) {
```
kuhar: Use zip_equals and drop the rvalue ref -- it doesn't help anything

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -178,7 +175,7 @@ static FailureOr<GPUMMASchedule> fitScheduleInSharedMemory(
            << schedule << "\nShrinking schedule...";
 
     auto decrementIfPossible =
-        [](SmallVector<int64_t> &sizes) -> LogicalResult {
+        [](SmallVector<int64_t, 2> &sizes) -> LogicalResult {
```
kuhar: ArrayRef

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -104,10 +105,36 @@ struct GPUMMASchedule {
...
+  int64_t getTotalNSubgroupCount() const {
+    return llvm::product_of(nSubgroupCounts);
+  }
+
+  // Check if all schedule dimensions are single-element.
+  bool hasSingleDimensions() const {
+    return mSubgroupCounts.size() == 1 && nSubgroupCounts.size() == 1 &&
+           mTileSizes.size() == 1 && nTileSizes.size() == 1 &&
+           kTileSizes.size() == 1 && mSizes.size() == 1 && nSizes.size() == 1 &&
+           kSizes.size() == 1;
```
kuhar: You can do `llvm::all_equal(size_t(1), mSubgroupCounts.size(), ...})`

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -52,11 +52,11 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
...
-                  llvm::product_of(schedule.mSubgroupCounts);
-  int64_t tileN = schedule.nSize * llvm::product_of(schedule.nTileSizes) *
-                  llvm::product_of(schedule.nSubgroupCounts);
-  int64_t tileK = schedule.kSize * llvm::product_of(schedule.kTileSizes);
+  int64_t tileM = schedule.getTotalMSize() * schedule.getTotalMTileSize() *
+                  schedule.getTotalMSubgroupCount();
+  int64_t tileN = schedule.getTotalNSize() * schedule.getTotalNTileSize() *
+                  schedule.getTotalNSubgroupCount();
+  int64_t tileK = schedule.getTotalKSize() * schedule.getTotalKTileSize();
   return (tileM * tileK * lhsBitwidth + numRhs * tileN * tileK * rhsBitwidth) /
```
Muzammiluddin-Syed-ECE, line 60: should we add the memory used for scales as well here?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Yu-Zhewen: The length of `intrinsicSizes` can actually be smaller than `sizes`, as explained in the comment and assertion above.


```
    // Multiply by intrinsic sizes, applying to the inner dimensions, as
    // the outer dimensions are unrolling factors. For example, if tileCount
    // = [a, b, c, d] and intrinsicSizes = [x, y], the result is [a, b, c*x,
    // d*y].
    assert(intrinsicSizes.size() <= sizes.size() &&
           "intrinsic sizes should not exceed tile count sizes");
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
Yu-Zhewen: Thanks, done

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Yu-Zhewen: I changed to `SmallVectorImpl<int64_t> &` instead of `ArrayRef` because the function modifies `sizes`

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Yu-Zhewen: Yeah, I think we're underestimating the usage here. I prefer adding a TODO and addressing this in a follow-up PR, as we need to know data types for scales to accurately calculate shared memory usage.

This will require updating the `GPUMatmulShapeType` structure (note: this PR fixes `GPUMMASchedule` instead):

https://github.com/iree-org/iree/blob/4f354bd7ef80cb5d12d9d3fab3158c9c0bd023f0/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h#L16-L24

We'd need to add scale type information, either by:
- Adding explicit `aScaleType` and `bScaleType`, or
- Making `aType`/`bType` arrays

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
kuhar: If it doesn't append we can do `MutableArrayRef`

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Muzammiluddin-Syed-ECE: I see, a broader scope rework might be helpful here and in other places in the code. A follow up PR makes sense, I'll make an issue.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Yu-Zhewen: I see, thank!

kuhar (approved):

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Muzammiluddin-Syed-ECE: https://github.com/iree-org/iree/issues/22785

PR #22740, opened 2025-11-24
[Codegen] Drop certain encoding attributes after type propagation

sommerlukas:
Drop encoding attributes that do not need further materialization during Codegen after type propagation has used the information.

Encodings can report eligible for being dropped by implementing a new interface method on `SerializableAttr`, with the default being that attributes are not dropped.

Dropping attributes happens as a second phase in the TypePropagationPass due to the fixpoint iteration.


compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
hanhanW, line 1: It seems to be too specific to packed_storage. I think you can just do hard check (i.e., packed_storage) in the pass; we add the interface methods when needed.

I also have concerns about the layering, because the encoding is attached to the type and the implementation is passing the type to the interface method.

compiler/src/iree/compiler/Codegen/Common/test/type_propagation_packed_storage.mlir
hanhanW, line 1: I think we just add tests to `type_propagation.mlir`.

I'm not sure why we have `type_propagation_packing.mlir`, but it should be collapsed into the main one, IMO.

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -545,39 +637,30 @@ struct TypePropagationPass final
...
+    DropEncodingTypeConverter dropConverter;
+    RewritePatternSet dropPatterns(context);
+    // We need less specific patterns here because we simply drop attributes
+    // from result and operand types.
+    dropPatterns
+        .insert<LegalizeBasicBlocks<func::FuncOp>, LegalizeResultElementType,
+                NamedOpTypePropagation<linalg::BatchMatmulOp>,
+                NamedOpTypePropagation<linalg::MatmulOp>,
+                NamedOpTypePropagation<linalg::MatvecOp>,
+                NamedOpTypePropagation<linalg::DotOp>>(dropConverter, context);
```
hanhanW, line 647: We don't use named op here. The plan is generailizing them into linalg.generic. I think we should try `linalg::LinalgOp` and maybe `LinalgExtOp` instead.

In this context, you can use OpInterfaceConversionPattern like:

https://github.com/iree-org/iree/blob/dd5f28a3fe285a63465ad0872922bc410de2af56/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp#L642-L667

And you can check if it is a linalg::GenericOp.

If so, we create the new generic op using new operands.

If not, we can create the op in a generic way, like [this](https://github.com/llvm/llvm-project/blob/6b58449b2c813588f9c251d4f0963762d6746881/mlir/lib/Dialect/Vector/Transforms/VectorTransforms.cpp#L463-L465):

```
        rewriter.create(op->getLoc(), op->getName().getIdentifier(),
                        bcastOp.getSource(), castResTy, op->getAttrs());
```

hanhanW (changes requested):

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
sommerlukas: I'm a bit surprised that we're not supposed to use named ops here. 

In the same file, as this, we insert patterns for named ops: https://github.com/iree-org/iree/blob/dd5f28a3fe285a63465ad0872922bc410de2af56/compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp#L535-L545

Should there be no patterns for named ops in this pass?

My rationale for this approach was to reuse the existing patterns with a different type converter instead of implementing dedicated patterns.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
sommerlukas: My understanding was that one goal of #21185 was to reduce special casing for `packed_storage` as much as possible, so I'm hesitant to introduce new special casing. That was my rationale to introduce the new interface method instead.

My rationale to add the type as parameter of the interface method was that some (future) encoding might want to make the decision on whether dropping is possible based some property of the type, e.g,. the tensor element type. 

I can remove that parameter from the interface function if you prefer.

hanhanW: I think you are making good progress for not having special casting already, because it removes some burdens and decouple deps. For this case, we can start with special check. There is a layering issue that I'm not sure how to fix atm. I just feel that it may belong to other interface, but not attribute interface. I need to think more about it.

Removing the parameter from the interface function does not address my concern as this is too specific to this attribute. So I suggest to just do the special check for now; we can come back to this topic when we need abstraction.

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
hanhanW: This pass is very old, and we are moving to generic op approach today. After thinking more, I think we can start with what you have here, since may run into rabbit holes if we try to move to my suggestion in one-step.

sommerlukas: Ok, I see. I agree that it makes sense to split updating this pass into a future PR.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
sommerlukas: Ok, I've added a special check and removed the new interface method. I've made a note on #21185 to remove the special case check in the future.

compiler/src/iree/compiler/Codegen/Common/test/type_propagation_packed_storage.mlir
sommerlukas: I've merged all three files into `type_propagation.mlir`.

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -73,6 +91,18 @@ static std::optional<Type> getLegalizedType(Type t) {
   return std::nullopt;
 }
 
+static std::optional<Type> dropEncodingWithoutMaterialization(Type t) {
```
hanhanW: nit: drop `optional`?

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -94,11 +124,33 @@ struct TypePropagationTypeConverter : public TypeConverter {
...
+struct DropEncodingTypeConverter : public TypeConverter {
+  DropEncodingTypeConverter() {
+    addConversion([](Type t) -> std::optional<Type> {
+      if (auto dispatchTensor =
+              dyn_cast<IREE::TensorExt::DispatchTensorType>(t)) {
+        Type oldBoundTy = dispatchTensor.getBoundType();
+        std::optional<Type> newBoundTy =
+            dropEncodingWithoutMaterialization(oldBoundTy);
+        if (!newBoundTy) {
+          return std::nullopt;
```
hanhanW: I think we can return `nullptr` instead. `optional` does not provide values to me in this case.

hanhanW (commented):
overall, lgtm.

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
sommerlukas: I've dropped it, we can bring it back if we need more fine-grained control in the future.

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
sommerlukas: I used `std::optional` here based on the documentation for the MLIR `TypeConverter`:
https://github.com/llvm/llvm-project/blob/b225907804af0264a60ca0d7458afa60723fcd71/mlir/include/mlir/Transforms/DialectConversion.h#L145-L153

But as I dropped the `optional` from `dropEncodingWithoutMaterialization`, I changed the return to `Type` now.

compiler/src/iree/compiler/Codegen/Common/TypePropagationPass.cpp
```diff
@@ -94,11 +124,30 @@ struct TypePropagationTypeConverter : public TypeConverter {
...
+  DropEncodingTypeConverter() {
+    addConversion([](Type t) -> Type {
+      if (auto dispatchTensor =
+              dyn_cast<IREE::TensorExt::DispatchTensorType>(t)) {
+        Type oldBoundTy = dispatchTensor.getBoundType();
+        std::optional<Type> newBoundTy =
+            dropEncodingWithoutMaterialization(oldBoundTy);
+        if (*newBoundTy == oldBoundTy) {
+          return t;
+        }
```
hanhanW: We don't need `optional` here, right?

sommerlukas: Yeah, missed it in the last commit, thanks for catching this!

hanhanW (approved):
Thanks, LGTM!

PR #22745, opened 2025-11-25
[Codegen] Test Cleanup 2/8: Common GPU tests

qedawkins:
Result of a scan over all tests in Codegen to cleanup common issues in tests. A summary of the results + a preamble approximating the issues to look for can be found here:

https://gist.github.com/qedawkins/40f9e604fd83745bf1ac20fd63a7a61f

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_pack_to_instrinsics.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_pack_to_instrinsics.mlir
@@ -95,46 +95,7 @@ module {
 //  CHECK-SAME:       affine_map<(d0, d1, d2, d3) -> (d1, d2)>
 //  CHECK-SAME:       affine_map<(d0, d1, d2, d3) -> (d0, d1)>
 //  CHECK-SAME:     lowering_config = #iree_gpu.lowering_config<{mma_kind = #iree_gpu.scaled_mma_layout<intrinsic = MFMA_SCALE_F32_16x16x128_B32, lhs_elem_type = f4E2M1FN, rhs_elem_type = f4E2M1FN, acc_elem_type = f32>}>
-// CHECK-SAME:      permutations = [array<i64: 0, 1, 2>, array<i64: 2, 0, 1>, array<i64: 0, 1>, array<i64: 1, 0>, array<i64: 0, 1>]
-//  CHECK-SAME:     : tensor<?x?x?x16x4x32xf4E2M1FN>, tensor<?x?x?x16x4x32xf4E2M1FN>, tensor<?x?x16x4xf8E8M0FNU>, tensor<?x?x16x4xf8E8M0FNU> into tensor<?x?x16x16xf32>
-
-// -----
-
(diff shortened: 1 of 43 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_pack_to_instrinsics.mlir
```diff
@@ -95,46 +95,7 @@ module {
...
-
-// -----
-
-#map = affine_map<(d0, d1, d2, d3) -> (d0, d2, d3)>
-#map1 = affine_map<(d0, d1, d2, d3) -> (d1, d2, d3)>
-#map2 = affine_map<(d0, d1, d2, d3) -> (d0, d2)>
-#map3 = affine_map<(d0, d1, d2, d3) -> (d1, d2)>
-#map4 = affine_map<(d0, d1, d2, d3) -> (d0, d1)>
-module {
-  func.func @scaled_mfma_16x16x128(%a: tensor<?x?x?xf4E2M1FN>, %b: tensor<?x?x?xf4E2M1FN>, %a_scales: tensor<?x?xf8E8M0FNU>, %b_scales: tensor<?x?xf8E8M0FNU>, %c: tensor<?x?xf32>) -> tensor<?x?xf32> {
```
kuhar, line 109: Why is this deleted?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_tile.mlir
```diff
@@ -54,8 +54,8 @@ func.func @innermost_reduction() {
 //      CHECK:     linalg.fill
 //      CHECK:     scf.for %{{.+}} = %[[C0]] to %[[C384]] step %[[C4]]
 //      CHECK:       linalg.generic
-// CHECK-SAME:         ins(%{{.+}}, %{{.+}} : tensor<4x4xf32>, tensor<4xf32>)
```
kuhar, line 57: What's wrong with this format?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_pack_to_instrinsics.mlir
qedawkins: This is an exact duplicate of the above test. The minor indentation change on the CHECK-SAME makes it look like it's not.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_tile.mlir
qedawkins: I shoot, I thought I removed these, good catch. Yeah claude went on an acid trip here and invented a new variant of wildcard matcher for some reason.

qedawkins: Ok I see what happened. Because the input and output tensor types are the same, the greedy wildcard match jumped all the way past the output. I changed it to match alphanumeric instead.

kuhar (approved):

PR #22755, opened 2025-11-25
build_tools: fix: ensure that iree-flatcc-cli and iree-c-embed-data are build for the target during cross-compilation

Manewing:
When running a cross-compilation of the runtime I noticed that the install includes binaries from the host binary directory that I specified, specifically: `iree-c-embed-data`. Looking at `third-party/iree/build_tools/embed_data/CMakeLists.txt:8`  you can see that the binary is only imported but never built for the cross-target.

When trying to also cross-compile the python bindings the cmake setup was failing with:
```
CMake Error at third-party/iree/runtime/bindings/python/CMakeLists.txt:334:EVAL:2 (install):
  install TARGETS given target "iree-c-embed-data" which does not exist.
Call Stack (most recent call first):
  third-party/iree/CMakeLists.txt:DEFERRED
```

This happens because the binary is not being build for the cross-target. The same issue happens for the `iree-flatcc-cli` tool.

This PR ensures that the host binaries are correctly used during the cross-compilation, as well as ensuring that the binaries are built for the target and included in the install.

### DIFF
--- a/build_tools/cmake/flatbuffer_c_library.cmake
+++ b/build_tools/cmake/flatbuffer_c_library.cmake
@@ -85,11 +85,17 @@ function(flatbuffer_c_library)
   endforeach()
   list(TRANSFORM _OUTS PREPEND "${CMAKE_CURRENT_BINARY_DIR}/")
 
+  # When cross-compiling execute the binary from the host build.
+  set(IREE_FLATCC_TOOL_BINARY iree-flatcc-cli)
+  if (IREE_HOST_BIN_DIR)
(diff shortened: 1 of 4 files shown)


ScottTodd (approved):
This cross-compile code has been a source of whack-a-mole issues for a while. If this solves issues for cases you care about then great.

I'm a bit skeptical here about building these tools but then not using them. Are the built tools functional?

---

I've also asked [over here on Discord](https://discord.com/channels/689900678990135345/690274711523164166/1443360489211170938) if we can have someone else who has more recently contributed to this code review this PR.

Manewing:
Hi @ScottTodd, thanks for your review. Both of the tools are included in the runtime install of the cross-build and I checked that they are working.

PR #22756, opened 2025-11-25
[Encoding] Test i1 mask attention with packed_storage

sommerlukas:
Test attention with an `i1` mask using the `packed_storage` encoding attribute instead of the experimental flag to avoid extending `i1` to `i8` during compilation.

Currently the test needs to drop the unit extent dimension compared to the other attention tests due to problems with FoldUnitExtentDims.

### DIFF
--- a/tests/e2e/linalg_ext_ops/BUILD.bazel
+++ b/tests/e2e/linalg_ext_ops/BUILD.bazel
@@ -81,6 +81,7 @@ VMVX_SRCS = enforce_glob(
     exclude = [
         "attention.mlir",
         "attention_i1_mask.mlir",
+        "attention_i1_mask_encoding.mlir",
     ],
 )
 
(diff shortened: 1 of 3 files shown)


sommerlukas:
The content of the new test is identical to [`attention_i1_mask.mlir`](https://github.com/iree-org/iree/blob/dc2c060f85066433db924c9d6d5267b3b8f81352/tests/e2e/linalg_ext_ops/attention_i1_mask.mlir), with two differences:
- The new test uses the `packed_storage` encoding instead of the experimental flag (intent of this PR)
- The new test drops all unit extent dimensions due to an only loosely related issue with `FoldUnitDims`. For this problem, I currently have an upstream PR up to fix it and a corresponding PR in IREE to integrate it (#22783). Once that is landed, I will bring back the unit-extent dimensions to match the existing attention e2e tests more closely. I would still like to merge this test in the meantime, as it will allow the work to proceed, e.g., to remove the experimental flag.

hanhanW (approved):
Nice progress!

PR #22757, opened 2025-11-25
[Encoding] Implement compatibility check for packed_storage

sommerlukas:
Implement the `SerializableAttr` interface method to check compatibility with other encodings for the `packed_storage` encoding.

In particular, `packed_storage` does not affect the shape of the tensor, so it's compatible with no encoding at all and `flow.tensor.bitcast` can be used to attach the encoding to a tensor.


hanhanW (approved):

hanhanW:
let's wait a day to see if Ben wants to comment or not. It does not block the draft PR (i.e., type propagation pass), so I think it is okay to wait.

hanhanW:
The change itself looks good to me, thanks!

hanhanW:
This does not change any stream specifics, and it adds the interface implementation to the attribute. I'm going to land the PR.

PR #22763, opened 2025-11-26
[Codegen][Tuner] Expose the python bindings for LinalgExt::inferScaledContractionDims and LinalgExt::isaScaledContractionOpInterface

Muzammiluddin-Syed-ECE:
Exposes bindings that enable tuner support for scaled contractions.


bangtianliu:
Missing Export Declarations.

you should add new defined functions to Files: compiler/src/iree/compiler/API/api_exports.{def,ld,macos.lst,c}

Without exports, these functions won't be visible when linking against the IREE compiler library. The Python bindings won't work.

compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
```diff
@@ -119,6 +119,24 @@ MLIR_CAPI_EXPORTED bool ireeCodegenHasIGEMMGenericConvDetails(MlirOperation op);
 MLIR_CAPI_EXPORTED ireeCodegenIGEMMGenericConvDetails
 ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op);
 
+typedef struct ireeCodegenScaledContractionDimensions {
+  MlirAttribute batch;
+  MlirAttribute m;
+  MlirAttribute n;
+  MlirAttribute k;
+  MlirAttribute kB;
+} ireeCodegenScaledContractionDimensions;
```
bangtianliu: All other structs in this file use struct without typedef. Consistency matters.

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -336,3 +336,279 @@ def test_igemm_conv_details():
 
     details = iree_codegen.get_igemm_generic_conv_details(matmul_op)
     assert details is None, "IGEMM details should be None for non-conv operation"
+
+
+@run
+def test_isa_scaled_contraction_op():
+    # Test 1: Regular matmul is not a scaled contraction
```
bangtianliu:
```suggestion
    # Test 1: Regular matmul is not a scaled contraction.
```

Add period, apply this all over this PR.

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -320,3 +321,87 @@ ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op) {
...
+
+  mlir::MLIRContext *ctx = maps[0].getContext();
+  auto toAttr = [ctx](llvm::ArrayRef<unsigned> vals) -> MlirAttribute {
+    mlir::Builder b(ctx);
+    llvm::SmallVector<mlir::Attribute, 2> attrs;
+    for (unsigned val : vals) {
+      attrs.push_back(b.getI32IntegerAttr(val));
+    }
+    return wrap(b.getArrayAttr(attrs));
+  };
```
bangtianliu:
```suggestion
  auto toAttr = [ctx](ArrayRef<unsigned> vals) -> MlirAttribute {
    return wrap(DenseI32ArrayAttr::get(ctx, llvm::to_vector_of<int32_t>(vals)));
  };
```

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -320,3 +321,87 @@ ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op) {
...
+    const MlirAffineMap *indexingMaps, size_t numMaps) {
+  ireeCodegenScaledContractionDimensions result{};
+  if (!indexingMaps || numMaps == 0) {
+    return result;
+  }
+
+  llvm::SmallVector<mlir::AffineMap, 3> maps;
+  for (size_t i = 0; i < numMaps; ++i) {
+    maps.push_back(unwrap(indexingMaps[i]));
+  }
```
bangtianliu:
```suggestion
  llvm::SmallVector<mlir::AffineMap, 5> maps = llvm::map_to_vector(
      llvm::ArrayRef(indexingMaps, numMaps),
      [](MlirAffineMap map) { return unwrap(map); });

```

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -320,3 +321,87 @@ ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op) {
...
 }
+
+bool ireeCodegenMlirOperationIsAScaledContractionOp(MlirOperation op) {
+  auto linalgOp = llvm::cast<mlir::linalg::LinalgOp>(unwrap(op));
+  return mlir::iree_compiler::IREE::LinalgExt::isaScaledContractionOpInterface(
+      linalgOp);
+}
+
+ireeCodegenScaledContractionDimensions
+ireeCodegenInferScaledContractionDimensions(MlirOperation op) {
```
bangtianliu, line 332: I’m not sure we really need two Python bindings to infer the same information, it looks a bit redundant.

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
Muzammiluddin-Syed-ECE: I decided to use an ArrayAttr instead of a DenseI32ArrayAttr to avoid adding updating the existing utility in IREECompilerDialectsModule.cpp:
```c++
static std::vector<int64_t> getIntArrayAttrValues(MlirAttribute attr) {
  if (mlirAttributeIsNull(attr) || !mlirAttributeIsAArray(attr))
    return {};

  std::vector<int64_t> result;
  size_t n = mlirArrayAttrGetNumElements(attr);
  result.reserve(n);
  for (size_t i = 0; i < n; ++i) {
    MlirAttribute elem = mlirArrayAttrGetElement(attr, i);
    int64_t val = mlirIntegerAttrGetValueInt(elem);
    result.push_back(val);
  }
  return result;
}
``` 
I can just update this one to support DenseI32ArrayAttr if you'd like (such as by checking mlirAttributeIsADenseI32Array).

Muzammiluddin-Syed-ECE: I think to maintain consistency with the rest of the file I will continue using the ArrayAttr (in the IGEMM we use `wrap(builder.getArrayAttr(dimMapAttrs));`) but I will use `map_to_vector` like you suggest.

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
```diff
@@ -320,3 +321,47 @@ ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op) {
...
+  if (failed(maybeDims)) {
+    return result;
+  }
+
+  const mlir::iree_compiler::IREE::LinalgExt::ScaledContractionDimensions
+      &scaledContractionDims = *maybeDims;
+  mlir::MLIRContext *ctx = linalgOp.getContext();
+
+  auto toAttr = [&ctx](llvm::ArrayRef<unsigned> vals) -> MlirAttribute {
+    mlir::Builder b(ctx);
```
bangtianliu: the Builder should be moved outside the lambda, otherwise new Builder will be created each call

compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
```diff
@@ -119,24 +119,20 @@ MLIR_CAPI_EXPORTED bool ireeCodegenHasIGEMMGenericConvDetails(MlirOperation op);
 MLIR_CAPI_EXPORTED ireeCodegenIGEMMGenericConvDetails
 ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op);
 
-typedef struct ireeCodegenScaledContractionDimensions {
+struct ireeCodegenScaledContractionDimensions {
   MlirAttribute batch;
```
bangtianliu, line 124: Maybe add comment say the actual type, like above example .

bangtianliu: I mean the `MlirAttribute batch;`

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
bangtianliu: You can go ahead to land in the way you want. 
I just want to provide more info here:
ArrayAttr is a general array that can hold any type of attribute, in which each element is a separate Attribute. 
DenseI32ArrayAttr is specialized for I32, since it knows all the types are in I32.

In future,  we can revisit it to clean all of them maybe.

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -336,3 +336,242 @@ def test_igemm_conv_details():
...
+    dims = iree_codegen.infer_scaled_contraction_dimensions(scaled_op)
+    assert dims is not None, "Should successfully infer dimensions"
+
+    # Verify the inferred dimensions match expected values.
+    # For the given indexing maps:
+    # d0 = M (parallel) -> m,
+    # d1 = N (parallel) -> n,
+    # d2 = Ko (reduction) -> k,
+    # d3 = Kb (reduction, block dim) -> kB.
+    assert list(dims.m) == [0], f"Expected m=[0], got {list(dims.m)}"
```
bangtianliu: No need `list(*)` ? you can have a try

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -336,3 +336,242 @@ def test_igemm_conv_details():
...
+    batched_ops = iree_codegen.get_tuner_root_ops(input_module_batched)
+    if len(batched_ops) == 1:
+        batched_op = batched_ops[0]
+        assert iree_codegen.isa_scaled_contraction_op(
+            batched_op
+        ), "Batched scaled matmul should be recognized"
+
+        dims_batched = iree_codegen.infer_scaled_contraction_dimensions(batched_op)
+        if dims_batched is not None:
+            # Expected: batch=[0], m=[1], n=[2], k=[3], kB=[4].
```
bangtianliu: This code comment is redundant, since below code itself reflects this. 

So remove all this kind of code comment.

compiler/bindings/python/test/api/tuner_api_test.py
bangtianliu: Yes, you can remove it since the returned type already is a list

bangtianliu:
```suggestion
    assert dims.m == [0], f"Got {list(dims.m)}"
```
The expected values are already visible in the assertion itself.

So apply this to all the tests.

compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp
Muzammiluddin-Syed-ECE: Sounds good I can put up another PR for it. in the meantime ive created an empty issue just to note it down https://github.com/iree-org/iree/issues/22774.

bangtianliu: Sure, thanks!

compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
```diff
@@ -119,6 +119,25 @@ MLIR_CAPI_EXPORTED bool ireeCodegenHasIGEMMGenericConvDetails(MlirOperation op);
 MLIR_CAPI_EXPORTED ireeCodegenIGEMMGenericConvDetails
 ireeCodegenGetIGEMMGenericConvDetails(MlirOperation op);
 
+struct ireeCodegenScaledContractionDimensions {
+  // Batch dimension for scaled contraction (ArrayAttr)
```
bangtianliu:
```suggestion
  // Batch dimension for scaled contraction (ArrayAttr).
```

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -85,7 +85,7 @@ def attention_op_detail():
     )  # (d0, d3, d2).
```
bangtianliu: Remove this kind of redundant code comment, since the code itself is obvious.

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -336,3 +336,230 @@ def test_igemm_conv_details():
...
+    root_op_list = iree_codegen.get_tuner_root_ops(input_module)
+    assert len(root_op_list) == 1, "Should have exactly one root op"
+    scaled_op = root_op_list[0]
+
+    # Verify it's a scaled contraction first.
+    assert iree_codegen.isa_scaled_contraction_op(
+        scaled_op
+    ), "Operation should be recognized as scaled contraction"
+
+    # Test dimension inference.
```
bangtianliu: Also this, no need

bangtianliu:
Here: https://releases.llvm.org/10.0.0/docs/CodingStandards.html?utm_source=chatgpt.com#commenting

> Comments are important for readability and maintainability. When writing comments, write them as English prose, using proper capitalization, punctuation, etc. Aim to describe what the code is trying to do and why, not how it does it at a micro level. Here are a few important things to document:
 
Not just for you, also for myself lol. Very easy to leave some unnecessary code comment there.

Muzammiluddin-Syed-ECE:
Sorry about the number of grammar mistakes in this PR, I'll be more careful 😅 and thank you for the feedback

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -336,3 +328,226 @@ def test_igemm_conv_details():
...
+                    %sum = arith.addf %acc, %prod : f32
+                    linalg.yield %sum : f32
+                } -> tensor<8x16x16xf32>
+                return %result : tensor<8x16x16xf32>
+            }
+        }
+    """
+    input_module_batched = ir.Module.parse(module_str_batched)
+    batched_ops = iree_codegen.get_tuner_root_ops(input_module_batched)
+    if len(batched_ops) == 1:
```
bangtianliu: Use assert to avoid if branch?

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -336,3 +328,226 @@ def test_igemm_conv_details():
...
+
+    dims = iree_codegen.infer_scaled_contraction_dimensions(scaled_op)
+    assert dims is not None, "Should successfully infer dimensions"
+
+    # Verify the inferred dimensions match expected values.
+    # For the given indexing maps:
+    # d0 = M (parallel) -> m,
+    # d1 = N (parallel) -> n,
+    # d2 = Ko (reduction) -> k,
+    # d3 = Kb (reduction, block dim) -> kB.
```
bangtianliu: Remove, since below code is very easy to understand, this code comment is unnecessary

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -336,3 +328,226 @@ def test_igemm_conv_details():
...
+                return %result : tensor<16x16xf32>
+            }
+        }
+    """
+    input_module = ir.Module.parse(module_str)
+    root_op_list = iree_codegen.get_tuner_root_ops(input_module)
+    assert len(root_op_list) == 1, "Should have exactly one root op"
+    scaled_op = root_op_list[0]
+
+    # Verify it's a scaled contraction first.
```
bangtianliu: Also here.

bangtianliu (approved):

bangtianliu:
Also update the PR tile, since it is misleading (looks like adding new transform matcher operations for scaled matmul). 

Maybe like

> [Codegen][Tuner] Expose the python bindings for LinalgExt::inferScaledContractionDims and  LinalgExt::isaScaledContractionOpInterface

PR #22767, opened 2025-11-26
[Stream] Add UnifyEncodingForGlobals pass. 1/n

hanhanW:
This pass identifies immutable globals that are encoded with multiple different encodings and unifies them to a single encoding. This reduces memory footprint when the same source data (e.g., model weights) is used in multiple dispatch operations with different encoding requirements.

The pass:
- Traces from stream.tensor.encode ops back to source globals
- Requires both source and encoded globals to be immutable
- Uses identity encoding as a conservative fallback
- Updates both stream.tensor.encode and stream.tensor.sizeof consistently

This is the minimal functioning implementation. The pass will be built incrementally with resolver integration and subgraph equivalence support.

Note that it does not work completely as dispatch sites and executables are not updated.

It is a step towards #22485.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.td
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.td
@@ -562,6 +562,42 @@ def SpecializeDispatchesPass :
   ];
 }
 
+def UnifyEncodingForGlobalsPass :
+    Pass<"iree-stream-unify-encoding-for-globals", "mlir::ModuleOp"> {
+  let summary = "Unifies multiple encodings of the same immutable global to reduce memory.";
+  let description = [{
+    When a single source global (e.g., a model weight) is encoded multiple
+    times with different encodings for different uses, this pass selects a
+    unified encoding and updates all references to reduce memory footprint.
+
+    The pass analyzes initializers to identify:
+    1. Source globals (e.g., stream.tensor.constant with parameters)
+    2. Encoded globals derived from the same source
+    3. Dispatch sites using these encoded globals
+
+    When multiple encoded globals share the same source, the pass queries
+    layout resolvers to determine a unified encoding. If a common encoding
+    is found, the pass updates:
+    1. The encoding operations in initializers
+    2. The corresponding stream.tensor.sizeof operations
+    3. The dispatch operand encodings
+    4. The executable binding types
+
+    Constraints:
+    - Only handles immutable globals with !stream.resource<constant> type
+    - Bails out if an encoded global can have multiple sources due to CFG
+    - Requires exactly one AffinityAnalysisDialectInterface implementation
+    - Skips hand-authored executables to avoid incorrect modifications
+
+    This pass should run after CombineInitializersPass and before
+    SpecializeEncodingsPass in the Stream pipeline.
+  }];
+  let dependentDialects = [
+    "IREE::Encoding::IREEEncodingDialect",
+    "IREE::Stream::StreamDialect",
+  ];
+}
+
 def SpecializeEncodingsPass :
     Pass<"iree-stream-specialize-encodings", "mlir::ModuleOp"> {
   let summary = "Specializes serializable encodings based on layout analysis.";
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -1,0 +1,330 @@
+// Copyright 2025 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Dialect/Encoding/IR/EncodingDialect.h"
+#include "iree/compiler/Dialect/Encoding/IR/EncodingTypes.h"
+#include "iree/compiler/Dialect/Stream/IR/StreamOps.h"
+#include "iree/compiler/Dialect/Stream/IR/StreamTypes.h"
+#include "iree/compiler/Dialect/Stream/Transforms/Passes.h"
+#include "iree/compiler/Dialect/Util/Analysis/GlobalTable.h"
+#include "iree/compiler/Dialect/Util/IR/UtilOps.h"
+#include "llvm/ADT/MapVector.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/DebugLog.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/SymbolTable.h"
+#include "mlir/Pass/Pass.h"
+
+namespace mlir::iree_compiler::IREE::Stream {
+
+#define DEBUG_TYPE "iree-stream-unify-encoding-for-globals"
+
+#define GEN_PASS_DEF_UNIFYENCODINGFORGLOBALSPASS
+#include "iree/compiler/Dialect/Stream/Transforms/Passes.h.inc"
+
+namespace {
+
+//===----------------------------------------------------------------------===//
+// Analysis.
+//===----------------------------------------------------------------------===//
+
+// Information about an encoded global and its relationship to its source.
+struct EncodedGlobalInfo {
+  IREE::Util::GlobalOpInterface encodedGlobal;
+  Attribute encodingAttr;
+  IREE::Stream::TensorEncodeOp encodeOp;
+  IREE::Stream::AffinityAttr affinityAttr;
+};
+
+// Information about a source global and all its encoded versions.
+struct SourceGlobalInfo {
+  IREE::Util::GlobalOpInterface sourceGlobal;
+  SmallVector<EncodedGlobalInfo> encodedVersions;
+};
+
+// Finds the global store op that uses the given value, traversing through
+// passthrough ops like clone.
+// TODO(#22485): Handle other passthrough ops (e.g., transfer).
+static IREE::Util::GlobalStoreOpInterface findStoreOp(Value value) {
+  for (auto user : value.getUsers()) {
+    if (auto store = dyn_cast<IREE::Util::GlobalStoreOpInterface>(user)) {
+      return store;
+    }
+    if (auto cloneOp = dyn_cast<IREE::Stream::AsyncCloneOp>(user)) {
+      if (auto store = findStoreOp(cloneOp.getResult())) {
+        return store;
+      }
+    }
+  }
+  return nullptr;
+}
+
+// Analyzes a module to find immutable globals that have multiple encoded
+// versions. Use run() to perform analysis, then query results with
+// getSourcesWithMultipleEncodings() or getSourceGlobals().
+class GlobalEncodingAnalyzer {
+public:
+  explicit GlobalEncodingAnalyzer(ModuleOp moduleOp)
+      : moduleOp(moduleOp), symbolTable(moduleOp) {}
+
+  LogicalResult run();
+
+  // Returns all source globals that have multiple distinct encodings.
+  SmallVector<StringRef> getSourcesWithMultipleEncodings() const {
+    SmallVector<StringRef> result;
+    for (const auto &[name, info] : sourceGlobals) {
+      llvm::SmallDenseSet<Attribute, 4> uniqueEncodings;
+      for (const EncodedGlobalInfo &encoded : info.encodedVersions) {
+        uniqueEncodings.insert(encoded.encodingAttr);
+      }
+      if (uniqueEncodings.size() > 1) {
+        result.push_back(name);
+      }
+    }
+    return result;
+  }
+
+  const llvm::MapVector<StringRef, SourceGlobalInfo> &getSourceGlobals() const {
+    return sourceGlobals;
+  }
+
+private:
+  // Walks all initializers to find encoding patterns and populates
+  // sourceGlobals map. Looks for patterns like:
+  //   %source = util.global.load @source_global
+  //   %encoded = stream.tensor.encode %source ...
+  //   util.global.store %encoded, @encoded_global
+  // Only considers immutable source and encoded globals.
+  LogicalResult collectGlobalEncodings();
+
+  // Traces from encode op's source operand back to a source global.
+  // Returns nullptr if tracing fails or source is mutable.
+  IREE::Util::GlobalOpInterface traceToSourceGlobal(Value value);
+
+  // Returns true if the type has a recognized encoding attribute.
+  bool hasRecognizedEncoding(Type type) const;
+
+  ModuleOp moduleOp;
+  SymbolTable symbolTable;
+
+  // Maps source global name to its info. Populated by run().
+  llvm::MapVector<StringRef, SourceGlobalInfo> sourceGlobals;
+};
+
+LogicalResult GlobalEncodingAnalyzer::run() {
+  LDBG() << "=== GlobalEncodingAnalyzer::run() ===";
+  if (failed(collectGlobalEncodings())) {
+    return failure();
+  }
+  LDBG_OS([&](llvm::raw_ostream &os) {
+    os << "Analysis complete:\n";
+    os << "  Source globals: " << sourceGlobals.size() << "\n";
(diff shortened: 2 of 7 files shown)


hanhanW:
I do my best to write design rationale to https://github.com/iree-org/iree/issues/22485#issue-3572150667


@benvanik @MaheshRavishankar  @sommerlukas can you help review? This is the minimal functioning implementation; I'll complete the pass incrementally in follow-ups.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.td
```diff
@@ -572,6 +572,42 @@ def SpecializeDispatchesPass :
   ];
 }
 
+def UnifyEncodingForGlobalsPass :
+    Pass<"iree-stream-unify-encoding-for-globals", "mlir::ModuleOp"> {
```
sommerlukas, line 566: Nit, maybe just personal preference: I think `iree-stream-unify-globals-encoding` would be a bit more concise. This would also apply to the class name etc.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,330 @@
...
+struct SourceGlobalInfo {
+  IREE::Util::GlobalOpInterface sourceGlobal;
+  SmallVector<EncodedGlobalInfo> encodedVersions;
+};
+
+// Finds the global store op that uses the given value, traversing through
+// passthrough ops like clone.
+// TODO(#22485): Handle other passthrough ops (e.g., transfer).
+static IREE::Util::GlobalStoreOpInterface findStoreOp(Value value) {
+  for (auto user : value.getUsers()) {
```
sommerlukas: Do we expect the `tensor.encode` operation to have multiple users in the initializer?

If it has multiple users, how do we ensure that the first store is actually the store we're interested in?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,330 @@
...
+      sourceInfo.encodedVersions.push_back(encodedInfo);
+    });
+  }
+  return success();
+}
+
+IREE::Util::GlobalOpInterface
+GlobalEncodingAnalyzer::traceToSourceGlobal(Value value) {
+  // TODO(#22485): Extend to general subgraph canonicalization.
+  // Currently only handles direct global loads + passthrough ops.
```
sommerlukas: Which passthroughs are currently handled? IIUC, only other globals initializers that directly store another loaded global are currently supported, right?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,330 @@
...
+        auto newResultType = RankedTensorType::get(
+            oldResultType.getShape(), oldResultType.getElementType(),
+            unifiedEncoding);
+        encodeOp.setResultEncodingAttr(TypeAttr::get(newResultType));
+        LDBG() << "  Updated encode op: " << encodeOp;
+
+        // Update stream.tensor.sizeof if it defines the result_size.
+        if (auto sizeofOp =
+                encodeOp.getResultSize()
+                    .getDefiningOp<IREE::Stream::TensorSizeOfOp>()) {
```
sommerlukas: Should this be checked with another prerequisite for candidate `tensor.encode` operations, i.e., that the result size is either static or directly coming from a `tensor.sizeof` operation and not from some other operation?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -0,0 +1,200 @@
...
+
+// -----
+
+// Test: mutable encoded global - should be skipped, encoding unchanged.
+
+#encoding1 = #iree_encoding.testing<[#iree_encoding.specialized<123>]>
+
+// CHECK: #[[$ENC:.+]] = #iree_encoding.testing<[#iree_encoding.specialized<123>]>
+// CHECK-LABEL: module @mutable_encoded_global_skipped
+module @mutable_encoded_global_skipped {
```
sommerlukas, line 123: Wouldn't it be better to have two encoded globals in this case? Otherwise, just based on the checks that the test performs, we won't be able to tell whether this case was skipped because of mutability or because there's only a single encoded global (same as test below).

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -0,0 +1,200 @@
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(iree-stream-unify-encoding-for-globals)' %s | FileCheck %s
+
```
sommerlukas, line 2: IIUC, the test currently does not yet test the case where the source global doesn't have an inline initializer and we need to trace through stores.

sommerlukas (commented):
Looks good, just a few nits and questions. I've also posted a question on the RFC in #22485.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
MaheshRavishankar: Similar question. Do we want to limit to the case where there is only one user?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+//===----------------------------------------------------------------------===//
+// Analysis.
+//===----------------------------------------------------------------------===//
+
+// Information about an encoded global and its relationship to its source.
+struct EncodedGlobalInfo {
+  IREE::Util::GlobalOpInterface encodedGlobal;
+  Attribute encodingAttr;
+  IREE::Stream::TensorEncodeOp encodeOp;
```
MaheshRavishankar, line 40: Why do we need the `encodeOp`? It could be a potential source of stale info?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,330 @@
...
+    if (!visited.insert(current).second) {
+      continue;
+    }
+    Operation *defOp = current.getDefiningOp();
+    if (!defOp) {
+      LDBG() << "      Bail: block argument";
+      return nullptr;
+    }
+    bool shouldContinue =
+        llvm::TypeSwitch<Operation *, bool>(defOp)
```
MaheshRavishankar, line 227: This seems like maybe something to move to its own function to make this more readable/managable long-term.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    // Unify encodings for each source global with multiple encodings.
+    for (auto sourceName : candidates) {
+      const auto &sourceInfo =
+          analyzer.getSourceGlobals().find(sourceName)->second;
+
+      // TODO(#22485): Select unified encoding via resolver. For now, use
+      // identity.
+      auto unifiedEncoding =
```
MaheshRavishankar, line 309: So this is just for initial commit right? Cause I dont know what "identity encoding" means.

MaheshRavishankar (approved):
Seems like an initial commit of the pass. By itself looks OK to me. Left a few comments, but nothing blocking. Thanks!

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Yes, this is just for initial commit. In some follow-ups, we'll let encoding resolver to provide the preferred encoding, using new interface methods. They are not developed yet, so we start with identity encoding. A tensor type with identity encoding will be resolved into the same tensor without encodings. I.e., I imagine that they will be resolved to no-encodings if the interface methods are not implemented yet. This is just a first step for testing; we'll improve this in follow-ups.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Good point. The corresponding `tensor.sizeof` op is usually generated during stream conversion. I.e., it generates two globals: one for the actual data and the other for the size. I think we can only handle this case, and we probably just collect the op in the struct and update it directly. If it is not the case, we bail out -- which is safer.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Yes, you're right. Let me update the comment.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: I may be over-designing the case when I evaluate control-flow support.. Let me strict it and we start with the IR chain we are seeing.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: good point, let me update the test.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: Inline is misleading, let me replace it with `with`. The `GlobalOp` has `initialValue` argument; `= #stream.parameter.named<"model"::"weight">` is the initial value.

https://github.com/iree-org/iree/blob/6a282845bd4731d5b222fa0c38e469a4bf63ba1f/compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.td#L1046-L1053

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.td
hanhanW: Does `unify-encoding-for-encoded-globals` sound better? The class name is a tracker for those globals, and it is part of the implementation. I think we want to describe what this pass is doing in the name.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: The IR chain of the problem we are solving is `encode -> clone -> store_into_encoded_global`. I was thinking if there could be other ops in between, but it may be over-designing when I looked at control flow support.

hanhanW:
I did not ask copilot for review.. not sure what's happening. I'll take a look.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.td
sommerlukas: > Does `unify-encoding-for-encoded-globals` sound better?

My aim was to shorten the name/option a bit, e.g., `iree-stream-unify-global-encodings` and `UnifyGlobalEncodingsPass`, but it's really just a nit and we can also just stick to the current name/option.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,338 @@
...
+// Information about a source global and all its encoded versions.
+struct SourceGlobalInfo {
+  IREE::Util::GlobalOpInterface sourceGlobal;
+  SmallVector<EncodedGlobalInfo> encodedVersions;
+};
+
+// Returns the global store op that uses the given value, traversing through
+// passthrough ops like clone. Returns nullptr if not found or multiple users.
+static IREE::Util::GlobalStoreOpInterface findStoreOp(Value value) {
+  if (!llvm::hasSingleElement(value.getUsers())) {
```
sommerlukas, line 52: Nit:
```suggestion
  if (!value.hasOneUse()) {
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -0,0 +1,338 @@
...
+IREE::Util::GlobalOpInterface
+GlobalEncodingAnalyzer::traceToSourceGlobal(Value value) {
+  // TODO(#22485): Extend to general subgraph canonicalization Currently only
+  // handles direct global loads.
+  SmallVector<Value> worklist;
+  DenseSet<Value> visited;
+  IREE::Util::GlobalOpInterface foundSourceGlobal;
+  worklist.push_back(value);
+  while (!worklist.empty()) {
+    Value current = worklist.pop_back_val();
```
sommerlukas: Do we expect the `worklist` to ever have more than one element in it? If so, does the order play a role (i.e., should we pop from the front if we `push_back`)?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
sommerlukas: This comment was specific to the test below, but to the file overall.

What I had in mind was a scenario like this:
```
global1 = <some initial value>

global2
g1 = load global1
store global2, g1

global3
g2 load global2
e1 = encode g2
store global3, e1
```

In this case, the source global would probably be `global1`. Do we expect such a case?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Since the initial support should be fairly simple, let me simplify the implementation for now. We can always extend based on our needs.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: I'll refactor them out once it starts being a monster. It is a little tricky because it maintains two states and the tracking value. It is doable, but I'll need to add more comments around the new method. The function currently has ~60 lines, so it's fine for now.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: They are not equivalent, right? It can have multiple uses and one user if a user uses it more than one time? E.g., `f(value, value)`. The behavior is the same because we only track store and clone ops for now, but I'd like to have clearer check here.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: This is not common in IREE, and we could see the case if someone authors it by hands. IIRC, I did see something like below, but it is not handled yet. It'd be easier to add such test case when we improve the pass.

```
global1 = <some initial value>

global2
g1 = load global1
d1 = dispatch(g1)
store global2, d1

global3
g1 load global1
e1 = encode g1
store global3, e1
```

sommerlukas (approved):

PR #22768, opened 2025-11-26
[LinalgExt] Added support for LinalgExt::IndexOp(s) inside of attention regions

keshavvinayak01:

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -2816,11 +2816,12 @@ LogicalResult IREE::LinalgExt::IndexOp::verify() {
         "expected parent op to be one of `iree_linalg_ext.custom_op`, "
         "`iree_linalg_ext.attention`");
   }
-  int64_t numLoops = customOp ? customOp.getNumLoops() : attentionOp.getNumLoops();
+  int64_t numLoops =
+      customOp ? customOp.getNumLoops() : attentionOp.getNumLoops();
   if (numLoops <= getDim()) {
     return emitOpError("expected dim (")
-           << getDim() << ") to be lower than the number of loops ("
-           << numLoops << ") of the enclosing CustomOp/AttentionOp";
+           << getDim() << ") to be lower than the number of loops (" << numLoops
+           << ") of the enclosing CustomOp/AttentionOp";
   }
   return success();
 }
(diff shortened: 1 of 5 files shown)


MaheshRavishankar (commented):
Could you add a roundtrip test please. An attention operation with a `linalg_ext.index` in the body.

hanhanW (commented):
Please add tests or explain why there are no new tests, if it is not a NFC change.

keshavvinayak01:
> Could you add a roundtrip test please. An attention operation with a `linalg_ext.index` in the body.

Done.

MaheshRavishankar (approved):

hanhanW (changes requested):
I think you should simplify the test case, because roundtrip test should not have torch ops.

hanhanW (approved):

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -2808,14 +2808,20 @@ CustomOp::reifyResultShapes(OpBuilder &builder,
 //===---------------------------------------------------------------------===//
 
 LogicalResult IREE::LinalgExt::IndexOp::verify() {
-  auto customOp = dyn_cast<CustomOp>(getOperation()->getParentOp());
-  if (!customOp) {
-    return emitOpError("expected parent op to be `iree_linalg_ext.custom_op`");
+  auto parentOp = getOperation()->getParentOp();
+  auto customOp = dyn_cast<CustomOp>(parentOp);
+  auto attentionOp = dyn_cast<AttentionOp>(parentOp);
+  if (!customOp && !attentionOp) {
```
hanhanW: Actually, I think this can be `isa<CustomOp, AttentionOp>`?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/invalid.mlir
```diff
@@ -1789,14 +1789,6 @@ func.func @custom_op_yield_type_mismatch(%arg0 : tensor<?xf32>, %arg1 : tensor<1
 
 // -----
 
-func.func @index_op_outside_custom_op() -> index {
-  // expected-error @+1 {{expected parent op to be `iree_linalg_ext.custom_op`}}
-  %0 = iree_linalg_ext.index 0 : index
-  return %0 : index
```
hanhanW, line 1795: Why not update the error message?

Groverkss (changes requested):
This will break tiling. Please add a failure to TilingInterface for attention if there is a linalg_ext.index present in the region.

MaheshRavishankar:
> This will break tiling. Please add a failure to TilingInterface for attention if there is a linalg_ext.index present in the region.

No, it wont break tiling :) . You can fix it up on tiling. Linalg operations do this.

Groverkss:
> > This will break tiling. Please add a failure to TilingInterface for attention if there is a linalg_ext.index present in the region.
> 
> No, it wont break tiling :) . You can fix it up on tiling. Linalg operations do this.

I mean that if you don't account for linalg_ext.index , then it's incorrect tiliing. By "break" i mean the tiled op will now be incorrect. We should add a failure for now if there are any index operations in the region and then fix it up.

keshavvinayak01:
Added the failure and a TODO for future implementation @Groverkss

Groverkss (approved):

PR #22769, opened 2025-11-26
[LinalgExt] Rewriter for Torch::HigherOrderFlexAttentionOp -> LinalgExt::AttentionOp

keshavvinayak01:

I ran the entire flex_attention_hop implementation with randomised input tensors, (Also see https://github.com/llvm/torch-mlir/pull/4366) through aot.export and compared against eager mode, and I noticed no accuracy losses (On CPU)


keshavvinayak01:
@Groverkss @hanhanW @MaheshRavishankar Pinging for long pending review :)

hanhanW:
> @Groverkss @hanhanW @MaheshRavishankar Pinging for long pending review :)

I don't really work on torch->linalg lowering and attention ops, but I can definitely help review. It'll take me some time to ramp up the context.

I know that you mentioned you ran the e2e test locally. My expectation of this PR is that it is the last PR to make FlexAttentionOp function e2e, but I don't see any e2e test enabled. Thus, my first question is: have we set up the test in [iree-test-suites/torch_ops](https://github.com/iree-org/iree-test-suites/tree/main/torch_ops)? It'd be a big plus if we can advance the test suite within this PR, like [this](https://github.com/iree-org/iree/commit/d3440737cc56a4d1b20c72181d9a37f194bd3ce5). Then it is much easier to capture the failures if something is wrong in our optimization.

keshavvinayak01:
> > @Groverkss @hanhanW @MaheshRavishankar Pinging for long pending review :)
> 
> I don't really work on torch->linalg lowering and attention ops, but I can definitely help review. It'll take me some time to ramp up the context.
> 
> I know that you mentioned you ran the e2e test locally. My expectation of this PR is that it is the last PR to make FlexAttentionOp function e2e, but I don't see any e2e test enabled. Thus, my first question is: have we set up the test in [iree-test-suites/torch_ops](https://github.com/iree-org/iree-test-suites/tree/main/torch_ops)? It'd be a big plus if we can advance the test suite within this PR, like [this](https://github.com/iree-org/iree/commit/d3440737cc56a4d1b20c72181d9a37f194bd3ce5). Then it is much easier to capture the failures if something is wrong in our optimization.

Currently I see tests for relu, gelu, transpose multiplies. I can add support for attention there, but is the goal to eventually keep adding support for complex ops in the same place as well? Just so I understand the rationale of that test suite.

keshavvinayak01 (commented):
@hanhanW Added the e2e test PR (tested locally).

keshavvinayak01:
Sorry, DCO check changes messed up the history!

keshavvinayak01:
I was unable to fix this PR's commit history, so I moved to https://github.com/iree-org/iree/pull/23292.

PR #22771, opened 2025-11-26
[Bazel] Migrate to bzlmod for LLVM compatibility

maxbartel:
LLVM has fully migrated to bzlmod and now requires bzlmod features. This change migrates IREE's Bazel build to use bzlmod.

Changes:
- Add MODULE.bazel with dependencies matching LLVM's versions
- Add build_tools/bazel/extensions.bzl module extension for creating IREE-specific repositories
- Enable bzlmod in iree.bazelrc
- Update flatcc BUILD.overlay to use includes= instead of hardcoded -Iexternal/... paths (bzlmod changes external paths)
- Update bitcode library paths for bzlmod repository naming
- Update llvm-project submodule to include upstream bzlmod changes
- Add BZLMOD_LLVM.md documentation for downstream consumers

The iree_extension creates llvm-raw only when IREE is the root module, allowing downstream consumers to provide their own LLVM by creating their own llvm-raw repository.


MODULE.bazel
```diff
@@ -0,0 +1,77 @@
...
+# Core Bazel dependencies - versions aligned with LLVM's MODULE.bazel
+bazel_dep(name = "apple_support", version = "1.24.1", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "bazel_skylib", version = "1.8.2")
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_cc", version = "0.2.11")
+bazel_dep(name = "rules_python", version = "1.6.3")
+bazel_dep(name = "rules_shell", version = "0.6.1")
+
+# Required by LLVM (for clang-fuzzer tools)
+bazel_dep(name = "protobuf", version = "31.1", repo_name = "com_google_protobuf")
```
ScottTodd, line 33: I don't think protobuf should be needed for IREE's LLVM build. I left a comment at https://github.com/llvm/llvm-project/pull/168928#discussion_r2566681186.

(If it is relatively painless to add here, fine, but I still think we should push back)

ScottTodd (commented):
Structurally LGTM.

Looks like there are some build errors: https://github.com/iree-org/iree/actions/runs/19711280974/job/56472322860?pr=22771

and pre-commit errors: https://github.com/iree-org/iree/actions/runs/19711280936/job/56472301274?pr=22771

MODULE.bazel
maxbartel: I can try removing it. Bzlmod should only download it anyway if there is a real dependency on it.

maxbartel: Took a closer look and as long as upstream carries the dependency in the clang BUILD file, IREE also needs to have it. If they refactor it out like you propose in the comment, we can also drop it.

bjacob (approved):

MODULE.bazel
ScottTodd: Could you/someone test with https://github.com/llvm/llvm-project/pull/170167 ?

PR #22775, opened 2025-11-26
[tests][e2e] Add custom mxfp4 gemm tests to verify shape of interest.

Muzammiluddin-Syed-ECE:
This patch adds a e2e matmul test for the shape `10000,16384,16384` which appears in models we care about. There was a choice between adding a singular custom test vs creating a new class of tests specifically for VERY_LARGE shapes which would be very problematic space wise for any other data type but mxfp4. But given that we want to test a very specific shape I didn't feel the need to go through with more changes than necessary. Open to discussion.

### DIFF
--- a/tests/e2e/matmul/CMakeLists.txt
+++ b/tests/e2e/matmul/CMakeLists.txt
@@ -1997,6 +1997,37 @@ iree_generated_e2e_runner_test(
     "requires-gpu-cdna4"
 )
 
+iree_generated_e2e_runner_test(
+  NAME
+    e2e_matmul_cdna4_mxfp4_llama
+  TEST_TYPE
+    matmul
+  GENERATOR
+    "generate_e2e_matmul_tests.py"
+  GENERATOR_ARGS
+    "--lhs_rhs_type=f4E2M1FN"
+    "--acc_type=f32"
+    "--mx_scale_type=f8E8M0FNU"
+    "--mx_block_size=32"
+    "--shapes=custom_mnk"
+    "--mnk=10000,16384,16384"
+    "--transpose_rhs"
+  TEST_RUNNER
+    iree_tools_testing_e2e_iree-e2e-matmul-test
+  TARGET_BACKENDS
+    "rocm"
+  DRIVERS
+    "hip"
+  COMPILER_FLAGS
+    ${IREE_HIP_TEST_COMPILER_FLAGS}
+  LABELS
+    "noasan"
+    "nomsan"
+    "notsan"
+    "noubsan"
+    "requires-gpu-cdna4"
+)
+
 endif()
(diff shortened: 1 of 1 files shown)


Muzammiluddin-Syed-ECE:
Also, I don't understand why all the other backends are explicitly included in [iree/tests/e2e/matmul/BUILD.bazel](https://github.com/iree-org/iree/blob/main/tests/e2e/matmul/BUILD.bazel) but rocm isn't.

bjacob (approved):
Yes, `custom_mnk` exists exactly for this reason.  A new shape collection can also be introduced, and it is not a problem if it is only used for some MXFP4 tests, but only becomes useful if it is going to contain multiple shapes. As long as it's a single shape, there's no reason not to use custom_mnk.

Regarding your other question about Bazel, note that it ramifies into 2 questions:
1. Why not run ROCm tests in the Bazel build?
2. Even if we are not going to run ROCm tests in the Bazel build, why not at least describe them in the Bazel build so that the CMakeLists can be autogenerated by `bazel_to_cmake`?

Regarding 1., the answer is a combination of weak motivation and difficulty:
* Weak motivation: hardly anyone cares, because these days hardly anyone actually uses Bazel to build, historically this was used at Google, Google isn't using IREE anymore and isn't using ROCm anyway.
* Difficulty: testing ROCm requires having configured for a particular target chip, as in the CMake variable `IREE_HIP_TEST_TARGET_CHIP`.  Bazel is notoriously *hostile* to configuration -- it doesn't have a configuration-logic step like CMake or any other build system. It's a purely declarative mapping of `bazel` command lines so build graph. So specifying the target chip could only be done as a config setting *on the bazel command line*.  And then you don't have an `if` statement in the Bazel language, instead of you have a `select` statement that looks more like a chain of ternary operators - it is an expression, not control-flow. That could be implemented, but it would look very different from the CMake code, so whichever way we approach this, the bazel-to-cmake translation could not be fully automated.

Regarding 2., there is room to do something here. Like explained above, it won't be possible to have all the chip-specializing logic in Bazel and have bazel-to-cmake translate it. But it should be possible to improve over the current situation where more than half of the CMakeLists is hand-authored.  As a strawman, we can use the Bazel function `iree_cmake_extra_content` to place the CMake `if` statements in a way that's opaque to Bazel.

Muzammiluddin-Syed-ECE:
Thanks for the explanation! I can see why the situation has progressed to this point. In my own experience, every time I make a change that requires touching CMake, it's never an amount of handwritten CMake code that I'm offended by, and so hand-authoring is tolerable. This means that for any given change, the benefit gained from something like bazel-to-cmake feels less important. 

Perhaps this would be worth making an issue about.

PR #22778, opened 2025-11-27
[LinalgExt] Added toggle for using useExp2 for onlineAttention Decomposition

keshavvinayak01:
Following the discussion from https://github.com/iree-org/iree/pull/22441

Depending on the backend, certain computations may benefit from directly using `exp` instead of `exp2`, since there might be accuracy losses due to FP-reassociation.  It's helpful to add flag incase the user tracks losses to this particular computation and might favour directly using `exp`.

The `use_exp2` flag is mostly unused in dialect conversions and passes, I presume it's used as a KernelOption. The changes here will not modify the default behavior.


MaheshRavishankar (changes requested):
I think it would be simpler to just add an optional attribute to the op itself to use exp2 for decomposition. This doesnt need to be a separate attribute that is not part of the op definition.

keshavvinayak01:
> I think it would be simpler to just add an optional attribute to the op itself to use exp2 for decomposition. This doesnt need to be a separate attribute that is not part of the op definition.

I was going to do that, but then I saw https://github.com/iree-org/iree/blob/70b2b4583fbf607b2be9873b297ef1ec0bf1db9b/compiler/src/iree/compiler/Codegen/Common/test/strip_compilation_info.mlir#L101

Where it's part of the decomposition config itself. So I thought I'd better refine the attribute and continue using it. Making it an optional attribute of the op itself might introduce redundancy since this already exists @MaheshRavishankar ?

MaheshRavishankar:
> > I think it would be simpler to just add an optional attribute to the op itself to use exp2 for decomposition. This doesnt need to be a separate attribute that is not part of the op definition.
> 
> I was going to do that, but then I saw
> 
> https://github.com/iree-org/iree/blob/70b2b4583fbf607b2be9873b297ef1ec0bf1db9b/compiler/src/iree/compiler/Codegen/Common/test/strip_compilation_info.mlir#L101
> 
> Where it's part of the decomposition config itself. So I thought I'd better refine the attribute and continue using it. Making it an optional attribute of the op itself might introduce redundancy since this already exists @MaheshRavishankar ?

I dont know the history of that, but we probably need to drop the old usage and just add an optional attribute here. @Groverkss  comments?

Groverkss:
> > > I think it would be simpler to just add an optional attribute to the op itself to use exp2 for decomposition. This doesnt need to be a separate attribute that is not part of the op definition.
> > 
> > 
> > I was going to do that, but then I saw
> > https://github.com/iree-org/iree/blob/70b2b4583fbf607b2be9873b297ef1ec0bf1db9b/compiler/src/iree/compiler/Codegen/Common/test/strip_compilation_info.mlir#L101
> > 
> > Where it's part of the decomposition config itself. So I thought I'd better refine the attribute and continue using it. Making it an optional attribute of the op itself might introduce redundancy since this already exists @MaheshRavishankar ?
> 
> I dont know the history of that, but we probably need to drop the old usage and just add an optional attribute here. @Groverkss comments?

I think it's okay to have a decomposition config dictionary and add these attributes to it. Attention op needs multiple configuration options so it's useful to have a dictionary.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/decompose_aggregate_op.mlir
```diff
@@ -416,3 +416,95 @@ func.func @online_attention_f8_masked(%query: tensor<192x1024x64xf8E4M3FNUZ>,
...
+        outs(%output, %max, %sum : tensor<192x1024x64xf32>, tensor<192x1024xf32>, tensor<192x1024xf32>) {
+                      ^bb0(%score: f32):
+                        iree_linalg_ext.yield %score: f32
+                     }
+        -> tensor<192x1024x64xf32>, tensor<192x1024xf32>, tensor<192x1024xf32>
+
+  return %out#0, %out#2 : tensor<192x1024x64xf32>, tensor<192x1024xf32>
+}
+
+// We want to check that we're correctly using exp
```
Groverkss, line 460: You don't need to FileCheck everything, just check if we are using exp/exp2 and the log multiplication

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Passes.td
```diff
@@ -104,6 +104,11 @@ def DecomposeAttentionPass :
     InterfacePass<"iree-linalg-ext-decompose-attention", "mlir::FunctionOpInterface"> {
   let summary =
       "Decomposes attention op into a sequence of linalg ops";
+  let options = [
+    Option<"useExp2", "use-exp2", "bool", /*default=*/"true",
+        "Use exp2 for computations; Tunable to allow for accuracte computations"
+        "in case of accuracy losses due to fp-reassociation.">,
+  ];
```
Groverkss, line 123: This shouldn't be a pass option, this already exists on the op

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeAttention.cpp
```diff
@@ -32,8 +32,16 @@ struct DecomposeAttentionPass final
 void DecomposeAttentionPass::runOnOperation() {
   MLIRContext *context = &getContext();
   IRRewriter rewriter(context);
+
+  SmallVector<NamedAttribute> decompositionConfigAttrs;
+  decompositionConfigAttrs.push_back(
+      rewriter.getNamedAttr("use_exp2", rewriter.getBoolAttr(useExp2)));
+  DictionaryAttr decompositionConfig =
+      rewriter.getDictionaryAttr(decompositionConfigAttrs);
+
```
Groverkss: Instead of breaking all tests, make the option to use_exp2 by default if not specified.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -1007,6 +1007,8 @@ def IREELinalgExt_OnlineAttentionOp : IREELinalgExt_Op<"online_attention",
     // Attributes to set on QK and PV matmul after decomposition.
     static StringRef getQKAttrStr() { return "qk_attrs"; }
     static StringRef getPVAttrStr() { return "pv_attrs"; }
+    // Flag to control whether to use exp2 (with log2(e) scaling) or exp.
+    static StringRef getUseExp2AttrStr() { return "use_exp2"; }
```
Groverkss, line 1105: Document `decomposition_config` in the op definition and add some documentation about this attribute there.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Passes.td
keshavvinayak01: When I try to compile the new `FlexAttention` or even the existing torch `SDPA` op, it typically converts to `tmTensor::AttentionOp` -> `LinalgExt::AttentionOp`, which converts to `LinalgExt::OnlineAttentionOp` finally. I thought it'd be easier to manually hack into the pass pipeline and add this option if needed, because in this flow, I'll have to modify the Torch op itself to pass this option to the last level. What do you think?

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeAttention.cpp
keshavvinayak01: Yes that's the default pass option.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -923,6 +923,16 @@ def IREELinalgExt_OnlineAttentionOp : IREELinalgExt_Op<"online_attention",
...
+
+    Decomposition Configuration:
+    The `decomposition_config` attribute is a DictionaryAttr that controls how
+    this operation is decomposed into lower-level operations. It supports:
+      - "qk_attrs": DictionaryAttr - Attributes to attach to the Q@K matmul
+        operation after decomposition (e.g., lowering_config, attention markers)
+      - "pv_attrs": DictionaryAttr - Attributes to attach to the P@V matmul
+        operation after decomposition
+      - "use_exp2": BoolAttr - If true, uses exp2 with log2(e) scaling instead
+        of exp. (Might be better accuracy-wise on some hardware)
```
Groverkss: Might be better accuracy-wise on some hardware > Gives better perf on some hardware, but trades off accuracy

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Passes.td
Groverkss: Just hack it into the pass for now, no need to add it as an option. Add a TODO to move it out and give the user control over it.

Groverkss (approved):

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Passes.td
keshavvinayak01: As discussed offline, the previous behaviour is maintained by the default pass option being true, and this gives the user control over turning it off.

Groverkss:
@MaheshRavishankar Can you have a look at this again?

MaheshRavishankar (approved):

keshavvinayak01 (commented):
Let's re-run CI on this and get it merged? @Groverkss

MaheshRavishankar:
What happened here. Why did you close this?

keshavvinayak01:
I was trying to rebase and push to trigger CI, but the git history got messed up. So I re-opened it https://github.com/iree-org/iree/pull/23211

hanhanW:
Next time, you can get to `#23211` branch and run `git checkout -B this-branch`. It will overwrite the current branch history; you can re-open this PR.

Merging PR like that is introducing overheads for future code tracking, IMO. You force people clicking into many links to figure out old review comments and the reason of making changes.

PR #22475, opened 2025-10-30
[ci] Add CI for MI355x machines

Groverkss:

PR #22780, opened 2025-11-27
[CI] Adding MI355x runner test

Muzammiluddin-Syed-ECE:
We have self-hosted runners that can run mxfp4 e2e tests so we should use them.


bjacob:
This looks similar to https://github.com/iree-org/iree/pull/22475 but I don't know why that wasn't merged. @Groverkss , @kuhar .

Issue #22370, opened 2025-10-21
[DataTiling] Support dynamic layout decisions


Max191:
# Background/Problem Statement #

Data tiling currently relies on static information to decide layouts. This makes it impossible to make decisions about layouts dynamically. The main use for this right now is in selecting ukernels for data tiled kernels. In GPU codegen, we have a few different sizes of ping pong ukernels, which can be selected based no the problem size. We do this because the larger ukernels may have too large of a tile to occupy all the CUs on the GPU.

In the non-data-tiling path, we do this via variant specialization within a hal.executable. This happens in the SpecializeExports pass, (tests here: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/test/specialize_exports.mlir).

In the data tiling path, this is much harder, because the ukernel that we select dictates the layout that we choose for data tiling. This means that the producer dispatches need to make the same decision independently about which ukernel will be chosen. As an example, say we have a fully dynamic matmul:

```mlir
%lhs = flow.tensor.encode %lhs_in : tensor<?x?xf32>-> tensor<?x?xf32, #lhs_encoding>
%rhs = flow.tensor.encode %rhs_in : tensor<?x?xf32>-> tensor<?x?xf32, #rhs_encoding>
%acc = flow.tensor.encode %acc_in : tensor<?x?xf32>-> tensor<?x?xf32, #result_encoding>
%matmul_dispatch = flow.dispatch.region ... {
  %matmul = linalg.matmul
    indexing_maps = [
      affine_map<(d0, d1, d2) -> (d0, d2)>,
      affine_map<(d0, d1, d2) -> (d1, d2)>,
      affine_map<(d0, d1, d2) -> (d0, d1)>
    ]
    ins(%lhs, %rhs : tensor<?x?xf16, #lhs_encoding>, tensor<?x?xf16, #rhs_encoding>)
    outs(%acc : tensor<?x?xf32, #result_encoding>) -> tensor<?x?xf32, #result_encoding>
  %unset_encoding = iree_encoding.unset_encoding %matmul : tensor<?x?xf32, #result_encoding> -> tensor<?x?xf32>
  flow.return %unset_encoding : tensor<?x?xf32>
}
```

The chosen ukernel will depend on the dynamic values of M, N, and K, but in each of the operands (`%lhs`, `%rhs`, `%acc`), only 2 of the 3 dimensions are present.

# Proposed Solution #

There are 2 problems above that need to be solved:

[...]

PR #22221, opened 2025-10-06
Carry encoding in the preferred storage type of a hoistable type

jtuyls:
This PR makes sure that the encoding attribute is included in the preferred storage type of a `HoistableTensorType`. Without this we can get a `tensor.bitcast` on an encoded type returning a non-encoded type, which will result in a compilation failure when converting into stream. 

This PR adds a `convertForBitcast` interface method to the SerializableAttr interface for encodings to specify how the types should be converted on a bitcast operation. This adds generality as different encodings can provide different implementations, allows them to perform correctness checks and would even allow propagating encodings further on the bitcasted type if needed in the future.

PR #22783, opened 2025-12-01
[DispatchCreation] Fold unit dims with encoding

sommerlukas:
Extend the `FoldUnitExtentDims` pass to handle tensors with encoding attached.

Encodings can opt into folding of unit extent dimensions by adding the `SerializableAttr` interface and implementing the new interface method `getCollapsedEncoding` that is used to query whether a tensor with encoding can be collapsed. If any of the tensor operands or results has an encoding that doesn't implement the interface, no folding happens to avoid issues with 0-entries in the indexing maps and passes that don't handle that case well.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
@@ -220,6 +220,23 @@ def IREEEncoding_SerializableAttr :
         assert(false && "unimplemented interface method");
         return {};
       }]
+    >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Returns the new encoding after collapsing the type from `origShape`
+        to `targetShape`. If the type with this encoding cannot be collapsed,
+        returns `nullptr`.
+      }],
+      /*retTy=*/"::mlir::Attribute",
+      /*methodName=*/"getCollapsedEncoding",
+      /*args=*/(ins
+        "::llvm::ArrayRef<int64_t>":$origShape,
+        "::llvm::ArrayRef<int64_t>":$targetShape
+      ),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        return nullptr;
+      }]
     >
   ];
 
@@ -337,32 +354,6 @@ def IREEEncoding_ContractionEncodingAttrInterface :
       /*methodBody=*/"",
       /*defaultImplementation=*/[{
         return std::nullopt;
-      }]
-    >
-  ];
-}
-
-def IREEEncoding_CollapsibleEncodingAttrInterface :
-  AttrInterface<"CollapsibleEncodingAttrInterface"> {
-  let cppNamespace = "::mlir::iree_compiler::IREE::Encoding";
-  let description = [{
-    Interface used to query whether a tensor with the encoding
-    can be collapsed, e.g., to fold away unit extent dimensions.
-  }];
-
-  let methods = [
-    InterfaceMethod<
-      /*desc=*/[{
-        Returns true if a tensor type with this encoding can be
-        collapsed.
-      }],
-      /*retTy=*/"bool",
-      /*methodName=*/"canCollapse",
-      /*args=*/(ins
-      ),
-      /*methodBody=*/"",
-      /*defaultImplementation=*/[{
-        return false;
       }]
     >
   ];
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
@@ -342,6 +342,32 @@ def IREEEncoding_ContractionEncodingAttrInterface :
   ];
 }
 
+def IREEEncoding_CollapsibleEncodingAttrInterface :
+  AttrInterface<"CollapsibleEncodingAttrInterface"> {
+  let cppNamespace = "::mlir::iree_compiler::IREE::Encoding";
+  let description = [{
+    Interface used to query whether a tensor with the encoding
+    can be collapsed, e.g., to fold away unit extent dimensions.
+  }];
+
+  let methods = [
+    InterfaceMethod<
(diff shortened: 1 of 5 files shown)


sommerlukas:
PR for the necessary changes to upstream MLIR: https://github.com/llvm/llvm-project/pull/170104

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
```diff
@@ -303,6 +303,32 @@ def IREEEncoding_ContractionEncodingAttrInterface :
...
+        Returns true if a tensor type with this encoding can be
+        collapsed.
+      }],
+      /*retTy=*/"bool",
+      /*methodName=*/"canCollapse",
+      /*args=*/(ins
+      ),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        return true;
```
hanhanW: It is better to return `false` by default, as it is safer to me. New encodings should implement it if they want the behavior.

hanhanW (commented):
I haven't reviewed the details yet, because I need to review the upstream one first. 

High-level comment: it is better to split it into two PRs:
- One is for LLVM integrate, which always returns false if encoding is present. I.e., it matches the current behavior on the main branch.
- The other is for adding the interface method and relevant tests, which is an improvement to IREE.

I'm treating this PR more like a demo.

sommerlukas:
After discussion with @hanhanW and @MaheshRavishankar, I took a different approach to exposing more control upstream, see the new PR https://github.com/llvm/llvm-project/pull/171796.

hanhanW:
> I haven't reviewed the details yet, because I need to review the upstream one first.
> 
> High-level comment: it is better to split it into two PRs:
> 
> * One is for LLVM integrate, which always returns false if encoding is present. I.e., it matches the current behavior on the main branch.
> * The other is for adding the interface method and relevant tests, which is an improvement to IREE.
> 
> I'm treating this PR more like a demo.

@sommerlukas can you prepare a NFC PR for integrate before you land the upstream change? https://github.com/llvm/llvm-project/pull/171796

Thanks!

sommerlukas:
> > I haven't reviewed the details yet, because I need to review the upstream one first.
> > High-level comment: it is better to split it into two PRs:
> > 
> > * One is for LLVM integrate, which always returns false if encoding is present. I.e., it matches the current behavior on the main branch.
> > * The other is for adding the interface method and relevant tests, which is an improvement to IREE.
> > 
> > I'm treating this PR more like a demo.
> 
> @sommerlukas can you prepare a NFC PR for integrate before you land the upstream change? [llvm/llvm-project#171796](https://github.com/llvm/llvm-project/pull/171796)
> 
> Thanks!

I've created an NFC PR for the integration in #22921. From this PR, only the last two commits need to be reviewed for now, as they add handling of encodings in `FoldUnitExtentDims` based on the new upstream API.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
```diff
@@ -303,6 +303,32 @@ def IREEEncoding_ContractionEncodingAttrInterface :
...
+      /*methodName=*/"canCollapse",
+      /*args=*/(ins
+      ),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        return false;
+      }]
+    >
+  ];
+}
```
hanhanW: I'm worried about having too many interfaces and some of them could be too specific. My preference is adding the new method to `SerializableAttr` interface, and break it into few more interfaces once we have better abstraction. For now, it seems to be a good home for the method, as it is about encoding properties. @jtuyls is going to extend the interface for his [dynamic layout decisions](https://github.com/iree-org/iree/issues/22370) work, which follows this working style.

btw, I'm expecting something more generic. What do you think if we make the interface method take `original shape` and `target shape`; return the new encoding attribute? The idea is more like: given the encoding you have with the original shape, what is the new encoding for the target shape if it's collapsible?

I.e.,

```
/// Returns nullptr on failure.
Attribute getCollapsedEncoding(ArrayRef<int64_t> origShape, ArrayRef<int64_t> targetShape);
```

(Potentially, it can be used by data-tiling `EncodingAttr`, as it also describes the indexing maps and it can return a new `EncodingAttr` with new indexing maps. Thus, it looks more generic to me. And the idea I shared above is how I interpret this method).

hanhanW (commented):
Thanks for working this out! I have a high-level comment for now; I'll review details once we are aligned. The rest of the code basically looks okay, as they are just building the control function.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
sommerlukas: Ok, I can integrate it into an existing interface. 

> btw, I'm expecting something more generic. What do you think if we make the interface method take original shape and target shape; return the new encoding attribute?

Sounds good to me. I was thinking instead of just the original shape, we could take the original type, so element type information can be considered for the new encoding. WDYT?

hanhanW: I'm concerned about passing a type because the encodings could mismatch. I recently raised a similar concern in a review: https://github.com/iree-org/iree/pull/22221#discussion_r2625523224

When do we need element type? We may need that when we have advanced encoding for non-packed-storage encoding, but the design is not flesh out. It is an easy change, so I suggest starting with the shapes. The core idea for folding unit dims with encoding is: given the encoding you have with the original shape, what is the new encoding for the target shape if it's collapsible.

My understanding is: the interface method returns an encoding, and the caller can use that to compose the new type.

sommerlukas: Ok, makes sense, I've implemented the version taking original and target shape and returning an encoding (or `nullptr`) now.

hanhanW (approved):
LGTM, thanks

PR #22788, opened 2025-12-01
[Codegen][GPU] Enable 3-stage pipelining with hipblaslt compute->write->read ordering

jerryyin:
3-stage uses different cluster ordering to maximize distance between read and use. Stage definitions remain the same. The changes are mostly trivial and the new pipelined code work out of box.

On MI300, the tracked gemm configs improve performance by 4.5% on average for NN and NT layouts.

Right now there's no easy way to experiment this through a compiler flag, and the only way to experiment this is to change the tablegen default number of stage to 3. I plan to expose it properly in the next PR.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -399,7 +399,6 @@ populateOpToStageMap(const StageClassification &stages, scf::ForOp forOp,
 static void
 populateOpToClusterMap(const StageClassification &stages, unsigned numStages,
                        llvm::DenseMap<Operation *, unsigned> &opToCluster) {
-
   unsigned clusterID = 0;
 
   if (numStages == 2) {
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -622,13 +647,6 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
     return forOp;
   }
 
-  // Multi-stage pipelining (numStages > 2) is not yet implemented.
-  if (numStages > 2) {
```
krzysz00, line 626: Can we keep this to check for N > 3 instead? I don't think we have support for that

krzysz00 (approved):
One comment, otherwise approved

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Yah, it's a good idea to cap it to 3 to be non-confusing.

Issue #22377, opened 2025-10-22
Remove `--iree-opt-generalize-matmul=false` from CLIP benchmarks


IanWood1:
This flag shouldn't be needed for CLIP. However, dropping the flag negatively impacts the number of dispatches raising them from 790 to 1049. This appears to be a problem with transpose propagation's handling of generic matmuls. 

From https://github.com/iree-org/iree/pull/20217#discussion_r1992574392:

> The performance regresses when generalizing matmuls, so I had to explicitly disable it. It's related to transpose propagation. We fuse matmul(a, transpose(b)) (transpose has multiple uses) only if matmul is a named op but not if it is a generic.

IanWood1:
Fixed by https://github.com/iree-org/iree/commit/ddf05030686396dc186ab75e0cf30e58474dbe6f

PR #22790, opened 2025-12-01
[GlobalOpt] Fuse transpose into matmul-looking linalg.generic

raayandhar:
We have this issue here: https://github.com/iree-org/iree/issues/22377

We have some `linalg.generic`'s that look like matmuls (in fact, we can convert them to matmul's using `inferContractionDims`). However, when there is a transpose as input to the matmul, we don't recognize that we can fuse them by changing the indexing maps. For example, when two transpose share the same transposed RHS, with the flag we are able to fuse the transpose into both matmuls. Without the flag we don't do this (and have `linalg.generic`'s instead).

Take this [reproducer.mlir](https://gist.github.com/raayandhar/30ca14e141a4b73c80f04b7f4721b998) and this [compile-script.sh](https://gist.github.com/raayandhar/fe2209b2431625ed25ef0afa9bae49d9), run something like `./compile-script.sh > out.mlir 2>&1` with and without the flag to see the difference. Also, you can see the difference without this flag and [without these changes](https://gist.github.com/raayandhar/5061a6b4380c5e91ac10ca530e05f265) vs. [with these changes](https://gist.github.com/raayandhar/52981c91b56e2c4c61d7eb945eeb708a), see line 66 on each where we effectively fuse the transpose by changing the indexing maps.

### DIFF
--- a/compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
@@ -1022,10 +1022,10 @@ static bool isTransposeOfLastTwoDims(ArrayRef<int64_t> perm) {
          perm[rank - 1] == static_cast<int64_t>(rank - 2);
 }
 
-// Fuses a transpose into a contraction (matmul-like) linalg.generic by
-// absorbing it into the indexing map. Only handles transposes that swap
-// the last two dimensions of an operand for now.
-class FuseTransposeThroughGenericContraction
+// Fuses a transpose into a reduction linalg.generic by absorbing it into the
+// indexing map. Only handles transposes that swap the last two dimensions of an
+// operand for now.
+class FuseTransposeThroughGenericReduction
     : public OpRewritePattern<linalg::GenericOp> {
 public:
   using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
@@ -1035,23 +1035,10 @@ public:
     if (!IREE::Flow::isNonNullAndOutsideDispatch(genericOp))
       return failure();
 
-    FailureOr<linalg::ContractionDimensions> maybeContractionDims =
-        linalg::inferContractionDims(genericOp);
-    if (failed(maybeContractionDims))
-      return rewriter.notifyMatchFailure(genericOp, "not a contraction");
-
-    linalg::ContractionDimensions contractionDims =
-        std::move(*maybeContractionDims);
-    if (contractionDims.m.size() != 1 || contractionDims.n.size() != 1 ||
-        contractionDims.k.size() != 1 || contractionDims.batch.size() > 1) {
-      return rewriter.notifyMatchFailure(genericOp,
-                                         "not a simple matmul contraction");
-    }
-
-    DenseSet<int64_t> mnkDims;
-    mnkDims.insert(contractionDims.m.begin(), contractionDims.m.end());
-    mnkDims.insert(contractionDims.n.begin(), contractionDims.n.end());
-    mnkDims.insert(contractionDims.k.begin(), contractionDims.k.end());
+    // Only try to fuse when the generic has at least one reduction dimension.
+    // This avoids elementwise cases that are already handled elsewhere.
+    if (genericOp.getNumParallelLoops() == genericOp.getNumLoops())
+      return rewriter.notifyMatchFailure(genericOp, "not a reduction");
 
     // Look for a transpose on any input that swaps the last two dimensions.
     for (int64_t inputIdx = 0; inputIdx < genericOp.getNumDpsInputs();
@@ -1071,11 +1058,6 @@ public:
       ArrayRef<AffineExpr> results = inputMap.getResults();
       size_t rank = results.size();
       if (rank < 2)
-        continue;
-      AffineDimExpr dim0 = dyn_cast<AffineDimExpr>(results[rank - 2]);
-      AffineDimExpr dim1 = dyn_cast<AffineDimExpr>(results[rank - 1]);
-      if (!dim0 || !dim1 || !mnkDims.contains(dim0.getPosition()) ||
-          !mnkDims.contains(dim1.getPosition()))
         continue;
 
       // Fuse by updating the indexing map to absorb the transpose.
@@ -1124,8 +1106,8 @@ struct PropagateLinalgTransposePass
 };
 } // namespace
 
-static void populateNamedOpSinkingPatterns(MLIRContext *context,
-                                           RewritePatternSet &sinkingPatterns) {
+static void populateMatmulSinkingPatterns(MLIRContext *context,
+                                          RewritePatternSet &sinkingPatterns) {
   sinkingPatterns.insert<NamedOpConversion</*OpType=*/linalg::MatmulOp,
                                            /*inputIdx=*/1>>(
       context, SmallVector<int64_t>{1, 0});
@@ -1138,7 +1120,7 @@ static void populateNamedOpSinkingPatterns(MLIRContext *context,
   sinkingPatterns.insert<NamedOpConversion</*OpType=*/linalg::BatchMatmulOp,
                                            /*inputIdx=*/0>>(
       context, SmallVector<int64_t>{0, 2, 1});
-  sinkingPatterns.insert<FuseTransposeThroughGenericContraction>(context);
+  sinkingPatterns.insert<FuseTransposeThroughGenericReduction>(context);
 }
 
 static void
@@ -1187,7 +1169,7 @@ void PropagateLinalgTransposePass::runOnOperation() {
     sinkingPatterns.insert<SinkTransposeThroughExtractSlice>(context);
     sinkingPatterns.insert<SinkTransposeThroughExpandShape>(
         context, enableEdgeReshapePropagation);
-    populateNamedOpSinkingPatterns(context, sinkingPatterns);
+    populateMatmulSinkingPatterns(context, sinkingPatterns);
     populateCommonCanonicalizationPatterns(context, sinkingPatterns);
     sinkingPatterns.add<SinkTransposeThroughUnaryElementwiseInput>(
         context, /*benefit=*/2);
@@ -1331,7 +1313,7 @@ void PropagateLinalgTransposePass::runOnOperation() {
     sinkingPatterns.insert<FuseTransposeWithLinalgOpConsumer>(
         context, enableAggressivePropagation, enableConvolutionPropagation);
     sinkingPatterns.insert<ComposeTransposes>(context);
-    populateNamedOpSinkingPatterns(context, sinkingPatterns);
+    populateMatmulSinkingPatterns(context, sinkingPatterns);
     populateCommonCanonicalizationPatterns(context, sinkingPatterns);
     sinkingPatterns.add<SinkTransposeThroughUnaryElementwiseInput>(
         context, /*benefit=*/2);
--- a/compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
@@ -280,7 +280,7 @@ public:
       for (int i = 0, e = transposedMap.getNumDims(); i < e; ++i) {
         if (transposedMap.isFunctionOfDim(i)) {
           interchange.push_back(
-              cast<AffineDimExpr>(transposedMap.getResult(permIdx))
+              llvm::cast<AffineDimExpr>(transposedMap.getResult(permIdx))
                   .getPosition());
           permIdx++;
           continue;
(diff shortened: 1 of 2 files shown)


raayandhar:
cc: @IanWood1

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,88 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    // Just taken from earlier (Specialize pattern)
+    if (!IREE::Flow::isNonNullAndOutsideDispatch(genericOp)) {
+      return failure();
+    }
+
+    auto maybeContractionDims = linalg::inferContractionDims(genericOp);
```
IanWood1: nit: can you use the type name here instead of auto

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,88 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+    if (failed(maybeContractionDims)) {
+      return rewriter.notifyMatchFailure(genericOp, "not a contraction");
+    }
+    auto contractionDims = *maybeContractionDims;
```
IanWood1: This should `std::move` or use a reference to prevent the copy

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,88 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+    // This is a suggestion from Claude. Does this make sense?
+    // Should we even care about batch vs. non-batch?
+    SmallVector<int64_t> expectedPerm;
+    if (contractionDims.batch.empty()) {
+      expectedPerm = {1, 0};
+    } else if (contractionDims.batch.size() == 1) {
+      expectedPerm = {0, 2, 1};
+    } else {
+      return rewriter.notifyMatchFailure(genericOp, "unsupported batch size");
+    }
```
IanWood1: To keep in sync with `NamedOpConversion`, I think we just want to make sure that the transpose isn't transposing batch dims. So, I think this works

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,88 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+        continue;
+      }
+
+      // This does the "fusion" part of the pattern.
+      // Not sure if this is actually correct way to accomplish what we want to
+      // do The other sinking patterns just replace the op with a new one
+      // basically It seems to work though; see line 66 in try.mlir (with these
+      // changes) vs/ line 66 in try-old.mlir (without these changes) After
+      // canonicalizing transpose in place (?) This is modifying the OP in order
+      // to fuse, is this what we want?
```
IanWood1: This looks good. It's similar to what `NamedOpConversion` does.

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,88 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+          AffineMap::get(inputMap.getNumDims(), inputMap.getNumSymbols(),
+                         newExprs, rewriter.getContext());
+      newIndexingMaps[inputIdx] = transposedMap;
+
+      rewriter.startOpModification(genericOp);
+      genericOp.setIndexingMapsAttr(
+          rewriter.getAffineMapArrayAttr(newIndexingMaps));
+      genericOp.setOperand(inputIdx, transpose.getInput());
+      rewriter.finalizeOpModification(genericOp);
+      return success();
```
IanWood1: Optional: It might be slightly more readable if this was moved out of the for loop.

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,88 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+                                         "not a simple matmul contraction");
+    }
+
+    // This is a suggestion from Claude. Does this make sense?
+    // Should we even care about batch vs. non-batch?
+    SmallVector<int64_t> expectedPerm;
+    if (contractionDims.batch.empty()) {
+      expectedPerm = {1, 0};
+    } else if (contractionDims.batch.size() == 1) {
+      expectedPerm = {0, 2, 1};
```
IanWood1: I think this assumes that the 0th dim is the batch dim in the linalg op. You will need to check the operand's indexing maps to make sure that it actually is.

IanWood1 (commented):
Looks good! From what I can tell, this should replicate the transformation but for `linalg.generic` ops. It would be good to add some tests, especially for the case mentioned here https://github.com/iree-org/iree/pull/22790#discussion_r2579150119


One thing I didn't consider is that linalg named matmul ops now have indexing maps (https://mlir.llvm.org/docs/Dialects/Linalg/#linalgmatmul-linalgmatmulop). It might be possible to completely erase the `NamedOpConversion` class and just use this class for the linalg generic & named ops. `inferContractionDims` should work with named ops too. But this could probably be done as a follow-up.

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
raayandhar: Yeah for now, we will check that the last dims are the ones being transposed

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
raayandhar: I added a new helper function that simplifies the entire pass logic just a little bit, what do you think?

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
raayandhar: Added tests for this

IanWood1:
@raayandhar one important difference between the reproducer and the model (CLIP) is that in the model the "rhs" is not an argument to the function. In fact, if you run the reproducer with `iree-compile reproducer.mlir --compile-to=dispatch-creation --iree-opt-level=O3` you will see that the `linalg.generic` is in the form of matmul transpose b (what we want). The transpose is fused later on in ElementwiseOpFusionPass.


I edited the reproducer to use a parameter [repro_with_param.mlir](https://gist.github.com/IanWood1/1f75871bf0cc544d93fcc05dcd965e05). Running with the same compile command, you will see that the `linalg.generic` matmul is just a normal matmul (bad) and the transpose gets hoisted into initializers. If the transpose isn't fused in PropagateLinalgTranspose, then it will be hoisted before ElementwiseOpFusion is run. But there is another case to consider, too. What if the matmul is already a matmul transpose b but there is still a transpose on the weights ([repro_with_param_mmtb.mlir](https://gist.github.com/IanWood1/0a13969bdddbae7745bdd6f1eefabcc2))? In this case, the current behavior would be better because we would want the transpose to be placed in the initializer and not fused with the matmul.

This feels like a very basic case of data tiling. Fortunately, the models we care about work well with the simple heuristics in this pass.


BTW I'm not trying to block this PR or say there is anything wrong with it. It just ports over the behavior from named ops to `linalg.generic` ops which is currently working for us. I'm just suggesting that there may be other areas to improve.

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,81 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+                                         "not a simple matmul contraction");
+    }
+
+    // Look for a transpose on any input that swaps the last two dimensions.
+    for (int64_t inputIdx = 0; inputIdx < genericOp.getNumDpsInputs();
+         ++inputIdx) {
+      auto transpose = genericOp.getDpsInputs()[inputIdx]
+                           .getDefiningOp<linalg::TransposeOp>();
+      if (!transpose || !isTransposeOfLastTwoDims(transpose.getPermutation()))
+        continue;
```
IanWood1: This won't work in the case where there is a batch dimension in in the last 2 dims. For example, if `d0` is the batch dim and the indexing map is `affine_map<(d0, d1, d2, d3) -> (d1, d3, d0)>`. I think you need to check that the last 2 dims in the indexing map are m, n, or k dims.

raayandhar: I added some changes for this, and also on the rank to simplify things

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
IanWood1: Because this now handles `linalg.generic` ops, could you rename this to something like `populateMatmulSinkingPatterns`?

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,101 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+      // In the case that the there is a batch dimension in the last two
+      // dimensions, we need to check that the last 2 dims in the indexing map
+      // are m, n, or k.
+      ArrayRef<AffineExpr> results = inputMap.getResults();
+      size_t rank = results.size();
+      if (rank < 2)
+        continue;
+      AffineDimExpr dim0 = dyn_cast<AffineDimExpr>(results[rank - 2]);
+      AffineDimExpr dim1 = dyn_cast<AffineDimExpr>(results[rank - 1]);
+      if (!dim0 || !dim1 || !mnkDims.contains(dim0.getPosition()) ||
```
IanWood1: I think it's a requirement for the map to be a projected permutation, meaning that all the results will be affine dim exprs. You could just use `cast<AffineDimExpr>` here and remove the `!dim0` checks from the if statement below.

IanWood1 (approved):
Looks good, but I'm seeing that the benchmark count is regressing in https://github.com/iree-org/iree/pull/22792. It would be good to figure that out before merging this.

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,83 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    if (!IREE::Flow::isNonNullAndOutsideDispatch(genericOp))
+      return failure();
+
+    // Only try to fuse when the generic has at least one reduction dimension.
+    // This avoids elementwise cases that are already handled elsewhere.
+    if (genericOp.getNumParallelLoops() == genericOp.getNumLoops())
+      return rewriter.notifyMatchFailure(genericOp, "not a reduction");
```
IanWood1: nit: add braces
```suggestion
    if (genericOp.getNumParallelLoops() == genericOp.getNumLoops()) {
      return rewriter.notifyMatchFailure(genericOp, "not a reduction");
    }
```

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,83 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+// operand for now.
+class FuseTransposeThroughGenericReduction
+    : public OpRewritePattern<linalg::GenericOp> {
+public:
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    if (!IREE::Flow::isNonNullAndOutsideDispatch(genericOp))
+      return failure();
```
IanWood1: Same here

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,83 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+      return false;
+  }
+  // Check that the last two dims are swapped.
+  return perm[rank - 2] == static_cast<int64_t>(rank - 1) &&
+         perm[rank - 1] == static_cast<int64_t>(rank - 2);
+}
+
+// Fuses a transpose into a reduction linalg.generic by absorbing it into the
+// indexing map. Only handles transposes that swap the last two dimensions of an
+// operand for now.
```
IanWood1: Since this now handles all reductions (not just matmuls), can we remove the `isTransposeOfLastTwoDims()` constraint?

compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -1007,6 +1007,83 @@ class NamedOpConversion : public OpRewritePattern<OpTy> {
...
+      AffineMap inputMap = genericOp.getMatchingIndexingMap(
+          genericOp.getDpsInputOperand(inputIdx));
+
+      // In the case that the there is a batch dimension in the last two
+      // dimensions, we need to check that the last 2 dims in the indexing map
+      // are m, n, or k.
+      ArrayRef<AffineExpr> results = inputMap.getResults();
+      size_t rank = results.size();
+      if (rank < 2)
+        continue;
```
IanWood1: I think this can be deleted

PR #22797, opened 2025-12-02
Adding stream.async.parameter.* ops.

benvanik:
These are equivalent to the cmd-phase ops but operate in the normal constraints of the async-phase (copy-on-write, timeline implied by use-def).

(most of this was authored with claude)


AWoloszyn (approved):

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/e2e/async_parameters.mlir
hanhanW, line 1: Do we move the pipeline test file to https://github.com/iree-org/iree/tree/main/compiler/src/iree/compiler/Dialect/Stream/Transforms/test ? It is rare to see a `e2e` under `IR/test`.

compiler/src/iree/compiler/Modules/IO/Parameters/Conversion/ParamsToVM/Patterns.cpp
hanhanW, line 1: It's not clear to me that why we need the change in this file? Is it a bug identified by claude when it implements the logic? What confuses me is that I don't see any change about `IREE::VM::CallOp`

compiler/src/iree/compiler/Dialect/Stream/Transforms/EncodeTensors.cpp
```diff
@@ -342,7 +342,7 @@ struct EncodeTensorConstantOp
              << alignedType << " does not have integral number of total bytes";
     }
     rewriter.replaceOpWithNewOp<IREE::Stream::AsyncConstantOp>(
-        op, op.getResult().getType(), encodedAttr, resultSize,
+        op, op.getResult().getType(), Value(), encodedAttr, resultSize,
```
hanhanW: It is better to add argument comment.

```suggestion
        op, op.getResult().getType(), /*await_timepoint=*/Value(), encodedAttr, resultSize,
```

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
```diff
@@ -2331,6 +2333,518 @@ void AsyncConcurrentOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+// Safely traverses through timepoint.await to find timeline-unsafe producer.
+// Only succeeds if single-use chain: producer -> await -> consumer.
+// This allows folding patterns to work with legal IR that properly awaits
+// timeline resources before consuming them.
+static Value tryLookThroughAwait(Value operand) {
+  auto awaitOp = operand.getDefiningOp<IREE::Stream::TimepointAwaitOp>();
+  if (!awaitOp)
+    return operand;
+  if (!operand.hasOneUse())
+    return operand;
```
hanhanW: style nit: add braces

```suggestion
  if (!awaitOp) {
    return operand;
  }
  if (!operand.hasOneUse()) {
    return operand;
  }
```

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
```diff
@@ -2331,6 +2333,518 @@ void AsyncConcurrentOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+  }
+  return operand;
+}
+
+// For result folding: finds consumer of timeline result through await barrier.
+// Returns {consumer_op, consumed_value} if single-use chain exists.
+static std::pair<Operation *, Value>
+findConsumerThroughAwait(Value timelineResult) {
+  if (!timelineResult.hasOneUse())
+    return {nullptr, nullptr};
```
hanhanW: style nit

```suggestion
  if (!timelineResult.hasOneUse()) {
    return {nullptr, nullptr};
  }
```

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
```diff
@@ -2331,6 +2333,518 @@ void AsyncConcurrentOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+//      : !stream.resource<constant>{%view_size}
+struct FoldAsyncParameterLoadResultSubview
+    : public OpRewritePattern<AsyncParameterLoadOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(AsyncParameterLoadOp op,
+                                PatternRewriter &rewriter) const override {
+    // Find consumer through optional await barrier (single-use chain).
+    auto [consumer, consumedValue] = findConsumerThroughAwait(op.getResult());
+    auto subviewOp =
+        dyn_cast_or_null<IREE::Stream::ResourceSubviewOp>(consumer);
```
hanhanW: style nit: prefer `dyn_cast_if_present`. I made a cleanup change recently, fyi: https://github.com/iree-org/iree/commit/33db935a33ddab6a8358418c0ae6dc263e56b0bb


```suggestion
    auto subviewOp =
        dyn_cast_if_present <IREE::Stream::ResourceSubviewOp>(consumer);
```

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOps.td
```diff
@@ -2937,6 +2978,385 @@ def Stream_AsyncConcurrentOp : Stream_Op<"async.concurrent", [
...
+    ```
+  }];
+
+  let arguments = (ins
+    Optional<Stream_Timepoint>:$await_timepoint,
+    // TODO(benvanik): change this to an attribute interface so we can support
+    // dynamic parameter selection in the future.
+    OptionalAttr<StrAttr>:$source_scope,
+    StrAttr:$source_key,
+    I64:$source_offset,
```
hanhanW, line 3034: Why is it I64, but not index type?

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOps.td
```diff
@@ -2937,6 +2978,385 @@ def Stream_AsyncConcurrentOp : Stream_Op<"async.concurrent", [
...
+    Loads a resource directly from a parameter archive file (or in-memory
+    buffer) at the specified byte offset. The loaded resource is allocated with
+    the specified size and populated with data from the parameter source. Note
+    that load operations _may_ become direct access/memory mapped resources if
+    the result resource type allows.
+
+    When no await_timepoint is specified the load may be hoisted and begin at
+    any point prior to the first use. When an await_timepoint is specified the
+    load will not begin until that timepoint is reached, allowing for explicit
+    ordering with other timeline operations.
```
hanhanW: I don't really follow what the difference is between the load op and the read op. If I read it correctly, the difference is around allocation? This is the other confusion that I have: why does load/read associate to an allocation?

hanhanW (commented):
I only skim through some files and ask few questions. It'd be easier if you split it into few commits/PRs. E.g.,

- Introduce ops
- Add folders
- Add relevant conversion/transformation and pipeline tests.

compiler/src/iree/compiler/Modules/IO/Parameters/Conversion/ParamsToVM/Patterns.cpp
benvanik: It has always been wrong, and never has been used. I don't know what you mean about not changing the call - this is changing the call.

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOps.td
benvanik: File offsets are always i64, as index may be 32-bits. This matches filesystem APIs (which is what this is).

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOps.td
benvanik: Reads read into an existing allocation, loads return a new allocation. This is structural (read does not return a resource and take a resource to write into, while load returns a new resource). The description notes why you may want to do this: direct access without copies/memory mapping without wiring. In unified memory situations this can be useful for staging (so you don't allocate staging memory, read, then allocate device memory, copy - you can just directly hit the memory). Use is situational, just like memory mapping/import in general.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/e2e/async_parameters.mlir
benvanik: Good Q! It really should have been Transforms/test/e2e/. The Transforms/test/e2e/ directory is intentionally for integration-like tests that span multiple pipeline stages or pipelines, as opposed to Transforms/test/ which tests individual passes in isolation. This test uses --iree-stream-transformation-pipeline to validate the full pipeline behavior (partitioning, scheduling, allocation, timepoint optimization) for the new parameter ops working together.

Added a comment to the test/e2e/ build file indicating the purpose.

PR #22800, opened 2025-12-02
[Codegen] Support dynamic offsets in collapse_shape fusion to interface stores

qedawkins:
Previously, `FoldCollapseShapeIntoInterfaceTensorStore` only handled static offsets when fusing `tensor.collapse_shape` into dispatch tensor stores. This change extends the pattern to also accept dynamic offsets when:

1. The offset comes from an `affine.apply` operation
2. The affine expression is provably divisible by the inner dimension product

This also fixes an issue where offset computation was done at the subspan insertion point, which could fail when the offset value is defined in a nested region (e.g., inside a loop). The offset affine.apply is now created at the store's insertion point where dominance is guaranteed.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
@@ -600,6 +600,16 @@ struct FoldCollapseShapeIntoInterfaceTensorStoreFullSlice
 ///       offsets = [%x * 286, 0, 0, 0], sizes = [3, 3, 1, 96]
 ///       strides = [1, 1, 1, 1] : tensor<3x3x1x96xf32> ->
 ///       !iree_tensor_ext.dispatch.tensor<writeonly:tensor<9x3x1x96xf32>>
+///
+/// Dynamic offsets are supported if they come from an affine.apply operation
+/// where the result is provably divisible by the inner dimension product. For
+/// example, if the inner dimensions are [4, 128] (product = 512), an offset of
+/// `affine_map<()[s0] -> (s0 * 512)>` can be transformed to `s0`.
+///
+/// The divisibility check handles Add and Mul expressions recursively:
+/// - For Mul: divisible if either operand is divisible (pessimistic)
+/// - For Add: divisible if both operands are divisible
+/// - For constants: divisible if value % innerDimProduct == 0
 struct FoldCollapseShapeIntoInterfaceTensorStore
     : OpRewritePattern<IREE::TensorExt::DispatchTensorStoreOp> {
   using OpRewritePattern<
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -820,6 +881,21 @@ struct FoldCollapseShapeIntoInterfaceTensorStore
     dispatchIndexOpFoldResults(expandedSizes, expandedDynamicDims,
                                expandedStaticDims);
     rewriter.setInsertionPoint(storeOp);
+
+    // Now compute the expanded offsets at the store's insertion point where the
+    // offset values are guaranteed to dominate.
+    SmallVector<OpFoldResult> expandedOffsets;
+    for (const auto &info : offsetInfos) {
```
jtuyls:
```suggestion
    for (const OffsetInfo &info : offsetInfos) {
```

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -688,9 +669,87 @@ struct FoldCollapseShapeIntoInterfaceTensorStore
...
       }
-      // In case of a single dynamic dimension, check that the subspan size is
-      // dynamic as well.
-      if (numDynamicDims == 1) {
+
+      // Compute the inner dimension product. This requires all inner dims to
+      // be static for the offset divisibility and subspan size checks below.
+      int64_t innerDimSize = 1;
+      bool hasStaticInnerDims = true;
+      for (auto j : llvm::drop_begin(group)) {
```
jtuyls:
```suggestion
      for (int64_t j : llvm::drop_begin(group)) {
```

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
jtuyls:
```suggestion
    for (const OffsetInfo &info : offsetInfos) {
```

compiler/src/iree/compiler/Codegen/Common/ReshapePatterns.cpp
```diff
@@ -717,17 +772,25 @@ struct FoldCollapseShapeIntoInterfaceTensorStore
     rewriter.setInsertionPoint(subspanOp);
     Location loc = collapseShape.getLoc();
     SmallVector<int64_t> expandedSubspanShape;
-    SmallVector<OpFoldResult> expandedOffsets;
     SmallVector<OpFoldResult> expandedSizes;
     SmallVector<Value> newSubspanDynamicDims;
+    // Store offset info for deferred computation. The offset value may be
+    // defined in a nested region that doesn't dominate the subspan op, so we
```
jtuyls, line 778: I think it would be useful to add a testcase for this as well or update one of the existing ones.

jtuyls (approved):
LGTM, thanks!

Issue #22121, opened 2025-09-26
[DT][gfx950] Bring up llama8b _fp8 on mi350


Abhishek-Varma:
1. Take the input IR from [here](https://gist.github.com/Abhishek-Varma/80a54ed0d159b8d3419c655ec80a5d30).
2. Run the following command with IREE commit `952da4192af` :-
```
iree-compile ~/input.mlir \
--iree-hal-target-backends=rocm --iree-hip-target=gfx950 --iree-hal-target-device=hip \
--iree-opt-level=O3 --iree-dispatch-creation-propagate-collapse-across-expands=true \
--iree-codegen-enable-default-tuning-specs=true --iree-hip-enable-tensor-ukernels \
--iree-hal-indirect-command-buffers=true --iree-stream-resource-memory-model=discrete \
--iree-hip-specialize-dispatches --iree-hal-memoization=true --iree-opt-data-tiling=false \
--iree-dispatch-creation-data-tiling --iree-hip-encoding-layout-resolver=data-tiling 
```

The above gives the following error as part of `GPUDistributePass` :-
```
error: unsupported non-normalized loops
```

NOTE: The input dispatch was obtained via Llama 8B f8e4m3fn on gfx950 - after disabling a check which prevents fusion of encoding ops for producers with multiple uses.

Steps to resolve :-
1. Replace `GPUDistributePass` [here](https://github.com/iree-org/iree/blob/5bf506f1410dcd36b090e7feac3fe9a74f4bf6ee/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L313) with `GPUDistributeForAllPass`.
2. Add `NormalizeLoopBoundsPass` normalizing `scf.forall` right above the aforementioned pass.

NOTE: Both resolution steps above are needed and it won't work with just step 2. Else it'd lead to :-
```
requires statically sized, normalized forall op
```

Now, the above resolution works because the input dispatch goes through `GPUVectorization` pipeline instead of `GPUTileAndFuse` pipeline. (CC: @Max191 )

But ideally we want the above dispatch to indeed go through `GPUTileAndFuse` pipeline. I tried hardcoding the pipeline to `GPUTileAndFuse` to see what happens and we get the following error :-
```
'vector.scatter' op write affecting operations on shared resources are restricted to lane or thread distributed contexts.
```

I'm adding this as an issue to drive the discussion on what the expectation for handling this dispatch is - can accordingly work towards the fix we all converge on.

[...]

PR #22801, opened 2025-12-02
[GPU] Add M dimension constraints for pingpong ukernel

### LINKED ISSUE #22799: [GPU][DT] `--iree-hip-enable-tensor-ukernels` is pessimizing llama performance
The [llama data tiling benchmarks](https://github.com/iree-org/iree/blob/516031078c9bc188111a27fc3faf43e13827a9e4/tests/external/iree-test-suites/torch_models/llama_8b_fp8/modules/llama_gfx942_data_tiling.json) see a dramatic performance improvement if `--iree-hip-enable-tensor-ukernels` is turned off. Locally, I found that decode went from 58ms to 23ms.

There are some dispatches that get very large padding sizes when `--iree-hip-enable-tensor-ukernels` is enabled. There is a reduction dispatch that takes the largest time for decode. In this instance, I found that it was packing/padding `4x4096` (16384 elems) to `1x128x256x32` (1048576 elems) which is a 64x size increase. With the option disabled it packs/pads `4x4096` (16384) to `1x64x16x64` (65536) which is only 4x increase.

Here is the IR/logs for the example above:

- `--iree-hip-enable-tensor-ukernels=true`
	- [executable source](https://gist.github.com/IanWood1/78d1e0a8f919352b3bfe212774cdd923)
	- [mlir dump](https://gist.github.com/IanWood1/940ac05777943b4c7255f141319c1635)
- `--iree-hip-enable-tensor-ukernels=false`
	- [executable source](https://gist.github.com/IanWood1/3cdcf2627edbefb4c3f8997f0707d5a7)
	- [mlir dump](https://gist.github.com/IanWood1/984acbeada54106c026b5afd8e3988d5)

I think the difference comes from the function `chooseDataTiledMMAAttr()`. When ukernels are enabled, [this logic](https://github.com/iree-org/iree/blob/516031078c9bc188111a27fc3faf43e13827a9e4/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp#L115-L123) will find a size that matches ukernel. Apparently, this will cause much more padding. With it disabled, it will instead of limit the excessive padding [here](https://github.com/iree-org/iree/blob/516031078c9bc188111a27fc3faf43e13827a9e4/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp#L250-L268). I verified this by commenting out the ukernel provider lines and running with ukernels enabled.

I don't think disabling `--iree-hip-enable-tensor-ukernels` is the fix here. Instead, it is exposing some underlying issue with ukernels/data tiling.

I'm not exactly sure why, but this is blocking https://github.com/iree-org/iree/pull/22011 which regresses llama decode performance only when ukernels are enabled. Otherwise, it also achieves 22ms.
IREE @ d2117d759174cd3e5118debc7f78b3eb9b6631ad

IanWood1:
The pingpong_dt_large_f8E4M3FNUZ and pingpong_dt_medium_f8E4M3FNUZ ukernels had no constraints on the M dimension, causing them to match matmuls with very small M sizes. This resulted in extreme padding overhead for small M sizes. This change adds constraints for the "large" and "medium" ukernels:

- pingpong_dt_large_f8E4M3FNUZ: Added M_min = 64 (≥25% utilization for M_tile=256)
- pingpong_dt_medium_f8E4M3FNUZ: Added M_min = 32 (≥25% utilization for M_tile=128)

This roughly thirds the runtime for data tiled llama decode and I have decreased the golden times accordingly. I also confirmed that the increase in dispatches is only from initializers.

closes https://github.com/iree-org/iree/issues/22799

ci-extra: test_torch

### DIFF
--- a/compiler/plugins/target/ROCM/test/materialize_encoding_ukernel_gfx942.mlir
+++ b/compiler/plugins/target/ROCM/test/materialize_encoding_ukernel_gfx942.mlir
@@ -170,3 +170,124 @@ func.func @matmul_f8_f8_f32_large_lowering_ukernel_provider(%arg0: tensor<?x?xf8
 // CHECK:      iree_codegen.inner_tiled
 // CHECK-SAME:     iterator_types = [#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>]
 // CHECK-SAME:     kind = #iree_gpu.data_tiled_mma_layout<intrinsic = MFMA_F32_16x16x32_F8E4M3FNUZ,  intrinsics_m = 8, subgroups_m = 2, intrinsics_n = 4, subgroups_n = 4, operands_interleaving_intrinsics_k = [0, 1]>
+
+// -----
+
+// Test that with M=4, no ukernel matches due to the M dimension constraint,
+// so it falls back to the default heuristic which should produce much smaller
+// tile sizes to avoid excessive padding.
+
+#executable_target_rocm_hsaco_fb = #hal.executable.target<"rocm", "rocm-hsaco-fb", {
+  abi = "hip",
+  iree.encoding.resolver = #iree_gpu.gpu_encoding_resolver<>,
+  iree_codegen.target_info = #iree_gpu.target<
+    arch = "gfx942",
+    features = "",
+    wgp = <compute =  fp64|fp32|fp16|int64|int32|int16|int8,
+    storage =  b64|b32|b16|b8,
+    subgroup =  shuffle|arithmetic,
+    dot =  dp4xi8toi32,
+    mma = [<MFMA_F32_16x16x16_BF16>, <MFMA_F32_32x32x8_BF16>, <MFMA_F32_16x16x32_F8E5M2FNUZ>,
+           <MFMA_F32_16x16x32_F8E5M2FNUZ_F8E4M3FNUZ>, <MFMA_F32_16x16x32_F8E4M3FNUZ>,
+           <MFMA_F32_16x16x32_F8E4M3FNUZ_F8E5M2FNUZ>, <MFMA_F32_32x32x16_F8E5M2FNUZ>,
+           <MFMA_F32_32x32x16_F8E5M2FNUZ_F8E4M3FNUZ>, <MFMA_F32_32x32x16_F8E4M3FNUZ>,
+           <MFMA_F32_32x32x16_F8E4M3FNUZ_F8E5M2FNUZ>, <MFMA_I32_16x16x32_I8>,
+           <MFMA_I32_32x32x16_I8>, <MFMA_F64_16x16x4_F64>, <MFMA_F32_16x16x4_F32>,
+           <MFMA_F32_16x16x16_F16>, <MFMA_F32_32x32x8_F16>
+          ],
+    subgroup_size_choices = [64],
+    max_workgroup_sizes = [1024, 1024, 1024],
+    max_thread_count_per_workgroup = 1024,
+    max_workgroup_memory_bytes = 65536,
+    max_workgroup_counts = [2147483647, 2147483647, 2147483647],
+    max_load_instruction_bits = 128,
+    simds_per_wgp = 4,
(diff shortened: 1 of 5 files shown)


Yu-Zhewen (approved):
Thanks for the fix! Yeah it makes complete sense to set a lower bound for M, as the ukernels here were originally designed for prefill, not decode.

Regarding the change in dispatch count, this also aligns with what @Abhishek-Varma observed (though his investigation was on gfx950 rather than gfx942):

https://github.com/iree-org/iree/issues/22121#issuecomment-3563023732

jtuyls (approved):
LGTM, thanks!

compiler/plugins/target/ROCM/test/materialize_encoding_ukernel_gfx942.mlir
```diff
@@ -170,3 +170,124 @@ func.func @matmul_f8_f8_f32_large_lowering_ukernel_provider(%arg0: tensor<?x?xf8
...
 // CHECK-SAME:     iterator_types = [#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>]
 // CHECK-SAME:     kind = #iree_gpu.data_tiled_mma_layout<intrinsic = MFMA_F32_16x16x32_F8E4M3FNUZ,  intrinsics_m = 8, subgroups_m = 2, intrinsics_n = 4, subgroups_n = 4, operands_interleaving_intrinsics_k = [0, 1]>
+
+// -----
+
+// Test that with M=4, no ukernel matches due to the M dimension constraint,
+// so it falls back to the default heuristic which should produce much smaller
+// tile sizes to avoid excessive padding.
+
+#executable_target_rocm_hsaco_fb = #hal.executable.target<"rocm", "rocm-hsaco-fb", {
```
kuhar: Would this test work without splits, with multiple functions per module, so that all these verbose attributes are shared?

Max191 (approved):

compiler/plugins/target/ROCM/test/materialize_encoding_ukernel_gfx942.mlir
IanWood1: Yes, it works. The only catch is that the encoding attrs need to be renamed to be unique. I pushed a commit to this PR to make that change. Let me know if it seems worth it.

compiler/plugins/target/ROCM/test/materialize_encoding_ukernel_gfx942.mlir
```diff
@@ -36,137 +40,94 @@
...
-#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
-#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
-#encoding_lhs = #iree_encoding.encoding<operand_index = 0, op_type = matmul, element_types = [f8E4M3FNUZ, f8E4M3FNUZ, f32], user_indexing_maps = [#map, #map1, #map2], iteration_sizes = [?, 2048, ?]>
-#encoding_rhs = #iree_encoding.encoding<operand_index = 1, op_type = matmul, element_types = [f8E4M3FNUZ, f8E4M3FNUZ, f32], user_indexing_maps = [#map, #map1, #map2], iteration_sizes = [?, 2048, ?]>
-#encoding_result = #iree_encoding.encoding<operand_index = 2, op_type = matmul, element_types = [f8E4M3FNUZ, f8E4M3FNUZ, f32], user_indexing_maps = [#map, #map1, #map2], iteration_sizes = [?, 2048, ?]>
-#pipeline_layout_3 = #hal.pipeline.layout<constants = 2, bindings = [
-  #hal.pipeline.binding<storage_buffer>,
-  #hal.pipeline.binding<storage_buffer>,
-  #hal.pipeline.binding<storage_buffer>
-]>
```
IanWood1, line 160: Note that I removed all the `#hal.pipeline.layout` attrs since they were not being used.

kuhar (approved):

Issue #22795, opened 2025-12-02
[GPU] Direct convolution failed to FuseAndHoistParallelLoops


yzhang93:
### What happened?

A forward convolution failed lowering through `--iree-codegen-llvmgpu-use-direct-convolution` with error:
`uses 155648 bytes of shared memory; exceeded the limit of 65536 bytes`.

The [dump IR](https://gist.github.com/yzhang93/b033520c59b0912705cd4855316414de) shows that the parallel loops were not hoisted successfully.

### Steps to reproduce your issue

Input IR:

```
#map = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7) -> (d0, d1 + d5, d2 + d6, d3, d7)>
#map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7) -> (d3, d4, d5, d6, d7)>
#map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7) -> (d0, d1, d2, d3, d4)>
#map3 = affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>
module {
  util.func public @fused_op_layout_cus_2171b288a05a4772972d26625f13c0cc74aedf80_128x24x48x384xbfloat16_384x3x1x128xbfloat16$async(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.fence, %arg3: !hal.fence) -> !hal.buffer_view attributes {inlining_policy = #util.inline.never, iree.abi.model = "coarse-fences", iree.abi.stub} {
    %cst = arith.constant 0.000000e+00 : f32
    %cst_0 = arith.constant 0.000000e+00 : bf16
    %0 = hal.tensor.import wait(%arg2) => %arg0 : !hal.buffer_view -> tensor<128x24x48x384xbf16>
    %1 = hal.tensor.import wait(%arg2) => %arg1 : !hal.buffer_view -> tensor<384x3x1x128xbf16>
    %expanded = tensor.expand_shape %0 [[0], [1], [2], [3, 4]] output_shape [128, 24, 48, 3, 128] : tensor<128x24x48x384xbf16> into tensor<128x24x48x3x128xbf16>
    %expanded_1 = tensor.expand_shape %1 [[0, 1], [2], [3], [4]] output_shape [3, 128, 3, 1, 128] : tensor<384x3x1x128xbf16> into tensor<3x128x3x1x128xbf16>
[...]

PR #22802, opened 2025-12-02
[LLVMGPU] Fix lowering strategy for direct convolution

yzhang93:
This fixes issue https://github.com/iree-org/iree/issues/22795.

When the convolution filter size is 3x1, the unit filter dimension is folded and the original innermost output image dimension now becomes a batch dimension.  Without this change, it tiles the innermost output image dimension to 1 while keeping the outermost dimension as non-unit, and caused bad code after PackToIntrinsics.


nirvedhmeshram (approved):
Thanks! just curious do you have plans to turn this on by default in certain situations?

yzhang93:
> Thanks! just curious do you have plans to turn this on by default in certain situations?

I have started a benchmark sweep with direct convolution, but there are some issues. Also it doesn't work for weight backwards currently. Let me fix those and re-evaluate and decide.

PR #22338, opened 2025-10-16
[Codegen] PCF Working Branch

qedawkins:
This introduces a new dialect for "Parallel Control Flow" (pcf a la scf). This is motivated by expressivity gaps in the current constructs we use to represent distributed compute for the next set of algorithms we're working towards supporting in the compiler. To name a few:
 - Stream-K
 - Warp/Clustered reductions
 - Named Barriers
 - Improved robustness of fusion support
 - Persistent-ish kernels (things that need to communicate across wkgps)

Naturally the base added here does not directly cover all of the above but should make it much easier to do so. The core idea behind the dialect hinges on three concepts:
1. Shaped references of unspecified layout and memspace
2. Execution scopes which threads are forked out of
3. Synchronization scopes which control access to associated memory

The dialect includes further documentation behind the above 3 points.

The initial proposal can be found here: https://hackmd.io/iHdqgU7MQnSvX6H_y_6JrQ (Much of the naming/representation has changed since this writeup)

PR #22804, opened 2025-12-03
[Codegen] Add PCF dialect

qedawkins:
This introduces a new dialect for "Parallel Control Flow" (named a la scf). This is motivated by expressivity gaps in the current constructs we use to represent distributed compute for the next set of algorithms we're working towards supporting in the compiler. To name a few:
 - Stream-K
 - Warp/Clustered reductions
 - Named Barriers
 - Improved robustness of fusion support
 - Persistent-ish kernels (things that need to communicate across wkgps)

Naturally the base added here does not directly cover all of the above but should make it much easier to do so. The core idea behind the dialect hinges on three concepts:
1. Shaped references of unspecified layout and memspace
2. Execution scopes which threads are forked out of
3. Synchronization scopes which control access to associated memory

The dialect includes further documentation behind the above 3 points.

The initial proposal can be found here: https://hackmd.io/iHdqgU7MQnSvX6H_y_6JrQ (Much of the naming/representation has changed since this writeup)


qedawkins:
This is now ready for review. This only includes the dialect skeleton. The full PR chain can be found here: #22338

Requoting from that PR:

> This was split into 4 pull requests: https://github.com/iree-org/iree/pull/22804, https://github.com/iree-org/iree/pull/22805, https://github.com/iree-org/iree/pull/22806, and https://github.com/iree-org/iree/pull/22807. This PR is now defunct and will only be used for CI.
>
> Depends on https://github.com/iree-org/iree/pull/22533, https://github.com/iree-org/iree/pull/22549, and https://github.com/iree-org/iree/pull/22800, as well as two pending LLVM changes.

When reviewing if there is missing usage/context in this PR, it might be addressed in one of the later commits, or feel free to just ask and I can hopefully point at it. I tried to split up the commit into more or less reviewable chunks but the chunks still ended up being fairly large, so I'm happy to take further requests to break down the change into smaller ones (especially the one introducing the major lowering passes/transforms).

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+
+  let assemblyFormat = [{
+    `(`$dynamicSizes`)` attr-dict `:` type($result)
+  }];
+
+  let extraClassDeclaration = [{
+    ShapedRefType getResultType() {
+      return cast<ShapedRefType>(getResult().getType());
+    }
+    SmallVector<OpFoldResult> getMixedSizes() {
```
kuhar: Nit: make this a declaration and move the implementation to the .cpp?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+       RecursiveMemoryEffects,
+       DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmBlockArgumentNames"]>,
+     ]> {
+  let summary = [{
+    Execute a set of workers in parallel on a region.
+  }];
+  let description = [{
+    Executes a region across a set of workers at a specified scope. When
+    control flow reaches this op, workers with count equal to the native
+    parallelism of the scope are spawned and begin executing the region. The
```
kuhar: I can't parse this sentence. Do you mean that `nproc` workers will be spawned, or that some workers from a larger pool whose count is `nproc` will be spawned?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+    2. Allocated by the op itself - dynamic sizes must be provided for
+       untied results with dynamic dimensions.
+
+    Basic example with tied results:
+    ```mlir
+      %0 = ... : tensor<4x8xf32>
+      %1 = pcf.generic scope(#foo.scope)
+        execute(%ref = %0)[%id: index, %num_workers: index]
+             : (!pcf.sref<4x8xf32, #foo.scope>) -> (tensor<4x8xf32>) {
+        // Each worker can read/write %ref
```
kuhar:
```suggestion
        // Each worker can read/write %ref.
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+    1. Tied to initial values (tensor or memref) - the init value provides the
+       initial contents and the result captures the final state.
+    2. Allocated by the op itself - dynamic sizes must be provided for
+       untied results with dynamic dimensions.
+
+    Basic example with tied results:
+    ```mlir
+      %0 = ... : tensor<4x8xf32>
+      %1 = pcf.generic scope(#foo.scope)
+        execute(%ref = %0)[%id: index, %num_workers: index]
```
kuhar, line 122: Why is the type of %0 omitted here? Is it always the same as the result type?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+    Example with initializer:
+    ```mlir
+      %result = pcf.generic scope(#foo.scope)
+        initialize {
+          %scratch = pcf.alloc() : !pcf.sref<16xf32, #foo.scope>
+          pcf.yield %scratch : !pcf.sref<16xf32, #foo.scope>
+        } -> (%scratch_arg: !pcf.sref<16xf32, #foo.scope>)
+        execute(%ref = %init)[%id: index, %num_workers: index]
+             : (!pcf.sref<4x8xf32, #foo.scope>) -> (tensor<4x8xf32>) {
+        // %scratch_arg is available here, initialized once
```
kuhar:
```suggestion
        // %scratch_arg is available here, initialized once.
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+      }
+    ```
+
+    Example with untied (allocated) results:
+    ```mlir
+      %d0, %d1 = ... : index
+      %result = pcf.generic scope(#foo.scope)
+        execute[%id: index, %num_workers: index]
+             : () -> (tensor<?x?xf32>{%d0, %d1}) {
+        // Result sref is allocated by the op, not tied to any init
```
kuhar:
```suggestion
        // Result sref is allocated by the op, not tied to any init.
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+                   CArg<"bool", "false">:$sync_on_return)>,
+  ];
+
+  let extraClassDeclaration = [{
+    int64_t getNumIterators() {
+      return getNumIndexArgs() / 2;
+    }
+    MutableArrayRef<BlockArgument> getLeadingArgs() {
+      return getRegion().getArguments().take_front(getNumLeadingArgs());
+    }
```
kuhar: Err, should we get mutable and non-mutable variants? Making the default mutable seems like a potential footgun.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+    MutableArrayRef<BlockArgument> getIdAndCountArgs() {
+      return getRegion().getArguments().take_back(getNumIndexArgs());
+    }
+    MutableArrayRef<BlockArgument> getRegionRefArgs() {
+      return getRegion().getArguments().drop_front(getNumLeadingArgs()).take_front(getNumResults());
+    }
+
+    bool isRegionRefArg(BlockArgument b) {
+      assert(b.getOwner() == &getRegion().front() && "unexpected non-entry block arg");
```
kuhar: Move longer methods to .cpp?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+       untied results with dynamic dimensions.
+
+    Basic example with 1D iteration:
+    ```mlir
+      %n = ... : index
+      %0 = ... : tensor<4x8xf32>
+      %1 = pcf.loop scope(#foo.scope) count(%n)
+        execute(%ref = %0)[%id: index]
+             : (!pcf.sref<4x8xf32, #foo.scope>) -> (tensor<4x8xf32>) {
+        // %id ranges from 0 to %n-1
```
kuhar:
```suggestion
        // %id ranges from 0 to %n-1.
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+      }
+    ```
+
+    Example with multi-dimensional iteration:
+    ```mlir
+      %m, %n = ... : index
+      %result = pcf.loop scope(#foo.scope) count(%m, %n)
+        execute(%ref = %init)[%i: index, %j: index]
+             : (!pcf.sref<?x?xf32, #foo.scope>) -> (tensor<?x?xf32>) {
+        // %i ranges from 0 to %m-1, %j ranges from 0 to %n-1
```
kuhar:
```suggestion
        // %i ranges from 0 to %m-1, %j ranges from 0 to %n-1.
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+      return getCount().size();
+    }
+    MutableArrayRef<BlockArgument> getIdArgs() {
+      return getRegion().getArguments().take_back(getNumIdArgs());
+    }
+    MutableArrayRef<BlockArgument> getRegionRefArgs() {
+      return getRegion().getArguments().take_front(getNumResults());
+    }
+
+    SmallVector<int64_t> getInitTiedResultIndices() {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+                   "ValueRange":$count, "ValueRange":$inits,
+                   "ValueRange":$dynamic_sizes, "ArrayRef<bool>":$is_tied,
+                   CArg<"bool", "false">:$sync_on_return)>,
+  ];
+
+  let extraClassDeclaration = [{
+    int64_t getNumIdArgs() {
+      return getCount().size();
+    }
+    MutableArrayRef<BlockArgument> getIdArgs() {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFTypes.cpp
```diff
@@ -0,0 +1,144 @@
...
+Type ShapedRefType::parse(AsmParser &parser) {
+  if (parser.parseLess()) {
+    return {};
+  }
+
+  llvm::SmallVector<int64_t> shape;
+  mlir::Type elementType;
+  Attribute scope;
+
+  auto shapeLoc = parser.getCurrentLocation();
```
kuhar: nit: spell out the type. Also below.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFTypes.cpp
```diff
@@ -0,0 +1,144 @@
...
+// #pcf.sref<...>
+//===----------------------------------------------------------------------===//
+
+Type ShapedRefType::parse(AsmParser &parser) {
+  if (parser.parseLess()) {
+    return {};
+  }
+
+  llvm::SmallVector<int64_t> shape;
+  mlir::Type elementType;
```
kuhar:
```suggestion
  SmallVector<int64_t> shape;
  Type elementType;
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFTypes.cpp
```diff
@@ -0,0 +1,144 @@
...
+  Attribute scope;
+
+  auto shapeLoc = parser.getCurrentLocation();
+  if (mlir::failed(parser.parseDimensionList(shape))) {
```
kuhar:
```suggestion
  if (failed(parser.parseDimensionList(shape))) {
```
also elsewhere below

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFTypes.cpp
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+  }
+
+  if (parser.parseGreater()) {
+    return {};
+  }
+
+  MLIRContext *context = parser.getContext();
+  return ShapedRefType::get(context, shape, elementType, cast<ScopeAttr>(scope),
```
kuhar: nit
```suggestion
  return ShapedRefType::get(parser.getContext(), shape, elementType, cast<ScopeAttr>(scope),
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFTypes.cpp
```diff
@@ -0,0 +1,144 @@
...
+                            syncScope);
+}
+
+void ShapedRefType::print(AsmPrinter &printer) const {
+  printer << "<";
+
+  auto shape = getShape();
+  for (int64_t dim : shape) {
+    if (mlir::ShapedType::isDynamic(dim))
```
kuhar: also here: we don't need the mlir namespace prefix

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFTypes.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+  MLIRContext *context = parser.getContext();
+  return ShapedRefType::get(context, shape, elementType, cast<ScopeAttr>(scope),
+                            syncScope);
+}
+
+void ShapedRefType::print(AsmPrinter &printer) const {
+  printer << "<";
+
+  auto shape = getShape();
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+    auto shapedType = cast<ShapedType>(resultType);
+    entryBlock.addArgument(
+        PCF::ShapedRefType::get(b.getContext(), shapedType.getShape(),
+                                shapedType.getElementType(), scope),
+        result.location);
+  }
+
+  // Thread count args.
+  Type indexType = b.getIndexType();
+  int64_t numCountArgs = count.size() == 0 ? 1 : count.size();
```
kuhar:
```suggestion
  int64_t numCountArgs = count.empty() ? 1 : count.size();
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+// Build a WriteSliceOp with dynamic entries.
+void WriteSliceOp::build(OpBuilder &b, OperationState &result, Value source,
+                         Value dest, ValueRange offsets, ValueRange sizes,
+                         ValueRange strides, ArrayRef<NamedAttribute> attrs) {
+  SmallVector<OpFoldResult> offsetValues = llvm::to_vector<4>(
+      llvm::map_range(offsets, [](Value v) -> OpFoldResult { return v; }));
+  SmallVector<OpFoldResult> sizeValues = llvm::to_vector<4>(
+      llvm::map_range(sizes, [](Value v) -> OpFoldResult { return v; }));
+  SmallVector<OpFoldResult> strideValues = llvm::to_vector<4>(
+      llvm::map_range(strides, [](Value v) -> OpFoldResult { return v; }));
```
kuhar: Use map_to_vector. You can do `auto offsetValues = llvm::map_to_vector(offsets, llvm::StaticCastTo<OpFoldResult>);`

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+
+void ReadSliceOp::build(OpBuilder &b, OperationState &result, Type resultType,
+                        Value source, ValueRange offsets, ValueRange sizes,
+                        ValueRange strides, ArrayRef<NamedAttribute> attrs) {
+  SmallVector<OpFoldResult> offsetValues = llvm::to_vector<4>(
+      llvm::map_range(offsets, [](Value v) -> OpFoldResult { return v; }));
+  SmallVector<OpFoldResult> sizeValues = llvm::to_vector<4>(
+      llvm::map_range(sizes, [](Value v) -> OpFoldResult { return v; }));
+  SmallVector<OpFoldResult> strideValues = llvm::to_vector<4>(
+      llvm::map_range(strides, [](Value v) -> OpFoldResult { return v; }));
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+
+void GetMemrefOp::build(OpBuilder &b, OperationState &result, Type resultType,
+                        Value source, ValueRange offsets, ValueRange sizes,
+                        ValueRange strides, ArrayRef<NamedAttribute> attrs) {
+  SmallVector<OpFoldResult> offsetValues = llvm::to_vector<4>(
+      llvm::map_range(offsets, [](Value v) -> OpFoldResult { return v; }));
+  SmallVector<OpFoldResult> sizeValues = llvm::to_vector<4>(
+      llvm::map_range(sizes, [](Value v) -> OpFoldResult { return v; }));
+  SmallVector<OpFoldResult> strideValues = llvm::to_vector<4>(
+      llvm::map_range(strides, [](Value v) -> OpFoldResult { return v; }));
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+
+OpFoldResult ReadSliceOp::fold(FoldAdaptor adaptor) {
+  bool changed = false;
+  OpBuilder builder(getContext());
+
+  SmallVector<OpFoldResult> mixedOffsets = getMixedOffsets();
+  SmallVector<OpFoldResult> mixedStrides = getMixedStrides();
+
+  // Try to fold dynamic offsets to static.
+  for (auto &offset : mixedOffsets) {
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+      std::optional<int64_t> constantValue = getConstantIntValue(value);
+      if (constantValue.has_value()) {
+        offset = builder.getI64IntegerAttr(*constantValue);
+        changed = true;
+      }
+    }
+  }
+
+  // Try to fold dynamic strides to static.
+  for (auto &stride : mixedStrides) {
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+  OpBuilder builder(getContext());
+
+  SmallVector<OpFoldResult> mixedOffsets = getMixedOffsets();
+  SmallVector<OpFoldResult> mixedStrides = getMixedStrides();
+
+  // Try to fold dynamic offsets to static.
+  for (auto &offset : mixedOffsets) {
+    if (auto value = dyn_cast<Value>(offset)) {
+      std::optional<int64_t> constantValue = getConstantIntValue(value);
+      if (constantValue.has_value()) {
```
kuhar:
```suggestion
      if (std::optional<int64_t> constantValue = getConstantIntValue(value)) {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+        changed = true;
+      }
+    }
+  }
+
+  // Try to fold dynamic strides to static.
+  for (auto &stride : mixedStrides) {
+    if (auto value = dyn_cast<Value>(stride)) {
+      std::optional<int64_t> constantValue = getConstantIntValue(value);
+      if (constantValue.has_value()) {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+
+OpFoldResult GetMemrefOp::fold(FoldAdaptor adaptor) {
+  bool changed = false;
+  OpBuilder builder(getContext());
+
+  SmallVector<OpFoldResult> mixedOffsets = getMixedOffsets();
+  SmallVector<OpFoldResult> mixedStrides = getMixedStrides();
+
+  // Try to fold dynamic offsets to static.
+  for (auto &offset : mixedOffsets) {
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+    if (auto value = dyn_cast<Value>(offset)) {
+      std::optional<int64_t> constantValue = getConstantIntValue(value);
+      if (constantValue.has_value()) {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,952 @@
...
+        changed = true;
+      }
+    }
+  }
+
+  // Try to fold dynamic strides to static.
+  for (auto &stride : mixedStrides) {
+    if (auto value = dyn_cast<Value>(stride)) {
+      std::optional<int64_t> constantValue = getConstantIntValue(value);
+      if (constantValue.has_value()) {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,894 @@
...
+//===----------------------------------------------------------------------===//
+// AllocOp
+//===----------------------------------------------------------------------===//
+
+def AllocOp : PCF_Op<"alloc", []> {
+  let summary = [{
+    Shaped ref allocation operation
+  }];
+  let description = [{
+    Allocates a `pcf.sref` with the given element type and shape. Dynamic
```
kuhar, line 41: Is it `pcf.sref` or `pcf.shaped_ref`?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
qedawkins: yes, it's always the same as the result type so in this case the type is `tensor<4x8xf32>`. It's similar to the shared_outs/init syntax on scf.forall.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
qedawkins: Removed all instances of `pcf.shaped_ref`, that was old and I missed it. I went with shortening the printed type name to keep the IR succinct, but I can see the argument for just using the long form name everywhere (tbc if you wanted the long form name that would be a bit of a pain updating all of the tests but whatever). My bot didn't struggle at all with the naming incongruity so I kept it.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
qedawkins: Yes, I essentially meant spawn `nproc` processes. I didn't like the wording either, `nproc` is much more succinct and clearer so changing it to that.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
qedawkins: Done, everything should have a mutable and non-mutable variant now.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1086 @@
...
+  ArrayRef<bool> isTied = op.getIsTied();
+  int64_t numResults = op.getNumResults();
+  if (isTied.size() != numResults) {
+    return op.emitOpError(
+               "`is_tied` mask length expected to match number of results ")
+           << numResults;
+  }
+
+  int64_t numInits =
+      std::accumulate(isTied.begin(), isTied.end(), (int64_t)(0));
```
kuhar: use `llvm::sum_of`

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1086 @@
...
+  auto [offsets, sizes, strides] = getOffsetsSizesAndStrides(ranges);
+  build(b, result, source, dest, offsets, sizes, strides, attrs);
+}
+
+// Build a WriteSliceOp with dynamic entries.
+void WriteSliceOp::build(OpBuilder &b, OperationState &result, Value source,
+                         Value dest, ValueRange offsets, ValueRange sizes,
+                         ValueRange strides, ArrayRef<NamedAttribute> attrs) {
+  auto offsetValues =
+      llvm::map_to_vector(offsets, [](Value v) -> OpFoldResult { return v; });
```
kuhar: use `llvm::StaticCastTo<OpFoldResult>` instead of the lambda? Also below.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1086 @@
...
+  setStaticOffsetsAttr(builder.getDenseI64ArrayAttr(staticOffsets));
+  setStaticStridesAttr(builder.getDenseI64ArrayAttr(staticStrides));
+  getOffsetsMutable().assign(dynamicOffsets);
+  getStridesMutable().assign(dynamicStrides);
+
+  return {};
+}
+
+LogicalResult GetMemrefOp::verify() {
+  auto resultType = getResultType();
```
kuhar: type

kuhar (approved):
The implementation looks good; I haven't reviewed the op semantics carefully, so you probably want to wait for another review that focuses on this.

qedawkins:
> The implementation looks good; I haven't reviewed the op semantics carefully, so you probably want to wait for another review that focuses on this.

Yeah there are some folks on vacation so I'm going to give time for everyone to take a look. Also I'm still waiting on other changes to get reviewed + land so I'm not in a rush. This commit has basically 0 conflicts with anything so it's easy to float :)

(thanks for the quick reviews though!)

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -60,8 +60,7 @@ static LogicalResult verifyParallelBodyOp(OpTy op, int64_t numLeadingArgs,
            << numResults;
   }
 
-  int64_t numInits =
-      std::accumulate(isTied.begin(), isTied.end(), (int64_t)(0));
+  int64_t numInits = llvm::sum_of(isTied, (int64_t)(0));
```
kuhar, line 53: You don't need to provide the 0 value.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+
+void ReadSliceOp::build(OpBuilder &b, OperationState &result, Type resultType,
+                        Value source, ValueRange offsets, ValueRange sizes,
+                        ValueRange strides, ArrayRef<NamedAttribute> attrs) {
+  auto offsetValues =
+      llvm::map_to_vector(offsets, [](Value v) -> OpFoldResult { return v; });
+  auto sizeValues =
+      llvm::map_to_vector(sizes, [](Value v) -> OpFoldResult { return v; });
+  auto strideValues =
+      llvm::map_to_vector(strides, [](Value v) -> OpFoldResult { return v; });
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+
+void GetMemrefOp::build(OpBuilder &b, OperationState &result, Type resultType,
+                        Value source, ValueRange offsets, ValueRange sizes,
+                        ValueRange strides, ArrayRef<NamedAttribute> attrs) {
+  auto offsetValues =
+      llvm::map_to_vector(offsets, [](Value v) -> OpFoldResult { return v; });
+  auto sizeValues =
+      llvm::map_to_vector(sizes, [](Value v) -> OpFoldResult { return v; });
+  auto strideValues =
+      llvm::map_to_vector(strides, [](Value v) -> OpFoldResult { return v; });
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
qedawkins: I do because `isTied` is a list of bools so unless I specify the accumulator type explicitly then it infers a bool.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,280 @@
...
+include "iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td"
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/BuiltinTypeInterfaces.td"
+include "mlir/IR/EnumAttr.td"
+
+//===----------------------------------------------------------------------===//
+// Parallel control flow dialect
+//===----------------------------------------------------------------------===//
+
+def PCF_Dialect : Dialect {
```
krzysz00, line 21: High-level question: once we've seen this working in real code / have had it settle down, do we think we might want to upstream this?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,280 @@
...
+  let useDefaultAttributePrinterParser = 1;
+}
+
+//===----------------------------------------------------------------------===//
+// Base pcf dialect op classes
+//===----------------------------------------------------------------------===//
+
+class PCF_Op<string mnemonic, list<Trait> traits = []> :
+    Op<PCF_Dialect, mnemonic, traits> {
+  let hasCustomAssemblyFormat = 1;
```
krzysz00: ... Huh

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,280 @@
...
+    between two coordinates can be made. Elements referenced by an `sref` may
+    not internally alias.
+
+    ```
+    template<size_t rank, typename eltype, typename alloc_scope_ty, sync_scope>
+    class ShapedRef {
+      // Access is pointwise within the coordinate space implied by the shape.
+      // Element type determines the minimum access bitwidth.
+      eltype *getElementPtr(int a, ...) // |rank| operands.
+      int shape[rank];
```
krzysz00: int64_t ? ssize_t?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,280 @@
...
+//===----------------------------------------------------------------------===//
+
+def PCF_DummyScopeAttr :
+    AttrDef<PCF_Dialect, "DummyScope", [
+      DeclareAttrInterfaceMethods<PCF_ScopeAttr>
+    ]> {
+  let mnemonic = "dummy_scope";
+  let summary = [{Dummy scope attribute used for testing }];
+  let description = [{
+    Scope that fails on all interface uses to use in testing where the scope
```
krzysz00:
```suggestion
    Scope that fails on all interface uses. For use in testing where the scope
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,280 @@
...
+  let summary = [{Attribute representing sequential execution}];
+  let description = [{
+    Scope that reuses the current process as the sole executor of a parallel
+    region.
+  }];
+  let parameters = (ins);
+  let assemblyFormat = [{}];
+
+  let extraClassDeclaration = [{
+    SmallVector<Value> getWorkerCounts(OpBuilder& builder, Location loc, int64_t numIds) {
```
krzysz00, line 227: ... Are we _sure_ this shouldn't be `SmallVector<OpFoldResult>` to allow constants to be constant?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,280 @@
...
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// Misc aliases
+//===----------------------------------------------------------------------===//
+
+def PCF_AnyRef: AnyTypeOf<[PCF_ShapedRef, AnyMemRef]>;
+def PCF_AnyShapedRef: AnyTypeOf<[PCF_ShapedRef]>;
+def PCF_AnyTensorOrMemRef: AnyTypeOf<[AnyRankedTensor, AnyMemRef]>;
```
krzysz00: We already have these in Utils/?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,280 @@
...
+    }
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// Misc aliases
+//===----------------------------------------------------------------------===//
+
+def PCF_AnyRef: AnyTypeOf<[PCF_ShapedRef, AnyMemRef]>;
+def PCF_AnyShapedRef: AnyTypeOf<[PCF_ShapedRef]>;
```
krzysz00: I think we might want to use the same constructions that AnyTensor / TensorOf and such out in upstream use.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
```diff
@@ -0,0 +1,180 @@
...
+
+//===----------------------------------------------------------------------===//
+// IREE::PCF::SyncScopeAttr
+//===----------------------------------------------------------------------===//
+
+def PCF_SyncScopeAttr : AttrInterface<"SyncScopeAttr"> {
+  let cppNamespace = "::mlir::iree_compiler::IREE::PCF";
+
+  let summary = [{Determines synchronization semantics.}];
+  let description = [{
```
krzysz00, line 129: Is it just me, or are we missing the mechanism to `await` a token?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+// AllocOps
+//===----------------------------------------------------------------------===//
+
+LogicalResult AllocOp::verify() {
+  if (getDynamicSizes().size() != getResultType().getNumDynamicDims()) {
+    return emitOpError(
+        "dimension operand count does not equal sref dynamic dimension count");
+  }
+
+  // TODO: Restrict legal parents for this op.
```
krzysz00: That can be done in TableGen, no?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+  return success();
+}
+
+SmallVector<OpFoldResult> AllocOp::getMixedSizes() {
+  SmallVector<OpFoldResult> result;
+  unsigned ctr = 0;
+  OpBuilder b(getContext());
+  for (int64_t i = 0, e = getResultType().getRank(); i < e; ++i) {
```
krzysz00: There are utility functions for this - see `Utils/StaticValueUtils.h` upstream

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+      if (failed(parser.parseArgumentList(regionLeadingArgs,
+                                          OpAsmParser::Delimiter::Paren,
+                                          /*allowType=*/true))) {
+        return failure();
+      }
+    }
+    numLeadingArgs = regionLeadingArgs.size();
+  }
+  if (failed(parser.parseKeyword("execute")))
+    return failure();
```
krzysz00, line 142: Ideally, we'd be doing with some SMLocs and better parse error messages

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+      return failure();
+    }
+  }
+
+  SmallVector<OpAsmParser::Argument> indexArgs;
+  if (failed(parser.parseLSquare())) {
+    return failure();
+  }
+
+  if (failed(parser.parseArgumentList(
```
krzysz00, line 176: Are you sure this couldn't be `OpAsmParser::Delimiter::Square` (or however you spell it)?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+
+  // If there is at least one region arg the arg types and op result types need
+  // to be parsed.
+  if (!regionRefArgs.empty()) {
+    if (failed(parser.parseColon()) || failed(parser.parseLParen())) {
+      return failure();
+    }
+
+    // Parse all types except the last followed by commas.
+    for (OpAsmParser::Argument &arg :
```
krzysz00: There's _definitely_ a utility for this

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+
+    int64_t numResults = isTied.size();
+    resultTypes.resize(numResults);
+    for (auto [i, isTied] : llvm::enumerate(isTied)) {
+      if (failed(parser.parseType(resultTypes[i]))) {
+        return failure();
+      }
+
+      auto shapedType = dyn_cast<ShapedType>(resultTypes[i]);
+      if (!shapedType) {
```
krzysz00, line 219: Check this in the verifier instead - the ShapedType-ness could, in theory, be added dynamically later

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+                   ValueRange inits, ValueRange dynamicSizes,
+                   ArrayRef<bool> isTied, bool syncOnReturn) {
+
+  result.addAttribute(LoopOp::getScopeAttrName(result.name), scope);
+  result.addOperands(count);
+  result.addOperands(inits);
+  result.addOperands(dynamicSizes);
+  result.addTypes(resultTypes);
+
+  result.addAttribute(
```
krzysz00: `result.getOrAddProperties().operandSegmentSizes - you don't need to make this an attribute, that just costs

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -0,0 +1,1085 @@
...
+                                 SmallVectorImpl<OpFoldResult> &results) {
+  bool changed = false;
+  OpBuilder builder(getContext());
+
+  SmallVector<OpFoldResult> mixedOffsets = getMixedOffsets();
+  SmallVector<OpFoldResult> mixedStrides = getMixedStrides();
+
+  // Try to fold dynamic offsets to static.
+  for (OpFoldResult &offset : mixedOffsets) {
+    if (Value value = dyn_cast<Value>(offset)) {
```
krzysz00: There are utility functions for this

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,788 @@
...
+
+    /// Erase the operand at 'index' from the operand list.
+    void eraseOperand(unsigned index);
+  }];
+
+  let assemblyFormat = [{
+    $condition $dest (`(` $dest_operands^ `:` type($dest_operands) `)`)? attr-dict
+  }];
+}
+
```
krzysz00, line 442: As a general flag, you may want to implement LoopLikeOpInterface so that integer range analysis works on these things?

krzysz00 (commented):
Initial review

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: I'm not particularly interested in moving it upstream because that just makes it harder for us to change the dialect and gives us less control over it. Also many of the interesting interface implementations are going to inevitably be tied to IREE stuff so moving it upstream leaves it stranded.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: good catch. I am shocked this never caused me issues...

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: I found it a lot easier implementation wise to leave it as `Value` so far, and none of my use cases have benefitted from OpFoldResult yet. Once we see a motivating use case I'm cool with changing it, but you can think of this interface method as more supposed to generate ops like `gpu.block_dim x/y/z` rather than a constant that we have annotated on the translation info (so this never actually generates constants right now).

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
qedawkins: Yes (sorta) :P

This attribute isn't necessarily backed by a token, it can be backed by whatever we want/invent. Right now the only version of this attribute that exists is one that waits on the barrier generated when returning from a `pcf.loop` or `pcf.generic`. It's just a scoped release + barrier to sync so no token waiting mechanism was needed yet.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
qedawkins: I would have to restrict it to specifically the initialization region of `pcf.generic` so I would have to handle that particular case in the verifier. Also in theory it could be nested within some control flow within that initialization region so I punted on adding a verifier to avoid the parent walks until we think it's actually worthwhile.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
qedawkins: These ops can't implement LoopLikeOpInterface for a few reasons:

1. The "iter_args", "init values", and "yielded" region values aren't actually produced the way that the LoopLikeOpInterface expects. The block arguments corresponding to an initial value is a completely different type.
2. LoopLikeOpInterface enables LICM which does NOT apply to these ops.

Good point on integer range analysis though, that definitely will need to be added here, I just haven't gotten around to that level of detail/optimization yet.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: This is pseudo-code :P

I made it size_t though :D

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
krzysz00: Ok, yeah, I can agree with that - we'll just need to make sure that we replace those with translation info constants _eventually_

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: I couldn't for the life of me figure out how to do it the upstream way without it removing the `#dialect.mnemonic` prefix from all the type printers, so I just left it as is.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
qedawkins: It's updating the regionRefArgs types in place so I can't use the simple utility that takes a SmallVector, but I switched to using parseCommaSeparatedList which cleaned up the code a bit.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: I didn't know about Utils, moved everything relevant over there.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: For sure. It really shouldn't be too bad to replace with constants later on. We can even support multiple entry points calling a shared function by hoisting id/count args towards entry points and then replacing.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
qedawkins: I'm casting in the parser since I need to know whether I need to parse the dynamic sizes. The verifier also checks for ShapedType in case it parsed from the generic representation. I'm adding better error logging though so that it's less of a surprise failure.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -200,26 +210,33 @@ static ParseResult parseParallelExecutionBody(
...
     resultTypes.resize(numResults);
     for (auto [i, isTied] : llvm::enumerate(isTied)) {
+      SMLoc resultTypeLoc = parser.getCurrentLocation();
       if (failed(parser.parseType(resultTypes[i]))) {
-        return failure();
+        return parser.emitError(resultTypeLoc, "failed to parse result type");
       }
 
-      auto shapedType = dyn_cast<ShapedType>(resultTypes[i]);
+      ShapedType shapedType = dyn_cast<ShapedType>(resultTypes[i]);
```
kuhar, line 218: auto

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,275 @@
...
+    bool isParentScopeOnlySync() const;
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// Base pcf attributes
+//===----------------------------------------------------------------------===//
+
+def PCF_DummyScopeAttr :
+    AttrDef<PCF_Dialect, "DummyScope", [
```
MaheshRavishankar: Please rename to `TestScope`.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
```diff
@@ -0,0 +1,180 @@
...
+    Attributes implementing this interface are tied to shaped refs and define
+    the synchronization semantics for that ref.
+  }];
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/[{
+        Returns the expanded types for this attribute that are introduced when
```
MaheshRavishankar, line 137: I dont think I fully understand how this gets used? The token gets resolved to a specific type?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
```diff
@@ -0,0 +1,180 @@
...
+      /*methodName=*/"getConcreteTypes",
+      /*args=*/(ins "MLIRContext*":$context)
+    >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Allocates the requisite data for this token. This must produce a range
+        with types matching `getConcreteTypes`.
+      }],
+      /*retTy=*/"SmallVector<Value>",
+      /*methodName=*/"allocate",
```
MaheshRavishankar, line 150: Can you give an example of a token that needs an allocation, and where that will get allocated?

MaheshRavishankar (approved):
Ok, did a review of the ops and left a few comments. I am approving cause I have no blocking comments.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
qedawkins: `pcf.sref` gets split with a 1-N conversion: `pcf.sref<..., #foo.sync_scope>` -> `(pcf.sref<...>, #foo.sync_scope.get_token_types())`. Currently the only implemented sync_scope is `parent_only` which only inserts a barrier at the end of the parent scope and otherwise gives no guarantees.

The idea is keeping the token tied to the `sref` type up until we're ready to separate and start resolving them.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
qedawkins: Two cases I had in mind were named barriers and producer/consumer queues as implemented here: https://huggingface.co/blog/mi300kernels

For named barriers we'd need to "allocate" the id of the barrier for that sref, so it might just be as simple as `arith.constant`, it could materialize as nothing (kept static), or we could use some special type depending on what impl we end up going for. For those mi300 kernels (originally shared by Jakub) they are managing their own producer/consumer queues in LDS so we'd actually have to allocate a little memory.

You can think of the sync_scope as a modifier that informs early optimization passes what we're allowed to do/assume about the memory referenced by the `pcf.sref` directly on the type. At some later point we separate the sync_scope from the memory and turn all of the memory operations into raw allocs/reads/writes + direct calls to synchronization stuff at which point we have to treat all remaining `pcf.sref`s like any other memory.

Groverkss:
I'm just back today from my vacation, if you can hold off landing until tommorow, I could give a review to this as well.

qedawkins:
> I'm just back today from my vacation, if you can hold off landing until tommorow, I could give a review to this as well.

I am going to fix a CI failure in my final e2e integration commit before landing so you have time (though the bug won't be in this PR). I also need to wait for integrate (this is currently set to merge into `users/qedawkins/pcf-base`).

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,274 @@
...
+    ShapedTypeInterface
+]> {
+  let mnemonic = "sref";
+
+  let summary = [{A shaped reference to a buffer.}];
+  let description = [{
+    A reference to a buffer with unspecified layout and physical storage.
+    Carries the shape and element type of the referenced region. Elements can
+    be accessed by index, though no assumptions about the physical relation
+    between two coordinates can be made. Elements referenced by an `sref` may
```
Groverkss: What does may not internally alias mean?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,274 @@
...
+    not internally alias.
```
Groverkss: I'm not sure what `Elements referenced by an `sref` may Comment view not internally alias.`

Are you talking about the same element? Is this about two different elements? I don't understand the "may not internally alias" part.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -0,0 +1,274 @@
...
+
+def PCF_SyncOnParentAttr :
+    AttrDef<PCF_Dialect, "SyncOnParent", [
+      DeclareAttrInterfaceMethods<PCF_SyncScopeAttr>
+    ]> {
+  let mnemonic = "sync_on_parent";
+  let summary = [{
+    Attribute indicating syncronization for the tied shape ref is dictated
+    solely by the parent.
+  }];
```
Groverkss, line 246: I'm a bit confused about how we can define a parent of a value. A value can be passed around. Does this mean you cannot outline a value like this ever?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
```diff
@@ -0,0 +1,180 @@
...
+#ifndef IREE_CODEGEN_DIALECT_PCF_INTERFACES
+#define IREE_CODEGEN_DIALECT_PCF_INTERFACES
+
+include "mlir/IR/OpBase.td"
+
+//===----------------------------------------------------------------------===//
+// IREE::PCF::ScopeAttr
+//===----------------------------------------------------------------------===//
+
+def PCF_ScopeAttr : AttrInterface<"ScopeAttr"> {
```
Groverkss: The naming is a bit weird. I would have expected this to be called PCF_ScopeAttrInterface not PCF_ScopeAttr. Same for other interface names.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFInterfaces.td
```diff
@@ -0,0 +1,180 @@
...
+      }]
+    >,
+  ];
+}
+
+//===----------------------------------------------------------------------===//
+// IREE::PCF::SyncScopeAttr
+//===----------------------------------------------------------------------===//
+
+def PCF_SyncScopeAttr : AttrInterface<"SyncScopeAttr"> {
```
Groverkss: This should really be PCF_SyncScopeAttrInterface

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,781 @@
...
+    control flow reaches this op, `nproc` workers of the specified scope are
+    spawned and begin executing the region. The scope is given by an attribute
+    implementing the `ScopeAttr` interface and is responsible for the semantics
+    of all pcf primitives at the same scope. Further details about scopes are
+    included in the docs for the interface.
+
+    The optional `initialize` region is executed once when control flow first
+    reaches the op. Values yielded from the initializer become block arguments
+    available to the execute region. This is useful for setting up per-op
+    state that persists across all worker invocations.
```
Groverkss, line 109: Is the initialize region executed before the workers are spawned or after?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
```diff
@@ -0,0 +1,781 @@
...
+
+    Example:
+
+    ```mlir
+    pcf.<scoped op> #foo.scope {
+      ^bb0(%0: !foo.type)
+        %1 = ... %0 : !foo.type
+        pcf.br.cond_return %cond ^bb0(%0: !foo.type)
+    }
+    ```
```
Groverkss, line 422: Why not use region control flow instead of block based control flow?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: It's the same restriction as for memrefs: https://github.com/llvm/llvm-project/blob/bfce02548d8ae3d20fd987c1b8aa56bd7578b1c9/mlir/include/mlir/IR/BuiltinTypes.td#L610

Since this type does not have an explicit layout, the wording is a little different but I can add an example to make it clearer.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: Left a comment on the first note, closing this one.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
Groverkss: Maybe this is an english problem an the correct word should be "must"? Do you mean "Elements at different corrdinates must not internally alias".

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
qedawkins: So I considered that for a while, but region based requires explicit passing of values between regions (i.e. no implicit captures across regions) and I found that to be more painful to write the transforms for so I went for block based for this op. I also wasn't sure how to make tied results work with it. With a single entry block I can require that the `pcf.sref` argument to the entry block IS the sole value we need to look at to find all accesses to a tied result. With region based stuff we either need to come up with restrictions on the types you're allowed to pass across region boundaries (restrict sref and leave it to the parent) or track uses across region boundaries instead of just SSA. I am open to region based CF once we need it (I could see a few good use cases) either by extending this op (not hard to just allow for more regions on this op) or another op (new ops of this kind aren't *too* hard to add).

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.td
qedawkins: Workers are spawned then all run the initialize region. I'd love for it to be workers are spawned *after*, but it has to be this way mainly because for the parallelism levels we're targeting and the fact that this is a region on an op that can go anywhere. Two other ways to do it would be to make just the initializer region isolated from above (I don't think that's possible though) or just make it reference a function with no arguments that gets called before threads are spawned. The issue with the function is it means during transformations nested on functions we can't make changes to the initializer region. Also both of those options are more awkward to implement dynamic shared mem for (you'd have to capture backwards program slices on the fly and verify uniformity of the allocation operands) so I'm just yoloing it right now :P.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: You can outline it, that is a good point though, perhaps "parent" is the wrong wording here. Maybe `SyncOnReturn` is clearer? Basically the value is only synchronized once we return from the worker.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
qedawkins: Yeah that, let me update the docs.

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
kuhar: also typo in synchroniztion

Groverkss (approved):

PR #22808, opened 2025-12-03
Adding AsyncUpdateOp elision support to ElideAsyncCopiesPass.

benvanik:
Adds support for eliding stream.async.update operations when the mutation has no observable effect. An update is safe to elide if:
- The target is not a by-reference function argument (mutation observable to caller)
- No subsequent operation reads from the mutated region

This enables optimizations where staging buffers can be bypassed when the update is immediately overwritten by another operation.

(most of this was authored with claude)


AWoloszyn (approved):

pdhirajkumarprasad:
@benvanik this change has caused ~400 failure in ONNX model for CPU only. I am working on creating minimum repro for this

PR #22809, opened 2025-12-03
[python] use pytest for tests

bangtianliu:
Existing Python tests like `dialects_test.py` use raw assert statements. 

This PR aims to transition them to pytest.

ci-extra: test_torch, windows_x64_msvc

### DIFF
--- a/build_tools/cmake/iree_python.cmake
+++ b/build_tools/cmake/iree_python.cmake
@@ -105,7 +105,7 @@ function(iree_local_py_test)
   add_test(
     NAME ${_NAME_PATH}
     COMMAND
-      "${Python3_EXECUTABLE}"
+      "${Python3_EXECUTABLE}" -m pytest
       "${CMAKE_CURRENT_SOURCE_DIR}/${_RULE_SRC}"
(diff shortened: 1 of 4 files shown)


build_tools/cmake/iree_python.cmake
```diff
@@ -105,7 +105,7 @@ function(iree_local_py_test)
   add_test(
     NAME ${_NAME_PATH}
     COMMAND
-      "${Python3_EXECUTABLE}"
+      "${Python3_EXECUTABLE}" -m pytest
       "${CMAKE_CURRENT_SOURCE_DIR}/${_RULE_SRC}"
       ${_RULE_ARGS}
   )
```
ScottTodd, line 111: Have you evaluated https://python-cmake.github.io/pytest-cmake/ ?

bangtianliu: not yet

bangtianliu: Thanks for pointing this, I will see how to use pytest-cmake.

PR #22816, opened 2025-12-03
[Encoding] Remove unneeded command line option

Muzammiluddin-Syed-ECE:
Previously, scaled GEMMs were unsupported in the data tiling pipeline and so a CL option was introduced to not generate scaled GEMM encodings by default. Now that we have support for data tiled scaled GEMMs, these encodings properly materialize, and this CL option can be removed.


Max191 (approved):
Please add a description explaining that scaled matmul DT is supported now before merging.

Max191 (approved):

kuhar (approved):

PR #22818, opened 2025-12-03
[Codegen][GPU] Replace prefetch_shared_memory with prefetch_num_stages in IREEGPUAttrs

jerryyin:
This PR looks big but can really condensed into a single file change, in `IREEGPUAttrs.td`

It replaced the
>   prefetch_shared_memory: Boolean option indicating whether or not to run
      the loop prefetching pass in the lowering pipeline.

With below, and default prefetch depth stays equivalent with current settings of 2.
>   prefetch_num_stages: Integer option controlling the software pipelining
      stages for shared memory prefetching. 0 or 1 disables prefetching, 2+
      enables prefetching with that many pipeline stages.

The goal is to replace the coarse grained bool control to integer such that the depth of software pipelining is configurable. All the rest of the changes are either fixing compilation failures or test failures. Once this PR is merged, I'd expose the to a --iree-compile flag so the integer can be set.

I've double checked with CC @bangtianliu that tuner doesn't touch this attribute yet, and it is left as default. (But this may result in some follow-up items once merged).


compiler/src/iree/compiler/API/Internal/IREEGPUDialectCAPI.cpp
```diff
@@ -24,15 +24,15 @@ bool ireeAttributeIsAGPUPipelineOptionsAttr(MlirAttribute attr) {
...
+ireeGPUPipelineOptionsAttrGet(MlirContext mlirCtx, int64_t *prefetchNumStages,
                               bool *noReduceSharedMemoryBankConflicts,
                               bool *useIgemmConvolution,
                               MlirAttribute *reorderWorkgroupsStrategy) {
   mlir::MLIRContext *ctx = unwrap(mlirCtx);
   mlir::Builder b(ctx);
-  auto prefetchSharedMemoryAttr = mlir::BoolAttr();
-  if (prefetchSharedMemory) {
-    prefetchSharedMemoryAttr = b.getBoolAttr(*prefetchSharedMemory);
+  auto prefetchNumStagesAttr = mlir::IntegerAttr();
```
kuhar:
```suggestion
  mlir::IntegerAttr prefetchNumStagesAttr;
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
```diff
@@ -930,15 +931,15 @@ def IREEGPU_GPUPipelineOptionsAttr : AttrDef<IREEGPU_Dialect, "GPUPipelineOption
   let cppNamespace = "::mlir::iree_compiler::IREE::GPU";
 
   let parameters = (ins
-    OptionalParameter<"BoolAttr">:$prefetch_shared_memory,
+    OptionalParameter<"IntegerAttr">:$prefetch_num_stages,
```
kuhar: Could we give this a fixed integer type so that we don't have to use specify it IR? IE:
`#iree_gpu.pipeline_options<prefetch_num_stages = 2 : i64, no_reduce_shared_memory_bank_conflicts = true`

==>

`#iree_gpu.pipeline_options<prefetch_num_stages = 2, no_reduce_shared_memory_bank_conflicts = true`

jerryyin: This is a really good feedback. I did some search and seems like `OptionalParameter<"std::optional<int64_t>">` is the way to go. Will address the review in the new commit.

kuhar (approved):

Issue #22815, opened 2025-12-03
[spirv] Lowering error in SPIRVSubgroupReduce pipeline on reduction + elementwise


kuhar:
### What happened?

At some point after release 3.5, we started performing more aggressive fusions and putting reductions + elementwise in the same dispatch regions. This fails to compile with SRPIV.

### Steps to reproduce your issue

```mlir
// RUN: iree-compile %s --iree-hal-target-device=vulkan --iree-vulkan-target=rdna4

func.func @main(
    %arg0: tensor<6144xf32>,
    %arg1: tensor<64x3x32xf32>
) -> (tensor<64x3x32xf32>, tensor<32x64x3xf32>) {
  %cst = arith.constant 0.000000e+00 : f32

  %result:2 = flow.dispatch.region -> (tensor<64x3x32xf32>, tensor<32x64x3xf32>) {
    %empty_f32 = tensor.empty() : tensor<f32>
    %empty_64x3x32 = tensor.empty() : tensor<64x3x32xf32>
    %empty_32x64x3 = tensor.empty() : tensor<32x64x3xf32>

    %filled = linalg.fill ins(%cst : f32) outs(%empty_f32 : tensor<f32>) -> tensor<f32>

    %reduction = linalg.generic {
      indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> ()>],
      iterator_types = ["reduction"]
    } ins(%arg0 : tensor<6144xf32>) outs(%filled : tensor<f32>) {
    ^bb0(%in: f32, %out: f32):
      %0 = arith.mulf %in, %in : f32
      %1 = arith.addf %out, %0 : f32
      linalg.yield %1 : f32
    } -> tensor<f32>

    %generic:2 = linalg.generic {
      indexing_maps = [
        affine_map<(d0, d1, d2) -> (d0, d1, d2)>,
        affine_map<(d0, d1, d2) -> ()>,
        affine_map<(d0, d1, d2) -> (d0, d1, d2)>,
        affine_map<(d0, d1, d2) -> (d2, d0, d1)>
      ],
      iterator_types = ["parallel", "parallel", "parallel"]
    } ins(%arg1, %reduction : tensor<64x3x32xf32>, tensor<f32>)
      outs(%empty_64x3x32, %empty_32x64x3 : tensor<64x3x32xf32>, tensor<32x64x3xf32>) {
    ^bb0(%in: f32, %in_reduction: f32, %out0: f32, %out1: f32):
[...]

PR #12661, opened 2023-03-16
Do not fuse producer-consumer if they have no common outer parallel loops

MaheshRavishankar:
Producer-consumer fusion tries to fuse producer with consumer when they have common outer-parallel loops.  This is done through use of affine_maps, but missed a case where the affine_map is empty which should be treated as a case with no common outer parallel loops (it was treated as a having common outer parallel loops).

Fixes #12431

PR #22819, opened 2025-12-03
[Dispatch Creation] Don't fuse if there are no common parallel loops

IanWood1:
This change prevents fusion when the root op does not share parallel loops with it's producers or consumers. Because we require the iteration space of reduction ops to match exactly, this can only occur when the non-root is all parallel. And when the non-root is all parallel, there will only ever be no common parallel loops if the root has no parallel loops. `newMap` is the mapping from the root op's to the non-root op's parallel loops. If there aren't any dims in the output, then there is no mapping between the loops.

This preserves the old behavior introduced by https://github.com/iree-org/iree/pull/12661.

Related https://github.com/iree-org/iree/issues/22815


kuhar (approved):
Thanks @IanWood1

compiler/src/iree/compiler/DispatchCreation/test/dispatch_linalg_ext_fusion.mlir
```diff
@@ -22,7 +22,7 @@ util.func public @linalgext_scatter_dispatch() -> tensor<8192x16x8x128xf32> {
     linalg.yield %10 : f32
   } -> tensor<4x1x16x8x128xf32>
 
-  %8 = iree_linalg_ext.scatter dimension_map = [0] unique_indices(false) ins(%7, %6 : tensor<4x1x16x8x128xf32>, tensor<4x1xi32>) outs(%4 : tensor<8192x16x8x128xf32>) {
+  %8 = iree_linalg_ext.scatter dimension_map = [0] unique_indices(true) ins(%7, %6 : tensor<4x1x16x8x128xf32>, tensor<4x1xi32>) outs(%4 : tensor<8192x16x8x128xf32>) {
```
IanWood1: unique_indices(false) sets the outer dimensions to reduction, meaning that there are no outer parallel loops. I'm not fully confident about this change. I need to look into it a bit more to figure out why the old logic always fused this (before the changes that caused the issue).

IanWood1:
CLIP dispatch count increased because this now gets split into 2 dispatches https://gist.github.com/IanWood1/73971d57bd8243f491291bfa4d4b8e36

MaheshRavishankar (approved):
The increase in clip dispatch is fine. This shouldnt have fused to start with.

PR #22822, opened 2025-12-04
[Codegen] Add MergeExtractSliceThroughBlockArgPattern

bangtianliu:
## Summary

This PR adds a new pattern, **`MergeExtractSliceThroughBlockArg`**, which merges consecutive `tensor.extract_slice` operations that flow through `scf.forall` block arguments. By collapsing these redundant slice chains, the pass removes non-tileable barriers and enables tile-and-fuse to correctly reach tileable producers (e.g., `linalg.fill`, `linalg.broadcast`).

This unlocks producer fusion that was previously blocked and eliminates race conditions caused by shared outer-loop temporaries.

---

### Problem

During the `TileAndDistributeToWorkgroupsUsingForallOpPass`, some `tensor.extract_slice` ops fail to fuse with their producers even when the original producer is tileable. The root cause is:

* `tensor.extract_slice` **does not implement** `TilingInterface`
* extract-slice chains flowing through `scf.forall` block arguments act as **fusion barriers**
* tileable producers (e.g., `linalg.fill`, `linalg.broadcast`) stay in the **outer loop**
* inner loops read from shared temporaries → **race conditions**

Example of the problematic structure:

```mlir
// Outer loop
%9 = linalg.fill ins(%cst : f16) outs(%8)
%ex0 = tensor.extract_slice %arg1[...]       // slice #1
%bc0 = linalg.broadcast ins(%9) outs(%ex0)
%ex1 = tensor.extract_slice %bc0[...]        // slice #2

// Inner loop: shared_outs come from slice #2
scf.forall shared_outs(%arg5 = %ex1) {
  %ex_inner = tensor.extract_slice %arg5[%arg3, %arg4] ...
  // Fusion attempts to reach producer of %arg5 → hits extract_slice → fails
}
```

**Impact:**
Fusion cannot look through `%ex1` to reach `%9` (the tileable `linalg.fill`).
Thus, the fill/broadcast ops execute once at dispatch scope, and all workgroups share the same memory slice → incorrect parallel behavior.

---

## Solution

The new rewrite merges nested extract-slice operations **through the block argument**, removing the non-tileable hop:

```mlir
// Before (blocked)
%arg5 = extract_slice( broadcast( fill(...) ) )

// After (merged)
%arg5 = fill(...)
```

After merging, the inner loop directly slices from the tileable producer:

```mlir
scf.forall (...) {
  %ex = tensor.extract_slice %arg1[%arg3, 0, %17] ...

  // Now tile-and-fuse can reach the producer:
  %tiled_fill = linalg.fill ins(%cst) outs(%tile_out)
  %tiled_bc   = linalg.broadcast ins(%tiled_fill) outs(...)
}
```

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/TileDispatchUsingForall.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/TileDispatchUsingForall.cpp
@@ -308,6 +308,7 @@ void TileAndDistributeToWorkgroupsUsingForallOpPass::runOnOperation() {
   // TODO(Max191): Replace populateSwapExtractWithExpandPattern with upstream
   // MLIR version once it is available (llvm-project/pull/126898).
   populateSwapExtractWithExpandPattern(cleanupPatterns);
+  populateMergeExtractSliceThroughBlockArgPattern(cleanupPatterns);
   // When fusing pads we do not want to generate zeroSliceGuards when doing
   // workgroup tiling. In `GPUApplyTilingLevelPass` we do have an option called
   // `allowZeroSlices` that can control this but we do not want these
--- a/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
@@ -779,6 +779,18 @@ struct MergeExtractSliceThroughBlockArg
           innerExtractOp, "shared_out is not defined by an extract_slice op");
     }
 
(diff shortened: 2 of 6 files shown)


bangtianliu:
Ok, I need to look into this CI error (llvm-cpu_split_reduction_using_tiling.mlir): https://github.com/iree-org/iree/actions/runs/19945814198/job/57201689113?pr=22822#step:5:2949

PR #22826, opened 2025-12-04
[Encoding] Use struct directive for TestingAttr assembly format

jtuyls:
The custom assembly format for encoding dialect's TestingAttr allowed specifying an testing attribute without layouts with `#iree_encoding.testing<>` instead of `#iree_encoding.testing<[]>`. By making use of the struct directive we can avoid the need for a custom assembly format while keeping the same behaviour, except for the parameter keys being added. However, these additional keys like `layouts =` are useful as well imo to improve the readability of the attribute.

Note that this also aligns the testing attribute assembly format with other related attributes:
https://github.com/iree-org/iree/blob/68a9309688a6605151d97d369a68cb5b620802bc/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.td#L140
https://github.com/iree-org/iree/blob/68a9309688a6605151d97d369a68cb5b620802bc/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td#L629
https://github.com/iree-org/iree/blob/68a9309688a6605151d97d369a68cb5b620802bc/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.td#L97

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/test/roundtrip.mlir
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/test/roundtrip.mlir
@@ -223,11 +223,11 @@ func.func @testing_without_layouts(%arg0: tensor<?x?xf32, #encoding>) -> tensor<
 
 // -----
 
-#encoding = #iree_encoding.testing<[#iree_encoding.specialization_resolver<123>]>
+#encoding = #iree_encoding.testing<layouts = [#iree_encoding.specialization_resolver<123>]>
 func.func @testing_with_layouts(%arg0: tensor<?x?xf32, #encoding>) -> tensor<?x?xf32, #encoding> {
   return %arg0 : tensor<?x?xf32, #encoding>
 }
-//      CHECK: #[[ENCODING:.+]] = #iree_encoding.testing<[#iree_encoding.specialization_resolver<123>]>
+//      CHECK: #[[ENCODING:.+]] = #iree_encoding.testing<layouts = [#iree_encoding.specialization_resolver<123>]>
 //      CHECK: func.func @testing_with_layouts(
 // CHECK-SAME:   %[[ARG0:.+]]: tensor<?x?xf32, #[[ENCODING]]>
 // CHECK         return %[[ARG0]]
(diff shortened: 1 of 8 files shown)


MaheshRavishankar (approved):

compiler/src/iree/compiler/Dialect/Encoding/IR/test/roundtrip.mlir
```diff
@@ -223,11 +223,11 @@ func.func @testing_without_layouts(%arg0: tensor<?x?xf32, #encoding>) -> tensor<
 
 // -----
 
-#encoding = #iree_encoding.testing<[#iree_encoding.specialization_resolver<123>]>
+#encoding = #iree_encoding.testing<layouts = [#iree_encoding.specialization_resolver<123>]>
```
hanhanW, line 226: huh, I did not catch this when I improved the testing encodings. It is better to use `iree_encoding.specialized<123>` instead.

hanhanW (commented):
It was intended to implement the custom parser to simplify the IR dump. When I implemented the encodings, it was the first encoding for testing purpose so I ended up with this bad situation.

Testing is a bad term and maybe we want to just replace `testing` with `layouts`.

compiler/src/iree/compiler/Dialect/Encoding/IR/test/roundtrip.mlir
jtuyls: I created a follow-up for this: https://github.com/iree-org/iree/pull/22851

jtuyls:
> It was intended to implement the custom parser to simplify the IR dump. When I implemented the encodings, it was the first encoding for testing purpose so I ended up with this bad situation.
> 
> Testing is a bad term and maybe we want to just replace `testing` with `layouts`.

I am not sure. `#iree_encoding.testing` seems useful to me for conciseness and clarity in lit tests. If we change the name to `layouts`, it's confusing because of the existing `#iree_encoding.layout`, which we might just want to reuse then. But still, imo, a dedicated `#iree_encoding.testing` is very clear and concise for testing purposes.

hanhanW:
> > It was intended to implement the custom parser to simplify the IR dump. When I implemented the encodings, it was the first encoding for testing purpose so I ended up with this bad situation.
> > Testing is a bad term and maybe we want to just replace `testing` with `layouts`.
> 
> I am not sure. `#iree_encoding.testing` seems useful to me for conciseness and clarity in lit tests. If we change the name to `layouts`, it's confusing because of the existing `#iree_encoding.layout`, which we might just want to reuse then. But still, imo, a dedicated `#iree_encoding.testing` is very clear and concise for testing purposes.

my bad, I forgot that we already have `#iree_encoding.layouts` attribute. And yes, having a testing encoding is clear and concise.

Issue #22695, opened 2025-11-18
Improvements to stablehlo.fft


pstarkcdpr:
### Request description

Currently, lowering `stablehlo.fft` has a number of issues, including:
1. Only `RFFT` is supported by the compiler, even though `FFT`, `IFFT`, and `IRFFT` are accepted by StableHLO.
2. The compiler does not intentionally fail `FFT`, `IFFT`, and `IRFFT` -- there are still compiler errors, but they happen more by accident than by design.
3. Large `RFFT` operations fail due to an error with readonly tensors (see issue [22473](https://github.com/iree-org/iree/issues/22473)).
4. Only 1D FFTs are supported.
5. If a multidimensional FFT is requested (e.g. `stablehlo.fft %0, type =  RFFT, length = [128, 128]`), a 1D FFT will be done on the last dimension, which is not the expected output.

We need to use FFTs for some of our models, so we've made some improvements to the current state. I've made a branch with these improvements here: https://github.com/pstarkcdpr/iree/tree/fft_1d

We addressed some, but not all, of the above issues:
1. Implemented the remaining FFT operations (`FFT`, `IFFT`, and `IRFFT`).
2. No need to fail now!
3. Changed the way that the all-zero imaginary input tensor is created to avoid it being marked as readonly, which fixes this issue.
4. Multidimensional FFTs are still not implemented.
5. Cause a compile error if FFT length is > 1. Note that you can continue to do a 1D FFT on the last dimension of a multidimensional tensor, but you still can't do a multidimensional FFT. It just fails now instead of silently giving the wrong answer.

I've never contributed to this project before so I'm not entirely certain of the process. I can make a PR from my fork. Internally, I also have a Python script that I'm using to test the FFTs. It compares the results to Numpy (within epsilon), as well as tests round trips within IREE (`input == ifft(fft(input))`). I'm not sure how IREE's testing works, so I don't know how to integrate something like this into the project.

[...]

Issue #22473, opened 2025-10-30
Getting "One or more operations with large vector sizes were found" error from linalg_ext.fft


pstarkcdpr:
### What happened?

[rfft_repro.zip](https://github.com/user-attachments/files/23226191/rfft_repro.zip)

I have a simple repro case where `linalg_ext.fft` causes the following compile error:
```
rfft_repro.mlir:25:12: error: One or more operations with large vector sizes (32768 bytes) were found:
```
See `rfft_repro.mlir`.

I tried tweaking the code a little bit and found that changing `%cst_5` from `arith.constant` to `tensor.empty` makes the compile error go away for some reason (I believe passing in a tensor argument instead of using `arith.constant` also achieves this). See `rfft_repro1.mlir` for this case.

While trying to make an even smaller repro, I found a 10 line file that compiled (see `rfft_repro2.mlir`, also quoted below):
```
module @rfft {
  util.func public @main(%arg0: tensor<128x153x8xf32>, %arg1: tensor<128x153x8xf32>) -> (tensor<128x153x8xf32>, tensor<128x153x8xf32>) {
    %cst1 = arith.constant 1 : index
    %cst_3 = arith.constant dense<0.000000e+00> : tensor<1xf32>
    %cst_4 = arith.constant dense<1.000000e+00> : tensor<1xf32>
    %0:2 = iree_linalg_ext.fft ins(%cst1, %cst_3, %cst_4: index, tensor<1xf32>, tensor<1xf32>) outs(%arg0, %arg1: tensor<128x153x8xf32>, tensor<128x153x8xf32>) : tensor<128x153x8xf32>, tensor<128x153x8xf32>
    util.return %0#0, %0#1 : tensor<128x153x8xf32>, tensor<128x153x8xf32>
  }
}
```
but it seemed brittle. Small perturbations caused issues. Not sure if they are related to the main issue or not, but returning only a single tensor `%0#0` caused a `op exceeded stack allocation limit` error (see `rfft_repro3.mlir`). Skipping the second and third parameters to `fft` (they are supposed to be optional) caused a compiler crash (see `rfft_repro4.mlir`).

I'm primarily concerned with the first case, but thought that I'd share some of the investigation I did in case it's useful.

The above tests were run on 3.8.0, although I originally saw the issue in 3.5.0 (the version we primarily use), and also tried 3.6.0.

### Steps to reproduce your issue

[...]

PR #22829, opened 2025-12-04
Implement missing stablehlo.fft operations

pstarkcdpr:
See issue [#22695](https://github.com/iree-org/iree/issues/22695). This change does the following:
- Causes a compile error when the fft length has more than 1 dimension (this was never supported but previously gave the wrong answer silently).
- Fixes an issue with RFFT where the imaginary part (a constant zero tensor) could cause a large stack allocation and fail to compile. Fixed by creating the zero imaginary tensor as part of the input shuffling step.
- Implements the remaining FFT operations: FFT, IFFT, IRFFT.
- Changes the function name `rewriteFft` to `rewriteRfft` since it only handles RFFT. The function `rewriteFft` is now used for the full FFT. Also adds `rewriteIrfft`.
- Updates Torch to use the new interface (i.e. changes calling `rewriteRfft` to `rewriteFft`).
- Updates the Torch tests to reflect changes in RFFT code gen.
- Updates the StableHLO tests to reflect changes in RFFT code gen.
- Adds StableHLO tests for FFT, IFFT, and IRFFT.
- Adds e2e tests for `stablehlo.fft`. Note that these are disabled in VMVX due to apparent issues with complex numbers. I mention this in the issue. Also disabled in Vulkan (again mentioned in the issue), but it's worth noting that the FFTs _do_ compile and run when using `sm_75`, but fail to compile with the Vulkan target is omitted. I haven't tried other Vulkan targets.


MaheshRavishankar:
Thanks for the change. This is a relatively big change and seems to do a lot of things at once. I dont want to add extra burden for you to contribute this to the project, so not going to hold you to it, but it is usually faster to merge small PR with targeted changes. Also easy to find missing test coverage that way. If you can "relatively easily" break this PR into smaller PRs that would be better. Ill still put this on my review stack though.

pstarkcdpr:
Thanks @MaheshRavishankar. The actual FFT implementation is one large change (`RewriteFft.cpp`) and a few minor ones. I don't think that can't really be separated out easily. Fixing the tests that break as a result probably needs to be done at the same time. Adding new tests could be a different PR though, if that helps. I just assumed that tests were required when adding new functionality, so made a point of finishing those before creating the PR 😄.

MaheshRavishankar:
Ok sounds good. Let me review this and see how it goes

pstarkcdpr:
Thanks @MaheshRavishankar. I'm very new to making changes in this project, so I wouldn't be surprised if there are things that could be done better.

The changes in `RewriteFft.cpp` still follows the same original algorithm, but I needed to refactor it to add the additional operations. Other than a bug fix, it still generates the same code as before for RFFT. If it helps, for when you get a chance to review, here's a high-level overview of the algorithm.

- FFT: (new) complex input and output. This is the canonical implementation and the rest are variations.
    - Shuffle and split input: FFT expects the inputs to be shuffled into bit-reversed order and have separate real and imaginary tensors. I do both of these at the same time in a single `linalg.generic`.
    - Run FFT: For each power-of-2, call `linalg_ext.fft` passing in generated coefficients as `arith.constant`s.
    - Complex output: Combine the real and imaginary parts back together into a complex tensor.

- IFFT: (new) complex input and output. Same as FFT with different coefficients and scaled by `1/N`.
    - Shuffle and split input: Same as FFT but additionally scales by `1/N`. FFTs are linear, so scaling the input or output has the same result. I figured that it was easiest to scale here while we're processing the input anyways instead of adding an additional step.
    - Run FFT: Same as FFT but with different coefficients.
    - Complex output: Same as FFT

- RFFT: (existed before but slightly modified) real input of size `N` to complex output of size `N/2+1`. Same as FFT but with real input (imaginary part is zero) and truncated.
    - Shuffle and generate: Shuffle input like with FFT, but no need to split since the input is not complex. We need the imaginary tensor of all zeroes though. Originally this was done with an `arith.constant`, but that had an issue with [large vector sizes](https://github.com/iree-org/iree/issues/22473) since it was readonly. I tried an `empty` + `fill`, but that didn't help. Instead, I'm generating the tensor of zeroes as part of the shuffle loop. It's a bit of a workaround and maybe there's a better way, but it works.
    - Run FFT: Same as FFT.
    - Truncate: `tensor.extract_slice` to size `N/2+1`
    - Complex output: Same as FFT

- IRFFT: (new) complex input of size `N/2+1` to real output of size `N`. Same as IFFT but need to expand the input back to size `N` and ignore the imaginary output.
    - Expand input: Make a tensor of size `N/2-1` that's mirrored and conjugated using a `linalg.generic`, then `tensor.concat` it to the input to make a tensor of size `N/2`.
    - Call IFFT.
    - Keep only the real tensor and ignore the imaginary one.
 
I hope this helps. Sorry for the big change. We needed IRFFT and that kind of necessitated implementing the whole family.

pstarkcdpr:
There was a CI failure due to a mistake I made with an excluded file in a `BUILD.bazel`. Should be fixed now. Not sure if there is a way to test Bazel locally.

pstarkcdpr:
Fixed the pre-commit formatting changes. It's my first PR in this repo so I had no idea that existed. I had been trying to stay within 80 character line limit by hand. This is much easier.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h
```diff
@@ -118,10 +118,43 @@ splitReduction(RewriterBase &rewriter, LinalgExt::TopkOp topkOp,
 /// %out = torch.aten.fft_rfft  %arg0, %none, %int-1, %none :
 ///           !torch.vtensor<[3,8,16],f32>, !torch.none, !torch.int, !torch.none
 ///           -> !torch.vtensor<[3,8,9],complex<f32>>
+FailureOr<std::pair<Value, Value>> rewriteRfft(Operation *op, Value operand,
```
MaheshRavishankar, line 121: I see why you added here, but ideally this should be in `compiler/Dialect/LinalgExt/IR/LinalgExtOps.h` . Not asking you to move it there, but just noting here.

tests/e2e/stablehlo_ops/fft_complex.mlir
```diff
@@ -0,0 +1,444 @@
...
+func.func @irfft_1d() {
+  %input = util.unfoldable_constant dense<[(16.588318,0.0), (-0.46602482,-2.0378456), (0.8568451,-0.33985636), (-1.1161734,0.6871074), (0.86731994,0.62438464), (-0.5296755,-0.90058994), (0.7260204,-1.6517396), (3.421573,2.8402321), (1.0612442,-2.7691374), (-0.33810878,0.4236927), (-0.7522839,-0.21823317), (-1.0411217,1.0264674), (0.09377679,-0.9341428), (-0.48347437,-0.76218516), (1.5533843,0.45277753), (-0.38116747,-0.24614418), (0.6452608,0.0)]> : tensor<17xcomplex<f32>>
+  %0 = stablehlo.fft %input, type = IRFFT, length = [32] : (tensor<17xcomplex<f32>>) -> tensor<32xf32>
+  check.expect_almost_eq_const(%0, dense<[0.75555766, 0.6689402, 0.3658616, 0.24478275, 0.78046405, 0.9256327, 0.42048004, 0.4443637, 0.98998463, 0.97323704, 0.21359204, 0.2344566, 0.38268244, 0.9832261, 0.81922436, 0.033826366, 0.87232935, 0.97022176, 0.795888, 0.05799026, 0.22904384, 0.42274877, 0.7001864, 0.46720487, 0.041910768, 0.26426968, 0.12391268, 0.6813819, 0.7870437, 0.16205639, 0.33862743, 0.43718916]> : tensor<32xf32>) : tensor<32xf32>
+  return
+}
+
+// 1D FFT on the innermost dimension of a 2D tensor.
+func.func @fft_2d() {
+  %input = util.unfoldable_constant dense<[
```
MaheshRavishankar, line 44: The tests before this seem fine to add here, but tests from here and below seem pretty large constants to add this way. I wonder if you can just use Torch.fft to check for correctness and add these tests to https://github.com/iree-org/iree-test-suites/tree/main/torch_ops . An even better solution would be to maybe have a test suite that is stablehlo based, and is setup similar to how the torch ops test suite is setup.

MaheshRavishankar (commented):
I dont know much about the FFT operation itself (apart from just basic understanding of it, and a lot of details have been flushed from my cache). I am not looking at the implementation details itself, but more the setup of the tests here. Left a few comment, but the comment is the test that is probably more blocking than anything else. Everything else looks pretty good. Thanks!

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h
pstarkcdpr: Ah. I just kept the functions where they were originally and then expanded from there. Worth pointing out this comment to @hanhanW since I think he's the one who added the original version and chose this location.

tests/e2e/stablehlo_ops/fft_complex.mlir
pstarkcdpr: Fair. I didn't implement the FFTs for Torch, just StableHLO (I maintained the Torch RFFT implementation though). How big is too big? I picked `32x32` for no obvious reason. I just wanted to test in 2D but make sure that the outer dimension was bigger than one. I could reduce this is `2x32` or `2x16` or something like that, in which case the constants wouldn't need to be any bigger than the 1D cases. What would be good?

MaheshRavishankar: 2x32 should be fine.

pstarkcdpr: Sounds good. I think I'll do `4x16` then, just so the outer dimension is non-trivial since the point is to test that it correctly does a 1D FFT even in a higher dimensional tensor.

pstarkcdpr: 2D tests updated. They are much smaller now.

MaheshRavishankar (approved):
Thanks!

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h
hanhanW: I may be the original author, but I'm not the one who moved it to this location. :) see https://github.com/iree-org/iree/commit/6d9f352dad5a7847daae84eb0890a468d9da76e3

MaheshRavishankar: Sigh! I was on the reviewers list but missed it. My bad!

hanhanW:
@pstarkcdpr do you need us to help land the PR? If so, can you fix the conflicts?

pstarkcdpr:
Yes, @hanhanW I could certainly use help getting this landed. I just wasn't sure what the next step was after the tests passed in December. I've resolved the conflicts now.

hanhanW:
There is a lint failure

pstarkcdpr:
Yeah I see it.

PR #22831, opened 2025-12-04
[tests][e2e] Add more llama related shapes

Muzammiluddin-Syed-ECE:
This patch adds other llama related shapes that were left out of [a previous PR](https://github.com/iree-org/iree/pull/22775). 

| N | K | 
|-------|-------|
| 1024 | 16384 |
| 16384 | 16384 |
| 53248 | 16384 |
| 53248 | 53248 |

Also, when running these tests locally, it was found that due to the extremely large size of the generated shapes, allocating space for accumulator arguments of type f32 was infeasible. So, a new argument is introduced `--no-accumulator` which allows you to generate tests which omit the accumulator as a function argument and instead use a constant set to zero in the IR.

### DIFF
--- a/tests/e2e/matmul/generate_e2e_matmul_tests.py
+++ b/tests/e2e/matmul/generate_e2e_matmul_tests.py
@@ -95,11 +95,9 @@ def get_test_shapes(shapes_id: ShapesId, accumulate=True):
         m, n, k = ShapesId.custom_mnk_values
         test_shapes = [TestShape(m=m, k=k, n=n, accumulate=False)]
         if accumulate:
-            test_shapes.extend(
-                [
-                    TestShape(m=m, k=k, n=n, accumulate=True),
-                ]
-            )
+            test_shapes += [
+                TestShape(m=m, k=k, n=n, accumulate=True),
+            ]
         return test_shapes
 
     raise ValueError(shapes_id)
--- a/tests/e2e/matmul/generate_e2e_matmul_tests.py
+++ b/tests/e2e/matmul/generate_e2e_matmul_tests.py
@@ -18,7 +18,7 @@ from tests.e2e.matmul.generate_code import *
 
 # Returns the list of TestShape's to use for the collection of shapes
 # identified by shapes_id.
-def get_test_shapes(shapes_id: ShapesId):
+def get_test_shapes(shapes_id: ShapesId, acc: bool):
     # Notes:
     # 1. Be conservative in adding more shapes, as that can increase both the
     #    build and execution latency of tests. The build latency is nearly the
@@ -92,10 +92,7 @@ def get_test_shapes(shapes_id: ShapesId):
         if ShapesId.custom_mnk_values is None:
             raise ValueError("Custom MNK values not set. Use --mnk=m,n,k")
         m, n, k = ShapesId.custom_mnk_values
-        return [
-            TestShape(m=m, k=k, n=n, accumulate=True),
-            TestShape(m=m, k=k, n=n, accumulate=False),
-        ]
+        return [TestShape(m=m, k=k, n=n, accumulate=acc)]
 
     raise ValueError(shapes_id)
 
@@ -109,6 +106,7 @@ def generate(
     shapes_id: ShapesId,
     transpose_rhs: bool,
     compilation_info_id: CompilationInfoId,
+    acc: bool,
 ):
     functions = {}
     calls = []
@@ -116,7 +114,7 @@ def generate(
     for compilation_info in get_test_compilation_infos(
         compilation_info_id, lhs_rhs_type
     ):
-        for shape in get_test_shapes(shapes_id):
+        for shape in get_test_shapes(shapes_id, acc):
             for dynamicities in get_dynamicities(shapes_id):
                 function = generate_function(
                     lhs_rhs_type=lhs_rhs_type,
@@ -251,6 +249,13 @@ def parse_arguments():
         help="Custom dynamicity mask for m,n,k. Format: dynamic|static,dynamic|static,dynamic|static (e.g., --mnk_dynamicities=dynamic,static,static)",
         required=False,
     )
+    parser.add_argument(
+        "--acc",
+        action="store_true",
+        help="Customize whether to add an accumulator argument to the function",
+        default=False,
+        required=False,
+    )
     return parser.parse_args()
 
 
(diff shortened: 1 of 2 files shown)


tests/e2e/matmul/generate_e2e_matmul_tests.py
```diff
@@ -18,7 +18,7 @@
 
 # Returns the list of TestShape's to use for the collection of shapes
 # identified by shapes_id.
-def get_test_shapes(shapes_id: ShapesId):
+def get_test_shapes(shapes_id: ShapesId, no_accumulator_arg=False):
```
kuhar: Can you explain what this flag does? Also, would it be more straightforward without the double negation?

tests/e2e/matmul/generate_e2e_matmul_tests.py
```diff
@@ -92,6 +92,8 @@ def get_test_shapes(shapes_id: ShapesId):
         if ShapesId.custom_mnk_values is None:
             raise ValueError("Custom MNK values not set. Use --mnk=m,n,k")
         m, n, k = ShapesId.custom_mnk_values
+        if no_accumulator_arg:
+            return [TestShape(m=m, k=k, n=n, accumulate=False)]
         return [
             TestShape(m=m, k=k, n=n, accumulate=True),
             TestShape(m=m, k=k, n=n, accumulate=False),
```
bjacob: Can you have `result = [TestShape(m=m, k=k, n=n, accumulate=False)]` unconditionally outside of the `if`, and then in the `if` conditionally extend with `TestShape(m=m, k=k, n=n, accumulate=True)` ?  To me it would show more clearly what the boolean switch is doing.

tests/e2e/matmul/generate_e2e_matmul_tests.py
```diff
@@ -251,6 +254,13 @@ def parse_arguments():
         help="Custom dynamicity mask for m,n,k. Format: dynamic|static,dynamic|static,dynamic|static (e.g., --mnk_dynamicities=dynamic,static,static)",
         required=False,
     )
+    parser.add_argument(
+        "--no_accumulator_arg",
```
bjacob: Let's find a better name for this flag. How about undoing the negation (drop the `no_` prefix) and do:

`--accumulate`

Notice I also dropped the `_arg` suffix. It's not just about args passed to the function, it's about changing what the function is doing.

bjacob (changes requested):

tests/e2e/matmul/generate_e2e_matmul_tests.py
Muzammiluddin-Syed-ECE: It would change every test except for the three new ones which is why I was hesitant, but we can do it this way.

bjacob: Don't need to change existing tests! Just need to default to `true`.

krzysz00: You could add `has_accumulatot=True`

tests/e2e/matmul/generate_e2e_matmul_tests.py
Muzammiluddin-Syed-ECE: Resolved in discussion: https://discord.com/channels/689900678990135345/689957613152239638/1446530233632166038

tests/e2e/matmul/generate_e2e_matmul_tests.py
Muzammiluddin-Syed-ECE: Done!

tests/e2e/matmul/generate_e2e_matmul_tests.py
Muzammiluddin-Syed-ECE: The issue is that there are two things we want:
- (1) We want to remove the accumulator tests from the new tests.
- (2) We want the accumulator tests on by default.

Lets say theres some option "x" which adds accumulator tests. 

At the command line level "x" must default to true to not edit other tests. So for all existing tests we omit this argument but for the tests being added we would do "x=false". This means at the point of test generation we need something like
```
add non-accumulator tests
if "x" then
    add accumulator tests.
```

So, the issue is that it is not recommended to [have bool arguments which default to true in argparse](https://discord.com/channels/689900678990135345/689957613152239638/1446532108226007202) (also "x=false" is not allowed at the CLI level). What argparse recommends you do is introduce a new arg called "no-x" which defaults to false instead of having "x" default to true.

What this means is that we now have a variable called "no-x" that defaults to false. So this means at the point of test generation we need something like 
```
add non-accumulator tests
if not "no-x" then 
    add accumulator tests.
```

So, there are two things that look like double negations here:
- "no-x" defaults to false
- the if condition is (not "no-x")

One way to avoid these would be to change what "x" does. For instance if there was some option "y" which removes accumulator tests. That would resolve both double negations but would require something something incredibly awkward like:
```
add non-accumulator tests
add accumulator tests
if "y" then:
    remove accumulator tests
```

So, the main issue is the requirement for "no defaults to true" in argparse and the requirement for us to not edit existing tests.

My workaround for this is to rename "no_x" as something that doesn't immediately scream double negation. Something like "remove_accumulator" which reframes the option from being not an action to its own action. if that makes sense.

krzysz00: So I claim that we don't actually need/want `--accumulator=false` and that it should be `--accumulator` and `--no-accumulator` (the usual Unix style) instead (defaulting to `--accumulator`)

Muzammiluddin-Syed-ECE: This worked thanks @krzysz00 !

tests/e2e/matmul/generate_e2e_matmul_tests.py
```diff
@@ -92,10 +93,14 @@ def get_test_shapes(shapes_id: ShapesId):
         if ShapesId.custom_mnk_values is None:
             raise ValueError("Custom MNK values not set. Use --mnk=m,n,k")
         m, n, k = ShapesId.custom_mnk_values
-        return [
-            TestShape(m=m, k=k, n=n, accumulate=True),
-            TestShape(m=m, k=k, n=n, accumulate=False),
-        ]
+        test_shapes = [TestShape(m=m, k=k, n=n, accumulate=False)]
+        if accumulate:
+            test_shapes.extend(
```
kuhar: I think += also works

kuhar (approved):

bjacob (approved):

PR #22834, opened 2025-12-05
[e2e][ukernel] Remove dead/duplicate tests

Yu-Zhewen:
These matmul ukernel tests were originally introduced in #21383, but they were unintentionally placed under the following condition:

https://github.com/iree-org/iree/blob/b973deabd2e4872c3127702b5c5b4b2b9c6d2992/tests/e2e/matmul/CMakeLists.txt#L2159

This condition conflicts with the test label `requires-gpu-cdna3`, making the tests never run in any configuration.

Later, #22446 added new tests that correctly cover the f16/bf16 cases, so the original ones are now redundant.


jtuyls (approved):
LGTM, nice find!

PR #22835, opened 2025-12-05
[Testing][ROCM][CI] Fix lit test timeout

jtuyls:
[For testing] Reverts the lit changes in https://github.com/iree-org/iree/pull/22801 to check whether the intermittent timeout issues we're seeing are caused by that, for example: https://github.com/iree-org/iree/actions/runs/19911373490/job/57080608780

### DIFF
--- a/compiler/plugins/target/ROCM/test/materialize_encoding_ukernel_gfx942.mlir
+++ b/compiler/plugins/target/ROCM/test/materialize_encoding_ukernel_gfx942.mlir
@@ -1,8 +1,4 @@
-// RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-materialize-device-encoding))" %s | FileCheck %s
-
-// ===----------------------------------------------------------------------===//
-// Shared Attributes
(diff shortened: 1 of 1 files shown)


kuhar (approved):

bjacob (changes requested):
This should be properly fixed by the latest iteration on https://github.com/iree-org/iree/pull/22843 though we're still waiting on CI to confirm.  If not, then at least https://github.com/iree-org/iree/pull/22849 is known to fix it.

jtuyls:
> This should be properly fixed by the latest iteration on #22843 though we're still waiting on CI to confirm. If not, then at least #22849 is known to fix it.

Yeah, this was just meant as a stopgap to avoid the issue interfering with people's work/PRs. I am pretty sure this hides the issue enough to be usable as it removes the parallelism in the tests. We should definitely fix the root cause but if we can't get it to work soon, I think we should merge a temporary solution. Preferably https://github.com/iree-org/iree/pull/22849 indeed, so closing this one.

PR #22836, opened 2025-12-05
Fix tests: `noubsan` was not being honored, and MXFP4 matmul tests are static-shape-only

bjacob:
This gfx950-only test failed to generate due to the static-shape limitation. Fixing that resulted in a [compilation failure](https://gist.github.com/bjacob/5a3015daaa027721337192de99311f8a) due to a stack frame exceeding the max size by ~2x when UBSan is enabled, which is not surprising due to the combination of a very large static shape and a sanitizer. The problem was that the existing `noubsan` test tag was not being honored.


kuhar (approved):

Muzammiluddin-Syed-ECE (approved):
LGTM

PR #22838, opened 2025-12-05
[Encoding] Add verifier for gpu/cpu/vmvx_encoding_resolver

jtuyls:
While debugging an issue in which I got an incompatible encoding with 3 dimensions on a ranked tensor with only 1 dimension, leading to cryptic segfaults/errors, I started adding some verifiers for encoding attributes. This is the first one and I will be adding a couple more.


hanhanW (commented):
LGTM, if we agree that using interface method for verification is better: https://github.com/iree-org/iree/pull/22850#pullrequestreview-3551107968

hanhanW (commented):
I don't touch tablegen for a while, so I'm not sure if we can add such implementation to an interface.

https://github.com/iree-org/iree/blob/d49d410ef5f56828a51c4b80766cb5c8337911f9/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td#L529-L553

If it is doable, it will benefit cpu encoding resolvers and vmvx encoding resolver.

jtuyls:
> I don't touch tablegen for a while, so I'm not sure if we can add such implementation to an interface.
> 
> https://github.com/iree-org/iree/blob/d49d410ef5f56828a51c4b80766cb5c8337911f9/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td#L529-L553
> 
> If it is doable, it will benefit cpu encoding resolvers and vmvx encoding resolver.

I added an interface method for reuse across gpu/cpu/vmvx now, however each resolver attribute still implements the VerifiableTensorEncoding interface separately. I tried to have the PackedLayoutMaterializerAttr inherit the VerifiableTensorEncoding interface directly, but that didn't work due to an 'interface not available' error during compilation. I also expect that some of these implementations will start to diverge as the gpu one could verify the swizzle as well (which I would like to do in a follow-up). Could you have another look?

Note that I also updated the implementation to also check the outerDimsPerm.

hanhanW (approved):

PR #22821, opened 2025-12-04
[Codegen][GPU] Preserve loop domain when collapsing dims in Conv to Matmul conversion

yzhang93:
This PR fixes an issue where `AffineMap::replace` dropped unused dims, leading to the loop-domain mismatch. For example, the following affine map

`affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d4, d1 + d5, d2 + d6, d3)>`

will become

`affine_map<(d0, d1, d2, d3, d4) -> (d4, d1, d2, d3)`

This PR rewrites the affected expressions manually while preserving the original loop domain.

Issue #22032, opened 2025-09-18
[GPU] Support direct convolution TileAndFuse pipeline for channel first layout


yzhang93:
Currently, direct convolution pipeline only supports NHWC layout. For a channel first layout, e.g., (NCHW or CHWN), it fails at `GPUTileAndConvertConvToMatmulPass`.

To reproduce for the following example:
```
func.func @conv_2d_nchw_fchw_fill(%arg0: tensor<2x16x130x130xf32>, %arg1: tensor<32x16x3x3xf32>) -> tensor<2x32x128x128xf32> {
    %cst = arith.constant 0.000000e+00 : f32
    %0 = tensor.empty() : tensor<2x32x128x128xf32>
    %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<2x32x128x128xf32>) -> tensor<2x32x128x128xf32>
    %2 = linalg.conv_2d_nchw_fchw {dilations = dense<1> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>} ins(%arg0, %arg1 : tensor<2x16x130x130xf32>, tensor<32x16x3x3xf32>) outs(%1 : tensor<2x32x128x128xf32>) -> tensor<2x32x128x128xf32>
    return %2 : tensor<2x32x128x128xf32>
  }
```
Command:
```
iree-compile --iree-hal-target-backends=rocm --iree-hip-target=gfx942 --iree-codegen-llvmgpu-use-direct-convolution=true --iree-codegen-llvmgpu-use-igemm=false nchw_test.mlir
```

PR #22840, opened 2025-12-05
[Codegen][GPU] Allow channel first layouts to lower through direct convolution path

yzhang93:
This PR removes some constraints for lowering through direct convolution pipeline, and closes the issue [#22032](https://github.com/iree-org/iree/issues/22032).

- With fix https://github.com/iree-org/iree/pull/22821, now channel first conv layouts, e.g., `nchw` or `chwn` can lower through direct convolution. Adding more tests to validate that.
- As non-unit dilation is natively supported in `GPUTileAndConvertConvToMatmul`, removing the constraint for that as well.


krzysz00 (approved):
Assuming this is good perf-wise, approved (some benchmarks would be nice)

nirvedhmeshram (approved):
LGTM

yzhang93:
@krzysz00 @nirvedhmeshram Performance-wise, many shapes currently underperform compared to the IGEMM path (the default lowering pipeline for convolutions), but the comparison isn’t entirely fair due to different lowering configs. I’ll include some benchmark results in my weekly report today. The goal of this PR isn’t performance-focused, and optimizations are coming next.

krzysz00:
Ok, since this is touching a non-default path, it can't regerses anything, approved

PR #22817, opened 2025-12-03
Add some nvidia tensor core support to TileAndFuse and VectorDistribute pipelines

ziereis:
Adds some nvidia tensor core support  to the newer LLVMGPU pipelines. This PR is a draft to collect some feedback :) I asked a long time ago about this on discord maybe @Groverkss can remember and i finally had some time and motivation to work on this :) 

At its core i added the layouts to support the[ `mma.sync.aligned.m16n8k16` ](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html?highlight=ldmatrix#warp-level-matrix-fragment-mma-16816-float)instruction with fp32 and fp16 accumulators.
In addition to this there are also the following changes on this branch:

- remove some old reference to nvidia tensor core pipelines that no longer exist
- add support for lowering loads from shmem to the matrix fragments with [ldmatrix](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html?highlight=ldmatrix#warp-level-matrix-instructions-ldmatrix) in the VectorDistribute pipeline
- makes the ReduceBankConflicts pass always pad atleast as many elements as the shmem allocation has alignment (ldmatrix loads always need to be aligned to 16 bytes)

if you guys wan't this in some form, i would definitely split this as this mixes a couple of changes together.

Most changes here are actually to support ldmatrix loads (adding only the mma.sync ops requires barely any changes), however from what im seeing so far it seems that this gives basically no performance in the current state and its also the part thats the most "hacky" in the current implementation. 

Would be nice if there is some way to implemented it so both pipelines could use it, but i couldn't really 
think of something that would work without putting too much effort, so i only implemented it for VectorDistribute as it seemed easiest to add there. 

I didn't do much benchmarking yet but at its current state this gives up to ~7-8x improvements for larger matmul shapes on my pc with a RTX 3070 (sm_86).

### Benchmark results on latest main (Contract Config)
```
====================================================================================================
Size                    Time (us)       Stddev     CV (%)       TFLOPS   Iterations
====================================================================================================
512x512x512                 98.50         0.64       0.65         2.73            5
1024x1024x1024             482.00         0.48       0.10         4.46            5
2048x2048x2048            3448.00        13.80       0.40         4.98            5
4096x4096x4096           25053.00        33.70       0.13         5.49            5
8192x8192x8192          196089.00       230.00       0.12         5.61            5
====================================================================================================
```
### Benchmark results for mma.sync with TileAndFuse (ldmatrix not implemented)
```
====================================================================================================
[...]

PR #22847, opened 2025-12-06
Add layout for mma.sync.m16n8k16

ziereis:
This PR adds only the mma layouts from here https://github.com/iree-org/iree/pull/22817 and some more tests.


compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -652,6 +671,20 @@ static VectorType getThreadVectorType(MLIRContext *context,
...
+      if (operandIndex == kMMAOperandLhs) {
+        return VectorType::get({4, 2}, elemType);
+      }
+      if (operandIndex == kMMAOperandRhs) {
+        return VectorType::get({2, 2}, elemType);
+      }
+      if (operandIndex == kMMAOperandAcc) {
+        return VectorType::get({2, 2}, elemType);
+      }
+    }
```
kuhar, line 690: Why can't this be calculated based of single subgroup layout?

ziereis: it can, have not though about this. Thanks!

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -652,6 +671,13 @@ static VectorType getThreadVectorType(MLIRContext *context,
   Type elemType = isIntrinsicLhs<MMAIntrinsicType>(operandIndex)   ? o.aType
                   : isIntrinsicRhs<MMAIntrinsicType>(operandIndex) ? o.bType
                                                                    : o.cType;
+  if constexpr (std::is_same_v<MMAIntrinsicType, MMAIntrinsic>) {
+    if (intrinsic == MMAIntrinsic::NV_MMA_SYNC_F32_16x8x16_F16 ||
+        intrinsic == MMAIntrinsic::NV_MMA_SYNC_F16_16x8x16_F16) {
+      return VectorType::get(
```
krzysz00, line 688: Why's this special-case needed?

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -749,6 +775,27 @@ static Value createMmaOp(OpBuilder &builder, Location loc,
                                   layout.nSize, layout.kSize, lhs, rhs, acc)
         .getResult();
   }
+  if (intrinsic == MMAIntrinsic::NV_MMA_SYNC_F32_16x8x16_F16 ||
+      intrinsic == MMAIntrinsic::NV_MMA_SYNC_F16_16x8x16_F16) {
+    // Transpose the two outer dimensions to model the column-major register
+    // ordering expected by mma.sync. The input shape differs between pipelines:
```
krzysz00, line 792: I don't know if this should be needed - I'd expect this can be modeled at a higher level by setting up the thread strides in getSingleSubgroupLayout. (the shape_cast is fine, though - it's the transpose I'm dubious about)

krzysz00 (commented):
Some notes on the codegen parts of this, thanks for poking at this and apologies for the review lag

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
ziereis: nvgpu intrinsics require specific shapes. https://mlir.llvm.org/docs/Dialects/NVGPU/#nvgpummasync-nvgpummasyncop. 

The existing logic would just linearize the shape to something that would not work with the intrinsic.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
ziereis: I think the SingleSubgroupLayout would need and additional outer "order" or strides if wan't to represent this.

The problem i have here is that i have an outer tile of 2x2 which will be placed in row major order (cant control this in the SingleSubgroupLayout), however the intrinsic requires column major ordering. 

I could try to add a "outer_stride" to the SingleSubgroupLayout if you think its the right way to solve this :)

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_contract_nvgpu.mlir
```diff
@@ -0,0 +1,163 @@
+// RUN: iree-opt --split-input-file --iree-transform-dialect-interpreter --canonicalize --cse %s | FileCheck %s
+
+// NV_MMA_SYNC_F32_16x8x16_F16: M=16, N=8, K=16, subgroup_size=32
+// From IREEGPUAttrs.cpp getSingleSubgroupLayout:
+//   LHS: outer=[2,2], thread=[8,4], strides=[4,1], element=[1,2]
+//   RHS: outer=[2,1], thread=[4,8], strides=[1,4], element=[2,1]
```
krzysz00, line 6: Is this meant to be 2, 2? Otherwise the numbers don't add up

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
krzysz00: Yeah, an extension to the single subgroup layout machinery might be called for here.

Or, alternatively, if you can set up the elements field and all you'd get is a reassociation along the K dimension ... that's quite fine, I'd argue (or at least there's precedent for it[

The reason I don't want to sneak in transposes late is because there're a bunch of passes that would like  to see this detail of the layout but won't get to.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_contract_nvgpu.mlir
ziereis: No its intentionally [2,1] The RHS fragment is only 8x16 and basically a vertical slice of the LHS fragment. 

ref: https://docs.nvidia.com/cuda/parallel-thread-execution/index.html?highlight=mma%2520sync#warp-level-matrix-fragment-mma-16816-float

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
ziereis: > Or, alternatively, if you can set up the elements field and all you'd get is a reassociation along the K dimension 

Could you elaborate on this a bit please? Its not clear to me what you mean by this.

Regarding the outer_strides,  i did look into this a bit on the weekend and could open a "draftish" PR for this or merge it into here if you want. However this might be more suited for a follow up, since this would change a lot of unrelated code to the changes here.

However i would like to clarify a couple of things just to be sure im not going into the wrong direction, especially about the TileAndFuse pipeline. 

~To my understanding the earliest point where i could introduce the transpose is after~ ~DistributeInnerTiledToLanesPass, there i would have a LHS that feeds into the inner_tiled op that is shaped~ ~like this~ `tensor<2x8x2x1x2x2xf16>`  ~where i would assume the shape corresponds to this: [XZY, XZY, OUTER, ELEM, OUTER, ELEM]~

~Then i could modify this pass to read the outer_strides, check if they are col major and insert a [0,1,4,5,2,3] transpose here.~

They way i implemented it now is by inserting the a [2, 1, 0, 3] transpose of the lhs`tensor<2x8x2x8xf16>`  in the ExpandUndistributedInnerTiles pass since i guess this is really the earliest where the outer shape is materialized. However its not quite apparent to me what the rest of the pipeline can now do with this transpose? Also did not properly look into how to handle this with the VectorDistributePipeline.

Is this inline with what you were thinking?

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -418,9 +418,12 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
   if (pipelineOptions.useIgemmConvolution) {
     funcPassManager.addPass(createConvolutionToIGEMMPass());
   }
+  // TODO: this shouldn't be required
```
ziereis: Some note on this:

I had to add this after https://github.com/iree-org/iree/commit/55430fded30d7ef4ab74ebb9659c571c1d07cca3, otherwise the matmul-accumulate TileAndFuse test case would no longer bufferize. I got a rough idea what the real problem is and i describe it with some IR here (https://gist.github.com/ziereis/3fe1d6077267c935f24299be3c81c3b0) however i couldn't really think of some good way to fix this.

ziereis:
@qedawkins @bjacob @kuhar @krzysz00 no idea if i'm pinging the right people but would be awesome if could get some more feedback on this. This is open now for quite a while and I would like to look more into codegen for nvidia but i feel like this is the bare minimum that has to exist before i can explore enabling more of the hardware.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPUDistributeContract.cpp
```diff
@@ -127,11 +127,11 @@ struct DistributeContract final : OpDistributionPattern<vector::ContractionOp> {
...
     // to use and it is attached as an attribute to this contract op.
-    auto mmaKind = contractOp->getAttrOfType<IREE::GPU::MmaInterfaceAttr>(
-        "iree.amdgpu.mma");
+    auto mmaKind =
+        contractOp->getAttrOfType<IREE::GPU::MmaInterfaceAttr>("iree.gpu.mma");
     if (!mmaKind) {
       return rewriter.notifyMatchFailure(
-          contractOp, "missing iree.amdgpu.mma intrinsic attribute");
+          contractOp, "missing iree.gpu.mma intrinsic attribute");
     }
```
Groverkss, line 136: It's a bit weird that this pass is called AMDGPUDistributeContract, but that's from days when we had to hardcode the mfma checks here. This is good, but eventually we should just move this pattern elsewhere.

Groverkss (approved):
VectorDistribute changes LGTM, please wait for others for reviews on GPU dialect stuff

Groverkss:
> @qedawkins @bjacob @kuhar @krzysz00 no idea if i'm pinging the right people but would be awesome if could get some more feedback on this. This is open now for quite a while and I would like to look more into codegen for nvidia but i feel like this is the bare minimum that has to exist before i can explore enabling more of the hardware.

Sorry for leaving this hanging. I pinged people on discord as well. I'll create a thread on discord where you can ping people if needed for more patches as well.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPUDistributeContract.cpp
Groverkss: Note that the reason I'm not asking to change the pass name is because I'm going to delete this pass soon anyway and switch VectorDistribute to use inner_tiled for codegen.

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -85,6 +85,10 @@ def ConcretizePadResultShapePass :
 def ConvertAccGEMMToGEMMPass :
     InterfacePass<"iree-convert-accgemm-to-gemm", "mlir::FunctionOpInterface"> {
   let summary = "Convert accumulating GEMMs to GEMMs post dispatch creation.";
+  let options = [
+    Option<"forceConversion", "force-conversion", "bool", /*default=*/"false",
+           "Force conversion even when the accumulator is read/written from the same buffer.">
+  ];
```
krzysz00: I assume this is for testing? If so, we should probably mark it as such?

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -564,14 +582,15 @@ struct OpaqueMmaLayout {
...
   }
-  return getUnsupportedMNKShape(intrinsic);
+  auto lhs = getSingleSubgroupLayout(intrinsic, kMMAOperandLhs);
+  auto rhs = getSingleSubgroupLayout(intrinsic, kMMAOperandRhs);
+  return {lhs.outer[0] * lhs.thread[0] * lhs.element[0],
+          rhs.outer[1] * rhs.thread[1] * rhs.element[1],
+          lhs.outer[1] * lhs.thread[1] * lhs.element[1]};
 }
 
 int64_t getMSize(MMAIntrinsic intrinsic) {
```
krzysz00, line 605: clang-tidy nit here

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
krzysz00: I see, this makes sense

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
krzysz00: Ok, now that I'm looking at this again, I think the current approach actually works out quite reasonably. Our vectors are fundamentally slices of the logical shape, and if the intrinsic wants them to be transposed from that, here would be a good place to do it, in the same way that we flatten things for `amdgpu.mfma` in this pass. Sorry about the confusion here

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
krzysz00: Could you pull this into an issue for followup?

krzysz00 (approved):
The changes to the mma attribute lgtm, apologies for the holdup

Now the `forecConversion` thing I'm very dubious about but I don't have immediate context for - @qedawkins @Max191 @MaheshRavishankar to those comments

compiler/src/iree/compiler/Codegen/Common/Passes.td
krzysz00: Note that there's been some discussion of what you've found on the Discord thread about this PR in case you haven't seen it yet

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
qedawkins: Yeah this isn't right. Even if there was an mma sync specific issue that was worth deferring, we wouldn't add it as a pipeline level switch like this. Discussion here on discord: https://discord.com/channels/689900678990135345/1467004077438271624/1467929727921688639

But TLDR the IR you shared looks fine to me, even after the point you're saying it looks wrong. When fusing the producing transpose we *can't* set the destination to the untransposed input. That transiently creates a race condition with loading and writing different data to an aliased destination. Perhaps something else is going wrong anyway because there is a follow up transpose that should fixup the destination.

ziereis: As per discussion on discord i moved this to a string attr on the lowering config which i set if the intrinsic type is one of the affected ones. I created this issue here as followup: https://github.com/iree-org/iree/issues/23370

compiler/src/iree/compiler/Codegen/Common/ConvertAccGEMMToGEMMPass.cpp
```diff
@@ -129,11 +136,13 @@ namespace {
 
 struct ConvertAccGEMMToGEMMPass final
     : impl::ConvertAccGEMMToGEMMPassBase<ConvertAccGEMMToGEMMPass> {
+  using impl::ConvertAccGEMMToGEMMPassBase<
+      ConvertAccGEMMToGEMMPass>::ConvertAccGEMMToGEMMPassBase;
```
kuhar:
```suggestion
  using Base::Base;
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -753,6 +786,26 @@ static Value createMmaOp(OpBuilder &builder, Location loc,
...
+    // VectorDistribute produces 2x2x1x2, TileAndFuse produces 2x1x2x2.
+    // Remove the unit dimension to simplify the transpose.
+    auto nonUnitVecType = VectorType::get({2, 2, 2}, builder.getF16Type());
+    auto reshaped =
+        vector::ShapeCastOp::create(builder, loc, nonUnitVecType, lhs);
+    auto permAttr = builder.getDenseI64ArrayAttr({1, 0, 2});
+    auto transposed = vector::TransposeOp::create(builder, loc, nonUnitVecType,
+                                                  reshaped, permAttr);
+    lhs = vector::ShapeCastOp::create(builder, loc, lhs.getType(), transposed);
+    SmallVector<Attribute> mmaShape{builder.getI64IntegerAttr(layout.mSize),
```
kuhar: Use a plain array, since we won't need to dynamically resize this
```suggestion
    Attribute mmaShape[] = {builder.getI64IntegerAttr(layout.mSize),
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.h
```diff
@@ -222,6 +222,9 @@ int64_t getKSize(MMAIntrinsic intrinsic);
 /// Returns the subgroup size used by the given MMA intrinsic.
 int64_t getIntrinsicSubgroupSize(MMAIntrinsic intrinsic);
 
+/// Returns true if the given MMA intrinsic is an NV_MMA_SYNC intrinsic
```
kuhar:
```suggestion
/// Returns true if the given MMA intrinsic is an NV_MMA_SYNC intrinsic.
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -926,6 +926,15 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
       {"subgroup", b.getI64ArrayAttr(subgroupTileSizes)},
       {"mma_kind", kind}};
 
+  // TODO(23370)
+  // At the moment we always have to convert accumulating gemms
+  // when using NV_MMA_SYNC intrinsics (NVIDIA mma.sync).
```
```suggestion
  // TODO(#23370): At the moment we always have to convert accumulating gemms
  // when using NV_MMA_SYNC intrinsics (NVIDIA mma.sync).
```

compiler/src/iree/compiler/Codegen/Common/Passes.td
ziereis: its removed now as per discussion on discord

kuhar (approved):
LGTM % a few clang-tidy warnings

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
ziereis: im pretty sure this is a false positive. This function is in the header and is used in other cpp files. not sure how it could be made static? :D

ziereis:
Thank you for all the reviews.

I could resolve some of the clang-tidy warning but the `misc-use-internal-linkage` ones that are left all seem like false positives to me.

kuhar:
> I could resolve some of the clang-tidy warning but the `misc-use-internal-linkage` ones that are left all seem like false positives to me.

yeah these are known false-positives in the diff mode

krzysz00:
Since we've got approvals and the only CI failures are the known-unrelated ones, @ziereis I think we can merge - if you don't have the rights to, ping me and I'll do it US tomorrow morning

ziereis:
@krzysz00 yes i can't merge myself. Would be very much appreciated if you could do it.

PR #22848, opened 2025-12-06
Add CDNA3 test filtering to CI

lialan:
Tests with `requires-gpu-cdna3` label lacked the filtering mechanism in `ctest_all.sh`. It could leak to run on non-cdna3 machines.

This adds `IREE_AMD_CDNA3_TESTS_DISABLE`: disabled by default, excludes labeled tests when set.

It should also enables CDNA3 tests in MI325 workflow.


ScottTodd (approved):
Please resolve the failing DCO check: https://github.com/iree-org/iree/pull/22848/checks?check_run_id=57339468971

PR #20507, opened 2025-04-09
[DT] Split out iree_encoding.layout from EncodingAttr.

hanhanW:
The revision implements `IREE::Encoding::LayoutAttr` and retires the `layouts` field from EncodingAttr, as it encodes much more information than needed.

It is not easy to do incremental because many layers are involved, including
- Encoding specialization.
- Encoding materialization.
- Lower `iree_codegen.query_tile_size` op for VMVX ukernel path.

To make review easier, there are five commits in the PR:
- Introduce iree_encoding.layout encoding attribute: https://github.com/iree-org/iree/commit/85d919941ebc9ea1fa17de77e2716d3f332236f4
- Implement SerializableEncodingInterface for Encoding::LayoutAttr: https://github.com/iree-org/iree/commit/a49a7263b471f8f8112e083b53e97540a6cf70b3
- Use iree_encoding.layout when the layouts are resolved: https://github.com/iree-org/iree/commit/1dc9007118a27ec32508277944acad373760a96c
- Retire the layouts field from EncodingAttr: https://github.com/iree-org/iree/commit/b2f0329047c4c819dd9c0fa85581a933c99969b3
- Fix the ukernel ops lowering for VMVX: https://github.com/iree-org/iree/commit/0a51344b21512329c3b82ba74ff696241ce0cf6f

In the third commit, we need to move `registerAttributes` to `EncodingAttr.cpp` because we want to skip default builders. 
 It is very tricky when the verifier is implemented, because an assertion is triggered. However, it should just return nullptr in this context. To bypass the crash in get method, we need to implement our own get methods.

Note: the root cause of `reigsterAtrributes` is that the full definition of the storage class needs to be visible whenever we register the attribute. 

Fixes https://github.com/iree-org/iree/issues/20489

PR #22850, opened 2025-12-08
[Encoding] Add verifier for iree_encoding.layouts

jtuyls:
Related to https://github.com/iree-org/iree/pull/22838, this PR adds a verifier for `#iree_encoding.layouts`.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp
@@ -93,6 +93,19 @@ Value LayoutAttr::calculateStorageSizeInBytes(Location loc, OpBuilder &builder,
   return res;
 }
 
+LogicalResult LayoutAttr::verifyEncoding(
+    llvm::ArrayRef<int64_t> shape, mlir::Type elementType,
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError) const {
+  for (Attribute attr : getLayouts()) {
+    auto verifiableAttr = dyn_cast<VerifiableTensorEncoding>(attr);
+    if (verifiableAttr &&
+        failed(verifiableAttr.verifyEncoding(shape, elementType, emitError))) {
+      return failure();
+    }
+  }
+  return success();
+}
+
 //===---------------------------------------------------------------------===//
 // iree_encoding.packed_storage
 //===---------------------------------------------------------------------===//
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.td
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.td
@@ -51,7 +51,8 @@ def LayoutAttr :
         "isSerialized",
         "isIdentityLayout",
         "calculateStorageSizeInBytes",
-      ]>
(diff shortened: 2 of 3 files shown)


hanhanW (commented):
Cool, I did not know that we have such interface. I had bad experience when I added verifier to encodings. Is this a better way for verification?

btw, should we move this snippet to the verifier?

https://github.com/iree-org/iree/blob/d49d410ef5f56828a51c4b80766cb5c8337911f9/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp#L52-L64

and we remove the below line?

https://github.com/iree-org/iree/blob/d49d410ef5f56828a51c4b80766cb5c8337911f9/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.td#L79

hanhanW:
here was my bad experience, btw: https://github.com/iree-org/iree/pull/20507#issuecomment-2790591646

jtuyls:
> Cool, I did not know that we have such interface. I had bad experience when I added verifier to encodings. Is this a better way for verification?
> 
> btw, should we move this snippet to the verifier?
> 
> https://github.com/iree-org/iree/blob/d49d410ef5f56828a51c4b80766cb5c8337911f9/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.cpp#L52-L64
> 
> and we remove the below line?
> 
> https://github.com/iree-org/iree/blob/d49d410ef5f56828a51c4b80766cb5c8337911f9/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingAttrs.td#L79

The main difference is that `verifyEncoding` is called from and with the surrounding type, so we can check whether the encoding is compatible with the type. We could move that 'internal' verifier in there as well and it might fix your issue as it won't be called from the builder itself anymore and just from the surrounding type. I didn't do that because I thought that we might want to leave the verification that doesn't depend on the type there to perform standalone and earlier verification. However, if we think that it's better to have a single place for all verification, we could move all verification in there as well indeed.

> here was my bad experience, btw: https://github.com/iree-org/iree/pull/20507#issuecomment-2790591646

Do you still know which tests were failing with the default builders? I gave it a try and reverted to default builders but didn't immediately find a failing test.

hanhanW (approved):
> Do you still know which tests were failing with the default builders? I gave it a try and reverted to default builders but didn't immediately find a failing test.

IIRC, it happens only if you create an invalid attribute: see https://github.com/llvm/llvm-project/blob/49496c531da2d1acf0d1e2e74dd84c190347c039/mlir/include/mlir/IR/StorageUniquerSupport.h#L173-L182

We don't need to invest this much, as it is not that important.

> The main difference is that verifyEncoding is called from and with the surrounding type, so we can check whether the encoding is compatible with the type. We could move that 'internal' verifier in there as well and it might fix your issue as it won't be called from the builder itself anymore and just from the surrounding type. I didn't do that because I thought that we might want to leave the verification that doesn't depend on the type there to perform standalone and earlier verification. However, if we think that it's better to have a single place for all verification, we could move all verification in there as well indeed.

I see, thanks for all the details, and I see the pros and cons for different approaches. I think the current way looks reasonable to me. It is not guaranteed that all the types that allow encoding will call the verifier; basic verification seems required for the encoding.

PR #22861, opened 2025-12-09
[Codegen][GPU] Apply FoldTensorEmptyPatterns with single use only

### LINKED ISSUE #22833: [regression-onnx] function 'main_graph$async_dispatch_0_conv_512x48x48x256x2x2_f32' uses 73728 bytes of shared memory; exceeded the limit of 65536 bytes
### What happened?

Due to https://github.com/iree-org/iree/commit/24b79f1daa30ab0412c20847db900ca96b9dd9a0 we are seeing regression in iree-compile for following simple IR

```mlir
module {
  func.func @main_graph(%arg1: !torch.vtensor<[1,256,96,96],f32>) -> !torch.vtensor<[1,512,48,48],f32> attributes {torch.onnx_meta.ir_version = 8 : si64, torch.onnx_meta.opset_version = 21 : si64, torch.onnx_meta.producer_name = "pytorch", torch.onnx_meta.producer_version = "2.1.0"} {
    %24 = torch.operator "onnx.Constant"() {torch.onnx.value = dense<1.0> : tensor<512x256x2x2xf32>} : () -> !torch.vtensor<[512,256,2,2],f32> 
    %25 = torch.operator "onnx.Constant"() {torch.onnx.value = dense<1.0> : tensor<512xf32>} : () -> !torch.vtensor<[512],f32> 
    %405 = torch.operator "onnx.Conv"(%arg1, %24, %25) {torch.onnx.dilations = [1 : si64, 1 : si64], torch.onnx.group = 1 : si64, torch.onnx.kernel_shape = [2 : si64, 2 : si64], torch.onnx.pads = [0 : si64, 0 : si64, 0 : si64, 0 : si64], torch.onnx.strides = [2 : si64, 2 : si64]} : (!torch.vtensor<[1,256,96,96],f32>, !torch.vtensor<[512,256,2,2],f32>, !torch.vtensor<[512],f32>) -> !torch.vtensor<[1,512,48,48],f32> 
    return %405 : !torch.vtensor<[1,512,48,48],f32>
  }
}
```

command : 
```
iree-compile test.mlir --iree-hal-target-backends=rocm --iree-hip-target=gfx942 -o test.vmfb
```

_No response_

### Version information

_No response_

### Additional context

_No response_

yzhang93:
This PR is to fix https://github.com/iree-org/iree/issues/22833.

Without this fix, the `extract_slice` was fused into the scf.forall loop while generating an extra empty tensor inside the loop. As a result, it lowered to an unnecessary allocation after bufferization.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -511,7 +511,8 @@ combineLayoutTransformation(MLIRContext *ctx, FunctionOpInterface funcOp,
                             CombineRelayoutOpsControlFnRef controlFn) {
   // Sink relayout operations to the end of the funcOp.
   RewritePatternSet propagationPatterns(ctx);
-  tensor::populateFoldTensorEmptyPatterns(propagationPatterns);
+  tensor::populateFoldTensorEmptyPatterns(propagationPatterns,
+                                          /*foldSingleUseOnly=*/true);
   tensor::ExpandShapeOp::getCanonicalizationPatterns(propagationPatterns, ctx);
   tensor::CollapseShapeOp::getCanonicalizationPatterns(propagationPatterns,
                                                        ctx);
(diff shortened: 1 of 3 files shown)


yzhang93:
Looks like this change would break other convolution tests. @Max191 Do you have any alternative suggestion to fix the issue while not break other tests?

For now, I will try to disable swapping `collapse_shape` and `extract_slice` ops for such case (when the `extract_slice` user is a `tensor.parallel_insert_slice`).

PR #22867, opened 2025-12-09
[Encoding] Add resolver swizzle verification

jtuyls:
Add swizzle verification as a follow up on the encoding verification added in https://github.com/iree-org/iree/pull/22838

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.cpp
@@ -7,9 +7,41 @@
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenDialect.h"
 #include "llvm/ADT/TypeSwitch.h"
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.cpp
```diff
@@ -6,10 +6,51 @@
...
+    if (permIdx < 0 || permIdx >= totalExpandedDims) {
+      return emitError() << "swizzle permutation index " << permIdx
+                         << " is out of bounds for " << totalExpandedDims
+                         << " expanded dimensions";
+    }
+    if (!seenIndices.insert(permIdx).second) {
+      return emitError() << "swizzle permutation contains duplicate index "
+                         << permIdx;
+    }
+  }
```
hanhanW: We can reuse the [isPermutationVector](https://github.com/llvm/llvm-project/blob/6960b633ee7633d0ed7e9853baea296dbe201ab2/mlir/include/mlir/Dialect/Utils/IndexingUtils.h#L237-L238). I think we don't need the detailed error message. Saying that it is not a permutation looks solid to me.

jtuyls: Thanks, that's better!

hanhanW (approved):

PR #22868, opened 2025-12-09
[Codegen][GPU]Skip prologue pipeline barriers only for non-nested pipelined loops

### LINKED ISSUE #22857: Regression in GEMM/Conv performance wrt to November 4
For this configuration
```
convbfp16 -n 1 -c 896 -H 59 -W 91 -k 896 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC -m conv -g 16 -F 1 -t 1
```
on MI300x if you checkout this old commit from Nov 4
```
git reset --hard c65dc6dc28 
git submodule update --init
```
it gives a mean time of 47.11 us, however on top of main today we get 72.26 us

On looking at the IR dump, the IR seems identical up to the prefetchSharedMemory pass and then diverges (not to say that is the issue but thats where we would need to start investigating)

Here are the dumps [old](https://sharkpublic.blob.core.windows.net/sharkpublic/nirvedh/output_old.mlir) and [new](https://sharkpublic.blob.core.windows.net/sharkpublic/nirvedh/output_curr.mlir)

jerryyin:
Add isInsideLoop() check to conditionally insert prologue barriers. Fixes potential non-determinism issues when pipelined loops are nested. Fixes: https://github.com/iree-org/iree/issues/22857. Also tested with https://github.com/iree-org/iree/issues/22468 and confirm no regression.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -596,6 +596,18 @@ static SharedBarrierState insertBarriersInRange(RewriterBase &rewriter,
   return state;
 }
 
+// Check if the operation is inside a loop (scf.for, scf.while, etc.)
+static bool isInsideLoop(Operation *op) {
+  Operation *parent = op->getParentOp();
+  while (parent) {
+    if (isa<scf::ForOp, scf::WhileOp, scf::ParallelOp>(parent)) {
+      return true;
+    }
+    parent = parent->getParentOp();
+  }
+  return false;
+}
+
 // Inserts synchronization barriers in the pipelined loop.
 static void insertPipelineBarriers(RewriterBase &rewriter,
                                    scf::ForOp newForOp) {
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -596,17 +596,48 @@ static SharedBarrierState insertBarriersInRange(RewriterBase &rewriter,
...
+  //   - Without a barrier, there's a data race between iterations
+  bool isNested = isInsideLoop(newForOp);
+
+  if (isNested) {
+    // Nested loop: insert barriers in prologue for correctness.
+    // Start with needBarrierBeforeWrite=true because the epilogue of the
+    // previous outer iteration may have read from shared memory.
+    state.needBarrierBeforeWrite = true;
+    state = insertBarriersInRange(rewriter, loc, parentBlock->begin(),
+                                  newForOp->getIterator(), state);
```
nirvedhmeshram, line 631: I think there is no lit test for this part of the logic, would be good to add one I think?

nirvedhmeshram (approved):
looks good % minor ask for extra test.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Yah good point

PR #22870, opened 2025-12-09
[Codegen] Do not swap extract_slice and collapse_shape for a special case

yzhang93:
This is a stop gap fix for https://github.com/iree-org/iree/issues/22833.

In this special case, the `collapse_shape` is from `unpack` op decomposition and there is no benefit to swap it with `extract_slice`. `collapse_shape` is then used as a blocker to avoid fusing `extract_slice` into `scf.forall`. Otherwise, an extra private allocation may be created. Some attempts were made to remove the extra `tensor.empty`, however, some other tests would fail as shown in https://github.com/iree-org/iree/pull/22861.


Max191 (commented):
If this is coming from unpack decomposition, can you achieve the same effect by avoiding the swap when the extract_slice is in the same Block as the collapse_shape?

My understanding of the pattern is that it is intended to be fusing the collapse_shape into the loop. For such cases, the ops should always be in different Blocks (one in the loop, and one outside of it).

Doing this would be less fragile, since we wouldn't need to match on specific ops.

Max191 (approved):

PR #22871, opened 2025-12-09
(Do not review) Play with transparent huge pages

bjacob:
This fails at the moment with `madvise` returning `EINVAL`, but this could change in the future with filesystems gaining support for huge pages. This is potentially a large performance improvement, so it's worth monitoring. https://discord.com/channels/689900678990135345/760577505840463893/1448036330591490293


bjacob:
@benvanik , one thing that I noticed was that `IREE_IO_FILE_MAPPING_FLAG_PRIVATE` was missing in a couple places. Worth taking separately ?

PR #22227, opened 2025-10-07
[python] Set up binding for preprocessing transform ops

bangtianliu:
This PR adds Python bindings for the Preprocessing Transform Extension ops.
It exposes the transform ops defined in that extension, such as `transform.iree.match.contraction`, as well as other commonly used ops like `transform.iree.match.cast_compatible_dag_from_root `and `transform.iree.match.dim_is_multiple_of`, etc.

The bindings are exposed using `declare_mlir_dialect_extension_python_bindings`. The logic mainly follows this llvm upstream PR: https://github.com/llvm/llvm-project/pull/159450.

PR #22872, opened 2025-12-09
[bindings] Add `iree_tensor_ext` to python bindings

rkayaith:
This is for use downstream in iree-turbine.

### DIFF
--- a/compiler/bindings/python/test/ir/dialects_test.py
+++ b/compiler/bindings/python/test/ir/dialects_test.py
@@ -58,6 +58,7 @@ from iree.compiler.dialects import (
     util,
     iree_codegen,
     iree_gpu,
+    iree_tensor_ext,
     preprocessing_transform,
 )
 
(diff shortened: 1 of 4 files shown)


compiler/bindings/python/test/ir/dialects_test.py
```diff
@@ -58,6 +58,7 @@ def decorator_builder(func):
     util,
     iree_codegen,
     iree_gpu,
+    iree_tensor_ext,
```
bangtianliu, line 61: Could you also add python tests here. 

Here is what I did: https://github.com/iree-org/iree/pull/22227/files#diff-75fe861126ab4077da88ce04af112bec6ee413b090b24036b4180a05cc19aff2R622

rkayaith: I added a very basic test: https://github.com/iree-org/iree/pull/22872/files#diff-75fe861126ab4077da88ce04af112bec6ee413b090b24036b4180a05cc19aff2R699-R702

Similar to `flow`, `stream`, etc. dialects, we're only using the auto-generated op bindings, so I don't think in depth testing is necessary

bangtianliu: Sure

bangtianliu (approved):

Issue #22528, opened 2025-11-03
Layernorm compiles to 3 dispatches instead of 1


rkayaith:
### What happened?

IR and dump: https://gist.github.com/rkayaith/e6f563114dff28385551dcc912694652
compile command: `iree-compile --iree-hal-target-backends=rocm --iree-hip-target=mi300x --iree-opt-level=O3`

### What component(s) does this issue relate to?

Compiler

### Version information

1e2abb8c

### Additional context

IREE BOO command: `iree-boo-driver layernorm --input 10x32x42x32`

IanWood1:
I took another look at this, and it looks like it was slightly improved to form two dispatches instead of three. Here is a small reproducer for why this still isn't forming a single dispatch. This reproducer will form 2 dispatches: one for `%3` and one for `%4`.


```mlir
util.func public @main(%arg0: tensor<32x32xf32>) -> (tensor<32xf32>, tensor<1024xf32>) {
  %cst = arith.constant 1.000000e+00 : f32
  %0 = tensor.empty() : tensor<32x32xf32>
  %1 = tensor.empty() : tensor<32xf32>
  %cst_0 = arith.constant 0.000000e+00 : f32
  %2 = linalg.fill ins(%cst_0 : f32) outs(%1 : tensor<32xf32>) -> tensor<32xf32>
  %3 = linalg.generic {indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = ["parallel", "parallel"]} ins(%arg0 : tensor<32x32xf32>) outs(%0 : tensor<32x32xf32>) {
  ^bb0(%in: f32, %out: f32):
    %5 = arith.addf %in, %cst : f32
    linalg.yield %5 : f32
  } -> tensor<32x32xf32>
  %collapsed = tensor.collapse_shape %3 [[0, 1]] : tensor<32x32xf32> into tensor<1024xf32>
  %4 = linalg.generic {indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0)>], iterator_types = ["parallel", "reduction"]} ins(%3 : tensor<32x32xf32>) outs(%2 : tensor<32xf32>) {
  ^bb0(%in: f32, %out: f32):
    %5 = arith.addf %in, %out : f32
    linalg.yield %5 : f32
  } -> tensor<32xf32>
[...]

Issue #22462, opened 2025-10-29
Enable fusion when producers have multiple uses


IanWood1:
We currently only try to fuse with a single consumer (except in the case of truncate consumers). This should be generalized to allow fusing with all users.

https://github.com/iree-org/iree/blob/ef4cde174622643e6534e098830b92e891b4bf7c/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp#L652-L661

Issues to resolve before this can be enabled:
- [x] https://github.com/iree-org/iree/pull/22461
- [x] https://github.com/iree-org/iree/issues/22576
- [x] https://github.com/iree-org/iree/pull/22011 Multi-use fusion can result in much larger dispatches. We need to limit the number of operands to a dispatch.

PR #22874, opened 2025-12-09
[Dispatch Creation] Ignore unfusable consumer

IanWood1:
This constraint was added back when we needed to fuse in dominance-order. It was possible to generate illegal IR if we skipped a consumer. This is now handled and the "break" can be removed after we now form multi-use dispatches https://github.com/iree-org/iree/issues/22462.

Results in 2 dispatches instead of 3 for https://github.com/iree-org/iree/issues/22528

ci-extra: test_torch


MaheshRavishankar (approved):
Nice!

PR #22884, opened 2025-12-11
[AMDGPU][LDS] Adding 1k, 2k, 4k, 8k static shape e2e tests for coalesced gather DMA op

lialan:
As the title mentioned, just adding a bunch of DMA e2e tests. Only for compile-time static shapes.

### DIFF
--- a/tests/e2e/matmul/CMakeLists.txt
+++ b/tests/e2e/matmul/CMakeLists.txt
@@ -1308,6 +1308,122 @@ iree_generated_e2e_runner_test(
 
 iree_generated_e2e_runner_test(
   NAME
+    e2e_matmul_${_CDNA_ARCH}_coalesced_dma_f32_1k
+  TEST_TYPE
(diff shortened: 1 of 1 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUPromoteMatmulOperands.cpp
```diff
@@ -30,6 +31,27 @@ namespace mlir::iree_compiler {
 #include "iree/compiler/Codegen/Common/GPU/Passes.h.inc"
 
 namespace {
+
+/// Check if a value originates from a tensor.pad operation by tracing back
+/// through the definition chain.
+static bool isFromPadOp(Value v) {
```
lialan: Not sure if it is enough to trace down the origin pad.

bjacob (approved):

PR #22886, opened 2025-12-11
[Stream] Update dispatch sites encodings in UnifyEncodingForGlobals pass 2/n

hanhanW:
Added functionality to update stream.tensor.dispatch operand encodings when unifying encodings for globals.

- Uses Explorer with walkTransitiveUses for cross-function tracking of encoded global uses.
- Follows stream.async.clone ops (BFS worklist) to collect updates. Currently, only stream.tensor.dispatch op is updated. Other ops that have TensorPhaseOp are not implemented yet.
- Applies all updates in one shot using `DenseMap<Operation*, DenseMap<unsigned, Attribute>>`. This makes update more efficient because we don't have to iterate all the encoding operands in every update.

New tests:
- Basic dispatch with encoded operand.
- Multiple encoded operands in same dispatch (@use_both_encodings).
- Cross-function tracking via `util.call` (@cross_function_call → @dispatch_helper).

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -284,30 +284,28 @@ static void applyTensorEncodingUpdates(TensorEncodingUpdates &updates) {
       continue;
     }
 
+    // The operand_encodings attribute has the same length as getMixedOperands().
+    // For non-affinity types (e.g., index), the encoding is just the type.
+    // For affinity types, the encoding is a RankedTensorType with encoding attr.
     SmallVector<Attribute> newOperandEncodings;
-    ArrayRef<Attribute> oldOperandEncodings =
-        dispatchOp.getOperandEncodings().getValue();
-
-    // The operand_encodings attribute only contains entries for resource
-    // operands (not index/scalar operands). We iterate through all mixed
-    // operands to map from mixed operand index (used in operandUpdates) to
-    // encoding index.
-    for (auto [idx, operand] : llvm::enumerate(dispatchOp.getMixedOperands())) {
-      if (!isa<IREE::Stream::ResourceType>(operand.getType())) {
+    for (auto [idx, operand, typeAttr] :
+         llvm::enumerate(dispatchOp.getMixedOperands(),
+                         dispatchOp.getOperandEncodings().getValue())) {
+      Type type = cast<TypeAttr>(typeAttr).getValue();
+      if (!isa<IREE::Stream::AffinityTypeInterface>(type)) {
+        newOperandEncodings.push_back(typeAttr);
         continue;
       }
-      Attribute oldEncoding = oldOperandEncodings[newOperandEncodings.size()];
       if (!operandUpdates.contains(idx)) {
-        newOperandEncodings.push_back(oldEncoding);
+        newOperandEncodings.push_back(typeAttr);
         continue;
       }
       Attribute newEncoding = operandUpdates.lookup(idx);
-      auto tensorType =
-          cast<RankedTensorType>(cast<TypeAttr>(oldEncoding).getValue());
+      auto tensorType = cast<RankedTensorType>(type);
       newOperandEncodings.push_back(
           TypeAttr::get(tensorType.cloneWithEncoding(newEncoding)));
-      LDBG() << "  Updated dispatch operand encoding at index "
-             << newOperandEncodings.size() - 1 << " to " << newEncoding;
+      LDBG() << "  Updated dispatch operand encoding at index " << idx
+             << " to " << newEncoding;
     }
     dispatchOp.setOperandEncodingsAttr(
         ArrayAttr::get(dispatchOp.getContext(), newOperandEncodings));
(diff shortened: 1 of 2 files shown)


hanhanW:
I found few issues when I added few more tests. I'll open it after I fix it.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -67,16 +69,126 @@ module @immutable_source_initialized_from_parameter {
...
+        %0 = stream.binding.subspan %arg0[%c0] : !stream.binding -> !iree_tensor_ext.dispatch.tensor<readonly:tensor<4096x4096xf32, #encoding1>>
+        %1 = stream.binding.subspan %arg2[%c0] : !stream.binding -> !iree_tensor_ext.dispatch.tensor<readonly:tensor<4096x4096xf32, #encoding2>>
+        %2 = stream.binding.subspan %arg4[%c0] : !stream.binding -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<16xf32>>
+        return
+      }
+    }
+  }
+
+  // CHECK-LABEL: util.func public @use_encoded_v1
+  util.func public @use_encoded_v1(%arg0: !stream.resource<*>, %arg1: index) -> !stream.resource<*> {
```
sommerlukas: Do we need this function at all? 

With `use_both_encodings` and `cross_function_call` (via `dispatch_helper`), we already have two uses of `#encoding1`. 

I'm generally a bit worried that we have so many uses of the globals in a single test. For example, if we want to test correct tracking across functions in `cross_function_call` and `dispatch_helper`, I think that should be the only use of one of the encodings. Otherwise, the other functions could already trigger unification of the encodings, even if the cross-functional tracking fails. 

Maybe it makes sense to test the different scenarios in separate tests (can still be in the same file)?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -270,9 +271,102 @@ GlobalEncodingAnalyzer::traceToSourceGlobal(Value value) {
...
+  }
+  dispatchOp.setOperandEncodingsAttr(
+      ArrayAttr::get(dispatchOp.getContext(), newOperandEncodings));
+}
+
+// Applies all cached encoding updates to tensor ops.
+static void applyTensorEncodingUpdates(TensorEncodingUpdates &updates) {
+  for (auto &[op, operandUpdates] : updates) {
+    // TODO: Handle other TensorPhaseOp ops (TensorFillOp, etc.) via TypeSwitch.
+    if (auto dispatchOp = dyn_cast<TensorDispatchOp>(op)) {
```
sommerlukas, line 315: Should we use `cast<>` here to get an assert if a currently unsupported op appears here?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -270,9 +271,102 @@ GlobalEncodingAnalyzer::traceToSourceGlobal(Value value) {
...
+
+      // Only stream tensor ops need to be updated. Skip other operations.
+      if (!user->hasTrait<OpTrait::IREE::Stream::TensorPhaseOp>()) {
+        return WalkResult::advance();
+      }
+
+      // TODO: Handle other tensor phase ops (TensorFillOp, etc.)
+      auto dispatchOp = dyn_cast<IREE::Stream::TensorDispatchOp>(user);
+      if (!dispatchOp) {
+        return WalkResult::advance();
```
sommerlukas, line 357: Is it correct to skip other `TensorPhaseOp` here? Wouldn't it be more correct to bail here if encounter a `TensorPhaseOp` that isn't handled yet?

sommerlukas (commented):
Implementation looks good apart from small things. 

The test is very dense, testing a lot of scenarios in one test, maybe it makes sense to split that more.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Some ops should be ignored, e.g., `stream.tensor.constant`. I haven't reviewed all the ops, and I will add tests for each op in the future. I think dyn_cast is fine for now, as we remain the encodings unchanged, which seems better than an assertion/crash?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: This is an experimental pass anyway, so it is always incorrect if other tensor ops are involved. I'd like to make code structure simpler, so I bail out when it is not a tensor dispatch op.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: Let me break it into two tests. I was worried about large code snippets in initializers.

> Otherwise, the other functions could already trigger unification of the encodings, even if the cross-functional tracking fails.

I don't quite follow this. All the dispatch ops need to update the encodings, once we decide the unified encoding for a global.

Anyway, I think it's better to split test cases, as it is always easier to debug/triage if one of them fails. I moved the cross-function one out. I'm keeping the other three in the same group because they are all trivial cases.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
sommerlukas: Ok, if there are operations that we can ignore, `dyn_cast` is fine.

sommerlukas (approved):
Test is less dense now, thanks for the update!

PR #22891, opened 2025-12-11
[Stream] Handle tied operand result encodings in UnifyEncodingForGlobals 3/n

hanhanW:
When a dispatch has a tied operand whose encoding is being unified, the result encoding must also be updated to match (verifier requirement). Additionally, insert a re-encode op after the dispatch to convert the result back to the original encoding for downstream users.

This is a rare case that exists primarily for correctness as a safe fallback. In practice, tied operands with encodings that need unification are uncommon.

I decide to not bail out the case in the first place because it'd add much more complex to the implementation, while the fallback can be easy: just add re-encode ops. It may result in the amount of memory footprint, but it is a rare case.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -283,9 +283,9 @@ using TensorEncodingUpdates =
     llvm::DenseMap<Operation *, OperandEncodingUpdates>;
 
 // Updates encoding attributes for a TensorDispatchOp.
-static void
-updateTensorDispatchOp(TensorDispatchOp dispatchOp,
-                       const OperandEncodingUpdates &operandUpdates) {
+static void updateTensorDispatchOp(TensorDispatchOp dispatchOp,
+                                   const OperandEncodingUpdates &operandUpdates,
+                                   IRRewriter &rewriter) {
   // Update operand encodings.
   // The operand_encodings attribute has the same length as getMixedOperands().
   // For non-affinity types (e.g., index), the encoding is just the type.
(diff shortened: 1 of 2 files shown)


hanhanW:
The PR depends on https://github.com/iree-org/iree/pull/22886

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -193,6 +193,85 @@ module @immutable_source_initialized_from_parameter {
...
+  util.global.store %0, @weight_size : index
+  // CHECK: %[[SOURCE:.+]] = util.global.load @weight
+  %source = util.global.load @weight : !stream.resource<constant>
+  %source_size = util.global.load @weight_size : index
+
+  // CHECK: stream.tensor.sizeof tensor<4096x4096xf32, #iree_encoding.identity>
+  // CHECK: stream.tensor.encode %[[SOURCE]] : {{.*}} -> tensor<4096x4096xf32, #iree_encoding.identity>
+  %size1 = stream.tensor.sizeof tensor<4096x4096xf32, #encoding1> : index
+  %enc1 = stream.tensor.encode %source : tensor<4096x4096xf32> in !stream.resource<constant>{%source_size} -> tensor<4096x4096xf32, #encoding1> in !stream.resource<*>{%size1}
+  %const1 = stream.async.clone %enc1 : !stream.resource<*>{%size1} -> !stream.resource<constant>{%size1}
```
sommerlukas: Do we need the `clone`s here? Or should we leave testing the handling of `clone` to other tests and focus on the scenario at hand here?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -193,6 +193,85 @@ module @immutable_source_initialized_from_parameter {
...
+  util.global.store %const1, @encoded_v1 : !stream.resource<constant>
+  util.global.store %size1, @encoded_v1_size : index
+
+  // CHECK: stream.tensor.sizeof tensor<4096x4096xf32, #iree_encoding.identity>
+  // CHECK: stream.tensor.encode %[[SOURCE]] : {{.*}} -> tensor<4096x4096xf32, #iree_encoding.identity>
+  %size2 = stream.tensor.sizeof tensor<4096x4096xf32, #encoding2> : index
+  %enc2 = stream.tensor.encode %source : tensor<4096x4096xf32> in !stream.resource<constant>{%source_size} -> tensor<4096x4096xf32, #encoding2> in !stream.resource<*>{%size2}
+  %const2 = stream.async.clone %enc2 : !stream.resource<*>{%size2} -> !stream.resource<constant>{%size2}
```
sommerlukas: Same question about `clone` as above.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -193,6 +193,85 @@ module @immutable_source_initialized_from_parameter {
...
+// Test tied operand: dispatch result tied to input operand.
+// When encoding changes, both operand and result encoding must change.
+// A re-encode op should be inserted after dispatch.
+// CHECK-LABEL: util.func public @dispatch_with_tied_operand
+// CHECK-SAME:    %[[N:[a-zA-Z0-9_]+]]: index
+// CHECK-SAME:    %[[M:[a-zA-Z0-9_]+]]: index
+util.func public @dispatch_with_tied_operand(%N: index, %M: index) -> !stream.resource<*> {
+  %encoded = util.global.load @encoded_v1 : !stream.resource<constant>
+  %encoded_size = util.global.load @encoded_v1_size : index
+  %cloned = stream.async.clone %encoded : !stream.resource<constant>{%encoded_size} -> !stream.resource<*>{%encoded_size}
```
sommerlukas, line 319: Same question as above about `clone`.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -193,6 +193,85 @@ module @immutable_source_initialized_from_parameter {
...
+  %size1 = stream.tensor.sizeof tensor<4096x4096xf32, #encoding1> : index
+  %enc1 = stream.tensor.encode %source : tensor<4096x4096xf32> in !stream.resource<constant>{%source_size} -> tensor<4096x4096xf32, #encoding1> in !stream.resource<*>{%size1}
+  %const1 = stream.async.clone %enc1 : !stream.resource<*>{%size1} -> !stream.resource<constant>{%size1}
+  util.global.store %const1, @encoded_v1 : !stream.resource<constant>
+  util.global.store %size1, @encoded_v1_size : index
+
+  // CHECK: stream.tensor.sizeof tensor<4096x4096xf32, #iree_encoding.identity>
+  // CHECK: stream.tensor.encode %[[SOURCE]] : {{.*}} -> tensor<4096x4096xf32, #iree_encoding.identity>
+  %size2 = stream.tensor.sizeof tensor<4096x4096xf32, #encoding2> : index
+  %enc2 = stream.tensor.encode %source : tensor<4096x4096xf32> in !stream.resource<constant>{%source_size} -> tensor<4096x4096xf32, #encoding2> in !stream.resource<*>{%size2}
```
sommerlukas: This question is unrelated to this PR (and doesn't need to be addressed as part of it), but should we even unify globals in a case like this, where the second encoding isn't even used in the program?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -306,14 +306,92 @@ updateTensorDispatchOp(TensorDispatchOp dispatchOp,
...
+  SmallVector<ValueRange> resultEncodingDimsMap(newResultEncodings.size());
+  ValueRange remainingDims = dispatchOp.getResultEncodingDims();
+  for (auto [idx, encodingType] : llvm::enumerate(newResultEncodings)) {
+    auto shapedType = cast<ShapedType>(encodingType);
+    int64_t dynamicDimCount = shapedType.getNumDynamicDims();
+    resultEncodingDimsMap[idx] = remainingDims.take_front(dynamicDimCount);
+    remainingDims = remainingDims.drop_front(dynamicDimCount);
+  }
+
+  rewriter.setInsertionPointAfter(dispatchOp);
```
sommerlukas, line 373: Should we use an `InsertionGuard` here in case we use the `rewriter` elsewhere?

sommerlukas (commented):
Implementation looks good apart from nit, some questions about the test.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Good catch! I think it is callee's responsibility to restore the insertion point. Otherwise, callers would need to understand if it will be moved or not. I totally missed it, and thanks for the catch!

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: good catch, removing it.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: I think they should be cleaned up in the entry of stream pipeline

https://github.com/iree-org/iree/blob/818e59f074ca6f62c9a06dd016a330115e4d837b/compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp#L78-L80

To be more specific, unused globals are erased by the FoldGlobals pass.

https://github.com/iree-org/iree/blob/818e59f074ca6f62c9a06dd016a330115e4d837b/compiler/src/iree/compiler/Dialect/Util/Transforms/FoldGlobals.cpp#L423-L427

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: removed

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: This one is a little tricky to me. We should not tie it at all, if we mark it `constant`. We may not have such verification in ops, and using `resource<*>` looks more correct to me.

https://github.com/iree-org/iree/blob/818e59f074ca6f62c9a06dd016a330115e4d837b/compiler/src/iree/compiler/Dialect/Stream/IR/StreamDialect.td#L77-L78

sommerlukas (approved):
LGTM after update.

PR #22892, opened 2025-12-11
[Codegen][ROCDL] Improve dynamic dimension bounds handling in ROCDLConfigureBufferInstructions

Yu-Zhewen:
This change replaces the custom `getDynamicSizeMax` function with a shared utility function `getDynamicUpperBound()` which uses `IntegerRangeAnalysis` with a fallback to `ValueBoundsConstraintSet` for complex cases.

For example, with data-tiling, the dynamic dimension goes through a `ceildiv` due to packing.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLConfigureBufferInstructions.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLConfigureBufferInstructions.cpp
@@ -14,6 +14,7 @@
 #include "llvm/Support/DebugLog.h"
 #include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Interfaces/FunctionInterfaces.h"
+#include "mlir/Interfaces/ValueBoundsOpInterface.h"
 
 #define DEBUG_TYPE "iree-codegen-rocdl-configure-buffer-instructions"
 #define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
@@ -67,26 +68,8 @@ static bool isDefinitelyWorkgroupUniform(Value arg) {
   });
 }
 
-/// Return the maximum value that has been `util.assume.int`'d about this value
-/// if there is one.
-/// TODO: it'd be nice to be able to run the IntRangeAnalysis just up to the
-/// value in question, but we don't have that, so we approximate it.
-static std::optional<int64_t> getDynamicSizeMax(Value size) {
-  size = stripIntegerCasts(size);
-  // Special case for constants that're still dynamic.
-  APInt constVal;
-  if (matchPattern(size, m_ConstantInt(&constVal))) {
-    return constVal.getZExtValue();
-  }
-  auto assumeOp = size.getDefiningOp<IREE::Util::AssumeIntOp>();
-  if (!assumeOp)
-    return std::nullopt;
-  std::optional<int64_t> maybeMax =
-      assumeOp.getUnionedUnsignedRange(cast<OpResult>(size).getResultNumber())
-          .second;
-  return maybeMax;
-}
-
+/// Return the maximum number of bytes spanned by the binding, or nullopt if
+/// unable to determine.
 static std::optional<int64_t>
 getSpannedBytes(IREE::HAL::InterfaceBindingSubspanOp binding) {
   int64_t maxNumElems = 1;
@@ -98,8 +81,10 @@ getSpannedBytes(IREE::HAL::InterfaceBindingSubspanOp binding) {
   if (!resultTy || !resultTy.hasRank())
     return std::nullopt;
   for (Value dynArg : binding.getResultDynamicDims(0)) {
-    std::optional<int64_t> dimMax = getDynamicSizeMax(dynArg);
-    if (!dimMax)
+    FailureOr<int64_t> dimMax = ValueBoundsConstraintSet::computeConstantBound(
+        presburger::BoundType::UB, dynArg,
+        /*stopCondition=*/nullptr, /*closedUB=*/true);
+    if (failed(dimMax))
       return std::nullopt;
     maxNumElems *= (*dimMax);
   }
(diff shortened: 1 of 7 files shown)


qedawkins (commented):
For a pass like this, we should be able to do this by loading the IntegerRangeAnalysis instead of ValueBounds. ValueBounds is expensive and uncacheable, so in general we prefer to use the analysis where possible.

Yu-Zhewen:
> For a pass like this, we should be able to do this by loading the IntegerRangeAnalysis instead of ValueBounds. ValueBounds is expensive and uncacheable, so in general we prefer to use the analysis where possible.

I found a  utility function in `PadDynamicAlloc.cpp` that uses `IntegerRangeAnalysis` with a `ValueBoundsConstraintSet` fallback. Moved it to a common place so both passes can use it.

qedawkins:
Nice! That sounds like exactly what we want, thanks for investigating and making the switch.

Yu-Zhewen:
CI is currently failing while we wait for llvm/llvm-project#171839

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLConfigureBufferInstructions.cpp
```diff
@@ -121,10 +107,20 @@ struct ROCDLConfigureBufferInstructionsPass final
...
     mlir::FunctionOpInterface funcOp = getOperation();
-    // Is this really he best way to skip this pass on non-rocdl targets?
+    // Is this really the best way to skip this pass on non-rocdl targets?
     IREE::GPU::TargetAttr target = getGPUTargetAttr(funcOp);
     if (!target || !target.isAMD())
       return;
+
+    // Initialize the DataFlowSolver with IntegerRangeAnalysis.
+    DataFlowSolver solver;
+    solver.load<dataflow::DeadCodeAnalysis>();
```
krzysz00, line 119: I think the usual incantations also want SCCP in here.

... someone really needs to clean that up.

MaheshRavishankar (approved):

MaheshRavishankar:
@qedawkins  and @krzysz00  if the code looks good you guys need to approve it. There is no point letting a PR hang this way for weeks.

compiler/src/iree/compiler/Codegen/Utils/Utils.cpp
```diff
@@ -1109,6 +1112,31 @@ int64_t getMinElementBitwidth(linalg::LinalgOp linalgOp) {
...
+            IntegerValueRange::getMaxRange(value).getValue().smax()) {
+      return range.getValue().smax().getSExtValue();
+    }
+  }
+  // Fallback to ValueBoundsConstraintSet for complex cases.
+  auto ub = ValueBoundsConstraintSet::computeConstantBound(
+      presburger::BoundType::UB, {value, std::nullopt},
+      /*stopCondition=*/nullptr, /*closedUB=*/true);
+  if (succeeded(ub))
+    return ub.value();
```
kuhar: nit: missing braces https://iree.dev/developers/general/contributing/#compiler

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLConfigureBufferInstructions.cpp
```diff
@@ -67,28 +71,11 @@ static bool isDefinitelyWorkgroupUniform(Value arg) {
...
-  APInt constVal;
-  if (matchPattern(size, m_ConstantInt(&constVal))) {
-    return constVal.getZExtValue();
-  }
-  auto assumeOp = size.getDefiningOp<IREE::Util::AssumeIntOp>();
-  if (!assumeOp)
-    return std::nullopt;
-  std::optional<int64_t> maybeMax =
-      assumeOp.getUnionedUnsignedRange(cast<OpResult>(size).getResultNumber())
-          .second;
```
kuhar, line 86: Does this still work with assumptions or did we lose it? I assume it does because I can see that we have relevant tests.

kuhar (approved):

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLConfigureBufferInstructions.cpp
Yu-Zhewen: Yes, they will still work without explicit handling, as they implement both [ValueBoundsOpInterface](https://github.com/iree-org/iree/blob/af624754d77851368e0b405033ee2692018ffcc6/compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp#L120) and [InferIntRangeInterface](https://github.com/iree-org/iree/blob/af624754d77851368e0b405033ee2692018ffcc6/compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.td#L452).

yash-amd:
Hi @Yu-Zhewen , due to this pr we are seeing numeric issue for the amd-shark-ai llama models

Yu-Zhewen:
> Hi @Yu-Zhewen , due to this pr we are seeing numeric issue for the amd-shark-ai llama models

Could you point me to the specific tests where you’re seeing the numeric issues? I can revert this PR while we investigate.

yash-amd:
Hi @Yu-Zhewen, 
For now I am providing you steps to run run_llm_vmfb which we run in amd-shark-ai to check for numerics:

For llama-8b-fp16
* you can find the irpa file, tokenizer.json, tokenizer_config in the ossci directory in amdsharkpublic

1, clone amd-shark-ai
```
git@github.com:nod-ai/amd-shark-ai.git
cd amd-shark-ai
```
2.  you can create a python env and run these:
```
pip install -r pytorch-rocm-requirements.txt
    pip install -r requirements.txt -r requirements-iree-pinned.txt -e amdsharktank/ -e shortfin/
    pip install -f https://iree.dev/pip-release-links.html --upgrade --pre iree-base-compiler iree-base-runtime iree-turbine
    pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
          iree-base-compiler iree-base-runtime --src deps \
          -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
    pip uninstall -y iree-base-compiler iree-base-runtime
```    
 
 3. now you can set up iree with your checkout


5. open amdsharktank/tests/e2e/configs/models.json  > search for llama-8b-fp16 > change the irpa, tokenizer.json, tokenizer_config.json path where you have downloaded it locally.


4. just run: 
```
 python -m amdsharktank.tools.e2e_model_test --model llama-8b-fp16 --stage validate_vmfb
```

above command would export, compile, and test numerics as well as show the commands its using.
Artifacts Directory: output_artifacts/output_llama-8b-fp16

Yu-Zhewen:
Thanks @yash-amd, I'm looking into it.

PR #22893, opened 2025-12-11
Support float type conversion between f32 and f64 in VM lists.

willghatch:
I had an issue where passing an f64 where an f32 was expected would result in a zero value, and this fixes it.

### DIFF
--- a/runtime/src/iree/vm/list.c
+++ b/runtime/src/iree/vm/list.c
@@ -615,6 +615,22 @@ static void iree_vm_list_convert_value_type(
         default:
           return;
       }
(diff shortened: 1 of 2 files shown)


willghatch:
This fixes an issue I found using Wave.  Full disclosure:  I used AI tools to help me with this, and I don't really understand much of the context here.  But it does fix the issue and seems reasonable.

willghatch:
Ping @benvanik

benvanik (approved):

willghatch:
Thanks!

PR #22898, opened 2025-12-12
[Stream] Selecting unified encodings from encoding resolvers.

hanhanW:
Add a new `getUnifiedEncoding` method to the `LayoutResolverAttr` interface that returns a unified encoding given multiple candidate encodings. This is used by the UnifyEncodingForGlobals pass to select an appropriate encoding when the same source data has multiple encoded versions.

Also refactor GlobalEncodingAnalyzer to:
- Set up the layout resolver from dialect interfaces internally
- Compute unified encodings as part of the analysis phase
- Provide a `getUnifiedEncoding(name)` getter for querying results

This simplifies the pass by moving analysis-related logic into the analyzer class, making the pass focused on applying transformations.

Update lit tests to include device definitions and affinity attributes on stream.tensor.* ops, which are required for the pass to resolve layout attributes properly. It also switches identity_resolver to specialization_resolver, which improves the test quality. Identity encoding is used in fallback solution.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
@@ -69,6 +69,29 @@ def IREEEncoding_LayoutResolverAttr :
         assert(false && "unimplemented interface method");
         return {};
       }]
+    >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Returns a unified encoding attribute given a list of candidate encoding
+        attributes. This is used when multiple encodings are used for the same
+        source data and need to be unified to a single encoding.
+
+        The resolver implementation can create any encoding based on
+        backend-specific heuristics. The returned encoding does not need to be
+        one of the input encodings - for example, a resolver might return an
+        identity encoding for simplicity, or synthesize a new encoding that
+        introduces less overheads in relayout.
+      }],
+      /*retTy=*/"::mlir::Attribute",
+      /*methodName=*/"getUnifiedEncoding",
+      /*args=*/(ins
+        "llvm::ArrayRef<::mlir::Attribute>":$encodings
+      ),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        assert(false && "unimplemented interface method");
+        return {};
+      }]
     >
   ];
 }
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
@@ -91,6 +91,8 @@ def IREEEncoding_LayoutResolverAttr :
       ),
       /*methodBody=*/"",
       /*defaultImplementation=*/[{
+        // TODO(hanchung): Remove the assertion once we know what to do for the
+        // default implementation.
         assert(false && "unimplemented interface method");
         return {};
       }]
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
@@ -76,6 +76,8 @@ def IREEEncoding_LayoutResolverAttr :
         attributes. This is used when multiple encodings are used for the same
         source data and need to be unified to a single encoding.
 
+        Returns nullptr if it fails to get a unified encoding.
+
         The resolver implementation can create any encoding based on
         backend-specific heuristics. The returned encoding does not need to be
         one of the input encodings - for example, a resolver might return an
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -613,8 +613,8 @@ struct UnifyEncodingForGlobalsPass
       SourceGlobalInfo sourceInfo = analyzer.getSourceGlobals(sourceName);
       // Update each encode op to use the unified encoding.
       Attribute unifiedEncoding = analyzer.getUnifiedEncoding(sourceName);
-      LDBG() << "Unifying encodings for source global: " << sourceName
-             << " to " << unifiedEncoding;
+      LDBG() << "Unifying encodings for source global: " << sourceName << " to "
+             << unifiedEncoding;
       for (EncodedGlobalInfo &encodedInfo : sourceInfo.encodedVersions) {
         auto encodeOp = encodedInfo.encodeOp;
         auto oldResultType =
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -34,8 +34,8 @@ namespace {
 template <typename T>
 SmallVector<const T *> gatherUsedDialectInterfaces(mlir::ModuleOp moduleOp) {
   SmallPtrSet<const T *, 4> resultSet;
-  for (auto dialect : moduleOp.getContext()->getLoadedDialects()) {
-    auto *dialectInterface = dialect->getRegisteredInterface<T>();
+  for (Dialect *dialect : moduleOp.getContext()->getLoadedDialects()) {
+    const T *dialectInterface = dialect->getRegisteredInterface<T>();
     if (!dialectInterface)
       continue;
     resultSet.insert(dialectInterface);
@@ -119,19 +119,13 @@ public:
   // Returns the SourceGlobalInfo for the given source global name. There is a
   // copy in the call, so it is not a cheap call.
   SourceGlobalInfo getSourceGlobals(StringRef name) {
-    assert(sourceGlobals.contains(name) &&
-           "Expected all the source globals are parsed");
-    return sourceGlobals.find(name)->second;
+    return sourceGlobals.at(name);
   }
 
   // Returns the unified encoding for the given source global name, or
   // std::nullopt if not found.
-  std::optional<Attribute> getUnifiedEncoding(StringRef name) const {
-    auto it = unifiedEncodings.find(name);
-    if (it != unifiedEncodings.end()) {
-      return it->second;
-    }
-    return std::nullopt;
+  Attribute getUnifiedEncoding(StringRef name) const {
+    return unifiedEncodings.at(name);
   }
 
 private:
@@ -618,7 +612,7 @@ struct UnifyEncodingForGlobalsPass
     for (StringRef sourceName : candidates) {
       SourceGlobalInfo sourceInfo = analyzer.getSourceGlobals(sourceName);
       // Update each encode op to use the unified encoding.
-      Attribute unifiedEncoding = *analyzer.getUnifiedEncoding(sourceName);
+      Attribute unifiedEncoding = analyzer.getUnifiedEncoding(sourceName);
       LDBG() << "Unifying encodings for source global: " << sourceName
              << " to " << unifiedEncoding;
       for (EncodedGlobalInfo &encodedInfo : sourceInfo.encodedVersions) {
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -116,13 +116,12 @@ public:
     return result;
   }
 
-  // Returns the SourceGlobalInfo for the given source global name, or
-  // std::nullopt if not found.
-  std::optional<SourceGlobalInfo> getSourceGlobals(StringRef name) const {
-    if (sourceGlobals.contains(name)) {
-      return sourceGlobals.find(name)->second;
-    }
-    return std::nullopt;
+  // Returns the SourceGlobalInfo for the given source global name. There is a
+  // copy in the call, so it is not a cheap call.
+  SourceGlobalInfo getSourceGlobals(StringRef name) {
+    assert(sourceGlobals.contains(name) &&
+           "Expected all the source globals are parsed");
+    return sourceGlobals.find(name)->second;
   }
 
   // Returns the unified encoding for the given source global name, or
@@ -170,7 +169,7 @@ private:
   llvm::MapVector<StringRef, SourceGlobalInfo> sourceGlobals;
 
   // Maps source global name to its unified encoding. Populated by run().
-  llvm::DenseMap<StringRef, Attribute> unifiedEncodings;
+  llvm::StringMap<Attribute> unifiedEncodings;
 };
 
 LogicalResult GlobalEncodingAnalyzer::run() {
@@ -220,10 +219,10 @@ LogicalResult GlobalEncodingAnalyzer::computeUnifiedEncodings() {
   // Build queries for layout resolution.
   SmallVector<IREE::Stream::AffinityAndOpPair> queries;
   for (StringRef sourceName : candidates) {
-    std::optional<SourceGlobalInfo> sourceInfo = getSourceGlobals(sourceName);
-    for (EncodedGlobalInfo &encodedInfo : sourceInfo->encodedVersions) {
-      queries.push_back({encodedInfo.encodeOp.getAffinityAttr(),
-                         encodedInfo.encodedGlobal});
+    SourceGlobalInfo sourceInfo = getSourceGlobals(sourceName);
+    for (EncodedGlobalInfo &encodedInfo : sourceInfo.encodedVersions) {
+      queries.push_back(
+          {encodedInfo.encodeOp.getAffinityAttr(), encodedInfo.encodedGlobal});
     }
   }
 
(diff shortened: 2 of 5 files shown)


hanhanW:
The PR depends on

- https://github.com/iree-org/iree/pull/22886
- https://github.com/iree-org/iree/pull/22891

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
```diff
@@ -69,6 +69,29 @@ def IREEEncoding_LayoutResolverAttr :
...
+      }],
+      /*retTy=*/"::mlir::Attribute",
+      /*methodName=*/"getUnifiedEncoding",
+      /*args=*/(ins
+        "llvm::ArrayRef<::mlir::Attribute>":$encodings
+      ),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        assert(false && "unimplemented interface method");
+        return {};
```
sommerlukas, line 97: Can't we just leave out the default definition if it's not an implementation, so missing implementations are flagged via a compiler error?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -28,6 +29,29 @@ namespace mlir::iree_compiler::IREE::Stream {
...
+  SmallPtrSet<const T *, 4> resultSet;
+  for (auto dialect : moduleOp.getContext()->getLoadedDialects()) {
+    auto *dialectInterface = dialect->getRegisteredInterface<T>();
+    if (!dialectInterface)
+      continue;
+    resultSet.insert(dialectInterface);
+  }
+
+  // NOTE: to ensure deterministic output we sort the result so that imports are
+  // always added in a consistent order.
```
sommerlukas, line 45: What determines the order without sorting? The order with which dialects where imported?
The sorted list would also not be consistent, if an additional dialect is imported, right?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -117,14 +161,23 @@ class GlobalEncodingAnalyzer {
...
+  // Layout resolver function from dialect interface.
+  IREE::Stream::ResolveLayoutAttrFn resolveLayoutAttr;
+
   // Maps source global name to its info. Populated by run(). The global name
   // must match the name of `sourceGlobal` inside SourceGlobalInfo. StringRef is
   // used for easier lookup, which works better with SymbolTable, etc.
   llvm::MapVector<StringRef, SourceGlobalInfo> sourceGlobals;
+
+  // Maps source global name to its unified encoding. Populated by run().
+  llvm::DenseMap<StringRef, Attribute> unifiedEncodings;
```
sommerlukas: I don't think `DenseMap` is the right container here, `StringMap` might be better suited.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -137,6 +190,89 @@ LogicalResult GlobalEncodingAnalyzer::run() {
...
+  if (candidates.empty()) {
+    LDBG() << "No source globals with multiple encodings found.";
+    return success();
+  }
+
+  // Build queries for layout resolution.
+  SmallVector<IREE::Stream::AffinityAndOpPair> queries;
+  for (StringRef sourceName : candidates) {
+    std::optional<SourceGlobalInfo> sourceInfo = getSourceGlobals(sourceName);
+    for (EncodedGlobalInfo &encodedInfo : sourceInfo->encodedVersions) {
```
sommerlukas: Shouldn't we check the `optional` before iterating over it?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -137,6 +190,89 @@ LogicalResult GlobalEncodingAnalyzer::run() {
...
+    LDBG() << "Failed to resolve layouts for a query";
+    return failure();
+  }
+
+  // Compute unified encoding for each source global.
+  MLIRContext *ctx = moduleOp.getContext();
+  for (StringRef sourceName : candidates) {
+    SetVector<Attribute> layoutResolvers;
+    SmallVector<Attribute> encodingAttrVersions;
+    std::optional<SourceGlobalInfo> sourceInfo = getSourceGlobals(sourceName);
```
sommerlukas: Same as above, do we need to check the `optional` before using it?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -38,14 +42,214 @@ module @immutable_source_with_initial_value {
...
+  // CHECK: %[[SOURCE:.+]] = util.global.load @weight
+  %source = util.global.load @weight : !stream.resource<constant>
+  %source_size = util.global.load @weight_size : index
+
+  // Device A encodes the shared source - should get identity encoding.
+  // CHECK: stream.tensor.sizeof on(#hal.device.affinity<@[[$DEVICE_A]]>) tensor<4096x4096xf32, #iree_encoding.identity>
+  // CHECK: stream.tensor.encode on(#hal.device.affinity<@[[$DEVICE_A]]>) %[[SOURCE]] : {{.*}} -> tensor<4096x4096xf32, #iree_encoding.identity>
+  %size1 = stream.tensor.sizeof on(#hal.device.affinity<@device_a>) tensor<4096x4096xf32, #encoding1> : index
+  %enc1 = stream.tensor.encode on(#hal.device.affinity<@device_a>) %source : tensor<4096x4096xf32> in !stream.resource<constant>{%source_size} -> tensor<4096x4096xf32, #encoding1> in !stream.resource<*>{%size1}
+  %const1 = stream.async.clone on(#hal.device.affinity<@device_a>) %enc1 : !stream.resource<*>{%size1} -> !stream.resource<constant>{%size1}
```
sommerlukas: Do we need all the `clone`s here? It increases the test code size and correct handling of `clone` is/should be tested by other tests.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
hanhanW: Good question. I honestly did not know what the behavior would be if we drop the default implementation. I just gave it a shot, and it will ask all the attribute to provide the implementation. E.g.,

```
In file included from ../compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp:33:
In file included from ../compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.h:12:
In file included from ../compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingOps.h:10:
In file included from ../compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h:89:
compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.h.inc:644:100: error: too few arguments to function call, expected 3, have 2
  644 |   return static_cast<const ConcreteAttr *>(impl)->getUnifiedEncoding(tablegen_opaque_val, encodings);
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                               ^
compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.h.inc:103:69: note: in instantiation of member function 'mlir::iree_compiler::IREE::Encoding::detail::LayoutResolverAttrInter
faceTraits::FallbackModel<mlir::iree_compiler::IREE::GPU::(anonymous namespace)::GPULayoutResolverAttr>::getUnifiedEncoding' requested here
  103 |     FallbackModel() : Concept{cloneWithSimplifiedConfig, getLayout, getUnifiedEncoding} {}
```

So it won't require us to add the implementation for all the attributes that implement the interface; it won't break downstream projects when they bump IREE.

In release build, it can be handled by the pass that has decent fallback solution (e.g., using identity encoding). Let me update the doc a bit.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: This was copied from other snippet. My understanding is that `getLoadedDialects` does not give you deterministic order, and we sort them by name. I.e., if you don't sort, you can get the same set of dialects that implement the interface but the order can be different.

We should already import all the dialects into the context. The importing order can be different. Does it make sense?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Nice, I know there is something like `StringMap`, but I did not remember. Thanks for reminding me!

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Let me add an assertion inside the implementation and drop the `optional`. The input looks wrong to me if it happens.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: Thanks for pointing it out! I removed the `clone`s.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
sommerlukas: Thanks for trying this out. The error is a bit different from what I had expected, I suppose I expected it to point more clearly to a missing implementation. 

To me, there's two ways this can be handled:
1. If in the future, we expect all encodings with `LayoutResolverAttr` interface to implement `getUnifiedEncoding`, we should eventually not have a default implementation. If you want to have a dummy default implementation for now while this series of PRs is still in progress, we can keep the current default implementation, but I'd suggest to add a `TODO` to remove it in the future.
2. If it's fine for encodings with `LayoutResolverAttr` interface to not implement `getUnifiedEncoding`, we should have a default implementation that doesn't `assert` (maybe return identity encoding instead).

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
sommerlukas: Ok, I see, so we mainly want to protect against different import orders here.

Maybe a naive question, but does the order really matter here? The only use of this function I can see in this file expects only a single dialect interface to be found and returns `failure()` otherwise, so can we just skip the sorting here?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -88,16 +116,28 @@ class GlobalEncodingAnalyzer {
     return result;
   }
 
-  // Returns the SourceGlobalInfo for the given source global name, or
+  // Returns the SourceGlobalInfo for the given source global name. There is a
+  // copy in the call, so it is not a cheap call.
+  SourceGlobalInfo getSourceGlobals(StringRef name) {
+    assert(sourceGlobals.contains(name) &&
+           "Expected all the source globals are parsed");
+    return sourceGlobals.find(name)->second;
```
sommerlukas:
```suggestion
    return sourceGlobals.at(name);
```
`::at()` aborts (i.e., `assert`s) if the entry doesn't exist.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -28,6 +29,29 @@ namespace mlir::iree_compiler::IREE::Stream {
 
 namespace {
 
+/// Returns a stably sorted list of dialect interfaces of T for all dialects
+/// used within the given module.
+template <typename T>
+SmallVector<const T *> gatherUsedDialectInterfaces(mlir::ModuleOp moduleOp) {
+  SmallPtrSet<const T *, 4> resultSet;
+  for (auto dialect : moduleOp.getContext()->getLoadedDialects()) {
```
jtuyls:
```suggestion
  for (Dialect *dialect : moduleOp.getContext()->getLoadedDialects()) {
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -28,6 +29,29 @@ namespace mlir::iree_compiler::IREE::Stream {
...
+    auto *dialectInterface = dialect->getRegisteredInterface<T>();
```
jtuyls:
```suggestion
    const T *dialectInterface = dialect->getRegisteredInterface<T>();
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -137,6 +189,93 @@ LogicalResult GlobalEncodingAnalyzer::run() {
...
+      const SetVector<Attribute> &resolvers =
+          cachedLayoutAttrs[IREE::Stream::AffinityAndOpPair(
+              encodedInfo.encodeOp.getAffinityAttr(),
+              encodedInfo.encodedGlobal)];
+      layoutResolvers.insert(resolvers.begin(), resolvers.end());
+      encodingAttrVersions.push_back(encodedInfo.encodingAttr);
+    }
+
+    // TODO: It is not clear which encoding to pick when there are multiple
+    // layout resolvers. For now, just fallback to identity encoding for safety.
```
jtuyls, line 247: This will regress current performance in for example llama as we have two different encodings for prefill and decode? Or is this somehow not called yet?

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
hanhanW: We should target (2). The interface method is pretty new, and other resolvers should catch up. The assertion is more like a sanity check atm. We used to have it in `isSerialized` method and we dropped it until it became mature: https://github.com/iree-org/iree/commit/8813333a37d474850011a595a3f043672ba6abc4

For now, I think we don't really know what the default behavior should be, so I'm being conservative. The default can be either using identity encoding or leave them as what they are (i.e., do not unify the encodings). We don't need to figure out all the things in one-shot as this pass is currently developed and it is totally new to IREE, IMO.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: I think we should not consider how it is used here. It is a local function, and it can be moved to common utils. Having the determinism is a plus to this function itself, IMO.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: The pass is not added to the pipeline yet. It will regress the performance if it is enabled. The plan in my mind is guarding the pass under an experimental flag. If we want to unify the encodings, we compile the model with the flag.

The next step is implementing the interface in GPU encoding resolver. I think we can decide the layout and return a serialized encoding. It definitely needs more discussion and exploration, and we can chat it in data-tiling sync.

hanhanW (commented):
I messed up the tests when I rebased the branch. I rework the tests in the last three commits. Hopefully it helps the review, and you can review them individually.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
jtuyls: Makes sense, thanks!

jtuyls (approved):
LGTM

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
sommerlukas: Ok, makes sense. Maybe we can put a TODO as a reminder.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
sommerlukas: IMHO, the sorting currently only adds unnecessary code complexity and compile-time overhead in a case that we want to fail anyway. We could remove the sorting now and bring it back when we actually turn this into a common utility. But if you think the sorting/determinism is going to be useful soon, we can also keep it.

sommerlukas (approved):
Test and implementation look good after updates, I'm going to leave the decision regarding sorting with you.

PR #22894, opened 2025-12-11
Bump torch-mlir

rkayaith:
Bump torch-mlir to https://github.com/iree-org/torch-mlir/commit/77d0c3e2fc5485c2585edd54aff46e85962966a7, which contains fixes for recent LLVM integrates.

PR #22902, opened 2025-12-12
Bump to llvm/torch-mlir@3cebce2

zjgarvey:
The recent bump in https://github.com/iree-org/iree/pull/22894 missed out on some commits we need.


sjain-stanford (approved):

IanWood1 (approved):

zjgarvey:
I'm going to squash and merge this with the failing tests because the failures should not be relevant. The backward conv torch-to-linalg pattern is not even active yet (we need to add an op to our backend legal ops set for `torch-decompose-complex-ops`). The other changes are strictly for scalar aten ops which are very uncommon, and I am confident they should not cause regressions in the tests which failed due to other reasons.

Issue #22405, opened 2025-10-24
Support Grouped ConvWGrad


a-sidorova:
Fusilli needs to support for grouped `ConvWGrad` (`torch.aten.convolution_backward`): https://github.com/nod-ai/shark-ai/issues/2587.

The input data:
```
n = 4, c = 16, h = 8, w = 8, k = 32, fc = 4 (group count = 4), r = 1, s = 1, stride = [1, 1] / [2, 2]
dy_shape = [n, k, h / stride, w / stride]
x_shape = [n, c, h, w]
dw_shape = [k, fc, r, s]
```
Command line:
```
iree-compile model.mlir -o model.vmfb --iree-hal-target-device=hip  --iree-hip-target=gfx942
```

- If `strides = [2, 2]` ([gconv_wgrad_g4_s2.mlir](https://gist.github.com/a-sidorova/860d31da9bf3c56b03351273f050ebdb#file-gconv_wgrad_g4_s2-mlir)), the MLIR is successfully lowered and compiled.

- But If `strides = [1, 1]` ([gconv_wgrad_g4_s1.mlir](https://gist.github.com/a-sidorova/860d31da9bf3c56b03351273f050ebdb#file-gconv_wgrad_g4_s1-mlir)), the pass `ConvertTorchToLinalg` is failed with the following logs:
```
model_nchw.mlir:49:63: error: 'linalg.conv_2d_ngchw_gfchw' op inferred input/output operand #1 has shape's dimension #2 to be 1, but found 4
    %grad_input_conv_wgrad, %dw_perm, %grad_bias_conv_wgrad = torch.aten.convolution_backward %dy_perm, %x_perm, %empty_conv_wgrad, %bias_conv_wgrad, %stride_conv_wgrad, %padding_conv_wgrad, %dilation_conv_wgrad, %transposed_conv_wgrad, %output_padding_conv_wgrad, %groups_conv_wgrad, %output_mask_conv_wgrad : !torch.vtensor<[4,32,8,8],f32>, !torch.vtensor<[4,16,8,8],f32>, !torch.vtensor<[32,4,1,1],f32>, !torch.none, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int, !torch.list<bool> -> !torch.none, !torch.vtensor<[32,4,1,1],f32>, !torch.none
                                                              ^
```

a-sidorova:
[...]

PR #22904, opened 2025-12-15
[Input][Torch] Direct lowering of torch convolution_backward to linalg

a-sidorova:
See https://github.com/llvm/torch-mlir/issues/4355 for more details

The PR disables decomposition  `torch.aten.convolution_backward` to the `torch` dialect - this op now will be directly lowered to `linalg` dialect. The direct lowering should improve performance of backward convolutions since we will assign reduction/parallel dims directly, instead of insertion transposes/expand/collapse ops around.

#### Issues:
- https://github.com/iree-org/iree/issues/22502
- https://github.com/iree-org/iree/issues/22405


zjgarvey:
Before merging this, I noticed an issue with the lowering in low-precision examples. Namely, if the `aten.convolution_backward` op takes `bfp16` tensors to `bfp16` output, the current lowering doesn't use the correct accumulator dtype of float32. 

I'll work on a patch for this NA time today in torch-mlir and I'll ping you for a review @a-sidorova

sjain-stanford:
@zjgarvey is this ready to land? Do we need another torch-mlir bump in IREE to go in this PR?

zjgarvey:
@sjain-stanford I responded in discord, but I'll reply here, too. 

The bump is up at #22934 

Once that lands, we should be good to merge this change.

sjain-stanford (approved):

zjgarvey (approved):
We are testing backward conv e2e in other IREE repos regularly (Turbine, Fusilli), so I'm fine not including any tests here. From my local investigation, this fixes quite a few compilation failures, but there are some places where performance is slightly worse.

I personally think we should land and iterate on performance.

MaheshRavishankar:
Looked at the performance profiles here. The final state of comparison is making sure that the torch + lowering to Linalg gets to the same state as the directy Linalg lowering we have been looking at for the Python API. So I have no issues landing this and iterating on it since we arent at the final state yet.

PR #22905, opened 2025-12-15
[Encoding] Add EncodingProperties retrieval to SerializableAttr

jtuyls:
Implements phase 1 for supporting dynamic encoding layout decisions: https://github.com/iree-org/iree/issues/22370

- **Add `EncodingProperties` struct** — Bundles encoding attribute with dynamic values (M, N, K). Returned when querying encodings from operations.
- **Extend `SerializableAttr` interface** — Add `getOperandEncodingProperties(Operation*)` that derives encodings + dynamic values for all operands. The operation is the source of truth.
- **Refactor `SetEncoding` pass** — Use the interface instead of manually constructing encodings. Initially passes zero dynamic values.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
@@ -63,6 +63,29 @@ MatmulNarrowDim getMatmulNarrowDim(linalg::LinalgOp linalgOp,
                                    int narrowThreshold);
 
 // The structs defined here because they are used by encoding_interfaces.td.
+
+/// Bundles an encoding attribute with its associated dynamic information.
+/// This is returned by interface methods that query encoding properties
+/// from operations. Dynamic values can include runtime information needed
+/// by the encoding (e.g., M, N, K dimensions for matmul encodings).
+struct EncodingProperties {
+  /// The encoding attribute for the operand/result.
+  Attribute encoding;
+  /// Dynamic values needed by the encoding. For matmul-like operations,
+  /// these typically correspond to the iteration domain dimensions (M, N, K).
+  /// TODO(#22370): Currently empty; will be populated to support dynamic
+  /// layout decisions.
+  SmallVector<Value> dynamicValues;
+};
+
+/// Bundles encoding properties for all operands and DPS init operands of an
+/// operation. This is returned by SerializableAttr::getEncodingProperties()
+/// when deriving encodings from an operation (e.g., linalg.matmul).
+struct OpEncodingProperties {
+  SmallVector<EncodingProperties> operands;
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
```diff
@@ -63,6 +63,29 @@ MatmulNarrowDim getMatmulNarrowDim(linalg::LinalgOp linalgOp,
...
+/// This is returned by interface methods that query encoding properties
+/// from operations. Dynamic values can include runtime information needed
+/// by the encoding (e.g., M, N, K dimensions for matmul encodings).
+struct EncodingProperties {
+  /// The encoding attribute for the operand/result.
+  Attribute encoding;
+  /// Dynamic values needed by the encoding. For matmul-like operations,
+  /// these typically correspond to the iteration domain dimensions (M, N, K).
+  /// TODO(jtuyls): Currently empty; will be populated to support dynamic
+  /// layout decisions. See https://github.com/iree-org/iree/issues/22370.
```
hanhanW: Linking with github issue directly looks better to me. IMO, we use handle when the corresponding issue is not present. It means that asking this person who has the context to me. See https://google.github.io/styleguide/cppguide.html#TODO_Comments


```suggestion
  /// TODO(#22370): Currently empty; will be populated to support dynamic
  /// layout decisions.
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
```diff
@@ -63,6 +63,29 @@ MatmulNarrowDim getMatmulNarrowDim(linalg::LinalgOp linalgOp,
...
+  SmallVector<Value> dynamicValues;
+};
+
+/// Bundles encoding properties for all operands and results of an operation.
+/// This is returned by SerializableAttr::getEncodingProperties() when
+/// deriving encodings from an operation (e.g., linalg.matmul).
+struct OpEncodingProperties {
+  SmallVector<EncodingProperties> operands;
+  SmallVector<EncodingProperties> results;
+};
```
hanhanW: I'm thinking if we use `inputs` and `inits`, so it connects to DPS ops better. `results` looks a little wonky to me, because you don't set encodings on results. (We unset encodings on results, though.) What do you think?

hanhanW (commented):
Thanks, it looks good to me! Max is out, so maybe we want an approval from Mahesh.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
jtuyls: Yeah, makes sense, updated it.

hanhanW (approved):

PR #22906, opened 2025-12-15
[LLVMGPU] Remove fold unit extent dims from vector distribute

Groverkss:
This unit dim folding was added early on when the pipeline was created. It helps reduce some cognitive load in tests, but adds a lot of implementation overhead for each new codegen op to implement unit dim extent folding. Vectorization completly fails if this unit dim folding doesn't work.

Remove unit dim folding from the codegen pipeline and handle it in vector distribute properly. We can drop a lot of code for droppping unit dims for codegen ops after this patch.

Eventually, this work will lead to dropping `outer_tile` from the layout, which will eventually reduce the number of final vector dimensions.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -825,29 +825,8 @@ void addGPUVectorDistributePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(createLLVMGPUConfigureTensorLayoutsPass());
   funcPassManager.addPass(createIREELoopInvariantCodeMotionPass());
 
-  // Generalize all named ops so that we can fold away unit extent dims. By this
-  // point, all tiling is finished so the tiling configurations on those ops can
-  // be safely dropped. This additionally allows vectorization of convolution to
-  // `vector.contract` as filter dimensions are expected to be tiled to 1 by
-  // this point.
-  funcPassManager.addPass(createLinalgGeneralizeNamedOpsPass());
-  {
-    IREE::LinalgExt::FoldUnitExtentDimsPassOptions options;
-    options.useReshapes = false;
-    funcPassManager.addPass(
-        IREE::LinalgExt::createFoldUnitExtentDimsPass(options));
-  }
-  funcPassManager.addPass(
-      IREE::VectorExt::createVectorExtFoldUnitExtentDimsPass());
-  {
-    LinalgFoldUnitExtentDimsPassOptions options;
-    options.useRankReducingSlices = true;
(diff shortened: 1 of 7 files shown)


kuhar (approved):

qedawkins (approved):
nice

krzysz00 (commented):
No objections here, though I think this might be another reason to land the drop unit dims pass late in codegen

nirvedhmeshram:
Any reason this didnt land? should it be closed if its not landing?

Groverkss:
> Any reason this didnt land? should it be closed if its not landing?

I want to get back to it, but there are some performance regressions. I'm in the middle of trying to fix some regressions before I land this.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -789,35 +787,17 @@ void addGPUVectorDistributePassPipeline(OpPassManager &funcPassManager,
...
-    funcPassManager.addPass(
-        IREE::LinalgExt::createFoldUnitExtentDimsPass(options));
-  }
-  funcPassManager.addPass(
-      IREE::VectorExt::createVectorExtFoldUnitExtentDimsPass());
-  {
-    LinalgFoldUnitExtentDimsPassOptions options;
-    options.useRankReducingSlices = true;
-    funcPassManager.addPass(mlir::createLinalgFoldUnitExtentDimsPass(options));
-    funcPassManager.addPass(createCanonicalizerPass());
```
keshavvinayak01, line 846: Adding back `FoldTensorSubsetOpsPass` should eliminate the CI regressions. It must be stray `tensor.insert_slice ops` that's causing unnecessary buffer allocation. I tested with this addition on the failing tests and it compiles.

Groverkss: Thanks! I tried adding that, let's see how it goes.

keshavvinayak01 (commented):
Once https://github.com/iree-org/iree/pull/23552 is in, I'll re-run CI on the currently failing test. Investigating the test and dumping out the IRs at several stages, I suspect the memory leak is occurring due to the multi-batch `iree_vector_ext.gather` being used as memory buffer instead of vector reads.

keshavvinayak01 (approved):

Groverkss:
I'm going to land this for now, but I am scared if the torch_models suite will pass on mi325. We can revert if needed if CI doesn't pass.

keshavvinayak01:
> I'm going to land this for now, but I am scared if the torch_models suite will pass on mi325. We can revert if needed if CI doesn't pass.

Yeah I tried multiple times to re-run the CI, it's flaky and refuses to run.

PR #22907, opened 2025-12-15
[Encoding] Add dynamic encoding dims to (Un)SetEncodingOp

jtuyls:
Implements part 1 of phase 2 for supporting dynamic encoding layout decisions: https://github.com/iree-org/iree/issues/22370.

Adds the dynamic `encoding_dims` parameter to `SetEncodingOp` and `UnsetEncdingOp`. They are not set yet.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
@@ -299,11 +299,12 @@ static Value generateEncodingTransferOps(RewriterBase &rewriter, Value src,
   Value value = src;
   if (srcType.getEncoding()) {
     value = IREE::Encoding::UnsetEncodingOp::create(
-        rewriter, src.getLoc(), srcType.dropEncoding(), value, dynamicDims);
+        rewriter, src.getLoc(), srcType.dropEncoding(), value, dynamicDims,
+        /*encodingDims=*/ValueRange{});
   }
   if (destType.getEncoding()) {
-    value = IREE::Encoding::SetEncodingOp::create(rewriter, src.getLoc(),
-                                                  destType, value);
+    value = IREE::Encoding::SetEncodingOp::create(
+        rewriter, src.getLoc(), destType, value, /*encodingDims=*/ValueRange{});
   }
   return value;
 }
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingOps.td
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingOps.td
@@ -23,13 +23,19 @@ def IREEEncoding_SetEncodingOp : IREEEncoding_PureOp<"set_encoding",[
     Operation to assign an encoding to a tensor. The operation does not change
     the rank or extent of a tensor. Instead it adds a LayoutResolverAttr
     attribute to the tensor type to represent a change in layout.
+
+    The optional `encoding_dims` operand carries dynamic values needed by the
+    encoding (e.g., M, N, K dimensions for matmul encodings). These values are
+    used for runtime layout selection based on problem size.
   }];
 
-  let arguments = (ins AnyRankedTensor:$source);
+  let arguments = (ins
+    AnyRankedTensor:$source,
+    Variadic<Index>:$encoding_dims);
   let results = (outs AnyRankedTensor:$result);
 
   let assemblyFormat = [{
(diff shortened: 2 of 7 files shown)


MaheshRavishankar (approved):
As it stands the PR is OK, but there is an implicit link between the dimensions expected by the encoding and the encoding_dims added here. I think we need an interface/method on that interface that allows the verifier of set_encoding to make sure that all the dynamic encoding_dims needed are specified.

Its OK to do that in a follow up, but I dont know if this is accounted in the original design.

jtuyls:
> As it stands the PR is OK, but there is an implicit link between the dimensions expected by the encoding and the encoding_dims added here. I think we need an interface/method on that interface that allows the verifier of set_encoding to make sure that all the dynamic encoding_dims needed are specified.
> 
> Its OK to do that in a follow up, but I dont know if this is accounted in the original design.

Here, I just introduce the changes to the ops before adding the logic to actually use them, because then I would also need to add the propagation logic changes and update a lot of lit tests (once the verifier is added). Therefore, I am staging this in (probably) 3 PRs:

1. Update the ops (this PR)
2. Add the logic for filling the encoding_dims + propagation updates (atomic to avoid breaking propagation)
3. Add verifier and update all lit tests

hanhanW (approved):
I think we will be able to drop `iteration_sizes` after we plumb it through, right?

Max191:
> I think we will be able to drop `iteration_sizes` after we plumb it through, right?

I don't think so, because the iteration_sizes still provide the semantic information required to interpret the encoding dims carried by the set/unset_encoding. The encoding dims carried by the encoding ops would basically just represent any question marks in the iteration_sizes.

Max191 (approved):
This LGTM. I was going to say the same thing that Mahesh mentioned about adding an encoding interface for verifying the correct number of carried encoding dimensions, but it is better as a followup once the encoding ops actually start carrying dynamic dims. I think it would add too much complexity to this PR, which is supposed to be simple.

hanhanW:
> > I think we will be able to drop `iteration_sizes` after we plumb it through, right?
> 
> I don't think so, because the iteration_sizes still provide the semantic information required to interpret the encoding dims carried by the set/unset_encoding. The encoding dims carried by the encoding ops would basically just represent any question marks in the iteration_sizes.

The example in https://github.com/iree-org/iree/issues/22370#issuecomment-3724210393 shows that it carries all the dim sizes. Anyway, I just found that it could be not needed for set/unset encoding ops, but we still need the information in gemm/generic op lowering.

jtuyls:
> > > I think we will be able to drop `iteration_sizes` after we plumb it through, right?
> > 
> > 
> > I don't think so, because the iteration_sizes still provide the semantic information required to interpret the encoding dims carried by the set/unset_encoding. The encoding dims carried by the encoding ops would basically just represent any question marks in the iteration_sizes.
> 
> The example in [#22370 (comment)](https://github.com/iree-org/iree/issues/22370#issuecomment-3724210393) shows that it carries all the dim sizes. Anyway, I just found that it could be not needed for set/unset encoding ops, but we still need the information in gemm/generic op lowering.

Sorry for the confusion. Initially, I indeed added all dimensions into the `encoding_dims`, hence the example in the issue, but after cleaning up, it now just contains the dynamic dimensions, because I didn't see an immediate use for all dimensions unless we drop `iteration_sizes` indeed, which I actually like because all encoding information would be in a single place. But we could still do that in a follow-up if we want or I can also implement it right now, WDYT?

hanhanW:
> > > > I think we will be able to drop `iteration_sizes` after we plumb it through, right?
> > > 
> > > 
> > > I don't think so, because the iteration_sizes still provide the semantic information required to interpret the encoding dims carried by the set/unset_encoding. The encoding dims carried by the encoding ops would basically just represent any question marks in the iteration_sizes.
> > 
> > 
> > The example in [#22370 (comment)](https://github.com/iree-org/iree/issues/22370#issuecomment-3724210393) shows that it carries all the dim sizes. Anyway, I just found that it could be not needed for set/unset encoding ops, but we still need the information in gemm/generic op lowering.
> 
> Sorry for the confusion. Initially, I indeed added all dimensions into the `encoding_dims`, hence the example in the issue, but after cleaning up, it now just contains the dynamic dimensions, because I didn't see an immediate use for all dimensions unless we drop `iteration_sizes` indeed, which I actually like because all encoding information would be in a single place. But we could still do that in a follow-up if we want or I can also implement it right now, WDYT?

The current version looks good to me. Let's land and iterate?

PR #22910, opened 2025-12-15
[Codegen] Add PCF Transforms boilerplate

qedawkins:
Adds the directory structure and boilerplate files for PCF dialect transformations:

- Transforms.h/cpp: Header and implementation for PCF transformations
- ConversionDialectInterface.h: Interface for dialect-specific lowering
- Passes.h/cpp/td: Pass infrastructure for PCF passes
- BUILD.bazel and CMakeLists.txt: Build configuration
- test/: Test directory structure

This commit provides the foundation for subsequent PCF transformation passes.

This is a placeholder PR that will be landed with the first pass from this directory that lands.

PR #22807, opened 2025-12-03
[Codegen] Replace workgroup loops with pcf.generic

qedawkins:
This commit adds the remaining PCF infrastructure and plumbing to replace all workgroup loops with pcf.generic ops:

1. Common passes:
   - ConvertWorkgroupForallToPCF: Converts workgroup-mapped scf.forall to pcf.loop
   - CombineSplitKWorkgroupLoop: Combines nested pcf.loops for split-K operations
   - FuseTensorToBufferConverters: Fuses tensor-to-buffer conversion patterns

2. Pipeline integration:
   - LLVMCPU: Adds PCF passes to pipelines with --iree-llvmcpu-test-pcf flag
   - LLVMGPU: Adds PCF lowering passes
   - SPIRV: Adds PCF lowering passes
   - VMVX: Adds PCF lowering passes

3. CodegenExternalModels: Adds workgroup scope PCF support

4. Codegen dialect extensions:
   - WorkgroupAttr implements PCF ScopeAttr interface
   - WorkgroupMappingAttr with linearize support

5. Bug fixes:
   - Fix null ArrayAttr crash in ConvertWorkgroupForallToPCF
   - Fix MLIRBufferizationInterfaces -> MLIRBufferizationDialect in CMake
   - Add --iree-llvmcpu-test-pcf flag to pipeline tests

PR #22911, opened 2025-12-15
[Codegen] Add ConvertForallToLoopsPass

qedawkins:
Adds the ConvertForallToLoopsPass which converts scf.forall ops to pcf.loop ops with sequential scope.

This is the entry point for converting parallel iteration constructs into PCF's parallel control flow representation. The pass:

- Converts scf.forall ops without mapping attributes to pcf.loop
- Linearizes/delinearizes iteration IDs based on DeviceMappingAttr
- Replaces tensor.parallel_insert_slice with pcf.write_slice
- Creates tied pcf.sref region arguments for shared output tensors

The underlying conversion pattern is exposed via convertForallToPCF() for use with custom execution scopes.


compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
```diff
@@ -0,0 +1,322 @@
...
+  }
+
+  SmallVector<int64_t> perm = getMappingPermutation(mappingAttr.value());
+  if (!isPermutationVector(perm)) {
+    return failure();
+  }
+  return perm;
+}
+
+static LogicalResult permuteForallIds(RewriterBase &rewriter,
```
MaheshRavishankar: I dont see why you need to change the original `scf.forall` seems unnecessary, and potentially if we for whatever reason dont convert the `scf.forall` to `pcf.*` then it just causes a change that is unexpected.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
```diff
@@ -0,0 +1,322 @@
...
+      // Add replacements in reverse to get fastest -> slowest.
+      auto delinResultsReverse = llvm::reverse(delin.getResults());
+      argReplacements.append(delinResultsReverse.begin(),
+                             delinResultsReverse.end());
+    } else {
+      argReplacements.append(loopOp.getIdArgs().begin(),
+                             loopOp.getIdArgs().end());
+    }
+
+    AffineExpr applyLbAndStep = (s0 + s1) * s2;
```
MaheshRavishankar: This might be wrong. The `lb` is  multiplied by the `step`, which I dont think is needed.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
```diff
@@ -0,0 +1,322 @@
...
+  SmallVector<OpFoldResult> mixedUbs = forallOp.getMixedUpperBound();
+  SmallVector<OpFoldResult> mixedLbs = forallOp.getMixedLowerBound();
+  SmallVector<OpFoldResult> mixedStep = forallOp.getMixedStep();
+
+  AffineExpr s0, s1, s2;
+  bindSymbols(rewriter.getContext(), s0, s1, s2);
+  AffineExpr numIters = (s0 - s1).ceilDiv(s2);
+  SmallVector<Value> iterationCounts;
+  for (auto [ub, lb, step] : llvm::zip_equal(mixedUbs, mixedLbs, mixedStep)) {
+    OpFoldResult iterCount = affine::makeComposedFoldedAffineApply(
```
MaheshRavishankar, line 192: Why do we need to do this here? Seems premature to me.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.td
```diff
@@ -9,4 +9,27 @@
...
+
+    The input is IR containing `scf.forall` ops with tensor results and
+    `tensor.parallel_insert_slice` terminators. Only forall ops without
+    mapping attributes are converted.
+
+    The output replaces each matching `scf.forall` with a `pcf.loop`:
+    - Iteration bounds come from the forall's upper/lower bounds and steps
+    - `tensor.parallel_insert_slice` ops become `pcf.write_slice` ops
+    - Shared output tensors become tied `pcf.sref` region arguments
+    - The scope is set to `#pcf.sequential` for sequential execution
```
MaheshRavishankar, line 26: Inst conversion from `scf.forall` to a sequential scope making this serialized loop (unless I am missing the context of `#pcf.sequential`.

MaheshRavishankar (changes requested):
Some initial comments.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.td
qedawkins: It's noted at the beginning of this pass description, but this is just a test pass for converting `scf.forall` without mapping (i.e. unspecified order w/ unmapped workers) to `pcf.sequential` (means the same thing). Naming of `pcf.sequential` aside, we'll never actually use this pass. Actual use cases will live in different directories and just call the transformation + provide the appropriate scope for the mapping.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: This is only matching the semantics of the `scf.forall`. `pcf.loop` requires `x, y, z` ordering of worker ids, so if the `scf.forall` specifies some permutation (e.g. `z, x, y`) then we need to account for it. That's what this helper is doing.

By the time we're calling this helper, we're already in the `Rewrite` section of `matchAndRewrite` and are guaranteed to be replacing the forall.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: pcf.loop only takes a list of iteration counts (i.e. it's always normalized) so we have to normalize the trip count here. The loop op already carried a lot of state so adding more just to defer this step wasn't really worth it.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: Good catch! I think you are right. I'll add a test with non-zero lb to make sure as well.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
MaheshRavishankar: I'd rather not do this explicitly by changing the IR. I would rather keep this information in the transformation itself. This might be just me, but I tend to think of scf.forall in two steps. Treat the iteration space as fastest varying innermost "logically". The mapping then tells you how to map the iteration space to the execution space and all permutations of traversals are valid. This is implicitly changing that, and it makes it confusing to reason about the transformations below.

MaheshRavishankar (changes requested):
I am reviewing the PR chain, this showed up first in my chain. I think this is OK except for 
1) I dont think the scf.forall should be modified before lowering it. Please fix. It will make it easier to understand AFAICS.
2) I think this still has the error with the apply lb and step being (s0 + s1) * s2 with {s0, s1, s2} mapped to {id, lb, step}. I think this should be s1 + s0 * s2.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: I changed it so that the forall's operands and attributes are never changed, but the body ivs are still permuted before moving the body over since it's easier to do it in that order.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: fixed

MaheshRavishankar (commented):
Ok, code changes look good. Looks like you added some changes w.r.t handling the mapping attribute, but none of the tests exercise that. Thats the only point left to address here.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
MaheshRavishankar: There are some subtle issues I have found in the past with normalized loops, but if you are going from an scf.forall post-tiling to a pcf.loop I think you wont hit those issues. I think it might be still useful to not have to normalize the loop, but we can come back to that later.

qedawkins:
Oh right, those were all exercised in the pass added here: https://github.com/iree-org/iree/pull/22807/files#diff-722928bfffa60578d77ae6bfd8518396ebd83391fb051567724000357069982d

I'll have to come up with a way to test this without pulling in too many more changes, maybe I'll add a "serializable" mapping to the codegen dialect for that.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+void ConvertForallToLoopsPass::runOnOperation() {
+  SmallVector<scf::ForallOp> opsToConvert;
+  getOperation()->walk([&](scf::ForallOp forallOp) {
+    std::optional<ArrayAttr> mapping = forallOp.getMapping();
+    // Empty mapping, no mapping, and local mapping all map to `pcf.sequential`.
+    // If it is a local mapping, then the lowering pattern will automatically
+    // handle any mapping permutation based on the mapping attribute's relative
+    // id.
+    if (!mapping || mapping->empty() || hasLocalMapping(forallOp)) {
```
Max191: small nit: maybe rename the util function to `hasEmptyOrLocalMapping` and return true for empty/null mappings. Then there's no need to duplicate the empty/null checks.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  SmallVector<int64_t> perm = getMappingPermutation(mappingAttr.value());
+  if (!isPermutationVector(perm)) {
+    return failure();
+  }
+  return perm;
+}
+
+LogicalResult matchForallConversion(scf::ForallOp forallOp) {
```
Max191: Should this function be static? I don't see it declared in the header.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
```diff
@@ -0,0 +1,312 @@
...
+      }
+    }
+  }
+  if (failed(getProcessorIdPermutation(forallOp))) {
+    return failure();
+  }
+  return success();
+}
+
+PCF::LoopOp convertForallToPCFImpl(RewriterBase &rewriter,
```
Max191: It seems this should also be static.

qedawkins:
> Ok, code changes look good. Looks like you added some changes w.r.t handling the mapping attribute, but none of the tests exercise that. Thats the only point left to address here.

I added new tests for mapping, PTAL

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
```diff
@@ -0,0 +1,312 @@
...
+  // to reconstruct the correct ids.
+  SmallVector<OpFoldResult> mixedUbs = forallOp.getMixedUpperBound();
+  applyPermutationToVector(mixedUbs, invPerm);
+  SmallVector<OpFoldResult> mixedLbs = forallOp.getMixedLowerBound();
+  applyPermutationToVector(mixedLbs, invPerm);
+  SmallVector<OpFoldResult> mixedStep = forallOp.getMixedStep();
+  applyPermutationToVector(mixedStep, invPerm);
+  // Permute the ivs of the body back to the original order by permuting the
+  // uses before we move it over to the new op.
+  permuteValues(rewriter, forallOp.getLoc(), forallOp.getInductionVars(), perm);
```
Max191: nit: Why not just permute the results of the pcf loop at the end of this function before replacing them? This way seems confusing to me, and the wording of the comment is strange because the forall results are already in the "original" order at this point.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: It's in an anonymous namespace. Same below.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: This isn't loop results, these are the induction variables. The main reason is a) there are two code paths that need to append the ivs, so permuting in two places is awkward, and b) it was easier to debug if I permuted first and could inspect the permuted op up front.

qedawkins: updated the comment to be clearer.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
qedawkins: Made them static too

MaheshRavishankar (approved):

MaheshRavishankar:
Thanks @qedawkins !

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToLoops.cpp
Max191: Oh, right. Idk why I said loop results, I meant the argReplacements (before inlining). Either way is fine to me, I just thought it's more clear to permute the list of replacements before inlining.

Max191 (approved):

qedawkins:
Changed the base to https://github.com/iree-org/iree/pull/22910 which is just adding boilerplate and was already approved so that I can land this.

PR #22912, opened 2025-12-15
[Codegen] Add FuseConsumersPass

qedawkins:
Adds the FuseConsumersPass which fuses consumer operations into the body of pcf.generic and pcf.loop ops.

This is a greedy pass that tiles consuming TilingInterface ops by their iteration domains and moves them into the producer bodies. Consumed results are replaced with writes to new allocations.

The pass supports two fusion modes:
1. Single result with multiple writes - the consumer only consumes one result but that result may be constructed from multiple writes
2. Multiple results with single writes - the consumer consumes multiple results but each result has a single write

Also adds:
- ConsumerFusionParams struct for tracking fusion parameters
- matchTilableConsumer/fuseTilableConsumer helper functions
- populatePCFDropUnusedResultPatterns for cleanup after fusion
- fuseExtractSliceIntoProducerLoop/Generic for slice extraction


compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
```diff
@@ -32,6 +38,72 @@ FailureOr<PCF::LoopOp> convertForallToPCF(RewriterBase &rewriter,
                                           PCF::ScopeAttrInterface scope,
                                           int64_t numIds = -1);
 
+struct ConsumerFusionParams {
+  // List of operands in the consumer that are fused along.
+  SmallVector<unsigned> operands;
```
MaheshRavishankar, line 43: I think both of these could be unified by just carrying the `OpOperand *`.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
```diff
@@ -32,6 +38,72 @@ FailureOr<PCF::LoopOp> convertForallToPCF(RewriterBase &rewriter,
...
+    slices.clear();
+  }
+};
+
+// Helpers to match a consumer as fusable into a producer. There are two
+// supported cases:
+//   1. The tilable |target| only consumes a single result of the producer but
+//      the produced operand may be constructed out of multiple writes within
+//      the producer.
+//   1. The tilable |target| consumes multiple results of the producer but only
```
MaheshRavishankar, line 68: I think this distinction is kind of artificial, and might suggest this isnt held correctly. You should be able to reason about multiple results from the producer (each produced by multiple writes, I assume that it is "not incorrect" to have multiple writes to the same result) being used in the consumer.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
```diff
@@ -32,6 +38,72 @@ FailureOr<PCF::LoopOp> convertForallToPCF(RewriterBase &rewriter,
...
+//      a single writing op constructs each consumed result.
+// Populates |params| with the matched information needed to perform a fusion
+// upon success. On failure |params| is cleared and a different tilable consumer
+// may be matched against.
+LogicalResult matchTilableConsumer(RewriterBase &rewriter,
+                                   PCF::GenericOp genericOp,
+                                   TilingInterface target,
+                                   ConsumerFusionParams &params);
+LogicalResult matchTilableConsumer(RewriterBase &rewriter, PCF::LoopOp loopOp,
```
MaheshRavishankar, line 94: Arent the `LoopOp` and `GenericOp` implementing the same interface? Then this should be a single function with the same interface right? Why two separate entry points?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.cpp
```diff
@@ -5,9 +5,268 @@
...
+  using Base::Base;
+  LogicalResult matchAndRewrite(IREE::PCF::LoopOp loopOp,
+                                PatternRewriter &rewriter) const override {
+    return dropUnusedResults(rewriter, loopOp);
+  }
+};
+} // namespace
+
+void populatePCFDropUnusedResultPatterns(RewritePatternSet &patterns) {
+  patterns.add<DropUnusedGenericResult, DropUnusedLoopResult>(
```
MaheshRavishankar: Why two patterns? They seem to be the same. Could just use a templated single pattern? Again, why cant we do this on the interface directly?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+      targetResults.insert(opResult.getResultNumber());
+    } else {
+      if (!dominanceInfo.dominates(operand.get(), producerOp)) {
+        return rewriter.notifyMatchFailure(
+            target, "unable to fuse due to operand dominance");
+      }
+    }
+  }
+
+  if (targetOperands.empty()) {
```
MaheshRavishankar, line 215: I dont know if I should return a failure here. I know this makes sense within a pattern rewriter, but as a stand-alone method, this should just return success.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+            "multiple operand fusion with multiple producers unsupported");
+      }
+
+      if (leader < 0) {
+        leader = currNumSlices;
+      } else {
+        PCF::WriteSliceOp currLeader = slices[leader];
+        PCF::WriteSliceOp next = slices.back();
+        if (next != currLeader) {
+          // Currently multiple producers split across block/region boundaries
```
MaheshRavishankar: This needs to be moved out to a header where it is easier to find. This is kind of buried in the implementation.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+        return rewriter.notifyMatchFailure(producerOp,
+                                           "non write slice producer");
+      }
+      if (slices.size() - currNumSlices != 1) {
+        return rewriter.notifyMatchFailure(
+            target,
+            "multiple operand fusion with multiple producers unsupported");
+      }
+
+      if (leader < 0) {
```
MaheshRavishankar, line 279: I am not convinced that this is possible to support. This is one of the drawbacks of having a write that does not connect to other writes. You have no way of guaranteeing anything, so once you have multiple writes pretty much all bets are off.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+  // Clone the op and replace all operands being fused along with unrealized
+  // conversion casts from the distributed producer tile to the undistributed
+  // tile. We will forward the input to the unrealized conversion cast directly
+  // to the tiled op once finished.
+  auto clonedOp = cast<TilingInterface>(rewriter.clone(*target));
+  SmallVector<UnrealizedConversionCastOp> unrealizedConversions;
+  for (auto [operand, slice] : llvm::zip_equal(operands, slices)) {
+    OpOperand &currOperand = clonedOp->getOpOperand(operand);
+    Type undistributedType = currOperand.get().getType();
+    auto conversion = UnrealizedConversionCastOp::create(
```
MaheshRavishankar, line 349: I am not sure why we want this. This seems like a anti-pattern. Unrealized conversion casts are for the most part an internal implementation detail of pattern rewrites. I am not sure using this way is a good idea.

I am not sure I follow what is happening here.

MaheshRavishankar (changes requested):
Ok, this is a big and involved change. I think this will take some significant time to review. I am kind-of disappointed that we have to reinvent a lot of the fusion logic. Its tricky and duplicative.

I left a few comments from an initial skim of these with things that stood out immediately. Need to understand this more. I think responses to those will help.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
qedawkins: I didn't add the interface yet. I should have scrubbed mentions of the interface from the dialect but may have missed a spot. We could add an interface but the methods would have to be quite rewrite heavy so I decided to hold off.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: Failure is what I would expect to return here. This matcher says "Is there at least one target consumer to fuse into this loop" which there are not if there are no valid fusable consumers.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: This is actually a pretty natural use of the op based on its name. We could add our own version of it for placeholder values if need be though. What this does is take this:
```
%0 = pcf.loop {
  ...
  %1 = ... : tensor<[tile]>
  pcf.write_slice %1 into %dest
} -> tensor<[full shape]>
foo.consumer %0
```
and temporarily moves it to this
```
pcf.loop {
  ...
  %1 = ... : tensor<[tile]>
  %ucst = unrealized_conversion_cast %1 : tensor<[tile]> to tensor<full shape>
  foo.consumer %ucst
} -> tensor<[full shape]>
```
Note that at this point, the unrealized_conversion_cast *is* exactly that, a pending conversion from the undistributed operand to the distributed one. This is needed all so that we can invoke the tiling interface rather than reinventing tiling customized for consumer fusion (one of the main goals with the implementation upstream as well).

We must introduce the unrealized conversion here because if for any reason the tiling interface implementation makes a mistake and doesn't load the correct input tile, we can emit a pass error simply by leaving the unrealized conversion in place. Otherwise we have no recourse other than an assert since patterns can't return fatal errors.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
qedawkins: It is somewhat artificial, yes, and I can add clarification as to why these two cases are currently un-combined, but the short answer is that the multiple producer + multiple consumer case is quite a bit more complex than these two cases. To give a few reasons:
1. Control flow actually can make the m-n case impossible if we have something like
```
%0:2 pcf.loop {
  scf.if %cond {
    pcf.write_slice %lhs_slice
  } else {
    pcf.write_slice %rhs_slice
  }
}
foo.consumer %0#0, %0#1
```
where in this case even if the slice offsets/sizes/strides are ostensibly consistent between the two operands, depending on `%cond` both aren't available at the same time.

2. Organizing multiple sets of potentially consistent writes across operands via interface methods is unwieldy and probably very inefficient.
```
%0:2 pcf.loop {
    pcf.write_slice %lhs_slice_0
    pcf.write_slice %rhs_slice_2
    pcf.write_slice %lhs_slice_1
    pcf.write_slice %rhs_slice_1
    pcf.write_slice %lhs_slice_2
    pcf.write_slice %rhs_slice_0
}
```
like in this example. Figuring out which slice matches with which operands by repeatedly querying the fusability query method with different combinations isn't a usable implementation.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+  };
+
+  if (auto dpsOp = dyn_cast<DestinationStyleOpInterface>(*target)) {
+    for (Value init : dpsOp.getDpsInits()) {
+      if (init.getDefiningOp() == producerOp) {
+        // There are two options if we are fusing along an init operand:
+        //  1. Create a new empty init with the same shape.
+        //  2. Use the init of the producer op.
+        //
+        // This ops for the latter because when fusing we'll replace the actual
```
MaheshRavishankar: Nit: Typo `s/ops/opts`.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+    // the producer's operands. To do this we replace all operands of |target|
+    // coming from the producer with equivalently shaped inits/tensor.empty ops
+    // and call reification on that. This is ostensibly a hack as there is no
+    // formal guarantee that swapping out operands like this is valid per the
+    // interface, however in practice this is valid for all known operations
+    // that implement the interface today.
+
+    SmallVector<Value> originalOperands;
+    rewriter.setInsertionPoint(producerOp);
+    for (unsigned operandIndex : params.operands) {
```
MaheshRavishankar, line 570: This seems pretty dangerous, but I see why you are doing this. (1), in tiling, I prefer to just clone things, and make the original operation dead on success. We arent going to be fusing a consumer that has a lot of operations. Cloning just makes things much e asier when you are cloning in ways that dont necessarily respect semantics. On failure, you have cloned a few things and then it just leaves those operations dead and you basically abort. You cant use PatternRewriterDriver here is the downside, but it is much easier to do a worklist than to shoe horn this into pattern rewriter driver. 

What you have here is still not enough though. You need to basically either move the consumer just after the producer, or move the producer just before the consumer. Chose the the former option (upstream does the latter which I should fix sometime). To move the consumer, you need to move the operations that define the operands of the consumer before the producer. You can do that using the `moveValueDefinitions`  method upstream for all the operands not defined by the producer and, where the insertion point to which you are moving the dependencies is the producer. If it fails, you cant clone this operation just before the producer.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
```diff
@@ -32,6 +38,72 @@ FailureOr<PCF::LoopOp> convertForallToPCF(RewriterBase &rewriter,
...
+  // interpretations. If |results| > 1, then |slices| == |operands| must hold
+  // and each slice corresponds to the sole write to the result consumed by the
+  // corresponding operand. If |results| == 1, then each slice is a different
+  // write to the same sole result.
+  //
+  // In the first case, slices[0] is the most dominant slice (and thus the
+  // insertion point for the fused op).
+  SmallVector<PCF::WriteSliceOp> slices;
+
+  void clear() {
```
MaheshRavishankar, line 56: Eh. We arent gaining much by "reusing" this. Just throw it away when you are done with it. I dont think you need a clear method.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
MaheshRavishankar: Ok, that is sort of interesting way of staging this. Ill go with this.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+    if (leader != 0) {
+      // Swap the most dominant slice to the beginning.
+      std::swap(slices[leader], slices[0]);
+      std::swap(targetOperands[leader], targetOperands[0]);
+    }
+  }
+  return success();
+}
+
+static void fuseIntoWriteSlices(RewriterBase &rewriter, TilingInterface target,
```
MaheshRavishankar, line 327: I think you are better off returning a `LogicalResult` here. Even if it means just returning `success` always. Then you are already setup for error propagation rather than having to do it later. Also then all callers of this method also are forced to propagate any errors.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+
+  // Get the iteration domain in terms of the operand tiles. This is required
+  // to fetch the result tile positions. This and all subsequent interface
+  // queries must succeed per the matcher check.
+  SmallVector<SmallVector<OpFoldResult>> allOffsets = llvm::map_to_vector(
+      slices, [](PCF::WriteSliceOp op) { return op.getMixedOffsets(); });
+  SmallVector<SmallVector<OpFoldResult>> allSizes = llvm::map_to_vector(
+      slices, [](PCF::WriteSliceOp op) { return op.getMixedSizes(); });
+  SmallVector<OpFoldResult> iterDomainOffsets, iterDomainSizes;
+  LogicalResult res = clonedOp.getIterationDomainTileFromOperandTiles(
```
MaheshRavishankar: To highlight the above statement with `void` return type. Just return the error here. I typically think its bad form to have `(void)res`.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
MaheshRavishankar: Also (here and in other `static` functions defined here) could you add some description of what the operands are expected to be. For example, not clear to me why `slices` is a `MutableArrayRef`

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
```diff
@@ -0,0 +1,916 @@
...
+static LogicalResult
+fuseExtractSliceIntoProducerImpl(RewriterBase &rewriter, OpTy producerOp,
+                                 tensor::ExtractSliceOp extractSliceOp) {
+  OpResult producerResult = cast<OpResult>(extractSliceOp.getSource());
+  if (!producerResult.hasOneUse()) {
+    return rewriter.notifyMatchFailure(producerOp,
+                                       "producer result has multiple uses");
+  }
+
+  // Only zero offset extract_slice ops are supported.
```
MaheshRavishankar, line 674: Does this not preclude most `extract_slices` that we usually see.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
```diff
@@ -32,6 +38,72 @@ FailureOr<PCF::LoopOp> convertForallToPCF(RewriterBase &rewriter,
...
+
+// Pattern set for dropping unused results from scoped ops. Due to memory
+// effects this requires cascading operation erasure and is unsuitable for
+// a canonicalization pattern.
+void populatePCFDropUnusedResultPatterns(RewritePatternSet &patterns);
+
+// Fuse a tensor.extract_slice consumer into a pcf.loop producer. This shrinks
+// the result to the extracted size and clamps all write_slice ops accordingly.
+LogicalResult
+fuseExtractSliceIntoProducerLoop(RewriterBase &rewriter, PCF::LoopOp loopOp,
```
MaheshRavishankar, line 113: Could you explain a bit more what this is used for. I havent seen this typically needed.

MaheshRavishankar (commented):
Phew! Finally got through this PR. In the end I get what it is doing. Left a few comments regarding implementation, but otherwise this looks good to go.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
qedawkins: This is run all the time in GPU today: https://github.com/iree-org/iree/blob/9273e925c1319312b1f7a4e9ad0d6b2f4c28995c/compiler/src/iree/compiler/Codegen/Common/GPU/GPUFuseAndHoistParallelLoops.cpp#L311

It's how we do unpadding since without it we're saddled with a copy into a the smaller result buffer.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: I can't return an error here since this is happening in the rewrite section of a pattern, but I can assert on it.

qedawkins: oh I already asserted, switched to [[maybe_unused]]

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: Same comment, this is matching this pattern: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/GPU/GPUFuseAndHoistParallelLoops.cpp#L311

We only support fusing "unpadding" extract slices there so I did the same here.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: This function can never be allowed to fail so I don't want to return a LogicalResult here. It's used as the rewrite part of a pattern rewrite *after* the match has happened. If we wanted a failure case in here, it should go in the matcher.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: > What you have here is still not enough though. You need to basically either move the consumer just after the producer, or move the producer just before the consumer.

This is good enough with the current impl because the matcher makes the most basic dominance guarantees needed for fusion (all operands to the consumer are either from the producer or dominate the producer). This means `moveValueDefinitions` use cases aren't supported just yet. If you want me to support that right now I can, but I was able to get away with not supporting it yet so I didn't.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
qedawkins: Using the clear method made it easier to write thiis matching loop across all users:
```
    ConsumerFusionParams params;
    TilingInterface fusionTarget;
    for (Operation *user : genericOp->getUsers()) {
      fusionTarget = dyn_cast<TilingInterface>(user);
      if (fusionTarget && succeeded(matchTilableConsumer(
                              rewriter, genericOp, fusionTarget, params))) {
        break;
      }
      fusionTarget = TilingInterface();
    }
```
Without clearing the `params` before each one, I'd have to do something like
```
    ConsumerFusionParams params;
    TilingInterface fusionTarget;
    for (Operation *user : genericOp->getUsers()) {
      fusionTarget = dyn_cast<TilingInterface>(user);
      ConsumerFusionParams tempParams;
      if (fusionTarget && succeeded(matchTilableConsumer(
                              rewriter, genericOp, fusionTarget, tempParams))) {
        params = tempParams;
        break;
      }
      fusionTarget = TilingInterface();
    }
```
So now I have 2 params structs which I thought looked ugly.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
qedawkins: I could but I would end up having to recover some information during the fusion rewrite implementation. The operands are stored as a list while the results are stored as a set since it's N results -> M operands with N <= M. This makes it easier to check how many results we're fusing along and adjust the logic accordingly.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
MaheshRavishankar: Ok, leaving a clear like that means there is a chance you dont clear and things mess up. What you have in the second case is no wastage of space if you `std::swap(params, tempParams)`. That paradigm allows not carrying state within objects that you want to "clear".

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
MaheshRavishankar: But that is not an error right?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
MaheshRavishankar: Missing comments on what the operands are?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
MaheshRavishankar: I am pretty sure you will hit the issue soon. But we can defer to that point for now.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
MaheshRavishankar: Ok, I wont hold this PR for that, but why this restriction at all?

MaheshRavishankar (approved):
Ok, just a few minor comments on missing comments. Approving but had a few more questions that I left here.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: It's not an error, it just means the match failed. Failure != error, just that the pattern doesn't apply.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: Just because the logic for fusing extract_slice ops doesn't support it yet (I pretty much just copied the logic from the existing pattern). I can extend it in a follow up.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FuseConsumers.cpp
qedawkins: Oh also forgot to explain why `slices` is a `MutableArrayRef`. It's because `slice.getSource()` is non-const so it needs to be mutable. I might have made a mistake in the op definition that resulted in this, if you know why a tablegen generated getter would be non-const.

qedawkins: Added a comment on the operands as well.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
qedawkins: Sorry missed this comment on my last pass, change is up here: #23065

too many boxes, though I got them all :(

PR #22915, opened 2025-12-15
[Codegen] Add LowerStructuralPCFPass

qedawkins:
Adds the LowerStructuralPCFPass which lowers structural PCF operations to SCF:

- pcf.generic is lowered to scf.execute_region
- pcf.loop is lowered to scf.forall
- pcf.branch_cond_return is lowered to cf.cond_br

The pass uses the scope attribute on each operation to determine how to generate worker IDs and counts for the lowered operations. Synchronization barriers are inserted when sync_on_return is set.

Also adds loadStructuralLoweringDependentDialects to PCFConversionDialectInterface, allowing external dialects to register dependent dialects needed during lowering.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
@@ -1,0 +1,255 @@
+// Copyright 2025 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Dialect/PCF/IR/PCF.h"
+#include "iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.h"
+#include "iree/compiler/Codegen/Dialect/PCF/Transforms/ConversionDialectInterface.h"
+#include "iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.h"
+#include "iree/compiler/Utils/RewriteUtils.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVectorExtras.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlow.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Support/WalkResult.h"
+#include "mlir/Transforms/WalkPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-pcf-lower-structural-pcf"
+
+namespace mlir::iree_compiler::IREE::PCF {
+
+#define GEN_PASS_DEF_LOWERSTRUCTURALPCFPASS
+#include "iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.h.inc"
+
+namespace {
+
+class LoadDependentDialectExtension : public DialectExtensionBase {
+public:
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(LoadDependentDialectExtension)
+
+  LoadDependentDialectExtension() : DialectExtensionBase(/*dialectNames=*/{}) {}
+
+  void apply(MLIRContext *context,
+             MutableArrayRef<Dialect *> dialects) const final {
+    for (Dialect *dialect : dialects) {
+      auto *iface = dyn_cast<PCFConversionDialectInterface>(dialect);
+      if (!iface) {
+        continue;
+      }
+      iface->loadStructuralLoweringDependentDialects(context);
+    }
+  }
+
+  /// Return a copy of this extension.
+  std::unique_ptr<DialectExtensionBase> clone() const final {
+    return std::make_unique<LoadDependentDialectExtension>(*this);
+  }
+};
+
+struct LowerStructuralPCFPass final
+    : impl::LowerStructuralPCFPassBase<LowerStructuralPCFPass> {
+  void getDependentDialects(DialectRegistry &registry) const override {
+    // Direct dialect deps.
+    registry.insert<iree_compiler::IREE::PCF::PCFDialect, scf::SCFDialect,
+                    cf::ControlFlowDialect>();
+    registry.addExtensions<LoadDependentDialectExtension>();
+  }
+  void runOnOperation() override;
+};
+
+struct LowerGenericOp : public OpRewritePattern<IREE::PCF::GenericOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(IREE::PCF::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    if (genericOp->getNumResults() != 0) {
+      return rewriter.notifyMatchFailure(
+          genericOp, "unexpected attempt to resolve generic op with results");
+    }
+
+    if (genericOp.getSyncOnReturn()) {
+      OpBuilder::InsertionGuard g(rewriter);
+      rewriter.setInsertionPointAfter(genericOp);
+      if (failed(genericOp.getScope().addBarrier(rewriter))) {
+        genericOp.emitOpError("failed to construct requested barrier");
+        return failure();
+      }
+    }
+
+    Location loc = genericOp.getLoc();
+    SmallVector<Value> indexArgs = genericOp.getScope().getWorkerIDs(
+        rewriter, loc, genericOp.getNumIterators());
+    assert(indexArgs.size() == genericOp.getNumIterators() &&
+           "expected num worker ids to match number of iterators");
+
+    indexArgs.append(genericOp.getScope().getWorkerCounts(
+        rewriter, loc, genericOp.getNumIterators()));
+    assert(indexArgs.size() == genericOp.getNumIndexArgs() &&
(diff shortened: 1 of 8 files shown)


MaheshRavishankar (commented):
Without reviewing this yet. I dont think we should be using `scf.execute_region` at all. Can you say whats the reason for this?

qedawkins:
> Without reviewing this yet. I dont think we should be using `scf.execute_region` at all. Can you say whats the reason for this?

I am open to alternatives, but that's pretty much the exact op I need here. I have
```
pcf.generic {
  /* some region to execute */
}
```
and need to execute its region without breaking up the parent block (because we haven't lowered scf to cf yet). This is kinda exactly what `scf.execute_region` is for IIUC (inlining function calls is the example given in the op docs) when you aren't able to break up the parent block yet (because the parent requires single block bodies).

krzysz00:
Wait, doesn't that make `execute_region` equivalent to just ... inlining the contents of the region? Or am I missing something?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.td
```diff
@@ -83,4 +83,31 @@ def FusePCFWritesPass : Pass<"iree-pcf-fuse-pcf-writes", ""> {
...
+
+    The input is expected to be IR where PCF ops no longer have tied results
+    (after `iree-pcf-convert-sref-to-memref`). The scope attribute on each op
+    determines how worker IDs and counts are materialized.
+
+    Op conversions:
+    - `pcf.generic` -> `scf.execute_region` with worker IDs from scope
+    - `pcf.loop` -> serialized `scf.forall` with iteration bounds from count operands
+    - `pcf.return` -> `scf.yield` (in generic) or `scf.forall.in_parallel` (in loop)
+    - `pcf.branch_cond_return` -> `cf.cond_br` to a return block
```
krzysz00, line 101: ... why do we need `cf` here? Couldn't we have `if` or `switch` or such?

krzysz00: Ah. That's why.

... Is it just me, or should this conditional branch in PCF be something more like a `switch`-type op that takes multiple regions to choose between? That means we aren't mixing structured and unstructured control flow

qedawkins:
> Wait, doesn't that make `execute_region` equivalent to just ... inlining the contents of the region? Or am I missing something?

if you have something like
```
scf.for {
  scf.execute_region {
    ^bb0: ...
    ^bb1: ...
  }
}
```
you cannot inline the body of the execute region because scf.for requires a single block body. So the solution is to use the execute_region to hold the region and then lower both at the same time.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
```diff
@@ -0,0 +1,253 @@
...
+        genericOp.getRegion().getBlocks(), [](Block &b) { return &b; });
+    for (Block *block : blocksToMove) {
+      rewriter.moveBlockBefore(block, &executeRegion.getRegion(),
+                               executeRegion.getRegion().end());
+    }
+
+    // Replace all pcf.return ops with scf.yield ops. We make this pattern
+    // responsible for terminator conversion to try to keep IR valid between
+    // pattern rewrites.
+    for (auto &block : executeRegion.getRegion()) {
```
kuhar: nit: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
```diff
@@ -0,0 +1,253 @@
...
+      }
+    }
+
+    rewriter.eraseOp(genericOp);
+
+    return success();
+  }
+};
+
+struct LowerLoopOp : public OpRewritePattern<IREE::PCF::LoopOp> {
```
kuhar:
```suggestion
struct LowerLoopOp final : OpRewritePattern<IREE::PCF::LoopOp> {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
```diff
@@ -0,0 +1,253 @@
...
+           "expected worker id count to match number of id args");
+
+    // Create an scf.forall op with no mapping. This is moderately more
+    // expressive than scf.for because it indicates that loop iterations
+    // are independent of one another.
+    //
+    // `scf.forall` orders its loop parameters from slowest to fasted varying,
+    // so reverse lbs/ubs/steps.
+    SmallVector<OpFoldResult> lbs(ids.rbegin(), ids.rend());
+    SmallVector<OpFoldResult> steps(workerCounts.rbegin(), workerCounts.rend());
```
kuhar: nit/optional: `auto bar = to_vector(reverse(foo));`

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
```diff
@@ -0,0 +1,253 @@
...
+  OpBuilder::InsertionGuard g(rewriter);
+  Block *newBlock =
+      rewriter.createBlock(region, region->end(), argTypes,
+                           SmallVector<Location>(argTypes.size(), loc));
+  PCF::ReturnOp::create(rewriter, loc);
+  return newBlock;
+}
+
+struct LowerBranchCondReturnOp
+    : public OpRewritePattern<IREE::PCF::BranchCondReturnOp> {
```
kuhar:
```suggestion
struct LowerBranchCondReturnOp final
    : OpRewritePattern<IREE::PCF::BranchCondReturnOp> {
```

kuhar (commented):
The logic looks good to me, I'll leave the exact lowering to scf for others to decide

kuhar (approved):

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
```diff
@@ -0,0 +1,253 @@
...
+        rewriter, loc, genericOp.getNumIterators());
+    assert(indexArgs.size() == genericOp.getNumIterators() &&
+           "expected num worker ids to match number of iterators");
+
+    indexArgs.append(genericOp.getScope().getWorkerCounts(
+        rewriter, loc, genericOp.getNumIterators()));
+    assert(indexArgs.size() == genericOp.getNumIndexArgs() &&
+           "expected num worker counts to match number of iterators");
+
+    auto executeRegion =
```
MaheshRavishankar, line 92: Why do you even need an `scf.execute_region` . Just inline the block where the `pcf` exists? ?

qedawkins: It's because `pcf.generic` takes a generic (multi-block) region, so
```
scf.for {
  pcf.generic {
    ^bb0
    ^bb1
  }
}
```
scf.for requires a single block body in this example so inlining the body of the pcf.generic is not allowed. We need an `scf.execute_region` to hold the multi-block region in the interim (until we lower scf).

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.td
qedawkins: Regions are limited because either we get no implicit capture within regions at the same scope, or we switch into a finer scope, in which case that's just reinventing one of the already existing `switch` ops. I'll admit that I haven't landed on a super strong use case for `pcf.branch_cond_return` yet, but I haven't implemented the use case I had in mind for it yet, so we will see. I am willing to remove it if prompted but in it's current state it does "work" :P.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
qedawkins: I might not be doing it right but
```
SmallVector<OpFoldResult> steps = llvm::to_vector(llvm::reverse(workerCounts));
```
doesn't work because `workerCounts` is a list of values and I'm turning it into a list of OpFoldResults and I don't see a way to change the return type on to_vector (only the size is templated).

kuhar: you can do `to_vector_of<OpFoldResult>`?

qedawkins: oh I had no idea that existed

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
MaheshRavishankar: `pcf.generic` needs a multi-block region? I thought it was a multi-region op with single block per region.... not sure if that makes a difference though.

MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/LowerStructuralPCF.cpp
qedawkins: It is multi-region and multi-block. It has a single initializer region that is single block and a single execution region which is multi-block. The problem with multi-region is you don't get implicit capture across blocks so I went with multiple blocks.

PR #22916, opened 2025-12-15
[Codegen] Add ConvertSRefToMemRefPass

qedawkins:
Adds the ConvertSRefToMemRefPass which converts pcf.sref types to memrefs.

The pass performs layout propagation analysis using DFX to determine the most specific memref layout for each sref value based on how it's produced and used. It then converts:
- pcf.alloc to memref.alloc
- pcf.write_slice to memref.subview + store operations
- pcf.read_slice to memref.subview + load operations
- pcf.get_memref to memref.subview with layout casts
- pcf.generic/loop result types from tensor to memref

Function signatures are updated to use memref types instead of sref types, with proper layout inference from call sites.

Also adds loadSRefLoweringDependentDialects to
PCFConversionDialectInterface, allowing external dialects to register dependent dialects needed during sref conversion.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
@@ -86,7 +86,7 @@ static bool isDefaultOrStrided(Attribute layout) {
 }
 
 // State that tracks floating point ranges and flags.
-struct StridedLayoutState : public DFX::AbstractState {
+struct StridedLayoutState : DFX::AbstractState {
   bool isValidState() const override { return isValid; }
   bool isAtFixpoint() const override { return !isValid || isFinalized; }
 
@@ -230,6 +230,11 @@ public:
 private:
   void initializeValue(Value value, DFX::Solver &solver) override;
   ChangeStatus updateValue(Value value, DFX::Solver &solver) override;
+  ChangeStatus updateOpResult(OpResult result, StridedLayoutState &newState,
+                              DFX::Solver &solver);
+  ChangeStatus updateBlockArgument(BlockArgument value,
+                                   StridedLayoutState &newState,
+                                   DFX::Solver &solver);
 };
 const char StridedLayoutValueElement::ID = 0;
 
@@ -241,124 +246,91 @@ void StridedLayoutValueElement::initializeValue(Value value,
   }
 }
 
-ChangeStatus StridedLayoutValueElement::updateValue(Value value,
-                                                    DFX::Solver &solver) {
-  StridedLayoutState newState = getState();
-
-  if (auto result = llvm::dyn_cast<OpResult>(value)) {
-    llvm::TypeSwitch<Operation *, void>(result.getOwner())
-        .Case([&](PCF::AllocOp allocOp) {
-          PCF::ShapedRefType resultType = allocOp.getResultType();
-          FailureOr<Attribute> memSpace =
-              resultType.getScope().getAllocMemSpace(allocOp.getContext());
-          if (failed(memSpace)) {
-            allocOp->emitOpError("failed to get memory space for allocation");
-            newState.invalidate();
-            return;
-          }
-          newState.setAssumed(MemRefType::get(
-              resultType.getShape(), resultType.getElementType(),
-              MemRefLayoutAttrInterface{}, memSpace.value()));
-          newState.indicateOptimisticFixpoint();
-        })
-        .Case([&](RegionBranchOpInterface regionOp) {
-          // For region branch ops get the result layout from the union of
-          // return sites.
-          if (solver.getExplorer().walkReturnOperands(
-                  regionOp.getOperation(), [&](OperandRange returnOperands) {
-                    auto returnOperand =
-                        returnOperands[result.getResultNumber()];
-                    auto returnState =
-                        solver.getElementFor<StridedLayoutValueElement>(
-                            *this, Position::forValue(returnOperand),
-                            DFX::Resolution::REQUIRED);
-                    newState ^= returnState;
-                    return WalkResult::advance();
-                  }) == TraversalResult::INCOMPLETE) {
-            newState.indicatePessimisticFixpoint();
-            return;
-          }
-        })
-        .Case([&](CallOpInterface callOp) {
-          // Give up pessimistically on indirect calls.
-          if (isa<Value>(callOp.getCallableForCallee())) {
-            newState.indicatePessimisticFixpoint();
-            return;
-          }
-          auto targetSymbol =
-              cast<SymbolRefAttr>(callOp.getCallableForCallee());
-          auto callableOp = solver.getExplorer()
-                                .getSymbolTables()
-                                .lookupNearestSymbolFrom<CallableOpInterface>(
-                                    callOp, targetSymbol);
-          assert(callableOp && "call target not found");
-          // For region branch ops get the result layout from the union of
-          // return sites.
-          if (solver.getExplorer().walkReturnOperands(
-                  callableOp, [&](OperandRange returnOperands) {
-                    auto returnOperand =
-                        returnOperands[result.getResultNumber()];
-                    auto returnState =
-                        solver.getElementFor<StridedLayoutValueElement>(
-                            *this, Position::forValue(returnOperand),
-                            DFX::Resolution::REQUIRED);
-                    newState ^= returnState;
-                    return WalkResult::advance();
-                  }) == TraversalResult::INCOMPLETE) {
-            newState.indicatePessimisticFixpoint();
-            return;
-          }
-        })
-        .Case([&](Util::OptimizationBarrierOp barrierOp) {
-          auto returnState = solver.getElementFor<StridedLayoutValueElement>(
-              *this,
-              Position::forValue(
-                  barrierOp.getOperand(result.getResultNumber())),
-              DFX::Resolution::REQUIRED);
-          newState ^= returnState;
-        });
-  } else if (auto bbArg = llvm::dyn_cast<BlockArgument>(value)) {
-    bool didUpdate = false;
-    if (bbArg.getParentBlock()->isEntryBlock()) {
-      didUpdate =
-          llvm::TypeSwitch<Operation *, bool>(bbArg.getOwner()->getParentOp())
-              .Case<PCF::GenericOp>([&](PCF::GenericOp genericOp) {
-                if (genericOp.isRegionRefArg(bbArg)) {
-                  auto resultType = dyn_cast<MemRefType>(
-                      genericOp.getTiedResult(bbArg).getType());
-                  if (!resultType ||
-                      !isDefaultOrStrided(resultType.getLayout())) {
-                    genericOp->emitOpError(
-                        "unexpected non-strided or default memref result type ")
-                        << resultType;
-                    newState.invalidate();
-                  } else {
-                    newState.setAssumed(resultType);
-                    newState.indicateOptimisticFixpoint();
-                  }
-                } else {
-                  // pcf.sref arguments must either be result tied or
-                  // initialized per the verifier.
-                  assert(genericOp.isInitializedArg(bbArg) &&
-                         "unexpected non-initialized arg");
-                  auto yield = cast<PCF::YieldOp>(
-                      genericOp.getInitializer().front().getTerminator());
-                  auto initializerState =
+ChangeStatus StridedLayoutValueElement::updateOpResult(
+    OpResult result, StridedLayoutState &newState, DFX::Solver &solver) {
+  llvm::TypeSwitch<Operation *, void>(result.getOwner())
+      .Case([&](PCF::AllocOp allocOp) {
+        PCF::ShapedRefType resultType = allocOp.getResultType();
+        FailureOr<Attribute> memSpace =
+            resultType.getScope().getAllocMemSpace(allocOp.getContext());
+        if (failed(memSpace)) {
+          allocOp->emitOpError("failed to get memory space for allocation");
+          newState.invalidate();
+          return;
+        }
+        newState.setAssumed(
+            MemRefType::get(resultType.getShape(), resultType.getElementType(),
+                            MemRefLayoutAttrInterface{}, memSpace.value()));
+        newState.indicateOptimisticFixpoint();
+      })
+      .Case([&](RegionBranchOpInterface regionOp) {
+        // For region branch ops get the result layout from the union of
+        // return sites.
+        if (solver.getExplorer().walkReturnOperands(
+                regionOp.getOperation(), [&](OperandRange returnOperands) {
+                  auto returnOperand = returnOperands[result.getResultNumber()];
+                  auto returnState =
                       solver.getElementFor<StridedLayoutValueElement>(
-                          *this,
-                          Position::forValue(
-                              yield->getOperand(bbArg.getArgNumber())),
+                          *this, Position::forValue(returnOperand),
                           DFX::Resolution::REQUIRED);
-                  newState ^= initializerState;
-                }
-                return true;
-              })
-              .Case<PCF::LoopOp>([&](PCF::LoopOp loopOp) {
-                auto resultType =
-                    dyn_cast<MemRefType>(loopOp.getTiedResult(bbArg).getType());
+                  newState ^= returnState;
+                  return WalkResult::advance();
+                }) == TraversalResult::INCOMPLETE) {
+          newState.indicatePessimisticFixpoint();
+          return;
+        }
+      })
+      .Case([&](CallOpInterface callOp) {
+        // Give up pessimistically on indirect calls.
+        if (isa<Value>(callOp.getCallableForCallee())) {
+          newState.indicatePessimisticFixpoint();
+          return;
+        }
+        auto targetSymbol = cast<SymbolRefAttr>(callOp.getCallableForCallee());
+        auto callableOp = solver.getExplorer()
+                              .getSymbolTables()
+                              .lookupNearestSymbolFrom<CallableOpInterface>(
+                                  callOp, targetSymbol);
+        assert(callableOp && "call target not found");
+        // For region branch ops get the result layout from the union of
+        // return sites.
+        if (solver.getExplorer().walkReturnOperands(
+                callableOp, [&](OperandRange returnOperands) {
+                  auto returnOperand = returnOperands[result.getResultNumber()];
+                  auto returnState =
+                      solver.getElementFor<StridedLayoutValueElement>(
+                          *this, Position::forValue(returnOperand),
+                          DFX::Resolution::REQUIRED);
+                  newState ^= returnState;
+                  return WalkResult::advance();
+                }) == TraversalResult::INCOMPLETE) {
+          newState.indicatePessimisticFixpoint();
+          return;
+        }
+      })
+      .Case([&](Util::OptimizationBarrierOp barrierOp) {
+        auto returnState = solver.getElementFor<StridedLayoutValueElement>(
+            *this,
+            Position::forValue(barrierOp.getOperand(result.getResultNumber())),
+            DFX::Resolution::REQUIRED);
+        newState ^= returnState;
+      });
+  return DFX::clampStateAndIndicateChange(getState(), newState);
+}
+
+ChangeStatus StridedLayoutValueElement::updateBlockArgument(
+    BlockArgument bbArg, StridedLayoutState &newState, DFX::Solver &solver) {
+  bool didUpdate = false;
+  if (bbArg.getParentBlock()->isEntryBlock()) {
+    didUpdate =
+        llvm::TypeSwitch<Operation *, bool>(bbArg.getOwner()->getParentOp())
+            .Case([&](PCF::GenericOp genericOp) {
+              if (genericOp.isRegionRefArg(bbArg)) {
+                auto resultType = dyn_cast<MemRefType>(
+                    genericOp.getTiedResult(bbArg).getType());
                 if (!resultType ||
                     !isDefaultOrStrided(resultType.getLayout())) {
-                  loopOp->emitOpError(
+                  genericOp->emitOpError(
                       "unexpected non-strided or default memref result type ")
                       << resultType;
                   newState.invalidate();
@@ -366,25 +338,65 @@ ChangeStatus StridedLayoutValueElement::updateValue(Value value,
                   newState.setAssumed(resultType);
                   newState.indicateOptimisticFixpoint();
                 }
-                return true;
-              })
-              .Default([&](Operation *) { return false; });
-    }
-    if (!didUpdate) {
-      solver.getExplorer().walkIncomingBranchOperands(
-          bbArg.getOwner(),
-          [&](Block *sourceBlock, OperandRange operands, size_t offset) {
-            auto bbArgState = solver.getElementFor<StridedLayoutValueElement>(
-                *this,
-                Position::forValue(operands[bbArg.getArgNumber() + offset]),
-                DFX::Resolution::REQUIRED);
-            newState ^= bbArgState;
-            return WalkResult::advance();
-          });
-    }
+              } else {
+                // pcf.sref arguments must either be result tied or
+                // initialized per the verifier.
+                assert(genericOp.isInitializedArg(bbArg) &&
+                       "unexpected non-initialized arg");
+                auto yield = cast<PCF::YieldOp>(
+                    genericOp.getInitializer().front().getTerminator());
+                auto initializerState =
+                    solver.getElementFor<StridedLayoutValueElement>(
+                        *this,
+                        Position::forValue(
+                            yield->getOperand(bbArg.getArgNumber())),
+                        DFX::Resolution::REQUIRED);
+                newState ^= initializerState;
+              }
+              return true;
+            })
+            .Case([&](PCF::LoopOp loopOp) {
+              auto resultType =
+                  dyn_cast<MemRefType>(loopOp.getTiedResult(bbArg).getType());
+              if (!resultType || !isDefaultOrStrided(resultType.getLayout())) {
+                loopOp->emitOpError(
+                    "unexpected non-strided or default memref result type ")
+                    << resultType;
+                newState.invalidate();
+              } else {
+                newState.setAssumed(resultType);
+                newState.indicateOptimisticFixpoint();
+              }
+              return true;
+            })
+            .Default([&](Operation *) { return false; });
+  }
+  if (!didUpdate) {
+    solver.getExplorer().walkIncomingBranchOperands(
+        bbArg.getOwner(),
+        [&](Block *sourceBlock, OperandRange operands, size_t offset) {
+          auto bbArgState = solver.getElementFor<StridedLayoutValueElement>(
+              *this,
+              Position::forValue(operands[bbArg.getArgNumber() + offset]),
+              DFX::Resolution::REQUIRED);
+          newState ^= bbArgState;
+          return WalkResult::advance();
+        });
+  }
+  return DFX::clampStateAndIndicateChange(getState(), newState);
+}
+
+ChangeStatus StridedLayoutValueElement::updateValue(Value value,
+                                                    DFX::Solver &solver) {
+  StridedLayoutState newState = getState();
+
+  // Handle op results.
+  if (auto result = dyn_cast<OpResult>(value)) {
+    return updateOpResult(result, newState, solver);
   }
 
-  return DFX::clampStateAndIndicateChange(getState(), newState);
+  // Else it is a block argument.
+  return updateBlockArgument(cast<BlockArgument>(value), newState, solver);
 }
 
 const std::string
@@ -568,8 +580,7 @@ struct ConvertGenericOp final : OpConversionPattern<PCF::GenericOp> {
   LogicalResult
   matchAndRewrite(PCF::GenericOp genericOp, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    if (llvm::any_of(genericOp.getResultTypes(),
-                     [](Type t) { return !isa<MemRefType>(t); })) {
+    if (!llvm::all_of(genericOp.getResultTypes(), llvm::IsaPred<MemRefType>)) {
       return rewriter.notifyMatchFailure(
           genericOp, "expected all parallel op results to be of memref type");
     }
@@ -846,7 +857,7 @@ struct ConvertOptimizationBarrier final
 
 /// Converts the operand and result types of the CallOp, used together with the
 /// FuncOpSignatureConversion.
-struct ConvertFuncOp final : public OpConversionPattern<func::FuncOp> {
+struct ConvertFuncOp final : OpConversionPattern<func::FuncOp> {
   ConvertFuncOp(TypeConverter &typeConverter, MLIRContext *context,
                 SRefLayoutAnalysis &mapping)
       : OpConversionPattern(typeConverter, context), mapping(mapping) {}
@@ -880,7 +891,7 @@ private:
   SRefLayoutAnalysis &mapping;
 };
 
-class ConvertReturnOp final : public OpConversionPattern<func::ReturnOp> {
+struct ConvertReturnOp final : OpConversionPattern<func::ReturnOp> {
 public:
   ConvertReturnOp(TypeConverter &typeConverter, MLIRContext *context,
                   SRefLayoutAnalysis &mapping)
@@ -1031,7 +1042,7 @@ struct ConvertWhileOp final : OpConversionPattern<scf::WhileOp> {
 };
 
 void ConvertSRefToMemRefPass::runOnOperation() {
-  auto *context = &getContext();
+  MLIRContext *context = &getContext();
 
   SRefLayoutAnalysis analysis(getOperation());
   if (failed(analysis.run())) {
--- a/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
@@ -85,7 +85,8 @@ static bool isDefaultOrStrided(Attribute layout) {
          isa<StridedLayoutAttr>(layout);
 }
 
-// State that tracks floating point ranges and flags.
+// State that tracks memref layouts per value. Lattice description is below
+// next to the state fields.
 struct StridedLayoutState : DFX::AbstractState {
   bool isValidState() const override { return isValid; }
   bool isAtFixpoint() const override { return !isValid || isFinalized; }
@@ -199,8 +200,26 @@ struct StridedLayoutState : DFX::AbstractState {
   }
 
 private:
+  // Assumed state for the value. Since shape consistency is already carried on
+  // the !pcf.sref type, only the layout and memory space is relevant for the
+  // analysis. The state starts uninitialized and is progressively unioned with
+  // all producers under the following cases:
+  // 1. If this state is uninitialized, take the incoming state.
+  // 2. If the incoming state is uninitialized, do nothing.
+  // 3. Else update the state with the minimally dynamic layout covering both
+  //    this and the incoming state.
+  // The lattice first initializes all values and then moves towards the least
+  // dynamic fixed point that satisfies all incoming memref types.
   MemRefType assumed = MemRefType();
+  // Indicates whether the state is at a fixed point. Only set for values that
+  // have a locally determined type (e.g. allocs and tied args).
   bool isFinalized = false;
+  // Indicates whether the state is valid. Invalid states are irrecoverable
+  // and result in a conversion failure. This happens on:
+  // 1. Irreconcilable memref layouts (e.g. different ranks).
+  // 2. Non-strided layouts.
+  // 3. Conflicting memory spaces.
+  // 4. Failure to converge to a fixed point.
   bool isValid = true;
 };
 
@@ -574,6 +593,29 @@ static void castIfMismatched(OpBuilder &b, Location loc,
   }
 }
 
+/// Convert all result tied values of a `pcf.generic`, e.g.
+///
+/// ```
+///   %src = ... : memref<?x?xi32, strided<[?, 1]>, 3>
+///   %0 = pcf.generic
+///     execute(%ref = %src : (!pcf.sref<?x?xi32>)
+///                         -> (memref<?x?xi32, strided<[?, 1]>, 3>) {
+///     // some use %ref : !pcf.sref<?x?xi32, #pcf.test_scope>
+///   }
+/// ```
+///
+/// Becomes:
+///
+/// ```
+///   %0 = pcf.generic execute {
+///     // some use %src : memref<?x?xi32, strided<[?, 1]>, 3>
+///   }
+/// ```
+///
+/// Additionally the initializer region is inlined into the main body of the
+/// pcf.generic as a part of this pattern as all allocations are resolved as a
+/// part of this pass.
+/// TODO(qedawkins): separate initializer handling out of this pass.
 struct ConvertGenericOp final : OpConversionPattern<PCF::GenericOp> {
   using Base::Base;
 
@@ -643,6 +685,24 @@ struct ConvertGenericOp final : OpConversionPattern<PCF::GenericOp> {
   }
 };
 
+/// Convert all result tied values of a `pcf.loop`, e.g.
+///
+/// ```
+///   %src = ... : memref<?x?xi32, strided<[?, 1]>, 3>
+///   %0 = pcf.loop
+///     execute(%ref = %src : (!pcf.sref<?x?xi32>)
+///                         -> (memref<?x?xi32, strided<[?, 1]>, 3>) {
+///     // some use %ref : !pcf.sref<?x?xi32, #pcf.test_scope>
+///   }
+/// ```
+///
+/// Becomes:
+///
+/// ```
+///   %0 = pcf.loop execute {
+///     // some use %src : memref<?x?xi32, strided<[?, 1]>, 3>
+///   }
+/// ```
 struct ConvertLoopOp final : OpConversionPattern<PCF::LoopOp> {
   using Base::Base;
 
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/convert_sref_to_memref.mlir
```diff
@@ -0,0 +1,464 @@
+// RUN: iree-opt %s --pass-pipeline="builtin.module(iree-pcf-convert-sref-to-memref)" --split-input-file --verify-diagnostics | FileCheck %s
```
kuhar: I don't see any diagnostics in this file? Do we need the flag?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/convert_sref_to_memref.mlir
```diff
@@ -0,0 +1,464 @@
...
+
+// -----
+
+util.func private @convert_memref_write_slice(%arg0: memref<?x?xi32>) {
+  pcf.generic scope(#pcf.test_scope)
+    execute(%ref = %arg0)[%id: index, %n: index]
+         : (!pcf.sref<?x?xi32, #pcf.test_scope>)
+        -> (memref<?x?xi32>) {
+    %src = memref.alloc() : memref<3x4xi32>
+    pcf.write_slice %src into %ref[1, 2] [3, 4] [1, 1] : memref<3x4xi32> into !pcf.sref<?x?xi32, #pcf.test_scope>
```
kuhar: Do we care that this is copying uninitialized memory? Is it possible it will get folded by something?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/convert_sref_to_memref.mlir
```diff
@@ -0,0 +1,464 @@
+// RUN: iree-opt %s --pass-pipeline="builtin.module(iree-pcf-convert-sref-to-memref)" --split-input-file --verify-diagnostics | FileCheck %s
+
+util.func private @convert_generic_with_init(%arg0: memref<?x?xi32, strided<[?, 1]>, 3>) {
```
kuhar, line 3: nit: do we care if these function are private or public? I think the default is `public` like with `func.func`?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/convert_sref_to_memref.mlir
```diff
@@ -0,0 +1,464 @@
...
+//   CHECK-DAG:     %[[CAST0:.+]] = memref.cast %[[ARG0]] : memref<?x?xi32, strided<[3, 1]>, 3> to memref<?x?xi32, strided<[?, 1]>, 3>
+//   CHECK-DAG:     %[[CAST1:.+]] = memref.cast %[[ARG1]] : memref<?x?xi32, strided<[4, 1]>, 3> to memref<?x?xi32, strided<[?, 1]>, 3>
+//  CHECK-NEXT:     scf.for {{.*}} iter_args(%[[ITER0:.+]] = %[[CAST0]], %[[ITER1:.+]] = %[[CAST1]])
+//  CHECK-NEXT:       %[[B:.+]]:2 = util.optimization_barrier %[[ITER1]], %[[ITER0]]
+//  CHECK-NEXT:       scf.yield %[[B]]#0, %[[B]]#1
+//       CHECK:     pcf.return
+
+// -----
+
+func.func @convert_alloc(%d0: index) -> !pcf.sref<?x5xi32, #pcf.sequential> {
```
kuhar, line 282: Why not `util.func` like elsewhere?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/convert_sref_to_memref.mlir
```diff
@@ -0,0 +1,464 @@
...
+//  CHECK-NEXT:     execute[{{.*}}] {
+//   CHECK-DAG:     %[[CAST:.+]] = memref.memory_space_cast %[[ARG0]] : memref<?x?xi32, strided<[?, 1]>, 3> to memref<?x?xi32, strided<[?, 1]>>
+//   CHECK-DAG:     %[[CAST2:.+]] = memref.cast %[[CAST]] : memref<?x?xi32, strided<[?, 1]>> to memref<?x?xi32, strided<[?, ?], offset: ?>>
+//   CHECK-DAG:     %[[SV:.+]] = memref.subview %[[CAST2]][0, 1] [%[[S0]], %[[S1]]] [1, 1] : memref<?x?xi32, strided<[?, ?], offset: ?>> to memref<?x?xi32, strided<[?, ?], offset: ?>>
+//       CHECK:     util.optimization_barrier %[[SV]]
+//       CHECK:     pcf.return
+
+// -----
+
+util.func private @convert_get_memref_no_space(%arg0: memref<?x?xi32, strided<[?, 1]>>, %s0: index, %s1: index) {
```
kuhar: nit: Maybe put it above the previous test case since this is the less specialized case?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
+// Copyright 2025 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <codecvt>
+#include <sstream>
```
kuhar: Why do we need these? Can you trim down the includes?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+//===----------------------------------------------------------------------===//
+
+static bool isDefaultOrStrided(Attribute layout) {
+  auto mapAttr = dyn_cast<AffineMapAttr>(layout);
+  return !layout || (mapAttr && mapAttr.isIdentity()) ||
+         isa<StridedLayoutAttr>(layout);
+}
+
+// State that tracks floating point ranges and flags.
+struct StridedLayoutState : public DFX::AbstractState {
```
kuhar:
```suggestion
struct StridedLayoutState final : DFX::AbstractState {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  }
+}
+
+ChangeStatus StridedLayoutValueElement::updateValue(Value value,
+                                                    DFX::Solver &solver) {
+  StridedLayoutState newState = getState();
+
+  if (auto result = llvm::dyn_cast<OpResult>(value)) {
+    llvm::TypeSwitch<Operation *, void>(result.getOwner())
+        .Case<PCF::AllocOp>([&](PCF::AllocOp allocOp) {
```
kuhar:
```suggestion
        .Case([&](PCF::AllocOp allocOp) {
```
also below

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+        .Case<Util::OptimizationBarrierOp>(
+            [&](Util::OptimizationBarrierOp barrierOp) {
+              auto returnState =
+                  solver.getElementFor<StridedLayoutValueElement>(
+                      *this,
+                      Position::forValue(
+                          barrierOp.getOperand(result.getResultNumber())),
+                      DFX::Resolution::REQUIRED);
+              newState ^= returnState;
+            });
```
kuhar: Can we return early here?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+          });
+    }
+  }
+
+  return DFX::clampStateAndIndicateChange(getState(), newState);
+}
+
+const std::string
+StridedLayoutValueElement::getAsStr(AsmState &asmState) const {
+  auto range = getAssumed();
```
kuhar: nit: spell this out

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+    explorer.initialize();
+  }
+
+  AsmState &getAsmState() { return solver.getAsmState(); }
+  Explorer &getExplorer() { return explorer; }
+
+  LogicalResult run() {
+    // Initialize all shaped ref values to maximally dynamic layouts.
+    explorer.walkValues([&](Value v) {
+      if (isa<PCF::ShapedRefType>(v.getType())) {
```
kuhar: `walkValuesOfType<PCF::ShapedRefType>`?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+        solver.getOrCreateElementFor<StridedLayoutValueElement>(
+            Position::forValue(v));
+      }
+      return WalkResult::advance();
+    });
+
+    // Run solver to completion.
+    auto result = solver.run();
```
kuhar: spell this out

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+    // Since the function op conversion pattern calls this function we're
+    // guaranteed the argument/result types will be unconverted when this is
+    // first called irrespective of pattern application order.
+    SmallVector<Type> argumentTypes;
+    Region *region = callableOp.getCallableRegion();
+    if (!region) {
+      return failure();
+    }
+    for (auto bbArg : callableOp.getCallableRegion()->getArguments()) {
```
kuhar: spell this out

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+static void castIfMismatched(OpBuilder &b, Location loc,
+                             MutableArrayRef<Value> vals,
+                             TypeRange targetTypes) {
+  for (auto [v, t] : llvm::zip_equal(vals, targetTypes)) {
+    v = castIfMismatched(b, loc, v, t);
+  }
+}
+
+struct ConvertGenericOp : public OpConversionPattern<PCF::GenericOp> {
```
kuhar:
```suggestion
struct ConvertGenericOp final : OpConversionPattern<PCF::GenericOp> {
```
also elsewhere below

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+    }
+    newArgs.append(replacements);
+    newArgs.append(newGenericOp.getRegion().getArguments().begin(),
+                   newGenericOp.getRegion().getArguments().end());
+    // Inline the entry block into the new region.
+    Block *entryBlock = &genericOp.getRegion().front();
+    rewriter.inlineBlockBefore(entryBlock, newEntry, newEntry->end(), newArgs);
+
+    // Move the remaining blocks into the new region.
+    for (auto &block : genericOp.getRegion()) {
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+      auto initProducedVals =
+          cast<PCF::YieldOp>(initializerBlock.getTerminator()).getOperands();
+      newArgs.append(initProducedVals.begin(), initProducedVals.end());
+      rewriter.eraseOp(initializerBlock.getTerminator());
+      rewriter.inlineBlockBefore(&initializerBlock, newEntry,
+                                 newEntry->begin());
+    }
+    newArgs.append(replacements);
+    newArgs.append(newGenericOp.getRegion().getArguments().begin(),
+                   newGenericOp.getRegion().getArguments().end());
```
kuhar: llvm::append_range

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+};
+
+struct ConvertLoopOp : public OpConversionPattern<PCF::LoopOp> {
+  using Base::Base;
+
+  LogicalResult
+  matchAndRewrite(PCF::LoopOp loopOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (llvm::any_of(loopOp.getResultTypes(),
+                     [](Type t) { return !isa<MemRefType>(t); })) {
```
kuhar:
```suggestion
    if (!llvm::all_of(loopOp.getResultTypes(), llvm::IsaPred<MemRefType>)) {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+    SmallVector<Value> replacements;
+
+    // Init iterator.
+    auto currInit = loopOp.getInits().begin();
+    ValueRange dynamicSizes = loopOp.getDynamicSizes();
+    for (auto [resultType, isTied] :
+         llvm::zip_equal(loopOp.getResultTypes(), loopOp.getIsTied())) {
+      if (isTied) {
+        replacements.push_back(*currInit);
+        ++currInit;
```
kuhar, line 729: nit: `continue` early

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+  LogicalResult
+  matchAndRewrite(PCF::WriteSliceOp writeOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Value destSlice = memref::SubViewOp::create(
+        rewriter, writeOp.getLoc(), adaptor.getDest(),
+        writeOp.getMixedOffsets(), writeOp.getMixedSizes(),
+        writeOp.getMixedStrides());
+    return llvm::TypeSwitch<Type, LogicalResult>(writeOp.getSourceType())
+        .Case<RankedTensorType>([&](RankedTensorType tensor) {
```
kuhar:
```suggestion
        .Case([&](RankedTensorType tensor) {
```
Also below

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+  LogicalResult
+  matchAndRewrite(PCF::ReadSliceOp readOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Value sourceSlice = memref::SubViewOp::create(
+        rewriter, readOp.getLoc(), adaptor.getSource(),
+        readOp.getMixedOffsets(), readOp.getMixedSizes(),
+        readOp.getMixedStrides());
+    return llvm::TypeSwitch<Type, LogicalResult>(readOp.getResultType())
+        .Case<RankedTensorType>([&](RankedTensorType tensor) {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+};
+
+struct ConvertAllocOp : public OpConversionPattern<PCF::AllocOp> {
+  using Base::Base;
+
+  LogicalResult
+  matchAndRewrite(PCF::AllocOp allocOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto allocType = cast_if_present<MemRefType>(
+        getTypeConverter()->convertType(allocOp.getResult()));
```
kuhar:
```suggestion
    auto allocType =
        getTypeConverter()->convertType<MemRefType>(allocOp.getResult());
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+    FunctionType newFunctionType = maybeNewFunctionType.value();
+
+    // Convert the original function types.
+    TypeConverter::SignatureConversion result(newFunctionType.getNumInputs());
+    for (auto [i, t] : llvm::enumerate(newFunctionType.getInputs())) {
+      result.addInputs(i, t);
+    }
+    if (failed(rewriter.convertRegionTypes(&funcOp.getFunctionBody(),
+                                           *getTypeConverter(), &result)))
+      return failure();
```
kuhar: nit: missing braces

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  auto isLegallyTypedOp = [&](Operation *op) -> bool {
+    for (Type type : op->getResultTypes()) {
+      if (isIllegalType(type))
+        return false;
+    }
+    for (Type type : op->getOperandTypes()) {
+      if (isIllegalType(type))
+        return false;
+    }
+    for (auto &region : op->getRegions()) {
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+      for (auto type : region.getArgumentTypes()) {
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+        if (isIllegalType(type))
+          return false;
+      }
+    }
+    if (auto funcInterface = dyn_cast<FunctionOpInterface>(op)) {
+      if (llvm::any_of(funcInterface.getArgumentTypes(),
+                       [&](Type t) { return isIllegalType(t); })) {
```
kuhar:
```suggestion
      if (llvm::any_of(funcInterface.getArgumentTypes(), isIllegalType)) {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+        return Type();
+      }
+      return maybeConvertedType.value();
+    }
+    // Passthrough for everything else.
+    return v.getType();
+  });
+
+  ConversionTarget target(*context);
+  auto isIllegalType = [&](Type t) { return isa<PCF::ShapedRefType>(t); };
```
kuhar:
```suggestion
  static constexpr auto isIllegalType = llvm::IsaPred<PCF::ShapedRefType>;
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+          return false;
+      }
+    }
+    if (auto funcInterface = dyn_cast<FunctionOpInterface>(op)) {
+      if (llvm::any_of(funcInterface.getArgumentTypes(),
+                       [&](Type t) { return isIllegalType(t); })) {
+        return false;
+      }
+      if (llvm::any_of(funcInterface.getResultTypes(),
+                       [&](Type t) { return isIllegalType(t); })) {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+        return false;
+      }
+    }
+    return true;
+  };
+  target.markUnknownOpDynamicallyLegal(isLegallyTypedOp);
+  ConversionConfig config;
+  config.allowPatternRollback = false;
+
+  patterns.insert<ConvertGenericOp, ConvertLoopOp, ConvertWriteSliceOp,
```
kuhar:
```suggestion
  patterns.add<ConvertGenericOp, ConvertLoopOp, ConvertWriteSliceOp,
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1128 @@
...
+                  ConvertReadSliceOp, ConvertGetMemrefOp, ConvertAllocOp,
+                  ConvertOptimizationBarrier>(typeConverter, context);
+
+  // Function related conversion patterns need the analysis to lookup function
+  // type conversions.
+  patterns.insert<ConvertFuncOp, ConvertReturnOp, ConvertCallOp>(
```
kuhar:
```suggestion
  patterns.add<ConvertFuncOp, ConvertReturnOp, ConvertCallOp>(
```

kuhar (commented):
Looks good overall, just some minor issues

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/convert_sref_to_memref.mlir
qedawkins: `util.func` requires specifying the function visibility, no default. I can change half the tests to func.func for better mixed coverage, but I mainly wanted to make sure multiple function types were supported.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/convert_sref_to_memref.mlir
qedawkins: Answered above, changed half so this makes more sense.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: AbstractStates are inherited whenever creating a state of a different kind of operation, so leaving this as is.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: It would just mean doing `clampStateAndIndicateChange` in two places, not sure this really looks better to me.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: There were a couple that needed to stay public, otherwise updated.

kuhar: public is the default with structs, it's entirely redundant (ack RE final)

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+//===----------------------------------------------------------------------===//
+
+static bool isDefaultOrStrided(Attribute layout) {
+  auto mapAttr = dyn_cast<AffineMapAttr>(layout);
+  return !layout || (mapAttr && mapAttr.isIdentity()) ||
+         isa<StridedLayoutAttr>(layout);
+}
+
+// State that tracks floating point ranges and flags.
+struct StridedLayoutState : public DFX::AbstractState {
```
kuhar:
```suggestion
struct StridedLayoutState : DFX::AbstractState {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+    indicatePessimisticFixpoint();
+    return;
+  }
+}
+
+ChangeStatus StridedLayoutValueElement::updateValue(Value value,
+                                                    DFX::Solver &solver) {
+  StridedLayoutState newState = getState();
+
+  if (auto result = llvm::dyn_cast<OpResult>(value)) {
```
kuhar:
```suggestion
  if (auto result = dyn_cast<OpResult>(value)) {
```
nit: we don't need these `llvm::` prefixes on casts / isa. also elsewhere

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+                  barrierOp.getOperand(result.getResultNumber())),
+              DFX::Resolution::REQUIRED);
+          newState ^= returnState;
+        });
+  } else if (auto bbArg = llvm::dyn_cast<BlockArgument>(value)) {
+    bool didUpdate = false;
+    if (bbArg.getParentBlock()->isEntryBlock()) {
+      didUpdate =
+          llvm::TypeSwitch<Operation *, bool>(bbArg.getOwner()->getParentOp())
+              .Case<PCF::GenericOp>([&](PCF::GenericOp genericOp) {
```
kuhar:
```suggestion
              .Case([&](PCF::GenericOp genericOp) {
```
also below

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+}
+
+struct ConvertGenericOp final : OpConversionPattern<PCF::GenericOp> {
+  using Base::Base;
+
+  LogicalResult
+  matchAndRewrite(PCF::GenericOp genericOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (llvm::any_of(genericOp.getResultTypes(),
+                     [](Type t) { return !isa<MemRefType>(t); })) {
```
kuhar:
```suggestion
    if (!llvm::all_of(genericOp.getResultTypes(), llvm::IsaPred<MemRefType>)) {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+  }
+};
+
+//===----------------------------------------------------------------------===//
+// Control Flow Conversion Patterns
+//===----------------------------------------------------------------------===//
+
+/// Converts the operand and result types of the CallOp, used together with the
+/// FuncOpSignatureConversion.
+struct ConvertFuncOp final : public OpConversionPattern<func::FuncOp> {
```
kuhar:
```suggestion
struct ConvertFuncOp final : OpConversionPattern<func::FuncOp> {
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+      rewriter.inlineRegionBefore(op.getRegion(i), dstRegion, dstRegion.end());
+    }
+
+    rewriter.replaceOp(op, newOp);
+    return success();
+  }
+};
+
+void ConvertSRefToMemRefPass::runOnOperation() {
+  auto *context = &getContext();
```
kuhar:
```suggestion
  MLIRContext *context = &getContext();
```

kuhar (approved):

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+
+      newLayout = StridedLayoutAttr::get(thisStridedLayout.getContext(),
+                                         newOffset, newStrides);
+    }
+    assumed = MemRefType::get(assumed.getShape(), assumed.getElementType(),
+                              newLayout, assumed.getMemorySpace());
+  }
+
+private:
+  MemRefType assumed = MemRefType();
```
MaheshRavishankar, line 213: I think the states need a bit more explanation. This is a fixed point, and these effectively define the lattice. So the lattice needs to be explained a bit for people to understand whats happening.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+
+void StridedLayoutValueElement::initializeValue(Value value,
+                                                DFX::Solver &solver) {
+  if (!isa<PCF::ShapedRefType>(value.getType())) {
+    indicatePessimisticFixpoint();
+    return;
+  }
+}
+
+ChangeStatus StridedLayoutValueElement::updateValue(Value value,
```
MaheshRavishankar: This method is too big and hard to reason. Id recommend following LLVM coding practices for early returns. So split this into two helper functions that handle OpResults and BlockArguments. Within those my personal style would be to have a separate method for each `Case` but I am not going to insist on that. I prefer that cause they you just can do
```
.Case<PCF::AllocOp, RegionBranchOpInterface... >([&](auto op)) { return updateOp(op, solver);}
```

and have methods for
```
ChangeStatus ...::updateOp(PCF::AllocOp allocOp, ...) {...}
ChangeStatus ...::updateOp(RegionBranchOpInterface branchOp, ...) {...}
```

and so on... This mostly a suggestion though... Mainly I think splitting OpResult and BlockArgument handling is useful.

MaheshRavishankar (commented):
Could you give a rationale as to why you are adding DFX based solver to do this. I see your tests and some of the tests have `scf.for` with multiple `iter_args` and `yields` but the values are not used, and I am not sure that this is a case we want to support. I dont think we want to pass around `memref`s like this into loops. It seems like a complex loop structure that only makes the analysis complicated (requiring DFX solver like you have here). The advantage of using some structured control flow (not to say scf.for/scf.forall or any current scf ops, but the concept of structured control flow) is that the op carries information representationally to avoid having to do such analysis in most cases. So I am trying to understand what path makes us need such handling of scf.for with memref args etc. Just the concept of merging layouts by making it dynamic makes me scared cause (a) you are losing information and (b) you might potentially be dealing wth the same object refering to multiple memrefs with different layouts. 

If we didnt have such complex control flow/memref passing through iter_args, a simple walk to find the layout should have sufficed right? So I am trying to connect what code generation path gets us to the kind of control flow that you are trying to use DFX to find the layout of a memref.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+        writeOp.getMixedOffsets(), writeOp.getMixedSizes(),
+        writeOp.getMixedStrides());
+    return llvm::TypeSwitch<Type, LogicalResult>(writeOp.getSourceType())
+        .Case([&](RankedTensorType tensor) {
+          rewriter.replaceOpWithNewOp<IREE::Codegen::StoreToBufferOp>(
+              writeOp, writeOp.getSource(), destSlice);
+          return success();
+        })
+        .Case([&](MemRefType memref) {
+          rewriter.replaceOpWithNewOp<memref::CopyOp>(
```
MaheshRavishankar, line 781: This might be a dumb question, but is the assumption that if the source and the destination are the same the copy gets eliminated?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1118 @@
...
+
+static void castIfMismatched(OpBuilder &b, Location loc,
+                             MutableArrayRef<Value> vals,
+                             TypeRange targetTypes) {
+  for (auto [v, t] : llvm::zip_equal(vals, targetTypes)) {
+    v = castIfMismatched(b, loc, v, t);
+  }
+}
+
+struct ConvertGenericOp final : OpConversionPattern<PCF::GenericOp> {
```
MaheshRavishankar, line 619: For all these patterns it would be useful to have some comments about what the transformation is doing here.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: I'd expect so yes, through standard copy folders.

qedawkins:
> Could you give a rationale as to why you are adding DFX based solver to do this. I see your tests and some of the tests have `scf.for` with multiple `iter_args` and `yields` but the values are not used, and I am not sure that this is a case we want to support. I dont think we want to pass around `memref`s like this into loops. It seems like a complex loop structure that only makes the analysis complicated (requiring DFX solver like you have here). The advantage of using some structured control flow (not to say scf.for/scf.forall or any current scf ops, but the concept of structured control flow) is that the op carries information representationally to avoid having to do such analysis in most cases. So I am trying to understand what path makes us need such handling of scf.for with memref args etc. Just the concept of merging layouts by making it dynamic makes me scared cause (a) you are losing information and (b) you might potentially be dealing wth the same object refering to multiple memrefs with different layouts.
> 
> If we didnt have such complex control flow/memref passing through iter_args, a simple walk to find the layout should have sufficed right? So I am trying to connect what code generation path gets us to the kind of control flow that you are trying to use DFX to find the layout of a memref.

Shortest answer: for completeness

Longer answer:
There is a chicken-and-egg problem with various forms of non-trivial control flow in our code generation, especially when it comes to handling of functions. If we want to be able to support function calls natively, we need the core passes to support it or we will never get started. This was especially apparent with ukernel work. We had to add custom function inlining logic due to our current pipeline structure because we don't support function calls.

w.r.t. the concern about iter_args, that's a semi-contrived example. We *technically* could choose to represent multi-buffering that way though if we wanted to, but that test case is primarily a check for proper analysis of structured ops. The main thing I wanted to make sure we supported was function calls so we can start writing in smaller sub-functions (e.g. a copy helper) that we call into rather than inline/transform everything in place.

> Just the concept of merging layouts by making it dynamic makes me scared cause (a) you are losing information and (b) you might potentially be dealing wth the same object refering to multiple memrefs with different layouts.

Very true, generating code like this isn't a goal. This is just something that this pass has to handle because of the types it's operating on. Note that the bufferization analysis does the exact same thing (because it has to!). You can think of the whole analysis here as *just* the layout portion of the analysis bufferization does.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: I did the OpResult + Block arg split. I can do the per-op split too if wanted, I don't really mind either way.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: Done

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: Added more of a description of each field + how the lattice is traversed.

qedawkins:
Merging and will wait for review comments in post if there are any (discussed offline). All outstanding comments have been addressed (to my knowledge).

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1214 @@
...
+// next to the state fields.
+struct StridedLayoutState : DFX::AbstractState {
+  bool isValidState() const override { return isValid; }
+  bool isAtFixpoint() const override { return !isValid || isFinalized; }
+
+  void invalidate() { isValid = false; }
+
+  ChangeStatus indicateOptimisticFixpoint() override {
+    isFinalized = true;
+    return ChangeStatus::UNCHANGED;
```
MaheshRavishankar, line 98: This confused me a bit. (I have been iterating with Claude for the review, and at the very least, Claude also didnt know for sure, so asking). It was confusing to me that setting something to optimistic fixed point, i.e. its state changed, and this method is returning `ChangeStatus::UNCHANGED`. First thought was shouldnt all the dependents know that this is an optimistic fixed point and update accordingly. The reason why this seems to work out is that the update methods below like `updateOpResult` are just comparing the old state to the new state and the actual return value here seems to be ignored. Could you clarify what is happening here.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
```diff
@@ -0,0 +1,1214 @@
...
+    // If no value is assumed yet, take RHS.
+    if (!assumed) {
+      assumed = rhs.getAssumed();
+      return;
+    }
+
+    if (!rhs.getAssumed()) {
+      // Two possibilities if the rhs is undefined. First, rhs is at a
+      // pessimistic fixed point, in which case we take it.
+      if (rhs.isAtFixpoint()) {
```
MaheshRavishankar, line 137: This is confusing since I am assuming this is the not valid + fixed point state for the RHS. But we have already checked that the RHS is invalid. So whats happening here. Looks like this rubric helps understand this better. Might be worth putting this somewhere.


```
 Pessimistic Fixpoint ≠ Invalid State                                                                                                                                                                                              
  ┌──────────────────────┬─────────┬─────────────┬──────────┬──────────────────────────────────────────────────────────────────┐                                                                                                    
  │        State         │ isValid │ isFinalized │ assumed  │                             Meaning                              │                                                                                                    
  ├──────────────────────┼─────────┼─────────────┼──────────┼──────────────────────────────────────────────────────────────────┤                                                                                                    
  │ Invalid              │ false   │ ?           │ any      │ Constraint violation (memory space mismatch, non-strided layout) │                                                                                                    
  ├──────────────────────┼─────────┼─────────────┼──────────┼──────────────────────────────────────────────────────────────────┤                                                                                                    
  │ Pessimistic Fixpoint │ true    │ true        │ empty    │ Can't determine layout (indirect call, incomplete traversal)     │                                                                                                    
  ├──────────────────────┼─────────┼─────────────┼──────────┼──────────────────────────────────────────────────────────────────┤                                                                                                    
  │ Optimistic Fixpoint  │ true    │ true        │ concrete │ Know exact layout (alloc, tied arg)                              │                                                                                                    
  └──────────────────────┴─────────┴─────────────┴──────────┴──────────────────────────────────────────────────────────────────┘          

  ```

MaheshRavishankar (commented):
Ok, I think I understand now, but just a few clarifying questions about the DFX solver setup.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: Two part answer:
The state setup was copy-pasted from other places and we do the same thing everywhere else and I wasn't paying enough attention to the return value here because it's ignored like you said.

Looking at all other instances of `indicateOptimisticFixpoint` we always seem to return `UNCHANGED` and ignore the return value so I'm unsure why we even return anything. Anyway, TLDR it's a copy-paste oversight, not some hidden reason that it returns unchanged. I can add a clarifying comment with the below.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertSRefToMemRef.cpp
qedawkins: Yeah the rubric is correct, thanks! I can add that.

PR #22920, opened 2025-12-16
[GlobalOpt] Restrict constexpr hoisting for index types

AGindinson:
Avoids crashes for operations like `tensor.dim` / `tensor.from_elements` when hoisting constant expressions. Improves index bitwidth assessment in general utilities for constexpr/hoistability analysis by propagating DataLayout information.

The only `HoistableTypeInterface`'s that are feasible to instantiate are the existing ones, for general tensors and for scalar indices. `tensor<index>` types need careful handling, as default utils for bitwidth assessment do not (and should not) assume the exact size of the index type.

For hoisting transforms that act on `ConstExprAnalsysis`, this blocks index-typed expressions altogether within `ConstExprHoistingPolicy`. This type of a transformation might cause issues if the original operation gets dispatch to a target that indexes in 64-bit, while the initializer runs on a 32-bit host, and the bounds of the hoisted constant don't get checked during materialization in the final stages of compilation.

Signed-off-by: Artem Gindinson <gindinson@roofline.ai>

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.cpp
+++ b/compiler/src/iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.cpp
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "iree/compiler/Dialect/Util/Analysis/Constant/ConstExpr.h"
+#include "iree/compiler/Dialect/Util/IR/UtilTypes.h"
 
 #include "iree/compiler/Dialect/Util/Analysis/Constant/OpOracle.h"
 #include "iree/compiler/Dialect/Util/Analysis/Explorer.h"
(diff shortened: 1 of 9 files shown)


MaheshRavishankar (commented):
It can be tricky handling index bits this way. There could be mismatches if the initializer runs on a machine where index is mapped to 32-bits, and the rest of it maps index to 64-bits (or vice-versa). I dont know off the top of my head what would happen today.

AGindinson:
@MaheshRavishankar Ah, that's a great catch. Intuitively, it seems that my optimization is plain incorrect if the host indexes in 32-bit. I'll have a go at querying TargetInfo then.

AGindinson:
@MaheshRavishankar, sorry for the very long turnaround on this one. Could you please take a look if this direction is OK, per the updated description?
Basically I'm blocking out the transformation for index types within `GlobalOpt`, but not touching the current hoisting behavior within ConstEval/JIT (compiler/src/iree/compiler/ConstEval/test/jit_globals.mlir).

What I'm very uncertain about: in the current form of the PR, I'm passing around `DataLayout` instances to try and un-hardcode the existing 64-bit assumptions for size estimation. However, my understanding is that DLTI dialect entries are actually not commonly encountered within the IREE flow, and all the extra queries in my change will evaluate to the default 64-bits anyway, also for GPU targets where bitwidths are hard-coded on the codegen level (LLVMGPU = 64, SPIRV = 32)?
So I'm unsure whether to leave the PR in its current shape or to revert to 64-bit assumptions everywhere, just retaining the checks and hoisting blockers for index constexprs.

Would appreciate your advice! Thanks.

MaheshRavishankar:
Hey @AGindinson  I am a bit behind on reviews, but I have this in my queue. Can I get back to yo on this next week?

AGindinson:
@MaheshRavishankar Absolutely, no rush at all – I took a while myself even to start the rework. Many thanks!

PR #22922, opened 2025-12-16
[Stream] Prioritize NamedParameterAttr in UnifyEncodingForGlobals source tracing. 4/n

hanhanW:
Extend the pass to handle cases where tensor constants are created inline via stream.tensor.constant with a NamedParameterAttr, without being stored to a source global first. The parameter name represents the actual data identity, so it should be used as the unification key.

Introduces SourceTraceResult to track either a source global or parameter attribute, with getSourceKey() prioritizing parameter names over global names.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -83,7 +83,7 @@ struct SourceTraceResult {
   std::string getSourceKey() {
     if (parameterAttr) {
       std::string key;
-      if (auto scope = parameterAttr.getScope()) {
+      if (StringAttr scope = parameterAttr.getScope()) {
         key = scope.str() + "::";
       }
       key += parameterAttr.getKey().str();
@@ -135,13 +135,13 @@ public:
   // Returns all source keys that have multiple distinct encodings.
   SmallVector<StringRef> getSourcesWithMultipleEncodings() const {
     SmallVector<StringRef> result;
-    for (const auto &entry : sourceGlobals) {
+    for (const auto &[name, info] : sourceGlobals) {
       llvm::SmallDenseSet<Attribute, 4> uniqueEncodings;
-      for (const EncodedGlobalInfo &encoded : entry.second.encodedVersions) {
+      for (const EncodedGlobalInfo &encoded : info.encodedVersions) {
         uniqueEncodings.insert(encoded.encodingAttr);
       }
       if (uniqueEncodings.size() > 1) {
-        result.push_back(entry.first());
+        result.push_back(name);
       }
     }
     return result;
@@ -317,7 +317,7 @@ bool GlobalEncodingAnalyzer::hasRecognizedEncoding(Type type) const {
 LogicalResult GlobalEncodingAnalyzer::collectGlobalEncodings() {
   LDBG() << "--- collectGlobalEncodings ---";
   // Walk all initializers to find encoding patterns:
-  //   %source = util.global.load @source_global  (or stream.tensor.constant)
+  //   %source = util.global.load @source_global (or stream.tensor.constant)
   //   %encoded = stream.tensor.encode %source ...
   //   util.global.store %encoded, @encoded_global
   for (auto initOp : moduleOp.getOps<IREE::Util::InitializerOp>()) {
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -65,9 +65,40 @@ struct EncodedGlobalInfo {
   IREE::Stream::TensorEncodeOp encodeOp;
 };
 
-// Information about a source global and all its encoded versions.
-struct SourceGlobalInfo {
+// Result of tracing from an encode op's source back to its origin.
+struct SourceTraceResult {
+  // The source global, if found (may be null if the source is from
+  // parameterAttr). If there are multiple source globals chain, it is the
+  // topmost one.
   IREE::Util::GlobalOpInterface sourceGlobal;
+  // The named parameter attribute, if the source is a TensorConstantOp with
+  // a NamedParameterAttr value.
+  IREE::Stream::NamedParameterAttr parameterAttr;
+
+  // Returns true if we found a valid source (either global or parameter).
+  explicit operator bool() const { return sourceGlobal || parameterAttr; }
+
+  // Returns a unique key identifying this source. Prioritizes parameter name
+  // over global name since parameters represent the actual data identity.
+  std::string getSourceKey() {
+    if (parameterAttr) {
+      std::string key;
+      if (auto scope = parameterAttr.getScope()) {
+        key = scope.str() + "::";
+      }
+      key += parameterAttr.getKey().str();
+      return key;
+    }
+    if (sourceGlobal) {
+      return sourceGlobal.getGlobalName().str();
+    }
+    return {};
+  }
+};
+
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -681,3 +681,43 @@ module @same_encoding_twice_no_unification {
...
+
+//       CHECK:   util.global private @[[$DEVICE_A:.+]] =
+util.global private @device_a = #device_target_local
+util.global private @encoded_v1 : !stream.resource<constant>
+util.global private @encoded_v2 : !stream.resource<constant>
+
+// CHECK: util.initializer
+util.initializer {
+  // The constant is created with parameter attribute, not loaded from a global.
+  // CHECK: %[[CST:.+]] = stream.tensor.constant
```
sommerlukas, line 705: Should we check here that the constant still uses the same name?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -101,31 +132,30 @@ class GlobalEncodingAnalyzer {
   // unified encodings.
   LogicalResult run();
 
-  // Returns all source globals that have multiple distinct encodings.
+  // Returns all source keys that have multiple distinct encodings.
   SmallVector<StringRef> getSourcesWithMultipleEncodings() const {
     SmallVector<StringRef> result;
-    for (const auto &[name, info] : sourceGlobals) {
+    for (const auto &entry : sourceGlobals) {
```
sommerlukas: Why can't we use the structured binding here anymore?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -101,31 +132,30 @@ class GlobalEncodingAnalyzer {
...
       llvm::SmallDenseSet<Attribute, 4> uniqueEncodings;
-      for (const EncodedGlobalInfo &encoded : info.encodedVersions) {
+      for (const EncodedGlobalInfo &encoded : entry.second.encodedVersions) {
         uniqueEncodings.insert(encoded.encodingAttr);
       }
       if (uniqueEncodings.size() > 1) {
-        result.push_back(name);
+        result.push_back(entry.first());
```
sommerlukas: If we can't use structured bindings (see above), can we use `.first` without `()` here, to keep it consistent with `.second`?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -288,7 +317,7 @@ bool GlobalEncodingAnalyzer::hasRecognizedEncoding(Type type) const {
 LogicalResult GlobalEncodingAnalyzer::collectGlobalEncodings() {
   LDBG() << "--- collectGlobalEncodings ---";
   // Walk all initializers to find encoding patterns:
-  //   %source = util.global.load @source_global
+  //   %source = util.global.load @source_global  (or stream.tensor.constant)
```
sommerlukas: Nit:
```suggestion
  //   %source = util.global.load @source_global (or stream.tensor.constant)
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -371,9 +401,9 @@ GlobalEncodingAnalyzer::traceToSourceGlobal(Value value) {
                 LDBG() << "      Bail: mutable source global";
                 return false;
               }
-              assert(!foundSourceGlobal &&
-                     "should only find one source global");
-              foundSourceGlobal = globalOp;
+              // It is okay to trace through globalOp because data is unchanged.
+              // I.e., there are no tensor dispatch ops in the chain.
```
sommerlukas, line 405: Do we need a test for this behavior that is now allowed?

sommerlukas (commented):
Mostly looks good, just a few small comments.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -65,9 +65,40 @@ struct EncodedGlobalInfo {
...
+
+  // Returns true if we found a valid source (either global or parameter).
+  explicit operator bool() const { return sourceGlobal || parameterAttr; }
+
+  // Returns a unique key identifying this source. Prioritizes parameter name
+  // over global name since parameters represent the actual data identity.
+  std::string getSourceKey() {
+    if (parameterAttr) {
+      std::string key;
+      if (auto scope = parameterAttr.getScope()) {
```
jtuyls: Nit: use explicit type.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
jtuyls: +1 And I think that could also check the prioritization of the NamedParameterAttr?

jtuyls (approved):
LGTM, just a few small comments.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: I should point it out in the first place. We already have such test case, and it was added in previous patches. :) E.g., `tensor.constant -> store_to_global -> load_from_global -> encode_data`.

https://github.com/iree-org/iree/blob/dd8b207711da856ca32173c345cf48d31ae54829/compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir#L115-L129

It was not an issue because we only trace until the first immutable global.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: The answer is no to me. It is clear that the pass does not update parameters.

And here is the guide: https://mlir.llvm.org/getting_started/TestingGuide/#filecheck-best-practices

> Tests should be minimal, and only check what is absolutely necessary.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Thanks, I dont remember what the old issue was, and I switched it back.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
sommerlukas: So, what you're basically saying is that the second test in this file already covers the new scenario where the `NamedParameter` gets prioritized?

sommerlukas (approved):
Just one question regarding testing.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Yes, this is what I'm saying, but I think it is not obvious. Let me add a global in the between for the new test, so people won't be confused. I also added a note saying that this is not common if a global cleanup is performed.

Issue #21659, opened 2025-08-12
[DT] Encode Weights in Initializers w/o Increased Memory Footprint


jtuyls:
With encoded weights, we see 2-3x increased memory footprint from initialization. This is because the same weights are used multiple times for prefill and decode and stored in separate globals. We should try to improve this to 1x. This requires that all encodings result in the same layout and avoid multiple globals.

**Immediate action items:**
1. Generate a simpler repro from llama model, mainly for compile-time issue.
2. Make the toy repro result in the same layout.

**Upcoming tasks:**
1. If we can't resolve them in the same layout during compile-time, we'll investigate and scope the work in goal (2).

cc @hanhanW @Max191

jtuyls:
For 2), here is a toy example: https://gist.github.com/jtuyls/fab1b5c9b6d2744df54968c40eb9abef

NOTE: To get the same layout, this needs `iterationSizes` to not be passed during encoding setting here: https://github.com/iree-org/iree/blob/6ef2c5e08fb1b56f19ccbf10627f35e253c81e65/compiler/src/iree/compiler/DispatchCreation/SetEncoding.cpp#L240

When compiling, this results in the following initializer (after iree-stream-materialize-encodings):
```
util.initializer {
    %c16777216 = arith.constant 16777216 : index
    %cst = stream.async.constant on(#hal.device.affinity<@__device_0>) : !stream.resource<constant>{%c16777216} = #stream.parameter.named<"model"::"blk.0.attn_q.weight:qs"> : tensor<4096x4096xf8E4M3FNUZ>
    util.global.store %cst, @"__auto.blk.0.attn_q.weight:qs" : !stream.resource<constant>
    util.global.store %c16777216, @"__auto.blk.0.attn_q.weight:qs__size" : index
    %__auto.blk.0.attn_q.weight3Aqs__size = util.global.load @"__auto.blk.0.attn_q.weight:qs__size" : index
    %__auto.blk.0.attn_q.weight3Aqs = util.global.load @"__auto.blk.0.attn_q.weight:qs" : !stream.resource<constant>
    %0 = stream.async.transfer %__auto.blk.0.attn_q.weight3Aqs : !stream.resource<constant>{%__auto.blk.0.attn_q.weight3Aqs__size} from(#hal.device.affinity<@__device_0>) -> to(#hal.device.affinity<@__device_0>) !stream.resource<*>{%__auto.blk.0.attn_q.weight3Aqs__size}
[...]

PR #22923, opened 2025-12-16
[GPU] Implement dummy getUnifiedEncoding method for GPU encoding resolver.

hanhanW:
It adds a test which is a simplified version from https://github.com/iree-org/iree/issues/21659

It also adds the pass to the pipeline. The pass is currently guarded under an experimental flag.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
+++ b/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
@@ -629,6 +629,16 @@ struct GPULayoutResolverAttr final
     MLIRContext *ctx = attr.getContext();
     return GPUEncodingResolverAttr::get(ctx, getPackedLayoutImpl(attr, type));
   }
+
+  // TODO: Improve the heuristic. This is a placeholder that demonstrates the
+  // encoding unification is working in practice.
+  Attribute getUnifiedEncoding(Attribute attr,
+                               ArrayRef<Attribute> encodings) const {
+    if (encodings.empty()) {
+      return nullptr;
+    }
+    return encodings[0];
+  }
 };
 
 struct GPUEncodingResolverVerifier
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
@@ -146,14 +146,9 @@ void buildStreamTensorPassPipeline(OpPassManager &passManager,
   }
 
   // Unify encodings for globals, which ensures that we don't increase memory
-  // footprint significantly, unless the program explicitly requires it. Then it
-  // folds the duplicate globals away.
+  // footprint significantly, unless the program explicitly requires it.
   if (clUnifyEncodingForGlobals) {
     passManager.addPass(IREE::Stream::createUnifyEncodingForGlobalsPass());
-    OpPassManager ipoPipeline(mlir::ModuleOp::getOperationName());
-    buildStreamCleanupPassPipeline(ipoPipeline, transformOptions);
-    passManager.addPass(
-        IREE::Util::createFixedPointIteratorPass(std::move(ipoPipeline)));
   }
 
   //----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


hanhanW:
I verified that this fixes all the llama memory footprint issue (3x->2x).

405b is a little different, so it took me some time to fix it. I also verified that (this PR + https://github.com/iree-org/iree/pull/22221) fixes the 405b memory footprint issue (3x->2x).

hanhanW:
This technically does not depend on https://github.com/iree-org/iree/pull/22922, but the dep looks better to me because we are adding the pass to the pipeline with this change.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
```diff
@@ -139,6 +145,17 @@ void buildStreamTensorPassPipeline(OpPassManager &passManager,
         IREE::Util::createFixedPointIteratorPass(std::move(ipoPipeline)));
   }
 
+  // Unify encodings for globals, which ensures that we don't increase memory
+  // footprint significantly, unless the program explicitly requires it. Then it
+  // folds the duplicate globals away.
```
sommerlukas: I think we should document the motivation to run a fixed-point iteration with the cleanups here again after just finishing that in the previous step.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
```diff
@@ -629,6 +629,16 @@ struct GPULayoutResolverAttr final
...
   }
+
+  // TODO: Improve the heuristic. This is a placeholder that demonstrates the
+  // encoding unification is working in practice.
+  Attribute getUnifiedEncoding(Attribute attr,
+                               ArrayRef<Attribute> encodings) const {
+    if (encodings.empty()) {
+      return nullptr;
+    }
+    return encodings[0];
```
sommerlukas, line 640: Is the order always consistent?

sommerlukas (commented):
Just two minor comments/questions.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
hanhanW: This is pretty close to the above comment, so I did not add the dup comments. If you read it from earlier snippet, then it is clear, right?

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
hanhanW: It's deterministic, but it may be inconsistent. It depends on the order in hoisting.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
sommerlukas: I do understand why we run it as a fix-point iteration based on the comment above (though something like "we use a fixpoint iteration for the same reason as above" wouldn't hurt to add), but I'm not 100% sure and think that we should document what we want to clean up here. Will we have redundant code? Is there any globals that can be eliminated entirely? IIUC, this pass would only change the encoding but not eliminate the unified globals.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
sommerlukas: If it can be inconsistent, don't we run the danger of having a flaky test? The test assumes that the first encoding is chosen, but if that can change from run to run, the test may fail in some runs.

hanhanW: I meant, it is consistent from run to run. However, the unification could change if someone makes updates in hoisting. It is still deterministic, and it can be caught by integration tests if we're at the position of tracking the performance. I think we're over-indexing on this as everything is under experimental pass. This is not yet ready to use, and I don't expect anyone other than us using the path. This is mostly for demonstrating that we can run the model, and it is a big milestone for the data-tiling work.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
hanhanW: Maybe I was not clear enough, by "it", I meant the whole `if` body. Anyway, after investigating more, I found that the globals are not folded away after the pipeline, but they are folded away in later passes. I think we can start with inserting the pass after we have the canonical globals here, and rely on later passes to deduplicate the globals.

sommerlukas (approved):

PR #22931, opened 2025-12-17
[LinalgExt] Enable tensor map_scatter lowering

### LINKED ISSUE #22697: Support docomposition for `iree_linalg_ext.map_scatter` with tensor output
Currently, we only accept `map_scatter` with buffer semantics to decompose.
https://github.com/sakupan102/iree/blob/8b98b015b5e90fba4df17aa80df050f26481be1e/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp#L438C3-L440

Due to the [upstream change](https://github.com/llvm/llvm-project/commit/7e7ea9c5357efcdf9ba6bd7ea3669e607a9af400), we should accept map_scatter with tensor semantics.

sakupan102:
Previously, vector.scatter didn’t support tensor outputs. Upstream changes added that support, so DecomposeMapScatter now handles the tensor case as well.

Closes https://github.com/iree-org/iree/issues/22697

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
@@ -166,16 +166,13 @@ performIndexAndMaskVectorization(MapScatterOp mapScatterOp,
   rewriter.setInsertionPoint(mapScatterOp);
   Value flatOutput;
   SmallVector<Value> strides;
-  SmallVector<OpFoldResult> outputSizes;
+  SmallVector<OpFoldResult> outputSizes =
+      getDims(rewriter, loc, mapScatterOp.getOutput());
   if (mapScatterOp.hasPureBufferSemantics()) {
-    outputSizes =
-        memref::getMixedSizes(rewriter, loc, mapScatterOp.getOutput());
     flatOutput = createFlatOutputBuffer(rewriter, loc, mapScatterOp.getOutput(),
                                         outputSizes, strides);
   } else {
     // For tensor outputs, create a flat output buffer as an empty tensor.
-    outputSizes =
-        tensor::getMixedSizes(rewriter, loc, mapScatterOp.getOutput());
     auto outputType = cast<TensorType>(mapScatterOp.getOutputType());
     SmallVector<ReassociationIndices> reassociations;
     reassociations.push_back(
@@ -388,9 +385,11 @@ static LogicalResult decomposeToScatter(MapScatterOp mapScatterOp,
   SmallVector<ReassociationIndices> reassociations;
   reassociations.push_back(llvm::to_vector(
       llvm::seq<int64_t>(mapScatterOp.getOutputType().getRank())));
-  auto expandOp =
-      tensor::ExpandShapeOp::create(rewriter, loc, mapScatterOp.getOutputType(),
-                                    scatterOp.getResult(), reassociations);
+  SmallVector<OpFoldResult> outputSizes =
+      tensor::getMixedSizes(rewriter, loc, mapScatterOp.getOutput());
+  auto expandOp = tensor::ExpandShapeOp::create(
+      rewriter, loc, mapScatterOp.getOutputType(), scatterOp.getResult(),
+      reassociations, outputSizes);
   rewriter.replaceOp(mapScatterOp, expandOp.getResult());
   return success();
 }
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
@@ -16,6 +16,9 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/TypeRange.h"
+#include "mlir/IR/Types.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
 #define DEBUG_TYPE "iree-decompose-map-scatter"
@@ -143,7 +146,7 @@ struct VectorizationResult {
   /// The vectorized mask values for conditional stores.
   Value maskVector;
   /// The flattened 1D output buffer.
-  Value flatOutputBuffer;
+  Value flatOutput;
 };
 
 /// Vectorize the index computation and mask evaluation for a `map_scatter` op.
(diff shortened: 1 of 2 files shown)


hanhanW:
Hi, thanks for the patch. @Max191 is currently on vacation, and it's better to have his eyes on this. IMO, this kind of work is better to have a concrete example and see if it truly fixes anything. We'll need to support map_scatter op better on CPU side, where all the pipelines are tensor-based pipeline. Let's see if it will help or not. For now, would you mind to wait until people are back from vacations?

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -161,11 +164,25 @@ performIndexAndMaskVectorization(MapScatterOp mapScatterOp,
   Location loc = mapScatterOp.getLoc();
   OpBuilder::InsertionGuard g(rewriter);
   rewriter.setInsertionPoint(mapScatterOp);
-  SmallVector<OpFoldResult> outputSizes =
-      memref::getMixedSizes(rewriter, loc, mapScatterOp.getOutput());
```
Max191, line 165: For the output sizes, there is a nice utility function to handle both memref and tensor types:

https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h#L49

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -161,11 +164,25 @@ performIndexAndMaskVectorization(MapScatterOp mapScatterOp,
...
+  } else {
+    // For tensor outputs, create a flat output buffer as an empty tensor.
+    outputSizes =
+        tensor::getMixedSizes(rewriter, loc, mapScatterOp.getOutput());
+    auto outputType = cast<TensorType>(mapScatterOp.getOutputType());
+    SmallVector<ReassociationIndices> reassociations;
+    reassociations.push_back(
+        llvm::to_vector(llvm::seq<int64_t>(outputType.getRank())));
+    flatOutput = tensor::CollapseShapeOp::create(
+        rewriter, loc, mapScatterOp.getOutput(), reassociations);
```
Max191, line 181: This is fine for now, since we don't have any concrete use cases yet, but adding collapse_shape ops on tensors can have more implications than it does for memrefs. It may be fine, but bufferization can often get tripped up by reshapes, which could end up creating extra large allocations. This is part of the challenge with vectorizing map_scatter on tensors, and we need to try it e2e to see if it works or not.

That said, I'm not sure what the best way to represent it will be, so we can reevaluate once we are able to test this in an e2e path.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -356,15 +373,30 @@ static LogicalResult decomposeToScatter(MapScatterOp mapScatterOp,
...
+
+  // For tensor outputs, expand the result back to the original shape.
+  auto scatterOp =
+      vector::ScatterOp::create(rewriter, loc, flatOutput.getType(), operands);
+  SmallVector<ReassociationIndices> reassociations;
+  reassociations.push_back(llvm::to_vector(
+      llvm::seq<int64_t>(mapScatterOp.getOutputType().getRank())));
+  auto expandOp =
+      tensor::ExpandShapeOp::create(rewriter, loc, mapScatterOp.getOutputType(),
+                                    scatterOp.getResult(), reassociations);
```
Max191: I believe this will only work for static shapes. To handle dynamic shapes too, you need to also pass a list of OpFoldResult for the resulting output shape, like what is done here:

https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp#L787-L788

You can get the output shape from the original map_scatter output tensor.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_map_scatter.mlir
```diff
@@ -31,6 +31,30 @@ func.func @identity_map_scatter(
 //       CHECK:   %[[EXTRACT_3:.+]] = vector.extract %[[INPUT]][3] : vector<16xf32> from vector<4x16xf32>
 //       CHECK:   vector.store %[[EXTRACT_3]], %[[FLAT_OUTPUT]][%[[EXTRACT_IDX_3]]] : memref<64xf32>, vector<16xf32>
 
+// -----
```
Max191, line 34: Please also add a test for dynamic shapes.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -388,7 +420,8 @@ static LogicalResult decomposeMapScatter(MapScatterOp mapScatterOp,
   const bool isMaskForwardSlice = maskOp && slice.contains(maskOp);
   const bool isUnitFunctionOfInnermostInputIdx =
       isUnitFunctionOf(innermostOutputIdx, innermostInputIdx);
-  if (!isMaskForwardSlice && isUnitFunctionOfInnermostInputIdx) {
+  if (!isMaskForwardSlice && isUnitFunctionOfInnermostInputIdx &&
+      mapScatterOp.hasPureBufferSemantics()) {
```
Max191, line 423: It shouldn't make a difference whether it is memref or tensor here. Can you support this case too (and add tests for it)?

Max191 (commented):
I am with adding this support now, since it is already mostly implemented in this PR. It still won't have any actual use cases yet, but we could try using on CPU and seeing if there are any issues e2e.

I left a few comments on the implementation, but looks good so far!

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
sakupan102: Since `vector.store` only supports memref as an output, we first need to add tensor support for the output.
I think it is better to disable the tensor case for now.

Max191: Sorry, I missed this response earlier. I think you can support the tensor case if you use `vector.transfer_write` instead of `vector.store`. The transfer_write should eventually lower into vector.store after vector lowering.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_map_scatter.mlir
```diff
@@ -33,6 +33,52 @@ func.func @identity_map_scatter(
...
+//   CHECK-DAG:   %[[CST_TRUE:.+]] = arith.constant dense<true> : vector<64xi1>
+//   CHECK-DAG:   %[[FLAT_OUTPUT:.+]] = tensor.collapse_shape %[[OUTPUT]] {{.*}} tensor<4x16xf32> into tensor<64xf32>
+//       CHECK:   %[[SCATTER:.+]] = vector.scatter
+//       CHECK:   %[[RESULT:.+]] = tensor.expand_shape %[[SCATTER]] {{.*}} : tensor<64xf32> into tensor<4x16xf32>
+//       CHECK:   return %[[RESULT]] : tensor<4x16xf32>
+
+
+// -----
+
+func.func @identity_map_scatter_tensor(
```
Max191, line 60: Give this test a different name (like identity_map_scatter_tensor_dynamic).

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
Max191: Can you also support the decomposeToLoadStore case by using vector.transfer_write?

PR #22932, opened 2025-12-17
[LLVMCPU][AArch64] New heuristic for matmul vector tile sizes

momchil-velikov:
Compute vector tile sizes using a heuristic that aims to keep the entire
ACC/OUT tile in registers, leave a few registers for LHS/RHS columns
or rows, and all that while not exceeding the number of available
registers.  The rationale is that a matrix multiplication typically
lowers to a loop nest in which the ACC/OUT tile remains live across all
iterations of the innermost loop, whereas the LHS and RHS operands live
for a single iteration and do not require the entire tiles to be
simultaneously resident in registers.

The base element type used is the element type of the output vector
under the assumption the operand types with smaller bitwidths
will be promoted to the output type and thus require more registers
for the same number of elements.

We have observed improvements in performance for on AArch64 targets
(Neoverse-V1 and Neoverse-V2 cores) for Neon and SVE configurations
without data tiling and peeling vector preprocessing strategy,
for example:
  * Neon, `batch.matmul`, f32: ~46% improvement (less time)
  * Neon, GPT2, f32: ~31% improvement

Signed-off-by: Momchil Velikov <momchil.velikov@arm.com>

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -1559,8 +1559,6 @@ static void getMatmulVectorSizesUsingFillRegisterFileHeuristic(
   }
 
   // Find the output element type of the matmul.
-  assert(op->getResultTypes().size() == 1 &&
-         "Expected single output type for matmul op");
   Type outputEltType = getElementTypeOrSelf(op->getResultTypes()[0]);
   if (!outputEltType.isSignlessIntOrFloat()) {
     return;
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -1549,6 +1549,14 @@ static void getMatmulVectorSizesUsingFillRegisterFileHeuristic(
     SmallVectorImpl<bool> &scalableSizeFlags) {
   assert(sizes.empty() && "Pre-condition enforced by caller");
 
+  // Check we have a matrix multiplication.
+  FailureOr<linalg::ContractionDimensions> cDims =
+      linalg::inferContractionDims(op);
+  if (failed(cDims) || cDims->m.size() != 1 || cDims->n.size() != 1 ||
+      cDims->k.size() != 1) {
+    return;
+  }
+
   // Find the output element type of the matmul.
   assert(op->getResultTypes().size() == 1 &&
          "Expected single output type for matmul op");
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -1528,54 +1528,54 @@ static FailureOr<Type> nonWideningLinalgElementType(linalg::LinalgOp op) {
   return inputAndOutputElementTypes[0];
 }
 
-/// Compute or adjust existing vector sizes using a generic heuristic that will
-/// aim to fill at least one full vector register for all the element types of
-/// the matmul. For now, the current heuristics only look at the N dimension but
-/// we would introduce logic to also consider unrolling trade-offs between the
-/// M, N and K.
-///
-/// Example: for an (i32 <- i8, i8) matmul and a 128-bit vector register, vector
-/// size N would be at least 128/8=16.
-///
-/// NOTE: This function should not contain target-specific conditional code.
-/// TODO: Currently it's only use on Aarch64. We should generalize it to other
-/// targets.
-static void getMatmulVectorSizesUsingFullVectorHeuristics(
+/// Compute vector tile sizes using a heuristic that aims to keep the entire
+/// ACC/OUT tile in registers, leave a few registers for LHS/RHS columns
+/// or rows, and all that while not exceeding the number of available registers.
+/// The rationale is that a matrix multiplication typically lowers to a loop
+/// nest in which the ACC/OUT tile remains live across all iterations of the
+/// innermost loop, whereas the LHS and RHS operands live for a single iteration
+/// and do not require the entire tiles to be simultaneously resident in
+/// registers.
+/// The base element type used is the element type of the output
+/// vector under the assumption the operand types with smaller bitwidths
+/// will be promoted to the output type and thus require more registers for the
+/// same number of elements.
+/// TODO: It might be worth extending the heuristic to consider target
+/// architecture features and operand types as well, e.g. for AArch64 FEAT_I8MM
+/// we might want to consider tile sizes that are multiples of 2x8.
+static void getMatmulVectorSizesUsingFillRegisterFileHeuristic(
     mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp op,
     int64_t vectorSize, SmallVectorImpl<int64_t> &sizes,
     SmallVectorImpl<bool> &scalableSizeFlags) {
-  if (sizes.empty()) {
-    getDefaultMatmulVectorSizes(op, vectorSize, sizes, scalableSizeFlags);
+  assert(sizes.empty() && "Pre-condition enforced by caller");
+
+  // Find the output element type of the matmul.
+  assert(op->getResultTypes().size() == 1 &&
+         "Expected single output type for matmul op");
+  Type outputEltType = getElementTypeOrSelf(op->getResultTypes()[0]);
+  if (!outputEltType.isSignlessIntOrFloat()) {
+    return;
   }
 
-  // Find the smallest type size in the matmul.
-  SmallVector<Type> matmulTypes;
-  auto operandTypes = op->getOperandTypes();
-  matmulTypes.append(operandTypes.begin(), operandTypes.end());
-  auto resultTypes = op->getResultTypes();
-  matmulTypes.append(resultTypes.begin(), resultTypes.end());
-
-  int64_t minSize = std::numeric_limits<int64_t>::max();
-  for (Type mmType : matmulTypes) {
-    if (auto shType = dyn_cast<ShapedType>(mmType)) {
-      mmType = shType.getElementType();
-    }
-
-    if (mmType.isSignlessIntOrFloat()) {
-      minSize = std::min(minSize, int64_t{mmType.getIntOrFloatBitWidth()});
-    }
-  }
-
-  LDBG() << "Smallest type found: " << minSize << " bits";
-  assert(minSize > 0 && minSize < std::numeric_limits<int64_t>::max() &&
-         "Min size couldn't be computed");
-
-  // Make sure that the smallest type can at least fill a full vector register
-  // given the tile size of the main vector dimension (N).
   constexpr int64_t byteSizeInBits = 8;
-  int64_t minNumElements =
-      (getNativeVectorSizeInBytes(entryPointFn) * byteSizeInBits) / minSize;
-  sizes[1] = std::max(sizes[1], minNumElements);
+  int64_t outBitWidth = outputEltType.getIntOrFloatBitWidth();
+  int64_t outNumElements =
+      (getNativeVectorSizeInBytes(entryPointFn) * byteSizeInBits) / outBitWidth;
+
+  // Numbers picked experimentally for a range of element types.
+  constexpr int64_t m = 8, n = 2, k = 1;
+
+  // Multiply "horizontal" extents by the number of elements that fit in a vector
+  // register.
+  sizes.append({m, n * outNumElements, k * outNumElements});
+
+  // Mark N dimension as scalable, if doing scalable vectorization.
+  auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(op);
+  scalableSizeFlags.resize(3, false);
+  if (isScalableVectorizationEnabled() &&
+      hasAnySVEFeature(targetAttr.getConfiguration())) {
+    scalableSizeFlags[1] = true;
+  }
 }
 
 /// Utility to compute the tile sizes for RISC-V Vector.
@@ -1701,7 +1701,7 @@ getMatmulVectorSizes(mlir::FunctionOpInterface entryPointFn,
     // Try to maximize the vector register utilization for all the matmul
     // element types.
     if (matmulTileSizes.empty()) {
-      getMatmulVectorSizesUsingFullVectorHeuristics(
+      getMatmulVectorSizesUsingFillRegisterFileHeuristic(
           entryPointFn, op, vectorSize, matmulTileSizes, matmulScalableFlags);
     }
   }
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -1577,8 +1577,8 @@ static void getMatmulVectorSizesUsingFillRegisterFileHeuristic(
   // Numbers picked experimentally for a range of element types.
   constexpr int64_t m = 8, n = 2, k = 1;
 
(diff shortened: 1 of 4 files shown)


banach-space:
Thank you for working on this, @momchil-velikov !

Approving as-is - we discussed this offline and the heuristics make sense to me. The experiments you ran (various matmul kernels + GPT-2) show clear benefit.

IIRC, the evaluation was done on [Neoverse V2](https://www.arm.com/products/silicon-ip-cpu/neoverse/neoverse-v2), right? The implementation itself doesn’t appear microarchitecture-specific, but it would be good to mention the evaluation hardware explicitly in the summary.

I appreciate that there might be some other (more generic?) approaches to this, but I don't have any specific suggestions myself. Given the observed performance improvements on the tested workloads, I am in favor.

It would be great to extend the PR summary with a bit more detail:
* Observed performance improvements,
* Which configurations are affected (e.g. DT, peeling),
* Hardware used for evaluation.

Thanks again! @hanhanW , what are your thoughts re the implementation? I've not really touched this code for a while, so don't have a strong opinion.

banach-space (approved):

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1553,6 +1553,57 @@ static void getMatmulVectorSizesUsingFullVectorHeuristics(
...
+    SmallVectorImpl<bool> &scalableSizeFlags) {
+  assert(sizes.empty() && "Pre-condition enforced by caller");
+
+  // Find the output element type of the matmul.
+  assert(op->getResultTypes().size() == 1 &&
+         "Expected single output type for matmul op");
+  auto shType = dyn_cast<ShapedType>(op->getResultTypes()[0]);
+  if (!shType)
+    return;
+  Type outputEltType = shType.getElementType();
```
hanhanW: We may use `getElementTypeOrSelf` here.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1671,7 +1722,7 @@ getMatmulVectorSizes(mlir::FunctionOpInterface entryPointFn,
     // Try to maximize the vector register utilization for all the matmul
     // element types.
     if (matmulTileSizes.empty()) {
-      getMatmulVectorSizesUsingFullVectorHeuristics(
```
hanhanW, line 1693: The function seems to be dead code, and we can delete it?

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1553,6 +1553,57 @@ static void getMatmulVectorSizesUsingFullVectorHeuristics(
...
+/// The rationale is that a matrix multiplication typically lowers to a loop
+/// nest in which the ACC/OUT tile remains live across all iterations of the
+/// innermost loop, whereas the LHS and RHS operands live for a single iteration
+/// and do not require the entire tiles to be simultaneously resident in
+/// registers.
+/// The base element type used is the element type of the output
+/// vector under the assumption the operand types with smaller bitwidths
+/// will be promoted to the output type and thus require more registers for the
+/// same number of elements.
+/// TODO: I might be worth extending the heuristic to consider target
```
hanhanW: typo?

```suggestion
/// TODO: It might be worth extending the heuristic to consider target
```

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1553,6 +1553,57 @@ static void getMatmulVectorSizesUsingFullVectorHeuristics(
...
+  // Multiply "horizontal" extents by the number of elements that fit in a vector
+  // register.
+  sizes.append({M, N * outNumElements, K * outNumElements});
+
+  // Mark N dimension as scalable, if doing scalable vectorization.
+  auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(op);
+  scalableSizeFlags.append({false,
+                            isScalableVectorizationEnabled() &&
+                                hasAnySVEFeature(targetAttr.getConfiguration()),
+                            false});
```
hanhanW: If you are passing `linalgOp` (not `MatmulOp`), it is better to infer M/N/K dims. The other approach can be bailing out in the entry if the op is not a matmul op.

I prefer returning the result instead of passing by reference, though. In any case, can you split it two main statements for readability? E.g.,

- Initialize with three false.
- Set N dim to true if it meets the condition.

```
vec.resize(3, false);
if (...) {
  vec[1] = true;
}
```

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1553,6 +1553,57 @@ static void getMatmulVectorSizesUsingFullVectorHeuristics(
...
+  if (!outputEltType.isSignlessIntOrFloat())
+    return;
+
+  constexpr int64_t byteSizeInBits = 8;
+  int64_t outBitWidth = outputEltType.getIntOrFloatBitWidth();
+  int64_t outNumElements =
+      (getNativeVectorSizeInBytes(entryPointFn) * byteSizeInBits) / outBitWidth;
+
+  // Numbers picked experimentally for a range of element types.
+  constexpr int64_t M = 8, N = 2, K = 1;
```
hanhanW: style nit: IREE follows camelCase naming, so they should be `m`, `n`, `k`.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1553,6 +1553,57 @@ static void getMatmulVectorSizesUsingFullVectorHeuristics(
...
+
+  // Find the output element type of the matmul.
+  assert(op->getResultTypes().size() == 1 &&
+         "Expected single output type for matmul op");
+  auto shType = dyn_cast<ShapedType>(op->getResultTypes()[0]);
+  if (!shType)
+    return;
+  Type outputEltType = shType.getElementType();
+  if (!outputEltType.isSignlessIntOrFloat())
+    return;
```
hanhanW: IREE style nit: wrap with braces. This is different from LLVM style guide. We wrap braces even if it is a single statement.


```suggestion
  if (!outputEltType.isSignlessIntOrFloat()) {
    return;
  }
```

hanhanW (changes requested):
> IIRC, the evaluation was done on [Neoverse V2](https://www.arm.com/products/silicon-ip-cpu/neoverse/neoverse-v2), right? The implementation itself doesn’t appear microarchitecture-specific, but it would be good to mention the evaluation hardware explicitly in the summary.

+1 for making summary better.

> Thanks again! @hanhanW , what are your thoughts re the implementation? I've not really touched this code for a while, so don't have a strong opinion.

I don't touch them for a while, either. IMO, some of the tile sizes were driven by old hardware or workload, so some of them are legacy to me. I don't have strong opinion for changing these code as long as it does not significant impact some benchmarks, although we dont have any benchmarks atm.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
momchil-velikov: >  infer M/N/K dims. 

I'm afraid I don't quite understand what does this mean.

Looking from the point of view of the function I'm replacing, there's no change in semantics, per se, just choosing different values with the convention that in the returned vectors the element 0 is M, the element 1 is N, and element 2 is K (M, N, and K themselves having conventional meaning). That looks common across most (all?) other functions in the file that suggest tile sizes.

> The other approach can be bailing out in the entry if the op is not a matmul op.

We know we have a contraction (`ContractionOpInterface`) with a single reduction dimension, is the choice going to be bad for *all* such contraction ops that are not a matmul?
Maybe for some, but then I'd rather enhance the heuristic if we stumble upon such cases.

banach-space: > I'm afraid I don't quite understand what does this mean.

See: https://github.com/llvm/llvm-project/blob/fa511cde48ea218dadfa8b35658ac06368f34607/mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.h?plain=1#L66

hanhanW: > Looking from the point of view of the function I'm replacing, there's no change in semantics, per se, just choosing different values with the convention that in the returned vectors the element 0 is M, the element 1 is N, and element 2 is K (M, N, and K themselves having conventional meaning). That looks common across most (all?) other functions in the file that suggest tile sizes.

Thanks @banach-space for the pointer! I was not clear about checking `MatmulOp` because I forgot that a MatmulOp op now takes indexing maps: https://github.com/llvm/llvm-project/commit/d15280894bd03c618ec1b4debcbfdb1fecb81594

Some codes are pretty old and they are not evolved because we're understaffing for some backends. IMO, we should bail out if we have such assumption. Otherwise, people may run into issues easily. E.g., you'll generate the same lowering config if you replace a matmul with the same matmul that has different indexing maps.

```
  // The shapes are MxK, NxK, MxN.
  %5 = linalg.matmul
    indexing_maps = [
      affine_map<(d0, d1, d2) -> (d0, d2)>,
      affine_map<(d0, d1, d2) -> (d1, d2)>,
      affine_map<(d0, d1, d2) -> (d0, d1)>
    ]
    ins(%1, %2 : tensor<512x128xi8>, tensor<512x128xi8>) outs(%fill : tensor<512x512xi32>) -> tensor<512x512xi32>
```

We may have a utility function to check if a `linalg::LinalgOp` is a common matmul form and use it here; it addresses my concern.

egebeysel:
I'll take a look at the code today or tomorrow, but I quickly gave this a try on my M5:

(latencies without the PR vs with the PR)

**Dynamic Matmul**

- 1024x1024x1024: 9.06 vs 7.63 ms
- (wide k) 1024x1024x4096:  52.0 vs 64.6 ms
- (wide n) 1024x4096x1024: 39.8 vs 45.9 ms
- (wide m) 4096x1024x1024: 25.6 vs 21.8 ms

**Static Matmul**

- 1024x1024x1024: 11.6 vs 7.45 ms
- (wide k) 1024x1024x4096:  55.5 vs 68.1 ms
- (wide n) 1024x4096x1024: 42.1 vs 40.4 ms
- (wide m) 4096x1024x1024: 33.9 vs 22.8 ms


(used the flags `--iree-llvmcpu-enable-ukernels=none --iree-opt-data-tiling=false --iree-dispatch-creation-data-tiling=false --iree-llvmcpu-target-cpu=host --iree-hal-target-backends=llvm-cpu` to compile)

momchil-velikov:
> I'll take a look at the code today or tomorrow, but I quickly gave this a try on my M5:
> 
> (latencies without the PR vs with the PR)
> 
> **Dynamic Matmul**
> 
> * 1024x1024x1024: 9.06 vs 7.63 ms
> * (wide k) 1024x1024x4096:  52.0 vs 64.6 ms
> * (wide n) 1024x4096x1024: 39.8 vs 45.9 ms
> * (wide m) 4096x1024x1024: 25.6 vs 21.8 ms
> 
> **Static Matmul**
> 
> * 1024x1024x1024: 11.6 vs 7.45 ms
> * (wide k) 1024x1024x4096:  55.5 vs 68.1 ms
> * (wide n) 1024x4096x1024: 42.1 vs 40.4 ms
> * (wide m) 4096x1024x1024: 33.9 vs 22.8 ms
> 
> (used the flags `--iree-llvmcpu-enable-ukernels=none --iree-opt-data-tiling=false --iree-dispatch-creation-data-tiling=false --iree-llvmcpu-target-cpu=host --iree-hal-target-backends=llvm-cpu` to compile)

I have also observed that no single set of tile parameters are best for all matrix sizes. Also regressions are only(?) in SVE code, Neon universally improves.
One issue is increase in cache misses, mostly when loading the RHS. The accesses to RHS are with a wide stride (for both the original and the patched compilers), however the original RHS tile rows are twice bigger (4 registers) than the modified ones (2 registers), which is one possible explanation. 
I have also experimented with tile parameters that increase `N` while still staying within register budget, but then the SVE register allocation  is quite suboptimal, with unnecessary spills.
There are other deficiencies with SVE codegen (for some of which we have initial downstream patches): 
* it does not use indexed FMLA (that potentially saves an explicit broadcast)
* there are not fully realized opportunities to use the `ld1rqw` for loading the LHS
* the addressing calculations for loading the strided rows (where you need to add the same constant a few times to a base address) are preformed quite badly (sometimes, may even result in spilling GPRs in the inner loop).

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
momchil-velikov: Done

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1506,51 +1506,62 @@ static FailureOr<Type> nonWideningLinalgElementType(linalg::LinalgOp op) {
...
-
-    if (mmType.isSignlessIntOrFloat())
-      minSize = std::min(minSize, int64_t{mmType.getIntOrFloatBitWidth()});
+  // Check we have a matrix multiplication.
+  FailureOr<linalg::ContractionDimensions> cDims =
+      linalg::inferContractionDims(op);
+  if (failed(cDims) || cDims->m.size() != 1 || cDims->n.size() != 1 ||
+      cDims->k.size() != 1) {
+    return;
   }
```
hanhanW, line 1564: You also need to check that `mDim == 0, nDim == 1, kDim == 2`.

hanhanW (commented):
I have a final nit. Please fix DCO issue, thanks.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
momchil-velikov: Done

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1528,54 +1528,63 @@ static FailureOr<Type> nonWideningLinalgElementType(linalg::LinalgOp op) {
...
 
-  // Find the smallest type size in the matmul.
-  SmallVector<Type> matmulTypes;
-  auto operandTypes = op->getOperandTypes();
-  matmulTypes.append(operandTypes.begin(), operandTypes.end());
-  auto resultTypes = op->getResultTypes();
-  matmulTypes.append(resultTypes.begin(), resultTypes.end());
+  // Find the output element type of the matmul.
+  assert(op->getResultTypes().size() == 1 &&
+         "Expected single output type for matmul op");
```
hanhanW: `inferContractionDims` guarantees that it has single result, because it is op interface definition. The implementation also reflects the requirement. So we can remove this assertion.

https://github.com/llvm/llvm-project/blob/91c4decc01522c0130d47f8270330194259f4207/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp#L483-L489

hanhanW (approved):
LGTM, just a final nit about assertion. I don't know why I did not point it out in the first place, sorry about that. Thanks for your patch!

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
momchil-velikov: Done

hanhanW:
Hi @momchil-velikov I'm not sure if you have permission to merge the PR. I'm happy to help land it if you can clear the lint issue. Thanks!

PR #22936, opened 2025-12-17
[CI][Torch Models] Tighten golden times 

efric:
Tighten golden times for benchmarks that exhibit significant % improvement over current goldens. The new values are those that have reasonable absolute margins and are derived from the average times of the 3 most recent passing commits with a 1.1x multiplier.


MaheshRavishankar (approved):
Thanks!

PR #22945, opened 2025-12-18
e2e matmul tests: fix some recent CDNA3-only tests

bjacob:
Some of theses tests were added to the Bazel build where they didn't belong and bazel-to-cmake couldn't handle the enablement logic for GPU-specific tests.

As GPU specific tests they are now correctly conditioned on IREE_HIP_TEST_TARGET_CHIP and also generalized to CDNA4 as there was no need to restrict them to CDNA3 only.


MaheshRavishankar (approved):
Awesome! Thanks!

krzysz00 (approved):

PR #22953, opened 2025-12-19
[Codegen] add FoldExtractSliceOfBroadcast pattern to TileAndDistributeToWorkgroups

bangtianliu:
This PR adds the FoldExtractSliceOfBroadcast pattern to TileAndDistributeToWorkgroupsUsingForall. This avoids mishandling of broadcasts with unit dimensions during this pass, which previously caused bufferization issues and subsequent numerical correctness problems.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
@@ -8,6 +8,7 @@
 #include "iree/compiler/Codegen/Common/CombineLayoutTransformation.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
 #include "llvm/ADT/ScopeExit.h"
+#include "llvm/ADT/SmallVectorExtras.h"
 #include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
@@ -374,7 +375,7 @@ namespace {
 ///   %extract is replaced by %in (tensor<4x1xf16>)
 struct FoldExtractSliceOfBroadcast final
     : OpRewritePattern<tensor::ExtractSliceOp> {
-  using OpRewritePattern::OpRewritePattern;
+  using Base::Base;
 
   LogicalResult matchAndRewrite(tensor::ExtractSliceOp extractOp,
                                 PatternRewriter &rewriter) const override {
@@ -411,9 +412,7 @@ struct FoldExtractSliceOfBroadcast final
     // shape as the input (essentially undoing the broadcast).
     SmallVector<OpFoldResult> offsets = extractOp.getMixedOffsets();
     SmallVector<OpFoldResult> sizes = extractOp.getMixedSizes();
-    if (llvm::any_of(offsets, [](OpFoldResult offset) {
-          return !isConstantIntValue(offset, 0);
-        })) {
+    if (!llvm::all_of(offsets, isZeroInteger)) {
       return rewriter.notifyMatchFailure(
           extractOp, "extract_slice offsets are not all zeros");
     }
@@ -423,8 +422,8 @@ struct FoldExtractSliceOfBroadcast final
     int64_t broadcastRank = broadcastOutputType.getRank();
 
     // Verify that for broadcast dimensions, the size is 1.
-    if (llvm::any_of(broadcastDims, [&](int64_t broadcastDim) {
-          return !isConstantIntValue(sizes[broadcastDim], 1);
+    if (!llvm::all_of(broadcastDims, [&](int64_t broadcastDim) {
+          return isOneInteger(sizes[broadcastDim]);
         })) {
       return rewriter.notifyMatchFailure(
           extractOp, "broadcast dimensions do not all have size 1");
@@ -432,9 +431,9 @@ struct FoldExtractSliceOfBroadcast final
 
     // Collect the indices of dimensions in the broadcast output that were not
     // broadcasted (i.e., dimensions that existed in the original input).
-    auto nonBroadcastDims = llvm::to_vector(llvm::make_filter_range(
+    auto nonBroadcastDims = llvm::filter_to_vector(
         llvm::seq<int64_t>(0, broadcastRank),
-        [&](int64_t i) { return !llvm::is_contained(broadcastDims, i); }));
+        [&](int64_t i) { return !llvm::is_contained(broadcastDims, i); });
 
     // Verify that for non-broadcast dimensions, sizes match input shape.
     if (llvm::any_of(llvm::enumerate(nonBroadcastDims), [&](auto pair) {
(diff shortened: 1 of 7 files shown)


bangtianliu:
The pattern was originally defined locally in FormSplitReductionDispatches.cpp. Since we now need the same pattern in TileAndDistributeToWorkgroupsUsingForall, I moved it to Transforms.cpp with a populateFoldExtractSliceOfBroadcastPattern() function and updated both passes to use the shared implementation.

MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/Common/test/tile_and_distribute_workgroups_using_forall.mlir
```diff
@@ -1364,3 +1364,60 @@ attributes {translation_info = #iree_codegen.translation_info<pipeline = LLVMGPU
...
+  %1 = hal.interface.binding.subspan layout(<bindings = [#hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, Indirect>, #hal.pipeline.binding<storage_buffer, Indirect>], flags = Indirect>) binding(1) alignment(64) offset(%c0) flags(Indirect) : memref<4x1x96xf16, #hal.descriptor_type<storage_buffer>>
+  %2 = amdgpu.fat_raw_buffer_cast %1 resetOffset : memref<4x1x96xf16, #hal.descriptor_type<storage_buffer>> to memref<4x1x96xf16, #amdgpu.address_space<fat_raw_buffer>>
+  %3 = hal.interface.binding.subspan layout(<bindings = [#hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, Indirect>, #hal.pipeline.binding<storage_buffer, Indirect>], flags = Indirect>) binding(2) alignment(64) offset(%c768) flags(Indirect) : memref<4x1x96xi32, strided<[96, 96, 1], offset: ?>, #hal.descriptor_type<storage_buffer>>
```
kuhar: Would it be possible to turn these into function arguments to simplify the test?

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -354,6 +355,108 @@ void populateSwapExtractWithExpandPattern(RewritePatternSet &patterns) {
...
+///   %broadcast = linalg.broadcast ins(%in : tensor<4x1xf16>)
+///                                 outs(%out : tensor<4x1x1xf16>) dimensions =
+///                                 [2]
+///   %extract = tensor.extract_slice %broadcast[0, 0, 0] [4, 1, 1] [1, 1, 1]
+///              : tensor<4x1x1xf16> to tensor<4x1xf16>
+/// ->
+///   %extract is replaced by %in (tensor<4x1xf16>)
+struct FoldExtractSliceOfBroadcast final
+    : OpRewritePattern<tensor::ExtractSliceOp> {
+  using OpRewritePattern::OpRewritePattern;
```
kuhar:
```suggestion
  using Base::Base;
```

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -354,6 +355,108 @@ void populateSwapExtractWithExpandPattern(RewritePatternSet &patterns) {
...
+          extractOp, "extract result type does not match broadcast input type");
+    }
+
+    // Verify that we're extracting from offset [0, 0, ..., 0] with the same
+    // shape as the input (essentially undoing the broadcast).
+    SmallVector<OpFoldResult> offsets = extractOp.getMixedOffsets();
+    SmallVector<OpFoldResult> sizes = extractOp.getMixedSizes();
+    if (llvm::any_of(offsets, [](OpFoldResult offset) {
+          return !isConstantIntValue(offset, 0);
+        })) {
```
kuhar:
```suggestion
    if (!llvm::all_of(offsets, isZeroInteger)) {
```
I think this aligns better with the match failure message.

(I realize this code is moved over, so fine to keep as-is in this PR)

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -354,6 +355,108 @@ void populateSwapExtractWithExpandPattern(RewritePatternSet &patterns) {
...
+    // Sizes should match input dimensions (accounting for broadcast dims).
+    ArrayRef<int64_t> broadcastDims = broadcastOp.getDimensions();
+    int64_t broadcastRank = broadcastOutputType.getRank();
+
+    // Verify that for broadcast dimensions, the size is 1.
+    if (llvm::any_of(broadcastDims, [&](int64_t broadcastDim) {
+          return !isConstantIntValue(sizes[broadcastDim], 1);
+        })) {
+      return rewriter.notifyMatchFailure(
+          extractOp, "broadcast dimensions do not all have size 1");
```
kuhar, line 429: Similar here

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -354,6 +355,108 @@ void populateSwapExtractWithExpandPattern(RewritePatternSet &patterns) {
...
+    }
+
+    // Collect the indices of dimensions in the broadcast output that were not
+    // broadcasted (i.e., dimensions that existed in the original input).
+    auto nonBroadcastDims = llvm::to_vector(llvm::make_filter_range(
```
kuhar: `llvm::filter_vector`

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
kuhar: this used to use `Base::Base` but got dropped after the move for some reason

compiler/src/iree/compiler/Codegen/Common/test/tile_and_distribute_workgroups_using_forall.mlir
```diff
@@ -1367,52 +1367,64 @@ attributes {translation_info = #iree_codegen.translation_info<pipeline = LLVMGPU
...
-      %21 = arith.cmpf ogt, %arg3, %arg4 : f16
-      iree_linalg_ext.yield %21 : i1
+  %empty_f16 = tensor.empty() : tensor<4x1x96xf16>
+  %empty_i32 = tensor.empty() : tensor<4x1x96xi32>
+  %init_f16 = tensor.empty() : tensor<4x1xf16>
+  %init_i32 = tensor.empty() : tensor<4x1xi32>
+  %fill_f16 = linalg.fill {lowering_config = #iree_codegen.lowering_config<tile_sizes = [[1, 128]]>}
+      ins(%cst : f16) outs(%init_f16 : tensor<4x1xf16>) -> tensor<4x1xf16>
+  %fill_i32 = linalg.fill {lowering_config = #iree_codegen.lowering_config<tile_sizes = [[1, 128]]>}
+      ins(%c0_i32 : i32) outs(%init_i32 : tensor<4x1xi32>) -> tensor<4x1xi32>
```
kuhar: Could these also be function arguments? Do we care about the lowering config (I don't know)

bangtianliu: Good catch! This test IR was derived from a real case where the race condition occurred and can certainly be simplified. Since this pattern (FoldExtractSliceOfBroadcast) doesn't specifically handle fill operations, the lowering_config isn't required. Will remove.

kuhar (approved):

PR #22921, opened 2025-12-16
[DispatchCreation][NFC] Integrate changes to `dropUnitDims`

sommerlukas:
Integrate upstream changes to the `dropUnitDims` API from https://github.com/llvm/llvm-project/pull/171796 into the `FoldUnitExtentDims` pass.

The behavior remains the same for now, therefore no test is changed. In a future PR, we will use the extended upstream API to allow folding for tensors with certain encodings.

This PR also fixes a small bug in the `DropUnitDimsFromCollapseOfExpand` pattern.

PR #22955, opened 2025-12-23
[GPU][Codegen] Skip dimension expansion for ops with nonfusable indexing maps

efric:
The `expand_dims` lowering config attribute relies on `linalg::populateFoldReshapeOpsByExpansionPatterns` to fuse `tensor.expand_shape`/`tensor.collapse_shape` into the linalg op in `GPUExpandDimensions`. This requires all indexing maps to be projected permutations. This patch adds a check to verify this before setting the attribute. 

Not sure why this wasn't caught in the PR CI but fixes compilation issue of `densenet-12` in [test_onnx_models::amdgpu_hip_rdna3](https://github.com/iree-org/iree/actions/runs/20449877963/job/58761041931#logs).


compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_vector_distribute_reduction_gfx942.mlir
```diff
@@ -442,3 +442,49 @@ func.func @batch_matvec_f16_f32() {
...
+  %filled = linalg.fill ins(%cst : f16) outs(%empty : tensor<4xf16>) -> tensor<4xf16>
+  %result = linalg.generic {
+    indexing_maps = [
+      affine_map<(d0, d1) -> (d1, 0, 0)>,
+      affine_map<(d0, d1) -> (d0, d1)>,
+      affine_map<(d0, d1) -> (d0)>
+    ], iterator_types = ["parallel", "reduction"]
+  } ins(%in0, %in1 : tensor<16384x1x1xf16>, tensor<4x16384xf16>)
+  outs(%filled : tensor<4xf16>) {
+  ^bb0(%a: f16, %b: f16, %out: f16):
```
Groverkss, line 472: I think something else is wrong if we are getting these kinds of indexing maps in our codegen. There is a pass that runs which should remove constants in indexing maps. Can you mentioned the full example instead?

Groverkss (changes requested):

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_vector_distribute_reduction_gfx942.mlir
efric: Ah, makes sense. This was observed here https://github.com/iree-org/iree/actions/runs/20449877963/job/58761041931#step:8:166 but I will look more into it today.

efric: After digging into it more, I think the root issue is from the most recent integrate #22943, likely involving the changes to `iree-dispatch-creation-fold-unit-extent-dims` from #22921. In particular, unit dims are not being properly folded away during GlobalOptimization. I think it didn’t surface until after the dimension expansion PR got merged because by luck we didn’t run into cases where this was a problem. Filed issue along with reproduction steps here: #22978. Closing this PR.

Issue #21865, opened 2025-09-05
[GPU][DT] Unexpected scratch memory usage from map_scatter


Yu-Zhewen:
To reproduce, run [this example](https://gist.github.com/Yu-Zhewen/a2b6110e8f5b39de4d2384f886380b9a) at main 3aa9c804084817094c27e541d04aaa428ef5696f  with

```
iree-compile matmul_large.mlir \
  -o matmul.vmfb \
  --iree-hal-target-backends=rocm \
  --iree-hip-target=gfx942 \
  --iree-hal-target-device=hip \
  --iree-opt-level=O3 \
  --iree-dispatch-creation-propagate-collapse-across-expands=true \
  --iree-codegen-enable-default-tuning-specs=true \
  --iree-hal-indirect-command-buffers=true \
  --iree-stream-resource-memory-model=discrete \
  --iree-hip-specialize-dispatches \
  --iree-hal-memoization=true \
  --iree-opt-data-tiling=false \
  --iree-dispatch-creation-data-tiling \
  --iree-hip-encoding-layout-resolver=data-tiling \
  --iree-llvmgpu-test-combine-layout-transformation \
  --iree-hal-dump-executable-intermediates-to=dt_asm \
  --iree-hip-enable-tensor-ukernels
```

Compare the assembly with the non-data-tiling version by removing 

```
  --iree-opt-data-tiling=false \
  --iree-dispatch-creation-data-tiling \
  --iree-hip-encoding-layout-resolver=data-tiling \
  --iree-llvmgpu-test-combine-layout-transformation \
```

It appears related to GPUFuseAndHoistParallelLoopsPass. The map_scatter op cannot be tiled on its output side, which causes a dynamic shape to be carried into the workgroup. This, in turn, triggers unexpected scratch memory usage. When the example is changed to a static shape, everything is fine.

Currently, this issue can be bypassed by this hack e3bab94a4f0fb26ae5c4390d566ce61daa0270f2.

Yu-Zhewen:
Updated results on llama (8B, prefill, FP8, seq length 2048)

main 3aa9c804084817094c27e541d04aaa428ef5696f + PR #21957
- non data-tiling + ukernel: 235ms
- data-tiling without ukernel: 345ms
- data-tiling + ukernel: 285ms

main 3aa9c804084817094c27e541d04aaa428ef5696f + PR #21957 + cherry-pick e3bab94a4f0fb26ae5c4390d566ce61daa0270f2
[...]

PR #22958, opened 2025-12-23
[LinalgExt] Simplify linearize-delinearize pairs in map_scatter

Yu-Zhewen:
Linearize-delinearize pairs often appear inside `map_scatter`. When these operations get lowered, they can create long multiply-add and divide-remainder chains, especially when a dynamic dimension occurs.

This PR adds a pattern to simplify such operation pairs in the `DecomposeMapScatter` pass. We can optimize away the redundant operations by matching dimension products and breaking them into smaller, more efficient chunks.

This optimization helps mitigate the VGPR spilling issue observed in https://github.com/iree-org/iree/issues/21865.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
@@ -148,8 +148,8 @@ struct SimplifyLinearizeDelinearizePairs final
     SmallVector<OpFoldResult> delinearizeBases = delinearizeOp.getMixedBasis();
     ValueRange linearizeInputs = linearizeOp.getMultiIndex();
 
-    rewriter.setInsertionPoint(delinearizeOp);
     Location loc = delinearizeOp.getLoc();
+    MLIRContext *ctx = rewriter.getContext();
 
     // Structure to store the inputs and bases for new linearize ops.
     struct LinearizeInfo {
@@ -157,9 +157,6 @@ struct SimplifyLinearizeDelinearizePairs final
       SmallVector<OpFoldResult> bases;
     };
     SmallVector<LinearizeInfo> newLinearizeInfos;
-
-    // Some temporary operations that are created during the matching process.
-    SmallVector<Operation *> toErase;
 
     // For each delinearize output dimension, try to match it with a product of
     // one or more linearize dimensions. We build up the product incrementally
@@ -169,41 +166,42 @@ struct SimplifyLinearizeDelinearizePairs final
     for (size_t delinIdx = 0;
          delinIdx < delinearizeBases.size() && linIdx < linearizeBases.size();
          ++delinIdx) {
-      // Get the delinearize basis for this dimension.
-      Value delinearizeBasisVal = getValueOrCreateConstantIndexOp(
-          rewriter, loc, delinearizeBases[delinIdx]);
-      if (delinearizeBasisVal.use_empty()) {
-        toErase.push_back(delinearizeBasisVal.getDefiningOp());
-      }
-
       LinearizeInfo newLinearizeInfo;
-      Value linearizeBasisProduct;
-
+      // Track operands for the AffineMap-based product expression.
+      SmallVector<Value> productOperands;
       // Accumulate dimensions from the linearize op until we find a match.
       while (linIdx < linearizeBases.size()) {
-        Value linearizeSizeVal = getValueOrCreateConstantIndexOp(
-            rewriter, loc, linearizeBases[linIdx]);
-        if (linearizeSizeVal.use_empty()) {
-          toErase.push_back(linearizeSizeVal.getDefiningOp());
-        }
-
         newLinearizeInfo.inputs.push_back(linearizeInputs[linIdx]);
         newLinearizeInfo.bases.push_back(linearizeBases[linIdx]);
-        ++linIdx;
 
-        // Build up the product of basis dimensions as a Value.
-        if (!linearizeBasisProduct) {
-          linearizeBasisProduct = linearizeSizeVal;
-        } else {
-          linearizeBasisProduct = arith::MulIOp::create(
-              rewriter, loc, linearizeBasisProduct, linearizeSizeVal);
-          toErase.push_back(linearizeBasisProduct.getDefiningOp());
+        // Build up the product of basis dimension using affine expressions.
+        AffineExpr productExpr = getAffineConstantExpr(1, ctx);
+        productOperands.clear();
+        for (OpFoldResult basis : newLinearizeInfo.bases) {
+          if (auto attr = dyn_cast<Attribute>(basis)) {
+            // Handle the static basis as a constant expression.
+            int64_t val = cast<IntegerAttr>(attr).getInt();
+            productExpr = productExpr * getAffineConstantExpr(val, ctx);
+          } else {
+            // Handle the dynamic basis as a symbol expression.
+            Value val = cast<Value>(basis);
+            productOperands.push_back(val);
+            productExpr = productExpr *
+                          getAffineSymbolExpr(productOperands.size() - 1, ctx);
+          }
         }
+        ++linIdx;
+        // Create Variable from the product expression.
+        AffineMap productMap =
+            AffineMap::get(0, productOperands.size(), productExpr, ctx);
+        ValueBoundsConstraintSet::Variable productVar(productMap,
+                                                      productOperands);
+        ValueBoundsConstraintSet::Variable delinearizeVar(
+            delinearizeBases[delinIdx]);
 
-        // Check if the delinearize basis matches the product of linearize
-        // bases.
-        FailureOr<bool> areEqual = ValueBoundsConstraintSet::areEqual(
-            delinearizeBasisVal, linearizeBasisProduct);
+        // Check equality.
+        FailureOr<bool> areEqual =
+            ValueBoundsConstraintSet::areEqual(productVar, delinearizeVar);
         if (succeeded(areEqual) && *areEqual) {
           // Match found!
           newLinearizeInfos.push_back(newLinearizeInfo);
@@ -232,12 +230,6 @@ struct SimplifyLinearizeDelinearizePairs final
       return success();
     }
 
-    // When the pattern fails to match, we need to explicitly clean up the
-    // temporary operations that were created during matching. This is necessary
-    // when patterns are applied in a greedy manner.
-    for (auto op : llvm::reverse(toErase)) {
-      rewriter.eraseOp(op);
-    }
     return rewriter.notifyMatchFailure(
         delinearizeOp, "could not match all delinearize outputs");
   }
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
@@ -16,6 +16,7 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/Interfaces/ValueBoundsOpInterface.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
 #define DEBUG_TYPE "iree-decompose-map-scatter"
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -105,6 +106,143 @@ struct FoldSubViewIntoMapScatter final : OpRewritePattern<MapScatterOp> {
...
+    // We only handle disjoint linearizations.
+    if (!linearizeOp.getDisjoint()) {
+      return rewriter.notifyMatchFailure(delinearizeOp,
+                                         "linearize op is not disjoint");
+    }
+    SmallVector<OpFoldResult> linearizeBases = linearizeOp.getMixedBasis();
+    SmallVector<OpFoldResult> delinearizeBases = delinearizeOp.getMixedBasis();
+    ValueRange linearizeInputs = linearizeOp.getMultiIndex();
+
+    rewriter.setInsertionPoint(delinearizeOp);
```
hanhanW: I thought the default insertion point is the same, i.e., right before the operation. Why do you need to set it again?

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_map_scatter.mlir
```diff
@@ -364,3 +364,37 @@ func.func @map_scatter_with_mask_on_inner_dim(
...
+    ^bb0(%idx0: index, %idx1: index, %idx2: index, %idx3: index, %idx4: index, %idx5: index, %idx6: index, %idx7: index, %idx8: index):
+      %mask = arith.constant true
+      %linearized_9d = affine.linearize_index disjoint [%idx0, %idx1, %idx2, %idx3, %idx4, %idx5, %idx6, %idx7, %idx8] by (%dim_ceildiv_128, 208, 2, 4, 4, 4, 2, 8, 16) : index
+      %delinearized_4d:4 = affine.delinearize_index %linearized_9d into (%dim_ceildiv_128, 208, 128, 256) : index, index, index, index
+      %linearized_4d = affine.linearize_index disjoint [%delinearized_4d#0, %delinearized_4d#2, %delinearized_4d#1, %delinearized_4d#3] by (%dim_ceildiv_128, 128, 208, 256) : index
+      %delinearized_2d:2 = affine.delinearize_index %linearized_4d into (%dim_aligned_128, 53248) : index, index
+      iree_linalg_ext.yield %delinearized_2d#0, %delinearized_2d#1, %mask : index, index, i1
+  } : vector<1x208x2x4x4x4x2x8x16xf16> into memref<?x53248xf16>
+  return
+}
```
hanhanW, line 419: Can you add a comment for the test? IIUC, it does not only test a pair, but also test the dynamic case (i.e., we are using SSA values in `linearized_4d`). Adding a comment helps people to understand why this is needed.

I think we also need a simple test for a simple pair, which helps future debugging easier as people won't have to analyze or break the test case.

After writing these down, I wonder if we can break the case into two test cases? One for a simple pair, and the other for dynamic. IIUC, the pattern-rewrite order is breaking `delinearized_4d` first, then `delinearized_9d`? In this context, I suggest having two test cases. One is `delinearized_4d`, and the other is the result (or simplified version) of first rewrite. It is clearer to me what we are testing.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -105,6 +106,143 @@ struct FoldSubViewIntoMapScatter final : OpRewritePattern<MapScatterOp> {
...
+
+    // Structure to store the inputs and bases for new linearize ops.
+    struct LinearizeInfo {
+      SmallVector<Value> inputs;
+      SmallVector<OpFoldResult> bases;
+    };
+    SmallVector<LinearizeInfo> newLinearizeInfos;
+
+    // Some temporary operations that are created during the matching process.
+    SmallVector<Operation *> toErase;
```
hanhanW: I understand why we need this, but it is really a workaround to me and it is fragile. IMO, we shouldn't create operations on failures, if possible. I think the main issue is that you create `arith::MulIOp` for analysis. I took a look at the `ValueBoundsConstraintSet` and I think we can avoid it this way:

The Variable class has a constructor that takes an AffineMap: https://github.com/llvm/llvm-project/blob/80b62cb9b979bf560ea17c127b1e85749376a6a2/mlir/include/mlir/Interfaces/ValueBoundsOpInterface.h#L136-L138

In this context, we can build an AffineExpr product and use Variable. Below is the code generated by my LLM agent, and it looks reasonable to me. Can you give it a shot?

```cpp
MLIRContext *ctx = rewriter.getContext();
SmallVector<OpFoldResult> linearizeBasisSlice;
SmallVector<Value> operands;

// Accumulate bases in this delinearize group
while (linIdx < linearizeBases.size()) {
  linearizeBasisSlice.push_back(linearizeBases[linIdx]);
  newLinearizeInfo.inputs.push_back(linearizeInputs[linIdx]);
  newLinearizeInfo.bases.push_back(linearizeBases[linIdx]);
  ++linIdx;
  // Build symbolic product as AffineExpr
  AffineExpr productExpr = getAffineConstantExpr(1, ctx);
  for (auto basis : linearizeBasisSlice) {
    if (auto attr = dyn_cast<Attribute>(basis)) {
      int64_t val = cast<IntegerAttr>(attr).getInt();
      productExpr = productExpr * getAffineConstantExpr(val, ctx);
    } else {
      // For dynamic values, create a symbol
      Value val = cast<Value>(basis);
      operands.push_back(val);
      productExpr = productExpr * getAffineSymbolExpr(operands.size()-1, ctx);
    }
  }

  // Create Variable from the product expression
  AffineMap productMap = AffineMap::get(0, operands.size(), productExpr, ctx);
  ValueBoundsConstraintSet::Variable productVar(productMap, operands);
  ValueBoundsConstraintSet::Variable delinearizeVar(delinearizeBases[delinIdx]);

  FailureOr<bool> areEqual =
    ValueBoundsConstraintSet::areEqual(productVar, delinearizeVar);
    if (succeeded(areEqual) && *areEqual) {
    newLinearizeInfos.push_back(newLinearizeInfo);
    break;
  }
}
```

hanhanW (changes requested):

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
Yu-Zhewen: Ah yeah, removed.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_map_scatter.mlir
Yu-Zhewen: Split into two separate cases now: static and dynamic, each with one linearize-delinearize pair

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
Yu-Zhewen: Thanks! That's actually what I was looking for.

Max191 (commented):
It would be nice to add these directly to the upstream canonicalizations for these ops:

https://github.com/llvm/llvm-project/blob/125948a058dcd89b7fe377872a5fc1a7f9d34e70/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L5254-L5257

https://github.com/llvm/llvm-project/blob/125948a058dcd89b7fe377872a5fc1a7f9d34e70/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L5691-L5692

There are already several cases handled by the upstream patterns, so it could potentially simplify some of this logic when it is consolidated there as well.

There shouldn't be anything special about linearize/delinearize ops inside of a map_scatter op, and we'd want to do this simplification elsewhere in the compiler if it ever shows up. Also, it will avoid any duplicated work if someone else wants to try and do the same thing upstream.

Yu-Zhewen:
I'm not sure this fits well upstream at the moment, as the existing canonicalizations for these ops bail out on dynamic dimensions in most cases ([example](https://github.com/llvm/llvm-project/blob/125948a058dcd89b7fe377872a5fc1a7f9d34e70/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L5200)), likely because handling dynamics can be expensive. I was thinking we could land this in IREE for now, and if there's strong demand from other users, we can upstream it later.

Happy to hear what others think though!

Max191:
> I'm not sure this fits well upstream at the moment, as the existing canonicalizations for these ops bail out on dynamic dimensions in most cases ([example](https://github.com/llvm/llvm-project/blob/125948a058dcd89b7fe377872a5fc1a7f9d34e70/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L5200)), likely because handling dynamics can be expensive. I was thinking we could land this in IREE for now, and if there's strong demand from other users, we can upstream it later.
> 
> Happy to hear what others think though!

I see, if the value bounds check is causing friction, then adding it in IREE is fine to me. We can add it to the codegen canonicalizer to still get the benefits of the pattern elsewhere in codegen:

https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/IREECodegenCanonicalizer.cpp

A couple of comments about this, since value bounds is involved, and we did something similar for subview ops:

1. I wonder if we need to actually use value bounds. It is common for canonicalizers + cse to resolve the dynamic values to the same SSA value, in which case the check can be trivial. If this is the case for your situation, you could try reusing some of the logic here to do the bounds checking, which will be more efficient: https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Codegen/Common/IREECodegenCanonicalizer.cpp#L26
2. If we do need the actual bounds checking, then IMO it is better to separate the bounds checking from the transformation patterns. This would be out of the scope of this PR, but it is worth adding a TODO. Over time we've been doing more complex index analysis, and I think we should have a better solution than calling value bounds or int range analysis in every pattern. I want to organize my thoughts into an issue, and then I can link it here (before or after this one lands).

krzysz00:
To put my two cents here: the dynamic-reasoning case should not be a canonicalization but should be upstream as a separate set of patterns and pass - something like `affine-simplify-with-bounds` which could also do some stuff with `affine.apply' in subsequent PRs

(The common upstream note, which I agree with in this case, is that canonicalizations shouldn't run complex analyses)

I'd also like to flag a pattern that could exist upstream but doesn't, which is a linearize(linearize)) combiner, or a thing that turns maps into linearize (disjoint) where possible, both of which will probably want to do some reasoning

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -105,6 +106,135 @@ struct FoldSubViewIntoMapScatter final : OpRewritePattern<MapScatterOp> {
...
+          newResults.push_back(newLinearized);
+        }
+      }
+      rewriter.replaceOp(delinearizeOp, newResults);
+      return success();
+    }
+
+    return rewriter.notifyMatchFailure(
+        delinearizeOp, "could not match all delinearize outputs");
+  }
```
hanhanW: two comments:

1. Use early-exist style to reduce the indent of main logic. I.e.,

```cpp
if (newLinearizeInfos.size() != delinearizeOp.getNumResults()) {
  return failure(...);
}
// main logic
```

2. The comment should be consistent. We either use `multi-to-one` or `many-to-one`.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -420,6 +550,7 @@ struct DecomposeMapScatterPass final
 
     RewritePatternSet patterns(context);
     patterns.add<FoldSubViewIntoMapScatter>(context);
+    patterns.add<SimplifyLinearizeDelinearizePairs>(context);
```
hanhanW: Let's combine two lines?

hanhanW (approved):
Overall looks good to me, thanks!

Please do not add it to canonicalization patterns because it is expensive. I don't have strong opion if this should live in IREE repo or upstream. Living in IREE repo could be better if we'd extend the analysis with IREE's specific interfaces. I don't have a concrete idea, so I'm okay with both.

hanhanW:
> We can add it to the codegen canonicalizer to still get the benefits of the pattern elsewhere in codegen

We can consider this, but definitely not adding it to generic canonicalization patterns.

Yu-Zhewen:
> > I'm not sure this fits well upstream at the moment, as the existing canonicalizations for these ops bail out on dynamic dimensions in most cases ([example](https://github.com/llvm/llvm-project/blob/125948a058dcd89b7fe377872a5fc1a7f9d34e70/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L5200)), likely because handling dynamics can be expensive. I was thinking we could land this in IREE for now, and if there's strong demand from other users, we can upstream it later.
> > Happy to hear what others think though!
> 
> I see, if the value bounds check is causing friction, then adding it in IREE is fine to me. We can add it to the codegen canonicalizer to still get the benefits of the pattern elsewhere in codegen:
> 
> https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/IREECodegenCanonicalizer.cpp
> 
> A couple of comments about this, since value bounds is involved, and we did something similar for subview ops:
> 
> 1. I wonder if we need to actually use value bounds. It is common for canonicalizers + cse to resolve the dynamic values to the same SSA value, in which case the check can be trivial. If this is the case for your situation, you could try reusing some of the logic here to do the bounds checking, which will be more efficient: https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Codegen/Common/IREECodegenCanonicalizer.cpp#L26
> 2. If we do need the actual bounds checking, then IMO it is better to separate the bounds checking from the transformation patterns. This would be out of the scope of this PR, but it is worth adding a TODO. Over time we've been doing more complex index analysis, and I think we should have a better solution than calling value bounds or int range analysis in every pattern. I want to organize my thoughts into an issue, and then I can link it here (before or after this one lands).

For this specific case, we do need ValueBounds checking because data-tiling introduces affine expressions that can't be resolved by SSA tracking alone. For example:

```
%dim_ceildiv_128 = affine.apply affine_map<()[s0] -> (s0 ceildiv 128)>()[%dim]
%dim_aligned_128 = affine.apply affine_map<()[s0] -> ((s0 ceildiv 128) * 128)>()[%dim]
```

I've filed #23032 to track the discussion about unifying index analysis utilities across passes, and added a TODO comment in the code referencing the issue.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -105,6 +106,136 @@ struct FoldSubViewIntoMapScatter final : OpRewritePattern<MapScatterOp> {
...
+            delinearizeBases[delinIdx]);
+
+        // Check equality.
+        FailureOr<bool> areEqual =
+            ValueBoundsConstraintSet::areEqual(productVar, delinearizeVar);
+        if (succeeded(areEqual) && *areEqual) {
+          // Match found!
+          newLinearizeInfos.push_back(newLinearizeInfo);
+          break;
+        }
```
hanhanW: I'd delete the comments because the code is already self-documenting to me.

(Sorry, I wanted to add this comment after I clicked submit in the previous run; I forgot. :/

Yu-Zhewen: Done

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/DecomposeMapScatter.cpp
```diff
@@ -105,6 +106,134 @@ struct FoldSubViewIntoMapScatter final : OpRewritePattern<MapScatterOp> {
...
+                                PatternRewriter &rewriter) const override {
+    // Find the linearize op that produces the input to this delinearize.
+    auto linearizeOp = delinearizeOp.getLinearIndex()
+                           .getDefiningOp<affine::AffineLinearizeIndexOp>();
+    if (!linearizeOp) {
+      return rewriter.notifyMatchFailure(
+          delinearizeOp, "delinearize op does not consume a linearize op");
+    }
+    // We only handle disjoint linearizations.
+    if (!linearizeOp.getDisjoint()) {
```
krzysz00, line 143: Followup note: we can probably add something to prove `disjoint`ness

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_map_scatter.mlir
```diff
@@ -364,3 +364,67 @@ func.func @map_scatter_with_mask_on_inner_dim(
...
+//  PREPROCESSING-SAME:     %[[INPUT:[a-zA-Z0-9_]+]]: vector<1x128x208x2x8x16xf32>
+//  PREPROCESSING-SAME:     %[[OUTPUT:[a-zA-Z0-9_]+]]: memref<?x208x256xf32>
+//  PREPROCESSING-SAME:     %[[DIM:[a-zA-Z0-9_]+]]: index
+//       PREPROCESSING:   %[[TRUE:.+]] = arith.constant true
+//       PREPROCESSING:   %[[DIM_CEILDIV_128:.+]] = affine.apply {{.*}}()[%[[DIM]]]
+//       PREPROCESSING:   iree_linalg_ext.map_scatter %[[INPUT]] into %[[OUTPUT]] {
+//       PREPROCESSING:     ^bb0(%[[IDX0:.+]]: index, %[[IDX1:.+]]: index, %[[IDX2:.+]]: index, %[[IDX3:.+]]: index, %[[IDX4:.+]]: index, %[[IDX5:.+]]: index):
+//       PREPROCESSING:       %[[LIN_0:.+]] = affine.linearize_index disjoint [%[[IDX0]], %[[IDX1]]] by (%[[DIM_CEILDIV_128]], 128)
+//       PREPROCESSING:       %[[LIN_1:.+]] = affine.linearize_index disjoint [%[[IDX3]], %[[IDX4]], %[[IDX5]]] by (2, 8, 16)
+//       PREPROCESSING:       iree_linalg_ext.yield %[[LIN_0]], %[[IDX2]], %[[LIN_1]], %[[TRUE]]
```
krzysz00, line 430: Can  we get some tests where the pattern shouldn't fire?

krzysz00 (approved):
Code looks correct, approved given testing.

Once this lands here, could you put up an upstream version? I think folks outside of IREE might find it useful.

Yu-Zhewen:
> Code looks correct, approved given testing.
> 
> Once this lands here, could you put up an upstream version? I think folks outside of IREE might find it useful.

Apologies for the delay. The upstream version: https://github.com/llvm/llvm-project/pull/187245

PR #22959, opened 2025-12-23
[Codegen] Avoid unnecessary masking in map_scatter 

Yu-Zhewen:
When folding `extract_slice` operations into `map_scatter`, the current behavior is always adding masking logic, without checking whether the `extract_slice` is an identity. For static shapes, this is fine because later optimizations will remove the unnecessary masking. However, for dynamic shapes, the masking code persists and adds runtime overhead. This PR therefore avoids inserting unnecessary masking at the first place.

This PR also adjusts the priority of linalg::CopyOp in root op select, so that when identity map_scatter is canonicalized to a copy, the copy won't be incorrectly selected as the root operation over more important compute ops (e.g., inner_tiled).

This PR helps mitigate the VGPR spilling issue observed in https://github.com/iree-org/iree/issues/21865.

Depends on https://github.com/llvm/llvm-project/pull/173356.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
@@ -449,4 +449,3 @@ func.func @unpack_no_padding_no_masking(%dim : index, %result : memref<?x16384xf
 // DISPATCH-SCOPE-LABEL: func @unpack_no_padding_no_masking
 // DISPATCH-SCOPE: iree_linalg_ext.map_scatter
 // DISPATCH-SCOPE-NOT: arith.cmpi ult
-// DISPATCH-SCOPE: return
--- a/compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
@@ -432,3 +432,21 @@ func.func @consumer_unfusable_due_to_init(%arg0: tensor<?xi32>, %arg1: tensor<?x
 //       DISPATCH-SCOPE:     scf.forall.in_parallel
 //       DISPATCH-SCOPE:       tensor.parallel_insert_slice
 //       DISPATCH-SCOPE:   linalg.generic
+
+// -----
+
+// Test that unpack without padding, so map_scatter doesn't need masking.
+func.func @unpack_no_padding_no_masking(%dim : index, %result : memref<?x16384xf32>) {
+  %assumed = util.assume.int %dim<umin = 1024, umax = 16384, udiv = 128> : index
+  %workload = iree_tensor_ext.dispatch.workload.ordinal %assumed, 0 : index
+  %tiled_dim = affine.apply affine_map<()[s0] -> (s0 ceildiv 128)>()[%workload]
+  %source = tensor.empty(%tiled_dim) : tensor<?x64x128x256xf32>
+  %dest = tensor.empty(%workload) : tensor<?x16384xf32>
+  %unpack = linalg.unpack %source outer_dims_perm = [0, 1] inner_dims_pos = [0, 1] inner_tiles = [128, 256] into %dest : tensor<?x64x128x256xf32> -> tensor<?x16384xf32>
+  iree_codegen.store_to_buffer %unpack, %result : tensor<?x16384xf32> into memref<?x16384xf32>
+  return
+}
+// DISPATCH-SCOPE-LABEL: func @unpack_no_padding_no_masking
+// DISPATCH-SCOPE: iree_linalg_ext.map_scatter
+// DISPATCH-SCOPE-NOT: arith.cmpi ult
+// DISPATCH-SCOPE: return
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -765,7 +765,9 @@ bool MapScatterOp::isIdentity() {
   return true;
 }
 namespace {
-struct FoldTensorIdentityMapScatter
+/// Convert an identity map_scatter to a copy operation. We keep the copy to
+/// preserve DPS semantics.
+struct ConvertIdentityMapScatterToCopy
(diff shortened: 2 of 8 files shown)


compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
```diff
@@ -432,3 +432,21 @@ func.func @consumer_unfusable_due_to_init(%arg0: tensor<?xi32>, %arg1: tensor<?x
...
+  %source = tensor.empty(%tiled_dim) : tensor<?x64x128x256xf32>
+  %dest = tensor.empty(%workload) : tensor<?x16384xf32>
+  %unpack = linalg.unpack %source outer_dims_perm = [0, 1] inner_dims_pos = [0, 1] inner_tiles = [128, 256] into %dest : tensor<?x64x128x256xf32> -> tensor<?x16384xf32>
+  iree_codegen.store_to_buffer %unpack, %result : tensor<?x16384xf32> into memref<?x16384xf32>
+  return
+}
+// DISPATCH-SCOPE-LABEL: func @unpack_no_padding_no_masking
+// DISPATCH-SCOPE: iree_linalg_ext.map_scatter
+// DISPATCH-SCOPE-NOT: arith.cmpi ult
+// DISPATCH-SCOPE: return
```
hanhanW: nit: the `return` check is unnecessary.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -779,18 +779,16 @@ struct ConvertIdentityMapScatterToCopy
     if (!mapScatterOp.hasPureTensorSemantics()) {
       return failure();
     }
-    auto copyOp = linalg::CopyOp::create(rewriter, mapScatterOp.getLoc(),
-                                         mapScatterOp.getInput(),
-                                         mapScatterOp.getOutput());
-    rewriter.replaceOp(mapScatterOp, copyOp.getResults());
```
hanhanW, line 785: I don't remember why I did not catch this in my review, but it makes sense to me.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -765,7 +765,7 @@ bool MapScatterOp::isIdentity() {
   return true;
 }
 namespace {
-struct ConvertIdentityMapScatterToCopy
+struct FoldIdentityMapScatter
```
hanhanW: How about renaming it to `FoldTensorIdentityMapScatter` or `FoldIdentityMapScatterOnTensors`?

hanhanW (approved):
Overall looks good to me, just few nits.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
Max191: The reason this was creating a copy is to preserve DPS. IIRC, some tests were failing without doing this.

I believe the issue was that for a simple copy dispatch, the copy would be converted to map_scatter, and then get folded away, which creates a dispatch with no compute op. The newer pipelines could probably handle this, but this would go down one of the simple catchall pipelines (I forget which one), and it would cause some failures.

IIRC, we don't convert simple copy dispatches into map_scatter anymore, so that may be why these failures don't happen anymore.

I'm curious what the lowering strategy selection issue was, though? Usually, we can collapse the copy op into the surrounding ops in order to maintain DPS. Since this is in the configuration phase, I'd be a bit wary of breaking DPS, since some old pipelines still rely on it.

That said, if the change is necessary, and the CI is green, then I think it is fine. I'm mostly curious about what the issue was.

Max191 (commented):
This LGTM. I just left one comment + question about the copy folding.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
Yu-Zhewen: I get a case like this: `iree_codegen.inner_tiled → linalg.generic → linalg.copy`, and the copy gets selected as the root op, which causes issues as its configuration conflicts with the inner_tiled requirement (in terms of number of subgroups).

Could potentially add the copy op to the rejection list, but CI seems to be happy if we remove it at the beginning.
https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp#L2431-L2436

Max191: Ok, yeah, that is an issue in the lowering strategy selection. Adding the copy to that list should be fine, and then you can just add a case for the copy here, so it gets picked up similarly to fully parallel generic ops:

https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp#L2466-L2469

Adding linalg::CopyOp to the list of ops to generalize here would also work I think: 
https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Codegen/Common/GPU/GPUGeneralizeNamedOps.cpp#L61-L63

Adding it to that list will make it a fully parallel generic op when it gets to lowering strategy selection, and then there would be no need for special logic on linalg:CopyOp in KernelConfig.

Yu-Zhewen: Ok, I've reverted `ConvertIdentityMapScatterToCopy` back to producing a copy (to preserve DPS) and added a comment explaining this.

I'm going with the first approach, by adding `linalg::CopyOp` to the rejection list in KernelConfig (with a new lit test as well).

Also tried `GPUGeneralizeNamedOps`, but that doesn't work here due to pass ordering (map_scatter hasn't been canonicalized to copy yet when that pass runs).

Yu-Zhewen: @Max191 Merged this PR. Let me know if you have any follow-up comment

Issue #22919, opened 2025-12-16
[regression-onnx] compilation timeout for beit_large_patch16_224_Opset16


PhaneeshB:
### What happened?

Compilation time increased at this commit - `1cbf99570f`
from `0m57.585s` to `3m34.501s` resulting in a timeout during tests.
time timeout for tests is set at 100s

Last Working Version: 3.9.0rc20251103 
First Failing Version: 3.9.0rc20251104

---

Data gathered from command `time iree-compile ...`

`1cbf99570f [Codegen][GPU] Allow intentional padding for non-K-major matmul layouts (#22486)`
```
real    3m33.391s
user    3m34.501s
sys     0m4.056s
```
`20149ebd43 [GPU] Clean up misc issues in IREEGPUAttrs. NFC. (#22531)` (one commit before)
```
real    1m14.580s
user    1m14.096s
sys     0m4.897s
```
`bdf9393d9e (tag: iree-3.9.0rc20251103, tag: iree-3.9.0rc20251102) Bump LLVM to llvm/llvm-project@04f87c693c7e (#22515)` (working release tag)
```
real    0m57.515s
user    0m57.585s
sys     0m3.916s
```

### Steps to reproduce your issue

mlir file : `/shark-dev/iree-bug/22919_beit_large_patch16_224_Opset16/model.mlir` (path in SharkMi300X-4 machine) 
file >2GB, stored locally on the SharkMi300X-4

compile command
```
iree-compile model.torch_onnx.mlir \
--iree-hal-target-backends=rocm --iree-hip-target=gfx942 -o compiled_model.vmfb
```

nirvedhmeshram:
@PhaneeshB @pdhirajkumarprasad there is some initial triage to do here someone on the codegen team can do it when they get to it but just bringing it up here in case someone on your end has cycles to do this.
To see which dispatch(es) got bad from the change, I see two ways to do it,
option 1: you could use tracy and see if dispatch times got worse for a particular set of dispatches, the regression in compile time will usually have a related performance degradation too (which we can spot in tracy)
option 2 : you could dump executable-configurations of dispatches and search for "padding" attribute and report the dispatches for which the attribute changed. The flags to dump are here https://iree.dev/developers/general/developer-tips/#dumping-executable-files

Also for testing sake you can easily revert this commit so you dont necessarily need to work on an older branch.

pdhirajkumarprasad:
[...]

Issue #19546, opened 2024-12-20
Support accumulating GEMMs in TileAndFuse with intrinsic without needing c promotion


nirvedhmeshram:
Currently for accumulating GEMM we fail to bufferize if we dont do c promotion in TileAndFuse pipeline when using intrinsics. See dump [here](https://gist.github.com/nirvedhmeshram/92ec867f33a27de37bac05fe05e31803) . I know there are some tranforms that are still in development but I wasnt sure they will serve this case as well.

nirvedhmeshram:
Here is what is causing this to fail to bufferize, after `GPUFuseAndHoistParallelLoopsPass`
We have the following access
```
%read_write_input = flow.dispatch.tensor.load  ... -> tensor<32x16x32x16xi32>
%workgroup_scf_forall = scf.forall ...  shared_outs(%arg2 = %read_write_input ) -> (tensor<32x16x32x16xi32>) {
  %11 = tensor.empty() : tensor<4x16x4x16xi32>
  %subgroup_scf_forall = scf.forall ... shared_outs(%arg5 = %11)  -> (tensor<4x16x4x16xi32>) {
    %extracted_slice = tensor.extract_slice %arg2
    %extracted_slice_0 = tensor.extract_slice %arg5
      %thread_scf_forall = scf.forall ...  shared_outs(%arg7 = %extracted_slice_0) -> (tensor<2x16x2x16xi32>) {
      %acc_input = tensor.extract_slice %extracted_slice
      %some_intrinsic_compute (...,  %acc_input )
       %acc_output= tensor.extract_slice %arg7
  ```
The point being that the accumlator input slice was derived from `%read_write_input` but accumlator  ouput is written back to a slice derived from an empty tensor.

Later after `EliminateEmptyTensorsPass`
we are left with almost the same access pattern except that we have
```
  %extracted_slice = tensor.extract_slice %arg2
  %subgroup_scf_forall = scf.forall ... shared_outs(%arg5 = %extracted_slice)  -> (tensor<4x16x4x16xi32>) {
```
[...]

PR #22975, opened 2025-12-23
[Codegen] Fix ConvertAccGemmToGemm to handle read-write arguments correctly.

MaheshRavishankar:
The current pass unconditionally converts from an accumulating GEMM to a non-accumulating GEMM. This transformation is only required when the `outs` arguments is coming from a read-only buffer. When coming from a read-write buffer, the accumulating gemm can be handled as is, and it does not need to converted to a non-accumulating gemm.

For a function input like

```
func.func @acc_gemm(%lhs : tensor<?x?xf32>, %rhs: tensor<?x?xf32>,
    %init : tensor<?x?xf32> {iree.abi.output = 0}) -> tensor<?x?xf32> {
  %0 = linalg.matmul ins(%lhs, %rhs : tensor<?x?xf32>, tensor<?x?xf32>)
      outs(%init : tensor<?x?xf32>) -> tensor<?x?xf32>
  return %0 : tensor<?x?xf32>
}
```

The dispatch sees an single `init` binding that is read-write. In those cases we dont need to convert the `linalg.matmul` into a non-accumulating GEMM. This case can be (and is currently) handled natively.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertAccGEMMToGEMMPass.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertAccGEMMToGEMMPass.cpp
@@ -31,6 +31,36 @@ namespace mlir::iree_compiler {
 #define GEN_PASS_DEF_CONVERTACCGEMMTOGEMMPASS
 #include "iree/compiler/Codegen/Common/Passes.h.inc"
 
+// Get the `iree_tensor_ext.dispatch.tensor.load` that this value is
+// populated with. This could potentially walk the use-def chain to get the load
+// operation, but for now it just returns the load op if that is the defining
+// operation for `v`.
+std::optional<IREE::TensorExt::DispatchTensorLoadOp>
+getDispatchTensorLoadOp(Value v) {
+  auto loadOp = v.getDefiningOp<IREE::TensorExt::DispatchTensorLoadOp>();
+  if (loadOp) {
+    return loadOp;
+  }
+  return std::nullopt;
+}
+
+// Get the `iree_tensor_ext.dispatch.tensor.store` that this value is
+// populated writes to. This could potentially walk the use-def chain of DPS
+// init operands to get the store operation, but for now it just returns the
+// store op if the result has a singe use and that use is the store op.
+std::optional<IREE::TensorExt::DispatchTensorStoreOp>
+getDispatchTensorStoreOp(Value v) {
+  if (v.getNumUses() != 1) {
+    return std::nullopt;
+  }
+  auto storeOp =
+      dyn_cast<IREE::TensorExt::DispatchTensorStoreOp>(*(v.getUsers().begin()));
+  if (storeOp) {
+    return storeOp;
+  }
+  return std::nullopt;
+}
+
 static bool accGemmToGemmPrecondition(Operation *op) {
   if (auto innerTiledOp = dyn_cast<IREE::Codegen::InnerTiledOp>(op)) {
     return isa<IREE::GPU::MmaInterfaceAttr, IREE::GPU::ScaledMMAAttr,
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/ConvertAccGEMMToGEMMPass.cpp
```diff
@@ -31,6 +31,36 @@ namespace mlir::iree_compiler {
...
+  if (loadOp) {
+    return loadOp;
+  }
+  return std::nullopt;
+}
+
+// Get the `iree_tensor_ext.dispatch.tensor.store` that this value is
+// populated writes to. This could potentially walk the use-def chain of DPS
+// init operands to get the store operation, but for now it just returns the
+// store op if the result has a singe use and that use is the store op.
```
jtuyls:
```suggestion
// store op if the result has a single use and that use is the store op.
```

compiler/src/iree/compiler/Codegen/Common/ConvertAccGEMMToGEMMPass.cpp
```diff
@@ -48,18 +78,51 @@ static bool accGemmToGemmPrecondition(Operation *op) {
...
+  // This currently would only work for very simple cases, but could be
+  // generalized further.
+  OpOperand *initValue = linalgOp.getDpsInitOperand(0);
+  std::optional<IREE::TensorExt::DispatchTensorLoadOp> initLoadOp =
+      getDispatchTensorLoadOp(initValue->get());
+  std::optional<IREE::TensorExt::DispatchTensorStoreOp> resultStoreOp =
+      getDispatchTensorStoreOp(linalgOp->getResult(0));
+  if (initLoadOp && resultStoreOp && initLoadOp->getSource() &&
+      resultStoreOp->getTarget()) {
+    // Check that the source and the result have a read/write tag. If they dont
```
jtuyls:
```suggestion
    // Check that the source and the result have a read/write tag. If they don't
```

jtuyls (approved):
LGTM

nirvedhmeshram:
I would like to verify this issue is not again caused by this PR https://github.com/iree-org/iree/issues/19546
I will update once I do a check, please hold on landing until then.

nirvedhmeshram (approved):
I will note though in practice with the {iree.abi.output = 0}) I am still seeing a readonly tensor and this pass is still making the elementwise, ideally I would have liked to be able to generate the readwrite tensor and see how the underlying codegen is handling it.
[here](https://gist.github.com/nirvedhmeshram/505cad9df432de13b24caede269f9b68) is the IR dump.

MaheshRavishankar:
> I will note though in practice with the {iree.abi.output = 0}) I am still seeing a readonly tensor and this pass is still making the elementwise, ideally I would have liked to be able to generate the readwrite tensor and see how the underlying codegen is handling it.
> [here](https://gist.github.com/nirvedhmeshram/505cad9df432de13b24caede269f9b68) is the IR dump.

I'll take a look. I tried a matmul and it was working as expected.

nirvedhmeshram (approved):
seems like the
```
/ -----// IR Dump After GPUCombineLayoutTransformationPass (iree-codegen-gpu-combine-layout-transformation) //----- //
```
pass is turning the readwrite tensor to read tensor, but I turned it off (with --iree-llvmgpu-test-combine-layout-transformation=false) and can confirm things workout for the shape 
but I tried a shape that does need padding and there is an issue, with a large private allocation see gist [here](https://gist.github.com/nirvedhmeshram/0d86df2e76f1cba2fc2c4d1b37416f7c)

However, turning it into a elementwise add also has the same issue as documented in https://github.com/iree-org/iree/issues/22919 so that is an unrelated problem.

MaheshRavishankar:
> seems like the
> 
> ```
> / -----// IR Dump After GPUCombineLayoutTransformationPass (iree-codegen-gpu-combine-layout-transformation) //----- //
> ```
> 
> pass is turning the readwrite tensor to read tensor, but I turned it off (with --iree-llvmgpu-test-combine-layout-transformation=false) and can confirm things workout for the shape but I tried a shape that does need padding and there is an issue, with a large private allocation see gist [here](https://gist.github.com/nirvedhmeshram/0d86df2e76f1cba2fc2c4d1b37416f7c)
> 
> However, turning it into a elementwise add also has the same issue as documented in #22919 so that is an unrelated problem.

Thanks @nirvedhmeshram . I looked further and adapted the pass to handle `iree_codegen.load_from_buffer` and `iree_codegen.store_to_buffer`. Hopefully that makes these cases easier. There is still a private allocation being created that needs padding. I think this is indeed a pre-existing problem, but hopefully this makes it easier to handle.

PR #22981, opened 2025-12-24
[ROCM][DT] Add mxfp4 pingpong ukernel

### LINKED ISSUE #21938: [GPU][DT] Implement mxfp4 MLIR ukernel

Yu-Zhewen:
This PR adds a new data-tiled scaled matmul ukernel for MXFP4 format using a pingpong (double buffering) strategy on gfx950. It uses tile size M=128, N=256, K=256 with the MFMA_SCALE_F32_16x16x128_B32 intrinsic.

The ukernel matching logic is updated to accommodate scaled matmul, and an e2e test is also added.

**Benchmark results (Llama 70B):**
|                        | ASM | Codegen | Data-tiling | Data-tiling+UKernel |
|------------------------|-----|---------|-------------|---------------------|
| Total f4gemm time (ms) | **254** | 627     | 577         | **394**                 |
| Total e2e time (ms)    | 742 | 990     | 1016        | 843                 |

The ukernel achieves ~64% of ASM performance (254/394). The remaining gap is mainly due to ASM's finer-grained unstructured scheduling that better hides memory latency.

Closes: #21938


Yu-Zhewen:
Note: As a separate issue, data-tiling shows slightly worse e2e time than codegen (1016 vs 990), because some encoding dispatches are not yet fused. I'm going to have a look at this.

image

compiler/plugins/target/ROCM/Dialect/ROCM/IR/ROCMAttrs.cpp
```diff
@@ -192,19 +199,27 @@ Attribute TensorUKernelProviderAttr::getDataLayoutForUKernel(
     // Depending on the type of data-tiled layout attribute, read the
     // appropriate kind of MMA-like intrinsic and check that it's supported by
     // the target.
-    if (auto dtMma = dyn_cast<GPU::DataTiledMMAAttr>(mma)) {
+    if (opType == IREE::Encoding::EncodingOpType::matmul) {
+      auto dtMma = dyn_cast<GPU::DataTiledMMAAttr>(mma);
+      if (!dtMma) {
+        continue;
+      }
```
jtuyls, line 206: Can't this just be a cast? It shouldn't happen that this attribute is not a `GPU::DataTiledMMAAttr>`?

compiler/plugins/target/ROCM/Dialect/ROCM/Transforms/ApplyBuiltinPDLPatterns.cpp
```diff
@@ -15,11 +15,13 @@
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
 #include "iree/compiler/Dialect/HAL/IR/HALDialect.h"
 #include "iree/compiler/Dialect/HAL/IR/HALTypes.h"
+#include "iree/compiler/Dialect/LinalgExt/Utils/MatchUtils.h"
 #include "iree/compiler/Dialect/Util/IR/UtilDialect.h"
 #include "iree/compiler/Utils/ShapeUtils.h"
 #include "llvm/ADT/SmallVectorExtras.h"
 #include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/Regex.h"
+#include "mlir/Dialect/AMDGPU/IR/AMDGPUDialect.h"
```
jtuyls, line 24: Do we need this?

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
```diff
@@ -315,8 +316,9 @@ struct MaterializeDeviceEncodingPass final
   using Base::Base;
 
   void getDependentDialects(DialectRegistry &registry) const override {
-    registry.insert<arith::ArithDialect, tensor::TensorDialect,
-                    vector::VectorDialect, IREE::Encoding::IREEEncodingDialect,
+    registry.insert<amdgpu::AMDGPUDialect, arith::ArithDialect,
```
jtuyls, line 319: Why do we need AMDGPUDialect?

jtuyls (commented):
Very nice!

bjacob (approved):

compiler/plugins/target/ROCM/Dialect/ROCM/IR/ROCMAttrs.cpp
Yu-Zhewen: The opType (line 202) comes from the encoding, while mma (line 195) comes from the ukernel info. Since we iterate over all ukernels, we will also encounter `DataTiledScaledMMAAttr` here, so we need to filter it out.

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
Yu-Zhewen: Yeah, this one is not needed. Forgot to clean it up. Removed now.

compiler/plugins/target/ROCM/Dialect/ROCM/Transforms/ApplyBuiltinPDLPatterns.cpp
Yu-Zhewen: Yes, we need it here. Same reason as the line below (`#include "mlir/Dialect/LLVMIR/ROCDLDialect.h"`), as the dialect is used inside ukernel. In this case, the new ukernel uses `amdgpu.gather_to_lds`.

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
Yu-Zhewen: Emmm, I'm wrong. This is also needed.

https://github.com/iree-org/iree/actions/runs/20789253316/job/59707141188?pr=22981

jtuyls (approved):
LGTM

compiler/plugins/target/ROCM/Dialect/ROCM/IR/ROCMAttrs.cpp
Yu-Zhewen: Also I refreshed the comments a bit to make it more clear.

jtuyls: Sg, thanks!

compiler/plugins/target/ROCM/builtins/mlir_ukernel/iree_uk_amdgpu_dt_scaled_matmul_f4E2M1FN.mlir
```diff
@@ -0,0 +1,322 @@
...
+
+        %rhs_scale_shared_num = arith.muli %buffer_num, %c2 : index
+        %rhs_scale_shared_outer = arith.addi %rhs_scale_shared_num, %subgroups#1 : index
+        %rhs_scale_shared_inner = arith.muli %ids#1, %c16 : index
+
+        amdgpu.gather_to_lds %rhs_scale_collapse[%i, %rhs_scale_buffer_inner], %rhs_scale_shared[%rhs_scale_shared_outer, %rhs_scale_shared_inner]
+          : !rhs_scale_copy_vec_ty, !rhs_scale_buffer_collapse_ty, !rhs_scale_shared_ty
+
+        // There are (8 + 1) gather_to_lds instructions above.
+        rocdl.s.waitcnt 9
```
krzysz00: ... I'm not sure this does what you think it does, and it _definitely_ waits for exports to complete for no reason. Please use the AMDGPU dialect wrapper here and everywhere.

krzysz00 (commented):
One note

compiler/plugins/target/ROCM/builtins/mlir_ukernel/iree_uk_amdgpu_dt_scaled_matmul_f4E2M1FN.mlir
Yu-Zhewen: Thanks, now changed to `amdgpu.memory_counter_wait load(9)`

compiler/src/iree/compiler/Codegen/Common/MaterializeEncoding.cpp
hanhanW: Why? Are we creating any AMDGPU op during the materialization?

https://mlir.llvm.org/docs/PassManagement/#dependent-dialects

> Dialects must be loaded in the MLIRContext before entities from these dialects (operations, types, attributes, …) can be created.

Yu-Zhewen: No, it's only because we need parse MLIR ukernel files for data layouts during materialization. I think the ROCM dialect is the right place for all these kind of dependencies. I may address it in a follow-up PR if that's ok, since there are other similar places that need cleanup as well.

hanhanW: Sure, I was just curious.

PR #22995, opened 2025-12-30
[AMDGPU][LDS] Unify DMA transfer linearization based on destination contiguity

lialan:
This PR refactors the `AMDGPULowerCoalescedDMAToGatherLDS` pass to use an unified and optimized approach for DMA transfers based on destination memref contiguity. Previously, the pass treated transfers row-by-row. 

This PR extends the transfer to cover multiple inner dimensions if it satisfy the contiguity condition. By doing so we can utilize larger DMA transfer sizes which will not be feasible previously.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -427,7 +427,7 @@ struct AMDGPULowerCoalescedDMAToGatherLDSPass final
 
     // dma_sizes is optional - if not specified, skip the size validation.
     ArrayRef<int64_t> dmaSizes;
-    if (auto dmaSizesAttr = target.getWgp().getDmaSizes()) {
+    if (DenseI64ArrayAttr dmaSizesAttr = target.getWgp().getDmaSizes()) {
       dmaSizes = dmaSizesAttr.asArrayRef();
     }
 
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -67,13 +67,180 @@ verifyMemoryLayoutContiguous(IREE::GPU::CoalescedGatherDMAOp dmaOp,
   return success();
 }
 
-// A segment of the innermost dimension that will be transferred
-// using a specific DMA size.
+/// Returns true if the memref type is fully contiguous, i.e., there are no
+/// gaps between elements across all dimensions. A memref is fully contiguous
+/// when the strides form a row-major layout where each stride equals the
+/// product of all inner dimension sizes.
+static bool isFullyContiguousMemRef(MemRefType memrefType) {
+  // A memref is fully contiguous if all dimensions are contiguous.
+  return memrefType.getNumContiguousTrailingDims() == memrefType.getRank();
+}
+
+/// Converts a linear index to multi-dimensional indices for a given shape.
+/// Uses row-major (C-style) layout where the last dimension varies fastest.
+///
+/// Example: shape = [4, 3, 2], linearIdx = 7
+///   strides = [6, 2, 1]
+///   indices[0] = 7 / 6 = 1, remainder = 7 % 6 = 1
+///   indices[1] = 1 / 2 = 0, remainder = 1 % 2 = 1
+///   indices[2] = 1 / 1 = 1
+///   Result: [1, 0, 1]
+static SmallVector<int64_t> linearToMultiIndex(int64_t linearIdx,
+                                               ArrayRef<int64_t> shape) {
+  SmallVector<int64_t> indices(shape.size());
+  for (int64_t i = shape.size() - 1; i >= 0; --i) {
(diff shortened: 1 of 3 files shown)


krzysz00:
I don't think this needs to be a binary "do we flatten into the linear case or no" decision - you can query the number of contiguous dimensions in the destination (capped at the first source dimension that has a gather index) and treat those as one big linear dimension in your tile offset code (where then there's a `delinearize` to get the offsets into source dimensions).

lialan:
> I don't think this needs to be a binary "do we flatten into the linear case or no" decision - you can query the number of contiguous dimensions in the destination (capped at the first source dimension that has a gather index) and treat those as one big linear dimension in your tile offset code (where then there's a `delinearize` to get the offsets into source dimensions).

Ahh, that is a good idea!

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -67,14 +67,136 @@ verifyMemoryLayoutContiguous(IREE::GPU::CoalescedGatherDMAOp dmaOp,
...
+  SmallVector<int64_t> sortedDmaSizes(dmaSizes);
+  llvm::sort(sortedDmaSizes, std::greater<>());
+
+  SmallVector<TransferSegment> segments;
+  int64_t remainingElements = totalElements;
+  int64_t currentOffset = 0;
+
+  for (int64_t dmaSize : sortedDmaSizes) {
+    // Calculate elements per lane for this DMA size.
+    if (dmaSize % elementBits != 0)
```
qedawkins: nit: missing braces

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -67,14 +67,136 @@ verifyMemoryLayoutContiguous(IREE::GPU::CoalescedGatherDMAOp dmaOp,
...
+             << " transfers, elementsPerLane=" << elementsPerLane
+             << ", startOffset=" << currentOffset
+             << ", elementsPerTransfer=" << elementsPerTransfer;
+
+      int64_t coveredElements = numTransfers * elementsPerTransfer;
+      remainingElements -= coveredElements;
+      currentOffset += coveredElements;
+    }
+
+    if (remainingElements == 0)
```
qedawkins: same here

qedawkins (approved):
The new output looks much more general, nice!

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -173,6 +185,55 @@ func.func @lower_coalesced_copy_dma_1d(
...
+}
+
+// -----
+
+#executable_target_rocm_hsaco_fb = #hal.executable.target<"rocm",
+  "rocm-hsaco-fb", {iree_codegen.target_info = #iree_gpu.target<
+  arch = "gfx950", features = "", wgp = <
+    compute = fp64|fp32|fp16|int64|int32|int16|int8,
+    storage = b64|b32|b16|b8, subgroup = shuffle|arithmetic,
+    dot = dp4xi8toi32, mma = [], subgroup_size_choices = [32, 32],
```
kuhar: Can we trim these down? For example, do these tests care about all the compute types/subgroup operations/dot product ops, or would `dot = none` work?

kuhar: Same elsewhere

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -126,74 +249,51 @@ struct LowerCoalescedGatherDMAPattern final
...
+    int64_t numLinearDims = destType.getNumContiguousTrailingDims();
+    // Always linearize at least the innermost dimension.
+    numLinearDims = std::max<int64_t>(numLinearDims, 1);
+    LDBG() << "  Number of linear dims: " << numLinearDims;
+
+    // Compute total elements in the linearized portion (last numLinearDims
+    // dims).
+    int64_t linearSize = 1;
+    for (int64_t i = destRank - numLinearDims; i < destRank; ++i) {
+      linearSize *= destShape[i];
```
krzysz00, line 284: I think you'll need to check if the dimensions in question have a static size as well?

krzysz00: ..., or is that already guaranteed by the verifier because you're relying on the staticness of the destination shape a bunch below too

krzysz00 (approved):
Overall, this change makes sense to me and we should go for it

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
lialan: Good catch! We should ensure static dims for linearized loading. added a check for that.

PR #23001, opened 2025-12-31
[CI] Allow macos checks to be triggered with ci-extra

kuhar:
This makes it easier to see if these are working, without having to push changes to a branch under iree-org/iree. We already have support for this for windows_x64_msvc and torch_ops.

ci-extra: macos_arm64_clang, macos_x64_clang

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,6 +247,18 @@ jobs:
     uses: ./.github/workflows/ci_linux_x64_clang_ubsan.yml
     secrets: inherit
 
+  macos_arm64_clang:
+    needs: setup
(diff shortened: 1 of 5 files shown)


kuhar:
My understanding is that the new jobs are not being picked up yet in this PR because github actions read the ci.yaml definition from main, so this will start working once merged.

kuhar:
The list of enabled jobs looks correct: https://github.com/iree-org/iree/actions/runs/20622038524/job/59225723539?pr=23001#step:5:7

amd-eochoalo (approved):
I think this looks good, maybe we can have a test run after it merges?

PR #23004, opened 2025-12-31
[CI] Drop macos job concurrency block

kuhar:
Remove `concurrency` block from macOS workflows - when called as reusable workflows, the `github.workflow` context inherits the caller's workflow name ("CI"), creating the same concurrency group as the parent workflow and causing the job to be skipped. Note that other optional checks like windows_x64_msvc don't use it either.

### DIFF
--- a/.github/workflows/ci_macos_arm64_clang.yml
+++ b/.github/workflows/ci_macos_arm64_clang.yml
@@ -13,16 +13,9 @@ on:
   workflow_call:
   workflow_dispatch:
 
-concurrency:
-  # A PR number if a pull request and otherwise the commit hash. This cancels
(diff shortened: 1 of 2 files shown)


kuhar:
Successful run here: https://github.com/iree-org/iree/actions/runs/20623632970

amd-eochoalo (commented):
I'm not sure why this concurrency is set for default jobs but needs to be deleted for non-default jobs. Aren't both groups reusable workflows?

kuhar:
I'm not sure what the exact semantics are, but this makes it so that macos jobs are never picked up. I think the default action behavior already handles cancellation properly.

amd-eochoalo (approved):
Got it. I think the default actions are not reusable workflows? They do not have `workflow_call`.

PR #23008, opened 2026-01-02
[CI] Upgrade to GCC 11 to fix stablehlo template instantiation error

Yu-Zhewen:
The GCC CI workflow was failing to build stablehlo due to a C++ template instantiation issue with GCC 9.

[StablehloAggressiveFolder.cpp](https://github.com/openxla/stablehlo/blob/4a6ea0d65a0ed97f91a04276fcda88fc516ae6de/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp#L499-L507) uses modern C++ patterns combining if constexpr with SFINAE-based type traits:
```
if constexpr (fold_unary::DirectFolderExists<Impl, APInt>::value) {
    return FoldIfImplemented<APInt>(operand); 
}
```

GCC 9 incorrectly evaluates these type traits eagerly before the if constexpr condition is resolved, causing "incomplete type" errors:
```
/__w/iree/iree/third_party/stablehlo/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp:501:48: error: no matching function for call to 'FoldIfImplemented<llvm::APInt>(llvm::APInt&)'

/__w/iree/iree/third_party/stablehlo/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp:464:78: error: incomplete type 'fold_unary::DirectFolderExists<Impl, llvm::APInt>' used in nested name specifier
```

Upgrade to GCC 11 resolves this problem.

Before: https://github.com/iree-org/iree/actions/runs/20654878139/job/59305852621

After: https://github.com/iree-org/iree/actions/runs/20660113060/job/59320617905?pr=23008

### DIFF
--- a/.github/workflows/ci_linux_x64_gcc.yml
+++ b/.github/workflows/ci_linux_x64_gcc.yml
@@ -27,7 +27,7 @@ jobs:
   linux_x64_gcc:
     if: ${{ github.repository_owner == 'iree-org' || github.event_name != 'schedule' }}
     runs-on: ubuntu-24.04
-    container: ghcr.io/iree-org/cpubuilder_ubuntu_jammy@sha256:78a558b999b230f7e1da376639e14b44f095f30f1777d6a272ba48c0bbdd4ccb
+    container: ghcr.io/iree-org/cpubuilder_ubuntu_jammy@sha256:dfaf99be6f3adf3d7e404ca8e41e545f5353ab694037f627d2170f49f688504f
     defaults:
       run:
         shell: bash
--- a/.github/workflows/ci_linux_x64_gcc.yml
+++ b/.github/workflows/ci_linux_x64_gcc.yml
@@ -40,8 +40,8 @@ jobs:
           submodules: true
       - name: "Building IREE with gcc"
         env:
-          CC: /usr/bin/gcc-9
-          CXX: /usr/bin/g++-9
+          CC: gcc-11
+          CXX: g++-11
           CMAKE_BUILD_TYPE: Release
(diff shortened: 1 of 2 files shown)


kuhar (approved):
Version bump sgtm, gcc11 came out almost 5 years ago.

Is there any documentation we maintain about supported tool chains that also needs updating?

Yu-Zhewen:
I found two places referencing GCC versions:

1. [docs/website/docs/developers/debugging/releases.md](https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/docs/website/docs/developers/debugging/releases.md?plain=1#L44-L48) mentions gcc 9.3.1 for manylinux, but I believe releases are actually using clang-17 now (see this recent [run](https://github.com/iree-org/iree/actions/runs/20688054596/job/59392088631)). This doc appears to be outdated.

2. [cpubuilder_ubuntu_jammy.Dockerfile](https://github.com/iree-org/base-docker-images/blob/187e081dcd9baa3f6357120f5ac4626bf4b202d7/dockerfiles/cpubuilder_ubuntu_jammy.Dockerfile#L18), the Docker image used by this GCC CI installs gcc-9. We could update it to gcc-11, or simply remove it since Ubuntu Jammy provides gcc-11 by default.

I can address these in separate PRs.

kuhar:
I think all of these should be updated together to keep a consistent state on gcc

Yu-Zhewen:
I've updated the documentation in this PR and created https://github.com/iree-org/base-docker-images/pull/39 for the Docker image change (separate repo). Once that PR is merged, I'll update the container SHA here to reference the new image.

PR #23011, opened 2026-01-02
[AMDGPU][LDS] Prefer contiguous subviews in `GPUConvertToCoalescedDMA`

lialan:
* When tiling for coalesced DMA operations, prefer keeping the innermost dimension whole to ensure contiguous memory access patterns.
* Redistributes warps to outer dimensions when default tiling would split the innermost dimension.
* Prevents generation of non-contiguous `memref.subview` operations with inefficient strided access.

## Example:
Previously, the pass would tile based on workgroup dimensions without considering memory contiguity. For a `64x128` tensor with 4 warps, default tiling would create `64x64` tiles, resulting in a strided subview:
```
memref.subview ... [64, 64] [1, 1] : memref<64x128xf32> to memref<64x64xf32, strided<[128, 1]>>
```

Instead, detect when tiling would split the innermost dimension and instead redistribute warps to outer dimensions:
Before: step (64, 64) → 64x64 non-contiguous subviews
After: step (16, 128) → 16x128 contiguous subviews
This ensures LDS subviews maintain the full innermost dimension for efficient coalesced memory access.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -642,34 +642,94 @@ private:
 
     // Compute tile sizes: divide the shape by number of warps.
     // This distributes the work across warps in each dimension.
+    //
+    // We prefer to keep the innermost dimension whole (not tiled) to ensure
+    // contiguous memory access patterns. If tiling the innermost dimension
+    // would create non-contiguous subviews, we redistribute warps to outer
+    // dimensions instead.
     SmallVector<OpFoldResult> tileSizes;
     int64_t numTiledDims = 0;
-    for (int64_t i = 0; i < rank; ++i) {
-      // Map dimensions: innermost dims map to innermost workgroup dims.
-      // For 2D: dim 0 (rows) -> wgSize[1], dim 1 (cols) -> wgSize[0].
-      int64_t warpDim =
-          (rank - 1 - i) < numWarps.size() ? numWarps[rank - 1 - i] : 1;
 
-      bool isInnermostDim = (i == rank - 1);
+    // Calculate total number of warps available.
+    // Note: numWarps may contain 0s for dimensions where wgSize < subgroupSize.
+    // We treat 0 as 1 for the purpose of counting total warps.
+    int64_t totalWarps = 1;
+    for (int64_t nw : numWarps) {
+      if (nw > 0) {
+        totalWarps *= nw;
+      }
+    }
 
-      // For innermost dimension: always tile if we need thread-level
-      // distribution, ensuring tile size is at least subgroup_size (but not
-      // exceeding the dimension size).
-      if (isInnermostDim && ShapedType::isStatic(shape[i])) {
-        int64_t tileSize =
-            (warpDim > 1) ? llvm::divideCeil(shape[i], warpDim) : shape[i];
-        // Ensure tile size is at least subgroup_size, but cap at dimension
-        // size.
-        tileSize = std::clamp(tileSize, *subgroupSize, shape[i]);
-        tileSizes.push_back(rewriter.getIndexAttr(tileSize));
-        ++numTiledDims;
-      } else if (warpDim > 1 && ShapedType::isStatic(shape[i])) {
-        // For other dimensions: only tile if warpDim > 1.
-        int64_t tileSize = llvm::divideCeil(shape[i], warpDim);
-        tileSizes.push_back(rewriter.getIndexAttr(tileSize));
-        ++numTiledDims;
-      } else {
-        tileSizes.push_back(rewriter.getIndexAttr(0));
+    // Check if default tiling would split the innermost dimension.
+    // Default mapping: innermost tensor dim → outermost workgroup dim.
+    int64_t innermostWarpDim = numWarps[0];
+    int64_t innermostDimSize = shape[rank - 1];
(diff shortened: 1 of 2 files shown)


lialan:
This should work together with https://github.com/iree-org/iree/pull/22995 so both lowering passes can maximize memory contiguity, and use large bitwidth loading.

krzysz00 (commented):
Trivial question - does this change only apply to the configuration for the copy or does it also change how the computation is tiled? If it's the latter, is there a way to avoid that?

lialan:
> Trivial question - does this change only apply to the configuration for the copy or does it also change how the computation is tiled? If it's the latter, is there a way to avoid that?

Good question, this change only applies to the copy/DMA configuration, not to computation tiling, so we are safe.

qedawkins (commented):
What is "default tiling?" Is there a reason we need to first infer *what* default tiling will pick and conditionally override? Why can't we always use the newly added logic here?

lialan:
> What is "default tiling?" Is there a reason we need to first infer _what_ default tiling will pick and conditionally override? Why can't we always use the newly added logic here?

We could have always used the new logic here; the default tiling was the old, naive way of doing tiling; there are no reason to to need to first infer the default tiling.

So this is purely an improvement. @qedawkins

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -560,6 +560,112 @@ struct GPUConvertToCoalescedDMAPass final
...
+          int64_t tileSize = llvm::divideCeil(shape[i], warpsForThisDim);
+          tileSizes.push_back(rewriter.getIndexAttr(tileSize));
+          // Update remaining parallelism for subsequent dimensions.
+          remainingWarps = llvm::divideCeil(remainingWarps, warpsForThisDim);
+          ++numTiledDims;
+        } else {
+          tileSizes.push_back(rewriter.getIndexAttr(0));
+        }
+      }
+    } else {
```
qedawkins: Do we need this else branch then? Can we just always take the `if`?

lialan: You are right, actually we just need the if part, and use the greedy approach, and then we are good.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -560,6 +560,54 @@ struct GPUConvertToCoalescedDMAPass final
...
+    for (int64_t nw : numWarps) {
+      if (nw > 0) {
+        totalWarps *= nw;
+      }
+    }
+
+    // Greedily distribute warps to outer dimensions, keeping innermost whole.
+    int64_t remainingWarps = totalWarps;
+    for (int64_t i = 0; i < rank; ++i) {
+      bool isInnermostDim = (i == rank - 1);
```
qedawkins: This is only used once AFAICS, just make it `if (isInnerMostDim)` and drop the var.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -560,6 +560,54 @@ struct GPUConvertToCoalescedDMAPass final
...
+                           ArrayRef<int64_t> numWarps) {
+    SmallVector<OpFoldResult> tileSizes;
+    int64_t numTiledDims = 0;
+    int64_t rank = shape.size();
+
+    // Calculate total number of warps available.
+    // Note: numWarps may contain 0s for dimensions where wgSize < subgroupSize.
+    // We treat 0 as 1 for the purpose of counting total warps.
+    int64_t totalWarps = 1;
+    for (int64_t nw : numWarps) {
```
qedawkins: nit/optional: You can use llvm::make_filter_range for this
```
llvm::accumulate(llvm::make_filter_range(numWarps, [](int64_t i) { return i > 0; }), 1l, std::multiplies<int64_t>());
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -560,6 +560,54 @@ struct GPUConvertToCoalescedDMAPass final
...
+        ++numTiledDims;
+      } else if (remainingWarps > 1 && ShapedType::isStatic(shape[i])) {
+        // Distribute remaining warps to this outer dimension.
+        int64_t warpsForThisDim = std::min(remainingWarps, shape[i]);
+        int64_t tileSize = llvm::divideCeil(shape[i], warpsForThisDim);
+        tileSizes.push_back(rewriter.getIndexAttr(tileSize));
+        // Update remaining parallelism for subsequent dimensions.
+        remainingWarps = llvm::divideCeil(remainingWarps, warpsForThisDim);
+        ++numTiledDims;
+      } else {
```
qedawkins, line 598: Can you add a short comment explaining what's happening in this branch.

qedawkins (approved):

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
kuhar: also use `llvm::product_of`

Issue #23005, opened 2025-12-31
[CodeGen][AMDGPU] ArgCompareOp VectorDistribute Pipeline Support


bangtianliu:
## Overview

This issue tracks the implementation of full VectorDistribute pipeline support for `ArgCompareOp` (argmax/argmin operations) on AMDGPU targets. The goal is to enable efficient GPU code generation for reduction operations that return both a selected value and its corresponding index.

### ArgCompareOp Overview

`ArgCompareOp` is defined in [LinalgExtOps.td:645-770](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td#L645-L770) and performs:
- A reduction over a specified dimension of a tensor
- Returns **two outputs**: the selected value AND its corresponding index
- Uses a **user-defined comparator region** that receives two values and returns `i1`

**Key Design Feature**: The comparator region provides flexibility to express:
- **argmax**: `arith.cmpf ogt, %a, %b` (greater than)
- **argmin**: `arith.cmpf olt, %a, %b` (less than)
- **Custom logic**: Any boolean predicate comparing two values

```mlir
// Example: argmax (select larger value)
iree_linalg_ext.arg_compare dimension(1)
  ins(%input : tensor<2x10xf32>)
  outs(%out_val, %out_idx : tensor<2xf32>, tensor<2xi32>) {
^bb0(%a: f32, %b: f32):
  %cmp = arith.cmpf ogt, %a, %b : f32
  iree_linalg_ext.yield %cmp : i1
}

// Example: argmin (select smaller value)
iree_linalg_ext.arg_compare dimension(1)
  ins(%input : tensor<2x10xf32>)
  outs(%out_val, %out_idx : tensor<2xf32>, tensor<2xi32>) {
^bb0(%a: f32, %b: f32):
  %cmp = arith.cmpf olt, %a, %b : f32
  iree_linalg_ext.yield %cmp : i1
}

// Example: custom comparator (select value with larger absolute value)
iree_linalg_ext.arg_compare dimension(1)
  ins(%input : tensor<2x10xf32>)
  outs(%out_val, %out_idx : tensor<2xf32>, tensor<2xi32>) {
^bb0(%a: f32, %b: f32):
  %abs_a = math.absf %a : f32
  %abs_b = math.absf %b : f32
  %cmp = arith.cmpf ogt, %abs_a, %abs_b : f32
[...]

PR #23015, opened 2026-01-05
[LinalgExt] Generalize ArgCompareOp in GPUGeneralizeNamedOpsPass

bangtianliu:
This PR adds `ArgCompareOp `generalization to `GPUGeneralizeNamedOpsPass`, converting it to a `linalg.generic` operation by inlining the comparator region. This enables `ArgCompareOp `to flow through the VectorDistribute pipeline.

## Background 
ArgCompareOp performs a reduction over a specified dimension, returning both the selected value and its index. It uses a user-defined comparator region that can express:
- **argmax**: arith.cmpf ogt, %a, %b
- **argmin**: arith.cmpf olt, %a, %b
- **Custom logic**: Any boolean predicate comparing two values

## Approach
The generalization inlines the comparator region directly into a linalg.generic body, using arith.select for both value and index updates. This approach:
- Avoids special-casing different comparison predicates
- Preserves the original comparison semantics from the region
- Produces a uniform output regardless of the specific comparator used
- Handles optional index_base offset for split reduction scenarios
- Skips arith.index_cast when output index type is already index

ci-extra: test_torch

Issue: https://github.com/iree-org/iree/issues/23005


compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -1249,4 +1249,214 @@ FailureOr<SmallVector<Value>> ExpReductionOp::decomposeOperation(OpBuilder &b) {
...
+  int64_t rank = inputType.getRank();
+
+  Value identityVal = arith::getIdentityValue(rmwKind, elemType, b, loc);
+  Value zeroIdx =
+      arith::ConstantOp::create(b, loc, b.getIntegerAttr(idxElemType, 0));
+
+  Value filledVal =
+      linalg::FillOp::create(b, loc, identityVal, outVal).getResult(0);
+  Value filledIdx =
+      linalg::FillOp::create(b, loc, zeroIdx, outIdx).getResult(0);
```
IanWood1: Shouldn't we be using the pre-existing values from `iree_linalg_ext.arg_compare`'s outs operands?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -1249,4 +1249,214 @@ FailureOr<SmallVector<Value>> ExpReductionOp::decomposeOperation(OpBuilder &b) {
...
+
+        Value idx =
+            linalg::IndexOp::create(nestedBuilder, nestedLoc, reductionDim);
+        Value currentIdx = arith::IndexCastOp::create(nestedBuilder, nestedLoc,
+                                                      idxElemType, idx);
+
+        Value newVal = arith::getReductionOp(rmwKind, nestedBuilder, nestedLoc,
+                                             inputElem, accVal);
+
+        Value cmp;
```
IanWood1: Why can't we just inline `iree_linalg_ext.arg_compare`'s region here and give that value to an `arith.select` to choose either `%in` or `%out`. That way, this transformation doesn't need to worry about the contents of the region at all.


Using your test case "arg_compare_argmax_f32" as an example:

```
func.func @arg_compare_argmax_f32(%input: tensor<4x128xf32>,
                                   %out_val: tensor<4xf32>,
                                   %out_idx: tensor<4xi32>)
    -> (tensor<4xf32>, tensor<4xi32>) {
  %result:2 = iree_linalg_ext.arg_compare
      dimension(1)
      ins(%input : tensor<4x128xf32>)
      outs(%out_val, %out_idx : tensor<4xf32>, tensor<4xi32>) {
    ^bb0(%a: f32, %b: f32):
      %cmp = arith.cmpf ogt, %a, %b : f32
  } -> tensor<4xf32>, tensor<4xi32>
  return %result#0, %result#1 : tensor<4xf32>, tensor<4xi32>
}
```

This would produce the following
```
%cmp = arith.cmpf ogt, %in, %out : f32
%result = arith.select %cmp, %in, %out : f32
```

Instead of what is currently being generated:

```
%result = arith.maximumf %in, %out : f32
```



The `arith.cmpf` + `arith.select` -> `arith.maximumf` optimization seems like a transformation that should be owned by another pass. It looks like a straight forward pattern match, so I don't think we are losing information this way. But I could be missing something here, so let me know if there is a reason you choose not to do it this way.

IanWood1: https://mlir.llvm.org/docs/Dialects/ArithOps/#arithmaximumf-arithmaximumfop

>Returns the maximum of the two arguments, treating -0.0 as less than +0.0. If one of the arguments is NaN, then the result is also NaN.

Actually, `arith.cmpf` + `arith.select` isn't equal to `arith.maximumf` in the two cases mentioned above. Maybe this wouldn't be a legal canonicalization, then.

bangtianliu: Good question! The reason we generate arith.maximumf + arith.cmpf ogt instead of just inlining the region is that we want the decomposed linalg.generic to match the pattern expected by [isArgmaxOp](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp#L904)

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
bangtianliu: Yes, we should use values from outs operands.

The reason behind that was also to match the requirement from isArgMaxOp: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp#L931

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
Max191: +1 on just inlining the region of the arg_compare if possible. Is there a reason why the form that matches `isArgmaxOp` is preferred at this point?

My understanding was that the point of having the arg_compare op was so that we don't need to implement the complex matching logic that would be required. If we are decomposing the arg_compare operation, then we are intentionally giving up the matching convenience, so we shouldn't care about matching anymore at the point of decomposition. Also, the form that isArgmaxOp expects is already somewhat arbitrary, and if something changes in either that matcher or in this decomposition, then the connection here will be lost.

The reason I like inlining is that it saves a lot of complexity and boilerplate from the decomposition, since you'd be able to get rid of all the switch cases. Unless there is a good reason for keeping the form of the `isArgmaxOp` function, then I think we should go for inlining.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/decompose_aggregate_op.mlir
```diff
@@ -496,3 +496,191 @@ func.func @exp_reduction(
 // CHECK:                             arith.mulf
 // CHECK:                             arith.addf
 // CHECK:      return %[[M]], %[[SUM]], %[[PV]]
+
```
Max191: If you go with the inlining idea, then I think you can also significantly reduce the number of tests, since there will be fewer functionally different cases.

Max191 (commented):
My main feedback is that I think we should avoid the big switch cases if possible. Inlining the arg_compare region would be ideal if we can.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
kuhar: > https://mlir.llvm.org/docs/Dialects/ArithOps/#arithmaximumf-arithmaximumfop
> 
> > Returns the maximum of the two arguments, treating -0.0 as less than +0.0. If one of the arguments is NaN, then the result is also NaN.
> 
> Actually, `arith.cmpf` + `arith.select` isn't equal to `arith.maximumf` in the two cases mentioned above. Maybe this wouldn't be a legal canonicalization, then.

Nobody cares about positive/negative zero in ML and most implementations don't respect it anyway: https://discourse.llvm.org/t/rfc-a-consistent-set-of-semantics-for-the-floating-point-minimum-and-maximum-operations/89006

This minor difference is safe to ignore.

bangtianliu: Thanks for your comments @IanWood1 @Max191 @kuhar, it makes sense very much.

I’m not defending the current approach, this discussion made me realize there’s a NaN-semantics concern.
`arith.maximumf` is NaN-propagating by definition. A lowering via `arith.cmpf + arith.select` is not equivalent: with ordered predicates (e.g., ogt), NaNs make the compare false, so whether NaN propagates becomes operand-order dependent (and ±0.0 handling differs too).

So the question is: do we want to preserve NaN-propagation semantics, or can we assume NaNs don’t appear (typical ML assumption)? If we can ignore NaNs, inlining the region is simpler.

bangtianliu: > +1 on just inlining the region of the arg_compare if possible. Is there a reason why the form that matches `isArgmaxOp` is preferred at this point?

The original intention behind this design choice was to first plumb through the argmax/argmin path along the VectorDistribute pipeline, then extend it to general cases. The existing `isArgMaxOp `is primarily used for ukernel support, so it's fine to inline the region here and implement a new matcher for my own purposes later if needed.

kuhar (approved):
LGTM but PR description seems out of sync with the code after the most recent changes.

Also please wait for @Max191 and @IanWood1 before merging

MaheshRavishankar (commented):
Hmmm... I am not sure this qualifies as an "AggregatedOp" in the sense it was used so far. Things like Softmax etc, that actually decompose to multiple linalg.generic operations is what these have been used for. Do we need to add this interface for arg_compare operation. Maybe we can have a different interface that is just for conversion to linalg.generic before vectorization.

bangtianliu:
> Hmmm... I am not sure this qualifies as an "AggregatedOp" in the sense it was used so far. Things like Softmax etc, that actually decompose to multiple linalg.generic operations is what these have been used for. Do we need to add this interface for arg_compare operation. Maybe we can have a different interface that is just for conversion to linalg.generic before vectorization.

Yes, the current implementation only produces a single linalg.generic. After thinking more about this and when I am about to work on how to support ArgCompareOp from vectorization, I realized that we could skip decomposition entirely and add dedicated support for ArgCompareOp directly without converting to linalg.generic. We could use a helper function to recognize argmax patterns for a specialized path, then generalize to other comparators later. What do you think? cc @MaheshRavishankar @kuhar @Max191 @IanWood1 @Groverkss for suggestions (the context is about plumbing through ArgCompareOp along vector distribute pipeline, I want to start from argmax first).

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -1249,4 +1249,77 @@ FailureOr<SmallVector<Value>> ExpReductionOp::decomposeOperation(OpBuilder &b) {
...
+      ValueRange{outVal, outIdx}, indexingMaps, iteratorTypes,
+      [&](OpBuilder &nestedBuilder, Location nestedLoc, ValueRange args) {
+        Value inputElem = args[0];
+        Value accVal = args[1];
+        Value accIdx = args[2];
+
+        Value idx =
+            linalg::IndexOp::create(nestedBuilder, nestedLoc, reductionDim);
+        Value currentIdx = arith::IndexCastOp::create(nestedBuilder, nestedLoc,
+                                                      idxElemType, idx);
```
IanWood1: This will create an illegal index cast if the output type is already `index`.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/decompose_aggregate_op.mlir
```diff
@@ -496,3 +496,47 @@ func.func @exp_reduction(
...
+// CHECK-SAME:    iterator_types = ["parallel", "reduction"]
+// CHECK-SAME:    ins(%[[INPUT]] : tensor<4x128xf32>)
+// CHECK-SAME:    outs(%[[OUT_VAL]], %[[OUT_IDX]] : tensor<4xf32>, tensor<4xi32>)
+// CHECK:         linalg.index 1
+// CHECK:         arith.index_cast
+// CHECK:         arith.cmpf ogt
+// CHECK:         arith.select
+// CHECK:         arith.select
+// CHECK:         linalg.yield
+// CHECK:       return %[[GENERIC]]#0, %[[GENERIC]]#1
```
IanWood1: Could you add a test for an op with `index_base` (check that it either doesn't get decomposed or that it is decomposed correctly)?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
```diff
@@ -1249,4 +1249,77 @@ FailureOr<SmallVector<Value>> ExpReductionOp::decomposeOperation(OpBuilder &b) {
...
+          nestedBuilder.clone(op, regionMap);
+        }
+
+        auto yieldOp = cast<IREE::LinalgExt::YieldOp>(srcBlock.getTerminator());
+        Value cmp = regionMap.lookup(yieldOp.getOperand(0));
+
+        Value newVal = arith::SelectOp::create(nestedBuilder, nestedLoc, cmp,
+                                               inputElem, accVal);
+        Value newIdx = arith::SelectOp::create(nestedBuilder, nestedLoc, cmp,
+                                               currentIdx, accIdx);
```
IanWood1: This doesn't handle the `index_base` arg. It should be easy to just create an `arith::AddIOP` to support it.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/decompose_aggregate_op.mlir
IanWood1: Also, maybe a test for https://github.com/iree-org/iree/pull/23015#discussion_r2665733876 should be added too.

Max191:
> Hmmm... I am not sure this qualifies as an "AggregatedOp" in the sense it was used so far

Good point. This is more similar to generalizing a named linalg op.

> I realized that we could skip decomposition entirely and add dedicated support for ArgCompareOp directly without converting to linalg.generic

Vectorizing the arg_compare op could prove difficult, since it has a region. It is quite similar to a linalg generic op, so we should try to make use of the linalg.generic vectorizer. I don't have context about how this interface was intended to fit into the lowering path for arg_compare, but I had figured the purpose was to make use of the linalg.generic vectorizer (and/or other infra around linalg ops).

> We could use a helper function to recognize argmax patterns for a specialized path, then generalize to other comparators later

Is there something special we want to do for argmax vs other types of comparators? I figured argmax would just be handled the same as any other arg_compare.

bangtianliu:
> Good point. This is more similar to generalizing a named linalg op.

Yeah,  then can other option is converting this PR to generalizing a named linalg op. Then most of code can reused. 

> 
> Vectorizing the arg_compare op could prove difficult, since it has a region. It is quite similar to a linalg generic op, so we should try to make use of the linalg.generic vectorizer. I don't have context about how this interface was intended to fit into the lowering path for arg_compare, but I had figured the purpose was to make use of the linalg.generic vectorizer (and/or other infra around linalg ops).
> 
> 
> Is there something special we want to do for argmax vs other types of comparators? I figured argmax would just be handled the same as any other arg_compare.

Regarding argmax vs other comparators: argmax/argmin are the most common use cases in practice, so I want to support them first and then extend to general cases.

Using the linalg.generic vectorizer is one approach. However, if we know it's an arg-reduction op, there may be more efficient lowering strategies (e.g., leveraging dpp/ballot/readlane intrinsics). This needs further investigation.

Also do you think we should define a `vector_ext.arg_compare` to do vectorization for this op since I noticed that we have support for gather op like that.

IanWood1 (approved):
LGTM in terms of correctness, but I'd wait until the "AggregatedOp" concerns raised by @MaheshRavishankar and @Max191 are resolved. Changing this to "generalizing a named linalg op" makes sense to me. I can review again if that path is chosen.

bangtianliu:
Per discussion with @Max191, this PR will be converted to generalize linalg.generic ops. I'll then evaluate VectorDistribute support and determine next steps based on the findings.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp
krzysz00: Re maximumf, I'd suggest taking a peek at what various backends of interest implement and wondering if we want `maxnumf` (or maybe it's spelled `maximumnumf`, I forget) instead

tests/external/iree-test-suites/torch_models/llama_8b_fp16/modules/scheduler_gfx942.json
```diff
@@ -1,11 +1,11 @@
...
   "compiler_flags": [
-      "--iree-hal-target-device=hip",
-      "--iree-hip-target=gfx942",
-      "--iree-opt-level=O1",
-      "--iree-stream-resource-memory-model=discrete",
-      "--iree-hal-indirect-command-buffers=true",
-      "--iree-hal-memoization=true"
+    "--iree-hal-target-device=hip",
+    "--iree-hip-target=gfx942",
+    "--iree-opt-level=O3",
```
Max191: Why is this changing to O3?

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/GeneralizeArgCompare.cpp
```diff
@@ -0,0 +1,90 @@
+// Copyright 2025 The IREE Authors
```
Max191: nit: Maybe change the filename to GeneralizeLinalgExtOps.cpp in case we have more ops that can be generalized like this in the future?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUGeneralizeNamedOps.cpp
```diff
@@ -13,6 +13,8 @@
```
Max191, line 13: Can you also update the description in the pass docs to indicate that this generalizes LinalgExt ops too?

https://github.com/iree-org/iree/blob/b71b345821ff8e50d4122f94545d5aa68df5027e/compiler/src/iree/compiler/Codegen/Common/GPU/Passes.td#L156-L162

Max191 (approved):
LGTM, just a couple of nits + one question

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_generalize_named_ops.mlir
```diff
@@ -168,3 +168,112 @@ func.func @reduce_op(%arg0: tensor<32x128xf32>, %arg1: tensor<32xf32>) -> tensor
...
+// CHECK-SAME:    %[[INPUT:[a-zA-Z0-9_]+]]: tensor<4x128xf32>
+// CHECK-SAME:    %[[OUT_VAL:[a-zA-Z0-9_]+]]: tensor<4xf32>
+// CHECK-SAME:    %[[OUT_IDX:[a-zA-Z0-9_]+]]: tensor<4xindex>
+// CHECK:        linalg.generic
+// CHECK-SAME:   indexing_maps = [#[[$MAP]], #[[$MAP1]], #[[$MAP1]]]
+// CHECK-SAME:   iterator_types = ["parallel", "reduction"]
+// CHECK-SAME:   outs(%[[OUT_VAL]], %[[OUT_IDX]] : tensor<4xf32>, tensor<4xindex>)
+// CHECK:        ^bb0(%[[IN:.+]]: f32, %[[ACC_VAL:.+]]: f32, %[[ACC_IDX:.+]]: index):
+// CHECK:          %[[IDX:.+]] = linalg.index 1 : index
+// CHECK-NOT:      arith.index_cast
```
kuhar: I don't think we need this: index to index cast would fail to verify anyway

compiler/src/iree/compiler/Codegen/Common/GPU/GPUGeneralizeNamedOps.cpp
```diff
@@ -68,6 +70,19 @@ struct GPUGeneralizeNamedOpsPass final
     if (failed(generalizeCandidates(&getContext(), namedOpCandidates))) {
       return signalPassFailure();
     }
+
+    SmallVector<IREE::LinalgExt::ArgCompareOp> argCompareOps;
+    funcOp.walk(
+        [&](IREE::LinalgExt::ArgCompareOp op) { argCompareOps.push_back(op); });
+
+    IRRewriter rewriter(&getContext());
+    for (auto argCompareOp : argCompareOps) {
```
kuhar: nit: spell out this type, per https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/GeneralizeArgCompare.cpp
kuhar: this should be 2026 now

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h
```diff
@@ -188,4 +188,39 @@ FailureOr<linalg::SplitReductionResult>
 splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
                      linalg::ControlSplitReductionFn controlSplitReductionFn);
 
+/// Generalize an ArgCompareOp to a linalg.generic by inlining its comparator
+/// region. This transforms the high-level arg_compare operation into a
+/// linalg.generic with explicit index tracking and selection logic.
+///
+/// Example:
+/// ```
```
kuhar:
```suggestion
/// ```mlir
```

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h
```diff
@@ -188,4 +188,39 @@ FailureOr<linalg::SplitReductionResult>
...
+/// %result:2 = iree_linalg_ext.arg_compare dimension(1)
+///     ins(%input : tensor<4x128xf32>)
+///     outs(%out_val, %out_idx : tensor<4xf32>, tensor<4xi32>) {
+///   ^bb0(%a: f32, %b: f32):
+///     %cmp = arith.cmpf ogt, %a, %b : f32
+///     iree_linalg_ext.yield %cmp : i1
+/// } -> tensor<4xf32>, tensor<4xi32>
+/// ```
+/// Becomes:
+/// ```
```
kuhar:
```suggestion
/// ```mlir
```

tests/external/iree-test-suites/torch_models/llama_8b_fp16/modules/scheduler_gfx942.json
kuhar: +1, this should be kept out

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_generalize_named_ops.mlir
```diff
@@ -168,3 +168,112 @@ func.func @reduce_op(%arg0: tensor<32x128xf32>, %arg1: tensor<32xf32>) -> tensor
...
+func.func @arg_compare_index_output(%input: tensor<4x128xf32>,
+                                    %out_val: tensor<4xf32>,
+                                    %out_idx: tensor<4xindex>)
+    -> (tensor<4xf32>, tensor<4xindex>) {
+  %result:2 = iree_linalg_ext.arg_compare
+      dimension(1)
+      ins(%input : tensor<4x128xf32>)
+      outs(%out_val, %out_idx : tensor<4xf32>, tensor<4xindex>) {
+    ^bb0(%a: f32, %b: f32):
+      %cmp = arith.cmpf ogt, %a, %b : f32
```
kuhar, line 257: Can we also have a testcase for at least one integer case?

tests/external/iree-test-suites/torch_models/llama_8b_fp16/modules/scheduler_gfx942.json
bangtianliu: For testing purpose (not belong to this PR), will undo it before landing. 

See [[CI][Torch]](https://github.com/iree-org/iree/pull/22379) for context, and discussion here: https://github.com/iree-org/iree/pull/22379#issuecomment-3437815189. Previously, we saw argmax compilation issues when using -O3.

I temporarily enabled -O3 to check whether the changes in the current PR introduce any compilation or numerical correctness regressions under that optimization level.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/GeneralizeArgCompare.cpp
bangtianliu: yeah, now it is 2026, sign

kuhar (approved):

MaheshRavishankar:
Ok, I dont have a strong enough opinion to hold this/revert this, but I think it is better to not use AggregateOpInterface for this. So any follow up to drop it from there would be appreciated.

bangtianliu:
> Ok, I dont have a strong enough opinion to hold this/revert this, but I think it is better to not use AggregateOpInterface for this. So any follow up to drop it from there would be appreciated.

The existing PR doesn't use AggregateOpInterface, but during my process of plumbing arg_compare op along the VectorDistribute pipeline, I found that working directly with ArgCompareOp (without converting it to linalg.generic) will be a better solution.

The reason is that VectorDistribute performs two-level tiling:
- Outer level: Partial reduction (PartialReductionOuterReduction)
- Inner level: Inter-thread reduction (subgroup-wise)

So a clear TODO for me is to add PartialReductionOuterReduction support for ArgCompareOp directly, in addition to the existing PartialReductionOuterParallel (split-reduction) support. Extending the existing linalg.generic reduction support to handle arg_reduce semantics would require drastic changes to the infrastructure (upstream tilinginterface).

MaheshRavishankar:
> > Ok, I dont have a strong enough opinion to hold this/revert this, but I think it is better to not use AggregateOpInterface for this. So any follow up to drop it from there would be appreciated.
> 
> The existing PR doesn't use AggregateOpInterface, but during my process of plumbing arg_compare op along the VectorDistribute pipeline, I found that working directly with ArgCompareOp (without converting it to linalg.generic) will be a better solution.
> 
> The reason is that VectorDistribute performs two-level tiling:
> 
> * Outer level: Partial reduction (PartialReductionOuterReduction)
> * Inner level: Inter-thread reduction (subgroup-wise)
> 
> So a clear TODO for me is to add PartialReductionOuterReduction support for ArgCompareOp directly, in addition to the existing PartialReductionOuterParallel (split-reduction) support. Extending the existing linalg.generic reduction support to handle arg_reduce semantics would require drastic changes to the infrastructure (upstream tilinginterface).

This totally makes sense!

Issue #21590, opened 2025-08-05
[CPU] SVE: workgroup count calculation injects vscale to initializers


egebeysel:
With SVE enabled, we get distribution loops of the form:

```mlir
  %1 = affine.apply affine_map<()[s0] -> (4096 ceildiv s0)>()[%c8_vscale]
  ...
  %4 = tensor.empty(%1, %c8_vscale) : tensor<?x4096x?x1xf32>
  %5 = scf.forall (%arg0, %arg1) = (0, 0) to (%1, 4096) step (64, 64) shared_outs(%arg2 = %4) -> (tensor<?x4096x?x1xf32>) {
   ...
   linalg.pack ... inner_tiles=[%c8_vscale, 1]
```
where the UB and therefore the workgroup count depends on the value of vscale, which is a runtime variable.

Therefore, in StreamToHALConversion while calculating the workgroup counts, we end up with `vector.vscale` ops in the initializers (and also in the memoize funcs) and this is marked as illegal and fails on the subsequent VM conversion pass.

I also have a rather long reproducer that I tried to strip down to the best of my abilities [here](https://gist.github.com/egebeysel/004f842901b400d63073aace0d274288).

I'd really like to hear your opinion on this one @benvanik.
Also cc @hanhanW @banach-space. Thanks in advance! Let me know if anything is unclear please :)

Side note: I believe this issue also connects to the storage size calculation issue: #21317.

Issue #21317, opened 2025-07-09
[CPU][DT] SVE: Storage Size Calculation


egebeysel:
As discussed briefly in https://github.com/iree-org/iree/pull/21304, storage size calculation in the presence of scalable tiles requires knowledge about the exact vector register length, which in the case of scalable vectors/matrices is a fixed _runtime_ value that can be queried using the `vector.vscale` op.

Although storage size calculation resides on the host-side, and we need a mechanism to query this from the target device. First ideas on how we could do this (from @banach-space and @hanhanW) include:

1. Conservatively using the maximum value of `vscale=16` for storage size calculation and allocating a big enough buffer. As stated by Andrzej in the issue above current devices usually have `vscale=1,2` whereas the maximum known implementation has `vscale=4`. Which means at best, we would be allocating x4 the buffer we actually need. But it is a step forward to unblock the issue.
2. For the cases where the host and device are actually identical, adding a mechanism for the host to actually query "itself" for the value of vscale. I believe this one is the best way to _unblock_ this for now - especially since the current devices tend to have the property of host and device being the same.
3. Probably the most involved one, but also the most general approach would be to have a mechanism for the host to actually query the device and store the value of vscale to use. This one probably would be the target for long-term.
4. Provide a compile-time flag/option/mechanism to specify `vscale`: essentially, scalable vectors and matrices aim to allow vector-length agnostic (VLA) programming and fully specifying the target vscale - which is fixed per hardware implementation AFAIK - on compile-time is sort of counter intuitive, although this would also unblock us for the time being and allow us to experiment.

Both 2 & 3 would probably require a wrapper around the vscale op on the vm side and changes in runtime. Also, I guess 2 can be used as a special case of 3 when host=device.

Here's a quick snippet to play around with this:

```mlir
util.initializer {
  %cst = stream.tensor.constant on(#hal.device.affinity<@__device_0>) : tensor<4096x4096xf32> in !stream.resource<constant> = dense_resource<__auto.constant_4096_4096_torch.float32> : tensor<4096x4096xf32>
  %0 = stream.resource.size %cst : !stream.resource<constant>
  util.global.store %cst, @__constant_tensor_4096x4096xf32 : !stream.resource<constant>
  util.global.store %0, @__constant_tensor_4096x4096xf32__size : index
  %__constant_tensor_4096x4096xf32__size = util.global.load @__constant_tensor_4096x4096xf32__size : index
[...]

PR #23023, opened 2026-01-05
[DT] [SVE] specify vscale for wg count and storage size calculations

egebeysel:
Add the CLI option to specify the value of vscale for _host-side_ code, e.g. stroge size calculation (#21317) or wg-count calculation (#21590). The flag is supposed to be a workaround and it and the corresponding tests can be removed once the issues are resolved.


compiler/src/iree/compiler/Utils/EquivalenceUtils.cpp
```diff
@@ -15,6 +18,23 @@
 
 namespace mlir::iree_compiler {
 
+static llvm::cl::opt<int> clVscaleValue(
+    "iree-util-vscale-value",
+    llvm::cl::desc("The value of vscale, this is a temporary flag #21317"),
+    llvm::cl::init(1));
```
banach-space: [nit] I'd initialise as e.g. `-1` and then perhaps add asserts to make sure that whenever this is used, it's been correctly set by a user.

compiler/src/iree/compiler/Utils/EquivalenceUtils.cpp
```diff
@@ -15,6 +18,23 @@
...
+
+unsigned getVscaleValue() { return clVscaleValue; }
+
+void mapVscaleOpToConstant(Operation &op, OpBuilder &builder, IRMapping &bvm,
```
banach-space:
```suggestion
void getMinVscaleAsCst(Operation &op, OpBuilder &builder, IRMapping &bvm,
```

I would assert here that `vscale >= 1`. Also, I would document that this will simply replace `vector.vscale` with the known min value for `vscale`.

compiler/src/iree/compiler/Utils/EquivalenceUtils.cpp
```diff
@@ -15,6 +18,23 @@
 
 namespace mlir::iree_compiler {
 
+static llvm::cl::opt<int> clVscaleValue(
```
banach-space:
```suggestion
static llvm::cl::opt<int> clVscaleValue(I'd rename to make it a bit more self-documenting.
```suggestion
static llvm::cl::opt<int> clVscaleMinVal(
```
Also, add a comment explaining that this is meant to help for the runtime/host-side code that currently cannot evaluate `vector.vscale` when querying sizes. 

It's also worth adding a note that this value will not be used during code-gen and that it should be at last as large as the actual runtime value.
```

File: compiler/src/iree/compiler/Utils/EquivalenceUtils.cpp [hunk 4]
```diff
@@ -15,6 +18,23 @@
...
+    "iree-util-vscale-value",
```
banach-space:
```suggestion
    "iree-vscale-min-val",
```

banach-space (commented):
Hey Ege, thanks for implementing this!

The overall logic makes sense to me. I’ve left a few minor inline suggestions, mostly around naming and documentation.

One higher-level comment: I think the change is framed a bit too much as _a temporary workaround_. To me, having a flag like this makes sense even long-term. We have a very similar concept in Clang/LLVM (see:
https://github.com/llvm/llvm-project/blob/188507e54218c91832714fc50b24bd2fc2fd9383/clang/include/clang/Options/Options.td?plain=1#L5480-L5483), and my naming suggestions are primarily about aligning with that precedent.

I’d also explicitly document that this flag:
* Is intended to help with host-side code generation
* Is not used for device-side code generation

The related issues (https://github.com/iree-org/iree/issues/21317 and https://github.com/iree-org/iree/issues/21590) provide very useful background and ideas for more general approaches, and are definitely worth keeping referenced. That said, I wouldn’t treat the newly added flag as something inherently temporary while those more generic solutions are explored.

Thanks again, +1 from me!

compiler/src/iree/compiler/Utils/EquivalenceUtils.cpp
hanhanW: We should add `experimental` to the flag, so the flag can is expected to be used by experts only.

So maybe `iree-experimental-vscale-min-val"`

compiler/src/iree/compiler/Utils/EquivalenceUtils.h
```diff
@@ -8,15 +8,28 @@
 #define IREE_COMPILER_UTILS_EQUIVALENCEUTILS_H_
 
 #include "llvm/ADT/SetVector.h"
+#include "mlir/IR/Builders.h"
 #include "mlir/IR/Operation.h"
 
 namespace mlir {
 class Block;
 class IRMapping;
+class OpBuilder;
```
hanhanW: I think you don't need this because you already include `Builders.h`.

compiler/src/iree/compiler/Utils/EquivalenceUtils.h
hanhanW, line 1: I haven't figured out where should we put these methods yet. But I'm curious that why do you put them here?

compiler/src/iree/compiler/Dialect/HAL/IR/HALOps.cpp
hanhanW, line 1: Instead of spreading workaround to other dialect and component. Have you considered creating constants in resolver implementation when the flag is on?

hanhanW (changes requested):

compiler/src/iree/compiler/Dialect/HAL/IR/HALOps.cpp
egebeysel: IIUC, you mean creating the constants directly during encoding resolvers and not materializing vscale ops at all, which would also indicate that I will be emitting non vector length agnostic code for device-side code as well, since resolver code is shared between and host and device.

These values occur _at the boundaries_ of host and device-side code, i.e. _spill_ from device-side codegen to host-side code, and the main purpose I have here is generating device-side code that is fully agnostic to the vector length - and eventually also host-side code, thought that seems to be more involved, #21317.

Please tell me if I'm misunderstanding something here and if I can just create the constants in the resolver implementation and _just_ utilize those for the host-side code.

compiler/src/iree/compiler/Utils/EquivalenceUtils.h
egebeysel: I also don't have a very clear picture of why exactly this file, but the main purpose of selecting `compiler/Utils` is that I didn't want to introduce any other dependency to HAL, i.e. `Codegen/Utils` or something like that. Might have also been its own file I guess, but I just thought having them here, as this also somewhat resonates with equivalence/equality made semi-sense. Happy to move it elsewhere if you have a better suggestion though!

compiler/src/iree/compiler/Utils/EquivalenceUtils.cpp
egebeysel: done

compiler/src/iree/compiler/Utils/EquivalenceUtils.cpp
egebeysel: done - though as discussed offline, not as the min value but rather the exact one

compiler/src/iree/compiler/Utils/EquivalenceUtils.h
egebeysel: yep, missed that one thanks

egebeysel:
Thanks for the reviews! @hanhanW @banach-space 

As per our offline discussion with @banach-space:

Currently, we cannot frame this as the _min_ known value for vscale, since we have use cases where this would lead to underestimation as well, e.g. for the underestimation of the workgroup count.

Therefore, I kept this value as the experimental, user-specified runtime value of `vscale` instead of a min/max known value. I added explicit comments stating that this flag is not to be used for device-side code and that device-side code is meant to be vector length agnostic. Also, added an assert to make sure that the user indeed explicitly specifies this value if it has to be used.

compiler/src/iree/compiler/Dialect/HAL/IR/HALOps.cpp
banach-space: >  the main purpose I have here is generating device-side code that is fully agnostic to the vector length 

This is key - this patch is primarily to help us with the **host** side of things. While we could use it for device-side (to implement Vector Length Specific/VLS vectorization for SVE), that's not really on our radar, is it? So, the idea is to:
* specialise on the host side,
* keep things vector-length-agnostics on the device side.

hanhanW: Yes, what @banach-space said is what I'm proposing. I'm suggesting moving the flag to `Codegen/ExternalInterfaces/Utils.cpp`, because `calculatePackedStorageSizeInBytesImpl` is the key method on the host side. Then we don't need to expose "workarounds/experimental codes" to other components like `HAL dialect` and `compiler/Utils`.

egebeysel: @hanhanW I'm sorry I might be missing something here, but we have two cases where the `vscale` operation gets injected into host-side code.
- Through `calculatePackedStorageSizeInBytesImpl`, which as you've mentioned lives in `Codegen/ExternalInterfaces` itself.
- Through the upper bounds of the distribution level loops to workgroup count calculation, which if I'm not missing something happens [here](https://github.com/iree-org/iree/blob/818f45f681923142b0dfd1a4b02deffdebfda707/compiler/src/iree/compiler/Dialect/HAL/IR/HALOps.cpp#L1893). The upper bound of the distribution loops is for example the outer N dimension of the mmt4d result, i.e. for the result shape `m0 x n0 x m1 x n1` where `n0 = ceildiv(n, n1)` and `n1` scalable, we get the loop upper bound being bound to the scalable `n1`. That makes `vscale` a part of the workgroup count calculation as well. If I'm not missing something, this mechanism is not connected with `calculatePackedStorageSizeInBytesImpl`.

I don't have anything against moving the logic to `Codegen/ExternalInterfaces/Utils.cpp`, though that means HAL would have a dependency on Codegen.

Let me know what you think and if I'm missing something!

egebeysel: Also, I'll fix the CI soon when I get a chance - sorry for the delay.

hanhanW: I may be missing something as well. 😅

I honestly am not aware of the second point. IIUC, it happens either in [lowerDispatchWorkgroupCountForDagRootOp](https://github.com/iree-org/iree/blob/56acf7ed012bd0432802891cbc8e3f30ae5deb4e/compiler/src/iree/compiler/Codegen/Common/TileAndDistributeToWorkgroupsPass.cpp#L128-L208) or [lowerWorkgroupCount](https://github.com/iree-org/iree/blob/56acf7ed012bd0432802891cbc8e3f30ae5deb4e/compiler/src/iree/compiler/Codegen/Common/TileAndDistributeToWorkgroupsPass.cpp#L210-L262). In either case, I think we can guard it with the same flag?

> I don't have anything against moving the logic to Codegen/ExternalInterfaces/Utils.cpp, though that means HAL would have a dependency on Codegen.

I don't follow this. Why does HAL start depending on Codegen (i.e., adding codegen dep to HAL/BUILD.bazel) if we go with this direction?

egebeysel: I thought the calculation did happen during the HAL conversion at [calculateWorkgroupCountFromRegion](https://github.com/iree-org/iree/blob/a41330552117c7c5f11db1b0e44dc25d0133bf4d/compiler/src/iree/compiler/Dialect/HAL/IR/HALOps.cpp#L1893), but now that I think about it, I probably have to check where that _region_ is actually created from the loop vars - makes more sense that it would happen somewhere at codegen and we can just tackle it there early on - thanks for the pointers and the review! :)

egebeysel:
@hanhanW Thanks for the review and the pointers - I've removed the code that spilled to HAL to handle it early on during workgroup count hint materialization  and put the flag right next to it's sibling :) Hope this looks better!

compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp
```diff
@@ -26,15 +26,28 @@ static llvm::cl::opt<bool> clEnableScalableVectorization(
...
+        "The runtime value of vscale. This will _only_ be used for host-side "
+        "code, e.g. to calculate storage sizes and workgroup counts. This is "
+        "due to a current limitation of the host-side code not being able to "
+        "properly query this value at runtime, see #21317 and #21590. Codegen "
+        "will be vector-length agnostic and will be querying the value of "
+        "vscale at runtime, as intended. For scalable vector code that "
+        "propagates vscale ops into the host-side code, this value has to be "
+        "explicitly set by the user, e.g. for SVE data-tiling on the AArch64 "
+        "backend."),
+    llvm::cl::init(-1));
```
hanhanW: Let's add `, llvm::cl::Hidden` for the experimental flag. I'm currently working on CPU options, and I think most of developer flags should be hidden. Viable options will be exposed via CPUCodegenOptions. ([WIP](https://github.com/iree-org/iree/pull/23259)).

It is okay if you don't add it atm. It is mostly an FYI that I'll eventually hide the option from CLI tools.

compiler/src/iree/compiler/Codegen/Transforms/Transforms.cpp
```diff
@@ -389,6 +382,16 @@ LogicalResult materializeSliceFromOrdinals(
...
+
+    // TODO(#21317, #21590): Currently, we cannot evaluate `vector.vscale` ops
+    // at host-side. Therefore, we map these values to the user-specified
+    // constants here at compile-time.
+    if (isa<vector::VectorScaleOp>(op)) {
+      Value constVal = arith::ConstantIndexOp::create(rewriter, op->getLoc(),
+                                                      getUserVscaleValue());
+      map.map(op->getResult(0), constVal);
+      continue;
+    }
```
hanhanW, line 394: IIUC, this is required because the HAL will clone the ops when `hal.executable.export`'s count region is materialized. There is no VectorToVM conversion, so the codegen has to kick in the workaround. In this context, LGTM. It's good to see that the workaround is scoped within `Codegen/`.

https://github.com/iree-org/iree/blob/a08d24b6b0443a3e0154564f30b5bf49478e7b9b/compiler/src/iree/compiler/Dialect/HAL/IR/HALOps.cpp#L1892-L1917

hanhanW (approved):
LGTM. @MaheshRavishankar it touches `materializeSliceFromOrdinals`, so maybe you want to take a look. I put my understanding in inlined comments, please take a look if you need more context.

Please wait a day to see if @MaheshRavishankar wants to review or not. The current revision looks okay to me. Thanks!

compiler/src/iree/compiler/Codegen/Transforms/Transforms.cpp
egebeysel: > IIUC, this is required because the HAL will clone the ops when `hal.executable.export`'s count region is materialized.

That's correct :)

> There is no VectorToVM conversion, so the codegen has to kick in the workaround.

Also correct, but for this specific op it's a bit complicated since the op is implicitly bound to the device that executes the original dispatch - which theoretically can be different from the host device (and we can have cases where host cannot even legalize this op, i.e. x86 host and aarch64 device with SVE).

> It's good to see that the workaround is scoped within Codegen/.

Thanks for pushing me to the right direction :)

hanhanW: This part is going to be tricky. I've been thinking it but I did not get a solution/workaround yet. FYI.

compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp
egebeysel: Makes sense, hid the flag :)

PR #23029, opened 2026-01-06
[Codegen] Resolve memref.dim ops inside narrow type emulation pass

jtuyls:
This fixes a compilation failure I encountered when compiling Llama 70B FP4 with data tiling enabled for gfx950. The issue was that memref.dim ops on narrow-type memrefs (like f4E2M1FN) through fat_raw_buffer_cast were marked illegal by the emulation pass but couldn't be resolved.

The fix adds a preprocessing step inside the narrow type emulation pass that runs `populateResolveRankedShapedTypeResultDimsPatterns` before the conversion.


hanhanW (approved):
LGTM, please wait for a day just in case if others have comments.

krzysz00 (approved):
No objection

PR #23030, opened 2026-01-06
[Util] Generalize FoldDimOp to trace through shape-preserving ViewLikeOpInterface

Yu-Zhewen:
This fixes data-tiling compilation of llama 70b fp4.


Yu-Zhewen:
Ah, just notice this is a duplicate with #23029.

PR #23031, opened 2026-01-06
[VM] Preserve mutability for globals and rely on fixup to update it.

### LINKED ISSUE #22880: onnx.ReduceMean operator is causing failure due to immutable operation on mutable object
### What happened?

We are seeing the following compiler error while compiling tf_efficientnet_b0_ns_Opset16 onnx model. Error is as follows : 
```
test_pass.mlir:252:12: error: 'vm.global.load.i64' op is marked as immutable but the global is mutable
    %248 = torch.operator "onnx.ReduceMean"(%246, %247) {torch.onnx.keepdims = 1 : si64} : (!torch.vtensor<[?,32,?,?],f32>, !torch.vtensor<[2],si64>) -> !torch.vtensor<[?,32,1,1],f32> 
           ^
test_pass.mlir:252:12: note: see current operation: %27 = "vm.global.load.i64"() <{global = @__hoisted_i64_21, is_immutable}> : () -> i64
```

This issue is introduced in the following commit : https://github.com/iree-org/iree/commit/3ac574a89d7a4e2ea1fe882927fc65934a7c3621 

1. Please download [test_medium.zip](https://github.com/user-attachments/files/24079144/test_medium.zip) file 
2. Unzip it to get test_medium.mlir
3. Run the following : 
```
iree-compile test_medium.mlir -o test.vmfb --iree-hal-target-backends=llvm-cpu --iree-llvmcpu-target-cpu=host
```

_No response_

### Version information

This error is first observed in : 
```
IREE (https://iree.dev):
  IREE compiler version 3.8.0rc20250916 @ e8050c352d82dfbd47076a750adce5408bebea4c
  LLVM version 22.0.0git
  Optimized build
```

Issue is also present in the latest version : 
```
IREE (https://iree.dev):
  IREE compiler version 3.10.0rc20251210 @ 0f64f41a0ae3c250409d8fe7068414d9666eca80
  LLVM version 22.0.0git
  Optimized build
```

### Additional context

Issue occured due to https://github.com/iree-org/iree/commit/3ac574a89d7a4e2ea1fe882927fc65934a7c3621, Local revarsal wokred on the main testcase and the medium testcase. 
Please find the main testcase here : [model.torch_onnx.zip](https://github.com/user-attachments/files/24079251/model.torch_onnx.zip)

The medium testcase passes till 3.8.0rc20250915, and started failing from 3.8.0rc20250916.
We have a smaller mlir to reproduce the original error, but this small testcase gives a linalg error in the passing version : [test_small.zip](https://github.com/user-attachments/files/24079330/test_small.zip)

Please reproduce and fix with medium testcase (attached in steps to reproduce) and please check the fix on the main testcase (attached here).

hanhanW:
The pass already runs fixupGlobalMutability method after initializing all globals with initial values. If the mutability is not preserved, it triggers a bug in [the fixup method](https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Dialect/VM/Transforms/GlobalInitialization.cpp#L98-L114).

If we always set mutable for the new globals, didChange is always true. Because it at least has one store.

```cpp
  explorer.forEachGlobal([&](const Explorer::GlobalInfo *globalInfo) {
    if (globalInfo->uses.empty())
      return;
    // TODO(benvanik): verify we want this behavior - we likely want to change
    // this to be mutable only if stores exist outside of initializers.
    //
    // If there are stores mark the global as mutable. We need to update all
    // of the loads if this changes anything.
    bool hasStores = !globalInfo->getStores().empty();
    bool didChange = globalInfo->op.isGlobalMutable() != hasStores;
    if (didChange) {
      globalInfo->op.setGlobalMutable(hasStores);
      for (auto loadOp : globalInfo->getLoads()) {
        loadOp.setGlobalImmutable(!hasStores);
      }
    }
  });
```

Fixes https://github.com/iree-org/iree/issues/22880

### DIFF
--- a/compiler/src/iree/compiler/Dialect/VM/Transforms/GlobalInitialization.cpp
+++ b/compiler/src/iree/compiler/Dialect/VM/Transforms/GlobalInitialization.cpp
@@ -224,7 +224,6 @@ class GlobalInitializationPass
       // initial value/initializer and avoid the work entirely.
       return success();
     }
-    globalOp.setGlobalMutable(true);
     return storePrimitiveGlobal(globalOp.getLoc(), globalOp.getGlobalName(),
                                 value, builder);
   }
--- a/compiler/src/iree/compiler/Dialect/VM/Transforms/test/global_initialization.mlir
+++ b/compiler/src/iree/compiler/Dialect/VM/Transforms/test/global_initialization.mlir
@@ -43,6 +43,8 @@ vm.module @mutability_change {
   vm.global.i32 private mutable @g1 = 123 : i32
   // CHECK: vm.global.i32 private mutable @g2 : i32
   vm.global.i32 private @g2 : i32
+  // CHECK: vm.global.i64 private mutable @g3
+  vm.global.i64 private @g3 = 12345 : i64
 
   vm.initializer {
     %c456 = vm.const.i32 456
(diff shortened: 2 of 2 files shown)


MaheshRavishankar:
I dont have context here. @benvanik  I think this needs your review.

benvanik (approved):

PR #23033, opened 2026-01-06
[CI] Add ci-extra flags for scheduled builds.

amd-eochoalo:
Allows the possibility of the following workflows to be triggered by ci-extra:

* linux_arm64_clang
* linux_x64_clang_debug
* linux_x64_clang_tsan
* linux_x64_clang_byollvm
* linux_x64_gcc

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,6 +235,24 @@ jobs:
     uses: ./.github/workflows/ci_linux_x64_clang.yml
     secrets: inherit
 
+  linux_x64_gcc:
+    needs: setup
+    if: contains(fromJson(needs.setup.outputs.enabled-jobs), 'linux_x64_gcc')
+    uses: ./.github/workflows/ci_linux_x64_gcc.yml
+    secrets: inherit
+
+  linux_arm64_clang:
+    needs: setup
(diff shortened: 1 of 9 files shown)


amd-eochoalo:
Some workflows are being built twice in this PR because of the rule:

```yaml
on:
  pull_request:
    paths:
      - ".github/workflows/ci_linux_arm64_clang.yml"
```

This means that they will be built when the file is modified. As we modified the file and added the ci-extra, then those workflows are being built twice.

kuhar (approved):

ScottTodd:
> Some workflows are being built twice in this PR because of the rule:
> 
> ```yaml
> on:
>   pull_request:
>     paths:
>       - ".github/workflows/ci_linux_arm64_clang.yml"
> ```
> 
> This means that they will be built when the file is modified. As we modified the file and added the ci-extra, then those workflows are being built twice.

That's worth fixing. You could remove the `pull_request` triggers and instead have `configure_ci.py` enable those jobs if those paths are modified.

ScottTodd (approved):
LGTM but would be nice to remove the `pull_request` triggers causing duplicate runs.

amd-eochoalo:
@ScottTodd ! Thanks! I removed the duplication.

PR #19244, opened 2024-11-21
Ignore multi-line git trailers in configure_ci.py.

ScottTodd:
Fixes https://github.com/iree-org/iree/issues/19240 (ugh string parsing).

PR #23034, opened 2026-01-06
[CI] Allow ci-extra to be set on integrates.

amd-eochoalo:
Allows ci-extra to be set on integrates.

### DIFF
--- a/build_tools/github_actions/configure_ci.py
+++ b/build_tools/github_actions/configure_ci.py
@@ -129,6 +129,11 @@ DEFAULT_SCHEDULE_ONLY_JOBS = frozenset(
     [
         "macos_arm64_clang",
         "macos_x64_clang",
+        "linux_arm64_clang",
+        "linux_x64_clang_byollvm",
+        "linux_x64_clang_debug",
+        "linux_x64_clang_tsan",
+        "linux_x64_gcc",
     ]
 )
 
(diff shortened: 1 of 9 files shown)


kuhar (changes requested):
I'm not sure this implementation works as intended. For example, integrate PR used to trigger the windows job that is off for regular PRs. I'd expect we need to add a union of extra jobs and (all_jobs - shedule_only) instead.

build_tools/github_actions/configure_ci.py
ScottTodd, line 1: There are unit tests for this file that should be updated: https://github.com/iree-org/iree/blob/main/build_tools/github_actions/configure_ci_test.py

Those tests might not be running on CI right now... some workflow refactoring may have excluded them...

kuhar (approved):
We should check the unit tests mentioned by Scott

build_tools/github_actions/configure_ci.py
amd-eochoalo: Thanks @ScottTodd , I took a look at them. It looks like two of them are failing. Both test_parse_trailer_map_multiline_trailer and test_parse_trailer_map_multiline_trailer_skip_first are failing. But these tests have been failing ever since introduced in https://github.com/iree-org/iree/pull/19244 (at least with my implementation of git, not sure if a different git implementation works differently.) I think we are fine deleting these two tests since they have been failing for a year without issue. I could parse the strings manually, let me know what you prefer. (In either case, I think this can be a follow up PR)

ScottTodd: Huh, those test cases pass for me, at least on Windows...

Fixing those can be a separate PR. This PR should just make sure that any relevant test cases for the modified code continue to pass (or add new tests as needed)

PR #23035, opened 2026-01-06
[Codegen/Common] Skip generating padding scf.forall loops when padding is effectively a no-op

### LINKED ISSUE #23016: [Codegen/Common] affine.delinearize_index op failure from degenerate padding loops during linalg.pack lowering
### Summary

Compiling a Qwen3 model with `--iree-opt-data-tiling=true` fails in Codegen with:
```
error: 'affine.delinearize_index' op no basis element may be statically non-positive
```
This appears to be triggered by degenerate / empty padding loops emitted during linalg.pack lowering when padding is provably unnecessary (e.g. padding size is zero or the padding iteration space is empty).

### Expected behavior

Compilation succeeds and produces output.vmfb.

### Actual behavior

Compilation failure with:

```
input.mlir:954:12: error: 'affine.delinearize_index' op no basis element may be statically non-positive
    %180 = linalg.batch_matmul ins(%reshape_57, %collapsed_58 : tensor<16x?x?xf32>, tensor<16x?x128xf32>) outs(%179 : tensor<16x?x128xf32>) -> tensor<16x?x128xf32>
           ^
input.mlir:954:12: note: see current operation: %31:3 = "affine.delinearize_index"(%arg0) <{static_basis = array<i64: 16, 0, 2>}> : (index) -> (index, index, index)
```

### Suspected root cause

The attached reproducer contains high-level IR. During codegen, the introduced `linalg.pack` operations with padding gets lowered into `iree_linalg_ext.map_scatter`, and padding materialization emits a separate distributed loop (e.g. `scf.forall`) to write padding values.

In cases where padding is provably a no-op (e.g. the packed dimension is already divisible by the tile size), the emitted padding loop can have a degenerate/empty iteration space (`lb >= ub`). This can later lead to invalid basis computation during affine delinearization (e.g., a 0 static basis element), triggering the failure above.

Example `linalg.pack` op pattern before it's lowered to `map_scatter`:

```
%inp = scf.forall (...) shared_outs(%arg3 = %10) -> (tensor<8x2x?x128xf16>) {
   ...
   linalg.generic
   ...
}
%collapsed = tensor.collapse_shape %inp {{.*}} : tensor<8x2x?x128xf16> into tensor<16x?x128xf16>
%pack = linalg.pack %collapsed
  padding_value(%cst : f16)
  outer_dims_perm = [0, 2, 1] 
  inner_dims_pos = [2, 1] 
  inner_tiles = [16, 1]
  into %dest : tensor<16x?x128xf16> -> tensor<16x8x?x16x1xf16>
  ...
```
When the packed dimension is already a multiple of the tile size (e.g. 128 divisible by 16), padding is a no-op, but padding loops are still emitted.

### Relevant existing behavior

`CombineLayoutTransformation` already contains a static check to avoid generating padding when it can prove padding is zero via constant/static reasoning (see: https://github.com/iree-org/iree/blob/f4b596d5b2a8813948f6dab18a36403e768eb5e1/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp#L371 ).

However, in this reproducer the relevant conditions are tied to dynamic dimensions, so the existing static check (and subsequent canonicalize/cse) does not reliably eliminate these padding loops, and degenerate/empty padding loops can survive into later passes.

### Potential fix

- Skip emitting padding loops when:
  - padding is provably zero, and/or
  - the padding loop iteration space is provably empty (lb >= ub)

This can be proven using ValueBoundsOpInterface which can discharge dynamic cases where purely static reasoning/canonicalization is insufficient.

Happy to reduce the reproducer further if needed :) 

1. Download attached [input.mlir](https://gist.github.com/phemashekar/6852a06fdeb73cc86ca2a6c10a80a28d)
2. Run
```
iree-compile input.mlir \
  --iree-hal-target-backends=llvm-cpu \
  --iree-opt-data-tiling=true \
  --iree-hal-local-host-device-backends=llvm-cpu \
  --iree-llvmcpu-target-cpu-features=host \
  --iree-llvmcpu-target-triple=x86_64-linux-gnu \
  --mlir-disable-threading \
  -o output.vmfb
```

Compiler

### Version information

- IREE commit: f4b596d5b2
- Backend: llvm-cpu
- OS: Linux (kernel 6.8.0-88-generic)

### Additional context

I have a patch that uses `ValueBoundsOpInterface` to prove padding is a no-op/empty-range and skips emitting degenerate padding loops in `CombineLayoutTransformation`. Filing this issue first to confirm the approach is acceptable before opening a PR.

phemashekar:
Fixes #23016

### TL;DR

Avoid generating padding materialization loops during `linalg.pack` lowering when the padding iteration space is provably empty, using `ValueBoundsOpInterface`.

### What changed?

- Use **ValueBounds** to prove padding is a no-op by proving that the derived half-open padding range `[lb, ub)` is empty (i.e., `lb >= ub`)
- Add helper predicate for “provably empty range”.
- Improve dimension canonicalization/resolution so no-op padding patterns are detected reliably.
- Skip emitting padding `scf.forall` when proven empty.
- Add tests where padding loops are not emitted when it's 0.

### How to test?

- Run

```
iree-opt compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir --pass-pipeline="builtin.module(func.func(iree-codegen-combine-layout-transformation{scope=dispatch}, canonicalize, cse))" --split-input-file
```

- The tests check no padding `scf.forall` is emitted when it isn't required

### Why?

Eliminating empty/degenerate padding loops:

- reduces IR bloat
- avoids issues from degenerate/empty iteration spaces (e.g., `scf.forall` with `lb == ub`) that can later materialize as invalid bases during index delinearization, causing failures in downstream passes.
- helps dynamic-shape cases where `ValueBounds` can prove padding is unnecessary even when static analysis can’t.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -18,7 +18,6 @@
 #include "mlir/IR/IRMapping.h"
 #include "mlir/Interfaces/ValueBoundsOpInterface.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
-#include "mlir/Transforms/Passes.h"
 
 #define DEBUG_TYPE "iree-codegen-combine-layout-transformation"
 
@@ -377,41 +376,38 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
                              Value ub) -> bool {
     using presburger::BoundType;
 
-    // Materialize diff = ub - lb and ask for a constant upper bound on diff.
+    // Model diff = ub - lb  as an affine map (d0, d1) -> (d0 - d1)
+    // and ask for a constant upper bound on diff.
     // If we can prove UB(diff) <= 0, then diff <= 0 => ub - lb <= 0 => lb >=
     // ub.
-    Value diff = arith::SubIOp::create(rewriter, loc, ub, lb);
+    MLIRContext *ctx = rewriter.getContext();
+    auto diffMap = AffineMap::get(
+        /*dimCount=*/2, /*symbolCount=*/0,
+        {getAffineDimExpr(0, ctx) - getAffineDimExpr(1, ctx)}, ctx);
 
     auto ubConstantBound = ValueBoundsConstraintSet::computeConstantBound(
-        BoundType::UB, ValueBoundsConstraintSet::Variable(diff),
+        BoundType::UB,
+        ValueBoundsConstraintSet::Variable(diffMap, ValueRange{ub, lb}),
         /*stopCondition=*/stopConditionFn,
         /*addConservativeSemiAffineBounds=*/true);
 
     return succeeded(ubConstantBound) && *ubConstantBound <= 0;
   };
 
-  // Prove v == 0 by proving both:
-  //   LB(v) == 0  and  UB(v) == 0
+  // Prove v == 0:
   //
   // This is intentionally strict (see stopConditionFn): if ValueBounds cannot
-  // prove either bound, it returns failure and we treat v as "not provably
+  // prove v == 0 exactly, it returns failure and we treat v as "not provably
   // zero".
   auto proveEqZero = [&](Value v) -> bool {
     using presburger::BoundType;
 
-    // Query constant upper bound of v.
-    FailureOr<int64_t> ub = ValueBoundsConstraintSet::computeConstantBound(
-        BoundType::UB, ValueBoundsConstraintSet::Variable(v),
+    FailureOr<int64_t> eq = ValueBoundsConstraintSet::computeConstantBound(
+        BoundType::EQ, ValueBoundsConstraintSet::Variable(v),
         /*stopCondition=*/stopConditionFn,
         /*addConservativeSemiAffineBounds=*/true);
 
-    // Query constant lower bound of v.
-    FailureOr<int64_t> lb = ValueBoundsConstraintSet::computeConstantBound(
-        BoundType::LB, ValueBoundsConstraintSet::Variable(v),
-        /*stopCondition=*/stopConditionFn,
-        /*addConservativeSemiAffineBounds=*/true);
-
-    return succeeded(ub) && succeeded(lb) && *ub == 0 && *lb == 0;
+    return succeeded(eq) && *eq == 0;
   };
 
   auto isProvablyZeroOfr = [&](OpFoldResult ofr) -> bool {
@@ -475,9 +471,8 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
                                    ubs, distConfigs, innerLoopBuilder);
     }
 
-    // If the pad amount is provably zero, skip generating loops.
-    // This reduces IR bloat and avoids degenerate scf.forall bounds for
-    // padding.
+    // Fast-path: if the pad amount is trivially/provably zero, skip generating
+    // loops.
     if (isProvablyZeroOfr(high)) {
       continue;
     }
@@ -694,16 +689,29 @@ combineLayoutTransformation(MLIRContext *ctx, FunctionOpInterface funcOp,
   // IR. This improves pattern matching and makes it easier for subsequent
   // rewrites/proofs to recognize identity/no-op cases.
   {
-    RewritePatternSet dimResolve(ctx);
-    memref::populateResolveRankedShapedTypeResultDimsPatterns(dimResolve);
-    if (failed(applyPatternsGreedily(funcOp, std::move(dimResolve)))) {
-      return failure();
+    RewritePatternSet patterns(ctx);
+    memref::populateResolveRankedShapedTypeResultDimsPatterns(patterns);
+
+    // TODO(phemashekar): This collects canonicalization patterns for all
+    // ops/dialects present in the function IR. Identify the minimal required
+    // subset for the ValueBounds-driven proofs and switch to an explicit
+    // pattern list.
+    DenseSet<Dialect *> dialectsInIR;
+    DenseSet<OperationName> opsInIR;
+    funcOp->walk([&](Operation *op) {
+      dialectsInIR.insert(op->getDialect());
+      opsInIR.insert(op->getName());
+    });
+
+    for (Dialect *dialect : dialectsInIR) {
+      dialect->getCanonicalizationPatterns(patterns);
     }
 
-    PassManager pm(ctx);
-    pm.addPass(createCanonicalizerPass());
-    pm.addPass(createCSEPass());
-    if (failed(pm.run(funcOp))) {
+    for (OperationName opName : opsInIR) {
+      opName.getCanonicalizationPatterns(patterns, ctx);
+    }
+
+    if (failed(applyPatternsGreedily(funcOp, std::move(patterns)))) {
       return failure();
     }
   }
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -369,24 +369,22 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
   //   ub = resultSize
   // and for a well-formed pad, resultSize = srcSize + low + high, so:
   //   ub - lb = high
-  // Therefore "lb >= ub" => "high == 0". Expressing the check in terms of lb/ub
+  // Therefore "lb >= ub" -> "high == 0". Expressing the check in terms of lb/ub
   // keeps it robust even when pad amount has been rewritten into a complex
   // form.
   auto proveEmptyRange = [&](RewriterBase &rewriter, Location loc, Value lb,
                              Value ub) -> bool {
-    using presburger::BoundType;
-
     // Model diff = ub - lb  as an affine map (d0, d1) -> (d0 - d1)
     // and ask for a constant upper bound on diff.
-    // If we can prove UB(diff) <= 0, then diff <= 0 => ub - lb <= 0 => lb >=
-    // ub.
+    // If we can prove UB(diff) <= 0, then "diff <= 0" -> "ub - lb <= 0" -> 
+    // "lb >= ub".
     MLIRContext *ctx = rewriter.getContext();
     auto diffMap = AffineMap::get(
         /*dimCount=*/2, /*symbolCount=*/0,
         {getAffineDimExpr(0, ctx) - getAffineDimExpr(1, ctx)}, ctx);
 
     auto ubConstantBound = ValueBoundsConstraintSet::computeConstantBound(
-        BoundType::UB,
+        presburger::BoundType::UB,
         ValueBoundsConstraintSet::Variable(diffMap, ValueRange{ub, lb}),
         /*stopCondition=*/stopConditionFn,
         /*addConservativeSemiAffineBounds=*/true);
@@ -400,10 +398,9 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
   // prove v == 0 exactly, it returns failure and we treat v as "not provably
   // zero".
   auto proveEqZero = [&](Value v) -> bool {
-    using presburger::BoundType;
 
     FailureOr<int64_t> eq = ValueBoundsConstraintSet::computeConstantBound(
-        BoundType::EQ, ValueBoundsConstraintSet::Variable(v),
+        presburger::BoundType::EQ, ValueBoundsConstraintSet::Variable(v),
         /*stopCondition=*/stopConditionFn,
         /*addConservativeSemiAffineBounds=*/true);
 
@@ -411,12 +408,9 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
   };
 
   auto isProvablyZeroOfr = [&](OpFoldResult ofr) -> bool {
-    // static 0?
     if (isConstantIntValue(ofr, 0)) {
       return true;
     }
-
-    // dynamic %v == 0?
     if (auto v = dyn_cast<Value>(ofr)) {
       return proveEqZero(v);
     }
@@ -458,7 +452,7 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
   // computation in order to have a more simplified kernel.
   for (auto [idx, low, high] :
        llvm::enumerate(padOp.getMixedLowPad(), padOp.getMixedHighPad())) {
-    // Create a distributed loop for the low padding if low pad is non-zero
+    // Create a distributed loop for the low padding if low pad is non-zero.
     if (!isProvablyZeroOfr(low)) {
       SmallVector<OpFoldResult> ubs(padResultSizes);
       SmallVector<OpFoldResult> lbs(ubs.size(), rewriter.getIndexAttr(0));
@@ -477,7 +471,7 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
       continue;
     }
 
-    // Create a distributed loop for the high padding
+    // Create a distributed loop for the high padding.
     SmallVector<OpFoldResult> ubs(padResultSizes);
     SmallVector<OpFoldResult> lbs(ubs.size(), rewriter.getIndexAttr(0));
     SmallVector<int64_t> shape(padOp.getSourceType().getShape());
@@ -699,7 +693,8 @@ combineLayoutTransformation(MLIRContext *ctx, FunctionOpInterface funcOp,
     DenseSet<Dialect *> dialectsInIR;
     DenseSet<OperationName> opsInIR;
     funcOp->walk([&](Operation *op) {
-      dialectsInIR.insert(op->getDialect());
+      if (Dialect *dialect = op->getDialect())
+        dialectsInIR.insert(dialect);
       opsInIR.insert(op->getName());
     });
 
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -18,6 +18,7 @@
 #include "mlir/IR/IRMapping.h"
 #include "mlir/Interfaces/ValueBoundsOpInterface.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Passes.h"
 
 #define DEBUG_TYPE "iree-codegen-combine-layout-transformation"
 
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -332,6 +334,75 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+    FailureOr<int64_t> ub = ValueBoundsConstraintSet::computeConstantBound(
+        BoundType::UB, ValueBoundsConstraintSet::Variable(v),
+        /*stopCondition=*/stopConditionFn,
+        /*addConservativeSemiAffineBounds=*/true);
+
+    // Query constant lower bound of v.
+    FailureOr<int64_t> lb = ValueBoundsConstraintSet::computeConstantBound(
+        BoundType::LB, ValueBoundsConstraintSet::Variable(v),
+        /*stopCondition=*/stopConditionFn,
+        /*addConservativeSemiAffineBounds=*/true);
```
Max191: You can use `BoundType::EQ` and just do a single check.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -332,6 +334,75 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+  // keeps it robust even when pad amount has been rewritten into a complex
+  // form.
+  auto proveEmptyRange = [&](RewriterBase &rewriter, Location loc, Value lb,
+                             Value ub) -> bool {
+    using presburger::BoundType;
+
+    // Materialize diff = ub - lb and ask for a constant upper bound on diff.
+    // If we can prove UB(diff) <= 0, then diff <= 0 => ub - lb <= 0 => lb >=
+    // ub.
+    Value diff = arith::SubIOp::create(rewriter, loc, ub, lb);
```
Max191: It is a little bit strange to be creating IR just as the anchor for value bounds like this. I think it works in this case, because the transform will not return failure afterwards, but there is a better way to do this. Could you try using an affine map variable:

https://github.com/llvm/llvm-project/blob/31b93d6e38dea9e7023101905d3471552e985540/mlir/include/mlir/Interfaces/ValueBoundsOpInterface.h#L138

You can create an affine map with a single result that represents that subtraction expression, and then use that along with the Value operands to the subtraction expression as the ValueBoundsConstraintSet variable.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -379,15 +450,41 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+      Value lbV = getValueOrCreateConstantIndexOp(rewriter, loc, lbs[idx]);
+      Value ubV = getValueOrCreateConstantIndexOp(rewriter, loc, ubs[idx]);
+
+      if (lbV == ubV) {
+        continue;
+      }
+
+      if (proveEmptyRange(rewriter, loc, lbV, ubV)) {
+        continue;
+      }
```
Max191, line 468: If you are checking for the empty range here, then why do you need to check for the zero high padding above?

Can we just have one or the other (I would assume checking for empty range is more powerful), and use the same one for checking both high and low pad?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -569,6 +666,24 @@ combineLayoutTransformation(MLIRContext *ctx, FunctionOpInterface funcOp,
...
+  {
+    RewritePatternSet dimResolve(ctx);
+    memref::populateResolveRankedShapedTypeResultDimsPatterns(dimResolve);
+    if (failed(applyPatternsGreedily(funcOp, std::move(dimResolve)))) {
+      return failure();
+    }
+
+    PassManager pm(ctx);
+    pm.addPass(createCanonicalizerPass());
+    pm.addPass(createCSEPass());
```
Max191: I think what you are looking for is `mlir::eliminateCommonSubExpressions`. You shouldn't need to create and run new passes here. Also, what is CSE needed for here? I wouldn't expect it to be needed in order for value bounds to work.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -569,6 +666,24 @@ combineLayoutTransformation(MLIRContext *ctx, FunctionOpInterface funcOp,
...
+  // rewrites/proofs to recognize identity/no-op cases.
+  {
+    RewritePatternSet dimResolve(ctx);
+    memref::populateResolveRankedShapedTypeResultDimsPatterns(dimResolve);
+    if (failed(applyPatternsGreedily(funcOp, std::move(dimResolve)))) {
+      return failure();
+    }
+
+    PassManager pm(ctx);
+    pm.addPass(createCanonicalizerPass());
```
Max191: For the canonicalizer, just run the canonicalization patterns that you need. The canonicalizer pass runs a lot of things, and people are always adding new things to it. This is more expensive than necessary, and we have less control over what is actually being run. Also, if possible, you should be running the canonicalization patterns at the same time as the resolve dims patterns above. That way the patterns can run in parallel.

Max191 (changes requested):

phemashekar:
Thanks for the detailed feedback! I’ll address these comments and update the PR soon.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: Good catch, I missed it somehow! I switched the check to use BoundType::EQ so we only run a single bound query.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: Agreed. I updated `proveEmptyRange` to use an affine-map-backed `ValueBoundsConstraintSet::Variable` to represent that without creating any transient IR.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: The empty-range proof does subsume the exact-zero case for well-formed padding, but I’m keeping the explicit `isProvablyZeroOfr(high)` check as a fast-path.

`isProvablyZeroOfr` first catches the common static 0 case and then a direct ValueBounds EQ proof on the pad amount. The empty-range check is more general but is slightly expensive and hence it's being used as a fallback mechanism incase it fails the first time

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: You're right, CSE isn't required for value bounds. I added it initially thinking that it might improve pattern-matching and indirectly help value bounds driven proofs. I’ll drop it to keep this transform lighter.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: Makes sense. I removed the nested PassManager and no longer run the full pass. Instead, I collect canonicalization patterns only from the ops/dialects present in this functions IR and apply them locally (alongside the resolve-dims patterns). This is narrower than running a full canonicalization pass, and I added a TODO to narrow this further to an explicit set of op-specific patterns.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -392,8 +434,8 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
   // computation in order to have a more simplified kernel.
   for (auto [idx, low, high] :
        llvm::enumerate(padOp.getMixedLowPad(), padOp.getMixedHighPad())) {
-    // Create a distributed loop for the low padding
-    if (!isConstantIntValue(low, 0)) {
+    // Create a distributed loop for the low padding if low pad is non-zero
```
hanhanW:
```suggestion
    // Create a distributed loop for the low padding if low pad is non-zero.
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -404,15 +446,40 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
-    // Create a distributed loop for the high padding
-    if (isConstantIntValue(high, 0)) {
+
+    // Fast-path: if the pad amount is trivially/provably zero, skip generating
+    // loops.
+    if (isProvablyZeroOfr(high)) {
       continue;
     }
+
+    // Create a distributed loop for the high padding
```
hanhanW:
```suggestion
    // Create a distributed loop for the high padding.
```

compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
```diff
@@ -433,19 +433,47 @@ func.func @consumer_unfusable_due_to_init(%arg0: tensor<?xi32>, %arg1: tensor<?x
 //       DISPATCH-SCOPE:       tensor.parallel_insert_slice
 //       DISPATCH-SCOPE:   linalg.generic
 
+
+// -----
+
+// Tests that no padding scf.forall loops are emitted when
+// padding in linalg.pack is effectively a no-op
```
hanhanW:
```suggestion
// Tests that no padding scf.forall loops are emitted when
// padding in linalg.pack is effectively a no-op.
```

compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
```diff
@@ -433,19 +433,47 @@ func.func @consumer_unfusable_due_to_init(%arg0: tensor<?xi32>, %arg1: tensor<?x
...
+    outer_dims_perm = [0, 2, 1]
+    inner_dims_pos = [2, 1]
+    inner_tiles = [16, 1]
+    into %dest : tensor<16x?x128xf16> -> tensor<16x8x?x16x1xf16>
+  iree_codegen.store_to_buffer %pack, %result : tensor<16x8x?x16x1xf16> into memref<16x8x?x16x1xf16>
+  return
+}
+// DISPATCH-SCOPE-LABEL: @pack_dynamic_dim_tile_size_1_no_pad_loop
+//       DISPATCH-SCOPE:   iree_linalg_ext.map_scatter
+// Verify no padding loops are generated
```
hanhanW:
```suggestion
// Verify no padding loops are generated.
```

compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation.mlir
```diff
@@ -433,19 +433,47 @@ func.func @consumer_unfusable_due_to_init(%arg0: tensor<?xi32>, %arg1: tensor<?x
...
+    into %dest : tensor<16x?xf16> -> tensor<2x?x8x1xf16>
+  iree_codegen.store_to_buffer %pack, %result : tensor<2x?x8x1xf16> into memref<2x?x8x1xf16>
   return
 }
-// DISPATCH-SCOPE-LABEL: func @unpack_no_padding_no_masking
-// DISPATCH-SCOPE: iree_linalg_ext.map_scatter
-// DISPATCH-SCOPE-NOT: arith.cmpi ult
+// DISPATCH-SCOPE-LABEL: @pack_divisible_static_dim_tile_size_8_no_pad_loop
+//       DISPATCH-SCOPE:   iree_linalg_ext.map_scatter
+// Verify no padding loops are generated
```
hanhanW:
```suggestion
// Verify no padding loops are generated.
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -357,6 +333,72 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+  };
+
+  // Prove that range [lb, ub) is empty (lb >= ub).
+  //
+  // For high padding, we have:
+  //   lb = low + srcSize
+  //   ub = resultSize
+  // and for a well-formed pad, resultSize = srcSize + low + high, so:
+  //   ub - lb = high
+  // Therefore "lb >= ub" => "high == 0". Expressing the check in terms of lb/ub
```
hanhanW: Let's use `->` to differentiate between `imply` and `inequalities`.


```suggestion
  // Therefore "lb >= ub" -> "high == 0". Expressing the check in terms of lb/ub
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -357,6 +333,72 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+  // keeps it robust even when pad amount has been rewritten into a complex
+  // form.
+  auto proveEmptyRange = [&](RewriterBase &rewriter, Location loc, Value lb,
+                             Value ub) -> bool {
+    using presburger::BoundType;
+
+    // Model diff = ub - lb  as an affine map (d0, d1) -> (d0 - d1)
+    // and ask for a constant upper bound on diff.
+    // If we can prove UB(diff) <= 0, then diff <= 0 => ub - lb <= 0 => lb >=
+    // ub.
```
hanhanW: ditto

```suggestion
    // If we can prove UB(diff) <= 0, then "diff <= 0" -> "ub - lb <= 0" ->
    // "lb >= ub".
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -357,6 +333,72 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+    if (isConstantIntValue(ofr, 0)) {
+      return true;
+    }
+
+    // dynamic %v == 0?
+    if (auto v = dyn_cast<Value>(ofr)) {
+      return proveEqZero(v);
+    }
+    return false;
+  };
```
hanhanW: nit: I'd remove the comment and blankline for this lambda. The code is self-documenting to me.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -357,6 +333,72 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+    // and ask for a constant upper bound on diff.
+    // If we can prove UB(diff) <= 0, then diff <= 0 => ub - lb <= 0 => lb >=
+    // ub.
+    MLIRContext *ctx = rewriter.getContext();
+    auto diffMap = AffineMap::get(
+        /*dimCount=*/2, /*symbolCount=*/0,
+        {getAffineDimExpr(0, ctx) - getAffineDimExpr(1, ctx)}, ctx);
+
+    auto ubConstantBound = ValueBoundsConstraintSet::computeConstantBound(
+        BoundType::UB,
```
hanhanW: Maybe just use `presburger::BoundType::UB` here, then we don't need the using statement?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -357,6 +333,72 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+  auto proveEqZero = [&](Value v) -> bool {
+    using presburger::BoundType;
+
+    FailureOr<int64_t> eq = ValueBoundsConstraintSet::computeConstantBound(
+        BoundType::EQ, ValueBoundsConstraintSet::Variable(v),
+        /*stopCondition=*/stopConditionFn,
+        /*addConservativeSemiAffineBounds=*/true);
+
+    return succeeded(eq) && *eq == 0;
+  };
```
hanhanW: Same, maybe just use `presburger::BoundType::EQ`; we can remove the blank lines.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -594,6 +661,37 @@ combineLayoutTransformation(MLIRContext *ctx, FunctionOpInterface funcOp,
...
+      opsInIR.insert(op->getName());
+    });
+
+    for (Dialect *dialect : dialectsInIR) {
+      dialect->getCanonicalizationPatterns(patterns);
+    }
+
+    for (OperationName opName : opsInIR) {
+      opName.getCanonicalizationPatterns(patterns, ctx);
+    }
```
hanhanW: I guess we only need the canonicalization patterns from tensor dialect, IREECodegen dialect, and Linalg dialect.

hanhanW (commented):
Thanks, I mostly have style nits.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: Thanks much for your feedback. I tried restricting canonicalization to the above mentioned dialects only (tensor/linalg/iree_codegen), but it wasn’t sufficient as the original failure comes back. My understanding is that dialect-wide canonicalization patterns are additive and don’t include op-level canonicalizations (which still cover a large portion of simplifications), and this function’s IR also contains ops from other dialects (e.g. arith/affine) that influence the shape/bounds expressions ValueBounds reasons about. I have added a TODO comment to further narrow it down to specific set of ops so that I can restrict the patterns to only those ops.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: Makes sense!

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Max191: This is not much different from running a full canonicalizer pass. I think we should still restrict the amount of canonicalization patterns that are being run here. The typical style in IREE is to statically add canonicalization patterns for the specific ops that we know we need.  Here is an example for reshape propagation: https://github.com/iree-org/iree/blob/9bcc19e0b4b2d704e632e4dcff51d3ee78412958/compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp#L418-L425
Dynamically walking the IR and adding the canonicalization patterns for every op and dialect can pull in a lot of unnecessary patterns.

Can you try limiting this to a smaller subset of specific operations that are able to handle your case?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Max191: Is it really that much less expensive? Even if the problem is simpler, I'd expect there to still be significant overhead from calling the value bounds interface. If you want a faster check that works in simple static cases, then how about just using the `isConstantIntValue` matcher and avoiding the value bounds check?

The issue with calling it twice is that the value bounds analysis is not cacheable, so calling it twice usually means roughly twice the overhead.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -357,6 +357,66 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+  // This is intentionally strict (see stopConditionFn): if ValueBounds cannot
+  // prove v == 0 exactly, it returns failure and we treat v as "not provably
+  // zero".
+  auto proveEqZero = [&](Value v) -> bool {
+
+    FailureOr<int64_t> eq = ValueBoundsConstraintSet::computeConstantBound(
+        presburger::BoundType::EQ, ValueBoundsConstraintSet::Variable(v),
+        /*stopCondition=*/stopConditionFn,
+        /*addConservativeSemiAffineBounds=*/true);
+
```
Max191: nit: Please remove these blank lines, since I don't think they help readability.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -357,6 +357,66 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
...
+    auto diffMap = AffineMap::get(
+        /*dimCount=*/2, /*symbolCount=*/0,
+        {getAffineDimExpr(0, ctx) - getAffineDimExpr(1, ctx)}, ctx);
+
+    auto ubConstantBound = ValueBoundsConstraintSet::computeConstantBound(
+        presburger::BoundType::UB,
+        ValueBoundsConstraintSet::Variable(diffMap, ValueRange{ub, lb}),
+        /*stopCondition=*/stopConditionFn,
+        /*addConservativeSemiAffineBounds=*/true);
+
```
Max191: nit: Please remove these blank lines too. The function is pretty small, so I don't think the blank lines improve readability.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
hanhanW: I'm considering if we should break this pass into two passes, and we run canonicalizer in between. We might need to implement the CPU specific version of CombineLayoutTransformation, because we may want different simplification. E.g., a pack op can be decomposed only if it does not have a lowering config. Otherwise, we fuse the op directly, instead of folding it into map_scatter.

btw, why do we need `populateResolveRankedShapedTypeResultDimsPatterns` patterns? Are we creating ops without making insertion point happy?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: yep that makes sense! I've removed it in the updated version

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: Thanks for pushing on this! After digging into it, it turned out that only a single op-level canonicalization pattern was actually needed here. I’ve updated the code and it's much simpler now :)

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -392,8 +426,8 @@ foldPadIntoMapScatter(RewriterBase &rewriter, tensor::PadOp padOp,
   // computation in order to have a more simplified kernel.
   for (auto [idx, low, high] :
        llvm::enumerate(padOp.getMixedLowPad(), padOp.getMixedHighPad())) {
-    // Create a distributed loop for the low padding
-    if (!isConstantIntValue(low, 0)) {
+    // Create a distributed loop for the low padding.
+    if (!(isConstantIntValue(low, 0))) {
```
hanhanW: We can remove one `()` pair, right?

hanhanW (commented):
I only have a final nit.

hanhanW (approved):

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
phemashekar: yep sorry, missed it!

Max191 (approved):
Nice, LGTM now!

phemashekar:
Thanks all! I don't have write access, could someone please merge when convenient? :)

Max191:
@phemashekar I think the description is a little out of date after addressing the review comments. If you update the description then I'll go ahead and merge for you.

phemashekar:
> @phemashekar I think the description is a little out of date after addressing the review comments. If you update the description then I'll go ahead and merge for you.

Thanks much, updated the description now!

Issue #22700, opened 2025-11-19
[LLVMGPU] Targeting transpose load instructions


Max191:
# Overview + Background #
This issue outlines a plan for targeting the transpose load instructions in MI350+.

The goal is to improve codegen for non transpose-b matmul layouts. Most of our codegen thus far has been focused on matmul_transpose_b variants (`MxK @ NxK`), but as we focus more on finer grained kernel generation, we need to improve variants where the K dimension is not the most contiguous dimension anymore. One relatively low-hanging improvement we can make is to use transpose load instructions to reduce the number of LDS transactions for the transposed operands.

## Current Codegen ##

In today's codegen, for basic `MxK @ KxN` matmuls (at a very high level), we will generate code that does the following:
1. Threads cooperatively copy LHS and RHS tiles of size `MxK` and `KxN` into shared memory, *keeping the original 2D layouts*. This is just a simple copy.
2. Threads load the data they need for the MFMA instruction(s) and execute MFMA.
    - Any packing/transposition to get into the correct MFMA layout happens as part of this load. Typically, this means the load of the RHS operand will not be contiguous, and will be unrolled into a bunch of single element loads.
3. Write the final result to global memory.

## How can transpose load help us? ##

Transpose loads can be used in step 2 in order to avoid unrolling the LDS loads into a bunch of scalar loads. The transpose load instructions work by having each thread index into a contiguous vector of data, but the loaded data will end up spread across different threads' VGPRs based on the transposed layout. On MI350, there are 4 instructions (`B4`, `B6`, `B8`, `B16`), but we can look at an example for the 8 bit element instruction:

For the `B8` transpose load instruction, each thread will be loading 64 bits of data, or 8 elements, and then those 8 elements will be spread across 8 different VGPRs once the instruction completes. AFAICS, once data is fetched from LDS, the transposition happens over 4 cycles in groups of 16 threads. Each group of 16 threads performs their own transpose independently from the other groups, so the data contained loaded (indexed) by threads 0-15 will remain in threads 0-15 after transposition (same for 16-31, 32-47, 48-63). For the 8 bit element type instruction (with 64 bit loads), this means that each group of 16 threads will be responsible for transposing a block of `8x16` 8 bit elements. The mapping for which thread loads (indexes) each element, and which thread receives (after transposition) each element is shown by the tables below:

### Data Loaded ###

[...]

PR #23037, opened 2026-01-06
[Codegen] Add index_hint op and lane hint attributes for GPU optimization

Max191:
Add an op to annotate index values with semantic information about how they vary across GPU lanes. This enables later optimization opportunities that have index pattern requirements like transpose loads.

The `iree_codegen.index_hint` op is a pure pass-through that wraps an index with a hint attribute. Two GPU-specific hint attributes are added:
- `#iree_gpu.lane_constant<N>`: index is uniform within N-lane groups
- `#iree_gpu.lane_increment<N>`: index increments by 1 within N-lane groups

These hints are always safe to remove and serve purely as optimization guides. The immediate use case is enabling `amdgpu.transpose_load` lowering, where row indices must be uniform and column indices must increment across lanes within 16-lane groups.

This is the final state of the index hint op proposed in https://github.com/iree-org/iree/issues/22700.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
@@ -585,4 +585,34 @@ def IREECodegen_WorkgroupCountHintOp :
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// IndexHintOp
+//===----------------------------------------------------------------------===//
+
+def IREECodegen_IndexHintOp : Op<IREECodegen_Dialect, "index_hint", [Pure]> {
+  let summary = "Compiler hint providing semantic information about an index";
+  let description = [{
+    Pure pass-through operation that annotates an index value with semantic
+    information about how it varies across parallel workers (e.g., GPU lanes).
+
+    The hint attribute describes the index behavior. Common hints include:
+    - `#iree_gpu.lane_constant<N>`: Index is uniform within groups of N lanes
+    - `#iree_gpu.lane_increment<N>`: Index increments by 1 within groups of N lanes
+
+    This operation is always safe to remove (replace with input). It exists
+    purely to guide optimization passes.
+
+    Example:
+    ```mlir
+    %row = iree_codegen.index_hint %idx {hint = #iree_gpu.lane_constant<16>} : index
+    %col = iree_codegen.index_hint %idx {hint = #iree_gpu.lane_increment<16>} : index
+    ```
+  }];
+
+  let arguments = (ins Index:$input, AnyAttr:$hint);
+  let results = (outs Index:$result);
+
+  let assemblyFormat = "$input ` ` `{` `hint` `=` $hint `}` attr-dict `:` type($input)";
+}
+
 #endif // IREE_CODEGEN_DIALECT_IREECODEGENOPS
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
@@ -604,15 +604,15 @@ def IREECodegen_IndexHintOp : Op<IREECodegen_Dialect, "index_hint", [Pure]> {
 
     Example:
     ```mlir
-    %row = iree_codegen.index_hint %idx {hint = #iree_gpu.lane_constant<16>} : index
-    %col = iree_codegen.index_hint %idx {hint = #iree_gpu.lane_increment<16>} : index
+    %row = iree_codegen.index_hint %idx(#iree_gpu.lane_constant<16>) : index
+    %col = iree_codegen.index_hint %idx(#iree_gpu.lane_increment<16>) : index
     ```
   }];
 
   let arguments = (ins Index:$input, AnyAttr:$hint);
   let results = (outs Index:$result);
 
-  let assemblyFormat = "$input ` ` `{` `hint` `=` $hint `}` attr-dict `:` type($input)";
+  let assemblyFormat = "$input `(` $hint `)` attr-dict `:` type($input)";
 }
 
 #endif // IREE_CODEGEN_DIALECT_IREECODEGENOPS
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/test/roundtrip.mlir
```diff
@@ -82,3 +82,38 @@ func.func @fusion_barrier(%arg0: tensor<?xf32>) -> tensor<?xf32> {
...
+func.func @index_hint_multiple(%idx0: index, %idx1: index, %idx2: index) -> (index, index, index) {
+  %row0 = iree_codegen.index_hint %idx0 {hint = #iree_gpu.lane_constant<16>} : index
+  %row1 = iree_codegen.index_hint %idx1 {hint = #iree_gpu.lane_constant<16>} : index
+  %col = iree_codegen.index_hint %idx2 {hint = #iree_gpu.lane_increment<16>} : index
+  return %row0, %row1, %col : index, index, index
+}
+// CHECK-LABEL: func.func @index_hint_multiple(
+// CHECK:         iree_codegen.index_hint {{.*}} {hint = #iree_gpu.lane_constant<16>} : index
+// CHECK:         iree_codegen.index_hint {{.*}} {hint = #iree_gpu.lane_constant<16>} : index
+// CHECK:         iree_codegen.index_hint {{.*}} {hint = #iree_gpu.lane_increment<16>} : index
```
Max191: Claude added this test and I forgot it was here. It doesn't really add anything useful, so I will remove it before landing.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -585,4 +585,34 @@ def IREECodegen_WorkgroupCountHintOp :
...
+    purely to guide optimization passes.
+
+    Example:
+    ```mlir
+    %row = iree_codegen.index_hint %idx {hint = #iree_gpu.lane_constant<16>} : index
+    %col = iree_codegen.index_hint %idx {hint = #iree_gpu.lane_increment<16>} : index
+    ```
+  }];
+
+  let arguments = (ins Index:$input, AnyAttr:$hint);
```
krzysz00, line 612: Uber-nit: You could get almost the same assembly format with `prop-dict attr-dict`

krzysz00 (approved):
I figure this makes sense overall so I can't see any reason not to land this, especially if there's a PR stack showing how it'll be used.

qedawkins (approved):
Will we be getting folders for adds/muls later or any other op definition changes from the rest of the pr stack? If so it's probably worth just adding now.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/test/roundtrip.mlir
```diff
@@ -82,3 +82,38 @@ func.func @fusion_barrier(%arg0: tensor<?xf32>) -> tensor<?xf32> {
 // CHECK-LABEL: func.func @fusion_barrier(
 // CHECK-SAME:    %[[ARG0:[a-zA-Z0-9_]+]]: tensor<?xf32>
 // CHECK:         iree_codegen.fusion_barrier %[[ARG0]] : tensor<?xf32>
+
+// -----
+
+func.func @index_hint_lane_constant(%idx: index) -> index {
+  %hinted = iree_codegen.index_hint %idx {hint = #iree_gpu.lane_constant<16>} : index
```
qedawkins: The asm format is redundant here, it's called `index_hint` and has a single `hint` field. Also by adding braces explicitly, the attribute dictionary parses a second set of braces which is misleading when trying to add discardable attrs. I would simplify it to
```
iree_codegen.index_hint %idx = #iree_gpu.lane_constant<16> : index
```
```

File: compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td [hunk 4]
```diff
@@ -959,4 +959,40 @@ def IREEGPU_GPUPipelineOptionsAttr : AttrDef<IREEGPU_Dialect, "GPUPipelineOption
...
+  let parameters = (ins "int64_t":$group_size);
+  let assemblyFormat = "`<` $group_size `>`";
+}
+
+def IREEGPU_LaneIncrementAttr : AttrDef<IREEGPU_Dialect, "LaneIncrement"> {
+  let mnemonic = "lane_increment";
+  let cppNamespace = "::mlir::iree_compiler::IREE::GPU";
+  let summary = "Index increments by 1 within groups of N lanes";
+  let description = [{
+    This attribute indicates that an index value increments by 1 within groups
```
qedawkins: Can we add the increment value as a default valued parameter as well. That way we can express contiguous vectorized loads as `<(group size), (vector width)>`.

qedawkins (commented):
Actually a couple comments

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -585,4 +585,34 @@ def IREECodegen_WorkgroupCountHintOp :
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// IndexHintOp
+//===----------------------------------------------------------------------===//
+
+def IREECodegen_IndexHintOp : Op<IREECodegen_Dialect, "index_hint", [Pure]> {
```
qedawkins, line 592: One more comment, since this is GPU only this op + the attributes should go in the iree_gpu dialect

Max191:
> Will we be getting folders for adds/muls later or any other op definition changes from the rest of the pr stack?

Not in this stack. I don't need to do anything complicated with it right now.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
Max191: Oh, I missed this comment. Nothing about the op is specific to GPU. It could be used anywhere. The attributes are GPU specific, though, so they are in the iree_gpu dialect.

qedawkins (approved):

krzysz00 (commented):
I'd like to flag that this should come with at least the trivial implementations of the two integer range inference mechanisms that just pass through argument information

Max191:
> I'd like to flag that this should come with at least the trivial implementations of the two integer range inference mechanisms that just pass through argument information

Interpreting these hints through the integer range analysis is not included in my PR stack. All that is needed is some trivial matching for now. I want to expand on what can be done with the hint ops later, but it's not in the scope of this PR.

krzysz00:
To clarify, the issue I'm worried about is that, if these ops don't get erased by the time integer analysis runs, they'll block information about the argument they're passing through.

Issue #23038, opened 2026-01-06
[LinalgExt][Codegen] iree_linalg_ext.map_gather Design Doc


Max191:
# Overview #

The `iree_linalg_ext.map_scatter` op was added to handle complex series of layout remapping operations (like pack, transpose, reshapes, etc.) before writing back to (global) memory. Now, we have some use cases for the inverse map_gather operation, which should behave very similarly, and can be used to handle complex series of layout remapping operations upon reading from memory.

## Current Use Case ##

The first use for map_gather will likely be in convolution code generation, where we propagate packed intrinsic layouts to elementwise consumers with multiple operands, like bias adds.

This leads to IR that looks like the following:
```mlir
%10 = iree_tensor_ext.dispatch.tensor.load %3 ... -> tensor<1x50x512xf32>
scf.forall // workgroup loop
    %extracted_slice = tensor.extract_slice %10  ... : tensor<1x50x512xf32> to tensor<1x50x64xf32>
    %padded = tensor.pad %extracted_slice low[0, 0, 0] high[0, 14, 0] {
    ^bb0(%arg3: index, %arg4: index, %arg5: index):
      tensor.yield %0 : f32
    } : tensor<1x50x64xf32> to tensor<1x64x64xf32>
    %expanded = tensor.expand_shape %padded [[0], [1, 2], [3, 4]] output_shape [1, 4, 16, 4, 16] : tensor<1x64x64xf32> into tensor<1x4x16x4x16xf32>
    scf.forall // thread loop
      scf.for{ //k loop
      }
      %extracted_slice_2 = tensor.extract_slice %expanded[0, %22, %18, %21, %17#2] [1, 2, 4, 2, 1] [1, 1, 1, 1, 1] : tensor<1x4x16x4x16xf32> to tensor<1x2x4x2x1xf32>
```

If we analyze the access pattern of the extract_slice at the end of the snippet with respect to the ultimate source tensor, it is taking a slice that will be strided in memory. In other words, the `%expanded = tensor.expand_shape` is required to represent the correct slice layout. Because of this, the expand_shape blocks fusions from happening during codegen, and we will get extra allocations when we bufferize. The above snippet is from a debug dump in this issue: https://github.com/iree-org/iree/issues/22919#issuecomment-3687528026

The map_gather op could be used to combine the pad and the expand_shape into a single fusable operation, which will then get fused into the thread loop with the bias add, avoiding the need for an extra allocation.

# MapGatherOp Definition #

[...]

PR #23041, opened 2026-01-06
[LinalgExt] Add map_gather op and verifier

nirvedhmeshram:
Adds a new op called `iree_linalg_ext.map_gather`, which represents a gathering operation from a source tensor to an output tensor by mapping from output indices to source indices. The index mapping is represented by a region containing index computation operations, yielding a set of source indices and a padding value.

The new op is the inverse of map_scatter and useful for representing long chains of index transformations in a single operation, particularly for handling complex sequences of layout transformations upon reading from memory, such as propagating packed intrinsic layouts to elementwise consumers.

This PR adds:
- Op definition
- Verifier for the new op
- Stub implementations for tiling interface methods
- Test cases in invalid.mlir (6 negative tests)
- Test cases in roundtrip.mlir (14 positive tests)

Related to issue #23038


compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/roundtrip.mlir
```diff
@@ -2254,3 +2254,165 @@ func.func @split_reduction_mapping(%arg0 : index,
...
+// CHECK: %[[RES:.+]] = iree_linalg_ext.map_gather %[[SOURCE]] into %[[OUTPUT]] {
+// CHECK: ^bb0(%[[IDX0:.+]]: index, %[[IDX1:.+]]: index):
+// CHECK: %[[PAD:.+]] = arith.constant 0.{{0+}}e+00 : f32
+// CHECK: iree_linalg_ext.yield %[[IDX0]], %[[IDX1]], %[[PAD]]
+// CHECK: } : tensor<?x?xf32> into tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK: return %[[RES]] : tensor<?x?xf32>
+
+// -----
+
+func.func @map_gather_result_sliced(
```
Max191:
```suggestion
func.func @map_gather_result_padded(
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/roundtrip.mlir
```diff
@@ -2254,3 +2254,165 @@ func.func @split_reduction_mapping(%arg0 : index,
...
+    %source: tensor<63xf32>, %output: tensor<64xf32>
+) -> tensor<64xf32> {
+  %0 = iree_linalg_ext.map_gather %source into %output {
+  ^bb0(%idx0: index):
+    %pad = arith.constant 0.0 : f32
+    iree_linalg_ext.yield %idx0, %pad : index, f32
+  } : tensor<63xf32> into tensor<64xf32> -> tensor<64xf32>
+  return %0 : tensor<64xf32>
+}
+// CHECK-LABEL: func.func @map_gather_result_sliced(
```
Max191:
```suggestion
// CHECK-LABEL: func.func @map_gather_result_padded(
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/roundtrip.mlir
```diff
@@ -2254,3 +2254,165 @@ func.func @split_reduction_mapping(%arg0 : index,
...
+  } : tensor<4x16x64xf32> into tensor<4x16x64xf32> -> tensor<4x16x64xf32>
+  return %0 : tensor<4x16x64xf32>
+}
+// CHECK-LABEL: func.func @map_gather_tensor_static(
+// CHECK-SAME: %[[SOURCE:[a-zA-Z0-9_]+]]
+// CHECK-SAME: %[[OUTPUT:[a-zA-Z0-9_]+]]
+// CHECK: %[[RES:.+]] = iree_linalg_ext.map_gather %[[SOURCE]] into %[[OUTPUT]] {
+// CHECK: ^bb0(%[[IDX0:.+]]: index, %[[IDX1:.+]]: index, %[[IDX2:.+]]: index):
+// CHECK: %[[PAD:.+]] = arith.constant 0.{{0+}}e+00 : f32
+// CHECK: iree_linalg_ext.yield %[[IDX0]], %[[IDX1]], %[[IDX2]], %[[PAD]]
```
Max191: nit: Add indentation to the block. Same for the other tests.

```suggestion
// CHECK:   %[[PAD:.+]] = arith.constant 0.{{0+}}e+00 : f32
// CHECK:   iree_linalg_ext.yield %[[IDX0]], %[[IDX1]], %[[IDX2]], %[[PAD]]
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/roundtrip.mlir
```diff
@@ -2254,3 +2254,165 @@ func.func @split_reduction_mapping(%arg0 : index,
...
+  ^bb0(%idx0: index):
+    %c63 = arith.constant 63 : index
+    %c0 = arith.constant 0 : index
+    %in_bounds = arith.cmpi ult, %idx0, %c63 : index
+    %clamped = arith.select %in_bounds, %idx0, %c0 : index
+    %pad = arith.constant 0.0 : f32
+    iree_linalg_ext.yield %clamped, %pad : index, f32
+  } : tensor<64xf32> into tensor<63xf32> -> tensor<63xf32>
+  return %0 : tensor<63xf32>
+}
```
Max191: This test doesn't make much sense with map_gather. There is no reason to clamp the index, since the iteration space of the result is always in bounds on the source. Good to have a slicing example, but just remove the clamping part.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -710,6 +711,102 @@ void MapScatterOp::insertTransformationAtStart(
...
+  Type paddingType = yieldOp.getOperandTypes()[getSourceRank()];
+  Type elementType = getSourceType().getElementType();
+  if (paddingType != elementType) {
+    return yieldOp.emitOpError("expected yielded padding value type to match "
+                               "source element type");
+  }
+  return success();
+}
+
+Value MapGatherOp::getSourceIndex(int64_t position) {
```
Max191: Save all the utility functions (including all the ones below) for the PRs where they are eventually used.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -2856,6 +2954,144 @@ DEFINE_OP_GET_EFFECTS(ExpReductionOp)
...
+  TypedAttr zeroAttr = cast<TypedAttr>(builder.getZeroAttr(elemType));
+  Value paddingValue = builder.create<arith::ConstantOp>(loc, zeroAttr);
+  yieldOperands.push_back(paddingValue);
+
+  builder.create<IREE::LinalgExt::YieldOp>(loc, yieldOperands);
+
+  return op;
+}
+
+LogicalResult MapGatherOp::getResultTilePosition(
```
Max191: All the tiling interface implementations belong in `TilingInterfaceImpl.cpp`. Also, let's save these for the PR that implements the full tiling interface.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -382,6 +382,119 @@ def IREELinalgExt_MapScatterOp : IREELinalgExt_Op<"map_scatter",
...
+    for each dimension in the `output` value. The block arguments represent the
+    index of a given element along the corresponding dimension (i.e., block arg
+    `i` represents the index along dimension `i` of the `output` value). The
+    block is terminated by an iree_linalg_ext.yield op, which yields one index
+    for each dimension in the `source`, and an additional padding value which
+    represents the value to use if the computed source indices are out of bounds.
+
+    This operation has padding semantics (as opposed to map_scatter's masking
+    semantics), and has gather semantics (fuses with consumers, as opposed to
+    map_scatter's scatter semantics which fuses with producers).
```
Max191: We can remove this part from the description.

Max191 (commented):
Cool! Mainly left comments about splitting up some of this for later PRs.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -587,6 +587,47 @@ void GatherOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
 
+//===----------------------------------------------------------------------===//
+// MapGatherOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult MapGatherOp::verify() {
+  if (getSourceType().getElementType() != getOutputType().getElementType()) {
+    return emitOpError("expected source and output element types to match");
+  }
+  if (getOutputType().getRank() == 0) {
```
krzysz00: How much does the trivial map_gather hurt us if we were to allow it? (I'd think it can be defined and then is a really weird way to spell extracting an element from a tensor)

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -272,6 +272,58 @@ def IREELinalgExt_GatherOp : IREELinalgExt_Op<"gather",
...
+    }
+
+    // Method to implement for specifying output range for
+    // DestinationStyleOpInterface
+    MutableOperandRange getDpsInitsMutable() {
+      return getOutputMutable();
+    }
+  }];
+}
+
```
krzysz00, line 326: Thinking from an optimization point of view, I'd like to propose that these map_gather ops can be annotated with an `inbounds` tag indicating that no padding will be performed so that we don't have to prove it later when we might not be able to.

We could even take a page from vector.transfer_read and make than annotation per-dimension, though then I'd call it something like `known_inbounds`.

(My micro-optimizing heart also wants `nneg`/`pad_right_only` for guarantees that we can just do one-sided comparisons for padding, but that's not actually as important as an `inbounds`)

krzysz00 (commented):
Some proactive thoughts

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
nirvedhmeshram: hmm, good point, but we didnt end up needing this for map_scatter, which parallels  transfer_write, so why should we need it for map_gather which parallels transfer_read? @Max191 do you think this would be useful?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
nirvedhmeshram: I have no strong opinion, however seems like allowing IR that we should never generate so why not disallow it instead.

Max191: I don't think we need to go as far as disallowing it, since it is still a valid gathering operation, but I can't think of a good reason for one way or the other, so I'll leave it up to you what to do here.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
Max191: I am not opposed to adding it. Albeit likely small, there could be a performance impact from checking padding bounds on every load.

That said, it's still the type of thing we can always add on later, and it would not be relevant to our current use case of map_gather (since we will always have actual padding), so I'm happy with adding it now or waiting until we need it later.

Max191 (approved):
This LGTM. I responded to the 2 comments from Krzystof, but I'll leave it up to you what decisions to make. I think all the options are fine.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
nirvedhmeshram: I would prefer to keep the op plumbing simple for now so I will skip this, good optimization to keep in mind though.

krzysz00: `map_scatter` should also grow an `inbounds`/`unmasked` as well for similar reasons, IMO, especially since there is a real overhead to buffer loads' bounds checking (per, for example, Triton) and we should avoid that overhead where possible.

PR #23050, opened 2026-01-07
[LinalgExt] Implement tiling interface for map_gather

Abhishek-Varma:
-- This PR implements the TilingInterface for `iree_linalg_ext.map_gather`.
-- `map_gather`'s iteration space is defined on the output/destination tensor i.e. 
    each output position gathers a value from a computed source location.
-- So the output operand gets tiled while the source remains full (to allow arbitrary index access).
-- The implementation includes basic tiling (getTiledImplementation, getIterationDomain, getLoopIteratorTypes, getResultTilePosition), producer fusion (generateResultTileValue), and consumer fusion from the output operand (getIterationDomainTileFromOperandTiles, getTiledImplementationFromOperandTiles).


compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -454,6 +454,167 @@ LogicalResult GatherOp::generateScalarImplementation(OpBuilder &b, Location loc,
...
+          memref::LoadOp::create(ifBuilder, ifLoc, getSource(), loadIndices);
+      memref::StoreOp::create(ifBuilder, ifLoc, loaded, getOutput(), ivs);
+      scf::YieldOp::create(ifBuilder, ifLoc);
+    };
+    auto elseBuilder = [&](OpBuilder &ifBuilder, Location ifLoc) {
+      memref::StoreOp::create(ifBuilder, ifLoc, paddingValue, getOutput(), ivs);
+      scf::YieldOp::create(ifBuilder, ifLoc);
+    };
+    scf::IfOp::create(nestedBuilder, nestedLoc, inBounds, thenBuilder,
+                      elseBuilder);
```
Max191: As per Nirvedh's suggestion on Discord, could we use an arith::SelectOp instead of the scf.if? It should be easier to handle without the regions that come with the if.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/convert_to_loops.mlir
```diff
@@ -1703,3 +1703,47 @@ func.func @map_scatter_memref(
...
+//      CHECK:     %[[SRC_IDX:.+]]:2 = affine.delinearize_index %[[IV]] into (%[[SRC_D0]], %[[SRC_D1]]) : index, index
+//      CHECK:     %[[BOUND_D0:.+]] = memref.dim %[[SOURCE]], %[[C0]]
+//      CHECK:     %[[GE_ZERO_0:.+]] = arith.cmpi sge, %[[SRC_IDX]]#0, %[[C0]] : index
+//      CHECK:     %[[LT_DIM_0:.+]] = arith.cmpi slt, %[[SRC_IDX]]#0, %[[BOUND_D0]] : index
+//      CHECK:     %[[IN_BOUNDS_0:.+]] = arith.andi %[[GE_ZERO_0]], %[[LT_DIM_0]]
+//      CHECK:     %[[BOUND_D1:.+]] = memref.dim %[[SOURCE]], %[[C1]]
+//      CHECK:     %[[GE_ZERO_1:.+]] = arith.cmpi sge, %[[SRC_IDX]]#1, %[[C0]] : index
+//      CHECK:     %[[LT_DIM_1:.+]] = arith.cmpi slt, %[[SRC_IDX]]#1, %[[BOUND_D1]] : index
+//      CHECK:     %[[IN_BOUNDS_1:.+]] = arith.andi %[[GE_ZERO_1]], %[[LT_DIM_1]]
+//      CHECK:     %[[IN_BOUNDS:.+]] = arith.andi %[[IN_BOUNDS_0]], %[[IN_BOUNDS_1]]
```
Max191: nit: Better to make all of the arith and dim ops `CHECK-DAG` instead, since the ordering doesn't matter for this test (i.e., dim 0 could be checked first or dim 1 could be checked first and both are correct). It will make it less sensitive to small code changes.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -454,6 +454,167 @@ LogicalResult GatherOp::generateScalarImplementation(OpBuilder &b, Location loc,
   return success();
```
Max191, line 454: Can you also add the implementation for fusion from a result tile? This is actually the most common fusion case we will see. You need to implement the `generateResultTileValue` function, and you should be able to just do the same as what GatherOp does:

https://github.com/iree-org/iree/blob/769716f67e5dd75578759c0702601d0ffc035316/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp#L411-L416

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -316,6 +324,37 @@ def IREELinalgExt_MapGatherOp : IREELinalgExt_Op<"map_gather"> {
...
     }
 
+    // Return the source index at the provided position.
+    Value getSourceIndex(int64_t position);
+
+    // Return the output index at the provided position.
+    Value getOutputIndex(int64_t position);
+
+    // Returns the padding value.
+    Value getPaddingValue();
```
Max191: Since these aren't used anywhere, I think it is better to delete them in this PR. We can add them in later if they are used by a future transformation.

Max191 (commented):
Looks great so far! Just a few comments.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
nirvedhmeshram: wont doing this in all cases risk an OOB access currently we emit something like this
```
      scf.if %7 {
        %8 = memref.load %arg0[%0#0, %0#1] : memref<?x?xf32>
        memref.store %8, %arg1[%arg2] : memref<?xf32>
      } else {
        memref.store %cst, %arg1[%arg2] : memref<?xf32>
      }
  ```
  but for the select we will need to unconditionally do the %8 load which maybe OOB? I was thinking that will need some special logic to check for address spaces where out of bounds is ok, so this can be added later unless I am missing something where the select is always safe?

File: compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp [hunk 5]
```diff
@@ -625,6 +625,90 @@ LogicalResult MapGatherOp::verify() {
...
+  SmallVector<Type> indexTypes(numOutputIndices, builder.getIndexType());
+  SmallVector<Location> locs(numOutputIndices, getLoc());
+
+  // Create the new block arguments for the new output indices, and transform
+  // them using the callback.
+  SmallVector<BlockArgument> newOutputIndices(
+      transformBody.addArguments(indexTypes, locs));
+  OpBuilder::InsertionGuard g(builder);
+  builder.setInsertionPointToStart(&transformBody);
+  SmallVector<Value> newOutputIndicesTransformed(
```
krzysz00, line 644: Shouldn't this also take an `OpBuilder`

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -625,6 +625,90 @@ LogicalResult MapGatherOp::verify() {
...
+      transformationBuilder(newOutputIndices));
+
+  // Replace the old output indices with the results of the transformation on
+  // the new output indices.
+  assert(oldOutputIndices.size() == newOutputIndicesTransformed.size() &&
+         "expected transformation to produce the same number of Values as the "
+         "previous number of output indices.");
+  for (auto [oldIdx, newIdx] :
+       llvm::zip_equal(oldOutputIndices, newOutputIndicesTransformed)) {
+    SmallVector<OpOperand *> uses(llvm::make_pointer_range(oldIdx.getUses()));
```
krzysz00: 1. Isn't this `replaceAllUsesWith`?
2. If you'll be iterating over a use list, you'll want an `early_inc_range`

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -454,6 +454,167 @@ LogicalResult GatherOp::generateScalarImplementation(OpBuilder &b, Location loc,
...
+      Value geZero = arith::CmpIOp::create(
+          nestedBuilder, nestedLoc, arith::CmpIPredicate::sge, idx, zero);
+      // Check: idx < dimSize
+      Value ltDim = arith::CmpIOp::create(
+          nestedBuilder, nestedLoc, arith::CmpIPredicate::slt, idx, dimSize);
+      // Combine: idx >= 0 && idx < dimSize
+      Value dimInBounds =
+          arith::AndIOp::create(nestedBuilder, nestedLoc, geZero, ltDim);
+
+      if (inBounds) {
```
krzysz00: To support the 0-D case and maybe clean this up a bit, start `inBounds` as a constant `true` and rely on the folders of `andi`.

krzysz00: Concurring that this shouldn't be a load-and-select since that can segfault. If we want to transparently get existing optimizations, it could be a vector.transfer_read with a mask?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -454,6 +454,167 @@ LogicalResult GatherOp::generateScalarImplementation(OpBuilder &b, Location loc,
...
+
+    // Create if-else: if in bounds, load from source; else use padding.
+    auto thenBuilder = [&](OpBuilder &ifBuilder, Location ifLoc) {
+      Value loaded =
+          memref::LoadOp::create(ifBuilder, ifLoc, getSource(), loadIndices);
+      memref::StoreOp::create(ifBuilder, ifLoc, loaded, getOutput(), ivs);
+      scf::YieldOp::create(ifBuilder, ifLoc);
+    };
+    auto elseBuilder = [&](OpBuilder &ifBuilder, Location ifLoc) {
+      memref::StoreOp::create(ifBuilder, ifLoc, paddingValue, getOutput(), ivs);
```
krzysz00: The store can be moved out of the if though.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
hanhanW: yep, there will be OOB access. I hit such issue before, like 4-5 years ago..

hanhanW (commented):
Please split this PR into two PRs. One for tiling implementation, and the other for `generateScalarImplementation` implementation. They are independent.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
Abhishek-Varma: Can you please elaborate more on this?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
Abhishek-Varma: Done.

Abhishek-Varma (commented):
Hi @Max191 @nirvedhmeshram @hanhanW @krzysz00 - I'm only keeping tiling related implementations in these PRs. The review comments pertaining to scalar implementation have been addressed too but I'll raise a separate PR for the same.

There was a bug in the tiling implementation which surfaced after the implementation of `generateResultTileValue` API.
```
BEFORE:

Operation *tiledOp = mlir::clone(builder, getOperation(), getResultTypes(),
                                   {getSource(), outputSlice->getResult(0)});
                                   
                                   
AFTER:
Value tiledOutput = outputSlice->getResult(0);
  SmallVector<Type> resultTypes;
  if (getNumResults()) {
    resultTypes.push_back(tiledOutput.getType());
  }
  Operation *tiledOp = mlir::clone(builder, getOperation(), resultTypes,
                                   {getSource(), tiledOutput});
```

Because of the state mentioned in `BEFORE` the tiled implementation was not accounting for the output slice and was incorrectly using the non-tiled output version. I have added a lit test for the same.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
Max191: I see, I forgot about this. Okay, let's stick to the scf.if for now.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
krzysz00: Something smelled funny about a function with that signature not having any `OpBuilder` or any other way to construct those `Value`s it's returning.

Abhishek-Varma: Oh okay. So it's mostly adapted from [MapScatter::insertTransformationAtStart's code](https://github.com/iree-org/iree/blob/56454c2892af920903c77a2a9478cb645f3beda5/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp#L722-L739). Let me know if something warrants a change and I'd do so. :)

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
Abhishek-Varma: We can take the scalar implementation's discussion to this newly raised PR : https://github.com/iree-org/iree/pull/23070 (I've tried addressing the comments in that PR). Can you take a look now?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
Abhishek-Varma: Done as part of : https://github.com/iree-org/iree/pull/23070

hanhanW (approved):
LGTM, thanks

hanhanW:
I think the PR desecription can be improved. See https://google.github.io/eng-practices/review/developer/cl-descriptions.html for guidance.

E.g., it is very helpful if we mention that the iteration space is defined on output/destination.

PR #23052, opened 2026-01-07
[Codegen] Implement workgroup execution scope

qedawkins:
This adds #iree_codegen.workgroup_scope as the scope for workgroup execution implementing the PCF dialects ScopeAttrInterface.

The interface is implemented as an external interface to avoid taking a HAL dep in the Codegen dialect (which currently results in a circular dep somehow?!?).


compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
```diff
@@ -872,4 +872,30 @@ void setUKernelDescriptor(Operation *op,
...
+// iree_codegen.workgroup_scope
+//===----------------------------------------------------------------------===//
+
+// Custom parser/printer to make the <> optional. It will either parse/print as
+//   #iree_codegen.workgroup_scope<linearize>
+// or without the <>.
+//   #iree_codegen.workgroup_scope
+Attribute IREE::Codegen::WorkgroupScopeAttr::parse(AsmParser &parser, Type) {
+  bool linearize = false;
+  if (parser.parseOptionalLess().succeeded()) {
```
kuhar: nit: use success(...)

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
```diff
@@ -664,4 +664,24 @@ def DenormalFpMathAttr :
...
 }
 
+//===---------------------------------------------------------------------===//
+// iree_codegen.workgroup_scope
+//===---------------------------------------------------------------------===//
+
+def WorkgroupScopeAttr : AttrDef<IREECodegen_Dialect, "WorkgroupScope", []> {
+  let mnemonic = "workgroup_scope";
+  let summary =
+      [{Attribute representing parallel execution across workgroups.}];
```
kuhar:
```suggestion
  let summary =
      "Attribute representing parallel execution across workgroups.";
```

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CodegenExternalModels.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    // If linearize is true, compute the product of all counts and replace the
+    // first ID with that.
+    if (linearize && !counts.empty()) {
+      Value one = counts.size() > 1
+                      ? arith::ConstantIndexOp::create(builder, loc, 1)
+                      : Value();
+      Value linearizedCount = counts[0];
+      for (int64_t i = 1, e = counts.size(); i < e; ++i) {
```
kuhar: Use `for (int64_t &count : llvm::drop_begin(counts))`?

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CodegenExternalModels.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    // Pad the outer most sizes with 1.
+    if (numIds > 3) {
+      Value one = arith::ConstantIndexOp::create(builder, loc, 1);
+      counts.append(numIds - 3, one);
+    }
+
+    return counts;
+  }
+  SmallVector<Value> getWorkerIDs(Attribute attr, OpBuilder &builder,
```
kuhar, line 67:
```suggestion
  }
  
  SmallVector<Value> getWorkerIDs(Attribute attr, OpBuilder &builder,
```

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CodegenExternalModels.cpp
```diff
@@ -0,0 +1,142 @@
...
+      // Linearize. IDs need to be reversed to match the counts and the builder
+      // can't take iterators.
+      SmallVector<Value> reverseIds(llvm::reverse(ids));
+      auto linearizeOp = affine::AffineLinearizeIndexOp::create(
+          builder, loc, reverseIds, dynamicCounts);
+
+      ids.front() = linearizeOp.getResult();
+      for (int64_t i = 1, e = ids.size(); i < e; ++i) {
+        ids[i] = one;
+      }
```
kuhar:
```suggestion
      llvm::fill(llvm::drop_begin(ids), one);
```
I think this should work

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CodegenExternalModels.h
```diff
@@ -0,0 +1,20 @@
+// Copyright 2025 The IREE Authors
```
kuhar:
```suggestion
// Copyright 2026 The IREE Authors
```

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CodegenExternalModels.cpp
```diff
@@ -0,0 +1,142 @@
+// Copyright 2025 The IREE Authors
```
kuhar:
```suggestion
// Copyright 2026 The IREE Authors
```

kuhar (approved):

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
qedawkins: We were supposed to be using `[{}]` everywhere even for summaries but it looks like this file had some slips. I'll make the indentation a bit nicer.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/lower_structural_pcf.mlir
```diff
@@ -112,3 +112,47 @@ util.func private @lower_loop(%d0: index, %d1: index) {
 //  CHECK-SAME:   %[[ARG1:[A-Za-z0-9_]+]]: index
 //       CHECK:   scf.forall (%[[ID1:.+]], %[[ID0:.+]]) in (%[[ARG1]], %[[ARG0]])
 //  CHECK-NEXT:     util.optimization_barrier %[[ID0]], %[[ID1]]
+
```
Max191, line 115: Can you add a test for the non-linearize case, but with > 3 IDs?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/lower_structural_pcf.mlir
```diff
@@ -112,3 +112,47 @@ util.func private @lower_loop(%d0: index, %d1: index) {
...
+//  CHECK-SAME:   %[[ARG0:[A-Za-z0-9_]+]]: index
+//  CHECK-SAME:   %[[ARG1:[A-Za-z0-9_]+]]: index
+//  CHECK-SAME:   %[[ARG2:[A-Za-z0-9_]+]]: index
+//  CHECK-SAME:   %[[ARG3:[A-Za-z0-9_]+]]: index
+//       CHECK:   %[[MUL0:.+]] = arith.muli %workgroup_count_x, %workgroup_count_y : index
+//       CHECK:   %[[MUL1:.+]] = arith.muli %[[MUL0]], %workgroup_count_z : index
+//       CHECK:   %[[LINEARIZE:.+]] = affine.linearize_index [%c0, %workgroup_id_z, %workgroup_id_y, %workgroup_id_x]
+//  CHECK-SAME:     by (1, %workgroup_count_z, %workgroup_count_y, %workgroup_count_x)
+//       CHECK:   scf.forall ({{.*}}) = (%c1, %c1, %c1, %[[LINEARIZE]])
+//  CHECK-SAME:     to (%[[ARG3]], %[[ARG2]], %[[ARG1]], %[[ARG0]]) step (%c1, %c1, %c1, %[[MUL1]])
```
Max191, line 187: What happens for cases where the product of workgroup counts is greater than the inner loop dimension? Based on the test it seems like many workgroups could be doing no work. I'd expect this to generate an scf.forall with a single linearized induction variable, and then the delinearize would be inside the scf.forall to remap the linear ID.

Max191 (commented):
My main concern is about how the linearization of the workgroup IDs is done. This is pretty different from what we have done before IIRC, and it seems like it could generate inefficient use of workgroups.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/lower_structural_pcf.mlir
qedawkins: Not sure I understand the question. The product of workgroup counts is determined by the value we request when forming the loop, so if there were more workgroups than necessary either something went wrong or there was another loop requesting more.

Once we reach this pass, the lowering is extremely mechanical. Take the `hal.interface.workgroup.count/id` ops and lower the loops based on them.

qedawkins: > I'd expect this to generate an scf.forall with a single linearized induction variable, and then the delinearize would be inside the scf.forall to remap the linear ID.

Ok I understand your question. This is the responsibility of earlier passes, once we get here we don't do anything fancy.

Max191: Ok, I think I see what you are doing when I look at https://github.com/iree-org/iree/pull/23055. You are creating PCF loops with a single ID, which should lower into the right thing.

So the semantics of the `linearize` flag are really just to say that all workgroup counts should be used exclusively for the innermost loop dimension.

Max191: You can ignore my comment, then. This looks okay to me.

Max191 (approved):
This LGTM now.

PR #23056, opened 2026-01-07
[AMDGPU][LDS] Support linearized DMA for small innermost dimensions

lialan:
When the linalg.copy output comes from tensor.empty(), the  `AMDGPULowerCoalescedDMAToGatherLDS` pass can linearize the destination memref. 

This patch allows coalesced DMA for tensors with small innermost dimensions (e.g., `128x16xf32` where `16 < minElementsPerTransfer=64`) by linearizing the tensor and using total elements instead of innermost dimension for the size check.

This change gates the optimization: The change in GPUConvertToCoalescedDMA simply allows copies with small innermost dimensions to proceed when linearization is safe, instead of rejecting them outright.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -58,6 +58,45 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
   return mapping;
 }
 
+/// Check if a value traces back to tensor.empty (possibly through forall args).
+static bool tracesToTensorEmpty(Value value) {
+  // Direct tensor.empty.
+  if (value.getDefiningOp<tensor::EmptyOp>()) {
+    return true;
+  }
+
+  // Check if value is an extract_slice from a forall block argument.
+  auto extractSlice = value.getDefiningOp<tensor::ExtractSliceOp>();
+  if (!extractSlice) {
+    return false;
+  }
+
+  auto blockArg = dyn_cast<BlockArgument>(extractSlice.getSource());
+  if (!blockArg) {
+    return false;
+  }
+
+  auto forallOp = dyn_cast<scf::ForallOp>(blockArg.getOwner()->getParentOp());
+  if (!forallOp) {
+    return false;
+  }
+
+  // Find the corresponding shared_out init value.
+  unsigned numIVs = forallOp.getInductionVars().size();
+  unsigned argIndex = blockArg.getArgNumber();
+  if (argIndex < numIVs) {
+    return false;
+  }
+
+  unsigned sharedOutIndex = argIndex - numIVs;
+  if (sharedOutIndex >= forallOp.getOutputs().size()) {
+    return false;
+  }
+
+  Value initValue = forallOp.getOutputs()[sharedOutIndex];
+  return initValue.getDefiningOp<tensor::EmptyOp>() != nullptr;
+}
+
 /// Helper to compute thread number of threads based on translation_info.
 /// Uses the subgroup_size from translation_info for thread-level tiling.
 static SmallVector<OpFoldResult>
@@ -117,9 +156,21 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
         std::min(minElementsPerTransfer, elementsPerTransfer);
   }
 
-  // If no valid DMA size found or innermost dim is too small, skip.
+  // Determine how many elements are available for coalesced access.
+  // For CopyOp with output tracing to tensor.empty(), we can linearize.
+  ArrayRef<int64_t> shape = outputType.getShape();
+  int64_t availableElements = innermostDim;
+  if (auto copyOp = dyn_cast<linalg::CopyOp>(op)) {
+    Value output = copyOp.getOutputs()[0];
+    if (tracesToTensorEmpty(output) &&
+        llvm::none_of(shape, ShapedType::isDynamic)) {
+      availableElements = ShapedType::getNumElements(shape);
+    }
+  }
+
+  // If no valid DMA size found or available elements are too small, skip.
   if (minElementsPerTransfer == std::numeric_limits<int64_t>::max() ||
-      innermostDim < minElementsPerTransfer) {
+      availableElements < minElementsPerTransfer) {
     return {};
   }
 
@@ -583,9 +634,11 @@ private:
     int64_t totalWarps = llvm::product_of(positiveWarps);
 
     // Greedily distribute warps to outer dimensions, keeping innermost whole.
+    // For 1D tensors, distribute across the single dimension (no inner/outer).
     int64_t remainingWarps = totalWarps;
     for (int64_t i = 0; i < rank; ++i) {
-      if (i == rank - 1) {
+      bool isInnermostOfMultiDim = (i == rank - 1) && (rank > 1);
+      if (isInnermostOfMultiDim) {
         // Keep innermost dimension whole (tile size = full dimension).
         tileSizes.push_back(rewriter.getIndexAttr(shape[i]));
         ++numTiledDims;
@@ -684,9 +737,23 @@ private:
           std::min(minElementsPerTransfer, elementsPerTransfer);
     }
 
-    // If no valid DMA size found or innermost dim is too small, skip.
+    // Determine how many elements are available for coalesced access.
+    // For CopyOp with tensor.empty() output, we can linearize all dimensions.
+    // Otherwise, we can only use the innermost dimension.
+    int64_t availableElements = innermostDim;
+    if (auto copyOp = dyn_cast<linalg::CopyOp>(op.getOperation())) {
+      Value output = copyOp.getOutputs()[0];
+      if (output.getDefiningOp<tensor::EmptyOp>()) {
+        // Can linearize all dimensions - compute total static elements.
+        if (llvm::none_of(shape, ShapedType::isDynamic)) {
+          availableElements = ShapedType::getNumElements(shape);
+        }
+      }
+    }
+
+    // If no valid DMA size found or available elements are too small, skip.
     if (minElementsPerTransfer == std::numeric_limits<int64_t>::max() ||
-        innermostDim < minElementsPerTransfer) {
+        availableElements < minElementsPerTransfer) {
       return failure();
     }
 
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
@@ -200,3 +200,218 @@ func.func @copy_prefer_contiguous_subview(%source: tensor<64x128xf32>, %init: te
 
   return %result : tensor<64x128xf32>
 }
+
+// -----
+
+// Test: Small innermost dimension with tensor.empty() output CAN be linearized.
+// When output comes from tensor.empty(), we can use total elements instead of
+// innermost dimension for the size check, enabling coalesced DMA.
+
+#gpu_target_linearize = #iree_gpu.target<arch = "gfx942", features = "", wgp = <
+  compute = fp32, storage = b32, subgroup = shuffle,
+  max_load_instruction_bits = 128, subgroup_size_choices = [64],
+  max_workgroup_sizes = [1024, 1024, 1024], max_thread_count_per_workgroup = 1024,
+  max_workgroup_memory_bytes = 65536, max_workgroup_counts = [2147483647, 2147483647, 2147483647],
+  dma_sizes = [32]
+>>
+
+#exec_target_linearize = #hal.executable.target<"rocm", "rocm-hsaco-fb", {iree_codegen.target_info = #gpu_target_linearize}>
+#translation_linearize = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse workgroup_size = [256, 1, 1] subgroup_size = 64>
+
+// CHECK-LABEL: func.func @copy_small_innermost_linearized
+// CHECK-SAME:    %[[SRC:[a-zA-Z0-9]+]]: tensor<128x16xf32>
+func.func @copy_small_innermost_linearized(%source: tensor<128x16xf32>) -> tensor<128x16xf32>
+  attributes {hal.executable.target = #exec_target_linearize, translation_info = #translation_linearize} {
+  %empty = tensor.empty() : tensor<128x16xf32>
+  %result = linalg.copy {lowering_config = #iree_gpu.use_global_load_dma}
+    ins(%source : tensor<128x16xf32>)
+    outs(%empty : tensor<128x16xf32>) -> tensor<128x16xf32>
+
+  // Innermost dimension (16) < minElementsPerTransfer (64), but since output is
+  // tensor.empty(), we use total elements (2048) for the check, which passes.
+  // With 4 warps (256/64), rows are tiled to 32 (128/4), columns kept whole at 16.
+
+  // CHECK: %[[EMPTY:.+]] = tensor.empty() : tensor<128x16xf32>
+
+  // Warp-level forall: step (32, 16) distributes 128 rows across 4 warps
+  // CHECK: %[[WARP_RESULT:.+]] = scf.forall (%[[IV0:.+]], %[[IV1:.+]]) = (0, 0) to (128, 16) step (32, 16)
+  // CHECK-SAME: shared_outs(%[[INIT_TILE:.+]] = %[[EMPTY]]) -> (tensor<128x16xf32>) {
+  // CHECK:   %[[SLICE_SRC:.+]] = tensor.extract_slice %[[SRC]][%[[IV0]], 0] [32, 16] [1, 1]
+  // CHECK-SAME:   : tensor<128x16xf32> to tensor<32x16xf32>
+  // CHECK:   %[[SLICE_DST:.+]] = tensor.extract_slice %[[INIT_TILE]][%[[IV0]], 0] [32, 16] [1, 1]
+  // CHECK-SAME:   : tensor<128x16xf32> to tensor<32x16xf32>
+
+  // Thread-level forall with 64 lanes
+  // CHECK:   %[[THREAD_RESULT:.+]] = scf.forall (%[[LANE:.+]]) in (64)
+  // CHECK-SAME:   shared_outs(%[[THREAD_INIT:.+]] = %[[SLICE_DST]]) -> (tensor<32x16xf32>) {
+  // CHECK:     scf.forall.in_parallel {
+  // CHECK:       iree_gpu.coalesced_gather_dma %[[SLICE_SRC]] into %[[THREAD_INIT]] lane(%[[LANE]])
+  // CHECK-SAME:       : tensor<32x16xf32>, tensor<32x16xf32>, index
+  // CHECK:     }
+  // CHECK:   } {mapping = [#iree_gpu.lane_id<0>]}
+
+  // CHECK:   scf.forall.in_parallel {
+  // CHECK:     tensor.parallel_insert_slice %[[THREAD_RESULT]] into %[[INIT_TILE]][%[[IV0]], 0] [32, 16] [1, 1]
+  // CHECK-SAME:     : tensor<32x16xf32> into tensor<128x16xf32>
+  // CHECK:   }
+  // CHECK: }
+
+  // CHECK: return %[[WARP_RESULT]]
+  // CHECK-NOT: linalg.copy
+
+  return %result : tensor<128x16xf32>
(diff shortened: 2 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,11 +58,39 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
   return mapping;
 }
 
+/// Compute total static elements in the shape. Returns 0 if any dimension
+/// is dynamic.
+static int64_t getTotalStaticElements(ArrayRef<int64_t> shape) {
+  int64_t total = 1;
+  for (int64_t dim : shape) {
+    if (ShapedType::isDynamic(dim))
```
qedawkins: nit: braces

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -116,15 +142,63 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
   }
 
-  // If no valid DMA size found or innermost dim is too small, skip.
-  if (minElementsPerTransfer == std::numeric_limits<int64_t>::max() ||
-      innermostDim < minElementsPerTransfer) {
+  if (minElementsPerTransfer == std::numeric_limits<int64_t>::max()) {
+    return {};
+  }
+
+  if (!hasEnoughElementsForDMA(shape, minElementsPerTransfer, canLinearize)) {
```
qedawkins: You can combine these two checks because they both return the same thing.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -116,15 +142,63 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
   }
 
   return {builder.getIndexAttr(*subgroupSize)};
 }
 
+/// Check if an extract_slice preserves contiguity for linearization.
+/// The slice must have unit strides. For rank > 1, the innermost dimension
+/// must be fully taken to ensure contiguous memory layout.
+static bool isContiguousSlice(tensor::ExtractSliceOp extractSlice) {
+  if (!extractSlice.hasUnitStride())
```
qedawkins: nit: braces here and below

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -116,15 +142,63 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
+    return false;
+
+  auto sourceType = extractSlice.getSourceType();
```
qedawkins:
```suggestion
  RankedTensorType sourceType = extractSlice.getSourceType();
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -116,15 +142,63 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
+
+/// Check if the copy's output can be linearized for DMA.
+/// Returns true if output is:
+/// - Direct tensor.empty (before subgroup tiling), or
+/// - Contiguous extract_slice (after subgroup tiling from tensor.empty)
+static bool canLinearizeCopy(linalg::CopyOp copyOp) {
+  Value output = copyOp.getOutputs()[0];
+
+  // Direct tensor.empty (before subgroup tiling).
+  if (output.getDefiningOp<tensor::EmptyOp>())
```
qedawkins: braces here and below too

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
kuhar: FYI, this is now automatically enforced through clang-format (and pre-commit in CI)

kuhar: github should pick it up as soon as you rebase @lialan

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,11 +58,39 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
   return mapping;
 }
 
+/// Compute total static elements in the shape. Returns 0 if any dimension
+/// is dynamic.
+static int64_t getTotalStaticElements(ArrayRef<int64_t> shape) {
+  int64_t total = 1;
```
kuhar: You could also do:
```c++
if (llvm::all_of(shape, ShapedType::isStatic)) {
  return llvm::product_of(shape);
} 
return 0;

File: compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp [hunk 7]
```diff
@@ -58,11 +58,36 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
...
+/// we check only the innermost dimension.
+static bool hasEnoughElementsForDMA(ArrayRef<int64_t> shape,
+                                    int64_t minElementsPerTransfer,
+                                    bool canLinearize) {
+  int64_t innermostDim = shape.back();
+  if (canLinearize) {
+    int64_t totalElements = getTotalStaticElements(shape);
+    return totalElements >= minElementsPerTransfer;
+  }
+  return innermostDim >= minElementsPerTransfer;
```
kuhar:
```suggestion
  if (canLinearize) {
    int64_t totalElements = getTotalStaticElements(shape);
    return totalElements >= minElementsPerTransfer;
  }
  return shape.back() >= minElementsPerTransfer;
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -684,9 +785,22 @@ struct GPUConvertToCoalescedDMAPass final
...
+    // Check if we have enough elements for coalesced DMA.
+    // Linearization (using total elements instead of innermost dim for size
+    // check) is only supported for linalg::CopyOp because it requires the
+    // destination to be a fresh allocation (tensor.empty) with no aliasing.
+    // GatherOp doesn't support linearization since its output may alias with
+    // existing data.
+    bool canLinearize = false;
+    if constexpr (std::is_same_v<OpTy, linalg::CopyOp>) {
+      canLinearize = canLinearizeCopy(op);
+    }
```
kuhar: Pass in `Operation *` and exit early inside the implementation if it's not a copy op. This avoids metaprogramming.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -713,10 +827,11 @@ struct GPUConvertToCoalescedDMAPass final
 
   LogicalResult applySubgroupTiling(FunctionOpInterface funcOp) {
     MLIRContext *context = &getContext();
-    SmallVector<Operation *> opsToTile;
+    IRRewriter rewriter(context);
```
kuhar: Undo this code motion? We prefer variables to be defined as close to their use as possible. We don't need a local variable for context either -- it's only used for the rewriter.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,11 +58,35 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
   return mapping;
 }
 
+/// Compute total static elements in the shape. Returns 0 if any dimension
+/// is dynamic.
+static int64_t getTotalStaticElements(ArrayRef<int64_t> shape) {
+  if (!llvm::all_of(shape, ShapedType::isStatic)) {
+    return 0;
+  }
```
kuhar: ubernit / optional
```suggestion
  if (llvm::any_of(shape, ShapedType::isDynamic)) {
    return 0;
  }
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -117,15 +139,76 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
+
+/// Check if an operation can be linearized for DMA.
+/// Linearization is only supported for linalg::CopyOp because it requires the
+/// destination to be a fresh allocation (tensor.empty) with no aliasing.
+static bool canLinearizeOp(Operation *op) {
+  auto copyOp = dyn_cast<linalg::CopyOp>(op);
+  if (!copyOp) {
+    return false;
+  }
+  return canLinearizeCopy(copyOp);
```
kuhar: ubernit / optional
```suggestion
  if (auto copyOp = dyn_cast<linalg::CopyOp>(op)) {
    return canLinearizeCopy(copyOp);;
  }
  return false;
```

kuhar: Or just merge these two overloads into a single function?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -578,14 +663,41 @@ struct GPUConvertToCoalescedDMAPass final
...
     // We treat 0 as 1 for the purpose of counting total warps.
-    auto positiveWarps =
-        llvm::make_filter_range(numWarps, [](int64_t n) { return n > 0; });
-    int64_t totalWarps = llvm::product_of(positiveWarps);
+    int64_t totalWarps = 1;
+    for (int64_t nw : numWarps) {
+      if (nw > 0) {
+        totalWarps *= nw;
+      }
+    }
```
kuhar: Why did you undo the previous logic? IMO the previous version was more idiomatic.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: good suggestion. merged into one.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -117,15 +139,75 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
+  if (rank <= 1) {
+    return true;
+  }
+
+  // For higher ranks, innermost dimension must be fully taken.
+  ArrayRef<int64_t> sourceShape = sourceType.getShape();
+  SmallVector<OpFoldResult> sizes = extractSlice.getMixedSizes();
+
+  int64_t innermostSourceDim = sourceShape[rank - 1];
+  if (ShapedType::isDynamic(innermostSourceDim)) {
```
krzysz00: // TODO we can try to check this?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -117,15 +139,75 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
+    return false;
+  }
+
+  std::optional<int64_t> innermostSliceSize =
+      getConstantIntValue(sizes[rank - 1]);
+  return innermostSliceSize && *innermostSliceSize == innermostSourceDim;
```
krzysz00: I'd generalize this a bit - have it return how many contiguous dimensions there are, so that we can linearize at much as possible even if we can't squish the whole tensor together

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -200,3 +200,162 @@ func.func @copy_prefer_contiguous_subview(%source: tensor<64x128xf32>, %init: te
...
+  // Innermost dimension (16) < minElementsPerTransfer (64), and output is not
+  // tensor.empty(), so linearization is not possible. The copy should remain.
+
+  // CHECK: %[[RESULT:.+]] = linalg.copy
+  // CHECK-SAME: ins(%[[SRC]] : tensor<128x16xf32>)
+  // CHECK-SAME: outs(%[[DST]] : tensor<128x16xf32>)
+  // CHECK: return %[[RESULT]]
+
+  return %result : tensor<128x16xf32>
+}
```
krzysz00, line 361: I think we're missing tests for having some extract_slice operations in the input, unless I misunderstood what this PR does.

lialan: test added

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -147,34 +147,44 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
-/// must be fully taken to ensure contiguous memory layout.
-static bool isContiguousSlice(tensor::ExtractSliceOp extractSlice) {
+/// Get the number of contiguous innermost dimensions in an extract_slice.
+/// The slice must have unit strides. Returns the count of innermost dimensions
+/// that are fully taken from the source, enabling partial linearization.
+/// Returns 0 if the slice is not contiguous (non-unit strides).
+static int64_t getNumContiguousDims(tensor::ExtractSliceOp extractSlice) {
   if (!extractSlice.hasUnitStride()) {
-    return false;
+    return 0;
```
krzysz00: ... This check can be weaker. If the strides are [2, 1, 1], that's up to two contiguous dimensions

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,11 +58,35 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
   return mapping;
 }
 
+/// Compute total static elements in the shape. Returns 0 if any dimension
+/// is dynamic.
+static int64_t getTotalStaticElements(ArrayRef<int64_t> shape) {
+  if (llvm::any_of(shape, ShapedType::isDynamic)) {
```
krzysz00: I claim that this should be something like `getContiguousStaticElements`, no?, and be a function of the contiguous dimension counts we compute

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,11 +58,35 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
...
+}
+
+/// Check if we have enough elements for coalesced DMA.
+/// When canLinearize is true (destination from tensor.empty), we check total
+/// elements since the downstream pass can linearize all dimensions. Otherwise,
+/// we check only the innermost dimension.
+static bool hasEnoughElementsForDMA(ArrayRef<int64_t> shape,
+                                    int64_t minElementsPerTransfer,
+                                    bool canLinearize) {
+  if (canLinearize) {
```
krzysz00: This condition shouldn't be binary like this - you should be checking if there's a suffix of the dimensions of this tensor that's long enough when linearized

krzysz00 (commented):
There're still ways to weaken the check here that were missed

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: Good comment. I made changes to check multiple inner dimensions.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: updated

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: You are right. Now it checks if the contiguous suffix.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,26 +58,34 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
...
-/// Compute total static elements in the shape. Returns 0 if any dimension
-/// is dynamic.
-static int64_t getTotalStaticElements(ArrayRef<int64_t> shape) {
-  if (llvm::any_of(shape, ShapedType::isDynamic)) {
+/// Compute static elements in the contiguous innermost suffix of the shape.
+/// Returns the product of the innermost `numContiguousDims` dimensions.
+/// Returns 0 if any of those dimensions is dynamic.
+static int64_t getContiguousStaticElements(ArrayRef<int64_t> shape,
+                                           int64_t numContiguousDims) {
+  if (numContiguousDims <= 0 || numContiguousDims > (int64_t)shape.size()) {
```
kuhar: Use c++ casts: https://llvm.org/docs/CodingStandards.html#prefer-c-style-casts

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,26 +58,34 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
...
     return 0;
   }
-  return ShapedType::getNumElements(shape);
+
+  auto suffix = shape.take_back(numContiguousDims);
```
kuhar: spell this out: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

qedawkins (approved):
LGTM after final approval from others

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -117,9 +156,22 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
 
-  // If no valid DMA size found or innermost dim is too small, skip.
+  // Determine how many elements are available for coalesced access.
+  // For CopyOp with output tracing to tensor.empty(), we can linearize.
+  ArrayRef<int64_t> shape = outputType.getShape();
+  int64_t availableElements = innermostDim;
+  if (auto copyOp = dyn_cast<linalg::CopyOp>(op)) {
+    Value output = copyOp.getOutputs()[0];
+    if (tracesToTensorEmpty(output)) {
+      if (llvm::none_of(shape, ShapedType::isDynamic)) {
```
krzysz00: Micro-nit: we could remove a level of nesting here

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -117,9 +156,22 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
+        availableElements = ShapedType::getNumElements(shape);
```
krzysz00: I feel the urge to ask - can't we stop the count at the rightmost dynamic dimension instead of only doing this for fully static shapes?

krzysz00: ... wait, not necessarily, the dimension at issue might be 0

krzysz00 (approved):
If we're sure that this only applies to fully static tensors, I don't see anything wrong with it

Issue #22940, opened 2025-12-18
[DT][GPU] Failed to compile 405b model when encoding unification is enabled


hanhanW:
Checkout to https://github.com/iree-org/iree/pull/22923 and run

```
iree-compile ~/llama3_405b_fp4.mlir \
  --iree-hip-target=gfx950 \
  --iree-hal-target-device=hip \
  --iree-opt-level=O3 \
  --iree-dispatch-creation-propagate-collapse-across-expands=true \
  --iree-hal-indirect-command-buffers=true \
  --iree-stream-resource-memory-model=discrete \
  --iree-hip-enable-tensor-ukernels=true \
  --iree-hip-specialize-dispatches \
  --iree-hal-memoization=true \
  --iree-stream-affinity-solver-max-iterations=1024 \
  --iree-dispatch-creation-data-tiling=true \
  --iree-stream-experimental-unify-encoding-for-globals \
  -o /tmp/repro.vmfb
```

It looks like different layout is chosen in `decode_bs4$async_dispatch_29`, which is suspicious. IIRC, they result in the same layout if tensor ukernel is enabled: https://github.com/iree-org/iree/issues/21659#issuecomment-3436635722

I need some time to investigate. I did not notice this issue because I only compiled them to Stream to verify the memory footprint.

hanhanW:
huh, I forgot to update the encodings of bindings in executables, that may be the root cause. Working on it.

hanhanW:
There are other tensor ops need to be updated. E.g., stream.tensor.clone.

```
'stream.tensor.clone' op clones changing tensor encoding from <<NULL ATTRIBUTE>> to #iree_encoding.encoding<operand_index = 1 : index, op_type =  scaled_matmul, element_types = [f4E2M1FN, f4E2M1FN, f8E8M0FNU, f8E8M0FNU, f32], user_indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0, d2, d3)>, affine_map<(d0, d1, d2, d3) -> (d1, d2, d3)>, affine_map<(d0, d1, d2, d3) -> (d0, d2)>, affine_map<(d0, d1, d2, d3) -> (d1, d2)>, affine_map<(d0, d1, d2, d3) -> (d0, d1)>], iteration_sizes = [4, 16384, 512, 32]>; not allowed
```

hanhanW:
> There are other tensor ops need to be updated. E.g., stream.tensor.clone.
> 
> ```
[...]

PR #23059, opened 2026-01-08
[Stream] Update executables in UnifyEncodingForGlobals pass.

hanhanW:
When the same executable is used by dispatch sites targeting different devices with different unified encodings (e.g., multi-device scenarios), the executable may have to be duplicated so each device gets bindings with its own encoding. If they result in the same encodings, duplication is not needed.

The revision also refacotrs reusable executable handling utilities from
SpecializeEncodings into Utils.h/Utils.cpp:
 - gatherUsedDialectInterfaces<T>()
 - getBindingLayoutAttrs()
 - recognizeDispatchEntryPoints()
 - updateBindingEncodings()
 - duplicateExecutablesPerLayoutVariant()

It is not a pure copy-paste refactoring. There are two minor changes:
- Function rename: recognizeEntryPoints → recognizeDispatchEntryPoints (to be more explicit about what it recognizes)
- Parameter change: SymbolTable symbolTable → SymbolTable &symbolTable (pass by reference instead of by value)

It is a step towards https://github.com/iree-org/iree/issues/22940

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/Utils.h
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/Utils.h
@@ -1,0 +1,92 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef IREE_COMPILER_DIALECT_STREAM_TRANSFORMS_UTILS_H_
+#define IREE_COMPILER_DIALECT_STREAM_TRANSFORMS_UTILS_H_
+
+#include "iree/compiler/Dialect/Stream/IR/StreamOps.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/SymbolTable.h"
+#include "mlir/Interfaces/FunctionInterfaces.h"
+
+namespace mlir::iree_compiler::IREE::Stream {
+
+//===----------------------------------------------------------------------===//
+// Dialect Interface Utilities
+//===----------------------------------------------------------------------===//
+
+/// Returns a stably sorted list of dialect interfaces of T for all dialects
+/// used within the given module.
+template <typename T>
+SmallVector<const T *> gatherUsedDialectInterfaces(mlir::ModuleOp moduleOp) {
+  SmallPtrSet<const T *, 4> resultSet;
+  for (auto dialect : moduleOp.getContext()->getLoadedDialects()) {
+    auto *dialectInterface = dialect->getRegisteredInterface<T>();
+    if (!dialectInterface)
+      continue;
+    resultSet.insert(dialectInterface);
+  }
+
+  // NOTE: to ensure deterministic output we sort the result so that imports are
+  // always added in a consistent order.
+  auto results = llvm::to_vector_of<const T *>(resultSet);
+  llvm::sort(
+      results, +[](const T *a, const T *b) {
+        return a->getDialect()->getNamespace().compare(
+                   b->getDialect()->getNamespace()) < 0;
+      });
+  return results;
+}
+
+//===----------------------------------------------------------------------===//
+// Executable Encoding Utilities
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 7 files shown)


hanhanW:
https://github.com/iree-org/iree/pull/23059/changes/7e0897481aa19d3884d1dcd2e768419cbad9f974 is the main implementation.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Utils.h
```diff
@@ -0,0 +1,92 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Returns the operands encodings and result encodings from the `dispatchOp` in
+/// |operands| + |results| order, i.e., it returns the stripped concatenated
+/// operand encodings and result encodings. If a result is tied to an operand,
+/// the result encoding is skipped because it shares the same binding with the
+/// tied operand.
+SmallVector<Attribute> getBindingLayoutAttrs(TensorDispatchOp dispatchOp);
+
+/// Returns true iff all the entry points are recognized by the pass:
```
sommerlukas: The documentation needs to be updated a bit here, `the pass` made sense in the old source file, but here it's unclear to what it relates.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Utils.h
```diff
@@ -0,0 +1,92 @@
...
+
+//===----------------------------------------------------------------------===//
+// Executable Encoding Utilities
+//===----------------------------------------------------------------------===//
+
+/// Returns the operands encodings and result encodings from the `dispatchOp` in
+/// |operands| + |results| order, i.e., it returns the stripped concatenated
+/// operand encodings and result encodings. If a result is tied to an operand,
+/// the result encoding is skipped because it shares the same binding with the
+/// tied operand.
```
sommerlukas, line 52: Why did you drop the example from the documentation? Would it make sense to keep it?

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -203,25 +209,92 @@ util.initializer {
...
+  %encoded_a_v1 = util.global.load immutable @encoded_a_v1 : !stream.resource<constant>
+  %encoded_a_v1_size = util.global.load immutable @encoded_a_v1_size : index
+  %operand_a_0 = stream.async.clone on(#hal.device.affinity<@device_a>)
+    %encoded_a_v1 : !stream.resource<constant>{%encoded_a_v1_size}
+    -> !stream.resource<*>{%encoded_a_v1_size}
+  %encoded_a_v2 = util.global.load immutable @encoded_a_v2 : !stream.resource<constant>
+  %encoded_a_v2_size = util.global.load immutable @encoded_a_v2_size : index
+  %operand_a_1 = stream.async.clone on(#hal.device.affinity<@device_a>)
+    %encoded_a_v2 : !stream.resource<constant>{%encoded_a_v2_size}
+    -> !stream.resource<*>{%encoded_a_v2_size}
```
sommerlukas: Do we need the `clone` here (and other places) to go from `constant` to `*` on the `resource`?

sommerlukas (approved):
Looks good, just a few nits in documentation.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Utils.h
hanhanW: Remove the `by the pass`, as below items clearly describe the conditions.

compiler/src/iree/compiler/Dialect/Stream/Transforms/Utils.h
hanhanW: Good catch! I asked claude to do the refactoring for me. I even asked if it is a pure copy-paste, but it seems like it screw up.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
hanhanW: No we don't, and removed. I was sticking it to the real use case, but it is okay to simplify the test case. thanks!

PR #23060, opened 2026-01-08
runtime: allow to collect host allocator memory usage statistics

Manewing:
This is a first PR to increase the visibility of memory usage by the IREE runtime.

It adds optional reporting for host allocation statistics to iree-run-module and iree-benchmark-module:
```
[[ iree_allocator_t memory statistics ]]
  HOST_ALLOC:     92504575B peak /   1623823585B allocated /   1531383426B freed /     92440159B live
[[ iree_hal_allocator_t memory statistics ]]
  HOST_LOCAL:            0B peak /            0B allocated /            0B freed /            0B live
DEVICE_LOCAL:        28640B peak /   1428451680B allocated /   1428451680B freed /            0B live
```

### DIFF
--- a/runtime/src/iree/base/BUILD.bazel
+++ b/runtime/src/iree/base/BUILD.bazel
@@ -56,6 +56,29 @@ iree_runtime_cc_library(
     ],
 )
 
+iree_runtime_cc_library(
+    name = "base_stats",
+    srcs = [
+        "allocator_stats.c",
+    ],
+    hdrs = ["allocator_stats.h"],
+    deps = [
+        ":base",
+        "//runtime/src/iree/base/internal:synchronization",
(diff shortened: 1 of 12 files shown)


benvanik (changes requested):
this doesn't really fit the existing code and needs a total rewrite ( it should be part of allocator.h/c/_test, not have separate build rules, be named consistently (you use stats/statistics interchangeably, we don't abbreviate like that), and just adds too much in the common path).

but, in general, this is something better served by your underlying allocator where you can get much more useful statistics - we _could_ provide a dumbed-down interface for it, but it would not be built like this.

if you'd like to keep this out of tree you can - it's not integrated - but this is not going to land in this form.

PR #23063, opened 2026-01-08
[DispatchCreation] Extend gather fusion to allow transpose producers

yzhang93:
This PR allows fusion of transpose op with `linalg.generic` of gather.

### DIFF
--- a/compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
@@ -17,6 +17,7 @@
 #include "iree/compiler/Dialect/LinalgExt/Utils/Utils.h"
 #include "iree/compiler/Dialect/Util/IR/UtilOps.h"
 #include "iree/compiler/GlobalOptimization/Passes.h"
+#include "iree/compiler/GlobalOptimization/Utils.h"
 #include "llvm/Support/Debug.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/GlobalOptimization/PropagateLinalgTranspose.cpp
```diff
@@ -17,6 +17,7 @@
 #include "iree/compiler/Dialect/LinalgExt/Utils/Utils.h"
 #include "iree/compiler/Dialect/Util/IR/UtilOps.h"
 #include "iree/compiler/GlobalOptimization/Passes.h"
+#include "iree/compiler/GlobalOptimization/Utils.h"
```
IanWood1: GlobalOpt already depends on Dispatch Creation so I don't think we can use this here. You might be able to replace this with a check that the body is just a yield.

yzhang93: Just see this after I pushed the change. I think we may want to limit the fusion with transpose op for safe. There maybe other generic ops that the body only contains yield, such as reduction, which we we don't want to fuse.

In the new commit, I moved the utility to `DispatchCreation/FusionUtils.cpp`, so it doesn't have the dependency issue.

IanWood1: Maybe it should be moved to https://github.com/iree-org/iree/blob/47fe908217caef04be21608cfcf45d44410f8dad/compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h#L201-L253. There are some other very similar here functions.

yzhang93: This is a good idea, thanks!

IanWood1 (approved):

PR #23069, opened 2026-01-09
[Stream] Handle all stream tensor ops in UnifyEncodingForGlobals

hanhanW:
Extend the pass to handle additional TensorPhaseOp ops that may appear in the encoded global use chain:

- TensorCloneOp: Insert re-encode op before clone if encoding mismatch
- TensorEncodeOp: Update source_encoding attribute directly
- TensorUpdateOp: Insert re-encode op before update for target/update operands

Added `maybeInsertReencode` helper for consistent re-encode op insertion.

The re-encode ops convert from unified encoding back to the encoding expected by each op, maintaining correctness while allowing later passes to fold them.

The rest of ops are not reachable, based on their op definition. Below is the analysis.

| Op | Reachable? | Reason |
|----|------------|----------|
| TensorDispatchOp | ✅ Yes | Can take loaded global as operands. |
| TensorCloneOp | ✅ Yes | Can clone loaded global. |
| TensorEncodeOp | ✅ Yes | Can re-encode the loaded global. |
| TensorUpdateOp | ✅ Yes | Can partially update the value with loaded global. |
| TensorSizeOfOp | ❌ No | No tensor operand - just a TypeAttr. |
| TensorEmptyOp | ❌ No | No tensor input - only produces a tensor. |
| TensorConstantOp | ❌ No | No tensor input - only produces a tensor. |
| TensorSplatOp | ❌ No | No tensor input - takes a scalar, produces a tensor. |
| TensorFillOp | ❌ No | Encoded globals are immutable constants, not fill targets. |
| TensorSliceOp | ❌ No | Slicing happens on resources, not on already-encoded globals. |
| TensorLoadOp | ❌ No | Loading from resource, not tensor. |
| TensorStoreOp | ❌ No | Storing value is scalar/vector, not tensor. |

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
@@ -15,6 +15,7 @@
 #include "iree/compiler/Dialect/Util/IR/UtilOps.h"
 #include "llvm/ADT/MapVector.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/DebugLog.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/IR/SymbolTable.h"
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -604,7 +735,36 @@ static void collectUpdatesForStreamTensorOps(Explorer &explorer,
...
+      // Handle TensorCloneOp - may need re-encode insertion.
+      if (auto cloneOp = dyn_cast<TensorCloneOp>(user)) {
+        LDBG() << "      Found TensorCloneOp";
+        updates[user][operand.getOperandNumber()] = newEncoding;
+        // Continue tracing through the clone's result.
+        worklist.push_back(cloneOp.getResult());
+        return WalkResult::advance();
+      }
+
+      // Handle TensorEncodeOp - may need re-encode insertion.
```
sommerlukas: IIUC, `TensorEncodeOp` isn't handled by re-encoding, but by changing the source encoding. Should this comment be updated to reflect that?

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -604,7 +735,36 @@ static void collectUpdatesForStreamTensorOps(Explorer &explorer,
         return WalkResult::advance();
       }
 
-      // TODO: Handle other tensor phase ops (TensorFillOp, etc.)
+      // Handle TensorCloneOp - may need re-encode insertion.
```
sommerlukas: Would it make sense to use a `TypeSwitch` here as well? That would also allow to relatively easily combine the identical handling of clone and encode op into a single case of the switch.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
```diff
@@ -560,14 +561,144 @@ static void updateTensorDispatchOp(TensorDispatchOp dispatchOp,
...
+  });
+  LDBG() << "  Updated TensorEncodeOp source encoding to " << newEncoding;
+}
+
+// Updates TensorUpdateOp by inserting re-encode if needed.
+static void updateTensorUpdateOp(TensorUpdateOp updateOp,
+                                 const OperandEncodingUpdates &operandUpdates,
+                                 IRRewriter &rewriter) {
+  // Handle target operand.
+  int operandNumber = updateOp.getTargetMutable().getOperandNumber();
```
sommerlukas: Below we just use operand number 0, does it make sense to have it consistent between the two places?

sommerlukas (commented):
Logic looks good, just a few small comments on code style.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: We should avoid magic numbers when possible. I think you refer to line 760? That was a bug, and I removed the code.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Good point, thanks.

compiler/src/iree/compiler/Dialect/Stream/Transforms/UnifyEncodingForGlobals.cpp
hanhanW: Remove the comments. They are not useful to me, because the function focus on what ops are collected but not how they will be updated.

sommerlukas (approved):

PR #23070, opened 2026-01-09
[LinalgExt] Implement generateScalarImplementation for map_gather

Abhishek-Varma:
-- This commit implements `generateScalarImplementation` for `map_gather` op.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -656,14 +656,16 @@ void MapGatherOp::insertTransformationAtStart(
   transformBody.eraseArguments(0, oldOutputIndices.size());
 }
 
-void MapGatherOp::inlineMapGatherBody(
-    OpBuilder &b, Location loc, ValueRange transformBodyIndices,
+/// Shared implementation for inlining the transformation body of map_gather
+/// and map_scatter ops.
+static void inlineMapGatherScatterBodyImpl(
+    OpBuilder &b, Location loc, Region &transformRegion,
+    ValueRange transformBodyIndices,
     function_ref<void(OpBuilder &, Location, ArrayRef<Value>)> bodyBuilder) {
-  Block &transformBlock = getTransformationRegion().front();
+  Block &transformBlock = transformRegion.front();
   IRMapping mapping;
   // Map the induction variables of the loop nest to the block arguments of the
-  // transformation body. The induction variables are the indices looping over
-  // the elements of output operand.
+  // transformation body.
   for (auto [idx, arg] : llvm::enumerate(transformBlock.getArguments())) {
     mapping.map(arg, transformBodyIndices[idx]);
   }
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -625,6 +625,36 @@ LogicalResult MapGatherOp::verify() {
...
+    }
+  }
+
+  // Get the cloned values that were yielded by the transformation body to pass
+  // to the bodyBuilder.
+  SmallVector<Value> mappedYieldedValues = llvm::map_to_vector(
+      transformBlock.getTerminator()->getOperands(),
+      [&](Value operand) -> Value { return mapping.lookupOrDefault(operand); });
+  bodyBuilder(b, loc, mappedYieldedValues);
+}
```
hanhanW: This is almost identical to MapScatterOp version. I think we should just refactor the logic to `inlineMapGatherScatterBodyImp` and both functions call the method.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -454,6 +454,92 @@ LogicalResult GatherOp::generateScalarImplementation(OpBuilder &b, Location loc,
...
+    // The last yielded Value is the padding, the rest are source indices.
+    Value paddingValue = yieldedValues.back();
+    ArrayRef<Value> loadIndices = yieldedValues.drop_back();
+
+    // Check bounds for each source dimension. Start with true so that
+    // for 0-D sources, inBounds is always true.
+    Value inBounds = arith::ConstantIntOp::create(nestedBuilder, nestedLoc,
+                                                  /*value=*/1, /*width=*/1);
+    for (auto [dim, idx] : llvm::enumerate(loadIndices)) {
+      Value zero = arith::ConstantIndexOp::create(nestedBuilder, nestedLoc, 0);
```
hanhanW: This will create many arith.constant op. I think we should use `nestedBuilder.createOrFold` here.

hanhanW (commented):
overall looks good, just a couple nits.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
Abhishek-Varma: Good point! Done.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
Abhishek-Varma: Done. Also pulled out Zero constant creation out of the loop.

hanhanW (approved):

hanhanW:
I think the op should be working e2e on CPU side. Can you add few test cases to https://github.com/iree-org/iree/tree/main/tests/e2e/linalg_ext_ops and enable them for CPU backends?

Abhishek-Varma:
> I think the op should be working e2e on CPU side. Can you add few test cases to https://github.com/iree-org/iree/tree/main/tests/e2e/linalg_ext_ops and enable them for CPU backends?

Yes, I'll raise a follow-up PR for this as the op needs to be added to PartitionableLoopsInterface - will add e2e tests as part of the same PR.

Going ahead with merging this.

PR #23072, opened 2026-01-09
[VectorDistribution] Distribute transfer_read to a single read.

sommerlukas:
The existing generic pattern distributes a `vector.transfer_read` to one `transfer_read` for each element vector and assembles the overall result through `insert_slice`. 

The alternative pattern introduced here distributes a `vector.transfer_read` to a single `transfer_read`. It achieves that by reinterpreting and transposing the underlying `memref` to the distributed shape and then reads the entire distributed vector in one `transfer_read`. 

The pattern has some prerequisites and will fall back to the generic pattern if these are not met.

Currently, it falls back to the existing pattern for permuted/transposed reads.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -529,12 +529,19 @@ struct DistributeTransferReadWithSingleRead final
 
     // Create a memref.reinterpret_cast operation to reinterpret the original
     // memref to the new, distributed shape.
+
+    // Collect the size operands. For static dimensions, this is just an
+    // attribute. For dynamic dimensions, we need to create a `memref.dim`
+    // operation.
     SmallVector<OpFoldResult> outputShapes(llvm::map_range(
         llvm::enumerate(expandedMemShape),
-        [&builder, &readOp](auto dimAndShape) -> OpFoldResult {
+        [&builder, &readOp,
+         undistributedDims](auto dimAndShape) -> OpFoldResult {
           size_t dim = dimAndShape.index();
           int64_t shape = dimAndShape.value();
           if (shape == ShapedType::kDynamic) {
+            assert(dim < undistributedDims &&
+                   "distributed dimensions should be static");
             Value constIndex = arith::ConstantIndexOp::create(builder, dim);
             return memref::DimOp::create(builder, readOp.getBase(), constIndex)
                 ->getResult(0);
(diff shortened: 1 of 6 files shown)


sommerlukas:
I added support for masks in the latest commit.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -300,8 +300,398 @@ static VectorValue extractSliceAsVector(RewriterBase &rewriter, Location loc,
   return cast<VectorValue>(slice);
 }
 
+void appendFirstN(llvm::SmallVectorImpl<int64_t> &appendTo,
+                  llvm::ArrayRef<int64_t> takeFrom, size_t N) {
+  appendTo.append(takeFrom.begin(), std::next(takeFrom.begin(), N));
```
kuhar: use `takeFrom.take_front(N)`

Also clang-tidy warnings above and below

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -300,8 +300,398 @@ static VectorValue extractSliceAsVector(RewriterBase &rewriter, Location loc,
...
+}
+
+void appendFirstNOperands(llvm::SmallVector<Value> &appendTo,
```
kuhar:
```suggestion
void appendFirstNOperands(llvm::SmallVectorImpl<Value> &appendTo,
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -300,8 +300,398 @@ static VectorValue extractSliceAsVector(RewriterBase &rewriter, Location loc,
...
+                          mlir::OperandRange takeFrom, size_t N) {
+  appendTo.append(takeFrom.begin(), std::next(takeFrom.begin(), N));
```
kuhar: similar here

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -300,8 +300,398 @@ static VectorValue extractSliceAsVector(RewriterBase &rewriter, Location loc,
...
+}
```
kuhar: Actually, I'm not convinced we need these helpers at all: just do `append_range(foo, bar.take_front(N))`

kuhar (commented):
Can you fix the clang-tidy warnings?

sommerlukas:
Closing this PR as there's a plan to use a different approach involving `transfer_gather` instead.

PR #23074, opened 2026-01-09
Revert "[LinalgExt] Generalize ArgCompareOp in GPUGeneralizeNamedOpsPass"

bangtianliu:
Reverts iree-org/iree#23015

**Reason**

During my process of plumbing arg_compare op along the VectorDistribute pipeline, I found that working directly with ArgCompareOp (without converting it to linalg.generic) will be a better solution.

The reason is that VectorDistribute performs two-level tiling:
 
 * Outer level: Partial reduction (PartialReductionOuterReduction)
 * Inner level: Inter-thread reduction (subgroup-wise)

Add PartialReductionOuterReduction support for ArgCompareOp directly, in addition to the existing PartialReductionOuterParallel (split-reduction) support. Extending the existing linalg.generic reduction support to handle arg_reduce semantics would require drastic changes to the infrastructure (upstream tilinginterface).

Will add the core implementation back when after tiling we find it is useful for the bufferization + vectorization + vector distribution phases.


kuhar (approved):
Alternatively, we could keep the generalization logic but not perform it by default. But maybe better to remove this code if it's not going to be used.

PR #23075, opened 2026-01-09
[Encoding] Propagate (Un)SetEncodingOp with dynamic encoding dims

jtuyls:
Stacked on top of: https://github.com/iree-org/iree/pull/22907.

This PR adds support for propagating dynamic encoding_dims through encoding operations during dispatch creation:
- **`EncodingTypes.cpp`**: Extract dynamic dimension values (Batch, M, N, K) from contractions, only for dimensions that are dynamic in iteration_sizes
-  **`Utils.cpp`**: Add rematerializeEncodingDims helper to recreate encoding dims at new insertion points during propagation (handles dominance, tensor.dim remapping, pure op cloning)
-  **`HoistEncodingOps.cpp`**: Rematerialize encoding dims when bubbling up set_encoding through generic ops
-  **`EncodingExternalModels.cpp`**: Rematerialize encoding dims when propagating through tensor.cast and similar ops

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.h
+++ b/compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.h
@@ -12,6 +12,7 @@
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/PatternMatch.h"
 
 namespace mlir::iree_compiler::IREE::Encoding {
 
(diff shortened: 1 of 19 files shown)


compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.h
```diff
@@ -80,6 +81,22 @@ MatmulNarrowDim getPo2MatmulNarrowDim(EncodingAttr encoding);
...
+///
+/// Returns the rematerialized dims if successful, or failure if some dims
+/// cannot be rematerialized.
+///
+/// Recursively rematerializes operands as needed. Supports:
+/// - Values that already dominate the insertion point (used as-is)
+/// - tensor.dim ops on the propagation source (recreated from newSource)
+/// - Any pure operation (cloned with recursively rematerialized operands)
+FailureOr<SmallVector<Value>>
+rematerializeEncodingDims(RewriterBase &builder, Operation *insertionPoint,
```
MaheshRavishankar, line 96: Not a blocking comment, but I think we have tried many times doing something similar. @IanWood1 do you have pointers to something similar. Can we make this a bit more re-usable utility (with potentially callbacks) to not be specific to the rematerializaiton of dims this way?

MaheshRavishankar (commented):
I have concerns about the recursive rematerialization. It can be pretty expensive. I think I need a little bit more context on this to understand the cost of rematerialization recursively. The patterns I am matching these back to typically use a fixed point/listeners to handle the cost getting out of hand. Can we discuss this offline first before I get back to reviewing this.

PR #22106, opened 2025-09-25
[Preprocessing] Add pass to sink transpose through pad

IanWood1:
Adds pass to sink transposes through pad ops. This is needed for eliminating transposes in https://github.com/iree-org/iree/issues/22091

PR #23080, opened 2026-01-09
Move SinkTransposeThroughPad preprocessing pass to PropagateLinalgTranspose

yzhang93:
Moved the `SinkTransposeThroughPad` pattern from the preprocessing `SinkTransposeThroughPadPass` to the global optimization `PropagateLinalgTransposePass`. This enables the pattern to work with the existing `SinkTransposeThroughExpandShape` pattern, allowing transposes to sink through `transpose->expand_shape->pad` sequences.'

The pattern is enabled behind a flag to avoid other models runtime regression. The flag `iree-global-opt-enable-sink-transpose-through-pad` is added to pipeline options.


IanWood1 (approved):
Looks good, but maybe @MaheshRavishankar should take a look. He originally suggested creating the preprocessing pass instead of putting the pattern in transpose propagation (https://github.com/iree-org/iree/pull/22106). I can't remember what the exact reasoning was or if the same concerns apply. Considering the new case we are seeing, I think it makes sense to move the pattern here.

MaheshRavishankar (approved):
Its behind a flag, so this is OK for now.

PR #23081, opened 2026-01-09
[LLVMGPU] Add pass to lower vector loads to amdgpu.transpose_load

Max191:
This pass identifies contiguous vector loads that can benefit from AMD's transpose load instruction, which provides better memory access patterns for certain GPU workloads.

The pass works in two phases:
1. Hint injection: Seeds thread ID values with lane hints and propagates these hints through arithmetic operations. Just affine.delinearize_index ops for now, but could be extended in the future.
2. Load transformation: Analyzes load operations to determine if indices follow a pattern suitable for transpose loads (consecutive lanes accessing consecutive memory with appropriate strides).

For now, this is not enabled for 6 bit or 4 bit types, since we have no real usecase for them and they would not be tested. However, the logic should work the same for those types. We just need use cases to test with.

Key components:
- LaneIncrementAttr: Indicates an index value that increments by a fixed amount within lane groups, with optional step and aligned parameters
- LaneConstantAttr: Indicates an index value that is constant within lane groups
- Pattern-based hint propagation through various ops including arith.addi, arith.muli, arith.divui/divsi, arith.remui/remsi, and affine.delinearize_index
- Validation of lane hint parameters (group_size, step must be positive and powers of 2)

The aligned field on LaneIncrementAttr tracks whether the base value at lane 0 is guaranteed to be a multiple of the group size, which is required for safe propagation through delinearize operations.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -585,6 +585,9 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(createFlattenSwizzleHintAllocsPass());
   funcPassManager.addPass(createPropagateDispatchSizeBoundsPass());
   funcPassManager.addPass(IREE::GPU::createLowerIREEGPUOpsPass());
+  if (forROCDL) {
+    funcPassManager.addPass(createROCDLLoadToTransposeLoadPass());
+  }
   funcPassManager.addPass(createUnrollAnnotatedLoopsPass());
   funcPassManager.addPass(createIREELoopInvariantCodeMotionPass());
   if (pipelineOptions.enableReduceSharedMemoryBankConflicts) {
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -585,9 +585,6 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(createFlattenSwizzleHintAllocsPass());
   funcPassManager.addPass(createPropagateDispatchSizeBoundsPass());
   funcPassManager.addPass(IREE::GPU::createLowerIREEGPUOpsPass());
-  if (forROCDL) {
-    funcPassManager.addPass(createROCDLLoadToTransposeLoadPass());
-  }
   funcPassManager.addPass(createUnrollAnnotatedLoopsPass());
   funcPassManager.addPass(createIREELoopInvariantCodeMotionPass());
   if (pipelineOptions.enableReduceSharedMemoryBankConflicts) {
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
@@ -1,0 +1,831 @@
+// Copyright 2025 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.h"
+#include "iree/compiler/Codegen/LLVMGPU/ROCDLPasses.h"
+#include "iree/compiler/Codegen/Utils/GPUUtils.h"
+#include "llvm/ADT/SetVector.h"
+#include "llvm/Support/Debug.h"
+#include "mlir/Dialect/AMDGPU/IR/AMDGPUDialect.h"
+#include "mlir/Dialect/AMDGPU/Utils/Chipset.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-rocdl-load-to-transpose-load"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_ROCDLLOADTOTRANSPOSELOADPASS
+#include "iree/compiler/Codegen/LLVMGPU/ROCDLPasses.h.inc"
+
+namespace {
+
+constexpr int64_t kTransposeLoadLaneGroupSize = 16;
+constexpr amdgpu::Chipset kGfx950 = amdgpu::Chipset(9, 5, 0);
+constexpr llvm::StringLiteral kPassLocalHintAttr = "__pass_local_hint";
+
+//===----------------------------------------------------------------------===//
+// Validation Helpers
+//===----------------------------------------------------------------------===//
+
+/// Returns true if the given memory space is workgroup (LDS) memory.
+static bool isWorkgroupMemory(MemRefType memrefType) {
+  Attribute memSpace = memrefType.getMemorySpace();
+  if (!memSpace) {
+    return false;
+  }
+  if (auto intMemSpace = dyn_cast<IntegerAttr>(memSpace)) {
+    return intMemSpace.getInt() == 3;
+  }
+  if (auto gpuMemSpace = dyn_cast<gpu::AddressSpaceAttr>(memSpace)) {
+    return gpuMemSpace.getValue() == gpu::AddressSpace::Workgroup;
+  }
+  return false;
+}
+
+/// Returns the required vector size for transpose_load given an element type.
+/// Returns std::nullopt if the element type is not supported.
+/// TODO(Max191): Add and test 4-bit and 6-bit element type support.
+static std::optional<int64_t> getTransposeLoadVectorSize(Type elementType) {
+  unsigned bitWidth = elementType.getIntOrFloatBitWidth();
+  switch (bitWidth) {
+  case 8:
+    return 8;
+  case 16:
+    return 4;
+  default:
+    return std::nullopt;
+  }
+}
+
+/// Given a permutation map, find which memref dimension corresponds to the
+/// specified result index in the permutation map.
+/// For example, for affine_map<(d0, d1) -> (d0, d1)> with resultIdx=1,
+/// returns 1 (d1's position).
+/// Asserts that the permutation map is valid and resultIdx is in bounds.
+static int64_t getMemrefDimFromMapResult(AffineMap permMap,
+                                         unsigned resultIdx) {
+  assert(resultIdx < permMap.getNumResults() &&
+         "resultIdx out of bounds for permutation map");
+
+  AffineExpr expr = permMap.getResult(resultIdx);
+  auto dimExpr = dyn_cast<AffineDimExpr>(expr);
+  assert(dimExpr && "permutation map result is not a dimension expression");
+
+  return dimExpr.getPosition();
+}
+
+/// Check if the specified memref dimension has stride 1 (is contiguous).
+static bool isDimensionContiguous(MemRefType memrefType, int64_t dim) {
+  assert(dim >= 0 && dim < memrefType.getRank() &&
+         "dim must be a valid memref dimension");
+
+  // getStridesAndOffset handles all standard layouts including identity.
+  SmallVector<int64_t> strides;
+  int64_t offset;
+  if (failed(memrefType.getStridesAndOffset(strides, offset))) {
+    // For layouts where strides cannot be computed, be conservative.
+    return false;
+  }
+
+  return strides[dim] == 1;
+}
+
+/// Checks if a row index value is uniform across lanes in a 16-lane group.
+/// A row index is uniform if it comes only from:
+/// - Constants
+/// - index_hint ops with lane_constant attribute (uniform across lanes)
+/// - Arithmetic/affine operations on these values
+///
+/// This function assumes that hint propagation has already been done, so all
+/// relevant thread_id ops should already have index_hint users with appropriate
+/// lane attributes.
+static bool isRowIndexUniform(Value rowIndex) {
+  LLVM_DEBUG(llvm::dbgs() << "isRowIndexUniform: checking value: " << rowIndex
+                          << "\n");
+  // Verify that the rowIndex is some function of constants and hint ops.
+  // Use SetVector as worklist to avoid revisiting the same value.
+  llvm::SetVector<Value> worklist;
+  worklist.insert(rowIndex);
+  while (!worklist.empty()) {
+    Value currentVal = worklist.pop_back_val();
+    LLVM_DEBUG(llvm::dbgs()
+               << "  checking worklist item: " << currentVal << "\n");
+    auto opResult = dyn_cast<OpResult>(currentVal);
+    if (!opResult) {
+      LLVM_DEBUG(llvm::dbgs() << "  FAIL: value is not an OpResult (likely a "
+                                 "block argument)\n");
+      return false;
+    }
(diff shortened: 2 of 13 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -581,6 +581,9 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
   // Step 9. Remaining post-bufferization optimizations/lowerings.
   funcPassManager.addPass(createPropagateDispatchSizeBoundsPass());
   funcPassManager.addPass(IREE::GPU::createLowerIREEGPUOpsPass());
+  if (forROCDL) {
+    funcPassManager.addPass(createROCDLLoadToTransposeLoadPass());
+  }
```
Max191: I am actually going to remove this for this PR. I will add it in a later PR, and I'll provide benchmark numbers there.

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,831 @@
...
+constexpr int64_t kTransposeLoadLaneGroupSize = 16;
+constexpr amdgpu::Chipset kGfx950 = amdgpu::Chipset(9, 5, 0);
+constexpr llvm::StringLiteral kPassLocalHintAttr = "__pass_local_hint";
+
+//===----------------------------------------------------------------------===//
+// Validation Helpers
+//===----------------------------------------------------------------------===//
+
+/// Returns true if the given memory space is workgroup (LDS) memory.
+static bool isWorkgroupMemory(MemRefType memrefType) {
```
krzysz00: ... isn't this already a utility function somewhere? I could've sworn we've got it upstream

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,831 @@
...
+
+/// Check if the specified memref dimension has stride 1 (is contiguous).
+static bool isDimensionContiguous(MemRefType memrefType, int64_t dim) {
+  assert(dim >= 0 && dim < memrefType.getRank() &&
+         "dim must be a valid memref dimension");
+
+  // getStridesAndOffset handles all standard layouts including identity.
+  SmallVector<int64_t> strides;
+  int64_t offset;
+  if (failed(memrefType.getStridesAndOffset(strides, offset))) {
```
krzysz00, line 124: Nit - I think a better way to phrase this check is asking if the last `rank - dim` dimensions are contiguous

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,831 @@
...
+                                           Location loc) {
+  if (sizes.empty()) {
+    return {};
+  }
+
+  SmallVector<Value> indices(sizes.size());
+  Value remaining = linearIdx;
+
+  // Process from innermost to outermost dimension
+  for (int64_t i = sizes.size() - 1; i >= 0; --i) {
```
krzysz00: ... why are we not just calling delinearize_index here?

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,831 @@
...
+    FunctionOpInterface funcOp = getOperation();
+
+    // Check if target supports transpose_load (gfx950 only)
+    IREE::GPU::TargetAttr target = getGPUTargetAttr(funcOp);
+    if (!target) {
+      return;
+    }
+    FailureOr<amdgpu::Chipset> chipset =
+        amdgpu::Chipset::parse(target.getArch());
+    if (failed(chipset) || *chipset != kGfx950) {
```
krzysz00, line 779: Thoughts on how this'll be generalized to future architectures?

krzysz00 (commented):
Minor commentary. I haven't reasoned out the logic in full, but nothing jumped out at me as fundamentally incorrect.

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
Max191: This is meant to just check that a single dimension is fully contiguous (stride of 1). We only care about the most contiguous dim for this pattern.

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
Max191: Thanks, I forgot to remove this in my cleanup before sending for review.

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
Max191: It should work for any architecture with the transpose load instructions, but we will need to take a look at the layouts for future architectures. Unless future versions of this instruction are vastly different, my guess would be it is some different parameterization of things like the row size of the transpose or the bitwidths of the loads.

krzysz00:
High-level note, this might end up having a rather nasty phase-ordering problem with something like the flattening pass in https://github.com/iree-org/iree/pull/23083, unless it ends up being the case that things get un-flatttened before they hit the transfer_read and that FoldMemRefAliasOps either hasn't run yet or won't ave made it harder to see the transpose

(yes, that particular pass is fp4 specific, but index flattening is still a transformation we want to do at some point before we hit LLVM, ideally somewhat early)

Max191:
> High-level note, this might end up having a rather nasty phase-ordering problem with something like the flattening pass in #23083, unless it ends up being the case that things get un-flatttened before they hit the transfer_read and that FoldMemRefAliasOps either hasn't run yet or won't ave made it harder to see the transpose
> 
> (yes, that particular pass is fp4 specific, but index flattening is still a transformation we want to do at some point before we hit LLVM, ideally somewhat early)

In the context of LDS swizzling, this op should be no different from a vector load. The indices it carries are exactly the indices that are accessed by the given lane. The only special thing about this op is where the data ends up in the end.

krzysz00:
To rephrase my concern, does this pass work if the index is already 1D - that is, instead of a 4x16xf32 memref I've made a 64xf32 one?

Max191:
> To rephrase my concern, does this pass work if the index is already 1D - that is, instead of a 4x16xf32 memref I've made a 64xf32 one?

Ah, I see. I don't think so, because it needs to look for strided vector loads, but if everything is flattened then that seems like we are running this pass too late. This pass is meant to be run while we still have the shape information about the buffers, which is fairly early on (shortly after bufferization).

krzysz00:
Ok, so we'll want a comment warning people about the shape info requirement.

Max191:
> Ok, so we'll want a comment warning people about the shape info requirement.

I'll add a note in the pass docs in the next round of review comments.

krzysz00 (commented):
I think the logic is fine but I haven't read it carefully - if you're confident in this pass I'm good to stamp it.

One note - does the transpose_load have expand_shape/collapse_shape/subview folders in the upstream patterns in AMDGPU?

krzysz00:
(Want a stamp? Or more of a "we hop on a call and you talk me through the logic to make sure it sounds good when you explain it" sort of deal?)

Max191:
> (Want a stamp? Or more of a "we hop on a call and you talk me through the logic to make sure it sounds good when you explain it" sort of deal?)

I think I have pretty good confidence with the pass now. I'm happy to walk through it with you if you'd like, but a stamp also works for me.

The delay was mostly just due to me testing/triaging on more GEMMs and some convolutions (plus some github issues distracting me). I found an issue with convolution, which I was prototyping a fix for, but then I just realized last night that the issue is actually earlier in the pipeline and not with this pass, so this PR should be pretty much good to go now.

There is one remaining issue which I will push up a quick fix for before merging. It's not valid to perform the transformation on subviews without analyzing the offsets of the subviews, so I just need to restrict the transformation to a chain of whitelisted ops beginning from memref.alloc().

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
+// Copyright 2025 The IREE Authors
```
kuhar:
```suggestion
// Copyright 2026 The IREE Authors
```

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+/// A row index is uniform if it comes only from:
+/// - Constants
+/// - index_hint ops with lane_constant attribute (uniform across lanes)
+/// - Arithmetic/affine operations on these values
+///
+/// This function assumes that hint propagation has already been done, so all
+/// relevant thread_id ops should already have index_hint users with appropriate
+/// lane attributes.
+static bool isRowIndexUniform(Value rowIndex) {
+  LLVM_DEBUG(llvm::dbgs() << "isRowIndexUniform: checking value: " << rowIndex
```
kuhar: Use LDBG from DebugLog.h

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+    }
+
+    ArrayRef<int64_t> basis = op.getStaticBasis();
+    Attribute inputHint = hintOp.getHint();
+    rewriter.setInsertionPointAfter(op);
+
+    return llvm::TypeSwitch<Attribute, LogicalResult>(inputHint)
+        // Case 1: lane_constant propagates to all results unchanged.
+        // Dynamic basis is allowed since we don't need basis values.
+        .Case<IREE::GPU::LaneConstantAttr>([&](auto laneConstant) {
```
kuhar:
```suggestion
        .Case([&](IREE::GPU::LaneConstantAttr laneConstant) {
```
also below

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+              // Not evenly divisible - inherit from inner result.
+              groupSizes[i] = lastValidGroupSize;
+            }
+            // Only grow currentGroupSize for static bases.
+            if (!ShapedType::isDynamic(basis[i])) {
+              currentGroupSize *= basis[i];
+            }
+          }
+
+          for (unsigned i = 0; i < op.getNumResults(); ++i) {
```
kuhar: nit: don't recalculate end

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+/// This pattern is designed to run in the same greedy driver as hint
+/// propagation patterns, allowing hints to propagate incrementally and
+/// transpose_load lowering to fire as soon as hints are available.
+///
+/// The pattern requires hints to already be present on the column index (via
+/// seedThreadIdHints + PropagateHintThroughDelinearize). The greedy driver
+/// iterates until fixpoint, so hints will propagate through all delinearize
+/// ops before this pattern can match.
+struct TransferReadToTransposeLoad : OpRewritePattern<vector::TransferReadOp> {
+  using OpRewritePattern::OpRewritePattern;
```
kuhar:
```suggestion
  using Base::Base;
```

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+    std::optional<SmallVector<int64_t>> workgroupSize =
+        getWorkgroupSize(funcOp);
+    if (workgroupSize) {
+      seedThreadIdHints(funcOp, rewriter, *workgroupSize);
+    }
+
+    // Phase 2: Propagate hints and lower transfer_reads via greedy patterns
+    RewritePatternSet patterns(funcOp.getContext());
+    patterns.add<PropagateHintThroughDelinearize>(funcOp.getContext());
+    patterns.add<TransferReadToTransposeLoad>(funcOp.getContext());
```
kuhar:
```suggestion
    patterns.add<PropagateHintThroughDelinearize, TransferReadToTransposeLoad>(funcOp.getContext());
```

kuhar (commented):
Just some drive-by nits.

Can you rebase this and see if the new clang-tidy workflow complains about anything? (I'm trying to tweak the configuration and find something that works well for us)

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+///   - Result K-1 (last) gets lane_increment<innermost_basis>
+///
+/// For an input with lane_constant<N>:
+///   - All results get lane_constant<N> (constant propagates through)
+///
+/// The pattern fails if any result already has an index_hint user, ensuring
+/// hints are only propagated once.
+struct PropagateHintThroughDelinearize
+    : OpRewritePattern<affine::AffineDelinearizeIndexOp> {
+  using OpRewritePattern::OpRewritePattern;
```
kuhar:
```suggestion
  using Base::Base;
```

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+///   - Results 0 to K-2 get lane_constant<product of bases after that position>
+///   - Result K-1 (last) gets lane_increment<innermost_basis>
+///
+/// For an input with lane_constant<N>:
+///   - All results get lane_constant<N> (constant propagates through)
+///
+/// The pattern fails if any result already has an index_hint user, ensuring
+/// hints are only propagated once.
+struct PropagateHintThroughDelinearize
+    : OpRewritePattern<affine::AffineDelinearizeIndexOp> {
```
kuhar:
```suggestion
struct PropagateHintThroughDelinearize final
    : OpRewritePattern<affine::AffineDelinearizeIndexOp> {
```

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -0,0 +1,792 @@
...
+///
+/// This pattern is designed to run in the same greedy driver as hint
+/// propagation patterns, allowing hints to propagate incrementally and
+/// transpose_load lowering to fire as soon as hints are available.
+///
+/// The pattern requires hints to already be present on the column index (via
+/// seedThreadIdHints + PropagateHintThroughDelinearize). The greedy driver
+/// iterates until fixpoint, so hints will propagate through all delinearize
+/// ops before this pattern can match.
+struct TransferReadToTransposeLoad : OpRewritePattern<vector::TransferReadOp> {
```
kuhar:
```suggestion
struct TransferReadToTransposeLoad final : OpRewritePattern<vector::TransferReadOp> {
```

Max191:
> One note - does the transpose_load have expand_shape/collapse_shape/subview folders in the upstream patterns in AMDGPU?

As far as I can tell, no. It works fine e2e so far, though, even with reshape ops in between, so it is at least functional without it.

krzysz00:
Said folding patterns will get added once I get around to the whole IndexedAccessOpInterface thing landing and make it happen for AMDGPU

krzysz00 (approved):

Max191:
> Can you rebase this and see if the new clang-tidy workflow complains about anything? (I'm trying to tweak the configuration and find something that works well for us)

@kuhar not sure how to get a link to the warning that actually works, but I'll paste a screenshot:
image

Seems like there is a false detection (this is in the big ROCDLLoadToTransposeLoad.cpp file)

PR #23087, opened 2026-01-11
[CI] update onnx_model ci to use huggingface token

Muzammiluddin-Syed-ECE:
Add a huggingface token to env to authenticate requests made to huggingface, avoiding unnecessary rate limitations. This will be used in onnx_model tests after https://github.com/iree-org/iree-test-suites/pull/142/ is merged.

### DIFF
--- a/.github/workflows/pkgci_test_onnx.yml
+++ b/.github/workflows/pkgci_test_onnx.yml
@@ -204,3 +204,5 @@ jobs:
               --timeout=120 \
               --durations=0 \
               --test-config-file=${CONFIG_FILE_PATH}
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}


.github/workflows/pkgci_test_onnx.yml
```diff
@@ -204,3 +204,5 @@ jobs:
               --timeout=120 \
               --durations=0 \
               --test-config-file=${CONFIG_FILE_PATH}
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
ScottTodd, line 208: Let's try to avoid using github secrets if at all possible. Secrets are not shared with workflow runs triggered by forks and are relatively easy to misuse (scope too broadly, allow exfiltration, forget to rotate regularly, etc.).

A few options:
* If these tests are running on persistent self-hosted runners, the runners themselves could have a `HF_TOKEN` environment variable set, ideally to a minimally scoped "read from public repositories" token, but perhaps to a token that grants read access to some gated repositories.
* Set up a service that grants short-lived tokens somehow (we do that in other projects for AWS / Azure / GCP storage access)

ScottTodd: (and if using a secret, we could name it more specifically, like "HF_PUBLIC_READONLY_TOKEN")

Muzammiluddin-Syed-ECE: Further discussion [on discord](https://discord.com/channels/689900678990135345/1166024193599615006/1460327218050896121)

Muzammiluddin-Syed-ECE:
After further discussion it seems that dealing with secrets is something that might be better done reactively rather than proactively. Given that the onnx model tests can rely on caching to alleviate the burden of rate limits from unauthenticated requests from PR / commit runs, this change may be unnecessary. Closing for now, may reopen in the future should the need be required.

PR #23097, opened 2026-01-12
[Codegen][GPU] Add subgroup and lane execution scopes

qedawkins:
Implements baseline execution scopes for subgroup and lane execution. At the moment both are quite minimal, supporting only basic allocations and barriers.

Lane scoped allocs + barriers are unimplemented but also likely never needed.


krzysz00:
Quick note: A lane scoped alloc is an alloca() and may be useful, and the barrier on such memory is a nop

compiler/src/iree/compiler/Codegen/Dialect/GPU/ExternalInterfaces/GPUScopeExternalModels.cpp
```diff
@@ -0,0 +1,152 @@
...
+    // Pad remaining ids with 0.
+    if (numIds > 1) {
+      Value zero = arith::ConstantIndexOp::create(builder, loc, 0);
+      llvm::fill(drop_begin(ids), zero);
+    }
+    return ids;
+  }
+
+  LogicalResult addBarrier(Attribute attr, OpBuilder &builder) const {
+    // Lane-level barriers are not yet supported (missing a generic fence).
```
krzysz00, line 126: Question for the future - is the subgroup-level barrier that is - iirc - wave_barrier on AMD a generally available concept that'd give a reason for a GPU op upstream?

krzysz00 (approved):
Looks fine to me overall

compiler/src/iree/compiler/Codegen/Dialect/GPU/ExternalInterfaces/GPUScopeExternalModels.cpp
qedawkins: Should be generic, I suspect no one has cared enough to do the work split out the memory fence from barrier over implementing their own and that's why we're here.

qedawkins:
> Quick note: A lane scoped alloc is an alloca() and may be useful, and the barrier on such memory is a nop

Yeah I left a brief note about the lane barrier somewhere, but lane scoped alloc would still have to be shared memory no? The "scope" of an allocation indicates that it is available to all workers, not local to that specific worker. This is different from the typical naming conventions, e.g. workgroup memory == memory accessible to all subgroups because when looking at, say, the "subgroup" scope in isolation we don't have a higher level of parallelism to refer to. So we end up with 2 kinds of allocations: thread local and shared ones. Thread local ones can always be created with an alloca, so allocating memory "at a scope" allocates the shared kind. At least that's how I reasoned about the naming, if you have better suggestions I'm open to them.

PR #23101, opened 2026-01-12
Integrate llvm-project@f9a8096067 [ours 31f0e3e644]

krzysz00:
Existing local reverts carried forward:
* Local revert of https://github.com/llvm/llvm-project/pull/169614 due to https://github.com/llvm/llvm-project/issues/172932.

Reverts dropped:
* https://github.com/llvm/llvm-project/pull/174084 and followups - it works now

Local workarounds dropped:
* Remove copying around some Python DLLs since upstream seems to put them in the right place now


Abhishek-Varma (approved):
Thanks!

PR #23102, opened 2026-01-12
[LinalgExt] Add OuterReduction tiling strategy for ArgCompareOp

bangtianliu:
This PR extends ArgCompareOp's PartialReductionOpInterface to support the OuterReduction tiling strategy in addition to the existing OuterParallel (Split-Reduction) strategy.

## Example
For `arg_compare` on `tensor<64x4096xf32>` with reduction `dim=1` and `tile_size=128`:

**OuterParallel** (existing) - each chunk writes to a separate slot:
```mlir
// Partial results: tensor<64x32xf32>, tensor<64x32xi32> (32 chunks)
%results:2 = scf.forall (%chunk_idx) = (0) to (4096) step (128)
    shared_outs(%val = %init_val, %idx = %init_idx) {
  %slice = tensor.extract_slice %input[0, %chunk_idx] [64, 128] [1, 1]
  %partial:2 = iree_linalg_ext.arg_compare dim(1) ins(%slice) outs(...)
  scf.forall.in_parallel {
    tensor.parallel_insert_slice %partial#0 into %val[0, %chunk_idx] [64, 1]
    tensor.parallel_insert_slice %partial#1 into %idx[0, %chunk_idx] [64, 1]
  }
}
%final:2 = linalg.reduce ins(%results#0, %results#1) dims=[1]
```
`OuterReduction `(this PR) - accumulates in place each iteration:
```mlir
// Partial results: tensor<64x128xf32>, tensor<64x128xi32> (tile shape)
%results:2 = scf.for %iv = 0 to 4096 step 128
    iter_args(%val = %init_val, %idx = %init_idx) {
  %slice = tensor.extract_slice %input[0, %iv] [64, 128] [1, 1]
  %updated:2 = linalg.generic ins(%slice) outs(%val, %idx) {
    ^bb0(%new: f32, %acc_val: f32, %acc_idx: i32):
      %global_idx = arith.addi %iv, %local_idx  // track position
      %cmp = arith.cmpf ogt, %new, %acc_val
      %sel_val = arith.select %cmp, %new, %acc_val
      %sel_idx = arith.select %cmp, %global_idx, %acc_idx
      linalg.yield %sel_val, %sel_idx
  }
  scf.yield %updated#0, %updated#1
}
%final:2 = linalg.reduce ins(%results#0, %results#1) dims=[1]
```

This is one necessary step for plumbing through ArgCompare along VectorDistribute pipeline. 

Issue: #23005

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
@@ -2013,8 +2013,7 @@ ArgCompareOp::mergeReductions(OpBuilder &b, Location loc,
       /*index_base=*/nullptr,
       /*dimension=*/reductionDim);
 
-  // Clone the comparator region from the original operRegion &targetRegion =
-  // mergeOp.getRegion();
+  // Clone the comparator region from the original operation.
   Region &targetRegion = mergeOp->getRegion(0);
   Region &sourceRegion = getRegion();
   IRMapping mapper;
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
@@ -1716,6 +1716,209 @@ ArgCompareOp::generateInitialTensorForPartialReduction(
                             indexBroadcastOp->getResult(0)};
 }
 
+// Helper function to create a tiled operation for the OuterReduction strategy.
+//
+// OuterReduction uses scf.for with iter_args to accumulate results across
+// iterations. Within each iteration, we perform element-wise comparisons (not
+// reductions), so we always use linalg.generic. The only difference between
+// explicit and implicit-index modes is whether indices come from the input
+// tensor or are computed from the tile offset.
+static std::pair<Operation *, SmallVector<Operation *>>
+createTiledOpForOuterReduction(ArgCompareOp op, OpBuilder &b, Location loc,
+                               ValueRange init, ArrayRef<OpFoldResult> offsets,
+                               SmallVector<Value> &tiledOperands,
+                               int64_t reductionDim, int64_t rank,
+                               bool isExplicitIndexMode) {
+  SmallVector<Operation *> slices;
+  Value initValResult = init[0];
+  Value initIdxResult = init[1];
+
+  SmallVector<Type, 2> resultTypes;
+  if (op.hasPureTensorSemantics()) {
+    resultTypes = {initValResult.getType(), initIdxResult.getType()};
+  }
+
+  // For OuterReduction, we always use linalg.generic to perform element-wise
+  // comparison, regardless of index mode. The difference is just how we obtain
+  // the indices: from input (explicit) or computed (implicit).
+  AffineMap identityMap = b.getMultiDimIdentityMap(rank);
+
+  // Base setup: input_value, accum_value, accum_index.
+  SmallVector<AffineMap> indexingMaps = {identityMap, identityMap, identityMap};
+  SmallVector<Value> inputs = {tiledOperands[0]};
+
+  if (isExplicitIndexMode) {
+    // Explicit-index mode: Also read indices from the sliced input_index
+    // tensor.
+    indexingMaps.push_back(identityMap);
+    inputs.push_back(tiledOperands[1]);
+  }
+
+  SmallVector<utils::IteratorType> iterators(rank,
+                                             utils::IteratorType::parallel);
+
+  auto genericOp = linalg::GenericOp::create(
+      b, loc, TypeRange{initValResult.getType(), initIdxResult.getType()},
+      inputs, ValueRange{initValResult, initIdxResult}, indexingMaps, iterators,
+      [&](OpBuilder &nestedBuilder, Location nestedLoc, ValueRange args) {
+        Value newVal, accVal, accIdx, newIdx;
+
+        newVal = args[0];
+        if (isExplicitIndexMode) {
+          // Args: [input_value, input_index, accum_value, accum_index].
+          newIdx = args[1];
+          accVal = args[2];
+          accIdx = args[3];
+        } else {
+          // Args: [input_value, accum_value, accum_index].
+          accVal = args[1];
+          accIdx = args[2];
+
+          // Compute the global index for this element.
+          Value localIdx =
+              linalg::IndexOp::create(nestedBuilder, nestedLoc, reductionDim);
+          Value offsetValue = getValueOrCreateConstantIndexOp(
+              nestedBuilder, nestedLoc, offsets[reductionDim]);
+          Value globalIdx = arith::AddIOp::create(nestedBuilder, nestedLoc,
+                                                  offsetValue, localIdx);
+
+          if (Value indexBase = op.getIndexBase()) {
+            globalIdx = arith::AddIOp::create(nestedBuilder, nestedLoc,
+                                              indexBase, globalIdx);
+          }
+
+          // Cast to match the output index tensor's element type.
+          Type idxElemType = accIdx.getType();
+          newIdx = globalIdx;
+          if (!isa<IndexType>(idxElemType)) {
+            newIdx = arith::IndexCastOp::create(nestedBuilder, nestedLoc,
+                                                idxElemType, globalIdx);
+          }
+        }
+
+        // Clone the user's comparison logic to decide whether to update.
+        Block &originalBlock = op.getRegion().front();
+        IRMapping mapper;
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1742,83 +1743,161 @@ FailureOr<TilingResult> ArgCompareOp::tileToPartialReduction(
...
-  for (int64_t i = 0; i < rank; ++i) {
-    if (i == reductionDim) {
-      continue;
+    SmallVector<int64_t> resultShape;
+    for (int64_t i = 0; i < rank; ++i) {
+      if (i == reductionDim) {
+        continue;
+      }
+      if (auto sizeAttr = dyn_cast<Attribute>(initSizes[i])) {
+        resultShape.push_back(cast<IntegerAttr>(sizeAttr).getInt());
```
kuhar: You can also use matchers to get the constant value directly (`m_Constant`)

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1742,83 +1743,161 @@ FailureOr<TilingResult> ArgCompareOp::tileToPartialReduction(
...
+  // OuterReduction requires a different IR structure than OuterParallel. We
+  // cannot use a tiled ArgCompareOp because each iteration must compare the
+  // new tile against the accumulated result, not perform an independent
+  // reduction. Instead, we emit a linalg.generic that applies the comparator
+  // elementwise and updates both value and index accumulators.
+  if (strategy == ReductionTilingStrategy::PartialReductionOuterReduction) {
+    Value tileInput = tiledOperands[0];
+
+    // All operands use identity maps: 1 input + 2 outputs (value, index).
+    AffineMap identityMap =
```
kuhar:
```suggestion
    auto identityMap =
```
See https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1722,9 +1722,10 @@ FailureOr<TilingResult> ArgCompareOp::tileToPartialReduction(
     ArrayRef<OpFoldResult> sizes,
```
kuhar, line 1925: Any change we can break this up? This function is getting really long and complicated

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1882,17 +1972,23 @@ LogicalResult ArgCompareOp::getPartialResultTilePosition(
...
   int64_t reductionDim = getDimension();
-  int64_t inputRank = getInputRank();
   resultOffsets.clear();
   resultSizes.clear();
-  for (int64_t i = 0; i < inputRank; ++i) {
+
+  bool isOuterReduction =
+      tilingStrategy == ReductionTilingStrategy::PartialReductionOuterReduction;
+
+  for (int64_t i = 0, e = getInputRank(); i < e; ++i) {
```
kuhar: You can do `for (auto [i, size, offset] : llvm::enumerate(sizes, offsets))`

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
hanhanW: You can use this method: [getConstantIntValue](https://github.com/llvm/llvm-project/blob/007f1af30eeb5604c04b0fd563af86c23dedbd5c/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L120C24-L120C43) It does the matching for you.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
bangtianliu: Yes, I agree this function is getting large.

I'm currently working on revisiting the split-reduction (OuterParallel) tiling strategy in a separate PR, where I'll implement merge reduction using arg_compare with explicit index input.

After that PR lands, I'll come back to this one and refactor tileToPartialReduction to better handle both strategies with the explicit index model in mind. I'll likely split the OuterReduction and OuterParallel implementations into separate helper functions.

kuhar (dismissed):
LGTM but please wait for another approval before merging

MaheshRavishankar (commented):
I was trying to review this, but I am confused why you chose to generate a `linalg.generic`. That will lose the semantics of this being an arg_compare which you will need later on?
Didnt you already change the implementation of arg_compare to take "current arg_compare" results. Maybe I was distracted for too long here and didnt follow the thread.

bangtianliu:
> I was trying to review this, but I am confused why you chose to generate a `linalg.generic`. That will lose the semantics of this being an arg_compare which you will need later on?
> Didnt you already change the implementation of arg_compare to take "current arg_compare" results. Maybe I was distracted for too long here and didnt follow the thread.

Good question! Let me use a concrete example to show why linalg.generic is needed for OuterReduction.

Setup: argmax on tensor<4xf32> with values [3, 7, 2, 9], tile_size=2 (2 chunks).

---
OuterParallell:
Chunk 0: [3, 7] → argmax → (val=7, idx=1)
Chunk 1: [2, 9] → argmax → (val=9, idx=3)

Partial results: tensor<2xf32> = [7, 9], tensor<2xi32> = [1, 3]
Merge: argmax([7, 9], [1, 3]) → (val=9, idx=3) ✓
Each tile reduces independently, then we merge. Tiled ArgCompareOp is correct here.

---
  OuterReduction:

  With scf.for iter_args, partial result shape = tile shape = tensor<2xf32>.

  Init: (val=[−inf, −inf], idx=[0, 0])

  Iter 0, tile [3, 7]:
    - If we used ArgCompareOp: reduces [3,7] → single (7, 1). WRONG shape!
    - What we need: elementwise compare [3, 7] vs [−inf, −inf]
      → (val=[3, 7], idx=[0, 1])

  Iter 1, tile [2, 9]:
    - Elementwise compare [2, 9] vs [3, 7]
      → position 0: 2 < 3, keep (3, 0)
      → position 1: 9 > 7, update (9, 3)  // idx = offset + pos = 2 + 1 = 3
    → (val=[3, 9], idx=[0, 3])

  Final merge: argmax([3, 9], [0, 3]) → (val=9, idx=3) ✓

  Key point: The inner linalg.generic has all parallel iterators, **and it's not a reduction** at all. It's a pure elementwise map that compares each position independently and updates the accumulator. The "reduction" aspect comes from the outer scf.for loop iterating over tiles, not from the generic itself.

  This is fundamentally different from ArgCompareOp which has a reduction iterator and contracts dimensions. We need a parallel elementwise update here, and linalg.generic with all parallel iterators expresses exactly that.

bangtianliu:
This PR should be rebased and updated according to some changes we made to arg_compare op (explict index mode).

MaheshRavishankar:
> This PR should be rebased and updated according to some changes we made to arg_compare op (explict index mode).

Ok your explanation makes sense. I can take a look, but please let me know after rebase etc. when it is ready for review.

bangtianliu:
Ok, now this PR is ready for review @MaheshRavishankar

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1859,16 +2007,15 @@ ArgCompareOp::mergeReductions(OpBuilder &b, Location loc,
...
+      b, loc, hasPureTensorSemantics() ? TypeRange(getDpsInits()) : TypeRange{},
       /*inputs=*/partialReduce,
       /*outputs=*/getDpsInits(),
-      /*indexBase=*/nullptr,
+      /*index_base=*/nullptr,
       /*dimension=*/reductionDim);
 
-  // Clone the comparator region from the original operation.
-  Region &targetRegion = mergeOp.getRegion();
+  // Clone the comparator region from the original operRegion &targetRegion =
```
MaheshRavishankar: This comment seems to have some code mixed with it.

MaheshRavishankar (approved):
I wouldnt say I fully understand the math here, but from what I could hold in my mind I think this works. Looks good to me. Obvioulys e2e tests will flush out any issues.

Not necessary to change, but just curious if we can reduce the "number of modes" the operation has. It has three modes now. Might need to revisit the operation design at some point. Just leaving some raw notes here/adding as comments might be useful to set the context as to why we need three modes of the operation.

Issue #22973, opened 2025-12-23
Assertion failure in function printer due to invalid tied operand index


sairam2661:
### What happened?

**Description**
The compiler crashes during IR printing when a function's `tied_operands` attribute references an operand index that exceeds the actual number of operands in the function signature.

**Output**
```
iree-opt: iree/third_party/llvm-project/llvm/include/llvm/ADT/STLExtras.h:1247: ReferenceT llvm::detail::indexed_accessor_range_base<mlir::TypeRange, llvm::PointerUnion<const mlir::Value *, const mlir::Type *, mlir::OpOperand *, mlir::detail::OpResultImpl *>, mlir::Type, mlir::Type, mlir::Type>::operator[](size_t) const [DerivedT = mlir::TypeRange, BaseT = llvm::PointerUnion<const mlir::Value *, const mlir::Type *, mlir::OpOperand *, mlir::detail::OpResultImpl *>, T = mlir::Type, PointerT = mlir::Type, ReferenceT = mlir::Type]: Assertion `Index < size() && "invalid index for value range"' failed.
Please report issues to https://github.com/iree-org/iree/issues and include the crash backtrace.
Stack dump:
0.      Program arguments: /workdir/builds/iree/tools/iree-opt --iree-global-opt-infer-numeric-narrowing -split-input-file test.mlir
 #0 0x000056070576874b backtrace (/workdir/builds/iree/tools/iree-opt+0x9174b)
 #1 0x00007fb1fcf46828 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /workdir/sources/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:842:13
 #2 0x00007fb1fcf3e5e3 llvm::sys::RunSignalHandlers() /workdir/sources/iree/third_party/llvm-project/llvm/lib/Support/Signals.cpp:109:18
 #3 0x00007fb1fcf492e4 SignalHandler(int, siginfo_t*, void*) /workdir/sources/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:429:38
 #4 0x00007fb1d705c520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #5 0x00007fb1d70b09fc pthread_kill (/lib/x86_64-linux-gnu/libc.so.6+0x969fc)
 #6 0x00007fb1d705c476 gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x42476)
[...]

Issue #22967, opened 2025-12-23
Assertion failure in Util::FuncOp printer due to out-of-bounds tied_operands index


sairam2661:
### What happened?

**Description**
The compiler crashes during IR printing when encountering a `util.func` with an invalid `tied_operands` attribute. 

**Output**
```
iree-opt: iree/third_party/llvm-project/llvm/include/llvm/ADT/ArrayRef.h:248: const T &llvm::ArrayRef<mlir::Attribute>::operator[](size_t) const [T = mlir::Attribute]: Assertion `Index < Length && "Invalid index!"' failed.
Please report issues to https://github.com/iree-org/iree/issues and include the crash backtrace.
Stack dump:
0.      Program arguments: /workdir/builds/iree/tools/iree-opt test.mlir
 #0 0x000055d9761d874b backtrace (/workdir/builds/iree/tools/iree-opt+0x9174b)
 #1 0x00007fc928ea2828 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /workdir/sources/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:842:13
 #2 0x00007fc928e9a5e3 llvm::sys::RunSignalHandlers() /workdir/sources/iree/third_party/llvm-project/llvm/lib/Support/Signals.cpp:109:18
 #3 0x00007fc928ea52e4 SignalHandler(int, siginfo_t*, void*) /workdir/sources/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:429:38
 #4 0x00007fc902fb8520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #5 0x00007fc90300c9fc pthread_kill (/lib/x86_64-linux-gnu/libc.so.6+0x969fc)
 #6 0x00007fc902fb8476 gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x42476)
 #7 0x00007fc902f9e7f3 abort (/lib/x86_64-linux-gnu/libc.so.6+0x287f3)
 #8 0x00007fc902f9e71b (/lib/x86_64-linux-gnu/libc.so.6+0x2871b)
 #9 0x00007fc902fafe96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
[...]

PR #23108, opened 2026-01-13
[Util] Introduce FuncOp::verify to check for out of bounds tied operands 

bruteforceboy:
Hi! This is my first IREE PR, so hopefully I did things right.

This is my attempt to fix issues #22973 and #22967 by introducing a custom verifier for `FuncOp`, where we check if valid tied operands match an argument.

I have also added a variation of the tests from the issues which previously caused a crash.

cc: @sairam2661


compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp
```diff
@@ -1738,6 +1738,27 @@ void FuncOp::print(OpAsmPrinter &p) {
...
 
+LogicalResult FuncOp::verify() {
+  Operation *op = getOperation();
+
+  auto tiedOperandsAttr = getTiedOperandsAttr();
+  if (tiedOperandsAttr) {
+    auto functionType = cast<FunctionType>(getFunctionTypeAttr().getValue());
+    auto numInputs = functionType.getInputs().size();
+
+    if (llvm::any_of(tiedOperandsAttr.getAsRange<IntegerAttr>(),
```
sommerlukas, line 1745: What happens if one of the attributes is not an `IntegerAttr`? Should this be checked first?

compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp
```diff
@@ -1738,6 +1738,27 @@ void FuncOp::print(OpAsmPrinter &p) {
   }
 }
 
+LogicalResult FuncOp::verify() {
+  Operation *op = getOperation();
+
+  auto tiedOperandsAttr = getTiedOperandsAttr();
+  if (tiedOperandsAttr) {
```
sommerlukas:
```suggestion
  if (auto tiedOperandsAttr = getTiedOperandsAttr()) {
```

compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp
```diff
@@ -1738,6 +1738,27 @@ void FuncOp::print(OpAsmPrinter &p) {
...
+    auto functionType = cast<FunctionType>(getFunctionTypeAttr().getValue());
+    auto numInputs = functionType.getInputs().size();
```
sommerlukas:
```suggestion
    auto numInputs = getFunctionType().getNumInputs();
```

compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp
```diff
@@ -1738,6 +1738,27 @@ void FuncOp::print(OpAsmPrinter &p) {
...
+
+    if (llvm::any_of(tiedOperandsAttr.getAsRange<IntegerAttr>(),
+                     [numInputs](IntegerAttr attr) {
+                       auto index = attr.getInt();
+                       return index !=
+                                  IREE::Util::TiedOpInterface::kUntiedIndex &&
+                              (index < 0 || index >= numInputs);
+                     }))
+      return op->emitError("tied_operands index out of range");
```
sommerlukas: I think if you call `emitError` directly on the `FuncOp` that this member function belongs to, you don't need to define `op` in the beginning of the function.

compiler/src/iree/compiler/Dialect/Util/IR/UtilTypes.cpp
```diff
@@ -409,6 +409,8 @@ detail::getTiedResultOperandIndex(Operation *op, unsigned resultIndex) {
     if (resultIndex >= indexAndLength.second)
       return std::nullopt;
   }
+  if (resultIndex >= valueAttrs.size())
+    return std::nullopt;
```
sommerlukas, line 413: Should we also check this condition in the verifier?

compiler/src/iree/compiler/Dialect/Util/IR/test/tied_operands_ops.mlir
sommerlukas, line 1: I'd suggest adding these tests to the existing `op_verification.mlir` in the same directory instead of creating a new file. The third test also doesn't have an expected error, should there be an error? It is also better to separate each error by putting the split marker (`// -----`) between them.

sommerlukas (changes requested):
Thank you for your contribution! I've added some comments, mostly code style.

compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp
bruteforceboy: From the table definitions [#1](https://github.com/iree-org/iree/blob/b50d762352709be8c3c3254fff6ac328c2f6246d/compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.td#L687) [#2](https://github.com/iree-org/iree/blob/b50d762352709be8c3c3254fff6ac328c2f6246d/compiler/src/iree/compiler/Dialect/Util/IR/UtilBase.td#L52C1-L68C1) and the use of the `tied_operands` attribute [#3](https://github.com/iree-org/iree/blob/b50d762352709be8c3c3254fff6ac328c2f6246d/compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp#L1750C1-L1753C10) [#4](https://github.com/iree-org/iree/blob/b50d762352709be8c3c3254fff6ac328c2f6246d/compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp#L1824C1-L1827C12), I don't think a check is necessary.

compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.cpp
bruteforceboy: right, this works. thanks!

compiler/src/iree/compiler/Dialect/Util/IR/UtilTypes.cpp
bruteforceboy: We can probably skip this in the verifier, because a result may not be tied to an argument, which is what this tries to check.

compiler/src/iree/compiler/Dialect/Util/IR/test/tied_operands_ops.mlir
bruteforceboy: Good idea, I'll move the tests. The third test doesn't have an expected error and that's intentional. `-1` as an index is [`IREE::Util::TiedOpInterface::kUntiedIndex`](https://github.com/iree-org/iree/blob/b50d762352709be8c3c3254fff6ac328c2f6246d/compiler/src/iree/compiler/Dialect/Util/IR/UtilInterfaces.td#L963C1-L964C48)

bruteforceboy:
@sommerlukas I've made a couple of changes here. Thanks for the review.

kuhar:
We determined we have to close this to comply with https://www.linuxfoundation.org/blog/navigating-global-regulations-and-open-source-us-ofac-sanctions

PR #23109, opened 2026-01-13
Fix CTS CheckStatusContains to compare only message portion, ignore stack traces

jtuyls:
This fixes the failing cts tests when run in debug build because stack traces are included:

Failing tests:
```
        1096 - iree/hal/drivers/hip/cts/hip_stream_semaphore_test (Failed)
        1097 - iree/hal/drivers/hip/cts/hip_stream_semaphore_submission_test (Failed)
        1106 - iree/hal/drivers/hip/cts/hip_graph_semaphore_test (Failed)
        1107 - iree/hal/drivers/hip/cts/hip_graph_semaphore_submission_test (Failed)
        1116 - iree/hal/drivers/hip/cts/hip_multi_queue_stream_semaphore_test (Failed)
        1117 - iree/hal/drivers/hip/cts/hip_multi_queue_stream_semaphore_submission_test (Failed)
        1126 - iree/hal/drivers/hip/cts/hip_multi_queue_graph_semaphore_test (Failed)
        1127 - iree/hal/drivers/hip/cts/hip_multi_queue_graph_semaphore_submission_test (Failed)
        1136 - iree/hal/drivers/hip/cts/hip_multi_queue_stream_queue_1_semaphore_test (Failed)
        1137 - iree/hal/drivers/hip/cts/hip_multi_queue_stream_queue_1_semaphore_submission_test (Failed)
        1146 - iree/hal/drivers/hip/cts/hip_multi_queue_graph_queue_1_semaphore_test (Failed)
        1147 - iree/hal/drivers/hip/cts/hip_multi_queue_graph_queue_1_semaphore_submission_test (Failed)
```

This is because the stack trace part differs:

**a_str:**
```
iree/runtime/src/iree/hal/cts/semaphore_test.h:440: CANCELLED; FailThenWait test.; stack:
  0x... hip_stream_semaphore_test <???> (event_semaphore.c:812)  ← clone site
  0x... hip_stream_semaphore_test <???> (semaphore.c:77)
  0x... hip_stream_semaphore_test <???> (semaphore_test.h:448)
  0x... hip_stream_semaphore_test <???> (gtest.cc:2664)
  ...
```

**b_str:**
```
iree/runtime/src/iree/hal/cts/semaphore_test.h:440: CANCELLED; FailThenWait test.; stack:
  0x... hip_stream_semaphore_test <???> (gtest.cc:2664)  ← original creation site
  0x... hip_stream_semaphore_test <???> (gtest.cc:2739)
  0x... hip_stream_semaphore_test <???> (gtest.cc:2885)
  ...
```

### DIFF
--- a/runtime/src/iree/hal/cts/cts_test_base.h
+++ b/runtime/src/iree/hal/cts/cts_test_base.h
@@ -277,7 +277,9 @@ class CTSTestBase : public BaseType, public CTSTestResources {
 
   // Check that a contains b.
   // That is the codes of a and b are equal and the message of b is contained
-  // in the message of a.
+  // in the message of a. Note: We only compare the message portion (before
+  // "; stack:") since iree_status_clone doesn't copy payloads like stack
+  // traces, so the full formatted strings will differ.
   void CheckStatusContains(iree_status_t a, iree_status_t b) {
     EXPECT_EQ(iree_status_code(a), iree_status_code(b));
     iree_allocator_t allocator = iree_allocator_system();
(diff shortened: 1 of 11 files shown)


runtime/src/iree/hal/cts/cts_test_base.h
```diff
@@ -277,7 +277,9 @@ class CTSTestBase : public BaseType, public CTSTestResources {
 
   // Check that a contains b.
   // That is the codes of a and b are equal and the message of b is contained
-  // in the message of a.
+  // in the message of a. Note: We only compare the message portion (before
+  // "; stack:") since iree_status_clone doesn't copy payloads like stack
+  // traces, so the full formatted strings will differ.
```
ScottTodd: Can you also mention `IREE_STATUS_FEATURE_STACK_TRACE` more directly here? Specifically, the contents/format of status string depends on which status features in https://github.com/iree-org/iree/blob/main/runtime/src/iree/base/status.h are enabled, but the check here is only interested in the "message" portion of the status string.

I also wonder if a parameter on `iree_status_to_string` that excludes the stack trace or a `iree_status_message` function would be useful (especially if multiple call sites want to do this sort of processing on the strings).

Thoughts @benvanik ?

ScottTodd (approved):

runtime/src/iree/hal/cts/cts_test_base.h
jtuyls: Yeah, that's nicer but involves a change to the signature and the corresponding usages. I updated the PR with this now, so could you have another look?

ScottTodd: I'll let @benvanik decide. I could go either way.

If the format flag were already there then the test changes are indeed simple and LGTM. I could also see a case for keeping the PR how you had it originally without changing the API and instead just adjusting the comments a bit to mention `IREE_STATUS_FEATURE_STACK_TRACE`.

AWoloszyn:
Shouldn't cloning the status ALSO clone the callstack. Having a differing stack trace on a cloned status is probably NOT what we actually want (since then we no longer know where the error actually occurred in the first place).

jtuyls:
> Shouldn't cloning the status ALSO clone the callstack. Having a differing stack trace on a cloned status is probably NOT what we actually want (since then we no longer know where the error actually occurred in the first place).

Yeah, that would also make sense to me, but here it explicitly says that payloads will not be cloned: https://github.com/iree-org/iree/blob/4745c4fe2d4413219df717de21d9231d14de8cfc/runtime/src/iree/base/status.h#L454C1-L457C41? Let me know what the right solution is here and I will update it to that.

PR #23110, opened 2026-01-13
[Codegen][GPU] Fix lane offset handling in coalesced DMA lowering

jtuyls:
This PR fixes incorrect lane offset handling in `AMDGPULowerCoalescedDMAToGatherLDS` that caused matmul failures when using `--iree-llvmgpu-use-direct-load` for certain matrix shapes (see sweep below).

**Issue explanation:**

When lowering `coalesced_gather_dma` operations to `amdgpu.gather_to_lds`, two issues existed:

1. **Lane offset was incorrectly added after delinearization:** The lane offset was added only to the innermost dimension after delinearization, causing out-of-bounds access when the linearized data spans multiple dimensions and the innermost dimension is smaller than `lanes × elementsPerLane`.

2. **Destination indices were not subgroup-uniform:** The `gather_to_lds` instruction requires the destination offset to be subgroup-uniform (same for all lanes), but both source and destination were using the same delinearized indices.

**Example with shape [16, 64], 32 lanes, 4 elements/lane:**
- Lane 16: offset = 64, but adding to innermost dim gives `[0, 64]` (overflow!)
- Correct: `delinearize(64)` → `[1, 0]` (row 1, col 0)

**Fix:**

The fix generates two separate delinearizations:
- **Source indices (divergent):** `delinearize(linearOffset + laneOffset)` — each lane reads from a different global memory location
- **Destination indices (uniform):** `delinearize(linearOffset)` — all lanes write to the same LDS base (hardware distributes internally)

This correctly distributes lane-parallel access across all dimensions while maintaining the required subgroup-uniform semantics for `gather_to_lds` destination offsets.

**Results on sweep:**

| Shape (M×K×N) | Before Fix | Rel Error | After Fix | Rel Error |
|---------------|------------|-----------|-----------|-----------|
| 256×256×112 | COMPILE_FAIL | - | COMPILE_FAIL | - |
| 256×256×128 | **FAIL** | 2.22e-01 | PASS | 3.00e-07 |
| 256×256×256 | **FAIL** | 3.56e-01 | PASS | 2.81e-07 |
| 1024×1024×512 | **FAIL** | 1.76e-01 | PASS | 3.83e-07 |
| 64×16×64 | PASS | 2.69e-07 | PASS | 2.69e-07 |
| 64×16×65 | PASS | 2.31e-07 | PASS | 2.31e-07 |
| 64×16×96 | PASS | 2.30e-07 | PASS | 2.30e-07 |
| 64×16×112 | COMPILE_FAIL | - | COMPILE_FAIL | - |
| 64×16×128 | **FAIL** | 1.12e+00 | PASS | 2.17e-07 |
| 128×8×128 | PASS | 2.38e-07 | PASS | 2.38e-07 |
| 128×16×128 | PASS | 2.54e-07 | PASS | 2.54e-07 |
| 128×32×128 | PASS | 2.99e-07 | PASS | 2.99e-07 |
| 128×48×128 | PASS | 2.86e-07 | PASS | 2.86e-07 |
| 128×64×128 | PASS | 2.89e-07 | PASS | 2.89e-07 |
| 256×32×256 | **FAIL** | 6.15e-01 | PASS | 3.09e-07 |
| 512×64×512 | **FAIL** | 4.45e-01 | PASS | 3.79e-07 |

**Note** some compilation failures remaining which are unrelated to this PR as they are happening before and after.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -146,52 +146,44 @@ computeTransferSegments(int64_t totalElements, int64_t elementBits,
 
 /// Generates source and destination indices for a GatherToLDS operation.
 ///
+/// The gather_to_lds instruction requires:
+///   - Source indices: per-lane divergent (each lane reads from different loc)
+///   - Destination indices: subgroup-uniform (all lanes write to same LDS base)
+///
 /// Index computation rules for each dimension:
 ///
-/// | Condition                 | dstIdx    | srcIdx                         |
-/// |---------------------------|-----------|--------------------------------|
-/// | Gather dimension          | dimOffset | load(indices[dim][dimOffset])  |
-/// | Non-gather, non-innermost | dimOffset | dimOffset                      |
-/// | Non-gather, innermost     | dimOffset | dimOffset + laneOffset         |
+/// | Condition        | dstIdx       | srcIdx                              |
+/// |------------------|--------------|-------------------------------------|
+/// | Gather dimension | dstDimOffset | load(indices[dim][srcDimOffset])    |
+/// | Non-gather       | dstDimOffset | srcDimOffset                        |
 ///
 /// Where:
-///   - dimOffset: position in this dimension from delinearization
-///   - laneOffset: lane_id * elementsPerLane (for parallel access)
+///   - srcDimOffset: position with lane offset (divergent per lane)
+///   - dstDimOffset: position without lane offset (uniform across subgroup)
 ///   - indices[dim]: index memref mapping dest positions to source positions
-///
-/// The innermost dimension always adds laneOffset so lanes access different
-/// elements in parallel.
 static std::pair<SmallVector<Value>, SmallVector<Value>>
 generateGatherIndices(PatternRewriter &rewriter, Location loc,
-                      ValueRange dimOffsets, Value laneOffset,
+                      ValueRange srcDimOffsets, ValueRange dstDimOffsets,
                       OperandRange indices) {
   SmallVector<Value> srcIndices;
   SmallVector<Value> dstIndices;
   size_t numIndexDims = indices.size();
-  int64_t destRank = dimOffsets.size();
 
-  for (auto [dim, dimOffset] : llvm::enumerate(dimOffsets)) {
-    bool isInnermost = (static_cast<int64_t>(dim) == destRank - 1);
+  for (auto [dim, srcDimOffset, dstDimOffset] :
+       llvm::enumerate(srcDimOffsets, dstDimOffsets)) {
+    // Destination uses uniform offset (no lane offset).
+    dstIndices.push_back(dstDimOffset);
 
-    // Destination always uses the dimension offset directly.
-    dstIndices.push_back(dimOffset);
-
-    // Build source index for this dimension.
+    // Build source index for this dimension (with lane offset).
     Value srcIdx;
     if (dim < numIndexDims) {
       // Gather dimension: load the source index from the index memref.
-      // The index memref maps destination positions to source positions.
+      // Use srcDimOffset (with lane offset) to look up the index.
       srcIdx = memref::LoadOp::create(rewriter, loc, indices[dim],
-                                      ValueRange{dimOffset});
+                                      ValueRange{srcDimOffset});
     } else {
-      // Non-gather dimension: source and dest use the same offset.
-      srcIdx = dimOffset;
-    }
-
-    // For innermost dimension: add lane offset so each lane accesses a
-    // different element in parallel. This distributes the work across lanes.
-    if (isInnermost) {
-      srcIdx = arith::AddIOp::create(rewriter, loc, srcIdx, laneOffset);
+      // Non-gather dimension: use srcDimOffset directly (with lane offset).
+      srcIdx = srcDimOffset;
     }
     srcIndices.push_back(srcIdx);
   }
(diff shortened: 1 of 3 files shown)


krzysz00:
From quickly skimming the test changes on my phone, I don't think this it correct - the lane offset shouldn't be added to the destination index because that needs to be subgroup-uniform

So if there's two additions, we need to get rid of the first one

krzysz00:
... Though I see the problem here and agree that the current code is also incorrect

jtuyls:
> From quickly skimming the test changes on my phone, I don't think this it correct - the lane offset shouldn't be added to the destination index because that needs to be subgroup-uniform
> 
> So if there's two additions, we need to get rid of the first one

Could you elaborate? I think there's only one lane offset addition here now.

Max191:
I believe the concern is with this:
```
    // CHECK: %[[LIN128:.+]] = arith.addi %[[C128]], %[[LANE_OFFSET]]
    // CHECK: %[[DELINEAR128:.+]]:2 = affine.delinearize_index %[[LIN128]] into (4, 128)
    // CHECK: amdgpu.gather_to_lds %[[SRC]][%[[DELINEAR128]]#0, %[[DELINEAR128]]#1], %[[DST]][%[[DELINEAR128]]#0, %[[DELINEAR128]]#1] : vector<4xf32>
```
The fact that the delinearized index (which contains the lane offset) is used for both the src and dst offsets seems wrong, because the gather_to_lds instruction is meant to have a subgroup uniform destination offset. It does seem surprising that this works (although maybe there is a hardware quirk that makes it work), and I think Krzystof is advocating for creating a second delinearize op for the destination index which does not depend on the lane ID. Something like:
```
    // CHECK: %[[LIN128:.+]] = arith.addi %[[C128]], %[[LANE_OFFSET]]
    // CHECK: %[[DELINEAR0_128:.+]]:2 = affine.delinearize_index %[[LIN128]] into (4, 128)
    // CHECK: %[[DELINEAR1_128:.+]]:2 = affine.delinearize_index %[[C128]] into (4, 128)
    // CHECK: amdgpu.gather_to_lds %[[SRC]][%[[DELINEAR0_128]]#0, %[[DELINEAR0_128]]#1], %[[DST]][%[[DELINEAR1_128]]#0, %[[DELINEAR1_128]]#1] : vector<4xf32>
```

jtuyls:
> I believe the concern is with this:
> 
> ```
>     // CHECK: %[[LIN128:.+]] = arith.addi %[[C128]], %[[LANE_OFFSET]]
>     // CHECK: %[[DELINEAR128:.+]]:2 = affine.delinearize_index %[[LIN128]] into (4, 128)
>     // CHECK: amdgpu.gather_to_lds %[[SRC]][%[[DELINEAR128]]#0, %[[DELINEAR128]]#1], %[[DST]][%[[DELINEAR128]]#0, %[[DELINEAR128]]#1] : vector<4xf32>
> ```
> 
> The fact that the delinearized index (which contains the lane offset) is used for both the src and dst offsets seems wrong, because the gather_to_lds instruction is meant to have a subgroup uniform destination offset. It does seem surprising that this works (although maybe there is a hardware quirk that makes it work), and I think Krzystof is advocating for creating a second delinearize op for the destination index which does not depend on the lane ID. Something like:
> 
> ```
>     // CHECK: %[[LIN128:.+]] = arith.addi %[[C128]], %[[LANE_OFFSET]]
>     // CHECK: %[[DELINEAR0_128:.+]]:2 = affine.delinearize_index %[[LIN128]] into (4, 128)
>     // CHECK: %[[DELINEAR1_128:.+]]:2 = affine.delinearize_index %[[C128]] into (4, 128)
>     // CHECK: amdgpu.gather_to_lds %[[SRC]][%[[DELINEAR0_128]]#0, %[[DELINEAR0_128]]#1], %[[DST]][%[[DELINEAR1_128]]#0, %[[DELINEAR1_128]]#1] : vector<4xf32>
> ```

Thanks! This makes a lot of sense! I guess the hardware might just disregard the non-uniform index within the subgroup as that's likely not even possible? I updated the PR now, could you have another look?

kuhar:
> Thanks! This makes a lot of sense! I guess the hardware might just disregard the non-uniform index within the subgroup as that's likely not even possible? I updated the PR now, could you have another look?

I think our implementation promotes VGPRs to SGPRs via readfirstlane which gives you the value with lane id 0.

kuhar:
> > Thanks! This makes a lot of sense! I guess the hardware might just disregard the non-uniform index within the subgroup as that's likely not even possible? I updated the PR now, could you have another look?
> 
> I think our implementation promotes VGPRs to SGPRs via readfirstlane which gives you the value with lane id 0.

If this is true and you don't want to dealinearize twice, you could use `gpu.subgroup_broadcast %x, first_active_lane` to explicitly promote the value from the first lane to a uniform value.

krzysz00:
What the compiler will do is it'll just insert a `readfirstlane` if there's a VGPR in the supposedly "uniform" argument - ideally we want to avoid this or do it ourselves if we absolutely have to 

I think, long-term, we need the backend to promote these sorts of wave ID computations to SGPRs eagerly

krzysz00:
Re the extra delinearize thing, one thing you could do is also delinearize the lane offset and then add that in pairwise - most of those'll be zeroes and fold away

jtuyls:
> Re the extra delinearize thing, one thing you could do is also delinearize the lane offset and then add that in pairwise - most of those'll be zeroes and fold away

Unless I am misunderstanding this, doesn't this reintroduce the bug? `delinearize(A + B, basis) != delinearize(A, basis) + delinearize(B, basis)`.

Also, `linearOffsetVal` is a compile-time constant so one of the delinearize ops should be folded away at some point and we will just have a single delinearize. Reference: https://github.com/iree-org/iree/blob/4745c4fe2d4413219df717de21d9231d14de8cfc/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp#L380

krzysz00:
... Yeah, you're right, my proposal would reintroduce the bug

(and yeah, the comman parts of the delinearization will, we expect, CSE once affine expressions are expanded)

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -366,29 +370,39 @@ struct LowerCoalescedGatherDMAPattern final
...
-          auto delinearize = affine::AffineDelinearizeIndexOp::create(
+
+          // Source indices: add lane offset before delinearization (divergent).
+          Value srcLinearOffset =
+              arith::AddIOp::create(rewriter, loc, linearOffsetVal, laneOffset);
+          auto srcDelinearize = affine::AffineDelinearizeIndexOp::create(
+              rewriter, loc, srcLinearOffset, basis, /*hasOuterBound=*/true);
+
+          SmallVector<Value> srcDimOffsets(outerDimOffsets);
+          for (Value v : srcDelinearize.getResults()) {
```
krzysz00: Nit: this can be something along the lines of llvm::append_range

krzysz00 (approved):
lgtm overall, thanks for finding the bug here

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -366,29 +370,39 @@ struct LowerCoalescedGatherDMAPattern final
...
+          // Destination indices: no lane offset (subgroup-uniform).
+          auto dstDelinearize = affine::AffineDelinearizeIndexOp::create(
               rewriter, loc, linearOffsetVal, basis, /*hasOuterBound=*/true);
-          for (Value v : delinearize.getResults()) {
-            dimOffsets.push_back(v);
+
+          SmallVector<Value> dstDimOffsets(outerDimOffsets);
+          for (Value v : dstDelinearize.getResults()) {
+            dstDimOffsets.push_back(v);
           }
```
kuhar:
```suggestion
          auto dstDimOffsets = llvm::to_vector_of<Value>(dstDelinearize.getResults());
}
```

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
kuhar: just do `auto srcDimOffsets = llvm::to_vector_of<Value>(...);`

jtuyls: `to_vector_of` doesn't work because we lose `outerDimOffsets` (unless using concat within, which is less readable imo). Updated it to use `append_range` now.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
jtuyls: Updated it to use `append_range` now, see comment above.

amd-eochoalo:
~~@jtuyls this broke the windows build https://github.com/iree-org/iree/actions/runs/21179959837/job/60919262533~~

Sorry, not this PR. Thanks!

PR #23111, opened 2026-01-13
[GPU] Account for scale operands in shared memory calculation

Yu-Zhewen:
This PR fixes the shared memory calculation for scaled matmul operations by properly accounting for scale operands in the GPU heuristics.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -72,6 +72,10 @@ static int64_t calculateOperandsSharedMemoryUsedInBytes(
   int64_t tileN = schedule.getTotalNSize() * schedule.getTotalNTileSize() *
                   schedule.getTotalNSubgroupCount();
 
+  // For scaled matmul, the K dimension is split into Ko (outer) and Kb (block),
+  // where elements in a Kb block share the same scale. For lhs and rhs we
+  // account for both Ko and Kb, while for scale operands, only Ko. For regular
+  // matmul, scale bitwidth is 0 so the scale terms below have no effect.
   int64_t tileK = schedule.getTotalKSize() * schedule.getTotalKTileSize();
   int64_t tileKb = schedule.kSizes.back() * schedule.kTileSizes.back();
   int64_t tileKo = tileK / tileKb;
@@ -688,13 +692,9 @@ FailureOr<GPUMMASchedule> deduceMMASchedule(
       int64_t rhsBitwidth = problem.bType.getIntOrFloatBitWidth();
       int64_t resultBitwidth = problem.cType.getIntOrFloatBitWidth();
       int64_t lhsScaleBitwidth =
-          problem.aScaleType.has_value()
-              ? problem.aScaleType->getIntOrFloatBitWidth()
-              : 0;
+          problem.aScaleType ? problem.aScaleType.getIntOrFloatBitWidth() : 0;
       int64_t rhsScaleBitwidth =
-          problem.bScaleType.has_value()
-              ? problem.bScaleType->getIntOrFloatBitWidth()
-              : 0;
+          problem.bScaleType ? problem.bScaleType.getIntOrFloatBitWidth() : 0;
       bool isAligned =
           isValidMMASchedule(problem, schedule, mustBeAligned, subgroupSize,
                              transposedLhs, transposedRhs);
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -17,6 +17,7 @@
 #include "llvm/Support/InterleavedRange.h"
 #include "llvm/Support/MathExtras.h"
 #include "llvm/Support/raw_ostream.h"
+#include "mlir/IR/Remarks.h"
 
 #define DEBUG_TYPE "iree-codegen-gpu-heuristics"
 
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -65,13 +65,24 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const GemmSize &gemmSize) {
...
     int64_t numRhs = 1) {
   int64_t tileM = schedule.getTotalMSize() * schedule.getTotalMTileSize() *
                   schedule.getTotalMSubgroupCount();
   int64_t tileN = schedule.getTotalNSize() * schedule.getTotalNTileSize() *
                   schedule.getTotalNSubgroupCount();
+
   int64_t tileK = schedule.getTotalKSize() * schedule.getTotalKTileSize();
-  return (tileM * tileK * lhsBitwidth + numRhs * tileN * tileK * rhsBitwidth) /
+  int64_t tileKb = schedule.kSizes.back() * schedule.kTileSizes.back();
+  int64_t tileKo = tileK / tileKb;
```
jtuyls, line 82: A comment here to explain what tileKo is and how it's needed to compute the scale memory size needed would be useful.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -34,11 +38,21 @@ struct GPUMatmulShapeType {
...
         cType(c), numHorizontallyFusedOps(numHorizontallyFusedOps) {}
+
+  GPUMatmulShapeType(ArrayRef<int64_t> m, ArrayRef<int64_t> n,
+                     ArrayRef<int64_t> k, ArrayRef<int64_t> batch, Type a,
+                     Type b, std::optional<Type> aScale,
+                     std::optional<Type> bScale, Type c,
+                     int64_t numHorizontallyFusedOps = 1)
+      : mSizes(m), nSizes(n), kSizes(k), batchSizes(batch), aType(a), bType(b),
+        aScaleType(aScale), bScaleType(bScale), cType(c),
+        numHorizontallyFusedOps(numHorizontallyFusedOps) {}
```
jtuyls: You could morph this constructor into the one above with default values for aScale and bScale.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -20,9 +20,13 @@ struct GPUMatmulShapeType {
   SmallVector<int64_t, 2> nSizes;
   SmallVector<int64_t, 2> kSizes;
   SmallVector<int64_t, 2> batchSizes;
+
   Type aType;
   Type bType;
+  std::optional<Type> aScaleType;
+  std::optional<Type> bScaleType;
```
jtuyls: Optional is not needed as Type is nullable as well.

jtuyls (approved):
I noticed the same, but didn't get to fixing yet. Thanks for this!

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -773,24 +773,24 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
 
-  Value lhs = operands[0];
-  Value rhs = operands[1];
-
-  Value init = operands[2];
+  Type lhsElemType = getElementTypeOrSelf(operands[0]);
+  Type rhsElemType = getElementTypeOrSelf(operands[1]);
+  Type initElemType = getElementTypeOrSelf(operands[2]);
+  Type lhsScaleType = nullptr;
+  Type rhsScaleType = nullptr;
```
```suggestion
  Type lhsScaleType;
  Type rhsScaleType;
```

kuhar (commented):
Can we test this?

Yu-Zhewen:
> Can we test this?

It's not easy to demonstrate a change in selected tile size, as scale operands only account for a small portion of the total shared memory (I also asked Claude to find an example, and it couldn't). Instead, I've used the MLIR remark infrastructure (similar to #22620) to directly verify the shared memory calculation.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -694,7 +715,15 @@ FailureOr<GPUMMASchedule> deduceMMASchedule(
...
              << "Predicted Shared Memory Used by Schedule: " << sharedMemoryUsed
              << " bytes";
-      return isAligned && sharedMemoryUsed <= sharedMemLimitInBytes;
+
+      bool isValid = isAligned && sharedMemoryUsed <= sharedMemLimitInBytes;
+      if (isValid) {
+        // Only emit remark for the shared memory usage of the valid schedule.
+        remark::analysis(loc, remark::RemarkOpts::name("SharedMemoryUsage")
+                                  .category("deduceMMASchedule"))
+            << std::to_string(sharedMemoryUsed);
```
kuhar, line 724: This is great, thanks! I wonder if we could further break down the calculation and print what is the size of each sub-allocation, but fine to leave as-is for this PR

kuhar (approved):

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Yu-Zhewen: Yeah I was also thinking that, but prefer not to add too much verbosity just for testing purposes (for debugging, it might be worth doing that). Will leave as-is for now.

PR #23112, opened 2026-01-13
Update iree-test-suites

amd-eochoalo:
* update iree-test-suite ref across the repo
* add golden times for square gemm benchmarks

This update will allow downloading from hugging face and not from github lfs for onnx models.

ci-extra: test_torch


amd-eochoalo:
The ONNX ops test suite that is failing has nothing to do with this update and it is also failing on main.

kuhar (approved):

PR #23113, opened 2026-01-13
[stablehlo] Drop noop sharding ops

b-chmiel:
This PR drops noop sharding `stablehlo.custom_call`s (conversion pattern `NoOpShardingDrop`).

All sharding ops with number of partitions and number of replicas equal to `1` are replaced by their argument, the rest of them are marked as unsupported.


b-chmiel:
@kuhar Would you mind taking a look?

Issue #23040, opened 2026-01-06
[Codegen][AMDGPU] Crash when compiling mixed precision matmul


IanWood1:
### What happened?

Attempting to compile a matmul with one fp32 operand and one bf16 operand results in a crash in `LLVMGPUSelectLoweringStrategyPass`.

It hits this assert: https://github.com/iree-org/iree/blob/6181754a8ab6e5b7ffa7e64e0fe1cc09ac01523b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp#L373-L374

```
iree-compile: iree/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp:378: std::optional<GPUMMASchedule> mlir::iree_compiler::IREE::GPU::getMmaScheduleFromProblemAndTarget(IREE::GPU::TargetAttr, GPUMatmulShapeType, bool, bool, bool, bool, bool, bool, int64_t): Assertion `problem.aType == problem.bType && "expected the same aType and bType."' failed.
Please report issues to https://github.com/iree-org/iree/issues and include the crash backtrace.
Stack dump:
 #0 0x000074090e38cc6e llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /home/ianwood2/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:842:13
 #1 0x000074090e38a7b5 llvm::sys::RunSignalHandlers() /home/ianwood2/iree/third_party/llvm-project/llvm/lib/Support/Signals.cpp:109:18
 #2 0x000074090e38da57 SignalHandler(int, siginfo_t*, void*) /home/ianwood2/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:429:38
 #3 0x0000740905642520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x00007409056969fc __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
 #5 0x00007409056969fc __pthread_kill_internal ./nptl/pthread_kill.c:78:10
 #6 0x00007409056969fc pthread_kill ./nptl/pthread_kill.c:89:10
 #7 0x0000740905642476 gsignal ./signal/../sysdeps/posix/raise.c:27:6
 #8 0x00007409056287f3 abort ./stdlib/abort.c:81:7
[...]

PR #23116, opened 2026-01-13
[LLVMGPU] Handle mixed-precision matmuls by returning nullopt

IanWood1:
Previously, `getMmaScheduleFromProblemAndTarget` would assert that both matmul operands have the same type (aType == bType). This assertion fails for mixed-precision operations like f32 x bf16, causing crashes in debug builds.

This change replaces the assertion with an explicit check that returns `std::nullopt` when operand types don't match, making the behavior consistent between debug and release builds. Mixed-precision operations will fall back to non-MMA tile-and-fuse configurations. This adds a test case in `config_tile_and_fuse.mlir` to verify that mixed-precision matmuls compile without crashing and use the non-MMA configuration.

Workaround for https://github.com/iree-org/iree/issues/23040


MaheshRavishankar (approved):
I am fine with this, cause it fixes a crash, but we probably need a better solution here. I am fine with going with this to unblock.

kuhar (approved):

PR #23118, opened 2026-01-13
[Codegen] Implement reshape propagation by expansion for inner_tiled

Max191:
Adds support for propagating tensor.collapse_shape and tensor.expand_shape
operations through iree_codegen.inner_tiled ops. This enables reshape fusion
to work with GPU MMA operations that use the inner_tiled abstraction.

Two patterns are introduced:
- FoldProducerCollapseShapeWithInnerTiled: Propagates collapse_shape through inner_tiled by expanding the operation and inserting a collapse on the result.
- FoldConsumerExpandShapeWithInnerTiled: Propagates expand_shape back through inner_tiled by expanding all operands.

Only outer (iteration) dimensions can be reshaped; inner dimensions that depend
on the MMA layout are preserved.

Also adds the patterns to the BlockDynamicDimensions pass and the PropagateReshapesByExpansion pass.


compiler/src/iree/compiler/Codegen/Dialect/Codegen/Transforms/ReshapeFusion.cpp
```diff
@@ -0,0 +1,313 @@
...
+  // Only single result inner_tiled ops are tested or used anywhere, so restrict
+  // to single result for now.
+  if (op->getNumResults() != 1)
+    return failure();
+
+  // Only outer dims can be expanded because inner dims depend on the `kind`
+  // attribute's implementation.
+  int64_t outerRank =
+      op.getIndexingMapsArray()[fusedOperand->getOperandNumber()]
+          .getNumResults();
```
nirvedhmeshram, line 38: Isnt this the "innerRank"?

nirvedhmeshram: oh my bad, i forgot the affine maps are for the outer dims

nirvedhmeshram (approved):
LGTM

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Transforms/ReshapeFusion.cpp
Max191: This would be the outer rank, since the indexing maps only track the outer (slowest varying) dimensions.

Issue #23105, opened 2026-01-13
[DT][CPU] compilation failure on unsupported fp types.


hanhanW:
IIRC, I did not see such issue in llama 405B model. However, I expect that this is needed for encode-parameter tool because you could relayout weights that have unsupported FP types on CPU. e.g., `f8E5M2FNUZ`, etc.

I have a local change that allows the emulation for CPU in `ConvertUnsupportedFloatArithPass.cpp`. However, there are more upstream work needed. E.g.,

1. We need a failing mode (i.e., `FailOnUnsupportedFP=false`) for [ConstrainedTruncFOpLowering](https://github.com/llvm/llvm-project/blob/393ba8c0dc98781ea2ada8e16c6c1f77d22008f7/mlir/lib/Conversion/ArithToLLVM/ArithToLLVM.cpp#L182C7-L184).
2. The lowering has to take destionation type into account, because `fp->i8` and `i8->fp` are not supported by `llvm.fptrunc` and `llvm.fpext` ops. We'll need some software emulation to handle the case.

PR #23119, opened 2026-01-14
[CPU] Add fp8 emulation support for CPU backend

hanhanW:
Extend ConvertUnsupportedFloatArithPass to handle the CPU backend by emulating fp8 arithmetic operations through f32. This is needed because LLVM doesn't natively support fptrunc/fpext for fp8 types.

- Add TruncFToFP8 pattern: emulates f32 -> fp8 using integer bit manipulation (extract sign/exp/mant, adjust bias, round, pack to i8).
- Add ExtFFromFP8 pattern: emulates fp8 -> f32 using integer bit manipulation (unpack from i8, adjust bias, pack to i32).
- Supported fp8 formats:
  * FNUZ (f8E5M2FNUZ, f8E4M3FNUZ): NaN=0x80, no Inf, no negative zero 
  * IEEE (f8E5M2): Has Inf and negative zero
  * FN (f8E4M3FN): No Inf, has negative zero
- Handle both scalar and vector types.
- Add lit tests for CPU target emulation.
- Add e2e tests (small_float_arith.mlir) for CPU backend.
- E2E tests: Round-to-nearest-even, denormal conversions, and special value handling

It is a step towards https://github.com/iree-org/iree/issues/23105

Signed-off-by: hanhanW <hanhan0912@gmail.com>

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
@@ -174,10 +174,13 @@ public:
       : rewriter(rewriter), loc(loc), vecType(dyn_cast<VectorType>(type)) {
     Type i32ScalarType = rewriter.getI32Type();
     Type i8ScalarType = rewriter.getI8Type();
+    Type f32ScalarType = rewriter.getF32Type();
     i32Type = vecType ? VectorType::get(vecType.getShape(), i32ScalarType)
                       : i32ScalarType;
     i8Type = vecType ? VectorType::get(vecType.getShape(), i8ScalarType)
                      : i8ScalarType;
+    f32Type = vecType ? VectorType::get(vecType.getShape(), f32ScalarType)
+                      : f32ScalarType;
   }
 
   /// Creates an i32 constant, splatted if working with vectors.
@@ -188,6 +191,16 @@ public:
       return rewriter.createOrFold<arith::ConstantOp>(loc, i32Type, splatAttr);
     }
     return rewriter.createOrFold<arith::ConstantOp>(loc, i32Type, attr);
+  }
+
+  /// Creates an f32 constant, splatted if working with vectors.
+  Value createF32Const(float value) {
+    auto attr = rewriter.getF32FloatAttr(value);
+    if (vecType) {
+      auto splatAttr = SplatElementsAttr::get(cast<ShapedType>(f32Type), attr);
+      return rewriter.createOrFold<arith::ConstantOp>(loc, f32Type, splatAttr);
+    }
+    return rewriter.createOrFold<arith::ConstantOp>(loc, f32Type, attr);
   }
 
   /// Extracts sign, exponent, and mantissa from an f32 value (as i32 bits).
@@ -232,6 +245,7 @@ public:
 
   Type getI32Type() const { return i32Type; }
   Type getI8Type() const { return i8Type; }
+  Type getF32Type() const { return f32Type; }
 
 private:
   RewriterBase &rewriter;
@@ -239,6 +253,7 @@ private:
   VectorType vecType;
   Type i32Type;
   Type i8Type;
+  Type f32Type;
 };
 
 //===----------------------------------------------------------------------===//
@@ -492,18 +507,12 @@ struct TruncFToFP8 final : public OpRewritePattern<arith::TruncFOp> {
 /// Emulates arith.extf from fp8 to f32 using integer bit manipulation.
 /// This implementation follows IREE's runtime/src/iree/base/internal/math.h.
 ///
-/// The runtime uses ldexpf() for denormals:
-///   return sign * ldexpf(mantissa, 1 - bias - mantissa_bits);
-///
-/// Since we can't use ldexpf in MLIR, we enumerate all possible denormal
-/// mantissa values (only 3-7 values for fp8) and compute their f32
-/// representations at pattern-match time. At runtime, we select the
-/// correct precomputed value based on the input mantissa.
+/// For normal values: adjust exponent bias and shift mantissa.
 ///
 /// For denormals (exp=0, mantissa!=0):
 ///   value = mantissa * 2^(1 - bias - mantissa_bits)
-///   f32_exp = kF32Bias + 1 - src_bias - src_mantissa_bits + leading_bit_pos
-///   f32_mantissa = (mantissa - leading_bit) << (23 - leading_bit_pos)
+/// We implement this using uitofp + mulf with a precomputed scale factor,
+/// which is simpler and more general than enumerating all possible values.
 struct ExtFFromFP8 final : public OpRewritePattern<arith::ExtFOp> {
   using Base::Base;
 
@@ -614,58 +623,17 @@ struct ExtFFromFP8 final : public OpRewritePattern<arith::ExtFOp> {
             normalF32Mant));
 
     // For denormals: value = mantissa * 2^(1 - bias - mant_bits)
-    // We need to find the leading 1 bit and normalize.
-    // For small mant_bits (2-3), we can enumerate cases.
-    //
-    // The arithmetic exponent for denormals is: 1 - bias
-    // For mantissa M with leading bit at position P (0-indexed from LSB):
-    //   f32_exp = f32_bias + (1 - src_bias) + P - src_mant_bits
-    //           = f32_bias + 1 - src_bias + P - src_mant_bits
-    //   f32_mant = (M - (1 << P)) << (f32_mant_bits - P)
-    //
-    // Simplified: f32_exp = 127 + 1 - src_bias - src_mant_bits + P
-    //                     = 128 - src_bias - src_mant_bits + P
-    int denormalBaseExp = 128 - srcInfo.bias - srcInfo.mantBits;
-
-    // Enumerate all possible denormal mantissa values and precompute their f32
-    // bit representations. We use enumeration instead of ldexpf() because:
-    //   1. MLIR arith dialect has no ldexp op (would need math dialect).
-    //      However, the pass is intended to be run late in the pipeline where
-    //      math dialect ops may not be allowed.
-    //   2. Our pattern works in integer domain (bit manipulation on i32).
-    //   3. fp8 has only 3-7 possible denormal values, so enumeration is
-    //      practical.
-    // For mantissa_bits=2 (E5M2): mantissa ∈ {1, 2, 3}
-    // For mantissa_bits=3 (E4M3): mantissa ∈ {1, 2, 3, 4, 5, 6, 7}
-    Value denormalResult = c0;
-    for (int m = (1 << srcInfo.mantBits) - 1; m >= 1; --m) {
-      // Find leading bit position P (0-indexed).
-      int leadingBitPos = 0;
-      for (int i = srcInfo.mantBits - 1; i >= 0; --i) {
-        if (m & (1 << i)) {
-          leadingBitPos = i;
-          break;
-        }
-      }
-
-      // Compute f32 representation for this denormal mantissa.
-      int f32Exp = denormalBaseExp + leadingBitPos;
-      // Remove leading 1 bit and shift to f32 mantissa position.
-      int mantWithoutLeading = m - (1 << leadingBitPos);
-      int f32Mant = mantWithoutLeading << (kF32MantBits - leadingBitPos);
-      int f32Bits = (f32Exp << kF32MantBits) | f32Mant;
-
-      Value cMantVal = helper.createI32Const(m);
-      Value cF32Bits = helper.createI32Const(f32Bits);
-      Value mantMatches = arith::CmpIOp::create(
-          rewriter, loc, arith::CmpIPredicate::eq, mantSrc, cMantVal);
-
-      // Add sign to the f32 value.
-      Value signedF32Bits =
-          arith::OrIOp::create(rewriter, loc, f32Sign, cF32Bits);
-      denormalResult = arith::SelectOp::create(rewriter, loc, mantMatches,
-                                               signedF32Bits, denormalResult);
-    }
+    // Use mulf with a precomputed scale factor instead of enumerating values.
+    Value mantF32 =
+        arith::UIToFPOp::create(rewriter, loc, helper.getF32Type(), mantSrc);
+    float scaleValue = std::ldexp(1.0f, 1 - srcInfo.bias - srcInfo.mantBits);
+    Value scale = helper.createF32Const(scaleValue);
+    Value denormalF32 = arith::MulFOp::create(rewriter, loc, mantF32, scale);
+    // Apply sign: bitcast to i32, OR with sign bit.
+    Value denormalI32 = arith::BitcastOp::create(
+        rewriter, loc, helper.getI32Type(), denormalF32);
+    Value denormalResult =
+        arith::OrIOp::create(rewriter, loc, f32Sign, denormalI32);
 
     // Select cascade for final result (same ordering logic as TruncF).
     // Later selects override earlier ones. NaN must be last (highest priority).
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
@@ -24,7 +24,7 @@
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Interfaces/FunctionInterfaces.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/WalkPatternRewriteDriver.h"
 
 #define DEBUG_TYPE "iree-convert-unsupported-float-arith"
 
@@ -352,7 +352,7 @@ private:
 ///
 /// 3. FN types (f8E4M3FN): No Inf, but has negative zero.
 ///    - Max exponent values are valid finite numbers (except NaN encoding).
-struct TruncFToFP8 final : public OpRewritePattern<arith::TruncFOp> {
+struct TruncFToFP8 final : OpRewritePattern<arith::TruncFOp> {
   using Base::Base;
 
   LogicalResult matchAndRewrite(arith::TruncFOp op,
@@ -572,7 +572,7 @@ struct TruncFToFP8 final : public OpRewritePattern<arith::TruncFOp> {
 ///   value = mantissa * 2^(1 - bias - mantissa_bits)
 /// We implement this using uitofp + mulf with a precomputed scale factor,
 /// which is simpler and more general than enumerating all possible values.
-struct ExtFFromFP8 final : public OpRewritePattern<arith::ExtFOp> {
+struct ExtFFromFP8 final : OpRewritePattern<arith::ExtFOp> {
   using Base::Base;
 
   LogicalResult matchAndRewrite(arith::ExtFOp op,
@@ -822,9 +822,7 @@ void ConvertUnsupportedFloatArithPass::runOnOperation() {
   if (isCPU) {
     RewritePatternSet emulationPatterns(context);
     emulationPatterns.add<TruncFToFP8, ExtFFromFP8>(context);
-    if (failed(applyPatternsGreedily(funcOp, std::move(emulationPatterns)))) {
-      return signalPassFailure();
-    }
+    walkAndApplyPatterns(funcOp, std::move(emulationPatterns));
   }
 }
 
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
@@ -61,62 +61,102 @@ void appendTypesIf(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
 }
 
 //===----------------------------------------------------------------------===//
-// Float type format parameters
+// Floating-point format constants and type info
 //===----------------------------------------------------------------------===//
+//
+// This follows the same approach as IREE's
+// runtime/src/iree/base/internal/math.h for floating-point conversions. The
+// compiler emits equivalent logic using MLIR arith ops instead of C control
+// flow.
+//
+// IEEE 754 floating-point format: [sign | exponent | mantissa]
+//   - Sign: 1 bit (0 = positive, 1 = negative)
+//   - Exponent: biased (stored = actual + bias)
+//   - Mantissa: fractional bits with implicit leading 1 for normal values
+//
+// Special values:
+//   - Zero: exp=0, mantissa=0 (signed zero if format supports it)
+//   - Denormal: exp=0, mantissa!=0, value = mantissa * 2^(1-bias-mantissa_bits)
+//   - Inf: exp=max, mantissa=0 (IEEE types only)
+//   - NaN: exp=max, mantissa!=0 (IEEE), or special encoding (FNUZ: 0x80)
+
+// F32 format constants (IEEE 754 binary32).
+constexpr int kF32MantBits = 23;
+constexpr int kF32Bias = 127;
+constexpr int kF32MantMask = (1 << kF32MantBits) - 1;
+constexpr int kF32MaxExp = 0xFF; // All exponent bits set (255).
 
 /// Parameters describing a floating-point format for conversion.
+/// Derived from APFloat semantics at runtime to support all MLIR float types.
 struct FloatTypeInfo {
-  unsigned expBits;     // Number of exponent bits.
-  unsigned mantBits;    // Number of mantissa bits (excluding implicit bit).
-  int bias;             // Exponent bias.
-  bool hasInf;          // Whether the format supports infinity.
-  bool hasNegZero;      // Whether the format supports negative zero.
+  unsigned expBits;  // Number of exponent bits.
+  unsigned mantBits; // Number of mantissa bits (excluding implicit leading 1).
+  int bias;          // Exponent bias: stored_exp = actual_exp + bias.
+  bool hasInf;       // Format supports infinity (exp=max, mantissa=0).
+  bool hasNan;       // Format supports NaN.
+  bool hasNegZero;   // Format supports negative zero.
+  bool nanAsNegZero; // NaN encoded as 0x80 (FNUZ types).
+  unsigned signMask; // Bit mask for sign bit.
+  unsigned expMask;  // Bit mask for exponent field.
+  unsigned mantMask; // Bit mask for mantissa field.
   unsigned nanEncoding; // Bit pattern for canonical NaN.
-  unsigned infEncoding; // Bit pattern for +Inf (0 if no Inf support).
+  unsigned infEncoding; // Bit pattern for +Inf (0 if no Inf).
+  unsigned maxFinite;   // Bit pattern for max finite value.
 };
 
-/// Returns format info for a float type by querying APFloat semantics. It
-/// assumes that the type can always be casted to FloatType.
+/// Returns format info for a float type by querying APFloat semantics.
 static FloatTypeInfo getFloatTypeInfo(Type type) {
   auto floatType = cast<FloatType>(type);
   const llvm::fltSemantics &sem = floatType.getFloatSemantics();
   FloatTypeInfo info;
 
   // Derive format parameters from APFloat semantics.
-  // Precision includes the implicit bit, so mantissa bits = precision - 1.
   info.mantBits = llvm::APFloat::semanticsPrecision(sem) - 1;
   unsigned totalBits = llvm::APFloat::semanticsSizeInBits(sem);
-  // Total bits = 1 (sign) + expBits + mantBits.
   info.expBits = totalBits - 1 - info.mantBits;
 
-  // Bias = 1 - minExponent. This works for both IEEE and FNUZ formats.
-  // For f32: minExp = -126, so bias = 1 - (-126) = 127.
-  // For f8E5M2FNUZ: minExp = -15, so bias = 1 - (-15) = 16.
+  // Compute masks.
+  info.signMask = 1u << (info.expBits + info.mantBits);
+  info.mantMask = (1u << info.mantBits) - 1;
+  info.expMask = ((1u << info.expBits) - 1) << info.mantBits;
+
+  // Bias from APFloat semantics.
   int minExp = llvm::APFloat::semanticsMinExponent(sem);
   info.bias = 1 - minExp;
 
   info.hasInf = llvm::APFloat::semanticsHasInf(sem);
-  // Check for negative zero support by seeing if getZero with negative=true
-  // produces a negative zero (vs NaN in FNUZ types where 0x80 encodes NaN).
+  info.hasNan = llvm::APFloat::semanticsHasNaN(sem);
+
+  // Check for FNUZ types where NaN is encoded as negative zero (0x80).
+  // These types have no negative zero and no infinity.
   llvm::APFloat negZero = llvm::APFloat::getZero(sem, /*Negative=*/true);
   info.hasNegZero = negZero.isZero() && negZero.isNegative();
+  info.nanAsNegZero = !info.hasNegZero && !info.hasInf && info.hasNan;
 
-  // Compute NaN and Inf encodings based on format type.
+  // Compute NaN and Inf encodings.
   unsigned maxExpCode = (1u << info.expBits) - 1;
-  unsigned mantMask = (1u << info.mantBits) - 1;
-
-  if (!info.hasNegZero && !info.hasInf) {
-    // FNUZ types: NaN = 0x80 (sign bit set, all else zero), no Inf.
-    info.nanEncoding = 1u << (totalBits - 1);
-    info.infEncoding = 0;
-  } else if (info.hasInf) {
-    // IEEE types: NaN = all exp bits + some mant bits, Inf = all exp bits.
-    info.infEncoding = maxExpCode << info.mantBits;
-    info.nanEncoding = info.infEncoding | mantMask;
+  if (info.nanAsNegZero) {
+    // FNUZ types: NaN = 0x80 (sign bit set, all else zero).
+    info.nanEncoding = info.signMask;
   } else {
-    // FN types (like f8E4M3FN): No Inf, but has NaN at max exp with mant != 0.
-    info.infEncoding = 0;
-    info.nanEncoding = (maxExpCode << info.mantBits) | mantMask;
+    // IEEE and FN types: NaN = all exp bits + some mantissa bits.
+    info.nanEncoding = info.expMask | info.mantMask;
+  }
+
+  // Inf encoding: only meaningful for types with infinity.
+  info.infEncoding = info.hasInf ? info.expMask : 0;
+
+  // Max finite value: for types with infinity, it's exp=(max-1), mantissa=all
+  // 1s. For types without infinity, max exp is valid, so exp=max, mantissa=all
+  // 1s (except for FN types where max mantissa is NaN).
+  if (info.hasInf) {
+    info.maxFinite = ((maxExpCode - 1) << info.mantBits) | info.mantMask;
+  } else if (info.hasNan && !info.nanAsNegZero) {
+    // FN types: max exp is valid but max mantissa is NaN.
+    info.maxFinite = info.expMask | (info.mantMask - 1);
+  } else {
+    // FNUZ or no-NaN types: all bit patterns except NaN are valid.
+    info.maxFinite = info.expMask | info.mantMask;
   }
 
   return info;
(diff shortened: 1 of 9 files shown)


hanhanW (commented):
This is mostly done by Claude with my guidance. We test special values (NaN/Inf/-0/+0/etc) in e2e tests, which demonstrates that the emulation is correct.

compiler/src/iree/compiler/Codegen/Common/test/convert_unsupported_float_arith.mlir
```diff
@@ -49,13 +49,13 @@ func.func @negf_f8_unsupported_ocp(%arg0 : f8E4M3FN) -> f8E4M3FN attributes
...
 // CHECK-LABEL: func.func @expand_f8_ocp(
-// CHECK-SAME: %[[ARG0:.*]]: f8E5M2
-// CHECK: %[[EXT0:.*]] = arith.extf %[[ARG0]] {{.*}} : f8E5M2 to f32
-// CHECK: %[[CST:.*]] = arith.constant 1.000000e+00 : f8E5M2
-// CHECK: %[[EXT1:.*]] = arith.extf %[[CST]] {{.*}} : f8E5M2 to f32
-// CHECK: %[[SUM:.*]] = arith.addf %[[EXT0]], %[[EXT1]] : f32
-// CHECK: %[[TRUNC:.*]] = arith.truncf %[[SUM]] {{.*}} : f32 to f8E5M2
-// CHECK: return %[[TRUNC]] : f8E5M2
+// CHECK-SAME:    %[[ARG0:.*]]: f8E5M2
+// CHECK:         %[[EXT0:.*]] = arith.extf %[[ARG0]] {{.*}} : f8E5M2 to f32
```
MaheshRavishankar, line 53: uber ultra nit: I usually try to align the `:`s :D

MaheshRavishankar (commented):
I mean the pass structure and stuff looks OK to me. I dont initimately know all the details about fp NaN/Inf semantics. So dont really have an opinion on that... but maybe someone who knows that can review.... Looking at tests coverage and how it gets the legal types on CPU and GPU, I think this looks reasonable to me. Ill come back and look at more details in a bit, but someone should review the flt semantics.

bjacob:
Have you considered reusing https://github.com/iree-org/iree/blob/a872b71de271a4e8f894fc4fa63f60c034f59999/runtime/src/iree/base/internal/math.h#L267-L611 as either microkernels or a template to port to compiler patterns (it supports more types, with full denormals and to-nearest-even support) ?

krzysz00 (changes requested):
... why aren't we using the upstream patterns that already exist for this? Or, if there are missing patterns upstream, why aren't we adding them there?

hanhanW:
> Have you considered reusing
> 
> https://github.com/iree-org/iree/blob/a872b71de271a4e8f894fc4fa63f60c034f59999/runtime/src/iree/base/internal/math.h#L267-L611
> 
> as either microkernels or a template to port to compiler patterns (it supports more types, with full denormals and to-nearest-even support) ?

I was not aware of this. I think you meant generating some symbols and linked into our codegened libraries (that you showed)? I tried, and I failed. It is tricky to set up the type converter and LLVM conversion. The fp8 types are not [LLVM compatible types](https://github.com/llvm/llvm-project/blob/7d5d4db834f2d982944c64cccbd84edd61d3cdb9/mlir/lib/Dialect/LLVMIR/IR/LLVMTypes.cpp#L788-L801), so the truncf lowering is off if [FailOnUnsupportedFP](https://github.com/llvm/llvm-project/blob/46de3200388c300ee871f5b4941dff0d97ba9480/mlir/lib/Conversion/ArithToLLVM/ArithToLLVM.cpp#L178-L181) is true. If we allow the failure, a `trunc(f32 -> i8)` op, which is invalid, is generated. Because we use i8 type for fp8 types in type conversion. Thus, I ended up with doing software emulation in compiler patterns.

I think I can follow the code and port it to compiler patterns. The current emulation is not too-hardcoded because it also uses expBits/mantBits to do the emulation, though it currently only work for fp8 types atm.

```cpp
/// Parameters describing a floating-point format for conversion.
struct FloatTypeInfo {
  unsigned expBits;     // Number of exponent bits.
  unsigned mantBits;    // Number of mantissa bits (excluding implicit bit).
  int bias;             // Exponent bias.
  bool hasInf;          // Whether the format supports infinity.
  bool hasNegZero;      // Whether the format supports negative zero.
  unsigned nanEncoding; // Bit pattern for canonical NaN.
  unsigned infEncoding; // Bit pattern for +Inf (0 if no Inf support).
};
```

There are few features are missing in current emulation. One is full denormals and the other is to-nearest-even support. I can port them to patterns as well.

The other benefit is that this can enable the emulation for all the backends but not only CPU. IMO, they should function even if the performance is bad.

> ... why aren't we using the upstream patterns that already exist for this? Or, if there are missing patterns upstream, why aren't we adding them there?

Because I did not find upstream patterns and I think they don't exist. If they exist, I'm happy to use them directly.

I don't contribute much to upstream now because the tests are fragile and there may be other noises that make me doing my work slowly. E.g., the transformation itself can be upstreamed, but how do we test it e2e in upstream? If someone makes mistake and it will make integrate painful as the regression can only be observed in IREE. Honestly, I made several changes when I tried to make sure all special values (e.g., NaN/Inf/etc) are handled correctly. If we can have same quality tests in upstream, I'd be more happy to upstream them. Otherwise, I'd keep them in IREE for a while.

bjacob:
> There are few features are missing in current emulation. One is full denormals and the other is to-nearest-even support. I can port them to patterns as well.

Thanks for the explanation. Given that (and pending confirmation that such patterns don't exist upstream yet), OK to carry on with either the current approach, or porting the runtime code to patterns.  The latter would be a form of deduplication but I don't want to make it a requirement as long as the pattern code is sound, I'm OK to review it.

Before I do get into reviewing code, yes we do need these features: denormals and to-nearest-even. While those features are minute detains on high-mantissa types such as fp32, for fp8, precisely due to the narrow mantissas, they become essential features. For example, with f8e5m2 having only 2 bits of mantissa, the representable values between 1.0 and 2.0 are {1.0, 1.25. 1.5. 1.75, 2.0}. So tie-breaks make differences of 0.25, which is too large to ignore.  Denormals I admit are a less critical feature for f8 types (they become more essential at even lower bit depths) but still, for f8e4m3, the smallest normal value is 1/64 and the denormal range is between 1/512 and 1/64, which is going to be more impactful than f32 denormals.

hanhanW:
> > There are few features are missing in current emulation. One is full denormals and the other is to-nearest-even support. I can port them to patterns as well.
> 
> Thanks for the explanation. Given that (and pending confirmation that such patterns don't exist upstream yet), OK to carry on with either the current approach, or porting the runtime code to patterns. The latter would be a form of deduplication but I don't want to make it a requirement as long as the pattern code is sound, I'm OK to review it.
> 
> Before I do get into reviewing code, yes we do need these features: denormals and to-nearest-even. While those features are minute detains on high-mantissa types such as fp32, for fp8, precisely due to the narrow mantissas, they become essential features. For example, with f8e5m2 having only 2 bits of mantissa, the representable values between 1.0 and 2.0 are {1.0, 1.25. 1.5. 1.75, 2.0}. So tie-breaks make differences of 0.25, which is too large to ignore. Denormals I admit are a less critical feature for f8 types (they become more essential at even lower bit depths) but still, for f8e4m3, the smallest normal value is 1/64 and the denormal range is between 1/512 and 1/64, which is going to be more impactful than f32 denormals.

Thanks! Let me convert it to draft and I'll open it once I complete the features.

bjacob:
SG! Take a look at how the linked code handles denormals. It's maybe simpler than one would fear.

krzysz00:
So, what I was referring to was the expansion patterns out it https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Arith/Transforms/ExpandOps.cpp , not the APFloat stuff . It looks like I was wrong and we don't actually have upstream patterns for fp8 ... but we should

compiler/src/iree/compiler/Codegen/Common/test/convert_unsupported_float_arith.mlir
hanhanW: I'm okay with either way as long as they are formated. :p

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -39,20 +760,32 @@ struct ConvertUnsupportedFloatArithPass final
...
 
+static void populateCPUSourceAndTargetType(MLIRContext *ctx, Operation *op,
+                                           SmallVectorImpl<Type> &sourceTypes,
+                                           Type &targetType) {
+  // For CPU backend, we emulate all the fp8 types to fp32. This supports any
+  // future new fp8 types automatically.
+  appendTypesIf<
+#define GET_TYPEDEF_LIST
+#include "mlir/IR/BuiltinTypes.cpp.inc"
+      >(ctx, sourceTypes);
```
hanhanW, line 732: We can also explicitly list out all the supported fp types.

hanhanW (commented):
The PR is ready for review. Below is a review note generated by claude:

```
This PR ports the fp8 conversion logic from runtime/src/iree/base/internal/math.h
to compiler patterns in ConvertUnsupportedFloatArithPass.cpp.

The compiler implementation follows the runtime exactly, with structural
differences due to SSA form:

1. **Control flow → Select cascades**: Runtime uses if/else with early returns;
   compiler computes all paths and uses arith.select to choose results.
   Select ordering matters: later selects override earlier ones, so we order
   from lowest to highest priority (NaN handling is always last).

2. **ldexpf() → Enumeration**: Runtime uses ldexpf() for denormals. Since MLIR
   arith dialect has no ldexp op and fp8 has only 3-7 possible denormal values,
   we precompute their f32 bit representations and use a select chain.
   (Also, this pass runs late in the pipeline where math dialect may not be allowed.)

3. **Rounding**: Matches runtime exactly:
   - Normal values: round-to-nearest-even (IEEE 754 default)
   - Subnormals: simple round-to-nearest (runtime's intentional choice)

Supported fp8 formats:
- FNUZ (f8E5M2FNUZ, f8E4M3FNUZ): NaN=0x80, no Inf, no negative zero
- IEEE (f8E5M2): Has Inf and negative zero
- FN (f8E4M3FN): No Inf, has negative zero

Key sections to review:
- Select cascade ordering (lines 458-518): Critical for FNUZ NaN correctness
- Round-to-nearest-even for normal values (lines 378-403)
- Denormal enumeration in ExtF (lines 689-717)
```

hanhanW:
> So, what I was referring to was the expansion patterns out it https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Arith/Transforms/ExpandOps.cpp , not the APFloat stuff . It looks like I was wrong and we don't actually have upstream patterns for fp8 ... but we should

I see. It's not an obvious place to find such patterns (to me). The upstream lowering is only for a few specific fp types, and they don't use generic logic. I think we prefer porting the runtime code to compiler patterns.

I believe that this should belong to IREE because as it is porting IREE's runtime code. It also has some IREE specific design idea (and I'm not sure if it still hold or not): IEEE 754 requires round-to-nearest-even for ALL values, including subnormals, but IREE runtime code uses simple round-to-nearest for subnormal:

https://github.com/iree-org/iree/blob/a872b71de271a4e8f894fc4fa63f60c034f59999/runtime/src/iree/base/internal/math.h#L400-L406

Furthermore, there are few options in software emulation:

- Enumerate values for ldexp.
- Use `math.powf`.
- Lookup-table.
- etc

I think we want better control, so leaving it in IREE sounds better to me.

krzysz00:
... I'm pretty sure those upstream patterns handle denormals correctly, so I wouldn't be worried about that.

I can agree that, given that we probably want these emulated operations to be out-of-line functions to save on code size and because we can generally implement extf more efficiently with a lookup table (which upstream can't quite do because adding constants to someone's context is non-trivial and would require communication to the user) these should be in IREE

However, if they are to be in IREE, they should replace the usage of the upstream expansion patterns in all cases, including for unsupported float types on GPUs and for fp4

hanhanW:
> ... I'm pretty sure those upstream patterns handle denormals correctly, so I wouldn't be worried about that.

I'm not saying that there are bugs in upstream patterns. I'm saying that the support is not generic and is not completed.

> I can agree that, given that we probably want these emulated operations to be out-of-line functions to save on code size and because we can generally implement extf more efficiently with a lookup table (which upstream can't quite do because adding constants to someone's context is non-trivial and would require communication to the user) these should be in IREE
> 
> However, if they are to be in IREE, they should replace the usage of the upstream expansion patterns in all cases, including for unsupported float types on GPUs and for fp4

These patterns are only used by CPU backends in the PR at the moment, but any backend can use them if they want. I don't follow how GPU handles the unsupported fp types, and people who work on that can pick this up. My main goal is filling the gap: IREE should be able to generate CPU code that is supported by other backends. IMO, CPU coverage should be a super set comparing to other backends.

bjacob:
Wow, thanks @hanhanW for acting on this so quick!  A couple notes but convergence is in sight.

> I believe that this should belong to IREE because as it is porting IREE's runtime code. It also has some IREE specific design idea (and I'm not sure if it still hold or not): IEEE 754 requires round-to-nearest-even for ALL values, including subnormals, but IREE runtime code uses simple round-to-nearest for subnormal

Thanks @hanhanW for the catch: this was not intentional! Sent a fix: https://github.com/iree-org/iree/pull/23134. Can you "rebase" your pattern on this ?

>  **ldexpf() → Enumeration**: Runtime uses ldexpf() for denormals. Since MLIR arith dialect has no ldexp op 

 All ldexpf does is multiply a float by a power of 2.  You can achieve the same with mulf.  Of course, this is less performant, but you don't care about performance here. This should result in simpler and more general code.

hanhanW:
> Wow, thanks @hanhanW for acting on this so quick! A couple notes but convergence is in sight.
> 
> > I believe that this should belong to IREE because as it is porting IREE's runtime code. It also has some IREE specific design idea (and I'm not sure if it still hold or not): IEEE 754 requires round-to-nearest-even for ALL values, including subnormals, but IREE runtime code uses simple round-to-nearest for subnormal
> 
> Thanks @hanhanW for the catch: this was not intentional! Sent a fix: #23134. Can you "rebase" your pattern on this ?
> 
> > **ldexpf() → Enumeration**: Runtime uses ldexpf() for denormals. Since MLIR arith dialect has no ldexp op
> 
> All ldexpf does is multiply a float by a power of 2. You can achieve the same with mulf. Of course, this is less performant, but you don't care about performance here. This should result in simpler and more general code.

All done, thanks!

bjacob:
Sounds great, please give me until tomorrow to review.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
 namespace {
 
+/// SFINAE (i.e., substitution failure is not an error) helper to detect if
+/// T::get(MLIRContext*) is valid.
+template <typename T, typename = void>
+struct HasContextGet : std::false_type {};
+template <typename T>
+struct HasContextGet<
+    T, std::void_t<decltype(T::get(std::declval<MLIRContext *>()))>>
+    : std::true_type {};
```
kuhar: use llvm::is_detected instead

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+
+/// Helpers to append types to a vector if they are f8 types.
+template <typename T>
+void maybeAppendType(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
```
kuhar:
```suggestion
static void maybeAppendType(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
```

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+  if constexpr (HasContextGet<T>::value) {
+    Type t = T::get(ctx);
+    if (isa<FloatType>(t) && t.getIntOrFloatBitWidth() == 8) {
+      types.push_back(t);
+    }
+  }
+}
+template <typename... Ts>
+void appendTypesIf(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
```
kuhar:
```suggestion
static void appendTypesIf(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
```

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+      auto splatAttr = SplatElementsAttr::get(cast<ShapedType>(f32Type), attr);
+      return rewriter.createOrFold<arith::ConstantOp>(loc, f32Type, splatAttr);
+    }
+    return rewriter.createOrFold<arith::ConstantOp>(loc, f32Type, attr);
+  }
+
+  /// Extracts sign, exponent, and mantissa from an f32 value (as i32 bits).
+  /// F32 format: 1 sign bit, 8 exponent bits, 23 mantissa bits.
+  /// Returns {sign, biasedExp, mantissa}.
+  std::tuple<Value, Value, Value> extractF32Fields(Value i32Val) {
```
kuhar: Can you use a struct instead and name the fields so that we don't have to guess which value corresponds to which component?

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+/// 1. FNUZ types (f8E5M2FNUZ, f8E4M3FNUZ): No Inf, no negative zero.
+///    - NaN is encoded as 0x80 (sign=1, exp=0, mantissa=0).
+///    - Overflow produces NaN; zero is always positive.
+///
+/// 2. IEEE types (f8E5M2): Has Inf and negative zero.
+///    - Standard IEEE-like encoding with Inf at max exponent.
+///
+/// 3. FN types (f8E4M3FN): No Inf, but has negative zero.
+///    - Max exponent values are valid finite numbers (except NaN encoding).
+struct TruncFToFP8 final : public OpRewritePattern<arith::TruncFOp> {
```
kuhar:
```suggestion
struct TruncFToFP8 final : OpRewritePattern<arith::TruncFOp> {
```
public is the default for structs

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+/// Emulates arith.extf from fp8 to f32 using integer bit manipulation.
+/// This implementation follows IREE's runtime/src/iree/base/internal/math.h.
+///
+/// For normal values: adjust exponent bias and shift mantissa.
+///
+/// For denormals (exp=0, mantissa!=0):
+///   value = mantissa * 2^(1 - bias - mantissa_bits)
+/// We implement this using uitofp + mulf with a precomputed scale factor,
+/// which is simpler and more general than enumerating all possible values.
+struct ExtFFromFP8 final : public OpRewritePattern<arith::ExtFOp> {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+constexpr int kF32MantBits = 23;
+constexpr int kF32Bias = 127;
+constexpr int kF32MantMask = (1 << kF32MantBits) - 1;
+constexpr int kF32MaxExp = 0xFF; // All exponent bits set (255).
+
+/// Parameters describing a floating-point format for conversion.
+/// Derived from APFloat semantics at runtime to support all MLIR float types.
+struct FloatTypeInfo {
+  unsigned expBits;  // Number of exponent bits.
+  unsigned mantBits; // Number of mantissa bits (excluding implicit leading 1).
```
bjacob: These two should be `int`, consistently with the above constants which are `int`.  Really avoid mixing integers of different signednesses, and prefer signed integers anyway unless representing bit-fields or specifically wanting modulo-2^32 arithmetic.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+  bool hasInf;       // Format supports infinity (exp=max, mantissa=0).
+  bool hasNan;       // Format supports NaN.
+  bool hasNegZero;   // Format supports negative zero.
+  bool nanAsNegZero; // NaN encoded as 0x80 (FNUZ types).
+  unsigned signMask; // Bit mask for sign bit.
+  unsigned expMask;  // Bit mask for exponent field.
+  unsigned mantMask; // Bit mask for mantissa field.
+  unsigned nanEncoding; // Bit pattern for canonical NaN.
+  unsigned infEncoding; // Bit pattern for +Inf (0 if no Inf).
+  unsigned maxFinite;   // Bit pattern for max finite value.
```
bjacob: Unsigned is good for bit-patterns, yes.  Reserving it for that usage makes intent more clear here.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+
+// F32 format constants (IEEE 754 binary32).
+constexpr int kF32MantBits = 23;
+constexpr int kF32Bias = 127;
+constexpr int kF32MantMask = (1 << kF32MantBits) - 1;
+constexpr int kF32MaxExp = 0xFF; // All exponent bits set (255).
+
+/// Parameters describing a floating-point format for conversion.
+/// Derived from APFloat semantics at runtime to support all MLIR float types.
+struct FloatTypeInfo {
```
bjacob: I believe that introducing this struct is unnecessary, being redundant with `fltSemantics`. See my comment below on getFloatTypeInfo.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+  unsigned signMask; // Bit mask for sign bit.
+  unsigned expMask;  // Bit mask for exponent field.
+  unsigned mantMask; // Bit mask for mantissa field.
+  unsigned nanEncoding; // Bit pattern for canonical NaN.
+  unsigned infEncoding; // Bit pattern for +Inf (0 if no Inf).
+  unsigned maxFinite;   // Bit pattern for max finite value.
+};
+
+/// Returns format info for a float type by querying APFloat semantics.
+static FloatTypeInfo getFloatTypeInfo(Type type) {
```
bjacob: The first thing that I notice about `getFloatTypeInfo` is that its `Type type` parameter is only used for one thing: to get the `fltSemantics`.  As far as I can see, from there, all the logic is based on `fltSemantics` only.  So at least, this function should be modified to take a `fltSemantics` argument instead of a `Type` argument.

But then, another thing becomes clear: this function and the FloatTypeInfo struct could be broken into separate functions each returning one value (one FloatTypeInfo member). Then the FloatTypeInfo struct can be removed.

For example:

```c++
int getMantissaBits(const  llvm::fltSemantics &sem) {
  return llvm::APFloat::semanticsPrecision(sem) - 1;
}
```

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+///   - Proper denormal/subnormal generation for underflow cases.
+///   - Correct handling of all fp8 format variants (IEEE, FN, FNUZ).
+///
+/// The conversion handles three categories of fp8 formats:
+///
+/// 1. FNUZ types (f8E5M2FNUZ, f8E4M3FNUZ): No Inf, no negative zero.
+///    - NaN is encoded as 0x80 (sign=1, exp=0, mantissa=0).
+///    - Overflow produces NaN; zero is always positive.
+///
+/// 2. IEEE types (f8E5M2): Has Inf and negative zero.
```
bjacob: I would call f8E5M2 "IEEE-like", but it's technically not an IEEE type.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+  Type i8Type;
+  Type f32Type;
+};
+
+//===----------------------------------------------------------------------===//
+// TruncF to small float emulation pattern
+//===----------------------------------------------------------------------===//
+
+/// Emulates arith.truncf from f32 to fp8 using integer bit manipulation.
+/// This implementation follows IREE's runtime/src/iree/base/internal/math.h.
```
bjacob: Maybe mention the specific functions there: `iree_math_truncate_f32_to_bits_rounding_to_nearest_even`.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+    // Check for NaN/Inf in source.
+    Value srcIsNanOrInf = arith::CmpIOp::create(
+        rewriter, loc, arith::CmpIPredicate::eq, f32Exp, cF32MaxExp);
+    Value srcIsNan = arith::AndIOp::create(
+        rewriter, loc, srcIsNanOrInf,
+        arith::CmpIOp::create(rewriter, loc, arith::CmpIPredicate::ne, f32Mant,
+                              c0));
+    Value srcIsInf = arith::AndIOp::create(
+        rewriter, loc, srcIsNanOrInf,
+        arith::CmpIOp::create(rewriter, loc, arith::CmpIPredicate::eq, f32Mant,
```
bjacob: Redundant CmpOp, maybe rather obtain srcIsNan from  srcIsInf .

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +36,632 @@ namespace mlir::iree_compiler {
...
+    return success();
+  }
+};
+
+//===----------------------------------------------------------------------===//
+// ExtF from small float emulation pattern
+//===----------------------------------------------------------------------===//
+
+/// Emulates arith.extf from fp8 to f32 using integer bit manipulation.
+/// This implementation follows IREE's runtime/src/iree/base/internal/math.h.
```
bjacob: Mention the specific function `iree_math_make_f32_from_bits`

hanhanW:
Review tips for @bjacob : I addressed the comments in two commits:

- Comments except `getFloatTypeInfo()` are addressed in https://github.com/iree-org/iree/pull/23119/changes/92a3ec6e1953c24495f9424e9f35d989dc5510f2
- Integrating float utils: https://github.com/iree-org/iree/pull/23119/changes/a3b1ccaea28a1af0770c993775d15b024bbc7c8a

Summary from Claude about the status of each comment:

```md
From kuhar:
  1. ✓ Use llvm::is_detected instead of custom HasContextGet - Done
  2. ✓ Use struct for extractF32Fields return (F32Fields) - Done
  3. ✓ Use struct for patterns (already were structs) - Done

From bjacob:
  1. ✓ Change expBits/mantBits to int - Done (now int in the helper's private members)
  2. ✓ Call f8E5M2 "IEEE-like" not "IEEE" - Done (line 351)
  3. ✓ Mention iree_math_truncate_f32_to_bits_rounding_to_nearest_even - Done (line 338)
  4. ✓ Remove redundant CmpOp, derive srcIsNan from srcIsInf - Done (using XOR)
  5. ✓ Mention iree_math_make_f32_from_bits - Done (line 571)
  6. ⏭ getFloatTypeInfo refactoring - User said they'd handle this separately, but we ended up removing it entirely by integrating into the helper

Additional improvements we made:
  - Removed FloatTypeInfo struct and getFloatTypeInfo() function entirely
  - Integrated float type queries into FloatEmulationHelper
  - Helper methods now return Value constants directly
  - Simplified constructor API to just take smallFloatType
```

EDIT: update commit hash after rebase.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
kuhar: not done

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
kuhar: not done

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -97,16 +800,31 @@ void ConvertUnsupportedFloatArithPass::runOnOperation() {
...
+  // Apply the standard float emulation patterns. This inserts extf/truncf pairs
+  // around unsupported float operations.
+  {
+    TypeConverter converter;
+    arith::populateEmulateUnsupportedFloatsConversions(converter, sourceTypes,
+                                                       targetType);
+    RewritePatternSet patterns(context);
+    arith::populateEmulateUnsupportedFloatsPatterns(patterns, converter);
+    ConversionTarget target(*context);
+    arith::populateEmulateUnsupportedFloatsLegality(target, converter);
```
kuhar, line 812: Is it valid to mix conversion and rewrite patterns? I remember it used to not always work a few years ago, but maybe Matthias fixed this

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -97,16 +800,31 @@ void ConvertUnsupportedFloatArithPass::runOnOperation() {
...
 
-  if (failed(applyPartialConversion(funcOp, target, std::move(patterns)))) {
-    signalPassFailure();
+  // For CPU, emulate extf/truncf to/from small float types using integer ops.
+  // This is needed because LLVM doesn't support fpext/fptrunc for fp8 types;
+  // the fp8 types eventually get lowered to i8 in LLVM IR.
+  if (isCPU) {
+    RewritePatternSet emulationPatterns(context);
+    emulationPatterns.add<TruncFToFP8, ExtFFromFP8>(context);
+    if (failed(applyPatternsGreedily(funcOp, std::move(emulationPatterns)))) {
```
kuhar: I think you can use the walk pattern rewrite driver if you don't have to rewrite newly created ops

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
hanhanW: It should be valid. I'm not aware of bugs, but I did see it working since a few years ago.

bjacob (approved):
Looks good, thanks!

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
hanhanW: Done. The checks are also updated because constants are created inline when needed rather than being hoisted to the beginning of the function if we use this rewrite.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -39,20 +721,32 @@ struct ConvertUnsupportedFloatArithPass final
 
 } // namespace
 
+static void populateCPUSourceAndTargetType(MLIRContext *ctx, Operation *op,
+                                           SmallVectorImpl<Type> &sourceTypes,
+                                           Type &targetType) {
+  // For CPU backend, we emulate all the fp8 types to fp32. This supports any
+  // future new fp8 types automatically.
+  appendTypesIf<
```
krzysz00, line 729: Nit: this should have a less generic name

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -97,16 +800,29 @@ void ConvertUnsupportedFloatArithPass::runOnOperation() {
...
+      return signalPassFailure();
+    }
+  }
 
-  if (failed(applyPartialConversion(funcOp, target, std::move(patterns)))) {
-    signalPassFailure();
+  // For CPU, emulate extf/truncf to/from small float types using integer ops.
+  // This is needed because LLVM doesn't support fpext/fptrunc for fp8 types;
+  // the fp8 types eventually get lowered to i8 in LLVM IR.
+  if (isCPU) {
```
krzysz00, line 822: Why is this constrained to CPU?

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +35,683 @@ namespace mlir::iree_compiler {
...
+  //===--------------------------------------------------------------------===//
+
+  /// Creates an i32 constant, splatted if working with vectors.
+  Value createI32Const(int64_t value) {
+    auto attr = rewriter.getIntegerAttr(rewriter.getI32Type(), value);
+    if (vecType) {
+      auto splatAttr = SplatElementsAttr::get(cast<ShapedType>(i32Type), attr);
+      return rewriter.createOrFold<arith::ConstantOp>(loc, i32Type, splatAttr);
+    }
+    return rewriter.createOrFold<arith::ConstantOp>(loc, i32Type, attr);
```
kuhar, line 139: Do we ever fold constants? What would these fold to?

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -30,6 +35,683 @@ namespace mlir::iree_compiler {
...
+  }
+
+  /// Creates an f32 constant, splatted if working with vectors.
+  Value createF32Const(float value) {
+    auto attr = rewriter.getF32FloatAttr(value);
+    if (vecType) {
+      auto splatAttr = SplatElementsAttr::get(cast<ShapedType>(f32Type), attr);
+      return rewriter.createOrFold<arith::ConstantOp>(loc, f32Type, splatAttr);
+    }
+    return rewriter.createOrFold<arith::ConstantOp>(loc, f32Type, attr);
```
kuhar, line 149: Also here

kuhar (approved):

krzysz00 (approved):
Overall, this looks good to me. However, I've got some thoughts for followup PRs

1. Since these truncations and extensions are non-trivial code blocks, would it make sense to have these generate functions instead and then call those functions as the rewrite pattern? Or at least to have the option?
2. Similarly, extf on fp8 and fp4 could just be generating a lookup table and then indexing into it - can we explore that? 
3. This should be enabled for GPU so that fp8 types that don't have native support at least kinda run

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
hanhanW: IIRC, it reuses the constants during op creation. E.g., if you use this rewriter to create two const_0, only one will be generated.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
hanhanW: Do you have any suggestion? To me, it is good enough because it is a local function to the pass..

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
hanhanW: Because I started the work for CPU; that's how I scope my work. I can definitely help other backends if someone is interested in it. It also means that it is other's responsibility to evaluate the option for such backend, because my scope (for now) is just for CPU backend.

hanhanW:
> Overall, this looks good to me. However, I've got some thoughts for followup PRs
> 
> 1. Since these truncations and extensions are non-trivial code blocks, would it make sense to have these generate functions instead and then call those functions as the rewrite pattern? Or at least to have the option?

I already created a class helper to reduce non-trivial code blocks. If you have anything concrete in your mind, I'm happy to make changes. I think passing those values would be another pain point as you could make too many fragment functions; the logic may become not straight-forward.

> 2. Similarly, extf on fp8 and fp4 could just be generating a lookup table and then indexing into it - can we explore that?

Wo don't care about performance here, so I don't see a point to explore that atm. The other point is that making it generic and be aligned with runtime code looks better to me.

> 3. This should be enabled for GPU so that fp8 types that don't have native support at least kinda run

I can help when I get cycles; I agree that we should support it for other backends. I'm starting with CPU because the functionality coverage of CPU should be a supper set like I mentioned in other comments.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
krzysz00: Yeah, this is the fold of "don't have 20 different copies of 0 running around"

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
krzysz00: I'd figure something like "appendSmallFloatTypes` or `appendCpuEmulatedTypes` or something - I'm mainly worried about the risk of someone grepping for this name

PR #23122, opened 2026-01-14
Adding libFuzzer-based fuzzing infrastructure.

benvanik:
This PR adds support for coverage-guided fuzz testing using libFuzzer. The infrastructure integrates with both Bazel and CMake build systems, automatically enabling AddressSanitizer for memory error detection.

New fuzz targets are defined using `iree_runtime_cc_fuzz` (or `iree_compiler_cc_fuzz` for compiler code). These targets implement the standard `LLVMFuzzerTestOneInput` entry point and are excluded from the default build to avoid slowing down normal development.

Two initial fuzz targets are included as proof-of-life: `unicode_fuzz` for UTF-8 and Unicode utilities, and `string_view_fuzz` for string parsing functions. Running these for just a few seconds immediately found two bugs in string_view.c: an out-of-bounds read in `iree_string_view_find_last_of` when given a position past the string length, and an exponential backtracking issue in
`iree_string_view_match_pattern` with pathological wildcard patterns like `*?*?*?*?*`. Both are fixed in this PR.

See `docs/website/docs/developers/debugging/fuzzing.md` for full documentation.

Bazel example:
```shell
bazel build --config=fuzzer //runtime/src/iree/base/internal:unicode_fuzz
./bazel-bin/runtime/src/iree/base/internal/unicode_fuzz -max_total_time=60
```

CMake example:
```shell
cmake -B build -DIREE_ENABLE_FUZZING=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo
cmake --build build --target unicode_fuzz
./build/runtime/src/iree/base/internal/unicode_fuzz -max_total_time=60
```

In the future supporting AFL is easy - for now, this with a to-be-landed iree-bazel-fuzz are effective enough.


ScottTodd (approved):
Clean implementation with docs and tests. Nice.

PR #23123, opened 2026-01-14
Adding iree-bazel-* wrapper tools for improved Bazel workflows.

benvanik:
Introduces a suite of wrapper tools in build_tools/bin/ that provide a
unified interface for IREE's Bazel builds:

- iree-bazel-configure: Platform detection and HAL driver configuration
  with environment variable support matching CMake conventions
- iree-bazel-build|test|run: Build, test, and run wrappers with watch
  mode support (-w flag) for rapid iteration
- iree-bazel-try: Quick compilation and execution of C/C++ snippets
  against IREE runtime with automatic dependency inference
- iree-bazel-query|fuzz: Query and fuzzing tool wrappers
- iree-bazel-lib: Shared library infrastructure for all tools

Key improvements:
- Driver configuration via environment variables (IREE_HAL_DRIVER_CUDA=ON)
- Watch mode for continuous rebuilds during development
- Simplified workflow: configure once, then build/test/run
- Automatic dependency resolution for experimentation with iree-bazel-try
- Consistent command-line interface across all tools
- --agent_md flag for easily adding to CLAUDE.local.md/etc

To support iree-bazel-try this commit adds three new harnesses for rapid
development:
1. gtest_harness.h - Google Test with IREE status matchers
2. gbenchmark_harness.h - Google Benchmark with IREE base types
3. MlirTransformHarness.h - MLIR transformation experiments w/ I/O

The MLIR harness uses standard MLIR patterns (mlir::openInputFile,
parseSourceFile) for transparent .mlir/.mlirbc support, llvm::cl for flag
parsing, and supports pass pipelines before/after transforms via
--pass-pipeline-before/--pass-pipeline-after flags.

Required some updates to configure_bazel.py to read HAL driver settings
from environment variables and write --iree_drivers flag to
configured.bazelrc, matching CMake's driver selection logic. In the
future I'll be cleaning up the --iree_drivers flag and adding some more
flags so that we have a greater equivalence between CMake and Bazel for
selecting feature subsets.  A 'localdev' bazel config for faster local iteration
was added to be used by the tools (or --config=localdev) that consistently
handles:
- Persistent disk cache to speed up clean builds
- Skymeld for better parallelism
- Ramdisk sandbox to reduce I/O bottleneck Support for 'mold' config
  required routing a CI flag around it.

Website documentation is updated to recommend bazelisk for version
management (as our CI uses) and documents the new tool-based workflow with
quickstart examples. Raw bazel instructions are left in as an advanced
section.

### DIFF
--- a/build_tools/bin/iree-bazel-build
+++ b/build_tools/bin/iree-bazel-build
@@ -1,0 +1,167 @@
+#!/bin/bash
+# Build IREE targets with Bazel.
+#
+# Usage: iree-bazel-build [options] [target] [bazel-args...]
+#
+# Examples:
+#   iree-bazel-build                      # Build all tools
+#   iree-bazel-build //compiler/...       # Build compiler
+#   iree-bazel-build --config=debug       # Debug build
+
+set -e
+
+# Source shared library.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/iree-bazel-lib"
+iree_bazel_init "iree-bazel-build"
+
+# Expand combined short flags (e.g., -nv -> -n -v).
+eval "set -- $(iree_expand_combined_flags "$@")"
+
+show_help() {
+    cat << 'EOF'
+iree-bazel-build - Build IREE with Bazel
+
+USAGE
+    iree-bazel-build [options] [target] [bazel-args...]
+
+OPTIONS
+    -n, --dry_run    Show the bazel command without executing
+    -v, --verbose    Show the bazel command before executing
+    -k, --keep_going Continue building after errors
+    -w, --watch      Watch mode: rebuild on file changes
+    -h, --help       Show this help
+
+    NOTE: Short flags can be combined: -nv is equivalent to -n -v
+
+ARGUMENTS
+    target       Bazel target (required)
+    bazel-args   Additional arguments passed to bazel build
+
+EXAMPLES
+    # Build the compiler tool
+    iree-bazel-build //tools:iree-compile
+
+    # Build multiple tools at once
+    iree-bazel-build //tools:iree-compile //tools:iree-opt
+
+    # Build the runtime module executor
+    iree-bazel-build //tools:iree-run-module
+
+    # Build all compiler components
+    iree-bazel-build //compiler/...
+
+    # Debug build
+    iree-bazel-build //tools:iree-compile --config=debug
+
+    # Continue building after errors
+    iree-bazel-build -k //compiler/...
+
+    # Watch and rebuild on changes
+    iree-bazel-build -w //tools:iree-opt
+
+    # Show command without executing (verbose dry-run)
+    iree-bazel-build -nv //tools:iree-compile
+
+COMMON TARGETS
+    //tools/...              All tools
+    //compiler/...           Compiler components
+    //runtime/...            Runtime components
+    //tools:iree-compile     Main compiler tool
+    //tools:iree-run-module  Module runner
+
+CONFIGURATIONS
+    --config=localdev        Local dev optimizations (disk cache, skymeld)
+    --config=debug           No optimizations, assertions enabled
+    --config=asserts         Optimized with assertions
+    --config=asan            Address Sanitizer
+    --config=msan            Memory Sanitizer
+    --config=tsan            Thread Sanitizer
+
+SEE ALSO
+    iree-bazel-configure, iree-bazel-test
+EOF
+}
+
+# Parse arguments.
+KEEP_GOING=0
+WATCH_MODE=0
+TARGET=""
+BAZEL_ARGS=()
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -h|--help)
+            show_help
+            exit 0
+            ;;
+        --agent-md|--agent_md)
+            iree_show_agent_md
+            exit 0
+            ;;
+        -n|--dry_run|--dry-run)
+            IREE_BAZEL_DRY_RUN=1
+            shift
+            ;;
+        -v|--verbose)
+            IREE_BAZEL_VERBOSE=1
+            shift
+            ;;
+        -k|--keep_going|--keep-going)
+            KEEP_GOING=1
+            shift
+            ;;
+        -w|--watch)
+            WATCH_MODE=1
+            shift
+            ;;
+        -*)
+            BAZEL_ARGS+=("$1")
+            shift
+            ;;
+        *)
+            if [[ -z "$TARGET" ]]; then
+                TARGET="$1"
+            else
+                BAZEL_ARGS+=("$1")
+            fi
+            shift
+            ;;
+    esac
+done
+
+# Set up worktree (after arg parsing so --help works anywhere).
+iree_setup_worktree
+
+# Target is required.
+if [[ -z "$TARGET" ]]; then
+    iree_error "Target is required for bazel build"
+    echo ""
+    echo "Examples:"
+    echo "  iree-bazel-build //tools:iree-compile"
+    echo "  iree-bazel-build //tools:iree-compile //tools:iree-opt"
+    echo "  iree-bazel-build //tools:iree-run-module"
+    echo "  iree-bazel-build //compiler/..."
+    exit 1
+fi
+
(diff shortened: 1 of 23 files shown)


docs/website/docs/developers/building/bazel.md
```diff
@@ -210,7 +232,13 @@ by git.
 === ":fontawesome-brands-windows: Windows"
 
     ```shell
-    build --disk_cache=c:/bazelcache
+    # Always use localdev optimizations
+    build --config=localdev
+
+    # Override disk cache path to avoid Windows path length issues.
+    # The default ~/.cache/bazel-iree can exceed the 260 character limit.
+    build:localdev --disk_cache=c:/bazelcache
```
ScottTodd: On Windows 11 we could also recommend using a Dev Drive: https://learn.microsoft.com/en-us/windows/dev-drive/ (I haven't tested it with Bazel, but for git operations and CMake builds it has been a nice improvement for me)

compiler/src/iree/compiler/Tools/MlirTransformHarness.h
```diff
@@ -0,0 +1,350 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Harness for one-shot MLIR transformations via iree-bazel-try.
+// Uses standard MLIR tool patterns for I/O, flag parsing, and pass pipelines.
```
ScottTodd, line 8: > These harnesses integrate with the iree-bazel-try tool (coming soon) for zero-boilerplate experimentation, but are useful on their own, too.

Do you want to lead with the `iree-bazel-try` tool first, instead of this PR? It feels weird referencing a tool that doesn't exist yet. Is this code used for anything else?

I'm also wondering what about this is bazel-specific. Is that try command invoking a build, or just running a tool? Would a CMake equivalent be possible?

(happy to review the code based on the future plans, just watching out for dead code that could go stale)

compiler/src/iree/compiler/Tools/MlirTransformHarness.h
```diff
@@ -0,0 +1,350 @@
...
+
+  // Parse command line options.
+  llvm::cl::ParseCommandLineOptions(
+      argc, argv,
+      "IREE Transform Harness - MLIR transformation experimentation tool\n");
+
+  // Register all IREE dialects and passes (same as iree-opt).
+  mlir::DialectRegistry registry;
+  mlir::iree_compiler::registerAllDialects(registry);
+  mlir::iree_compiler::registerAllPasses();
```
ScottTodd, line 167: Looks like everything in this file is agnostic to the IREE compiler except for these registrations.

I wonder if there is appetite for this sort of tool in upstream LLVM/MLIR. Fine to keep it here in IREE and just consider that later, or let someone else fork it and upstream it.

benvanik:
Working on the iree-bazel-* tools - will add them to this PR as another commit - wanted to get your thoughts on this part before I cleaned them up, so thank you! <3

compiler/src/iree/compiler/Tools/MlirTransformHarness.h
benvanik: Added iree-bazel-try and the other iree-bazel-* tools. It does a cquery to find required dependencies, generates a bazel file, writes out the file, runs a bazel build, and then executes the result. It's possible to have a cmake try, but it'd be way dumber and slower unless you're just linking against a unified static/shared library bundle of the project (I originally started with iree-cmake-try, then gave up because agents kept using stale libraries x_x).

docs/website/docs/developers/building/bazel.md
benvanik: Ah nice - I moved this into iree-bazel-configure so we can do things like detect those cases. I'll let a Windows dev give it a shot (which may be future me whenever I setup a Windows VM again!)

compiler/src/iree/compiler/Tools/MlirTransformHarness.h
benvanik: Could be, though it's hard because everything has to be registered so it's probably easiest to just clone the idea. This is most useful paired with agents and the iree-bazel-try wrapper, so keeping it here.

benvanik:
Ok, iree-bazel-* tools added so the whole end-to-end is here!

docs/website/docs/developers/building/bazel.md
ScottTodd, line 1: FYI @maxbartel I think you guys were active users of the Bazel build system when last I checked. This PR is adding some new tooling, so you may want to review this too.

configure_bazel.py
```diff
@@ -62,11 +62,70 @@ def write_platform(bazelrc):
...
+def cmake_bool_is_true(value):
+    """Check if a CMake-style bool value is true."""
+    if not value:
+        return False
+    return value.upper() in ("ON", "YES", "TRUE", "Y", "1")
+
+
+def get_hal_driver_defaults():
+    """Get HAL driver defaults matching CMake logic (CMakeLists.txt:275-318)."""
+    defaults_enabled = True  # IREE_HAL_DRIVER_DEFAULTS
```
ScottTodd: Rather than reference lines in a source file, can this reference some text, like a specific identifier/variable name? The lines in the source file will change over time.

build_tools/bin/iree-bazel-build
ScottTodd, line 1: Can you add a README.md to this folder? Could link to the bazel doc

build_tools/bin/iree-bazel-try
```diff
@@ -0,0 +1,1017 @@
...
+            sort -n | head -n "$excess" | cut -d' ' -f2- | \
+            while read -r old_slot; do
+                rm -rf "$old_slot"
+                iree_debug "Evicted cache slot: $old_slot"
+            done
+    fi
+}
+
+# Ensure .iree-bazel-try is in .git/info/exclude (local exclude, not .gitignore).
+ensure_ignores() {
```
ScottTodd: Why not add to the shared .gitignore? If these tools are going in the shared repository I think it makes sense to put files they generate in the shared .gitignore too, then remove this extra code that modifies what is normally a developer-authored (and not tool-authored) file.

build_tools/bin/iree-bazel-try
```diff
@@ -0,0 +1,1017 @@
...
+
+    # Look for undefined reference errors.
+    while IFS= read -r symbol; do
+        if [[ -n "$symbol" ]]; then
+            local dep
+            if dep=$(find_dep_for_symbol "$symbol" "$worktree" 2>/dev/null); then
+                if [[ -z "${suggested_deps[$dep]:-}" ]]; then
+                    suggested_deps[$dep]=1
+                    echo "$dep"
+                fi
```
ScottTodd, line 437: style nit: I would prefer brace-delimiting all variables, per the recommendations at https://google.github.io/styleguide/shellguide.html#variable-expansion

It might also be worth running these the shell scripts through shellcheck or even adding a shellcheck (or another tool) pre-commit hook: https://github.com/koalaman/shellcheck-precommit

In general I'd recommend moving over to Python after a shell script reaches a certain complexity level, but these scripts are pretty self-contained and consistently structured, so I won't push that too hard here.

ScottTodd (approved):

build_tools/bin/iree-bazel-configure
```diff
@@ -0,0 +1,129 @@
...
+        # Unix: Use XDG cache directory.
+        CACHE_DIR="$HOME/.cache/bazel-iree"
+    fi
+
+    cat > user.bazelrc << EOF
+# Local Bazel configuration.
+# This file is not version-controlled.
+
+# Disk cache for faster rebuilds (override with IREE_BAZEL_CACHE_DIR).
+build --disk_cache=$CACHE_DIR
```
maxbartel: I would also add garbage collection for the disk cache https://bazel.build/remote/caching#disk-cache-gc We had instances where rebuilding while trying out options filled the whole disk in a matter of hours...

maxbartel (approved):
LGTM, considering they don't seem to be a requirement, but rather wrappers that add functionality.

PR #23124, opened 2026-01-14
[LinalgExt] Add map_gather e2e tests for CPU/VMVX/ROCM

Abhishek-Varma:
It attaches PartitionableLoops interfaces to the op and adds e2e tests for map_gather op on CPU/VMVX/ROCM.
No additional lit tests are added because they are templated and identical to other LinalgExt ops. Adding lit tests do not provide additional value because all the code paths are covered via other LinalgExt op tests.


hanhanW (approved):
Nice!

PR #23125, opened 2026-01-14
[Encoding] Propagate (Un)SetEncodingOp with dynamic encoding dims

jtuyls:
This PR adds support for propagating dynamic encoding_dims through encoding operations during dispatch creation:
- **`EncodingTypes.cpp`**: Extract dynamic dimension values (Batch, M, N, K) from contractions, only for dimensions that are dynamic in iteration_sizes
-  **`Utils.cpp`**: Add rematerializeEncodingDims helper to recreate encoding dims at new insertion points during propagation (handles dominance, tensor.dim remapping, pure op cloning)
-  **`HoistEncodingOps.cpp`**: Rematerialize encoding dims when bubbling up set_encoding through generic ops
-  **`EncodingExternalModels.cpp`**: Rematerialize encoding dims when propagating through tensor.cast and similar ops

**Note:** The previous version was closed automatically as I hadn't stacked it correctly: https://github.com/iree-org/iree/pull/23075

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
@@ -10,7 +10,10 @@
 #include "llvm/ADT/SmallVector.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/AffineMap.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinAttributes.h"
@@ -49,6 +52,23 @@ static Type getContractionInputTypeWithSignedness(OpBuilder &builder,
   return elemType;
 }
 
+/// Extract dynamic dimension values for a linalg operation in loop order.
+/// Returns values only for dimensions that are dynamic.
+static SmallVector<Value> getDynamicLoopDims(OpBuilder &builder,
+                                             linalg::LinalgOp linalgOp) {
+  SmallVector<Range> loopRanges =
+      linalgOp.createLoopRanges(builder, linalgOp.getLoc());
+
+  SmallVector<Value> dynamicDims;
+  for (Range &range : loopRanges) {
+    // OpFoldResult is either Value (dynamic) or Attribute (static).
+    if (auto value = dyn_cast<Value>(range.size)) {
+      dynamicDims.push_back(value);
+    }
+  }
+  return dynamicDims;
+}
+
 FailureOr<OpEncodingProperties>
 SerializableAttr::getEncodingProperties(Operation *op) {
   auto linalgOp = dyn_cast<linalg::LinalgOp>(op);
@@ -57,7 +77,7 @@ SerializableAttr::getEncodingProperties(Operation *op) {
   }
 
   MLIRContext *ctx = op->getContext();
-  OpBuilder builder(ctx);
+  OpBuilder builder(op);
 
   OpEncodingProperties props;
   SmallVector<Type> elemTypes;
@@ -65,11 +85,14 @@ SerializableAttr::getEncodingProperties(Operation *op) {
   SmallVector<int64_t> iterationSizes = linalgOp.getStaticLoopRanges();
   EncodingOpType opType;
 
+  // Extract dynamic loop dimensions.
+  SmallVector<Value> dynamicDims = getDynamicLoopDims(builder, linalgOp);
+
   auto addEncoding = [&](int64_t operandIndex) {
     return EncodingProperties{EncodingAttr::get(ctx, operandIndex, opType,
                                                 elemTypes, maps,
                                                 iterationSizes),
-                              /*dynamicValues=*/{}};
+                              dynamicDims};
   };
 
   // Return encoding properties for contraction operations.
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
@@ -67,14 +67,22 @@ MatmulNarrowDim getMatmulNarrowDim(linalg::LinalgOp linalgOp,
 /// Bundles an encoding attribute with its associated dynamic information.
 /// This is returned by interface methods that query encoding properties
 /// from operations. Dynamic values can include runtime information needed
-/// by the encoding (e.g., M, N, K dimensions for matmul encodings).
+/// by the encoding (e.g., Batch, M, N, K dimensions for matmul encodings).
 struct EncodingProperties {
   /// The encoding attribute for the operand/result.
   Attribute encoding;
-  /// Dynamic values needed by the encoding. For matmul-like operations,
-  /// these typically correspond to the iteration domain dimensions (M, N, K).
-  /// TODO(#22370): Currently empty; will be populated to support dynamic
-  /// layout decisions.
+  /// Optional dynamic values needed by the encoding. This field may be empty
+  /// for operations that don't require dynamic dimension tracking.
+  ///
+  /// For operations implementing IndexingMapOpInterface (e.g., linalg ops),
+  /// these correspond to the dynamic entries in iteration_sizes, ordered by
+  /// loop dimension index. For example:
+  /// - matmul with iteration_sizes=[?, ?, ?] -> [M, N, K]
+  /// - batch_matmul with iteration_sizes=[?, ?, ?, ?] -> [Batch, M, N, K]
+  /// - matmul with iteration_sizes=[?, 128, ?] -> [M, K] (N is static)
+  ///
+  /// For operations that do not implement IndexingMapOpInterface, the meaning
+  /// and ordering of these values is defined by downstream users.
   SmallVector<Value> dynamicValues;
 };
 
@@ -89,6 +97,10 @@ struct OpEncodingProperties {
 struct PropagationEncoding {
   SmallVector<Attribute> operandEncodings;
   SmallVector<Attribute> resultEncodings;
+  /// Optional dynamic encoding dimension values that should be carried through
+  /// during propagation. See EncodingProperties::dynamicValues for details on
+  /// the ordering and semantics of these values.
+  SmallVector<Value> encodingDims;
 };
 
 struct PropagationResult {
--- a/compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
@@ -6,11 +6,18 @@
 
 #include "iree/compiler/Dialect/Encoding/Utils/Utils.h"
 
+#include <variant>
+
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/DebugLog.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/IR/AffineMap.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinTypeInterfaces.h"
+#include "mlir/IR/Dominance.h"
+
+#define DEBUG_TYPE "iree-encoding-utils"
 
 namespace mlir::iree_compiler::IREE::Encoding {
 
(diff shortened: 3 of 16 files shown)


hanhanW:
Sorry, I forgot to review this PR. I'll take a look tomorrow.

jtuyls:
Adding back the review comments from @MaheshRavishankar 

> I have concerns about the recursive rematerialization. It can be pretty expensive. I think I need a little bit more context on this to understand the cost of rematerialization recursively. The patterns I am matching these back to typically use a fixed point/listeners to handle the cost getting out of hand. Can we discuss this offline first before I get back to reviewing this.

As per offline discussion, the recursive function will return early in case an intermediate value with multiple uses in the use-def chain is hit multiple, so this should only rematerialize every value in the chain only once.

> Not a blocking comment, but I think we have tried many times doing something similar. @IanWood1 do you have pointers to something similar. Can we make this a bit more re-usable utility (with potentially callbacks) to not be specific to the rematerializaiton of dims this way?

Happy to use an existing utility if that exists or generalize this one to take in a callback for the special `tensor.dim` rematerialization case.

Also, not exactly the same, but I have tried using the upstream `moveValueDefinitions` function (updated with cross-region support: https://github.com/llvm/llvm-project/pull/176343) here as well, but that only works if we don't have intermediate `tensor.dim` ops depending of the source of the propagation that need to be rematerialized before it:
```
%cast = tensor.cast %src : tensor<1024x?xf16> to tensor<?x512xf16>
// This dim depends on the cast result (propagation source).
// It gets rematerialized to use the new source (src).
%m = tensor.dim %cast, %c0 : tensor<?x512xf16>
%c512 = arith.constant 512 : index
%0 = iree_encoding.set_encoding %cast encoding_dims{%m, %c512} : tensor<?x512xf16> -> tensor<?x512xf16, #encoding_dim_on_source>
util.return %0 : tensor<?x512xf16, #encoding_dim_on_source>
```
If that cross-region support is upstreamed, we could use it here as well though to first try moving the encoding dims and falling back to rematerialization if not possible.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
hanhanW, line 1: The `getContractionDynamicDims` method does not match loop dimension ordering because sorting is not involved.

I think what you're looking for is calling [LinalgOp::createLoopRanges](https://github.com/llvm/llvm-project/blob/fae5760fd1ae807a4b480bc0de47c21102146dd5/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp#L1134-L1149) and filter out the dynamic (i.e., Value) dimension. All the fields in `Range` have `OpFoldResult` type.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
```diff
@@ -67,14 +67,19 @@ MatmulNarrowDim getMatmulNarrowDim(linalg::LinalgOp linalgOp,
...
-  /// TODO(#22370): Currently empty; will be populated to support dynamic
-  /// layout decisions.
+  /// these correspond to the dynamic entries in iteration_sizes, ordered
+  /// by loop dimension index. For example:
+  /// - matmul with iteration_sizes=[?, ?, ?] -> [M, N, K]
+  /// - batch_matmul with iteration_sizes=[?, ?, ?, ?] -> [Batch, M, N, K]
+  /// - matmul with iteration_sizes=[?, 128, ?] -> [M, K] (N is static)
+  ///
+  /// The order matches the loop dimensions as they appear in the operation's
+  /// indexing maps, filtered to only include dynamic dimensions.
```
hanhanW: Let's expand this a bit. This is the case when an op implements IndexingMapOpInterface. If the target operation does not implement the interface, it is a custom field defined by downstream users.

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
```diff
@@ -7,10 +7,15 @@
 #include "iree/compiler/Dialect/Encoding/Utils/Utils.h"
 
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Debug.h"
```
hanhanW: Include `DebugLog.h` and use `DBGS()`.

compiler/src/iree/compiler/DispatchCreation/HoistEncodingOps.cpp
```diff
@@ -96,6 +97,21 @@ bubbleUpSetEncodingThroughGenericOp(RewriterBase &rewriter,
...
+  Value genericOutput = encodingOp.getSource();
+  Value firstInput = genericOp.getDpsInputOperand(0)->get();
+  FailureOr<SmallVector<Value>> rematerializedDims =
+      IREE::Encoding::rematerializeEncodingDims(rewriter, genericOp,
+                                                encodingOp.getEncodingDims(),
+                                                genericOutput, firstInput);
+  if (failed(rematerializedDims)) {
+    return rewriter.notifyMatchFailure(encodingOp,
+                                       "cannot rematerialize encoding_dims");
+  }
```
hanhanW, line 112: This can be problematic because we could run into the case that we create operations but return a failure. We should avoid creating operations on failure when we do pattern-rewrite. (I checked that it is called by a OpPattern.)

I think we can break the method into two methods:
- Fetch all the ops that need the `clone`, so it returns a list of `(Operation*, InsertionPoint)`; it returns `std::nullopt` on failure.
- Apply the clone. Either you implement it in a method, or you just loop over the list and clone the operation here.

hanhanW (changes requested):

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
hanhanW: We also need to document that this is an optional field, because scaled_matmul does not use the field.

jtuyls: > Let's expand this a bit. This is the case when an op implements IndexingMapOpInterface. If the target operation does not implement the interface, it is a custom field defined by downstream users.

I updated the doc for this.

> because scaled_matmul does not use the field.

Not sure what you mean here, scaled_matmul should also use the field? I added a test for that as well.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
jtuyls: Thanks, that's better!

compiler/src/iree/compiler/DispatchCreation/HoistEncodingOps.cpp
jtuyls: Good point! I updated the method to have two stages now. Everything is still in a single function, but it will return early now without creating any new ops in case of failure.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
hanhanW: I misread the code, I thought that it's only used by matmul but not scaled matmul. Sorry about that.

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
```diff
@@ -140,4 +145,206 @@ bool isNarrowNResult(EncodingAttr encoding) {
...
+
+  Kind kind;
+  Value originalValue;
+  // For CreateDimOp: location and index for the new tensor.dim
+  LocationAttr dimLoc;
+  Value dimIndex;
+  // For CloneOp: the operation to clone and which result index to use
+  Operation *opToClone = nullptr;
+  unsigned resultIndex = 0;
+};
```
hanhanW: I dont like that there are two modes in a struct, so I asked LLM to provide a suggestion. Below is the response, and I like the using `std::variant` idea.

---

The issue is this is a "tagged union" implemented as a flat struct with unused fields depending on Kind. This wastes memory and lacks type safety.

Suggested improvement using `std::variant`:

```cpp
struct UseExistingAction {
  Value originalValue;
};

struct CreateDimOpAction {
  Value originalValue;
  LocationAttr dimLoc;
  Value dimIndex;
};

struct CloneOpAction {
  Value originalValue;
  Operation *opToClone;
  unsigned resultIndex;
};

using RematerializationAction =
    std::variant<UseExistingAction, CreateDimOpAction, CloneOpAction>;
```

With std::variant, the plan construction looks better:

```cpp
  // UseExisting
  plan.push_back(UseExistingAction{value});

  // CreateDimOp
  plan.push_back(CreateDimOpAction{value, dimOp.getLoc(), dimOp.getIndex()});

  // CloneOp
  plan.push_back(CloneOpAction{value, definingOp, resultIndex});
```

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
```diff
@@ -140,4 +145,206 @@ bool isNarrowNResult(EncodingAttr encoding) {
...
+  }
+
+  /// Apply a single action and return the resulting value.
+  Value applyAction(RewriterBase &builder,
+                    const RematerializationAction &action) {
+    // Check if already applied.
+    auto it = applied.find(action.originalValue);
+    if (it != applied.end()) {
+      return it->second;
+    }
```
hanhanW: optional: I would do `if (applied.contains(...)) { return applied[...]; } style, because it also avoid leaking `it` variable to other snippet.

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
```diff
@@ -140,4 +145,206 @@ bool isNarrowNResult(EncodingAttr encoding) {
...
+    unsigned resultIndex =
+        std::distance(definingOp->getResults().begin(), resultIt);
+
+    analyzed.insert(value);
+    plan.push_back({RematerializationAction::Kind::CloneOp,
+                    value,
+                    {},
+                    {},
+                    definingOp,
+                    resultIndex});
```
hanhanW: So by the time we get to the CloneOp case, we've already verified that the value has a defining op (i.e., it's not a block argument). This means value can be cast to `OpResult`, and we can use it directly in the struct. Then we no longer need to compute the `resultIndex` ourself. The `OpResult` should have all the information we need.

Example code changes from LLM:

```c++
  // In analyzeValue:
  plan.push_back(CloneOpAction{cast<OpResult>(value)});

  // In applyAction:
  } else if constexpr (std::is_same_v<T, CloneOpAction>) {
    Operation *cloned = builder.clone(*act.originalResult.getOwner());
    for (auto [idx, operand] :
         llvm::enumerate(act.originalResult.getOwner()->getOperands())) {
      if (auto it = applied.find(operand); it != applied.end()) {
        cloned->setOperand(idx, it->second);
      }
    }
    return cloned->getResult(act.originalResult.getResultNumber());
  }
```

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.h
```diff
@@ -80,6 +81,22 @@ MatmulNarrowDim getPo2MatmulNarrowDim(EncodingAttr encoding);
...
+/// when propagating encodings through operations where the original encoding
+/// dims don't dominate the new insertion point.
+///
+/// Returns the rematerialized dims if successful, or failure if some dims
+/// cannot be rematerialized. On failure, no operations are created.
+///
+/// Recursively rematerializes operands as needed. Supports:
+/// - Values that already dominate the insertion point (used as-is)
+/// - tensor.dim ops on propagationSource (recreated using newSource)
+/// - Pure operations (cloned with rematerialized operands)
```
hanhanW: Style nit: add proper punctuation to comments.

```suggestion
/// - Values that already dominate the insertion point (used as-is).
/// - tensor.dim ops on propagationSource (recreated using newSource).
/// - Pure operations (cloned with rematerialized operands).
```

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
```diff
@@ -140,4 +145,206 @@ bool isNarrowNResult(EncodingAttr encoding) {
...
+  Value dimIndex;
+  // For CloneOp: the operation to clone and which result index to use
+  Operation *opToClone = nullptr;
+  unsigned resultIndex = 0;
+};
+
+/// Helper class for analyzing and executing rematerialization of encoding dims.
+/// Split into two phases to avoid creating operations on failure:
+/// 1. analyze() - builds a plan without creating any operations
+/// 2. apply() - executes the plan and creates operations
```
hanhanW: Style nit: add proper punctuation to comments. Same review comment for other comments. (here and above and maybe other places.)

```suggestion
/// Helper class for analyzing and executing rematerialization of encoding dims.
/// Split into two phases to avoid creating operations on failure:
/// 1. analyze() - builds a plan without creating any operations.
/// 2. apply() - executes the plan and creates operations.
```

compiler/src/iree/compiler/ExternalInterfaces/EncodingExternalModels.cpp
```diff
@@ -43,6 +45,23 @@ static IREE::Encoding::PropagationResult propagateThroughEncodingCastableOp(
...
+  // input as their first operand.
+  Value sourceOperand = op->getOperand(0);
+  FailureOr<SmallVector<Value>> rematerializedDims =
+      IREE::Encoding::rematerializeEncodingDims(
+          builder, op, encodings.encodingDims, propagationSource->get(),
+          sourceOperand);
+  if (failed(rematerializedDims)) {
+    return failure();
+  }
+  // Update encodings with rematerialized dims.
```
hanhanW: optional: this comment is redundant to me because the above mega comment already states it.

hanhanW (changes requested):
It looks better to me. The only main concern is about `RematerializationAction`, please take a look.

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
jtuyls: Nice suggestion, thanks!

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
jtuyls: Updated it

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
jtuyls: Updated it, thanks!

compiler/src/iree/compiler/Dialect/Encoding/Utils/Utils.cpp
jtuyls: Checked here and other comments

hanhanW (approved):
LGTM, thanks!

jtuyls:
@hanhanW I needed to make one more larger change here to fix the e2e tests, see last commit. Could you help review?

hanhanW (approved):

PR #23139, opened 2026-01-15
[Codegen] Add pass to remove iree_codegen.index_hint ops

Max191:
Adds a pass to remove iree_codegen.index_hint operations. The pass unconditionally drops all index_hint ops, and should be used once the compiler is done using them for optimizations. The ops can get in the way of later optimizations, so this pass should be used to drop them once they are no longer needed.

The pass is not added to any pipelines, because we are not generating index_hint ops anywhere yet, but this pass will be added later once index_hints start to be used.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/RemoveIndexHints.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/RemoveIndexHints.cpp
@@ -24,14 +24,12 @@ struct RemoveIndexHintsPass final
     FunctionOpInterface funcOp = getOperation();
     IRRewriter rewriter(funcOp.getContext());
 
-    // Collect all index_hint operations
     SmallVector<IREE::Codegen::IndexHintOp> indexHintOps;
     funcOp.walk([&](IREE::Codegen::IndexHintOp hintOp) {
       indexHintOps.push_back(hintOp);
     });
 
     for (auto hintOp : indexHintOps) {
-      // Replace hint result with the original input (pass-through)
       hintOp.getResult().replaceAllUsesWith(hintOp.getInput());
       rewriter.eraseOp(hintOp);
     }
--- a/compiler/src/iree/compiler/Codegen/Common/RemoveIndexHints.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/RemoveIndexHints.cpp
@@ -1,0 +1,43 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_REMOVEINDEXHINTSPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+
+/// Pass to remove all iree_codegen.index_hint operations by replacing them
+/// with their input values.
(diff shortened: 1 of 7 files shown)


MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/Common/RemoveIndexHints.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+/// Pass to remove all iree_codegen.index_hint operations by replacing them
+/// with their input values.
+struct RemoveIndexHintsPass final
+    : impl::RemoveIndexHintsPassBase<RemoveIndexHintsPass> {
+  void runOnOperation() override {
+    FunctionOpInterface funcOp = getOperation();
+    IRRewriter rewriter(funcOp.getContext());
+
+    // Collect all index_hint operations
```
hanhanW:
```suggestion
    // Collect all index_hint operations.
```

Or just remove the comment. I think the code is self-documenting.

compiler/src/iree/compiler/Codegen/Common/RemoveIndexHints.cpp
```diff
@@ -0,0 +1,43 @@
...
+    SmallVector<IREE::Codegen::IndexHintOp> indexHintOps;
+    funcOp.walk([&](IREE::Codegen::IndexHintOp hintOp) {
+      indexHintOps.push_back(hintOp);
+    });
+
+    for (auto hintOp : indexHintOps) {
+      // Replace hint result with the original input (pass-through)
```
hanhanW:
```suggestion
      // Replace hint result with the original input (pass-through).
```

same comment here.

compiler/src/iree/compiler/Codegen/Common/test/remove_index_hints.mlir
```diff
@@ -0,0 +1,34 @@
...
+// CHECK-NOT: iree_codegen.index_hint
+// CHECK: return %arg0
+func.func @remove_lane_increment_hint(%arg0: index) -> index {
+  %hint = iree_codegen.index_hint %arg0(#iree_gpu.lane_increment<16>) : index
+  return %hint : index
+}
+
+// -----
+
+// Test: Multiple hints in sequence are all removed
```
hanhanW:
```suggestion
// Test: Multiple hints in sequence are all removed.
```

compiler/src/iree/compiler/Codegen/Common/test/remove_index_hints.mlir
```diff
@@ -0,0 +1,34 @@
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-codegen-remove-index-hints))' %s | FileCheck %s
+
+// Test: index_hint with lane_constant is removed
```
hanhanW:
```suggestion
// Test: index_hint with lane_constant is removed.
```

compiler/src/iree/compiler/Codegen/Common/test/remove_index_hints.mlir
```diff
@@ -0,0 +1,34 @@
...
+// CHECK-NOT: iree_codegen.index_hint
+// CHECK: return %arg0
+func.func @remove_lane_constant_hint(%arg0: index) -> index {
+  %hint = iree_codegen.index_hint %arg0(#iree_gpu.lane_constant<16>) : index
+  return %hint : index
+}
+
+// -----
+
+// Test: index_hint with lane_increment is removed
```
hanhanW:
```suggestion
// Test: index_hint with lane_increment is removed.
```

hanhanW (approved):
LGTM if the review comments are addressed.

krzysz00 (commented):
I'd still like to see these index_hint ops implementing the integer inference interfaces

Max191:
> I'd still like to see these index_hint ops implementing the integer inference interfaces

We can add it, although I realized that for the current attributes that we have, the implementation would mostly just be to avoid blocking the analysis. The hints give very little information about the ranges/divisibilities of the index values, since they are more about how the index varies across lanes.

(other hint attributes could give more information, of course)

PR #23149, opened 2026-01-15
[bazel][NFC] Add `# keep sorted` to enforce_glob calls in BUILD.bazel files.

hanhanW:
This change adds buildifier's `# keep sorted` directive to all enforce_glob calls in `compiler/` and `tests/` directories, ensuring file lists stay alphabetically sorted. Also converts one remaining glob() call to enforce_glob() in the ROCM ukernel BUILD.bazel.


kuhar (approved):
I wish there was some automatic way to make it the default

PR #23153, opened 2026-01-15
[LinalgExt] Extend arg_compare to support both value and index provided ( explicit-index mode)

bangtianliu:
This PR extends the `arg_compare `operation to accept an optional second input for pre-computed indices, addressing a critical limitation discovered while implementing VectorDistribute pipeline support for argmax/argmin operations.
 
## Motivation
 While working on arg_compare support along the VectorDistribute pipeline, we encountered a fundamental issue with the merge reduction phase after partial reduction. Each tile produces a (max value, index) pair, and merging them creates this pattern:
 
 ```mlir
   %19:2 = linalg.generic {
    indexing_maps = [...], iterator_types = ["reduction"]
  } ins(%18#0, %18#1 : tensor<2048xf16>, tensor<2048xi32>)
    outs(%12, %13 : tensor<f16>, tensor<i32>) {
  ^bb0(%in: f16, %in_3: i32, %out: f16, %out_4: i32):
    %20 = arith.cmpf ogt, %in, %out : f16
    %21 = arith.select %20, %in, %out : f16
    %22 = arith.select %20, %in_3, %out_4 : i32
    linalg.yield %21, %22 : f16, i32
  }
 ```
 The upstream MLIR vectorizer cannot handle this pattern. The paired-select reduction (selecting both value and index based on the same condition) doesn't fit the single-combiner model that generic vectorization expects. 
 
This PR extends arg_compare to optionally accept index inputs, enabling it to handle both:
  1. Initial argmax: Compute (value, index) from scratch using iteration variables (linalg.index)
  2. Merge reduction: Merge partial (value, index) pairs from distributed tiles

This PR is mainly update the iree_linalg_ext.arg_compare op definition and verification logic, along with roundtrip and invalid IR tests.                                                          
                                                                                                                                                                                                                      
  After this PR, there will be a few PRs about                                                                                                                                                                        
  - Add tiling support                                                                                                                                                                                                
  - Add convert_to_loops lowering support                                                                                                                                                                             
  - Update partial reduction interface support.  
  - Update e2e test.   

Refer to discord discussion for more details: https://discordapp.com/channels/689900678990135345/1461079481803608158/1461079488770212096

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -1247,12 +1247,18 @@ TopkOp::reifyResultShapes(OpBuilder &b,
 LogicalResult ArgCompareOp::verify() {
   Operation *op = getOperation();
 
-  unsigned numInputVals = llvm::size(getInputs());
-  if (numInputVals != 1) {
-    return op->emitOpError(
-               "expected exactly one tensor input operand, but got ")
-           << numInputVals;
+  // The operation supports two modes based on the number of inputs:
+  // - Implicit-index mode (1 input): Computes indices from iteration variables.
+  // - Explicit-index mode (2 inputs): Uses pre-existing (value, index) pairs.
+  unsigned numInputs = llvm::size(getInputs());
+  if (numInputs < 1 || numInputs > 2) {
+    return op->emitOpError("expected 1 or 2 input operands, but got ")
+           << numInputs;
   }
+
+  bool isExplicitIndexMode = numInputs == 2;
+  ShapedType inputValueType = getInputType();
+  Type inputElemType = inputValueType.getElementType();
 
   unsigned numOutputs = getNumDpsInits();
   if (numOutputs != 2) {
@@ -1261,21 +1267,48 @@ LogicalResult ArgCompareOp::verify() {
            << numOutputs;
   }
 
-  uint64_t dim = getDimension();
-  int64_t rank = getInputRank();
-  if (dim >= rank) {
-    return op->emitOpError("reduction dimension exceeds or equals input rank. ")
-           << "got dimension: " << dim << ", but input rank is: " << rank;
-  }
-
-  ShapedType inputType = getInputType();
   auto outputValueType = getOutputValueType();
   auto outputIndexType = getOutputIndexType();
 
-  if (inputType.getElementType() != outputValueType.getElementType()) {
+  if (isExplicitIndexMode) {
+    ShapedType inputIndexType = cast<ShapedType>(getInputIndex().getType());
+
+    if (failed(verifyCompatibleShape(inputValueType, inputIndexType))) {
+      return op->emitOpError(
+                 "explicit-index mode: value and index inputs must have same "
+                 "shape. ")
+             << "Value shape: "
+             << llvm::interleaved_array(inputValueType.getShape())
+             << ", index shape: "
+             << llvm::interleaved_array(inputIndexType.getShape());
+    }
+
+    if (!isa<IntegerType, IndexType>(inputIndexType.getElementType())) {
+      return op->emitOpError(
+                 "explicit-index mode: index input must have integer or index "
+                 "element type, but got ")
+             << inputIndexType.getElementType();
+    }
+
+    if (inputIndexType.getElementType() != outputIndexType.getElementType()) {
+      return op->emitOpError(
+                 "explicit-index mode: input and output index element types "
+                 "must match. ")
+             << "Input index type: " << inputIndexType.getElementType()
+             << ", output index type: " << outputIndexType.getElementType();
+    }
+  }
+
+  if (inputValueType.getElementType() != outputValueType.getElementType()) {
     return op->emitOpError("input and output value element types must match. ")
-           << "Input type: " << inputType.getElementType()
+           << "Input type: " << inputValueType.getElementType()
            << ", output value type: " << outputValueType.getElementType();
+  }
+
+  if (!isa<IntegerType, IndexType>(outputIndexType.getElementType())) {
+    return op->emitOpError(
+               "output index must have integer or index element type, but got ")
+           << outputIndexType.getElementType();
   }
 
   if (failed(verifyCompatibleShape(outputValueType, outputIndexType))) {
@@ -1286,10 +1319,17 @@ LogicalResult ArgCompareOp::verify() {
            << llvm::interleaved_array(outputIndexType.getShape());
   }
 
+  uint64_t dim = getDimension();
+  int64_t rank = getInputRank();
+  if (dim >= rank) {
+    return op->emitOpError("reduction dimension exceeds or equals input rank. ")
+           << "got dimension: " << dim << ", but input rank is: " << rank;
+  }
+
   SmallVector<int64_t> expectedShape;
   for (int64_t i = 0; i < rank; ++i) {
     if (i != dim) {
-      expectedShape.push_back(inputType.getDimSize(i));
+      expectedShape.push_back(inputValueType.getDimSize(i));
     }
   }
   if (!llvm::equal(expectedShape, outputValueType.getShape())) {
@@ -1307,13 +1347,13 @@ LogicalResult ArgCompareOp::verify() {
     return op->emitOpError("region block should have 2 arguments, but got ")
            << numArgs;
   }
-  Type inputElemType = inputType.getElementType();
+
   Type arg0Type = block.getArgument(0).getType();
   Type arg1Type = block.getArgument(1).getType();
 
   if (arg0Type != inputElemType || arg1Type != inputElemType) {
     return op->emitOpError(
-               "comparator region arguments must match input element type. ")
+               "comparator arguments must match input value element type. ")
            << "Expected: " << inputElemType << ", but got: " << arg0Type
            << " and " << arg1Type;
   }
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -752,14 +752,24 @@ def IREELinalgExt_ArgCompareOp : IREELinalgExt_Op<"arg_compare", [
...
+    1. Implicit-index mode (single input): Computes (value, index) pairs from scratch.
+       Indices are implicitly computed from the reduction dimension's iteration index
+       (using linalg.index on the reduction dimension). The comparator
+       region receives two candidate values and returns a single `i1` result indicating
+       whether the first argument should be preferred.
+
+    2. Explicit-index mode (two inputs): Reduces existing (value, index) pairs.
+       Indices are explicitly provided as input. The comparator region receives
+       two candidate values and returns `i1` to select the preferred value.
+       This mode is used to combine results from split-reduction operations.
```
IanWood1, line 766: Here's an alternative design that I think might be cleaner and more general. Instead of special-casing a second "index input," what if we treat all inputs uniformly? In the case of merge reduction, the arg compare op would receive the partial values and indices and treat them both as normal inputs:


```mlir
%0:3 = iree_linalg_ext.arg_compare dimension(1)
  ins(%partial_vals, %partial_idxs : tensor<2x4xf32>, tensor<2x4xi32>)
  outs(%max_val, %max_idx, %out_idx : tensor<2xf32>, tensor<2xi32>, tensor<2xi32>) {
  ^bb0(%val_a: f32, %idx_a: i32, %val_b: f32, %idx_b: i32):
    %cmp = arith.cmpf ogt, %val_a, %val_b : f32
}
```

The third output (`%out_idx`) would be left unused because we don't care about the index into the partial reduction tensors. This approach seems more natural to me because in the merge step we are just applying an arg max over a tuple of values. Also, it might be easier to write code for this op since handling two different modes might be tricky. However, the trade off is extra block args and an unused result. What do you think?

bangtianliu: > what if we treat all inputs uniformly

 No, that's not the semantics of `arg_compare`. The inputs should be either a value array (implicit-index mode), or a value-index pair with matching shapes (explicit-index mode). The index input has special semantics as position tracking - it's not just another arbitrary input.
  
  I'd also like to clarify the semantics of the merge step. It's not about applying an argmax over a tuple of values. Rather, each tile produces a single (value, index) pair, and the merge step performs a pairwise reduction using the same predicate logic on the values. The indices are simply carried along - whichever value wins the comparison, its corresponding index is selected.

  Regarding the region: in our design, the comparator always takes 2 arguments (just values), keeping it uniform across both modes:
  ```mlir
  ^bb0(%val_a: f32, %val_b: f32):
    %cmp = arith.cmpf ogt, %val_a, %val_b : f32
   ```
The region is just used to define predication logic. why would we need to expose indices to the comparator? For argmax/argmin, the comparison is purely value-based.
  
I think making it too general would blur its purpose and potentially complicate both the implementation and understanding.

IanWood1 (approved):

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1247,37 +1247,70 @@ TopkOp::reifyResultShapes(OpBuilder &b,
...
-  unsigned numInputVals = llvm::size(getInputs());
-  if (numInputVals != 1) {
-    return op->emitOpError(
-               "expected exactly one tensor input operand, but got ")
-           << numInputVals;
+  // The operation supports two modes based on the number of inputs:
+  // - Implicit-index mode (1 input): Computes indices from iteration variables.
+  // - Explicit-index mode (2 inputs): Uses pre-existing (value, index) pairs.
+  unsigned numInputs = llvm::size(getInputs());
+  if (numInputs < 1 || numInputs > 2) {
```
kuhar:
```suggestion
  if (numInputs == 1 || numInputs == 2) {
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1247,37 +1247,70 @@ TopkOp::reifyResultShapes(OpBuilder &b,
...
 
-  unsigned numInputVals = llvm::size(getInputs());
-  if (numInputVals != 1) {
-    return op->emitOpError(
-               "expected exactly one tensor input operand, but got ")
-           << numInputVals;
+  // The operation supports two modes based on the number of inputs:
+  // - Implicit-index mode (1 input): Computes indices from iteration variables.
+  // - Explicit-index mode (2 inputs): Uses pre-existing (value, index) pairs.
+  unsigned numInputs = llvm::size(getInputs());
```
kuhar, line 1253: Maybe add this as an extra class declaration instead?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1247,37 +1247,70 @@ TopkOp::reifyResultShapes(OpBuilder &b,
...
-           << "got dimension: " << dim << ", but input rank is: " << rank;
-  }
-
-  ShapedType inputType = getInputType();
   auto outputValueType = getOutputValueType();
   auto outputIndexType = getOutputIndexType();
 
-  if (inputType.getElementType() != outputValueType.getElementType()) {
+  if (isExplicitIndexMode) {
+    ShapedType inputIndexType = cast<ShapedType>(getInputIndex().getType());
```
kuhar: Also here: maybe add this as a helper in ODS?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1247,37 +1247,70 @@ TopkOp::reifyResultShapes(OpBuilder &b,
...
+
+    if (failed(verifyCompatibleShape(inputValueType, inputIndexType))) {
+      return op->emitOpError(
+                 "explicit-index mode: value and index inputs must have same "
+                 "shape. ")
```
kuhar: does compatible == the same, or do we allow for some wiggle room?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1247,37 +1247,70 @@ TopkOp::reifyResultShapes(OpBuilder &b,
...
+    }
+
+    if (!isa<IntegerType, IndexType>(inputIndexType.getElementType())) {
+      return op->emitOpError(
+                 "explicit-index mode: index input must have integer or index "
+                 "element type, but got ")
+             << inputIndexType.getElementType();
+    }
+
+    if (inputIndexType.getElementType() != outputIndexType.getElementType()) {
```
kuhar: Can we define these through ODS?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1247,37 +1247,70 @@ TopkOp::reifyResultShapes(OpBuilder &b,
...
+  }
+
+  if (inputValueType.getElementType() != outputValueType.getElementType()) {
     return op->emitOpError("input and output value element types must match. ")
-           << "Input type: " << inputType.getElementType()
+           << "Input type: " << inputValueType.getElementType()
            << ", output value type: " << outputValueType.getElementType();
   }
 
+  if (!isa<IntegerType, IndexType>(outputIndexType.getElementType())) {
```
kuhar: Also here

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1247,37 +1247,70 @@ TopkOp::reifyResultShapes(OpBuilder &b,
...
+  // The operation supports two modes based on the number of inputs:
+  // - Implicit-index mode (1 input): Computes indices from iteration variables.
+  // - Explicit-index mode (2 inputs): Uses pre-existing (value, index) pairs.
+  unsigned numInputs = llvm::size(getInputs());
+  if (numInputs < 1 || numInputs > 2) {
+    return op->emitOpError("expected 1 or 2 input operands, but got ")
+           << numInputs;
   }
 
+  bool isExplicitIndexMode = numInputs == 2;
```
kuhar: You already have a helper for this: `hasExplicitIndexInput`

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -820,6 +846,19 @@ def IREELinalgExt_ArgCompareOp : IREELinalgExt_Op<"arg_compare", [
...
+    // when a second input operand provides pre-computed indices.
+    bool hasExplicitIndexInput() {
+      return getInputs().size() == 2;
+    }
+
+    // Returns the index input operand. This method should only be called when
+    // the operation is in explicit-index mode.
+    Value getInputIndex() {
+      assert(hasExplicitIndexInput() && "Only valid in explicit-index mode");
+      return getDpsInputOperand(1)->get();
```
kuhar: alternatively you could return nullptr when there's none and not assert

kuhar (approved):

PR #23162, opened 2026-01-16
Fix test failures revealed by reverse iteration (nondeterminism)

kuhar:
Using https://github.com/iree-org/iree/pull/23161


amd-eochoalo (approved):
I think this looks good, but it is my first time looking into these files. The changes in AutomaticReferenceCounting.cpp are pretty self contained so that looks good! For the changes of HoistIntoGlobals, my question would be why aren't we using dominance instead of adding opOrder? Is it too expensive?

kuhar:
Because two ops in sibling regions wouldn't dominate one another. I think we need total order to stay deterministic.

amd-eochoalo:
> Because two ops in sibling regions wouldn't dominate one another. I think we need total order to stay deterministic.

Thanks!

PR #23165, opened 2026-01-16
[Codegen] Create CombineSourceLayoutTransformation pass for MapGatherOp

Abhishek-Varma:
-- This commit creates a separate pass `iree-codegen-combine-source-layout-transformation`
    adding support of folding relayout operations into map_gather op.
-- The following ops' folding are being supported through changes in this PR :-
    1. `linalg::CopyOp`
    2. `linalg::TransposeOp`
    3. `tensor::ExpandShapeOp`
    4. `tensor::CollapseShapeOp`
    5. `tensor::ExtractSliceOp`
    6. `tensor::PadOp`
-- Consequently `getPaddingValue` and `createIdentityMapGather` member methods were added to `MapGatherOp`.
-- The implementations for the `iree-codegen-combine-source-layout-transformation` pass reside in the same file as `MapScatter`'s relayout folding implementations (which has been renamed as `iree-codegen-combine-result-layout-transformation`).


compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -452,6 +453,212 @@ FailureOr<MapScatterOp> foldIntoMapScatter(RewriterBase &rewriter,
...
+  if (extractSliceOp.getSourceType().getRank() !=
+      extractSliceOp.getResultType().getRank()) {
+    return rewriter.notifyMatchFailure(
+        extractSliceOp, "rank reducing extract_slice op is not supported");
+  }
+  // TODO(Max191): Support non-unit strides.
+  if (!areAllConstantIntValue(extractSliceOp.getMixedStrides(), 1)) {
+    return rewriter.notifyMatchFailure(extractSliceOp,
+                                       "non-unit strides are not supported");
+  }
```
hanhanW: `TODO(Max191)`? Copy-paste problem?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -452,6 +453,212 @@ FailureOr<MapScatterOp> foldIntoMapScatter(RewriterBase &rewriter,
...
+      }
+    }
+  }
+
+  if (isIdentity) {
+    // Identity extract_slice: just replace the source without any changes.
+    rewriter.modifyOpInPlace(
+        mapGatherOp, [&]() { mapGatherOp.getSourceMutable().assign(source); });
+    return mapGatherOp;
+  }
```
hanhanW: Why dont we combine them?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -452,6 +453,212 @@ FailureOr<MapScatterOp> foldIntoMapScatter(RewriterBase &rewriter,
...
+static MapGatherOp insertIdentityMapGather(RewriterBase &rewriter,
+                                           OpResult root) {
+  Location loc = root.getLoc();
+  SetVector<OpOperand *> originalUses;
+  for (OpOperand &use : root.getUses()) {
+    originalUses.insert(&use);
+  }
+  OpBuilder::InsertionGuard g(rewriter);
+  rewriter.setInsertionPointAfterValue(root);
+  Type elementType = cast<RankedTensorType>(root.getType()).getElementType();
```
hanhanW: why cast?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -347,6 +347,31 @@ def IREELinalgExt_MapGatherOp : IREELinalgExt_Op<"map_gather",
...
+    // receives the current yielded source indices (excluding padding) and
+    // should return the new source indices to yield.
+    void insertTransformationAtEnd(
+        OpBuilder &builder,
+        function_ref<SmallVector<Value>(ValueRange)> transformationBuilder);
+
+    // Return true if the map_gather op has vector semantics.
+    bool isVectorized() {
+      return isa<VectorType>(getOutputType());
+    }
```
hanhanW: unused?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUCombineLayoutTransformation.cpp
```diff
@@ -74,7 +74,16 @@ gpuPadDistributionConfigFn(ArrayRef<int64_t> iterationBounds,
...
+///
+/// Note: This control function currently only handles map_scatter.
+/// map_gather is not yet supported in the GPU pass.
+static bool gpuRelayoutCombinationControlFn(OpResult leaf,
+                                            RelayoutOpKind kind) {
+  // TODO: Since we changed the API to include RelayoutOpKind, only handle
+  // map_scatter. Will add support for map_gather in GPU Codegen in a future PR.
+  if (kind == RelayoutOpKind::Gather) {
+    return false;
+  }
```
hanhanW: Why do we need this? This is not documented in PR description.

hanhanW (changes requested):
I did not review all the code yet, but I have a first high-level question: do we want to make both transformation (gather and scatter) be in a single pass? They look orthogonal to me, and it is not clear that which folding is prioritized.

The other high level comment is passing bool-ish argument (i.e., your new enum) to functions. It makes this much more difficult to follow (especially in what is already difficult code). This comment depends on the first mega comment about if we want to split it into two passes or not.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Updated.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUCombineLayoutTransformation.cpp
Abhishek-Varma: This had to be done because of the API signature needing the enum.

But I've removed it now since we are creating a new pass altogether for map_gather now.

nirvedhmeshram:
> -- Unlike map_scatter, we don't allow tensor.pad op in the relayout op chain

I think we want to allow this as the use case mentioned in https://github.com/iree-org/iree/issues/23038 has pad op in it?

hanhanW:
Hey @Abhishek-Varma I did not get a chance to review it today. I'll prioritize it tomorrow. You can continue working on other changes and use stacked PR features. (or just send the PR out and ask me/others to review the last commit.)

hanhanW (changes requested):
## Code Review: Add map_gather Support in CombineLayoutTransformation

### Summary
This PR adds support for `iree_linalg_ext.map_gather` op to combine layout transformation ops (transpose, expand_shape, collapse_shape, extract_slice, copy) into a single map_gather operation, similar to the existing map_scatter support. The implementation creates a new pass `CombineLayoutTransformationForMapGatherPass` for Dispatch scope.

### CI Status
Most checks pass. The failing `Test Sharktank / sharktank_model_tests :: rocm_hip_w7900` appears unrelated to this PR.

---

### Critical Issues

**1. Code Duplication Concern**
The diff creates a separate file `CombineLayoutTransformationForMapGather.cpp` that duplicates significant portions of `CombineLayoutTransformation.cpp`. This includes:
- `simplifyComplexRelayoutOpsForGather` vs `simplifyComplexRelayoutOps`
- `isSupportedGatherRelayoutOp` vs `isSupportedRelayoutOp`
- Pattern structure for `InsertMapGatherOpPattern` vs `InsertMapScatterOpPattern`

Consider refactoring to share common code. A template-based or policy-based approach could reduce duplication.

---

### Medium Issues

**2. Missing `affine::AffineDialect` Dependency**
The new pass creates `affine::AffineLinearizeIndexOp` and `affine::AffineDelinearizeIndexOp` in the transformation body. The dependent dialects in `Passes.td` should include:
```tablegen
let dependentDialects = [
    "affine::AffineDialect",  // Missing!
    "iree_compiler::IREE::LinalgExt::IREELinalgExtDialect",
    "scf::SCFDialect",
    "tensor::TensorDialect"
];
```

---

### Minor Issues

**3. Comment Inconsistency**
The function comment at line 289:
```cpp
// Insert identity map_gather op after the root and replace uses.
```
Should probably say "before the root" or clarify the position more accurately, since you're intercepting the chain of relayout ops.

**4. Test Coverage**
The tests look good and cover the main cases:
- transpose folding
- expand_shape folding
- collapse_shape folding  
- extract_slice with offset folding
- copy folding (resulting in identity cleanup)

Consider adding tests for failure cases:
- Chained relayout ops (e.g., transpose + expand_shape)
- Non-unit stride extract_slice (should fail gracefully)
- Rank-reducing extract_slice (should fail gracefully)

---

### Architectural Question

**5. Why Separate Pass vs Extending Existing?**
The PR creates a completely new pass `CombineLayoutTransformationForMapGatherPass` rather than adding map_gather support to the existing `CombineLayoutTransformationPass`. 

The `Transforms.cpp` changes add `FoldRelayoutOpIntoMapGatherPattern` alongside the scatter pattern, suggesting the intent is to have both work together. But then a separate pass is also created. Could you clarify in the PR description:
- When should `CombineLayoutTransformationPass` be used vs `CombineLayoutTransformationForMapGatherPass`?
- Is the intent for these to eventually merge?

---

### LinalgExtOps.cpp Changes Review

The new methods added to `MapGatherOp` look correct:
- `insertTransformationAtEnd`
- `isIdentity`
- `createIdentityMapGather`

These mirror the map_scatter equivalents and are well-implemented.

---

### Recommendation

Please address:
1. Add missing `affine::AffineDialect` to dependentDialects
2. Add clarification in PR description about pass design choice (separate pass vs extending existing)
3. Consider adding tests for failure cases

The core implementation logic for folding layout ops into map_gather appears correct and follows the patterns established by map_scatter.

hanhanW:
I don't follow why a review comment is generated by LLM agents. I hided it and deleted the app.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,522 @@
+// Copyright 2025 The IREE Authors
```
hanhanW: 2026

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.h
```diff
@@ -0,0 +1,56 @@
+// Copyright 2025 The IREE Authors
```
hanhanW: 2026

We actually dont need this header. The methods can go into `CombineLayoutTransformation.h`, which collects all the transformation methods for `CombineLayout*`.

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -175,6 +175,40 @@ def CombineLayoutTransformationPass :
...
+           ::llvm::cl::values(
+             clEnumValN(IREE::Codegen::GatherRelayoutCombinationScope::Dispatch, "dispatch",
+               "Combine relayout ops starting from iree_codegen.load_from_buffer"))
+           }]>
+  ];
+  let dependentDialects = [
+    "iree_compiler::IREE::LinalgExt::IREELinalgExtDialect",
+    "scf::SCFDialect",
+    "tensor::TensorDialect"
+  ];
```
hanhanW, line 207: We should include `affine::AffineDialect` because the pass creates affine::AffineLinearizeIndexOp and affine::AffineDelinearizeIndexOp. And also `arith::ArithDialect`.

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -175,6 +175,40 @@ def CombineLayoutTransformationPass :
...
+    Starting from the specified root ops, iteratively combine consumer
+    layout/indexing transformation ops (linalg.transpose, tensor.collapse_shape,
+    tensor.expand_shape, tensor.extract_slice, etc.) into a single
+    iree_linalg_ext.map_gather operation.
+
+    Note: tensor.pad is NOT supported for map_gather. Unlike map_scatter which
+    can write padding values directly to the output buffer (obtained from
+    store_to_buffer), map_gather reads from a source buffer and produces a
+    tensor result. There is no output memref available to write padding values
+    to at this stage.
```
hanhanW: need update

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -175,6 +175,40 @@ def CombineLayoutTransformationPass :
...
+    Option<"scope", "scope",
+           "::mlir::iree_compiler::IREE::Codegen::GatherRelayoutCombinationScope",
+           /*default=*/"IREE::Codegen::GatherRelayoutCombinationScope::Dispatch",
+           "Specify the roots for layout transformation combination.",
+           [{
+           ::llvm::cl::values(
+             clEnumValN(IREE::Codegen::GatherRelayoutCombinationScope::Dispatch, "dispatch",
+               "Combine relayout ops starting from iree_codegen.load_from_buffer"))
+           }]>
+  ];
```
hanhanW: Do we really need the option? Do we know what to do with the options? Can we introduce them later when we understand what is needed?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,522 @@
...
+}
+
+// This is only desirable in the dispatch scope but not in the workgroup scope.
+static bool shouldDoReshapesByExpansion(
+    IREE::Codegen::GatherRelayoutCombinationScope scope) {
+  if (scope == IREE::Codegen::GatherRelayoutCombinationScope::Dispatch) {
+    return true;
+  }
+  return false;
+}
```
hanhanW: It is dead code to me. We should introduce such method when needed.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,522 @@
...
+  // support for gather). This adds both scatter and gather folding patterns,
+  // but since we only insert map_gather ops, only those will be matched.
+  populateCombineRelayoutOpPatterns(relayoutCombinationPatterns,
+                                    /*padDistributionConfigFn=*/nullptr);
+  memref::populateResolveRankedShapedTypeResultDimsPatterns(
+      relayoutCombinationPatterns);
+  if (failed(applyPatternsGreedily(funcOp,
+                                   std::move(relayoutCombinationPatterns)))) {
+    return failure();
+  }
```
hanhanW: Most of other things in this function are not used at all. Can we start with these, and build the pass based on what we're seeing?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+  // Clean up any identity map_gather ops after combining.
+  funcOp->walk([&](MapGatherOp mapGatherOp) {
+    if (mapGatherOp.isIdentity()) {
+      rewriter.replaceOp(mapGatherOp, mapGatherOp.getSource());
+    }
+  });
```
hanhanW: This belong to a canonicalization pattern and it does not belong to this PR.

hanhanW (changes requested):
I haven't reviewed the core logic yet, but can we clean up first? I think some of them are not used at all, and it is adding tech debt and overheads to reviewers.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.h
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/Passes.td
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/Passes.td
Abhishek-Varma: Sure. I've removed them for now. Will add later.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
Abhishek-Varma: Done. Can you check now ?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
Abhishek-Varma: Sure. I have raised a PR : https://github.com/iree-org/iree/pull/23240 - not removing from the current commit to allow the tests to run.

compiler/src/iree/compiler/Codegen/Common/test/combine_layout_transformation_for_map_gather.mlir
```diff
@@ -0,0 +1,147 @@
...
+}
+// CHECK-LABEL: @fold_transpose
+//  CHECK-SAME:   %[[BUFFER:[a-zA-Z0-9_]+]]
+//       CHECK:   %[[SOURCE:.+]] = iree_codegen.load_from_buffer %[[BUFFER]]
+//       CHECK:   %[[DEST:.+]] = tensor.empty() : tensor<4x16x2xf32>
+//   CHECK-NOT:   linalg.transpose
+//       CHECK:   %[[MAP_GATHER:.+]] = iree_linalg_ext.map_gather
+//  CHECK-SAME:     %[[SOURCE]] into %[[DEST]] {
+//  CHECK-NEXT:   ^bb0(%[[IDX0:.+]]: index, %[[IDX1:.+]]: index, %[[IDX2:.+]]: index):
+// For perm=[1,2,0]: result[i0,i1,i2] = input[i1,i2,i0], so yield [idx1,idx2,idx0]
```
hanhanW:
```suggestion
// For perm=[1,2,0]: result[i0,i1,i2] = input[i1,i2,i0], so yield [idx1,idx2,idx0].
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -751,6 +774,37 @@ void MapGatherOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+      builder.createBlock(&region, region.end(), indexTypes, blockArgLocs);
+  SmallVector<Value> yieldedValues(block->getArguments());
+  // Add a zero padding value (the identity transformation shouldn't need it
+  // since source and output have the same shape).
+  Type elementType = outputType.getElementType();
+  TypedAttr zeroAttr = builder.getZeroAttr(elementType);
+  Value padding = arith::ConstantOp::create(builder, loc, zeroAttr);
+  yieldedValues.push_back(padding);
+  IREE::LinalgExt::YieldOp::create(builder, loc, yieldedValues);
+  return mapGatherOp;
```
hanhanW, line 791: nit: be consistent at least within a function. You have a blank line before the comment.


```suggestion
  SmallVector<Value> yieldedValues(block->getArguments());
  
  // Add a zero padding value (the identity transformation shouldn't need it
  // since source and output have the same shape).
  Type elementType = outputType.getElementType();
  TypedAttr zeroAttr = builder.getZeroAttr(elementType);
  Value padding = arith::ConstantOp::create(builder, loc, zeroAttr);
  yieldedValues.push_back(padding);
  IREE::LinalgExt::YieldOp::create(builder, loc, yieldedValues);
  return mapGatherOp;
```

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -164,12 +164,33 @@ struct FoldPadOpIntoMapScatterPattern
...
+  }
+};
+
 } // namespace
 
 void populateCombineRelayoutOpPatterns(
     RewritePatternSet &patterns,
     PadDistributionConfigFn padDistributionConfigFn) {
   patterns.add<FoldRelayoutOpIntoMapScatterPattern>(patterns.getContext());
+  patterns.add<FoldRelayoutOpIntoMapGatherPattern>(patterns.getContext());
```
hanhanW: If we are isolating two passes, what is the point of adding both patterns here?

I think my original question was not answered, or we do not have an answer yet. This brings the question back: what is the priority of folding ops? Do we prioritize map_scatter or map_gather? We need an answer. This is why I suggested isolating two passes in the first place. Hopefully, we can get the answer when we plumb through the support.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,378 @@
...
+    for (auto [dim, sliceSize] : llvm::enumerate(sliceSizes)) {
+      ValueBoundsConstraintSet::Variable sourceDimVar(source, dim);
+      FailureOr<bool> areEqual =
+          ValueBoundsConstraintSet::areEqual(sliceSize, sourceDimVar);
+      if (!succeeded(areEqual) || !*areEqual) {
+        return false;
+      }
+    }
+    return true;
+  };
```
hanhanW: It seems to be a common util used by both map_gather and map_scatter? Can we refactor?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,378 @@
...
+    if (isa<IREE::Codegen::LoadFromBufferOp>(producer)) {
+      return true;
+    }
+    if (!isSupportedRelayoutOp(producer)) {
+      return false;
+    }
+    current = producer;
+  }
+  return false;
+}
```
hanhanW: The `getOperand(0)` is fragile to me. This is a different method; you are spreading assumptions to other methods. You should at least have an assertion and document it in function comments.

btw, we can rename `isSupportedRelayoutOp` to `isSupportedSingleInputRelayoutOp`, which makes it less fragile.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+/// Insert identity map_gather ops after the given operation if it is a valid
+/// leaf op of a relayout op chain. A relayout op chain is a sequence of
+/// relayout ops (defined by `isSupportedRelayoutOp`) for which the only
+/// users of the ops in the chain are relayout ops, except for the leaves of the
+/// chain. The leaves are simply relayout ops that have non relayout op users.
+struct InsertMapGatherOpPattern : public RewritePattern {
+  InsertMapGatherOpPattern(MLIRContext *context, PatternBenefit benefit = 1)
+      : RewritePattern(MatchAnyOpTypeTag(), benefit, context) {}
```
hanhanW: Why not use the Base? I.e., something like `using Base::Base`? You are silently prioritizing this pattern. Why?

hanhanW (changes requested):

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
Abhishek-Varma: Hi @hanhanW - I forgot separating it from this part of the code when creating a separate pass. I've done that now. Can you please check?

Abhishek-Varma:
Hi @hanhanW.

I've now changed the algo. I'm creating identity map_gather at the result of load_from_buffer and now iteratively fusing in consumer relayout ops instead of the earlier algo of creating an identity map_gather at the end of a relayout op chain. It's more intuitive now.

I've also refactored the code because most of the foldings followed a similar structure - so reviewing it will be easier now!

Locally, I've included this pass at https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L544-L545 and have [switched off the temporary workaround flag `iree-codegen-test-c-promtion`](https://github.com/iree-org/iree/blob/3824de72c2fd6a14a5cdd4c6a9eae45c91c795cc/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp#L37-L41) :-

I used the above setup to test the original regression with the [repro dispatch](https://github.com/iree-org/iree/issues/22919#issuecomment-3687432773) and I confirm it resolves the issue :-

```
real    0m10.902s
user    0m10.810s
sys     0m0.163s
```

NOTE: I confirmed first by simply switching off the `iree-codegen-test-c-promtion` flag - the compilation of the repro dispatch was taking > 10 minutes. This was done as a way to "revert" the temporary workaround first and confirm the issue exists. Then I followed the above setup and observed that with the pass introduced through this PR the compilation takes place successfully.

I've even verified the output of the dispatch. Even isolated a few cases like `transpose` op and verified the e2e correctness too.

I'm adding the [output.mlir and input.mlir](https://gist.github.com/Abhishek-Varma/ac74334090d9366633b60ba4235b97f7) for reference that the pass can handle complex relayout chains.

So, once this PR merges - I can raise another PR that removes the temporary workaround and adds this pass in the `addGPUTileAndFusePassPipeline`.

CC: @MaheshRavishankar @nirvedhmeshram @jtuyls @Max191

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,387 @@
+// Copyright 2026 The IREE Authors
```
Max191: Suggestion for pass organization + naming:

I think the map_gather and map_scatter implementations are similar enough that they can be in the same file. This should make sharing code much easier. Can you move the implementation of the map_gather pass into CombineLayoutTransformation.cpp?

Also, to keep naming consistent, can we change the names of both passes? I was thinking we can call them `CombineSourceLayoutTransformation` (map_gather) and `CombineResultLayoutTransformation` (map_scatter).

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,387 @@
...
+///
+/// This handles the shared boilerplate:
+/// 1. Create new dest tensor from consumer result shape
+/// 2. Clone map_gather with new dest and region
+/// 3. Apply index transformation via insertTransformationAtStart
+/// 4. Optionally update fill value
+/// 5. Replace consumer with new map_gather result
+template <typename ConsumerOpTy>
+static FailureOr<MapGatherOp> foldConsumerIntoMapGatherImpl(
+    RewriterBase &rewriter, ConsumerOpTy consumerOp, MapGatherOp mapGatherOp,
```
Max191: nit: The template looks unnecessary here, since you are just using standard Operation methods on the `consumerOp`. You should be able to just use `Operation *` instead of `ConsumerOpTy`.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,387 @@
...
+  rewriter.replaceOp(consumerOp, newMapGather.getResult(0));
+  LDBG() << "Folded consumer " << consumerOp->getName().getStringRef()
+         << " into map_gather:\n"
+         << newMapGather;
+  return newMapGather;
+}
+
+/// Fold a consumer `copyOp` into a producer `mapGatherOp`.
+/// Copy doesn't change indices, so we just replace uses.
+static FailureOr<MapGatherOp> foldCopyIntoMapGather(RewriterBase &rewriter,
```
Max191: nit: Rename to `foldIdentityLikeOpIntoMapGather` for consistency with the map_scatter implementation? (or renaming the map_scatter one to copy is also fine).

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+  ArrayRef<int64_t> perm = transposeOp.getPermutation();
+  SmallVector<int64_t> inversePerm = invertPermutationVector(perm);
+
+  auto indexTransformBuilder =
+      [&](ArrayRef<BlockArgument> newOutputIndices) -> SmallVector<Value> {
+    SmallVector<Value> indexValues(newOutputIndices.begin(),
+                                   newOutputIndices.end());
+    return applyPermutation(indexValues, inversePerm);
+  };
```
Max191: nit: If you combine the impls into the single CombineLayoutTransformation.cpp file, then you could create a shared index transformation builder for transpose, since this is the same as for map_scatter.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,387 @@
...
+      [&](ArrayRef<BlockArgument> newOutputIndices) -> SmallVector<Value> {
+    SmallVector<Value> indices(newOutputIndices.begin(),
+                               newOutputIndices.end());
+    auto linearizeIndexOp = affine::AffineLinearizeIndexOp::create(
+        rewriter, loc, indices, resultDims, /*disjoint=*/true);
+    auto delinearizeIndexOp = affine::AffineDelinearizeIndexOp::create(
+        rewriter, loc, linearizeIndexOp.getResult(), srcDims,
+        /*hasOuterBound=*/true);
+    return delinearizeIndexOp->getResults();
+  };
```
Max191: nit: Same here. You could create a shared transformation builder for reshapes.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+  return foldConsumerIntoMapGatherImpl(rewriter, extractSliceOp, mapGatherOp,
+                                       indexTransformBuilder);
+}
+
+/// Fold a consumer `padOp` into a producer `mapGatherOp`.
+/// Index transformation: source_idx = new_idx - low_pad
+/// Fill value is set to the pad value.
+static FailureOr<MapGatherOp> foldPadIntoMapGather(RewriterBase &rewriter,
```
Max191: This function needs to verify that there is not already padding with a different pad value, or else it will overwrite the old padding. For example:
```mlir
%pad0 = tensor.pad ... low[0] high[2] {
  %pad_val0 = arith.constant 0.0 : f32
  tensor.yield %pad_val0 : f32
}
%pad1 = tensor.pad %pad0 low[0] high[2] {
  %pad_val1 = arith.constant 1.0 : f32
  tensor.yield %pad_val1 : f32
}
```
This creates something like: `result tensor = [..., 0.0, 0.0, 1.0, 1.0]`.

To represent this with map_gather, we need to use non-constant (index dependent) padding values.

You don't need to support this case, but this function should at least fail if it runs into this case. Unfortunately, there isn't a nice way of detecting whether or not a map_gather already has padding, since it always carries a padding value, but there are a couple of options:

1. Make the padding value optional for map_gather. When there is no padding value, this means there is no padding needed in the gather.
2. For non-padded map_gather, you could return a `ub.poison` as the padding value. This would carry the same semantics as having no padding value, and it would eventually either be folded away (when the map_gather is lowered) or replaced by another padding value (while doing layout transformation combination).

Once we can represent non-padded map_gather ops, this transform should verify that the map_gather is not padded before doing the transformation.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
```diff
@@ -0,0 +1,387 @@
...
+    // Check that the load doesn't already have a map_gather user (avoid
+    // infinite loop).
+    bool hasMapGatherUser =
+        llvm::any_of(loadOp->getUsers(), [](Operation *user) {
+          return isa<IREE::LinalgExt::MapGatherOp>(user);
+        });
+    if (hasMapGatherUser) {
+      return failure();
+    }
+    (void)insertIdentityMapGather(rewriter, loadOp->getResult(0));
```
Max191: Can you add a control function like we have for map_scatter:
https://github.com/iree-org/iree/blob/fa6527609cccf190df6e78cb07a8c49899d4a304/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp#L576-L578

There will be cases where we want to prevent the transformation, even though it is possible, since map_gather is generally less efficient than some simple relayout ops if they can be handled by codegen.

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -188,6 +188,25 @@ def CombineLayoutTransformationPass :
...
+    Starting from `iree_codegen.load_from_buffer` ops, iteratively combine
+    consumer layout/indexing transformation ops (linalg.transpose,
+    tensor.collapse_shape, tensor.expand_shape, tensor.extract_slice,
+    tensor.pad, etc.) into a single iree_linalg_ext.map_gather operation.
+  }];
+  let dependentDialects = [
+    "affine::AffineDialect",
+    "arith::ArithDialect",
+    "iree_compiler::IREE::LinalgExt::IREELinalgExtDialect",
+    "scf::SCFDialect",
```
Max191: nit: I don't think we need scf dialect. It was there for map_scatter because we created loops to distribute padding, but we don't need to do it for map_gather (I don't see any scf ops being created anywhere in the PR).

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -347,6 +347,14 @@ def IREELinalgExt_MapGatherOp : IREELinalgExt_Op<"map_gather",
       OpBuilder &b, Location loc, ValueRange transformBodyIndices,
       function_ref<void(OpBuilder &, Location, ArrayRef<Value>)> bodyBuilder);
 
+    // Helper to apply transformations to the yielded source indices at the
+    // end of the transformation body. The `transformationBuilder` function
+    // receives the current yielded source indices (excluding padding) and
+    // should return the new source indices to yield.
+    void insertTransformationAtEnd(
+        OpBuilder &builder,
+        function_ref<SmallVector<Value>(ValueRange)> transformationBuilder);
```
Max191: This doesn't seem to be used anywhere. Can you remove it?

Max191 (commented):
Overall, looks good. I left one main comment about the tensor.pad handling, and then mostly nits.

One more high level question: Have you verified numerical correctness with any e2e examples using the CombineLayoutTransformationForMapGather pass yet?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
hanhanW: +1, it is a deadcode now.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
hanhanW: Eventually, they can belong the same file, so I'm okay with moving the implementation to the same file, as long as they are two separate passes for now. I'm not saying that they can't be a single pass, but we'll need to understand of the priorities when we do that -- E.g., looking at an op, should we fold it into map_scatter or map_gather. Example 2, what if we have map_gather -> map_scatter seq.

We don't need to answer all the questions and solve all the problems in one PR. So starting with a new pass sounds better to me.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
hanhanW: Just in case, since I asked for removing many code in my review. This is optional to me, becasue we are at the stage of building the transformation; it will happen when we integrate the pass into the pipeline.

I think it is better to change when we integrate the pass. Otherwise, it is dead code.

hanhanW (commented):
Mostly looks good. I was thinking if I should ask refactoring, or defer it when we're evalulating if we want to collapse two passes or not. Moving them to the same file and keep them two separate pass looks good to me.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
Abhishek-Varma: Done. Moved the implementations to one file, but keeping both as separate passes. Also renamed the passes.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
Abhishek-Varma: Done. Nice!

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformationForMapGather.cpp
Abhishek-Varma: Very good point! I chose the `ub.poison` approach while inserting an identity `map_gather` op at the result of `load_from_buffer`. Have added a lit test too. Can you take a look now?

Abhishek-Varma:
> One more high level question: Have you verified numerical correctness with any e2e examples using the CombineLayoutTransformationForMapGather pass yet?

Hi @Max191 - Yes, I did in fact check e2e correctness as reported [here](https://github.com/iree-org/iree/pull/23165#issuecomment-3822674149). I verified the repro's output as well as a standalone transpose op example.

And today I also checked a bunch of complex relayout op chains' e2e correctness too.

Here's what I did :-
1. Formed a complex chain relayout ops.
2. Applied the current pass.
3. Added the output to [e2e map_gather test already checked in](https://github.com/iree-org/iree/blob/main/tests/e2e/linalg_ext_ops/map_gather.mlir) along with corresponding `check.expect*` ops with what tensor values are expected.

The tests passed! :D

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -24,21 +25,19 @@
...
 // Preprocessing Utilities
 //===----------------------------------------------------------------------===//
 
-/// Convert complex ops into simpler ops by decomposing or raising to a named
-/// op.
-///  - `PackOp`s and `UnPackOp`s are decomposed.
-///  - Transpose `linalg::GenericOp`s are raised to `linalg::TransposeOp`s.
-static void simplifyComplexRelayoutOps(RewriterBase &rewriter,
-                                       FunctionOpInterface funcOp) {
+void simplifyComplexRelayoutOps(RewriterBase &rewriter,
```
hanhanW: We no longer need to expose this method to public, right?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -78,6 +77,36 @@ static void simplifyComplexRelayoutOps(RewriterBase &rewriter,
...
+          ArrayRef<BlockArgument> indices) -> SmallVector<Value> {
+        SmallVector<Value> indexValues(indices.begin(), indices.end());
+        auto linearizeIndexOp = affine::AffineLinearizeIndexOp::create(
+            rewriter, loc, indexValues, linearizeDims, /*disjoint=*/true);
+        auto delinearizeIndexOp = affine::AffineDelinearizeIndexOp::create(
+            rewriter, loc, linearizeIndexOp.getResult(), delinearizeDims,
+            /*hasOuterBound=*/true);
+        return delinearizeIndexOp->getResults();
+      };
+}
```
hanhanW: This is a very weird function. You're returning a lambda that captures `rewriter` by reference. Now all the callers need to understand it and it is quite dangerous.

I think we should inline the logic at call sites - it's only used in two places (foldReshapeIntoMapScatter and foldReshapeIntoMapGather). The code isn't that long. I don't see the value of refactoring it, because it makes nesting level deeper and readability worse.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -78,6 +77,36 @@ static void simplifyComplexRelayoutOps(RewriterBase &rewriter,
...
 
+/// Builds an index transformation that applies a permutation to the indices.
+/// Used by both map_scatter and map_gather transpose folding.
+static auto buildTransposeIndexTransform(ArrayRef<int64_t> perm) {
+  return [perm = SmallVector<int64_t>(perm)](
+             ArrayRef<BlockArgument> indices) -> SmallVector<Value> {
+    SmallVector<Value> indexValues(indices.begin(), indices.end());
+    return applyPermutation(indexValues, perm);
+  };
+}
```
hanhanW: Same here. This is a 4-line lambda. The "factory" function doesn't provide meaningful abstraction - it just adds indirection.

compiler/src/iree/compiler/Codegen/Common/Passes.td
hanhanW: We also need ub dialect now.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
```diff
@@ -0,0 +1,176 @@
...
+  %source = iree_codegen.load_from_buffer %buffer : memref<4x16xf32> -> tensor<4x16xf32>
+  %init = tensor.empty() : tensor<4x16xf32>
+  %copied = linalg.copy ins(%source : tensor<4x16xf32>) outs(%init : tensor<4x16xf32>) -> tensor<4x16xf32>
+  return %copied : tensor<4x16xf32>
+}
+// CHECK-LABEL: @fold_copy
+//  CHECK-SAME:   %[[BUFFER:[a-zA-Z0-9_]+]]
+//       CHECK:   %[[SOURCE:.+]] = iree_codegen.load_from_buffer %[[BUFFER]]
+//       CHECK:   linalg.copy
+//   CHECK-NOT:   iree_linalg_ext.map_gather
```
hanhanW: This is why I don't like canonicalization run on pass tests. Anyway, I think you need to update the case, so map_gather is not converted to a copy op. The test is not straightforward enough to the implementation, because you are testing more than we expect.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -798,5 +814,356 @@ struct CombineLayoutTransformationPass final
...
+    for (auto [idx, offset, stride] :
+         llvm::zip_equal(newOutputIndices, offsets, strides)) {
+      Value offsetValue =
+          getValueOrCreateConstantIndexOp(rewriter, loc, offset);
+      Value strideValue =
+          getValueOrCreateConstantIndexOp(rewriter, loc, stride);
+      // original_idx = offset + idx * stride
+      Value scaledIdx = arith::MulIOp::create(rewriter, loc, idx, strideValue);
+      Value originalIdx =
+          arith::AddIOp::create(rewriter, loc, offsetValue, scaledIdx);
```
hanhanW: Let's add `arith::IntegerOverflowFlags::nsw`, which enables the downstream folding for constants (e.g., in LLVM).

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -798,5 +814,356 @@ struct CombineLayoutTransformationPass final
...
+  if (!padValue) {
+    return rewriter.notifyMatchFailure(
+        padOp, "non-constant padding value is not supported");
+  }
+
+  // Check if the map_gather already has a real (non-poison) padding value.
+  // If so, we cannot safely replace it as that would lose the existing padding.
+  Block &transformBody = mapGatherOp.getTransformationRegion().front();
+  auto yieldOp = cast<IREE::LinalgExt::YieldOp>(transformBody.getTerminator());
+  Value currentPadValue = yieldOp.getOperand(yieldOp.getNumOperands() - 1);
```
hanhanW: I think it's worth having a `getPaddingValue()` method in `MapGather` op.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -798,5 +814,356 @@ struct CombineLayoutTransformationPass final
...
+  Location loc = padOp->getLoc();
+  SmallVector<OpFoldResult> lowPad = padOp.getMixedLowPad();
+
+  auto indexTransformBuilder =
+      [&](ArrayRef<BlockArgument> newOutputIndices) -> SmallVector<Value> {
+    SmallVector<Value> sourceIndices;
+    for (auto [idx, low] : llvm::zip_equal(newOutputIndices, lowPad)) {
+      Value lowValue = getValueOrCreateConstantIndexOp(rewriter, loc, low);
+      // source_idx = idx - low_pad
+      Value sourceIdx = arith::SubIOp::create(rewriter, loc, idx, lowValue);
```
hanhanW: ditto, let's add `arith::IntegerOverflowFlags::nsw`

hanhanW (commented):
Mostly looks good, just some nits. I think we over-abstract the "factor" functions.

hanhanW:
The PR title and description need to be updated.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Yeah. Updated.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
Abhishek-Varma: I understand. I've updated the test case.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/Passes.td
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -36,8 +36,8 @@ using IREE::LinalgExt::MapScatterOp;
 // Preprocessing Utilities
 //===----------------------------------------------------------------------===//
 
-void simplifyComplexRelayoutOps(RewriterBase &rewriter,
-                                FunctionOpInterface funcOp) {
+static void simplifyComplexRelayoutOps(RewriterBase &rewriter,
```
hanhanW, line 43: We need to add the comment back.

hanhanW (approved):
just one final nit. lgtm

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -798,5 +798,369 @@ struct CombineLayoutTransformationPass final
...
+/// 1. Create new dest tensor from consumer result shape
+/// 2. Clone map_gather with new dest and region
+/// 3. Apply index transformation via insertTransformationAtStart
+/// 4. Optionally update fill value
+/// 5. Replace consumer with new map_gather result
+static FailureOr<MapGatherOp> foldConsumerIntoMapGatherImpl(
+    RewriterBase &rewriter, Operation *consumerOp, MapGatherOp mapGatherOp,
+    function_ref<SmallVector<Value>(ArrayRef<BlockArgument>)>
+        indexTransformBuilder,
+    Value newFillValue = nullptr) {
```
Max191: nit: Make newFillValue `std::optional<Value>`. It makes it more obvious that it is optional.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -729,10 +730,12 @@ bool MapGatherOp::isIdentity() {
   if (getSourceType() != getOutputType()) {
     return false;
   }
+
   // Bail out on dynamic shapes.
   if (!getSourceType().hasStaticShape()) {
     return false;
   }
+
```
Max191: nit: I don't think the new lines here are needed.

Max191 (approved):
LGTM, nice work! Just a couple nits.

PR #23168, opened 2026-01-16
[CI] Update torch_ops config file and run tests when config files are modified

amd-eochoalo:
* Updates torch_ops configuration file to skip running some tests (new tests added without golden_value and a new failing that was not skipped).
* Adds a new rule to configure_ci.py to run torch tests whenever configuration files are modified. This is because otherwise one needs to remember to add ci-extra to test relevant tests. (onnx and sharktank are not included here since they are always run on pre-submit)


amd-eochoalo:
The relevant torch_ops  amdgpu_hip_gfx1201_O3 tests have passed.

kuhar (approved):

PR #23172, opened 2026-01-16
[AArch64] Disable breaking up transfer+extract into multiple scalar loads for AArch64

momchil-velikov:
Disable the transformation that breaks up vector loads followed by an extract into multiple scalar loads for the AArch64 target.

With this transformation disabled we keep the efficient vector loads while opening opportunities for using indexed forms of instructions, for example
https://developer.arm.com/documentation/ddi0602/2025-12/SIMD-FP-Instructions/FMLA--by-element---Floating-point-fused-multiply-add-to-accumulator--by-element--?lang=en

We have observed 8-12% performance improvements on various matrix multiplication workloads and GPT2 with Neon.

Signed-off-by: Momchil Velikov <momchil.velikov@arm.com>


hanhanW (approved):
There are already some legacy `isXXXArch` tech debt, so I'm fine with having one more. 
(plus I don't have a good suggestion atm, so it's fine to me.)

momchil-velikov:
@hanhanW  Hello, could you, please, commit this?

momchil-velikov:
Thanks!

PR #23175, opened 2026-01-16
[Codegen][GPU] Enable swizzling for scaled matmuls

Muzammiluddin-Syed-ECE:
This is the fourth of a series of PRs that together implement support in
IREE for XOR swizzling through the SwizzleHintOp.

There are four PRs that need to be merged:
1) Allow rank > 1 swizzle hint op operands and add a pass to flatten
swizzle hint allocs.
2) Add patterns which can fold reshapes and `extract_slice` ops into
empty ops through swizzle hint ops.
3) Add swizzle hint attribute to be set in `lowering_config` and
consumed in `GPUPromoteMatmulOperandsPass`.
4) Update `LLVMGPUSelectLoweringStrategy` Pass to set xor swizzles for
MXFP4 GEMMs.

This is PR 4, which does three things:
- Expresses the row width as a function of CacheLineSizeInBits and the element type of the chosen intrinsic's operands.
- Adds swizzle attribute to promotion type.
- Adds a test for the swizzle attribute which should have been added in PR 3.

We see an average 4.8% geomean improvement over top of main in the mxfp4 gemms tested. See [full data here](https://gist.github.com/Muzammiluddin-Syed-ECE/71c517206d89018f8706d661c94294b6/).

|shape|Feature_Throughput (TFLOps)|Top_of_main_Throughput(TFLOps)|improvement_ratio|improvement_percent|
|-----|---------------------------|------------------------------|-----------------|-------------------|
|1000_1024_8192_512|96.044057|86.408117|1.111517|11.151661|
|16300_1024_8192_512|1282.544470|1219.842766|1.051401|5.140146|
|500_16384_26624_1664|713.748860|723.841063|0.986057|-1.394257|
|53200_53248_8192_512|1194.983741|1107.453688|1.079037|7.903721|
|8100_16384_8192_512|1388.739043|1301.227179|1.067253|6.725333|
|8192_512_256_16|305.040291|301.612872|1.011364|1.136364|
|8192_53248_8192_512|1734.982869|1728.772840|1.003592|0.359216|
|...|...|...|...|...|
|OVERALL_GEOMEAN|||1.049751|4.975053|

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -941,10 +941,10 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
     // TODO(#23329): Do not swizzle shapes that have no bank conflicts.
     FailureOr<Attribute> lhsSwizzleAttr =
-        getXORShuffleAttr(context, defaultConfigAttr, target, kind,
+        getXorShuffleAttr(context, defaultConfigAttr, target, kind,
                           schedule->kTileSizes, kMMAOperandLhs);
     FailureOr<Attribute> rhsSwizzleAttr =
-        getXORShuffleAttr(context, defaultConfigAttr, target, kind,
+        getXorShuffleAttr(context, defaultConfigAttr, target, kind,
                           schedule->kTileSizes, kMMAOperandRhs);
     if (failed(lhsSwizzleAttr) || failed(rhsSwizzleAttr)) {
       promotionArray = {};
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -937,7 +937,6 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     // TODO(#22119): We don't use global load DMA for scaled matmuls, because
     // compilation doesn't support it. Once this is fixed, we should use global
     // load DMA here when possible.
-
     promotionList.append({2, 3});
     auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
     FailureOr<Attribute> lhsSwizzleAttr =
@@ -947,10 +946,11 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
         getXORShuffleAttr(context, defaultConfigAttr, target, kind,
                           schedule->kTileSizes, kMMAOperandRhs);
     if (failed(lhsSwizzleAttr) || failed(rhsSwizzleAttr)) {
-      return failure();
+      promotionArray = {};
+    } else {
+      promotionArray = {*lhsSwizzleAttr, *rhsSwizzleAttr, defaultConfigAttr,
+                        defaultConfigAttr};
     }
-    promotionArray = {*lhsSwizzleAttr, *rhsSwizzleAttr, defaultConfigAttr,
-                      defaultConfigAttr};
   }
   if ((!mustBeAligned || couldNeedPadding) && cPromoteIfPadding) {
     // If needed then add C operand which would be operand 2 or 4 for unscaled
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -941,8 +941,8 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
     // TODO(#23329): Do not swizzle shapes that have no bank conflicts.
     FailureOr<Attribute> lhsSwizzleAttr =
-    getXORShuffleAttr(context, defaultConfigAttr, target, kind,
-      schedule->kTileSizes, kMMAOperandLhs);
+        getXORShuffleAttr(context, defaultConfigAttr, target, kind,
+                          schedule->kTileSizes, kMMAOperandLhs);
     FailureOr<Attribute> rhsSwizzleAttr =
         getXORShuffleAttr(context, defaultConfigAttr, target, kind,
                           schedule->kTileSizes, kMMAOperandRhs);
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -937,8 +937,20 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     // TODO(#22119): We don't use global load DMA for scaled matmuls, because
     // compilation doesn't support it. Once this is fixed, we should use global
     // load DMA here when possible.
-    promotionArray = {};
+
     promotionList.append({2, 3});
+    auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
+    FailureOr<Attribute> lhsSwizzleAttr =
+        getXORShuffleAttr(context, defaultConfigAttr, target, kind,
+                          schedule->kTileSizes, kMMAOperandLhs);
+    FailureOr<Attribute> rhsSwizzleAttr =
+        getXORShuffleAttr(context, defaultConfigAttr, target, kind,
+                          schedule->kTileSizes, kMMAOperandRhs);
+    if (failed(lhsSwizzleAttr) || failed(rhsSwizzleAttr)) {
+      return failure();
+    }
+    promotionArray = {*lhsSwizzleAttr, *rhsSwizzleAttr, defaultConfigAttr,
+                      defaultConfigAttr};
   }
   if ((!mustBeAligned || couldNeedPadding) && cPromoteIfPadding) {
     // If needed then add C operand which would be operand 2 or 4 for unscaled
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -939,9 +939,10 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     // load DMA here when possible.
     promotionList.append({2, 3});
     auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
+    // TODO(#23329): Do not swizzle shapes that have no bank conflicts.
     FailureOr<Attribute> lhsSwizzleAttr =
-        getXORShuffleAttr(context, defaultConfigAttr, target, kind,
-                          schedule->kTileSizes, kMMAOperandLhs);
+    getXORShuffleAttr(context, defaultConfigAttr, target, kind,
+      schedule->kTileSizes, kMMAOperandLhs);
     FailureOr<Attribute> rhsSwizzleAttr =
         getXORShuffleAttr(context, defaultConfigAttr, target, kind,
                           schedule->kTileSizes, kMMAOperandRhs);
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -603,11 +603,11 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(IREE::GPU::createLowerIREEGPUOpsPass());
   funcPassManager.addPass(createUnrollAnnotatedLoopsPass());
   funcPassManager.addPass(createIREELoopInvariantCodeMotionPass());
-  // if (pipelineOptions.enableReduceSharedMemoryBankConflicts) {
-  //   GPUReduceBankConflictsPassOptions options = {};
-  //   options.paddingBits = 64;
-  //   funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
-  // }
+  if (pipelineOptions.enableReduceSharedMemoryBankConflicts) {
+    GPUReduceBankConflictsPassOptions options = {};
+    options.paddingBits = 64;
+    funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
+  }
   funcPassManager.addPass(createHoistStaticallyBoundAllocationsPass());
   if (forROCDL && pipelineOptions.prefetchNumStages >= 2) {
     funcPassManager.addPass(createFissionTransferOpsInControlFlowPass());
@@ -866,11 +866,11 @@ void addGPUVectorDistributePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(createCanonicalizerPass());
   funcPassManager.addPass(createCSEPass());
 
-  // if (options.enableReduceSharedMemoryBankConflicts) {
-  //   GPUReduceBankConflictsPassOptions options = {};
-  //   options.paddingBits = 64;
-  //   funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
-  // }
+  if (options.enableReduceSharedMemoryBankConflicts) {
+    GPUReduceBankConflictsPassOptions options = {};
+    options.paddingBits = 64;
+    funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
+  }
   if (forROCDL && options.prefetchNumStages >= 2) {
     ROCDLPrefetchSharedMemoryPassOptions prefetchOpts;
     prefetchOpts.numStages = options.prefetchNumStages;
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -603,11 +603,11 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(IREE::GPU::createLowerIREEGPUOpsPass());
   funcPassManager.addPass(createUnrollAnnotatedLoopsPass());
   funcPassManager.addPass(createIREELoopInvariantCodeMotionPass());
-  if (pipelineOptions.enableReduceSharedMemoryBankConflicts) {
-    GPUReduceBankConflictsPassOptions options = {};
-    options.paddingBits = 64;
-    funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
-  }
+  // if (pipelineOptions.enableReduceSharedMemoryBankConflicts) {
+  //   GPUReduceBankConflictsPassOptions options = {};
+  //   options.paddingBits = 64;
+  //   funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
+  // }
   funcPassManager.addPass(createHoistStaticallyBoundAllocationsPass());
   if (forROCDL && pipelineOptions.prefetchNumStages >= 2) {
     funcPassManager.addPass(createFissionTransferOpsInControlFlowPass());
@@ -866,11 +866,11 @@ void addGPUVectorDistributePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(createCanonicalizerPass());
   funcPassManager.addPass(createCSEPass());
 
-  if (options.enableReduceSharedMemoryBankConflicts) {
-    GPUReduceBankConflictsPassOptions options = {};
-    options.paddingBits = 64;
-    funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
-  }
+  // if (options.enableReduceSharedMemoryBankConflicts) {
+  //   GPUReduceBankConflictsPassOptions options = {};
+  //   options.paddingBits = 64;
+  //   funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
+  // }
   if (forROCDL && options.prefetchNumStages >= 2) {
     ROCDLPrefetchSharedMemoryPassOptions prefetchOpts;
     prefetchOpts.numStages = options.prefetchNumStages;
--- a/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
@@ -579,7 +579,9 @@ Value getCombiningIdentityValue(Location loc, OpBuilder &builder,
 gpu::AllReduceOperation combiningKindToAllReduce(vector::CombiningKind kind) {
   switch (kind) {
 #define MAP_CASE(X)                                                            \
+  \ 
   case vector::CombiningKind::X:                                               \
+  \ 
     return gpu::AllReduceOperation::X
 
     MAP_CASE(ADD);
@@ -800,8 +802,43 @@ getTotalTileElems(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
          llvm::product_of(layout.element);
 }
 
-// Disabling clang-tidy for the following function, as it will be externally
-// linked to the CAPI in a future PR. 
+// TODO(muzasyed): Add more intrinsics for gfx950.
+static FailureOr<std::pair<int64_t, int64_t>> getXorShuffleParamsForGfx950(
+    IREE::GPU::TargetAttr target,
+    IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
+  if (auto smma = dyn_cast<IREE::GPU::ScaledMMAAttr>(intrinsic)) {
+    switch (smma.getIntrinsic()) {
+    case IREE::GPU::ScaledMMAIntrinsic::MFMA_SCALE_F32_16x16x128_B32:
+      return std::pair(/*row_width*/ int64_t(256),
+                       /*access_width*/ int64_t(32));
+    default:
+      return failure();
+    }
+  }
+  return failure();
+}
+
+/// Validate the XOR shuffle parameters for the given intrinsic and operand
+/// index. If the parameters produce an XOR Shuffle that is not valid, return
+/// failure. Else, return the swizzle parameters.
+static FailureOr<std::pair<int64_t, int64_t>>
+validateXorShuffle(FailureOr<std::pair<int64_t, int64_t>> swizzle,
+                   IREE::Codegen::InnerTileDescAttrInterface intrinsic,
+                   int operandIndex) {
+  FailureOr<int64_t> maybeTotalTileElems =
+      getTotalTileElems(intrinsic, operandIndex);
+  if (failed(maybeTotalTileElems)) {
+    return failure();
+  }
+  int64_t totalTileElems = *maybeTotalTileElems;
+  if (!isXORShuffleValid(swizzle->first, swizzle->second, totalTileElems)) {
+    return failure();
+  }
+  return swizzle;
+}
+
+// Disabling clang-tidy for the following functions, as it will be externally
+// linked to the CAPI in a future PR.
 // NOLINTBEGIN
 FailureOr<std::pair<int64_t, int64_t>>
 getXorShuffleBounds(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
@@ -815,7 +852,6 @@ getXorShuffleBounds(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
   }
   return std::pair(*maybeMinimumAccessElems, *maybeTotalTileElems);
 }
-// NOLINTEND
 
 bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
                        int64_t totalTileElems) {
@@ -853,41 +889,6 @@ bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
     return false;
   }
   return true;
-}
-
-/// Validate the XOR shuffle parameters for the given intrinsic and operand
-/// index. If the parameters produce an XOR Shuffle that is not valid, return
-/// failure. Else, return the swizzle parameters.
-static FailureOr<std::pair<int64_t, int64_t>>
-validateXorShuffle(FailureOr<std::pair<int64_t, int64_t>> swizzle,
-                   IREE::Codegen::InnerTileDescAttrInterface intrinsic,
-                   int operandIndex) {
-  FailureOr<int64_t> maybeTotalTileElems =
-      getTotalTileElems(intrinsic, operandIndex);
-  if (failed(maybeTotalTileElems)) {
-    return failure();
-  }
-  int64_t totalTileElems = *maybeTotalTileElems;
-  if (!isXORShuffleValid(swizzle->first, swizzle->second, totalTileElems)) {
-    return failure();
-  }
-  return swizzle;
-}
-
-// TODO(muzasyed): Add more intrinsics for gfx950.
-static FailureOr<std::pair<int64_t, int64_t>> getXorShuffleParamsForGfx950(
-    IREE::GPU::TargetAttr target,
-    IREE::Codegen::InnerTileDescAttrInterface intrinsic, int operandIndex) {
-  if (auto smma = dyn_cast<IREE::GPU::ScaledMMAAttr>(intrinsic)) {
-    switch (smma.getIntrinsic()) {
-    case IREE::GPU::ScaledMMAIntrinsic::MFMA_SCALE_F32_16x16x128_B32:
-      return std::pair(/*row_width*/ int64_t(256),
-                       /*access_width*/ int64_t(32));
-    default:
-      return failure();
-    }
-  }
-  return failure();
 }
 
 FailureOr<std::pair<int64_t, int64_t>> getXorShuffleParamsForTunedChipset(
@@ -959,6 +960,7 @@ getXorShuffleParams(IREE::GPU::TargetAttr target,
   }
   return xorShuffleAttr;
 }
+// NOLINTEND
 
 FailureOr<Attribute>
 getXorShuffleAttr(MLIRContext *context, Attribute baseConfigAttr,
@@ -975,9 +977,8 @@ getXorShuffleAttr(MLIRContext *context, Attribute baseConfigAttr,
       context, effectiveRowWidth, numAccessElems,
       /*row_stride=*/int64_t(0),
       /*per_phase=*/int64_t(0));
-  Attribute swizzleOperand =
-      IREE::GPU::SwizzleOperandAttr::get(context, baseConfigAttr, swizzleAttr);
-  return swizzleOperand;
+  return IREE::GPU::SwizzleOperandAttr::get(context, baseConfigAttr,
+                                            swizzleAttr);
 }
 
 //===----------------------------------------------------------------------===//
--- a/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
@@ -828,24 +828,38 @@ bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
   // The number of total tile elements we want to swizzle must be greater than
   // or equal to the number of elements in a row.
   if (totalTileElems < numRowElems) {
+    LDBG() << "The number of row elements exceed the total elements in the tile"
+           << "\n";
     return false;
   }
   // We need at least one column of access elements to swizzle within a row.
-  if (numAccessElems < numRowElems) {
+  if (numRowElems < numAccessElems) {
+    LDBG()
+        << "The number of access elements exceed the total elements in the row"
+        << "\n";
     return false;
   }
   // The size of a row must evenly divide the total number of tile elements.
   if (totalTileElems % numRowElems != 0) {
+    LDBG() << "The number of row elements does not evenly divide the total "
+              "elements in the tile"
+           << "\n";
     return false;
   }
   // The number of access elements must evenly divide the number of row
   // elements.
   if (numRowElems % numAccessElems != 0) {
+    LDBG() << "The number of access elements does not evenly divide the number "
+              "of row elements"
+           << "\n";
     return false;
   }
   // The number of columns in a row must evenly divide the total number of tile
   // elements. This is to avoid incomplete rows at the end of the tile.
-  if (((numRowElems / numAccessElems) % totalTileElems) != 0) {
+  if (totalTileElems % (numRowElems / numAccessElems) != 0) {
+    LDBG() << "The number of columns in a row does not evenly divide the total "
+              "elements in the tile"
+           << "\n";
     return false;
   }
   return true;
--- a/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
@@ -839,7 +839,7 @@ validateXorShuffle(FailureOr<std::pair<int64_t, int64_t>> swizzle,
 
 // Disabling clang-tidy for the following functions, as it will be externally
 // linked to the CAPI in a future PR.
-// NOLINTBEGIN
+// NOLINTBEGIN(misc-use-internal-linkage)
 FailureOr<std::pair<int64_t, int64_t>>
 getXorShuffleBounds(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
                     int operandIndex) {
@@ -960,7 +960,7 @@ getXorShuffleParams(IREE::GPU::TargetAttr target,
   }
   return xorShuffleAttr;
 }
-// NOLINTEND
+// NOLINTEND(misc-use-internal-linkage)
 
 FailureOr<Attribute>
 getXorShuffleAttr(MLIRContext *context, Attribute baseConfigAttr,
--- a/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
@@ -764,18 +764,17 @@ static FailureOr<int64_t>
 getKSize(IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
   SmallVector<VectorType> undistributedTypes;
   intrinsic.getUndistributedTileTypes(undistributedTypes);
-  if (undistributedTypes.empty()) {
-    return failure();
-  }
-  // For matmul-like operations (C += A * B), the K dimension is the last
-  // dimension of the LHS operand (operand 0).
-  ArrayRef<int64_t> lhsShape = undistributedTypes[0].getShape();
-  if (lhsShape.size() < 2) {
-    return failure();
-  }
-  // Return the product of all reduction dimensions (everything after the M
-  // dim).
-  return llvm::product_of(lhsShape.drop_front(1));
+  auto newIndices =
+          llvm::to_vector(llvm::map_range(
+              zip_equal(intrinsic.getOperandIteratorTypes(), undistributedTypes),
+              [&](auto pair) { 
+                utils::IteratorType iteratorType = std::get<0>(pair)
+                if (iteratorType == utils::IteratorType::reduction) {
+                  return std::get<1>(pair).getShape().size();
+                }
+                return 1;
+              }));
+  return llvm::product_of(newIndices[0]);
 }
 
 /// Returns the number of elements each thread accesses for the given intrinsic
--- a/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
@@ -749,75 +749,62 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
 static FailureOr<int64_t>
 getOperandBitwidth(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
                    int operandIndex) {
-  if (auto smma = dyn_cast<IREE::GPU::ScaledMMAAttr>(intrinsic)) {
-    SmallVector<Type> elementTypes;
-    smma.getElementTypes(elementTypes);
-    return elementTypes[operandIndex].getIntOrFloatBitWidth();
+  SmallVector<Type> elementTypes;
+  intrinsic.getElementTypes(elementTypes);
+  if (operandIndex < 0 || operandIndex >= elementTypes.size()) {
+    return failure();
   }
-  if (auto mma = dyn_cast<IREE::GPU::MMAAttr>(intrinsic)) {
-    auto [aType, bType, _] = mma.getABCElementTypes();
-    return operandIndex == IREE::GPU::kMMAOperandLhs
-               ? aType.getIntOrFloatBitWidth()
-               : bType.getIntOrFloatBitWidth();
-  }
-  return failure();
+  return elementTypes[operandIndex].getIntOrFloatBitWidth();
 }
 
-/// Returns the size of the total K dimension for the given intrinsic.
-/// For intrinsics with multiple K dimensions, returns the product of all K
-/// dimensions.
+/// Returns the size of the total K dimension (reduction dimension) for the
+/// given intrinsic. For intrinsics with multiple K dimensions, returns the
+/// product of all K dimensions. This is determined by looking at all reduction
+/// dimensions in the undistributed tile shape.
 /// Note this is for a single intrinsic.
 static FailureOr<int64_t>
 getKSize(IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
-  if (auto smma = dyn_cast<IREE::GPU::ScaledMMAAttr>(intrinsic)) {
-    return getKSize(smma.getIntrinsic()) * getKbSize(smma.getIntrinsic());
+  SmallVector<VectorType> undistributedTypes;
+  intrinsic.getUndistributedTileTypes(undistributedTypes);
+  if (undistributedTypes.empty()) {
+    return failure();
   }
-  if (auto mma = dyn_cast<IREE::GPU::MMAAttr>(intrinsic)) {
-    return getKSize(mma.getIntrinsic());
+  // For matmul-like operations (C += A * B), the K dimension is the last
+  // dimension of the LHS operand (operand 0).
+  ArrayRef<int64_t> lhsShape = undistributedTypes[0].getShape();
+  if (lhsShape.size() < 2) {
+    return failure();
   }
-  return failure();
+  // Return the product of all reduction dimensions (everything after the M
+  // dim).
+  return llvm::product_of(lhsShape.drop_front(1));
 }
 
 /// Returns the number of elements each thread accesses for the given intrinsic
-/// and operand index.
+/// and operand index. This is determined by the `element` field of the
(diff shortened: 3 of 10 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -914,8 +914,28 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
     promotionList.append({2, 3});
+    // The row width is the number of elements across which we want to swizzle
+    // groups over.To avoid bank conflicts, we want to set the row width to be
+    // the number of elements that fill all the cache lines.
+    auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
+    int64_t lhsBitwidth = lhsElemType.getIntOrFloatBitWidth();
+    int64_t rhsBitwidth = rhsElemType.getIntOrFloatBitWidth();
+    int64_t lhsRowWidth = kCacheLineSizeBits / lhsBitwidth;
+    int64_t rhsRowWidth = kCacheLineSizeBits / rhsBitwidth;
+    int64_t accessWidth = schedule->kSizes.back();
```
kuhar: What is the unit? I thought that kSizes has the number of elements, not bits?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -956,6 +976,7 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   }
   auto configDict = DictionaryAttr::get(context, attrs);
   auto loweringConfig = IREE::GPU::LoweringConfigAttr::get(context, configDict);
+  llvm::errs() << "loweringConfig: " << loweringConfig << "\n";
```
kuhar: drop this debug print

kuhar (commented):
What's the speedup after we land this?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -914,8 +914,28 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
     // compilation doesn't support it. Once this is fixed, we should use global
     // load DMA here when possible.
-    promotionArray = {};
     promotionList.append({2, 3});
+    // The row width is the number of elements across which we want to swizzle
+    // groups over.To avoid bank conflicts, we want to set the row width to be
+    // the number of elements that fill all the cache lines.
+    auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
+    int64_t lhsBitwidth = lhsElemType.getIntOrFloatBitWidth();
+    int64_t rhsBitwidth = rhsElemType.getIntOrFloatBitWidth();
```
kuhar: Are we always accessing a single element from lhs and rhs?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -914,8 +914,28 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
+    int64_t lhsRowWidth = kCacheLineSizeBits / lhsBitwidth;
+    int64_t rhsRowWidth = kCacheLineSizeBits / rhsBitwidth;
```
kuhar: LDS != cache. This seems like the wrong constant to use even if the value will end up being the same.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: Right both access width and row width take number of elements, just like kSizes.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: No, we are accessing 32 elements at a time as described by the access width. Perhaps I don't understand the question.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: Will add a new constant.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
kuhar: I don't understand your reply. Can you update variable names to clearly indicate if the unit is bits or elements?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
kuhar: It shouldn't be a constant, it should be based on the target env. We have something similar already added to pad encoding.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -914,8 +914,28 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     // TODO(#22119): We don't use global load DMA for scaled matmuls, because
     // compilation doesn't support it. Once this is fixed, we should use global
     // load DMA here when possible.
-    promotionArray = {};
     promotionList.append({2, 3});
+    // The row width is the number of elements across which we want to swizzle
```
kuhar: If it's elements call the related variables something like `numRowElems` and `numAccessElems`. To me width is usually means bits/bytes.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: Done

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -44,6 +44,7 @@ namespace mlir::iree_compiler::IREE::GPU {
 
 constexpr int64_t kCacheLineSizeBits = 128 * 8;
 constexpr int64_t kPreferredCopyNumBits = 128;
+constexpr int64_t kLDSBankWidthBits = 32 * 4 * 8;
```
kuhar: This needs to be a target env attribute. For example mi300 has 32 banks while mi355 has 64 banks.

kuhar (changes requested):

Muzammiluddin-Syed-ECE:
> What's the speedup after we land this?

Roughly a Geomean speedup of %4.4 across shapes that previously had bank conflicts. Those that had no bank conflicts to begin with were unchanged.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -914,8 +915,28 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
     promotionList.append({2, 3});
+    // The row width is the number of elements across which we want to swizzle
+    // groups over.To avoid bank conflicts, we want to set the row width to be
+    // the number of elements that fill all the cache lines.
+    auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
+    int64_t lhsBitwidth = lhsElemType.getIntOrFloatBitWidth();
+    int64_t rhsBitwidth = rhsElemType.getIntOrFloatBitWidth();
+    int64_t lhsNumRowElems = kLDSBankWidthBits / lhsBitwidth;
+    int64_t rhsNumRowElems = kLDSBankWidthBits / rhsBitwidth;
+    int64_t numAccessElems = schedule->kSizes.back();
```
krzysz00: Is that the right number here? We want "the number of elements we'll be getting each time we go to LDS", or, probably equivalently, "the number of elements we'll be feeding to one intrinsic"

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -914,8 +915,28 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
+        /*per_phase=*/int64_t(0));
+    auto rhsSwizzleAttr = IREE::Codegen::XORShuffleAttr::get(
+        context, rhsNumRowElems, numAccessElems, /*row_stride=*/int64_t(0),
+        /*per_phase=*/int64_t(0));
+    Attribute lhsSwizzleOperand = IREE::GPU::SwizzleOperandAttr::get(
+        context, defaultConfigAttr, lhsSwizzleAttr);
+    Attribute rhsSwizzleOperand = IREE::GPU::SwizzleOperandAttr::get(
+        context, defaultConfigAttr, rhsSwizzleAttr);
+    promotionArray = {lhsSwizzleOperand, rhsSwizzleOperand, defaultConfigAttr,
+                      defaultConfigAttr};
```
krzysz00: This code needs to be extracted into a form the tuner can call, I claim. @bangtianliu for thoughts on what that could look like.

bangtianliu: Currently, we already have `pipeline_options` support in the tuner, as shown in the following code: https://github.com/nod-ai/amd-shark-ai/blob/main/amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py#L678-L685.

This is how it looks like in the test:

``` python
def test_gpu_pipeline_options(tuner_ctx: common.TunerContext) -> None:
    options = iree_gpu.PipelineOptionsAttr.get()
    assert str(options) == "#iree_gpu.pipeline_options<>"

    options = iree_gpu.PipelineOptionsAttr.get(prefetch_num_stages=2)
    assert str(options) == "#iree_gpu.pipeline_options<prefetch_num_stages = 2>"

    options = iree_gpu.PipelineOptionsAttr.get(
        prefetch_num_stages=2, no_reduce_shared_memory_bank_conflicts=False
    )
    assert (
        str(options)
        == "#iree_gpu.pipeline_options<prefetch_num_stages = 2, no_reduce_shared_memory_bank_conflicts = false>"
    )

    options = iree_gpu.PipelineOptionsAttr.get(
        reorder_workgroups_strategy=iree_gpu.ReorderWorkgroupsStrategyAttr.get(
            iree_gpu.ReorderWorkgroupsStrategy.Transpose
        )
    )
    assert (
        str(options)
        == "#iree_gpu.pipeline_options<reorder_workgroups_strategy = <Transpose>>"
    )
```

Given this, we can extract the swizzle operand construction logic into a reusable utility function via IREE’s Python bindings, similar to how we already expose other GPU attributes such as `MMAAttr `and `PipelineOptionsAttr`.

Since the target information is already available on the tuner side, once the Python binding is exposed, we should be able to support tuning workflows like the following::

```python
        pipeline_options_search_space = rocm_dispatch_constraints.PipelineOptionsSearchSpace(
            prefetch_num_stages=args.prefetch_num_stages_options,
            no_reduce_shared_memory_bank_conflicts=args.no_reduce_shared_memory_bank_conflicts_options,
        )
```

This does not necessarily have to go through pipeline_options specifically. However, we can follow the same general approach to enumerate all possible configurations, regardless of whether the constraints come from the target architecture or from the computation itself.

bangtianliu: > This code needs to be extracted into a form the tuner can call, I claim. 

For this Q itself,  this swizzle operand construction code should be extracted into a helper function. That way, when we expose it through Python bindings, we can call that function directly. Here is one example:
https://github.com/iree-org/iree/blob/8b6f3392ebee1275250a139baa3fd4380647b1d6/compiler/src/iree/compiler/API/Internal/IREECodegenDialectCAPI.cpp#L209-L222

kuhar: We can land bindings separately to keep this PR small

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: As discussed ill add a comment as a warning that details that for these scaled matmul intrinsics, the access element number is equal to the last element of the ksizes, specifically only in these intrinsics and does not apply generally.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: Added functions that can be used in this future PR.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: Updated.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+           llvm::product_of(layout.element);
+  }
+  if (auto mma = dyn_cast<IREE::GPU::MMAAttr>(intrinsic)) {
+    layout =
+        IREE::GPU::getSingleSubgroupLayout(mma.getIntrinsic(), operandIndex);
+    return llvm::product_of(layout.outer) * llvm::product_of(layout.thread) *
+           llvm::product_of(layout.element);
+  }
+  return failure();
+}
```
Muzammiluddin-Syed-ECE, line 816: These would be simpler if the `getIntrinsic()` method was part of the `InnerTileDescAttrInterface`. The only thing preventing this is the fact that `VirtualMMAIntrinsic` doesnt have an intrinsic paramter of type `MMAIntrinsicType`.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  }
+  auto [effectiveRowWidth, numAccessElems] = xorShuffleParams.value();
+  auto swizzleAttr = IREE::Codegen::XORShuffleAttr::get(
+      context, effectiveRowWidth, numAccessElems,
+      /*row_stride=*/int64_t(0),
+      /*per_phase=*/int64_t(0));
+  Attribute swizzleOperand =
+      IREE::GPU::SwizzleOperandAttr::get(context, baseConfigAttr, swizzleAttr);
+  return swizzleOperand;
+}
```
Muzammiluddin-Syed-ECE: Should I move all these to a new file called SwizzleUtils.cpp

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  // Figure out how many elements can fit across all banks of LDS.
+  IREE::GPU::TargetWgpAttr wgp = target.getWgp();
+  FailureOr<int64_t> bitwidth = getOperandBitwidth(intrinsic, operandIndex);
+  if (failed(bitwidth)) {
+    return failure();
+  }
+  // Assumptions:
+  // - Each bank is 4 bytes wide (32 bits).
+  // - If not specified, assume 64 banks.
+  constexpr int64_t defaultBankWidthBits = 32;
```
Muzammiluddin-Syed-ECE: We talked about listing the assumptions but is this sufficient or would a more visible place be better

Muzammiluddin-Syed-ECE:
> > What's the speedup after we land this?
> 
> Roughly a Geomean speedup of %4.4 across shapes that previously had bank conflicts. Those that had no bank conflicts to begin with were unchanged.

Here's a gist showing the perf against most recent top of main: [data](https://gist.github.com/Muzammiluddin-Syed-ECE/71c517206d89018f8706d661c94294b6).

kuhar:
> > > What's the speedup after we land this?
> > 
> > 
> > Roughly a Geomean speedup of %4.4 across shapes that previously had bank conflicts. Those that had no bank conflicts to begin with were unchanged.
> 
> Here's a gist showing the perf against most recent top of main: [data](https://gist.github.com/Muzammiluddin-Syed-ECE/71c517206d89018f8706d661c94294b6).

I don't know how to read this data -- there are no units and it's unclear if positive change means improvement or regression.

Do you have data about bank conflicts? (Doesn't have to be exhaustive, just sample a dozen random shapes and see if the bank conflict counters align with the observed delta).

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -850,11 +850,11 @@ void addGPUVectorDistributePassPipeline(OpPassManager &funcPassManager,
...
-  if (options.enableReduceSharedMemoryBankConflicts) {
-    GPUReduceBankConflictsPassOptions options = {};
-    options.paddingBits = 64;
-    funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
-  }
+  // if (options.enableReduceSharedMemoryBankConflicts) {
+  //   GPUReduceBankConflictsPassOptions options = {};
+  //   options.paddingBits = 64;
+  //   funcPassManager.addPass(createGPUReduceBankConflictsPass(options));
+  // }
```
kuhar: Seems like this is still work in progress?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+    auto [aType, bType, _] = mma.getABCElementTypes();
+    return operandIndex == IREE::GPU::kMMAOperandLhs
+               ? aType.getIntOrFloatBitWidth()
+               : bType.getIntOrFloatBitWidth();
+  }
+  return failure();
+}
+
+static FailureOr<int64_t>
+getKSize(IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
```
kuhar, line 763: Can you add a comment explaining what this returns in both cases?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  }
+  if (auto mma = dyn_cast<IREE::GPU::MMAAttr>(intrinsic)) {
+    return getKSize(mma.getIntrinsic());
+  }
+  return failure();
+}
+
+static FailureOr<int64_t>
+getNumAccessElems(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
+                  int operandIndex) {
```
kuhar, line 784: Is this per thread or in total?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+    layout =
+        IREE::GPU::getSingleSubgroupLayout(mma.getIntrinsic(), operandIndex);
+    return llvm::product_of(layout.outer) * llvm::product_of(layout.thread) *
+           llvm::product_of(layout.element);
+  }
+  return failure();
+}
+
+FailureOr<std::pair<int64_t, int64_t>>
+getXORShuffleBounds(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
```
kuhar: clang-tidy warnings in a few places below

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+                    int operandIndex) {
+  FailureOr<int64_t> maybeMinimumAccessElems =
+      getNumAccessElems(intrinsic, operandIndex);
+  FailureOr<int64_t> maybeTotalTileElems =
+      getTotalTileElems(intrinsic, operandIndex);
+  if (failed(maybeMinimumAccessElems) || failed(maybeTotalTileElems)) {
+    return failure();
+  }
+  return std::make_pair(*maybeMinimumAccessElems, *maybeTotalTileElems);
```
kuhar:
```suggestion
  return std::pair(*maybeMinimumAccessElems, *maybeTotalTileElems);
```

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+}
+
+bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
+                       int64_t totalTileElems) {
+  // The number of total tile elements we want to swizzle must be greater than
+  // or equal to the number of elements in a row.
+  if (totalTileElems < numRowElems) {
+    LDBG() << "The number of row elements exceed the total elements in the tile"
+           << "\n";
```
kuhar: Can you pre-commit cleanup to this file and switch to DebugLog.h?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+    LDBG() << "The number of columns in a row does not evenly divide the total "
+              "elements in the tile"
+           << "\n";
+    return false;
+  }
+  return true;
+}
+
+static FailureOr<std::pair<int64_t, int64_t>>
+validatedShuffle(FailureOr<std::pair<int64_t, int64_t>> swizzle,
```
kuhar: Add a comment explaining what it means for a shuffle to be valid and what the result is.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+    ArrayRef<int64_t> reductionTileSizes, int operandIndex) {
+  // Compute XOR shuffle swizzle parameters for bank conflict avoidance.
+  // - row_width: Select entirety of K Tile size, may not prevent bank
+  //              conflicts if the K tile size is too small.
+  // - access_width: number of contiguous elements each thread accesses,
+  //                 derived from the MMA intrinsic's element layout.
+  int64_t numAccessElems = getNumAccessElems(intrinsic, operandIndex).value();
+
+  // Calculate K tile size (total K elements in shared memory) to use as row
+  // width. For small K tiles, this may not reduce bank conflicts effectively.
```
kuhar, line 920: Don't we have a helper for this above?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  int64_t kTileSize =
+      llvm::product_of(reductionTileSizes) * getKSize(intrinsic).value();
+  int64_t kTilePow2 = llvm::PowerOf2Ceil(kTileSize);
```
kuhar: why do we need to calculate this?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,254 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  IREE::GPU::TargetWgpAttr wgp = target.getWgp();
+  FailureOr<int64_t> bitwidth = getOperandBitwidth(intrinsic, operandIndex);
+  if (failed(bitwidth)) {
+    return failure();
+  }
+  // Assumptions:
+  // - Each bank is 4 bytes wide (32 bits).
+  // - If not specified, assume 64 banks.
+  constexpr int64_t defaultBankWidthBits = 32;
+  constexpr int64_t defaultBankCount = 64;
```
kuhar: I'd say that if we query bank count from the target env and it's not known, we should bail out instead and not swizzle at all. For example nvidia hw has 32 banks.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
Muzammiluddin-Syed-ECE: oh sorry that was from a bad rebasing last night

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: will add a comment explaining

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: If i understand correctly its asking for external linkage, but this is used in ConfigUtils.cpp so I dont understand why the error persists.

Muzammiluddin-Syed-ECE:
> 
> I don't know how to read this data -- there are no units and it's unclear if positive change means improvement or regression.
> 
> Do you have data about bank conflicts? (Doesn't have to be exhaustive, just sample a dozen random shapes and see if the bank conflict counters align with the observed delta).

Sorry I shouldve added to the file that the units were throughput so TFLOPs (higher better).

Also here is some data about bank conflicts on top of main. 
https://gist.github.com/Muzammiluddin-Syed-ECE/88e8d52c4a9ecdf9f28c824c3388c937 .

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: The helper is for a single intrinsic whereas this is the number of K elements processed in a subgroup tile.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: Oh right i guess this would already be a power of 2. Will remove.

Muzammiluddin-Syed-ECE: The logic was that because the total number of elements is always a power of 2, we need row width to also be a power of 2 for it to divide the total number of elements evenly.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: added in `GPUUtils.h`.

Muzammiluddin-Syed-ECE:
# Summary of improvements

From the [collected data](https://gist.github.com/Muzammiluddin-Syed-ECE/88e8d52c4a9ecdf9f28c824c3388c937) we see that there is on average a 4.8% geomean improvement in throughput (TFLOps) across various shapes of MXFP4 gemms when XOR swizzling is enabled. 

image


We can see that for several keys shapes in llama 405b there were an unacceptable number of cycles lost to bank conflicts (which is what this Rocprof counter `SQ_LDS_BANK_CONFLICT` measures). These are completely removed in this PR.

|shape (M_N_Ko_Kb)|SQ_LDS_BANK_CONFLICT (ToM)| SQ_LDS_BANK_CONFLICT (Wtih XOR Swizzle)|
|-----|--------------------------|-------------------------------|
|512_16384_26624_1664|20447232.0|0.0|
|1024_16384_8192_512|12582912.0|0.0|
|53248_1024_8192_512|40894464.0|0.0|0|
|1024_16384_26624_1664|40894464.0|0.
|8192_53248_8192_512|327155712.0|0.0|
|16300_53248_8192_512|654311424.0|0.0|
|53248_16384_26624_1664|2126512128.0|0.0|

This is a truncated list of all the shapes tested, see more [here](https://gist.github.com/Muzammiluddin-Syed-ECE/88e8d52c4a9ecdf9f28c824c3388c937).

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -937,8 +935,21 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     // TODO(#22119): We don't use global load DMA for scaled matmuls, because
     // compilation doesn't support it. Once this is fixed, we should use global
     // load DMA here when possible.
-    promotionArray = {};
     promotionList.append({2, 3});
+    auto defaultConfigAttr = IREE::GPU::DerivedThreadConfigAttr::get(context);
+    // TODO(#23329): Do not swizzle shapes that have no bank conflicts.
+    FailureOr<Attribute> lhsSwizzleAttr =
+        getXORShuffleAttr(context, defaultConfigAttr, target, kind,
```
krzysz00: My brain wants this called `getXorShuffleAttr` but that's not actually important

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,251 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
   return std::nullopt;
 }
 
+static FailureOr<int64_t>
+getOperandBitwidth(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
```
krzysz00, line 750: ... are we sure there isn't a generic way to write this? I could've sworn there's an interface method that'll basically give you these element types

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,251 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+/// Note this is for a single intrinsic.
+static FailureOr<int64_t>
+getKSize(IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
+  if (auto smma = dyn_cast<IREE::GPU::ScaledMMAAttr>(intrinsic)) {
+    return getKSize(smma.getIntrinsic()) * getKbSize(smma.getIntrinsic());
+  }
+  if (auto mma = dyn_cast<IREE::GPU::MMAAttr>(intrinsic)) {
+    return getKSize(mma.getIntrinsic());
+  }
+  return failure();
```
krzysz00: This might be something you can determine generically by looking at all the reduction dimensions, if I've remembered what an inner tile descriptor lets you look up?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,251 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  if (auto mma = dyn_cast<IREE::GPU::MMAAttr>(intrinsic)) {
+    layout =
+        IREE::GPU::getSingleSubgroupLayout(mma.getIntrinsic(), operandIndex);
+    return llvm::product_of(layout.element);
+  }
+  return failure();
+}
+
+static FailureOr<int64_t>
+getTotalTileElems(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
```
krzysz00, line 794: I think this can be read off of something like getDistributedShape()

I'm trying to prevent having casing for [scaled]mmaAttr so that this all just works for, say, the skinny gemm stuff. Or that virtual mma we use in attention.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: done

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: done

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: done

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: done

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
 }
 
+static FailureOr<int64_t>
+getOperandBitwidth(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
+                   int operandIndex) {
+  SmallVector<Type> elementTypes;
+  intrinsic.getElementTypes(elementTypes);
+  if (operandIndex < 0 || operandIndex >= elementTypes.size()) {
+    return failure();
+  }
```
kuhar: When could this happen? This seems like something that we should assert against instead?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  return elementTypes[operandIndex].getIntOrFloatBitWidth();
+}
+
+/// Returns the size of the total K dimension (reduction dimension) for the
```
kuhar: Can you explain what you mean by size? Number of elements?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+getKSize(IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
+  SmallVector<VectorType> undistributedTypes;
+  intrinsic.getUndistributedTileTypes(undistributedTypes);
+  if (undistributedTypes.empty()) {
+    return failure();
+  }
+  // For matmul-like operations (C += A * B), the K dimension is the last
+  // dimension of the LHS operand (operand 0).
+  ArrayRef<int64_t> lhsShape = undistributedTypes[0].getShape();
+  if (lhsShape.size() < 2) {
```
kuhar: What if we are doing `transpose(A) * B`?  Don't we have to check indexing maps in order to find the reduction dims?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+    return failure();
+  }
+  // Return the product of all reduction dimensions (everything after the M
+  // dim).
+  return llvm::product_of(lhsShape.drop_front(1));
```
kuhar: There should be multiple M dims, no?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+getTotalTileElems(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
+                  int operandIndex) {
+  IREE::GPU::MMASingleSubgroupLayout layout =
+      IREE::GPU::getSingleSubgroupLayout(intrinsic, operandIndex);
+  return llvm::product_of(layout.outer) * llvm::product_of(layout.thread) *
+         llvm::product_of(layout.element);
+}
+
+FailureOr<std::pair<int64_t, int64_t>>
+getXorShuffleBounds(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
```
kuhar, line 842: clang-tidy warning below. Is this a false-positive?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
kuhar: This seems like a local function and I don't see docs. Also, we should start function names with action verbs.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+}
+
+// TODO(muzasyed): Add more intrinsics for gfx950.
+static FailureOr<std::pair<int64_t, int64_t>> getXorShuffleParamsForGfx950(
+    IREE::GPU::TargetAttr target,
+    IREE::Codegen::InnerTileDescAttrInterface intrinsic, int operandIndex) {
+  if (auto smma = dyn_cast<IREE::GPU::ScaledMMAAttr>(intrinsic)) {
+    switch (smma.getIntrinsic()) {
+    case IREE::GPU::ScaledMMAIntrinsic::MFMA_SCALE_F32_16x16x128_B32:
+      return std::pair(/*row_width*/ 256l, /*access_width*/ 32l);
```
kuhar: long don't necessarily have to be the same as int64

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: I think this indicates a need for a generalized version of a function like `getMNKShape` or `getMNKKbShapeFromScaledIntrinsic` in the `InnerTiledDescAttrInterface`. 

Since we don't have any MMA intrinsics at the moment where the LHS is represented as KxM, we don't have any functions in the flavour of "getIndexingMapOfOperand'. So I'll just submit a separate PR to add this interface function instead.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: AFAIK we don't have multiple M dimensions in any MMA intrinsic.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: This is going to be linked with the CAPI in the future to expose to the tuner, so I don't want to make it static. I can add a line to tell clang-tidy to ignore this line.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
kuhar: ah disregard, I thought this looks at arbitrary shapes

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+    return failure();
+  }
+  auto [effectiveRowWidth, numAccessElems] = xorShuffleParams.value();
+  auto swizzleAttr = IREE::Codegen::XORShuffleAttr::get(
+      context, effectiveRowWidth, numAccessElems,
+      /*row_stride=*/int64_t(0),
+      /*per_phase=*/int64_t(0));
+  Attribute swizzleOperand =
+      IREE::GPU::SwizzleOperandAttr::get(context, baseConfigAttr, swizzleAttr);
+  return swizzleOperand;
```
kuhar:
```suggestion
  return IREE::GPU::SwizzleOperandAttr::get(context, baseConfigAttr, swizzleAttr);
```

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+/// Returns the total number of elements in the shape of the given intrinsic
+/// with the given operand index. This is the product of all dimensions in the
+/// distributed shape: outer × thread × element.
+static FailureOr<int64_t>
+getTotalTileElems(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
+                  int operandIndex) {
+  IREE::GPU::MMASingleSubgroupLayout layout =
+      IREE::GPU::getSingleSubgroupLayout(intrinsic, operandIndex);
+  return llvm::product_of(layout.outer) * llvm::product_of(layout.thread) *
+         llvm::product_of(layout.element);
```
kuhar, line 799: @krzysz00 does this math work with rdna3 instrinsics that require some elements to be duplicated?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -800,7 +800,42 @@ getTotalTileElems(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
          llvm::product_of(layout.element);
 }
 
-// Disabling clang-tidy for the following function, as it will be externally
+// TODO(muzasyed): Add more intrinsics for gfx950.
+static FailureOr<std::pair<int64_t, int64_t>> getXorShuffleParamsForGfx950(
+  IREE::GPU::TargetAttr target,
+  IREE::Codegen::InnerTileDescAttrInterface intrinsic, int operandIndex) {
```
kuhar: operandIndex seems unused

kuhar (commented):
We shouldn't disable lints in entire functions -- just ignore warnings about internal linkage. There may be other useful warnings.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+getKSize(IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
+  int64_t kSize = 1;
+  int64_t lhsOperandIndex = 0;
+  SmallVector<VectorType> undistributedTypes;
+  intrinsic.getUndistributedTileTypes(undistributedTypes);
+  auto shape = undistributedTypes[lhsOperandIndex].getShape();
+  auto lhsIteratorTypes = intrinsic.getOperandIteratorTypes()[lhsOperandIndex];
+  assert(lhsIteratorTypes.size() == shape.size() &&
+         "Iterator types and shape must have the same size");
+  for (int i = 0; i < lhsIteratorTypes.size(); i++) {
```
krzysz00: This could be an llvm::enumerate() Failing that, the usual nitpickery about ++i and not using `int` when there's a fixed size available - for one thing, this should be `size_t`

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
krzysz00: The duplication lives on `tstrides` . This'll give you what I'd expect to be the correct amount of data to be stored in LDS

krzysz00 (approved):
Nitpicks, but I think this makes sense overall as a step for making this feature happen

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,238 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
   return std::nullopt;
 }
 
+static FailureOr<int64_t>
+getOperandBitwidth(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
+                   int operandIndex) {
+  SmallVector<Type> elementTypes;
+  intrinsic.getElementTypes(elementTypes);
+  assert(elementTypes.size() > operandIndex && "Operand index out of bounds");
```
kuhar: nit/optional: small vector already asserts on OOB accesses, so this assert is mostly redundant

kuhar: if anything, I'd check it's positive

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
kuhar: also don't re-evaluate size

kuhar: You don't need enumerate here, `zip_equals` will do and also internally assert that sizes are consistent

kuhar (commented):
Looks good overall. One more thing to consider is replacing all these `pair<int64_t, int64_t>` with a custom struct so that each integer is named

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.h
```diff
@@ -177,6 +177,13 @@ Value getCombiningIdentityValue(Location loc, OpBuilder &builder,
 mlir::gpu::AllReduceOperation
 combiningKindToAllReduce(vector::CombiningKind kind);
 
+struct XorShuffleParams {
+  int64_t row_width;
+  int64_t access_width;
+  XorShuffleParams(int64_t row_width, int64_t access_width)
+      : row_width(row_width), access_width(access_width) {}
```
kuhar: clang tidy warnings above

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.h
```diff
@@ -176,6 +177,75 @@ Value getCombiningIdentityValue(Location loc, OpBuilder &builder,
...
```
kuhar: You don't need this constructor, something equivalent will be autogenerated for you for POD structs

```suggestion
```

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+/// dimensions) of the given intrinsic. For intrinsics with multiple K
+/// dimensions, returns the product of all K dimensions. This is determined by
+/// looking at all reduction dimensions in the undistributed tile shape.
+static FailureOr<int64_t>
+getKSize(IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
+  int64_t kSize = 1;
+  int64_t lhsOperandIndex = 0;
+  SmallVector<VectorType> undistributedTypes;
+  intrinsic.getUndistributedTileTypes(undistributedTypes);
+  auto shape = undistributedTypes[lhsOperandIndex].getShape();
```
kuhar: spell out the type

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+  auto lhsIteratorTypes = intrinsic.getOperandIteratorTypes()[lhsOperandIndex];
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+
+// TODO(muzasyed): Add more intrinsics for gfx950.
+static FailureOr<XorShuffleParams> getXorShuffleParamsForGfx950(
+    IREE::GPU::TargetAttr target,
+    IREE::Codegen::InnerTileDescAttrInterface intrinsic) {
+  if (auto smma = dyn_cast<IREE::GPU::ScaledMMAAttr>(intrinsic)) {
+    switch (smma.getIntrinsic()) {
+    case IREE::GPU::ScaledMMAIntrinsic::MFMA_SCALE_F32_16x16x128_B32:
+      return XorShuffleParams({/*rowElems*/ int64_t(256),
+                               /*accessElems*/ int64_t(32)});
```
kuhar: This way clang-tidy / clangd can check that the argument names match the comments
```suggestion
      return XorShuffleParams({/*rowElems=*/256,
                               /*accessElems=*/ 32});
```
Also you don't need the explicit upcast to i64 since we switched to a struct

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+    default:
+      return failure();
```
kuhar, line 812: Maybe move the TODO here?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -745,6 +746,240 @@ std::optional<SmallVector<int64_t>> getWmmaNativeVectorSize(Operation *op) {
...
+                  IREE::GPU::TargetAttr target,
+                  IREE::Codegen::InnerTileDescAttrInterface intrinsic,
+                  ArrayRef<int64_t> reductionTileSizes, int operandIndex) {
+  FailureOr<XorShuffleParams> xorShuffleParams =
+      getXorShuffleParams(target, intrinsic, reductionTileSizes, operandIndex);
+  if (failed(xorShuffleParams)) {
+    return failure();
+  }
+  auto effectiverowElems = xorShuffleParams.value().rowElems;
+  auto numAccessElems = xorShuffleParams.value().accessElems;
```
kuhar: spell out the types

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.h
```diff
@@ -176,6 +177,73 @@ Value getCombiningIdentityValue(Location loc, OpBuilder &builder,
 mlir::gpu::AllReduceOperation
 combiningKindToAllReduce(vector::CombiningKind kind);
 
+struct XorShuffleParams {
```
kuhar, line 209: Maybe add a comment explaining this matches the attribute defined in <blah>.td?

Muzammiluddin-Syed-ECE:
the clang-tidy warnings are false positives (`getXorShuffleAttr` is used in ConfigUtils.cpp) or not related to this PR (`getNSize` and `getKSize`).

kuhar (approved):

PR #23180, opened 2026-01-17
Simplify quantifiers all_of, any_of, none_of. NFC.

kuhar:
* Use `llvm::IsaPred<T>` instead of lambdas where possible
* `!any_of` --> `none_of`


jtuyls (approved):
LGTM

Issue #23092, opened 2026-01-12
[DT] Unfused `set_encoding` in llama 70B/405B


Yu-Zhewen:
To reproduce, add `--iree-dispatch-creation-data-tiling` to https://github.com/nod-ai/iree-model-benchmark/blob/6cf44644c23ea53a20fd22f18486d1ac78788c3d/llama3/compile-70b-fp4-base.sh#L30

IR: 
```ll
// Quantization
%834:2 = flow.dispatch.region -> (tensor<?xi8>{%833}, tensor<?x32xf4E2M1FN>{%833}) {...}

%expanded_11 = tensor.expand_shape %834#0 [[0, 1, 2]] output_shape [4, %812, 256] : tensor<?xi8> into tensor<4x?x256xi8>
%835 = iree_tensor_ext.bitcast %834#1 : tensor<?x32xf4E2M1FN>{%833} -> tensor<?x16xi8>{%833}
%836 = iree_tensor_ext.compute_barrier.start %835 : tensor<?x16xi8>{%833} -> tensor<?x16xi8>
%expanded_12 = tensor.expand_shape %836 [[0, 1, 2], [3]] output_shape [4, %812, 256, 16] : tensor<?x16xi8> into tensor<4x?x256x16xi8>
%collapsed_13 = tensor.collapse_shape %expanded_12 [[0], [1], [2, 3]] : tensor<4x?x256x16xi8> into tensor<4x?x4096xi8>
%cast = tensor.cast %collapsed_13 : tensor<4x?x4096xi8> to tensor<?x?x?xi8>
%837 = iree_tensor_ext.bitcast %cast : tensor<?x?x?xi8>{%c4, %812, %c4096} -> tensor<?x?x?xf4E2M1FN>{%c4, %812, %c8192}
%838 = iree_tensor_ext.compute_barrier.start %837 : tensor<?x?x?xf4E2M1FN>{%c4, %812, %c8192} -> tensor<?x?x?xf4E2M1FN>
%839 = iree_tensor_ext.bitcast %expanded_11 : tensor<4x?x256xi8>{%812} -> tensor<4x?x256xf8E8M0FNU>{%812}
%840 = iree_tensor_ext.compute_barrier.start %839 : tensor<4x?x256xf8E8M0FNU>{%812} -> tensor<4x?x256xf8E8M0FNU>
%841 = iree_tensor_ext.bitcast %3 : tensor<8192x256xi8> -> tensor<8192x256xf8E8M0FNU>
[...]

PR #23181, opened 2026-01-18
[DispatchCreation] Include TensorExt ops in compute regions for barrier insertion

Yu-Zhewen:
One use case is `iree_tensor_ext.bitcast`, and the wrong position of barriers can block dispatch fusion in data-tiling (https://github.com/iree-org/iree/issues/23092#issuecomment-3746370586).


IanWood1 (approved):
LGTM!

hanhanW (approved):

PR #23182, opened 2026-01-18
[TensorExt] Propagate and fold iree_tensor_ext.bitcast 

Yu-Zhewen:
In Llama 70B/405B compilation with data-tiling, we encounter IR like:

```
%0 = iree_tensor_ext.bitcast %arg0 : tensor<?x32xf4E2M1FN> -> tensor<?x16xi8>
%1 = tensor.expand_shape %0 [[0, 1, 2], [3]] output_shape [4, %d1, 512, 16]
    : tensor<?x16xi8> into tensor<4x?x512x16xi8>
%2 = tensor.collapse_shape %1 [[0], [1], [2, 3]]
    : tensor<4x?x512x16xi8> into tensor<4x?x8192xi8>
%3 = iree_tensor_ext.bitcast %2 : tensor<4x?x8192xi8> -> tensor<4x?x16384xf4E2M1FN>
```

This PR therefore adds the following patterns to propagate and fold `bitcast` operation:
   - `expand_shape(bitcast(x))` → `bitcast(expand_shape(x))`
   - `bitcast(collapse_shape(x))` → `collapse_shape(bitcast(x))`
   - `bitcast(bitcast(x))` → `bitcast(x)`

The first two patterns are added to the `BubbleUpExpandShapes` pass, and the third pattern is added as a canonicalization for the `iree_tensor_ext.bitcast` op.


compiler/src/iree/compiler/Dialect/TensorExt/IR/test/tensor_ext_folding.mlir
```diff
@@ -45,3 +45,101 @@ util.func public @barrierEndFoldDuplicate(%arg0: tensor<4x8xf32>) -> tensor<4x8x
...
+// -----
+
+// CHECK-LABEL: @foldConsecutiveBitCasts
+// CHECK-SAME:  %[[ARG0:[A-Za-z0-9]+]]: tensor<4x?x16384xf4E2M1FN>
+// CHECK-SAME:  %[[D0:[A-Za-z0-9]+]]: index
+util.func public @foldConsecutiveBitCasts(%arg0: tensor<4x?x16384xf4E2M1FN>, %d0: index) -> tensor<4x?x16384xf4E2M1FN> {
+  // CHECK-NOT: iree_tensor_ext.bitcast
+  // CHECK: util.return %[[ARG0]]
+  %0 = iree_tensor_ext.bitcast %arg0 : tensor<4x?x16384xf4E2M1FN>{%d0} -> tensor<4x?x8192xi8>{%d0}
+  %1 = iree_tensor_ext.bitcast %0 : tensor<4x?x8192xi8>{%d0} -> tensor<4x?x16384xf4E2M1FN>{%d0}
```
IanWood1, line 104: This should already be handled by the folder, right? https://github.com/iree-org/iree/blob/20701db8707b2227c1ea233e2c005aa67a025d5e/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp#L112-L116

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
```diff
@@ -100,6 +97,204 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
...
+        collapseOp, collapseResultType, newCollapseOp,
+        bitcastOp.getSourceDims(), bitcastOp.getResultDims());
+    rewriter.eraseOp(bitcastOp);
+    return success();
+  }
+};
+
+/// Fold consecutive bitcasts that cancel each other out.
+/// bitcast(bitcast(x : A -> B) : B -> A) -> x
+struct FoldConsecutiveBitCasts final : OpRewritePattern<BitCastOp> {
```
IanWood1: I think this might be incorrect in the case where the input dynamic dims != output dynamic dims (see `compareShapesEqual` used in the folder). It's a bit unclear if the bitcast also allows reassociating dynamic dimensions.

IanWood1 (changes requested):
I don't think `PropagateBitCastThroughExpandShape` and `PropagateBitCastThroughCollapseShape` are a good fit for canonicalization patterns. When comparing `expand_shape(bitcast(x))` with `bitcast(expand_shape(x))`, neither is more "canonical" than the other. 

These patterns should be moved under the `Transforms/` directory and then used in a pass similar to [BubbleUpExpandShapes](https://github.com/iree-org/iree/blob/20701db8707b2227c1ea233e2c005aa67a025d5e/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp). This is what is done for LinalgExt ops [here](https://github.com/iree-org/iree/blob/20701db8707b2227c1ea233e2c005aa67a025d5e/compiler/src/iree/compiler/Codegen/Dialect/Codegen/Transforms/ReshapeFusion.cpp).

compiler/src/iree/compiler/Dialect/TensorExt/IR/test/tensor_ext_folding.mlir
IanWood1: Ignore this, I was wrong. The folder just looks at a single bitcast.

Yu-Zhewen:
> I don't think `PropagateBitCastThroughExpandShape` and `PropagateBitCastThroughCollapseShape` are a good fit for canonicalization patterns. When comparing `expand_shape(bitcast(x))` with `bitcast(expand_shape(x))`, neither is more "canonical" than the other.
> 
> These patterns should be moved under the `Transforms/` directory and then used in a pass similar to [BubbleUpExpandShapes](https://github.com/iree-org/iree/blob/20701db8707b2227c1ea233e2c005aa67a025d5e/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp). This is what is done for LinalgExt ops [here](https://github.com/iree-org/iree/blob/20701db8707b2227c1ea233e2c005aa67a025d5e/compiler/src/iree/compiler/Codegen/Dialect/Codegen/Transforms/ReshapeFusion.cpp).

Good point! I've addressed this by creating `SinkBitCastPass`.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
Yu-Zhewen: I've reworked this pattern. Instead of only handling the special case where two bitcasts cancel out completely (`A -> B -> A`), I changed it to chain any two consecutive bitcasts into one. The `compareShapesEqual` check for complete cancellation (when the result is a no-op) is left to `BitCastOp::fold`, which already handles that case.

compiler/src/iree/compiler/DispatchCreation/Passes.td
```diff
@@ -210,6 +210,20 @@ def FusionPreprocessingPass :
   ];
 }
 
+def SinkBitCastPass :
```
IanWood1: Do we need to create a new pass for this?

compiler/src/iree/compiler/DispatchCreation/Passes.cpp
```diff
@@ -285,6 +285,15 @@ static void addDispatchRegionCreationPasses(OpPassManager &passManager,
...
   FunctionLikeNest(passManager)
+      // Sink bitcast operations through reshapes to enable better fusion
+      // of encoding ops into dispatch regions.
+      .addPass(DispatchCreation::createSinkBitCastPass)
+      // Canonicalize to fold chained bitcasts.
+      .addPass([&]() {
+        IREE::Flow::CanonicalizePassOptions options;
+        options.cseConstants = false;
+        return IREE::Flow::createCanonicalizePass(options);
+      })
```
IanWood1: Do we need add a new pass here? What if these patterns were used in BubbleUpExpandShapes?

compiler/src/iree/compiler/DispatchCreation/SinkBitCast.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    rewriter.replaceOpWithNewOp<BitCastOp>(
+        expandOp, expandResultType, newExpandOp, newDynSrcDims, newDynDstDims);
+    rewriter.eraseOp(bitcastOp);
+    return success();
+  }
+};
+
+/// Propagate bitcast through collapse_shape: collapse_shape(bitcast(x)) ->
+/// bitcast(collapse_shape(x)).
```
IanWood1: Another note on moving this to BubbleUpExpandShapes: this should be "sinking" `tensor.collapse_shape` ops. AKA both patterns should be increasing the dimensionality of the bitcast.

IanWood1 (commented):
The bitcast folding pattern looks good to me. I haven't thoroughly reviewed the reshape patterns yet, since I'm trying to understand why we need them and where they should run in the pipeline. If you want, the consecutive bitcast pattern could be moved to a new PR so that it could be merged faster.

My main concern with the SinkBitCastPass is that it feels like too much of a one-off pass. BubbleUpExpandShapesPass seems like a natural place to put these patterns (note that the pass "bubbles" expand shapes and "sinks" collapse shapes). The goal of that pass is to move reshapes to the edges of the program so that they don't get stuck between ops. It's main use case is for fusion but it will also clear the reshapes so that the bitcasts can fold. However, I don't have the best understanding of the encoding pipeline so I could be missing something here. Let me know you don't think this will work

Yu-Zhewen:
> The bitcast folding pattern looks good to me. I haven't thoroughly reviewed the reshape patterns yet, since I'm trying to understand why we need them and where they should run in the pipeline. If you want, the consecutive bitcast pattern could be moved to a new PR so that it could be merged faster.
> 
> My main concern with the SinkBitCastPass is that it feels like too much of a one-off pass. BubbleUpExpandShapesPass seems like a natural place to put these patterns (note that the pass "bubbles" expand shapes and "sinks" collapse shapes). The goal of that pass is to move reshapes to the edges of the program so that they don't get stuck between ops. It's main use case is for fusion but it will also clear the reshapes so that the bitcasts can fold. However, I don't have the best understanding of the encoding pipeline so I could be missing something here. Let me know you don't think this will work

Thanks @IanWood1 for the suggestion! I've addressed your feedback:

- Moved patterns to `BubbleUpExpandShapes` instead of a new `SinkBitCastPass`.
- Changed the collapse pattern direction as you suggested. The collapse pattern now "sinks" `collapse_shape` through `bitcast`.

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
```diff
@@ -377,6 +380,197 @@ struct BubbleExpandThroughConcat final
...
+    }
+
+    // For rank > 1, the new expand output shape's last dim is derived from the
+    // bitcast source's last dim (scaled by bitwidth ratio). We don't support
+    // dynamic last dim yet since we'd need affine expressions to scale it.
+    // For rank 1, the bitwidth ratio is 1:1, so it doesn't matter.
+    if (bitcastSrcType.getRank() > 1 &&
+        ShapedType::isDynamic(bitcastSrcType.getShape().back())) {
+      return failure();
+    }
```
IanWood1: I'm not sure what's special about the rank 1 case. For example, we need affine expressions for:

```mlir
func.func @bubbleExpand1dThroughBitCastI32ToI8(%arg0: tensor<?xi32>, %d0: index, %d1: index) -> tensor<4x512x?xi8> {
  %c4 = arith.constant 4 : index
  %d0_mul_4 = arith.muli %d0, %c4 : index
  %0 = iree_tensor_ext.bitcast %arg0 : tensor<?xi32>{%d0} -> tensor<?xi8>{%d0_mul_4}
  %1 = tensor.expand_shape %0 [[0, 1, 2]] output_shape [4, 512, %d1]
      : tensor<?xi8> into tensor<4x512x?xi8>
  return %1 : tensor<4x512x?xi8>
}
```

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
```diff
@@ -377,6 +380,197 @@ struct BubbleExpandThroughConcat final
...
+    RankedTensorType expandResultType = expandOp.getResultType();
+    SmallVector<int64_t> newOutputShape(expandResultType.getShape());
+    newOutputShape.back() =
+        (newOutputShape.back() * bitcastDstType.getElementTypeBitWidth()) /
+        bitcastSrcType.getElementTypeBitWidth();
+
+    SmallVector<OpFoldResult> oldMixedOutputShape =
+        expandOp.getMixedOutputShape();
+    SmallVector<OpFoldResult> newMixedOutputShape(oldMixedOutputShape);
+    newMixedOutputShape.back() = rewriter.getIndexAttr(newOutputShape.back());
```
IanWood1: Related to the comment above: I think we need to check that the last dim in the output shape is static.

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
```diff
@@ -377,6 +380,197 @@ struct BubbleExpandThroughConcat final
...
+      if (lastGroup.size() != 1) {
+        return failure();
+      }
+    }
+
+    // Compute the new expand output shape. Same as original but with the last
+    // dim scaled according to the bitcast source and destination element type
+    // bitwidths.
+    RankedTensorType expandResultType = expandOp.getResultType();
+    SmallVector<int64_t> newOutputShape(expandResultType.getShape());
```
IanWood1: This might be clearer.

```suggestion
    SmallVector<int64_t> newExpandedShape(expandResultType.getShape());
```

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
```diff
@@ -377,6 +380,197 @@ struct BubbleExpandThroughConcat final
...
+    // Build the dynamic dims for the new bitcast.
+    int64_t expandResultRank = expandResultType.getRank();
+    SmallVector<Value> newDynSrcDims;
+    for (int64_t i = 0; i < expandResultRank; ++i) {
+      if (ShapedType::isStatic(newOutputShape[i])) {
+        continue;
+      }
+      Value dimVal = getValueOrCreateConstantIndexOp(
+          rewriter, bitcastOp.getLoc(), newMixedOutputShape[i]);
+      newDynSrcDims.push_back(dimVal);
```
IanWood1: Is there ever a case where `newOutputShape[i]` is dynamic but `newMixedOutputShape[i]` is static? If not, we can replace `getValueOrCreateConstantIndexOp` with a dyn_cast<Value>.

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
```diff
@@ -377,6 +380,197 @@ struct BubbleExpandThroughConcat final
...
+    RankedTensorType bitcastDstType = bitcastOp.getResult().getType();
+
+    // All dims except last must match for bitcast (last dim can differ due to
+    // element type size change). This check also filters out rank mismatches.
+    if (!compareMixedShapesEqualExceptLast(
+            collapseDstType, bitcastOp.getSourceDims(), bitcastDstType,
+            bitcastOp.getResultDims())) {
+      return failure();
+    }
+
```
IanWood1, line 510: You need to add a check similar to what is done for expand_shape to ensure that the collapse doesn't collapse the last dim.


e.g.

```
func.func @collapseSourceLastDimNotDivisible(%arg0: tensor<4x5xf32>) -> tensor<10xi64> {
  %0 = tensor.collapse_shape %arg0 [[0, 1]]
      : tensor<4x5xf32> into tensor<20xf32>
  %1 = iree_tensor_ext.bitcast %0 : tensor<20xf32> -> tensor<10xi64>
  return %1 : tensor<10xi64>
}
```

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
```diff
@@ -377,6 +380,197 @@ struct BubbleExpandThroughConcat final
...
+    auto newBitcastResultType = RankedTensorType::get(
+        newBitcastResultShape, bitcastDstType.getElementType());
+
+    // Build dynamic dims for the new bitcast. For dims that are in their own
+    // reassociation group, we can reuse the corresponding source dims.
+    SmallVector<Value> newBitcastSrcDims;
+    SmallVector<Value> newBitcastDstDims;
+    for (int64_t i = 0; i < collapseSrcType.getRank(); ++i) {
+      if (collapseSrcType.isDynamicDim(i)) {
+        Value dimVal = tensor::DimOp::create(rewriter, collapseOp.getLoc(),
```
IanWood1:
```suggestion
        Value dimVal = tensor::DimOp::createOrFold(rewriter, collapseOp.getLoc(),
```

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
Yu-Zhewen: Thanks for catching this! You're right that we need a check here. However, I believe the issue isn't specifically about whether the last dimension is collapsed with others, but rather whether the bitwidth scaling produces an integer dimension size.

Your example fails because: last dim after scaling: 5 × 32 bits / 64 bits = 2.5 (fractional!)

I've added a divisibility check `(lastDimSize × srcBits) % dstBits == 0` and also included the negative test case.

(Note that we already have a check ensuring that dynamic dimensions are in their own reassociation groups, so we don't need to worry about that.)

compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
Yu-Zhewen: You're right, I was overthinking this. Fixed.

IanWood1 (approved):
LGTM, thanks!

PR #23183, opened 2026-01-18
[TensorExt] Improves `BitCastOfTensorCastStaticInfo` to handle constant dynamic dims

Yu-Zhewen:
Handles cases like `tensor<?x?x?xi8>{%c4, %d0, %c4096}` where dynamic dims are actually constants.

These patterns originated from https://github.com/nod-ai/amd-shark-ai/blob/fca16186c53e663435934a9ab2aab6bfb6f93e0a/amdsharktank/amdsharktank/kernels/gemm_fp4.py#L55.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
@@ -52,8 +52,10 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
       return failure();
     }
 
-    // TODO: Support partial static info incorporation.
-    if (bitcastOp.getSourceDims() != bitcastOp.getResultDims()) {
+    // All dims except last must match (last dim can differ due to element type
+    // size change).
+    if (bitcastOp.getSourceDims().drop_back() !=
+        bitcastOp.getResultDims().drop_back()) {
       return failure();
     }
 
@@ -61,7 +63,8 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
     RankedTensorType resTensorType = bitcastOp.getResult().getType();
     ArrayRef<int64_t> resShape = resTensorType.getShape();
 
-    SmallVector<Value> newDynamicDims;
+    SmallVector<Value> newDynSrcDims;
+    SmallVector<Value> newDynDestDims;
     int64_t intermediateDynamicDim = 0;
     int64_t resDynamicDim = 0;
     SmallVector<int64_t> newResultSizes(resShape);
@@ -76,9 +79,15 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
         ++resDynamicDim;
       }
 
+      APInt cst;
       if (ShapedType::isDynamic(castSize)) {
-        newDynamicDims.push_back(
+        newDynSrcDims.push_back(
             bitcastOp.getSourceDims()[intermediateDynamicDim]);
+        newDynDestDims.push_back(
+            bitcastOp.getResultDims()[intermediateDynamicDim]);
+      } else if (matchPattern(bitcastOp.getResultDims()[intermediateDynamicDim],
+                              m_ConstantInt(&cst))) {
+        newResultSizes[resDynamicDim] = cst.getSExtValue();
       } else {
         newResultSizes[resDynamicDim] = castSize;
       }
@@ -91,7 +100,7 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
                               resTensorType.getEncoding());
     Value newBitcast = BitCastOp::create(rewriter, bitcastOp.getLoc(), newType,
                                          tensorCastOp.getOperand(),
-                                         newDynamicDims, newDynamicDims);
+                                         newDynSrcDims, newDynDestDims);
     // We create a new cast to continue propagating static information.
     rewriter.replaceOpWithNewOp<tensor::CastOp>(bitcastOp, resTensorType,
                                                 newBitcast);
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
@@ -72,14 +72,14 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
     SmallVector<OpFoldResult> newResultMixed;
     int64_t rank = resTensorType.getRank();
     for (int64_t i = 0; i < rank - 1; ++i) {
-      int64_t castSize = tensorCastSrcType.getShape()[i];
-      OpFoldResult resOfr = resultMixed[i];
       // Try cast source static info first.
+      int64_t castSize = tensorCastSrcType.getShape()[i];
       if (!ShapedType::isDynamic(castSize)) {
         newResultMixed.push_back(rewriter.getIndexAttr(castSize));
         continue;
       }
       // Try result constant.
+      OpFoldResult resOfr = resultMixed[i];
       if (std::optional<int64_t> resConst = getConstantIntValue(resOfr)) {
         newResultMixed.push_back(rewriter.getIndexAttr(*resConst));
         continue;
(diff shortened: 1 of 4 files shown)


jtuyls (approved):
LGTM

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
```diff
@@ -52,33 +50,41 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
...
     }
 
-    // TODO: Support partial static info incorporation.
-    if (bitcastOp.getSourceDims() != bitcastOp.getResultDims()) {
+    // All dims except last must match (last dim can differ due to element type
+    // size change).
+    if (bitcastOp.getSourceDims().drop_back() !=
+        bitcastOp.getResultDims().drop_back()) {
       return failure();
     }
```
hanhanW, line 62: Please add a comment about this also filters out the case that the ranks mismatch. It is allowed to have different rank between source and destination. E.g.,

https://github.com/iree-org/iree/blob/61e6df0d9c9b337e1e9a5c3e065dce31f8af86d8/compiler/plugins/input/Torch/InputConversion/test/bitcast_tensor.mlir#L73-L78

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
```diff
@@ -52,33 +50,41 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
...
+        newDynSrcDims.push_back(
             bitcastOp.getSourceDims()[intermediateDynamicDim]);
+        newDynDestDims.push_back(
+            bitcastOp.getResultDims()[intermediateDynamicDim]);
+      } else if (matchPattern(bitcastOp.getResultDims()[intermediateDynamicDim],
+                              m_ConstantInt(&cst))) {
+        newResultSizes[resDynamicDim] = cst.getSExtValue();
       } else {
         newResultSizes[resDynamicDim] = castSize;
       }
```
hanhanW, line 112: Given that the rank is the same, I'm thinking something more general. Can we infer the shape more aggressively?

E.g., you can use [getConstantIntValue](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h) method on both srcDim and destDim. If one of them is constant, the dim size can be constant. Otherwise, we preserve the original dynamic value. The last dimension needs to be handled specially though. In this context, we can complete the folding for plain tensor types, and people wont need to update the code anymore. E.g., we will be able to support `tensor<4x?x4096xi8> -> tensor<?x?x?xi4>`.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
hanhanW: oh, maybe it is already covered by `canFoldIntoConsumerOp`, but not this check.

Yu-Zhewen: I realized `bitcastOp.getSourceDims().drop_back() != bitcastOp.getResultDims().drop_back()` is not enough, as it only checks dynamic dimension values and their relative order. Created a new utility function `compareShapesEqualExceptLastDim` which wraps the existing `compareShapesEqual`.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
Yu-Zhewen: Done. Also added a test for the case of `tensor<4x128x4096xi8> -> tensor<?x?x?xi4>`

compiler/src/iree/compiler/Utils/ShapeUtils.cpp
```diff
@@ -5,37 +5,81 @@
...
+        return false;
+      }
+      rhsDynDimIdx++;
     } else {
+      // Both static - compare sizes directly.
       if (lhsType.getDimSize(i) != rhsType.getDimSize(i)) {
-        // Static dimensions differ.
         return false;
       }
     }
```
hanhanW, line 37: I'd restructure it a bit if we introduce `allowCasting`.

```
for (...) {
   // Check static first.
   if (lhsIsStatic && rhsIsStatic) // compare and return.
   if (!lhsIsStatic && !rhsIsStatic) // compare SSA value and return.
   if (!allowCasting) { return false; }
   // The rest of code, and you can use isEqualConstantIntOrValue method.
}
```

isEqualConstantIntOrValue: https://github.com/llvm/llvm-project/blob/51e2c82023d9fd1345448cdab504ae1b2d05f9f8/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L134-L140

The other approach can be adding a new function for `allowCasting = true` case. And the implementation can be simple:

- Use [dispatchIndexOpFoldResults](https://github.com/llvm/llvm-project/blob/51e2c82023d9fd1345448cdab504ae1b2d05f9f8/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L65-L70) to construct `SmallVector<OpFoldResult>` for each.
- Return the result from [isEqualConstantIntOrValueArray](https://github.com/llvm/llvm-project/blob/51e2c82023d9fd1345448cdab504ae1b2d05f9f8/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L134-L140).

I'm +1 to the second approach which creates a new function because passing boolean-ish argument usually makes code harder to follow. The implementation is simple, so I don't see a strong reason to make it work with existing function.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
hanhanW: Simliar to the other comment. I'm thinking if we should just
1. Dispatch them to OpFoldResults, using [getMixedValues](https://github.com/llvm/llvm-project/blob/51e2c82023d9fd1345448cdab504ae1b2d05f9f8/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L157-L164) method.
2. Use source array as default.
3. Continue if the source item is known as static.
4. Check the corresponding item from result OpFoldResults and update it if it is a constant.
5. Use [decomposeMixedValues](https://github.com/llvm/llvm-project/blob/51e2c82023d9fd1345448cdab504ae1b2d05f9f8/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L166-L170C1) to construct the final shape and dynamic values.

The last dimension needs to be handled specially, and I believe that you can figure it out -- if the above idea makes sense to you.

The concern is about having too many branches in the implementation and it is not easy to follow the logic.

hanhanW (changes requested):

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
Yu-Zhewen: Thanks for the detailed suggestions!

hanhanW: I did not realize that the source is Tensor::CastOp's source. You made a reasonable change to me, thanks!

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpFolders.cpp
```diff
@@ -52,46 +53,72 @@ struct BitCastOfTensorCastStaticInfo final : OpRewritePattern<BitCastOp> {
...
+    MLIRContext *ctx = bitcastOp.getContext();
+    SmallVector<OpFoldResult> resultMixed = getMixedValues(
+        resTensorType.getShape(), bitcastOp.getResultDims(), ctx);
+
+    // Build new result shape by propagating static information.
+    SmallVector<OpFoldResult> newResultMixed;
+    int64_t rank = resTensorType.getRank();
+    for (int64_t i = 0; i < rank - 1; ++i) {
+      int64_t castSize = tensorCastSrcType.getShape()[i];
+      OpFoldResult resOfr = resultMixed[i];
```
hanhanW: optional: I'd move it before where it is used, i.e., after the `Try result constants.` comment.

hanhanW (approved):
LGTM, thanks

Issue #7737, opened 2021-11-24
Enable mutable binding fusion by propagating noalias into codegen.


benvanik:
To reduce the total amount of overhead in the system it's important that we are able to alias bindings provided to dispatches. Unfortunately we are not able to express to some of the backends that aliased bindings do not imply overlapping regions of memory and when we do induce the aliasing we lose vectorization :(

To work around this we currently do a conservative form of alias induction by way of the FuseDispatchBindings pass with aliasMutableBindings=false resulting in double the number of bindings being required as we must provide both read and write bindings aliasing the same memory instead of just providing the one read/write:
https://github.com/google/iree/blob/23194ac085f8e604007d85a600cc456f8b79a67d/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp#L97-L103

But our memory model matches that of Vulkan/SPIR-V: https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#_a_id_aliasingsection_a_aliasing
> The Simple, GLSL, and Vulkan memory models can assume that aliasing is generally not present between the memory object declarations. Specifically, the consumer is free to assume aliasing is not present between memory object declarations, unless the memory object declarations explicitly indicate they alias. Aliasing is indicated by applying the Aliased decoration to a memory object declaration’s <id>, for OpVariable and OpFunctionParameter.

It'd be good then if we could emit code that observed this behavior so that we had a consistent memory model we were targeting and didn't need to worry about slipping off the fast path on accident.  In SPIR-V we don't need to do anything, then, unless we ever switch to the OpenCL model (which requires a `Restrict` decoration), but on the LLVM CPU/GPU paths we are currently stuck with the inverse assumption that everything may alias unless explicitly told otherwise.

The root issue on the LLVM side is that the `noalias` attribute is ignored on anything but function arguments even though the spec allows `restrict` to be placed on any pointer value (and gcc and icc support it): https://en.cppreference.com/w/c/language/restrict
> Struct members
> The scope of the aliasing assertion made by a restrict-qualified pointer that is a member of a struct is the scope of the identifier used to access the struct.
> Even if the struct is declared at file scope, when the identifier used to access the struct has block scope, the aliasing assertions in the struct also have block scope; the aliasing assertions are only in effect within a block execution or a function call, depending on how the object of this struct type was created.

[...]

PR #22236, opened 2025-10-07
[Codegen] Drop Descriptor flag uses on `hal.binding.subspan` and use `MemoryAccessBitfield` to represent subspan access

RiverDave:

PR #23184, opened 2026-01-19
Propagate binding correlation information to LLVM codegen for mutable binding fusion

kimm240:
## Problem

When `--iree-stream-resource-alias-mutable-bindings=true` is enabled, the `FuseDispatchBindings` pass fuses multiple bindings that refer to the same underlying storage. However, the correlation information about which bindings alias or don't alias was not propagated to LLVM code generation, causing LLVM's alias analysis to be overly conservative. This prevents optimizations like vectorization, especially in cases with multiple in-place operands (e.g., the `sort_3d` test case with two in-place operands of different types).

This addresses issue #7737.

## Solution

We propagate binding correlation information from the Stream dialect through to LLVM code generation by:

1. **Storing correlation attributes in Stream dialect**: The `FuseDispatchBindings` pass now attaches `stream.binding_correlation` and `stream.binding_noalias` attributes to function arguments, indicating which bindings are in the same correlation group (same resource, different offsets) or point to different resources.

2. **Reading attributes in LLVM codegen**: The `ConvertFunc` class in `LLVMGPU/ConvertToLLVM.cpp` reads these attributes and applies LLVM `noalias` attributes to function arguments, enabling LLVM's alias analysis to make better optimization decisions.

### FuseDispatchBindings Pass (`FuseDispatchBindings.cpp`)

In `updateExecutableSignature`, we now generate two types of attributes:
- `stream.binding_correlation`: Indicates bindings in the same correlation group (same resource, different offsets - they don't alias each other)
- `stream.binding_noalias`: Indicates bindings pointing to different resources (guaranteed not to alias)

```cpp
// For bindings in different correlation groups (different resources)
if (binding.value().correlationMap != otherBinding.value().correlationMap) {
  noaliasIndices.push_back(IntegerAttr::get(..., otherBinding.index()));
}
if (!noaliasIndices.empty()) {
  funcOp.setArgAttr(bindingArg.getArgNumber(), "stream.binding_noalias",
                   ArrayAttr::get(funcOp.getContext(), noaliasIndices));
}
```

### LLVMGPU ConvertToLLVM (`LLVMGPU/ConvertToLLVM.cpp`)

The `ConvertFunc` class reads the correlation attributes from the original function and applies LLVM `noalias` attributes:

```cpp
// Read noalias information
if (auto noaliasAttr = funcOp.getArgAttrOfType<ArrayAttr>(
        binding, "stream.binding_noalias")) {
  // ... process and store ...
}

// Apply LLVM noalias attribute
newFuncOp.setArgAttr(idx, LLVM::LLVMDialect::getNoAliasAttrName(), unit);
```

## Verification

The implementation has been verified with the `sort_3d` test case:

IR was made by
`
../iree-build/tools/iree-compile --iree-hal-target-device=cuda --iree-cuda-target=sm_86 --mlir-print-ir-after-all test.mlir -o test.vmfb 2> ir_dump.log
`.

`test.mlir` is it. which is `@sort_3d` test in `dispatch_linalg_on_tensors.mlir` :
```
util.func public @sort_3d(%arg0: tensor<?x?x?xi32>, %arg1 : tensor<?x?x?xf32>)  
    -> (tensor<?x?x?xi32>, tensor<?x?x?xf32>) {  
  %0, %1 = iree_linalg_ext.sort dimension(0)  
      outs(%arg0, %arg1 : tensor<?x?x?xi32>, tensor<?x?x?xf32>) {  
      ^bb0(%arg2: i32, %arg3: i32, %arg4 : f32, %arg5 : f32):  
        %2 = arith.cmpf ogt, %arg4, %arg5 : f32  
        iree_linalg_ext.yield %2 : i1  
      } -> tensor<?x?x?xi32>, tensor<?x?x?xf32>  
  util.return %0, %1 : tensor<?x?x?xi32>, tensor<?x?x?xf32>  
}    
```

1. **Stream dialect**: `stream.binding_noalias` attributes are correctly generated:
   ```mlir
   func.func @dispatch(
     %arg0: !stream.binding {stream.binding_noalias = [1 : i32]},
     %arg1: !stream.binding {stream.binding_noalias = [0 : i32]}
   )
   ```
   -  It is in after `FuseDispatchBindingsPass`.

2. **LLVM dialect**: `llvm.noalias` attributes are correctly applied:
   ```mlir
   llvm.func @dispatch(
     %arg0: !llvm.ptr<1> {llvm.noalias, llvm.nonnull, llvm.noundef},
     %arg1: !llvm.ptr<1> {llvm.noalias, llvm.nonnull, llvm.noundef}
   )
   ```
   - It is  in after `ConvertToNVVMPass`.

## Future Work

- Implement similar support for LLVMCPU backend (currently only LLVMGPU is supported)
- Consider using alias.scope metadata for more precise alias information (as mentioned in the TODO comments)


compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -190,15 +190,51 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
...
+    // don't alias each other).
+    // 
+    // For bindings in different correlation groups (pointing to different
+    // resources), we also store this information so that LLVM codegen can
+    // infer that they don't alias each other (noalias).
+    SmallVector<Attribute> correlatedIndices;
+    SmallVector<Attribute> noaliasIndices;
+    for (auto otherBinding : llvm::enumerate(bindings)) {
+      if (binding.index() == otherBinding.index())
+        continue;
```
bangtianliu: We use braces with single-line if statements and loops. https://iree.dev/developers/general/contributing/#compiler

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
```diff
@@ -341,13 +341,76 @@ class ConvertFunc : public ConvertOpToLLVMPattern<func::FuncOp> {
...
+    DenseMap<int64_t, SmallVector<int64_t>> bindingNoaliasMap;
+    for (auto subspan : subspans) {
+      int64_t binding = subspan.getBinding().getSExtValue();
+      // Try to read correlation information from the original function.
+      // Note: This assumes the correlation attribute was preserved through
+      // Stream → HAL conversion. If not, we'll need to preserve it explicitly.
+      if (auto correlationAttr = funcOp.getArgAttrOfType<ArrayAttr>(
+              binding, "stream.binding_correlation")) {
+        SmallVector<int64_t> correlatedBindings;
+        for (auto attr : correlationAttr) {
```
krzysz00: Can we spell out the types here?

krzysz00: ... more importantly

1. Can we turn this as corralatedBindings.getAsValueRange<IntegerAttr>
2. If you're worried about the types, you can add a dialect attribute verifier to `stream` tAt'll check the format of these attributes

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
```diff
@@ -341,13 +341,76 @@ class ConvertFunc : public ConvertOpToLLVMPattern<func::FuncOp> {
...
+      // - Correlation: bindings in the same group point to the same resource
+      //   but with different offsets (noalias).
+      // - Noalias: bindings pointing to different resources (noalias).
+      auto correlationIt = bindingCorrelationMap.find(idx);
+      auto noaliasIt = bindingNoaliasMap.find(idx);
+      
+      if (correlationIt != bindingCorrelationMap.end() && !correlationIt->second.empty()) {
+        // This binding is in a correlation group. The bindings in the same
+        // group don't alias each other (different offsets).
+        // TODO: In the future, we could use alias.scope metadata to be more
```
krzysz00: Ah, this is where the code will go for annotating the different subspans of a binding as noasias etc.

Some of that would be made simpler by something like https://discourse.llvm.org/t/n-ary-separate-objects-non-argument-noalias-groups/88420 landing.

I can also flag #22236 as an interesting related piece of work

krzysz00 (commented):
Is this going to be stacked on top of another PR that implements these TODOs? Otherwise, I figure gesturing at a feature that doesn't quite exist doesn't really make sense

MaheshRavishankar (commented):
Not reviewed the code yet, but I think for the code that lowers to LLVM GPU, we should be OK just adding noalias to all operands. The fact that they are the same bindings shouldnt matter, cause the generated LLVM function takes a different pointer for each argument. One or more pointers could have been set by using the same base but different offsets, but LLVM shouldnt care about this.

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
kimm240: I fix it in commit 4c56b2c.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
kimm240: I fix it in commit commit  0638ea0.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
kimm240: Thanks for the guidance! I'm very grateful for the high-level context and the links you shared.

Regarding the TODOs: I agree that gesturing at features that don't yet exist can be confusing. I've removed those comments in commit faf30dd.

However, seeing the discussions in the LLVM community you shared, I’m open to exploring the implementation of that.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
```diff
@@ -349,13 +341,58 @@ class ConvertFunc : public ConvertOpToLLVMPattern<func::FuncOp> {
...
     // Set argument attributes.
     Attribute unit = rewriter.getUnitAttr();
+
+    // Build a map from HAL binding index to correlation group and noalias
+    // bindings. This allows us to determine which bindings are in the same
+    // correlation group (i.e., point to the same resource but with different
+    // offsets) and which bindings point to different resources (noalias).
+    DenseMap<int64_t, SmallVector<int64_t>> bindingCorrelationMap;
+    DenseMap<int64_t, SmallVector<int64_t>> bindingNoaliasMap;
+    for (auto subspan : subspans) {
```
kuhar: spell out the type

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
```diff
@@ -349,13 +341,58 @@ class ConvertFunc : public ConvertOpToLLVMPattern<func::FuncOp> {
...
       // aligned.
       newFuncOp.setArgAttr(idx, LLVM::LLVMDialect::getAlignAttrName(),
                            rewriter.getI32IntegerAttr(16));
+
+      // Check if this binding has correlation or noalias information.
+      // - Correlation: bindings in the same group point to the same resource
+      //   but with different offsets (noalias).
+      // - Noalias: bindings pointing to different resources (noalias).
+      auto correlationIt = bindingCorrelationMap.find(idx);
+      auto noaliasIt = bindingNoaliasMap.find(idx);
```
kuhar: These seem unused

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -191,15 +190,50 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
   SmallVector<BlockArgument> newBindingArgs;
   auto bindingType = IREE::Stream::BindingType::get(funcOp.getContext());
   auto offsetType = IndexType::get(funcOp.getContext());
-  for (auto &binding : bindings) {
+  for (auto binding : llvm::enumerate(bindings)) {
```
kuhar: Use structured bindings for the index and the value

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -191,15 +190,50 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
...
     SmallVector<Location> locs;
-    for (unsigned oldIdx : binding.correlationMap.set_bits()) {
+    for (unsigned oldIdx : binding.value().correlationMap.set_bits()) {
```
kuhar:
```suggestion
    for (unsigned oldIdx : binding->correlationMap.set_bits()) {
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -191,15 +190,50 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
...
+    // This tells us which other bindings are in the same correlation group
+    // (i.e., point to the same resource but with different offsets, so they
+    // don't alias each other).
+    //
+    // For bindings in different correlation groups (pointing to different
+    // resources), we also store this information so that LLVM codegen can
+    // infer that they don't alias each other (noalias).
+    SmallVector<Attribute> correlatedIndices;
+    SmallVector<Attribute> noaliasIndices;
+    for (auto otherBinding : llvm::enumerate(bindings)) {
```
kuhar: use structured bindings for the index and the value

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -191,15 +190,50 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
...
+      if (binding.index() == otherBinding.index()) {
+        continue;
+      }
+
+      if (binding.value().correlationMap ==
+          otherBinding.value().correlationMap) {
+        // Same correlation group: same resource, different offsets (noalias)
+        correlatedIndices.push_back(IntegerAttr::get(
+            IntegerType::get(funcOp.getContext(), 32), otherBinding.index()));
```
kuhar: create a `Builder` and do something like `b.getI32Attr(...);`

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -191,15 +190,50 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
...
+      } else {
+        // Different correlation groups: different resources (noalias)
+        noaliasIndices.push_back(IntegerAttr::get(
+            IntegerType::get(funcOp.getContext(), 32), otherBinding.index()));
```
kuhar: Also here

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -191,15 +190,50 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
...
+    SmallVector<Attribute> correlatedIndices;
+    SmallVector<Attribute> noaliasIndices;
+    for (auto otherBinding : llvm::enumerate(bindings)) {
+      if (binding.index() == otherBinding.index()) {
+        continue;
+      }
+
+      if (binding.value().correlationMap ==
+          otherBinding.value().correlationMap) {
+        // Same correlation group: same resource, different offsets (noalias)
```
kuhar:
```suggestion
        // Same correlation group: same resource, different offsets (noalias).
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
```diff
@@ -191,15 +190,50 @@ static void updateExecutableSignature(IREE::Stream::ExecutableOp executableOp,
...
+        correlatedIndices.push_back(IntegerAttr::get(
+            IntegerType::get(funcOp.getContext(), 32), otherBinding.index()));
+      } else {
+        // Different correlation groups: different resources (noalias)
```
kuhar:
```suggestion
        // Different correlation groups: different resources (noalias).
```

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
kimm240: I fix it in commit 4cf303d

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
kimm240: I fix it in  commit [632dc52](https://github.com/iree-org/iree/pull/23184/commits/632dc5269509278854152da375e3e85db25aeadc)

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
kimm240: Ifix it in commit [78705c9](https://github.com/iree-org/iree/pull/23184/commits/78705c981d46553f7222766c7009411f02df940b).

Other  structure  binding  issue is also fixed in that commit.

compiler/src/iree/compiler/Dialect/Stream/Transforms/FuseDispatchBindings.cpp
kimm240: I  fix it in commit [0c8135e](https://github.com/iree-org/iree/pull/23184/commits/0c8135ed3ff800afdae71befbd4baad7d690f3e1).

Other Builder issue is also fixed in that commit.

kimm240:
@MaheshRavishankar 
The reason I implemented the correlation tracking is to be conservative and safe. While most GPU dispatch arguments don't overlap, I was concerned about cases where multiple bindings might point to the exact same resource (true aliasing).

If we can guarantee that IREE's high-level flow/stream always produces non-overlapping subspans for separate bindings, then simply applying noalias to all operands would be simpler.

MaheshRavishankar:
> @MaheshRavishankar 
> The reason I implemented the correlation tracking is to be conservative and safe. While most GPU dispatch arguments don't overlap, I was concerned about cases where multiple bindings might point to the exact same resource (true aliasing).
> 
> If we can guarantee that IREE's high-level flow/stream always produces non-overlapping subspans for separate bindings, then simply applying noalias to all operands would be simpler.

Yes. It is guaranteed.

kuhar:
Two subspans into the same binding may overlap, distinct bindings do not

kimm240:
I completely agree that applying a blanket noalias to all pointer arguments is a simpler approach and works effectively under IREE's current guarantee that distinct bindings never overlap.

However, I have preserved the correlation tracking logic in this PR to prepare for the following future scenarios:
Foundation for `alias.scope` Metadata: 

As discussed in the [LLVM forum thread](https://www.google.com/url?sa=E&q=https%3A%2F%2Fdiscourse.llvm.org%2Ft%2Fn-ary-separate-objects-non-argument-noalias-groups%2F88420), the next step for maximizing GPU performance is moving beyond function attributes to granular `!alias.scope` and `!noalias` metadata. And  My implementation provides the exact 'pointer grouping' information required to emit this metadata.

While distinct bindings don't overlap, subspans within the same binding might. The correlation logic is the only way to distinguish these cases. By keeping this infrastructure, we can safely optimize instructions even when multiple subspans are fused into a single binding argument.

Relying on a global "no-overlap" guarantee is fine for now, but explicit correlation makes the codegen much more resilient to future changes in resource management or binding fusion strategies.

I’d appreciate it if we could keep this logic as a foundation for upcoming metadata work.

MaheshRavishankar:
> Two subspans into the same binding may overlap, distinct bindings do not

Are you sure? Two subspans into the same binding I think do not overlap. Thats how the combine bindings in Stream works. @benvanik  could you confirm that in a dispatch the binding subspans do not overlap.

kuhar:
> > Two subspans into the same binding may overlap, distinct bindings do not
> 
> Are you sure? Two subspans into the same binding I think do not overlap.

I remember debugging an elementwise dispatch that had two inputs coming from the same buffer at different base offsets.

MaheshRavishankar:
Ok, maybe I am getting confused. I haven't looked at in a while. The subspans are independent, that is guaranteed. Whether the offset is added to the base inside the kernel or outside. I don't remember now. Would be good to check

krzysz00:
... the subspan condition is actually slightly weaker, per a bunch of previous discussions I had with @benvanik on the `writeonly` stuff. The summary of that is that, as best as I can recall, we obey Rust rules here: a given region of memory can either be covered by N immutable subspans (including 0) or exactly 1 mutable subspan.

So multiple read-only subspans aliasing each other is allowed, though one could probably get away with `noalias`ing them anyway.

MaheshRavishankar:
> ... the subspan condition is actually slightly weaker, per a bunch of previous discussions I had with @benvanik on the `writeonly` stuff. The summary of that is that, as best as I can recall, we obey Rust rules here: a given region of memory can either be covered by N immutable subspans (including 0) or exactly 1 mutable subspan.
> 
> So multiple read-only subspans aliasing each other is allowed, though one could probably get away with `noalias`ing them anyway.

Well, then if there is a way to say the reads dont alias with the writes, thats enough I guess. I am not blocking on this PR... but the logic seems more complicated, and I thought we could just use some existing gaurantees. I am not reviewing this actively. If @kuhar  and @krzysz00  are fine with this, it works for me.

kimm240:
Hi all! Regarding the previous discussion, as @kuhar and @krzysz00 mentioned, read-only subspans can indeed overlap, so I believe the current correlation tracking logic is necessary for safety and for the future alias.scope metadata work.

If you think any further refinements or additional implementations are needed to better align this with IREE's architecture, please let me know. I’m more than happy to follow up and implement them.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
```diff
@@ -349,13 +341,46 @@ class ConvertFunc : public ConvertOpToLLVMPattern<func::FuncOp> {
...
+      int64_t binding = subspan.getBinding().getSExtValue();
+      // Try to read correlation information from the original function.
+      // Note: This assumes the correlation attribute was preserved through
+      // Stream → HAL conversion. If not, we'll need to preserve it explicitly.
+      if (auto correlationAttr = funcOp.getArgAttrOfType<ArrayAttr>(
+              binding, "stream.binding_correlation")) {
+        SmallVector<int64_t> correlatedBindings;
+        for (IntegerAttr intAttr : correlationAttr.getAsRange<IntegerAttr>()) {
+          correlatedBindings.push_back(intAttr.getInt());
+        }
```
kuhar: nit/optional: this can be


```suggestion
        auto correlatedBindings = llvm::map_to_vector(correlationAttr.getAsRange<IntegerAttr>(), [] (IntegerAttr attr) { return attr.getInt(); });
```

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
```diff
@@ -349,13 +341,46 @@ class ConvertFunc : public ConvertOpToLLVMPattern<func::FuncOp> {
...
+        if (!correlatedBindings.empty()) {
+          bindingCorrelationMap[binding] = std::move(correlatedBindings);
+        }
+      }
+      // Read noalias information (bindings pointing to different resources).
+      if (auto noaliasAttr = funcOp.getArgAttrOfType<ArrayAttr>(
+              binding, "stream.binding_noalias")) {
+        SmallVector<int64_t> noaliasBindings;
+        for (IntegerAttr intAttr : noaliasAttr.getAsRange<IntegerAttr>()) {
```
kuhar: similar here

kuhar (commented):
Can we add some tests for this?

kimm240:
@kuhar 
This test confirms that `FuseDispatchBindings` correctly generates `stream.binding_noalias` when distinct resources are allocated and `alias_mutable_bindings` is set to false. It ensures the analysis logic properly identifies non-aliasing bindings at the Stream level.

Please Give me any feedback if there is an issue  in test.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
kimm240: I fix it in  commit 07c1c37.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_nvvm_noalias_e2e.mlir
```diff
@@ -0,0 +1,41 @@
+// RUN: iree-compile --iree-hal-target-device=cuda --iree-cuda-target=sm_86 \
+// RUN:     --mlir-print-ir-after-all --mlir-disable-threading=true %s -o /dev/null 2>&1 \
```
kuhar: We generally don't test against the full debug output in IREE -- either break it up to test individual passes or check against the output of a full pipeline

kimm240: @kuhar 
Apologies, I accidentally included a test case intended for local debugging in my previous push. I have removed it in the latest commit and updated the PR.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
```diff
@@ -349,13 +341,44 @@ class ConvertFunc : public ConvertOpToLLVMPattern<func::FuncOp> {
...
+    // bindings. This allows us to determine which bindings are in the same
+    // correlation group (i.e., point to the same resource but with different
+    // offsets) and which bindings point to different resources (noalias).
+    DenseMap<int64_t, SmallVector<int64_t>> bindingCorrelationMap;
+    DenseMap<int64_t, SmallVector<int64_t>> bindingNoaliasMap;
+    for (IREE::HAL::InterfaceBindingSubspanOp subspan : subspans) {
+      int64_t binding = subspan.getBinding().getSExtValue();
+      // Try to read correlation information from the original function.
+      // Note: This assumes the correlation attribute was preserved through
+      // Stream → HAL conversion. If not, we'll need to preserve it explicitly.
```
kuhar: nit: stick to ascii
```suggestion
      // Stream -> HAL conversion. If not, we'll need to preserve it explicitly.
```

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_nvvm_noalias.mlir
```diff
@@ -0,0 +1,87 @@
...
+// groups and should have stream.binding_noalias attributes.
+//
+// STREAM-LABEL: util.func public @dispatch
+// STREAM-SAME: (%[[ARG0:.+]]: !stream.binding {stream.binding_noalias = [1 : i32]},
+// STREAM-SAME:  %[[ARG1:.+]]: !stream.binding {stream.binding_noalias = [0 : i32]}
+
+// After ConvertToNVVMPass: llvm.noalias attributes should be applied.
+// LLVM-LABEL: llvm.func @dispatch
+// LLVM-SAME: (%[[LLVM_ARG0:.+]]: !llvm.ptr {llvm.align = 16 : i32, llvm.noalias, llvm.nonnull, llvm.noundef},
+// LLVM-SAME:  %[[LLVM_ARG1:.+]]: !llvm.ptr {llvm.align = 16 : i32, llvm.noalias, llvm.nonnull, llvm.noundef}
```
kuhar: These are never checked

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_nvvm_noalias.mlir
```diff
@@ -0,0 +1,87 @@
...
+
+// Test that stream.binding_noalias attributes are correctly generated in Stream
+// dialect and propagated to llvm.noalias attributes in LLVM dialect.
+//
+// This test verifies the complete pipeline as described in the PR:
+// 1. Stream dialect: After FuseDispatchBindings pass, stream.binding_noalias
+//    attributes are generated on func.func arguments for distinct bindings
+//    (different resources)
+// 2. LLVM dialect: After ConvertToNVVMPass, llvm.noalias attributes are
+//    applied to function arguments
```
kuhar: this one is missing from the test

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp
kimm240: I fix it in commit [ca20bad](https://github.com/iree-org/iree/pull/23184/commits/ca20bad2b30809f7c15fac83a78859f82edaa7e4)

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_nvvm_noalias.mlir
kimm240: @kuhar 
I add test for LLVM dialect in commit 6c07283

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_nvvm_noalias.mlir
kimm240: I delete it for clarity.

kimm240:
Hi @kuhar @krzysz00, just a gentle ping on this PR.
I addressed all @kuhar 's previous feedback, including adding the requested NVVM/LLVM dialect tests and fixing the ascii comments, back in early February.
Could you please take another look when you have a chance? Please let me know if there are any further refinements needed. Thanks for your time!

kuhar:
Can you rebase this? We plan to have a release in a couple of days, I'd hold off with merging this until it's out.

kimm240:
Hi @kuhar, thank you for the update. I have already rebased this PR onto the latest main. I'm happy to wait until the upcoming release is out. Thanks again for the review!

kimm240:
Hi @kuhar! As requested, I've rebased the PR onto the latest main branch again. Please let me know when you have a chance to take a look. Thanks!

kimm240:
@kuhar  @krzysz00  @bangtianliu 
Prev CI is reproducibly failing on llama_8b_fp16/quality_gfx942 with an output mismatch, so this may indicate a correctness issue. (Latest Commit may solve it, but I can't sure what it is.)

If this is confirmed as a correctness bug, we likely need to decide whether this is:
1) acceptable drift from aliasing-related optimization (and therefore an expected/weight update case), or
2) a logic issue that requires making noalias propagation more conservative.

I can’t reliably reproduce this exact test locally in my current environment, so I can’t confidently conclude which direction is correct yet.

Guidance on the preferred direction would be greatly appreciated.

kimm240:
Change FfileCheck for this pr

krzysz00:
Re the CI failures, you may have run into a pile of flakes - could you try rebasing? (Sorry for the lack of movement here, was travelling etc.)

kimm240:
Hi @krzysz00! I've rebased the PR onto the latest main. Please let me know if there are any further issues to address. Also, hope you had a great trip!

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_nvvm.mlir
```diff
@@ -31,9 +31,9 @@ hal.executable @abs_ex_dispatch_0 {
   }
 }
 // CHECK-LABEL: llvm.func @abs_ex_dispatch_0
-//  CHECK-SAME: (%[[ARG0:.+]]: !llvm.ptr {llvm.align = 16 : i32, llvm.noalias, llvm.nonnull, llvm.noundef},
-//  CHECK-SAME:  %[[ARG1:.+]]: !llvm.ptr {llvm.align = 16 : i32, llvm.noalias, llvm.nonnull, llvm.noundef, llvm.readonly},
-//  CHECK-SAME:  %[[ARG2:.+]]: !llvm.ptr {llvm.align = 16 : i32, llvm.noalias, llvm.nonnull, llvm.noundef})
+//  CHECK-SAME: (%[[ARG0:.+]]: !llvm.ptr {llvm.align = 16 : i32, llvm.nonnull, llvm.noundef},
+//  CHECK-SAME:  %[[ARG1:.+]]: !llvm.ptr {llvm.align = 16 : i32, llvm.nonnull, llvm.noundef, llvm.readonly},
```
krzysz00: ... these should still be `noalias` - each binding is distinct.

... really, this should be an NFC PR if it even exists at all, now that I think about, since over in LLVM it's (last I checked, by way of "the rust people do this, I'm pretty sure") fine to `noalias` a read-only pointer even if something aliasing exists in the scope so long as there's no modifications.

krzysz00: .... Or, more accurately, the solution to this test is to add the `stream.noalias`es in here

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_nvvm.mlir
krzysz00, line 195: These probably should be given their stream attributes

krzysz00 (commented):
Reading again, I'm not sure that the function attributes with the suffixed number work. This probably should be an array of arrays on the pipeline or something like that.

(Sorry for pulling this around, the design issue must have slipped my mind last time)

kimm240:
@krzysz00 Thanks for the design feedback! As you suggested, I've completely updated the logic to use a single function-level 'Array of Arrays' attribute instead of the individual stream.binding_noalias_<idx> attributes.(commit ebb73ad)

I've also updated all the test cases to align with this new structure. With this change, llvm.noalias is now properly restored for the bindings in use. (commit  6e897f6)

Please let me know if there is anything else that needs to be addressed!

krzysz00 (commented):
High-level question with this PR: if I take multiple subspans from the same `readonly` binding, will that still be marked `noalias`?

(Observe `rustc -O` emitting something similar on that case https://godbolt.org/z/TnxPdWYMr )

That is, let's add a test for this.

(I think one way to phrase that would be saying that readonly bindings are noalis with *themselves*)

krzysz00:
And thank you so much for your work on this!

kimm240:
@krzysz00 Thanks for the suggestion! I've added a new test case, `same_binding_readonly_multi_subspan`, to verify the scenario you mentioned in commit 91e3b67.

- Regarding multiple subspans within the same binding:
The new test confirms that multiple ReadOnly subspans from the same binding are correctly deduplicated into a single kernarg. 

- Regarding automatic noalias for ReadOnly bindings:
It may carry a risk of data inconsistency if a writeable binding in the same function aliases with those ReadOnly bindings.

Given the need for careful safety checks between readers and writers, I believe it would be better to address this specific 'readonly-only' optimization in a separate follow-up PR. I’d appreciate your thoughts on this approach.

Also, I wanted to express my gratitude for this opportunity. I’ve learned so much through this process thanks to your valuable insights and guidance. It is truly an honor to contribute to a project like IREE!

krzysz00:
FYI your CI failures are you hitting asserts during tests and seem to be real

kimm240:
@krzysz00  I found that the invalid argument number assertion was triggered during the HAL-to-LLVM conversion because func.func in the HAL dialect often has zero arguments, while my logic was attempting to access attributes via binding indices directly.
 
I'll monitor the remaining CI checks closely. Please let me know if you spot any other potential issues! Thanks!

Q. Is result of test_mi325 real?

krzysz00:
>  It may carry a risk of data inconsistency if a writeable binding in the same function aliases with those ReadOnly bindings.

It's an invariant that a region of memory is covered by 1 writable binding or N read-only bindings - that is, we have Rust semantics

kimm240:
@krzysz00 Most of the CI runs are green except for the AMD MI325 and MI355 tests. Is there a recommended way or a guide to debug these target-specific failures? Any guidance would be greatly appreciated.

PR #23186, opened 2026-01-19
[Encoding] Drop experimental i1 packing flag

sommerlukas:
Remove the experimental `--iree-experimental-packed-i1-storage` flag. The experimental flag was used to indicate packing of `i1` values (instead of extending them individually to `i8`).

Frontend or users that want packing of `i1` should apply the `packed_storage` encoding to relevant tensors instead of using the option.


sommerlukas:
This pull request is currently draft as it depends on https://github.com/llvm/llvm-project/pull/176427 to be integrated via LLVM integration.

hanhanW:
> This pull request is currently draft as it depends on [llvm/llvm-project#176427](https://github.com/llvm/llvm-project/pull/176427) to be integrated via LLVM integration.

If it depends on an upstream change, it seems to be a functionality change somewhere? Why do we need that? (I haven't reviewed the code yet because it is draft. I only read the description and comments.)

sommerlukas:
The IR where this becomes relevant is the following (see `@truncate_i1_2` in `subbyte_types.mlir` for the full code).

We start off with a `tensor.empty<4x4xi1, #iree_encoding.packed_storage>`.

Early on, we have a pass that collapses tensor shapes. This gives us something like:
```
%0 = tensor.empty<4x4xi1, #iree_encoding.packed_storage>
%1 = tensor.collapse_shape %0 ... : tensor<16xi1, #iree_encoding.packed_storage>
```

Without the change in upstream, this gets folded into:
```
%0 = tensor.empty<16xi1>
%1 = tensor.cast %0 to tensor<16xi1, #iree_encoding.packed_storage>
```

After the changes from this PR `TypePropagation` relies on tensors having the right encoding if they are meant to be packed, but `%0` doesn't have the encoding.

The IR without the upstream change would then be transformed into:
```
%0 = tensor.empty<16xi8> // No packing applied here, because there is no encoding on the tensor.
%1 = tensor.cast %0 to tensor<16xi1> // This is obviously wrong, we can't cast 16xi8 to 16xi1
```

With the change in upstream, we get:
```
%0 = tensor.empty<16xi1, #iree_encoding.packed_storage
```

Which is correctly transformed by `TypePropagation`.

The relevant functionality change is that `TypePropagation` needs the encoding to be present, if there is no global option telling it to pack all `i1`s.

This is also the reason why CI fails on `subbyte_types.mlir` without the upstream change.

sommerlukas:
Necessary upstream changes have landed through integrate, this PR is ready for review.

hanhanW (approved):
Nice cleanup, thanks for all the contribution for the workstream!

lialan (approved):
LGTM!

PR #21138, opened 2025-06-19
[LinalgExt] support converting argcompare to loops.

bangtianliu:
On top of #21106.

This PR Adds `generateScalarImplementation` for ArgCompareOp, defining the scalar behavior within the innermost loop during loop lowering.

The lowering performs the following: 
- Loads the current best value and corresponding index (accumulator) from the output tensor.
- Loads the candidate value at the current loop indices.
- Invokes the provided comparator region (T a, T b) -> i1 to compare values.
- Selects the new best value and index based on the comparison result.
- Supports an optional index_base operand, which is added to the computed index before storing.

Next, will add tiling support.

PR #21077, opened 2025-06-11
[LinalgExt] add TilingInterface support for ArgCompareOp

bangtianliu:
This PR adds tiling support for the `iree_linalg_ext.argmax` operation by implementing the tiling interface. The following methods are introduced:
- `getLoopIteratorTypes`
- `getResultTilePosition`
- `getTiledImplementation`
- `generateResultTileValue`

Additionally, corresponding test cases are provided to verify the tiling behavior for both tensor and memref cases.

PR #23193, opened 2026-01-19
[LinalgExt] Extend arg_compare tiling interface for explicit-index mode

bangtianliu:
This PR extends the tiling interface implementation for `ArgCompareOp` to support explicit-index mode (2 inputs: value + pre-computed indices), which was introduced in PR #23153.

 **Key changes:**
  - Extended `getTiledImplementation`  (introduced in PR #21077) to slice both value and index inputs in explicit-index mode
  - Extended `generateScalarImplementation` (convert_to_loops) (introduced in PR #21138) to load indices from input instead of computing from iteration variables in explicit-index mode

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
@@ -1553,6 +1553,13 @@ ArgCompareOp::getTiledImplementation(OpBuilder &builder,
   tiledOperands.push_back(inputSlice->getResult(0));
   slices.push_back(inputSlice);
 
+  if (hasExplicitIndexInput()) {
+    Operation *inputIndexSlice =
+        getSlice(builder, loc, getInputIndex(), offsets, sizes, strides);
+    tiledOperands.push_back(inputIndexSlice->getResult(0));
+    slices.push_back(inputIndexSlice);
+  }
+
   SmallVector<OpFoldResult> outputOffsets, outputSizes;
   if (failed(getResultTilePosition(builder, 0, offsets, sizes, outputOffsets,
                                    outputSizes))) {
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1646,18 +1653,30 @@ LogicalResult ArgCompareOp::generateScalarImplementation(OpBuilder &b,
...
-                                             castedIndex);
+
+  Value candidateIndex;
+  if (hasExplicitIndexInput()) {
+    // Explicit-index mode: load from input.
+    // The verifier ensures input and output index types match, so no cast
+    // needed.
+    candidateIndex = memref::LoadOp::create(b, loc, getInputIndex(), ivs);
+  } else {
+    // Implicit-index mode: compute from iteration variable.
```
kuhar:
```suggestion
    // Implicit-index mode: compute from induction variable.
```

? If that's not it, maybe it's worth renaming this function argument.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1646,18 +1653,30 @@ LogicalResult ArgCompareOp::generateScalarImplementation(OpBuilder &b,
...
+    Value indexOffset = ivs[reductionDim];
+    if (getIndexBase()) {
+      indexOffset = arith::AddIOp::create(b, loc, getIndexBase(), indexOffset);
+    }
+    candidateIndex = indexOffset;
+
+    // Cast if iteration variable type doesn't match output index type.
```
kuhar: also here

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1646,18 +1653,30 @@ LogicalResult ArgCompareOp::generateScalarImplementation(OpBuilder &b,
...
-  if (getIndexBase()) {
-    indexOffset = arith::AddIOp::create(b, loc, getIndexBase(), indexOffset);
-  }
-  Value castedIndex = indexOffset;
-  if (castedIndex.getType() != bestIndexSoFar.getType()) {
-    castedIndex = arith::IndexCastOp::create(b, loc, bestIndexSoFar.getType(),
-                                             castedIndex);
+
+  Value candidateIndex;
+  if (hasExplicitIndexInput()) {
```
IanWood1, line 1658: If we don't use index base when in explicit index mode, should we make it illegal in the verifier?

IanWood1 (approved):
Looks good, but maybe we want to disallow index_base when in explicit index mode?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
bangtianliu: Absolutely, that is what I plan to do in the next PR.

kuhar (approved):

PR #23197, opened 2026-01-19
New distribution tile heuristic for CPU data-tiled matmuls.

bjacob:
Looked at our performance on a hundred matmul shapes; hadn't done so in a long time. The new heuristic provides substantial [speedups](https://docs.google.com/spreadsheets/d/1TiLAsmQuEqbze0BvjS6g_KskiCPMUsSNiHBrq9KPxks/edit?usp=sharing) across many shapes (> 1.2x on larger shapes, sometimes > 1.5x, never a regression). It also makes more sense heuristically.

A key finding was that the tile sizes that we were getting out of `getDefaultDistributedLevelTileSizes` were not at all the `maxTileSizes` that we gave it as its input. The effective tile sizes were chosen to be divisors of the tensors' dimensions. As a result, the only way to have precise control over tile sizes is to directly choose them as divisors. Then we don't need to call `getDefaultDistributedLevelTileSizes` anymore. This PR does that, which is why the new heuristic explicitly iterates over divisors.

The previous heuristic was meant to be based on two cache-size-ish parameters but that never really worked. The new heuristic has a single cache-size-ish parameter, `iree-llvmcpu-matmul-tile-bytes`, and tuning shows that it really works as intended. The docstring on the command-line flag has details.

There is a second command-line parameter, `iree-llvmcpu-matmul-tile-undercount-whole-matrix`, to allow growing tiles past the limit for the sake of fitting entire matrices in one tile.

Data collected on EPYC 9575F, single-threaded, bf16 matmuls: https://docs.google.com/spreadsheets/d/1TiLAsmQuEqbze0BvjS6g_KskiCPMUsSNiHBrq9KPxks/edit?usp=sharing

The focus on single-threaded is intentional: that is the primary server use case. However, as far as this heuristic is concerned, to a reasonable first approximation, tuning for N threads is a simple matter of dividing the `iree-llvmcpu-matmul-tile-bytes` parameter by N. There is some wiggle room though; the default 4M tile size is smaller than many CPUs' L3 cache and thus will still work unchanged up to several threads. Just be aware that if you ever want to tune for multi-thread matmuls on CPU, you may need to reduce `iree-llvmcpu-matmul-tile-bytes`.

A vaguely reasonable default "cache size" is estimated based on the target architecture, which is very little information but is all we have. Users are expected to override `iree-llvmcpu-matmul-tile-bytes` anyway, but do feel free to improve the defaults.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -2055,9 +2055,9 @@ getMmt4dLoweringConfig(linalg::LinalgOp op, DictionaryAttr targetConfig) {
   auto staticSizeOr = [](int64_t value, int64_t valueIfDynamic) {
     return ShapedType::isDynamic(value) ? valueIfDynamic : value;
   };
-  const int64_t M1 = staticSizeOr(lhsShape[mmt4dDimBase + 0], 1024);
-  const int64_t N1 = staticSizeOr(rhsShape[mmt4dDimBase + 0], 1024);
-  const int64_t K1 = staticSizeOr(lhsShape[mmt4dDimBase + 1], 1024);
+  const int64_t M1 = staticSizeOr(lhsShape[mmt4dDimBase + 0], 64);
+  const int64_t N1 = staticSizeOr(rhsShape[mmt4dDimBase + 0], 64);
+  const int64_t K1 = staticSizeOr(lhsShape[mmt4dDimBase + 1], 64);
 
   // M0, N0, K0 should almost always be static, but there could be an exception
   // on targets like Arm SVE, so just to be safe, also normalize.
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -1988,33 +1988,30 @@ static bool adjustVectorSizesForScalableVectorization(
   return false;
 }
 
-static SmallVector<std::pair<int, int>> getDivisorPairs(int n) {
-  SmallVector<std::pair<int, int>> divisorPairs;
-  for (int i = 1; i * i <= n; ++i) {
-    int j = n / i;
+static SmallVector<std::pair<int64_t, int64_t>> getDivisors(int64_t n) {
+  SmallVector<std::pair<int64_t, int64_t>> divisors;
+  for (int64_t i = 1; i * i <= n; ++i) {
+    int64_t j = n / i;
     if (i * j == n) {
-      divisorPairs.push_back({i, j});
+      divisors.push_back({i, j});
       if (j != i) {
-        divisorPairs.push_back({j, i});
+        divisors.push_back({j, i});
       }
     }
   }
-  llvm::sort(divisorPairs, [](auto p, auto q) { return p.first < q.first; });
-  return divisorPairs;
+  llvm::sort(divisors, llvm::less_first{});
+  return divisors;
 }
 
 static int getMatmulTileBytes(Operation *op) {
   if (clMatmulTileBytes) {
     return clMatmulTileBytes;
   }
-  IREE::HAL::ExecutableTargetAttr target =
-      IREE::HAL::ExecutableTargetAttr::lookup(op);
   // Generally we want to return a value that is between L2 and L3 cache sizes.
   // We don't have that information at compile time, so we have to make a
   // reasonable guess.
-  if (target) {
-    auto config = target.getConfiguration();
-    if (config) {
+  if (auto target = IREE::HAL::ExecutableTargetAttr::lookup(op)) {
+    if (DictionaryAttr config = target.getConfiguration()) {
       if (isX86_64(config) && hasFeature(config, "+avx512f")) {
         // Recent or server X86.  This value was manually tuned in benchmarks on
         // EPYC 9575F on single-threaded bf16 matmuls.
@@ -2027,6 +2024,7 @@ static int getMatmulTileBytes(Operation *op) {
       }
     }
   }
+
   // Default, low-ish value.
   return 1 * 1024 * 1024;
 }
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1975,25 +1988,67 @@ static bool adjustVectorSizesForScalableVectorization(
...
+  for (int i = 1; i * i <= n; ++i) {
+    int j = n / i;
+    if (i * j == n) {
+      divisorPairs.push_back({i, j});
+      if (j != i) {
+        divisorPairs.push_back({j, i});
+      }
+    }
+  }
+  llvm::sort(divisorPairs, [](auto p, auto q) { return p.first < q.first; });
```
kuhar:
```suggestion
  llvm::sort(divisorPairs, llvm::less_first{});
```

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1975,25 +1988,67 @@ static bool adjustVectorSizesForScalableVectorization(
...
+    return clMatmulTileBytes;
+  }
+  IREE::HAL::ExecutableTargetAttr target =
+      IREE::HAL::ExecutableTargetAttr::lookup(op);
+  // Generally we want to return a value that is between L2 and L3 cache sizes.
+  // We don't have that information at compile time, so we have to make a
+  // reasonable guess.
+  if (target) {
+    auto config = target.getConfiguration();
+    if (config) {
```
kuhar:
```suggestion
    if (auto config = target.getConfiguration();) {
```
I'd also spell this out

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -2002,34 +2057,62 @@ getMmt4dLoweringConfig(linalg::LinalgOp op, DictionaryAttr targetConfig) {
...
-  auto getMatmulTileSize = [](int64_t targetTileBytes, int bitWidth,
-                              int64_t reductionSize, int64_t tile0Size) {
-    int64_t targetRhsTileElems = targetTileBytes * 8 / bitWidth;
-    int64_t targetRhsTileNSize = targetRhsTileElems / reductionSize;
-    int64_t tileSize = llvm::divideCeil(targetRhsTileNSize, tile0Size);
-    tileSize = std::max(tileSize, int64_t{1});
-    return tileSize;
+
+  auto divisorPairsOfM1 = getDivisorPairs(M1);
+  auto divisorPairsOfN1 = getDivisorPairs(N1);
```
kuhar: nit: spell this out or use structured bindings?

kuhar (commented):
Just drive-by nits, I don't know this code

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1975,25 +1988,67 @@ static bool adjustVectorSizesForScalableVectorization(
   return false;
 }
 
+static SmallVector<std::pair<int, int>> getDivisorPairs(int n) {
```
hanhanW: most of dim sizes are using `int64_t`, should we use `int64_t` here as well?

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -1975,25 +1988,67 @@ static bool adjustVectorSizesForScalableVectorization(
...
+      if (isAArch64(config) &&
+          (hasFeature(config, "+sve") || hasFeature(config, "+i8mm"))) {
+        // Recent Arm64. Value set from looking up some recent SoC specs.
+        return 4 * 1024 * 1024;
+      }
+    }
+  }
+  // Default, low-ish value.
+  return 1 * 1024 * 1024;
+}
```
hanhanW, line 2038: nit: add a new line between the two methods.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -2002,34 +2057,62 @@ getMmt4dLoweringConfig(linalg::LinalgOp op, DictionaryAttr targetConfig) {
...
+  int64_t selectedTileN = 1;
+  int64_t selectedCost = evalTraversalCost(M1, N1);
+  const int matmulTileBytes = getMatmulTileBytes(op);
+
+  // Iterate over all candidate tile shapes, which are the divisors of (M1, N1).
+  for (auto [tileM, numTilesM] : divisorPairsOfM1) {
+    for (auto [tileN, numTilesN] : divisorPairsOfN1) {
+      // Compute candidate tile size in bytes.
+      int64_t lhsBytes = tileM * M0 * K1 * K0 * lhsTypeBits / 8;
+      int64_t rhsBytes = tileN * N0 * K1 * K0 * rhsTypeBits / 8;
```
hanhanW, line 2154: What happens when `K1` is dynamic? Should we add a guard like `if (ShapedType::isDynamic(K1)) { ... } `?

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
hanhanW: Just a slight improvement that sorting is not needed if we use two vectors.

```suggestion
  SmallVector<std::pair<int, int>> smallDivisorFirst;
  SmallVector<std::pair<int, int>> largeDivisorFirst;
  for (int i = 1; i * i <= n; ++i) {
    int j = n / i;
    if (i * j == n) {
      smallDivisorFirst.push_back({i, j});
      if (j != i) {
        largeDivisorFirst.push_back({j, i});
      }
    }
  }
  smallDivisorFirst.append(largeDivisorFirst.rbegin(), largeDivisorFirst.rend());
```

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
bjacob: Good catch. The old code handled that; making the new code also handle that.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
bjacob: Sorting results in simpler, shorter code and more easy to read invariants: now the array is guaranteed to be sorted, not depending on the code being correct.

The performance impact of sorting is negligible here.

Note also that sorting doesn't matter much. It only ever makes a difference when there are ties, since the first met entry will take precedence. The only purpose of sorting here is to ensure that results don't accidentally change when the implementation of this function changes.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
```diff
@@ -115,7 +115,7 @@ func.func @mmt4d_tensors(%arg0: tensor<32x128x8x1xf32>, %arg1 : tensor<?x128x?x1
   %mmt4d = linalg.mmt4d ins(%arg0, %arg1 : tensor<32x128x8x1xf32>, tensor<?x128x?x1xf32>) outs(%init : tensor<32x?x8x?xf32>) -> tensor<32x?x8x?xf32>
   return %mmt4d : tensor<32x?x8x?xf32>
 }
-// CHECK-DAG:  #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [4, 1, 0, 0, 0, 0], vector_common_parallel = [1, 1, 0, 8, [8], 0], vector_reduction = [0, 0, 1, 0, 0, 1]>
```
hanhanW, line 118: The proper fix for SVE is that you should pass the actual M0, N0 to `adjustVectorSizesForScalableVectorization`. I.e., line 2136 in the file on your branch.

```
  bool scalableTilesFound = false;
  // If scalable vectorization is enabled, adjust the vector tile sizes and the
  // corresponding scalable flags.
  if (targetConfig && isScalableVectorizationEnabled()) {
    scalableTilesFound = adjustVectorSizesForScalableVectorization(
        op, targetConfig, M0, N0, vecTileSizes, vecScalableTileFlags);
  }
```

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
hanhanW: > Sorting results in simpler, shorter code and more easy to read invariants: now the array is guaranteed to be sorted, not depending on the code being correct.
>
> The performance impact of sorting is negligible here.

I agree that the performance impact is negligible. They are both simple to me, honestly. :p Both options are okay, I'll let you make the decision.

> The only purpose of sorting here is to ensure that results don't accidentally change when the implementation of this function changes.

This is confusing to me because the sorting happens in this function. I like the idea of sorting because it aligns with my intuition when I see the function. I think we can document it a bit in the function comment.

hanhanW (approved):
LGTM, just a critical comment for SVE fix.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
hanhanW: Then the function can pick the scalable vector size correctly (ShapedType::isDynamic(n0)):

https://github.com/iree-org/iree/blob/c815c2fcdef52d8d1c026a04ba413761e78068f4/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp#L1964-L1974

hanhanW:
cc @banach-space @egebeysel for the mmt4d config changes, just in case.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
bjacob: OK, added a comment.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
bjacob: Thank you! Done. Looking good now? Notice how I had to swap the order of the logic in this function: it used to compute distribution tile sizes first, then vectorization tile sizes second. Now it is the opposite, as the distribution tile sizes depend on the vectorization tile sizes.

hanhanW (approved):
LGTM, thanks!

amd-eochoalo:
@bjacob there looks to be a regression on  [Test Sharktank / sharktank_tests :: cpu_task](https://github.com/iree-org/iree/actions/runs/21225275655/job/61071227428#logs)

```
 check 584.0361014124937 <= 181.50000000000003
 ```

I know we focus on GPUs but unclear if just increasing the golden time is appropriate in this case, @MaheshRavishankar ?

hanhanW:
> @bjacob there looks to be a regression on [Test Sharktank / sharktank_tests :: cpu_task](https://github.com/iree-org/iree/actions/runs/21225275655/job/61071227428#logs)
> 
> ```
>  check 584.0361014124937 <= 181.50000000000003
> ```
> 
> I know we focus on GPUs but unclear if just increasing the golden time is appropriate in this case, @MaheshRavishankar ?

It is a big regression (`584.0361014124937 ms (golden time 165.0 ms)`), so we'll need a basic triaging. @bjacob can you take a lookt?

amd-eochoalo:
@hanhanW I'll create a ticket and assign it to @bjacob

PR #23203, opened 2026-01-20
Add initial clang-tidy configuration

kuhar:
Add `.clang-tidy` files for compiler directories. These are based on the MLIR configuration with iree-specific tweaks. I ran these on the past 100 commits to confirm these settings produce reasonable warnings.

Sample command:
```shell
git diff HEAD~100 -U0 --relative -- . \
  | ../third_party/llvm-project/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py \
     -clang-tidy-binary=~/iree/main/build/llvm-project/bin/clang-tidy \
     -config-file=.clang-tidy  -p1 -j64 -quiet
```

Also add support for building clang tidy from source (`-DIREE_BUILD_CLANG_TOOLS_EXTRA=ON`). We may benefit from this instead of random system installations because the `llvm-*` constains mlir-specific checks.

The end goal is to have these run in CI (pre-commit only) and produce warnings (without blocking merging), so that I don't have to leave so many code review nits.


benvanik (approved):

kuhar:
I ran clang tidy on the whole LLVMGPU codegen directory and all warnings seem reasonable with the current version of `.clang-tidy`

PR #23212, opened 2026-01-20
Adding configurable compiler plugins for Bazel builds.

benvanik:
This implements configurable static compiler plugin selection for Bazel builds. Previously, Bazel hardcoded the contents of `StaticLinkedPlugins.inc`, so every Bazel-built compiler tool linked the same fixed plugin set with no supported way to exclude unavailable toolchains, speed up local iteration, or add downstream compiler plugins.

The core mechanism mirrors the shape of the existing Bazel HAL driver selection: `//compiler/plugins:enabled_plugins` is a `string_list_flag`, exposed through the `--iree_compiler_plugins` alias, and `compiler/plugins/plugins.bzl` generates the per-plugin config settings, `.inc` fragments, and aggregate dependency targets consumed by `compiler/src/iree/compiler/PluginAPI/Config`.

`iree-bazel-configure` now resolves `IREE_COMPILER_PLUGINS` from the local environment. It supports:

* `IREE_COMPILER_PLUGINS=all` for every buildable plugin.
* `IREE_COMPILER_PLUGINS=all,-plugin_id` for all buildable plugins except selected exclusions.
* `IREE_COMPILER_PLUGINS=plugin_a,plugin_b` for explicit selection.

The wrapper defaults to `all` so local developer builds get a batteries-included compiler when the relevant SDKs and submodules are available. Explicit plugin selections fail loudly when a requested plugin is unknown or cannot be built.

Toolchain and SDK detection now covers:

* CUDA, including `IREE_CUDA_TOOLKIT_ROOT`, `IREE_CUDA_DEPS_DIR`, common CUDA environment variables, and `nvcc` discovery, with the selected repository environment persisted for `@iree_cuda`.
* ROCm/HIP, including `ROCM_PATH`, `ROCM_ROOT`, `ROCM_HOME`, `HIP_PATH`, and `hipcc`.
* StableHLO and torch-mlir submodule availability.
* Vulkan SDK discovery for LLVM's `@vulkan_sdk//:sdk` repository rule, using `VULKAN_SDK`, `vulkaninfo`, and common SDK install locations.

The Torch input plugin is now wired into Bazel. Bazel-built tools can link `input_torch` and process Torch dialect input through the existing Torch input conversion tests. The implementation uses torch-mlir's upstream Bazel `torch_mlir_configure` rule instead of carrying a copied overlay implementation. The Torch InputConversion CMake files are also generated from their BUILD files; the root Torch CMake file remains hand-maintained because it owns CMake-only torch-mlir vendor setup and plugin registration.

The plugin registry is also root-loadable for out-of-tree users. `//compiler/plugins` loads `@//build_tools/bazel:default_compiler_plugins.bzl`, so a downstream Bazel root can provide that file to register additional compiler plugins, replace registration targets, or choose different default plugin IDs without editing IREE. In the same spirit, IREE now creates bundled LLVM, StableHLO, and torch-mlir source repositories only when IREE is the root module; downstream roots can provide their own MLIR-adjacent repositories.

This PR intentionally does not add GPU architecture autodetection. Vulkan SDK discovery is just enough to make LLVM's Vulkan SDK repository rule work when a local SDK is available; choosing GPU-specific target defaults should be a separate policy discussion.

Validation performed locally:

* `iree-bazel-configure`
* `iree-bazel-build //compiler/src/iree/compiler/PluginAPI/Config:StaticLinkedPlugins`
* `iree-bazel-build //tools:iree-compile`
* `iree-bazel-build //compiler/src/iree/compiler/PluginAPI/Config:StaticLinkedPlugins --iree_compiler_plugins=hal_target_llvm_cpu,input_tosa`
* select-aware `bazel cquery` checks confirming disabled Torch/StableHLO plugins are absent from the configured graph
* `iree-bazel-test //compiler/plugins/... --test_output=errors`
* CMake build/test coverage for the generated Torch InputConversion CMake files
* `pre-commit run --all`
* `git diff --check`


build_tools/bazel/workspace.bzl
```diff
@@ -67,6 +68,78 @@ cuda_auto_configure = repository_rule(
...
+            "BUILD.bazel",
+            content = """# Stub: torch-mlir disabled (IREE_INPUT_TORCH != ON)
+package(default_visibility = ["//visibility:public"])
+
+# Provide empty targets that dependent code can reference.
+# These will fail at build time if actually used.
+""",
+        )
+
+torch_mlir_auto_configure = repository_rule(
```
maxbartel, line 127: Since this is largely a copy of the torch-mlir rule, I think we can make the setup a bit easier. What about this approach:
In `build_tools/bazel/extensions.bzl` we do a 
```
new_local_repository(
        name = "torch-mlir-raw",
        build_file_content = "# empty - BUILD files overlaid by torch_mlir_configure",
        path = "third-party/torch-mlir",
)
```
and in MODULE.bazel the use_repo rule would then export `torch-mlir-raw` instead of `torch-mlir`

To setup the overlay we then to also in MODULE.bazel
```
torch_mlir_configure = use_repo_rule(
    "@torch-mlir-raw//utils/bazel:configure.bzl",
    "torch_mlir_configure",
)

torch_mlir_configure(name = "torch-mlir")
```

The actual conditional would then be moved to the plugin registration itself, which at least for me would also be a bit more intuitive. Bzlmod should be smart enough to not execute the setup if there is no dependency in the tree. WDYT?

build_tools/bazel/extensions.bzl
```diff
@@ -94,6 +94,12 @@ def _iree_extension_impl(module_ctx):
         iree_repo_alias = "@iree_core",
     )
 
+    # torch-mlir auto-configuration (conditional on IREE_INPUT_TORCH)
+    torch_mlir_auto_configure(
```
maxbartel: Can we move this into the if statement for llvm please? That way we can inject our own torch-mlir more easily. IMO this should be possible for LLVM, Torch-MLIR and StableHLO. Basically the projects with a dependency on MLIR

compiler/plugins/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
...
+        "hal_target_vmvx",
+        "hal_target_vulkan_spirv",
+        "example",
+        "simple_io_sample",
+    ],
+    plugins = {
+        # Input plugins
+        "input_stablehlo": "//compiler/plugins/input/StableHLO:registration",
+        "input_tosa": "//compiler/plugins/input/TOSA:registration",
+        # TODO(#xxxx): Enable when torch-mlir Bazel changes land upstream.
```
maxbartel: This landed right?

compiler/plugins/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
...
+        "hal_target_llvm_cpu",
+        "hal_target_local",
+        "hal_target_metal_spirv",
+        "hal_target_rocm",
+        "hal_target_vmvx",
+        "hal_target_vulkan_spirv",
+        "example",
+        "simple_io_sample",
+    ],
+    plugins = {
```
maxbartel: This here would be the place were we can build an extension point if iree is not the root of the project.

We could return this dictionary from a macro in a different file, let's say `build_tools/bazel/default_compiler_plugins.bzl`

If we import it as `@//build_tools/bazel:default_compiler_plugins.bzl`, the import happens from the workspace root. Another project could then have their file exporting the dictionary at the same relative path to its root.

maxbartel (commented):
The plugin infra looks solid. I have a suggestion on how to import torch-mlir though. Pretty cool!

krzysz00 (commented):
As a comment, can we get similar autodetection of whether there's a Vulkan SDK installed?

... and as a followup, if there'll be such autodetection, it might make sense to detect the gfx number of the running machine so that we can build e2e tests and the like? But that's probably a subsequent PR

build_tools/bazel/workspace.bzl
```diff
@@ -67,6 +68,78 @@ cuda_auto_configure = repository_rule(
     },
 )
 
+def torch_mlir_auto_configure_impl(repository_ctx):
+    """Conditionally configures torch-mlir based on IREE_INPUT_TORCH env var."""
+    env = repository_ctx.os.environ
+    iree_repo_alias = repository_ctx.attr.iree_repo_alias
+    enabled = env.get(IREE_INPUT_TORCH_ENV_KEY, "OFF").upper() in ["ON", "TRUE", "1", "YES"]
```
ScottTodd, line 75: Does this (torch_mlir in Bazel) do anything in the IREE repository right now? We don't have any BUILD.bazel files under https://github.com/iree-org/iree/tree/main/compiler/plugins/input/Torch like we do for StableHLO and TOSA.

compiler/src/iree/compiler/PluginAPI/Config/BUILD.bazel
```diff
@@ -12,28 +12,17 @@ package(
     licenses = ["notice"],  # Apache 2.0
 )
 
+# Generates StaticLinkedPlugins.inc by concatenating .inc files from all enabled
+# plugins. Each .inc file contains a single HANDLE_PLUGIN_ID(plugin_id) macro.
+#
+# Plugin selection is controlled by --//compiler/plugins:enabled_plugins flag.
+# Use the --iree_compiler_plugins flag alias for convenience:
+#   bazel build --iree_compiler_plugins=input_stablehlo,hal_target_llvm_cpu //tools:iree-compile
```
ScottTodd, line 20: Can you document this flag at https://iree.dev/developers/building/bazel/?

Could update existing sections as needed: 

Image

benvanik:
feedback addressed and vulkan sdk check added

PR #23211, opened 2026-01-20
[LinalgExt] Added toggle for using useExp2 for onlineAttention Decomposition

keshavvinayak01:
Following the discussion from https://github.com/iree-org/iree/pull/22441

Depending on the backend, certain computations may benefit from directly using `exp` instead of `exp2`, since there might be accuracy losses due to FP-reassociation.  It's helpful to add flag incase the user tracks losses to this particular computation and might favour directly using `exp`.

The `use_exp2` flag is mostly unused in dialect conversions and passes, I presume it's used as a KernelOption. The changes here will not modify the default behavior.

PR #23219, opened 2026-01-20
[LinalgExt] MSVC Bug fix - useExp / useExp2 in AggregatedOpInterfaceImpl

keshavvinayak01:
The original code returns two types, and MSVC refuses to pick a common type. By switching the conditional to return a Value, the types are identical, so MSVC should accept it.

Solves the CI bug raised by https://github.com/iree-org/iree/pull/23211


amd-eochoalo:
@keshavvinayak01 can you make sure to build windows_msvc before merging? You can use ci-extra: windows_x64_msvc

keshavvinayak01:
https://github.com/iree-org/iree/actions/runs/21183682586/job/60932314903

amd-eochoalo (approved):

Issue #23194, opened 2026-01-19
BubbleUpExpandShapesPass creates asymmetric tensor ranks in linalg.generic, causing performance degradation


yzhang93:
The `BubbleUpExpandShapesPass` in dispatch creation can introduce asymmetric tensor ranks when bubbling `tensor.expand_shape` operations that add trailing unit dimensions. When these operations bubble up through or into multi-operand `linalg.generic` operations, they can cause one operand to have a different rank than others, leading to performance degradation.

For example, the following IR:

```
util.func public @main$async(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer_view, %arg3: !hal.fence, %arg4: !hal.fence) attributes {inlining_policy = #util.inline.never, iree.abi.model = "coarse-fences", iree.abi.stub} {
  %cst = arith.constant 0.000000e+00 : f32
  %0 = hal.tensor.import wait(%arg3) => %arg1 : !hal.buffer_view -> tensor<16x48x32x480xbf16>
  %1 = iree_tensor_ext.compute_barrier.start %0 : tensor<16x48x32x480xbf16> -> tensor<16x48x32x480xbf16>
  %2 = hal.tensor.import wait(%arg3) => %arg2 : !hal.buffer_view -> tensor<16x48x32x480xbf16>
  %3 = iree_tensor_ext.compute_barrier.start %2 : tensor<16x48x32x480xbf16> -> tensor<16x48x32x480xbf16>
  %expanded = tensor.expand_shape %1 [[0], [1], [2], [3, 4]] output_shape [16, 48, 32, 5, 96] : tensor<16x48x32x480xbf16> into tensor<16x48x32x5x96xbf16>
  %expanded_0 = tensor.expand_shape %3 [[0], [1], [2], [3, 4]] output_shape [16, 48, 32, 5, 96] : tensor<16x48x32x480xbf16> into tensor<16x48x32x5x96xbf16>
  %4 = tensor.empty() : tensor<5x96x96xf32>
  %5 = linalg.fill ins(%cst : f32) outs(%4 : tensor<5x96x96xf32>) -> tensor<5x96x96xf32>
[...]

PR #23220, opened 2026-01-20
[DispatchCreation] Block asymmetric unit-dim expansions in BubbleUpExpandShapes

yzhang93:
Fixed a performance regression in `BubbleUpExpandShapesPass` where bubbling unit-dimension `tensor.expand_shape` operations through multi-operand `linalg.generic` operations created asymmetric tensor ranks, leading to suboptimal code generation. Details of the issue: https://github.com/iree-org/iree/issues/23194.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/BubbleUpExpandShapes.cpp
@@ -423,6 +423,53 @@ static bool canCauseReshapingLoopByExpansion(Operation *producer,
          !llvm::hasSingleElement(producer->getUses());
 }
 
+/// Check if bubbling up a unit-dim expand_shape would create asymmetric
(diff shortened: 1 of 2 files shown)


IanWood1 (changes requested):
I don't think we want to limit BubbleUpExpandShapes here. This might cause us to miss fusion in the case of a bit-truncate consumer. Instead, I think we need to improve CollapseDimensions' handling of `scf.forall` ops. This is the exact problem that pass is designed to solve, but it lacks good handling of scf ops.

yzhang93:
> I don't think we want to limit BubbleUpExpandShapes here. This might cause us to miss fusion in the case of a bit-truncate consumer. Instead, I think we need to improve CollapseDimensions' handling of `scf.forall` ops. This is the exact problem that pass is designed to solve, but it lacks good handling of scf ops.

Yeah, I can see the issue it caused on other tests. Another thinking, can we add/move `CollapseDimensions` before applying split reduction?

Issue #22942, opened 2025-12-18
torch operator onnx.SoftMax is failing with error : 'func.func' op failed to distribute


prosenjitdhole:
### What happened?

Please consider the following MLIR code : 
```
module {
  func.func @torch_jit(%arg0: !torch.vtensor<[1,16,576,576],f32>) -> !torch.vtensor<[1,576,576,16],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 21 : si64, torch.onnx_meta.producer_name = "pytorch", torch.onnx_meta.producer_version = "1.12.1"} {
    %816 = torch.operator "onnx.Softmax"(%arg0) {torch.onnx.axis = -1 : si64} : (!torch.vtensor<[1,16,576,576],f32>) -> !torch.vtensor<[1,16,576,576],f32>
    %817 = torch.operator "onnx.Transpose"(%816) {torch.onnx.perm = [0 : si64, 2 : si64, 3 : si64, 1 : si64]} : (!torch.vtensor<[1,16,576,576],f32>) -> !torch.vtensor<[1,576,576,16],f32>
    return  %817: !torch.vtensor<[1,576,576,16], f32>
  }
}
```

This MLIR is failing with the following error : 
```
test.mlir:3:12: error: 'func.func' op failed to distribute
    %816 = torch.operator "onnx.Softmax"(%arg0) {torch.onnx.axis = -1 : si64} : (!torch.vtensor<[1,16,576,576],f32>) -> !torch.vtensor<[1,16,576,576],f32>
```

### Steps to reproduce your issue

Please use the above mlir as test.mlir : 
```
iree-compile test.mlir --iree-hal-target-backends=rocm --iree-hip-target=gfx942 -o test.vmfb
```

### Version information

Fails in latest version : 
```
IREE (https://iree.dev):
  IREE compiler version 3.10.0rc20251218 @ f32209fdff313e00940f3222a65427cee4c62d25
  LLVM version 22.0.0git
  Optimized build
```

### Additional context

This testcase used to **pass** with the following version : 
```
IREE (https://iree.dev):
  IREE compiler version 3.8.0rc20251010 @ afe01bda52cf977a8e371caf147e38a38542622d
  LLVM version 22.0.0git
  Optimized build
```

And started failing from : 
```
IREE (https://iree.dev):
  IREE compiler version 3.8.0rc20251011 @ cdce159aa49d934c036051bd63dced24bba8eee6
[...]

PR #23221, opened 2026-01-20
[Dispatch Creation] Move iter space tracking to LinalgExt

IanWood1:
Moves `computeIterationSpaceMapping` and `getOuterParallelLoops` into LinalgExt. Implements the fusion interface on SetEncoding and UnSetEncoding ops to prevent cyclic dependency from their use in `getOuterParallelLoops`. https://github.com/iree-org/iree/issues/22942


MaheshRavishankar (approved):
I think this is just a move of code. Looks good to me!

PR #23224, opened 2026-01-20
[plugins][Torch] Add a flag to externalize transients during torch-to-IREE conversion

zjgarvey:
This is currently set up to add a `!hal.buffer` program argument to both the async func and the sync wrapper during `FuncConversionPass`, and treat it as a transient memory user allocation for all outputs (inplace or exported).

The intended use:

`iree-compile <usual-args> --iree-externalize-transients`

When invoking the main compiled program, the user will be expected to query the required transient buffer size and provide an additional `!hal.buffer` input.

-------

Edit: some notes about why we are using a flag.

We are choosing to append a buffer input through this pass using flag, rather than starting with a torch program with an extra mutable dynamic tensor arg and indicating the purpose of the extra tensor with an op like `torch.overwrite.tensor.contents`.

The reason a torch op doesn't make sense is that there isn't a clear import path for this. There isn't a torch op that says "use this tensor as scratch memory", and we'd need to put together something rather contrived in order to trace a function with the correct signature to match the desired IREE-input IR, when we have no intention of doing anything within torch-mlir proper to handle this special op.

### DIFF
--- a/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
@@ -651,9 +651,6 @@ public:
     Location loc = torchFunc.getLoc();
     // Determine whether to build pure async or async + sync wrapper.
     bool generateSyncWrapper = true;
-    // Determine whether to include a new transient buffer input to the
-    // generated async function and corresponding sync wrapper.
-    bool externalizeTransientMemory = externalizeTransients;
     StringRef originalName = torchFunc.getName();
     std::string asyncFunctionName = originalName.str();
     if (generateSyncWrapper) {
@@ -672,14 +669,6 @@ public:
                                              torchFuncType.getInputs().end());
     convertedFuncInfo.torchResultTypes.append(
         torchFuncType.getResults().begin(), torchFuncType.getResults().end());
-    // When externalizing transient memory, put the input buffer before fences.
-    if (externalizeTransientMemory) {
-      convertedFuncInfo.torchInputTypes.push_back(bufferType);
-    }
-    // For the coarse-fences ABI, we add two fences to the end. Treat these as
-    // original types so that the lists line up.
-    convertedFuncInfo.torchInputTypes.push_back(fenceType);
-    convertedFuncInfo.torchInputTypes.push_back(fenceType);
     SmallVector<Type> ireeInputTypes(convertedFuncInfo.torchInputTypes);
     SmallVector<Type> ireeResultTypes(convertedFuncInfo.torchResultTypes);
     convertedFuncInfo.inputDispositions.resize(ireeInputTypes.size());
@@ -700,6 +689,19 @@ public:
       }
     }
 
+    // When externalizing transient memory, put the input buffer before fences.
+    if (externalizeTransients) {
+      convertedFuncInfo.torchInputTypes.push_back(bufferType);
+      ireeInputTypes.push_back(bufferType);
+      convertedFuncInfo.inputDispositions.push_back(
+          TypeDisposition::TRANSIENT_BUFFER);
+    }
+    // For the coarse-fences ABI, we add two fences to the end. Treat these as
+    // original types so that the lists line up.
+    convertedFuncInfo.torchInputTypes.append({fenceType, fenceType});
+    ireeInputTypes.append({fenceType, fenceType});
+    convertedFuncInfo.inputDispositions.append(
+        {TypeDisposition::FENCE, TypeDisposition::FENCE});
     // Build tied operands index mapping results back to operands.
     SmallVector<int64_t> tiedOperands;
     bool anyTiedOperands = false;
--- a/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
@@ -652,7 +652,7 @@ public:
     // Determine whether to build pure async or async + sync wrapper.
     bool generateSyncWrapper = true;
     // Determine whether to include a new transient buffer input to the
-    // generated async wrapper.
+    // generated async function and corresponding sync wrapper.
     bool externalizeTransientMemory = externalizeTransients;
     StringRef originalName = torchFunc.getName();
     std::string asyncFunctionName = originalName.str();
--- a/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
@@ -577,7 +577,7 @@ void createCoarseFencesSyncWrapper(StringRef syncFunctionName,
 
 } // namespace
 
-class FuncConversionPass
+class FuncConversionPass final
     : public impl::FuncConversionPassBase<FuncConversionPass> {
 public:
   using Base::Base;
--- a/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
@@ -125,6 +125,7 @@ enum class TypeDisposition {
   MUTABLE_TENSOR,
   TORCH_PRIMITIVE,
   PASSTHROUGH,
+  TRANSIENT_BUFFER,
   FENCE,
 };
 
@@ -149,6 +150,8 @@ struct ConvertedAsyncFunctionInfo {
   SmallVector<Value> barrierInputs;
   // Meta data per barrier input: storage, torchType, returnIndex (or -1)
   SmallVector<BarrierResult> barrierResultMeta;
+  // Transient buffer
+  BlockArgument transientBuffer;
 
   LogicalResult postProcess();
   LogicalResult convertImmutableTensorArg(BlockArgument argValue,
@@ -164,6 +167,11 @@ struct ConvertedAsyncFunctionInfo {
         torchType,
         returnIndex,
     });
+  }
+
+  void setTransientBuffer(BlockArgument bufferArg) {
+    assert(!transientBuffer && "Cannot reassign existing transient buffer.");
+    transientBuffer = bufferArg;
   }
 
   Attribute getTorchArgAttr(BlockArgument argValue, StringRef attrName) {
@@ -240,6 +248,9 @@ LogicalResult ConvertedAsyncFunctionInfo::postProcess() {
     case TypeDisposition::PASSTHROUGH:
       // Do nothing.
       break;
+    case TypeDisposition::TRANSIENT_BUFFER:
+      setTransientBuffer(argValue);
+      break;
     case TypeDisposition::FENCE:
       // Do nothing.
       break;
@@ -307,6 +318,7 @@ LogicalResult ConvertedAsyncFunctionInfo::postProcess() {
     SmallVector<Value> aliasedResults;
     for (auto [barrierInput, meta] :
          llvm::zip_equal(barrierInputs, barrierResultMeta)) {
+      Value aliasResult;
       if (meta.storage) {
         // Use the wait fence indicating when the storage is available for
         // mutation. We need to ensure that no writes are made to the storage
@@ -318,13 +330,21 @@ LogicalResult ConvertedAsyncFunctionInfo::postProcess() {
         Value waitFence = waitSignalFences->first;
         auto barrierInputDims = IREE::Util::buildDynamicDimsForValue(
             barrierInput.getLoc(), barrierInput, postambleBuilder);
-        aliasedResults.push_back(IREE::HAL::TensorAliasOp::create(
+        aliasResult = IREE::HAL::TensorAliasOp::create(
             postambleBuilder, barrierInput.getLoc(), barrierInput.getType(),
             barrierInput, barrierInputDims, meta.storage, waitFence,
-            storageAffinityAttr));
+            storageAffinityAttr);
       } else {
-        aliasedResults.push_back(barrierInput);
+        aliasResult = barrierInput;
       }
+      if (transientBuffer) {
+        auto sourceDims = IREE::Util::buildDynamicDimsForValue(
+            aliasResult.getLoc(), aliasResult, postambleBuilder);
+        aliasResult = IREE::HAL::TensorTransientsOp::create(
+            postambleBuilder, aliasResult.getLoc(), aliasResult.getType(),
+            aliasResult, sourceDims, transientBuffer, Attribute{});
+      }
+      aliasedResults.push_back(aliasResult);
     }
     auto barrierOp = IREE::HAL::TensorBarrierOp::create(
         postambleBuilder, funcOp.getLoc(), aliasedResults, coarseSignalFence);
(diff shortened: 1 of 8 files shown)


compiler/plugins/input/Torch/InputConversion/test/func_conversion_transients.mlir
```diff
@@ -0,0 +1,136 @@
...
+//   CHECK-DAG: %[[TORCH_ARG0:.+]] = torch_c.from_builtin_tensor %[[WAIT_ARG0]]
+//   CHECK-DAG: %[[WAIT_ARG1:.+]] = hal.tensor.import on(#hal.device.promise<@dev_b>) wait(%arg3) => %arg1
+//   CHECK-DAG: %[[TORCH_ARG1:.+]] = torch_c.from_builtin_tensor %[[WAIT_ARG1]]
+//   CHECK-DAG: %[[TORCH_RESULT0:.+]] = torch.operator "other_calc"(%[[TORCH_ARG0]])
+//   CHECK-DAG: %[[TORCH_RESULT1:.+]] = torch.operator "mutate_inplace"(%[[TORCH_ARG1]])
+//   CHECK-DAG: %[[TENSOR_ARG0:.+]] = torch_c.to_builtin_tensor %[[TORCH_RESULT0]]
+//   CHECK-DAG: %[[TENSOR_ARG1:.+]] = torch_c.to_builtin_tensor %[[TORCH_RESULT1]]
+//       CHECK: %[[EXPORT_ALIAS1:.+]] = hal.tensor.alias on(#hal.device.promise<@dev_b>) wait(%arg3) => %[[TENSOR_ARG1]] : tensor<5x4xf32> to %arg1 : !hal.buffer_view
+//   CHECK-DAG: %[[TRANSIENTS_CALL1:.+]] = hal.tensor.transients %[[EXPORT_ALIAS1]] : tensor<5x4xf32> from %arg2 : !hal.buffer
+//   CHECK-DAG: %[[TRANSIENTS_CALL0:.+]] = hal.tensor.transients %[[TENSOR_ARG0]] : tensor<4x5xi32> from %arg2 : !hal.buffer
```
zjgarvey: Does something about the device promise need to be propagated here, or is this only for the final output?

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -502,6 +522,8 @@ void createCoarseFencesSyncWrapper(StringRef syncFunctionName,
   // The coarse fences wrapper has the same signature as the async variant
   // but with the last two inputs (wait, signal fence) sliced off.
   FunctionType asyncFuncType = asyncFuncOp.getFunctionType();
+  // Note: If we externalize a transient buffer, we are currently including it
+  // as the final input for the sync wrapper.
```
zjgarvey, line 526: If we need to preserve the number of arguments for the main program- i.e., the sync wrapper- I'm not sure exactly how we should be setting up the buffer generation inside the sync wrapper, given that the size query functions are not yet generated.

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -555,9 +577,15 @@ void createCoarseFencesSyncWrapper(StringRef syncFunctionName,
 
 } // namespace
 
-class FuncConversionPass final
+class FuncConversionPass
```
zjgarvey: This looks accidental.

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -623,6 +651,9 @@ class FuncConversionPass final
     Location loc = torchFunc.getLoc();
     // Determine whether to build pure async or async + sync wrapper.
     bool generateSyncWrapper = true;
+    // Determine whether to include a new transient buffer input to the
+    // generated async wrapper.
```
zjgarvey:
```suggestion
    // generated async func and corresponding sync wrapper.
```

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -766,6 +802,13 @@ class FuncConversionPass final
       return success();
     }
 
+    if (isa<IREE::HAL::BufferType>(torchType)) {
+      // Are there other situations where we might have buffer inputs?
+      ireeType = torchType;
+      disp = TypeDisposition::TRANSIENT_BUFFER;
+      return success();
+    }
```
zjgarvey, line 812: Seems kinda flaky. Maybe we generate the annotations and types in one go instead.

compiler/plugins/input/Torch/InputConversion/Passes.h
```diff
@@ -36,6 +36,13 @@ struct TorchToIREELoweringPipelineOptions
   Option<bool> decompose{*this, "decompose",
                          llvm::cl::desc("Decompose complex torch operations."),
                          llvm::cl::init(true)};
+  Option<bool> externalizeTransients{
+      *this, "externalize-transients",
+      llvm::cl::desc(
```
zjgarvey, line 41: Maybe indicate that this option only has an effect on torch/torch-onnx programs.

compiler/plugins/input/Torch/InputConversion/Passes.h
```diff
@@ -36,6 +36,13 @@ struct TorchToIREELoweringPipelineOptions
   Option<bool> decompose{*this, "decompose",
                          llvm::cl::desc("Decompose complex torch operations."),
                          llvm::cl::init(true)};
+  Option<bool> externalizeTransients{
```
zjgarvey, line 39: Maybe add a lit test for `iree-compile --iree-externalize-transients --compile-to=input`?

zjgarvey: This should be covered by the provided lit tests, but it might be good to test these options align properly.

zjgarvey (commented):
Self-review comments.

Some of these are questions, and I'd appreciate some feedback from reviewers.

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
MaheshRavishankar: Good question. I think if we are using the flag, then the sync function also needs to get an extra argument for the transients. I dont see any other way. That argument can be a !hal.buffer, and this method needs to add the attributes related to transients (the size and function) to the sync wrapper as well?

compiler/plugins/input/Torch/PluginRegistration.cpp
```diff
@@ -38,6 +39,13 @@ struct TorchOptions {
     binder.opt<bool>("iree-torch-decompose-complex-ops", decompose,
                      llvm::cl::cat(category),
                      llvm::cl::desc("Decompose complex torch operations."));
+    binder.opt<bool>(
+        "iree-externalize-transients", externalizeTransients,
```
MaheshRavishankar: `iree-torch-externalize-transients`?

zjgarvey: Yeah, that's not a bad idea to enforce it only does something for torch conversion.

zjgarvey: With regards to your other comment above:

I don't think the sync wrapper will have the "transients" attributes like the `transients.size.constant` (I don't think it will "know" that the buffer arg is a transient memory buffer. The async func will, because we are calling `hal.tensor.transients` in that func). I could be wrong. I'll check in a bit.

If the sync wrapper doesn't have the transients size attribute, it might make sense to try and deal with this a bit more carefully, but I'm not sure. I'll ping Ben and see if he has any advice or opinions.

benvanik: yeah, this needs to be torch-specific

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
benvanik: yes - sync and async must behave consistently (coarse-fences async is just sync + 2 fences, so if you add a buffer to async you need to be adding it to sync)

compiler/plugins/input/Torch/InputConversion/test/func_conversion_transients.mlir
```diff
@@ -0,0 +1,136 @@
...
+//   CHECK-DAG: %[[WAIT_ARG0:.+]] = hal.tensor.import on(#hal.device.promise<@dev_a>) wait(%arg3) => %arg0
+//   CHECK-DAG: %[[TORCH_ARG0:.+]] = torch_c.from_builtin_tensor %[[WAIT_ARG0]]
+//   CHECK-DAG: %[[WAIT_ARG1:.+]] = hal.tensor.import on(#hal.device.promise<@dev_b>) wait(%arg3) => %arg1
+//   CHECK-DAG: %[[TORCH_ARG1:.+]] = torch_c.from_builtin_tensor %[[WAIT_ARG1]]
+//   CHECK-DAG: %[[TORCH_RESULT0:.+]] = torch.operator "other_calc"(%[[TORCH_ARG0]])
+//   CHECK-DAG: %[[TORCH_RESULT1:.+]] = torch.operator "mutate_inplace"(%[[TORCH_ARG1]])
+//   CHECK-DAG: %[[TENSOR_ARG0:.+]] = torch_c.to_builtin_tensor %[[TORCH_RESULT0]]
+//   CHECK-DAG: %[[TENSOR_ARG1:.+]] = torch_c.to_builtin_tensor %[[TORCH_RESULT1]]
+//       CHECK: %[[EXPORT_ALIAS1:.+]] = hal.tensor.alias on(#hal.device.promise<@dev_b>) wait(%arg3) => %[[TENSOR_ARG1]] : tensor<5x4xf32> to %arg1 : !hal.buffer_view
+//   CHECK-DAG: %[[TRANSIENTS_CALL1:.+]] = hal.tensor.transients %[[EXPORT_ALIAS1]] : tensor<5x4xf32> from %arg2 : !hal.buffer
```
benvanik: why 1? and why are these two call1 and call0? (here and above)

compiler/plugins/input/Torch/InputConversion/test/func_conversion_transients.mlir
```diff
@@ -0,0 +1,136 @@
...
+builtin.module @return_immutable_arg {
+func.func @main(%arg0: !torch.vtensor<[4,5],si32>) -> !torch.vtensor<[4,5],si32>  {
+  return %arg0 : !torch.vtensor<[4,5],si32>
+}
+}
+
+// -----
+// CHECK-LABEL: @mutable_input_overwrite_no_return
+//       CHECK: util.func public @main$async(
+//  CHECK-SAME:     %arg0: !hal.buffer_view, %arg1: !hal.buffer_view,
```
benvanik: as you do above, capture values here (not raw SSA identifiers)

compiler/plugins/input/Torch/InputConversion/test/func_conversion_transients.mlir
```diff
@@ -0,0 +1,136 @@
...
+  %1 = torch.operator "mutate_inplace"(%0) : (!torch.vtensor<[5,4],f32>) -> !torch.vtensor<[5,4],f32>
+  %2 = torch.operator "other_calc"(%arg0) : (!torch.vtensor<[4,5],si32>) -> !torch.vtensor<[4,5],si32>
+  torch.overwrite.tensor.contents %1 overwrites %arg1 : !torch.vtensor<[5,4],f32>, !torch.tensor<[5,4],f32>
+  return %2 : !torch.vtensor<[4,5],si32>
+}
+}
+
+// -----
+// CHECK-LABEL: @immutable_import_export
+// CHECK: hal.buffer_view.dim<%arg0
```
benvanik: capture values, etc

benvanik (changes requested):

compiler/plugins/input/Torch/InputConversion/test/func_conversion_transients.mlir
zjgarvey: I'm not sure I understand your question. There are two parallel computations for this test (one inplace), so they are disambiguated with 0/1.

 Is "call" is misleading? I can rename to "TRANS0"/"TRANS1" if that is less misleading.

My understanding is that this op "marks" or "flags" a value to say "all intermediate memory required to get here should come from the provided buffer" I don't know a good word for that.

--------

I did do a bit of copy/paste to get some initial tests. The original test:

https://github.com/iree-org/iree/blob/8e0aa2bbae1786d7f2e19347eb146e3de336913b/compiler/plugins/input/Torch/InputConversion/test/func_conversion.mlir#L64C15-L64C16

I didn't think too hard about the existing conventions there.

zjgarvey: I decided to just take a minute to clean up all the lit tests included here and use better naming.

I also decided to include a bit more on the dynamic dim example, since it seemed important to include what the IR looks like for that.

Let me know if this addresses your initial comment.

zjgarvey:
@benvanik @MaheshRavishankar brought up a concern about the sync wrapper use from a practical standpoint:

If someone wants to use this flag and call the sync wrapper, the sync wrapper won't get any attributes for the `transients_size`. They would need to look up the attributes from the async function to figure out how much memory to pass in for the buffer.

I would like to add some support for propagating size query attributes during the `MaterializeTransientsSizeQueriesPass`, but I personally think we could do this in a separate patch. What do you think?

MaheshRavishankar:
> @benvanik @MaheshRavishankar brought up a concern about the sync wrapper use from a practical standpoint:
> 
> If someone wants to use this flag and call the sync wrapper, the sync wrapper won't get any attributes for the `transients_size`. They would need to look up the attributes from the async function to figure out how much memory to pass in for the buffer.
> 
> I would like to add some support for propagating size query attributes during the `MaterializeTransientsSizeQueriesPass`, but I personally think we could do this in a separate patch. What do you think?

Just so that I am sure, is it more involved than just "copying" the attributes to the sync wrapper?

zjgarvey:
> Just so that I am sure, is it more involved than just "copying" the attributes to the sync wrapper?

A little bit. I looked into it on Friday, and it may be too hacky to just copy/paste the attributes. The materialize size query pass looks for pack ops, which don't reference the individual buffer arg directly. When lifting the size query up to another func op, we should be at least be sure that the buffer being used for transient memory is being directly forwarded to the func call where the pack ops are located.

I'm not sure how robust we need all of this to be, but it would probably be ideal to look at a full function call graph for the module, and essentially make a parallel `_transients_size` call graph. E.g., funcA could have an external buffer arg, and could call both funcB and funcC, each of which rely on this common external memory buffer for transients. Would we need to take the max of the size queries, or the sum, or something more subtle?

MaheshRavishankar:
> > Just so that I am sure, is it more involved than just "copying" the attributes to the sync wrapper?
> 
> A little bit. I looked into it on Friday, and it may be too hacky to just copy/paste the attributes. The materialize size query pass looks for pack ops, which don't reference the individual buffer arg directly. When lifting the size query up to another func op, we should be at least be sure that the buffer being used for transient memory is being directly forwarded to the func call where the pack ops are located.
> 
> I'm not sure how robust we need all of this to be, but it would probably be ideal to look at a full function call graph for the module, and essentially make a parallel `_transients_size` call graph. E.g., funcA could have an external buffer arg, and could call both funcB and funcC, each of which rely on this common external memory buffer for transients. Would we need to take the max of the size queries, or the sum, or something more subtle?

Ok, yeah I understand. Ok if we dont need to think about the sync wrapper right now to connect things, then lets do that in a follow up. Anyway this is under a flag now

benvanik (approved):

PR #23225, opened 2026-01-21
[CI] Use ROCM docker for AMDGPU workflows

jtuyls:


.github/workflows/pkgci_test_amd_mi250.yml
```diff
@@ -55,7 +73,28 @@ jobs:
             --fetch-gh-workflow=${{ inputs.artifact_run_id }}
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
+      - name: Display environment info
+        run: |
+          echo "=== ROCm Version (container) ==="
+          cat /opt/rocm/.info/version || echo "version file not found"
```
ScottTodd: I'd suggest putting code like this in a script under https://github.com/iree-org/iree/tree/main/build_tools/scripts. See the `check_cuda.sh` and `check_vulkan.sh` scripts for reference.

PR #23228, opened 2026-01-21
[ReductionVectorDistribute] Avoid adding lowering configs on failure

rkayaith:
The logic here was attaching lowering configs before validating that all ops were actually supported. This was causing issues where the reduction vector distribution pipeline would leave lowering configs in the IR even though it wasn't being used, resulting in compilation failures in later passes.


nirvedhmeshram (approved):
LGTM

kuhar (approved):

PR #23230, opened 2026-01-21
[AMDGPU][LDS] Enable global load DMA by default on gfx950+

lialan:
* Automatically use coalesced global load DMA for matmul/IGEMM on CDNA4+ architectures.
* Falls back to standard promotion and avoid using LDS DMA when source comes from `tensor.pad` or when padding is required.
 
ci-extra: test_amd_mi355

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -426,10 +426,17 @@ struct AMDGPULowerCoalescedDMAToGatherLDSPass final
       return;
     }
 
-    // dma_sizes is optional - if not specified, skip the size validation.
-    ArrayRef<int64_t> dmaSizes;
-    if (DenseI64ArrayAttr dmaSizesAttr = target.getWgp().getDmaSizes()) {
-      dmaSizes = dmaSizesAttr.asArrayRef();
+    // dma_sizes is mandatory - if not specified, we cannot lower.
+    DenseI64ArrayAttr dmaSizesAttr = target.getWgp().getDmaSizes();
+    if (!dmaSizesAttr || dmaSizesAttr.empty()) {
+      LDBG() << "No DMA sizes specified in target, skipping pass";
+      return;
+    }
+    ArrayRef<int64_t> dmaSizes = dmaSizesAttr.asArrayRef();
+    // Require dma_sizes to contain 128 bits or higher.
+    if (llvm::none_of(dmaSizes, [](int64_t size) { return size >= 128; })) {
+      LDBG() << "No DMA size >= 128 bits in target, skipping pass";
+      return;
     }
 
     MLIRContext *context = &getContext();
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -98,17 +98,38 @@ static bool tracesToTensorEmpty(Value value) {
   return initValue.getDefiningOp<tensor::EmptyOp>() != nullptr;
 }
 
+/// Check if the source of an operation comes directly from global memory.
+/// Returns false if the source goes through tensor.pad or other local
+/// computation that would prevent using global load DMA.
+static bool sourceIsFromGlobalMemory(Operation *op) {
+  Value source;
+  if (auto copyOp = dyn_cast<linalg::CopyOp>(op)) {
+    source = copyOp.getInputs()[0];
+  } else if (auto gatherOp = dyn_cast<IREE::LinalgExt::GatherOp>(op)) {
+    source = gatherOp.getSource();
+  } else {
+    return false;
+  }
+
+  // Trace through extract_slice operations to find the origin.
+  while (auto extractOp = source.getDefiningOp<tensor::ExtractSliceOp>()) {
+    source = extractOp.getSource();
+  }
+
+  // If the source comes from tensor.pad, it's not directly from global memory.
+  if (source.getDefiningOp<tensor::PadOp>()) {
+    return false;
+  }
+
+  // Otherwise, assume it's from global memory (e.g., dispatch tensor load).
+  return true;
+}
+
 /// Helper to compute thread number of threads based on translation_info.
 /// Uses the subgroup_size from translation_info for thread-level tiling.
 static SmallVector<OpFoldResult>
 computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
                             RankedTensorType outputType) {
-  // Check that this operation has the use_global_load_dma config.
-  auto dmaConfig = getLoweringConfig<IREE::GPU::UseGlobalLoadDMAAttr>(op);
-  if (!dmaConfig) {
-    return {};
-  }
-
   // Get the function containing this operation.
   auto funcOp = op->getParentOfType<FunctionOpInterface>();
   if (!funcOp) {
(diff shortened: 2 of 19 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -426,6 +427,20 @@ struct AMDGPULowerCoalescedDMAToGatherLDSPass final
...
+    FailureOr<amdgpu::Chipset> maybeChipset =
+        amdgpu::Chipset::parse(targetArch);
+    if (failed(maybeChipset)) {
+      LDBG() << "Failed to parse chipset from target arch: " << targetArch;
+      return;
+    }
+    constexpr amdgpu::Chipset kGfx950{9, 5, 0};
+    if (*maybeChipset < kGfx950) {
+      LDBG() << "Target arch " << targetArch << " is not CDNA4+, skipping pass";
+      return;
```
kuhar: Can we make sure this is not accidentally enabled on rdna cards? Would it be possible to have a lit test for this?

kuhar (commented):
There are some clang-tidy warnings

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
lialan: Indeed, so let's just restrict it to gfx950 only. I will add a test.

krzysz00:
Just going to stick a call for benchmarks here

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
lialan: Added some guard to limit to gfx950+ but also arch that has global load lds instructions.

kuhar: Is there something else that tells us if DMA is available? Maybe we could check for the `dma_sizes` target attribute?

lialan: now only enabled when it is both:
* gfx 950+
* have DMA size >= 128bit.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -426,10 +427,24 @@ struct AMDGPULowerCoalescedDMAToGatherLDSPass final
...
+
+    // dma_sizes is mandatory - if not specified, we cannot lower.
+    DenseI64ArrayAttr dmaSizesAttr = target.getWgp().getDmaSizes();
+    if (!dmaSizesAttr || dmaSizesAttr.empty()) {
+      LDBG() << "No DMA sizes specified in target, skipping pass";
+      return;
+    }
+    ArrayRef<int64_t> dmaSizes = dmaSizesAttr.asArrayRef();
+    // Require dma_sizes to contain 128 bits or higher.
+    if (!llvm::any_of(dmaSizes, [](int64_t size) { return size >= 128; })) {
```
kuhar:
```suggestion
    if (llvm::none_of(dmaSizes, [](int64_t size) { return size >= 128; })) {
```

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1514,7 +1509,8 @@ static LogicalResult setContractConfig(IREE::GPU::TargetAttr target,
       SmallVector<NamedAttribute, 1> pipelineAttrs;
       auto pipelineOptions = IREE::GPU::GPUPipelineOptionsAttr::get(
           context, /*prefetch_num_stages=*/0,
-          /*no_reduce_shared_memory_bank_conflicts=*/true,
+          /*no_reduce_shared_memory_bank_conflicts=*/
+          IREE::GPU::targetSupportsGlobalLoadDMA(target),
```
qedawkins: We shouldn't be changing this value for targets/configs where we aren't using DMAs. I see changes to gfx942 tests when we're only enabling the feature for 950+.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -100,6 +101,11 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
     return {};
   }
 
+  // Global load DMA is only supported on CDNA4+ (gfx950+) architectures.
```
qedawkins: We don't need to mention the targets that support it everywhere (and don't want to call out a specific architecture in what is supposed to be a generic pass). In this instance the code is probably clear enough on its own to not need the comment.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -666,6 +672,11 @@ struct GPUConvertToCoalescedDMAPass final
       return failure();
     }
 
+    // Global load DMA is only supported on CDNA4+ (gfx950+) architectures.
+    if (!IREE::GPU::targetSupportsGlobalLoadDMA(target)) {
```
qedawkins: This check looks redundant, we're already checking in `computeNumThreadsImpl`?

qedawkins (changes requested):
I'm closing my eyes to the amdgpu chipset hardcoding in the GPU dialect since the whole op needs to move anyway, but there's no way to do that right now.

I'll also echo Krzysztof's request. Providing some basic benchmarking results with a feature enablement acts as proof that the feature is working as intended. Easily reproducible results is best, but even a small hand picked sweep is enough.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
krzysz00: We could actually remove the gfx950 check - "DMA sizes >= 128 bits" might actually be a good condition to use here

It could make things a touch awkward if we want to not use this op on gfx1250 and need to do phase ordering about it, but  for now that's a sufficient condition that doesn't involve mentioning an architecture by name

lialan:
@qedawkins @krzysz00

Column Z and Column AA are baseline and the new results.

Overall it is positive, but very much diminished by a lot of regressions. So I am investigating those regressions.

qedawkins:
> @qedawkins @krzysz00 here are the [benchmark numbers](https://amdcloud-my.sharepoint.com/:x:/r/personal/nmeshram_amd_com/_layouts/15/Doc.aspx?sourcedoc=%7B6C4DC316-C731-4574-BCCD-BEE53E84F4D3%7D&file=%5Bmiopen_shapes_iree.xlsx%5D(https://amdcloud-my.sharepoint.com/:x:/r/personal/nmeshram_amd_com/_layouts/15/Doc.aspx?sourcedoc=%7B6C4DC316-C731-4574-BCCD-BEE53E84F4D3%7D&file=miopen_shapes_iree.xlsx&action=default&mobileredirect=true)&action=default&mobileredirect=true) using turbine.
> 
> Column Z and Column AA are baseline and the new results.
> 
> Overall it is positive, but very much diminished by a lot of regressions. So I am investigating those regressions.

Awesome, thanks for running the sweep. Getting a head start on the regressions sounds great, thanks!

lialan:
https://github.com/iree-org/iree/pull/23365 tries to enable DMA for unaligned cases, so we should see if we can merge that before we merge this one.

PR #23231, opened 2026-01-21
[CI][Torch] Enable split reduction and O3 for llama_8b_fp16 quality tests with gfx942 config

bangtianliu:
This PR enables `-opt-level=O3` for the llama 8b fp16 quality test. 

"--iree-dispatch-creation-enable-split-reduction" will be removed when it is set to true by default. 

ci-extra: test_torch


kuhar (approved):

MaheshRavishankar:
Do we know what the impact is on LLaMa?

bangtianliu:
> Do we know what the impact is on LLaMa?

This PR is mainly addressing the issue here: https://github.com/iree-org/iree/pull/22379#pullrequestreview-3370476772 based on my understanding, mainly for correctness checking. cc @Groverkss here.

PR #23237, opened 2026-01-22
[CI] Add clang-tidy workflow

kuhar:
Runs clang-tidy on pull request and leaves comments on modified lines only.

I made it separate from the regular lint: probably too heavyweight to run locally, and I wanted it to be non-blocking.


kuhar:
Sample comments posted by the workflow:
image

compiler/src/.clang-tidy
```diff
@@ -1,13 +0,0 @@
-Checks: >
```
kuhar, line 1: This has been replaced by a new clang-tidy file in https://github.com/iree-org/iree/pull/23203

amd-eochoalo (approved):
Looks good!

PR #23238, opened 2026-01-22
[Codegen] Add an option to enable fp8/fp4 software emulation for all GPU targets.

hanhanW:
Previously, using fp8 types on chips without hardware support would produce a compile error (e.g., f8E5M2 on gfx942 or f8E5M2FNUZ on gfx950).

This change adds a flag to enable software emulation as a fallback, allowing any small float type to be used on any chip. If the flag is not enabled, it emits the error message with the hint of adding the flag.

On ConvertUnsupportedFloatArith pass side, it marks arith.scaling_extf/truncf as legal since they have their own expansion patterns that run in later passes. The revision also fixes a bug that gfx950 does not have FNUZ hardware support. So `<=` is replaced with `<`.

The GPU specific tests are refreshed; all the executable target attributes are removed from existing tests. Three of existing GPU tests are removed because we already have the same tests.

AMDGPU support matrix:

| Type         | gfx908/90a | gfx942 | gfx950 | gfx11xx | gfx12xx |
|--------------|------------|--------|--------|---------|---------|
| f8E4M3FNUZ   |   emu      |   hw   |  emu   |   emu   |   emu   |
| f8E5M2FNUZ   |   emu      |   hw   |  emu   |   emu   |   emu   |
| f8E4M3FN     |   emu      |  emu   |   hw   |   emu   |   hw    |
| f8E5M2       |   emu      |  emu   |   hw   |   emu   |   hw    |
| f4E2M1FN     |   emu      |  emu   |   hw   |   emu   |   hw    |

(hw = hardware instructions, emu = software emulation)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
@@ -369,8 +369,7 @@ private:
 ///    - f4E2M1FN has no NaN and no negative zero.
 struct TruncFToSmallFloat final : OpRewritePattern<arith::TruncFOp> {
   TruncFToSmallFloat(MLIRContext *ctx, ArrayRef<Type> sourceTypes)
-      : OpRewritePattern(ctx),
-        sourceTypes(sourceTypes.begin(), sourceTypes.end()) {}
+      : OpRewritePattern(ctx), sourceTypes(sourceTypes) {}
 
   LogicalResult matchAndRewrite(arith::TruncFOp op,
                                 PatternRewriter &rewriter) const override {
@@ -603,8 +602,7 @@ private:
 /// which is simpler and more general than enumerating all possible values.
 struct ExtFFromSmallFloat final : OpRewritePattern<arith::ExtFOp> {
   ExtFFromSmallFloat(MLIRContext *ctx, ArrayRef<Type> sourceTypes)
-      : OpRewritePattern(ctx),
-        sourceTypes(sourceTypes.begin(), sourceTypes.end()) {}
+      : OpRewritePattern(ctx), sourceTypes(sourceTypes) {}
 
   LogicalResult matchAndRewrite(arith::ExtFOp op,
                                 PatternRewriter &rewriter) const override {
@@ -845,7 +843,7 @@ void ConvertUnsupportedFloatArithPass::runOnOperation() {
   // the small float types need the emulation. For GPU targets, some types may
   // have hardware conversion support and can be skipped.
   SmallVector<Type> typesNeedingConversionEmulation = allSmallFloatTypes;
-  if (auto gpuAttr = getGPUTargetAttr(funcOp)) {
+  if (IREE::GPU::TargetAttr gpuAttr = getGPUTargetAttr(funcOp)) {
     typesNeedingConversionEmulation =
         getTypesNeedingConversionEmulationForGPU(context, gpuAttr);
   }
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
@@ -41,7 +41,8 @@ using hasContextGet = decltype(T::get(std::declval<MLIRContext *>()));
 
 /// Helpers to append types to a vector if they are small float types (fp4/fp8).
 template <typename T>
-static void maybeAppendType(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
+static void maybeAppendSmallFloatType(MLIRContext *ctx,
+                                      SmallVectorImpl<Type> &types) {
   if constexpr (llvm::is_detected<hasContextGet, T>::value) {
     Type t = T::get(ctx);
     if (isa<FloatType>(t)) {
@@ -52,9 +53,11 @@ static void maybeAppendType(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
     }
   }
 }
+
 template <typename... Ts>
-static void appendTypesIf(MLIRContext *ctx, SmallVectorImpl<Type> &types) {
-  (maybeAppendType<Ts>(ctx, types), ...);
+static void appendSmallFloatTypes(MLIRContext *ctx,
+                                  SmallVectorImpl<Type> &types) {
+  (maybeAppendSmallFloatType<Ts>(ctx, types), ...);
 }
 
 //===----------------------------------------------------------------------===//
@@ -126,8 +129,12 @@ public:
     // Check for FNUZ types where NaN is encoded as sign bit only
     // (e.g., 0x80 for fp8, 0x8 for fp4). These types have no negative zero
     // and no infinity.
-    llvm::APFloat negZero = llvm::APFloat::getZero(sem, /*Negative=*/true);
-    smallHasNegZero = negZero.isZero() && negZero.isNegative();
+    if (llvm::APFloat::semanticsHasZero(sem)) {
+      llvm::APFloat negZero = llvm::APFloat::getZero(sem, /*Negative=*/true);
+      smallHasNegZero = negZero.isZero() && negZero.isNegative();
+    } else {
+      smallHasNegZero = false;
+    }
     smallNanIsNegZero = !smallHasNegZero && !smallHasInf && smallHasNan;
   }
 
@@ -361,7 +368,9 @@ private:
 ///    - Max exponent values are valid finite numbers (except NaN encoding).
 ///    - f4E2M1FN has no NaN and no negative zero.
 struct TruncFToSmallFloat final : OpRewritePattern<arith::TruncFOp> {
-  using Base::Base;
+  TruncFToSmallFloat(MLIRContext *ctx, ArrayRef<Type> sourceTypes)
+      : OpRewritePattern(ctx),
+        sourceTypes(sourceTypes.begin(), sourceTypes.end()) {}
 
   LogicalResult matchAndRewrite(arith::TruncFOp op,
                                 PatternRewriter &rewriter) const override {
@@ -372,6 +381,11 @@ struct TruncFToSmallFloat final : OpRewritePattern<arith::TruncFOp> {
     unsigned resultBitWidth = resultElemType.getIntOrFloatBitWidth();
     if (!isa<Float32Type>(inputElemType) ||
         (resultBitWidth != 4 && resultBitWidth != 8)) {
+      return failure();
+    }
+
+    // Only match types that are in our source types list.
+    if (!llvm::is_contained(sourceTypes, resultElemType)) {
       return failure();
     }
 
@@ -567,6 +581,9 @@ struct TruncFToSmallFloat final : OpRewritePattern<arith::TruncFOp> {
     rewriter.replaceOp(op, result);
     return success();
   }
+
+private:
+  SmallVector<Type> sourceTypes;
 };
 
 //===----------------------------------------------------------------------===//
@@ -585,7 +602,9 @@ struct TruncFToSmallFloat final : OpRewritePattern<arith::TruncFOp> {
 /// We implement this using uitofp + mulf with a precomputed scale factor,
 /// which is simpler and more general than enumerating all possible values.
 struct ExtFFromSmallFloat final : OpRewritePattern<arith::ExtFOp> {
-  using Base::Base;
+  ExtFFromSmallFloat(MLIRContext *ctx, ArrayRef<Type> sourceTypes)
+      : OpRewritePattern(ctx),
+        sourceTypes(sourceTypes.begin(), sourceTypes.end()) {}
 
   LogicalResult matchAndRewrite(arith::ExtFOp op,
                                 PatternRewriter &rewriter) const override {
@@ -597,6 +616,11 @@ struct ExtFFromSmallFloat final : OpRewritePattern<arith::ExtFOp> {
     unsigned inputBitWidth = inputElemType.getIntOrFloatBitWidth();
     if ((inputBitWidth != 4 && inputBitWidth != 8) ||
         !isa<Float32Type>(resultElemType)) {
+      return failure();
+    }
+
+    // Only match types that are in our source types list.
+    if (!llvm::is_contained(sourceTypes, inputElemType)) {
       return failure();
     }
 
@@ -728,6 +752,9 @@ struct ExtFFromSmallFloat final : OpRewritePattern<arith::ExtFOp> {
 
     return success();
   }
+
+private:
+  SmallVector<Type> sourceTypes;
 };
 
 struct ConvertUnsupportedFloatArithPass final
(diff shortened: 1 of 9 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
```diff
@@ -137,41 +137,6 @@ static void populateSwapSetPrioWithMFMAPatterns(RewritePatternSet &patterns) {
...
-    auto pred = containsAPred<Float8E5M2FNUZType, Float8E4M3FNUZType>;
-    if (llvm::any_of(op->getOperandTypes(), pred) ||
-        llvm::any_of(op->getResultTypes(), pred)) {
-      return op->emitOpError(
-          "F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 "
-          "(MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.");
-    }
-  }
-  return success();
-}
```
hanhanW, line 173: We can do a warning, though it should be guarded under LDBG. IMO, we dont expect warnings when the input program is valid.

kuhar:
Can we make sure there's a big fat warning when this happens? Also, how do we handle MMA over unsupported types?

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
kuhar: I'd prefer a real mlir warning -- even if the input program is valid, the user is shooting themselves in the foot

hanhanW:
> Can we make sure there's a big fat warning when this happens? Also, how do we handle MMA over unsupported types?

They should already become intrinsics when you reach here. This is a very end step before LLVM conversion. Currently the pass is run right before the conversion.

https://github.com/iree-org/iree/blob/5aa6453d7c52f30e0f79c5fa0cf028f2c67e3957/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L1060-L1069

kuhar:
So we will still error out on any non-trivial inputs? Can we have a test that tries to run matmul with unsupported element types and use `--verify-diagnostics` to document the behavior?

hanhanW:
> So we will still error out on any non-trivial inputs? Can we have a test that tries to run matmul with unsupported element types and use `--verify-diagnostics` to document the behavior?

I have to try, but I don't expect error out on GPU side as long as the codegen path is generic enough. You never choose MMA attribute for such matmul, so it is just a linalg generic op. No matter it is vectorized or not, it becomes a seq of arith ops; the emulation handles both scalar and vector code.

We do have CPU working for scaled matmul e2e; CPU does not support the type natively. It is already enabled on main branch since yesterday.

Sorry, I misread the first comment. I'll try and report back tomorrow. But above is what I'd expect.

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -737,109 +764,98 @@ struct ConvertUnsupportedFloatArithPass final
...
-    arith::populateEmulateUnsupportedFloatsConversions(converter, sourceTypes,
-                                                       targetType);
+    arith::populateEmulateUnsupportedFloatsConversions(
+        converter, allSmallFloatTypes, targetType);
     RewritePatternSet patterns(context);
     arith::populateEmulateUnsupportedFloatsPatterns(patterns, converter);
     ConversionTarget target(*context);
     arith::populateEmulateUnsupportedFloatsLegality(target, converter);
 
+    // Mark scaling ops as legal - they have their own expansion patterns in
```
krzysz00, line 832: I'd suggest adding them here and actually disabling the upstream pass if we're migrating to this system.

On the GPU side, this'll probably mean running the scaling_extf/truncf => intrinsics patterns a bit earlier, or moving this later, both of which are fine, I'd think

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
krzysz00: +1 for a warning for setting off the float emulator

Also, like with f32 on RDNA3, I'd expect a mma or scaled mma that isn't natively supported to go to vector.contract and FMA chains ... or maybe scalar code. Both of which deserve a warning in the small float case.

kuhar:
> I have to try, but I don't expect error out on GPU side as long as the codegen path is generic enough. You never choose MMA attribute for such matmul, so it is just a linalg generic op. No matter it is vectorized or not, it becomes a seq of arith ops; the emulation handles both scalar and vector code.

If we can't find a matching intrinsic, the performance with scalar math is going to be an order of magnitude slower than expected (or more) -- we should emit a warning somewhere, and have a test that makes sure we get it on a matmul/conv dispatch through `// expected-warning@below`.

If someone uses the wrong fp types, it's almost always a user error, not something intentional.

krzysz00:
The main exception to "it'll be slow if we can't find an intrinsic" is f32 on RDNA, where there's no intrinsic because, paraphrasing, "it wouldn't be faster than just doing the FMAs and we don't want to waste an opcode"

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
hanhanW: How it works on CPU is that we have `ArithExpandOps` pass in the pipeline, so they are expanded before we reach to this pass. We can do something similar for GPU.

https://github.com/iree-org/iree/blob/63d1f649e3a6b3a5be322d7758b756f6bbefcd0f/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L622-L628

On GPU side, I think you'll need a custom pass that kick in few AMDGPU conversion. I can follow-up on that later.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
hanhanW: I gave it a shot, and you'll see a lot of warnings (like ~70) when you build `iree-test-deps`. E.g.,

```
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2 and F8E4M3FN types are not supported on gfx942 (MI-300) or older chipsets; try F8E5M2FNUZ or F8E4M3FNUZ instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
<unknown>:0: warning: F8E5M2FNUZ and F8E4M3FNUZ types are not supported on non-gfx942 (MI-300) chipsets; try F8E5M2 or F8E4M3FN instead.
```

I still dont think that we should have such behavior. Do we have any flag (like `--verbose`) that we can guard this behavior under that?

krzysz00: ... we should fix our tests to either expect diagnostics in those cases or to not run on these platforms? Or we can probably suppress the warnings somehow - maybe via said diagnostics checker or a 2>&1) if they're intentional

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
krzysz00: ... I'll point out that those settings for the expansion are causing your fp4 emulation to not fire

hanhanW:
> > I have to try, but I don't expect error out on GPU side as long as the codegen path is generic enough. You never choose MMA attribute for such matmul, so it is just a linalg generic op. No matter it is vectorized or not, it becomes a seq of arith ops; the emulation handles both scalar and vector code.
> 
> If we can't find a matching intrinsic, the performance with scalar math is going to be an order of magnitude slower than expected (or more) -- we should emit a warning somewhere, and have a test that makes sure we get it on a matmul/conv dispatch through `// expected-warning@below`.
> 
> If someone uses the wrong fp types, it's almost always a user error, not something intentional.

> The main exception to "it'll be slow if we can't find an intrinsic" is f32 on RDNA, where there's no intrinsic because, paraphrasing, "it wouldn't be faster than just doing the FMAs and we don't want to waste an opcode"

I think it is already happening on main branch I think. E.g., if you compile f32f32f32 matmul for gfx1100, you use vector<4xf32> in the lowering. There are no intrinsics: https://gist.github.com/hanhanW/2d893b240c5e9cf1aeed3a0b21d6d258

I think small float types are not different in this context. It is easy to provide the bad shape/types that the hardware does not have native support.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
hanhanW: Why not run on CI? It is intended to run. Otherwises, we don't need the software emulation. I don't tink `2>&1` is a good idea. It is too specific for this requirement.

kuhar:
Maybe the solution is to keep your emulation as-is and warn around the kernel configuration time, @krzysz00 @hanhanW ?

krzysz00:
I think there's a meaningful difference between "no intrrinsic" (happens sometimes) and "we're falling back to expensive software emulation between because we don't have this datatype"

(heck, there's probably a parallel warning about f64 being lossy)

kuhar:
The practical difference is that many people can get confused and use the wrong mlir without realizing they are using the unsupported set of fp8/fp4 types. We've had a few situations like this and should warn about it. As for f32 -> f32, or f64, if they are using wide types like this, most likely they don't know what they are doing anyway, so I wouldn't warn.

hanhanW:
> The practical difference is that many people can get confused and use the wrong mlir without realizing they are using the unsupported set of fp8/fp4 types. We've had a few situations like this and should warn about it. As for f32 -> f32, or f64, if they are using wide types like this, most likely they don't know what they are doing anyway, so I wouldn't warn.

I think this kind of things have been happening for a long time. We have been seeing it on CPU side as well, e.g., people don't specify the target cpu and generic is slow. E.g., people don't use data-tiling flag and complain about the performance.

Users usually copy the commands either from docs or codebase. The best I can think of is guarding the warning under a `aggressively-warning` flag. It emits warning only when the flag is on; we expect users to use the flag when they copy the commands. Thus, we still don't have warnings on default path; users can spot the warning because they just copy the commands from somewhere.

hanhanW:
> > The practical difference is that many people can get confused and use the wrong mlir without realizing they are using the unsupported set of fp8/fp4 types. We've had a few situations like this and should warn about it. As for f32 -> f32, or f64, if they are using wide types like this, most likely they don't know what they are doing anyway, so I wouldn't warn.
> 
> I think this kind of things have been happening for a long time. We have been seeing it on CPU side as well, e.g., people don't specify the target cpu and generic is slow. E.g., people don't use data-tiling flag and complain about the performance.
> 
> Users usually copy the commands either from docs or codebase. The best I can think of is guarding the warning under a `aggressively-warning` flag. It emits warning only when the flag is on; we expect users to use the flag when they copy the commands. Thus, we still don't have warnings on default path; users can spot the warning because they just copy the commands from somewhere.

@kuhar I'm going to emit performance warnings like https://github.com/iree-org/iree/pull/23259. I'll introduce a GPU version and emit warnings when the flag is on. We can discuss in a call if we need bandwidth. For now, I'm planning to enable it for O3, most of people would see it because they usually only care performance and use O3. CI does not widely use O3, so it is still clean for our standard build.

hanhanW:
> > > The practical difference is that many people can get confused and use the wrong mlir without realizing they are using the unsupported set of fp8/fp4 types. We've had a few situations like this and should warn about it. As for f32 -> f32, or f64, if they are using wide types like this, most likely they don't know what they are doing anyway, so I wouldn't warn.
> > 
> > 
> > I think this kind of things have been happening for a long time. We have been seeing it on CPU side as well, e.g., people don't specify the target cpu and generic is slow. E.g., people don't use data-tiling flag and complain about the performance.
> > Users usually copy the commands either from docs or codebase. The best I can think of is guarding the warning under a `aggressively-warning` flag. It emits warning only when the flag is on; we expect users to use the flag when they copy the commands. Thus, we still don't have warnings on default path; users can spot the warning because they just copy the commands from somewhere.
> 
> @kuhar I'm going to emit performance warnings like #23259. I'll introduce a GPU version and emit warnings when the flag is on. We can discuss in a call if we need bandwidth. For now, I'm planning to enable it for O3, most of people would see it because they usually only care performance and use O3. CI does not widely use O3, so it is still clean for our standard build.

@kuhar @krzysz00 may I prefetch if this sounds okay to you? I'd like to sort out the issues in advance. Otherwise, I'm working on something not related. Thanks for your time.

kuhar:
We talked about this during the codegen sync. I like @bjacob's suggestion the most: have it off by default and add a verbose flag to enable emulation of unsupported types. Then in the hard error on unsupported types print out the flag name, so that users who want to opt into it can discover it easily.

This also solves the issue of our own tests emitting too many warnings.

hanhanW:
> We talked about this during the codegen sync. I like @bjacob's suggestion the most: have it off by default and add a verbose flag to enable emulation of unsupported types. Then in the hard error on unsupported types print out the flag name, so that users who want to opt into it can discover it easily.
> 
> This also solves the issue of our own tests emitting too many warnings.

Thanks for the summary. I think flipping between `crash + verbose message to enable it with a flag` and `inefficient performance + warning` is an easy switch, so I don't have strong opinion on which one should be default. We can easily flip them based on our needs. I'll go ahead to create a flag to unblock this PR.

---

### Side note

Later on, I can move the flag to GPUCodegenOptions after I complete the support: https://github.com/iree-org/iree/pull/23290

I discussed with Mahesh about CPU one and the overall direction looks good. We'll have the GPU support soon. It is independent to this PR, so this is mostly a note.

hanhanW:
Okay, this is ready for review. @kuhar @krzysz00 can you take a look? Thanks.

compiler/src/iree/compiler/Codegen/Common/test/convert_unsupported_float_arith.mlir
```diff
@@ -350,8 +282,143 @@ func.func @extf_fnuz_nan_handling(%arg0 : f8E5M2FNUZ) -> f32 attributes
...
+// CHECK-LABEL: func.func @gpu_fnuz_gfx942_hw_conversion
+// CHECK-SAME:    (%[[ARG0:.*]]: f8E4M3FNUZ)
+// Arithmetic emulation: wrap with extf/truncf
+// CHECK:         %[[EXT:.*]] = arith.extf %[[ARG0]] {{.*}} : f8E4M3FNUZ to f32
+// CHECK:         %[[NEG:.*]] = arith.negf %[[EXT]] : f32
+// CHECK:         %[[TRUNC:.*]] = arith.truncf %[[NEG]] {{.*}} : f32 to f8E4M3FNUZ
+// No bit manipulation (no arith.bitcast) - conversions use hardware
+// CHECK-NOT:     arith.bitcast
+// CHECK:         return %[[TRUNC]]
+#executable_target_gfx942 = #hal.executable.target<"rocm", "rocm-hsaco-fb", {abi = "hip", iree_codegen.target_info = #iree_gpu.target<arch = "gfx942", features = "", wgp = <compute =  fp64|fp32|fp16|int64|int32|int16|int8, storage =  b64|b32|b16|b8, subgroup =  shuffle|arithmetic, dot =  dp4xi8toi32, subgroup_size_choices = [64], max_workgroup_sizes = [1024, 1024, 1024], max_thread_count_per_workgroup = 1024, max_workgroup_memory_bytes = 65536, max_workgroup_counts = [2147483647, 2147483647, 2147483647], max_load_instruction_bits = 128, simds_per_wgp = 4, vgpr_space_bits = 16384>>, ukernels = "none"}>
```
kuhar: nit: can we trim this down? For example, we shouldn't care about the exact subgroup / dot prod operations `= none`

compiler/src/iree/compiler/Codegen/Common/test/convert_unsupported_float_arith.mlir
```diff
@@ -350,8 +282,143 @@ func.func @extf_fnuz_nan_handling(%arg0 : f8E5M2FNUZ) -> f32 attributes
...
+//
+// With emulation disabled, extf/truncf remain unconverted (no bitcast).
+// NOEMU-LABEL: func.func @gpu_ocp_gfx942_emulated
+// NOEMU-SAME:    (%[[ARG0:.*]]: f8E4M3FN)
+// NOEMU:         %[[EXT:.*]] = arith.extf %[[ARG0]] {{.*}} : f8E4M3FN to f32
+// NOEMU:         %[[NEG:.*]] = arith.negf %[[EXT]] : f32
+// NOEMU:         %[[TRUNC:.*]] = arith.truncf %[[NEG]] {{.*}} : f32 to f8E4M3FN
+// NOEMU-NOT:     arith.bitcast
+// NOEMU:         return %[[TRUNC]]
+#executable_target_gfx942_ocp = #hal.executable.target<"rocm", "rocm-hsaco-fb", {abi = "hip", iree_codegen.target_info = #iree_gpu.target<arch = "gfx942", features = "", wgp = <compute =  fp64|fp32|fp16|int64|int32|int16|int8, storage =  b64|b32|b16|b8, subgroup =  shuffle|arithmetic, dot =  dp4xi8toi32, subgroup_size_choices = [64], max_workgroup_sizes = [1024, 1024, 1024], max_thread_count_per_workgroup = 1024, max_workgroup_memory_bytes = 65536, max_workgroup_counts = [2147483647, 2147483647, 2147483647], max_load_instruction_bits = 128, simds_per_wgp = 4, vgpr_space_bits = 16384>>, ukernels = "none"}>
```
kuhar: same here

compiler/src/iree/compiler/Codegen/Common/test/convert_unsupported_float_arith.mlir
```diff
@@ -350,8 +282,143 @@ func.func @extf_fnuz_nan_handling(%arg0 : f8E5M2FNUZ) -> f32 attributes
...
+// CHECK-LABEL: func.func @gpu_ocp_gfx950_hw_conversion
+// CHECK-SAME:    (%[[ARG0:.*]]: f8E4M3FN)
+// Arithmetic emulation: wrap with extf/truncf
+// CHECK:         %[[EXT:.*]] = arith.extf %[[ARG0]] {{.*}} : f8E4M3FN to f32
+// CHECK:         %[[NEG:.*]] = arith.negf %[[EXT]] : f32
+// CHECK:         %[[TRUNC:.*]] = arith.truncf %[[NEG]] {{.*}} : f32 to f8E4M3FN
+// No bit manipulation (no arith.bitcast) - conversions use hardware
+// CHECK-NOT:     arith.bitcast
+// CHECK:         return %[[TRUNC]]
+#executable_target_gfx950 = #hal.executable.target<"rocm", "rocm-hsaco-fb", {abi = "hip", iree_codegen.target_info = #iree_gpu.target<arch = "gfx950", features = "", wgp = <compute =  fp64|fp32|fp16|int64|int32|int16|int8, storage =  b64|b32|b16|b8, subgroup =  shuffle|arithmetic, dot =  dp4xi8toi32, subgroup_size_choices = [64], max_workgroup_sizes = [1024, 1024, 1024], max_thread_count_per_workgroup = 1024, max_workgroup_memory_bytes = 65536, max_workgroup_counts = [2147483647, 2147483647, 2147483647], max_load_instruction_bits = 128, simds_per_wgp = 4, vgpr_space_bits = 16384>>, ukernels = "none"}>
```
kuhar: same here

compiler/src/iree/compiler/Codegen/Common/test/convert_unsupported_float_arith.mlir
```diff
@@ -350,8 +282,143 @@ func.func @extf_fnuz_nan_handling(%arg0 : f8E5M2FNUZ) -> f32 attributes
...
+//
+// With emulation disabled, extf/truncf remain unconverted (no bitcast).
+// NOEMU-LABEL: func.func @gpu_fnuz_gfx950_emulated
+// NOEMU-SAME:    (%[[ARG0:.*]]: f8E4M3FNUZ)
+// NOEMU:         %[[EXT:.*]] = arith.extf %[[ARG0]] {{.*}} : f8E4M3FNUZ to f32
+// NOEMU:         %[[NEG:.*]] = arith.negf %[[EXT]] : f32
+// NOEMU:         %[[TRUNC:.*]] = arith.truncf %[[NEG]] {{.*}} : f32 to f8E4M3FNUZ
+// NOEMU-NOT:     arith.bitcast
+// NOEMU:         return %[[TRUNC]]
+#executable_target_gfx950_fnuz = #hal.executable.target<"rocm", "rocm-hsaco-fb", {abi = "hip", iree_codegen.target_info = #iree_gpu.target<arch = "gfx950", features = "", wgp = <compute =  fp64|fp32|fp16|int64|int32|int16|int8, storage =  b64|b32|b16|b8, subgroup =  shuffle|arithmetic, dot =  dp4xi8toi32, subgroup_size_choices = [64], max_workgroup_sizes = [1024, 1024, 1024], max_thread_count_per_workgroup = 1024, max_workgroup_memory_bytes = 65536, max_workgroup_counts = [2147483647, 2147483647, 2147483647], max_load_instruction_bits = 128, simds_per_wgp = 4, vgpr_space_bits = 16384>>, ukernels = "none"}>
```
kuhar: same here

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -361,7 +368,9 @@ class FloatEmulationHelper {
 ///    - Max exponent values are valid finite numbers (except NaN encoding).
 ///    - f4E2M1FN has no NaN and no negative zero.
 struct TruncFToSmallFloat final : OpRewritePattern<arith::TruncFOp> {
-  using Base::Base;
+  TruncFToSmallFloat(MLIRContext *ctx, ArrayRef<Type> sourceTypes)
+      : OpRewritePattern(ctx),
+        sourceTypes(sourceTypes.begin(), sourceTypes.end()) {}
```
kuhar: I think this constructor should work
```suggestion
        sourceTypes(sourceTypes) {}
```

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -585,7 +602,9 @@ struct TruncFToSmallFloat final : OpRewritePattern<arith::TruncFOp> {
 /// We implement this using uitofp + mulf with a precomputed scale factor,
 /// which is simpler and more general than enumerating all possible values.
 struct ExtFFromSmallFloat final : OpRewritePattern<arith::ExtFOp> {
-  using Base::Base;
+  ExtFFromSmallFloat(MLIRContext *ctx, ArrayRef<Type> sourceTypes)
+      : OpRewritePattern(ctx),
+        sourceTypes(sourceTypes.begin(), sourceTypes.end()) {}
```
kuhar: same here

compiler/src/iree/compiler/Codegen/Common/ConvertUnsupportedFloatArithPass.cpp
```diff
@@ -737,109 +764,99 @@ struct ConvertUnsupportedFloatArithPass final
...
 
-  // For CPU, emulate extf/truncf to/from small float types using integer ops.
-  // This is needed because LLVM doesn't support fpext/fptrunc for fp4/fp8
-  // types; these types eventually get lowered to small integers in LLVM IR.
-  if (isCPU) {
+  // Determine which types need software conversion emulation. By default, all
+  // the small float types need the emulation. For GPU targets, some types may
+  // have hardware conversion support and can be skipped.
+  SmallVector<Type> typesNeedingConversionEmulation = allSmallFloatTypes;
+  if (auto gpuAttr = getGPUTargetAttr(funcOp)) {
```
kuhar: nit: spell this out

kuhar (approved):

krzysz00 (approved):
Overall setup seems reasonable - can we make sure that scaling_extf/scaling_truncf get emulated too? I'm wanting to make sure there isn't phase ordering issues between your code and the upstream ArithExpandOps 's handling of those ops {and of fp4)

hanhanW:
> Overall setup seems reasonable - can we make sure that scaling_extf/scaling_truncf get emulated too? I'm wanting to make sure there isn't phase ordering issues between your code and the upstream ArithExpandOps 's handling of those ops {and of fp4)

`scaling_extf/scaling_truncf` must be emulated. Otherwise, you won't pass `fp4_f32_conversion.mlir` tests.

```mlir
func.func @f4E2M1FN_scaled_to_f32() {
  // [0, 1.0, 2.0, 0.5] : f4E2M1FN
  %input = util.unfoldable_constant dense<[0, 2, 4, 1]> : tensor<4xi8>
  %scale = util.unfoldable_constant dense<2.0> : tensor<4xf8E8M0FNU>
  %init0 = tensor.empty() : tensor<4xf32>
  %res = linalg.generic {indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>], iterator_types = ["parallel"]}
    ins(%input, %scale : tensor<4xi8>, tensor<4xf8E8M0FNU>) outs(%init0 : tensor<4xf32>) {
  ^bb0(%in: i8, %s: f8E8M0FNU, %out: f32):
    %0 = arith.trunci %in : i8 to i4
    %1 = arith.bitcast %0 : i4 to f4E2M1FN
    %2 = arith.scaling_extf %1, %s : f4E2M1FN, f8E8M0FNU to f32
    linalg.yield %2 : f32
  } -> tensor<4xf32>

  check.expect_eq_const(%res, dense<[0.0, 2.0, 4.0, 1.0]> : tensor<4xf32>) : tensor<4xf32>
  return
}

func.func @f32_to_f4E2M1FN_scaled() {
  %input = util.unfoldable_constant dense<[0.0, 2.0, 4.0, 1.0]> : tensor<4xf32>
  %scale = util.unfoldable_constant dense<2.0> : tensor<4xf8E8M0FNU>
  %init0 = tensor.empty() : tensor<4xi8>
  %res = linalg.generic {indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>], iterator_types = ["parallel"]}
    ins(%input, %scale : tensor<4xf32>, tensor<4xf8E8M0FNU>) outs(%init0 : tensor<4xi8>) {
  ^bb0(%in: f32, %s: f8E8M0FNU, %out: i8):
    %0 = arith.scaling_truncf %in, %s : f32, f8E8M0FNU to f4E2M1FN
    %1 = arith.bitcast %0 : f4E2M1FN to i4
    %2 = arith.extui %1 : i4 to i8
    linalg.yield %2 : i8
  } -> tensor<4xi8>

  // [0.0, 1.0, 2.0, 0.5] : tensor<4xf4E2M1FN>
  check.expect_eq_const(%res, dense<[0, 2, 4, 1]> : tensor<4xi8>) : tensor<4xi8>
  return
}
```

It is handled in `ConvertToROCDLPass` in GPU backends.

https://github.com/iree-org/iree/blob/570967eef9bbe767da82c2eb50d81221f062d9b4/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp#L288

krzysz00:
@hanhanW The code you linked shows the exact phase ordering problem I'm worried about! That code populates the **upstream** emulation patterns for fp4 and f8E8M0FNU, not the ones you added, and can *introduce* extf and truncf operations.

Therefore, those scaled ext/scaled trunc expansions need to be run *before* (or alongside) your patterns, and the two populate calls afterwards need to be removed to ensure consistency.

krzysz00:
(To give a concrete example, if I were passing f8E5M2 to those expansion tests and not an fp4, they wouldn't work, as far as I can tell)

hanhanW:
It seems to be pre-existing issue as it was added to late phase. I noticed something simliar when I made changes in CPU, and maybe we follow it: https://github.com/iree-org/iree/commit/6202dcd16c1cfbf4ff8f6fee9f4d61655f1284a3

PR #23240, opened 2026-01-22
[LinalgExt] Add pattern to canonicalize identity map_gather into a linalg.copy

Abhishek-Varma:
-- This commit implements pattern to canonicalize identity map_gather
    to linalg.copy op.
-- For this `isIdentity` member method for map_gather is also created.
   A map_gather is considered identity if it's just reading from source
   without any index transformation ie. this is true if source and output
   have the same type and the mapping from output to source index is
   identity.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -695,6 +695,47 @@ void MapGatherOp::inlineMapGatherBody(
                                  transformBodyIndices, bodyBuilder);
 }
 
+bool MapGatherOp::isIdentity() {
+  if (getSourceType() != getOutputType()) {
+    return false;
+  }
+  // Check that the block arguments are directly yielded in the order that they
+  // are defined in the block (excluding padding).
+  Block &transformBody = getTransformationRegion().getBlocks().front();
+  auto yieldOp = cast<IREE::LinalgExt::YieldOp>(transformBody.getTerminator());
+  for (unsigned i = 0; i < getSourceRank(); ++i) {
+    auto yieldedBbArg = dyn_cast<BlockArgument>(yieldOp.getOperand(i));
+    if (yieldedBbArg != transformBody.getArgument(i)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+namespace {
+/// Replace an identity map_gather with its source.
+struct ReplaceIdentityMapGather : public OpRewritePattern<MapGatherOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(MapGatherOp op,
+                                PatternRewriter &rewriter) const override {
+    if (!op.isIdentity()) {
+      return failure();
+    }
+    if (!op.hasPureTensorSemantics()) {
+      return failure();
+    }
+    rewriter.replaceOp(op, op.getSource());
+    return success();
+  }
+};
+} // namespace
+
+void MapGatherOp::getCanonicalizationPatterns(RewritePatternSet &results,
+                                              MLIRContext *ctx) {
+  results.add<ReplaceIdentityMapGather>(ctx);
+}
+
 //===----------------------------------------------------------------------===//
 // MapScatterOp
 //===----------------------------------------------------------------------===//
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -587,6 +587,38 @@ void GatherOp::getCanonicalizationPatterns(RewritePatternSet &results,
   results.add<ConvertGatherToExtract>(ctx);
 }
 
+namespace {
+/// Convert an identity map_gather or map_scatter to a copy operation.
+/// We keep the copy to preserve DPS semantics.
+template <typename OpTy>
+struct ConvertIdentityMapGatherScatterToCopy : public OpRewritePattern<OpTy> {
+  using OpRewritePattern<OpTy>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(OpTy op,
+                                PatternRewriter &rewriter) const override {
+    if (!op.isIdentity()) {
+      return failure();
+    }
+    if (op.isVectorized()) {
+      return failure();
+    }
+    if (!op.hasPureTensorSemantics()) {
+      return failure();
+    }
+    Value source;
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -695,6 +695,47 @@ void MapGatherOp::inlineMapGatherBody(
...
+    if (yieldedBbArg != transformBody.getArgument(i)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+namespace {
+/// Replace an identity map_gather with its source.
+struct ReplaceIdentityMapGather : public OpRewritePattern<MapGatherOp> {
```
hanhanW: Let's follow the naming from scatter version (`ConvertIdentityMapScatterToCopy`)?

hanhanW (approved):
LGTM, just one nit for the naming.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
Abhishek-Varma: Sure. I've renamed it to `ConvertIdentityMapGatherToSource` - can you please check this before I merge ?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -695,6 +695,47 @@ void MapGatherOp::inlineMapGatherBody(
...
+
+  LogicalResult matchAndRewrite(MapGatherOp op,
+                                PatternRewriter &rewriter) const override {
+    if (!op.isIdentity()) {
+      return failure();
+    }
+    if (!op.hasPureTensorSemantics()) {
+      return failure();
+    }
+    rewriter.replaceOp(op, op.getSource());
```
hanhanW: Thanks for pinging me offline. I think we should have `linalg.copy` for now (for consistency). See https://github.com/iree-org/iree/pull/22959#discussion_r2661643308 for more details.

> The reason this was creating a copy is to preserve DPS. IIRC, some tests were failing without doing this.
>
> I believe the issue was that for a simple copy dispatch, the copy would be converted to map_scatter, and then get folded away, which creates a dispatch with no compute op. The newer pipelines could probably handle this, but this would go down one of the simple catchall pipelines (I forget which one), and it would cause some failures.
>
> IIRC, we don't convert simple copy dispatches into map_scatter anymore, so that may be why these failures don't happen anymore.
>
> I'm curious what the lowering strategy selection issue was, though? Usually, we can collapse the copy op into the surrounding ops in order to maintain DPS. Since this is in the configuration phase, I'd be a bit wary of breaking DPS, since some old pipelines still rely on it.
>
> That said, if the change is necessary, and the CI is green, then I think it is fine. I'm mostly curious about what the issue was.

Max191: Yeah, let's have it convert into a copy for now. We can revisit it later if the copies cause any issues, but in general I think we shouldn't see identity map_scatter/map_gather unless there was already a copy op in the dispatch (if we do then that's probably a bug), so we should be folding back to copy when it is identity.

Abhishek-Varma: I've changed the implementation now to canonicalize it to a copy op. Have reused MapScatter's pattern and templatized it.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -587,6 +587,38 @@ void GatherOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+    }
+    Value source;
+    if constexpr (std::is_same_v<OpTy, MapGatherOp>) {
+      source = op.getSource();
+    } else {
+      source = op.getInput();
+    }
+    auto copyOp =
+        linalg::CopyOp::create(rewriter, op.getLoc(), source, op.getOutput());
+    rewriter.replaceOp(op, copyOp.getResults());
```
Max191: nit: Use `replaceOpWithNewOp<linalg::CopyOp>`

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -695,6 +727,28 @@ void MapGatherOp::inlineMapGatherBody(
                                  transformBodyIndices, bodyBuilder);
 }
 
+bool MapGatherOp::isIdentity() {
+  if (getSourceType() != getOutputType()) {
+    return false;
+  }
```
Max191, line 731: I just realized that this is actually not enough to prove identity for dynamic cases, since the map_gather could be taking a slice. The isIdentity function has the same bug in map_scatter currently.

You could use the ValueBoundsOpInterface for any dynamic dimensions of the source/output to check if they are equal to each other, and then that should be enough to prove identity: https://github.com/llvm/llvm-project/blob/6f8cea3f90596c1a80e89e2c1e2d2aa75cc9e4e8/mlir/lib/Interfaces/ValueBoundsOpInterface.cpp#L821

The bug already exists in map_scatter, so I am also okay with changing this PR to just bail on dynamic dimensions for now, and then following up with a second PR to fix the bug in map_scatter and add support for dynamic dimensions in map_gather.

hanhanW: ValueBoundsOpInterface is expensive and I think we should exclude it from canonicalizations. Maybe we bail out for dynamic cases, and add such support to the specific pass.

hanhanW: Good catch!

Max191: Yeah, I was thinking the same. We definitely want dynamic support, though, since many cases will have dynamic outer dims (maybe not for conv right now, but definitely for DT, and probably conv in the future).

We could keep the canonicalization as static only and have a special pattern or IR walk that handles dynamic dims using the int range analysis (or maybe value bounds). Int range analysis is preferred, since we can just run it once, but I'm not sure if it has the type of equality checks that we want.

Abhishek-Varma: Good catch, indeed! I've made the updates bailing out on dynamic case - have added a lit test for the same.

Max191 (approved):
Nice, LGTM now!

hanhanW (approved):

PR #23241, opened 2026-01-22
[Bazel][ConstEval] Add missing tool dependency to LITs

AGindinson:
In some Bazel environments, this guards against:
```
error: required embedded linker tool (typically `lld`) not found after searching:
  * --iree-llvmcpu-embedded-linker-path= flag
  * IREE_LLVM_EMBEDDED_LINKER_PATH environment variable
  * common locations at relative file paths
  * system PATH
```


MaheshRavishankar (approved):
I think this is OK. I am not sure I have all the know how here, but seems OK to me.

PR #23244, opened 2026-01-22
[Codegen] Apply clang-tidy fixes to KernelConfig. NFC.

kuhar:
Checking if the new clang-tidy config is reasonable -- IMO looks good

Assisted-by: clang-tidy

PR #21525, opened 2025-07-29
[Codegen][NFC] Switch to new LDBG macro.

hanhanW:
It replaces `LDBG(...);` with `LDBG() ...;`; it drops unused `DBGS()` macro.

PR #23248, opened 2026-01-22
[GPU] Add divisibility comparision to buffer optimization

nirvedhmeshram:
Adds a new capability to go from masked loads to no masked loads for buffer loads by using divisibility analysis. This should not have a big impact by itself at the moment as the heuristics need to be updated to take advantage of this feature. That will be done in a follow up PR that will share performance details on some tracked shapes.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
@@ -7,9 +7,14 @@
 #include "iree/compiler/Codegen/LLVMGPU/Passes.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
 #include "iree/compiler/Codegen/Utils/Utils.h"
+#include "iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.h"
 #include "llvm/Support/DebugLog.h"
+#include "mlir/Analysis/DataFlow/ConstantPropagationAnalysis.h"
+#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
+#include "mlir/Analysis/DataFlowFramework.h"
 
 #define DEBUG_TYPE "iree-codegen-rocdl-buffer-instructions-optimization"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
 
 namespace mlir::iree_compiler {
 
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
```diff
@@ -7,9 +7,14 @@
...
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
 #include "iree/compiler/Codegen/Utils/Utils.h"
+#include "iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.h"
 #include "llvm/Support/DebugLog.h"
+#include "mlir/Analysis/DataFlow/ConstantPropagationAnalysis.h"
+#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
+#include "mlir/Analysis/DataFlowFramework.h"
 
 #define DEBUG_TYPE "iree-codegen-rocdl-buffer-instructions-optimization"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
```
kuhar: Use DebugLog.h instead

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
```diff
@@ -28,10 +33,39 @@ Value createI1And(Location loc, ArrayRef<Value> values, OpBuilder &builder) {
...
+// Check if the innermost mask index is divisible by the mask size using
+// divisibility analysis. Returns true if divisible or if analysis is
+// inconclusive.
+static bool isInnermostMaskIndexDivisible(Value maskIndex, int64_t maskSize,
+                                          DataFlowSolver &solver) {
+  auto *lattice =
+      solver.lookupState<IREE::Util::IntegerDivisibilityLattice>(maskIndex);
+
+  if (lattice && !lattice->getValue().isUninitialized()) {
+    const auto &div = lattice->getValue().getValue();
```
kuhar: spell this out

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
```diff
@@ -119,13 +206,23 @@ struct ROCDLBufferInstructionsOptimizationPass final
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     FunctionOpInterface funcOp = getOperation();
+
+    // Setup divisibility analysis.
+    DataFlowSolver solver;
+    solver.load<dataflow::DeadCodeAnalysis>();
+    solver.load<IREE::Util::IntegerDivisibilityAnalysis>();
+    solver.load<dataflow::SparseConstantPropagation>();
```
krzysz00: Uber-nit: I usually see this is sccp, dead code, then the analysis you're interested in

krzysz00 (approved):
Nifty! Do we know how often this comes up in practice? (approved pending Jakub's commentary)

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
nirvedhmeshram: can you provide an example? This is consistent with one of your recent PRs
https://github.com/iree-org/iree/pull/23244/changes

It is using DebugLog.h AFAICS

kuhar: Sure: https://github.com/iree-org/iree/pull/21525

> This is consistent with one of your recent PRs https://github.com/iree-org/iree/pull/23244/changes

This PR only applied clang-tidy fixes

nirvedhmeshram: Thanks!

kuhar (approved):

nirvedhmeshram:
> Do we know how often this comes up in practice?

@krzysz00 this should not come up much for matmul transpose b type layouts at the moment as the inner dimension is K and the heuristic will adjust the tile size to match the vector load size for example for f16 they are both 8. However for matmul the N dimension is inner for rhs which is padded since [this ](https://github.com/iree-org/iree/pull/22486) PR and that should benefit from this. You will notice this was a bit of a mixed bag by this comment https://github.com/iree-org/iree/pull/22486#issuecomment-3473677028 and now I wouldnt expect those regressions.  However, I will share all quantification of perf gain when I relax the heuristics for this to benefit all layouts, not just one specific case of matmul.

Issue #19072, opened 2024-11-08
[RFC] Formally supporting some suite of "-O*" type flags


qedawkins:
## Motivation

In the process of trying to land #18474 I ran into a number of correctness issues related to LLVM backends where the only saving grace was having a reference that gave correct numerics to compare against. I often found myself hacking in adjustments to the LLVM optimization level for comparison, or string parsing on dispatch names for finer grain control over various pipelines.

Moving forward we still have a fairly large suite of models with correctness issues across multiple backends: https://github.com/nod-ai/e2eshark-reports/blob/main/2024-11-07/ci_reports_onnx/rocm/combined-reports/summary.md + https://github.com/nod-ai/e2eshark-reports/blob/main/2024-11-07/ci_reports_onnx/llvm-cpu/combined-reports/summary.md

We need to find a way to be much better about numerics if we want consistent users of the compiler. Focusing on a few of the models that we care about for only the most optimized paths is not going to help get us there.

Currently we have a set of flags scattered throughout the code base that control various optimizations, most of which are leftover from when the underlying feature was under development. To name a few:

- `--iree-opt-aggressively-propagate-transposes`
- `--iree-opt-outer-dim-concat`
- `--iree-opt-data-tiling`
- `--iree-opt-numeric-precision-reduction`
- `--iree-dispatch-creation-enable-aggressive-fusion`
- `--iree-scheduling-optimize-bindings`
- `--iree-llvmcpu-disable-distribution`
- `--iree-llvmcpu-disable-vector-peeling`
- `--iree-codegen-llvmgpu-use-vector-distribution`
- `--iree-codegen-llvmgpu-use-igemm`

99% of this issue is focused at codegen (as that's where 99%+ of the numerics issues come from) but other phases (e.g. DispatchCreation) also includes a large space of optimization choices.

## Proposal

What I'm proposing is:

- Limit the proliferation of "developer flags." This includes being clearer about the naming when it is for enabling an in-progress feature that is intended to be turned on by default (i.e. `-test-` or `-experimental-` naming prefixes). I don't know what the opinion of `llvm::cl::hidden` is, but that is another option to avoid exposing every new/developer flag to the user. This cleanup will help with formalizing the compiler's APIs for controlling optimizations.
- Expose [codegen] backend optimization level flags (mainly for LLVM based backends, I don't think there is any equivalent in SPIR-V) as compiler options (not floating flags)
[...]

PR #23259, opened 2026-01-23
[CPU] Introduce optimization level to CPU backend.

hanhanW:
The revision adds CPUCodegenOptions and moves few CPU flags to the struct:
- iree-llvmcpu-disable-distribution: always default `off`.
- iree-llvmcpu-fail-on-out-of-bounds-stack-allocation: always default `on`.
- iree-llvmcpu-reassociate-fp-reductions: default `off` for `O0`, `on` for `O2+`
- Merged a pass level test to pipeline test because disabling distribution is a global option.

Note that `iree-opt-level` is the master flag that also controls `iree-llvmcpu-mlir-opt-level` unless you explicitly specify a value.

The opt-level fix causes split_reduction_using_tiling.mlir to fail because FP reassociation is now correctly disabled at O0. The test validates that split reduction produces similar results to normal reduction - a relative tolerance (rtol 0.01%) is more appropriate than absolute tolerance since FP rounding error scales with value magnitude.

It is a step towards https://github.com/iree-org/iree/issues/19072

Assisted-by: Claude


compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
```diff
@@ -110,11 +112,6 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   pipelineOpts.enableAArch64I8mm =
       isAArch64(targetConfig) && hasI8mmFeature(targetConfig);
   pipelineOpts.enablePeeling = isOptEnabled(funcOp, getEnableLoopPeelingStr());
-  if (loweringConfig &&
-      llvm::all_of(loweringConfig.getWorkgroupTileSizes(),
-                   [](int64_t tileSize) { return tileSize == 0; })) {
-    pipelineOpts.disableDistribution = true;
-  }
```
IanWood1, line 117: Is it okay that this was removed?

hanhanW:
I did not enable binding, so I did not get the test failure. It is intended. I was wrong that I thought `iree-opt-level` overwrite other `*-opt-level`. Let me remove the test.

https://github.com/iree-org/iree/blob/f8162058e7ae34951a8b0cce2e5da16f1477f98f/compiler/bindings/python/test/api/api_test.py#L62-L66

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+struct CPUCodegenOptions {
+  llvm::OptimizationLevel optLevel = llvm::OptimizationLevel::O0;
+
+  // Disable thread distribution in codegen.
+  bool disableDistribution = false;
+
+  // Fail if the upper bound of dynamic stack allocation cannot be solved.
+  bool failOnOutOfBoundsStackAllocation = true;
```
IanWood1, line 21: This defaults to `true` but the PR description says:

>iree-llvmcpu-fail-on-out-of-bounds-stack-allocation: always default `off`.

Looks like this was always default `true`, so maybe the description needs to be fixed?

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
```diff
@@ -0,0 +1,61 @@
...
+      llvm::cl::desc("Fail if the upper bound of dynamic stack allocation "
+                     "cannot be solved."),
+      llvm::cl::cat(category));
+
+  binder.opt<bool>("iree-llvmcpu-reassociate-fp-reductions",
+                   reassociateFpReductions,
+                   {init_at_opt(llvm::OptimizationLevel::O0, false),
+                    init_at_opt(llvm::OptimizationLevel::O1, true)},
+                   llvm::cl::desc("Enables reassociation for FP reductions."),
+                   llvm::cl::cat(category));
```
IanWood1, line 37: Should the `iree-llvmcpu-*` options be in the LLVMCPU dir?

compiler/bindings/python/test/api/api_test.py
```diff
@@ -82,6 +82,23 @@ def testOptFlags(self):
...
+            self.assertIn("--iree-llvmcpu-opt-level=O2", flags)
+
+        def testCodegenOptLevelOverride(self):
+            # Override just the CPU opt level, verify it doesn't affect others.
+            session = Session()
+            session.set_flags("--iree-opt-level=O2")
+            session.set_flags("--iree-llvmcpu-opt-level=O0")
+            flags = session.get_flags()
+            self.assertIn("--iree-codegen-opt-level=O2", flags)
+            self.assertIn("--iree-llvmcpu-opt-level=O0", flags)
```
IanWood1: The failure here is expected. If you look at `testOptFlags()` test above, `--iree-global-optimization-opt-level` and `--iree-opt-strip-assertions` are never modified. This test doesn't check that the options "cascade" but rather that they are not modified.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -296,6 +297,12 @@ void GlobalInit::registerCommandLineOptions() {
   clGlobalOptimizationOptions = &GlobalOptimizationOptions::FromFlags::get();
   clParameterOptions = &ParameterOptions::FromFlags::get();
   clDispatchCreationOptions = &DispatchCreationOptions::FromFlags::get();
+
+  // Codegen options (inherit from global opt level).
+  // Note: Registration order matters for inheritance.
+  (void)CodegenOptions::FromFlags::get();
+  (void)CPUCodegenOptions::FromFlags::get();
```
IanWood1: I think we want to handle this in LLVMCPU's plugin registration. It should be possible to use the `createUninitializedSession(OptionsBinder &localOptionsBinder)` hook to bind the options. Then, we should be able to plumb through the options to the passes here:

https://github.com/iree-org/iree/blob/bc8099297cc78a3a542faa72bca7cf1dd60b4cf0/compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp#L865-L870



This should be better because this will allow us to set the options per session.

compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
```diff
@@ -359,8 +360,9 @@ struct DecomposeMismatchEncodingTensorLoadOp
           loadOp, "the source type and the result type match after conversion");
     }
 
-    LDBG() << "Performance warning: decomposing mismatched encoding load op: "
-           << loadOp;
+    if (CodegenOptions::FromFlags::get().emitPerformanceWarnings) {
+      loadOp.emitWarning("decomposing mismatched encoding load op");
+    }
```
IanWood1: This should be plumbed through as a member on the option & a pass option for the passes that use this pattern. The main reason is that `FromFlags::get()` doesn't work with sessions. But it also makes it easier to reproduce the warning if it's a pass option.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -532,7 +528,7 @@ getDefaultDistributionTileSizes(ArrayRef<int64_t> lbs, ArrayRef<int64_t> ubs,
   size_t numDims = lbs.size();
   // Set all the distribution tile sizes to zero if thread distribution is
   // disabled.
-  if (clDisableDistribution) {
+  if (CPUCodegenOptions::FromFlags::get().disableDistribution) {
```
IanWood1: Same here

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
```diff
@@ -95,6 +96,7 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   LoweringConfigAttrInterface loweringConfig = getRootLoweringConfig(funcOp);
   auto pipeline = translationInfo.getDispatchLoweringPassPipeline();
   LLVMCPUPipelineOptions pipelineOpts;
+  pipelineOpts.cpuOpts = CPUCodegenOptions::FromFlags::get();
```
IanWood1: This should be a pass option, too. But it might be a bit trickier.

compiler/bindings/python/test/api/api_test.py
IanWood1: Also, it might be worth it to keep these to ensure that the code here is working correctly.

https://github.com/iree-org/iree/blob/bc8099297cc78a3a542faa72bca7cf1dd60b4cf0/compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp#L994-L1001

IanWood1 (commented):
Nice! I'm excited to see codegen options using the optimization level defaults.

My main feedback is about session compatibility. Currently, the pipeline uses CPUCodegenOptions::FromFlags::get(), which only accesses global CLI flags. This means when the compiler is invoked through a session (e.g., via C API), any flags set via session.setFlags() will be ignored. Only the initial global CLI flags will be used. Getting this working will also make it easier for other backends/plugins to implement similar options.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
hanhanW: Yes, it was okay to remove. We set the tile sizes based on the flag, however the flag is only accesable from KernelDispatch.cpp. This is how it recovers the check, so we could remove it.

I'm adding the support for sessions, and I think it is better to clear the tile sizes after tile size selection.

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
hanhanW: Good question. Initially, I was planning to put each option to their own directories. However, I found that there is an option used by both `LLVMCPU/` and `ExternalInterfaces/`, which determines whether scalable lowering should be enabled or not: [clEnableScalableVectorization](https://github.com/iree-org/iree/blob/56acf7ed012bd0432802891cbc8e3f30ae5deb4e/compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp#L23C28-L27). I'm considering to expose this to public option.

The other point is that putting all the options together is less fragmented, IMO. So you can browse all the options for all the backends in a single file.

Thus, I ended up with having them altogether in this file; it is also easier to skim through all the options because they all locate in a single file. I think I should mention this in PR description. What do you think?

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.h
hanhanW: Good catch, fixed the description. Thanks!

compiler/bindings/python/test/api/api_test.py
hanhanW: I updated the test which makes sure that one of the option is captured by sessions.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
hanhanW: done

hanhanW:
I tested with a [python program](https://gist.github.com/hanhanW/502773855b27359d6d58a7187c10e8a2) that ensures the distribution option is working as expected:

```
Compiling with distribution enabled...
Compiling with distribution disabled...

Running benchmarks...

With distribution (parallel):    9.48 ms
Without distribution (sequential): 202.56 ms
Speedup from distribution: 21.37x
```

compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp
```diff
@@ -857,17 +860,52 @@ class LLVMCPUTargetBackend final : public TargetBackend {
...
     targets.add("llvm-cpu", [=]() {
-      return std::make_shared<LLVMCPUTargetBackend>(options.getTargetOptions());
+      return std::make_shared<LLVMCPUTargetBackend>(options.getTargetOptions(),
+                                                    codegenOptions);
     });
   }
+
+  // Override Registration to also bind CPUCodegenOptions to the session.
+  struct Registration : PluginSession::Registration {
+    using PluginSession::Registration::Registration;
```
hanhanW, line 884: This using statement is required for https://github.com/iree-org/iree/blob/7edade77a7abfe95b3aa5b6c9e95125eb1cb0c8b/compiler/src/iree/compiler/PluginAPI/Client.h#L267-L268

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
```diff
@@ -748,6 +742,10 @@ namespace {
 } // namespace
 
 void registerCodegenLLVMCPUPasses() {
+  // Register codegen options so they're available from the command line.
+  // This ensures the options are registered for both iree-opt and iree-compile.
+  (void)CPUCodegenOptions::FromFlags::get();
```
IanWood1: Is this still needed after setting up the binder through plugin registration?

bjacob (approved):

hanhanW:
@IanWood1 I have a question. Why isn't `iree-opt-level` a master flag? E.g., I'd expect it sets default optimization level for sub-categories like global-opt, codegen, etc. Is it just not implemented yet, or it is intended?

(I'm mainly looking at [this comment](https://github.com/iree-org/iree/issues/19072#issuecomment-2887611959) about having a master flag, i.e., `iree-llvmcpu-opt-level`, for CPU backend, but I think the `iree-opt-level` can serve the role?)

IanWood1 (commented):
Looks good to me in terms of setting up the Binder/Options. However, I'm not familiar enough with this part of the codebase so I'll hold of on approving. One thing that I just noticed is that there already exists an options struct for LLVMCPU named `LLVMCPUTargetCLOptions`:



https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.h#L174


Binder method:
https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.cpp#L481



Given this other struct exists, why do we need the new `CPUCodegenOptions` struct? This isn't blocking. I'm just trying to get an idea of whats going on.

IanWood1:
> @IanWood1 I have a question. Why isn't `iree-opt-level` a master flag? E.g., I'd expect it sets default optimization level for sub-categories like global-opt, codegen, etc. Is it just not implemented yet, or it is intended?
> 
> (I'm mainly looking at [this comment](https://github.com/iree-org/iree/issues/19072#issuecomment-2887611959) about having a master flag, i.e., `iree-llvmcpu-opt-level`, for CPU backend, but I think the `iree-opt-level` can serve the role?)

Maybe I'm misunderstanding what you mean by master flag but `--iree-opt-level` should work as a master flag. For example, if you set `--iree-opt-level=O3` then `--iree-global-optimization-opt-level` will also get set to `O3`. Additionally, when you have a flag that is linked to the optimization level, it will get set based on the pipeline opt level. For example, `--iree-opt-strip-assertions` gets set when `--iree-global-optimization-opt-level` is >O0 (either through setting the global opt flag or the master `--iree-opt-level` flag).

https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/tools/test/compile_flags.mlir demonstrates this behavior.

hanhanW:
> Looks good to me in terms of setting up the Binder/Options. However, I'm not familiar enough with this part of the codebase so I'll hold of on approving. One thing that I just noticed is that there already exists an options struct for LLVMCPU named `LLVMCPUTargetCLOptions`:
> 
> https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.h#L174
> 
> Binder method:
> 
> https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.cpp#L481
> 
> Given this other struct exists, why do we need the new `CPUCodegenOptions` struct? This isn't blocking. I'm just trying to get an idea of whats going on.

How the codegen works in IREE is that we have some lowering/transformation/optimization that takes linalg ops as input and lower them to low-level dialect. On CPU side, they are lowered to LLVM dialect; we conver them to LLVM module and rely on LLVM to do further lowering. Having a new option allows us to set different optimization level on MLIR lowering and LLVM lowering. This is why I'm adding a new option for MLIR part.

> > @IanWood1 I have a question. Why isn't `iree-opt-level` a master flag? E.g., I'd expect it sets default optimization level for sub-categories like global-opt, codegen, etc. Is it just not implemented yet, or it is intended?
> > (I'm mainly looking at [this comment](https://github.com/iree-org/iree/issues/19072#issuecomment-2887611959) about having a master flag, i.e., `iree-llvmcpu-opt-level`, for CPU backend, but I think the `iree-opt-level` can serve the role?)
> 
> Maybe I'm misunderstanding what you mean by master flag but `--iree-opt-level` should work as a master flag. For example, if you set `--iree-opt-level=O3` then `--iree-global-optimization-opt-level` will also get set to `O3`. Additionally, when you have a flag that is linked to the optimization level, it will get set based on the pipeline opt level. For example, `--iree-opt-strip-assertions` gets set when `--iree-global-optimization-opt-level` is >O0 (either through setting the global opt flag or the master `--iree-opt-level` flag).
> 
> https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/tools/test/compile_flags.mlir demonstrates this behavior.

I see, thanks for the sharing on lit tests which demonstrates that it is a master flag. I was confused by the `api_test.py`. I thought that the `iree-global-optimization-opt-level` flag was not set to O2 in this case. I may be misunderstanding about what this test is doing.

https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/bindings/python/test/api/api_test.py#L62-L66

hanhanW (commented):
I found a "bug" that `iree-opt-level` does not work as master flag for `iree-llvmcpu-mlir-opt-level`. Claude helps me the fix: https://github.com/iree-org/iree/pull/23259/changes/f1e7a09ed5edd6b771a6e0083bd0ec5540905b28

Thus, we no longer need the change in `docs/`. The original suggestion includes `iree-opt-level` in the compile command.

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
hanhanW: No, we don't need it. Deleted, thanks!

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.h
```diff
@@ -0,0 +1,32 @@
...
+
+  // Enables reassociation for FP reductions.
+  bool reassociateFpReductions = true;
```
hanhanW: I found the issue because this is set wrongly. After thinking a while, should we just drop all the default values from this struct? The real default values are defined in `.cpp`, IIUC.

hanhanW:
> > Looks good to me in terms of setting up the Binder/Options. However, I'm not familiar enough with this part of the codebase so I'll hold of on approving. One thing that I just noticed is that there already exists an options struct for LLVMCPU named `LLVMCPUTargetCLOptions`:
> > https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.h#L174
> > 
> > Binder method:
> > https://github.com/iree-org/iree/blob/7aa7be5fd3f237f3ef885437f3d1be3f8798dac6/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.cpp#L481
> > 
> > Given this other struct exists, why do we need the new `CPUCodegenOptions` struct? This isn't blocking. I'm just trying to get an idea of whats going on.
> 
> How the codegen works in IREE is that we have some lowering/transformation/optimization that takes linalg ops as input and lower them to low-level dialect. On CPU side, they are lowered to LLVM dialect; we conver them to LLVM module and rely on LLVM to do further lowering. Having a new option allows us to set different optimization level on MLIR lowering and LLVM lowering. This is why I'm adding a new option for MLIR part.

Actually, there is a chance to have a single binder. We can declare a field for CPUCodegenOption in the old struct and reuse the binder. Let me give it a shot.

hanhanW:
I gave it a shot, and I learned the setup better. If we move the CPUOptions into `LLVMCPUTargetCLOptions` and we reuse the binder, we will hit a dependency violation that core codegen will depend on plugin. Because the default value of CPUOptions in `LLVMCPULowerExecutableTargetPass` pass will be `LLVMTargetOptions::FromFlags::get().cpuOpts`. I think core (`Codegen/LLVMCPU/`) can't depend on plugin code (`plugins/target/LLVMCPU/`).

Thus, if we want to keep CPUOptions controled by global without the weird dependency, we'll need to the registeration separately, which leads to current situation.

Summary:

- Option A (current state): Keep CPUCodegenOptions::FromFlags in core. Custom Registration needed to avoid double-registration. Global CLI works everywhere.
- Option B: Remove FromFlags from CPUCodegenOptions. We will lose the global cli control at pass level, which makes logic disconnect.

@IanWood1 does it answer your question?

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -446,6 +452,7 @@ Session::Session(GlobalInit &globalInit)
     cTargetOptions = IREE::VM::getCTargetOptionsFromFlags();
 #endif
   }
+  pluginSession.emplace(globalInit.pluginManager, binder, pluginManagerOptions);
```
IanWood1: Nit: maybe part of the comment added above could be moved here.

IanWood1 (approved):
LGTM on setting up the binder with the plugin.

From the description:
>The opt-level fix causes split_reduction_using_tiling.mlir to fail because FP reassociation is now correctly disabled at O0. The test validates that split reduction produces similar results to normal reduction - a relative tolerance (rtol 0.01%) is more appropriate than absolute tolerance since FP rounding error scales with value magnitude.

Do we have any way to test the binder is working correctly when >O0? It would be nice to have a test, but this might be a difficult thing to check.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
hanhanW: I think the current one looks better because they belong the same code block. Below is the code snippet if we expand the section. Does it make sense?

Image

hanhanW:
> LGTM on setting up the binder with the plugin.
> 
> From the description:
> 
> > The opt-level fix causes split_reduction_using_tiling.mlir to fail because FP reassociation is now correctly disabled at O0. The test validates that split reduction produces similar results to normal reduction - a relative tolerance (rtol 0.01%) is more appropriate than absolute tolerance since FP rounding error scales with value magnitude.
> 
> Do we have any way to test the binder is working correctly when >O0? It would be nice to have a test, but this might be a difficult thing to check.

Good point! Tests are added in https://github.com/iree-org/iree/pull/23259/commits/3fb271bfd7b25915828248bf240bbb81d582e385. Thanks that we have pipeline tests for split reduction flag.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
IanWood1: That makes sense.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -390,7 +390,12 @@ struct Session {
   // PluginManagerOptions must initialize first because the session depends on
   // it.
   PluginManagerOptions pluginManagerOptions;
-  PluginManagerSession pluginSession;
+
+  // Plugin session is constructed after applyOptimizationDefaults() so that
+  // opt-level-dependent defaults are applied before options are copied. The
+  // `std::optional` is required because PluginManagerSession does not have
+  // default constructor. Using `std::optional` allows deferring construction.
+  std::optional<PluginManagerSession> pluginSession;
```
MaheshRavishankar: The only issue I have with this PR is this `std::optional` thing. I cant suggest an alternative though cause I couldnt come up with a "cleaner" way of doing this.

tests/e2e/regression/split_reduction_using_tiling.mlir
```diff
@@ -38,7 +38,7 @@ func.func @simple_large_reduction_tiling_static() {
       %0 = arith.addf %b0, %b1 : f32
       linalg.yield %0 : f32
   } -> tensor<128xf32>
-  check.expect_almost_eq (%normal_reduce, %split_reduction, atol 0.1) : tensor<128xf32>
+  check.expect_almost_eq (%normal_reduce, %split_reduction, atol 0.0, rtol 0.0001) : tensor<128xf32>
```
MaheshRavishankar, line 41: atol of 0? really that is pretty stringent. I guess you turned off reassociation? Its good though if we can keep it this way.

MaheshRavishankar (approved):
I am approving this change. I highlighted one issue I found here with the use of `std::optional` but I cant suggest an alternative. 

@benvanik if you have thoughts here please let us know.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
hanhanW: I gave it another shot and it turns out that we don't need the optional. Because the constructor of `pluginSession` stores reference for `PluginManagerOptions` and it doesn't read the values, so the ordering does not matter here. It is safe as long as the values are not run until `initializePlugins`.

EDIT: I was wrong again, so reverted it back.

tests/e2e/regression/split_reduction_using_tiling.mlir
hanhanW: `rtol` matters here. The default is zero, which is too small. The values are not small, so atol does not play a big role here.

hanhanW:
I fixed a default value in https://github.com/iree-org/iree/pull/23259/commits/384b2e7867d00a2ec557583ea3be486a4ed6fd3c, but it triggered bugs. Then I realized that the `iree-opt-level` never works with `iree-opt` unless we have https://github.com/iree-org/iree/pull/23259/commits/bb0d837c445d5eaa27b266c334523d942d27a987

@IanWood1 can you take a look?

PR #23267, opened 2026-01-23
[GPU] Don't swap expand with slice in the same block

Max191:
Prevents swapping tensor.extract_slice ops with tensor.expand_shape ops when they are in the same block during GPUFuseAndHoistParallelLoops. This pattern is intended to be a tiling fusion pattern for tensor.expand_shape operations, which always begin with the ops in different blocks. When the ops are in the same block, this does not open up more fusions, and instead moves the extract_slice further compute ops that produce loads. This can block later optimizations that fold the slice into loading operations like vector.transfer_read.

This restriction is done through a control function to maintain the ability to swap same-block slice + expand ops, in case it is needed for other use cases.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUFuseAndHoistParallelLoops.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUFuseAndHoistParallelLoops.cpp
@@ -383,7 +383,17 @@ void GPUFuseAndHoistParallelLoopsPass::runOnOperation() {
     populateFuseTilableForallConsumersPattern(patterns);
     patterns.add<FuseCollapseShapeConsumers>(context);
     patterns.add<FuseExtractSliceConsumers>(context);
-    populateSwapExtractWithExpandPattern(patterns);
+    // Only swap expand_shape with extract_slice when they are in different
+    // blocks. This acts as a fusion pattern for the expand_shape, which should
+    // only apply when they are in different blocks (i.e., one inside a loop,
+    // and one outside). Swapping in other cases can interfere with later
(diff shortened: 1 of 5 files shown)


kuhar (approved):

nirvedhmeshram (approved):
Looks like some pipeline tests need looking over/fixing but the idea makes sense

compiler/src/iree/compiler/Codegen/Common/GPU/GPUFuseAndHoistParallelLoops.cpp
```diff
@@ -383,7 +383,17 @@ void GPUFuseAndHoistParallelLoopsPass::runOnOperation() {
     populateFuseTilableForallConsumersPattern(patterns);
     patterns.add<FuseCollapseShapeConsumers>(context);
     patterns.add<FuseExtractSliceConsumers>(context);
-    populateSwapExtractWithExpandPattern(patterns);
+    // Only swap expand_shape with extract_slice when they are in different
+    // blocks. This acts as a fusion pattern for the expand_shape, which should
+    // only apply when they are in different blocks (i.e., one inside a loop,
+    // and one outside). Swapping in other cases can interfere with later
+    // optimizations that fold the slice into consumer load operations.
+    auto swapControlFn = [](OpOperand *operand) {
```
MaheshRavishankar, line 391: Just a FYI, so no action necessary. Everytime I see a control-fn in a pattern to me that implies this pass/transformation is not suitable for pattern rewriter usage.

Max191:
Fixing the pipeline tests requires an improvement to the pattern matching in GPUReduceBankConflicts. I'll convert this to a draft until I resolve it.

PR #23274, opened 2026-01-24
[LLVMGPU] Update seeds to increase KElementCountPerSubgroup

Muzammiluddin-Syed-ECE:
This change aims to improve our ability to reduce bank conflicts for scaled gemms. Currently the seeds lead to a reduction tile of [1, 1], leading to a K tile size of 128 elements. Given that each thread access 32 elements at a time, this leads to a tile size that is too small to benefit from swizzling.

For context see: https://hackmd.io/@ZFjpPhi8Q5ug3nawAd1acw/S12vLtgLWx

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -264,8 +264,7 @@ getGemmHeuristicSeeds(GemmSize gemmSize, int64_t inBitWidth, bool scaled) {
           {/*bestSubgroupCountPerWorkgroup=*/8,
            /*bestMNTileCountPerSubgroup=*/32,
            /*bestKTileCountPerSubgroup=*/4,
-           /*bestKElementCountPerSubgroup=*/kCacheLineSizeBits / 2 /
-               inBitWidth});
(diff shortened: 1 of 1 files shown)


Muzammiluddin-Syed-ECE:
The PR is in draft while I rerun my benchmarking scripts to sweep seeds with bank conflict management to show necessity.

Max191 (commented):
Just FYI, we should really be trying to set these seeds based on hardware details. Increasing this may benefit the newer chips, but could cause regressions on other chips. If we have enough benchmark data to support this change then it is okay for right now IMO, but I just want to note that moving forward we should be trying to compute our heuristics as a function of hardware parameters as opposed to some arbitrary constants based on benchmarks that we collected at some point in time.

Muzammiluddin-Syed-ECE:
> Just FYI, we should really be trying to set these seeds based on hardware details. Increasing this may benefit the newer chips, but could cause regressions on other chips. If we have enough benchmark data to support this change then it is okay for right now IMO, but I just want to note that moving forward we should be trying to compute our heuristics as a function of hardware parameters as opposed to some arbitrary constants based on benchmarks that we collected at some point in time.

This only changes the seeds for scaled matmuls which currently only impact our MI35x chips. So I think it should be easier to prove that for all affected chips this is a reasonable change. 

However, in this case turns out its not a reasonable change 😅 this severely worsens performance for matmuls where the M dimension isnt a power of 2. So ill close this for now.

PR #23280, opened 2026-01-26
[Codegen] Use safer hoisting in OptimizeTensorInsertExtractSlices

Max191:
Use the `moveLoopInvariantCodeFromGuaranteedLoops` transform instead of the `moveLoopInvariantCode` transform in the OptimizeTensorInsertExtractSlices pass. This transform is safer, because it validates that loops will be executed at least once before hoisting loop invariant code. Hoisting from loops that may not execute is not an optimization, so this is a better version of the transformation.

The new safer transform also hoists from linalg.generic ops, so the `moveLoopInvariantCodeFromGenericOps` is removed, since it is no longer used.

This PR also removes the `_batch_matmul_narrow_n_2_dispatch_4_unpack_i32` test, which was doing nothing but checking that a tensor.empty op gets hoisted from an scf.for loop (which cannot be guaranteed to execute). Hoisting empty tensors is not the job of this pass, and the test is verbose, so the test is simply removed.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
@@ -225,23 +225,6 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
   }
 }
 
-void moveLoopInvariantCodeFromGenericOps(Operation *op) {
-  // linalg.generic operations are also loop-like, but they don't have
-  // LoopLikeOpInterface implemented for them.
-  op->walk([&](linalg::GenericOp genericOp) {
-    moveLoopInvariantCode(
-        &genericOp.getBodyRegion(),
-        [&](Value value, Region *) {
-          return !genericOp->isAncestor(value.getParentRegion()->getParentOp());
-        },
-        [&](Operation *op, Region *) {
-          return !isa<linalg::IndexOp>(op) && isMemoryEffectFree(op) &&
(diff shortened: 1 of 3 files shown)


MaheshRavishankar (approved):

krzysz00 (approved):

Max191:
I'm going to convert to draft until I have the pipeline tests under control. There is an annoying issue with broadcast in the VectorDistribute pipeline.

Max191:
Here's the issue with broadcast in VectorDistribute: https://github.com/iree-org/iree/issues/23283

Max191:
The hack we could do is to prevent hoisting from scf.forall ops in OptimizeTensorInsertExtractSlices, since that is what is causing the bug, but it is just masking the real issue IMO. Otherwise, maybe this deserves a longer form discussion.

Groverkss:
> The hack we could do is to prevent hoisting from scf.forall ops in OptimizeTensorInsertExtractSlices, since that is what is causing the bug, but it is just masking the real issue IMO. Otherwise, maybe this deserves a longer form discussion.

I commented on the issue a potential fix that should fix the dps chain. This is fixing a real issue on vector distribute where we aren't hoisting out the redundant computation so we should try to fix it. If it doesn't work, maybe add an option to not fuse out of scf.forall and enable it in VectorDistribute and i'll have a look at it.

Max191:
The suggestion from @Groverkss worked, so once https://github.com/iree-org/iree/pull/23285 is merged, I'll rebase this one and it should be good to go. Thanks for the quick suggestion Kunwar!

PR #23282, opened 2026-01-26
[ROCm][gfx1250] Add e2e matmul tests for gfx1250

kuhar:
Also fix the filter for rdna4 tests.

### DIFF
--- a/tests/e2e/matmul/compilation_info.py
+++ b/tests/e2e/matmul/compilation_info.py
@@ -38,6 +38,7 @@ class CompilationInfoId(enum.Enum):
     LLVMGPUVectorDistributeMFMA = "LLVMGPUVectorDistributeMFMA"
     LLVMGPUVectorDistributeWMMAR3 = "LLVMGPUVectorDistributeWMMAR3"
     LLVMGPUVectorDistributeWMMAR4 = "LLVMGPUVectorDistributeWMMAR4"
+    LLVMGPUVectorDistributeWMMA1250 = "LLVMGPUVectorDistributeWMMA1250"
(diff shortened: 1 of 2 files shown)


tests/e2e/matmul/compilation_info.py
```diff
@@ -271,10 +274,46 @@ def get_rocm_test_compilation_infos(
             MMASchedule("WMMAR4_I32_16x16x16_I8", 2, 4, 2, 1, 2),
             MMASchedule("WMMAR4_I32_16x16x16_I8", 4, 2, 4, 2, 2),
         ]
+    elif intrinsic == "WMMA1250":
+        # gfx1250 WMMA intrinsics: 16x16 tiles with various K sizes.
+        # F16: K=32, F8E4M3FN: K=64 or K=128, I8: K=64
```
krzysz00, line 279: Should we also have bf16? f8E5M2?

krzysz00 (approved):
Looks good overall

tests/e2e/matmul/compilation_info.py
kuhar: I mean, it depends on how exhaustive we want to be -- I doubt we are interested in verifying every single intrinsic out there. Do we realistically expect f16 to work but b16 to fail, or similar with different fp8 variants?

kuhar: (Not sure where the line should be, I picked something that seemed to give us a good enough coverage without being super verbose)

bjacob:
Post-merge LGTM :-)

PR #23286, opened 2026-01-26
[e2e] Increase test timeout for gfx1250

kuhar:
Expose timeout as an optional argument in `iree_native_test` in matmul tests. Regression tests already know how to translate the bazel timeout parameter to seconds.

Assisted-by: claude

### DIFF
--- a/tests/e2e/regression/CMakeLists.txt
+++ b/tests/e2e/regression/CMakeLists.txt
@@ -174,6 +174,8 @@ iree_check_single_backend_test_suite(
     "notsan"
     "noubsan"
     "requires-gpu-amd"
+  TIMEOUT
+    300
 )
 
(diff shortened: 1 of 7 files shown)


tests/e2e/regression/CMakeLists.txt
```diff
@@ -174,6 +174,8 @@ iree_check_single_backend_test_suite(
     "notsan"
     "noubsan"
     "requires-gpu-amd"
+  TIMEOUT
```
krzysz00, line 177: Can we leave a note pointing out that we should remove this after the chip is out for real?

krzysz00 (approved):

bjacob:
Post-merge LGTM :-)

PR #23290, opened 2026-01-27
[GPU] Introduce and plumb through GPUCodegenOptions.

hanhanW:
The revision mirrors the CPUCodegenOptions pattern from the LLVMCPU backend. For now, it only expose `enablePrefetch` to public because it is the only flag used in `tests/external/iree-test-suites`. Other flags can be exposed base on needs. The remaining flags are supposed to be hidden from CLI helpers, since they are either for testing purpose or experiments.

It also drops `=true` from `tests/external/iree-test-suites/*` because it is redundant and we want consistency.

### DIFF
--- a/compiler/plugins/target/ROCM/ROCMTarget.cpp
+++ b/compiler/plugins/target/ROCM/ROCMTarget.cpp
@@ -19,6 +19,7 @@
 #include "iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.h"
 #include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
 #include "iree/compiler/Codegen/LLVMGPU/Passes.h"
+#include "iree/compiler/Codegen/Utils/CodegenOptions.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
 #include "iree/compiler/Codegen/Utils/Utils.h"
 #include "iree/compiler/Dialect/Encoding/IR/EncodingTypes.h"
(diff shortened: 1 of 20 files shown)


hanhanW:
It depends on https://github.com/iree-org/iree/pull/23259. It is better to wait @IanWood1 to review the other one, just in case if there are red flags or not. (I don't want to waste reviewers' time.)

krzysz00:
Does controlling LLVM's opt optimization level and the LLC optimization level also fall into these options or are they handled separately?

hanhanW:
> Does controlling LLVM's opt optimization level and the LLC optimization level also fall into these options or are they handled separately?

Coming from CPU's world and the LLVM lowering path, I think we eventually will have at least two optimization flags, one is for MLIR level codegen that IREE is doing, and the other is LLVM opt level for LLVM module. E.g., https://github.com/iree-org/iree/issues/19072#issuecomment-2887568131

```
  // Optimization level to be used by the LLVM optimizer (middle-end).
  llvm::OptimizationLevel optimizerOptLevel;
  // Optimization level to be used by the LLVM code generator (back-end).
  llvm::CodeGenOptLevel codeGenOptLevel;
```

This one is mainly for MLIR level that we control our optimization in IREE.

compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -372,7 +373,8 @@ static void checkRegisterSpilling(IREE::HAL::ExecutableVariantOp &variantOp,
 
 class ROCMTargetBackend final : public TargetBackend {
 public:
-  ROCMTargetBackend(const ROCMOptions &options) : options(options) {}
+  ROCMTargetBackend(const ROCMOptions &options, GPUCodegenOptions codegenOpts)
```
kuhar, line 376: Is the warning above is actionable?

kuhar (approved):

compiler/plugins/target/ROCM/ROCMTarget.cpp
hanhanW: Yes, I think so. We should move these classes into an anonymous namespace. There are several classes need the fix (e.g., AMDGPUTargetDevice, HIPTargetDevice, and other backends, etc.), so I'll send a follow-up to fix them.

PR #23292, opened 2026-01-27
[LinalgExt] Rewriter for Torch::HigherOrderFlexAttentionOp -> LinalgExt::OnlineAttentionOp

keshavvinayak01:

I ran the entire flex_attention_hop implementation with randomised input tensors, (Also see https://github.com/llvm/torch-mlir/pull/4366) through aot.export and compared against eager mode, and I noticed no accuracy losses (On CPU)

Test: [Torch ops test PR ](https://github.com/iree-org/iree-test-suites/pull/149)

### DIFF
--- a/compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
@@ -397,24 +397,20 @@ struct FlexAttentionOpConversion
     Value rowRedEmpty =
         tensor::EmptyOp::create(rewriter, loc, rowRedShape, floatType);
 
-    Value accInit = arith::ConstantOp::create(
-        rewriter, loc, rewriter.getFloatAttr(floatType, 0.0));
-    Value maxInit = arith::ConstantOp::create(
-        rewriter, loc,
-        rewriter.getFloatAttr(floatType,
-                              APFloat::getLargest(floatType.getFloatSemantics(),
-                                                  /*Negative=*/true)));
-    Value sumInit = arith::ConstantOp::create(
-        rewriter, loc, rewriter.getFloatAttr(floatType, 0.0));
+    Value zeroInit = arith::getIdentityValue(arith::AtomicRMWKind::addf,
+                                             floatType, rewriter, loc);
+    Value maxInit = arith::getIdentityValue(arith::AtomicRMWKind::maximumf,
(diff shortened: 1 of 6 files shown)


MaheshRavishankar:
@keshavvinayak01  moving PRs around makes it hard to track what is new and what has been up for a while. It is disruptive for reviewers. Can we keep this a bit more stable?

keshavvinayak01:
@Groverkss Could you close reviews on this?

keshavvinayak01 (commented):
Re-opening with the rewriter converting to `online_attention` directly instead of AttentionOp. We might also use this torch op in fusilli cc @sjain-stanford so pulling you in for reviews.

cc @MaheshRavishankar @Groverkss

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -162,6 +172,340 @@ struct FftRfftOpConversion : OpRewritePattern<torch::Torch::AtenFftRfftOp> {
...
+static Value convertToBuiltinTensor(PatternRewriter &rewriter, Location loc,
+                                    Value torchTensor) {
+  auto tensorType = cast<torch::Torch::ValueTensorType>(torchTensor.getType());
+  return torch::TorchConversion::ToBuiltinTensorOp::create(
+      rewriter, loc, tensorType.toBuiltinTensor(), torchTensor);
+}
+
+/// Inline a single-block torch function's body at the current insertion point.
+/// Returns the result value. Falls back to func.call for multi-block functions.
+static Value inlineTorchFunction(PatternRewriter &rewriter, Location loc,
```
MaheshRavishankar: Probably want to make it return a `SmallVector<Value>` to be able to handle multiple return values from a function (even though this isnt the case here).
I'd expect this to already be a utility function somewhere?

MaheshRavishankar (approved):
Overall looks fine to me.

@rsuderman  can you review this PR and the follow ups on this if no one gets to it.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: No utility at this level. There's `llvm::InlineFunction`, `mlir::inlineCall`, and also a `CastAndCallOp` but I doubt we can use it here.

Issue #21955, opened 2025-09-12
[CodeGen] Teach linalg conv transformation with generalized form.


hanhanW:
IREE tends to generalize linalg operations, which improves the fusion. E.g., [Improve transpose fusion for conv](https://github.com/iree-org/iree/pull/21778) generalizes all the convolution ops to improve the fusion. However, it regresses backends that can't handle generalized form. The main reason is that they match named ops when they set lowering configs, and some transformations do not support `linalg.generic` form yet.

E.g., CPU backend specialize the lowering configs for the set of op. One missing feature is to have matchers for the ops.

https://github.com/iree-org/iree/blob/f9b7bf1e7a32c5b28cceb8732beb7526223c136b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp#L2887-L2894

For those convolution ops, they all go to [ConvTileAndDecomposeExpert](https://github.com/iree-org/iree/blob/515f29290ce16e140e92c670e8cc17acff378904/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L325C9-L379) pipeline. There are two passes that need to be improved; they live in upstream.

- [DecomposeConvolutionToLowerDimOpsPass](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/DecomposeConvolutionToLowerDimOps.cpp): this is the pass that downscale 2D convolution to 1D convolution when either H or W dimensions are ones. The work include updating [the upstream decompose convolution patterns](https://github.com/llvm/llvm-project/blob/580fdeb6ff55fcd54be16ed8555eaaa6a9aee1c0/mlir/lib/Dialect/Linalg/Transforms/Transforms.cpp#L1631-L1650) to handle generic form.
[...]

PR #23294, opened 2026-01-27
[Codegen][Common] Allow generic conv ops to decompose to lower dim ops

Abhishek-Varma:
-- This commit makes use of `isaConvolutionOpInterface` which works
   for both named and generic convolution ops.
-- This is needed to enable generalizing convolution ops as part of
   https://github.com/iree-org/iree/issues/21955

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/decompose_conv2d.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/decompose_conv2d.mlir
@@ -1,7 +1,7 @@
 // RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-decompose-convolution-to-lower-dim-ops))" --split-input-file %s | FileCheck %s
 // Test the same patterns on generic convolution ops by first generalizing the
 // named ops. This ensures decomposition works on both named and generic convs.
-// RUN: iree-opt --pass-pipeline="builtin.module(func.func(linalg-generalize-named-ops,iree-codegen-decompose-convolution-to-lower-dim-ops))" --split-input-file %s | FileCheck %s --check-prefix=GENERIC
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(linalg-generalize-named-ops,iree-codegen-decompose-convolution-to-lower-dim-ops))" --split-input-file %s | FileCheck %s
 
 #config = #iree_codegen.lowering_config<tile_sizes = [[0, 0, 0, 0, 0, 0], [1, 1, 1, 4, 0, 0], [0, 0, 0, 0, 1, 4], [0, 0, 0, 0, 0, 0]]>
 module {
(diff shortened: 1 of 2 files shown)


Max191 (approved):

compiler/src/iree/compiler/Codegen/Common/test/decompose_conv2d.mlir
```diff
@@ -1,4 +1,7 @@
 // RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-decompose-convolution-to-lower-dim-ops))" --split-input-file %s | FileCheck %s
+// Test the same patterns on generic convolution ops by first generalizing the
+// named ops. This ensures decomposition works on both named and generic convs.
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(linalg-generalize-named-ops,iree-codegen-decompose-convolution-to-lower-dim-ops))" --split-input-file %s | FileCheck %s --check-prefix=GENERIC
```
hanhanW: This one can use `CHECK`, because it aims to decompose (down ranking is a better term IMO) conv ops.

Abhishek-Varma: So the generalizing pass upstream does not preserve attributes of an op. The lit test input had lowering_config which got missed because of that and `CHECK` was testing for it.

Since this pass is not dependent on the lowering_config, I've removed it from the input and the corresponding `CHECK` line.

Now, we can reuse the rest of the `CHECK` statements.

Let me know if it looks okay to you - I can merge this accordingly.

hanhanW: We should preserve the lowering config. ARM folks used to rely on this (cc @banach-space )

Can you update the logic?

https://github.com/iree-org/iree/blob/fc06a5d009c4068b3ef06b7d79fd9a97e879620a/compiler/src/iree/compiler/Codegen/Common/DecomposeConvolutionToLowerDimOps.cpp#L59-L77

hanhanW (changes requested):
put a blocker

compiler/src/iree/compiler/Codegen/Common/test/decompose_conv2d.mlir
Abhishek-Varma: Done. Can you check now ?

hanhanW (approved):
Thanks!

yash-amd:
Due to this commit we are seeing numeric failure for some HF onnx models in amd-shark-testsuite including hf_densenet121.ra_in1k

for repro:

```
git clone git@github.com:nod-ai/AMD-SHARK-TestSuite.git
cd AMD-SHARK-TestSuite/alt_e2eamdshark
export CACHE_DIR=<some-dir>
python -m venv test_suite.venv && source test_suite.venv/bin/activate
pip install --upgrade pip
pip install -r ./base_requirements.txt
pip install -r ./hf_requirements.txt
build iree on your commit 
source /path/to/iree-build/.env && export PYTHONPATH
```

run :
```
python run.py -va -b llvm-cpu -d local-task -t hf_densenet121.ra_in1k --report
```
after this runs, the model will show failure at numerics,
detailed logs are in test-run/hf_densenet121.ra_in1k/


we can also find which operator is causing the issue -> we have support for that in test-suite

hanhanW:
> Due to this commit we are seeing numeric failure for some HF onnx models in amd-shark-testsuite including hf_densenet121.ra_in1k
> 
> for repro:
> 
> ```
> git clone git@github.com:nod-ai/AMD-SHARK-TestSuite.git
> cd AMD-SHARK-TestSuite/alt_e2eamdshark
> export CACHE_DIR=<some-dir>
> python -m venv test_suite.venv && source test_suite.venv/bin/activate
> pip install --upgrade pip
> pip install -r ./base_requirements.txt
> pip install -r ./hf_requirements.txt
> build iree on your commit 
> source /path/to/iree-build/.env && export PYTHONPATH
> ```
> 
> run :
> 
> ```
> python run.py -va -b llvm-cpu -d local-task -t hf_densenet121.ra_in1k --report
> ```
> 
> after this runs, the model will show failure at numerics, detailed logs are in test-run/hf_densenet121.ra_in1k/
> 
> we can also find which operator is causing the issue -> we have support for that in test-suite

@Abhishek-Varma can you take a look? I think you can isolate the dispatch and see what's happening.

Tips:

- You can use `--iree-hal-dump-executable-sources-to` to dump the dispatches and check which one has a convolution op.
- You can dump the inputs and outputs for the specific dispatch to create the small repro.

hanhanW:
Oh wait, this one was reverted, and got relanded in https://github.com/iree-org/iree/pull/23455. The upstream fix was landed, and @Abhishek-Varma confirmed that the other issue (https://github.com/iree-org/iree/issues/23382) was fixed. @yash-amd can you try with the fresh build? Which IREE version are you using?

yash-amd:
@hanhanW , I tried with 19th of feb IREE release as well, I will try with latest iree as well and will confirm here.

yash-amd:
Hi @hanhanW,
I tested the same model with the latest commit and its still giving failure at numerics.

hanhanW:
> Hi @hanhanW, I tested the same model with the latest commit and its still giving failure at numerics.

@Abhishek-Varma please take a look

hanhanW:
@yash-amd can you file an issue and assign it to @Abhishek-Varma ? It is better to track the problem using Github issue, instead of using PR.

PR #23295, opened 2026-01-27
[LLVMCPU] Allow generic conv ops too in KernelDispatch

Abhishek-Varma:
-- Named convolution ops are assigned `CPUConvTileAndDecomposeExpert`
   and we'd need the same to be in the case of generic convolution
   ops as well. Changes in this patch achieves are required to enable
   that.
-- This is needed to enable generalizing convolution ops as part of
   https://github.com/iree-org/iree/issues/21955

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -3195,14 +3195,6 @@ setRootConfigImpl(mlir::FunctionOpInterface entryPointFn, Operation *op,
                 IREE::LinalgExt::WinogradInputTransformOp,
                 IREE::LinalgExt::WinogradOutputTransformOp>(
               [&](auto op) { return setWinogradRootConfig(entryPointFn, op); })
-          .Case<linalg::Conv2DNhwcHwcfOp, linalg::Conv2DNchwFchwOp,
-                linalg::PoolingNhwcSumOp, linalg::PoolingNhwcMaxOp,
-                linalg::PoolingNhwcMaxUnsignedOp, linalg::PoolingNhwcMinOp,
-                linalg::PoolingNhwcMinUnsignedOp, linalg::PoolingNchwSumOp,
-                linalg::PoolingNchwMaxOp, linalg::DepthwiseConv2DNhwcHwcOp>(
-              [&](auto op) {
-                return setConvInterfaceRootConfig(entryPointFn, op);
-              })
           .Default([&](Operation *op) { return failure(); });
   if (succeeded(result)) {
     return result;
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_lowering_strategy.mlir
```diff
@@ -1,4 +1,8 @@
 // RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' --split-input-file %s | FileCheck %s
+// Test the same lowering strategy selection on generic convolution ops by first
+// generalizing the named ops. This ensures convolution pipeline selection works
+// on both named and generic convs.
+// RUN: iree-opt --pass-pipeline='builtin.module(func.func(linalg-generalize-named-ops),iree-llvmcpu-select-lowering-strategy)' --split-input-file %s | FileCheck %s --check-prefix=GENERIC
```
hanhanW, line 5: Why do we need a new `check-prefix`? Is it because that the matmul and other operations get generalized which causes failing?

Abhishek-Varma: Yes, the named ops get generalized and the corresponding `CHECK`-line fails.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -3199,7 +3208,14 @@ setRootConfigImpl(mlir::FunctionOpInterface entryPointFn, Operation *op,
...
   }
 
+  // Check for linalg.generic ops that match convolution patterns using
+  // isaConvolutionOpOfType. This allows generalized conv ops to use the
+  // specialized CPUConvTileAndDecomposeExpert pipeline.
   if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
+    if (is2DConvOp(linalgOp) || is2DDepthConvOp(linalgOp) ||
+        is2DPoolingOp(linalgOp)) {
+      return setConvInterfaceRootConfig(entryPointFn, linalgOp);
+    }
```
hanhanW, line 3210: I think we should delete line 3198 case. I.e., the conv checks above. So we make this become the single entry.

hanhanW (approved):
LG, just one nit

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
Abhishek-Varma: Nice catch!

PR #23301, opened 2026-01-27
[GPU] Support some collapse_shape ops in GPUReduceBankConflicts

Max191:
GPUReduceBankConflicts would always fail if there was a memref.collapse_shape in the use chain of the memref.alloc op. However, there are some cases where padding the allocation is possible, but the pattern matching was too conservative.

This PR relaxes the constraint by verifying that `replaceMemrefUsesAndPropagateType` will succeed before generating any new IR. If the type propagation is possible, then the padded allocation and subview are generated and the new strided type is propagated.

There was also a bug in the original implementation of `replaceMemrefUsesAndPropagateType` for collapse_shape, which was exposed by a failing test in pad_dynamic_alloc.mlir. The test was testing an invalid transformation, so this PR updates the test as well.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/pad_dynamic_alloc.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/pad_dynamic_alloc.mlir
@@ -57,15 +57,15 @@ func.func @dynamic_alloc_collapse_consumer(%id : index) {
   %c0 = arith.constant 0 : index
   %cst = arith.constant 0.000000e+00 : f32
   %0 = util.assume.int %id<umin = 0, umax =  32> : index
-  %1 = memref.alloc(%0, %0) : memref<?x?xf32, 3>
-  %2 = memref.collapse_shape %1 [[0, 1]] : memref<?x?xf32, 3> into memref<?xf32, 3>
+  %1 = memref.alloc(%0) : memref<?x32xf32, 3>
+  %2 = memref.collapse_shape %1 [[0, 1]] : memref<?x32xf32, 3> into memref<?xf32, 3>
   memref.store %cst, %2[%c0] : memref<?xf32, 3>
   return
 }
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Utils/Utils.cpp
```diff
@@ -1430,84 +1430,152 @@ void addConfigDenormalFpMathF32(MLIRContext *context,
...
+  Operation *newOp =
+      mlir::clone(rewriter, user, llvm::to_vector(*resultTypes), newOperands);
+  LDBG() << "\t\tNew user : " << *newOp;
+  return llvm::to_vector_of<Value>(newOp->getResults());
+}
+
+LogicalResult canReplaceMemrefUsesAndPropagateType(Value origValue,
+                                                   Type replacementType) {
+  SmallVector<std::pair<Value, Type>> worklist;
+  worklist.push_back({origValue, replacementType});
```
kuhar:
```suggestion
  SmallVector<std::pair<Value, Type>> worklist = {{origValue, replacementType}}
```

kuhar: test reply via vscode extension

compiler/src/iree/compiler/Codegen/Common/test/pad_dynamic_alloc.mlir
```diff
@@ -57,15 +57,15 @@ func.func @dynamic_alloc_collapse_consumer(%id : index) {
   %c0 = arith.constant 0 : index
```
Max191, line 57: The old implementation had a bug, which was exposed by this PR. This used to be an invalid transformation, so I updated the test in this PR now that the bug is fixed.

nirvedhmeshram (approved):
LGTM

PR #23307, opened 2026-01-28
[VectorDistribution] Add support for `shape_cast`

sommerlukas:
Add support for vector distribution of `vector.shape_cast` operations. This includes support for propagating vector layouts through `shape_cast` in vector layout analysis as well as patterns to lower `vector.shape_cast` to a vector-distributed version.

This PR is based on a prototype implementation by @Groverkss.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -96,6 +96,7 @@ void LayoutInfo::propagateLayoutForward(Value val) {
       Operation *parentOp = cast<vector::MaskOp>(yieldOp->getParentOp());
       Value result = parentOp->getResult(operandIdx);
       setLayoutIfUnset(result, layout);
+      continue;
     }
 
     if (OpTrait::hasElementwiseMappableTraits(user)) {
@@ -183,6 +184,13 @@ void LayoutInfo::propagateLayoutForward(Value val) {
       AffineMap maskMap =
           inversePermutation(compressUnusedDims(write.getPermutationMap()));
       setLayoutOrClone(&mask, layout.apply(maskMap));
+      continue;
+    }
+
+    if (auto shapeCast = dyn_cast<vector::ShapeCastOp>(user)) {
+      setLayoutIfUnset(
+          shapeCast.getResult(),
+          layout.reshape(shapeCast.getResultVectorType().getShape()));
       continue;
     }
   }
@@ -290,10 +298,21 @@ void LayoutInfo::propagateLayoutBackward(Value val) {
     }
     return;
   }
+
+  if (auto shapeCast = dyn_cast<vector::ShapeCastOp>(defOp)) {
+    setLayoutOrClone(
+        &shapeCast.getSourceMutable(),
+        layout.reshape(shapeCast.getSourceVectorType().getShape()));
+    return;
+  }
 }
 
 void LayoutInfo::setLayoutOrClone(OpOperand *val,
                                   VectorLayoutInterface layout) {
+  if (!layout) {
+    // No layout to set.
+    return;
+  }
   if (!isa<ShapedType>(val->get().getType())) {
     // Don't set layouts on non-shaped types. This would anyway be an empty
     // layout.
@@ -322,12 +341,15 @@ void LayoutInfo::setLayoutOrClone(OpOperand *val,
     return;
   }
 
-  // Otherwise, create a to_layout op to change the layout.
-  Value v = val->get();
-  Value layourtedV = ToLayoutOp::create(b, v.getLoc(), v, layout);
-  val->set(layourtedV);
-  layouts[layourtedV] = layout;
-  return;
+  if (getLayout(val->get()) != layout) {
+    // Create `to_layout` op to change layout if it's not the same as the
+    // existing.
+    Value v = val->get();
+    Value layourtedV = ToLayoutOp::create(b, v.getLoc(), v, layout);
+    val->set(layourtedV);
+    layouts[layourtedV] = layout;
+    return;
+  }
 }
 
 LogicalResult propagateVectorLayoutInfo(
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
@@ -21,6 +21,8 @@
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/IR/PatternMatch.h"
 
+#define DEBUG_TYPE "iree-vector-ext-layout-attr"
+
 using namespace mlir;
 
 namespace mlir::iree_compiler::IREE::VectorExt {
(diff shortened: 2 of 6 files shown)


compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -96,6 +96,7 @@ void LayoutInfo::propagateLayoutForward(Value val) {
       Operation *parentOp = cast<vector::MaskOp>(yieldOp->getParentOp());
       Value result = parentOp->getResult(operandIdx);
       setLayoutIfUnset(result, layout);
+      continue;
```
sommerlukas, line 99: While looking through the implementation, I noticed that this is the only case where we don't `continue` to the next loop iteration. I assume that's a simple oversight and therefore added it.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -98,6 +100,158 @@ VectorLayoutInterface NestedLayoutAttr::apply(AffineMap map) const {
...
+  // We can only go to a lower level if this is a new dimension, and we have
+  // to start from the lowest level (element).
+  //
+  // When merging a thread/subgroup level, we need to also ensure that the
+  // strides can be merged into a single stride.
+  SmallVector<int64_t> remainingLevels = {1, 1, 1, 1, 1};
+  SmallVector<int64_t> remainingStrides = {0, 0};
+  SmallVector<bool> unitDims = {true, true, true, true, true};
+  int64_t currDim = getRank();
+  int64_t currLevel = 5;
```
sommerlukas, line 139: @Groverkss In your prototype, this was initialized to `0`. However, this means that on the first 5 iterations, we will just hit the `if (unitDims[currLevel]) {...` case, because we didn't initialize `remainingLevels` etc. Starting with `5`, we should initialize them on the first iteration (`if (curLevel == 5)...`). The result should be the same, but we don't need the first five iterations.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -98,6 +100,157 @@ VectorLayoutInterface NestedLayoutAttr::apply(AffineMap map) const {
                                subgroupStrides, threadStrides);
 }
 
+namespace {
+static SmallVector<int64_t> shuffle(unsigned rank, ArrayRef<int64_t> input,
+                                    bool simple) {
+  if (simple) {
+    SmallVector<int64_t> result(input);
+    std::reverse(result.begin(), result.end());
+    return result;
```
kuhar:
```suggestion
    return llvm::to_vector(llvm::reverse(input));
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -98,6 +100,157 @@ VectorLayoutInterface NestedLayoutAttr::apply(AffineMap map) const {
...
+  }
+  SmallVector<int64_t> result;
+  result.reserve(input.size());
+  for (unsigned r = 0; r < rank; ++r) {
+    for (unsigned i = r; i < input.size(); i += rank) {
```
kuhar: nit: don't re-evaluate end
```suggestion
    for (unsigned i = r, e = input.size(); i < e; i += rank) {
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -98,6 +100,157 @@ VectorLayoutInterface NestedLayoutAttr::apply(AffineMap map) const {
                                subgroupStrides, threadStrides);
 }
 
+namespace {
+static SmallVector<int64_t> shuffle(unsigned rank, ArrayRef<int64_t> input,
```
kuhar: Can you add a documentation comment? The name is quite generic.

sommerlukas: The `shuffle` function was left over from an earlier experiment that I did. I've removed the entire function now, thanks for catching that!

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: yeah this is good to add.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -98,6 +100,137 @@ VectorLayoutInterface NestedLayoutAttr::apply(AffineMap map) const {
...
+  SmallVector<int64_t> subgroupCount;
+  SmallVector<int64_t> batchCount;
+  SmallVector<int64_t> outerCount;
+  SmallVector<int64_t> threadCount;
+  SmallVector<int64_t> elementCount;
+  SmallVector<int64_t> subgroupStrides;
+  SmallVector<int64_t> threadStrides;
+  // To reshape, we need to ensure that we always go up in the layout level.
+  // We can only go to a lower level if this is a new dimension, and we have
+  // to start from the lowest level (element).
```
Groverkss: I think it will be useful to explain what does up/down mean. Maybe an example walkthrough would be useful (can generate one using claude or something)

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
Groverkss: That makes sense.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -98,6 +100,137 @@ VectorLayoutInterface NestedLayoutAttr::apply(AffineMap map) const {
...
+        }
+        --currDim;
+        currLevel = 0;
+        remainingLevels = llvm::to_vector(
+            llvm::reverse(getPackedShapeForUndistributedDim(currDim)));
+        // printShape("New rem levels", remainingLevels);
+        remainingStrides = {getThreadStrides()[currDim],
+                            getSubgroupStrides()[currDim]};
+        // printShape("New rem strides", remainingStrides);
+        unitDims = llvm::map_to_vector(remainingLevels,
```
Groverkss: Remove debug prints

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -98,6 +100,137 @@ VectorLayoutInterface NestedLayoutAttr::apply(AffineMap map) const {
...
+  //
+  // When merging a thread/subgroup level, we need to also ensure that the
+  // strides can be merged into a single stride.
+  SmallVector<int64_t> remainingLevels = {1, 1, 1, 1, 1};
+  SmallVector<int64_t> remainingStrides = {0, 0};
+  SmallVector<bool> unitDims = {true, true, true, true, true};
+  int64_t currDim = getRank();
+  int64_t currLevel = 5;
+  int64_t minLevel = 0;
+  for (int64_t dim : llvm::reverse(newShape)) {
```
Groverkss, line 142: I'd add a high level sketch of what you are trying to do above this loop and an example of how it should look like. Otherwise it's hard to understand what we are doing.

Groverkss (commented):
Mostly LGTM, I think the hardest part of this PR is just the reshaping function. Documenting that properly with what it's doing and examples, will help us in future.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -168,7 +243,11 @@ NestedLayoutAttr::reshape(ArrayRef<int64_t> newShape) const {
       }
 
       if (currLevel < minLevel) {
-        // Cannot go below the minimum level, bail out.
+        // Check that invariant that we're always moving upwards in the levels
+        // of the layout (see above for definition). Bail out if the invariant
+        // doesn't hold.
+        LLVM_DEBUG(llvm::dbgs() << "invariant violated, trying to move below "
```
kuhar: nit: can we switch to DebugLog.h instead?

kuhar (approved):
I don't see any issues but I haven't reviewed carefully. Please wait for others before merging.

Groverkss (approved):

PR #23308, opened 2026-01-28
[GlobalOptimization] Fix output_shape handling in SinkTransposeThroughExpandShape

qedawkins:
The SinkTransposeThroughExpandShape pattern was not properly setting the output_shape field when constructing the new tensor.expand_shape op. This caused issues when the expand_shape had dynamic dimensions, as the output_shape values need to be permuted along with the result type.


stellaraccident:
Thanks. Should fix https://github.com/stellaraccident/frank-models/issues/2

Max191 (approved):

stellaraccident:
Thanks. Works now!

PR #23309, opened 2026-01-28
[Codegen][GPU] Modify constraint on swizzle_hint ops to allow for multiple types of swizzles

Muzammiluddin-Syed-ECE:
At the tensor level, when you apply the same swizzle hint on a matmul transpose b with operands of the same shape CSE can cause the following pattern to emerge.

```
%extracted_slice_0 = tensor.extract_slice %global_load_operand_0
%extracted_slice_1 = tensor.extract_slice %global_load_operand_1
%empty = tensor.empty()
%swizzle = iree_codegen.swizzle_hint **XORShuffle<256, 32>** %empty
%copy_0 = linalg.copy {lowering_config = ...} ins(%extracted_slice_0) outs(%swizzle)
%pack_0 = linalg.pack %copy_0  {...}
%copy_1 = linalg.copy {...} ins(%extracted_slice_1) outs(%swizzle)
%pack_1 = linalg.pack %copy_1  {...}
%matmul = matmul_transpose_b(%pack_0, %pack_2) 
```

However, this only applies if the XOR shuffle is the same across both operands. The CSE happens differently when different shuffles apply to different operands. In this case we get the following pattern.

```
%extracted_slice_0 = tensor.extract_slice %global_load_operand_0
%extracted_slice_1 = tensor.extract_slice %global_load_operand_1
%empty = tensor.empty()
%swizzle_0 = iree_codegen.swizzle_hint **XORShuffle<256, 32>** %empty
%swizzle_1 = iree_codegen.swizzle_hint **XORShuffle<128, 16>** %empty
%copy_0 = linalg.copy {lowering_config = ...} ins(%extracted_slice_0) outs(%swizzle_0)
%pack_0 = linalg.pack %copy_0  {...}
%copy_1 = linalg.copy {...} ins(%extracted_slice_1) outs(%swizzle_1 )
%pack_1 = linalg.pack %copy_1  {...}
%matmul = matmul_transpose_b(%pack_0, %pack_2) 
```

This means that for any given tensor.empty op that is swizzled, we must allow it to have multiple swizzle hint op users (this PR) or remove CSE for tensor.empty ops (no way).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_fuse_and_hoist_forall.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_fuse_and_hoist_forall.mlir
@@ -1,4 +1,4 @@
-// RUN: iree-opt %s --pass-pipeline='builtin.module(func.func(iree-codegen-gpu-fuse-and-hoist-parallel-loops))' --split-input-file --verify-diagnostics | FileCheck %s
+// RUN: iree-opt /home/muzasyed/iree/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_fuse_and_hoist_forall.mlir --pass-pipeline='builtin.module(func.func(iree-codegen-gpu-fuse-and-hoist-parallel-loops))' --split-input-file --verify-diagnostics | FileCheck %s
 
 #translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse workgroup_size = [64, 1, 1] subgroup_size = 64>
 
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp
```diff
@@ -132,9 +132,13 @@ static FailureOr<Value> createSharedAllocDestination(RewriterBase &rewriter,
...
-           "swizzle hint as its only user");
+    bool onlySwizzle = llvm::all_of(
+        empty->getUsers(), llvm::IsaPred<IREE::Codegen::SwizzleHintOp>);
+    if (!onlySwizzle) {
+      return swizzleHintOp->emitOpError(
+          "a tensor.empty op with a swizzle hint applied, should have only "
+          "swizzle hint ops as its users");
+    }
     auto newSwizzle = IREE::Codegen::SwizzleHintOp::create(
         rewriter, loc, allocTensor.getResult(), swizzleHintOp.getSwizzle());
```
Max191: This is not necessarily the correct swizzle hint op. It is just the first user of the tensor.empty.

You should instead save the swizzleOp var that you get from the destination above, and use that here.

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp
```diff
@@ -132,9 +132,13 @@ static FailureOr<Value> createSharedAllocDestination(RewriterBase &rewriter,
   // of the `tensor.empty` op.
   if (auto swizzleHintOp =
           dyn_cast<IREE::Codegen::SwizzleHintOp>(*empty->getUsers().begin())) {
-    assert(empty->hasOneUse() &&
-           "a tensor.empty op with a swizzle hint applied, should have the "
-           "swizzle hint as its only user");
+    bool onlySwizzle = llvm::all_of(
+        empty->getUsers(), llvm::IsaPred<IREE::Codegen::SwizzleHintOp>);
+    if (!onlySwizzle) {
```
Max191: Do we need to have the only users be swizzle ops? This transformation creates an entirely new alloc_tensor, leaving the other users of the tensor.empty untouched. It shouldn't matter what the other users are because this fusion becomes independent of them once the new alloc_tensor is created.

Max191 (changes requested):
Can you add a test for this change?

Also, I'm not sure that we even need the other tensor.empty users to also be swizzle hint ops. Would be good to get @qedawkins eyes here since I haven't touched swizzle hints much.

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp
Muzammiluddin-Syed-ECE: These swizzle ops are created when the alloc is first created at GPUPromoteMatmulOperands so I dont think there is a case where you should have any other user except the swizzle ops, if an alloc is swizzled unless there is some weird CSE going on. But I take your point, i guess it isnt necessary here (but is necessary post bufferization which is why we have a check at ResolveSwizzleHintsPass).

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_promote_matmul_operands.mlir
```diff
@@ -392,3 +392,28 @@ func.func @swizzle_operand_no_promote_fill(%b: tensor<128x128xf32>) -> tensor<4x
...
+    ins(%a, %b : tensor<64x64xf32>, tensor<64x64xf32>) outs(%fill : tensor<64x64xf32>) -> tensor<64x64xf32>
+  return %mm : tensor<64x64xf32>
+}
+
+// SwizzleOperand attribute creates swizzle_hint op with xor_shuffle
+// and flattens/expands the tensor for shared memory swizzling.
+// CHECK-LABEL: func.func @promote_with_multiple_swizzle_operand
+//       CHECK:   %[[EMPTY_A:.+]] = tensor.empty() : tensor<4096xf32>
+//       CHECK:   %[[SWIZZLE_A:.+]] = iree_codegen.swizzle_hint %[[EMPTY_A]][#iree_codegen.xor_shuffle<128, 16>] : tensor<4096xf32>
+//       CHECK:   %[[SWIZZLE_B:.+]] = iree_codegen.swizzle_hint %[[EMPTY_A]][#iree_codegen.xor_shuffle<256, 32>] : tensor<4096xf32>
```
Max191: Wait, this test seems to just be testing the CSE?

This test isn't necessary, since it is not really relevant to the code change. Instead, please add a test to gpu_fuse_and_hoist_forall.mlir where there is an empty op with multiple swizzle hints.

Muzammiluddin-Syed-ECE: done!

kuhar (approved):

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_fuse_and_hoist_forall.mlir
```diff
@@ -961,3 +961,35 @@ func.func @no_swap_same_block_expand_slice(%arg0: tensor<64xf16>) -> tensor<4x4x
...
+// -----
+
+// Check that one tensor.empty can have multiple different swizzle hints.
+// CHECK-LABEL: func.func @multi_swizzle
+// CHECK: %[[EMPTY:.+]] = tensor.empty()
+// CHECK: %[[SWIZZLE1:.+]] = iree_codegen.swizzle_hint %[[EMPTY]][#iree_codegen.xor_shuffle<64, 8>]
+// CHECK: %[[SWIZZLE2:.+]] = iree_codegen.swizzle_hint %[[EMPTY]][#iree_codegen.xor_shuffle<64, 16>]
+// CHECK: scf.forall {{.*}} shared_outs(%{{.+}} = %[[SWIZZLE1]])
+// CHECK: scf.forall {{.*}} shared_outs(%{{.+}} = %[[SWIZZLE2]])
+func.func @multi_swizzle(%arg0: tensor<128x128xf16>, %arg1: tensor<128x128xf16>) -> (tensor<128x128xf16>, tensor<128x128xf16>) {
```
Max191: This test is still not testing the transform AFAICT. The createSharedAllocDestination function is called when we fuse multiple scf.forall loops together, but this test does not have any fusion.

I think what you want is to create a third scf.forall that has users of the first 2 scf.forall results inside its region. Then, the transforms should fuse both of the first 2 scf.forall loops into the third one, and create bufferization.alloc_tensor ops with the appropriate swizzles. Can you update the test and check that all 3 foralls get fused together and the alloc_tensors are created properly?

Max191 (approved):

qedawkins (approved):

PR #23311, opened 2026-01-28
[Encoding] Add `iree_encoding.dim` op and reification

jtuyls:
Implements phase 3 for adding specialization support on dynamic values for data-tiling: https://github.com/iree-org/iree/issues/22370. This adds the `iree_encoding.dim` operation and reification patterns to query encoding dimensions. For example:

```
// Encoding with dynamic M dimension (iteration_sizes = [?, 64, 128])
// encoding_dims has one element: the M value
%encoded = iree_encoding.set_encoding %input encoding_dims{%m}
    : tensor<?x128xf32> -> tensor<?x128xf32, #encoding>

// Query the first (and only) dynamic encoding dimension
%m_value = iree_encoding.dim %encoded[0] : tensor<?x128xf32, #encoding>
```

Assisted-by: Claude


compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
```diff
@@ -544,4 +544,82 @@ def IREEEncoding_PropagationOpInterface :
...
+
+def IREEEncoding_EncodingDimReificationInterface :
+  OpInterface<"EncodingDimReificationInterface"> {
+  let cppNamespace = "::mlir::iree_compiler::IREE::Encoding";
+  let description = [{
+    Interface for operations that can provide encoding dimension values.
+
+    This interface enables reification of `iree_encoding.encoding_dim` operations
+    by tracing through producer chains to find where encoding dimension values
+    were originally captured.
```
hanhanW, line 559:
```suggestion
    This interface enables reification of `iree_encoding.dim` operations by tracing
    through producer chains to find where encoding dimension values were
    originally captured.
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingOps.td
```diff
@@ -90,4 +95,72 @@ def IREEEncoding_UnsetEncodingOp : IREEEncoding_PureOp<"unset_encoding", [
...
+    - `encoding_dims` contains two values: M, K (in iteration order)
+    - index 0 returns M
+    - index 1 returns K
+
+    This operation is resolved through reification patterns that trace the
+    producer chain:
+    - `set_encoding` implements `EncodingDimReificationInterface` and returns
+      the corresponding `encoding_dims` value
+    - `tensor.cast` and DPS ops (like `linalg.fill`, `linalg.generic`) forward
+      the query to their source/init operands
```
hanhanW, line 128:
```suggestion
    - `set_encoding` implements `EncodingDimReificationInterface` and returns
      the corresponding `encoding_dims` value.
    - `tensor.cast` and DPS ops (like `linalg.fill`, `linalg.generic`) forward
      the query to their source/init operands.
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingOps.td
```diff
@@ -90,4 +95,72 @@ def IREEEncoding_UnsetEncodingOp : IREEEncoding_PureOp<"unset_encoding", [
...
+  }];
+
+  let arguments = (ins
+    AnyRankedTensor:$source,
+    IndexAttr:$index
+  );
+  let results = (outs Index:$result);
+
+  let assemblyFormat = [{
+    attr-dict $source `[` $index `]` `:` type($source)
```
hanhanW, line 150: Making attribute list after source and index seems more common?


```suggestion
    $source `[` $index `]` attr-dict `:` type($source)
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
```diff
@@ -544,4 +544,82 @@ def IREEEncoding_PropagationOpInterface :
...
+          Returns:
+          - Success with the value if the dimension can be resolved directly
+          - Failure if the operation cannot directly provide the value
+            (caller should use `getEncodingDimSource` to trace through)
+        }],
+        /*retTy=*/"::mlir::FailureOr<::mlir::Value>",
+        /*methodName=*/"reifyEncodingDim",
+        /*args=*/(ins
+          "::mlir::OpBuilder &":$builder,
+          "unsigned":$resultIndex,
```
hanhanW, line 591: Do we pass `OpResult` instead? It usually provides more information and passing it is not expensive as my understanding is that it is a pointer-like value.

I don't have a full picture about how you'd use the interface, so I'll leave the decision to you.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
```diff
@@ -544,4 +544,82 @@ def IREEEncoding_PropagationOpInterface :
...
+          dimension value directly if available at the given index.
+
+          For pass-through operations: Returns failure, indicating that the
+          pattern should use `getEncodingDimSource` to trace through to the
+          actual source.
+
+          Returns:
+          - Success with the value if the dimension can be resolved directly
+          - Failure if the operation cannot directly provide the value
+            (caller should use `getEncodingDimSource` to trace through)
```
hanhanW, line 585:
```suggestion
          - Success with the value if the dimension can be resolved directly.
          - Failure if the operation cannot directly provide the value.
            (caller should use `getEncodingDimSource` to trace through)
```

Do we need `FailureOr`? Do we just follow the other method that returns either a Value or null?

> caller should use `getEncodingDimSource` to trace through

Can you collaborate a bit more? Does it mean that caller uses wrong method if it returns failure?

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
```diff
@@ -544,4 +544,82 @@ def IREEEncoding_PropagationOpInterface :
...
+          Returns the source value whose encoding dimensions should be queried
+          for the given result index.
+
+          This is used by pass-through operations (like `tensor.cast`, DPS ops)
+          that don't originate encoding dimensions but forward them from another
+          value.
+
+          Returns:
+          - The source value to trace through (for pass-through operations)
+          - Null value if this operation is not a pass-through for encoding dims
```
hanhanW, line 610:
```suggestion
          - The source value to trace through (for pass-through operations).
          - Null value if this operation is not a pass-through for encoding dims.
```

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingPatterns.h
hanhanW, line 1: It is weird to see `*Patterns` under `IR/`. Are they only used by canonicalization patterns? If so, can you move them to `EncodingOps.cpp`?

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingPatterns.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+/// Resolves iree_encoding.dim using EncodingDimReificationInterface.
+///
+/// This pattern handles operations that implement the interface in two ways:
+/// 1. Operations that directly provide encoding dims (like set_encoding):
+///    The pattern calls reifyEncodingDim() and replaces with the result.
+/// 2. Operations that forward encoding dims from a source (like tensor.cast):
+///    The pattern calls getEncodingDimSource() and creates a new dim op on
+///    that source.
+///
```
hanhanW, line 31: I'd drop this blank comment.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingPatterns.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  // This pattern creates a new DimOp, so we need to bound recursion.
+  void initialize() { setHasBoundedRewriteRecursion(); }
+
+  LogicalResult matchAndRewrite(DimOp dimOp,
+                                PatternRewriter &rewriter) const override {
+    OpResult result = dyn_cast<OpResult>(dimOp.getSource());
+    if (!result) {
+      return failure();
+    }
```
hanhanW, line 102:
```suggestion
    auto result = dyn_cast<OpResult>(dimOp.getSource());
    if (!result) {
      return failure();
    }
```

Please also replace `return failure()` with more meaningful message if possible. I.e., `return rewriter.notifyMatchFailure(...)`. The error message is also a self-comment, which looks better to me.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingPatterns.cpp
```diff
@@ -0,0 +1,153 @@
...
+    // Verify encodings match.
+    auto resultType = dyn_cast<RankedTensorType>(result.getType());
+    auto initType = dyn_cast<RankedTensorType>(tiedInit->get().getType());
+    if (!resultType || !initType) {
+      return failure();
+    }
+
+    if (resultType.getEncoding() != initType.getEncoding()) {
+      return failure();
+    }
```
hanhanW, line 129: My intuition told me that it should be checked by the interface; yes, I confirmed it: https://github.com/llvm/llvm-project/blob/52dfcab327fe959074563603b6ebaaed314e9677/mlir/lib/Interfaces/DestinationStyleOpInterface.cpp#L51-L59

```cpp
  for (OpOperand *opOperand : outputTensorOperands) {
    OpResult result = dstStyleOp.getTiedOpResult(opOperand);
    if (result.getType() != opOperand->get().getType())
      return op->emitOpError("expected type of operand #")
             << opOperand->getOperandNumber() << " ("
             << opOperand->get().getType() << ")"
             << " to match type of corresponding result (" << result.getType()
             << ")";
  }
```

They have the same type, which indicates that they have the same encoding.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingInterfaces.td
hanhanW, line 1: Why don't we have a unified interface method? Then the next question is, why do we need the interface? Can't it just be a single method?

(I may be missing how it is used in other places. I haven't reached to that state yet.)

hanhanW (changes requested):
I feel that adding the interface may be over-design. What you're trying to do in this PR can just have a single canonicalization pattern? Below is an implementation example from Claude.

I think the main question is: why do we need this interface? In practice, the encodings are either come from SetEncoding op or frontend. Is it because you need such support for Flow::EncodeOp and Stream::EncodeOp as well?

```cpp
  struct ReifyEncodingDim : public OpRewritePattern<DimOp> {
    LogicalResult matchAndRewrite(DimOp dimOp,
      auto result = dyn_cast<OpResult>(dimOp.getSource());
      if (!result)
        return failure();

      Operation *producer = result.getOwner();
      int64_t dimIndex = dimOp.getConstantIndex();

      // Source: set_encoding directly provides encoding dims.
      if (auto setEnc = dyn_cast<SetEncodingOp>(producer)) {
        ValueRange encodingDims = setEnc.getEncodingDims();
        if (dimIndex >= encodingDims.size())
          return failure();
        rewriter.replaceOp(dimOp, encodingDims[dimIndex]);
        return success();
      }

      // Pass-through: tensor.cast forwards to source.
      if (auto castOp = dyn_cast<tensor::CastOp>(producer)) {
        rewriter.replaceOpWithNewOp<DimOp>(dimOp, castOp.getSource(),
                                           dimIndex);
        return success();
      }

      // Pass-through: DPS ops forward to tied init.
      if (auto dpsOp = dyn_cast<DestinationStyleOpInterface>(producer)) {
        if (auto *tiedInit = dpsOp.getTiedOpOperand(result)) {
          rewriter.replaceOpWithNewOp<DimOp>(dimOp, tiedInit->get(),
                                             dimIndex);
          return success();
        }
      }

      return failure();
    }
  };
```

btw, please expand the context in PR description. I was not aware of an interface is added until I review the code. Here is a good guidance: https://google.github.io/eng-practices/review/developer/cl-descriptions.html#informative

jtuyls:
> I feel that adding the interface may be over-design. What you're trying to do in this PR can just have a single canonicalization pattern?

Imo, the interface is appropriate as for the cost of a bit additional code we get:
- Separation of concerns: No op-specific special casing in the transformation pattern. New op support can be added through the the interface instead of new `if` blocks for each supported op (SetEncodingOp, tensor.cast, DPS ops, and possibly expand_shape, collapse_shape, extract_slice, etc.). So, we separate the transformation pattern from the op specific logic.
- External models / downstream extensibility: Projects building on IREE can extend to their custom ops by registering an external model.
- Explicit contract: The interface definition documents clearly what ops must provide.

Per my understanding, interfaces exist precisely to avoid encoding op-specific knowledge into transformation logic, and this keeps future additions isolated.

PR #23315, opened 2026-01-28
[CodeGen] Fix Attention GPU pipeline for shapes not requiring padding

keshavvinayak01:
Small attention shapes (e.g., `1×3×4` from `tests/e2e/linalg_ext_ops/attention.mlir`) failed to compile on GPU with error:

```Padding OnlineAttention without existing mask is not yet supported```

Added an `isPaddingNeeded()` check in `TensorMaskingOpInterface.cpp` that verifies whether padding is actually necessary before attempting it. When all dimensions are already multiples of their respective tile sizes, the pass skips padding entirely, allowing small, aligned shapes to compile successfully.

## Testing
Compile command: `iree-compile --iree-hal-target-device=hip --iree-hip-target=gfx1201 tests/e2e/linalg_ext_ops/attention.mlir -o /tmp/attention_gfx1201.vmfb`

Numerics check:
```
keshavvinayak@Shark23:~/iree$ iree-check-module --device=hip --module=/tmp/attention_gfx1201.vmfb 
[==========] Running 6 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 6 tests from module
[ RUN      ] module.attention1x3x4
[       OK ] module.attention1x3x4 (1545 ms)
[ RUN      ] module.causal_attention1x3x4
[       OK ] module.causal_attention1x3x4 (92 ms)
[ RUN      ] module.attention1x4x4_i1_mask_all_ones
[       OK ] module.attention1x4x4_i1_mask_all_ones (71 ms)
[ RUN      ] module.softcap_attention1x3x4
[       OK ] module.softcap_attention1x3x4 (75 ms)
[ RUN      ] module.attention1x4x4
[       OK ] module.attention1x4x4 (72 ms)
[ RUN      ] module.attention3x3x4
[       OK ] module.attention3x3x4 (71 ms)
[----------] 6 tests from module (1928 ms total)

[----------] Global test environment tear-down
[==========] 6 tests from 1 test suite ran. (1928 ms total)
[  PASSED  ] 6 tests.
``` 
Issue: https://github.com/iree-org/iree/issues/23305

### CODE REVIEW

File: compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp [hunk 1]
```diff
@@ -29,6 +29,34 @@ static bool isIterationCarriedArg(OpOperand &arg, linalg::GenericOp op) {
   return op.isDpsInit(inputOperand);
 }
 
+static bool isPaddingNeeded(OpBuilder &b, TilingInterface op,
+                            ArrayRef<OpFoldResult> padMultiples) {
```
hanhanW, line 39: Please add a doc.

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
```diff
@@ -29,6 +29,34 @@ static bool isIterationCarriedArg(OpOperand &arg, linalg::GenericOp op) {
...
+  SmallVector<Range> iterationDomain = op.getIterationDomain(b);
+
+  // Mismatched sizes.
+  if (iterationDomain.size() < padMultiples.size()) {
+    return true;
+  }
+
+  size_t numDims = std::min(iterationDomain.size(), padMultiples.size());
```
hanhanW: Mismatching sizes looks invalid to me. Do we need these checks? Or can they be an assrtion?

hanhanW (changes requested):
@keshavvinayak01 Please always add tests unless it is intended. We don't rely on local testing to claim a fix. If a test is not added, please explain why in PR description. I'd expect a lit test somewhere.

btw, if this does fix the issue, can you enable the test? You can move the file from exclude_list to the first list and run `bazel_to_cmake.py` script.

https://github.com/iree-org/iree/blob/8c014e5d907f7336f9c4aa86a1ec6c14a300a6e5/tests/e2e/linalg_ext_ops/BUILD.bazel#L112-L131

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
keshavvinayak01: If we return a failure here, then we won't generate the expected diagnostics for 

https://github.com/iree-org/iree/blob/fce95cd99184bb458d191edc2c0d1de7aa9940f7/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_apply_padding_online_attention.mlir#L12C1-L41C2

keshavvinayak01:
> @keshavvinayak01 Please always add tests unless it is intended. We don't rely on local testing to claim a fix. If a test is not added, please explain why in PR description. I'd expect a lit test somewhere.
> 
> btw, if this does fix the issue, can you enable the test? You can move the file from exclude_list to the first list and run `bazel_to_cmake.py` script.
> 
> https://github.com/iree-org/iree/blob/8c014e5d907f7336f9c4aa86a1ec6c14a300a6e5/tests/e2e/linalg_ext_ops/BUILD.bazel#L112-L131

I don't think we need to add any other lit tests. The file already contains the cases I'm handing. Removing `attention.mlir` from the exclude list is sufficient?

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
```diff
@@ -29,6 +29,40 @@ static bool isIterationCarriedArg(OpOperand &arg, linalg::GenericOp op) {
...
+///   - A dimension size is not evenly divisible by its corresponding multiple.
+/// A pad multiple of 0 indicates that no padding is required for that
+/// dimension.
+static bool isPaddingNeeded(OpBuilder &b, TilingInterface op,
+                            ArrayRef<OpFoldResult> padMultiples) {
+  SmallVector<Range> iterationDomain = op.getIterationDomain(b);
+
+  // Mismatched sizes. Return true (assume padding is needed) to allow later
+  // validation stages to emit proper diagnostics.
+  if (iterationDomain.size() < padMultiples.size()) {
```
MaheshRavishankar: This seems like this should be an assert on an error. If the iteration domain size is not the same as the pad multiples list then this analysis is tricky.

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
hanhanW: Doesn't it mean that the preset lowering config is wrong? There are five dimensions, and you only set four dimensions. If you don't want to tile the last dimension, just make it zero. Do I miss something?

hanhanW:
> > @keshavvinayak01 Please always add tests unless it is intended. We don't rely on local testing to claim a fix. If a test is not added, please explain why in PR description. I'd expect a lit test somewhere.
> > btw, if this does fix the issue, can you enable the test? You can move the file from exclude_list to the first list and run `bazel_to_cmake.py` script.
> > https://github.com/iree-org/iree/blob/8c014e5d907f7336f9c4aa86a1ec6c14a300a6e5/tests/e2e/linalg_ext_ops/BUILD.bazel#L112-L131
> 
> I don't think we need to add any other lit tests. The file already contains the cases I'm handing. Removing `attention.mlir` from the exclude list is sufficient?

Which file? `attention.mlir`? I still think we need unit test. You are updating some heuristic, and it is not reflected in unit test. We should not rely on integration tests to test such behavior. "Be able to compile and run" does not mean that your logic is fully tested. We could silently flip to other path and it still compiles in the future. Then, your change here becomes tech debts or dead code, right? No one can reason if we can delete the code or not; it will likely remain until we re-implement the thing. People won't know when can we delete the code.

> When all dimensions are already multiples of their respective tile sizes, the pass skips padding entirely, allowing small, aligned shapes to compile successfully.

I'd ask which pass? Is the logic exercised by the corresponding test? If so, you can explicitly mention it in your PR description.

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
keshavvinayak01: In that case, all the lit tests will fail in https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_apply_padding_online_attention.mlir#L1C1-L136C2, since the `iterationDomain` in the affine_map size is 5 but `padMultiplies` / lowering_config length is 4. That's wrong if I'm not mistaken, and it is what confused me to keeps this a soft failure instead of a compilation error.

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
keshavvinayak01: @hanhanW Yes, I don't know why all the lowering_configs in that file only have 4 dims, they're all wrong, should be 5. I assumed that is present for a reason and wrote code accordingly.

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
```diff
@@ -221,12 +251,22 @@ struct OnlineAttentionOpInterface final
...
+
+    if (needsPadding && !onlineAttentionOp.getMask()) {
       op->emitError(
           "Padding OnlineAttention without existing mask is not yet supported");
       return failure();
     }
 
+    if (!needsPadding) {
+      return failure();
+    }
```
hanhanW, line 266: Remove the blank lines because they belong the same code block? See https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

>
> Do not add blank lines where indentation already provides clear delineation, such as at the start or end of a code block. Do use blank lines to separate code into closely related chunks, analogous to paragraph breaks in prose. Within a statement or declaration, usually only insert line breaks to stay within the [line length limit](https://google.github.io/styleguide/cppguide.html#Line_Length), or to attach a comment to only part of the contents.

hanhanW (approved):

keshavvinayak01:
> > > @keshavvinayak01 Please always add tests unless it is intended. We don't rely on local testing to claim a fix. If a test is not added, please explain why in PR description. I'd expect a lit test somewhere.
> > > btw, if this does fix the issue, can you enable the test? You can move the file from exclude_list to the first list and run `bazel_to_cmake.py` script.
> > > https://github.com/iree-org/iree/blob/8c014e5d907f7336f9c4aa86a1ec6c14a300a6e5/tests/e2e/linalg_ext_ops/BUILD.bazel#L112-L131
> > 
> > 
> > I don't think we need to add any other lit tests. The file already contains the cases I'm handing. Removing `attention.mlir` from the exclude list is sufficient?
> 
> Which file? `attention.mlir`? I still think we need unit test. You are updating some heuristic, and it is not reflected in unit test. We should not rely on integration tests to test such behavior. "Be able to compile and run" does not mean that your logic is fully tested. We could silently flip to other path and it still compiles in the future. Then, your change here becomes tech debts or dead code, right? No one can reason if we can delete the code or not; it will likely remain until we re-implement the thing. People won't know when can we delete the code.
> 
> > When all dimensions are already multiples of their respective tile sizes, the pass skips padding entirely, allowing small, aligned shapes to compile successfully.
> 
> I'd ask which pass? Is the logic exercised by the corresponding test? If so, you can explicitly mention it in your PR description.

Added a lit test in addition to the iteration dims fix, it would've generated the `Padding OnlineAttention without existing mask is not yet supported` error before, but would sneak through now because padding isn't required.

PR #23319, opened 2026-01-29
[Codegen][NFC] Drop the workaround from EmulateNarrowType pass.

### LINKED ISSUE #20645: [Integrate] Upstream narrow type emulation is breaking iree test
### What happened?

 https://github.com/llvm/llvm-project/pull/133231 seems to be breaking our subbyte emulation tests. 

A simple repro is https://gist.github.com/raikonenfnu/1ea07b7e231d8997bfa1c29502df637d

original test https://github.com/iree-org/iree/blob/1a8d229431e62b50eb297e75ca4bf1dba3b67f65/tests/e2e/linalg/fp_to_subbyte.mlir

Based on the above test, currently IREE is generating code that has different trailing dim and "virtually"/fake non constant offset into the store `vector.store %0[%2]` where %2 is actually a constant,  but since it is inside a branch region, %c0 becomes %2 which is a region argument and is now a non-constant.

The condition above leads our program into this code path where it cannot determine `!foldedNumFrontPadElems` and fails this lowering. as seen in:
https://github.com/llvm/llvm-project/blob/2de936b6eb38e7a37224a97c2a22aa79b9dfb9dc/mlir/lib/Dialect/Vector/Transforms/VectorEmulateNarrowType.cpp#L619-L629

IIUC, that PR is somewhat correct, as long as the trailing dim does not match, we may need partial stores. 

I don't think I have enough context on the best way to solve this. We can:
1. Fold away that region since the condition to the regions are constants or somehow constant fold S.T we have constant offset in vector.store
2.  Add some code in vector dialect's EmulateNarrowType vector::StoreOp conversion to handle non constant cases through series of bitcasting and memref.generic_atomic_rmws. (this seems much harder and require some more thinking)

1. wget https://[gist.githubusercontent.com/raikonenfnu/1ea07b7e231d8997bfa1c29502df637d/raw/6e3311f6794dd5afeaeac0f82422cbe498d67272/emulate_failure.mlir](https://gist.githubusercontent.com/raikonenfnu/1ea07b7e231d8997bfa1c29502df637d/raw/6e3311f6794dd5afeaeac0f82422cbe498d67272/emulate_failure.mlir)

2. iree-opt --iree-codegen-emulate-narrow-type emulate_failure.mlir -o out.mlir

> emulate.mlir:18:3: error: failed to legalize operation 'vector.store' that was explicitly marked illegal

_No response_

### Version information

_No response_

### Additional context

_No response_

hanhanW:
https://github.com/llvm/llvm-project/commit/2de936b6eb38e7a37224a97c2a22aa79b9dfb9dc is a reasonable support, but it breaks IREE. The reason is that it is conversative about offsets, while IREE usually tiles the dimensions to be aligned with native vector size. In this context, the stores are always aligned store. The upstream has `assumeAligned` mode, so we drop the workaround.

Below is an example (from https://github.com/iree-org/iree/issues/20645) that shows the dynamic offset `%arg2` is always aligned with bytes.

```
func.func @main(%arg0: memref<8xf32>, %arg1: memref<8xi4>) {
  %c4 = arith.constant 4 : index
  %c8 = arith.constant 8 : index
  scf.for %arg2 = %c0 to %c8 step %c4 {
    %0 = vector.load %arg0[%arg2] : memref<8xf32>, vector<4xf32>
    %1 = arith.fptoui %0 : vector<4xf32> to vector<4xi32>
    %2 = arith.trunci %1 : vector<4xi32> to vector<4xi4>
    vector.store %2, %arg1[%arg2] : memref<8xi4>, vector<4xi4>
  }
  return
}
```

Closes https://github.com/iree-org/iree/issues/20645


hanhanW:
It requires https://github.com/llvm/llvm-project/pull/178565

compiler/src/iree/compiler/Codegen/Common/EmulateNarrowType.cpp
```diff
@@ -637,12 +163,12 @@ LogicalResult emulateNarrowType(
...
   memref::populateFlattenMemrefsPatterns(patterns);
 
-  patterns.insert<IREEConvertVectorStore>(ctx, /*disableAtomicRMW=*/false,
-                                          /*benefit=*/100);
   arith::populateArithNarrowTypeEmulationPatterns(typeConverter, patterns);
   memref::populateMemRefNarrowTypeEmulationPatterns(typeConverter, patterns);
   populateIREEResolveExtractStridedMetadataPatterns(patterns);
-  vector::populateVectorNarrowTypeEmulationPatterns(typeConverter, patterns);
+  vector::populateVectorNarrowTypeEmulationPatterns(typeConverter, patterns,
+                                                    /*disableAtomicRMW=*/false,
```
MaheshRavishankar, line 170: Are you sure you want to use AtomicRMW. That can be expensive? My understanding is that we should just vectorize to a size where we dont need atomics.

hanhanW: This is default option; we already use it. We can switch in a follow up.

hanhanW: I forgot to say that I agree with you and it should be happening on CPU side. We'll need the switch or expose the option.

I also have a local patch that makes VMVX happy for all the subtypes we have, and it doesn't use atomicRMW.

MaheshRavishankar: THis is common CPU/GPU path. We can tolerate some regression on CPU side, but regression on GPU side is more problematic. If there was a known correctness issue, that is one thing, but do we have a known correctness issue here on GPU?

hanhanW: I did not see any failure locally when I did the switch. I need to test more, but GPU test coverage is lower than I expected. I'm not aware of any correctness issue and I did not find any on Github issues. GPU does not work for other fp8 types, btw. See the table in https://github.com/iree-org/iree/pull/23238#issue-3840904362

| Type         | gfx908/90a | gfx942 | gfx950 | gfx11xx | gfx12xx |
|--------------|------------|--------|--------|---------|---------|
| f8E4M3FNUZ   |   emu      |   hw   |  emu   |   emu   |   emu   |
| f8E5M2FNUZ   |   emu      |   hw   |  emu   |   emu   |   emu   |
| f8E4M3FN     |   emu      |  emu   |   hw   |   emu   |   hw    |
| f8E5M2       |   emu      |  emu   |   hw   |   emu   |   hw    |
| f4E2M1FN     |   emu      |  emu   |   hw   |   emu   |   hw    |

`emu` means that it is not supported without the PR.

What I wanted to say is that it's been using AtomicRMW mode for a long time. I'll create a PR to do the experiment, let's move forward to drop the workaround?

I think the tolerance won't be impacted if the tiling config is correct. The heuristic should take it into account, like what I observed in https://github.com/iree-org/iree/issues/20645#issuecomment-2921182970. If not, it is a bug to me and we need to fix it. Or we expose the option and we always enable the flag on GPU.

hanhanW: sharktank tests are red because of azure issue, other tests look okay: https://github.com/iree-org/iree/pull/23344

MaheshRavishankar (approved):

Issue #23314, opened 2026-01-28
C API Integration for --iree-codegen-tuning-spec-path Flag


bangtianliu:
The tuning spec flags are currently defined as standalone LLVM command-line options in 
https://github.com/iree-org/iree/blob/d9aec69f563dafe1b70309d6a45570f9eb099906/compiler/src/iree/compiler/Codegen/Common/MaterializeTuningSpecsPass.cpp#L45-L49

## Request Description

The `--iree-codegen-tuning-spec-path` flag should be refactored into a structured options class to improve C API integration, so that we can use for [fusilli](https://github.com/iree-org/fusilli).

PR #23320, opened 2026-01-29
Add C API support for --iree-codegen-tuning-spec-path flag

bangtianliu:
This PR moves `--iree-codegen-tuning-spec-path` flag to `ROCMOptions`, which uses IREE's OptionsBinder so that C-API support can be enabled. Then the tuning spec path is converted as one pass option on `MaterializeTuningSpecsPass`

Previously, this flag was defined as a standalone `llvm::cl::opt<>` which prevented C API users from setting tuning specs programmatically.

I have only added C API support for `--iree-codegen-tuning-spec-path` . I can extend the same support to `iree-codegen-enable-default-tuning-specs` and `iree-codegen-dump-tuning-specs-to` if needed. 

Assisted-by: claude

Issue: #23314


MaheshRavishankar:
@IanWood1  you should probably add yourself as the CODEOWNER for these files.

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -1000,6 +1007,18 @@ bool Invocation::runPipeline(enum iree_compiler_pipeline_t pipeline) {
     }
   });
 
+  // Only pipelines that use MaterializeTuningSpecsPass need tuning spec
+  // options. STD and HAL_EXECUTABLE both run buildHALTransformPassPipeline
+  // which includes ConfigureExecutablesPass that adds
+  // MaterializeTuningSpecsPass.
+  bool needsTuningSpecs = (pipeline == IREE_COMPILER_PIPELINE_STD ||
+                           pipeline == IREE_COMPILER_PIPELINE_HAL_EXECUTABLE);
+  if (needsTuningSpecs) {
```
kuhar: naming nit: these don't strictly need tuning specs, should we instead say these *support* tuning specs?

compiler/src/iree/compiler/Codegen/Common/MaterializeTuningSpecsPass.cpp
```diff
@@ -63,6 +58,19 @@ llvm::cl::opt<std::string> clCodegenTuningSpecDumpDir(
...
 
+// Global accessor for tuning spec options.
+// This is set by the compiler driver before pipeline execution and cleared
+// after completion.
+static TuningSpecOptions *globalTuningSpecOptions = nullptr;
+
+static const TuningSpecOptions &getTuningSpecOptions() {
+  if (globalTuningSpecOptions) {
+    return *globalTuningSpecOptions;
+  }
```
kuhar: Relying on global variables seems like a major red flag. Is this even thread-safe? I'm thinking of having multiple compiler objects created by different threads, each setting its own options.

kuhar (changes requested):

IanWood1 (commented):
Would it be possible to make the file path a pass option for MaterializeTuningSpecsPass and then add a new flag to `ROCMOptions`:


https://github.com/iree-org/iree/blob/8db6cb5c33945937233c145d8d850c445494de37/compiler/plugins/target/ROCM/ROCMTarget.cpp#L79


We would have to rename `--iree-codegen-tuning-spec-path` to be backend specific, but I think that should be fine because tuning is only used by ROCM.

bangtianliu:
> Would it be possible to make the file path a pass option for MaterializeTuningSpecsPass and then add a new flag to `ROCMOptions`:
> 
> https://github.com/iree-org/iree/blob/8db6cb5c33945937233c145d8d850c445494de37/compiler/plugins/target/ROCM/ROCMTarget.cpp#L79
> 
> We would have to rename `--iree-codegen-tuning-spec-path` to be backend specific, but I think that should be fine because tuning is only used by ROCM.

Thanks for the insightful comment! This also addresses the thread-safety concern since we no longer rely on global variables. 
If there is a need to extend tuning support to other backends like SPIRV in the future (only concern in my mind), I can revisit and extend the support then.

compiler/src/iree/compiler/API/test/tuning_spec_flags_test.c
```diff
@@ -0,0 +1,36 @@
+// Copyright 2026 The IREE Authors
```
IanWood1: This test should probably be removed + it will fail if building without the rocm backend.

compiler/src/iree/compiler/API/Internal/BUILD.bazel
```diff
@@ -23,6 +23,7 @@ iree_compiler_cc_library(
     ],
     deps = [
         "//compiler/bindings/c:headers",
+        "//compiler/src/iree/compiler/Codegen/Common:PassHeaders",
```
IanWood1: This can be deleted

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -41,6 +41,7 @@
 #include <limits>
 
 #include "iree/compiler/API/Internal/Diagnostics.h"
+#include "iree/compiler/Codegen/Common/Passes.h"
```
IanWood1: Same here

compiler/src/iree/compiler/Codegen/Common/Options.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2026 The IREE Authors
```
IanWood1: This file can be deleted

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -1089,10 +1090,8 @@ bool Invocation::runTextualPassPipeline(const char *textPassPipeline) {
                                      llvm::errs()))) {
     return false;
   }
-  if (failed(passManager->run(parsedModule))) {
-    return false;
-  }
-  return true;
+
+  return succeeded(passManager->run(parsedModule));
```
IanWood1: This isn't a bad change, but given the current state of the PR it's a bit random. This PR shouldn't require any changes to `API/`.  I'm not sure if it should be kept or not.

IanWood1 (approved):
LGTM minus a few cleanups from the old implementation.

compiler/src/iree/compiler/API/test/tuning_spec_flags_test.c
bangtianliu: Then follow-up Q is that How can we add a test for this?

bangtianliu: Ok moved the test to `compiler/plugins/target/ROCM/test/`

bangtianliu:
cc @kuhar for another look

compiler/plugins/target/ROCM/test/tuning_spec_c_api_test.c
```diff
@@ -0,0 +1,35 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Test that verifies the tuning spec path flag can be set via C API.
+// This test is in the ROCM plugin directory because the flag is registered
+// by ROCMOptions and only available when ROCM is enabled.
+
+#include <stdio.h>
+#include "iree/compiler/embedding_api.h"
+
+int main(int argc, char **argv) {
```
kuhar: these seem unused
```suggestion
int main() {
```

compiler/plugins/target/ROCM/test/tuning_spec_c_api_test.c
```diff
@@ -0,0 +1,35 @@
...
+  ireeCompilerGlobalInitialize();
+  iree_compiler_session_t *session = ireeCompilerSessionCreate();
+
+  const char *flags[] = {"--iree-codegen-tuning-spec-path=/tmp/spec.mlir"};
+
+  iree_compiler_error_t *err = ireeCompilerSessionSetFlags(session, 1, flags);
+  if (err) {
```
kuhar: nit: there's only one flag, so no need for the array
```suggestion
  const char *flag = "--iree-codegen-tuning-spec-path=/tmp/spec.mlir";
  iree_compiler_error_t *err = ireeCompilerSessionSetFlags(session, 1, &flag);
  if (err) {
```

compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -93,6 +93,7 @@ struct ROCMOptions {
       IREE::Codegen::DenormalFpMath::None;
   bool enableRegSpillWarning = false;
   bool debugSymbols = false;
+  std::string tuningSpecPath = "";
```
kuhar: Maybe move it higher up closer to the other string options?

compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -183,6 +184,10 @@ struct ROCMOptions {
     binder.opt<bool>(
         "iree-hip-emit-debug-info", debugSymbols, cl::cat(category),
         cl::desc("Generate and embed debug information (DWARF) for HIP."));
+    binder.opt<std::string>(
```
kuhar: similar here

compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -183,6 +184,10 @@ struct ROCMOptions {
...
+        "iree-codegen-tuning-spec-path", tuningSpecPath, cl::cat(category),
+        cl::desc("File path to a module containing a tuning spec (transform "
```
kuhar:
```suggestion
        cl::desc("Path to a module containing a tuning spec (transform "
```

compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -183,6 +184,10 @@ struct ROCMOptions {
...
+                 "dialect library)."));
```
kuhar: Maybe it would be worth documenting if this needs to be a text file or if mlir bytecode is also accepted

compiler/src/iree/compiler/Codegen/Common/MaterializeTuningSpecsPass.cpp
```diff
@@ -175,6 +170,9 @@ serializeTuningSpecToAttr(ModuleOp tuningSpec) {
 
 struct MaterializeTuningSpecsPass final
     : impl::MaterializeTuningSpecsPassBase<MaterializeTuningSpecsPass> {
+  using Base = impl::MaterializeTuningSpecsPassBase<MaterializeTuningSpecsPass>;
+  using Base::Base;
```
kuhar, line 173: the `Base` type alias should be tablegened for us
```suggestion
struct MaterializeTuningSpecsPass final
: impl::MaterializeTuningSpecsPassBase<MaterializeTuningSpecsPass> {
  using Base::Base;
```

compiler/src/iree/compiler/Codegen/Common/MaterializeTuningSpecsPass.cpp
```diff
@@ -279,4 +278,5 @@ struct MaterializeTuningSpecsPass final
 };
 
 } // namespace
+
```
kuhar: nit: undo this?

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -793,6 +792,11 @@ def MaterializeTuningSpecsPass : Pass<"iree-codegen-materialize-tuning-specs", "
     This attribute is expected to be short-lived and removed by
     `iree-codegen-materialize-user-configs`.
   }];
+  let options = [
+    Option<"tuningSpecPath", "tuning-spec-path", "std::string",
+           /*default=*/"\"\"",
+           "File path to a module containing a tuning spec (transform dialect library).">,
```
kuhar: also here: explain if text/bytecode is accepted
```suggestion
           "Path to a module containing a tuning spec (transform dialect library).">,
```

kuhar (commented):
Do we also have to update https://iree.dev/reference/tuning/#usage-in-iree ?

compiler/plugins/target/ROCM/test/BUILD.bazel
```diff
@@ -4,13 +4,25 @@
...
 package(
     features = ["layering_check"],
     licenses = ["notice"],  # Apache 2.0
 )
 
+# C API test for tuning spec flag - only available when ROCM is enabled.
+iree_compiler_cc_test(
+    name = "tuning-spec-c-api-test",
+    testonly = True,
+    srcs = ["tuning_spec_c_api_test.c"],
```
kuhar: In the future we can add more c api tests specific to rocm, so I wouldn't make the name specific to the flag we are testing right now

compiler/plugins/target/ROCM/test/tuning_spec_c_api_test.c
```diff
@@ -0,0 +1,34 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Test that verifies the tuning spec path flag can be set via C API.
```
kuhar: AFAICT, this test is only compiled and executed, but nothing checks that the 'PASSED` message is printed, right?

Why aren't we using gtest here?  We can implement C api tests in C++, that's completely fine.

docs/website/docs/reference/tuning.md
```diff
@@ -253,9 +253,16 @@ flags:
 * `--iree-codegen-enable-default-tuning-specs` -- enables or disables the
   default tuning specs shipped with the compiler.
 * `--iree-codegen-tuning-spec-path` -- loads a user-specified tuning spec.
+  Accepts both MLIR text (`.mlir`) and bytecode (`.mlirbc`) formats.
 * `--iree-codegen-dump-tuning-specs-to` -- dumps final tuning specs to a
   directory or standard output.
 
+!!! note
+    The `--iree-codegen-tuning-spec-path` flag is part of `ROCMOptions` and
```
kuhar:
```suggestion
    The `--iree-codegen-tuning-spec-path` flag is a part of `ROCMOptions` and
```

docs/website/docs/reference/tuning.md
```diff
@@ -253,9 +253,16 @@ flags:
...
+    only available for ROCM/HIP backends. It can be set programmatically via
```
kuhar:
```suggestion
    only available for the ROCM/HIP backends. It can be set programmatically via
```

compiler/plugins/target/ROCM/test/tuning_spec_c_api_test.c
kuhar: For example, here you have unit tests for runtime code (IREE runtime is implemented in pure C): https://github.com/iree-org/iree/blob/main/runtime/src/iree/tokenizer/wordpiece_test.cc

compiler/plugins/target/ROCM/test/rocm_c_api_test.cc
```diff
@@ -0,0 +1,39 @@
...
+
+  void TearDown() override {
+    ireeCompilerSessionDestroy(session);
+    ireeCompilerGlobalShutdown();
+  }
+
+  iree_compiler_session_t *session = nullptr;
+};
+
+TEST_F(ROCMCAPITest, TuningSpecPathFlagAccepted) {
```
kuhar, line 28: Maybe worth adding a second testcase that uses some non-existing flag? Just to make sure we this API produces errors as expected. If there already other tests for this, maybe worth adding a comment that explains it's tested elsewhere.

compiler/plugins/target/ROCM/test/rocm_c_api_test.cc
```diff
@@ -0,0 +1,39 @@
...
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// C API tests for ROCM-specific flags. This test is in the ROCM plugin
+// directory because these flags are registered by ROCMOptions and only
+// available when ROCM is enabled.
+
+#include <gtest/gtest.h>
```
kuhar, line 11:
```suggestion
#include "iree/testing/gtest.h"
```

kuhar (approved):

compiler/plugins/target/ROCM/test/rocm_c_api_test.cc
bangtianliu: After applying this change, it causes bazel test error: https://github.com/iree-org/iree/actions/runs/21682118908/job/62519644825?pr=23320#step:5:1707

The `iree/testing/gtest.h` header is in the runtime directory (runtime/src/iree/testing/), so using it here would require a compiler to runtime dependency.

I checked other compiler unit tests like [compiler/src/iree/compiler/Utils/unittests/UtilsTest.cpp](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Utils/unittests/UtilsTest.cpp), [compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/unittests/UtilsTest.cpp](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/unittests/UtilsTest.cpp), they all use <gtest/gtest.h> directly.

so I will still use `#include <gtest/gtest.h>`

PR #23322, opened 2026-01-29
[Encoding] Add encoding_dims flow/stream conversion

jtuyls:
Implements phase 4 for adding specialization support on dynamic values for data-tiling: https://github.com/iree-org/iree/issues/22370. This carries the encoding dynamic values through Flow and Stream dialects.

Assisted-by: Claude

### DIFF
--- a/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
+++ b/compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
@@ -677,6 +677,9 @@ struct GPULayoutResolverAttr final
     return GPUEncodingResolverAttr::get(ctx, getPackedLayoutImpl(attr, type));
   }
 
+  // Selects a unified encoding from multiple candidates. Prefers static
+  // encodings (no dynamic encoding dims) over dynamic ones, since dynamic
+  // encoding dims cannot be tracked across function boundaries in initializers.
   // TODO: Improve the heuristic. This is a placeholder that demonstrates the
   // encoding unification is working in practice.
   Attribute getUnifiedEncoding(Attribute attr,
@@ -684,6 +687,21 @@ struct GPULayoutResolverAttr final
     if (encodings.empty()) {
       return nullptr;
     }
+
+    // Prefer static encodings (getNumDynamicEncodingDims() == 0).
+    for (Attribute encoding : encodings) {
+      if (auto serializableAttr =
+              dyn_cast<IREE::Encoding::SerializableAttr>(encoding)) {
+        std::optional<int64_t> numDynamicDims =
+            serializableAttr.getNumDynamicEncodingDims();
+        if (numDynamicDims && *numDynamicDims == 0) {
+          return encoding;
+        }
+      }
+    }
+
+    // No static encoding found, return the first one.
+    // The caller will verify and fall back to identity if needed.
     return encodings[0];
   }
 };
--- a/compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp
@@ -6,6 +6,7 @@
 
 #include "iree/compiler/Dialect/Flow/IR/FlowOps.h"
 
+#include "iree/compiler/Dialect/Encoding/IR/EncodingTypes.h"
 #include "iree/compiler/Dialect/TensorExt/IR/TensorExtOps.h"
 #include "iree/compiler/Dialect/TensorExt/IR/TensorExtTypes.h"
 #include "iree/compiler/Dialect/Util/IR/ClosureOpUtils.h"
@@ -1656,6 +1657,18 @@ LogicalResult TensorEncodeOp::verify() {
   if (failed(mlir::verifyCompatibleShape(operandType.getShape(),
                                          resultType.getShape()))) {
     return emitOpError("the operand shape and result shape are not compatible");
+  }
+  // Verify encoding_dims count matches what the encoding expects.
+  // Check both operand (for decode) and result (for encode) encodings.
+  for (Type type : {operandType, resultType}) {
+    std::optional<int64_t> expectedDims =
+        IREE::Encoding::getNumDynamicEncodingDims(type);
+    if (expectedDims.has_value() &&
+        static_cast<int64_t>(getEncodingDims().size()) != *expectedDims) {
+      return emitOpError() << "encoding expects " << *expectedDims
+                           << " encoding dim(s), but "
+                           << getEncodingDims().size() << " provided";
+    }
   }
   return success();
 }
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/MaterializeEncodings.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/MaterializeEncodings.cpp
@@ -66,6 +66,7 @@ static std::string getDispatchFuncName(IREE::Stream::TensorEncodeOp encodeOp) {
 ///   - source_binding
 ///   - dynamic dimension sizes of the source type
 ///   - dynamic dimension sizes of the destination type
+///   - encoding_dims (M, N, K for matmul encodings)
 ///   - destination binding
 static func::FuncOp createWorkgroupFunc(IREE::Stream::TensorEncodeOp encodeOp,
                                         StringRef functionName) {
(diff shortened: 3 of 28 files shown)


compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp
```diff
@@ -1657,6 +1658,22 @@ LogicalResult TensorEncodeOp::verify() {
...
+      std::optional<int64_t> expectedDims =
+          encodingAttr.getNumDynamicEncodingDims();
+      if (expectedDims.has_value() &&
+          static_cast<int64_t>(getEncodingDims().size()) != *expectedDims) {
+        return emitOpError() << "encoding expects " << *expectedDims
+                             << " encoding dim(s), but "
+                             << getEncodingDims().size() << " provided";
+      }
+    }
+  }
```
hanhanW, line 1672: Maybe we can create a method in `EncodingTypes.h` and reuse it for Encoding/Flow/Stream ops.

compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp
```diff
@@ -1657,6 +1658,22 @@ LogicalResult TensorEncodeOp::verify() {
...
     return emitOpError("the operand shape and result shape are not compatible");
   }
+  // Verify encoding_dims count matches what the encoding expects.
+  // Check both operand (for decode) and result (for encode) encodings.
+  for (RankedTensorType type : {operandType, resultType}) {
+    if (IREE::Encoding::SerializableAttr encodingAttr =
+            dyn_cast_or_null<IREE::Encoding::SerializableAttr>(
+                type.getEncoding())) {
+      std::optional<int64_t> expectedDims =
+          encodingAttr.getNumDynamicEncodingDims();
```
hanhanW: I've seen this pattern at least three times. Do we consider introducing a method in Encoding dialect? E.g., `IREE::Encoding::getNumDynamicEncodingDims(Type)`.

compiler/src/iree/compiler/Dialect/Stream/Transforms/MaterializeEncodings.cpp
```diff
@@ -250,25 +280,33 @@ replaceEncodeOpWithDispatchOp(RewriterBase &rewriter,
...
-    dynamicDims.push_back(argument);
+    workload.push_back(argument);
+  }
+  for (Value argument : encodeOp.getEncodingDims()) {
+    workload.push_back(argument);
   }
   rewriter.replaceOpWithNewOp<IREE::Stream::AsyncDispatchOp>(
       encodeOp, exportOp,
-      /*workload=*/dynamicDims, encodeOp.getResult().getType(), operands,
+      /*workload=*/workload, encodeOp.getResult().getType(), operands,
```
hanhanW: nit: remove the argument comment


```suggestion
      workload, encodeOp.getResult().getType(), operands,
```

compiler/src/iree/compiler/Dialect/Stream/Transforms/SpecializeEncodings.cpp
hanhanW, line 1: We need a test

compiler/src/iree/compiler/Dialect/Stream/Transforms/MaterializeEncodings.cpp
hanhanW, line 1: I think this is broken when an encoding op has encoding_dims while the other does not. (assuming the same encoding attributes are identical.)

The encodingSignature does not take encoding_dims into account.

https://github.com/iree-org/iree/blob/ab71aabfc875530b835ea750b09c960ccddfc029/compiler/src/iree/compiler/Dialect/Stream/Transforms/MaterializeEncodings.cpp#L302-L313

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
```diff
@@ -817,12 +817,12 @@ util.initializer {
 #map = affine_map<(m, n, k) -> (m, k)>
 #map1 = affine_map<(m, n, k) -> (k, n)>
 #map2 = affine_map<(m, n, k) -> (m, n)>
+// encoding1 has dynamic M, encoding2 has static M=4. The GPU resolver's
+// getUnifiedEncoding returns the first encoding, so both should be unified
+// to #encoding1 (with iteration_sizes = [?, 4096, 4096]).
```
hanhanW: It looks broken to me. How does encoding dimensions get propagated to other functions (outside initializer)?

I feel that we'll need to create new globals and store these values; perform updates for dispatch ops. However, the executables are already outlined, which makes this pass broken when the number of encoding dims do not match.

I don't have a solution atm. We probably can't do encoding unification if the number of encoding_dims is not the same.

Then the next design question pops out in my mind is that we may have to carry all the values (including c128) in this case. I need to think more about it. Thanks for raising the PR and providing the examples.

hanhanW (changes requested):
TLDR: I think it will break encoding unification; I need to re-think about the design. See inline comments for a little more details.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
jtuyls: > It looks broken to me. How does encoding dimensions get propagated to other functions (outside initializer)?

Do we really need that? This is before SpecializeEncodings, so we have unserialized encodings at this point and we don't  update the internals of the executables right? After SpecializeEncodings, we have serialized encodings and shouldn't need the encoding dims anymore? Maybe I am missing something though and this may need a higher bw discussion?

hanhanW: I had such question when I saw the result of MaterializeEncoding pass:

```
stream.async.dispatch on(#{{.+}}) @[[$EX]]::@[[$ENTRY]][%[[D0]], %[[D1]], %[[D0]], %[[D1]], %[[M]], %[[N]], %[[K]]]
```

I forgot that we'll need to pass in the encoding dims to dispatch op. If the unified encoding does not carry the same encoding dims, the dispatch op is broken, right?

In this example, the dispatch op that uses `#encoding1` takes `%M` and the dispatch op that uses `#encoding2` takes zero encoding dims. If the result of encoding unification is `#encoding1`, the latter dispatch op has to be updated with taking one more encoding dimension. Otherwise, you're not able to do your specialization later on.

I may be missing something. Your specialization (about layouts) is exercised inside the executable, no?

jtuyls: > In this example, the dispatch op that uses `#encoding1` takes `%M` and the dispatch op that uses `#encoding2` takes zero encoding dims. If the result of encoding unification is `#encoding1`, the latter dispatch op has to be updated with taking one more encoding dimension. Otherwise, you're not able to do your specialization later on.

I see, you mean that this would be broken when we want to specialize on these encoding dimension (not this PR)? Yes, in that case, we would need to have access to encoding dimensions in the way it's currently set up as the actual specialization happens later. We could solve that issue though by adding the specialization values to all the dispatches (that could be encoding-specific logic at the point where we insert the specialization values/markers)? It's not really about the encoding dims, but the specialization values, in that case. These could be the same, but not necessarily (specialization values could be a subset for example).

compiler/src/iree/compiler/Dialect/Stream/Transforms/SpecializeEncodings.cpp
jtuyls: Added it.

compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp
jtuyls: Yes, good idea, I created a utility for this now.

compiler/src/iree/compiler/Dialect/Stream/Transforms/MaterializeEncodings.cpp
jtuyls: Removed it.

compiler/src/iree/compiler/Dialect/Stream/Transforms/test/unify_encoding_for_globals.mlir
jtuyls: As discussed offline, we can now do these unifications:
- Static + static -> static
- Dynamic + static -> static
- Dynamic + dynamic -> static
- Dynamic + dynamic -> dynamic, only if the dynamic values are the same.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
```diff
@@ -677,13 +677,31 @@ struct GPULayoutResolverAttr final
...
+    // Prefer static encodings (getNumDynamicEncodingDims() == 0).
+    for (Attribute encoding : encodings) {
+      if (auto serializableAttr =
+              dyn_cast<IREE::Encoding::SerializableAttr>(encoding)) {
+        std::optional<int64_t> numDynamicDims =
+            serializableAttr.getNumDynamicEncodingDims();
+        if (numDynamicDims && *numDynamicDims == 0) {
+          return encoding;
+        }
+      }
```
hanhanW: Just a random idea when I saw the code: I'm thinking if we create a static method to `SerializableAttr`, so we will be able to use `IREE::Encoding::SerializableAttr::getNumDynamicEncodingDims(Attribute)` here.

PR #23332, opened 2026-01-30
[Codegen][PCF] Add convertForallToGenericNest transform

qedawkins:
Add a transform that converts scf.forall operations into multi-level pcf.generic nests. This is required in cases where a single scf.forall mapping type needs to map to multiple scopes. The immediate case this arises from is converting thread mapped scf.forall ops to combined subgroup + lane scopes. In this case, we can't simply convert to pcf.loop because of the way that automatic redistribution of pcf.loop's lowering works. We need to redistribute the iterations of the scf.forall to *all* workers across both scopes. This means that if we make one of the two scopes (subgroup or lane) a pcf.loop this fails to create the correct IR structure. If we make it loop over subgroups, this fails to predicate the lanes, and if we make it loop over lanes, we end up with more thread divergence than normal (normally only a single subgroup may exhibit divergence).

The transform creates an outer loop over workers and an inner scf.forall over the per-worker iteration range. For multi-dimensional cases, affine.linearize_index and affine.delinearize_index are used to flatten/unflatten indices appropriately.

Additionally adds a new method getNativeNumProcessorIds to ScopeAttrInterface. This is needed to query for the number of ids to generate.


qedawkins:
Gentle ping for review

hanhanW:
> Gentle ping for review

I'll get to this after I finish the implementation for dispatch_config chain, i.e, likely later today, sorry for the delay. (It's also a good time for me to learn these new things!)

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
```diff
@@ -0,0 +1,662 @@
...
+  if (scopes.empty()) {
+    return forallOp.emitError("at least one scope required");
+  }
+
+  if (failed(matchForallConversion(forallOp))) {
+    return failure();
+  }
+
+  return convertForallToGenericNestImpl(rewriter, forallOp, scopes);
+}
```
hanhanW, line 701: The comments are redundant to me because the function name already documents it.

EDIT: actually most of block comments are redundant to me, e.g.,

```
// scf.forall -> pcf.generic nest Pass
```

and

```
// scf.forall -> pcf.generic nest Implementation

static PCF::GenericOp
convertForallToGenericNestImpl(RewriterBase &rewriter, scf::ForallOp forallOp,
                               ArrayRef<PCF::ScopeAttrInterface> scopes) {
```

and so on

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
```diff
@@ -0,0 +1,662 @@
...
+        arith::MulIOp::create(rewriter, loc, totalWorkers, allCounts[i]);
+  }
+
+  // Compute total iteration count from forall bounds.
+  SmallVector<OpFoldResult> mixedUbs = forallOp.getMixedUpperBound();
+  Value totalIters = arith::ConstantIndexOp::create(rewriter, loc, 1);
+  for (OpFoldResult ub : mixedUbs) {
+    Value dim = getValueOrCreateConstantIndexOp(rewriter, loc, ub);
+    totalIters = arith::MulIOp::create(rewriter, loc, totalIters, dim);
+  }
```
hanhanW: Should we take `lbs` and `steps` into account like the other function?

```
  for (auto [ub, lb, step] : llvm::zip_equal(mixedUbs, mixedLbs, mixedStep)) {
    OpFoldResult iterCount = affine::makeComposedFoldedAffineApply(
        rewriter, loc, numIters, ArrayRef<OpFoldResult>{ub, lb, step});
    iterationCounts.push_back(
        getValueOrCreateConstantIndexOp(rewriter, loc, iterCount));
  }
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
```diff
@@ -0,0 +1,662 @@
...
+//===----------------------------------------------------------------------===//
+// scf.forall -> pcf.loop Implementation
+//===----------------------------------------------------------------------===//
+
+static PCF::LoopOp convertForallToPCFLoopImpl(RewriterBase &rewriter,
+                                              scf::ForallOp forallOp,
+                                              PCF::ScopeAttrInterface scope,
+                                              int64_t numIds) {
+  assert(succeeded(matchForallConversion(forallOp)) &&
+         "converting unsupported forall op");
```
hanhanW: I'd drop the assertion because it is local function and it is already checked in public functions. Having duplicated checks in assertion could lead to slow iree-compile/iree-opt in debug build.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
```diff
@@ -0,0 +1,662 @@
...
+  if (!delinearizationBasis.empty()) {
+    AffineExpr mul = s0 * s1;
+    Value total = delinearizationBasis.front();
+    total = std::accumulate(
+        delinearizationBasis.begin() + 1, delinearizationBasis.end(), total,
+        [&](Value l, Value r) {
+          OpFoldResult acc =
+              affine::makeComposedFoldedAffineApply(rewriter, loc, mul, {l, r});
+          return getValueOrCreateConstantIndexOp(rewriter, loc, acc);
+        });
```
hanhanW, line 243: I'd just use loop which looks easier to me. 

```
  Value total = delinearizationBasis.front();
  for (Value dim : llvm::drop_begin(delinearizationBasis)) {
    total = arith::MulIOp::create(rewriter, loc, total, dim); // or using affine op.
  }
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
```diff
@@ -0,0 +1,662 @@
...
+  Attribute syncScope = PCF::SyncOnReturnAttr::get(rewriter.getContext());
+  for (BlockArgument regionRefArg : loopOp.getRegionRefArgs()) {
+    auto srefType = cast<PCF::ShapedRefType>(regionRefArg.getType());
+    auto newSrefType = PCF::ShapedRefType::get(
+        rewriter.getContext(), srefType.getShape(), srefType.getElementType(),
+        srefType.getScope(), syncScope);
+    regionRefArg.setType(newSrefType);
+  }
+
+  rewriter.setInsertionPoint(terminator);
```
hanhanW: It's better to have `InsertionGuard` when you move insertion point in functions. Otherwise, the callers need to track the insertion point themselves, which is painful.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
```diff
@@ -0,0 +1,662 @@
...
+  rewriter.replaceOpWithNewOp<PCF::ReturnOp>(terminator);
+
+  // Use the inits as the replacements for the shared outs bbargs to appease
+  // `inlineBlockBefore`. By this point all of their users have been replaced
+  // or erased so it doesn't matter what goes here.
+  argReplacements.append(inits.begin(), inits.end());
+  rewriter.inlineBlockBefore(forallOp.getBody(), loopOp.getBody(),
+                             loopOp.getBody()->end(), argReplacements);
+
+  rewriter.replaceOp(forallOp, loopOp);
```
hanhanW: Should we be consistent between two functions? This one performs op replacement but the other does not. I'm +1 on letting callers to do the replacement, so users can control transformations better.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
```diff
@@ -0,0 +1,662 @@
...
+  // Unspecified mappings indicate sequential foralls which we can choose the
+  // iteration order for.
+  if (!mappingAttr) {
+    return llvm::to_vector(llvm::reverse(llvm::seq(forallOp.getRank())));
+  }
+  // Empty mappings are unsupported at the moment. It's unclear when a forall
+  // with an empty mapping would be useful or important.
+  if (mappingAttr.value().empty()) {
+    return SmallVector<int64_t>{};
+  }
```
hanhanW, line 90: Should we return a failure instead?

compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFBase.td
```diff
@@ -207,6 +207,10 @@ def PCF_TestScopeAttr
       assert(false && "why are you here?");
       return {};
     }
+    int64_t getNativeNumProcessorIds() {
+      assert(false && "why are you here?");
```
hanhanW: Maybe we replace the error message with `TestScopeAttr is for structural tests only; querying hardware properties is not expected`.

hanhanW (commented):
Looks okay, here is the first round of review comments.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
qedawkins: I like them because it helps me locate where things are in the file quickly when scrolling through but I can pare them down a bit.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
qedawkins: This is also folding as it goes so I'm keeping the composed affine apply.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/ConvertForallToPCF.cpp
qedawkins: Done

hanhanW (approved):

PR #23334, opened 2026-01-30
[LinalgExt] Add mixed tensor-buffer semantics for map_scatter

qedawkins:
This commit allows for mixed tensor-buffer semantics of map_scatter. This is required to avoid materializing the full scattered result of the map_scatter when distributed.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
@@ -3220,3 +3220,24 @@ func.func @drop_fusion_barrier() -> memref<6xf32> {
 // CHECK-LABEL: func.func @drop_fusion_barrier
 // CHECK:         %[[ALLOC:.+]] = memref.alloc() : memref<6xf32>
 // CHECK:         return %[[ALLOC]]
+
+// -----
+
+// Test bufferization of map_scatter with mixed tensor-buffer semantics.
+// The tensor input should be bufferized while the memref output stays as-is.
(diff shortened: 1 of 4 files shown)


hanhanW (commented):
> This is required to avoid materializing the full scattered result of the map_scatter when distributed.

Can you ellaborate a bit more? When do we partially bufferize the op?

qedawkins:
> > This is required to avoid materializing the full scattered result of the map_scatter when distributed.
> 
> Can you ellaborate a bit more? When do we partially bufferize the op?

Currently never, but we will be soon. The main issue is that stuff like this is a race:
```
foo.my_parallel_loop -> !full_tensor_ty {
  %0 = map_scatter : !thread_local_tensor_ty -> !full_tensor_ty
  write %0
}
```
but the nature of scatters are that they *are* the write itself, so either map_scatter or a distributed version of it needs to exist to do one of:
a) Act as the terminator of an scf.forall
b) Accept memref types
c) Accept sref types

This is doing b) because it's the most straightforward to implement and works nicely with the most common path we have for map_scatter (immediately consumed by a store_to_buffer op so we can compose them directly).

Max191 (approved):
This is great that we are fixing this issue, but I'm curious how this partial bufferization will happen in the codegen pipeline? We have a couple of different paths where map_scatter gets introduced, so I'm wondering how it will all fit together.

qedawkins:
> This is great that we are fixing this issue, but I'm curious how this partial bufferization will happen in the codegen pipeline? We have a couple of different paths where map_scatter gets introduced, so I'm wondering how it will all fit together.

The easiest way is just composing `store_to_buffer` with `map_scatter`. If you scatter and then store the result to memory, you might as well just scatter straight into memory. In the future we'll have other ways to get there though.

PR #23335, opened 2026-01-30
[GPU] Become more honest about barrier semantics (and drop stale code)

krzysz00:
Recently, I added an upstream PR
( https://github.com/llvm/llvm-project/pull/177425 ) that adds the ability for gpu barriers to declare what address spaces they fence on in addition to allowing the __syncthreads() semantics.

This PR uses this support pervasively in IREE's GPU code generation, either by explicitly specifying the address spaces fenced by barrier ops we create or by inhereting it from address spaces on memrefs (which we know will b in shared memory).

We were **already** using such barriers on AMD GPU targets, where we had a rewrite pattern to unconditionally rewrite `gpu.barrie' to `amdgpu.lds_barrier`, which always only fences workgroup memory and doesn't synthronize global memory. This commit removes that rewrite, which allows us to be more explicit abuht the semantics we're using **and** will allow us to synchronize global memory accesses within a workgrop if we ever need to do so.

This PR also adds explanatory notes explaining the memory semantics we're expecting where needed.

While I'm here, this PR removes an unmanitained and stale copy of the gpu barrier elimination logic from the IREE transform ops and delegates to the upstream version instead.


krzysz00:
@qedawkins I'd like to note that the tests for value_barrier and the like aren't creating explicitly-private {as seems to be implied by the small size of the allocation) or explicitly-LDS memory and so I'm needing to treat those alloc_tensor results as "it could be in any address space". Do you have a sense of how we might better test realistic behavior?

qedawkins:
> @qedawkins I'd like to note that the tests for value_barrier and the like aren't creating explicitly-private {as seems to be implied by the small size of the allocation) or explicitly-LDS memory and so I'm needing to treat those alloc_tensor results as "it could be in any address space". Do you have a sense of how we might better test realistic behavior?

This change looks nice! Value barrier probably needs the same kind of address space now so that we can set it properly when we create it. Later on we won't know what it's for anymore like you're seeing and I can't think of a good way to get that information back.

krzysz00:
@qedawkins I'd've thought you can get it off the address spaces of the input memrefs, no?

qedawkins:
Yes and no. value_barrier is allowed to take vector types too to indicate we wait until that value is available. In those cases we don't know what memory space to wait on. In the memref cases for sure though.

krzysz00:
So re `value_barrier` - is that in this PR? A followup PR? A prerequisite PR?

And how will we get address space information at the tensor level?

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
```diff
@@ -49,6 +50,53 @@ getBuffers(RewriterBase &rewriter, const MutableOperandRange &operands,
...
+            })
+            .Case([&](IREE::HAL::DescriptorTypeAttr) {
+              addressSpaces.insert(globalSpace);
+              return true;
+            })
+            .Case([&](amdgpu::AddressSpaceAttr) {
+              addressSpaces.insert(globalSpace);
+              return true;
+            })
+            .Default([&](Attribute) { return false; });
```
kuhar:
```suggestion
            .Default(false);
```

compiler/src/iree/compiler/Codegen/Common/GPU/VectorReductionToGPU.cpp
```diff
@@ -194,7 +194,8 @@ struct WarpOpBarrier final : OpRewritePattern<gpu::WarpExecuteOnLane0Op> {
     }
 
     rewriter.setInsertionPointAfter(warpOp);
-    (void)gpu::BarrierOp::create(rewriter, barrierOp.getLoc());
+    (void)gpu::BarrierOp::create(rewriter, barrierOp.getLoc(),
```
kuhar: drop the void cast since none of the other calls do that?

kuhar (approved):

PR #23343, opened 2026-01-30
[DispatchCreation] Support split reduction on weight backward CNHW layout

yzhang93:
For channel first NCHW convolutions, the weight backward layout is CNHW. Enable split reduction for these layouts to improve the performance.


MaheshRavishankar (approved):
Nice!

PR #23344, opened 2026-01-30
[Codegen] Switch narrow type emulation to use nonAtomic operations.

hanhanW:
ci-extra: test_torch

Signed-off-by: hanhanW <hanhan0912@gmail.com>


hanhanW:
I talked to Mahesh, and we agree that we need to have a verification pass, if we want to disable atomic patterns. The idea is check the offsets after linearization. Ideally, the linearization should happen isolated from the emulation. Then the check can happen in between.

However, it requires more work. Thus, we can check it at the end of the pass for now.

PR #23350, opened 2026-02-02
[DRAFT][VectorExt] Implements new Gather/Scatter ops in VectorExt

keshavvinayak01:
Addresses https://github.com/iree-org/iree/issues/23409


keshavvinayak01:
Thinking of bifurcating the Tablegen/Verifiers and the Folders/Canonicalizations. @Groverkss

PR #23351, opened 2026-02-02
[Codegen][Common] Add InsertBatchDimForBatchlessConv pass for 2D Conv

Abhishek-Varma:
Most of it has been taken from : https://github.com/iree-org/iree/issues/21955#issuecomment-3807719144 (CC: @hanhanW )- refactored the logic, cleaned up and tested for the generic conv tests locally.

Context: 
Upstream MLIR's linalg::inferConvolutionDims and related APIs (matchConvolutionOpOfType, isaConvolutionOpInterface) expect convolutions to have a batch dimension. These APIs are used by DownscaleConv patterns and vectorization to recognize and optimize convolution operations.

However, IREE's dispatch formation pipeline strips unit dimensions (including N=1 batch dimensions) via fold-unit-extent-dims. So after generalization step, a conv_2d_nhwc_hwcf with N=1 becomes a 6-loop generic op instead of the expected 7-loop structure, causing the upstream APIs to fail pattern matching.

This pass restores the batch dimension for such "batchless" 2D convolutions by inserting tensor.expand_shape on inputs/outputs and tensor.collapse_shape on results. This allows the upstream convolution detection APIs to recognize the operation, enabling DownscaleConv and vectorization patterns to apply.

The reshape operations are propagated to dispatch boundaries and folded into dispatch tensor load/store operations, resulting in zero runtime cost.

Lit tests added for supported operations:
- Conv2DNhwcHwcf, Conv2DNchwFchw
- PoolingNhwcSum/Max/Min, PoolingNhwcMaxUnsigned/MinUnsigned
- PoolingNchwSum/Max
- DepthwiseConv2DNhwcHwc

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
@@ -4,22 +4,6 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-//===----------------------------------------------------------------------===//
-// InsertBatchDimForBatchlessConv Pass
-//===----------------------------------------------------------------------===//
-//
-// This pass detects convolution operations that have been generalized and
-// had their batch dimension (N=1) stripped by IREE codegen transformations.
-// It restores the batch dimension by inserting tensor.expand_shape and
-// tensor.collapse_shape operations, enabling upstream convolution-specific
-// patterns (like DownscaleConv and vectorization) to match and apply.
-//
-// The reshape operations are expected to be folded into dispatch tensor
-// load/store operations by the PropagateReshapesByExpansion pass, resulting
-// in zero runtime cost.
-//
-//===----------------------------------------------------------------------===//
-
 #include "iree/compiler/Codegen/Common/Passes.h"
 #include "iree/compiler/Codegen/Common/Transforms.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
@@ -108,13 +92,12 @@ static Value prependUnitDimToTensor(RewriterBase &rewriter, Location loc,
 
 /// Shifts all existing dimensions in an affine map by 1 and prepends d0.
 /// (d0, d1, ...) -> (...) becomes (d0, d1, d2, ...) -> (d0, ...)
-static AffineMap shiftAndPrependDimToMap(AffineMap oldMap, unsigned newNumDims,
-                                         MLIRContext *ctx) {
+static AffineMap shiftAndPrependDimToMap(AffineMap oldMap, MLIRContext *ctx) {
   AffineMap shifted = oldMap.shiftDims(1);
   SmallVector<AffineExpr> newResults;
   newResults.push_back(getAffineDimExpr(0, ctx)); // New leading dim.
   llvm::append_range(newResults, shifted.getResults());
-  return AffineMap::get(newNumDims, 0, newResults, ctx);
+  return AffineMap::get(oldMap.getNumDims() + 1, 0, newResults, ctx);
 }
 
 /// Inserts a unit batch dimension into a batchless convolution operation.
@@ -130,7 +113,6 @@ static linalg::GenericOp insertUnitBatchDimension(RewriterBase &rewriter,
                                                   linalg::GenericOp op) {
   Location loc = op.getLoc();
   MLIRContext *ctx = rewriter.getContext();
-  unsigned newNumLoops = op.getNumLoops() + 1;
 
   // Process all input operands.
   // - Operand 0 (image): expand tensor and prepend dim to map.
@@ -145,7 +127,7 @@ static linalg::GenericOp insertUnitBatchDimension(RewriterBase &rewriter,
     if (inputOperand->getOperandNumber() == 0) {
       // Image input: expand and prepend dim to map.
       newInputs.push_back(prependUnitDimToTensor(rewriter, loc, input));
-      newMaps.push_back(shiftAndPrependDimToMap(oldMap, newNumLoops, ctx));
+      newMaps.push_back(shiftAndPrependDimToMap(oldMap, ctx));
     } else {
       // Other inputs (filter, zero points): keep as-is, just shift map.
       newInputs.push_back(input);
@@ -158,8 +140,8 @@ static linalg::GenericOp insertUnitBatchDimension(RewriterBase &rewriter,
   Value output = outputOperand->get();
   auto outputType = cast<RankedTensorType>(output.getType());
   Value expandedOutput = prependUnitDimToTensor(rewriter, loc, output);
-  AffineMap newOutputMap = shiftAndPrependDimToMap(
-      op.getMatchingIndexingMap(outputOperand), newNumLoops, ctx);
+  AffineMap newOutputMap =
+      shiftAndPrependDimToMap(op.getMatchingIndexingMap(outputOperand), ctx);
   newMaps.push_back(newOutputMap);
 
   // New iterator types: prepend parallel (batch) to existing types.
--- a/compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
@@ -1,0 +1,278 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//===----------------------------------------------------------------------===//
+// InsertBatchDimForBatchlessConv Pass
+//===----------------------------------------------------------------------===//
+//
+// This pass detects 2D convolution operations that have been generalized and
+// had their batch dimension (N=1) stripped by IREE codegen transformations.
+// It restores the batch dimension by inserting tensor.expand_shape and
+// tensor.collapse_shape operations, enabling upstream convolution-specific
+// patterns (like DownscaleConv and vectorization) to match and apply.
+//
+// Note: This pass currently only handles 2D convolutions (conv_2d, pooling_2d,
+// depthwise_conv_2d).
+//
+// The reshape operations are expected to be folded into dispatch tensor
+// load/store operations by the PropagateReshapesByExpansion pass, resulting
+// in zero runtime cost.
+//
+//===----------------------------------------------------------------------===//
+
+#include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Common/Transforms.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Dialect/Linalg/Utils/Utils.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/AffineExpr.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-codegen-insert-batch-dim-for-batchless-conv"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_INSERTBATCHDIMFORBATCHLESSCONVPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+
+/// Detects if the given linalg op is a batch-less 2D convolution (a convolution
+/// where the batch dimension N=1 was stripped by IREE's unit dim folding).
+///
+/// Returns success if this is a batch-less 2D convolution that should be
+/// transformed, failure otherwise.
+static LogicalResult isBatchlessConv(linalg::LinalgOp op) {
+  // Must be a convolution operation (uses upstream MLIR's convolution
+  // detection)
+  if (!linalg::isaConvolutionOpInterface(op)) {
+    return failure();
+  }
+
+  auto dimsOr = linalg::inferConvolutionDims(op);
+  if (failed(dimsOr)) {
+    return failure();
+  }
+  linalg::ConvolutionDimensions dims = *dimsOr;
+
+  // Must be 2D spatial convolution
+  if (dims.outputImage.size() != 2 || dims.filterLoop.size() != 2) {
+    return failure();
+  }
+
+  // Detect operation kind based on channel dimensions
+  bool isRegularConv =
+      !dims.inputChannel.empty() && !dims.outputChannel.empty();
+  bool isDepthwise = !dims.depth.empty();
+  bool isPooling = dims.inputChannel.empty() && dims.outputChannel.empty() &&
+                   dims.depth.empty();
+
+  // Check if batch dimension is missing
+  bool isBatchless = false;
+
+  if (isRegularConv || isDepthwise) {
+    // For conv/depthwise: batch should be non-empty when present
+    isBatchless = dims.batch.empty();
+  } else if (isPooling) {
+    // For pooling: batch contains [N, C], without N it's just [C]
+    // So batch.size() == 1 means only channel, no real batch
+    isBatchless = (dims.batch.size() == 1);
+  }
+
+  if (!isBatchless) {
+    return failure();
+  }
+
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
+// Helper functions for batch dimension insertion
+//===----------------------------------------------------------------------===//
+
+/// Builds reassociation indices to prepend a unit batch dimension at position
+/// 0. For a tensor of rank R, produces: [[0, 1], [2], [3], ..., [R]]
+static SmallVector<ReassociationIndices> buildBatchReassociation(int64_t rank) {
+  SmallVector<ReassociationIndices> reassoc;
+  reassoc.push_back({0, 1}); // New batch dim groups with first existing dim
+  for (int64_t i = 1; i < rank; ++i) {
+    reassoc.push_back({i + 1});
+  }
+  return reassoc;
+}
+
+/// Expands a tensor by prepending a unit batch dimension at position 0.
+/// tensor<AxBxC> -> tensor<1xAxBxC>
+static Value expandWithBatchDim(PatternRewriter &rewriter, Location loc,
+                                Value tensor) {
+  auto tensorType = cast<RankedTensorType>(tensor.getType());
+  SmallVector<int64_t> newShape = {1};
+  llvm::append_range(newShape, tensorType.getShape());
+  auto newType = RankedTensorType::get(newShape, tensorType.getElementType());
+  auto reassoc = buildBatchReassociation(tensorType.getRank());
+  return tensor::ExpandShapeOp::create(rewriter, loc, newType, tensor, reassoc);
+}
+
+/// Transforms an indexing map to account for a new batch dimension.
+/// Shifts all existing dimensions by 1 and prepends d0 (batch).
+/// (d0, d1, ...) -> (...) becomes (d0, d1, d2, ...) -> (d0, ...)
+static AffineMap prependBatchDimToMap(AffineMap oldMap, unsigned newNumLoops,
+                                      MLIRContext *ctx) {
+  AffineMap shifted = oldMap.shiftDims(1);
+  SmallVector<AffineExpr> newResults;
+  newResults.push_back(getAffineDimExpr(0, ctx)); // batch dim
+  llvm::append_range(newResults, shifted.getResults());
+  return AffineMap::get(newNumLoops, 0, newResults, ctx);
+}
+
+//===----------------------------------------------------------------------===//
+// InsertBatchDimForConvPattern
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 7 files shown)


compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+//===----------------------------------------------------------------------===//
+// Pass definition
+//===----------------------------------------------------------------------===//
+
+struct InsertBatchDimForBatchlessConvPass final
+    : impl::InsertBatchDimForBatchlessConvPassBase<
+          InsertBatchDimForBatchlessConvPass> {
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<linalg::LinalgDialect, tensor::TensorDialect>();
+  }
```
hanhanW: delete them, because it is handled by tablegen.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+    MLIRContext *context = &getContext();
+
+    RewritePatternSet patterns(context);
+    patterns.add<InsertBatchDimForConvPattern>(context);
+
+    if (failed(applyPatternsGreedily(getOperation(), std::move(patterns)))) {
+      return signalPassFailure();
+    }
+
+    // Run reshape propagation to push the expand/collapse shapes to boundaries
```
hanhanW:
```suggestion
    // Run reshape propagation to push the expand/collapse shapes to boundaries.
```

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+      tensor::CollapseShapeOp::getCanonicalizationPatterns(reshapePatterns,
+                                                           context);
+      tensor::ExpandShapeOp::getCanonicalizationPatterns(reshapePatterns,
+                                                         context);
+
+      if (failed(applyPatternsGreedily(getOperation(),
+                                       std::move(reshapePatterns)))) {
+        return signalPassFailure();
+      }
+    }
```
hanhanW, line 265: I'd expect populateFoldReshapeOpsByCollapsingPatterns as well.

  so the actual dispatch was

```mlir
linalg.fill
linalg.generic // conv
linalg.generic // elem
```

after inserting the batch dimension:

```mlir
linalg.fill
tensor.expand_shape
linalg.generic // conv
tensor.collapse_shape
linalg.generic // elem
```

after reshape propagation:

```mlir
tensor.expand_shape
linalg.fill
linalg.generic // conv
linalg.generic // elem
tensor.collapse_shape
```

Can you add such test case?

btw, have you tried if we can drop the flag with this pass? I think this should be the final PR, right?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
hanhanW, line 1: Please add proper punctuation for comments. You miss periods for several comments.

https://google.github.io/styleguide/cppguide.html#Punctuation_Spelling_and_Grammar

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+#define GEN_PASS_DEF_INSERTBATCHDIMFORBATCHLESSCONVPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+
+/// Detects if the given linalg op is a batch-less 2D convolution (a convolution
+/// where the batch dimension N=1 was stripped by IREE's unit dim folding).
+///
+/// Returns success if this is a batch-less 2D convolution that should be
+/// transformed, failure otherwise.
```
hanhanW: Why not just return a boolean? I always prefer boolean and std::optional if builders/rewriters are not involved. It is cleaner, and it does not need to propagate failure messages to the builders.

(Then the comment can be `Returns true if ...`

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+  // detection)
+  if (!linalg::isaConvolutionOpInterface(op)) {
+    return failure();
+  }
+
+  auto dimsOr = linalg::inferConvolutionDims(op);
+  if (failed(dimsOr)) {
+    return failure();
+  }
+  linalg::ConvolutionDimensions dims = *dimsOr;
```
hanhanW: Spell out the type for `auto`, and I think naming it as `maybeConvDims` is better. 

optional: we don't need `dims`.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+  if (!isBatchless) {
+    return failure();
+  }
+
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
+// Helper functions for batch dimension insertion
+//===----------------------------------------------------------------------===//
```
hanhanW: The way you categorize methods make the above one along. Maybe we just delete this comment?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+/// tensor<AxBxC> -> tensor<1xAxBxC>
+static Value expandWithBatchDim(PatternRewriter &rewriter, Location loc,
+                                Value tensor) {
+  auto tensorType = cast<RankedTensorType>(tensor.getType());
+  SmallVector<int64_t> newShape = {1};
+  llvm::append_range(newShape, tensorType.getShape());
+  auto newType = RankedTensorType::get(newShape, tensorType.getElementType());
+  auto reassoc = buildBatchReassociation(tensorType.getRank());
+  return tensor::ExpandShapeOp::create(rewriter, loc, newType, tensor, reassoc);
+}
```
hanhanW: Use `RewriterBase`.

btw, I found a common issue when I reviewed your PRs recently: Function/method names should describe **what they do**, not **why the caller uses them**. So maybe you should rename these methods. If you looking at the implementation, there is no "batch" semantics at all. E.g., this is a method that inserts a unit dimension at position `0`. We should abstract functions with `what they are doing`, not `how you want to call it`.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+//===----------------------------------------------------------------------===//
+//
+// This pass detects 2D convolution operations that have been generalized and
+// had their batch dimension (N=1) stripped by IREE codegen transformations.
+// It restores the batch dimension by inserting tensor.expand_shape and
+// tensor.collapse_shape operations, enabling upstream convolution-specific
+// patterns (like DownscaleConv and vectorization) to match and apply.
+//
+// Note: This pass currently only handles 2D convolutions (conv_2d, pooling_2d,
+// depthwise_conv_2d).
```
hanhanW: Is it correct? I thought that the implementation is dimension-count-agnostic

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,278 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Pattern to insert batch dimension for batch-less convolutions.
+///
+/// This transforms a 6D linalg.generic (3 parallel + 3 reduction) representing
+/// a batch-less conv into a 7D linalg.generic (4 parallel + 3 reduction) with
+/// the batch dimension restored:
+///
+///   expand_shape(image_input) -> linalg.generic (with batch) ->
+///   collapse_shape(output)
```
hanhanW: Is it an example or the truth? I thought that the implementation is dimension-count-agnostic. Where is the 6-D check? I'm +1 on generalizing it, but not only stick it on 2d conv. Your current implementation looks okay to me.

hanhanW (commented):
Thanks for the patch!

Have you checked if this is the last needed PR? If so, can we enable it in CPU pipeline and discard the flag? (Or at least you can create a PR to demonstrate it. It gives us a better picture of the status.)

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: I've added the test case. But adding `populateFoldReshapeOpsByCollapsingPatterns` led to an infinite loop because of expansion vs collapsing.

Could you please take a look at the lit test?

NOTE: The structure of the output is :-
```
linalg.fill   (because the expanded shape got subsumed into tensor.empty)
tensor.expand_shape (for input)
linalg.generic // conv
linalg.generic // elem
tensor.collapse_shape
```

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: I understand. That makes sense. I've updated it.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: I did constraine it to just 2D Conv, but have now lifted up the constraint and added a 3D Conv lit test. Can you check now?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: Addressed in the [next thread](https://github.com/iree-org/iree/pull/23351#discussion_r2763780149).

Abhishek-Varma:
> Thanks for the patch!
> 
> Have you checked if this is the last needed PR? If so, can we enable it in CPU pipeline and discard the flag? (Or at least you can create a PR to demonstrate it. It gives us a better picture of the status.)

Hi @hanhanW - Yes, I've checked that with this pass we can enable generalization of linalg.conv in CPU pipeline. Prior to this CPUSelectLoweringStrategy pass was not assigning `CPUConvTileAndDecomposeExpert` - but now it does and I have even compared the vector codes and they match as reported [here](https://github.com/iree-org/iree/issues/21955#issuecomment-3804285976) :-

Comparison | Named Conv | Generic Conv
-- | -- | --
vector<8xf32> | 949 | 949
vector<16xf32> | 138 | 138
vector<8xf16> | 130 | 130
vector<128xf32> | 26 | 26
vector<128xf16> | 26 | 26
vector<16xf16> | 10 | 10

Can you point me what other test to add besides the `lowering-strategy` test to show the correctness (because individual PRs checked in for decomposition and vectorization prove same vector codes already) ? I can accordingly add to the PR that I've raised on top of current PR : https://github.com/iree-org/iree/pull/23377

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
hanhanW: I think we can assume (and bail out if it does not match) that there is only one convolution-like op. Because it is typically the root op and we seldom fuse two reduction ops into a single dispatch. (softmax is the exception.)

Then you can use `op->isBeforeInBlock(newConvOp)` in the control function. Example code from Claude:

```cpp
    // Bubble expand_shape upward: only for ops BEFORE the conv
    linalg::ControlFusionFn expandControlFn = [&](OpOperand *fusedOperand) {
      Operation *op = fusedOperand->getOwner();
      if (op->getBlock() != newConvOp->getBlock())
        return false;
      return op->isBeforeInBlock(newConvOp);
    };

    // Sink collapse_shape downward: only for ops AFTER the conv
    linalg::ControlFusionFn collapseControlFn = [&](OpOperand *fusedOperand) {
      Operation *op = fusedOperand->getOwner();
      if (op->getBlock() != newConvOp->getBlock())
        return false;
      return newConvOp->isBeforeInBlock(op);
    };
```

In this context, we can walk all the operation and call the function once. We dont need a pattern here because it is only applied once.

Abhishek-Varma: Hi @hanhanW - I've added populateFoldReshapeOpsByCollapsingPatterns. The structure of the output is still the same and I believe that's expected. Can you please take a look at it now?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,277 @@
...
+// had their batch dimension (N=1) stripped by IREE codegen transformations.
+// It restores the batch dimension by inserting tensor.expand_shape and
+// tensor.collapse_shape operations, enabling upstream convolution-specific
+// patterns (like DownscaleConv and vectorization) to match and apply.
+//
+// The reshape operations are expected to be folded into dispatch tensor
+// load/store operations by the PropagateReshapesByExpansion pass, resulting
+// in zero runtime cost.
+//
+//===----------------------------------------------------------------------===//
```
hanhanW: Honestly, I don't like the pass description in C++ file because they are mostly duplicated to the description in tablegen. The file comment can be dropped, IMO.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+/// Shifts all existing dimensions in an affine map by 1 and prepends d0.
+/// (d0, d1, ...) -> (...) becomes (d0, d1, d2, ...) -> (d0, ...)
+static AffineMap shiftAndPrependDimToMap(AffineMap oldMap, unsigned newNumDims,
+                                         MLIRContext *ctx) {
+  AffineMap shifted = oldMap.shiftDims(1);
+  SmallVector<AffineExpr> newResults;
+  newResults.push_back(getAffineDimExpr(0, ctx)); // New leading dim.
+  llvm::append_range(newResults, shifted.getResults());
+  return AffineMap::get(newNumDims, 0, newResults, ctx);
```
hanhanW: Do you need `newNumDims`? I think we can query `oldMap.getNumDims()` and plus `1`?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,277 @@
...
+      RewritePatternSet reshapePatterns(context);
+      populateReshapeToInterfaceTensorPatterns(reshapePatterns);
+
+      // Bubble expand_shape upward: only for ops BEFORE the conv.
+      linalg::ControlFusionFn expandControlFn = [&](OpOperand *fusedOperand) {
+        Operation *op = fusedOperand->getOwner();
+        if (op->getBlock() != newConvOp->getBlock()) {
+          return false;
+        }
+        return op != newConvOp.getOperation();
```
hanhanW: Should this be `return op->isBeforeInBlock(newConvOp);`?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
```diff
@@ -0,0 +1,277 @@
...
+      };
+
+      // Sink collapse_shape downward: only for ops AFTER the conv.
+      linalg::ControlFusionFn collapseControlFn = [&](OpOperand *fusedOperand) {
+        Operation *op = fusedOperand->getOwner();
+        if (op->getBlock() != newConvOp->getBlock()) {
+          return false;
+        }
+        return op->isBeforeInBlock(newConvOp);
```
hanhanW: Should it be `newConvOp->isBeforeInBlock(op)`?

I'm not sure if it introduces infinite loop because the folding can introduce new reshapes. If so, maybe we should just run them separately. The first round only bubbling up the expand shape ops if they are before conv. The second round only sinking down the collapse shape ops if they are after conv op.

Can you check what's happening now?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
hanhanW: What do you mean by "The structure of the output is still the same"?

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: Correct. Removed.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: So I tried setting the previous controlFn as `return op->isBeforeInBlock(newConvOp);` for `ExpansionPatterns` and `newConvOp->isBeforeInBlock(op);` for `CollapsingPatterns` - although it didn't lead to infinite loop but it didn't push the collapse shape beyond the elementwise.

I separated them into two phases but to no avail.

Keeping the controlFn's as what you suggested, instead of using `CollapsingPatterns` for the second phase if I use `ExpansionPatterns` - then that did seem to work.

Having said that, the current implementation does indeed achieve what we're looking for as is demonstrated via the lit tests.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: I meant that in the expected output as mentioned [here]() we should get :-
```
tensor.expand_shape     ------ 1 Input/Filter
linalg.fill             ------ 2 OUT
linalg.generic // conv
linalg.generic // elem
tensor.collapse_shape
```

But I get :-
```
linalg.fill             ------ 1 OUT
tensor.expand_shape     ------ 2 Input/Filter
linalg.generic // conv
linalg.generic // elem
tensor.collapse_shape
```

I believe the order in this case doesn't look wrong to me as both operands are mutually exclusive.

hanhanW: Ah, it looks okay. As long as they are propagated to the boundary, i.e., the source is the binding, then it is okay.

What you have looks good, I did not form it correctly in my snippet. (There should be three expand_shape, but I only wrote one.)

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
hanhanW: I may be over-indexing on this, since I was looking for a robust solution. My concern was also making future dispatch happy, e.g., there can be dequant op before the conv op.

However, as a temporally workaround, it is okay to be like this -- we don't need to spend much time on this as long as we are building good techniques; I believed that https://github.com/llvm/llvm-project/pull/180586 is one step towards our goal.

Thanks for trying it out. I think we need to make the above comment (i.e., the expand shape one) better. I'm confused that why `return op != newConvOp.getOperation();` does not bubble up the expand shape that creates in sinking collapse_shape patterns. Are we hitting the max number of iteration and luckily get the result?

> Keeping the controlFn's as what you suggested, instead of using CollapsingPatterns for the second phase if I use ExpansionPatterns - then that did seem to work.

If it seemed to work, what is the problem of the switch? I think the algorithm is simple:

- Bubble up all the expand_shape ops that are before the conv op. (Note: this phase only has expand_shape patterns.)
- Sinking down all the collapse_shape ops that are after the conv op. The collapse_shape ops that created in the previous won't be sinked down because they are before the conv op. (Note: this phase onlyu has collapse_shape patterns.)

hanhanW (commented):
Many thanks for having several commits, which makes the review easier! I only have a thing to clarify, and the PR mostly looks good to me.

compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
Abhishek-Varma: I've now implemented two phased bubbling up of reshape ops to the boundaries.

Can you check now?

hanhanW (approved):

MaheshRavishankar:
I just found this while compiling release notes. Great step forward here by plumbing everything through. Now that this works, a bit of an anti-pattern is the fact that this needs to be done at all. The transformations/recognizers should all handle the case where there is no batch dimension seemlesly. Maybe something to address in due course and remove this workaround.

hanhanW:
> I just found this while compiling release notes. Great step forward here by plumbing everything through. Now that this works, a bit of an anti-pattern is the fact that this needs to be done at all. The transformations/recognizers should all handle the case where there is no batch dimension seemlesly. Maybe something to address in due course and remove this workaround.

@Abhishek-Varma is working on removing the workaround, where the workstream listed in https://github.com/iree-org/iree/issues/21955#issuecomment-3807719144

This PR is part of reaching the milestone, and the remaining work has been started.

Issue #23242, opened 2026-01-22
[CI] Issues in the CI


amd-eochoalo:
## `EACCESS` issue on `test_mi_250`

* [First seen on 2026/01/22.](https://github.com/iree-org/iree/actions/runs/21224680272/job/61069079131)
```
Cleaning the repository
Warning: Unable to clean or reset the repository. The repository will be recreated instead.
Deleting the contents of '/groups/aig_sharks/actions-runner-iree/_work/iree/iree'
Error: File was unable to be removed Error: EACCES: permission denied, unlink '/groups/aig_sharks/actions-runner-iree/_work/iree/iree/.git/HEAD'
```

## `EACCESS` issue on `amdgpu_rocm_250_gfx90a`

* [First seen on 2026/01/22](https://github.com/iree-org/iree/actions/runs/21226874644/job/61076732258)
```
Cleaning the repository
Warning: Unable to clean or reset the repository. The repository will be recreated instead.
Deleting the contents of '/groups/aig_sharks/actions-runner-iree/_work/iree/iree'
Error: File was unable to be removed Error: EACCES: permission denied, unlink '/groups/aig_sharks/actions-runner-iree/_work/iree/iree/.git/HEAD'
```

## cts tests failing on `test_w7900`

* [Link to action](https://github.com/iree-org/iree/actions/runs/21246291514/job/61137040691)
```
The following tests FAILED:
	 71 - iree/hal/drivers/hip/cts/hip_stream_buffer_mapping_test (Subprocess aborted)
	 73 - iree/hal/drivers/hip/cts/hip_stream_command_buffer_dispatch_constants_test (Failed)
	 74 - iree/hal/drivers/hip/cts/hip_stream_command_buffer_fill_buffer_test (Subprocess aborted)
	 81 - iree/hal/drivers/hip/cts/hip_graph_buffer_mapping_test (Subprocess aborted)
	 83 - iree/hal/drivers/hip/cts/hip_graph_command_buffer_dispatch_constants_test (Failed)
	 84 - iree/hal/drivers/hip/cts/hip_graph_command_buffer_fill_buffer_test (Subprocess aborted)
	 91 - iree/hal/drivers/hip/cts/hip_multi_queue_stream_buffer_mapping_test (Subprocess aborted)
	 93 - iree/hal/drivers/hip/cts/hip_multi_queue_stream_command_buffer_dispatch_constants_test (Failed)
[...]

PR #23352, opened 2026-02-02
[CI] Remove MI250 tests

amd-eochoalo:
As discussed in a meeting, we can remove the MI250 tests if these tests run somewhere else. Partially addresses issue #23242

Assisted-by: Claude


.github/workflows/pkgci_test_amd_mi250.yml
```diff
@@ -1,71 +0,0 @@
...
-      BUILD_DIR: build-tests
-      VENV_DIR: ${{ github.workspace }}/.venv
-      GH_TOKEN: ${{ github.token }}
-      CC: clang
-      CXX: clang++
-      IREE_CPU_DISABLE: 1
-      IREE_VULKAN_DISABLE: 1
-      IREE_CUDA_ENABLE: 0
-      IREE_HIP_ENABLE: 1
-      IREE_HIP_TEST_TARGET_CHIP: "gfx90a"
```
ScottTodd, line 34: > As discussed in a meeting, we can remove the MI250 tests if these tests run somewhere else.

Is gfx90a coverage still being provided somewhere, or do you believe that coverage provided by gfx942, gfx950, gfx1100, and gfx1201 is sufficient for IREE these days?

amd-eochoalo: Yes, this is something that I asked Mahesh and I understood from our conversation that he was ok with the coverage provided by the other targets. If there was some miscommunication, I'd be happy if @MaheshRavishankar corrects me here.

MaheshRavishankar: I think so. I think gfx942, gfx1100 and gfx1201 run the same tests, and MI250 isnt adding too much value. @kuhar to confirm as well.

We have no control of the server that is using MI250 and we dont really have good setup for issues with that server. So this seems like a diminishing return setup for me.

ScottTodd: Makes sense to me. Thanks for the context.

ScottTodd (approved):

kuhar (approved):

PR #23354, opened 2026-02-02
[Codegen][GPU] Add multi-buffering support for gather_to_lds async copy mode

jerryyin:
This is a follow-up to https://github.com/llvm/llvm-project/pull/176941. Since upstream limitation on multi-buffering is addressed, I can then manually clone all lds subview like ops inside the K loop and start to multi-buffer un-ambiguously. Note, this PR does multi-buffering only. The new pipelining code that take advantage of multi-buffering in async copy mode will come in successive PRs. Brief summary of changes:
- Add detection for `gather_to_lds` async copy patterns vs. stream copy patterns (global read + shared write)
- Bail on mixed-mode loops where both patterns exist
- Skip multi-buffering when `gather_to_lds` is inside conditional regions (e.g., `scf.if`)
- Clone view-like ops (subview, etc.) into the loop body when needed for `memref::multiBuffer` to work
- Skip multi-buffering when view ops have uses outside the target loop

Added unit tests covering:
- Multi-buffering with multiple LDS allocations - typical GEMM K loop case with two operands
- Conditional `gather_to_lds` (no multi-buffering)
- Mixed async/stream copy (no pipelining)
- View op escape scenarios (no multi-buffering)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -18,6 +18,7 @@
 #include "mlir/Dialect/AMDGPU/IR/AMDGPUDialect.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/MemRef/Transforms/Transforms.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/SCF/Transforms/Transforms.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
@@ -27,6 +28,7 @@
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/IR/Visitors.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
+#include "mlir/Interfaces/ViewLikeInterface.h"
 
 #define DEBUG_TYPE "iree-codegen-llvmgpu-prefetch-shared-memory-copy"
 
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -690,8 +903,31 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
...
+    LDBG() << "Loop has both gather_to_lds and stream copy ops - "
+           << "skipping pipelining entirely";
+    return forOp;
+  }
+
+  if (hasAsyncCopy) {
+    LDBG() << "Async copy mode detected (gather_to_lds present)";
+
+    // Multi-buffer LDS allocations for double buffering
+    if (failed(multiBufferLDSAllocations(forOp, /*numBuffers=*/2))) {
```
lialan: nit: better avoid magic numbers

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -41,6 +43,217 @@ struct StageClassification {
 };
 } // namespace
 
+/// Checks if a loop contains gather_to_lds operations directly in the loop
+/// body. Returns false if any gather_to_lds is inside a nested region (e.g.,
+/// scf.if), as conditional async copies can't be reliably pipelined.
+static bool hasGatherToLDS(scf::ForOp forOp) {
```
lialan: we are finding gather to lds nodes which is immediate child to the for loop, so the name could be a little bit confusing. Suggest to rename the function to reduce confusion.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -41,6 +43,217 @@ struct StageClassification {
...
+      return forOp->isAncestor(use.getOwner());
+    });
+
+    // Add to mapping so dependent ops will use the cloned version
+    mapping.map(origResult, cloneResult);
+
+    opsToErase.push_back(op);
+  }
+
+  // Erase original ops (in reverse order to handle dependencies).
```
lialan, line 212: Here is what my copilot tells me:
```
  The ops are added to opsToErase in forward order (same as clone order), then erased in reverse. However, viewOps was already sorted
  topologically (dependencies first). Erasing in reverse of that order should work, but only if all uses were successfully replaced.
  Consider adding a safety check:

  for (auto it = opsToErase.rbegin(); it != opsToErase.rend(); ++it) {
    Operation *op = *it;
    if (op->use_empty()) {
      op->erase();
    } else {
      LDBG() << "Warning: could not erase op with remaining uses: " << *op;
    }
  }
```

lialan (approved):
LGTM % very small nits.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -41,6 +43,217 @@ struct StageClassification {
 };
 } // namespace
 
+/// Checks if a loop contains gather_to_lds operations directly in the loop
+/// body. Returns false if any gather_to_lds is inside a nested region (e.g.,
+/// scf.if), as conditional async copies can't be reliably pipelined.
+static bool hasGatherToLDS(scf::ForOp forOp) {
+  bool found = false;
+  forOp.getBody()->walk([&](amdgpu::GatherToLDSOp gatherOp) {
```
krzysz00: ... instead of doing this, just iterate over the operations of the for op's region

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -41,6 +43,217 @@ struct StageClassification {
...
+    });
+
+    // Add to mapping so dependent ops will use the cloned version
+    mapping.map(origResult, cloneResult);
+
+    opsToErase.push_back(op);
+  }
+
+  // Erase original ops (in reverse order to handle dependencies).
+  for (auto it = opsToErase.rbegin(); it != opsToErase.rend(); ++it) {
```
krzysz00: I'd use rewriter methods for the erase. Also, there's LLVM helpers for the resersed range. Also also consider if make_early_inc_range is useful here.

krzysz00 (approved):
Similar notes to Alan, makes sense overall

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Renamed hasGatherToLDS → hasDirectGatherToLDS to clarify it only detects top-level (immediate children) gather_to_lds ops. Also separated the nested detection into a new function hasNestedGatherToLDS for clarity.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Changed from hardcoded 2 to numStages. This ties the buffer count directly to the pipeline depth, which was what I was planning to do. Note: I've only tested 2 buffers offline which is the reason why I hard-coded 2 in the buffer field. Your review reminded me to just wire it through since it is low risky code anyways.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Added an assertion `assert(op->use_empty() && "expected all uses to be replaced")` rather than a conditional check with warning. The invariant should always hold because `collectViewOpsToClone()` validates that all uses are inside the loop and will be replaced. An assertion catches logic bugs immediately during development, which is preferable to silently continuing with potentially broken state.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Agreed! Changed to direct iteration, this is cleaner (Also, split out nested gather to lds op detection in a standalone function to bail out when that happens).

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Regarding rewriter.eraseOp(): This function is a static helper (cloneViewOpsInsideLoop) without access to a rewriter. We could thread a rewriter through, but since this is called from multiBufferLDSAllocations which doesn't have a rewriter either (In fact, the upstream multi-buffer call doesn't carry a rewriter too - which make sense since it'd just ended up rewriting the buffer ending in an infinite loop).

`make_early_inc_range` works for ops in place but not useful here since all ops are already in a vector.

PR #23365, opened 2026-02-03
[AMDGPU][LDS] Add in_bounds attribute to CoalescedGatherDMAOp for tensor.pad fusion

lialan:
## Summary
- Add `in_bounds` attribute to `CoalescedGatherDMAOp` to indicate which dimensions may have out-of-bounds reads
- Fuse `tensor.pad` into `coalesced_gather_dma` by tracing through to the original source tensor
- Enable coalesced DMA for unaligned matmuls that require padding

## Motivation

When matmul dimensions don't align with tile sizes, `tensor.pad` is inserted to pad operands. Previously, this caused `coalesced_gather_dma` to fail because:
1. The padded tensor becomes a private memory allocation during bufferization
2. `amdgpu.gather_to_lds` requires source to be `fat_raw_buffer` or global memory

This PR fuses `tensor.pad` into the DMA operation, allowing it to read directly from buffer memory. AMD hardware's OOB behavior (returns 0 for out-of-bounds reads with `boundsCheck=true`) provides implicit padding.

## Example

Before:
```
%extracted_slice_0 = tensor.extract_slice %6[%22, %19] [%17, %20] [1, 1] : tensor<65x121xf32> to tensor<?x?xf32>
%padded_3 = tensor.pad %extracted_slice_2 low[0, 0] high[%21, 0] {
      ^bb0(%arg4: index, %arg5: index):
        tensor.yield %cst : f32
} : tensor<?x64xf32> to tensor<4x64xf32>
%26 = tensor.empty() : tensor<4x64xf32>
%27 = linalg.copy {lowering_config = #iree_gpu.use_global_load_dma} ins(%padded_3 : tensor<4x64xf32>) outs(%26 : tensor<4x64xf32>) -> tensor<4x64xf32> 
```
We cannot handle such case because:
* When inferring address space of `tensor.pad`, it will always be assigned to `private`.
* global load DMA cannot handle `private` sources.

Notice that the semantics of the op is basically to extract a slice of an unaligned tensor to an aligned tensor. We can still do a DMA from the buffer pointer to utilize the address clamping, as long as the source is inferred to use fat raw pointer.

So the solution is to just fuse `tensor.pad` into `coalesced_gather_dma` when we are converting `linalg.copy` to `iree_gpu.coalesced_gather_dma`:
```
%extracted_slice_0 = tensor.extract_slice %6[%22, %19] [%17, %20] [1, 1] : tensor<65x121xf32> to tensor<?x?xf32>
%23 = tensor.empty() : tensor<32x4xf32>
%24 = scf.forall (%arg4, %arg5) = (0, 0) to (32, 4) step (32, 4) shared_outs(%arg6 = %23) -> (tensor<32x4xf32>) {
    %extracted_slice_7 = tensor.extract_slice %arg6[%arg4, %arg5] [32, 4] [1, 1] : tensor<32x4xf32> to tensor<32x4xf32>
    %30 = scf.forall (%arg7) in (64) shared_outs(%arg8 = %extracted_slice_7) -> (tensor<32x4xf32>) {
     ...
     %extracted_slice_9 = tensor.extract_slice %arg8[0, %arg7] [32, %33] [1, 1] : tensor<32x4xf32> to tensor<32x?xf32>
      ...
      scf.forall.in_parallel {
        iree_gpu.coalesced_gather_dma %extracted_slice_0 into %arg8 lane(%arg7) in_bounds [false, false] : tensor<?x?xf32>, tensor<32x4xf32>, index
      }
    } {mapping = [#iree_gpu.lane_id<0>]}
    scf.forall.in_parallel {
      tensor.parallel_insert_slice %30 into %arg6[%arg4, %arg5] [32, 4] [1, 1] : tensor<32x4xf32> into tensor<32x4xf32>
    }
} {mapping = [#gpu.warp<linear_dim_1>, #gpu.warp<linear_dim_0>]}
```

* The implicit fusing happens when we convert `linalg.copy` to `iree_gpu.coalesced_gather_dma` op (along with tiling).
* `iree_gpu.coalesced_gather_dma` now has attribute `in_bounds` which tells whether we are loading in bounds, so it now supports masking.
* make sure we source tensor is inferred to use buffer pointer.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -408,10 +408,11 @@ private:
           // Fix: when any non-outermost source index exceeds its dimension,
           // replace the outermost index with sourceShape[0] to force the
           // linearized offset past the buffer end → hardware returns 0.
-          if (inBoundsAttr) {
-            auto sourceType = cast<MemRefType>(source.getType());
+          auto sourceType = cast<MemRefType>(source.getType());
+          if (inBoundsAttr && hasAMDGPUFatRawBufferAddressSpace(sourceType)) {
             ArrayRef<int64_t> sourceShape = sourceType.getShape();
-            Value anyNonOutermostOOB;
+            Value anyNonOutermostOOB = arith::ConstantOp::create(
+                rewriter, loc, rewriter.getBoolAttr(false));
 
             for (int64_t dim = 1; dim < sourceType.getRank(); ++dim) {
               if (dim >= static_cast<int64_t>(inBoundsAttr->size())) {
@@ -435,26 +436,19 @@ private:
                                                   arith::CmpIPredicate::uge,
                                                   srcIndices[dim], dimSize);
 
-              if (anyNonOutermostOOB) {
-                anyNonOutermostOOB = arith::OrIOp::create(
-                    rewriter, loc, anyNonOutermostOOB, isOOB);
-              } else {
-                anyNonOutermostOOB = isOOB;
-              }
+              anyNonOutermostOOB = arith::OrIOp::create(
+                  rewriter, loc, anyNonOutermostOOB, isOOB);
             }
 
-            if (anyNonOutermostOOB) {
-              Value oobOuterIdx;
-              if (ShapedType::isDynamic(sourceShape[0])) {
-                oobOuterIdx = memref::DimOp::create(rewriter, loc, source, 0);
-              } else {
-                oobOuterIdx = arith::ConstantIndexOp::create(rewriter, loc,
-                                                             sourceShape[0]);
-              }
-              srcIndices[0] =
-                  arith::SelectOp::create(rewriter, loc, anyNonOutermostOOB,
-                                          oobOuterIdx, srcIndices[0]);
+            Value oobOuterIdx;
+            if (ShapedType::isDynamic(sourceShape[0])) {
+              oobOuterIdx = memref::DimOp::create(rewriter, loc, source, 0);
+            } else {
+              oobOuterIdx =
+                  arith::ConstantIndexOp::create(rewriter, loc, sourceShape[0]);
             }
+            srcIndices[0] = arith::SelectOp::create(
+                rewriter, loc, anyNonOutermostOOB, oobOuterIdx, srcIndices[0]);
           }
 
           amdgpu::GatherToLDSOp::create(rewriter, loc, source, srcIndices, dest,
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -288,6 +288,20 @@ struct LowerCoalescedGatherDMAPattern final
     }
     SmallVector<TransferSegment> segments = std::move(*segmentsOrFailure);
 
+    // OOB padding requires fat_raw_buffer for hardware OOB clamping.
+    if (std::optional<ArrayAttr> inBounds = dmaOp.getInBounds()) {
+      auto srcType = cast<MemRefType>(source.getType());
+      if (!hasAMDGPUFatRawBufferAddressSpace(srcType)) {
+        for (Attribute attr : *inBounds) {
+          if (!cast<BoolAttr>(attr).getValue()) {
+            dmaOp.emitOpError("in_bounds with OOB dimensions requires "
+                              "fat_raw_buffer address space on source");
+            return failure();
+          }
+        }
+      }
+    }
+
     // Set up for code generation.
     rewriter.setInsertionPoint(dmaOp);
     TypedValue<IndexType> laneId = dmaOp.getLane();
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -300,44 +300,6 @@ struct LowerCoalescedGatherDMAPattern final
           }
         }
       }
-
-      // For non-outermost dims with OOB (in_bounds=false), the vector read
-      // must not cross row boundaries. Each lane reads `elementsPerLane`
-      // contiguous elements from the source buffer. If the source dim size is
-      // not a multiple of elementsPerLane, a vector read near the end of a row
-      // will wrap into the next row instead of returning zeros.
-      // Example: source 64x62xf32, dest 64x64xf32, vector<4xf32>:
-      //   Lane at [0, 60] reads 4 elements at flat offsets 60..63.
-      //   Offset 62 wraps to [1, 0] instead of returning 0.
-      ArrayRef<int64_t> sourceShape = srcType.getShape();
-      for (int64_t dim = 1; dim < srcType.getRank(); ++dim) {
-        if (dim >= static_cast<int64_t>(inBounds->size())) {
-          break;
-        }
-        bool dimInBounds = cast<BoolAttr>((*inBounds)[dim]).getValue();
-        if (dimInBounds) {
-          continue;
-        }
-        // This non-outermost dim has padding. Check that source dim size is
-        // a multiple of elementsPerLane for every segment to prevent row
-        // crossing.
-        if (ShapedType::isDynamic(sourceShape[dim])) {
-          return rewriter.notifyMatchFailure(
-              dmaOp, "non-outermost OOB dim " + Twine(dim) +
-                         " has dynamic source size; cannot verify vector "
-                         "reads do not cross row boundaries");
-        }
-        for (const TransferSegment &segment : segments) {
-          if (sourceShape[dim] % segment.elementsPerLane != 0) {
-            return rewriter.notifyMatchFailure(
-                dmaOp, "non-outermost OOB dim " + Twine(dim) +
-                           " has source size " + Twine(sourceShape[dim]) +
-                           " not divisible by elementsPerLane " +
-                           Twine(segment.elementsPerLane) +
-                           "; vector reads would cross row boundaries");
-          }
-        }
-      }
     }
 
     // Set up for code generation.
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -300,6 +300,44 @@ struct LowerCoalescedGatherDMAPattern final
           }
         }
       }
+
+      // For non-outermost dims with OOB (in_bounds=false), the vector read
+      // must not cross row boundaries. Each lane reads `elementsPerLane`
+      // contiguous elements from the source buffer. If the source dim size is
+      // not a multiple of elementsPerLane, a vector read near the end of a row
+      // will wrap into the next row instead of returning zeros.
+      // Example: source 64x62xf32, dest 64x64xf32, vector<4xf32>:
+      //   Lane at [0, 60] reads 4 elements at flat offsets 60..63.
+      //   Offset 62 wraps to [1, 0] instead of returning 0.
+      ArrayRef<int64_t> sourceShape = srcType.getShape();
+      for (int64_t dim = 1; dim < srcType.getRank(); ++dim) {
+        if (dim >= static_cast<int64_t>(inBounds->size())) {
+          break;
+        }
+        bool dimInBounds = cast<BoolAttr>((*inBounds)[dim]).getValue();
+        if (dimInBounds) {
+          continue;
+        }
+        // This non-outermost dim has padding. Check that source dim size is
+        // a multiple of elementsPerLane for every segment to prevent row
+        // crossing.
+        if (ShapedType::isDynamic(sourceShape[dim])) {
+          return rewriter.notifyMatchFailure(
+              dmaOp, "non-outermost OOB dim " + Twine(dim) +
+                         " has dynamic source size; cannot verify vector "
+                         "reads do not cross row boundaries");
+        }
+        for (const TransferSegment &segment : segments) {
+          if (sourceShape[dim] % segment.elementsPerLane != 0) {
+            return rewriter.notifyMatchFailure(
+                dmaOp, "non-outermost OOB dim " + Twine(dim) +
+                           " has source size " + Twine(sourceShape[dim]) +
+                           " not divisible by elementsPerLane " +
+                           Twine(segment.elementsPerLane) +
+                           "; vector reads would cross row boundaries");
+          }
+        }
+      }
     }
 
     // Set up for code generation.
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -507,7 +507,11 @@ struct AMDGPULowerCoalescedDMAToGatherLDSPass final
     // ops to be successfully lowered. In the future, a fallback lowering path
     // (e.g., using global_load) could handle ops that don't match the pattern.
     WalkResult result = funcOp.walk([&](IREE::GPU::CoalescedGatherDMAOp op) {
-      op.emitOpError("failed to lower coalesced_gather_dma op");
+      op.emitOpError(
+          "failed to lower to gather_to_lds; possible causes: source "
+          "lacks fat_raw_buffer address space for OOB padding, destination "
+          "is not contiguous, or element sizes are incompatible with "
+          "dma_sizes");
       return WalkResult::interrupt();
     });
     if (result.wasInterrupted()) {
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -304,7 +304,8 @@ struct LowerCoalescedGatherDMAPattern final
     }
 
     emitTransfers(rewriter, loc, source, dest, destShape, numLinearDims,
-                  elementType, indices, segments, segmentLaneOffsets);
+                  elementType, indices, segments, segmentLaneOffsets,
+                  dmaOp.getInBounds());
 
     rewriter.eraseOp(dmaOp);
     return success();
@@ -337,7 +338,8 @@ private:
                      Value dest, ArrayRef<int64_t> destShape,
                      int64_t numLinearDims, Type elementType,
                      OperandRange indices, ArrayRef<TransferSegment> segments,
-                     ArrayRef<Value> segmentLaneOffsets) const {
+                     ArrayRef<Value> segmentLaneOffsets,
+                     std::optional<ArrayAttr> inBoundsAttr) const {
     int64_t destRank = destShape.size();
     int64_t numOuterDims = destRank - numLinearDims;
     LDBG() << "Emitting transfers: " << numOuterDims << " outer dims, "
@@ -399,6 +401,61 @@ private:
 
           auto [srcIndices, dstIndices] = generateGatherIndices(
               rewriter, loc, srcDimOffsets, dstDimOffsets, indices);
+
+          // Raw buffer OOB clamping is 1D (linear): it returns 0 only when the
+          // byte offset >= total buffer size. For non-outermost dimensions,
+          // an OOB index wraps into the next row instead of returning 0.
+          // Fix: when any non-outermost source index exceeds its dimension,
+          // replace the outermost index with sourceShape[0] to force the
+          // linearized offset past the buffer end → hardware returns 0.
+          if (inBoundsAttr) {
+            auto sourceType = cast<MemRefType>(source.getType());
+            ArrayRef<int64_t> sourceShape = sourceType.getShape();
+            Value anyNonOutermostOOB;
+
+            for (int64_t dim = 1; dim < sourceType.getRank(); ++dim) {
+              if (dim >= static_cast<int64_t>(inBoundsAttr->size())) {
+                break;
+              }
+              bool dimInBounds =
+                  cast<BoolAttr>((*inBoundsAttr)[dim]).getValue();
+              if (dimInBounds) {
+                continue;
+              }
+
+              Value dimSize;
+              if (ShapedType::isDynamic(sourceShape[dim])) {
+                dimSize = memref::DimOp::create(rewriter, loc, source, dim);
+              } else {
+                dimSize = arith::ConstantIndexOp::create(rewriter, loc,
+                                                         sourceShape[dim]);
+              }
+
+              Value isOOB = arith::CmpIOp::create(rewriter, loc,
+                                                  arith::CmpIPredicate::uge,
+                                                  srcIndices[dim], dimSize);
+
+              if (anyNonOutermostOOB) {
+                anyNonOutermostOOB = arith::OrIOp::create(
+                    rewriter, loc, anyNonOutermostOOB, isOOB);
+              } else {
+                anyNonOutermostOOB = isOOB;
+              }
+            }
+
+            if (anyNonOutermostOOB) {
+              Value oobOuterIdx;
+              if (ShapedType::isDynamic(sourceShape[0])) {
+                oobOuterIdx = memref::DimOp::create(rewriter, loc, source, 0);
+              } else {
+                oobOuterIdx = arith::ConstantIndexOp::create(rewriter, loc,
+                                                             sourceShape[0]);
+              }
+              srcIndices[0] =
+                  arith::SelectOp::create(rewriter, loc, anyNonOutermostOOB,
+                                          oobOuterIdx, srcIndices[0]);
+            }
+          }
 
           amdgpu::GatherToLDSOp::create(rewriter, loc, source, srcIndices, dest,
                                         dstIndices,
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -294,9 +294,9 @@ struct LowerCoalescedGatherDMAPattern final
       if (!hasAMDGPUFatRawBufferAddressSpace(srcType)) {
         for (Attribute attr : *inBounds) {
           if (!cast<BoolAttr>(attr).getValue()) {
-            dmaOp.emitOpError("in_bounds with OOB dimensions requires "
-                              "fat_raw_buffer address space on source");
-            return failure();
+            return rewriter.notifyMatchFailure(
+                dmaOp, "in_bounds with OOB dimensions requires "
+                       "fat_raw_buffer address space on source");
           }
         }
       }
@@ -503,7 +503,6 @@ struct AMDGPULowerCoalescedDMAToGatherLDSPass final
 
     walkAndApplyPatterns(funcOp, std::move(patterns));
 
-#ifndef NDEBUG
     // Verify all CoalescedGatherDMAOps were lowered. Currently, we require all
     // ops to be successfully lowered. In the future, a fallback lowering path
     // (e.g., using global_load) could handle ops that don't match the pattern.
@@ -514,7 +513,6 @@ struct AMDGPULowerCoalescedDMAToGatherLDSPass final
     if (result.wasInterrupted()) {
       return signalPassFailure();
     }
-#endif // NDEBUG
   }
 };
 } // namespace
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -409,8 +409,8 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
       }
     }
 
-    // Fallback: original behavior without tensor.pad fusion.
-    // Only trace through ONE level of extract_slice (the immediate input).
+    // Fallback: no tensor.pad fusion. The input is an extract_slice from
+    // tiling; trace through it to get the actual source.
     if (!source) {
       if (auto extractSlice = input.getDefiningOp<tensor::ExtractSliceOp>()) {
         source = extractSlice.getSource();
@@ -545,19 +545,19 @@ struct ConvertPadFusionCopyToCoalescedDMA
 
   LogicalResult matchAndRewrite(linalg::CopyOp copyOp,
                                 PatternRewriter &rewriter) const override {
-    // Only match copies with use_global_load_dma config
+    // Only match copies with use_global_load_dma config.
     auto config = getLoweringConfig<IREE::GPU::UseGlobalLoadDMAAttr>(copyOp);
     if (!config) {
       return failure();
     }
 
-    // Check if this is a tensor.pad fusion case
+    // Check if this is a tensor.pad fusion case.
     auto pad = traceToTensorPad(copyOp.getInputs()[0]);
     if (!pad) {
       return failure(); // Not a pad fusion case
     }
 
-    // Check if padding exists (non-zero low/high pad)
+    // Check if padding exists (non-zero low/high pad).
     bool hasPadding = false;
     for (auto [low, high] :
          llvm::zip(pad.getMixedLowPad(), pad.getMixedHighPad())) {
@@ -974,7 +974,7 @@ private:
     bool isPadFusion = false;
     if (auto copyOp = dyn_cast<linalg::CopyOp>(op.getOperation())) {
       if (auto pad = traceToTensorPad(copyOp.getInputs()[0])) {
-        // Check if padding exists (non-zero low/high pad)
+        // Check if padding exists (non-zero low/high pad).
         for (auto [low, high] :
              llvm::zip(pad.getMixedLowPad(), pad.getMixedHighPad())) {
           if (!isConstantIntValue(low, 0) || !isConstantIntValue(high, 0)) {
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -23,6 +23,7 @@
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/Matchers.h"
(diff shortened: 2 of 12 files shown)


krzysz00:
I'd like to flag that we'll need to handle the slow path (address less than 4 bytes from the edge of the buffer) 

I'll also note that we should check the semantics of gather to LDS within an if statement because it's possible this might work for global loads as well

lialan:
> I'll also note that we should check the semantics of gather to LDS within an if statement because it's possible this might work for global loads as well

@krzysz00 Did you mean we avoid using buffer read but stick with global loads? I am confused.

krzysz00:
What I mean is that we should confirm that

```
x = ...; 
if (isInBounds(x)) {
  gather_to_lds global[x], lds[...] , ...
}
```
works in a way we can work with

If the answer is something like "yes, that causes the lanes that don't take the if to not write a value / write 0", then that'll be useful as a fallback for when we can't guarantee bufferability

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -415,3 +408,171 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
   return %result : tensor<64x128xf32>
 }
+
+// -----
+
+// Test: tensor.pad fusion into coalesced_gather_dma.
+// When linalg.copy reads from tensor.pad, trace through to the original source
+// and set in_bounds attribute based on padding.
+
+#gpu_target_pad = #iree_gpu.target<arch = "gfx942", features = "", wgp = <
```
krzysz00: gfx950 for the target - let's not lie about what LDS operations are available in tests

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -415,3 +408,171 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
+    ins(%padded : tensor<4x64xf32>)
+    outs(%init : tensor<4x64xf32>) -> tensor<4x64xf32>
+
+  // Key check: tensor.pad is fused - source is the extract_slice result, not the padded tensor
+  // in_bounds = [false, true] because M dim has dynamic padding, K dim has no padding
+  // CHECK: %[[EXTRACTED:.+]] = tensor.extract_slice %[[SRC]]
+  // CHECK: scf.forall {{.*}} shared_outs(%[[OUTER_INIT:.+]] = %[[INIT]])
+  // CHECK:   scf.forall (%[[LANE:.+]]) in (64) shared_outs(%[[INNER_INIT:.+]] = %[[OUTER_INIT]])
+  // CHECK:     scf.forall.in_parallel {
+  // CHECK:       iree_gpu.coalesced_gather_dma %[[EXTRACTED]] into %[[INNER_INIT]] lane(%[[LANE]]) in_bounds [false, true]
```
krzysz00, line 492: High-level comment: where is the padding element? What if I want to pad with -inf or 1 or something?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -415,3 +408,171 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
+}
+
+// -----
+
+// Test: tensor.pad fusion bails out when source row size is not DWORD-aligned.
+// On AMD CDNA, per-component range checking is performed for each DWORD.
+// If a DWORD is partially out-of-bounds, the entire DWORD returns zero,
+// causing incorrect results. We bail out to avoid the slow path.
+
+#gpu_target_pad_unaligned = #iree_gpu.target<arch = "gfx942", features = "", wgp = <
```
krzysz00: ... Ok, general note about the target

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -58,6 +59,15 @@ static SmallVector<Attribute> getThreadMapping(MLIRContext *ctx) {
   return mapping;
 }
 
+/// Trace through extract_slice operations to find an underlying tensor.pad.
+/// Returns the PadOp if found, nullptr otherwise.
+static tensor::PadOp traceToTensorPad(Value source) {
+  while (auto extractSlice = source.getDefiningOp<tensor::ExtractSliceOp>()) {
```
krzysz00, line 67: Are there other operations that could go here?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -300,14 +310,74 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
...
+             llvm::zip(pad.getMixedLowPad(), pad.getMixedHighPad())) {
+          bool isInBounds =
+              isConstantIntValue(low, 0) && isConstantIntValue(high, 0);
+          inBoundsVec.push_back(isInBounds);
+        }
+      }
+    }
+
+    // Fallback: original behavior without tensor.pad fusion.
+    // Only trace through ONE level of extract_slice (the immediate input).
```
krzysz00: Why this restriction?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -421,6 +498,58 @@ struct ConvertCopyToCoalescedDMA
...
+      if (!isConstantIntValue(low, 0) || !isConstantIntValue(high, 0)) {
+        hasPadding = true;
+        break;
+      }
+    }
+    if (!hasPadding) {
+      return failure(); // No actual padding
+    }
+
+    // This is a tensor.pad fusion case. Convert directly to
```
krzysz00, line 636: ... why?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -757,9 +889,42 @@ struct GPUConvertToCoalescedDMAPass final
...
+          }
+        }
+      }
+    }
+
+    SmallVector<OpFoldResult> tileSizes;
+    int64_t numTiledDims = 0;
+
+    if (isPadFusion) {
+      // For tensor.pad fusion, create a single-iteration wrapper forall
```
krzysz00: (that is, I don't understand why we can't also tile to subgroups here)

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
```diff
@@ -300,6 +338,18 @@ LogicalResult CoalescedGatherDMAOp::verify() {
       continue;
     }
 
+    // If in_bounds is present and this dimension allows OOB (in_bounds=false),
+    // skip the size matching check - hardware OOB returns 0 for padding.
```
krzysz00: I'm not sure that's actually as correct as you want it to be.

Consider a 64x62xf32 region of memory padded out to 64x64xf32

If I read <4 x f32> starting at [0, 60], I'll get (0, 60) (0, 61), (1, 0), (1, 1) from the underlying tensor

If I read from [63, 60], I'll get (63, 60), (63, 61), 0, 0 because the OOB handler will kick in

krzysz00 (dismissed):
Hold on this PR because I have correctness concerns

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -319,13 +320,15 @@ def IREEGPU_CoalescedGatherDMAOp : Op<IREEGPU_Dialect, "coalesced_gather_dma", [
     AnyRankedTensorOrMemRef:$source,
     Variadic<RankedTensorOrVectorOrMemRefOf<[I32, Index]>>:$indices,
     AnyRankedTensorOrMemRef:$init,
-    Index:$lane
+    Index:$lane,
+    OptionalAttr<BoolArrayAttr>:$in_bounds
```
qedawkins, line 337: Op description needs to be updated to describe the semantics of `in_bounds` (the defaults if not present, what the op does if it's false etc.)

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
qedawkins: We shouldn't be mentioning hardware here, we should be talking in terms of the semantics of the op on its own (which we happen to pick semantics for that match the hardware).

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
```diff
@@ -205,6 +205,43 @@ Operation *CoalescedGatherDMAOp::getIteratingParent() {
...
+  // For tensor form with result, the write is captured in the result value
+  // For tensor form without result (combiner case in forall.in_parallel),
+  // we must declare a write effect to prevent DCE from eliminating the op.
+  if (isa<MemRefType>(init.getType())) {
+    effects.emplace_back(MemoryEffects::Write::get(),
+                         &getOperation()->getOpOperand(initOperandIdx),
+                         SideEffects::DefaultResource::get());
+  } else if (isa<RankedTensorType>(init.getType()) &&
+             getOperation()->getNumResults() == 0) {
+    // Tensor combiner case: declare write effect to prevent DCE
```
qedawkins: nit: Here and above
```suggestion
    // Tensor combiner case: declare write effect to prevent DCE.
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -415,3 +408,171 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
+
+#exec_target_pad = #hal.executable.target<"rocm", "rocm-hsaco-fb", {iree_codegen.target_info = #gpu_target_pad}>
+#translation_pad = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse workgroup_size = [64, 1, 1] subgroup_size = 64, {gpu_pipeline_options = #iree_gpu.pipeline_options<prefetch_num_stages = 2, no_reduce_shared_memory_bank_conflicts = true, use_igemm_convolution = false>}>
+
+// CHECK-LABEL: func.func @copy_with_tensor_pad_fusion
+// CHECK-SAME:    %[[SRC:[a-zA-Z0-9]+]]: tensor<121x64xf32>
+// CHECK-SAME:    %[[INIT:[a-zA-Z0-9]+]]: tensor<4x64xf32>
+func.func @copy_with_tensor_pad_fusion(%source: tensor<121x64xf32>, %init: tensor<4x64xf32>, %off: index, %sz: index, %high: index) -> tensor<4x64xf32>
+  attributes {hal.executable.target = #exec_target_pad, translation_info = #translation_pad} {
+  // Extract a dynamic slice
```
qedawkins: nit: Here and below.
```suggestion
  // Extract a dynamic slice.
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1025,3 +1025,142 @@ func.func @lower_coalesced_dma_lane_offset_regression(
   } {mapping = [#gpu.thread<linear_dim_0>]}
   return
 }
+
```
qedawkins, line 1028: I can't find where the changes to the amdgpu lowering were added. Don't we need to switch on the offset to force it out of bounds for inner row? For example `in_bounds = [false, true]` means that row 0 will go out of bounds, but won't be out of bounds of the buffer, meaning we have to manually force it to "read" out of bounds to actually get the padding behavior we want. That was my expectation anyway, but I don't see that happening in these tests, nor do I see code changes that suggest we're doing that. Am I missing something here?

qedawkins (commented):
A few comments on top of what Krzysztof added.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
lialan: we can't fuse it if -inf or 1 or anything other than 0.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: so far only extrace slice appears between pad in our tiling pipeline. 

Let's be conservative so if anything changes we will be notified by tests

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: different path?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
krzysz00: Not into the gather to LDS directly, no, but it's just a select on the loaded values from LDS

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: Reason being ...When tiling to ubgroups, the extract_slice from tiling loses the padding context — the DMA sees a statically-sized sub-tile rather than the full padded buffer. We need the DMA to see the full shape to correctly determine which lanes are OOB. Could be improved thou by propagating `in_bounds` through the tiling.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
lialan: As discussed, added a guard to load OOB values to ensure correctness ... for now.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
lialan: added.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
krzysz00: Yeah, prop getting the in_bounds through tiling seems like the right way to do this

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -400,6 +402,61 @@ struct LowerCoalescedGatherDMAPattern final
...
+
+              if (anyNonOutermostOOB) {
+                anyNonOutermostOOB = arith::OrIOp::create(
+                    rewriter, loc, anyNonOutermostOOB, isOOB);
+              } else {
+                anyNonOutermostOOB = isOOB;
+              }
+            }
+
+            if (anyNonOutermostOOB) {
```
krzysz00: You need to check address spaces here - this won't work if we're not using buffers

Also. you can simplify a bunch of this by using identity values like false and letting IR folding clean up the generated code

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
lialan: The force-OOB code is in `AMDGPULowerCoalescedDMAToGatherLDS.cpp` at lines 405-458, added in commit 7720ffa1df. For non-outermost dimensions where `in_bounds[dim] = false`, it compares `srcIndices[dim] >= sourceShape[dim]` and replaces the outermost index with `sourceShape[0]` to force the linearized offset past the buffer end, triggering hardware OOB loading of 0. 

So we are actually good here. The loads are already guarded by checks.

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1025,3 +1025,277 @@ func.func @lower_coalesced_dma_lane_offset_regression(
...
+//   - Elements per lane = 64 / 64 = 1 (each lane reads 1 f32)
+//   - Delinearization basis = (4, 64)
+//   - 4 transfers per lane (one per row)
+//
+// This verifies correct row access pattern: all 4 rows (0-3) are accessed,
+// not just row 0 repeated 4 times (which was the bug before the fix).
+
+#executable_target_rocm_hsaco_fb_unaligned = #hal.executable.target<"rocm",
+  "rocm-hsaco-fb", {iree_codegen.target_info = #iree_gpu.target<
+  arch = "gfx942", features = "", wgp = <
```
Yu-Zhewen: Same as the other comment from @krzysz00, please use gfx950 here.

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1025,3 +1025,277 @@ func.func @lower_coalesced_dma_lane_offset_regression(
...
+    compute = fp32, storage = b32, subgroup = shuffle, dot = none, mma = [],
+    subgroup_size_choices = [64, 64],
+    max_workgroup_sizes = [1024, 1024, 1024],
+    max_thread_count_per_workgroup = 1024,
+    max_workgroup_memory_bytes = 65536,
+    max_workgroup_counts = [2147483647, 2147483647, 2147483647],
+    max_load_instruction_bits = 128, simds_per_wgp = 4,
+    vgpr_space_bits = 8192, dma_sizes = [32]>>}>
```
Yu-Zhewen:
```suggestion
    vgpr_space_bits = 8192, dma_sizes = [32, 128]>>}>
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1025,3 +1025,277 @@ func.func @lower_coalesced_dma_lane_offset_regression(
...
+#executable_target_rocm_hsaco_fb_pad = #hal.executable.target<"rocm",
+  "rocm-hsaco-fb", {iree_codegen.target_info = #iree_gpu.target<
+  arch = "gfx950", features = "", wgp = <
+    compute = fp32, storage = b32, subgroup = none, dot = none, mma = [], subgroup_size_choices = [32, 32],
+    max_workgroup_sizes = [1024, 1024, 1024],
+    max_thread_count_per_workgroup = 1024,
+    max_workgroup_memory_bytes = 65536,
+    max_workgroup_counts = [2147483647, 2147483647, 2147483647],
+    max_load_instruction_bits = 128, simds_per_wgp = 4,
+    vgpr_space_bits = 8192, dma_sizes = [32]>>}>
```
Yu-Zhewen:
```suggestion
    vgpr_space_bits = 8192, dma_sizes = [32, 128]>>}>
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -415,3 +408,171 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
+#exec_target_pad_unaligned = #hal.executable.target<"rocm", "rocm-hsaco-fb", {iree_codegen.target_info = #gpu_target_pad_unaligned}>
+#translation_pad_unaligned = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse workgroup_size = [64, 1, 1] subgroup_size = 64, {gpu_pipeline_options = #iree_gpu.pipeline_options<prefetch_num_stages = 2, no_reduce_shared_memory_bank_conflicts = true, use_igemm_convolution = false>}>
+
+// CHECK-LABEL: func.func @copy_with_tensor_pad_unaligned_row
+// CHECK-SAME:    %[[SRC:[a-zA-Z0-9]+]]: tensor<65x121xf16>
+// CHECK-SAME:    %[[INIT:[a-zA-Z0-9]+]]: tensor<4x124xf16>
+func.func @copy_with_tensor_pad_unaligned_row(%source: tensor<65x121xf16>, %init: tensor<4x124xf16>, %off: index, %sz: index, %high_m: index) -> tensor<4x124xf16>
+  attributes {hal.executable.target = #exec_target_pad_unaligned, translation_info = #translation_pad_unaligned} {
+  // Extract a dynamic slice: tensor<?x121xf16>
+  // Row size = 121 * 2 bytes = 242 bytes, NOT 4-byte aligned
```
Yu-Zhewen:
```suggestion
  // Extract a dynamic slice: tensor<?x121xf16>.
  // Row size = 121 * 2 bytes = 242 bytes, NOT 4-byte aligned.
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -415,3 +408,171 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
+  %extracted = tensor.extract_slice %source[%off, 0] [%sz, 121] [1, 1]
+      : tensor<65x121xf16> to tensor<?x121xf16>
+
+  // Pad to static size
```
Yu-Zhewen:
```suggestion
  // Pad to static size.
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -415,3 +408,171 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
+  %cst = arith.constant 0.0 : f16
+  %padded = tensor.pad %extracted low[0, 0] high[%high_m, 3] {
+  ^bb0(%arg0: index, %arg1: index):
+    tensor.yield %cst : f16
+  } : tensor<?x121xf16> to tensor<4x124xf16>
+
+  // Copy from padded tensor
```
Yu-Zhewen:
```suggestion
  // Copy from padded tensor.
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -300,14 +310,74 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
...
     Value input = innerOp.getInputs()[0];
-    if (auto extractSlice = input.getDefiningOp<tensor::ExtractSliceOp>()) {
-      source = extractSlice.getSource();
-    } else {
-      return failure();
+
+    // After tiling, the input is typically:
+    //   tensor.extract_slice %padded[...] [...] [1, 1]
+    // We need to trace through extract_slice to find if source is tensor.pad.
+    if (auto pad = traceToTensorPad(input)) {
```
Yu-Zhewen:
```suggestion
    if (tensor::PadOp pad = traceToTensorPad(input)) {
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -757,9 +889,42 @@ struct GPUConvertToCoalescedDMAPass final
       return failure();
     }
 
-    // Compute tile sizes for subgroup-level distribution.
-    auto [tileSizes, numTiledDims] =
-        computeSubgroupTileSizes(rewriter, shape, numWarps);
+    // Check if this is a tensor.pad fusion case.
+    bool isPadFusion = false;
+    if (auto copyOp = dyn_cast<linalg::CopyOp>(op.getOperation())) {
+      if (auto pad = traceToTensorPad(copyOp.getInputs()[0])) {
```
Yu-Zhewen:
```suggestion
      if (tensor::PadOp pad = traceToTensorPad(copyOp.getInputs()[0])) {
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
```diff
@@ -380,12 +380,22 @@ struct CoalescedGatherDMAOpBufferizationInterface
       }
     }
 
-    rewriter.setInsertionPoint(gatherOp);
+    // Insert the memref operation in the forall body, before the in_parallel
+    // terminator (not inside the in_parallel region which will be removed).
+    auto inParallelOp = gatherOp->getParentOfType<scf::InParallelOp>();
+    if (inParallelOp) {
+      // Insert before the in_parallel terminator (in the forall body)
```
Yu-Zhewen:
```suggestion
      // Insert before the in_parallel terminator (in the forall body).
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
```diff
@@ -380,12 +380,22 @@ struct CoalescedGatherDMAOpBufferizationInterface
...
+      rewriter.setInsertionPoint(inParallelOp);
+    } else {
+      // Not in in_parallel, just insert at current location
```
Yu-Zhewen:
```suggestion
      // Not in in_parallel, just insert at current location.
```

tests/e2e/matmul/CMakeLists.txt
```diff
@@ -1426,6 +1426,97 @@ iree_generated_e2e_runner_test(
     "requires-gpu-${_CDNA_ARCH}"
 )
 
+# Unaligned matmul tests for coalesced DMA with tensor.pad fusion
```
Yu-Zhewen, line 1537: Would it be better to move those tests under

https://github.com/iree-org/iree/blob/753cd7fd88ffba2b64d0d7aa1c55146b4ecbefe6/tests/e2e/matmul/CMakeLists.txt#L2288

since they are gfx950 specific.

Yu-Zhewen (commented):
Thanks! Some Nits.

tests/e2e/matmul/CMakeLists.txt
lialan: They are actually guarded by:
https://github.com/iree-org/iree/blob/753cd7fd88ffba2b64d0d7aa1c55146b4ecbefe6/tests/e2e/matmul/CMakeLists.txt#L1191

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: let me just add a TODO for this.

lialan: the change will be non-trivial, I added a TODO here. 

But it does not impact functionalities.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
lialan: added TODO.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -408,10 +408,11 @@ struct LowerCoalescedGatherDMAPattern final
           // Fix: when any non-outermost source index exceeds its dimension,
           // replace the outermost index with sourceShape[0] to force the
           // linearized offset past the buffer end → hardware returns 0.
-          if (inBoundsAttr) {
-            auto sourceType = cast<MemRefType>(source.getType());
+          auto sourceType = cast<MemRefType>(source.getType());
+          if (inBoundsAttr && hasAMDGPUFatRawBufferAddressSpace(sourceType)) {
```
krzysz00, line 426: If there are any non-in-bounds dimensions and we're not using buffers, we either need to fail or put an if statement around the gather_to_lds to guard the copying.

(If we go the if statement route, we need a test for that - either a dynamic buffer we can't prove is < 2 GB or a buffer that is statically > 2 GB)

lialan: Good suggestion. I added a guard a test for that.

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1025,3 +1025,276 @@ func.func @lower_coalesced_dma_lane_offset_regression(
...
+    // CHECK: %[[LANE_OFFSET:[a-zA-Z0-9_]+]] = arith.muli %[[LANE_ID]], %[[C4]]
+    //
+    // Transfer 1: row 0
+    // CHECK: amdgpu.gather_to_lds {{.+}} : vector<4xf32>
+    // Transfer 2: row 1
+    // CHECK: amdgpu.gather_to_lds {{.+}} : vector<4xf32>
+    // Transfer 3: row 2 (OOB in source, hardware returns 0)
+    // CHECK: amdgpu.gather_to_lds {{.+}} : vector<4xf32>
+    // Transfer 4: row 3 (OOB in source, hardware returns 0)
+    // CHECK: amdgpu.gather_to_lds {{.+}} : vector<4xf32>
```
qedawkins: This test is not testing much it looks like to me? It's just testing that we got 4 `gather_to_lds` ops. Shouldn't we be checking the operands to the gathers? What did the `in_bounds` attribute do here?

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1025,3 +1025,276 @@ func.func @lower_coalesced_dma_lane_offset_regression(
...
+    //
+    // Bounds check: compare srcIndices[1] >= 62 (source inner dim size).
+    // CHECK: %[[FALSE:.+]] = arith.constant false
+    // CHECK: %[[C62:.+]] = arith.constant 62 : index
+    // CHECK: %[[CMP:.+]] = arith.cmpi uge, %[[SRC_DELIN0]]#1, %[[C62]] : index
+    // CHECK: %[[OOB:.+]] = arith.ori %[[FALSE]], %[[CMP]] : i1
+    // Replace outermost index with 64 (source dim 0 size) to force hardware OOB.
+    // CHECK: %[[C64_OOB:.+]] = arith.constant 64 : index
+    // CHECK: %[[FIXED_IDX:.+]] = arith.select %[[OOB]], %[[C64_OOB]], %[[SRC_DELIN0]]#0 : index
+    // CHECK: amdgpu.gather_to_lds %[[SRC]][%[[FIXED_IDX]], %[[SRC_DELIN0]]#1], %[[DST]][%[[DST_DELIN0]]#0, %[[DST_DELIN0]]#1] : vector<4xf32>
```
qedawkins, line 1270: This is the partial read off the end case. Does this work how we expect cc @krzysz00?

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1025,3 +1025,276 @@ func.func @lower_coalesced_dma_lane_offset_regression(
...
+
+#translation_64 = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse workgroup_size = [64, 1, 1] subgroup_size = 32>
+
+func.func @no_lower_oob_without_fat_raw_buffer(
+    %source: memref<2x128xf32>,
+    %dest: memref<4x128xf32, #gpu.address_space<workgroup>>)
+    attributes {hal.executable.target = #executable_target_rocm_hsaco_fb,
+                translation_info = #translation_64} {
+  scf.forall (%arg6) in (64) {
+    // expected-error @+1 {{failed to lower coalesced_gather_dma op}}
```
qedawkins: Can we maybe get a little more descriptive failure message?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -451,3 +444,328 @@ func.func @copy_with_extract_slice_input(%large_source: tensor<256x128xf32>) ->
...
+  attributes {hal.executable.target = #exec_target_fat_raw, translation_info = #translation_fat_raw} {
+  %source = iree_codegen.load_from_buffer %buf
+      : memref<64x512xbf16, #amdgpu.address_space<fat_raw_buffer>> -> tensor<64x512xbf16>
+  %result = linalg.copy {lowering_config = #iree_gpu.use_global_load_dma}
+    ins(%source : tensor<64x512xbf16>)
+    outs(%init : tensor<64x512xbf16>) -> tensor<64x512xbf16>
+
+  // fat_raw_buffer source: sourceIsNotFromFatRawBuffer returns false, DMA applied.
+  // 2 warps (128/64), 64 rows → step = 32 rows, 512 cols whole.
+  // CHECK: iree_gpu.coalesced_gather_dma
```
qedawkins: Since you're matching the SSA values above, match them on here too. Same below

qedawkins (approved):
I'm closing my eyes to the chipset leak. This should go in an AMD specific directory if it's going to be AMD specific.

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
lialan: Added more descriptive errors.

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
lialan: Added detailed operand checking instead of just counting gather_to_lds ops

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
lialan: Added SSA value matching.

PR #23369, opened 2026-02-03
[CI] Update iree-test-suite ref

amd-eochoalo:
Updates the iree-test-suite to a ref that does not make use of git-lfs.

### DIFF
--- a/.github/workflows/pkgci_test_onnx.yml
+++ b/.github/workflows/pkgci_test_onnx.yml
@@ -107,7 +107,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           repository: iree-org/iree-test-suites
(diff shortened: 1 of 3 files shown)


amd-eochoalo:
From the failing tests:

1. Test AMD MI355: interesting numerical difference, will re-run. These definitely don't affect this test, so if it is something more than a fluke, it is not this change.
2. Test ONNX: Timeout outs. Hopefully rerunning will get rid of them
3. RISC-V: has happened before on main
4. Sharktank tests: We achieved what we wanted, which is moved past the git-lfs issue
5. Test Torch Ops: small tolerance issue, will follow up on another PR. Progressed passed git-lfs issue.
6. Torch models: Azure access issue (also seen on main)

Muzammiluddin-Syed-ECE (approved):

amd-eochoalo:
Rerunning fixed:

* Test AMD MI355
* Test ONNX: Timeout outs.

This is an improvement over the current state of the CI.

Issue #23371, opened 2026-02-03
[VMVX] compilation failure on unsupported fp types and narrow types (i4).


hanhanW:
VMVX has been behind in terms of sub-byte and small float support. The support on CPU side is completed, and we can also make VMVX work. The final prototype can be found in https://github.com/hanhanW/iree/commits/users/hanhanW/vmvx-subbyte/ which enables those e2e tests for VMVX.

PR #23372, opened 2026-02-03
[VM] Add sub-byte integer support to ArithToVM.

hanhanW:
Generalize ExtUI/ExtSI/TruncI conversions from i1-only to handle all power-of-two sub-byte widths (i1, i2, i4):

- ExtUI: mask-and-AND for zero extension.
- ExtSI: shift-left/arithmetic-shift-right for sign extension (i2, i4); keep select-based for i1.
- TruncI: mask to the target bit width.

Only power-of-two bit widths are supported to avoid issues with non-standard sizes like i5 or i6.

It is a step towards https://github.com/iree-org/iree/issues/23371


benvanik (changes requested):
a few valid bugprone things (you're testing isInteger(32) and then again below && isInteger(32)

benvanik (approved):
actually, those may be preexisting - yuck

PR #23374, opened 2026-02-03
[VM] Add buffer ops for 8-bit and 16-bit floats in UtilToVM.

hanhanW:
Add buffer fill/load/store support for small float types:
- 8-bit floats (f8 variants): use i8 buffer ops, byte-addressed
- 16-bit floats (f16, bf16): use i16 buffer ops, element-addressed

This mirrors how integer types are handled (i8 uses byte ops, i16 uses element ops).

It is a step towards https://github.com/iree-org/iree/issues/23371


hanhanW:
It depends on https://github.com/iree-org/iree/pull/23372 and https://github.com/iree-org/iree/pull/23373

benvanik (approved):

PR #23375, opened 2026-02-04
[VMVX] Enable sub-byte and small float support in VMVX pipeline.

### LINKED ISSUE #23371: [VMVX] compilation failure on unsupported fp types and narrow types (i4).

hanhanW:
Add passes to VMVX codegen pipline to enable subbyte and small float support.

Pipeline changes:
- Add ArithExpandOps with bf16/f8E8M0 options
- Add ConvertUnsupportedFloatArith pass
- Add EmulateNarrowType pass

Enable the previously excluded e2e tests for VMVX:
- fp4_f32_conversion
- fp_to_subbyte
- small_float_arith
- subbyte_to_fp

The revision exposes an option to disable atomic ops in EmulateNarrowType pass. It is safe to disable it on VMVX because VMVX usually picks tile_size=64 for distribution. There are no races because it does not use cyclic distrubtion. Ideally, we should do an analysis to determine whether atomicRMW ops are needed or not, but it is very involved atm. The pass does not only apply emulation but also linearization.

Fixes https://github.com/iree-org/iree/issues/23371

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/EmulateNarrowType.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/EmulateNarrowType.cpp
@@ -102,6 +102,8 @@ static void populateIreeNarrowTypeEmulationPatterns(
 
 struct EmulateNarrowTypePass final
     : impl::EmulateNarrowTypePassBase<EmulateNarrowTypePass> {
+  using Base::Base;
+
   void getDependentDialects(DialectRegistry &registry) const override {
     registry
         .insert<arith::ArithDialect, func::FuncDialect, memref::MemRefDialect,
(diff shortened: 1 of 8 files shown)


hanhanW:
It depends on
- https://github.com/iree-org/iree/pull/23238
- https://github.com/iree-org/iree/pull/23319
- https://github.com/iree-org/iree/pull/23344
- https://github.com/iree-org/iree/pull/23372
- https://github.com/iree-org/iree/pull/23373
- https://github.com/iree-org/iree/pull/23374

bjacob (approved):

benvanik (approved):

PR #23380, opened 2026-02-04
Introduce `iree_cpu_set_t`, a better `cpu_set_t`

bjacob:
The POSIX standard structure, `cpu_set_t`, has issues:
* It was designed in an era of low core counts, when something like uint64_t felt like it would be enough. It gradually grew to its current size of 1024 bits (glibc) while dynamic-size extensions were also introduced, resulting in a worst-of-both-worlds: it has a large static size AND it still had to mode-switch to dynamic.
* Its bit layout is opaque, so bits can only be addressed individually, which is inefficient unless the compiler is able to coalesce 32 or 64 consecutive accesses, which is not trivial as it relies on loop tiling.

`iree_cpu_set_t` addresses these issues:
* Explicit layout by 64-bit words with bit-to-word mapping given by division (word_idx = bit_idx / 64) as in `iree_bitmap_t`, aka "little endian" word order.
* Direct word addressing.
* Static size just a single word. Storage is inline up to bit count 56, dynamic after that. When using dynamic storage, the bit-count is stored in the first word of dynamic storage.
  * This relies on bit 63 being 0 in pointers: the top bit is reserved to indicate inline storage.  Thus the only assumption here is that pointers never have bit 63 set.
     * The bit-count is intentionally NOT stored in the static storage so as to avoid any issues with large address space hosts (5-level paging => 57-bit address space).
     * Pointers stored in memory are still true pointers, not tagged pointers: bit 63 is set only when *not* storing a pointer.
     * On 32-bit pointer architectures, the whole 32 bits are available for the pointer, which is stored in the low 32 bits, not overlapping with bit 63.

### DIFF
--- a/runtime/src/iree/base/internal/BUILD.bazel
+++ b/runtime/src/iree/base/internal/BUILD.bazel
@@ -378,6 +378,24 @@ iree_runtime_cc_test(
     ],
 )
 
+iree_runtime_cc_library(
+    name = "cpu_set",
+    hdrs = ["cpu_set.h"],
+    deps = [
+        "//runtime/src/iree/base",
+    ],
+)
+
+iree_runtime_cc_test(
+    name = "cpu_set_test",
+    srcs = ["cpu_set_test.cc"],
+    deps = [
+        ":cpu_set",
+        "//runtime/src/iree/testing:gtest",
+        "//runtime/src/iree/testing:gtest_main",
+    ],
+)
+
 #===------------------------------------------------------------------------===#
 # Utilities with thread dependencies
 #===------------------------------------------------------------------------===#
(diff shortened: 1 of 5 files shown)


amd-eochoalo (approved):
I think this looks good in my opinion. Looking forward to see where these functions will be used.

bjacob:
Following dumping my thoughts in https://github.com/iree-org/iree/issues/23395 , I'm not sure that I'll get back to this myself, so I'm not sure that I'd have a short-term plan to use this at all, myself.  It would get used whenever https://github.com/iree-org/iree/issues/23395 is acted on, or possibly by @benvanik for his own NUMA purposes.  I'm not sure that we should merge this if it's not going to be used, so we can close this for now, unless someone is thinking about doing the above things.

PR #23386, opened 2026-02-04
[VectorExt] Add iree_vector_ext.arg_compare operation

bangtianliu:
**Context**: I am in the process of adding vectorization support for `iree_linalg_ext.arg_compare` operation. 
I plan to split the work into two PRs: 
- adds the `iree_vector_ext.arg_compare` operation (this PR)
- vectorizes `iree_linalg_ext.arg_compare`, converted to `iree_vector_ext.arg_compare`.

Refer to Issue: #23005 and [discord discussion](https://discord.com/channels/689900678990135345/1461079481803608158) for more context.  

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
@@ -684,17 +684,100 @@ Type TransferGatherOp::getExpectedMaskType() {
 }
 
 //===----------------------------------------------------------------------===//
+// YieldOp
+//===----------------------------------------------------------------------===//
+
+static LogicalResult verifyYieldForArgCompare(YieldOp yieldOp,
+                                              ArgCompareOp argCompareOp) {
+  unsigned numOperands = yieldOp.getNumOperands();
+  if (numOperands != 1) {
+    return yieldOp.emitOpError("expected 1 yield operand, but got ")
+           << numOperands;
+  }
+
+  Type yieldType = yieldOp.getOperand(0).getType();
+  if (!yieldType.isInteger(1)) {
+    return yieldOp.emitOpError(
+               "expected yield operand to have type i1, but got ")
+           << yieldType;
+  }
+
+  return success();
+}
+
+LogicalResult YieldOp::verify() {
+  // ParentOneOf<["ArgCompareOp"]> ODS trait ensures parent is ArgCompareOp.
+  auto argCompareOp = cast<ArgCompareOp>((*this)->getParentOp());
+  return verifyYieldForArgCompare(*this, argCompareOp);
+}
+
+//===----------------------------------------------------------------------===//
 // ArgCompareOp
 //===----------------------------------------------------------------------===//
 
 LogicalResult ArgCompareOp::verify() {
   Operation *op = getOperation();
 
-  // Most constraints are enforced in ODS; only conditional logic remains here.
+  VectorType inputType = getInputValueType();
+  VectorType initValueType = getInitValueType();
+  VectorType initIndexType = getInitIndexType();
+
+  int64_t inputRank = inputType.getRank();
+  int64_t initValueRank = initValueType.getRank();
+  int64_t initIndexRank = initIndexType.getRank();
+  int64_t dimension = getDimension();
+
+  if (dimension < 0 || dimension >= inputRank) {
+    return op->emitOpError("dimension ")
+           << dimension << " is out of range [0, " << inputRank << ")";
+  }
+
+  if (initValueRank != inputRank - 1) {
+    return op->emitOpError("init value rank (")
+           << initValueRank << ") must be input rank - 1 (" << (inputRank - 1)
+           << ")";
+  }
+
+  if (initIndexRank != inputRank - 1) {
+    return op->emitOpError("init index rank (")
+           << initIndexRank << ") must be input rank - 1 (" << (inputRank - 1)
+           << ")";
+  }
+
+  SmallVector<int64_t> expectedShape;
+  for (int64_t i = 0; i < inputRank; ++i) {
+    if (i != dimension) {
+      expectedShape.push_back(inputType.getDimSize(i));
+    }
+  }
+
+  ArrayRef<int64_t> initValueShape = initValueType.getShape();
+  if (expectedShape != initValueShape) {
+    return op->emitOpError(
+               "init value shape must match input shape with reduction "
+               "dimension removed. ")
+           << "Expected: " << llvm::interleaved_array(expectedShape)
+           << ", but got: " << llvm::interleaved_array(initValueShape);
+  }
+
+  ArrayRef<int64_t> initIndexShape = initIndexType.getShape();
+  if (expectedShape != initIndexShape) {
+    return op->emitOpError(
+               "init index shape must match input shape with reduction "
+               "dimension removed. ")
+           << "Expected: " << llvm::interleaved_array(expectedShape)
+           << ", but got: " << llvm::interleaved_array(initIndexShape);
+  }
+
+  Type initIndexElementType = initIndexType.getElementType();
+  if (!isa<IntegerType, IndexType>(initIndexElementType)) {
+    return op->emitOpError(
+               "init index must have integer or index element type, but got ")
+           << initIndexElementType;
+  }
+
   if (hasExplicitIndexInput()) {
-    VectorType inputType = getInputValueType();
     VectorType inputIndexType = getInputIndexType();
-    VectorType initIndexType = getInitIndexType();
     ArrayRef<int64_t> inputIndexShape = inputIndexType.getShape();
     ArrayRef<int64_t> inputValueShape = inputType.getShape();
 
@@ -729,37 +812,22 @@ LogicalResult ArgCompareOp::verify() {
     }
   }
 
-  Region &comparator = getRegion();
-  if (comparator.getNumArguments() != 2) {
+  // Region structure is enforced by ODS (SizedRegion<1> and
+  // SingleBlockImplicitTerminator), so we can directly access it.
+  Block &block = getRegion().front();
+  if (block.getNumArguments() != 2) {
     return op->emitOpError("comparator region must have exactly 2 arguments");
   }
 
-  VectorType inputType = getInputValueType();
   Type inputElementType = inputType.getElementType();
-  Type arg0Type = comparator.getArgument(0).getType();
-  Type arg1Type = comparator.getArgument(1).getType();
+  Type arg0Type = block.getArgument(0).getType();
+  Type arg1Type = block.getArgument(1).getType();
 
   if (arg0Type != inputElementType || arg1Type != inputElementType) {
     return op->emitOpError(
                "comparator arguments must match input value element type. ")
            << "Expected: " << inputElementType << ", but got: " << arg0Type
            << " and " << arg1Type;
-  }
-
-  Block &block = comparator.front();
-  Operation *terminator = block.getTerminator();
-
-  unsigned numYieldOperands = terminator->getNumOperands();
-  if (numYieldOperands != 1) {
-    return op->emitOpError(
-               "expected comparator region to yield 1 operand, but got ")
-           << numYieldOperands;
-  }
-
-  Type yieldType = terminator->getOperand(0).getType();
-  if (!yieldType.isInteger(1)) {
-    return op->emitOpError("comparator region must yield i1, but got ")
-           << yieldType;
   }
 
   // Since ArgCompareOp is marked Pure, all operations in the comparator must
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
@@ -695,10 +695,9 @@ LogicalResult ArgCompareOp::verify() {
            << numInputs;
   }
 
-  unsigned numOutputs = llvm::size(getOutputs());
-  if (numOutputs != 2) {
-    return op->emitOpError("expected 2 output operands, but got ")
-           << numOutputs;
+  unsigned numInits = llvm::size(getInits());
+  if (numInits != 2) {
+    return op->emitOpError("expected 2 init operands, but got ") << numInits;
   }
 
   unsigned numResults = llvm::size(getResults());
@@ -707,12 +706,12 @@ LogicalResult ArgCompareOp::verify() {
   }
 
   VectorType inputType = getInputType();
-  VectorType outputValueType = getOutputValueType();
-  VectorType outputIndexType = getOutputIndexType();
+  VectorType initValueType = getInitValueType();
+  VectorType initIndexType = getInitIndexType();
 
   int64_t inputRank = inputType.getRank();
-  int64_t outputValueRank = outputValueType.getRank();
-  int64_t outputIndexRank = outputIndexType.getRank();
+  int64_t initValueRank = initValueType.getRank();
+  int64_t initIndexRank = initIndexType.getRank();
   int64_t dimension = getDimension();
 
   if (dimension < 0 || dimension >= inputRank) {
@@ -720,15 +719,15 @@ LogicalResult ArgCompareOp::verify() {
            << dimension << " is out of range [0, " << inputRank << ")";
   }
 
-  if (outputValueRank != inputRank - 1) {
-    return op->emitOpError("output value rank (")
-           << outputValueRank << ") should be input rank - 1 (" << inputRank
+  if (initValueRank != inputRank - 1) {
+    return op->emitOpError("init value rank (")
+           << initValueRank << ") should be input rank - 1 (" << inputRank
            << " - 1)";
   }
 
-  if (outputIndexRank != inputRank - 1) {
-    return op->emitOpError("output index rank (")
-           << outputIndexRank << ") should be input rank - 1 (" << inputRank
+  if (initIndexRank != inputRank - 1) {
+    return op->emitOpError("init index rank (")
+           << initIndexRank << ") should be input rank - 1 (" << inputRank
            << " - 1)";
   }
 
@@ -740,38 +739,38 @@ LogicalResult ArgCompareOp::verify() {
     }
   }
 
-  ArrayRef<int64_t> outputValueShape = outputValueType.getShape();
-  if (!llvm::equal(expectedShape, outputValueShape)) {
+  ArrayRef<int64_t> initValueShape = initValueType.getShape();
+  if (!llvm::equal(expectedShape, initValueShape)) {
     return op->emitOpError(
-               "output value shape must match input shape with reduction "
+               "init value shape must match input shape with reduction "
                "dimension removed. ")
            << "Expected: " << llvm::interleaved_array(expectedShape)
-           << ", but got: " << llvm::interleaved_array(outputValueShape);
+           << ", but got: " << llvm::interleaved_array(initValueShape);
   }
 
-  ArrayRef<int64_t> outputIndexShape = outputIndexType.getShape();
-  if (!llvm::equal(expectedShape, outputIndexShape)) {
+  ArrayRef<int64_t> initIndexShape = initIndexType.getShape();
+  if (!llvm::equal(expectedShape, initIndexShape)) {
     return op->emitOpError(
-               "output index shape must match input shape with reduction "
+               "init index shape must match input shape with reduction "
                "dimension removed. ")
            << "Expected: " << llvm::interleaved_array(expectedShape)
-           << ", but got: " << llvm::interleaved_array(outputIndexShape);
+           << ", but got: " << llvm::interleaved_array(initIndexShape);
   }
 
   Type inputElementType = inputType.getElementType();
-  Type outputValueElementType = outputValueType.getElementType();
+  Type initValueElementType = initValueType.getElementType();
 
-  if (inputElementType != outputValueElementType) {
-    return op->emitOpError("input and output value element types must match. ")
+  if (inputElementType != initValueElementType) {
+    return op->emitOpError("input and init value element types must match. ")
            << "Input type: " << inputElementType
-           << ", output value type: " << outputValueElementType;
+           << ", init value type: " << initValueElementType;
   }
 
-  Type outputIndexElementType = getOutputIndexElementType();
-  if (!isa<IntegerType, IndexType>(outputIndexElementType)) {
+  Type initIndexElementType = initIndexType.getElementType();
+  if (!isa<IntegerType, IndexType>(initIndexElementType)) {
     return op->emitOpError(
-               "output index must have integer or index element type, but got ")
-           << outputIndexElementType;
+               "init index must have integer or index element type, but got ")
+           << initIndexElementType;
   }
 
   if (hasExplicitIndexInput()) {
@@ -792,12 +791,12 @@ LogicalResult ArgCompareOp::verify() {
              << inputIndexElementType;
     }
 
-    if (inputIndexElementType != outputIndexElementType) {
+    if (inputIndexElementType != initIndexElementType) {
       return op->emitOpError(
-                 "explicit-index mode: input and output index element types "
+                 "explicit-index mode: input and init index element types "
                  "must match. ")
              << "Input index type: " << inputIndexElementType
-             << ", output index type: " << outputIndexElementType;
+             << ", init index type: " << initIndexElementType;
     }
 
     if (getIndexBase()) {
@@ -838,12 +837,16 @@ LogicalResult ArgCompareOp::verify() {
   }
 
   TypeRange resultTypes = getResultTypes();
-  if (resultTypes[0] != outputValueType) {
-    return op->emitOpError("result type 0 doesn't match output value type");
+  if (resultTypes[0] != initValueType) {
+    return op->emitOpError("result type 0 doesn't match init value type. ")
+           << "Result type: " << resultTypes[0]
+           << ", init value type: " << initValueType;
   }
 
-  if (resultTypes[1] != outputIndexType) {
-    return op->emitOpError("result type 1 doesn't match output index type");
+  if (resultTypes[1] != initIndexType) {
+    return op->emitOpError("result type 1 doesn't match init index type. ")
+           << "Result type: " << resultTypes[1]
+           << ", init index type: " << initIndexType;
   }
 
   return success();
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
@@ -689,23 +689,7 @@ Type TransferGatherOp::getExpectedMaskType() {
 LogicalResult ArgCompareOp::verify() {
   Operation *op = getOperation();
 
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +242,141 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
   let hasVerifier = 1;
 }
 
+def IREEVectorExt_ArgCompareOp : IREEVectorExt_PureOp<"arg_compare", [
```
IanWood1, line 275: You need `SingleBlockImplicitTerminator<"::mlir::iree_compiler::IREE::LinalgExt::YieldOp">` or to manually verify.


Actually, this would require adding LinalgExt as a dependency which would be a bit strange. This might need a new terminator.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/test/invalid.mlir
```diff
@@ -81,3 +81,260 @@ func.func @indexing_map_invalid_index_vector_shape(%indices: vector<128x64xindex
...
+func.func @arg_compare_no_inputs(%out_val: vector<4xf32>,
+                                 %out_idx: vector<4xi32>)
+    -> (vector<4xf32>, vector<4xi32>) {
+  // expected-error @+1 {{expected 1 or 2 input operands, but got 0}}
+  %result:2 = iree_vector_ext.arg_compare
+      dimension(0)
+      outs(%out_val, %out_idx : vector<4xf32>, vector<4xi32>) {
+    ^bb0(%a: f32, %b: f32):
+      %cmp = arith.cmpf ogt, %a, %b : f32
+      iree_linalg_ext.yield %cmp : i1
```
IanWood1: It's a bit strange to use an iree_linalg_ext op here.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +242,141 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+    Variadic<AnyVectorOfAnyRank>:$results
+  );
+
+  let regions = (region SizedRegion<1>:$region);
+
+  let assemblyFormat = [{
+    attr-dict
+    `dimension` `(` $dimension `)`
+    (`ins` `(` $inputs^ `:` type($inputs) `)`)?
+    `outs` `(` $outputs `:` type($outputs) `)`
```
IanWood1: Can these be renamed since this isn't a DPS op. Maybe `acc` or `init` instead of outs?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
bangtianliu: Yeah, I added `IREEVectorExt_YieldOp` in the latest commit.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,180 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+    The `index_base` is optional and is preserved from the tensor version
+    when applicable.
+
+    The `inits` operands provide the initial accumulator values for the
+    reduction (initial max/min value and index).
+  }];
+
+  let arguments = (ins
+    Variadic<AnyVectorOfAnyRank>:$inputs,
+    Variadic<AnyVectorOfAnyRank>:$inits,
```
IanWood1: Would it be possible to have named inputs and inits instead of using variadic (with optional input). This may cut down some of the verification logic and reduce the number of `extraClassDeclaration`.

IanWood1 (commented):
LGTM in terms of correctness, but I can't speak to the broader design/lowering.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,180 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+  SingleBlockImplicitTerminator<"::mlir::iree_compiler::IREE::VectorExt::YieldOp">
+]> {
+  let summary = "Vectorized arg-reduction using a user-defined comparator.";
+  let description = [{
+    The `iree_vector_ext.arg_compare` op is the vectorized form of
+    `iree_linalg_ext.arg_compare`. It performs a reduction over a given
+    dimension of a vector, returning both the selected value and its
+    corresponding index.
+
+    This operation is created by vectorizing `iree_linalg_ext.arg_compare`
```
qedawkins: This comment is about the lowering, we don't need it as a part of the op's semantics description.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,180 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+    through the GenericVectorizationPass. It wraps the operation with
+    vector.transfer_read/write but keeps the core reduction semantics intact.
+
+    The comparator region defines the selection rule (e.g., "greater than"
+    for argmax or "less than" for argmin).
```
qedawkins: Can you expand on the semantics for the execution for the comparator region here? I'd assume there are restrictions on the kinds of operations allowed inside (only pure ops maybe?)

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,180 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+      ins(%partial_vals, %partial_idxs : vector<4x32xf32>, vector<4x32xi32>)
+      inits(%out_val, %out_idx : vector<4xf32>, vector<4xi32>) {
+    ^bb0(%a: f32, %b: f32):
+      %cmp = arith.cmpf ogt, %a, %b : f32
+      iree_vector_ext.yield %cmp : i1
+    } -> vector<4xf32>, vector<4xi32>
+    ```
+
+    The `index_base` is optional and is preserved from the tensor version
+    when applicable.
```
qedawkins: Can you describe what `index_base` means here without referencing the tensor version.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
qedawkins: +1 lets definitely do that. The LinalgExt approach to having unnamed variadics for all operands is really poor.

qedawkins (commented):
Ok a few notes about the op description, otherwise the definition here makes sense to me. Definite +1 to Ian's comment about the tablegen though, lets not import whatever is going on in LinalgExt.

qedawkins (approved):

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +682,161 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+  ArrayRef<int64_t> inputShape = inputType.getShape();
+  SmallVector<int64_t> expectedShape;
+  for (int64_t i = 0; i < inputRank; ++i) {
+    if (i != dimension) {
+      expectedShape.push_back(inputType.getDimSize(i));
+    }
+  }
+
+  ArrayRef<int64_t> initValueShape = initValueType.getShape();
+  if (!llvm::equal(expectedShape, initValueShape)) {
```
kuhar:
```suggestion
  if (expectedShape != initValueShape) {
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +682,161 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+
+  ArrayRef<int64_t> inputShape = inputType.getShape();
+  SmallVector<int64_t> expectedShape;
+  for (int64_t i = 0; i < inputRank; ++i) {
+    if (i != dimension) {
+      expectedShape.push_back(inputType.getDimSize(i));
+    }
+  }
+
+  ArrayRef<int64_t> initValueShape = initValueType.getShape();
```
kuhar, line 754: Can we move shape verification to ODS?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +682,161 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+  if (!llvm::equal(expectedShape, initValueShape)) {
+    return op->emitOpError(
+               "init value shape must match input shape with reduction "
+               "dimension removed. ")
+           << "Expected: " << llvm::interleaved_array(expectedShape)
+           << ", but got: " << llvm::interleaved_array(initValueShape);
+  }
+
+  ArrayRef<int64_t> initIndexShape = initIndexType.getShape();
+  if (!llvm::equal(expectedShape, initIndexShape)) {
```
kuhar: same here

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +682,161 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+               "init index shape must match input shape with reduction "
+               "dimension removed. ")
+           << "Expected: " << llvm::interleaved_array(expectedShape)
+           << ", but got: " << llvm::interleaved_array(initIndexShape);
+  }
+
+  Type inputElementType = inputType.getElementType();
+  Type initValueElementType = initValueType.getElementType();
+
+  if (inputElementType != initValueElementType) {
```
kuhar: Also here -- can we move to ODS?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +682,161 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+    return op->emitOpError("input and init value element types must match. ")
+           << "Input type: " << inputElementType
+           << ", init value type: " << initValueElementType;
+  }
+
+  Type initIndexElementType = initIndexType.getElementType();
+  if (!isa<IntegerType, IndexType>(initIndexElementType)) {
```
kuhar, line 773: Can we move to ODS?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +682,161 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+  }
+
+  Type yieldType = terminator->getOperand(0).getType();
+  if (!yieldType.isInteger(1)) {
+    return op->emitOpError("comparator region must yield i1, but got ")
+           << yieldType;
+  }
+
+  Type resultValueType = getResult(0).getType();
+  if (resultValueType != initValueType) {
```
kuhar: Can we move to ODS?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +682,161 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+    return op->emitOpError("result value type doesn't match init value type. ")
+           << "Result type: " << resultValueType
+           << ", init value type: " << initValueType;
+  }
+
+  Type resultIndexType = getResult(1).getType();
+  if (resultIndexType != initIndexType) {
```
kuhar: Can we move to ODS?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,176 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
 }
 
+//===----------------------------------------------------------------------===//
+// Terminator ops.
+//===----------------------------------------------------------------------===//
+
+def IREEVectorExt_YieldOp : IREEVectorExt_PureOp<"yield", [
+    Pure,
+    ReturnLike,
+    Terminator
```
kuhar: I think there's also another trait to check the parent op, something like `ParentOneOf`

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,176 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+
+  let assemblyFormat = "attr-dict ($values^ `:` type($values))?";
+}
+
+//===----------------------------------------------------------------------===//
+// LinalgExt-style ops.
+//===----------------------------------------------------------------------===//
+
+def IREEVectorExt_ArgCompareOp : IREEVectorExt_PureOp<"arg_compare", [
+  Pure,
```
kuhar, line 276: Is this always pure? For example, what if the comparator performs division?

bangtianliu:
cc @kuhar @IanWood1 for another look

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,225 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+  AttrSizedOperandSegments,
+  SingleBlockImplicitTerminator<"::mlir::iree_compiler::IREE::VectorExt::YieldOp">,
+  AllElementTypesMatch<["input_value", "init_value"]>,
+  AllTypesMatch<["init_value", "result_value"]>,
+  AllTypesMatch<["init_index", "result_index"]>,
+  PredOpTrait<"dimension must be in range [0, input_rank)",
+    CPred<"$dimension >= 0 && $dimension < "
+          "::llvm::cast<::mlir::VectorType>($input_value.getType()).getRank()">>,
```
IanWood1: I think this is done just to be safe, but it should be possible to use the normal namespace style. e.g.
```suggestion
          "cast<VectorType>($input_value.getType()).getRank()">>,
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/test/invalid.mlir
```diff
@@ -81,3 +81,243 @@ func.func @indexing_map_invalid_index_vector_shape(%indices: vector<128x64xindex
...
   return %out : vector<128x64xf16>
 }
+
+// -----
+
+func.func @arg_compare_dimension_out_of_bounds(%input: vector<4x128xf32>,
+                                               %out_val: vector<4xf32>,
+                                               %out_idx: vector<4xi32>)
+    -> (vector<4xf32>, vector<4xi32>) {
+  // expected-error @+1 {{failed to verify that dimension must be in range [0, input_rank)}}
```
kuhar: would be nice to print the actual input rank

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -682,6 +683,97 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+    return op->emitOpError(
+               "expected comparator region to yield 1 operand, but got ")
+           << numYieldOperands;
+  }
+
+  Type yieldType = terminator->getOperand(0).getType();
+  if (!yieldType.isInteger(1)) {
+    return op->emitOpError("comparator region must yield i1, but got ")
+           << yieldType;
+  }
```
kuhar: I wonder if yield should verify this itself -- how do other upstream ops handle this?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,225 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+                 "input_value", "init_value",
+                 "ArgCompareOp::getExpectedInitType(::llvm::cast<::mlir::VectorType>($_self), "
+                 "getDimension(), $init_value.getType())">,
+  TypesMatchWith<"init index shape must match input shape with reduction dimension removed",
+                 "input_value", "init_index",
+                 "ArgCompareOp::getExpectedInitType(::llvm::cast<::mlir::VectorType>($_self), "
+                 "getDimension(), $init_index.getType())">,
+  // Indices must be integer or index type.
+  PredOpTrait<"init index must have integer or index element type",
+    CPred<"::mlir::isa<::mlir::IntegerType>("
```
kuhar:
```suggestion
    CPred<"::llvm::isa<::mlir::IntegerType>("
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,225 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+          "::llvm::cast<::mlir::VectorType>($init_index.getType()).getElementType()) || "
+          "::mlir::isa<::mlir::IndexType>("
```
kuhar:
```suggestion
          "::mlir::llvm<::mlir::IndexType>("
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/test/invalid.mlir
```diff
@@ -81,3 +81,243 @@ func.func @indexing_map_invalid_index_vector_shape(%indices: vector<128x64xindex
...
+  return %result#0, %result#1 : vector<4xf32>, vector<4xi32>
+}
+
+// -----
+
+func.func @arg_compare_wrong_output_rank(%input: vector<4x128xf32>,
+                                         %out_val: vector<4x8xf32>,
+                                         %out_idx: vector<4xi32>)
+    -> (vector<4x8xf32>, vector<4xi32>) {
+  // expected-error @+1 {{failed to verify that init value rank must be input rank - 1}}
```
kuhar:
```suggestion
  // expected-error @+1 {{failed to verify that init value rank must be input_rank - 1}}
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,225 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+      return cast<VectorType>(getInputValue().getType());
+    }
+
+    VectorType getInitValueType() {
+      return cast<VectorType>(getInitValue().getType());
+    }
+
+    VectorType getInitIndexType() {
+      return cast<VectorType>(getInitIndex().getType());
+    }
```
IanWood1, line 387: I think you can remove the cast. It should already be a VectorType.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
bangtianliu: Good catch! 

Mainly two approaches there:
  - `scf.yield`: No custom verifier,  parent ops verify everything
  - `linalg.yield`: has `hasVerifier = 1` and implements its own verification:
  
so that I can follow the approach from `linalg.yield` (closest to my use case here), I can move this verification to `iree_vector_ext.yield`.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,225 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+
+  let extraClassDeclaration = [{
+    // Helper for TypesMatchWith: compute expected init type based on input type and reduction dimension.
+    // If dimension is out of bounds, return initType unchanged to let the dimension bounds check fail.
+    static Type getExpectedInitType(VectorType inputType, int64_t dimension, Type initType) {
+      int64_t rank = inputType.getRank();
+      // Return initType if dimension is invalid - let the dimension bounds check catch this.
+      if (dimension < 0 || dimension >= rank) {
+        return initType;
+      }
```
IanWood1: This should probably be asserted or return nullptr. Is this an ordering issue with the tablegen verification?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,225 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+    `ins` `(` $input_value (`,` $input_index^)? `:` type($input_value) (`,` type($input_index)^)? `)`
+    `inits` `(` $init_value `,` $init_index `:` type($init_value) `,` type($init_index) `)`
+    (`index_base` `(` $index_base^ `:` type($index_base) `)`)?
+    $region `->` type($result_value) `,` type($result_index)
+  }];
+
+  let extraClassDeclaration = [{
+    // Helper for TypesMatchWith: compute expected init type based on input type and reduction dimension.
+    // If dimension is out of bounds, return initType unchanged to let the dimension bounds check fail.
+    static Type getExpectedInitType(VectorType inputType, int64_t dimension, Type initType) {
```
IanWood1: Should `initType` be a VectorType?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -242,4 +243,225 @@ def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
...
+  TypesMatchWith<"init index shape must match input shape with reduction dimension removed",
+                 "input_value", "init_index",
+                 "ArgCompareOp::getExpectedInitType(::llvm::cast<::mlir::VectorType>($_self), "
+                 "getDimension(), $init_index.getType())">,
+  // Indices must be integer or index type.
+  PredOpTrait<"init index must have integer or index element type",
+    CPred<"::mlir::isa<::mlir::IntegerType>("
+          "::llvm::cast<::mlir::VectorType>($init_index.getType()).getElementType()) || "
+          "::mlir::isa<::mlir::IndexType>("
+          "::llvm::cast<::mlir::VectorType>($init_index.getType()).getElementType())">>
```
IanWood1: Not blocking, but I don't see the benefit of doing verification in the td file instead of in the verify method. It makes sense in cases where the checks are easily reusable (`SingleBlockImplicitTerminator`). This just seems more difficult to read, especially since there is a mix of c++ and table gen.

I'd be open to hearing the benefits of this approach. I could be missing something.

IanWood1 (approved):

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
bangtianliu: cc @kuhar here

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
bangtianliu: Ok here is linalg.yield 
```C++
LogicalResult linalg::YieldOp::verify() {
  auto *parentOp = (*this)->getParentOp();
  if (parentOp->getNumRegions() != 1 || parentOp->getRegion(0).empty())
    return emitOpError("expected single non-empty parent region");

  if (auto linalgOp = dyn_cast<LinalgOp>(parentOp))
    return verifyYield(*this, linalgOp);

  return emitOpError("expected parent op with LinalgOp interface");
}
```
 linalg.yield verification is split into two parts:
  - General (in YieldOp::verify()): Structural sanity checks (parent has regions, parent is correct type)
  - Specific (in verifyYield() helper): Parent-specific requirements (operand counts, types)
  
 so for my case, it is kind of similar. We have general cases for` iree_vector_ext.yield`,  `yield i1` is specific to ArgCompare Op.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
bangtianliu: Yes, it is about  TableGen verification ordering. 

` PredOpTrait<"dimension must be in range [0, input_rank)", ...>` runs before   `TypesMatchWith<"init value shape must match...", "ArgCompareOp::getExpectedInitType(...)">`.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
kuhar: This split into general and specific makes most sense to me

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
kuhar: I prefer using traits when available, if we have to write CPred and won’t be reusing them across multiple ops it probably makes more sense to keep in ::verify. I don’t remember which traits are generic and available here, so moving some of these back to .cpp may make more sense.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -683,18 +683,108 @@ Type TransferGatherOp::getExpectedMaskType() {
...
+
+  // ParentOneOf ODS trait restricts valid parent operations.
+  if (auto argCompareOp = dyn_cast<ArgCompareOp>(parentOp)) {
+    return verifyYieldForArgCompare(*this, argCompareOp);
+  }
+
+  return emitOpError("unexpected parent op '")
+         << parentOp->getName() << "';"
+         << " iree_vector_ext.yield must be used within a supported operation";
+}
```
kuhar: Isn't this already checked in ODS? `ParentOneOf`

kuhar (approved):
LGTM % one minor issue

PR #21817, opened 2025-09-02
[Codegen][LLVMGPU] Later scf-to-cf to support math.erf 

newling:
Before this PR, `math-transform` ran after `cf-to-scf`. But `math-transform` introduces `scf.if` when there is a `math.erf`. 

This PR moves `scf-to-cf` later, so that the scf.if can be converted. Fixes https://github.com/iree-org/iree/issues/21813

PR #23388, opened 2026-02-04
Move HoistStaticallyBoundAllocations after CF.

amd-eochoalo:
There's a comment that says that this pass must run after CF lowering, but it is actually ran before CF lowering. PR #21817 7cdfc13c36578453f sank CF lowering. PR #23369 86395216 revealed the presence of a bug in a convolution test and was bisected to this change.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -1027,9 +1027,6 @@ static void addLowerToLLVMGPUPasses(OpPassManager &modulePassManager,
         auto getIndexBitwidth = [](mlir::FunctionOpInterface) { return 64; };
         return createGPUCheckResourceUsagePass(getIndexBitwidth);
       })
-      // Hoist allocations back into the entry block, as lowering to CF may have
-      // split the block at a point before the allocation.
-      .addPass(createHoistStaticallyBoundAllocationsPass)
       // Handle complex operation conversion.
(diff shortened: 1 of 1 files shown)


amd-eochoalo:
`test_torch_ops` on gfx1201 is now passing [here](https://github.com/iree-org/iree/actions/runs/21691842053/job/62558705175?pr=23388). `test_torch_ops` on gfx1201 on main is [failing](https://github.com/iree-org/iree/actions/runs/21684851518/job/62546216122).

amd-eochoalo:
[Here](https://gist.github.com/amd-eochoalo/1e31375fbb38ab4fb832b3f69e1adf40) is the gist after CSE with the bad IR. [Here](https://gist.github.com/amd-eochoalo/f91887d9ba9df0ebd706adab027f41d3) is the gist after hoisting static allocations with the good IR. And here is the [diff](https://gist.github.com/amd-eochoalo/c89fa0a39ce73f2b16b41e0495e0f70d).

I am not yet 100% sure why there is an error, I know that allocas must be in the entry point in order to be static and allocas in other basic blocks are considered dynamic and can grow the stack which can lead to stack overflows when allocating in a loop. But this is a numeric issue, so maybe it has something else to do with it being executed on a GPU (instead of on a CPU) or it leads to an error in a different optimization further along the pipeline.

Muzammiluddin-Syed-ECE (approved):

amd-eochoalo:
@Muzammiluddin-Syed-ECE thanks for the review! :)

PR #23392, opened 2026-02-05
[Stream] Fix EmplaceTransientsPass to handle zero allocas case.

### LINKED ISSUE #23384: EmplaceTransientsPass leaves orphaned stream.resource.transients when no allocas exist
## Description

When using `--iree-torch-externalize-transients` with a graph that has no intermediate transient allocations (e.g., a single-dispatch graph writing directly to output), compilation fails with:

```
error: failed to legalize operation 'stream.resource.transients' that was explicitly marked illegal
```

## Root Cause

The `--iree-torch-externalize-transients` flag creates `stream.resource.transients` ops during torch-to-IREE conversion. However, `EmplaceTransientsPass` only processes these ops when there are `stream.resource.alloca` ops to emplace into the user-provided buffer.

When there are no allocas (because all outputs write directly to user-provided buffers), the `stream.resource.transients` op is left behind and fails to lower to HAL.

## Expected Behavior

Compilation should succeed with `iree.abi.transients.size.constant = 0` when no transient allocations are needed.

## Reproduction

**IREE version:** `3.11.0rc20260203 @ ca60cabdcd1615499706a6a2c951b41f541e969b`

**Repro for GPU backend:** https://gist.github.com/sjain-stanford/d1a1ccd4d11764ccef10cdec2b693787

**Repro for CPU backend:** https://gist.github.com/sjain-stanford/add9a9414a40c5054ff596e598007355

## Notes

- Multi-dispatch graphs that DO have intermediate allocations compile successfully with the flag
- The issue is specifically when the graph has zero transient allocations

## Suggested Fix

Either:
1. `EmplaceTransientsPass` should remove `stream.resource.transients` ops when there are no allocas to emplace
2. Or the torch plugin should not create the op when transient size would be 0
3. Or add a fallback lowering pattern that removes the op when unused

MaheshRavishankar:
When using `--iree-torch-externalize-transients` with a graph that has no intermediate transient allocations, the `stream.resource.transients` op was left behind and failed to lower to HAL with:
  "failed to legalize operation 'stream.resource.transients'"

The fix removes the transients ops by forwarding their resource and timepoint values when there are no allocas to emplace.

PR Co-authored by using Claude-Opus-4.6

Fixes #23384

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/EmplaceTransients.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/EmplaceTransients.cpp
@@ -1278,9 +1278,38 @@ processTransientResource(FunctionOpInterface funcOp,
                          TransientResource &transientResource,
                          Explorer &explorer) {
   if (transientResource.allocaOps.empty()) {
-    // No allocations - just remove the transients ops without creating a pack.
+    // No allocations - create a trivial pack op with zero slices so that the
+    // MaterializeTransientSizeQueriesPass can still generate a size query
+    // function (which will return 0). This ensures consistent ABI behavior.
     LLVM_DEBUG(LDBG() << "no allocations found, "
-                         "removing transients ops\n");
+                         "creating trivial zero-size pack\n");
+
+    // Find the first transients op to use as insertion point.
+    auto firstTransientsOp = transientResource.transientsOps.front();
+    OpBuilder builder(firstTransientsOp);
+
+    // Create a pack op with zero slices - it will fold to total_length = 0.
+    auto indexType = builder.getIndexType();
+    SmallVector<int64_t> lifetimeIntervals; // Empty for zero slices.
+    SmallVector<Value> sliceSizes;          // Empty for zero slices.
+    SmallVector<Type> packedOffsetTypes;    // Empty for zero slices.
+    auto packOp = IREE::Stream::ResourcePackOp::create(
+        builder, firstTransientsOp.getLoc(), indexType, packedOffsetTypes,
+        /*offset=*/nullptr, builder.getIndexArrayAttr(lifetimeIntervals),
+        sliceSizes, transientResource.affinity);
+
+    // Add the stream.experimental.transients attribute so
+    // MaterializeTransientSizeQueriesPass can find this pack op and generate
+    // a size query function.
+    packOp->setAttr("stream.experimental.transients", builder.getUnitAttr());
+
+    LLVM_DEBUG({
+      LDBG() << "created trivial pack op: ";
+      packOp->print(llvm::dbgs(), explorer.getAsmState());
+      llvm::dbgs() << "\n";
+    });
+
+    // Remove the transients ops by forwarding their values.
     for (auto transientsOp : transientResource.transientsOps) {
       // Forward the result timepoint to the await timepoint.
       transientsOp.getResultTimepoint().replaceAllUsesWith(
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/EmplaceTransients.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/EmplaceTransients.cpp
@@ -1278,10 +1278,17 @@ processTransientResource(FunctionOpInterface funcOp,
                          TransientResource &transientResource,
                          Explorer &explorer) {
   if (transientResource.allocaOps.empty()) {
-    // No allocations - valid case (e.g., zero allocations with transients
-    // annotation).
+    // No allocations - just remove the transients ops without creating a pack.
     LLVM_DEBUG(LDBG() << "no allocations found, "
-                         "nothing to pack\n");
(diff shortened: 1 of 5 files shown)


zjgarvey (commented):
My only concern with this approach is that without the pack ops, we won't be generating a size query function during the size query materialization pass. 

We could "try" to grab the attribute from the reflection, and if we fail, we can pass an empty buffer, but it may be better to still generate the size query and make sure it's returning zero. 

Can we instead try to generate a trivial pack op if the transients aren't used, so at least we generate the size query? I'm not 100% familiar with this dialect, so I'm not sure if that's the right move or not.

tests/e2e/regression/zero_transients.mlir
```diff
@@ -0,0 +1,14 @@
...
+    %arg2 : tensor<128x512xf32> {iree.abi.output  = 0 : index}) -> tensor<128x512xf32> {
+  %cst = arith.constant 0.0 : f32
+  %0 = tensor.empty() : tensor<128x512xf32>
+  %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<128x512xf32>) -> tensor<128x512xf32>
+  %2 = linalg.matmul ins(%arg0, %arg1 : tensor<128x256xf32>, tensor<256x512xf32>)
+      outs(%1 : tensor<128x512xf32>) -> tensor<128x512xf32>
+  util.return %2 : tensor<128x512xf32>
+}
+//      CHECK: transient-memory-size
+// CHECK-SAME: 0
```
sjain-stanford: Thanks for adding this here. Will add the same in Fusilli.

tests/e2e/regression/zero_transients.mlir
```diff
@@ -0,0 +1,14 @@
+// RUN: iree-compile --iree-hal-target-backends=rocm --iree-rocm-target=mi300x --iree-opt-level=O3 --iree-scheduling-dump-statistics-file=- --iree-scheduling-dump-statistics-format=json --compile-to=stream %s | FileCheck %s
+
+// For this case no transient memory allocations should be required.
```
zjgarvey: I'm not sure I understand this one. We don't have an input annotated with `iree.abi.transients` so why would this get a transients size in the first place?

Ah, even if we don't have any transients ops at all, then we are still generating the pack op? Personally, I think there should be a transients size query if and only if there is a transients buffer as a function argument. For now, I think just restricting to when a transients op is present would be good.

compiler/src/iree/compiler/Dialect/Stream/Transforms/EmplaceTransients.cpp
```diff
@@ -1278,10 +1278,46 @@ processTransientResource(FunctionOpInterface funcOp,
...
-    // annotation).
+    // No allocations - create a trivial pack op with zero slices so that the
+    // MaterializeTransientSizeQueriesPass can still generate a size query
+    // function (which will return 0). This ensures consistent ABI behavior.
     LLVM_DEBUG(LDBG() << "no allocations found, "
-                         "nothing to pack\n");
+                         "creating trivial zero-size pack\n");
+
+    // Find the first transients op to use as insertion point.
+    auto firstTransientsOp = transientResource.transientsOps.front();
```
zjgarvey, line 1288: What happens if this is empty (as in the lit test in `zero_transients.mlir`?

zjgarvey (commented):
I think it would make sense to only generate the transients pack if we actually have transient buffers in the first place. Otherwise, this looks good to me.

tests/e2e/regression/zero_transients.mlir
MaheshRavishankar: I dont think I fully understand the comment. First this test was just a guard against us not using any transients in a case we dont need it. Having said that, I tried to study a bit of what happens here. This change only kicks in when there is buffer passed in with transients pre-allocated. For that case it generates an empty pack operation which later becomes size 0. So I tried these four combinations

1. No transients needed + no external buffer passed in for transients --> Works as expected. Nothing happens with this change.
2. No transients needed + external buffer passed in for transients --> Generates an empty pack operation which later translates to size of 0
3. Transients needed + no external buffer passed in for transients --> Works as expected. Nothing happens with this change.
4. Transients needed + external buffer passed in for transients --> Works as expected. Nothing happens with this change.


Happy to share the IR dump with you if you want.

compiler/src/iree/compiler/Dialect/Stream/Transforms/EmplaceTransients.cpp
MaheshRavishankar: I dont think this code path is executed if this is empty.

zjgarvey: Ah, I see. This function only gets called on existing transient resources. Makes sense.

I think I got confused by the earlier lit test. Thanks for adding more.

zjgarvey (approved):
This looks good to me, thanks.

PR #23393, opened 2026-02-05
[GlobalOpt] Add NHWC_FHWC support to 1x1 conv to matmul pass

HendrikKlug-synthara:
For FHWC filter layout [Cout, Kh, Kw, Cin], a transpose is required to reorder dimensions to [Cin, Kh, Kw, Cout] so that after unit extent dims are folded, the filter has the correct [Cin, Cout] shape for matmul.

### DIFF
--- a/compiler/src/iree/compiler/GlobalOptimization/Convert1X1FilterConv2DToMatmul.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/Convert1X1FilterConv2DToMatmul.cpp
@@ -7,6 +7,8 @@
 #include "iree/compiler/GlobalOptimization/Passes.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/IR/AffineExpr.h"
 #include "mlir/IR/PatternMatch.h"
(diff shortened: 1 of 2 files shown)


yzhang93:
I'm not sure what is the original usage of this pass. From the old tests, it's not just converting to a strict meaning "matmul", but any generic form of gemm it corresponds to (examples are nchw layout). P.S. I believe this pass no longer belongs to GlobalOpt, as it's only used in preprocessing. CC: @IanWood1

HendrikKlug-synthara:
This is currently useful for me because when importing tflite networks to tosa, linear layers get translated to 1x1 conv ops. I'm using this pass to convert the 1x1 convs to matmul ops (converting the linalg.generic to linalg.matmul using `linalg-specialize-generic-ops`)

HendrikKlug-synthara:
Hey, @yzhang93 looks like you moved this pass in 922310651fafc4d16668bb14739e49069914d362 Should I rebase this patch or Is adding support for other filter layouts not wanted upstream?

yzhang93:
> Hey, @yzhang93 looks like you moved this pass in [9223106](https://github.com/iree-org/iree/commit/922310651fafc4d16668bb14739e49069914d362) Should I rebase this patch or Is adding support for other filter layouts not wanted upstream?

IMO, adding a transpose in this pass is slightly out of scope. Using this pass, the convolution should lower directly to the corresponding matmul-like op without modifying the order of dimensions in affine maps. Previously, we handled transposes in separate passes, such as `ConvertConvToChannelsLastPass` or `ConvertConvFilterToChannelsLastPass`.

PR #23394, opened 2026-02-05
[VectorDistribute] Consider all compute ops for thread tile size

### LINKED ISSUE #23340: LayerNorm NHWC with scale and bias hangs at the compilation stage
### What happened?

The MLIR can be found [here](https://gist.github.com/a-sidorova/54f712e91b6214c28f4d0f658e1eab9c).

Command line:
```
iree-compile layernorm_nhwc.mlir --iree-hal-target-backends=rocm \
                     --iree-hip-target=gfx950 --iree-opt-level=O3 -o iree-compile-output.vmfb
```

The compilation process is stuck. No responses from the compiler.

Noticed that it happens when `C < 8`.

For example, in Fusilli we have timeout in 30 seconds. The compilation process definitely doesn't fit into that time frame.

IREE Compiler

### Version information

Can be successfully reproduced on the IREE commits: https://github.com/iree-org/iree/commit/18fe0926981becffb1332ed727aaf916a9727b13 (30/01/26) and https://github.com/iree-org/iree/commit/4031503a5536c232c5dc0a0412bbf7086db38207 (14/01/26).

### Additional context

_No response_

sommerlukas:
So far, selection of the thread tile size (`threadLoads`) in reduction VectorDistribute lowering strategy selection was purely based on the root operation. If the selected size wasn't suitable for other compute operations in the same dispatch, selection of the VectorDistribute strategy would fail, falling back to other strategies.

This change addresses an existing TODO to consider the constraints of the other compute ops in the same dispatch when choosing the thread tile size. This may lead to smaller tile sizes (worst case one element per thread) if the other compute ops have additional constraints, but allows to use the VectorDistribute pipeline for such reductions.

This fixes https://github.com/iree-org/iree/issues/23340.

Assisted-by: Claude Code

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
@@ -741,6 +741,9 @@ LogicalResult setReductionConfig(IREE::GPU::TargetAttr target,
         threadLoads /= 2;
       }
     }
+    if (threadLoads <= 1) {
+      break;
+    }
   }
 
   std::optional<int64_t> parallelSize = 1;
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
@@ -676,7 +676,7 @@ LogicalResult setReductionConfig(IREE::GPU::TargetAttr target,
   // reductions in scaled matmul with the last dimension being the block size
   // (32 for gfx950).
   int64_t reductionSize = bounds[reductionDims.back()];
-  if (!ShapedType::isDynamic(reductionSize) &&
+  if (ShapedType::isStatic(reductionSize) &&
       reductionSize % target.getPreferredSubgroupSize() != 0) {
     // Consider the entire reduction dimension.
     reductionSize = 1;
@@ -736,7 +736,7 @@ LogicalResult setReductionConfig(IREE::GPU::TargetAttr target,
   // dispatch.
   for (linalg::LinalgOp linalgOp : *computeOps) {
     int64_t constraint = getThreadLoadsConstraint(linalgOp);
-    if (!ShapedType::isDynamic(constraint)) {
+    if (ShapedType::isStatic(constraint)) {
       while (threadLoads > 1 && constraint % threadLoads != 0) {
         threadLoads /= 2;
       }
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
@@ -130,6 +130,21 @@ static FailureOr<int64_t> getBitWidth(linalg::LinalgOp op) {
   return largestInput.elementBitwidth;
 }
 
+/// Returns the dimension size that threadLoads must divide for the given op.
+static int64_t getThreadLoadsConstraint(linalg::LinalgOp op) {
+  SmallVector<unsigned> parallelDims, reductionDims;
+  op.getParallelDims(parallelDims);
+  op.getReductionDims(reductionDims);
+  SmallVector<int64_t> bounds = op.getStaticLoopRanges();
+
+  if (reductionDims.empty()) {
+    // Parallel-only op: threadLoads must divide last parallel dim
+    return bounds[parallelDims.back()];
+  }
+  // Reduction op: threadLoads must divide last reduction dim
+  return bounds[reductionDims.back()];
+}
+
 /// Check if the reduction op has a single combiner operation.
 static LogicalResult checkSingleCombiner(linalg::LinalgOp op) {
   bool foundSingleReductionOutput = false;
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_vector_distribute_reduction_gfx942.mlir
```diff
@@ -179,78 +179,78 @@ func.func @test_multiple_reduction() {
 
 // -----
 
-// The same IR as 'test_multiple_reduction' but with an unsupported operation,
```
sommerlukas, line 182: The rendering of the diff isn't great, but this essentially removes the negative test from https://github.com/iree-org/iree/pull/23228 (which is not a negative test anymore, as we now support this scenario) and adds a test based on https://github.com/iree-org/iree/issues/23340.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -717,6 +732,17 @@ LogicalResult setReductionConfig(IREE::GPU::TargetAttr target,
     }
   }
 
+  // Adjust threadLoads to satisfy constraints from all compute ops in the
+  // dispatch.
+  for (linalg::LinalgOp linalgOp : *computeOps) {
+    int64_t constraint = getThreadLoadsConstraint(linalgOp);
+    if (!ShapedType::isDynamic(constraint)) {
```
efric:
```suggestion
    if (ShapedType::isStatic(constraint)) {
```

efric (approved):
Thanks, this looks good to me. We have something similar in [SPIRV](https://github.com/iree-org/iree/blob/133860abfd355f99559558bb20449c9279c6f16b/compiler/src/iree/compiler/Codegen/SPIRV/KernelConfig.cpp#L1230) and it makes sense to do the same here. I saw the alternative solutions you proposed in the issue and agree that this is the best minimally invasive solution.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_vector_distribute_reduction_gfx942.mlir
```diff
@@ -179,78 +179,78 @@ func.func @test_multiple_reduction() {
...
+  %2 = hal.interface.binding.subspan layout(#pipeline_layout) binding(2) alignment(64) offset(%c0) flags("ReadOnly|Indirect") : !iree_tensor_ext.dispatch.tensor<readonly:tensor<16384x3xbf16>>
+  %3 = hal.interface.binding.subspan layout(#pipeline_layout) binding(3) alignment(64) offset(%c0) flags(Indirect) : !iree_tensor_ext.dispatch.tensor<writeonly:tensor<2x16384x3xbf16>>
+  %4 = iree_tensor_ext.dispatch.tensor.load %0, offsets = [0, 0, 0], sizes = [2, 16384, 3], strides = [1, 1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<2x16384x3xbf16>> -> tensor<2x16384x3xbf16>
+  %5 = iree_tensor_ext.dispatch.tensor.load %1, offsets = [0, 0], sizes = [16384, 3], strides = [1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<16384x3xbf16>> -> tensor<16384x3xbf16>
+  %6 = iree_tensor_ext.dispatch.tensor.load %2, offsets = [0, 0], sizes = [16384, 3], strides = [1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<16384x3xbf16>> -> tensor<16384x3xbf16>
+  %7 = tensor.empty() : tensor<2xbf16>
+  %8 = linalg.fill ins(%cst : bf16) outs(%7 : tensor<2xbf16>) -> tensor<2xbf16>
```
kuhar, line 208: BTW, any chance to simplify this test by moving these to function arguments? (Could be a separe PR)

kuhar (approved):

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -717,6 +732,17 @@ LogicalResult setReductionConfig(IREE::GPU::TargetAttr target,
     }
   }
 
+  // Adjust threadLoads to satisfy constraints from all compute ops in the
+  // dispatch.
+  for (linalg::LinalgOp linalgOp : *computeOps) {
+    int64_t constraint = getThreadLoadsConstraint(linalgOp);
+    if (ShapedType::isStatic(constraint)) {
+      while (threadLoads > 1 && constraint % threadLoads != 0) {
+        threadLoads /= 2;
```
keshavvinayak01, line 741: nitpick
Can exploit bit manipulation if threadloads is always power of 2, something like:
`threadLoads =
        std::min<int64_t>(threadLoads, constraint & (-constraint));`

keshavvinayak01 (approved):

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
keshavvinayak01: And a 
```
if (threadLoads <= 1)
    break;
```
before.

sommerlukas: Thanks for the feedback. I've added the early end of the loop as you suggested.

As for the other suggestion, I'd prefer readability here and leave bit-manipulation tricks to the compiler. I'll still understand what `/= 2` means in six months, I'm not so sure the bit-manipulation will be obvious to a future reader of this code, me included.

krzysz00: Agreed on the readability there - turning power-of-two divisions into bit manipulation is what -O2 and -O3 are for, no?

sommerlukas:
Currently holding off on merging this PR as @Groverkss plans to address this as part of an overhaul of the vector distribute reduction configuration logic overhaul.

Groverkss (approved):
I don't think this is the correct solution but ok to land as a hack. We are at the limit of what the reduction configuration can do. I'm working on some fixes to improve this.

PR #23397, opened 2026-02-05
[StableHLO][LinalgExt] Fix suffix scan lowering in StableHLO reduce_window conversion

### LINKED ISSUE #23381: Incorrect results coming from cumsum
### What happened?

We have a model written in JAX which is exported to StableHLO and then compiled and run in IREE. We noticed that the output from JAX didn't match the output from IREE. After some investigation, we found that switching `jnp.cumsum` to an equivalent matmul fixed the problem. So we suspect that the `cumsum` is somehow being compiled incorrectly in IREE.

I have included a small repro case. There are 2 identical models except that one uses `cumsum` and the other uses `matmul`. The models have 3 outputs: `trajectory`, `loss`, and `gradient`. The models are StableHLO MLIR files and can be run using IREE or `stablehlo-translate`. I have included the outputs for JAX, `stablehlo-translate`, and IREE for reference. Note that all outputs for both models in all three environments agree except that IREE generates an almost entirely zero tensor for `gradient` in the `cumsum` model.

Given that the same MLIR file gives correct results using `stablehlo-translate` but not when compiled and run in IREE, we think that this likely an IREE compiler codegen issue.

The following can be used to compile and run in IREE
```
iree-compile \
    --mlir-print-debuginfo \
    --iree-hal-executable-debug-level=3 \
    --iree-input-type=auto \
    --iree-hal-target-device=local \
    --iree-hal-local-target-device-backends=llvm-cpu \
    --iree-llvmcpu-target-cpu=host \
    -o "mlir_cumsum/decode_forward_grad_cumsum.vmfb" "mlir_cumsum/decode_forward_grad_cumsum.mlir"

iree-run-module --device=local-sync \
                --function=main \
                --module="mlir_cumsum/decode_forward_grad_cumsum.vmfb" \
                --input="@inputs/action.npy" \
                --input="@inputs/init.npy" \
                --input="@inputs/goal.npy" \
                --output="@mlir_cumsum/iree_outputs/trajectory.npy" \
                --output="@mlir_cumsum/iree_outputs/loss.npy" \
                --output="@mlir_cumsum/iree_outputs/gradient.npy"
```
A similar command can be used for the matmul version.

The following command can be used to generate the same outputs using the StableHLO interpreter (alas `stablehlo-translate` doesn't accept NumPy files as input):
```
INPUT="[dense<[[1.916238814592E-01, 2.310134619474E-01, 2.436317652464E-01, 1.996245682240E-01, 2.110414803028E-01, 1.060300841928E-01, 2.544803619385E-01, 2.929546535015E-01, 8.231372237206E-01, 7.897616028786E-01, 5.683182179928E-02, 6.374107003212E-01, 1.192909240723E+00, 1.778607219458E-01, -1.271262168884E+00, -8.249371647835E-01, -3.690034747124E-01, 1.177057361603E+01, 7.209615230560E+00, -1.795608162880E+00, -2.729878234863E+01, 2.818708717823E-01, -9.266345977783E+00, 7.578706264496E+00, -9.665021896362E+00, -9.124464416504E+01, 7.105489349365E+01, 5.542745971680E+01, 6.869965362549E+01, -8.773067474365E+01, 2.022275733948E+01, 4.710940933228E+01],[-7.788905641064E-04, 1.656853593886E-02, -1.245126873255E-02, -3.119949717075E-03, 1.641701720655E-02, 9.237653762102E-02, -1.862861067057E-01, 4.323945567012E-02, 2.964200675488E-01, -1.089849472046E+00, -4.397805035114E-01, 6.102857589722E-01, 2.287607908249E+00, 4.556356370449E-01, -1.335853695869E+00, 2.484302520752E+00, 1.525308990479E+01, -4.617755413055E+00, 1.275899982452E+01, 1.945725822449E+01, -8.071454048157E+00, -1.477600097656E+01, -5.644944667816E+00, 3.242681980133E+00, -4.340805053711E+01, -5.501557922363E+01, 6.276083755493E+01, -1.089413452148E+01, -1.711965942383E+01, 9.033758544922E+01, -4.330717468262E+01, 4.940369033813E+01],[2.306439448148E-03, 1.671823300421E-02, -1.738532260060E-02, 3.972386941314E-03, -5.823666602373E-02, -1.392383277416E-01, 3.667024970055E-01, -2.757268212736E-02, 7.174304127693E-01, -1.721415519714E-01, -7.506300210953E-01, 2.048193812370E-01, 1.026321053505E+00, -1.302021384239E+00, 1.717482566833E+00, -9.351584315300E-01, 5.545004367828E+00, 2.200427055359E+01, -6.222821712494E+00, 4.278964042664E+00, -7.285795211792E-01, -5.548461914062E+00, 1.764404106140E+01, -2.333568954468E+01, 4.029774475098E+01, 1.705682754517E+01, 6.596599578857E+01, 5.522461414337E+00, -1.093027877808E+02, -1.149425048828E+02, -1.330261077881E+02, -5.921913909912E+01]]> : tensor<3x32xf32>, dense<[0.0, 0.0, 0.0, 0.0, 0.0, 1.0]> : tensor<6xf32>, dense<[1.0, 0.0]> : tensor<2xf32>]"

stablehlo-translate "mlir_cumsum/decode_forward_grad_cumsum.mlir" --interpret --probe-output-dir="mlir_cumsum/slo_outputs" --args="$INPUT"
```
(Also note that I left some commented out `interpreter.probe` instructions in the models to simplify getting outputs from `stablehlo-translate`).

Here's the repro case:
[cumsum_bug.zip](https://github.com/user-attachments/files/25079433/cumsum_bug.zip)

1. Unzip repro
2. Note that compiling and running `decode_forward_grad_cumsum.mlir` gives the wrong `gradient` output when using IREE.

Compiler

### Version information

3.10 (although we've seen the same issue in 3.5 and 3.9)

### Additional context

_No response_

keshavvinayak01:
Fixes #23381

The gradient computation uses a suffix scan (reverse accumulation), which `stablehlo.reduce_window` represents via postfix padding ([0, N] instead of [N, 0]). The existing `ScanOpConversion` pattern correctly detected `isPostfix` but still mapped it directly to a forward `IREE::LinalgExt::ScanOp`.


keshavvinayak01:
@MaheshRavishankar @rsuderman Can I get some reviews on this?

MaheshRavishankar (commented):
I dont know off the top of my head what reduce_window represents in StableHLO. Isnt this more similar to pooling operation?

keshavvinayak01:
> I dont know off the top of my head what reduce_window represents in StableHLO. Isnt this more similar to pooling operation?

Yes, it's pooling. But the JAX configuration in the original issue lowers `cumsum` to `reduce_window` with stride=1 and a full window, making it a prefix sum effectively.

The above works, but the back-propagation algorithm would require a working reverse accumulation for the gradient. Since it was producing an empty tensor, I tracked it down here and simply ran the prefix sum on a reversed input to make it worked.

compiler/plugins/input/StableHLO/Conversion/StableHLOToLinalgExt.cpp
```diff
@@ -567,10 +567,29 @@ struct ScanOpConversion final
       outputTys.push_back(output.getType());
     }
 
+    // IREE::LinalgExt::ScanOp only supports purely inclusive (prefix) scans.
```
MaheshRavishankar, line 570: I am fine with this solution, and I am fine with landing this as a stop-gap, but ideally, we can have the scan op handle postfix as well.

MaheshRavishankar (approved):

PR #23398, opened 2026-02-05
[VectorDistribute] Support permutation in single transfer-read distribution

sommerlukas:
Extend the new pattern from https://github.com/iree-org/iree/pull/23072 that distributes `transfer_reads` by reshaping the memref and issuing a single `transfer_read` to now also handle permutations. In https://github.com/iree-org/iree/pull/23072, the pattern still bailed out if the permutation map on the `transfer_read` was not a minor identity map. With the changes from this PR, the pattern now supports non-identity permutations and broadcast.


sommerlukas:
This PR is currently marked as draft, as it's stacked on top of https://github.com/iree-org/iree/pull/23072.

sommerlukas:
Closing this PR as there's a plan to use a different approach involving `transfer_gather` instead.

PR #23399, opened 2026-02-05
Add additional search paths for tools and bitcode

AaronStGeorge:
When `libIREECompiler.so` is distributed as part of [`TheRock`](https://github.com/ROCm/TheRock) tools and bitcode will be in different default locations. Add additional fallback search paths corresponding to install locations used in `TheRock`.

### DIFF
--- a/compiler/src/iree/compiler/Utils/ToolUtils.cpp
+++ b/compiler/src/iree/compiler/Utils/ToolUtils.cpp
@@ -286,7 +286,7 @@ std::string findPlatformLibDirectory(StringRef platformName) {
   }
 
   SmallString<256> libPath(dylibPath);
-  // Trim patth to lib dir: some/path/lib/libIREECompiler.so -> some/path/lib
+  // Trim path to lib dir: some/path/lib/libIREECompiler.so -> some/path/lib
   llvm::sys::path::remove_filename(libPath);
 
   // Try IREE's convention: lib/iree_platform_libs/<platform>/
--- a/compiler/src/iree/compiler/Utils/ToolUtils.cpp
+++ b/compiler/src/iree/compiler/Utils/ToolUtils.cpp
@@ -187,6 +187,19 @@ std::string findToolFromDylibDir(SmallVector<std::string> toolNames) {
     return toolPath;
   }
 
+  // Then search in an nested llvm/bin/ directory. This is the ROCm/TheRock
+  // standard location:
+  //   lib/
+  //     libIREECompiler.so
+  //     llvm/bin/
+  //       lld
+  toolPath = findToolAtPath(normalizedToolNames, dylibDir + "/llvm/bin/");
+  if (!toolPath.empty()) {
+    LLVM_DEBUG(llvm::dbgs() << "Found tool in ROCm llvm/bin directory at path "
+                            << toolPath << "\n");
+    return toolPath;
+  }
+
   // Then search in an adjacent tools/ directory. Build trees are
   // organized this way for reasons:
   //   lib/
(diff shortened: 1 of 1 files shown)


compiler/src/iree/compiler/Utils/ToolUtils.cpp
```diff
@@ -215,6 +215,18 @@ std::string findToolFromDylibDir(SmallVector<std::string> toolNames) {
...
+  //   lib/
+  //     libIREECompiler.so
+  //     llvm/bin/
+  //       lld
+  toolPath = findToolAtPath(normalizedToolNames, dylibDir + "/llvm/bin/");
+  if (!toolPath.empty()) {
+    LLVM_DEBUG(llvm::dbgs() << "Found tool in ROCm llvm/bin directory at path "
+                            << toolPath << "\n");
+    return toolPath;
+  }
```
ScottTodd: I'm fine with moving this up, to just after
```
  // Then search in an adjacent bin/ directory. Binary installs are
  // packaged this way:
```

So the search order would keep "installs" ahead of "builds" as:
1. Directly next to the library (for python packages)
2. adjacent bin/ (for IREE standalone installs)
3. nested llvm/bin (for TheRock / ROCm installs)
4. adjacent tools/ (for Bazel builds, IIRC)
5. adjacent llvm-project/bin (for CMake builds)

compiler/src/iree/compiler/Utils/ToolUtils.cpp
```diff
@@ -274,15 +286,26 @@ std::string findPlatformLibDirectory(StringRef platformName) {
...
+    return std::string(ireePath);
   }
-  llvm::sys::fs::make_absolute(path);
-  (void)llvm::sys::path::remove_dots(path, /*remove_dot_dot=*/true);
 
-  std::string pathStr(path);
-  return pathStr;
+  // Fallback: ROCm/TheRock standard location lib/llvm/amdgcn/bitcode/
+  SmallString<256> rocmPath(path);
+  llvm::sys::path::append(rocmPath, "llvm", "amdgcn", "bitcode");
```
ScottTodd: The comment and code disagree here? I think there should be a "lib" component in the path.

Note that there is a "llvm" symlink at the top level dist/rocm/ install path _on Linux only_. Windows does not have that symlink.

AaronStGeorge: > The comment and code disagree here? I think there should be a "lib" component in the path.

IIUC the `lib` should already be part of `path`
```
  std::string dylibPath = getCurrentDylibPath();
  SmallString<256> path(dylibPath);
  // Find lib dir some/path/lib/libIREECompiler.so -> some/path/lib
  llvm::sys::path::remove_filename(path);
```
I'll update the variable names to make this code clearer

> Note that there is a "llvm" symlink at the top level dist/rocm/ install path on Linux only. Windows does not have that symlink

`dist/rocm/llvm` -> symlink, `dist/rocm/lib/llvm` -> real - correct? We should be reading from `dist/rocm/lib/llvm`.

compiler/src/iree/compiler/Utils/ToolUtils.cpp
AaronStGeorge: Updated 👍

compiler/src/iree/compiler/Utils/ToolUtils.cpp
```diff
@@ -272,17 +285,31 @@ std::string findPlatformLibDirectory(StringRef platformName) {
...
   }
 
-  SmallString<256> path(dylibPath);
-  llvm::sys::path::remove_filename(path);
-  llvm::sys::path::append(path, "iree_platform_libs", platformName);
-  if (!llvm::sys::fs::is_directory(path)) {
-    return {};
+  SmallString<256> libPath(dylibPath);
+  // Trim patth to lib dir: some/path/lib/libIREECompiler.so -> some/path/lib
+  llvm::sys::path::remove_filename(libPath);
```
ScottTodd: nit: typo
```suggestion
  // Trim path to lib dir: some/path/lib/libIREECompiler.so -> some/path/lib
  llvm::sys::path::remove_filename(libPath);
```

ScottTodd (approved):

PR #23404, opened 2026-02-06
[CodeGen][NFCI] Introduce bass CodegenOption and move tuningSpecPath into it.

hanhanW:
The revision creates a bass struct that holds common optimization options, like tuningSpecPath.

Assisted-by: Claude

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
@@ -6,13 +6,34 @@
 
 #include "iree/compiler/Codegen/Utils/CodegenOptions.h"
 
+#include <mutex>
+
 IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::CPUCodegenOptions);
 IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::GPUCodegenOptions);
 
 namespace mlir::iree_compiler {
 
+std::string CodegenOptions::tuningSpecPath = "";
+
+void CodegenOptions::bindOptions(OptionsBinder &binder) {
+  static llvm::cl::OptionCategory category("IREE Codegen Options");
+
+  // Use std::call_once to ensure the option is registered exactly once,
+  // even when multiple derived classes (CPU/GPU) call this method.
+  static std::once_flag bindFlag;
+  std::call_once(bindFlag, [&]() {
+    binder.opt<std::string>(
+        "iree-codegen-tuning-spec-path", tuningSpecPath,
+        llvm::cl::cat(category),
+        llvm::cl::desc("Path to a module containing a tuning spec (transform "
+                       "dialect library). Accepts MLIR text (.mlir) and "
+                       "bytecode (.mlirbc) formats."));
+  });
+}
+
 void CPUCodegenOptions::bindOptions(OptionsBinder &binder) {
   static llvm::cl::OptionCategory category("IREE CPU Codegen Options");
+  CodegenOptions::bindOptions(binder);
 
   auto initAtOpt = binder.optimizationLevel(
       "iree-llvmcpu-mlir-opt-level", optLevel,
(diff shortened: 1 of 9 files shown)


hanhanW:
It depends on https://github.com/iree-org/iree/pull/23290

@MaheshRavishankar this demonstrates that we can have common option for backends. As discussed, the tuning spec path can be moved to here, as it is not llvmgpu specific.

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
```diff
@@ -6,13 +6,34 @@
...
 
+std::string CodegenOptions::tuningSpecPath = "";
+
+void CodegenOptions::bindOptions(OptionsBinder &binder) {
+  static llvm::cl::OptionCategory category("IREE Codegen Options");
+
+  // Use std::call_once to ensure the option is registered exactly once,
+  // even when multiple derived classes (CPU/GPU) call this method.
+  static std::once_flag bindFlag;
+  std::call_once(bindFlag, [&]() {
```
kuhar: I think this needs to be stored in the context / dialect, no? Otherwise we won't be able to set different options if we run multiple compilations in parallel?

kuhar: Ah no, this merely registers the flag, the flag handling logic should be the same as any other flag later on

kuhar (approved):

bangtianliu:
I guess you also need to move this test: https://github.com/iree-org/iree/pull/23320/files#diff-4e21d837a6194bd068c03bfec814207696c35c5a30a0a9e73cb4ad4dd5cb7486R17

kuhar:
> I guess you also need to move this test: https://github.com/iree-org/iree/pull/23320/files#diff-4e21d837a6194bd068c03bfec814207696c35c5a30a0a9e73cb4ad4dd5cb7486R17

If this flag is only registered with the rocm backend plugin, it needs to remain there -- otherwise nothing else will register it. The test failure seems real though.

bangtianliu:
> If this flag is only registered with the rocm backend plugin, it needs to remain there -- otherwise nothing else will register it. The test failure seems real though.

This PR moves `tuningSpecPath` to `CodegenOptions`, which supports both CPU/GPU backend and is not ROCM-specific. 
Therefore, the `rocm_c_api_test.cc`  should be renamed and moved a proper location, sth like `compiler/src/iree/compiler/Codegen/Utils/test/codegen_options_c_api_test.cc `.  

On the other hand, The CI failure indicates the refactoring is incomplete so that `CodegenOptions` is not registered with C API successfully.

bangtianliu (approved):
Thanks.

hanhanW:
Although I got two approvals, but I want to say that this depends on https://github.com/iree-org/iree/pull/23290. :p

compiler/src/iree/compiler/Utils/OptionUtils.h
```diff
@@ -114,6 +114,12 @@ class OptionsBinder {
     const Deprecated *dep = filterDeprecated(Ms...);
     auto [changedCallback, clCallback] = makeChangedCallback<V>(name, dep);
     OptionInfo &info = getOptionsStorage()[name];
+    // Skip if this option is already registered in this binder's storage.
+    // This allows multiple option structs to share common options via
+    // inheritance without causing duplicate registration errors.
+    if (info.option) {
+      return;
+    }
```
IanWood1, line 129: For this to work, `opt` needs to be called with the same `value` both times, right? The binder won't update `value` in both locations and this would lead to one of the options being silently un-updated. Maybe we should store `value`'s address so we can error out if they don't match.

MaheshRavishankar (approved):
For the specific comment we discussed @hanhanW  this makes sense. But ill let  @IanWood1  approve.

compiler/src/iree/compiler/Utils/OptionUtils.h
hanhanW: Implemented in https://github.com/iree-org/iree/pull/23438

IanWood1 (approved):

hanhanW:
> For the specific comment we discussed @hanhanW this makes sense. But ill let @IanWood1 approve.

Thanks @IanWood1 for reviewing the code. The change was landed in https://github.com/iree-org/iree/pull/23438, and this PR itself is cleaner now.

PR #23405, opened 2026-02-06
[Runtime] Fix RISC-V test

Muzammiluddin-Syed-ECE:
Add UNIMPLEMENTED as a status to be ignored like PERMISSION DENIED.

Assisted By: composer-1


amd-eochoalo (approved):
Awesome.

PR #23407, opened 2026-02-06
Simplify type switch cases. NFC.

kuhar:
Applied llvm-type-switch-case-types clang-tidy check. Also simplified trivial `.Default` lambdas returning `failure()`, `success()`, or `false` to direct values by hand.


jtuyls (approved):
LGTM

PR #23416, opened 2026-02-06
[GPU][LDS] Only enable coalesced DMA when elements are aligned to minimum transfer size

Yu-Zhewen:
Require divisibility so coalesced DMA doesn't fail in further lowering.


kuhar (approved):

krzysz00 (approved):
Approved, though @lialan this sort of thing is one of the reasons I wanted us to have a generic fallback path. It would've been better if lowering couldn't fail, just become less performant because we had to materialize the copies into registers.

lialan (approved):

lialan:
> Approved, though @lialan this sort of thing is one of the reasons I wanted us to have a generic fallback path. It would've been better if lowering couldn't fail, just become less performant because we had to materialize the copies into registers.

@krzysz00 yes it would be great of we have a general fallback path. But that does not give a lot of benefits in the immediate future. We are prioritizing making DMA work better (without failing of course). We will get to the slow path later.

Issue #18792, opened 2024-10-16
Rename `iree-hip-` compiler flags to `iree-rocm-` when they apply to codegen.


benvanik:
The various ROCM codegen flags like `iree-hip-target` and `iree-hip-enable-ukernels` are not HIP-specific and need to be renamed back to ROCM. The AMDGPU target uses the same ROCM codegen and we shouldn't have any hip naming in the shared code paths.

ScottTodd:
Can we alias the flags? Changing the flag names once caused some disruption.

qedawkins:
My vote would be for `llvm-amdgpu` or maybe just `amdgpu` (but that might be unclear w.r.t. SPIRV) so that it's clear that we're using LLVM's amdgpu backend.

benvanik:
llvm-amdgpu sgtm

kuhar:
Renaming done, we should drop the `--iree-hip-` flag aliases after the v3.11 release.

PR #23420, opened 2026-02-06
Rename iree-hip-* compiler flags to iree-rocm-*

kuhar:
These flags are not HIP-specific and the naming should be aligned with the backend name: ROCm. Also rename the IREE_HIP_TEST_TARGET_CHIP CMake flag to IREE_ROCM_TEST_TARGET_CHIP.

The old flag names are kept as deprecated aliases that emit warnings when used. We can drop the aliases after the next release.

Issue: https://github.com/iree-org/iree/issues/18792


benvanik:
thank you!

benvanik (approved):

qedawkins (approved):

PR #23423, opened 2026-02-07
[DispatchCreation] Update split reduction heuristics for GEMM

yzhang93:
There are two types of GEMM with large reduction dimensions need split reduction which are convolution-like shapes (y=1, x=1) and standard GEMM shapes. The current split heuristics were mostly derived from convolutions with x=1, y=1, which typically have relatively small m/n sizes and an extremely large k size. However, for GEMM the shape characteristics are different. The goal of this PR is to update the heuristics to improve the performance for GEMM shapes while keeping the performance for convolution-like shapes unchanged. The new decision parameters were derived from extensive experiments on Mi355.

With the new heuristics, the performance for convolution-like shapes (include all Prod and Proxy lists, also added new customer shapes) is as below:

```
Total tests: 381

Baseline total:     62729.17 us
New total:  61222.39 us
Overall speedup:    1.02x
Overall improvement: +2.40%

Tests improved:  67/381 (17.6%)
Tests neutral:   303/381 (79.5%)
Tests regressed: 11/381 (2.9%)
```

And for large-k GEMM shapes:

```
Total tests: 38

Baseline total:     127169.12 us
New total:  97637.32 us
Overall speedup:    1.30x
Overall improvement: +23.22%

Tests improved:  17/38 (44.7%)
Tests neutral:   18/38 (47.4%)
Tests regressed: 3/38 (7.9%)
```

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -353,19 +353,27 @@ getMatmulLikeReductionSizes(PartialReductionOpInterface op,
   int64_t mSize = getSizeAt(mDims);
   int64_t nSize = getSizeAt(nDims);
   int64_t kSize = getSizeAt(kDims);
+
+  // TODO(vivian): Take element type into account, and use total bytes instead
+  // of output size.
   int64_t outputSize = mSize * nSize * batchSize;
   int64_t ratio = kSize / std::sqrt(mSize * nSize) / batchSize;
 
+  // The constants below are determined based on empirical data.
+  const int64_t largeOutputSize = 2048 * 4096;
+  const int64_t largeKSize = 18000;
+  const int64_t ratioThreshold = 48;
+
   // When the output size is large, the workload tends to distributed across
   // many workgroups, making split reduction little to no effect.
-  if (outputSize > 2048 * 4096) {
+  if (outputSize > largeOutputSize) {
     LDBG() << "skipping op; large output size";
     return std::nullopt;
   }
 
   // When the reduction size is small relative to the M/N sizes, split
   // reduction often has no effect or even degrades performance.
-  if (kSize < 18000 && ratio < 48) {
+  if (kSize < largeKSize && ratio < ratioThreshold) {
     LDBG() << "skipping op; small reduction size";
     return std::nullopt;
   }
--- a/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
@@ -116,8 +116,7 @@ getOuterReductionSizes(PartialReductionOpInterface op,
 /// These convolutions have a special CHWN or CNHW layout, where the filter
 /// sizes (corresponding to output image sizes in forward convolutions) are
 /// typically large, while the output spatial dimensions are small. This makes
-/// the split reduction strategy particularly effective. Currently, splitting
-/// is only applied along the input channel dimension.
+/// the split reduction strategy particularly effective.
 static std::optional<SmallVector<int64_t>>
 getWeightBackwardReductionSizes(PartialReductionOpInterface op,
                                 int64_t splitReductionTargetSize) {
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -116,8 +116,7 @@ getOuterReductionSizes(PartialReductionOpInterface op,
 /// These convolutions have a special CHWN or CNHW layout, where the filter
 /// sizes (corresponding to output image sizes in forward convolutions) are
 /// typically large, while the output spatial dimensions are small. This makes
-/// the split reduction strategy particularly effective. Currently, splitting
-/// is only applied along the input channel dimension.
+/// the split reduction strategy particularly effective.
```
nirvedhmeshram, line 119: Is `getWeightBackwardReductionSizes` the right name for this function looks like it handles all convolutions?

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -354,23 +353,19 @@ getMatmulLikeReductionSizes(PartialReductionOpInterface op,
...
-  const int64_t ratioThreshold = 384;
-  const int64_t largeKSize = 24576;
-  const int64_t largeMNSize = 1024;
-
-  // When the M or N size is large, the workload tends to distributed across
+  // When the output size is large, the workload tends to distributed across
   // many workgroups, making split reduction little to no effect.
-  if (mSize > largeMNSize || nSize > largeMNSize) {
-    LDBG() << "skipping op; large M or N size";
+  if (outputSize > 2048 * 4096) {
```
nirvedhmeshram: I preferred the code before in that it had named constants can we still keep things like largeMNSize.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -354,23 +353,19 @@ getMatmulLikeReductionSizes(PartialReductionOpInterface op,
...
+    LDBG() << "skipping op; large output size";
     return std::nullopt;
   }
 
   // When the reduction size is small relative to the M/N sizes, split
   // reduction often has no effect or even degrades performance.
-  int64_t ratio = kSize / std::sqrt(mSize * nSize) / batchSize;
-  if (ratio <= ratioThreshold && kSize < largeKSize) {
+  if (kSize < 18000 && ratio < 48) {
```
nirvedhmeshram: same here we should keep using constants with name for better readibility.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
```diff
@@ -354,23 +353,19 @@ getMatmulLikeReductionSizes(PartialReductionOpInterface op,
   int64_t mSize = getSizeAt(mDims);
   int64_t nSize = getSizeAt(nDims);
   int64_t kSize = getSizeAt(kDims);
+  int64_t outputSize = mSize * nSize * batchSize;
```
nirvedhmeshram, line 359: maybe out of scope for this PR, but we should make this a function of the element types and do this in terms of total bytes rather than the size, how the lowering config does here https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp#L395-L408 
Perhaps worth adding aa TODO about that?

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
yzhang93: There is a confusion because of the folding codes. The changes in this PR are actually within another function `getMatmulLikeReductionSizes`.

compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp
yzhang93: Done.

nirvedhmeshram (approved):
LGTM

PR #22881, opened 2025-12-10
[LLVMCPU] Add MaterializeTuningSpecsPass to codegen pipeline

gglangg:
Add `MaterializeTuningSpecsPass` to LLVMCPU codegen pipeline, mirroring the LLVMGPU pipeline structure.

This enables tuning spec support for CPU targets via `--iree-codegen-tuning-spec-path`.

PR #23424, opened 2026-02-07
[CPU] Plumb through tuning spec support for LLVMCPU.

hanhanW:
Closes https://github.com/iree-org/iree/pull/22881

### DIFF
--- a/compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp
+++ b/compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp
@@ -243,7 +243,7 @@ public:
   void
   buildConfigurationPassPipeline(IREE::HAL::ExecutableTargetAttr targetAttr,
                                  OpPassManager &passManager) override {
-    buildLLVMCPUCodegenConfigurationPassPipeline(passManager);
+    buildLLVMCPUCodegenConfigurationPassPipeline(passManager, codegenOptions_);
   }
 
   void buildTranslationPassPipeline(IREE::HAL::ExecutableTargetAttr targetAttr,
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
@@ -655,12 +655,14 @@ static void addLowerToLLVMPasses(OpPassManager &modulePassManager,
 }
 
 void buildLLVMCPUCodegenConfigurationPassPipelineImpl(
-    OpPassManager &modulePassManager) {
+    OpPassManager &modulePassManager, const CPUCodegenOptions &cpuOpts) {
   {
     FunctionLikeNest funcPassManager(modulePassManager);
     addCommonTargetExecutablePreprocessingPasses(funcPassManager,
                                                  clUseSoftmaxInterFusion);
   }
+  modulePassManager.addPass(createMaterializeTuningSpecsPass(
+      MaterializeTuningSpecsPassOptions{cpuOpts.tuningSpecPath}));
   modulePassManager.addPass(createMaterializeUserConfigsPass());
   FunctionLikeNest(modulePassManager)
       .addPass(createMaterializeDeviceEncodingPass)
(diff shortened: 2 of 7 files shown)


bangtianliu (approved):
LGTM! LLVMCPU tuning isn’t currently supported in the tuner, though.

hanhanW:
> LGTM! LLVMCPU tuning isn’t currently supported in the tuner, though.

Is it because the tuner does not understand CPU pipeline yet? I think that's fine for now. People will be able to use the tuning pass now.

kuhar:
> Is it because the tuner does not understand CPU pipeline yet? I think that's fine for now. People will be able to use the tuning pass now.

Correct. You can write tuning specs by hand but the tuner itself doesn't know how to produce these

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.h
kuhar, line 23: What happens if I invoke the compiler concurrently in two threads with different flags? Is this supported?

kuhar: @hanhanW can we discuss this before merging?

hanhanW: Sorry, I missed the comment and it looks like the comment should be dropped in the other PR.

I don't quite follow your question. My understanding is (and it ties to LLVM cl::opt design) that each flag binds to a storage. There may be multiple flags bind to the same storage like what you did in https://github.com/iree-org/iree/commit/740405ce0bac4eef1df06970523038a257c8d1ba I added few unit tests in https://github.com/iree-org/iree/commit/6744f1b97d479edc2379c056582b6551fbe8706c today that might answer your question?

If there are two threads using the same session, you likely hit races. It depends on how you set up the env. Can you elaborate the scenario more? Techniqually, the PR itself does not impact how the mechanism works. The OptionsBinder seems to follow LLVM cl design, if I understand the code correctly.

@IanWood1 has more context, FYI.

kuhar: > Sorry, I missed the comment and it looks like the comment should be dropped in the other PR.

Ah right, I saw a dangling comment and did not realize this was rebased, so no blockers here.

> If there are two threads using the same session, you likely hit races. It depends on how you set up the env. Can you elaborate the scenario more? 

I think this is the key property should be that this is not worse than any other llvm::cl flag registration / storage.

The scenario I'm thinking about is when you do multi-threaded JIT compilation with different options -- I imagine this could be a future scenario for fusilli tuning, where we spawn a few threads, each gets its own LLVM / MLIRContext, but they may pick different compiler options. I think anything that relies on values stored in global/static variables is fundamentally incompatible with it, unless the underlying storage is thread-local (or context-local).

hanhanW: These values are not context-local. They are process-global because this is a fundamental limitation of `LLVM's cl::opt system`. Your concern happens to all the projects that use `LLVM cl::opt system`. So yeah, this is not worse than any other llvm::cl flag registration / storage.

In practice, I don't think that we'd do JIT compilation with different options. They should be consistent to me. E.g., you don't ask a dispatch to use `O3` and the other to use `O0`, right? If you look at how IREE lower executables to LLVM, it already collects all the function into a single llvm::Module and do the lowering. It implies that we are consistent in terms of this level of optimization.

https://github.com/iree-org/iree/blob/4802bac98f12c68ca86014e1aaef225202e72300/compiler/plugins/target/ROCM/ROCMTarget.cpp#L848-L851

where the `llvmModule` is from `hal.executable.variant`:

https://github.com/iree-org/iree/blob/4802bac98f12c68ca86014e1aaef225202e72300/compiler/plugins/target/ROCM/ROCMTarget.cpp#L607

The IREE executable structure is

```
hal.executable.variant {
  hal.executable.export public @a_func
  hal.executable.export public @b_func
  hal.executable.export public @c_func
  builtin.module {
    llvm.func @internal_funcs...like_iree_uk_mmt4d // this is resolved during link-time.
    llvm.func @a_func
    llvm.func @b_func
    llvm.func @c_func
  }
}
```

kuhar: You can specify different tuning specs in each thread -- that's what I'm thinking about

> E.g., you don't ask a dispatch to use O3 and the other to use O0, right

this sounds like a reasonable idea too, but seems like it's not possible today

hanhanW: yeah, unfortunately I think we hit the limitation here. Just adding a bit that this is a pre-existing issue. We already had such constraint when we introduce this flag to IREE.

> You can specify different tuning specs in each thread -- that's what I'm thinking about

I don't know what we'd need here, but I imagine that we can have all the matcher and spec in a single file unless you want different strategy for the same kernel. I think we'll figure it out when we run into this problem. :)

PR #23430, opened 2026-02-08
[VectorDistribute] add support padding integer attention masks 

ziereis:
Currently compiling attention with  an integer/bool mask crashes. This fix addresses this by selecting the neutral element based on the type of the mask.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
+++ b/compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
@@ -285,6 +285,9 @@ struct OnlineAttentionOpInterface final
                                        paddedOp.getValueMap());
     if (paddedOp.getMask()) {
       Type maskElementType = getElementTypeOrSelf(paddedOp.getMask().getType());
+      // The attention mask can be either float or integer. For float masks,
+      // we use maximumf whose identity is -INF.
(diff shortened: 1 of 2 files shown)


Groverkss (approved):

compiler/src/iree/compiler/Codegen/Interfaces/TensorMaskingOpInterface.cpp
```diff
@@ -284,8 +284,12 @@ struct OnlineAttentionOpInterface final
                                        arith::AtomicRMWKind::addf, bounds,
                                        paddedOp.getValueMap());
     if (paddedOp.getMask()) {
+      Type maskElementType = getElementTypeOrSelf(paddedOp.getMask().getType());
+      arith::AtomicRMWKind maskNeutralKind =
+          isa<FloatType>(maskElementType) ? arith::AtomicRMWKind::maximumf
+                                          : arith::AtomicRMWKind::addi;
```
bangtianliu, line 293: The current logic assumes only two cases: float → maximumf, everything else → addi

1. Could we also add a test with a wider integer type (e.g., i8) to confirm the integer path generalizes beyond boolean masks?
2. Add a comment to explain using addi for the else branch to make it more friendly for readers. My understanding is that addi is just used to get 0s.

bangtianliu (approved):

PR #23431, opened 2026-02-08
Enable TensorDynamicDimAnalysis to propagate through `iree_codegen.load_from_buffer`

ziereis:
Extended `TensorDynamicDimAnalysis` to propagate information through `iree_codegen.load_from_buffer` ops by reading the dynamic dimensions from the source memref. 

Without this change the block-dynamic-dims pass was only able to block the dynamic dimension of Q for IR that looks similar to the test that is included here. The reason for this is that Q could be read from the `empty` op of the result,  however the other dims are not exposed via the result of the attention op.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
@@ -5,6 +5,8 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.h"
+#include <cstdint>
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
 #include "iree/compiler/Dialect/TensorExt/IR/TensorExtOps.h"
 #include "iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.h"
 #include "llvm/Support/Debug.h"
@@ -123,6 +125,44 @@ static void updateTensorDimInfo(
   }
 }
 
+// Update the tensor dimension information for result of an
+// `iree_codegen.load_from_buffer` operation.
+static void updateTensorDimInfo(
+    IREE::Codegen::LoadFromBufferOp loadFromBufferOp,
+    const DataFlowSolver &solver,
+    TensorDynamicDimAnalysis::TensorDimDivisibilityInfo &divisibilityInfo,
+    TensorDynamicDimAnalysis::TensorDimRangeInfo &rangeInfo) {
+  RankedTensorType resultType =
+      cast<RankedTensorType>(loadFromBufferOp.getTensor().getType());
+  if (resultType.hasStaticShape()) {
+    return;
+  }
+
+  auto bufferType =
+      dyn_cast<MemRefType>(loadFromBufferOp.getBuffer().getType());
+  if (!bufferType) {
+    return;
+  }
+
+  std::optional<ValueRange> maybeBufferDynamicDims =
+      IREE::Util::findDynamicDims(loadFromBufferOp.getBuffer());
+  if (!maybeBufferDynamicDims.has_value()) {
+    return;
+  }
+  ValueRange bufferDynamicDims = maybeBufferDynamicDims.value();
+
+  Value result = loadFromBufferOp.getResult();
+  uint64_t dynamicDimIndex = 0;
+  for (auto [dimIndex, dimSize] : llvm::enumerate(bufferType.getShape())) {
+    if (!ShapedType::isDynamic(dimSize)) {
+      continue;
+    }
+    Value dynamicDim = bufferDynamicDims[dynamicDimIndex++];
+    updateTensorDimInfo(result, dimIndex, dynamicDim, solver, divisibilityInfo,
+                        rangeInfo);
+  }
+}
+
 // Update the tensor dimension information for result of a `tensor.empty`
 // operation.
 static void updateTensorDimInfo(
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
```diff
@@ -123,6 +125,44 @@ static void updateTensorDimInfo(
...
 
+// Update the tensor dimension information for result of an
+// `iree_codegen.load_from_buffer` operation.
+static void updateTensorDimInfo(
+    IREE::Codegen::LoadFromBufferOp loadFromBufferOp,
+    const DataFlowSolver &solver,
+    TensorDynamicDimAnalysis::TensorDimDivisibilityInfo &divisibilityInfo,
+    TensorDynamicDimAnalysis::TensorDimRangeInfo &rangeInfo) {
+  RankedTensorType resultType =
+      cast<RankedTensorType>(loadFromBufferOp.getTensor().getType());
```
hanhanW: Use `auto` because `cast` spells the type.

compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
```diff
@@ -123,6 +125,44 @@ static void updateTensorDimInfo(
...
+  if (resultType.hasStaticShape()) {
+    return;
+  }
+
+  auto bufferType =
+      dyn_cast<MemRefType>(loadFromBufferOp.getBuffer().getType());
+  if (!bufferType) {
+    return;
+  }
```
hanhanW: I think you don't need the check. By definition, it should always be MemRefType. E.g.,

https://github.com/iree-org/iree/blob/6f2612b7f92b791ef89925e16468f4f0711366a8/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp#L81-L91

compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
```diff
@@ -123,6 +125,44 @@ static void updateTensorDimInfo(
...
+      IREE::Util::findDynamicDims(loadFromBufferOp.getBuffer());
+  if (!maybeBufferDynamicDims.has_value()) {
+    return;
+  }
+  ValueRange bufferDynamicDims = maybeBufferDynamicDims.value();
+
+  Value result = loadFromBufferOp.getResult();
+  uint64_t dynamicDimIndex = 0;
+  for (auto [dimIndex, dimSize] : llvm::enumerate(bufferType.getShape())) {
+    if (!ShapedType::isDynamic(dimSize)) {
```
hanhanW: nit: avoid double negative. We can use `isStatic` here.

compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
```diff
@@ -123,6 +125,44 @@ static void updateTensorDimInfo(
...
+      continue;
+    }
+    Value dynamicDim = bufferDynamicDims[dynamicDimIndex++];
+    updateTensorDimInfo(result, dimIndex, dynamicDim, solver, divisibilityInfo,
+                        rangeInfo);
```
hanhanW: The `dynamicDim` variable does not add much value, e.g., readability, to me, so maybe we collapse two statements?

compiler/src/iree/compiler/Codegen/Common/test/block_dynamic_dims.mlir
```diff
@@ -101,6 +101,93 @@ func.func @block_attention_dims() {
 //       CHECK:   %[[GENERIC:.+]] = linalg.generic
 //       CHECK:   iree_tensor_ext.dispatch.tensor.store %[[GENERIC]], %[[OUTPUT_BINDING]]
 
+func.func @block_attention_dims_load_from_buffer() {
```
hanhanW, line 112: You're missing a separator (ie `// -----`) above.

compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
hanhanW: I was expecting `MemRefType bufferType = loadFromBufferOp.getBuffer().getType()`.

The method returns TypedValue result.

> ::mlir::TypedValue<::mlir::MemRefType> getBuffer() {

ziereis: Sorry! i didn't properly look at this. Should be resolved now.

compiler/src/iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.cpp
```diff
@@ -123,6 +126,39 @@ static void updateTensorDimInfo(
...
+    const DataFlowSolver &solver,
+    TensorDynamicDimAnalysis::TensorDimDivisibilityInfo &divisibilityInfo,
+    TensorDynamicDimAnalysis::TensorDimRangeInfo &rangeInfo) {
+  auto resultType =
+      cast<RankedTensorType>(loadFromBufferOp.getTensor().getType());
+  if (resultType.hasStaticShape()) {
+    return;
+  }
+
+  MemRefType bufferType = loadFromBufferOp.getBuffer().getType();
```
hanhanW: picky nit: I'd move it to where it is use, i.e., line 152. It was here for verification, but we no longer need it.

hanhanW (approved):

hanhanW:
This breaks https://github.com/iree-org/iree/actions/runs/21916571112/job/63286562266

Issue #14048, opened 2023-06-11
[metal] Probe metal offline toolchain in PATH


antiagainst:
(Issue created to track https://github.com/openxla/iree/pull/12713#discussion_r1214557107 fixes.)

Right now in the Metal SPIR-V target we check whether the current host platform is Apple to decide whether invoking offline toolchain to generate MTLLibrary. This should be fixed to check `PATH` given the metal offline toolchain can also be available on other platforms like Windows.

PR #23433, opened 2026-02-09
[MetalSPIRV] Auto-detect Metal tools and fall back to MSL source embedding

dbrll:
On macOS without Xcode (Command Line Tools only), `iree-compile` with
`--iree-hal-target-backends=metal-spirv` fails because `xcrun metal` is
not found and the compiler errors out instead of falling back.

This patch makes the Metal compiler backend gracefully degrade:

- Add `findMetalToolchain()` that resolves `xcrun` once and verifies both
  `metal` and `metallib` are reachable, avoiding redundant PATH lookups
- When tools are absent, emit a remark and embed MSL source instead of
  erroring out. The Metal HAL runtime already handles this path via
  `MTLDevice.makeLibrary(source:)`
- Replace `system()` with `llvm::sys::ExecuteAndWait` and capture stderr
  so compilation failures report actual compiler diagnostics (follows the
  CUDA/ptxas pattern)
- CMake: replace aggressive Metal disabling with an informational status
  message. The HAL driver and compiler backend both work without Xcode

Addresses TODO [14048](https://github.com/iree-org/iree/issues/14048): "[metal] Probe metal offline toolchain in PATH".

## Test plan

- macOS with Xcode: produces metallib (existing behavior preserved)
- macOS without Xcode: emits remark, embeds MSL source, succeeds
- Linux: embeds MSL source (existing cross-compilation behavior)
- `--iree-metal-compile-to-metallib=false`: embeds MSL source regardless

ci-extra: macos_arm64_clang

Assisted-by: Claude

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,6 +294,20 @@ set(IREE_HAL_DRIVER_METAL_DEFAULT ${IREE_HAL_DRIVER_DEFAULTS})
 # Right now only support Apple silicon devices.
 if(NOT APPLE OR NOT ${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm64")
   set(IREE_HAL_DRIVER_METAL_DEFAULT OFF)
+else()
+  # Probe for Metal offline compilation tools (optional, for ahead-of-time
+  # compilation). The Metal compiler backend will automatically fall back to
+  # embedding MSL source for runtime compilation when these tools are absent.
+  execute_process(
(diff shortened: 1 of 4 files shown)


dbrll:
@antiagainst This PR auto-detects the Metal toolchain at compile time and falls back to MSL source embedding when Xcode isn't available (addresses #14048). Rebased on latest main. Would appreciate your feedback when you get a chance.

PR #23400, opened 2026-02-05
[Codege][GPU] Add async copy mode pipelining for gather_to_lds

jerryyin:
This PR build upon the multi-buffer PR [here](https://github.com/iree-org/iree/pull/23354). It extends the async copy pipelining to support configurable pipeline depths (2 or 3 stages), following Triton's approach of using empty intermediate stages to achieve N-1 prologue iterations.

- **N-stage support**: Compute ops assigned to `stage = numStages - 1`, creating empty intermediate stages that cause the pipeliner to generate N-1 prologue iterations
- **Multi-buffering**: Buffer count matches pipeline depth (2-stage → 2 buffers, 3-stage → 3 buffers)
- **Preserved execution order**: Loop body maintains `load → ds_read → compute` order regardless of stage count
- **Stage assignment strategy**:
  - Stage 0: `gather_to_lds` (N-1 iterations ahead)
  - Stage 1 to N-2: empty (for N > 2)
  - Stage N-1: `transfer_read` + compute (current iteration)

Behavior Summary:
 - 2 Stage pipeline has 2 buffers, 1 load in prologue and 1 compute in epilogue
 - 3 Stage pipeline has 3 buffers, 2 loads in prologue and 2 computes in epilogue

PR #23437, opened 2026-02-09
[Codegen][GPU] Add LDS alias scope metadata for LoadToLDS operations

jerryyin:
This PR is a follow-up to https://github.com/iree-org/iree/pull/23400. I find that the backend will conservatively insert vmcnt(0) that kills the multi-buffering async load performance. Cursor did the investigation and help me understand that we need to attach alias scopes so that the backend can differentiate from multiple buffers that they are not alias to each other so no conservative vmcnt(0) added.

Detailed Context: In pipelined loops with multi-buffered LDS, the LLVM backend's SIInsertWaitcnts pass conservatively inserts s_waitcnt vmcnt(0) before ds_read when it can't prove that in-flight buffer_load_lds operations target a different LDS region. With alias scopes, the backend knows that LoadToLDS ops to different buffers don't alias, eliminating one redundant vmcnt(0) per loop iteration.

The pass traces each LoadToLDS LDS pointer back through GEP and extractvalue chains to find the root allocation, creates a unique scope per distinct root, and annotates each op with alias_scopes (self) and noalias_scopes (all others).

In end to end multi-buffered pipeline test, this pass can provide enough hint for backend to skip overly conservative vmcnt(0).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/BUILD.bazel
@@ -110,6 +110,7 @@ iree_compiler_cc_library(
         "LLVMGPUVectorLowering.cpp",
         "LLVMGPUVectorToGPU.cpp",
         "Passes.cpp",
+        "ROCDLAddLDSAliasScopes.cpp",
         "ROCDLAnnotateKernelForTranslation.cpp",
         "ROCDLBufferInstructionsOptimization.cpp",
         "ROCDLConfigureBufferInstructions.cpp",
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/CMakeLists.txt
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/CMakeLists.txt
@@ -89,6 +89,7 @@ iree_cc_library(
     "LLVMGPUVectorLowering.cpp"
     "LLVMGPUVectorToGPU.cpp"
     "Passes.cpp"
+    "ROCDLAddLDSAliasScopes.cpp"
     "ROCDLAnnotateKernelForTranslation.cpp"
     "ROCDLBufferInstructionsOptimization.cpp"
     "ROCDLConfigureBufferInstructions.cpp"
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
@@ -9,6 +9,7 @@
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUDialect.h"
 #include "iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.h"
 #include "iree/compiler/Codegen/LLVMGPU/Passes.h"
+#include "iree/compiler/Codegen/LLVMGPU/ROCDLPasses.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
 #include "llvm/Support/DebugLog.h"
 #include "mlir/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.h"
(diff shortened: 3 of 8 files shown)


krzysz00:
I'm going to note that once the explicit async support lands, we might not need this anymore

krzysz00:
I'm also going to flag that this reads as fundamentially insufficient - if you're multi-buffered in LDS, you'll have multiple logically distinct buffers in one allocation.

jerryyin:
> once the explicit async support lands, we might not need this anymore

Could you let me know what explicit async support is? Triton style explicit token and waits? And who is working on it? 

If "explicit async support" simply means generating s_waitcnt vmcnt(N) instructions at the IR level with the correct count, that alone doesn't solve the problem. The SIInsertWaitcnts pass will still run its own alias analysis, still conclude that it can't prove the LoadToLDS targets don't alias, and still compute vmcnt(0). So at the end of the day you still need one of the following:
 1. Alias scope, as implemented here to tell SIInsertWaitcnts to compute a non-conservative vmcnt
 2. A way to suppress SIInsertWaitcnts, which I don't see how

>  if you're multi-buffered in LDS, you'll have multiple logically distinct buffers in one allocation.

I thought this is exactly what this implementation tell the backend? Basically telling the backend that the buffer used to prefetch isn't a alias with the buffer used to ds_read/MFMA, therefore eliminating the defensive vmcnt(0). What is fundamentally insufficient?

jerryyin:
> fundamentally insufficient" — multiple logically distinct buffers in one allocation

I took some additional time looking at this and think I understand what you meant know. After some poking, it looks to me the approach is more robust than I expected:

After `memref.multiBuffer`, both slots live within a single enlarged `memref.alloc`. However, the pass operates after memref-to-LLVM lowering, and the way `memref.subview` lowers guarantees that each slot gets its own descriptor struct. Specifically:

1. `memref.multiBuffer` creates `memref.subview %mb_alloc[%buf_idx, 0, 0] [1, 32, 128] [1, 1, 1]` per slot, with a different dynamic `%buf_idx` per slot.
2. `expand-strided-metadata` + `finalize-memref-to-llvm` lowers each `memref.subview` by constructing a **fresh LLVM struct descriptor** via `llvm.mlir.poison` → chain of `llvm.insertvalue` ops. This is structural — the subview descriptor has different rank/offset/strides than the parent, so a new struct must be built.
3. At the LLVM dialect level, slot 0's pointer comes from `llvm.extractvalue %desc_slot0[1]` and slot 1's from `llvm.extractvalue %desc_slot1[1]`. Since `%desc_slot0` and `%desc_slot1` are different SSA values (different `insertvalue` chains), `traceToLDSBase` correctly identifies two distinct roots.

I verified this by running a minimal multi-buffer-like test through the lowering pipeline — each subview consistently produces its own descriptor. The key invariant is that memref-to-LLVM lowering always builds a per-subview descriptor struct because the offset metadata differs between slots.

That said, I'll concede that if the lowering infrastructure were fundamentally different — e.g., bypassing descriptors entirely and computing slot addresses via GEP arithmetic from a shared base pointer — then both slots would trace to the same root and the pass would silently produce no alias scopes. But that is not the case today and would be a significant change to the lowering infrastructure.

krzysz00:
@jerryyin The asyncmark / asyncwait stuff is https://github.com/llvm/llvm-project/pull/180466 and https://github.com/llvm/llvm-project/pull/180466 . This is a compiler feature that's very likely landing soon (might need a few pokes of LLVM folks to stamp it) which'll bypass that vmcnt() insertion for these global loads to LDS by making it something we manage explicitly. It'll also be how we're supposed to deal with the gfx1250 async operations.

---

And re the thing with subviews, this means I'm holding this PR for correctness concerns.

1. Why were those subviews not folded away by `fold-memref-alias-ops`???
2. If different subviews are being marked `noalias` with each other without proof of disjointness, that's incorrect.

krzysz00 (changes requested):
(correctness hold)

jerryyin:
> . This is a compiler feature that's very likely landing soon (might need a few pokes of LLVM folks to stamp it) which'll bypass that vmcnt() insertion

Good to know! I'm hoping that can make my life easier.

> 1. Why were those subviews not folded away by fold-memref-alias-ops???
> 2. If different subviews are being marked noalias with each other without proof of disjointness, that's incorrect.

Legit concerns. I'm not familiar with 1). For 2), right, just infer noalias from SSA isn't enough. It probably mean the noalias information needs to be attached at the subview level when we know those are indeed disjoint from each other. But given the PR you've referenced, it probably doesn't make sense to go down the rabbit hole from this PR.

jerryyin:
Superseded by https://github.com/iree-org/iree/pull/23648 and closing

PR #23443, opened 2026-02-10
[e2e] Fix missed iree-hip-* to iree-rocm-* flag renames 

Yu-Zhewen:
Follow-up to https://github.com/iree-org/iree/pull/23420


kuhar (approved):
Thanks!

PR #23447, opened 2026-02-10
[PCF] Add producer fusion into pcf.generic/loop ops

qedawkins:
Add FuseProducersPass (iree-pcf-fuse-producers) that fuses DPS producers into pcf.generic and pcf.loop ops through their tied init arguments.

For each tied init that is the single result of a TilingInterface + DestinationStyleOpInterface producer (e.g. linalg.fill, linalg.transpose), the pass:
1. Replaces the scoped op's init with the producer's DPS init
2. At each pcf.read_slice on the corresponding sref, generates a tiled version of the producer via generateResultTileValue
3. For vector-typed read_slices, converts the tiled tensor result to a vector via vector.transfer_read
4. Erases the original producer if it has no remaining uses

The pass requires SyncOnReturn semantics on the sref, under which overlapping reads and writes are undefined behavior, allowing all reads to assume as though they see the init value.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
@@ -110,9 +110,9 @@ struct ProducerFusionParams {
 // through its tied init argument. The producer must:
 //   1. Implement TilingInterface and DestinationStyleOpInterface.
 //   2. Have a single result.
-//   3. Have all operands dominating the scoped op.
-//   4. Feed into a tied init of the scoped op whose corresponding sref arg
-//      has only read_slice and write_slice users.
+//   3. Have all operands dominating the generic/loop op.
+//   4. Feed into a tied init of the generic/loop op whose corresponding sref
+//      arg has only read_slice and write_slice users.
 //
 // On success, |params| is populated with the result index, producer op, and
 // list of read_slice ops to replace with tiled producer computations.
@@ -122,7 +122,7 @@ LogicalResult matchTilableProducer(RewriterBase &rewriter,
 LogicalResult matchTilableProducer(RewriterBase &rewriter, PCF::LoopOp loopOp,
                                    ProducerFusionParams &params);
 
-// Fuses the matched producer into the scoped op by:
+// Fuses the matched producer into the pcf.generic/loop op by:
 //   1. Replacing the tied init with the producer's DPS init.
 //   2. Generating tiled producer computations at each read_slice site via
 //      TilingInterface::generateResultTileValue.
--- a/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
@@ -96,6 +96,42 @@ void fuseTilableConsumer(RewriterBase &rewriter, PCF::LoopOp loopOp,
                          TilingInterface target,
                          const ConsumerFusionParams &params);
 
+struct ProducerFusionParams {
+  // Which result index of the scoped op has a fusable producer init.
+  unsigned resultIndex;
+  // The producer op to fuse (implements TilingInterface + DPS). Stored as
+  // Operation* to avoid requiring the full TilingInterface definition here.
+  Operation *producer;
+  // Read sites on the corresponding sref arg that consume the init value.
+  SmallVector<PCF::ReadSliceOp> readSlices;
+};
+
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
```diff
@@ -96,6 +96,42 @@ void fuseTilableConsumer(RewriterBase &rewriter, PCF::LoopOp loopOp,
...
+};
+
+// Helpers to match a DPS producer as fusable into a pcf.generic/loop op
+// through its tied init argument. The producer must:
+//   1. Implement TilingInterface and DestinationStyleOpInterface.
+//   2. Have a single result.
+//   3. Have all operands dominating the scoped op.
+//   4. Feed into a tied init of the scoped op whose corresponding sref arg
+//      has only read_slice and write_slice users.
+//
```
Max191:
```suggestion
//   3. Have all operands dominating the pcf.generic/loop op.
//   4. Feed into a tied init of the pcf.generic/loop op whose corresponding sref arg
//      has only read_slice and write_slice users.
//
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h
```diff
@@ -96,6 +96,42 @@ void fuseTilableConsumer(RewriterBase &rewriter, PCF::LoopOp loopOp,
...
+// On success, |params| is populated with the result index, producer op, and
+// list of read_slice ops to replace with tiled producer computations.
+LogicalResult matchTilableProducer(RewriterBase &rewriter,
+                                   PCF::GenericOp genericOp,
+                                   ProducerFusionParams &params);
+LogicalResult matchTilableProducer(RewriterBase &rewriter, PCF::LoopOp loopOp,
+                                   ProducerFusionParams &params);
+
+// Fuses the matched producer into the scoped op by:
```
Max191:
```suggestion
// Fuses the matched producer into the pcf.generic/loop op by:
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/test/fuse_producers.mlir
```diff
@@ -0,0 +1,275 @@
...
+//       CHECK:    %[[EXT:.+]] = tensor.extract_slice %[[DEST]][%[[ID0]], %[[ID1]]] [4, 8] [1, 1]
+//       CHECK:    %[[TILED_FILL:.+]] = linalg.fill ins(%[[CST]]{{.*}} outs(%[[EXT]]
+//       CHECK:    %[[VEC:.+]] = vector.transfer_read %[[TILED_FILL]][%[[C0]], %[[C0]]], %[[POISON]] {in_bounds = [true, true]}
+//       CHECK:    pcf.write_slice %[[VEC]] into %[[REF]][%[[ID0]], %[[ID1]]] [4, 8] [1, 1]
+//       CHECK:    pcf.return
+//       CHECK:  return %[[GENERIC]]
+
+// -----
+
+// Negative Tests:
```
Max191, line 253: Maybe add a negative test for non-SyncOnReturn scope?

Max191 (approved):
Overall LGTM, just some nits.

One question: Is fusion through inits expected to happen separately from fusion via implicit capture? Why does this pass only perform fusion through the init args?

qedawkins:
> Overall LGTM, just some nits.
> 
> One question: Is fusion through inits expected to happen separately from fusion via implicit capture? Why does this pass only perform fusion through the init args?

Fusion through implicit capture doesn't need anything special since it's just `extract_slice(tilable op)` you can use the same pattern everywhere.

Max191:
> Fusion through implicit capture doesn't need anything special since it's just `extract_slice(tilable op)` you can use the same pattern everywhere.

So then we won't have pcf.read_slice ops on anything but the sref block args?

qedawkins:
> > Fusion through implicit capture doesn't need anything special since it's just `extract_slice(tilable op)` you can use the same pattern everywhere.
> 
> So then we won't have pcf.read_slice ops on anything but the sref block args?

You'd have to convert to sref somewhere, and that wouldn't be via the pcf.generic/loop since fusion through its tied inits is what this handles and that's the only source of sref the loop provides. So if we did have to handle pcf.read_slice somewhere else, it would be orthogonal to this. In general, I don't think making management of implicit capture the responsibility of the capturing op is a good idea outside of converstion to/from isolated from above. It breaks the core MLIR assumption that all uses are the same.

Issue #22089, opened 2025-09-24
Allow TileAndFuse to handle dynamic-shape matmuls


bjacob:
Details: https://github.com/iree-org/iree/issues/22087#issuecomment-3329869825

The testcase being referred to is in the preceding comment. It's just any matmul with a dynamic inner dim.

Tentatively assigning to Max based on the TODO comment :-)

nirvedhmeshram:
I think one of the key things to figure out is how to find the bounds of the GEMM shape when the shape is dynamic, maybe in most real world dispatches there will be some bounds that we can figure out with the value bounds/integer range analysis but what to do if such information is not present?

Max191:
> but what to do if such information is not present?

If a fully dynamic matmul comes in with no information about bounds, we probably want to just treat it as a very large GEMM. The "bounds" that we would see would techincally be very large (max int value), so I think it's better not to have a special case for this, and instead just treat it as if it is a very large GEMM. That means we will pad it to the largest square tile size possible.

nirvedhmeshram:
I did some very rough experimentation [here ](https://github.com/nirvedhmeshram/iree/tree/dynamic_experiment) to see if we can handle dynamic dims and the answer was yes, but I will work with @YashDeshpande25 on properly making this work

nirvedhmeshram:
Prefetching a test case

Case 1 (no bounds)
```
util.func @matmul_DYNxDYNxf16_times_DYNxDYNxf16_into_DYNxDYNxf32(%lhs: tensor<?x?xf16>, %rhs: tensor<?x?xf16>) -> tensor<?x?xf32> {
  %c0 = arith.constant 0 : index
  %c1 = arith.constant 1 : index
  %acc_dim0 = tensor.dim %lhs, %c0 : tensor<?x?xf16>
  %acc_dim1 = tensor.dim %rhs, %c1 : tensor<?x?xf16>
  %init_acc = tensor.empty(%acc_dim0, %acc_dim1) : tensor<?x?xf32>
  %c0_acc_type = arith.constant 0.0: f32
  %acc = linalg.fill ins(%c0_acc_type : f32) outs(%init_acc : tensor<?x?xf32>) -> tensor<?x?xf32>
[...]

PR #23448, opened 2026-02-10
Infer dynamic loop bounds from IntegerRangeAnalysis for GPU tile-and-fuse

YashDeshpande25:
Added function to determine bounds of dynamic dims in GEMMs using InterRangeAnalysis in Tile & Fuse Pipeline.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/BlockDynamicDimensions.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/BlockDynamicDimensions.cpp
@@ -10,6 +10,8 @@
 #include "iree/compiler/Codegen/Transforms/Transforms.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
 #include "iree/compiler/Dialect/LinalgExt/Transforms/Transforms.h"
+#include "iree/compiler/Dialect/LinalgExt/Utils/MatchUtils.h"
+#include "iree/compiler/Dialect/LinalgExt/Utils/Utils.h"
 #include "iree/compiler/Dialect/Util/IR/UtilDialect.h"
 #include "iree/compiler/Dialect/Util/IR/UtilOps.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
@@ -25,7 +27,7 @@ static llvm::cl::opt<bool> clEnableBlockedMatmuls(
     "iree-codegen-block-dynamic-dimensions-of-contractions",
     llvm::cl::desc("developer flag to gaurd blocking dynamic dimensions of "
                    "contraction-like ops"),
-    llvm::cl::Hidden, llvm::cl::init(true));
+    llvm::cl::Hidden, llvm::cl::init(false));
 
 namespace mlir::iree_compiler {
 
@@ -303,7 +305,9 @@ blockDynamicDimensions(RewriterBase &rewriter,
                                       attentionOp);
       })
       .Case([&](linalg::LinalgOp linalgOp) {
-        if (clEnableBlockedMatmuls) {
+        if (clEnableBlockedMatmuls ||
+            !(linalg::isaContractionOpInterface(linalgOp) ||
+              IREE::LinalgExt::isaScaledContractionOpInterface(linalgOp))) {
           return blockDynamicDimensions(rewriter, dynamicDimAnalysis, linalgOp);
         }
         return success();
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -7,6 +7,7 @@
 #include "iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.h"
 
 #include "iree/compiler/Codegen/Common/GPU/GPUHeuristics.h"
+#include "iree/compiler/Codegen/Common/TensorDynamicDimAnalysis.h"
 #include "iree/compiler/Codegen/Common/TileInferenceUtils.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
@@ -20,10 +21,15 @@
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
 #include "iree/compiler/Dialect/LinalgExt/Utils/MatchUtils.h"
 #include "iree/compiler/Dialect/LinalgExt/Utils/Utils.h"
+#include "llvm/ADT/DenseSet.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/DebugLog.h"
 #include "llvm/Support/InterleavedRange.h"
+#include "mlir/Analysis/DataFlow/ConstantPropagationAnalysis.h"
+#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
+#include "mlir/Analysis/DataFlow/IntegerRangeAnalysis.h"
+#include "mlir/Analysis/DataFlowFramework.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/IR/Attributes.h"
@@ -44,6 +50,14 @@ namespace mlir::iree_compiler::IREE::GPU {
 
 constexpr int64_t kCacheLineSizeBits = 128 * 8;
 constexpr int64_t kPreferredCopyNumBits = 128;
+
+// Sentinel value used by IntegerRangeAnalysis when bounds are unknown.
+static constexpr uint64_t MAX_DIM_VALUE = (static_cast<uint64_t>(1) << 53) - 1;
+
+// Fallback bound when IntegerRangeAnalysis cannot determine the actual value.
+// Kept small (2^14) to avoid int64_t overflow when dimensions are multiplied
+// together in heuristic calculations.
+static constexpr uint64_t MAX_BOUND_VALUE = static_cast<uint64_t>(1) << 14;
 
 //===----------------------------------------------------------------------===//
 // Lowering Config Selection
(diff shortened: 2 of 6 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1112,6 +1119,122 @@ LogicalResult setIGEMMConvolutionLoweringConfig(
...
+          if (operandDef) {
+            for (Value v : operandDef->getOperands()) {
+              if (v.getType().isIndex()) {
+                collectIndexValues(v, indexValues, visited);
+              }
+            }
+          }
+        }
+      }
+    };
```
nirvedhmeshram: do we we need to do this approach, looking at what compiler/src/iree/compiler/Codegen/Common/BlockDynamicDimensions.cpp
does, it uses `dynamicDimAnalysis.getDivisibilityInfo(v, index);`

you can similarly use
`dynamicDimAnalysis.getRangeInfo(v, index);` ?

nirvedhmeshram (commented):
I think the outer logic is what we want, however I think we can largely simplify the logic in `getLoopBoundsWithRangeAnalysis` by using `TensorDynamicDimAnalysis` provided in the compiler, you can closely follow `BlockDynamicDimensions` pass for this.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1122,7 +1196,19 @@ LogicalResult setMatmulLoweringConfig(IREE::GPU::TargetAttr target,
...
   }
 
-  SmallVector<int64_t> bounds = linalgOp.getStaticLoopRanges();
+  // Use IntegerRangeAnalysis to get better bounds for dynamic shapes.
+  bool boundsUsingAnalysis = false;
+  FailureOr<SmallVector<int64_t>> maybeBounds =
+      getLoopBoundsWithRangeAnalysis(linalgOp, entryPoint);
+  SmallVector<int64_t> bounds;
+  if (succeeded(maybeBounds)) {
+    boundsUsingAnalysis = true;
```
nirvedhmeshram: I think this bool will be set true even when the bounds you find are equal to the bounds of getStaticLoopRanges, but what we want is to only set this to true if you are making bounds that are different from the bounds  of getStaticLoopRanges.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1112,6 +1121,71 @@ LogicalResult setIGEMMConvolutionLoweringConfig(
...
+  TensorDynamicDimAnalysis dynamicDimAnalysis(entryPoint);
+  if (failed(dynamicDimAnalysis.run())) {
+    return linalgOp.getStaticLoopRanges();
+  }
+
+  SmallVector<int64_t> bounds = linalgOp.getStaticLoopRanges();
+  SmallVector<AffineMap> indexingMaps = linalgOp.getIndexingMapsArray();
+
+  // Sentinel value used by IntegerRangeAnalysis when bounds are unknown.
+  constexpr int64_t unboundedSentinel = 9007199254740991;
```
nirvedhmeshram: I think this is the proper way to define this
```
plugins/input/Torch/InputConversion/BindSymbolicShapes.cpp:39:static constexpr uint64_t MAX_DIM_VALUE = (static_cast<uint64_t>(1) << 53) - 1;
src/iree/compiler/Dialect/HAL/IR/HALOps.cpp:36:static constexpr uint64_t MAX_DIM_VALUE = (static_cast<uint64_t>(1) << 53) - 1;
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1112,6 +1121,71 @@ LogicalResult setIGEMMConvolutionLoweringConfig(
...
+      }
+
+      if (boundRefined) {
+        break;
+      }
+    }
+
+    // If we couldn't refine the bound, set it to a large value.
+    if (!boundRefined && ShapedType::isDynamic(bounds[loopIdx])) {
+      bounds[loopIdx] = 1 << 20;
```
nirvedhmeshram: Lets make this a constant like MAX_DIM_VALUE, lets call it `MAX_BOUND_VALUE` for this explain in a comment why we are making it 1<<20, the reason being these can be multiplied with other dimensions in the heursitic logic so we dont want something too large, I also think 1<<14 (which is 16384) is big enough.

krzysz00: ... yeah, I'm suspicious of a maximum that implies a dimension bound fits in 32 bits, but also it sounds like this is in a heuristic context

nirvedhmeshram: yup this is just to coax the heuristic to let it give us a nice tile and also be safe for it to multiply the dims.

compiler/src/iree/compiler/Codegen/Common/BlockDynamicDimensions.cpp
```diff
@@ -25,7 +25,7 @@ static llvm::cl::opt<bool> clEnableBlockedMatmuls(
     "iree-codegen-block-dynamic-dimensions-of-contractions",
     llvm::cl::desc("developer flag to gaurd blocking dynamic dimensions of "
                    "contraction-like ops"),
-    llvm::cl::Hidden, llvm::cl::init(true));
+    llvm::cl::Hidden, llvm::cl::init(false));
```
nirvedhmeshram: we dont want to change the default behavior of the pass, we want the pass to leave contraction ops alone as the TAF pipeline can handle them without blocking dynamic dims.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1112,6 +1129,68 @@ LogicalResult setIGEMMConvolutionLoweringConfig(
       workgroupSize, targetSubgroupSize, pipelineConfig);
 }
 
+static FailureOr<SmallVector<int64_t>>
```
nirvedhmeshram: I dont think you are every returning failure so you dont need the `FailureOr`? 
also you can instead return
std::pair<SmallVector<int64_t>, bool> where the bool is if any bound was refined or not. I think you will need to `||` the current refinedBounds that you have per dim and also set them as true when you use the `MAX_BOUND_VALUE`

and then on the caller side you can do

```
std::pair<SmallVector<int64_t>, bool> inferredBounds =
      getLoopBoundsWithRangeAnalysis(linalgOp, entryPoint);

SmallVector<int64_t> bounds = inferredBounds.first
bool boundsUsingAnalysis = inferredBounds.second
```

nirvedhmeshram (commented):
Looking good, just need to fix the config tests and address the comments here.

compiler/src/iree/compiler/Codegen/Common/BlockDynamicDimensions.cpp
nirvedhmeshram: oh wait I misunderstood what this flag was doing, I think doing it this way is fine, but lets add back the tests I asked you to remove since they will still work.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
krzysz00: I'm just worried this is a footgun waiting to happen

nirvedhmeshram: Basically if the user doesnt provide meaningful bounds we have to make some assumptions, we agreed on doing this here https://github.com/iree-org/iree/issues/22089#issuecomment-3338690539

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1122,7 +1204,11 @@ LogicalResult setMatmulLoweringConfig(IREE::GPU::TargetAttr target,
     return failure();
   }
 
-  SmallVector<int64_t> bounds = linalgOp.getStaticLoopRanges();
+  // Use IntegerRangeAnalysis to get better bounds for dynamic shapes.
+  std::pair<SmallVector<int64_t>, bool> inferredBounds =
```
krzysz00, line 1207: To rephrase my concern, that function name is making me very concerned that someone's going to think these values are precise when they really aren't.

I'd use that 1 << 53 maximum if you must, and go prevasively rework the heuristics to use saturating math ... or use the upstream thing where dynamic * anything == dynamic

But that's a fairly big change, and I think the short-term fix here is an appropriate pile of warning signs

nirvedhmeshram: This is not user visible anywhere though, its just for us to come up with tile sizes, the underlying size of the problem is dynamic anyway so I don't actually see your concern. To elaborate let's say umax was 1024. We are fine with it being treated as 1024 when in reality it could be anything from 1 to 1024. I don't see this case as any different. We have to estimate a bound and then get a tile size based on that. In the example I gave I do want the heuristic logic to still precisely think its 1024 and make the best tile size we can. Similarly in this case I want a logic for a large bound so I reasonable power of 2 is ok. If the user program care about this they should provide better bounds otherwise dynamic size bounds are acceptable to be best effort.

nirvedhmeshram: Anyway once this is ready for review let's have @Max191 take a look to have a third opinion.

krzysz00: To rephrase what I take an issue with, a function that returns upper bounds that doesn't actually return upper bounds and might instead return a value like 1 << 14 or 1 << 20 (which is, I claim, well within reasonable sizes for a matmul dimension) is a dangerous thing to have the name and interface it has

If this function was called `inferBoundsTreatingDynamicAsLargeConstant` , I'd probably drop the complaints here - that's a more accurate name that is long enough to set off the reader's danger sense.

nirvedhmeshram: That name is not true though, it would have to be `inferBoundsTreatingUnboundedDynamicAsLargeConstant` which sounds better to be made a comment than a function name in my opinion. 
Also 1 << 20 has no value over 1 << 14. It would find the same tile sizes but would be more likely to have overflow in the heuristics. Note we can have 1 << 14 as a bound and the dynamic dim might end up 1 << 20 at runtime and things will still work.

krzysz00: Right, but this particular means of handling of unbounded dynamic dimensions ... and I'll suggest that the 1 << 53 bound is, in fact, "unbounded" and should be made into 1 << 14 here - is useful for heuristics but unsound generally so it should be *very* obvious to the reader that there's some trickery going on here.

PR #23451, opened 2026-02-10
[GPU] Add iree_gpu.global_subgroup_barrier op

qedawkins:
Add a synchronization-only barrier op that has no memory fence semantics. The key distinction between this and `gpu.barrier` is that it's semantically global. That is, a subgroup is let through a barrier once all subgroups have reached *any* instance of the barrier, not just a specific one. Note that this is a dramatically more restrictive condition for optimizing the barriers themselves and should only be preferred in situations where it is expressly required. This op also does not fence memory and expects that to be handled separately.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
@@ -343,4 +343,23 @@ def IREEGPU_CoalescedGatherDMAOp : Op<IREEGPU_Dialect, "coalesced_gather_dma", [
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// GlobalSubgroupBarrierOp
+//===----------------------------------------------------------------------===//
+
+def IREEGPU_GlobalSubgroupBarrierOp : Op<IREEGPU_Dialect,
+    "global_subgroup_barrier", []> {
+  let summary = "Synchronization-only barrier across all subgroups.";
+  let description = [{
+    All subgroups in the workgroup must reach any instance of this op before
+    any can proceed past it. Unlike `gpu.barrier`, this op has no memory
+    fence semantics - memory operations can be freely reordered with respect
+    to it. Fences are handled separately.
+  }];
+
+  let arguments = (ins);
+  let results = (outs);
+  let assemblyFormat = "attr-dict";
+}
+
 #endif // IREE_CODEGEN_DIALECT_IREEGPUOPS
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
@@ -355,6 +355,10 @@ def IREEGPU_GlobalSubgroupBarrierOp : Op<IREEGPU_Dialect,
     any can proceed past it. Unlike `gpu.barrier`, this op has no memory
     fence semantics - memory operations can be freely reordered with respect
     to it. Fences are handled separately.
+
+    This operation operates on a single, global workgroup-wide barrier object
+    that can implement such dynamic join and must not be used if such a barrier
+    isn't available.
   }];
 
   let arguments = (ins);
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToNVVM.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToNVVM.cpp
@@ -45,9 +45,9 @@ namespace {
 
 // Lower iree_gpu.global_subgroup_barrier to just the hardware barrier
 // instruction, with NO memory fences. Fences are handled separately.
-struct LowerGlobalSubgroupBarrier
-    : public OpRewritePattern<IREE::GPU::GlobalSubgroupBarrierOp> {
-  using OpRewritePattern::OpRewritePattern;
+struct LowerGlobalSubgroupBarrier final
+    : OpRewritePattern<IREE::GPU::GlobalSubgroupBarrierOp> {
+  using Base::Base;
 
   LogicalResult matchAndRewrite(IREE::GPU::GlobalSubgroupBarrierOp op,
                                 PatternRewriter &rewriter) const override {
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToNVVM.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToNVVM.cpp
@@ -7,6 +7,7 @@
 #include "iree/compiler/Codegen/Common/GPU/GPUPatterns.h"
 #include "iree/compiler/Codegen/Common/Transforms.h"
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUDialect.h"
+#include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.h"
 #include "iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
 #include "mlir/Conversion/ArithToLLVM/ArithToLLVM.h"
@@ -41,6 +42,24 @@ namespace mlir::iree_compiler {
 #include "iree/compiler/Codegen/LLVMGPU/Passes.h.inc"
 
 namespace {
+
+// Lower iree_gpu.global_subgroup_barrier to just the hardware barrier
+// instruction, with NO memory fences. Fences are handled separately.
+struct LowerGlobalSubgroupBarrier
+    : public OpRewritePattern<IREE::GPU::GlobalSubgroupBarrierOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(IREE::GPU::GlobalSubgroupBarrierOp op,
+                                PatternRewriter &rewriter) const override {
+    rewriter.replaceOpWithNewOp<NVVM::Barrier0Op>(op);
+    return success();
+  }
+};
+
+static void
+populateLowerGlobalSubgroupBarrierPatterns(RewritePatternSet &patterns) {
+  patterns.add<LowerGlobalSubgroupBarrier>(patterns.getContext());
+}
 
 /// A pass that replaces all occurrences of GPU device operations with their
 /// corresponding NVVM equivalent.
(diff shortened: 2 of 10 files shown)


qedawkins:
@krzysz00 I figure you're the best to take a look at this. Basically we need to distinguish between "The" barrier and "a" barrier. gpu.barrier (mainly per the canonicalization that drops adjacent ones) semantically requires all threads to reach that specific barrier to proceed, while we need the semantics "all threads must reach any *instance* of the barrier to proceed" for wave specialization on targets that don't have named barriers.

If you have ideas other than a new op I'd be interested to hear your thoughts.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_rocdl_gfx1200.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: iree-opt --iree-gpu-test-target=gfx1200 --iree-convert-to-rocdl %s | FileCheck %s
```
kuhar: Why test against gfx1200 instead of gfx1201? I mean it's fine since it's a LIT tests, but we don't really have any gfx1200 hw for runtime tests

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToNVVM.cpp
```diff
@@ -42,6 +43,24 @@ namespace mlir::iree_compiler {
 
 namespace {
 
+// Lower iree_gpu.global_subgroup_barrier to just the hardware barrier
+// instruction, with NO memory fences. Fences are handled separately.
+struct LowerGlobalSubgroupBarrier
+    : public OpRewritePattern<IREE::GPU::GlobalSubgroupBarrierOp> {
+  using OpRewritePattern::OpRewritePattern;
```
kuhar:
```suggestion
struct LowerGlobalSubgroupBarrier final
    : OpRewritePattern<IREE::GPU::GlobalSubgroupBarrierOp> {
  using Base::Base;
```

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
```diff
@@ -61,6 +62,58 @@ static llvm::cl::opt<int> clROCMIndexingBitsDeprecated(
...
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    constexpr amdgpu::Chipset kGfx90a(9, 0, 0x0a);
+
+    if (chipset < kGfx90a) {
+      // Pre-gfx90a: use inline asm.
+      auto asmDialectAttr = LLVM::AsmDialectAttr::get(rewriter.getContext(),
+                                                      LLVM::AsmDialect::AD_ATT);
+      const char *asmStr = ";;;WARNING: BREAKS DEBUG WATCHES\ns_barrier";
+      rewriter.replaceOpWithNewOp<LLVM::InlineAsmOp>(
```
kuhar, line 88: Would it be worth moving to rocdl?

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_rocdl_gfx1200.mlir
qedawkins: Oh I didn't know that. This was just a chipset that exercised the behavior I wanted, happy to change to 1201.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
qedawkins: WDYM moving to rocdl? Moving the op to rocdl?

compiler/src/iree/compiler/Codegen/LLVMGPU/test/convert_to_rocdl_gfx1200.mlir
kuhar: 9070xt / r9700 is gfx1201

kuhar: you could actually have two run lines, doesn't matter, just commented because it's a bit unexpected

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
kuhar: I'm generally leaning towards not having inline asm in iree and using rocdl for these low-level intrinsics

krzysz00:
@qedawkins This is `gpu.barrier memfence []'

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
```diff
@@ -61,6 +62,58 @@ static llvm::cl::opt<int> clROCMIndexingBitsDeprecated(
...
+  LogicalResult matchAndRewrite(IREE::GPU::GlobalSubgroupBarrierOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    constexpr amdgpu::Chipset kGfx90a(9, 0, 0x0a);
+
+    if (chipset < kGfx90a) {
+      // Pre-gfx90a: use inline asm.
+      auto asmDialectAttr = LLVM::AsmDialectAttr::get(rewriter.getContext(),
+                                                      LLVM::AsmDialect::AD_ATT);
+      const char *asmStr = ";;;WARNING: BREAKS DEBUG WATCHES\ns_barrier";
```
krzysz00, line 87: 1. No it doesn't break debug watches
2. Use the intrinsic on pain of getting a stern talking to from the compiler team unless you know a good reason not to

krzysz00 (changes requested):
Holding

qedawkins:
Not quite,
```
gpu.barrier memfence []
gpu.barrier memfence []
```
this folds because the barriers are unique, meaning we can assume the second barrier is redundant. The same folder is invalid for global barriers since other workers can have a critical region between the two.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
krzysz00: I'll note that the standing position of the compiler folks is "if you have felt the need to use inline asm, please don't, but if  you must, tell us about it so we can see if your usecase can be eliminated"

krzysz00:
> Not quite,
> 
> ```
> gpu.barrier memfence []
> gpu.barrier memfence []
> ```
> 
> this folds because the barriers are unique, meaning we can assume the second barrier is redundant. The same folder is invalid for global barriers since other workers can have a critical region between the two.

In which case, can you go upstream and patch the folder to special-case the no-memfence case?

qedawkins:
`amdgpu.lds_barrier` folds for the same reason (as does any version of this op with memfences on it). The issue is one of semantics, unless we want to change that.

krzysz00:
Re-reading, I can see why you'd want this - it's to make sure we have correctness around pingpong

I'd still want to re-use upstream implementations here and see if maybe we don't want to fiddle with the folders up there instead of adding a new op that'l awkwardly diverge

krzysz00:
`gpu.barrier memfence []` is specifically "a barrier with no memory fencing semantics" and so its folders should probably be updated accordingly

qedawkins:
> `gpu.barrier memfence []` is specifically "a barrier with no memory fencing semantics" and so its folders should probably be updated accordingly

The memory fence is a separate discussion. Whether or not `gpu.barrier` fences memory is independent of the folder, so it really is a semantics change to the upstream barrier (either the folder never applies or always applies to sequential identical barriers). If we want to go with changing the semantics of gpu.barrier and dropping the folder I'm good with that. Though I do prefer not having to engage with discussions about it upstream because the current semantics + folder are also completely reasonable as a use case.

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
qedawkins: https://github.com/llvm/llvm-project/blob/f9bca14a33888e70e2f9f1aa1f4a4edd3d7e5dfd/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp#L601

This is just a direct copy-paste of this. If you guys want to tell me what to put here I'd be happy to put whatever.

krzysz00:
@qedawkins Ok, I can see what you're getting at now, though I'm at least going to argue for wandering upstream and either explicitly adding a carve-out to the uniqueness thing when you specify `memfence []` or doing something interesting like allowing for named barriers that don't allow the folding (barriers can have an optional `id` for example - or we'd do a named barrier object sort of thing and just only support the trivial usage on gfx9)

compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
krzysz00: On re-reading, it's probably fine to keep the MI-100 workaround in here, and that comment's even correct.

qedawkins:
> @qedawkins Ok, I can see what you're getting at now, though I'm at least going to argue for wandering upstream and either explicitly adding a carve-out to the uniqueness thing when you specify `memfence []` or doing something interesting like allowing for named barriers that don't allow the folding (barriers can have an optional `id` for example - or we'd do a named barrier object sort of thing and just only support the trivial usage on gfx9)

I'm not interested in adding a flag to gpu.barrier for these semantics just to kill a folder. And justifying the semantics I'm adding here in a vacuum upstream does not sound appealing to me. Re: named barriers, I'm expecting we'll need something shaped fairly different since typically you need a signal and an await for those no?

krzysz00:
SPIR-V has, IIRC, named barriers but not a split signal/wait - but that's off-topic

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -343,4 +343,23 @@ def IREEGPU_CoalescedGatherDMAOp : Op<IREEGPU_Dialect, "coalesced_gather_dma", [
...
+
+def IREEGPU_GlobalSubgroupBarrierOp : Op<IREEGPU_Dialect,
+    "global_subgroup_barrier", []> {
+  let summary = "Synchronization-only barrier across all subgroups.";
+  let description = [{
+    All subgroups in the workgroup must reach any instance of this op before
+    any can proceed past it. Unlike `gpu.barrier`, this op has no memory
+    fence semantics - memory operations can be freely reordered with respect
+    to it. Fences are handled separately.
+  }];
```
krzysz00, line 362:
```suggestion

This operation operates on a single, global workgroup-wide barrier object that can implement such dynamic join and must not be used if such a barrier isn't available.
  }];
```

krzysz00 (approved):
Approved now that we've got more context on this one

kuhar (approved):

PR #23452, opened 2026-02-11
[Codegen][PCF] Add foldForallIntoPCFLoop for split-k loop handling

qedawkins:
Adds a generic `foldForallIntoPCFLoop` function that folds an `scf.forall` containing a `pcf.loop` into a single `pcf.generic` operation. This enables incorporating user expressed extra levels of parallelism into a scope. The immediate use case this enables is split-k by folding the split-k loop into the workgroup loop.

Key changes:
- Add `foldForallIntoPCFLoop` API in PCF/Transforms/Transforms.h
- Implement structural matching helpers (matchFoldTerminator, matchFoldPCFLoop, matchFoldWriteSlices) that validate requirements without scope-specific logic
- Add TestFoldForallIntoPCFLoopPass for unit testing with local_mapping + sequential scope
- Add FoldSplitKWorkgroupLoop pattern in ConvertWorkgroupForallToPCF.cpp that matches split_reduction_mapping + workgroup_scope and calls the generic fold
- Run fold pattern as second pass in ConvertWorkgroupForallToPCFPass

The fold operation:
1. Creates pcf.generic with same scope as inner pcf.loop
2. Linearizes forall iteration space and delinearizes inside generic
3. Converts pcf.loop to a nested scf.forall loop to handle spillover
4. Composes tensor.parallel_insert_slice with pcf.write_slice ops

The reason this pattern has to generate a pcf.generic instead of pcf.loop is to avoid extra execution of code that was inside
the scf.forall but not inside the pcf.loop.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertWorkgroupForallToPCF.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertWorkgroupForallToPCF.cpp
@@ -9,6 +9,10 @@
 #include "iree/compiler/Codegen/Dialect/PCF/IR/PCF.h"
 #include "iree/compiler/Codegen/Dialect/PCF/Transforms/Transforms.h"
 #include "iree/compiler/Codegen/Transforms/Transforms.h"
+#include "iree/compiler/Codegen/Utils/GPUUtils.h"
+#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtDialect.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Arith/Utils/Utils.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/WalkPatternRewriteDriver.h"
@@ -20,6 +24,17 @@ namespace mlir::iree_compiler {
 
 namespace {
 struct ConvertWorkgroupForall : OpRewritePattern<scf::ForallOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(scf::ForallOp op,
+                                PatternRewriter &rewriter) const override;
+};
+
+/// Folds an scf.forall with split-reduction mapping containing a pcf.loop
+/// with workgroup scope into a single pcf.generic. This handles the "split-k"
+/// pattern where the outer forall represents additional parallelism from
+/// K-dimension splitting and the inner pcf.loop represents the original
+/// workgroup-level iteration.
+struct FoldSplitKWorkgroupLoop : OpRewritePattern<scf::ForallOp> {
   using Base::Base;
   LogicalResult matchAndRewrite(scf::ForallOp op,
                                 PatternRewriter &rewriter) const override;
@@ -69,10 +84,94 @@ ConvertWorkgroupForall::matchAndRewrite(scf::ForallOp op,
   return success();
 }
 
+LogicalResult
+FoldSplitKWorkgroupLoop::matchAndRewrite(scf::ForallOp op,
+                                         PatternRewriter &rewriter) const {
+  // Match scf.forall with split-reduction mapping.
+  if (!forallOpHasMappingType<IREE::LinalgExt::SplitReductionMappingAttr>(op)) {
+    return failure();
+  }
+
+  // Find pcf.loop with workgroup scope in the forall body.
+  IREE::PCF::LoopOp loopOp = nullptr;
+  for (Operation &bodyOp : op.getBody()->without_terminator()) {
+    if (auto loop = dyn_cast<IREE::PCF::LoopOp>(&bodyOp)) {
+      if (isa<IREE::Codegen::WorkgroupScopeAttr>(loop.getScope())) {
+        loopOp = loop;
+        break;
+      }
+    }
+  }
+  if (!loopOp) {
+    return failure();
+  }
+
+  // Capture values needed for workgroup count computation before folding.
+  // The forall upper bounds and loop counts are defined outside the forall
+  // so they remain valid after the fold erases it.
+  Location loc = op.getLoc();
+  SmallVector<OpFoldResult> upperBounds = op.getMixedUpperBound();
+  SmallVector<Value> loopCounts(loopOp.getCount());
+
+  // Fold forall + pcf.loop into pcf.generic.
+  FailureOr<IREE::PCF::GenericOp> result =
+      IREE::PCF::foldForallIntoPCFLoop(rewriter, op);
+  if (failed(result)) {
+    return failure();
+  }
+
+  // Compute total workgroup count after folding (forall iterations * loop
+  // count). Generate IR before the pcf.generic.
+  rewriter.setInsertionPoint(*result);
+
+  Value forallCount = nullptr;
+  for (OpFoldResult ub : upperBounds) {
+    Value ubVal = getValueOrCreateConstantIndexOp(rewriter, loc, ub);
+    if (!forallCount) {
+      forallCount = ubVal;
+    } else {
+      forallCount = arith::MulIOp::create(rewriter, loc, forallCount, ubVal);
+    }
+  }
+
+  Value totalLoopCount = nullptr;
+  for (Value count : loopCounts) {
+    if (!totalLoopCount) {
+      totalLoopCount = count;
+    } else {
+      totalLoopCount =
+          arith::MulIOp::create(rewriter, loc, totalLoopCount, count);
+    }
+  }
+
+  Value totalCount =
+      arith::MulIOp::create(rewriter, loc, forallCount, totalLoopCount);
+
+  // Create workgroup count hint for the folded generic.
+  SmallVector<OpFoldResult> counts = {OpFoldResult(totalCount)};
+  [[maybe_unused]] LogicalResult hintRes = createWorkgroupCountHint(
+      rewriter, loc, counts, /*maxWorkgroupParallelDims=*/1,
+      /*reverse=*/false);
+  assert(succeeded(hintRes) &&
+         "Unexpected failure to construct workgroup count hint");
+
+  return success();
+}
+
 void ConvertWorkgroupForallToPCFPass::runOnOperation() {
-  RewritePatternSet patterns(&getContext());
-  patterns.add<ConvertWorkgroupForall>(&getContext());
-  walkAndApplyPatterns(getOperation(), std::move(patterns));
+  // First pass: Convert workgroup foralls to pcf.loop.
+  {
+    RewritePatternSet patterns(&getContext());
+    patterns.add<ConvertWorkgroupForall>(&getContext());
+    walkAndApplyPatterns(getOperation(), std::move(patterns));
+  }
+
+  // Second pass: Fold split-k foralls containing pcf.loop into pcf.generic.
+  {
+    RewritePatternSet patterns(&getContext());
+    patterns.add<FoldSplitKWorkgroupLoop>(&getContext());
+    walkAndApplyPatterns(getOperation(), std::move(patterns));
+  }
 }
 
 } // namespace mlir::iree_compiler
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertWorkgroupForallToPCF.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertWorkgroupForallToPCF.cpp
@@ -11,6 +11,7 @@
 #include "iree/compiler/Codegen/Transforms/Transforms.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtDialect.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Arith/Utils/Utils.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
@@ -124,13 +125,32 @@ FoldSplitKWorkgroupLoop::matchAndRewrite(scf::ForallOp op,
   // count). Generate IR before the pcf.generic.
   rewriter.setInsertionPoint(*result);
 
+  // Account for lower bounds and steps when computing iteration counts.
+  SmallVector<OpFoldResult> lowerBounds = op.getMixedLowerBound();
+  SmallVector<OpFoldResult> steps = op.getMixedStep();
+
+  AffineExpr s0, s1, s2;
+  bindSymbols(rewriter.getContext(), s0, s1, s2);
+  AffineExpr numItersExpr = (s0 - s1).ceilDiv(s2);
+
   Value forallCount = nullptr;
-  for (OpFoldResult ub : upperBounds) {
-    Value ubVal = getValueOrCreateConstantIndexOp(rewriter, loc, ub);
+  for (int64_t i = 0, e = upperBounds.size(); i < e; ++i) {
+    OpFoldResult lb = i < (int64_t)lowerBounds.size()
+                          ? lowerBounds[i]
+                          : rewriter.getIndexAttr(0);
+    OpFoldResult ub = upperBounds[i];
+    OpFoldResult step =
+        i < (int64_t)steps.size() ? steps[i] : rewriter.getIndexAttr(1);
+
+    Value iterCount = getValueOrCreateConstantIndexOp(
+        rewriter, loc,
+        affine::makeComposedFoldedAffineApply(rewriter, loc, numItersExpr,
+                                              {ub, lb, step}));
     if (!forallCount) {
-      forallCount = ubVal;
+      forallCount = iterCount;
     } else {
-      forallCount = arith::MulIOp::create(rewriter, loc, forallCount, ubVal);
+      forallCount =
+          arith::MulIOp::create(rewriter, loc, forallCount, iterCount);
     }
   }
 
--- a/compiler/src/iree/compiler/Codegen/Common/ConvertWorkgroupForallToPCF.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ConvertWorkgroupForallToPCF.cpp
@@ -108,10 +108,12 @@ FoldSplitKWorkgroupLoop::matchAndRewrite(scf::ForallOp op,
   }
 
   // Capture values needed for workgroup count computation before folding.
-  // The forall upper bounds and loop counts are defined outside the forall
-  // so they remain valid after the fold erases it.
+  // The fold erases the forall op, so any mixed bounds/steps must be
+  // materialized up front.
   Location loc = op.getLoc();
+  SmallVector<OpFoldResult> lowerBounds = op.getMixedLowerBound();
   SmallVector<OpFoldResult> upperBounds = op.getMixedUpperBound();
+  SmallVector<OpFoldResult> steps = op.getMixedStep();
   SmallVector<Value> loopCounts(loopOp.getCount());
 
   // Fold forall + pcf.loop into pcf.generic.
@@ -124,10 +126,6 @@ FoldSplitKWorkgroupLoop::matchAndRewrite(scf::ForallOp op,
   // Compute total workgroup count after folding (forall iterations * loop
   // count). Generate IR before the pcf.generic.
   rewriter.setInsertionPoint(*result);
-
-  // Account for lower bounds and steps when computing iteration counts.
-  SmallVector<OpFoldResult> lowerBounds = op.getMixedLowerBound();
-  SmallVector<OpFoldResult> steps = op.getMixedStep();
 
   AffineExpr s0, s1, s2;
   bindSymbols(rewriter.getContext(), s0, s1, s2);
--- a/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
@@ -1,0 +1,546 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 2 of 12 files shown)


compiler/src/iree/compiler/Codegen/Common/ConvertWorkgroupForallToPCF.cpp
```diff
@@ -69,10 +84,94 @@ ConvertWorkgroupForall::matchAndRewrite(scf::ForallOp op,
...
+    return failure();
+  }
+
+  // Compute total workgroup count after folding (forall iterations * loop
+  // count). Generate IR before the pcf.generic.
+  rewriter.setInsertionPoint(*result);
+
+  Value forallCount = nullptr;
+  for (OpFoldResult ub : upperBounds) {
+    Value ubVal = getValueOrCreateConstantIndexOp(rewriter, loc, ub);
```
MaheshRavishankar: I dont know if thats enough. You need to account for steps in the forall and also the lower bound. You probably just want to normalize the scf.forall before doing this?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,546 @@
...
+static PCF::GenericOp foldForallIntoPCFLoopImpl(RewriterBase &rewriter,
+                                                scf::ForallOp forallOp,
+                                                PCF::LoopOp loopOp) {
+  Location loc = forallOp.getLoc();
+  scf::InParallelOp terminator = forallOp.getTerminator();
+
+  // Replace RegionIterArgs with initial values (except in terminator).
+  for (auto [iterArg, init] :
+       llvm::zip(forallOp.getRegionIterArgs(), forallOp.getOutputs())) {
+    rewriter.replaceUsesWithIf(iterArg, init, [&](OpOperand &use) {
```
MaheshRavishankar: Just for me to understand, do we really care about this case? This is a race AFAICS.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,546 @@
...
+      /*scope=*/loopOp.getScope(),
+      /*inits=*/forallOp.getOutputs(),
+      /*dynamic_sizes=*/ValueRange{},
+      /*is_tied=*/SmallVector<bool>(forallOp.getNumResults(), true),
+      /*num_iterators=*/1);
+
+  // Set sync scope to SyncOnReturn for the pcf.generic sref arguments.
+  Attribute syncScope = PCF::SyncOnReturnAttr::get(rewriter.getContext());
+  for (auto regionRefArg : genericOp.getRegionRefArgs()) {
+    auto srefType = cast<PCF::ShapedRefType>(regionRefArg.getType());
```
MaheshRavishankar: Did you already check earlier that everything has syncScope. Then isnt it better to just "transfer" the scope that existed originally?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,546 @@
...
+  for (int64_t i = 0, e = numDims; i < e; ++i) {
+    iterCountValues.push_back(
+        getValueOrCreateConstantIndexOp(rewriter, loc, iterCountOFRs[i]));
+  }
+  OpFoldResult totalItersOFR;
+  {
+    AffineExpr prod = s0 * s1;
+    totalItersOFR = iterCountOFRs[0];
+    for (int64_t i = 1, e = numDims; i < e; ++i) {
+      totalItersOFR = affine::makeComposedFoldedAffineApply(
```
MaheshRavishankar: You probably want to use `affine.linearize` here since linearize/delinearize pairs might better cancel out.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,546 @@
...
+      }
+      writesByResult[idx].push_back(writeOp);
+    }
+  }
+
+  return success();
+}
+
+/// Core implementation of foldForallIntoPCFLoop after matching succeeds.
+static PCF::GenericOp foldForallIntoPCFLoopImpl(RewriterBase &rewriter,
```
MaheshRavishankar: This method is way too big. Would be more readable if you could break it up into smaller functional pieces.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,546 @@
...
+        }
+
+        auto expandedType =
+            RankedTensorType::get(expandedShape, sourceType.getElementType());
+        expandedSource = tensor::ExpandShapeOp::create(
+            rewriter, writeOp.getLoc(), expandedType, writeOp.getSource(),
+            reassociation);
+      }
+
+      // Compose offsets, sizes, and strides.
```
MaheshRavishankar: Looks like we are doing this composition a few times....

  FusePCFWrites.cpp (composeWriteSliceWithParallelInsert):                                                                                                                                                                                                                                                                                                              
  - offsets: writeOffset + insertOffset * writeStride                                                                                                                                                                                                                                                                                                                   
  - sizes: insertSizes                                                                                                                                                                                                                                                                                                                                                  
  - strides: writeStride * insertStride

  FoldForallIntoPCFLoop.cpp (the PR):
  - offsets: writeOffset + insertOffset (just addition, no scaling)
  - sizes: writeSizes (not insert sizes)
  - strides: writeStride * insertStride

  Two differences:
  1. Offsets: The PR doesn't scale insertOffset by writeStride. This gives wrong results when the write stride is non-unit — the insert offset is in the destination coordinate space, so it needs to be scaled by the write stride.
  2. Sizes: The PR takes sizes from the write_slice, while FusePCFWrites takes them from the insert_slice.

  The FusePCFWrites formula matches the upstream affine::mergeOffsetsSizesAndStrides convention (producer_offsets * consumer_strides + consumer_offsets). The PR's version looks like a bug for the non-unit stride case

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+  Value totalWorkers = genericOp.getCountArgs()[0];
+  // Each worker handles ceil(totalWorkers / loopCount) iterations.
+  AffineExpr ceilDiv = s0.ceilDiv(s1);
+  Value outerStep = getValueOrCreateConstantIndexOp(
+      rewriter, loc,
+      affine::makeComposedFoldedAffineApply(rewriter, loc, ceilDiv,
+                                            {totalWorkers, loopCount}));
+
+  auto outerForall = scf::ForallOp::create(
```
MaheshRavishankar: I am really confused why we are creating this `outerForall` loop.

MaheshRavishankar (commented):
I did some review here, but I think some refactoring could really help follow the lgoic better (for someone who doesnt have the pcf semantics burnt into their heads). Please make the main method more readable and abstract out functionality into smaller methods instead of inlining all the logic into one giant function.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: I don't follow how this is a race? Nearly 100% of forall uses match this check
```
scf.forall shared_outs(%out) {
  %0 = tensor.extract_slice %out // != terminator
  scf.forall.in_parallel {
    tensor.parallel_insert_slice %0 into %out // == terminator
  }
```

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: There's no original scope to copy over here. The `scf.forall` is the outer op and this code is setting the syncScope of the generic op we're creating that returns the values the scf.forall originally returned (which doesn't have a syncScope).

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: It's because `pcf.generic` runs exactly as many iterations as threads while `pcf.loop` and `scf.forall` maps the number of threads to the number of iterations. So we need to resolve the over/undersubscription part of the `pcf.loop` and `scf.forall` while doing the fusion and that can either be done with an `scf.for` or an `scf.forall` with no mapping. The `innerForall` is for the loop and this one is for the `scf.forall`. I've been using mapping-less `scf.forall` since it's easier to create a multi-nest and captures slightly more information if we ever want to optimize around it.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: Fixed, I can look into changing everything to use `mergeOffsetsSizesAndStrides` too as a separate change.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: Broke it up a bit splitting out index related stuff and keeping the loop structure construction here. Let me know if you want to further partition it.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.td
```diff
@@ -32,6 +32,36 @@ def ConvertForallToLoopsPass : Pass<"iree-pcf-convert-forall-to-loops", ""> {
...
+    The input is IR containing `scf.forall` ops with `iree_codegen.local_mapping`
+    attributes that contain a `pcf.loop` with `#pcf.sequential` scope as the
+    last operation before the terminator.
+
+    Structural requirements validated:
+    - All ops in `scf.forall.in_parallel` are `tensor.parallel_insert_slice`.
+    - All insert sources come from the same `pcf.loop` result.
+    - All insert destinations are `scf.forall` shared_outs.
+    - The `pcf.loop` is the last op before terminator in forall body.
+    - The `pcf.loop` has a single count argument (linearized).
```
MaheshRavishankar, line 50: I am guessing this requirement is because we linearize the loop to be able to just launch a 1D grid, and you are working off of that?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,607 @@
...
+/// - Loop is last op before terminator.
+static LogicalResult matchFoldPCFLoop(scf::ForallOp forallOp,
+                                      PCF::LoopOp loopOp) {
+  // Single count argument required.
+  if (loopOp.getCount().size() != 1) {
+    return failure();
+  }
+
+  // Count must dominate the forall.
+  Value countValue = loopOp.getCount()[0];
```
MaheshRavishankar: You should just be using the method to check dominance. Better yet, you can just move the definition before `scf.forall`. There is a utility method upstream to do this.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,607 @@
...
+      forallOp.getRegion().front().getTerminator()->getPrevNode();
+  if (lastOp != loopOp) {
+    return failure();
+  }
+
+  return success();
+}
+
+/// Validates pcf.loop region ref args:
+/// - All users are pcf.write_slice ops.
```
MaheshRavishankar, line 107: Just for my understanding, when would we not do that? Is that even defined?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,607 @@
...
+  SmallVector<OpFoldResult> steps = forallOp.getMixedStep();
+
+  int64_t numDims = upperBounds.size();
+  AffineExpr s0, s1, s2;
+  bindSymbols(rewriter.getContext(), s0, s1, s2);
+  AffineExpr numItersExpr = (s0 - s1).ceilDiv(s2);
+
+  SmallVector<OpFoldResult> iterCountOFRs;
+  for (int64_t i = 0, e = numDims; i < e; ++i) {
+    OpFoldResult lb = i < (int64_t)lowerBounds.size()
```
MaheshRavishankar: unnecessary check here and below. I think lowerBounds.size() is always equal to upperBounds.size() and same for steps.size()

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,607 @@
...
+    auto loopResult = cast<OpResult>(insertOp.getSource());
+    unsigned resultIdx = loopResult.getResultNumber();
+    auto destArg = cast<BlockArgument>(insertOp.getDest());
+    unsigned argIdx = destArg.getArgNumber() - forallOp.getRank();
+    resultToRefArgIdx[resultIdx] = argIdx;
+  }
+
+  AffineExpr s0, s1, s2;
+  bindSymbols(rewriter.getContext(), s0, s1, s2);
+  // Composition formula (note: opposite direction from FusePCFWrites):
```
MaheshRavishankar: I dont quite udnerstand the opposite direction comment. You have two sets of offsets, sizes and strides. The "direction" could be fixed, and the caller decides which set to pass as the first set and the second set?

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+      SmallVector<OpFoldResult> writeOffsets = writeOp.getMixedOffsets();
+      SmallVector<OpFoldResult> writeStrides = writeOp.getMixedStrides();
+      SmallVector<OpFoldResult> writeSizes = writeOp.getMixedSizes();
+
+      SmallVector<OpFoldResult> insertOffsets = insertOp.getMixedOffsets();
+      SmallVector<OpFoldResult> insertSizes = insertOp.getMixedSizes();
+      SmallVector<OpFoldResult> insertStrides = insertOp.getMixedStrides();
+
+      // Expand source to match sref rank by adding unit dims.
```
MaheshRavishankar, line 247: Is this because the write op doesnt allow reduced rank inserts? I understand the pain it causes, but it does take those annoying unit-dims out of the picture.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,607 @@
...
+static PCF::GenericOp foldForallIntoPCFLoopImpl(RewriterBase &rewriter,
+                                                scf::ForallOp forallOp,
+                                                PCF::LoopOp loopOp) {
+  Location loc = forallOp.getLoc();
+  scf::InParallelOp terminator = forallOp.getTerminator();
+
+  // Replace RegionIterArgs with initial values (except in terminator).
+  for (auto [iterArg, init] :
+       llvm::zip(forallOp.getRegionIterArgs(), forallOp.getOutputs())) {
+    rewriter.replaceUsesWithIf(iterArg, init, [&](OpOperand &use) {
```
MaheshRavishankar, line 309: actually, I am confused why we even have iter_args after bufferization? IIRC the previous "folding" would happen after bufferization to really avoid this complication.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
MaheshRavishankar: Right, I think I had in mind that we should be doing all of this post bufferization, so wires got crossed. Ignore this, let me start a different comment on my current review stack.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+  Value totalWorkers = genericOp.getCountArgs()[0];
+  // Each worker handles ceil(totalWorkers / loopCount) iterations.
+  AffineExpr ceilDiv = s0.ceilDiv(s1);
+  Value outerStep = getValueOrCreateConstantIndexOp(
+      rewriter, loc,
+      affine::makeComposedFoldedAffineApply(rewriter, loc, ceilDiv,
+                                            {totalWorkers, loopCount}));
+
+  auto outerForall = scf::ForallOp::create(
```
MaheshRavishankar, line 394: So is this is the general way you deal with pcf.generic. You have a loop inside it to represent over-subscription?

MaheshRavishankar (approved):
Ok, I think this looks good to me. Left a few comments, but not blockers.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: We could have other flavors of writes (e.g. scatters) instead which aren't supported by this pattern.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: Yeah, pretty much. The only reason this is an issue is because tensor ops allow it and this mixes the two semantics. I'm thinking that unit dims shouldn't be such a problem as long as we're consistent everywhere.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: Yes, that's the basic idea. And I use scf.forall since the loop iterations are independent of one another so I might as well use it over scf.for.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+  // Step 2: Validate pcf.loop structure.
+  if (failed(matchFoldPCFLoop(forallOp, loopOp))) {
+    return failure();
+  }
+
+  // Step 3: Validate write_slice ops.
+  if (failed(matchFoldWriteSlices(loopOp))) {
+    return failure();
+  }
```
hanhanW, line 553: It's better if we use `rewriter.notifyMatchFailure()` and inline the comment into the failure messages.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
```diff
@@ -0,0 +1,561 @@
...
+    SmallVector<scf::ForallOp> forallOps;
+    getOperation()->walk([&](scf::ForallOp forallOp) {
+      // Only match foralls with local_mapping attribute.
+      if (!hasEmptyOrLocalMapping(forallOp)) {
+        return;
+      }
+      // Check if there's a pcf.loop with sequential scope inside.
+      scf::InParallelOp terminator = forallOp.getTerminator();
+      Operation *lastOp = terminator->getPrevNode();
+      auto loopOp = dyn_cast_or_null<PCF::LoopOp>(lastOp);
```
hanhanW: Prefer `dyn_cast_if_present`, see https://github.com/iree-org/iree/commit/33db935a33ddab6a8358418c0ae6dc263e56b0bb

hanhanW (approved):
Just two nits, the idea looks good to me!

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/Passes.td
qedawkins: Correct, without a 1D grid it's hard to fold the extra iterations from the split-k loop evenly across the multiple dims of the pcf.loop. This can probably be relaxed in the future, I may have to rework the iteration mapping semantics of pcf.loop, we'll see.

compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms/FoldForallIntoPCFLoop.cpp
qedawkins: Made it actually use a shared utility this time

PR #23453, opened 2026-02-11
[Target][NFC] Mark internal classes final and wrap within anonymous namespaces

hanhanW:
It moves the internal classes into anonymous namespaces and add final specifiers to the classes and methods.

It also trims the unused headers and delete the empty line between mega namespace and anonymous namepsace.

### DIFF
--- a/compiler/plugins/target/CUDA/CUDATarget.cpp
+++ b/compiler/plugins/target/CUDA/CUDATarget.cpp
@@ -51,7 +51,6 @@
 #include "mlir/Target/LLVMIR/Export.h"
 
 namespace mlir::iree_compiler::IREE::HAL {
-
 namespace {
 struct CUDAOptions {
   std::string clTarget = "sm_60";
@@ -114,7 +113,6 @@ struct CUDAOptions {
     return success();
   }
 };
-} // namespace
 
 static constexpr char kPtxasCompilerName[] = "ptxas";
 
@@ -386,7 +384,7 @@ public:
 
   IREE::HAL::DeviceTargetAttr
   getDefaultDeviceTarget(MLIRContext *context,
-                         const TargetRegistry &targetRegistry) const override {
+                         const TargetRegistry &targetRegistry) const final {
     Builder b(context);
     auto deviceConfigAttr = b.getDictionaryAttr({});
     auto executableConfigAttr = b.getDictionaryAttr({});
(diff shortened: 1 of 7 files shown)


hanhanW:
@kuhar this is the follow-up for one of your review comments: https://github.com/iree-org/iree/pull/23290#discussion_r2784134144

I also made few changes for quality, please take a look. Thanks!

compiler/plugins/target/CUDA/CUDATarget.cpp
```diff
@@ -386,7 +384,7 @@ class CUDATargetDevice final : public TargetDevice {
 
   IREE::HAL::DeviceTargetAttr
   getDefaultDeviceTarget(MLIRContext *context,
-                         const TargetRegistry &targetRegistry) const override {
+                         const TargetRegistry &targetRegistry) const final {
```
kuhar, line 387: I don't think we need to change these since the whole class is marked as final.

compiler/plugins/target/CUDA/CUDATarget.cpp
```diff
@@ -718,16 +716,15 @@ class CUDATargetBackend final : public TargetBackend {
   const CUDAOptions &options;
 };
 
-namespace {
-struct CUDASession
+struct CUDASession final
```
kuhar, line 719: I would only add final here and leave methods with override

compiler/plugins/target/CUDA/CUDATarget.cpp
```diff
@@ -718,16 +716,15 @@ class CUDATargetBackend final : public TargetBackend {
...
     : public PluginSession<CUDASession, CUDAOptions,
```
kuhar, line 720: public is redundant for structs

hanhanW: Honestly, I think these should be class but not struct. Maybe we make the switch? Struct is more like a container that do not have methods to me.

compiler/plugins/target/CUDA/CUDATarget.cpp
hanhanW: True. We can use `override`, but it will force you to scroll up to verify if the class has `final` keyword or not. It provides local reasoning, so you don't have to jump to the class declaration to verify it.

I think there is no convention in practice, and I'm open to both.

compiler/plugins/target/CUDA/CUDATarget.cpp
kuhar: almost all patterns are structs for brevity so 🤷

kuhar (approved):

PR #23457, opened 2026-02-11
Clean up RISC-V toolchain files

HanKuanChen:
Remove unnecessary CMake path variables (CMAKE_FIND_ROOT_PATH, CMAKE_PREFIX_PATH, CMAKE_SYSTEM_LIBRARY_PATH) as they are not used in IREE's cross-compilation workflow.

Remove explicit linker flags for standard libraries (stdc++, pthread, libm, ldl) since CMake handles these automatically via the sysroot.

Switch to CMAKE_*_FLAGS_INIT instead of CMAKE_*_FLAGS to follow CMake convention.


mshockwave (commented):
> Remove explicit linker flags for standard libraries (stdc++, pthread, libm, ldl) since CMake handles these automatically via the sysroot

just want to double check: will cmake add those linker flags automatically upon seeing them in sysroot?

HanKuanChen:
I think my previous comment was ambiguous. CMake will add those options automatically if the dependency is described correctly. For example, libm is enabled by default, so we do not need to add it ourselves.
https://github.com/iree-org/iree/blob/main/build_tools/cmake/iree_copts.cmake#L410

mshockwave (approved):
LGTM

hanhanW (approved):
Approving based on @mshockwave's review

PR #23459, opened 2026-02-11
[LinalgExt] Cleanup TD definitions to avoid variadic everywhere

### LINKED ISSUE #17511: LinalgExt ops should not have Variadic ins/outs
LinalgExt ops have a set number of inputs and outputs. Most of these ops have a defined set of inputs/outputs, causing long handwritten verifiers in c++ which should be auto generated from the tablegen definition.

qedawkins:
For some reason the tablegen for this dialect stuck out like a sore thumb where many of the operands to the ops in this dialect were using an array of ExtraClassDeclarations to provide the getters/setters for operands instead of using the declarative formats like every other dialect.

Also drops the verifier checks for operand counts since it's enforced by the tablegen and a few invalid tests because they fail as parser errors now and aren't as important to test.

Fixes https://github.com/iree-org/iree/issues/17511

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
@@ -45,8 +45,7 @@ def OpGroupNonStructuredOps : OpDocGroup {
 let opDocGroup = OpGroupNonStructuredOps in {
 
 def IREELinalgExt_ScatterOp : IREELinalgExt_Op<"scatter",
-    [AttrSizedOperandSegments,
-     DeclareOpInterfaceMethods<ReifyRankedShapedTypeOpInterface, ["reifyResultShapes"]>,
+    [DeclareOpInterfaceMethods<ReifyRankedShapedTypeOpInterface, ["reifyResultShapes"]>,
      DeclareOpInterfaceMethods<LinalgFusionInterface,
       ["getIndexingMapsForResults", "getIndexingMapsForOperands",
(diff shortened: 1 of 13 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -2026,18 +1894,12 @@ def IREELinalgExt_WinogradFilterTransformOp : IREELinalgExt_Op<"winograd.filter_
     `output_tile_size` `(` $output_tile_size `)`
     `kernel_size` `(` $kernel_size `)`
     `kernel_dimensions` `(` $kernel_dimensions `)`
-    `ins` `(` $inputs `:` type($inputs) `)`
-    `outs` `(` $outputs `:` type($outputs) `)`
+    `ins` `(` $input `:` type($input) `)`
+    `outs` `(` $output `:` type($output) `)`
```
Groverkss, line 1898: I honestly don't know if we should keep the ins/outs notation in the assembly format but I also like less churn so won't block on this.

Groverkss (approved):
So beautiful

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
qedawkins: Yeah I didn't want to change the assembly format in this. Agree, it's strange for many of these ops and changing that too makes sense to me.

kuhar:
woo, thanks Quinn!

PR #23466, opened 2026-02-11
[Util] Integrate `DenseIntElementsAttr` switch to unpacked i1 data

### LINKED ISSUE #23463: Update i1 DenseElementsAttr handling after upstream LLVM change
(Issue generated by claude)

## Summary

LLVM commit llvm/llvm-project@c6964b1b4dcda0f19f91cef186314f4cc2e0feb8 ([mlir][IR] `DenseElementsAttr`: Remove `i1` dense packing special case](https://github.com/llvm/llvm-project/pull/180397)) changed how `i1` elements are stored in `DenseElementsAttr`:

- **Before**: i1 elements were densely packed (1 bit per element, 8 values per byte)
- **After**: i1 elements are padded to 1 byte per element (0x00 or 0x01)

IREE's `serializeBitIntegerValuesAsBytes` function in `compiler/src/iree/compiler/Dialect/Util/IR/UtilAttrs.cpp` assumes the old bit-packed format. This causes incorrect results when processing i1 tensors.

## Failing Tests

- `compiler/src/iree/compiler/ConstEval/test/compile_regressions.mlir`
- `tests/e2e/regression/i1_inlined_constant.mlir` (llvm-cpu and vmvx backends)

## Suggested Fix

Update `serializeBitIntegerValuesAsBytes` to pass through raw data directly since i1 is no longer bit-packed:

```cpp
// Serializes i1 values as 0/1 byte values in the output.
// Note: As of LLVM commit c6964b1b4dcd, DenseElementsAttr no longer bit-packs
// i1 values. Each i1 element is stored as a full byte (0x00 or 0x01).
static LogicalResult serializeBitIntegerValuesAsBytes(DenseIntElementsAttr attr,
                                                      llvm::raw_ostream &os) {
  auto rawData = attr.getRawData();
  os.write(rawData.data(), rawData.size());
  return success();
}
```

The comments at lines 393-398 and 460-462 should also be updated to reflect that i1 is no longer bit-packed.

## Related

- Upstream PR: https://github.com/llvm/llvm-project/pull/180397
- Discourse discussion: https://discourse.llvm.org/t/denseelementsattr-i1-element-type/62525

rkayaith:
As of LLVM commit llvm/llvm-project@c6964b1b4dcd, DenseElementsAttr no longer bit-packs i1
values. Each i1 element is stored as a full byte (0x00 or 0x01), matching
the behavior of DenseResourceElementsAttr.

This drops our revert and removes the now-unnecessary `serializeBitIntegerValuesAsBytes`
function which previously unpacked bit-packed i1 values.]

closes https://github.com/iree-org/iree/issues/23463

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/IR/UtilAttrs.cpp
+++ b/compiler/src/iree/compiler/Dialect/Util/IR/UtilAttrs.cpp
@@ -371,7 +371,6 @@ static LogicalResult serializeGenericElementData(Location loc,
     case 1:
       // NOTE: i1 is treated as i8 in a lot of places in MLIR/IREE and will need
       // a larger cleanup to serialize as a sub-byte value like the others.
-      return serializeRawData(loc, attr, os);
     case 8:
       return serializeRawData(loc, attr, os);
     case 16:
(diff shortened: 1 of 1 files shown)


MaheshRavishankar (approved):
This looks OK to me

MaheshRavishankar:
But maybe wait to land it next week

PR #23469, opened 2026-02-12
Fix wrong output_shape of tensor.expand_shape ops.

hanhanW:


hanhanW:
My guess is that there were wrong fixes when people did the integrate. There are no correctness issues because they are static values and the lowerings do not use them.

IanWood1 (approved):

krzysz00 (approved):

PR #23472, opened 2026-02-12
[LDS] Add pre-check to ensure all copies are DMA-convertible before converting any

lialan:
This patch ensures we either use DMA for all promoted operands or none. This is to ensure we can have async buffer pipelines.

* Add isCopyDMAConvertible() to check DMA viability without modifying IR.
* In GPUConvertToCoalescedDMAPass, collect all linalg.copy ops with use_global_load_dma or derived_thread_config. If ALL are DMA-convertible, upgrade them all to use_global_load_dma. If ANY fails, downgrade them all to derived_thread_config.
* Add new lit tests for paired copy conversion and mixed attribute scenarios.


qedawkins (approved):
Makes sense to me

PR #23475, opened 2026-02-12
[DispatchCreation] Fix producer fusion with nested region uses

### LINKED ISSUE #23339: Failed to LayerNorm with unit batch with enabled split reduction
### What happened?

The MLIR can be found [here](https://gist.github.com/a-sidorova/1f1abd464f98232f2aa0a97211c7058c).

Command line:
```
iree-compile layernorm_single_batch.mlir --iree-hal-target-backends=rocm --iree-hip-target=gfx950 \
             --iree-opt-level=O3 --iree-preprocessing-pass-pipeline="builtin.module(util.func(iree-preprocessing-convert-conv-filter-to-channels-last))" \
             --iree-flow-enable-pad-handling --iree-global-opt-propagate-transposes-through-conv \
             --iree-global-opt-enable-sink-transpose-through-pad --iree-dispatch-creation-enable-fuse-padding-into-linalg-consumer-ops \
             --iree-dispatch-creation-enable-aggressive-reshape-movement --iree-dispatch-creation-enable-split-reduction \
              -o iree-compile-output.vmfb
```

Output:
```
layernorm_single_batch.mlir:22:29: error: operand #0 does not dominate this use
    %y_layernorm_fwd_perm = torch.aten.layer_norm %x_layernorm_fwd_perm, %normalized_shape_layernorm_fwd, %none_scale_layernorm_fwd, %none_bias_layernorm_fwd, %eps_layernorm_fwd, %cudnn_enable_layernorm_fwd : !torch.vtensor<[1,3,128,128],f32>, !torch.list<int>, !torch.none, !torch.none, !torch.float, !torch.bool -> !torch.vtensor<[1,3,128,128],f32>
                            ^
layernorm_single_batch.mlir:22:29: note: see current operation: %32 = "tensor.extract_slice"(%17#1, %arg12, %arg13) <{operandSegmentSizes = array<i32: 1, 2, 0, 0>, static_offsets = array<i64: -9223372036854775808, -9223372036854775808, 0>, static_sizes = array<i64: 1, 8, 128>, static_strides = array<i64: 1, 1, 1>}> : (tensor<3x128x128xf32>, index, index) -> tensor<1x8x128xf32>
layernorm_single_batch.mlir:22:29: note: operand defined here (op in a parent region)
```

But I remove the key `--iree-dispatch-creation-enable-split-reduction` from compilation command line, the MLIR is successfully built and executed.

IREE Compiler

### Version information

Can be successfully reproduced on the IREE commits: https://github.com/iree-org/iree/commit/18fe0926981becffb1332ed727aaf916a9727b13 (30/01/26) and https://github.com/iree-org/iree/commit/4031503a5536c232c5dc0a0412bbf7086db38207 (14/01/26).

### Additional context

_No response_

nirvedhmeshram:
In `fuseRootsWithProducers`, the pass only checked same-block uses via `getFusableUses`, missing uses from above where a producer's result is captured in nested regions. This caused dominance violations when the producer was fused into a later dispatch, breaking intervening ops that used it.

The fix walks up from each use to the parent op in the same block as root and verifies the root dominates it, blocking fusion if not.

Fixes: https://github.com/iree-org/iree/issues/23339

Co-authored-by: GitHub Copilot <noreply@github.com>

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
@@ -761,28 +761,26 @@ static bool isFusableWithProducer(OpOperand &operand,
 static bool hasUsesBetweenProducerAndRoot(Operation *producer, Operation *root,
                                           const FusionGroup &fusionGroup,
                                           const DominanceInfo &dominanceInfo) {
-  for (Value result : producer->getResults()) {
-    for (OpOperand &use : result.getUses()) {
-      Operation *user = use.getOwner();
-      // Walk up to the parent in the same block as the producer/root.
-      while (user->getBlock() != root->getBlock()) {
-        user = user->getParentOp();
-        if (!user) {
-          break;
-        }
-      }
+  for (OpOperand &use : producer->getUses()) {
+    Operation *user = use.getOwner();
+    // Walk up to the parent in the same block as the producer/root.
+    while (user->getBlock() != root->getBlock()) {
+      user = user->getParentOp();
       if (!user) {
-        continue;
+        break;
       }
-      // If the user is in the fusion group, it will be moved too.
-      if (fusionGroup.contains(user) || user == producer) {
-        continue;
-      }
-      // If the user does not come after the root, then moving the producer
-      // into the dispatch at root's position would break dominance.
-      if (!dominanceInfo.properlyDominates(root, user)) {
-        return true;
-      }
+    }
+    if (!user) {
+      continue;
+    }
+    // If the user is in the fusion group, it will be moved too.
+    if (fusionGroup.contains(user)) {
+      continue;
+    }
+    // If the user does not come after the root, then moving the producer
+    // into the dispatch at root's position would break dominance.
+    if (!dominanceInfo.properlyDominates(root, user)) {
+      return true;
     }
   }
   return false;
--- a/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
@@ -754,6 +754,40 @@ static bool isFusableWithProducer(OpOperand &operand,
   return true;
 }
 
+/// Check if moving the producer into the dispatch at the root's position would
+/// break any existing uses. Returns true if there are uses between the producer
+/// and the root that are not in the fusion group, which would cause a dominance
+/// violation.
+static bool hasUsesBetweenProducerAndRoot(Operation *producer, Operation *root,
+                                          const FusionGroup &fusionGroup,
+                                          const DominanceInfo &dominanceInfo) {
+  for (Value result : producer->getResults()) {
+    for (OpOperand &use : result.getUses()) {
+      Operation *user = use.getOwner();
+      // Walk up to the parent in the same block as the producer/root.
+      while (user->getBlock() != root->getBlock()) {
+        user = user->getParentOp();
+        if (!user) {
+          break;
+        }
+      }
+      if (!user) {
+        continue;
+      }
+      // If the user is in the fusion group, it will be moved too.
+      if (fusionGroup.contains(user) || user == producer) {
+        continue;
+      }
+      // If the user does not come after the root, then moving the producer
+      // into the dispatch at root's position would break dominance.
+      if (!dominanceInfo.properlyDominates(root, user)) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
 /// Starting from the `root` op, traverse the operand use-def chain
 /// in reverse to fuse with producers.
 static void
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -754,6 +754,40 @@ static bool isFusableWithProducer(OpOperand &operand,
...
 
+/// Check if moving the producer into the dispatch at the root's position would
+/// break any existing uses. Returns true if there are uses between the producer
+/// and the root that are not in the fusion group, which would cause a dominance
+/// violation.
+static bool hasUsesBetweenProducerAndRoot(Operation *producer, Operation *root,
+                                          const FusionGroup &fusionGroup,
+                                          const DominanceInfo &dominanceInfo) {
+  for (Value result : producer->getResults()) {
+    for (OpOperand &use : result.getUses()) {
```
IanWood1: Looks like these two loops can be simplified to just `producer->getUsers()`

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -754,6 +754,40 @@ static bool isFusableWithProducer(OpOperand &operand,
   return true;
 }
 
+/// Check if moving the producer into the dispatch at the root's position would
+/// break any existing uses. Returns true if there are uses between the producer
+/// and the root that are not in the fusion group, which would cause a dominance
+/// violation.
+static bool hasUsesBetweenProducerAndRoot(Operation *producer, Operation *root,
```
IanWood1, line 761: Side note: ideally, we would have something like `hasTransitiveDependencyOnFusionGroup` but for the producer case. Then `movePrecedingOpsIntoDispatchRegion` would need to move uses of producer that dominate the root.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
IanWood1, line 822: We might be able to remove this now.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -754,6 +754,40 @@ static bool isFusableWithProducer(OpOperand &operand,
...
+        }
+      }
+      if (!user) {
+        continue;
+      }
+      // If the user is in the fusion group, it will be moved too.
+      if (fusionGroup.contains(user) || user == producer) {
+        continue;
+      }
+      // If the user does not come after the root, then moving the producer
```
IanWood1: I don't see when this would be true `user == producer`.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
nirvedhmeshram: maybe better to do that in a separate PR? I wouldnt want this one getting reverted in case there is some use case where this is not yet safe.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
nirvedhmeshram: Makes sense, probably want to file an issue for this and use this PR as a workaround for the compilation bug in the time being?

IanWood1: Yeah that makes sense, I'm using https://github.com/iree-org/iree/issues/22528 to track this (specifically https://github.com/iree-org/iree/issues/22528#issuecomment-3745446056).

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
IanWood1: Good point, there's no reason to risk a revert.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -754,6 +754,38 @@ static bool isFusableWithProducer(OpOperand &operand,
...
+                                          const DominanceInfo &dominanceInfo) {
+  for (OpOperand &use : producer->getUses()) {
+    Operation *user = use.getOwner();
+    // Walk up to the parent in the same block as the producer/root.
+    while (user->getBlock() != root->getBlock()) {
+      user = user->getParentOp();
+      if (!user) {
+        break;
+      }
+    }
```
IanWood1: optional nit:
```suggestion
    while (user && user->getBlock() != root->getBlock()) {
      user = user->getParentOp();
    }
```

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -754,6 +754,38 @@ static bool isFusableWithProducer(OpOperand &operand,
...
+    if (!user) {
+      continue;
+    }
+    // If the user is in the fusion group, it will be moved too.
+    if (fusionGroup.contains(user)) {
+      continue;
+    }
```
IanWood1: Optional:
```suggestion
    // If the user is in the fusion group, it will be moved too.
    if (!user || fusionGroup.contains(user)) {
      continue;
    }
```

IanWood1 (approved):
LGTM!

Issue #23436, opened 2026-02-09
Handle differences in how subgroup IDs should work depending on what the target has


krzysz00:
Some targets (such as AMD's `gfx1250`) have a "efficient" subgroup ID operation - in particular, one that puts that ID in a uniform register and doesn't require us to compute it. Other targets, such as older AMD GPUs, do not have this - they require subgroup IDs to be computed by lineaarizing together thread/workitem IDs and dividing by the subgroup size.

Currently, IREE always runs patterns that decompose the `gpu.subgroup_id` operation into such a linearized computation. Unless/until there's pattern matches in relevant compilers (LLVM, mainly) that will take that decomposed form and convert it into accesses to the subgroup ID register, the usage of such patterns should be controlled by target information. I suggest that we extend WGPInfo with `hardwareSubgroupId = false|true` (defaulting to false).

If `hardwareSubgroupId` is false, we'd emit the decompositions as normal, and could go ahead and add a `readfirstlane`-like operation (something that copies said subgroup ID into uniform registers explicitly, so as to make computations that are functions of the subgroup ID be placed into uniform registers more aggressively).

If `hardwareSubgroupId` is true, we should not only preserve `gpu.subgroup_id` operations but we should actively introduce them - supposing that the subgroup size is 32 and there are 128 threads, then, with a hardware subgroup ID, we should rewrite

```
%x = gpu.thread_id x upper_bound 128
%ids:3 = affine.delinearize_index %x into (2, 2, 32) : index, index, index
```
into
```
%sg = gpu.subgroup_id upper_bound 4
%ids#0, %ids#1 = affine.delinearize_index %sg into (2, 2) : index, index
%ids#2 = gpu.lane_id upper_bound 32
```

jiwonj05:
Hi @krzysz00 , I'd like to work on this! 

My plan is to:
1. Extend WGPInfo with the hardwareSubgroupId attribute.
2. Implement the logic to preserve gpu.subgroup_id for hardware that supports it (like gfx1250).
3. Add the patterns to delinearize thread IDs into gpu.subgroup_id + gpu.lane_id when the flag is true.

Could you please assign this to me?

krzysz00:
I think now that I've cleaned up how subgroup IDs work and removed the hardcoded call to creating the thread_id ... expression over in #23863, this is closed as too complicated and unnecessary

PR #23479, opened 2026-02-16
[Codegen][GPU] Add GPUConvertAllIDsToThreadIDs pass

qedawkins:
Add a pass that canonicalizes GPU ID ops using known workgroup and
subgroup sizes from translation info if present. This allows us to
normalize earlier lowerings to always emit dynamic values that we clean
up later on.

Additionally includes rewrites for gpu.subgroup_id and gpu.lane_id to
thread ID linearization. The reasoning behind this is two-fold. First,
on some targets these operations lower to unfavorable intrinsics and it
ends up being faster to just lower based on the thread id. Second, the
backend compiler has a hard time optimizing expressions involving both
the lane and subgroup ids because the relationship between the two is
not clear, whereas if they are expressed as delinearizations of the
thread id, it's easier for the backend compiler to optimize around them.

Concretely, ds_read2_b64 instructions require a shared base VGPR address
between the two loads (with a constant offset difference between the
two) to target. Without this pass, we end up with mbcnt intrinsics which
block the ability for LLVM to derive a single shared base. It is
possible there are improvements to the LLVM backend that can allow us to
drop the linearization patterns in this pass, but for now this maintains
the status quo.


compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_canonicalize_ids.mlir
```diff
@@ -0,0 +1,143 @@
+// RUN: iree-opt %s --split-input-file --mlir-print-local-scope \
+// RUN:   --pass-pipeline="builtin.module(func.func(iree-codegen-gpu-canonicalize-ids, canonicalize, cse))" | \
+// RUN:   FileCheck %s
+
+// Test: gpu.subgroup_size folds to constant.
+func.func @fold_subgroup_size() -> index
+    attributes {
+      translation_info = #iree_codegen.translation_info<pipeline = None workgroup_size = [128, 2, 1] subgroup_size = 64>
+    } {
+  %0 = gpu.subgroup_size : index
```
krzysz00: This is already handled by `Codegen/Common/PropagateDispatchSizeBounds.cpp` . (and maybe it's worth combining your pass with that one / putting yours right after that one, etc. )

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_canonicalize_ids.mlir
```diff
@@ -0,0 +1,143 @@
...
+//       CHECK:   return %[[C128]], %[[C2]], %[[C1]]
+
+// -----
+
+// Test: gpu.thread_id z folds to 0 when z-dim is 1.
+func.func @fold_thread_id_to_zero() -> (index, index, index)
+    attributes {
+      translation_info = #iree_codegen.translation_info<pipeline = None workgroup_size = [128, 2, 1] subgroup_size = 64>
+    } {
+  %0 = gpu.thread_id x
```
krzysz00: ... this shouldn't be needed, given dispatch size bound propagation followed by integer range analysis

compiler/src/iree/compiler/Codegen/Common/GPU/GPUCanonicalizeIDs.cpp
```diff
@@ -0,0 +1,255 @@
...
+/// them to constants.
+///   linear = tid.x + block_dim.x * (tid.y + block_dim.y * tid.z)
+static Value createLinearThreadId(OpBuilder &builder, Location loc) {
+  Value tidX = gpu::ThreadIdOp::create(builder, loc, gpu::Dimension::x);
+  Value tidY = gpu::ThreadIdOp::create(builder, loc, gpu::Dimension::y);
+  Value tidZ = gpu::ThreadIdOp::create(builder, loc, gpu::Dimension::z);
+  Value dimX = gpu::BlockDimOp::create(builder, loc, gpu::Dimension::x);
+  Value dimY = gpu::BlockDimOp::create(builder, loc, gpu::Dimension::y);
+  // linear = tid.x + dimX * (tid.y + dimY * tid.z)
+  Value inner = arith::MulIOp::create(builder, loc, tidZ, dimY);
```
krzysz00: This is `affine.linearize disjoint` to pick up on its canonicalizaitons etc.

krzysz00 (changes requested):
I think  all this code is redundant with existing passes and doesn't really get at the complexities I identified around subgroup ID handling

qedawkins:
Ok so looking at this https://github.com/iree-org/iree/issues/23436

Where does the linearization happen because I don't see it happening anywhere for me.

krzysz00:
@qedawkins There are upstream patterns for this - they get run *quite* late, if I remember right. Might even be in the gpu-to-rocdl bit.

krzysz00:
(and I misread the PR - I thought it was the external PR for that issue)

krzysz00:
(that there ticket is for "if we do have a subgroup_id operation that's better than the obvious formula, we should use that - and not only should we use that, we should be using it preferentially in any cases where we might be delinearizing a thread ID")

qedawkins:
> @qedawkins There are upstream patterns for this - they get run _quite_ late, if I remember right. Might even be in the gpu-to-rocdl bit.

Rewrote it so that the only pattern is the linearizing one. PTAL. I'm seeing artifacts in the final asm without this change (mbcnt intrinsics) so if there is something supposed to kick in later, it doesn't appear to be.

krzysz00:
@qedawkins Can you take a diff of the LLVM IR we output with / without this change?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertAllIDsToThreadIDs.cpp
```diff
@@ -0,0 +1,94 @@
...
+    Value sgSize =
+        gpu::SubgroupSizeOp::create(rewriter, loc, rewriter.getIndexType(),
+                                    /*upper_bound=*/nullptr);
+    rewriter.replaceOpWithNewOp<arith::DivUIOp>(op, linear, sgSize);
+    return success();
+  }
+};
+
+//===----------------------------------------------------------------------===//
+// Rewrite gpu.lane_id using thread_id linearization.
```
krzysz00, line 57: ... Ok, I can see why this causes changes

lane_id gets lowered to mbcnt intrinsics because those are the canonical way to get a lane ID on AMD chips, and I think there's an argument for keeping that behavior

(the issue with thread ID mod subgroup size is that you've only got certain conditions in which it's the right move)

qedawkins: What are those conditions? Do they always hold for us, because if not we should change everywhere we currently linearize to use lane_id.

krzysz00: ... Right, no, if you use the full linearization it's always fine

The thing that isn't guaranteed is `lane_id <=> (thread_id x) % subgroup_size` , which requires the X dimension to be well-behaved (either it's something like N * subgroup_size or all the other dimensions are 1)

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertAllIDsToThreadIDs.cpp
```diff
@@ -0,0 +1,94 @@
...
+//===----------------------------------------------------------------------===//
+
+struct RewriteLaneId final : OpRewritePattern<gpu::LaneIdOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(gpu::LaneIdOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value linear = createLinearThreadId(rewriter, loc);
+    Value sgSize =
```
krzysz00, line 67: Note to someone: if there isn't already a `rem(linearize))` canonicalization pattern, add one

qedawkins:
> @qedawkins Can you take a diff of the LLVM IR we output with / without this change?

Here are all the dumps: https://gist.github.com/qedawkins/94e0341bb3d903c9129b62bfc380ac9b

pcf_no_canonicalize is the problematic version that generates fewer `ds_read2st64` instructions. Baseline is what we generate on main today (which linearizes all the way when we form loops) and `pcf_ukernel` is what gets generated with this pass.
```
  ┌──────────────────┬───────────────────────┬───────────────────────┬───────────────────────┐                                                                                                                                                                                                                                                                                             
  │      Metric      │ Baseline (scf.forall) │ PCF + CanonicalizeIDs │ PCF - CanonicalizeIDs │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ Total lines      │ 923                   │ 995                   │ 1117 (+22%)           │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ v_mfma           │ 256                   │ 256                   │ 256                   │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ ds_read2st64_b64 │ 32                    │ 32                    │ 14 (-56%)             │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ ds_read_b64      │ 32                    │ 32                    │ 68 (+112%)            │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ ds_write_b64     │ 32                    │ 32                    │ 32                    │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ buffer_load      │ 16                    │ 16                    │ 16                    │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ buffer_store     │ 32                    │ 64                    │ 64                    │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ VGPRs            │ 252                   │ 252                   │ 256 (+4)              │
  ├──────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
  │ SGPRs            │ 26                    │ 30                    │ 26                    │
  └──────────────────┴───────────────────────┴───────────────────────┴───────────────────────┘
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertAllIDsToThreadIDs.cpp
qedawkins: Oh yeah no, we always linearize all of them for that reason for sure

krzysz00 (commented):
Those differences are interesting - I think there's an investigation due, especially if stuff isn't - or is - being constant-folded compared to the old way of looking at things

I'm also going to flag that, on gfx12, the rewrite here is actively harmful, since there's a dedicated "subgroup ID" SGPR on that chip onward.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertAllIDsToThreadIDs.cpp
krzysz00: ... though all the existing canonicalizations will behave if this is a two-output delinearize instead, which ... yeah it looks silly but

qedawkins:
@krzysz00 I did a search of the codebase and it turns out we don't actually generate subgroup_id ever. The only places I found it were upstream:

  mlir/lib/Dialect/XeGPU/Transforms/XeGPUWgToSgDistribute.cpp — 5 sites:                                                                                                                                                                                                                                                                                           
  - Line 104                                                                                                                                                                                                                                                                                                                                                       
  - Line 831                                                                                                                                                                                                                                                                                                                                                       
  - Line 1041 
  - Line 1406
  - Line 1592

  mlir/test/lib/Dialect/XeGPU/TestXeGPUTransforms.cpp — 1 site:
  - Line 209

  The other references in llvm-project are all consumers or converters of SubgroupIdOp, not creators:
  - GPUToROCDL/LowerGpuOpsToROCDLOps.cpp:173 — converts it to ROCDL intrinsics (mbcnt etc.)
  - GPUToSPIRV/GPUToSPIRV.cpp:842 — SPIR-V lowering
  - GPUToLLVMSPV/GPUToLLVMSPV.cpp:427 — LLVM-SPV lowering
  - GPU/Transforms/SubgroupIdRewriter.cpp:24 — upstream's own rewrite pattern (similar this pass)
  - GPU/IR/InferIntRangeInterfaceImpls.cpp:228 — range inference
  - GPU/IR/ValueBoundsOpInterfaceImpl.cpp:106 — bounds interface
  
 So it's only XeGPU that generates this, meaning this pass is actually maintaining the status quo for us atm. So If you're ok with it, we can just land this as is and then do https://github.com/iree-org/iree/issues/23436 in a follow up. Then I don't have to steal a "good first issue" either :)

qedawkins:
Also I tried converting lane_id only and it actually made it worse. So at least up through mi350 it looks like converting is best.

krzysz00:
@qedawkins I could've sworn I've seen some subgroup_id and/or subgroup_size in stuff like handwritten kernels or Wave

And re the conversion of lane ID, do you mean that it's better to convert things into lane_id or to rewrite lane_id to the linearized thread ID thing?

Also, can you see what happens (in the new world where subgroup ID gets used more) if you explicitly `gpu.broadcast_something_something firstlane` it (I forget the op name)? Might need me to go add trivial integer and value ranges to that operation.

qedawkins:
> @qedawkins I could've sworn I've seen some subgroup_id and/or subgroup_size in stuff like handwritten kernels or Wave
> 
> And re the conversion of lane ID, do you mean that it's better to convert things into lane_id or to rewrite lane_id to the linearized thread ID thing?
> 
> Also, can you see what happens (in the new world where subgroup ID gets used more) if you explicitly `gpu.broadcast_something_something firstlane` it (I forget the op name)? Might need me to go add trivial integer and value ranges to that operation.

Ah right I forgot about Wave. In general hidden users of a lowering pipeline which isn't exactly a public API is hard for us to manage, but I won't go break them then. Sorry I was trying to cheat a bit :P

I think I'll pick up https://github.com/iree-org/iree/issues/23436 in a few days/next week and roll it into this if the new contributor hasn't yet.

re: firstlane broadcasting, I'd rather not have to investigate/try everything all at once, I'm trying to keep changes self contained (i.e. this one from my perspective was maintaining the status quo but was deferring a lowering).

krzysz00:
Got it 

{also, some of that readfirstlane placement has been turned into a compiler ticket so I wouldn't focus on that too much - there're tweaks to the "where should the readfirstlane actually be" code that probably need to happen)

qedawkins:
Closing this since I've decided to take a different route.

PR #23480, opened 2026-02-16
Wire Gemmini linalg lowering in IREE and add matmul IR dump

101sparsh:
Summary

First step towards wiring Buddy’s Gemmini dialect into IREE and confirming that linalg.matmul can lower to a Gemmini op. This PR is mainly to get the wiring + IR dump in place so we have a concrete starting point for the full pipeline + backend work.

Changes
	•	Vendor Buddy Gemmini bits under compiler/src/iree/compiler/ThirdParty/buddy_gemmini/:
	•	Gemmini dialect (Gemmini.td, dialect/ops headers, registration)
	•	Lowering passes:
	•	LowerLinalgToGemmini (convert-linalg-to-gemmini)
	•	LowerGemmini skeleton (stub for later LLVM/legalization work)
	•	Simple test drivers (test-gemmini-opt.cpp, test-gemmini-simple.cpp)
	•	CMake wiring for the above.
	•	Register Gemmini and its passes in the IREE compiler so they are available through iree-opt.
	•	Add a tiny linalg matmul example under compiler/src/iree/compiler/Codegen/Common/test/gemmini_ir_dump.mlir.
	•	Add BUDDY_GEMMINI_INTEGRATION.md with notes on how this integration is structured and how to reproduce the IR dump.

How I tested

From the IREE build tree:

`cat <<'EOF' >/tmp/test_matmul.mlir
module {
  func.func @matmul(%a: tensor<4x4xf32>, %b: tensor<4x4xf32>) -> tensor<4x4xf32> {
    %c0 = arith.constant 0.0 : f32
    %init = tensor.empty() : tensor<4x4xf32>
    %fill = linalg.fill ins(%c0 : f32) outs(%init : tensor<4x4xf32>) -> tensor<4x4xf32>
    %0 = linalg.matmul ins(%a, %b : tensor<4x4xf32>, tensor<4x4xf32>) outs(%fill : tensor<4x4xf32>) -> tensor<4x4xf32>
    return %0 : tensor<4x4xf32>
  }
}
EOF`

`./build/tools/iree-opt /tmp/test_matmul.mlir \`
 ` --pass-pipeline='builtin.module(one-shot-bufferize,convert-linalg-to-gemmini)' `
 ` -o - | tee /tmp/gemmini_lowered.mlir | grep -n 'gemmini\.'`

This produces:
`11:    gemmini.tile_matmul %1 %0 %alloc %alloc_0 : memref<4x4xf32, ...> memref<4x4xf32, ...> memref<4x4xf32> memref<4x4xi32>`

and the full IR in /tmp/gemmini_lowered.mlir shows bufferization.to_buffer + allocs + a gemmini.tile_matmul op, confirming that convert-linalg-to-gemmini is wired and firing.

Next steps / scope
	•	This PR is only about wiring the dialect and proving we can lower a simple matmul to gemmini.tile_matmul via iree-opt.
	•	Not yet hooking Gemmini passes into the full IREE compile pipeline or any backend.
	•	Follow-up work:
	•	Decide the right place in the full pipeline (most likely early in the global opt stage) to insert convert-linalg-to-gemmini (and lower-gemmini once fully implemented).
	•	Make sure the full pipeline remains valid when these passes are enabled.
	•	Link Gemmini into the backend and start experimenting with real kernels / configs.


qedawkins (changes requested):
I don't think we want this in tree, or at least you're going to have to share more information about what you're trying to do. There is a plugin system where you can author extensions like downstream which might be a better fit for what you're trying to do.

101sparsh:
Thanks for the feedback, that makes sense.
This PR was mainly a prototype to vendor Buddy’s Gemmini dialect into IREE and check that linalg.matmul can be bufferized and lowered to gemmini.tile_matmul via iree-opt, so we’d have a concrete reference for later work.
I agree this is better as an out-of-tree extension using the compiler plugin system, so I’ll close this and move the integration into a downstream plugin instead.

PR #23482, opened 2026-02-16
[VectorDistribute] Remove special convolution handling

Groverkss:
Add a pass to tile convolutions to matmuls so we never encounter a conv op in vector distribute during layout configuration. This is effectively an NFC, as this was what the unit dim folding was accomplishing.


krzysz00 (commented):
Have we checked perf / impact on final generated code?

Groverkss:
This is a NFC. There should be no perf changes.

kuhar (approved):
LGTM % typo in the PR title

PR #23483, opened 2026-02-16
[VectorDistribute] Add distribution pattern for iree_codegen.inner_tiled

Groverkss:
Add support for distributing iree_codegen.inner_tiled in the GPU vector distribution pipeline, enabling a future switch from vector.contract to inner_tiled for MMA distribution.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -7,6 +7,8 @@
 #include <cstdint>
 #include "iree/compiler/Codegen/Common/GPU/GPUPatterns.h"
 #include "iree/compiler/Codegen/Common/GPU/GPUVectorDistribution.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.h"
 #include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -2236,6 +2238,100 @@ struct DistributeShapeCast final : OpDistributionPattern<vector::ShapeCastOp> {
...
+    if (tiledOp.hasTensorSemantics()) {
+      return rewriter.notifyMatchFailure(tiledOp, "requires vector semantics");
+    }
+
+    Location loc = tiledOp.getLoc();
+    MLIRContext *ctx = tiledOp.getContext();
+
+    // Get layouts, distributed forms, and unpacked forms for all inputs.
+    SmallVector<Value> unpackedInputs;
+    for (auto input : tiledOp.getInputs()) {
```
qedawkins: nit: spell out types, here and below

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -2236,6 +2238,100 @@ struct DistributeShapeCast final : OpDistributionPattern<vector::ShapeCastOp> {
   }
 };
 
+struct DistributeInnerTiled final
```
qedawkins, line 2246: Looking at the impl here, it's just assuming that the layouts were set properly (correct me if not). Can we add a short explanation of that above the pattern here.

qedawkins (approved):

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: added comment.

PR #23484, opened 2026-02-16
[Codegen][GPU] Add FuseSubgroupConsumers pass

qedawkins:
Add a pass that fuses tilable consumers and extract_slice ops into subgroup-scoped pcf.generic/loop ops.

Right now this pass isn't added to any pipeline to separate pipeline changes from pass introductions. Ultimately this pass will be added right around where FuseAndHoistParallelLoops is in the LLVMGPUTileAndFuse pass pipeline.


krzysz00 (approved):
Looks reasonable to me

PR #23486, opened 2026-02-16
[Codegen][GPU] Add BankConflictPaddingHintOp and common hint op interface

Max191:
Adds a new BankConflictPaddingHintOp, which behaves similarly to SwizzleHintOp, but represents padding of the allocation instead of swizzling of the load which consume it. A new interface, called AllocationHintOpInterface now contains the relevant implementation for preserving the hint ops through the codegen pipeline, until they are ready to be used somewhere after bufferization.

The new BankConflictPaddingHintOp will be used for 2 purposes:
1. To improve the GPUReduceBankConflicts pass implementation to work per-operand. This is important for transpose_load targeting on MI350, which requires mixed padding amounts per operand.
2. To enable reasoning about shared memory constraints due to bank conflict padding while selecting the lowering config. This is an improvement to the existing path, which will just bail on bank conflict avoidance if the padding ends up being too large.


compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
```diff
@@ -778,4 +778,26 @@ def IREECodegen_TargetInfoAttrInterface :
...
+    generic handling of hint ops during transformations like fusion.
+  }];
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/[{
+        Returns the operand that holds the source allocation being hinted.
+      }],
+      /*retTy=*/"::mlir::OpOperand &",
+      /*methodName=*/"getHintedOperand",
```
krzysz00: I'm tempted to suggest that this should be read-only and that you'd want `getHintedOperandMutable()` for the mutable version, but I don't know if that's a strong enough convention in all of IREE to justify the change.

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -265,14 +270,19 @@ struct SwizzleHintOpInterface final
   LogicalResult bufferize(Operation *op, RewriterBase &rewriter,
                           const BufferizationOptions &options,
                           bufferization::BufferizationState &state) const {
-    auto hintOp = cast<IREE::Codegen::SwizzleHintOp>(op);
+    auto hintOp = cast<IREE::Codegen::AllocationHintOpInterface>(op);
     FailureOr<Value> maybeBuffer =
-        getBuffer(rewriter, hintOp.getOperand(), options, state);
+        getBuffer(rewriter, hintOp.getHintedOperand().get(), options, state);
```
krzysz00: Ok, yeah, this sort of things looks weird and justifies the split above, I think.

krzysz00 (approved):
LGTM after one nit

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
Max191: I added the mutable and read only versions, although I kept just the mutable version as an interface function. The non-mutable one is derived from the mutable one so there is less boilerplate when implementing the interface.

Muzammiluddin-Syed-ECE (approved):

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
```diff
@@ -778,4 +778,34 @@ def IREECodegen_TargetInfoAttrInterface :
   ];
 }
 
+def IREECodegen_AllocationHintOpInterface :
+  OpInterface<"AllocationHintOpInterface"> {
+  let cppNamespace = "::mlir::iree_compiler::IREE::Codegen";
+  let description = [{
+    Interface for ops that annotate an allocation with a hint (e.g. swizzle
```
kuhar:
```suggestion
    Interface for ops that annotate an allocation with a hint (e.g., swizzle
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -343,4 +344,45 @@ def IREEGPU_CoalescedGatherDMAOp : Op<IREEGPU_Dialect, "coalesced_gather_dma", [
...
+
+def IREEGPU_BankConflictPaddingHintOp : Op<IREEGPU_Dialect,
+    "bank_conflict_padding_hint", [
+    SameOperandsAndResultType, Pure,
+    DeclareOpInterfaceMethods<IREECodegen_AllocationHintOpInterface>]> {
+  let summary = [{Hint to apply bank conflict padding to a shared memory allocation.}];
+  let description = [{
+    Wraps a tensor or memref to indicate the desired bank conflict padding.
+    Consumed by GPUReduceBankConflicts which reads the padding amount and
+    applies it to the underlying allocation.
```
kuhar: Can you explain how this affects indexing? Is the padding applied to the innermost dimension only? What if the innermost dimension is less than or greater than dword * num_banks?

kuhar (commented):
Looks good modulo documentation

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -343,4 +344,45 @@ def IREEGPU_CoalescedGatherDMAOp : Op<IREEGPU_Dialect, "coalesced_gather_dma", [
...
+  }];
+
+  let arguments = (ins AnyRankedTensorOrMemRef:$operand,
+                       I64Attr:$padding_bits);
+  let results = (outs AnyRankedTensorOrMemRef:$result);
+
+  let assemblyFormat = [{
+    $operand `[` `padding_bits` `=` $padding_bits `]` attr-dict `:` type($result)
+  }];
+}
```
kuhar, line 438: Can we do some verification to make sure the values are reasonable? For example, zero padding (allowed?), negative padding, very large padding...

krzysz00:
Late ping - is this work still going?

PR #23487, opened 2026-02-16
[GPU] Support BankConflictPaddingHintOp in GPUReduceBankConflicts

Max191:
Adds support for handling the BankConflictPaddingHintOp in GPUReduceBankConflicts, while maintaining the original pass option as a fallback for pipelines that are not yet using the hint op. Setting the pass option to zero will effectively create the desired effect of using exclusively hint ops for padding. This will be done later in the LLVMGPUTileAndFuse pipeline to start.

The shared memory limit constraints are still handled as part of this pass, and it is left as TODO to move it to lowering config selection, but this requires transitioning all pipelines to use the hint op.


Max191:
Based on https://github.com/iree-org/iree/pull/23486

compiler/src/iree/compiler/Codegen/Common/GPU/test/reduce_bank_conflicts.mlir
```diff
@@ -182,3 +179,99 @@ func.func @pad_if_below_limit() {
...
+      : memref<4x32x64xf32, #gpu.address_space<workgroup>>
+  return
+}
+
+// -----
+
+// Mixed: hinted alloc gets hint padding, unhinted alloc gets fallback.
+// CHECK-LABEL: func.func @mixed_hinted_and_unhinted
+// CHECK-DAG:     memref.alloc() : memref<2x16x65xf32, #gpu.address_space<workgroup>>
+// CHECK-DAG:     memref.alloc() : memref<2x16x66xf32, #gpu.address_space<workgroup>>
```
krzysz00, line 218: Is the future plan that the default behavior is "no padding"?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,103 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+    Value source = hintOp.getOperand();
+    memref::AllocOp allocOp = findUnderlyingAlloc(source);
+    if (!allocOp || !hasSharedMemoryAddressSpace(allocOp.getType())) {
+      return;
+    }
+
+    auto [it, inserted] =
+        allocPaddingMap.try_emplace(allocOp, hintOp.getPaddingBits());
+    if (!inserted) {
+      // Multiple hints for the same alloc — disable padding.
```
krzysz00: I'd go with slightly more complex logic - multiple *conflicting* hints are a disable ... and a warning to the user

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,103 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+    }
 
-      unsigned bitWidth = 64; // IREE's default bitWidth for indexes
-      auto elemType = allocType.getElementType();
-      if (!elemType.isIndex()) {
-        bitWidth = IREE::Util::getTypeBitWidth(elemType);
-      }
-      unsigned elemSize = bitWidth / 8;
-      unsigned extraElements = paddingBits / bitWidth;
+    int outerProduct = 1;
```
krzysz00: int64_t

krzysz00 (approved):
lgtm % nits

compiler/src/iree/compiler/Codegen/Common/GPU/test/reduce_bank_conflicts.mlir
Max191: Yes, that is the plan, but this pass is used across many pipelines (like SPIRV, for example), so we will need to update them all to support the hint ops before removing the old default behavior.

compiler/src/iree/compiler/Codegen/Common/GPU/test/reduce_bank_conflicts.mlir
```diff
@@ -182,3 +179,99 @@ func.func @pad_if_below_limit() {
...
+func.func @hint_zero_padding() {
+  %0 = memref.alloc() : memref<4x32x64xf32, #gpu.address_space<workgroup>>
+  %hinted = iree_gpu.bank_conflict_padding_hint %0 [padding_bits = 0]
+      : memref<4x32x64xf32, #gpu.address_space<workgroup>>
+  return
+}
+
+// -----
+
+// Multiple hints on the same alloc result in no padding.
```
Muzammiluddin-Syed-ECE, line 244: why do we choose to allow multiple hints on the same alloc result at the memref level and emit only a warning?

kuhar: This is not something we can check in the verifier

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,106 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+  return nullptr;
+}
+
+/// Collect padding hints from BankConflictPaddingHintOp wrapping shared memory
+/// allocations. Returns a map from alloc to the desired padding in bits.
+/// If multiple hints map to the same alloc with conflicting padding values,
+/// the padding is set to 0 (no padding applied).
+static DenseMap<memref::AllocOp, unsigned>
+collectPaddingHints(FunctionOpInterface funcOp) {
+  DenseMap<memref::AllocOp, unsigned> allocPaddingMap;
```
kuhar, line 135: nit: maybe call it `allocToPadding` to focus on what maps to what; we know it's map based on the type

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,106 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+
+/// Remove all BankConflictPaddingHintOp from the function, replacing their
+/// uses with their operands.
+static void removePaddingHints(FunctionOpInterface funcOp) {
+  SmallVector<IREE::GPU::BankConflictPaddingHintOp> hintsToRemove;
+  funcOp.walk([&](IREE::GPU::BankConflictPaddingHintOp hintOp) {
+    hintsToRemove.push_back(hintOp);
+  });
+
+  for (auto hintOp : hintsToRemove) {
```
kuhar, line 166: nit: type

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,106 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+    }
 
-      unsigned bitWidth = 64; // IREE's default bitWidth for indexes
-      auto elemType = allocType.getElementType();
-      if (!elemType.isIndex()) {
-        bitWidth = IREE::Util::getTypeBitWidth(elemType);
-      }
-      unsigned elemSize = bitWidth / 8;
-      unsigned extraElements = paddingBits / bitWidth;
+    int64_t outerProduct = 1;
```
kuhar, line 200: use `llvm::product_of(shape.drop_back())`

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,106 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+    for (size_t i = 0; i < shape.size() - 1; ++i) {
+      outerProduct *= shape[i];
+    }
 
-      totalExtra += outerProduct * extraElements * elemSize;
+    unsigned bitWidth = 64; // IREE's default bitWidth for indexes
```
kuhar, line 205: nit: missing `.`

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,106 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+    auto elemType = allocType.getElementType();
```
kuhar, line 206: nit: type

kuhar (commented):
Do we support both padding and swizzle over the same allocations?

compiler/src/iree/compiler/Codegen/Common/GPU/test/reduce_bank_conflicts.mlir
Muzammiluddin-Syed-ECE: Sorry this is a general question that is more related to code below. let me move the comment.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUReduceBankConflicts.cpp
```diff
@@ -121,35 +111,106 @@ static int64_t computeSharedMemoryUsage(mlir::FunctionOpInterface funcOp) {
...
+    unsigned padding = hintOp.getPaddingBits();
+    auto [it, inserted] = allocPaddingMap.try_emplace(allocOp, padding);
+    if (!inserted && it->second != padding) {
+      // Conflicting hints for the same alloc — disable padding.
+      hintOp->emitWarning()
+          << "conflicting bank conflict padding hints for " << *allocOp << ": "
+          << it->second << " bits vs " << padding << " bits";
+      it->second = 0;
+    }
+  });
```
Muzammiluddin-Syed-ECE, line 153: why do we choose to ignore hints all together in the case of multiple hints on the same alloc? Is this a pattern we expect to happen? i guess what im asking is why not throw an error.

qedawkins: This is my concern as well. A swizzle hint has nothing to do with the memory it operates on while this affects operations not involved in it's chain. The way I would expect to expose this is instead as the allocation op itself, or simply with dynamic allocation sizes.

PR #23489, opened 2026-02-17
[DispatchCreation] Normalize reduction dim order in TransposeGenericOps

### LINKED ISSUE #23196: Fix reduction dim order in `PropagateLinalgTransposePass`
```
func.func @main(%arg0: !torch.tensor<[16,48,1,32,288],bf16>, %arg1: !torch.vtensor<[288,2,288,1,1],bf16>, %arg2: !torch.vtensor<[16,2,48,32,288],bf16>) attributes {torch.assume_strict_symbolic_shapes} {
  %none = torch.constant.none
  %false = torch.constant.bool false
  %0 = torch.prim.ListConstruct  : () -> !torch.list<int>
  %int1 = torch.constant.int 1
  %int2 = torch.constant.int 2
  %int1_0 = torch.constant.int 1
  %int1_1 = torch.constant.int 1
  %1 = torch.prim.ListConstruct %int2, %int1_0, %int1_1 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %int0 = torch.constant.int 0
  %int0_2 = torch.constant.int 0
  %int0_3 = torch.constant.int 0
  %2 = torch.prim.ListConstruct %int0, %int0_2, %int0_3 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %int1_4 = torch.constant.int 1
  %int1_5 = torch.constant.int 1
  %int1_6 = torch.constant.int 1
  %3 = torch.prim.ListConstruct %int1_4, %int1_5, %int1_6 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %int0_7 = torch.constant.int 0
  %int4 = torch.constant.int 4
  %int1_8 = torch.constant.int 1
  %int2_9 = torch.constant.int 2
  %int3 = torch.constant.int 3
  %4 = torch.prim.ListConstruct %int0_7, %int4, %int1_8, %int2_9, %int3 : (!torch.int, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %5 = torch.aten.permute %arg2, %4 : !torch.vtensor<[16,2,48,32,288],bf16>, !torch.list<int> -> !torch.vtensor<[16,288,2,48,32],bf16>
  %int0_10 = torch.constant.int 0
  %int2_11 = torch.constant.int 2
  %int1_12 = torch.constant.int 1
  %int3_13 = torch.constant.int 3
  %int4_14 = torch.constant.int 4
  %6 = torch.prim.ListConstruct %int0_10, %int2_11, %int1_12, %int3_13, %int4_14 : (!torch.int, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %7 = torch.aten.permute %arg1, %6 : !torch.vtensor<[288,2,288,1,1],bf16>, !torch.list<int> -> !torch.vtensor<[288,288,2,1,1],bf16>
  %8 = torch.aten.convolution %5, %7, %none, %1, %2, %3, %false, %0, %int1 : !torch.vtensor<[16,288,2,48,32],bf16>, !torch.vtensor<[288,288,2,1,1],bf16>, !torch.none, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int -> !torch.vtensor<[16,288,1,48,32],bf16>
  %int0_15 = torch.constant.int 0
  %int3_16 = torch.constant.int 3
  %int2_17 = torch.constant.int 2
  %int4_18 = torch.constant.int 4
  %int1_19 = torch.constant.int 1
  %9 = torch.prim.ListConstruct %int0_15, %int3_16, %int2_17, %int4_18, %int1_19 : (!torch.int, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %10 = torch.aten.permute %8, %9 : !torch.vtensor<[16,288,1,48,32],bf16>, !torch.list<int> -> !torch.vtensor<[16,48,1,32,288],bf16>
  torch.overwrite.tensor.contents %10 overwrites %arg0 : !torch.vtensor<[16,48,1,32,288],bf16>, !torch.tensor<[16,48,1,32,288],bf16>
  return
}
```
For this input we end with a swapped reduction dim order that causes codegen perf issue

IanWood1:
Add NormalizeReductionDimsPattern to sort reduction dimensions into ascending order within indexing maps.

Closes https://github.com/iree-org/iree/issues/23196

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/TransposeGenericOps.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/TransposeGenericOps.cpp
@@ -119,12 +119,77 @@ struct TransposeGenericOpPattern final
   }
 };
 
+/// Normalize the order of reduction dimensions in generic ops to make them
+/// appear in ascending order in the indexing maps.
+struct NormalizeReductionDimsPattern final
+    : public OpRewritePattern<linalg::GenericOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    SmallVector<unsigned> reductionDims;
+    for (auto [idx, iterType] :
+         llvm::enumerate(genericOp.getIteratorTypesArray())) {
+      if (linalg::isReductionIterator(iterType)) {
+        reductionDims.push_back(idx);
+      }
+    }
+    if (reductionDims.size() < 2) {
+      return rewriter.notifyMatchFailure(genericOp, "<2 reduction dims");
+    }
+
+    // Pre-compute dim positions within each indexing map.
+    SmallVector<DenseMap<unsigned, unsigned>> posMaps;
+    for (AffineMap map : genericOp.getIndexingMapsArray()) {
+      DenseMap<unsigned, unsigned> posMap;
+      for (auto [pos, expr] : llvm::enumerate(map.getResults())) {
+        if (auto dimExpr = dyn_cast<AffineDimExpr>(expr)) {
+          posMap[dimExpr.getPosition()] = pos;
+        }
+      }
+      posMaps.push_back(std::move(posMap));
+    }
+
+    // Sort reduction dims by their positions in the indexing maps. Only
+    // reorder a pair if ALL maps that contain both dims unanimously agree
+    // on the ordering; conflicting or unknown pairs keep their original
+    // relative order (guaranteed by stable_sort).
+    SmallVector<unsigned> sorted(reductionDims);
+    llvm::stable_sort(sorted, [&](unsigned dimA, unsigned dimB) {
+      bool anyInfo = false;
+      for (const auto &posMap : posMaps) {
+        auto itA = posMap.find(dimA);
+        auto itB = posMap.find(dimB);
+        if (itA == posMap.end() || itB == posMap.end()) {
+          continue;
+        }
+        anyInfo = true;
+        if (itA->second >= itB->second) {
+          return false;
+        }
+      }
+      return anyInfo;
+    });
+
+    if (sorted == reductionDims) {
+      return rewriter.notifyMatchFailure(genericOp, "already normalized");
+    }
+
+    // Build interchange: identity permutation with reduction dims reordered.
+    auto interchange =
+        llvm::to_vector(llvm::seq<unsigned>(0, genericOp.getNumLoops()));
+    for (size_t i = 0; i < reductionDims.size(); ++i) {
+      interchange[reductionDims[i]] = sorted[i];
+    }
+    return interchangeGenericOp(rewriter, genericOp, interchange);
+  }
+};
+
 struct TransposeGenericOpsPass final
     : public impl::TransposeGenericOpsPassBase<TransposeGenericOpsPass> {
   void runOnOperation() override {
     RewritePatternSet patterns(&getContext());
-    patterns.add<MakeReductionInnermostPattern, TransposeGenericOpPattern>(
-        &getContext());
+    patterns.add<MakeReductionInnermostPattern, TransposeGenericOpPattern,
+                 NormalizeReductionDimsPattern>(&getContext());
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/TransposeGenericOps.cpp
```diff
@@ -119,12 +119,77 @@ struct TransposeGenericOpPattern final
...
+    // Pre-compute dim positions within each indexing map.
+    SmallVector<DenseMap<unsigned, unsigned>> posMaps;
+    for (AffineMap map : genericOp.getIndexingMapsArray()) {
+      DenseMap<unsigned, unsigned> posMap;
+      for (auto [pos, expr] : llvm::enumerate(map.getResults())) {
+        if (auto dimExpr = dyn_cast<AffineDimExpr>(expr)) {
+          posMap[dimExpr.getPosition()] = pos;
+        }
+      }
+      posMaps.push_back(std::move(posMap));
```
MaheshRavishankar, line 149: nit: `posMaps.emplace_back(...)`.

MaheshRavishankar (commented):
I wanted to just check that this pattern does not destructively interfere with the other patterns added here. In general these are a bit of "kitchen sink" patterns and historically these are here cause I couldnt figure out a better place to do this. So these were here to maintain status quo while I/We cleaned up the other parts of the dispatch creation pipeline.

Can you also run the torch_tests with these as well as the benchmark CI that @pdhirajkumarprasad  has setup?

IanWood1:
> I wanted to just check that this pattern does not destructively interfere with the other patterns added here.

This shouldn't interfere with the other patterns:
- `TransposeGenericOpPattern` only transforms ops without reduction loops.
- `MakeReductionInnermostPattern` interchanges `reduction <-> parallel` loops. This pattern just interchanges parallel loops.


> In general these are a bit of "kitchen sink" patterns and historically these are here cause I couldnt figure out a better place to do this. So these were here to maintain status quo while I/We cleaned up the other parts of the dispatch creation pipeline.

This is my current understanding of these patterns. After the semi-recent changes to dispatch creation, we should be able to remove `TransposeGenericOpPattern` (but this is very likely to cause codegen issues). `MakeReductionInnermostPattern` shouldn't really be in dispatch creation because this pipeline doesn't care about the ordering of `iterator_types`. This is just cleanup for codegen similar to what `NormalizeReductionDimsPattern` from this PR is.


> Can you also run the torch_tests with these as well as the benchmark CI that @pdhirajkumarprasad has setup?

Yes, I reached out to @pdhirajkumarprasad to point me to the CI.

compiler/src/iree/compiler/DispatchCreation/TransposeGenericOps.cpp
IanWood1: `push_back` and `emplace_back` should be exactly the same here. @hanhanW left a comment a while ago regarding when to use emplace_back and the gist of it that `emplace_back` is only useful when you want to call a constructor that is not the move/copy constructor. This calls the move constructor for `DenseMap<unsigned, unsigned>` so `push_back` should be fine.

hanhanW: Here is the read: https://abseil.io/tips/112

An example is:

> The first line is quite straightforward: it adds the number 1048576 to the end of the vector. The second, however, is not so clear. Without knowing the type of the vector, we don’t know what constructor it’s invoking, so we can’t really say what that line is doing. For example, if vec2 is a std::vector<int>, that line merely adds 1048576 to the end, as with the first line, but if vec2 is a std::vector<std::vector<int>>, that second line constructs a vector of over a million elements, allocating several megabytes of memory in the process.

```
vec1.push_back(1<<20);
vec2.emplace_back(1<<20);
```

The behavior is different depending on the types.

MaheshRavishankar (approved):

PR #23490, opened 2026-02-17
[PCF] Add MemoryEffectsOpInterface to WriteSliceOp

qedawkins:
Without declared memory effects, WriteSliceOp (which has no results) appears side-effect-free to OptimizeVectorTransferPass. This causes eraseDeadAllocAndStores to eliminate writes as dead code.

Declare Read on the source operand and Write on the dest (pcf.sref).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
@@ -919,6 +919,19 @@ LogicalResult WriteSliceOp::fold(FoldAdaptor adaptor,
   return success();
 }
 
+void WriteSliceOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  if (isa<MemRefType>(getSourceType())) {
+    // Source memrefs are read from.
+    effects.emplace_back(MemoryEffects::Read::get(), &getSourceMutable(),
+                         SideEffects::DefaultResource::get());
+  }
+  // The dest operand is written to.
+  effects.emplace_back(MemoryEffects::Write::get(), &getDestMutable(),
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.cpp
```diff
@@ -919,6 +919,19 @@ LogicalResult WriteSliceOp::fold(FoldAdaptor adaptor,
   return success();
 }
 
+void WriteSliceOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  if (isa<MemRefType>(getSourceType())) {
```
krzysz00, line 925: I take it this if statement is what prevents you from tablegening this?

qedawkins: I honestly didn't even realize we can set these in tablegen, but yeah, I don't see a way to set it conditionally like this in tablegen.

krzysz00: Yeah, it's the `Arg<>` syntax ... and this leads me to the question - is this not a read if the source is a non-memref?

qedawkins: Yeah the input can be a vector tensor or memref, and I don't think you can't have read effects on value semantic types, unless memory effects don't mean what I think.

krzysz00: ... Ok, yeah, having looked at TransferRead, it does what you're doing here

krzysz00 (approved):
lgtm

Issue #23415, opened 2026-02-06
[VectorDistribute] Improve handling of hardware-unfriendly shapes


sommerlukas:
Not all shapes are well aligned to the sizes of hardware primitives (e.g., width of loads, matrix intrinsics, ...) or it is unclear due to dynamic shapes. In some cases, the vector distribute pipeline handles that by reducing the number of elements processed by each thread to better accommodate the shape of the input. 

However, in these cases it would be more efficient to continue using the full vector-width per thread and instead mask out some threads. This can happen through padding of the shapes to a more hardware-friendly shape and masked operation on the padded tensor.

There is however room for improvement in the vector distribute pipeline for these cases, concretely:

- Propagate padding from a root operation (e.g., reduction) to its producers and consumers.
- Improve the lowering of "unmasking" as the opposite operation of padding, bringing the tensor back to its original shape. This can be done with `extract_slice`, which however may require additional intermediate allocations during lowering. Instead, the goal is to introduce an `unmask` (bike shedding TBD) operation with clear semantics that lowers more efficiently.

For the `unmask` operation, some prior discussion of the background can be found in [this Discord thread](https://discord.com/channels/689900678990135345/1436113909151957082).

sommerlukas:
After discussion with @Groverkss, I'm currently testing a different approach for the second step.

Instead of explicitly padding the operations early in pipeline, the necessary padding/masking for the `reduction` and `serial` lowering levels is happening during code generation.

Overall, the approach is:
1. Apply tiling for `reduction` and `serial` levels early in the pipeline without prior padding. If the size of the tensor is dynamic or isn't divisible by the tile size, this will yield a dynamic dimension in the tensor.
2. After this step, the tensor will have a dynamic size in that dimension, but the information about the tile size is still present in the lowering config. 
3. After vectorization, operations should work on vectors of the tile size. That means that during vectorization of `linalg.generic`, the necessary masking needs to be introduced. The information on the tile sizes is taken from operations with a lowering config and propagated to operations without a lowering config through the `VectorLayoutAnalysis`, which, despite its name, supports all shaped types. The analysis needs to be extended for propagating through `linalg.generic`.
4. The masked, vectorized code can now be distributed and further lowered. As distribution of masked operations, in particular `vector.contract` hasn't been exercised much so far, some additions will be necessary here.

With this approach, it's not necessary to introduce the `unmask` operation as padding/unpadding is handled during vectorization.

[...]

PR #23493, opened 2026-02-17
[VectorDistribute] Reify shape to enable folding

sommerlukas:
During vectorization of `to_layout`, use shape reification if possible. 

`to_layout` is vectorized to a pair of masked `transfer_read` and `transfer_write`. If padding is required, the vectorization of `linalg.generic` will also use masked `transfer_read/write`. On the operands, the resulting chain is already successfully folded before this change.

By using shape reification, the same mask SSA value will be used for the entire `transfer_write -> transfer_read -> transfer_write` chain on the result of `linalg.generic`, enabling the same folding to take place and avoiding insertion of `memref.copy` during bufferization.

Assisted-by: Claude Code


sommerlukas:
This is not needed anymore after taking another direction for https://github.com/iree-org/iree/issues/23415, closing this PR.

PR #23494, opened 2026-02-17
Initial plumbing for inner_tiled on CPU with data-tiled MMA attribute.

bjacob:
This is just the initial round of plumbing to get to the point where one can at least build and validate a `inner_tiled` op on CPU, meaning with a `kind` parameter that is a CPU-specific data-tiled MMA layout attr with an intrinsic enum that designates a SIMD intrinsic.

Most of this was written by AI.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
@@ -4,14 +4,20 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include "iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.h"
 #include "iree/compiler/Codegen/Dialect/CPU/IR/IREECPUTypes.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
 #include "iree/compiler/Dialect/Encoding/IR/EncodingTypes.h"
 #include "llvm/ADT/STLExtras.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
 #include "mlir/IR/Attributes.h"
+#include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/ValueRange.h"
 
+#include "iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.cpp.inc"
 #define GET_ATTRDEF_CLASSES
 #include "iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp.inc"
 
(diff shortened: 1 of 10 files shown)


compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.td
```diff
@@ -0,0 +1,132 @@
...
+//   - 3 is RISC-V.
+// * B denotes the instruction family / ISA extension set.
+//   - For x86:
+//     - 1 is SSE.
+//     - 2 is AVX/AVX2.
+//     - 3 is AVX-512.
+//   - For Arm:
+//     - 1 is NEON.
+//     - 2 is SVE/SVE2.
+//     - 3 is SME/SME2.
```
egebeysel, line 87: In these cases, the intrinsics would require scalable (dynamic) sizes for the M/N dimensions. Please correct me if I'm wrong, but there is currently no way** of modeling that with the current `inner_tiled` op, right?

** I think the operation semantics currently do not allow that: https://github.com/iree-org/iree/blob/038bebb17c015f1352d414cb471e61d6d48df9ae/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp#L319-L325

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.td
```diff
@@ -0,0 +1,132 @@
...
+//      lower to a cast followed by a matmul on the wider type. This would be
+//      indicated by a _CAST_I16 or _CAST_F32 suffix.
+//
+// Values are 0xABCD where:
+// * A denotes the architecture:
+//   - 0 is reserved for the None-value.
+//   - 1 is x86.
+//   - 2 is Arm.
+//   - 3 is RISC-V.
+// * B denotes the instruction family / ISA extension set.
```
egebeysel, line 79: AFAIK, the RISC-V currently only has the V extension over here that would make sense, but I still think it's sensible to enumerate that right now and give it a default. I'm currently working on scalable data-tiling for RVV, but there is support for the static case via the `+zvl*` flags. So a potential second additional _family_ here might be `RVV (scalable)`. Also IIRC, the matrix extension is baking as well and might come here in the future :)

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -277,6 +283,162 @@ SmallVector<bool> LoweringConfigAttr::getVectorScalableFlags() const {
   return result;
 }
 
+//===----------------------------------------------------------------------===//
+// CPU MMA intrinsic layout (MxNxK shape and element types)
+//===----------------------------------------------------------------------===//
+
+struct CPUOpaqueMmaLayout {
+  int64_t mSize = 0;
```
egebeysel, line 293: same concern with the scalable tiles, maybe we should model these as  in for the CPU: https://github.com/iree-org/iree/blob/038bebb17c015f1352d414cb471e61d6d48df9ae/compiler/src/iree/compiler/Codegen/Utils/Utils.h#L334-L335

egebeysel (commented):
Very interesting stuff so I self-appointed myself to review if you don't mind :) I don't really have much comments regarding the static cases - it LGTM from that front.

Though I'm actively working on scalable vector ISAs (SVE, SME, RVV) and have some concerns about the current semantics of the operation to model them. I left my thoughts as comments, but if I'm not mistaken - the operation currently does not allow them at all.

I'm willing to work on this and bolt in the missing pieces myself - I think it's important that we do this early rather than later after switching to use this op instead of the `mmt4d`. Though I'm not sure how that would play out with the current infrastructure built around the `inner_tiled` op. Let me know what you think!

Also, more of a question rather than a review, but what is the incentive of switching from `mmt4d` to `inner_tiled` for the CPU pipeline? Is it more to unify the data-tiling work for the CPU and GPU pipelines or is there an _effective_ semantic gap between these op for the CPU?

egebeysel:
cc @banach-space

bjacob:
Thanks @egebeysel  for all the comments. We had a conversation offline and then here: https://discord.com/channels/689900678990135345/1456679894812590326/1474440363199434843

The summary is that we agree that further changes are needed anyway, but the present PR lays some groundwork that is useful regardless. I would like to go ahead and get this merged just to take it off of my stack of WIP PRs.

krzysz00 (approved):
Interface implementations and such lgtm

egebeysel (approved):
Thanks for the offline discussion. LGTM  as well :)

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/test/iree_cpu_inner_tiled_ops.mlir
hanhanW, line 1: nit: can we add a blank line after the separator (i.e., `// -----`)?

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.h
```diff
@@ -0,0 +1,16 @@
+// Copyright 2024 The IREE Authors
```
hanhanW: 2026

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.td
```diff
@@ -0,0 +1,132 @@
+// Copyright 2024 The IREE Authors
```
hanhanW: 2026

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.td
```diff
@@ -8,8 +8,92 @@
...
+//===----------------------------------------------------------------------===//
+//
+// CPU analogue of IREEGPU_DataTiledMMAAttr, for use with
+// iree_codegen.inner_tiled. Like the GPU case, this wraps an intrinsic-enum and
+// some intrinsics_{m,n,k} unrolling factor. Unlike the GPU case, there is no
+// thread-distribution, no concept of subgroups and no interleaving of
+// intrinsics' layout.
+//
+// Some GPU-specific methods in IREECodegen_InnerTileDescAttrInterface are left
+// here but are unused.
```
hanhanW, line 61: Do we move this to `descripction`? The benefit is that developers can eaisly find the comment in website doc like: https://iree.dev/reference/mlir-dialects/IREEGPU/#conditionaltransposeattr

CPU dialect is missing, and I'll raise a PR for it.

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -277,6 +283,162 @@ SmallVector<bool> LoweringConfigAttr::getVectorScalableFlags() const {
...
+  int64_t nSize = 0;
+  int64_t kSize = 0;
+  Type aType;
+  Type bType;
+  Type cType;
+};
```
hanhanW, line 299: Clang-tidy nit: this can be moved into an anonymous namespace to enforce internal linkage.

hanhanW (approved):
Just few nits. I don't know what the final picture is, but I'm sure that it is going to be great. I'm looking forward to it, and thanks for starting the work!

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.td
hanhanW: will be done in https://github.com/iree-org/iree/pull/23546

PR #23502, opened 2026-02-18
[Codegen][GPU] Rework GPUVerifyDistribution to use PreOrder walk with skip

qedawkins:
Rework the pass to walk in pre-order and skip into distributed regions (thread/lane-mapped scf.forall, pcf.generic, pcf.loop) rather than checking parent ops for every write. This is simpler, more efficient, and correctly handles PCF ops.

Add test cases for thread-mapped forall writes, undistributed writes, pcf.generic writes, pcf.loop writes, and writes outside pcf.generic regions.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVerifyDistribution.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVerifyDistribution.cpp
@@ -7,6 +7,7 @@
 #include "iree/compiler/Codegen/Common/GPU/Passes.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.h"
+#include "iree/compiler/Codegen/Dialect/PCF/IR/PCFOps.h"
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUVerifyDistribution.cpp
```diff
@@ -44,26 +48,62 @@ struct GPUVerifyDistributionPass final
...
-                                "parent subgroup distributed loop.");
-          return WalkResult::interrupt();
-        }
-        return WalkResult::advance();
-      }
-      if (auto memoryEffectOp = dyn_cast<MemoryEffectOpInterface>(op)) {
-        if (!operationHasParentForallOfMappingType<
-                mlir::gpu::GPUThreadMappingAttr, IREE::GPU::LaneIdAttr>(op)) {
+    WalkResult res =
+        funcOp->walk<WalkOrder::PreOrder>([&](Operation *op) -> WalkResult {
```
kuhar, line 52: yikes, what is clang-tidy smoking here xD

kuhar: we should file a bug

kuhar (approved):

PR #23506, opened 2026-02-18
[VectorDistribute] Support linalg.generic in layout analysis

sommerlukas:
Add support for forward and backward propagation of layouts through `linalg.generic`. Forward propagation will always be able to determine the layout of the result, if all operand layouts are known. In backward propagation, if only the result layout is known and there's a reduction involved, not all operand layouts can be determined.

Despite its name the `VectorLayoutAnalysis` works on all shaped types and adding support for `linalg.generic` is a preparatory step to use the analysis earlier in the pipeline as part of https://github.com/iree-org/iree/issues/23415.

Assisted-by: Claude Code

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -52,6 +52,10 @@ struct LayoutInfo {
   }
   bool hasLayout(Value val) const { return layouts.contains(val); }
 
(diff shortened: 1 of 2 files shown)


sommerlukas:
Marked as draft for now to check feasibility of the planned approach for https://github.com/iree-org/iree/issues/23415.

sommerlukas:
This is not needed anymore after taking another direction for https://github.com/iree-org/iree/issues/23415, closing this PR.

PR #23509, opened 2026-02-18
[NFC][Codegen] Update InnerTiledDescAttrInterface to not use void methods when unnecessary

Muzammiluddin-Syed-ECE:
Within the `InnerTileDescAttrInterface` we have some interface methods which have return values, and others which are void and update output parameters. For interface methods like `populateOperandOffsetsSizesStrides` this makes sense because it has multiple outputs and this is a valid alternative to making a struct to house three `SmallVector`s. However, for interface methods updating only a single argument (`getUndistributedTileTypes`) this does not make sense and is inconsistent with the other interface methods with return values (`getUndistributedTileDimExpansion`).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
@@ -117,26 +117,20 @@ def IREECodegen_InnerTileDescAttrInterface :
         ```
         where `tA`, `tB`, and `tC` are the element types of matrices A, B, and C,
         respectively.
-
-        The returned values are placed into the `result` vector, which will have
-        its existing contents overwritten if there is any.
       }],
-      /*retType=*/"void",
+      /*retType=*/"::llvm::SmallVector<::mlir::VectorType, 4>",
       /*methodName=*/"getUndistributedTileTypes",
-      /*args=*/(ins "::llvm::SmallVectorImpl<::mlir::VectorType>&":$result)
+      /*args=*/(ins)
     >,
     InterfaceMethod<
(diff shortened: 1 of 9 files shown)


krzysz00:
As a counterpoint, returning into a SmallVectorImpl is fairly common practice in LLVM because it lets the caller choose the size of the SmallVector depending on their expectations. I'm pretty sure there's style guide about this

Muzammiluddin-Syed-ECE:
> As a counterpoint, returning into a SmallVectorImpl is fairly common practice in LLVM because it lets the caller choose the size of the SmallVector depending on their expectations. I'm pretty sure there's style guide about this

Ah yes I see it now in the ProgrammersManual.rst

```
#. Many LLVM APIs take a ``SmallVectorImpl`` as an out parameter (see the note
   below).

#. ``SmallVector`` with N equal to 0 is smaller than ``std::vector`` on 64-bit
   platforms, since it uses ``unsigned`` (instead of ``void*``) for its size
   and capacity.

.. note::

   Prefer to use ``ArrayRef<T>`` or ``SmallVectorImpl<T>`` as a parameter type.

   It's rarely appropriate to use ``SmallVector<T, N>`` as a parameter type.
   If an API only reads from the vector, it should use :ref:`ArrayRef
   <dss_arrayref>`.  Even if an API updates the vector the "small size" is
   unlikely to be relevant; such an API should use the ``SmallVectorImpl<T>``
   class, which is the "vector header" (and methods) without the elements
   allocated after it. Note that ``SmallVector<T, N>`` inherits from
   ``SmallVectorImpl<T>`` so the conversion is implicit and costs nothing. E.g.
```

Muzammiluddin-Syed-ECE:
No need for the change if this is preferred.

PR #23510, opened 2026-02-18
[VectorExt] Refactor transfer_gather to use unified indexing_maps

Groverkss:
Refactors transfer_gather to use a single `indexing_maps` attribute that encodes source, index vector, and mask mappings in a unified AffineMap array. This replaces the previous separate indexed/permutation_map attributes with a more expressive representation.

Example — embedding lookup: reading rows from a 3D source where the row
index is gathered from an index vector, while the column is contiguous:

```mlir
// result[i, j] = base[0, indices[i], j]
%result = iree_vector_ext.transfer_gather %base[%c0, %c0, %c0]
  [%indices : vector<16xindex>], %pad {
    indexing_maps = [
      affine_map<(d0, d1)[s0] -> (0, s0, d1)>,
      affine_map<(d0, d1)[s0] -> (d0)>
    ]
  } : memref<4096x512x8xf16>, vector<16x8xf16>
```

The main problem with the previous representation was it required the indexing maps to be iterated over the memory space. This doesn't work at all for "ragged tensors", where you are doing something like:

```mlir
// result[i, j] = base[0, indices[i], j]
%result = iree_vector_ext.transfer_gather %base[%c0]
  [%indices : vector<16x8xindex>], %pad {
    indexing_maps = [
      affine_map<(d0, d1)[s0] -> (s0)>,
      affine_map<(d0, d1)[s0] -> (d0, d1)>
    ]
  } : memref<?xf16>, vector<16x8xf16>
```

As we move to handle more complex gathers, we need this form of transfer_gather representation. Note that this form of transfer_gather is more expressive than the previous form, and should handle everything the previous form did.

This form also enables us to directly vectorize a tensor.extract inside a linalg.generic to a transfer_gather, rather than vector.gather.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -676,9 +676,28 @@ struct DistributeTransferGather final
           gatherOp, "warp or thread tiles have overlapping strides");
     }
 
-    ValueRange indices = gatherOp.getIndices();
-    AffineMap permMap = gatherOp.getPermutationMap();
+    ValueRange indices = gatherOp.getOffsets();
     SmallVector<int64_t> strides(rank, 1);
+
+    // Derive the permutation map from the source indexing map.
+    // The source map is (vector_dims)[symbols] -> (source_dims).
+    // We need to build a permutation map (source_dims) -> (vector_dims)
+    // for getTransferIndicesFromNestedLayout.
+    AffineMap sourceMap = gatherOp.getIndexingMapsArray().front();
+    int64_t sourceRank =
+        cast<ShapedType>(gatherOp.getBase().getType()).getRank();
+    MLIRContext *ctx = rewriter.getContext();
+    SmallVector<AffineExpr> permExprs(rank, getAffineConstantExpr(0, ctx));
+    SmallVector<bool> isGatheredSourceDim(sourceRank, false);
+    for (auto [j, expr] : llvm::enumerate(sourceMap.getResults())) {
+      if (auto dimExpr = dyn_cast<AffineDimExpr>(expr)) {
+        unsigned vecDim = dimExpr.getPosition();
+        permExprs[vecDim] = getAffineDimExpr(j, ctx);
+      } else if (isa<AffineSymbolExpr>(expr)) {
+        isGatheredSourceDim[j] = true;
+      }
+    }
+    AffineMap permMap = AffineMap::get(sourceRank, 0, permExprs, ctx);
 
     SmallVector<SmallVector<int64_t>> allMaskOffsets;
     std::vector<StaticTileOffsetRange::IteratorTy> allIndexVecOffsets;
@@ -696,19 +715,16 @@ struct DistributeTransferGather final
           StaticTileOffsetRange(vecDistShape, vecTileShape).begin());
     }
 
-    SmallVector<bool> indexed =
-        llvm::to_vector(gatherOp.getIndexed().getAsValueRange<BoolAttr>());
-
     for (auto [idx, offsets] :
          llvm::enumerate(StaticTileOffsetRange(distShape, tileShape))) {
       SmallVector<Value> slicedIndices = getTransferIndicesFromNestedLayout(
           rewriter, indices, offsets, vectorLayout, permMap, warpIndices,
           threadIndices);
 
-      // Only take the sliced indices for the non-indexed values.
+      // Only take the sliced indices for the non-indexed (non-gathered) values.
       for (auto [dim, slicedIndex, index] :
            llvm::enumerate(slicedIndices, indices)) {
-        if (indexed[dim]) {
+        if (isGatheredSourceDim[dim]) {
           slicedIndex = index;
         }
       }
@@ -737,9 +753,8 @@ struct DistributeTransferGather final
 
       VectorValue slicedGather = IREE::VectorExt::TransferGatherOp::create(
           rewriter, gatherOp.getLoc(), innerVectorType, gatherOp.getBase(),
-          slicedIndices, slicedIndexVecs, gatherOp.getIndexed(),
-          gatherOp.getIndexedMaps(), gatherOp.getPermutationMapAttr(),
-          gatherOp.getPadding(), slicedMask, gatherOp.getInBoundsAttr());
+          slicedIndices, slicedIndexVecs, gatherOp.getIndexingMapsAttr(),
+          gatherOp.getPadding(), slicedMask);
 
       if (acc.getType().getRank() == 0) {
         // TODO: This should really be a folding pattern in
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -233,11 +233,11 @@ static LogicalResult populateWarpAndThreadIndices(
   return success();
 }
 
-static VectorValue getSlicedPermutedMask(PatternRewriter &rewriter,
-                                         Location loc,
-                                         ArrayRef<int64_t> offsets,
-                                         NestedLayoutAttr vectorLayout,
-                                         VectorValue mask) {
+static VectorValue getSlicedPermutedValue(PatternRewriter &rewriter,
+                                          Location loc,
+                                          ArrayRef<int64_t> offsets,
+                                          NestedLayoutAttr vectorLayout,
+                                          VectorValue mask) {
   SmallVector<int64_t> sliceMaskOffsets =
       getDistributedTransferOffsetsFromNestedLayout(offsets, vectorLayout);
   SmallVector<int64_t> strides(vectorLayout.getElementTile().size(), 1);
@@ -245,19 +245,6 @@ static VectorValue getSlicedPermutedMask(PatternRewriter &rewriter,
       rewriter, loc, mask, sliceMaskOffsets, vectorLayout.getElementTile(),
       strides);
   return slicedMask;
-}
-
-static VectorValue getSlicedIndexVec(PatternRewriter &rewriter, Location loc,
-                                     ArrayRef<int64_t> offsets,
-                                     NestedLayoutAttr vectorLayout,
-                                     VectorValue indexVec) {
-  SmallVector<int64_t> sliceMaskOffsets =
-      getDistributedTransferOffsetsFromNestedLayout(offsets, vectorLayout);
-  SmallVector<int64_t> strides(vectorLayout.getElementTile().size(), 1);
-  VectorValue slicedIndexVec = vector::ExtractStridedSliceOp::create(
-      rewriter, loc, indexVec, sliceMaskOffsets, vectorLayout.getElementTile(),
-      strides);
-  return slicedIndexVec;
 }
 
 /// Project a vector based on a provided projection map.
@@ -393,8 +380,8 @@ struct DistributeTransferRead final
         SmallVector<int64_t> maskTileShape =
             getElementVectorTileShape(maskLayout);
         SmallVector<int64_t> maskOffsets = allMaskOffsets[idx];
-        slicedMask = getSlicedPermutedMask(rewriter, readOp.getLoc(),
-                                           maskOffsets, maskLayout, mask);
+        slicedMask = getSlicedPermutedValue(rewriter, readOp.getLoc(),
+                                            maskOffsets, maskLayout, mask);
       }
 
       VectorValue slicedRead = vector::TransferReadOp::create(
@@ -576,8 +563,8 @@ struct DistributeTransferWrite final
         SmallVector<int64_t> maskTileShape =
             getElementVectorTileShape(maskLayout);
         SmallVector<int64_t> maskOffsets = allMaskOffsets[idx];
-        slicedMask = getSlicedPermutedMask(rewriter, writeOp.getLoc(),
-                                           maskOffsets, maskLayout, mask);
+        slicedMask = getSlicedPermutedValue(rewriter, writeOp.getLoc(),
+                                            maskOffsets, maskLayout, mask);
       }
 
       vector::TransferWriteOp::create(rewriter, writeOp.getLoc(), slicedVector,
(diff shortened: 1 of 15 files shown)


compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -117,128 +117,95 @@ def IREEVectorExt_ToSIMTOp : IREEVectorExt_PureOp<"to_simt",
...
+                   "base", "padding", "getElementTypeOrSelf($_self)">
   ]> {
   let arguments = (ins AnyShaped:$base,
-                       Variadic<Index>:$indices,
-                       Variadic<VectorOfAnyRankOf<[Index]>>:$index_vecs,
-                       BoolArrayAttr:$indexed,
-                       AffineMapArrayAttr:$indexed_maps,
-                       AffineMapAttr:$permutation_map,
+                       Variadic<Index>:$offsets,
+                       Variadic<AnyTypeOf<[Index, VectorOfAnyRankOf<[Index]>]>>:$index_vecs,
```
amd-eochoalo, line 129: Very minor comment: I noticed that index_vecs can be an Index but I couldn't find a test where there it is a scalar index. The most similar test is one in canonicalize with the following:

```
  [%scalar, %indices : vector<1xindex>, vector<64x1xindex>], %cst0 {
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -612,37 +357,147 @@ OpFoldResult TransferGatherOp::fold(FoldAdaptor adaptor) {
...
+  }
+
+  return result;
+}
+
 struct FoldSingleElementIndexVec final : OpRewritePattern<TransferGatherOp> {
   using Base::Base;
 
-  LogicalResult matchAndRewrite(TransferGatherOp xferOp,
+  LogicalResult matchAndRewrite(TransferGatherOp op,
```
amd-eochoalo, line 436: Minor comment: I believe this could be split into two patterns where one pattern focuses on making vector<1xindex> into index and the second pattern would focus on adding it to the base offset.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -676,10 +676,29 @@ struct DistributeTransferGather final
...
+    SmallVector<AffineExpr> permExprs(rank, getAffineConstantExpr(0, ctx));
+    SmallVector<bool> isGatheredSourceDim(sourceRank, false);
+    for (auto [j, expr] : llvm::enumerate(sourceMap.getResults())) {
+      if (auto dimExpr = dyn_cast<AffineDimExpr>(expr)) {
+        unsigned vecDim = dimExpr.getPosition();
+        permExprs[vecDim] = getAffineDimExpr(j, ctx);
+      } else if (isa<AffineSymbolExpr>(expr)) {
+        isGatheredSourceDim[j] = true;
+      }
+    }
```
amd-eochoalo: Would it be a good idea to outline this into its own function?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -612,37 +357,147 @@ OpFoldResult TransferGatherOp::fold(FoldAdaptor adaptor) {
   return OpFoldResult();
 }
 
+/// Invert the source indexing map to get a permutation_map for
```
amd-eochoalo: Is this similar to that other function I suggested to outline?

amd-eochoalo (approved):
I'm not too familiar with this part of the code yet. From what I understand this makes sense. I only had some minor comments but you might want to wait for another reviewer who is more familiar with the code base.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
Groverkss: That does make sense. For some reason I thought that's what it was doing.

Groverkss: Actually, it just adds more overhead. It's easier to just do it in one go.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: It's not being reused so I don't think it's worth it. Maybe when we add transfer_scatter as well, then we can outline it.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
Groverkss: No this is different. This assumes that the indexing map doesn't have any symbols.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: Maybe I should still do it in preperation for the transfer_scatter stuff.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
Groverkss: Actually, it is, now that I think of it, just don't know how to share it other than putting it as an extraClassDeclaration method i guess.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: done

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
Groverkss: added

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
Groverkss: done

Groverkss:
Landing for now if CI passes, would appreciate post commit reviews.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution.mlir
```diff
@@ -1268,11 +1268,10 @@ func.func @paged_transfer_gather(%indices: vector<16xindex>,
   %c0 = arith.constant 0 : index
 
   %out = iree_vector_ext.transfer_gather %source[%c0, %c0, %c0]
-  [None, %indices: vector<16xindex>, None], %cst0 { indexed_maps = [
-                                             affine_map<(d0, d1, d2) -> (d1)>],
-    permutation_map = affine_map<(d0, d1, d2) -> (d1, d2)>,
-    in_bounds = [true, true] }
-  : memref<4096x512x8xf16>, vector<16x8xf16>
+  [%indices : vector<16xindex>], %cst0 {
+    indexing_maps = [affine_map<(d0, d1)[s0] -> (0, s0, d1)>,
```
krzysz00, line 1272: Ok so why is there an s0 term here? I'm trying to work out what these maps pean

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -63,404 +61,115 @@ void TransferGatherOp::getEffects(
...
+    return emitOpError("expected ")
+           << expectedNumIndexingMaps
+           << " indexing maps, got: " << indexingMaps.size();
+  }
+
+  int64_t vectorRank = vector.getType().getRank();
+  int64_t indexSyms = indexVecs.size();
+  for (AffineMap map : indexingMaps) {
+    if (map.getNumDims() != vectorRank) {
+      return emitOpError("expected all indexing maps to have number of dims "
```
krzysz00, line 92: Ok so, just for my notes, this is one of the main departures from how transfer_read works

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -469,83 +178,112 @@ static int64_t getVectorRank(Type type) {
   return 0;
 }
 
-struct IndexVecFoldResult {
-  Value indexVec;
-  AffineMap indexMap;
+struct IndexingMapFoldResult {
```
krzysz00, line 181: Tidy warning here

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -612,37 +357,131 @@ OpFoldResult TransferGatherOp::fold(FoldAdaptor adaptor) {
...
+  // Build broadcast shape: [broadcast dim sizes..., source shape...]
+  SmallVector<int64_t> broadcastShape;
+  for (int64_t i = 0; i < targetRank; ++i) {
+    if (targetDimToSourceDim[i] == -1) {
+      broadcastShape.push_back(targetShape[i]);
+    }
+  }
+  for (int64_t i = 0; i < sourceRank; ++i) {
+    broadcastShape.push_back(sourceType.getDimSize(i));
+  }
```
krzysz00, line 407: Isn't this loop just an append_rsnge on get shape()?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -117,128 +117,114 @@ def IREEVectorExt_ToSIMTOp : IREEVectorExt_PureOp<"to_simt",
 }
 
 def IREEVectorExt_TransferGatherOp : IREEVectorExt_PureOp<"transfer_gather", [
-    DeclareOpInterfaceMethods<VectorTransferOpInterface>,
```
krzysz00, line 120: Does this break any pattern matches that might otherwise have run? And what motivated this?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -117,128 +117,114 @@ def IREEVectorExt_ToSIMTOp : IREEVectorExt_PureOp<"to_simt",
...
+      return SmallVector<AffineMap>(
+          maps.begin() + 1, maps.begin() + 1 + getIndexVecs().size());
+    }
+
+    std::optional<AffineMap> getMaskIndexingMap() {
+      if (!getMask()) return std::nullopt;
+      return getIndexingMapsArray().back();
+    }
+
+    /// Invert the source indexing map to get a permutation map suitable for
```
krzysz00, line 205: Should this code be in the .cpp instead? (I know there isn't consistent style here)

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution.mlir
krzysz00: ... Ok, having read further, I can see how this is meant to work

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td
```diff
@@ -117,128 +117,114 @@ def IREEVectorExt_ToSIMTOp : IREEVectorExt_PureOp<"to_simt",
...
-    For every dimension that is gathered, the operation defines how it is
-    gathered. For each gathered dimension, the operation expects a vector of
-    indices in `index_vecs` to act as a source of indices for that dimension
-    and an AffineMap in `index_maps` describing how this source of indices is
-    indexed. For example, for the following gather:
+    - Map 0 (source map): `(vector_dims)[symbols] -> (source_dims)`.
+      A dim expr means the source dimension is contiguous (iterated in
+      lockstep with the vector dimension). A symbol expr means the source
+      dimension is gathered (looked up via the corresponding index vector).
+      A constant 0 means the source dimension is broadcast (always reads
```
krzysz00, line 159: I'm not sure "broadcast" is always the right word here but I can't think of a better one.

krzysz00 (commented):
Sorry for delayed review, I think this reworking makes sense, just a few questions

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -63,404 +61,115 @@ void TransferGatherOp::getEffects(
...
-      return emitOpError(
-          "requires formal padding and source of the same elemental type");
+    // Scalar index: map must have 0 results and type must be plain index.
+    if (isa<IndexType>(indexVecs[i].getType())) {
+      if (!expectedShape.empty()) {
+        return emitOpError("expected empty indexing map for scalar index vec "
+                           "at position ")
+               << i;
+      }
+      continue;
```
Max191, line 138: Is there any point in allowing plain indices for index vectors? Isn't that equivalent to just adding the index to the offsets into the source?

It doesn't really matter all that much, since it probably lowers the same either way, but maybe it makes other logic simpler to have this constraint.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -63,404 +61,115 @@ void TransferGatherOp::getEffects(
...
+      return emitOpError("Mismatched mask shape")
+             << ". Expected: [" << expectedShape << "]" << ", got: ["
+             << actualShape << "]";
     }
   }
 
   return success();
 }
 
-ParseResult TransferGatherOp::parse(OpAsmParser &parser,
```
Max191, line 356: It brings me joy to see all this printing/parsing logic go away :D

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -469,83 +178,112 @@ static int64_t getVectorRank(Type type) {
...
+      mask = newMask;
+      maskMap = newMap;
+    }
+  }
   if (!changed) {
     return Value();
   }
 
-  OpBuilder b(gatherOp);
+  OpBuilder b(op);
```
Max191, line 235: I know this was pre-existing, but why do we create a new builder here instead of passing the builder from the caller? This can prevent rewrite listeners from detecting changes. Is that intentional?

Max191 (commented):
Overall, LGTM, just a few post-merge comments:

PR #23511, opened 2026-02-18
[Common] Update CombineSourceLayoutTransform pass to fold broadcast

Abhishek-Varma:
This commit makes following updates :-
1. Folds named/generic broadcast op into MapLoadOp.
2. Inserts identity MapLoadOp only for "complex" relayout => chains length ≥ 2, has reshape (expand/collapse), and has non-extract_slice op (transpose, pad, copy, broadcast). 
3. Avoids map_load for simple chains (e.g. single extract_slice) to prevent large memref.alloca and stack size issues.
4. Adds pass flag/option `test-combine-non-complex-chains` for lit-testing each op in isolation.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -19,6 +19,7 @@
 #include "mlir/Dialect/UB/IR/UBOps.h"
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/IRMapping.h"
+#include "mlir/Interfaces/DestinationStyleOpInterface.h"
 #include "mlir/Interfaces/ValueBoundsOpInterface.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
@@ -553,6 +554,34 @@ bool isSupportedSingleInputRelayoutOpForSource(Operation *op) {
   return linalgOp && linalg::isaBroadcastOpInterface(linalgOp).has_value();
 }
 
+/// Returns true if `user` is a supported relayout op that uses `result` as a
+/// valid source for extending the relayout chain.
+///
+/// We use different logic for DPS vs non-DPS ops so neither case needs to
+/// consider operand numbers:
+/// - DPS ops (e.g. linalg broadcast): Iterate getDpsInputOperands(). Only
+///   consider uses as an *input* operand.
+/// - Non-DPS ops (expand_shape, collapse_shape, transpose, etc.): Caller
+///   iterates result.getUsers(), so we know user consumes result. These ops
+///   have a single input.
+static bool isRelayoutChainExtension(Operation *user, Value result) {
+  if (!isSupportedSingleInputRelayoutOpForSource(user)) {
+    return false;
+  }
+  auto dpsOp = dyn_cast<DestinationStyleOpInterface>(user);
+  if (dpsOp) {
+    for (OpOperand *input : dpsOp.getDpsInputOperands()) {
+      if (input->get() == result) {
+        return true;
+      }
+    }
+    return false;
+  }
+  // Non-DPS: caller iterates result.getUsers(), so user consumes result.
+  // Single-input relayout ops have only one place for it.
+  return true;
+}
+
 /// Collects all relayout ops in the chain starting from `relayoutOp`
 /// (inclusive). The chain extends through result->user edges where the user is
 /// a supported relayout op.
@@ -563,8 +592,7 @@ static void collectRelayoutChain(Operation *relayoutOp,
   }
   Value result = relayoutOp->getResult(0);
   for (Operation *user : result.getUsers()) {
-    if (isSupportedSingleInputRelayoutOpForSource(user) &&
-        user->getOperand(0) == result) {
+    if (isRelayoutChainExtension(user, result)) {
       collectRelayoutChain(user, chain);
     }
   }
@@ -586,16 +614,14 @@ static bool isComplexRelayoutChain(Operation *relayoutOp) {
   if (chain.size() < 2) {
     return false;
   }
-  bool hasReshape = llvm::any_of(chain, [](Operation *op) {
-    return isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp>(op);
-  });
+  bool hasReshape = llvm::any_of(
+      chain, llvm::IsaPred<tensor::ExpandShapeOp, tensor::CollapseShapeOp>);
   // Need at least one op that is not reshape and not extract_slice (e.g. copy,
   // transpose, broadcast, pad).
-  bool hasOtherNonExtractSlice = llvm::any_of(chain, [](Operation *op) {
-    return !isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp,
-                tensor::ExtractSliceOp>(op);
-  });
-  return hasReshape && hasOtherNonExtractSlice;
+  bool allReshapeOrExtractSlice = llvm::all_of(
+      chain, llvm::IsaPred<tensor::ExpandShapeOp, tensor::CollapseShapeOp,
+                           tensor::ExtractSliceOp>);
+  return hasReshape && !allReshapeOrExtractSlice;
 }
 
 /// Collects direct relayout op users of `loadResult` that start complex
@@ -606,8 +632,7 @@ getComplexChainRelayoutUsers(Value loadResult,
                              bool combineNonComplexChains = false) {
   SmallPtrSet<Operation *, 4> complexUsers;
   for (Operation *user : loadResult.getUsers()) {
-    if (isSupportedSingleInputRelayoutOpForSource(user) &&
-        user->getOperand(0) == loadResult &&
+    if (isRelayoutChainExtension(user, loadResult) &&
         (combineNonComplexChains || isComplexRelayoutChain(user))) {
       complexUsers.insert(user);
     }
@@ -1177,8 +1202,7 @@ struct FoldConsumerRelayoutIntoMapLoadPattern
     Operation *consumerOp = nullptr;
     Value mapLoadResult = mapLoadOp.getResult(0);
     for (Operation *user : mapLoadOp->getUsers()) {
-      if (isSupportedSingleInputRelayoutOpForSource(user) &&
-          user->getOperand(0) == mapLoadResult) {
+      if (isRelayoutChainExtension(user, mapLoadResult)) {
         consumerOp = user;
         break;
       }
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -13,6 +13,7 @@
 #include "llvm/Support/DebugLog.h"
 #include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
 #include "mlir/Dialect/MemRef/Transforms/Transforms.h"
 #include "mlir/Dialect/UB/IR/UBOps.h"
@@ -539,9 +540,46 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
 }
 
 bool isSupportedSingleInputRelayoutOp(Operation *op) {
-  return isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp,
-             tensor::ExtractSliceOp, tensor::PadOp, linalg::CopyOp,
-             linalg::TransposeOp>(op);
+  if (isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp,
+          tensor::ExtractSliceOp, tensor::PadOp, linalg::CopyOp,
+          linalg::TransposeOp>(op)) {
+    return true;
+  }
+  auto genericOp = dyn_cast<linalg::GenericOp>(op);
+  return genericOp && linalg::isaBroadcastOpInterface(genericOp).has_value();
+}
+
+/// Returns true if the relayout op starts a "complex" chain.
+/// A "complex" chain is :-
+/// - a chain of relayout ops with length >= 2,
+/// - or a chain of length 1 with one of the supported linalg relayout ops.
+static bool isComplexRelayoutChain(Operation *relayoutOp) {
+  assert(isSupportedSingleInputRelayoutOp(relayoutOp) &&
+         "expected a supported relayout op");
+  Value result = relayoutOp->getResult(0);
+  bool hasRelayoutUser = llvm::any_of(result.getUsers(), [](Operation *user) {
+    return isSupportedSingleInputRelayoutOp(user);
+  });
+  // Chain length >= 2 -> complex.
+  if (hasRelayoutUser) {
+    return true;
+  }
+  // Chain length 1: complex only if the op is a linalg op.
+  return isa<linalg::LinalgOp>(relayoutOp);
+}
+
+/// Collects direct relayout op users of `loadResult` that start a complex
+/// relayout chain.
+static SmallPtrSet<Operation *, 4>
+getComplexChainRelayoutUsers(Value loadResult) {
+  SmallPtrSet<Operation *, 4> complexUsers;
+  for (Operation *user : loadResult.getUsers()) {
+    if (isSupportedSingleInputRelayoutOp(user) &&
+        user->getOperand(0) == loadResult && isComplexRelayoutChain(user)) {
+      complexUsers.insert(user);
+    }
+  }
+  return complexUsers;
 }
 
 // This is only desirable in the dispatch scope but not in the workgroup scope.
@@ -996,6 +1034,30 @@ foldExtractSliceIntoMapLoad(RewriterBase &rewriter,
                                      indexTransformBuilder);
 }
 
+/// Fold a consumer broadcast `linalg.generic` into a producer `map_load`.
+static FailureOr<MapLoadOp> foldBroadcastGenericIntoMapLoad(
+    RewriterBase &rewriter, linalg::GenericOp genericOp, MapLoadOp mapLoadOp) {
+  assert(genericOp.getDpsInputs()[0] == mapLoadOp.getResult(0) &&
+         "expected map_load to be the producer of genericOp input");
+  if (!linalg::isaBroadcastOpInterface(genericOp).has_value()) {
+    return rewriter.notifyMatchFailure(genericOp,
+                                       "generic op is not a broadcast");
+  }
+
+  AffineMap inputMap = genericOp.getIndexingMapsArray()[0];
+  return foldConsumerIntoMapLoadImpl(
+      rewriter, genericOp, mapLoadOp,
+      [inputMap](ArrayRef<BlockArgument> indices) -> SmallVector<Value> {
+        SmallVector<Value> sourceIndices;
+        sourceIndices.reserve(inputMap.getNumResults());
+        for (AffineExpr expr : inputMap.getResults()) {
+          unsigned pos = cast<AffineDimExpr>(expr).getPosition();
+          sourceIndices.push_back(indices[pos]);
+        }
+        return sourceIndices;
+      });
+}
+
 /// Fold a consumer `padOp` into a producer `mapLoadOp`.
 /// Index transformation: source_idx = new_idx - low_pad
 /// Fill value is set to the pad value.
@@ -1063,6 +1125,9 @@ FailureOr<MapLoadOp> foldIntoMapLoad(RewriterBase &rewriter, Operation *op,
       .Case<tensor::PadOp>([&](tensor::PadOp padOp) {
         return foldPadIntoMapLoad(rewriter, padOp, mapLoadOp);
       })
+      .Case<linalg::GenericOp>([&](linalg::GenericOp genericOp) {
+        return foldBroadcastGenericIntoMapLoad(rewriter, genericOp, mapLoadOp);
+      })
       .Default([](Operation *) { return failure(); });
 }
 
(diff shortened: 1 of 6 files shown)


Abhishek-Varma:
Ping for review.

hanhanW:
> Ping for review.

I'll take a look later today. I need to catch up context from the issue.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -532,9 +533,46 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
-             tensor::ExtractSliceOp, tensor::PadOp, linalg::CopyOp,
-             linalg::TransposeOp>(op);
+  if (isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp,
+          tensor::ExtractSliceOp, tensor::PadOp, linalg::CopyOp,
+          linalg::TransposeOp>(op)) {
+    return true;
+  }
+  auto genericOp = dyn_cast<linalg::GenericOp>(op);
+  return genericOp && linalg::isaBroadcastOpInterface(genericOp).has_value();
+}
```
hanhanW: I'm now concerned about this function because they are shared between source folding and destination folding. I think it is better to have two functions (on per each) because now they diverge. It seems easy to diverge and it is just ~6 lines, so we don't over abstract it.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -532,9 +533,46 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+
+/// Returns true if the relayout op starts a "complex" chain.
+/// A "complex" chain is :-
```
hanhanW: style nit: having `:-` is not common in IREE codebase.

```suggestion
/// A "complex" chain is :
```

hanhanW (changes requested):
High-overview question:

> This is done to prevent creating MapLoadOp for simple primitives
like tensor.extract_slice since it unneccesarily ends up creating an
empty tensor for MapLoad's destination -> which in turn leads to
creation of big memref.alloca ops later in the pipeline causing stack
size limit issue.

Do we really need the folding for extract_slice op? The implementation is looking for chained relayout op, should we explicitly match the case to reflect what you're saying here?

I approved the previous PR that allows extract_slice op because it is a mirror of the destionatino folding, but I wonder if we need it for source folding.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -532,9 +533,46 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+/// - a chain of relayout ops with length >= 2,
+/// - or a chain of length 1 with one of the supported linalg relayout ops.
+static bool isComplexRelayoutChain(Operation *relayoutOp) {
```
Max191, line 609: This could still capture some cases that we don't want to fold to map_gather. For example:
```
%load = iree_codegen.load_from_buffer ...
scf.for {
  %extract_slice = tensor.extract_slice %load
  %copy = linalg.copy ins(%extract_slice) ...
}
```
Or even just:
```
%load = iree_codegen.load_from_buffer ...
%copy = linalg.copy ins(%load) ...
```

It would be safer to check more specifically for op sequences that we know will be difficult to handle. I'm thinking we can look for chains that contain at least one reshape op (expand/collapse_shape) and another type of op that isn't extract_slice.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -983,6 +1021,30 @@ foldExtractSliceIntoMapLoad(RewriterBase &rewriter,
                                      indexTransformBuilder);
 }
 
+/// Fold a consumer broadcast `linalg.generic` into a producer `map_load`.
+static FailureOr<MapLoadOp> foldBroadcastGenericIntoMapLoad(
```
Max191: Could we support the named `linalg.broadcast` op as well? The logic here should work for either one if you just change the type to `LinalgOp`.

Max191 (commented):
A couple of quick comments

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Hi @hanhanW @Max191 - I've updated the code to now look identify a chain as "complex" (and thus a candidate to be folded into `map_load`) based on Max's suggestion. So now trivial cases of extract_slices, etc will not get folded into map_load anymore.

Can you please check now?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Good point. Done.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
```diff
@@ -182,3 +131,126 @@ func.func @nested_pads_different_values(%buffer : memref<8x16xf32>) -> tensor<14
...
+
+// -----
+
+// Due to applyPatternsGreedily collapse(expand(x)) gets folded to x, so we end up
+// with load->transpose before our patterns see the chain. And since single transpose
+// is not complex - so no map_load op is inserted.
+func.func @complex_relayout_chain(%buffer : memref<8x16xf32>) -> tensor<16x8xf32> {
+  %source = iree_codegen.load_from_buffer %buffer : memref<8x16xf32> -> tensor<8x16xf32>
+  %expanded = tensor.expand_shape %source [[0, 1], [2]] output_shape [2, 4, 16] : tensor<8x16xf32> into tensor<2x4x16xf32>
+  %collapsed = tensor.collapse_shape %expanded [[0, 1], [2]] : tensor<2x4x16xf32> into tensor<8x16xf32>
```
Max191: nit: The expand + collapse here are inverse of each other, so they can fold with each other. I'd either put the transpose in between the ops, or give the expand/collapse different sets of reassociation indices so that they are not direct inverses of each other.

The way it is right now doesn't really test anything except the canonicalizer that folds them away.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
```diff
@@ -1,111 +1,77 @@
 // RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-combine-source-layout-transformation,canonicalize,cse))" \
```
Max191, line 1: With the recent change to the complex relayout chain matching, many of these ops are no longer being folded, so the test coverage got much worse now. We typically fix issues like these by adding a special pass option for testing.

For example: https://github.com/iree-org/iree/blob/8df2db9ade97f812c311ad11d00c602c8b695444/compiler/src/iree/compiler/GlobalOptimization/Passes.td#L109-L112

You can add a pass option to this pass like `test-combine-non-complex-chains`, and then ignore the complex relayout matching when this pass option is enabled. Then we can have multiple RUN statements in the test for the different options. One to test that only complex chains are folded, and one to verify the transformation for each op in isolation. Here's an example for the options I showed above: https://github.com/iree-org/iree/blob/8df2db9ade97f812c311ad11d00c602c8b695444/compiler/src/iree/compiler/GlobalOptimization/test/propagate_linalg_transpose.mlir#L3-L4

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -531,12 +532,76 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+}
+
+/// Collects direct relayout op users of `loadResult` that start a complex
+/// relayout chain.
+static SmallPtrSet<Operation *, 4>
+getComplexChainRelayoutUsers(Value loadResult) {
+  SmallPtrSet<Operation *, 4> complexUsers;
+  for (Operation *user : loadResult.getUsers()) {
+    if (isSupportedSingleInputRelayoutOpForSource(user) &&
+        user->getOperand(0) == loadResult && isComplexRelayoutChain(user)) {
```
Max191: nit: The `user->getOperand(0) == loadResult` check is not necessary here, because `SingleInputRelayoutOp`s only have one tensor operand, and the loadResult will be a tensor Value because it comes from load_from_buffer. There is no need to check that loadResult is consumed by operand 0 specifically. Same with the other places where you have this check. I'll point them out as well.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -1060,10 +1154,12 @@ struct FoldConsumerRelayoutIntoMapLoadPattern
...
   LogicalResult matchAndRewrite(IREE::LinalgExt::MapLoadOp mapLoadOp,
                                 PatternRewriter &rewriter) const override {
-    // Find a consumer relayout op.
+    // Find a consumer relayout op (one that uses map_load result as its input).
     Operation *consumerOp = nullptr;
+    Value mapLoadResult = mapLoadOp.getResult(0);
     for (Operation *user : mapLoadOp->getUsers()) {
-      if (isSupportedSingleInputRelayoutOp(user)) {
+      if (isSupportedSingleInputRelayoutOpForSource(user) &&
+          user->getOperand(0) == mapLoadResult) {
```
Max191: nit: Same here about `user->getOperand(0) == mapLoadResult`.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -531,12 +532,76 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+/// a supported relayout op.
+static void collectRelayoutChain(Operation *relayoutOp,
+                                 SmallPtrSetImpl<Operation *> &chain) {
+  if (!chain.insert(relayoutOp).second) {
+    return;
+  }
+  Value result = relayoutOp->getResult(0);
+  for (Operation *user : result.getUsers()) {
+    if (isSupportedSingleInputRelayoutOpForSource(user) &&
+        user->getOperand(0) == result) {
```
Max191: nit: Same here about `user->getOperand(0) == result`.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -531,12 +532,76 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+      collectRelayoutChain(user, chain);
+    }
+  }
+}
+
+/// Returns true if the relayout op starts a "complex" chain.
+/// A "complex" chain is one that is difficult for bufferization to handle:
+/// - chain length >= 2,
+/// - contains at least one reshape op (expand_shape or collapse_shape),
+/// - and contains at least one op that is not extract_slice.
```
Max191:
```suggestion
/// A "complex" chain is one that is difficult for bufferization to handle. The conditions are:
/// - The chain length must be >= 2.
/// - The chain must contain at least one reshape op (expand_shape or collapse_shape).
/// - The chain must contain at least one op that is not extract_slice or a reshape op.
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -531,12 +532,76 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+  // transpose, broadcast, pad).
+  bool hasOtherNonExtractSlice = llvm::any_of(chain, [](Operation *op) {
+    return !isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp,
+                tensor::ExtractSliceOp>(op);
+  });
+  return hasReshape && hasOtherNonExtractSlice;
+}
+
+/// Collects direct relayout op users of `loadResult` that start a complex
+/// relayout chain.
```
Max191:
```suggestion
/// Collects direct relayout op users of `loadResult` that start complex
/// relayout chains.
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.h
```diff
@@ -55,8 +55,13 @@ CombineRelayoutOpsControlFn
...
 
 /// Returns true if the `op` type has a folding pattern into
-/// iree_linalg_ext.map_store or iree_linalg_ext.map_load.
-bool isSupportedSingleInputRelayoutOp(Operation *op);
+/// iree_linalg_ext.map_store (used by CombineResultLayoutTransformationPass).
+bool isSupportedSingleInputRelayoutOpForResult(Operation *op);
+
+/// Returns true if the `op` type has a folding pattern into
+/// iree_linalg_ext.map_load (used by CombineSourceLayoutTransformationPass).
+/// Includes broadcast GenericOp in addition to the ops supported for Result.
```
Max191:
```suggestion
/// Includes broadcast LinalgOps in addition to the ops supported in
/// isSupportedSingleInputRelayoutOpForResult.
```

Max191 (commented):
Nice! It looks good now (% some nits) aside from the testing. I think we just need a pass option to test non-complex chains and then testing will be much easier.

Max191:
Also, please update the description to be consistent with the new changes. It is still referring to the old implementation.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
Abhishek-Varma: Done. Can you please check now ?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: So, `.copy` and `.broadcast` ops will have more than one operand. I understand logically it doesn't make sense for  these ops to have the result of the load on their `outs` - but semantically it's still possible.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: I've added [a lit test](https://github.com/Abhishek-Varma/iree/blob/9cc29f479f8502a2de973a07d62931bbe481c3f2/compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir#L306) which necessitated the inclusion of this check.
It wasn't required earlier but yesterday while re-testing the [regressions](https://github.com/iree-org/iree/issues/23038#issuecomment-3879226180) I saw the above case.

The input IR right before this pass runs is [here](https://gist.github.com/Abhishek-Varma/b029eee81f83c4f5310d202f84cdcd8f) - `%19` is a `.copy` op which consumes the output of `.broadcast` op `%14`.

I haven't triaged why it got introduced and why it wasn't required earlier (possibly some new changes to the codegen pipeline). The above is what I observed for the regression that required broadcast op: https://github.com/iree-org/iree/issues/23038#issuecomment-3908514491

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Same reasoning as https://github.com/iree-org/iree/pull/23511#discussion_r2894798274 and https://github.com/iree-org/iree/pull/23511#discussion_r2894797948

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Max191: Ohh, right. I forgot we have DPS ops in here. Okay, this is fine then, you can ignore my comment.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
```diff
@@ -1,111 +1,100 @@
...
-//  CHECK-SAME:   %[[BUFFER:[a-zA-Z0-9_]+]]
-//       CHECK:   %[[SOURCE:.+]] = iree_codegen.load_from_buffer %[[BUFFER]]
-//       CHECK:   %[[DEST:.+]] = tensor.empty() : tensor<4x16x2xf32>
-//   CHECK-NOT:   linalg.transpose
-//       CHECK:   %[[MAP_GATHER:.+]] = iree_linalg_ext.map_load
-//  CHECK-SAME:     %[[SOURCE]] into %[[DEST]] {
-//  CHECK-NEXT:   ^bb0(%[[IDX0:.+]]: index, %[[IDX1:.+]]: index, %[[IDX2:.+]]: index):
-// For perm=[1,2,0], inverse_perm=[2,0,1]: output[i,j,k] = input[k,i,j], so yield [idx2,idx0,idx1].
-//       CHECK:     iree_linalg_ext.yield %[[IDX2]], %[[IDX0]], %[[IDX1]],
-//       CHECK:   } : tensor<2x4x16xf32> into tensor<4x16x2xf32> -> tensor<4x16x2xf32>
```
Max191, line 20: We are losing the checks that confirm the correctness of the transformation now. Could you add these more verbose checks back for the folded checks?

Same for the ops below.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
```diff
@@ -182,3 +162,148 @@ func.func @nested_pads_different_values(%buffer : memref<8x16xf32>) -> tensor<14
...
+}
+// CHECK-LABEL: @broadcast_generic
+//  CHECK-SAME:   %[[BUFFER:.+]]:
+//       CHECK:   iree_codegen.load_from_buffer %[[BUFFER]]
+//       CHECK:   linalg.generic
+//   CHECK-NOT:   iree_linalg_ext.map_load
+// FOLD-LABEL: @broadcast_generic
+//  FOLD-SAME:   %[[BUFFER:.+]]:
+//   FOLD-NOT:   linalg.generic
+//       FOLD:   iree_linalg_ext.map_load
```
Max191: Same here. Can you have some more verbose checks so we are testing that the transformation is done correctly?

Max191 (approved):
Nice work, it LGTM now. Just make sure to address the comments about the tests before landing.

Max191:
Also, let's wait for @hanhanW to give one more review before landing.

compiler/src/iree/compiler/Codegen/Common/test/combine_source_layout_transformation.mlir
Abhishek-Varma: Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
hanhanW: In this case, should we have two branches here? One for DPS ops and the other for non-DPS ops.

For DPS ops, we iterate the `getInputs()`; for the non-DPS ops, we just add user. Then both cases no longer have to consider the operand number.

It is not clear to me when I look at the code and comment, so I think it is easier if we can make the change and document it. I think the core reason is that DPS perserves the semantics for destination while non-DPS ops do not have such semantic, thus we don't fold when it appears to DPS's inits, right?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -531,12 +532,82 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+  }
+  bool hasReshape = llvm::any_of(chain, [](Operation *op) {
+    return isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp>(op);
+  });
+  // Need at least one op that is not reshape and not extract_slice (e.g. copy,
+  // transpose, broadcast, pad).
+  bool hasOtherNonExtractSlice = llvm::any_of(chain, [](Operation *op) {
+    return !isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp,
+                tensor::ExtractSliceOp>(op);
+  });
```
hanhanW: We can use `llvm::IsaPred` like below:

https://github.com/iree-org/iree/blob/fa5580cd08e2f237b5156ead7d2d8776a10ef207/compiler/src/iree/compiler/Dialect/Flow/Transforms/InitializeEmptyTensors.cpp#L36-L38

The `any_of` can be `!all_of` I think:

```cpp
  bool allReshapeOrExtractSlice = llvm::all_of(chain, llvm::IsaPred<
      tensor::ExpandShapeOp, tensor::CollapseShapeOp, tensor::ExtractSliceOp>);
  return hasReshape && !allReshapeOrExtractSlice;
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -531,12 +532,82 @@ static MapStoreOp insertIdentityMapStore(RewriterBase &rewriter,
...
+  SmallPtrSet<Operation *, 4> complexUsers;
+  for (Operation *user : loadResult.getUsers()) {
+    if (isSupportedSingleInputRelayoutOpForSource(user) &&
+        user->getOperand(0) == loadResult &&
+        (combineNonComplexChains || isComplexRelayoutChain(user))) {
+      complexUsers.insert(user);
+    }
+  }
+  return complexUsers;
+}
```
hanhanW, line 641: Same here, consider using two branches, so it is easier for read and documentation.

hanhanW (commented):
Just one question for op collection and one nit for implementation.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Makes sense. Done.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Done.

hanhanW (approved):
Thanks!

Issue #22957, opened 2025-12-23
Resnest14 model fails with `error: function main uses 73728 bytes of shared memory; exceeded the limit of 65536 bytes`


vivekkhandelwal1:
### What happened?

After this commit https://github.com/iree-org/iree/commit/c3385bf59f4750c083d0b0729b36f7211d361472 got merged the Resnest14 model started failing during compilation with the below error:
```
error: function 'main_graph$async_dispatch_22_conv_2d_ngchw_gfchw_1x2x128x56x56x64x3x3_f32' uses 73728 bytes of shared memory; exceeded the limit of 65536 bytes
    %94 = torch.operator "onnx.Conv"(%93, %28, %29) {torch.onnx.dilations = [1 : si64, 1 : si64], torch.onnx.group = 2 : si64, torch.onnx.kernel_shape = [3 : si64, 3 : si64], torch.onnx.pads = [1 : si64, 1 : si64, 1 : si64, 1 : si64], torch.onnx.strides = [1 : si64, 1 : si64]} : (!torch.vtensor<[1,128,56,56],f32>, !torch.vtensor<[256,64,3,3],f32>, !torch.vtensor<[256],f32>) -> !torch.vtensor<[1,256,56,56],f32> 
          ^
```

### Steps to reproduce your issue

Get the repro IR from [here](https://gist.github.com/vivekkhandelwal1/d6f6f22ca2b4d86e0940083a8783b99f).

Run:
```
iree-compile repro.mlir --iree-hal-target-backends=rocm --iree-hip-target=gfx942 -o repro.vmfb
```

### What component(s) does this issue relate to?

Compiler

### Version information

Commit: c3385bf59f4750c083d0b0729b36f7211d361472

Max191:
@vivekkhandelwal1 I am back from vacation now, but I can't reproduce this issue anymore. The repro IR is out of date, but after fixing up the IR it compiles normally. Can you verify whether or not this issue still exists? If not, then let's close the issue.

MaheshRavishankar:
cc @YashDeshpande25  can you take a look.

pdhirajkumarprasad:
@Max191 Please use below IR to reproduce the issue

```mlir
hal.executable public @main_graph$async_dispatch_1 {
[...]

PR #23512, opened 2026-02-18
[Flow] Fold broadcast with empty tensor input to empty tensor

Abhishek-Varma:
-- Adds a canonicalization in the Flow phase that folds `broadcast(empty tensor)` into a single tensor.empty() with the broadcast's result shape.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Flow/Transforms/Canonicalize.cpp
+++ b/compiler/src/iree/compiler/Dialect/Flow/Transforms/Canonicalize.cpp
@@ -8,6 +8,8 @@
 
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/Utils.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tensor/Transforms/Transforms.h"
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/Pass/Pass.h"
@@ -84,6 +86,53 @@ struct FoldFullInsertSlice : public OpRewritePattern<tensor::InsertSliceOp> {
   }
 };
 
+/// Returns true if linalgOp is a broadcast. Handles both the named
+/// linalg.broadcast op and broadcast-like linalg.generic.
+static bool isBroadcastLinalgOp(linalg::LinalgOp linalgOp) {
+  if (isa<linalg::BroadcastOp>(linalgOp.getOperation())) {
+    return true;
+  }
+  auto genericOp = dyn_cast<linalg::GenericOp>(linalgOp.getOperation());
+  return genericOp && linalg::isaBroadcastOpInterface(genericOp).has_value();
+}
+
+/// For a broadcast LinalgOp (named or generic), returns the single input and
+/// init values.
+static std::pair<Value, Value> getBroadcastInputAndInit(linalg::LinalgOp op) {
+  if (auto broadcastOp = dyn_cast<linalg::BroadcastOp>(op.getOperation())) {
+    return {broadcastOp.getInput(), broadcastOp.getInit()};
+  }
+  auto genericOp = cast<linalg::GenericOp>(op.getOperation());
+  return {genericOp.getDpsInputOperand(0)->get(),
+          genericOp.getDpsInitOperand(0)->get()};
+}
+
+/// Fold broadcast(tensor.empty()) -> tensor.empty() with the broadcast result
+/// shape. Handles both linalg.broadcast and broadcast-like linalg.generic.
+struct FoldBroadcastWithEmptyTensor
+    : public OpInterfaceRewritePattern<linalg::LinalgOp> {
+  using OpInterfaceRewritePattern<linalg::LinalgOp>::Base::Base;
+
+  LogicalResult matchAndRewrite(linalg::LinalgOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (!isBroadcastLinalgOp(linalgOp)) {
+      return rewriter.notifyMatchFailure(linalgOp, "not a broadcast op");
+    }
+    auto [input, init] = getBroadcastInputAndInit(linalgOp);
+    if (!input.getDefiningOp<tensor::EmptyOp>()) {
+      return rewriter.notifyMatchFailure(linalgOp,
+                                         "input not defined by tensor.empty");
+    }
+    SmallVector<OpFoldResult> resultSizes =
+        tensor::getMixedSizes(rewriter, linalgOp.getLoc(), init);
+    auto resultType = cast<RankedTensorType>(init.getType());
+    Value newEmpty = tensor::EmptyOp::create(
+        rewriter, linalgOp.getLoc(), resultSizes, resultType.getElementType());
+    rewriter.replaceOp(linalgOp.getOperation(), newEmpty);
+    return success();
+  }
+};
+
 /// Convert an "affine.apply" operation into a sequence of arith ops.
 class AffineApplyLowering : public OpRewritePattern<affine::AffineApplyOp> {
 public:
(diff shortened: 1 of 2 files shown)


Groverkss (commented):
Where are we getting inputs that have tensor.empty() as inputs?

Abhishek-Varma:
> Where are we getting inputs that have tensor.empty() as inputs?

The input was shared as part of a regression filed : https://github.com/iree-org/iree/issues/22957#issuecomment-3767133872

We see tensor.empty() as broadcast's input there.
And this PR implements what @MaheshRavishankar suggested here for the Flow::Canonicalizer : https://github.com/iree-org/iree/issues/22957#issuecomment-3910850250

Groverkss (changes requested):

Groverkss:
> > Where are we getting inputs that have tensor.empty() as inputs?
> 
> The input was shared as part of a regression filed : [#22957 (comment)](https://github.com/iree-org/iree/issues/22957#issuecomment-3767133872)
> 
> We see tensor.empty() as broadcast's input there. And this PR implements what @MaheshRavishankar suggested here for the Flow::Canonicalizer : [#22957 (comment)](https://github.com/iree-org/iree/issues/22957#issuecomment-3910850250)

The input in that regression doesn't have tensor.empty() as an input to the broadcast.

```mlir
 %11 = tensor.empty() : tensor<2x64x28x28xf32>
  %12 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0, d1)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%8 : tensor<2x64xf32>) outs(%11 : tensor<2x64x28x28xf32>) {
        ^bb0(%in: f32, %out: f32):
          linalg.yield %in : f32
        } -> tensor<2x64x28x28xf32>
  %13 = linalg.generic ... outs(%12 : tensor<2x64x28x28xf32>)  -> tensor<2x64x28x28xf32>
```

This is a perfectly legal thing to do. There is nothing wrong with it. At max, the broadcast can fold with the linalg.generic, but there is no tensor.empty() being fed as an input to the broadcast.

compiler/src/iree/compiler/Dialect/Flow/Transforms/Canonicalize.cpp
```diff
@@ -84,6 +86,53 @@ struct FoldFullInsertSlice : public OpRewritePattern<tensor::InsertSliceOp> {
...
+  if (isa<linalg::BroadcastOp>(linalgOp.getOperation())) {
+    return true;
+  }
+  auto genericOp = dyn_cast<linalg::GenericOp>(linalgOp.getOperation());
+  return genericOp && linalg::isaBroadcastOpInterface(genericOp).has_value();
+}
+
+/// For a broadcast LinalgOp (named or generic), returns the single input and
+/// init values.
+static std::pair<Value, Value> getBroadcastInputAndInit(linalg::LinalgOp op) {
```
MaheshRavishankar, line 101: You dont need to special case on BroadCastOp and GenericOp. You can just do this on LinalgOp.

compiler/src/iree/compiler/Dialect/Flow/Transforms/Canonicalize.cpp
```diff
@@ -84,6 +86,53 @@ struct FoldFullInsertSlice : public OpRewritePattern<tensor::InsertSliceOp> {
   }
 };
 
+/// Returns true if linalgOp is a broadcast. Handles both the named
+/// linalg.broadcast op and broadcast-like linalg.generic.
+static bool isBroadcastLinalgOp(linalg::LinalgOp linalgOp) {
+  if (isa<linalg::BroadcastOp>(linalgOp.getOperation())) {
```
MaheshRavishankar, line 92: I think the linalg::isaBroadcastOpInterface should cover this as well. We dont need to special case this I think?

MaheshRavishankar (approved):
Looks good.

MaheshRavishankar:
> > > Where are we getting inputs that have tensor.empty() as inputs?
> > 
> > 
> > The input was shared as part of a regression filed : [#22957 (comment)](https://github.com/iree-org/iree/issues/22957#issuecomment-3767133872)
> > We see tensor.empty() as broadcast's input there. And this PR implements what @MaheshRavishankar suggested here for the Flow::Canonicalizer : [#22957 (comment)](https://github.com/iree-org/iree/issues/22957#issuecomment-3910850250)
> 
> The input in that regression doesn't have tensor.empty() as an input to the broadcast.
> 
> ```mlir
>  %11 = tensor.empty() : tensor<2x64x28x28xf32>
>   %12 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0, d1)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%8 : tensor<2x64xf32>) outs(%11 : tensor<2x64x28x28xf32>) {
>         ^bb0(%in: f32, %out: f32):
>           linalg.yield %in : f32
>         } -> tensor<2x64x28x28xf32>
>   %13 = linalg.generic ... outs(%12 : tensor<2x64x28x28xf32>)  -> tensor<2x64x28x28xf32>
> ```
> 
> This is a perfectly legal thing to do. There is nothing wrong with it. At max, the broadcast can fold with the linalg.generic, but there is no tensor.empty() being fed as an input to the broadcast.

I think you are mixing this. Whether the result of the broadcast becomes input or not, using an empty + broadcast is kind of strange thing to do. Its just a new empty. This seems like a perfectly fine canonicalization to do.

Groverkss:
> > > > Where are we getting inputs that have tensor.empty() as inputs?
> > > 
> > > 
> > > The input was shared as part of a regression filed : [#22957 (comment)](https://github.com/iree-org/iree/issues/22957#issuecomment-3767133872)
> > > We see tensor.empty() as broadcast's input there. And this PR implements what @MaheshRavishankar suggested here for the Flow::Canonicalizer : [#22957 (comment)](https://github.com/iree-org/iree/issues/22957#issuecomment-3910850250)
> > 
> > 
> > The input in that regression doesn't have tensor.empty() as an input to the broadcast.
> > ```mlir
> >  %11 = tensor.empty() : tensor<2x64x28x28xf32>
> >   %12 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0, d1)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%8 : tensor<2x64xf32>) outs(%11 : tensor<2x64x28x28xf32>) {
> >         ^bb0(%in: f32, %out: f32):
> >           linalg.yield %in : f32
> >         } -> tensor<2x64x28x28xf32>
> >   %13 = linalg.generic ... outs(%12 : tensor<2x64x28x28xf32>)  -> tensor<2x64x28x28xf32>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is a perfectly legal thing to do. There is nothing wrong with it. At max, the broadcast can fold with the linalg.generic, but there is no tensor.empty() being fed as an input to the broadcast.
> 
> I think you are mixing this. Whether the result of the broadcast becomes input or not, using an empty + broadcast is kind of strange thing to do. Its just a new empty. This seems like a perfectly fine canonicalization to do.

empty + broadcast is not a new empty if the empty is the init operand right? I'm confused. The computation you mention is just a normal broadcast...

The reason I'm a bit against this PR is that we don't have proper propagation rules for tensor.empty. tensor.empty is effectively same as a poison tensor, but we need to define some propagation rules for tensor.empty before we do something like this.

I'm not seeing how it's solving the problem there.

MaheshRavishankar:
Ok, This pattern by itself is good I think, but actually I mis-read the input IR on the issue here. This wont "fix" the issue you were seeing, but as it stands this is a good pattern to have I think . Ill comment on the issue.

compiler/src/iree/compiler/Dialect/Flow/Transforms/Canonicalize.cpp
Abhishek-Varma: It currently doesn't work. I've raised a PR upstream to cover this case : https://github.com/llvm/llvm-project/pull/182806

PR #23513, opened 2026-02-18
Use new vector.multi_reduction populate methods

amd-eochoalo:
`populateVectorMultiReductionLoweringPatterns` is being deprecated upstream. See https://github.com/llvm/llvm-project/pull/180750 for reference. This populate method has been replaced for three populate methods with a finer level of granularity.

* `populateVectorMultiReductionReorderAndExpandPatterns`
* `populateVectorMultiReductionFlatteningPatterns`
* `populateVectorMultiReductionUnrollingPatterns`


krzysz00 (approved):
lgtm

PR #23516, opened 2026-02-19
[GPU][Codegen] Implement virtual sparse mfma (VSMFMA)

efric:
This PR introduces virtual sparse MFMAs (VSMFMA), which implement the sparse trick for skinny GEMM (M=8) workloads originally described in the hugging face article [Creating custom kernels for the AMD MI300](https://huggingface.co/blog/mi300kernels). Specifically, this patch implements `VSMFMA_F32_16x16x32_F16` and `VSMFMA_I32_16x16x64_I8`. Plumbing `VSMFMA` through MMA selection and support for other variants will be follow-up PRs.

Sparse MFMA (V_SMFMAC) instructions perform MMA on an imbalanced pair of operands: a 4:2 structured-sparse matrix A and a dense matrix B. The instruction also takes a sparsity index that encodes which 2 of every 4 elements along K are non-zero within the sparse matrix A. The trick exploits this by pairing even/odd lanes to jointly describe a full dense row. For example, with `i8` (`K=64`, 16 elements per thread along K):

```
Logical row 0 (16 elements along K):
K0 K1 K2 K3 | K4 K5 K6 K7 | K8 K9 K10 K11 | K12 K13 K14 K15

Lane 0 (even, sparse index 0x4444 -> selects positions {0,1} per group of 4):
K0 K1  _  _ | K4 K5  _  _ | K8  K9   _   _  | K12 K13  _   _

Lane 1 (odd,  sparse index 0xEEEE -> selects positions {2,3} per group of 4):
_  _ K2 K3 |  _  _ K6 K7 |  _   _  K10 K11 |  _   _  K14 K15

```

The lane-pairing layout maps each of the 8 logical M-rows to a pair of adjacent physical rows (`row 2i` and `2i+1` for logical row `i`). Within each pair, the even lane supplies positions
  `{0,1}` from each K-group of 4 and the odd lane supplies positions `{2,3}`. The hardware interprets each physical row as having 2:4 structured sparsity and computes a partial dot product over only its non-zero elements. Summing the two physical rows' results reconstructs the full dense dot product for the logical row. This yields a semantic `M=8` matmul from a physical `16×16` instruction.

The benefit over the current approach of padding M=8 to M=16 for dense MFMA is that sparse MFMA processes twice the K-depth of the equivalent dense MFMA in the same number of cycles. The overhead of packing the sparse A operand and collapsing the accumulator partials is outweighed by the throughput gain. In HIP microbenchmarks on gfx942, the sparse trick achieves ~1.5–2x throughput improvement over padding to M=16 for M=8 skinny GEMMs.

For more context: #22863
HIP implementation: https://github.com/nod-ai/hip-matmul/pull/17

Assisted by: Claude

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_contract_amdgpu.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_contract_amdgpu.mlir
@@ -1149,3 +1149,199 @@ builtin.module attributes { transform.with_named_sequence } {
 // CHECK-SAME:    %[[R1]]#0, %[[R1]]#1, %[[R1]]#2, %[[R1]]#3
 // CHECK:       %[[R_SIMD:.+]] = iree_vector_ext.to_simd %[[B_OUT]] : vector<1x2x1x1x4x1xf32> -> vector<32x32xf32>
 // CHECK:       return %[[R_SIMD]]
+
+// -----
+
+// Sparse-trick VSMFMA_F32_16x16x32_F16 (M=8 skinny GEMM via smfmac)
+
+#map1 = affine_map<(m, n, k) -> (m, k)>
+#map2 = affine_map<(m, n, k) -> (k, n)>
+#map3 = affine_map<(m, n, k) -> (m, n)>
+
+#layout_a = #iree_vector_ext.nested_layout<
+  subgroup_tile = [1, 1],
+  batch_tile    = [1, 1],
+  outer_tile        = [1, 1],
+  thread_tile       = [8, 4],
+  element_tile     = [1, 8],
+
+  subgroup_strides        = [1, 1],
+  thread_strides          = [2, 16]
+>
+
+#layout_b = #iree_vector_ext.nested_layout<
+  subgroup_tile = [1, 1],
+  batch_tile    = [1, 1],
+  outer_tile        = [1, 1],
+  thread_tile       = [4, 16],
+  element_tile     = [8, 1],
+
+  subgroup_strides        = [1, 1],
+  thread_strides          = [16, 1]
+>
+
+#layout_c = #iree_vector_ext.nested_layout<
+  subgroup_tile = [1, 1],
+  batch_tile    = [1, 1],
+  outer_tile        = [1, 1],
+  thread_tile       = [4, 16],
+  element_tile     = [2, 1],
+
+  subgroup_strides        = [1, 1],
+  thread_strides          = [16, 1]
+>
+
+func.func @contract_to_vsmfma_f16_8x16x32_mm(%a : vector<8x32xf16>, %b : vector<32x16xf16>, %c : vector<8x16xf32>) -> vector<8x16xf32> {
+  %A = iree_vector_ext.to_layout %a to layout(#layout_a) : vector<8x32xf16>
+  %B = iree_vector_ext.to_layout %b to layout(#layout_b) : vector<32x16xf16>
+  %C = iree_vector_ext.to_layout %c to layout(#layout_c) : vector<8x16xf32>
+
+  %output = vector.contract {
+    indexing_maps = [#map1, #map2, #map3],
+    iterator_types = ["parallel", "parallel", "reduction"],
+    kind = #vector.kind<add>,
+    iree.gpu.mma = #iree_gpu.virtual_mma_layout<VSMFMA_F32_16x16x32_F16>
+  } %A, %B, %C : vector<8x32xf16>, vector<32x16xf16> into vector<8x16xf32>
+
+  %O = iree_vector_ext.to_layout %output to layout(#layout_c) : vector<8x16xf32>
+  return %O : vector<8x16xf32>
+}
+
+builtin.module attributes { transform.with_named_sequence } {
+  transform.named_sequence @__transform_main(%variant_op: !transform.any_op {transform.readonly}) {
+    %top_level_func = transform.structured.match ops{["func.func"]} in %variant_op : (!transform.any_op) -> !transform.any_op
+    transform.iree.test_gpu_vector_distribution %top_level_func : !transform.any_op
+    transform.yield
+  }
+}
+
+// Key checks:
+// 1. Lane parity computation.
+// 2. vector.shuffle to select sparse A elements (even/odd halves).
+// 3. arith.select to choose between even/odd paths based on lane parity.
+// 4. Acc expand (vector.interleave) and collapse (vector.shuffle + pairwise addf).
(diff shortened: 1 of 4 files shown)


Groverkss:
It would be good to understand where this work aligns. My understanding would be that this work would align more when the skinny dimension is something between 4-16, but not < 4, as in the < 4 case, you are memory bound, using more compute doesn't help.

efric:
> It would be good to understand where this work aligns. My understanding would be that this work would align more when the skinny dimension is something between 4-16, but not < 4, as in the < 4 case, you are memory bound, using more compute doesn't help.

Yes you’re right, this is targeting the case when the skinny dimension is exactly 8. 8 is the ideal extent to take advantage of the 4:2 sparsity nature too. <8 means we have to pad to do this or apply the trick twice, which means more overhead and each smfmac call only doing a fraction of the real work.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -1322,21 +1337,192 @@ LogicalResult VirtualMMAAttr::populateOperandOffsetsSizesStrides(
...
+//
+// The layout and distribution infrastructure operate on the collapsed vector
+// (e.g vector<2xf32>). buildVSMFMAOp handles the translation: it expands
+// a collapsed accumulator into the 4-element physical form before the smfmac,
+// then collapses the result back afterward.
+static LogicalResult buildVSMFMAOp(OpBuilder &builder, Location loc,
+                                   const VSMFMAConfig &config,
+                                   ValueRange inputs, Value acc,
+                                   SmallVectorImpl<Value> &results) {
+  Value smfmacAcc = expandAccumulator(builder, loc, acc);
```
efric, line 1497: Atm this means that with kBatch > 1, we will have do this “expand/collapse” for each VSMFMA. We should be able to hoist this out to the caller(s) as future optimization and just do one collapse at the end of the loop.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUEnums.td
```diff
@@ -373,13 +373,17 @@ def VMFMA_F32_16x16x32_F16  : I32EnumAttrCase<"VMFMA_F32_16x16x32_F16", 0>;
 def VMFMA_F32_32x32x16_F16  : I32EnumAttrCase<"VMFMA_F32_32x32x16_F16", 1>;
 def VMFMA_F32_16x16x32_F8E4M3FNUZ  : I32EnumAttrCase<"VMFMA_F32_16x16x32_F8E4M3FNUZ", 2>;
 def VMFMA_F32_32x32x16_F8E4M3FNUZ  : I32EnumAttrCase<"VMFMA_F32_32x32x16_F8E4M3FNUZ", 3>;
+def VSMFMA_F32_16x16x32_F16  : I32EnumAttrCase<"VSMFMA_F32_16x16x32_F16", 4>;
+def VSMFMA_I32_16x16x64_I8   : I32EnumAttrCase<"VSMFMA_I32_16x16x64_I8", 5>;
```
kuhar, line 377: I'd call these `VDMFMA_T32_8x16x...` -- I think what we care about more is what the actual dense size is, and the sparse expansion is more of an implementation detail.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_contract_amdgpu.mlir
```diff
@@ -1149,3 +1149,199 @@ builtin.module attributes { transform.with_named_sequence } {
...
+// CHECK:       %[[AND:.+]] = arith.andi %[[LANE_ID]]
+// CHECK:       %[[IS_ODD:.+]] = arith.cmpi ne, %[[AND]]
+//
+// CHECK:       %[[EVEN_A:.+]] = vector.shuffle %[[A_CAST]], %[[A_CAST]] [0, 1, 4, 5] : vector<8xf16>, vector<8xf16>
+// CHECK:       %[[ODD_A:.+]] = vector.shuffle %[[A_CAST]], %[[A_CAST]] [2, 3, 6, 7] : vector<8xf16>, vector<8xf16>
+// CHECK:       %[[SPARSE_A:.+]] = arith.select %[[IS_ODD]], %[[ODD_A]], %[[EVEN_A]] : vector<4xf16>
+//
+// CHECK:       %[[SPARSE_IDX:.+]] = arith.select %[[IS_ODD]], %{{.+}}, %{{.+}} : vector<4xi8>
+//
+// CHECK:       %[[SMFMA:.+]] = amdgpu.sparse_mfma 16x16x32 %[[SPARSE_A]] * %[[B_CAST]] + %[[ACC_EXP]] sparse(%[[SPARSE_IDX]] : vector<4xi8>) : vector<4xf16>, vector<8xf16>, vector<4xf32>
```
kuhar, line 1247: In this expansion, the data in neighboring lanes is duplicated, no?

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUEnums.td
efric: Fair point

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_contract_amdgpu.mlir
efric: Yes that's right, it's duplicated and then we select the portion we want depending on lane parity.

kuhar: @krzysz00 can we do something to get rid of these shuffles through a different single_subgroup_layout?

kuhar:
> It would be good to understand where this work aligns. My understanding would be that this work would align more when the skinny dimension is something between 4-16, but not < 4, as in the < 4 case, you are memory bound, using more compute doesn't help.

we may also be able to use it for shapes that are a multiple of 8 but not 16 -- on paper, sparse mfma does the same amount of math per cycle as dense mfma, so it all boils down to the packing/swizzling overhead compared to masking / padding

efric:
Post meeting notes with @kuhar, @krzysz00, @qedawkins. 

Initial optimization step is folding two smfmacs into one VDMFMA to get unique 128b loads from neighboring lanes and also halve expand/collapse pattern. 

@qedawkins floated ideas regarding performing the reduction after the K loop but it's quite tricky, so current plan is the proceed with the naive optimization first.

PR #23517, opened 2026-02-19
[VectorExt] Add transfer_gather unrolling

Groverkss:
Unroll transfer_gather ops from dim 0 inward until rank 1, keeping
the innermost vector dimension for vectorized loads. After unrolling,
existing canonicalizations (FoldSingleElementIndexVec, FoldContigous-
GatherToTransferRead) convert symbol-free gathers to transfer_read.

Also add FoldIndexVecAddBroadcast canonicalization to absorb
broadcast(scalar) addends from index vectors into base offsets.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
@@ -1,0 +1,240 @@
+// Copyright 2025 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.h"
+#include "iree/compiler/Codegen/Dialect/VectorExt/Transforms/Transforms.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/PatternMatch.h"
+
+using namespace mlir;
+using namespace mlir::iree_compiler::IREE::VectorExt;
+
+namespace {
+
+/// Remove dim 0 from an AffineMap by:
+/// 1. Replacing AffineDimExpr(0) with AffineConstantExpr(0)
+/// 2. Renumbering AffineDimExpr(k) where k > 0 to AffineDimExpr(k-1)
+/// 3. Reducing numDims by 1
+static AffineMap removeDim0FromMap(AffineMap map) {
+  MLIRContext *ctx = map.getContext();
+  SmallVector<AffineExpr> newResults;
+  for (AffineExpr expr : map.getResults()) {
+    if (auto dimExpr = dyn_cast<AffineDimExpr>(expr)) {
+      unsigned pos = dimExpr.getPosition();
+      if (pos == 0) {
+        newResults.push_back(getAffineConstantExpr(0, ctx));
+      } else {
+        newResults.push_back(getAffineDimExpr(pos - 1, ctx));
+      }
+    } else {
+      newResults.push_back(expr);
+    }
+  }
+  return AffineMap::get(map.getNumDims() - 1, map.getNumSymbols(), newResults,
+                        ctx);
+}
+
+/// Remove dim 0 references from an index vec map. Returns the new map with
+/// results that referenced dim 0 dropped, and the axis positions in the index
+/// vec that need to be sliced.
+static AffineMap removeDim0FromIndexVecMap(AffineMap map,
+                                           SmallVectorImpl<int64_t> &axes) {
+  MLIRContext *ctx = map.getContext();
+  SmallVector<AffineExpr> newResults;
+  for (auto [resultIdx, expr] : llvm::enumerate(map.getResults())) {
+    if (auto dimExpr = dyn_cast<AffineDimExpr>(expr)) {
+      unsigned pos = dimExpr.getPosition();
+      if (pos == 0) {
+        axes.push_back(resultIdx);
+        continue; // Drop this result.
+      }
+      newResults.push_back(getAffineDimExpr(pos - 1, ctx));
+    } else {
+      newResults.push_back(expr);
+    }
+  }
+  return AffineMap::get(map.getNumDims() - 1, map.getNumSymbols(), newResults,
+                        ctx);
+}
+
+/// Extract a slice from a vector at position `idx` along the given `axis`.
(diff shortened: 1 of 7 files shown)


Groverkss:
This is stacked on top of https://github.com/iree-org/iree/pull/23510 , please review the top commit only.

kuhar:
@amd-eochoalo can you help review?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
```diff
@@ -0,0 +1,240 @@
...
+  return vector::ShapeCastOp::create(b, loc, newType, slice);
+}
+
+//===----------------------------------------------------------------------===//
+// UnrollTransferGatherDim
+//===----------------------------------------------------------------------===//
+
+/// Unrolls dim 0 of a transfer_gather, reducing vector rank by 1 each
+/// application. Stops at rank 1.
+struct UnrollTransferGatherDim : public OpRewritePattern<TransferGatherOp> {
```
amd-eochoalo, line 103: nit: Consider using MaskableOpRewritePattern.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/vector_lowering.mlir
```diff
@@ -222,3 +222,27 @@ func.func @int_contract_add_no_chain_fma(
...
+  %indices: vector<4xindex>) -> vector<4x64xf16> {
+  %cst = arith.constant 0.0 : f16
+  %c0 = arith.constant 0 : index
+  %out = iree_vector_ext.transfer_gather %source[%c0, %c0]
+  [%indices : vector<4xindex>], %cst {
+    indexing_maps = [affine_map<(d0, d1)[s0] -> (s0, d1)>,
+                     affine_map<(d0, d1)[s0] -> (d0)>]
+  } : memref<4096x64xf16>, vector<4x64xf16>
+  return %out : vector<4x64xf16>
+}
```
amd-eochoalo, line 242: I think more tests are needed

1. masked
2. something that exercises the extract_strided_slice path in extractVecSlice.
3. a test with more dimensions where some index vectors don't reference d0.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
```diff
@@ -0,0 +1,240 @@
...
+    for (AffineMap &m : newIndexVecMaps) {
+      newAllMaps.push_back(m);
+    }
+    if (mask) {
+      newAllMaps.push_back(newMaskMap);
+    }
+
+    // Initialize accumulator.
+    Value acc = arith::ConstantOp::create(rewriter, loc, resultType,
+                                          rewriter.getZeroAttr(resultType));
```
amd-eochoalo: Why not poison?

amd-eochoalo (changes requested):
I think adding a couple of more tests would be good. Some minor questions too.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+  if (rank == 1) {
+    // Extracting from rank-1 along axis 0 gives a scalar.
+    return vector::ExtractOp::create(b, loc, vec, SmallVector<int64_t>{idx});
+  }
+
+  if (axis == 0) {
+    // vector.extract %vec[idx] drops the leading dim.
+    return vector::ExtractOp::create(b, loc, vec, SmallVector<int64_t>{idx});
+  }
```
amd-eochoalo: Maybe consider merging this into a single branch.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
Groverkss: if a transfer_gather op has a vector.mask around it, we are doing something wrong. We shouldn't have it. The op has a mask operand.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/vector_lowering.mlir
Groverkss: I want to add more tests for 3. , but I want to wait until I have support for lowering it to vector.gather/vector.scatter . Once I have that, I can add support for that case.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
Groverkss: good catch, thanks

Groverkss: actually, it should be the padding to use.

Groverkss: hmm, actually it should just be poison, because the padding would always get written to anyway by the subops.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
Groverkss: i guess if rank == 1, then axis == 0 will always be true.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherOps.cpp
amd-eochoalo: transfer_gather has MaskableOpInterface. I think since it can be represented at this stage in the compiler, this code should at least avoid matching it if it is masked.

amd-eochoalo: ~~But also, aren't you getting Value mask = op.getMask here?~~ Nevermind this, I understood what you said above. However, I think it should still avoid being successfully matched if there is a mask.

Groverkss: I think i'll just remove MaskableOpInterface from transfer_gather, it's an anti-pattern to have it, the only reason i added it earlier was to match transfer_read/transfer_gather but it's a complete anti-pattern.

Groverkss: I'll send another pr that removes it, can you approve this one and I can send another to remove maskable op interface.

amd-eochoalo (approved):

PR #23518, opened 2026-02-19
[Codegen][GPU] Improve CombineBarrierRegionsPass

Groverkss:
Before this patch, combine barrier regions pass would only look for adjacent barrier regions to combine. This was restrictive and didn't always work for more complex things. The CombineValueBarrierOps pass already had logic to combine ops even when there were other ops in between.

This patch combines both the passes, sharing the logic of moving operations between the two barrier ops, combining them into one.

This pass adds both the barrier combining transformations into one pass, and renames it to GPUCombineValueSemanticBarriersPass.


compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_combine_value_semantic_barriers.mlir
```diff
@@ -0,0 +1,259 @@
...
+
+// CHECK-LABEL: func.func @barrier_diamond_chain
+// CHECK-COUNT-3: value_barrier
+// CHECK-NOT: value_barrier
+
+// -----
+
+// barrier_region tests
+
+func.func @combine_barrier_region(%arg0: tensor<6xf32>, %arg1: tensor<7xf32>) -> (tensor<1xf32>, tensor<2xf32>) {
```
nirvedhmeshram, line 133: nit: I dont think this test is adding anything over the three region test below?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_combine_value_semantic_barriers.mlir
```diff
@@ -0,0 +1,259 @@
...
+
+// -----
+
+func.func @dont_combine_dependent_barriers(%arg0: tensor<6xf32>) -> (tensor<1xf32>, tensor<1xf32>) {
+  %0 = iree_gpu.barrier_region ins(%arg0 : tensor<6xf32>) {
+  ^bb0(%intermediate: tensor<6xf32>):
+    %slice = tensor.extract_slice %intermediate[1] [1] [1] : tensor<6xf32> to tensor<1xf32>
+    iree_gpu.yield %slice : tensor<1xf32>
+  } : tensor<1xf32>
+  %1 = iree_gpu.barrier_region ins(%0 : tensor<1xf32>) {
```
nirvedhmeshram, line 203: can we add a dummy op chain to this test in the middle that is consumer/producer to  the barrier regions.

nirvedhmeshram (approved):
LGTM, i noticed my nits are in existing tests anyway so optional if you want to address those.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_combine_value_semantic_barriers.mlir
Groverkss: These are tests carried over from the previous implementation.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_combine_value_semantic_barriers.mlir
Groverkss: I thought about this, but it should be handled properly because of how backward slice works. (I'm being a bit lazy here).

PR #23519, opened 2026-02-19
[Codegen][GPU] Move hoisting to interface and add it for barrier ops

Groverkss:
iree_gpu.barrier_region ops prevent hoisting and only resolve after bufferization. This means that before bufferization, we effectively cannot do any hoisting from loops because they are usually in a barrier region. After bufferization, it's extremely hard to figure out if we can do hoisting because of side effects.

This leads to better codegen in general in TileAndFuse pipeline, generating better barriers in general, because the barrier region is much more restricted. This was manually verified.

The test changes in igemm pipeline are mainly the test being wrong earlier, it was matching some private memory load, instead of the shared memory loads. There was bad barrier placement before this patch because the barrier regions were capturing everything, leading to an extra barrier which got matched.

The patch also moves the hoisting implementation to an interface. This is required, otherwise all codegen patterns start depending on iree gpu dialect, which shouldn't be the case for non gpu code.


Groverkss:
Depends on https://github.com/iree-org/iree/pull/23518 , please only review the top commit

krzysz00 (approved):
lgtm here

nirvedhmeshram (approved):

compiler/src/iree/compiler/Codegen/Interfaces/Interfaces.cpp
```diff
@@ -46,6 +49,35 @@
...
+// linalg.generic operations are also loop-like, but they don't have
+// LoopLikeOpInterface implemented for them.
+// Ideally, we should be checking if the linalg.generic op has a trip
+// count of zero, but while that is possible and can be written using
+// ValueBoundsConstraintSet, it is usually not needed. Unlike loops,
+// which can have arbitrary operations inside them, the loop invariant
+// operations inside a linalg.generic operation are usually
+// operations performed on scalars. Hoisting scalar constants does not
+// have a big cost even if the trip count is zero.
+struct LinalgGenericHoistableRegionModel final
```
krzysz00, line 63: Could this be implemented for all linalg ops that have regions? I vaguely recall there being some other intermediate ones. Not critical for this PR.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_igemm_tile_and_fuse.mlir
```diff
@@ -151,8 +151,8 @@ hal.executable private @main {
 //          CHECK:   scf.forall ({{.*}}) in (17, 1, 81) {
 //          CHECK:     %[[LOOP:.+]] = scf.for {{.+}} = %[[C0]] to %[[C721]] step %[[C1]] {{.*}} -> (vector<1x1x1x1x4x1xf32>)
 //          CHECK:       gpu.barrier memfence [#gpu.address_space<workgroup>]
-//      CHECK-DAG:       vector.transfer_read {{.*}} vector<4xf16>
-//      CHECK-DAG:       vector.transfer_read {{.*}} vector<4xf16>
+//          CHECK:       vector.transfer_read {{.*}} vector<4xf16>
+//          CHECK:       vector.transfer_read {{.*}} vector<4x1x1xf16>
```
krzysz00, line 155: This change seems unrelated?

krzysz00 (commented):
Interface lgtm, curious about where those extra unit dimensions came from and if they'll break any pattern matches later down the line

hanhanW (approved):
Partially Approve. The move of interface looks reasonable to me, thanks for the ping on reviews!

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_igemm_tile_and_fuse.mlir
Groverkss: From pr description:

> The test changes in igemm pipeline are mainly the test being wrong earlier, it was matching some private memory load, instead of the shared memory loads. There was bad barrier placement before this patch because the barrier regions were capturing everything, leading to an extra barrier which got matched.

compiler/src/iree/compiler/Codegen/Interfaces/Interfaces.cpp
Groverkss: Not really. linalg named ops have a region, but the region is really fake a shouldn't be there.

PR #23520, opened 2026-02-19
Integrate llvm/llvm-project@f56efa2

efric:


Muzammiluddin-Syed-ECE (approved):

amd-eochoalo:
Hey, one thing to note is that there was a discussion about the naming convention for the branch on discord a while back. (Discussion starts [here](https://discord.com/channels/689900678990135345/1080178290188374049/1464336120887115776)) I think that the naming convention is technically sm-iree-integrates/llvm-YYYYMMDD as all branches prefixed with sm-iree are protected. Not a big deal, but some tools which other people use depend on this naming convention. See the playbook on LLVM integrates for more detail (I don't use those tools).

PR #23528, opened 2026-02-20
`iree/tokenizer/`: Streaming tokenizer with full HuggingFace compatibility

benvanik:
For the humans, from the human: this is a few weeks of deep feature branch work with a team of @claude's. It's gone through dozens of review cycles with mixed model teams and had quite a bit of stress testing (including extensive fuzzing, benchmarking against other libraries, and _a lot_ of compatibility testing). v1 was me learning what a tokenizer was and all the edge cases, and v2 is the proper implementation. I'm proud that this is now the most compatible and most performant tokenizer (and the only streaming one with training support) while also remaining absolutely tiny (KB of RAM for infinite streaming) and self-contained (zero dependencies). Future changes will add additional file formats as well as our own optimized for mmap and precompilation, and keeping up with new tokenizer changes is fairly easy (everything is abstracted and changes to any particular stage are independent). For a motivated embedder, it's possible to LTO out all of the code not required for a particular tokenizer configuration (by default, the JSON loader naturally needs to link them all in), which gives similar footprint to a hand-rolled tokenizer library while not sacrificing any features. Neat.

Note, this is single-threaded (intended to be run in request/response handlers): all the modern rust implementations seem to focus on complex multi-threading setups and still cannot match the numbers here. Turns out, optimizing for L1 is the right approach.

---

Replaces the experimental tokenizer with a production implementation. The previous version validated the architecture — normalize, segment, encode, decode as a streaming pipeline — but was designed for batch operation with streaming bolted on afterward. This is a ground-up rewrite where streaming is the foundation, not a feature.

### Why

Streaming through an architecture designed for batching does not work. v1 could produce output incrementally, but the guarantees were wrong: partial UTF-8 characters leaked between stages, memory usage scaled with input size, and edge cases in BPE merge interactions silently produced incorrect tokens on vocabularies we were not testing against. These are not fixable with patches — the streaming model has to be structural.

The other discovery was that nearly every tokenizer implementation we tested against had differences and outright bugs — and most do not even support the full configuration surface. The HuggingFace Rust tokenizer, the most widely deployed implementation, falls back to Python for a number of models. Other implementations support a subset of normalizers, a subset of pre-tokenizers, or only BPE. We picked HuggingFace Tokenizers (Rust + Python) as our golden reference because it is the most used, but "matches HuggingFace output on the top 170 models by downloads this ~year" required implementing capabilities that most tokenizer libraries simply do not have.

### Streaming architecture

The pipeline has four stages — normalization, segmentation, encoding, decoding — each processing incrementally with fixed-size state.

Normalizers tile input into 64-byte chunks processed through ping-pong scratch buffers. Each normalizer's output contract guarantees UTF-8 codepoint boundary alignment: a normalizer may buffer internally, but every byte sequence it emits is complete UTF-8. This contract propagates through the pipeline — segmenters receive valid UTF-8, encoders receive complete segments, decoders receive valid token sequences. No stage needs to handle partial characters from upstream.

Encode state is ~1-2KB, decode state is ~1KB, regardless of input size. Run-length encoded offset tracking maps output token positions back to source byte positions through the full pipeline without per-token allocation. Offset maps are available for both encode and decode — a requirement for training data preparation that many tokenizer implementations omit.

Both encode and decode stream: encode produces token IDs incrementally as input arrives, decode produces text incrementally as token IDs arrive. This matters for serving pipelines where tokens flow between stages — speculative decoding, streaming generation, pipeline-parallel inference — and blocking on complete sequences adds latency that the tokenizer should not be responsible for.

### What's here

**Encoding algorithms** (`model/`): BPE with backtracking and merge reachability analysis — the merge heap processes pairs by priority, and a backtracking pass resolves cases where greedy longest-match produces tokens whose suffixes are blocked by context-dependent merge interactions. Handles merge reachability (unreachable decomposition paths filtered during table construction), suffix blocking (boundary merges consuming prefix operands), and compound preemption (multi-step merge chains consuming tokens that appeared available). WordPiece with greedy longest-match for the BERT family. Unigram with Viterbi decoding for T5 and XLM-RoBERTa.

**Normalizers** (`normalizer/`): NFC, NFD, NFKD Unicode normalization. Lowercase, strip, strip accents. BERT normalization (accent stripping + lowercasing as a single pass). HuggingFace precompiled normalization tables. String replace with multi-byte UTF-8 safety across processing tile boundaries. Regex replace using the built-in DFA engine. Prepend (SentencePiece space prefix). Sequence chaining with streaming passthrough. 25 test and fuzz targets.

**Segmenters** (`segmenter/`): BERT (whitespace + punctuation + CJK splitting), standalone whitespace and punctuation splitting, metaspace (SentencePiece encoding), digit grouping, generic pattern split, regex split, sequence chaining. 14 test and fuzz targets.

**Decoders** (`decoder/`): Byte-level encoding reversal, byte fallback for unknown tokens, metaspace decoding, WordPiece `##` stripping, CTC blank token removal, replace, strip, sequence chaining. 16 test and fuzz targets.

**Regex engine** (`regex/`): Self-contained DFA regex implementation — no external dependency. Patterns compile to NFA then DFA at tokenizer load time. O(n) matching with no backtracking, safe for untrusted patterns. Used by regex replace normalizers, split segmenters, and regex decoders.

**Vocabulary** (`vocab/`): Token ID ↔ string mapping with hash table lookup and longest-prefix-match trie. Separate merge pair hash for BPE. Incremental construction for streaming model loading.

**HuggingFace loader** (`format/huggingface/`): Complete `tokenizer.json` parser covering the full configuration surface — model parameters, normalizer chains, pre-tokenizer chains, decoder chains, post-processor templates, added tokens with special token flags. 9 test and fuzz targets. More loaders will be added in the future.

### Validation

Tested against 170 HuggingFace models with 1,667 encode/decode golden tests — GPT-2 through DeepSeek, BERT through XLM-RoBERTa, Whisper, T5, CodeLlama, vision models, multilingual variants. All passing. Golden expected values are generated by HuggingFace Tokenizers (Rust) as the reference implementation.

43 fuzz targets cover every parser (JSON loader, regex compiler, vocab builder) and every processing stage (each normalizer, segmenter, encoder, decoder). Fuzz dictionaries seed the corpus with format-specific tokens.

### Performance

Faster than HuggingFace Rust tokenizer, rust-tokenizers, rust-bert, tiktoken-rs, and llama.cpp's tokenizer, while supporting features none provide in their totality: streaming encode and decode, source byte offset tracking through the full pipeline, and the complete HuggingFace normalizer/pre-tokenizer/decoder configuration surface. The performance follows from architectural decisions — cache-friendly tiled processing, bounded state, no per-token allocation, DFA regex — rather than post-hoc optimization of a batch design.

Benchmark infrastructure (`tools/comprehensive_benchmark.cc`) measures throughput across model families and input sizes.

**Performance — GPT-2 r50k_base (50K vocab, encode MiB/s):**

| | IREE (C) | Rust bpe | HF (Rust) | tiktoken-rs |
|---|---|---|---|---|
| ASCII | **44.7** | 39.9 | 4.0 | 15.0 |
| CJK | **37.9** | 37.6 | 4.2 | 15.4 |
| Code | **52.1** | 30.6 | 3.9 | 12.7 |

**Performance — GPT-2 r50k_base (50K vocab, decode MiB/s):**

| | IREE (C) | Rust bpe | HF (Rust) | tiktoken-rs |
|---|---|---|---|---|
| ASCII | **655** | 481 | 47.6 | 310 |
| CJK | **1,558** | 320 | 30.5 | 231 |
| Code | **2,428** | 490 | 36.4 | 338 |

**Performance — LLaMA 3.1 (128K vocab, encode MiB/s):**

The Rust `bpe` crate cannot load LLaMA 3's vocabulary (its rank-based invariant check fails on LLaMA 3's 280K merge rules for 128K tokens).

| | IREE (C) | HF (Rust) | Speedup |
|---|---|---|---|
| ASCII | **42.8** | 4.4 | **9.7x** |
| CJK | **23.0** | 5.8 | **4.0x** |
| Code | **57.1** | 4.5 | **12.7x** |

**Performance — LLaMA 3.1 (128K vocab, decode MiB/s):**

| | IREE (C) | HF (Rust) | Speedup |
|---|---|---|---|
| ASCII | **655** | 46.9 | **14.0x** |
| CJK | **1,558** | 44.6 | **34.9x** |
| Code | **2,428** | 49.4 | **49.1x** |

### Scale

~39K lines implementation, ~46K lines test, ~8K lines fuzz, ~4.5K lines benchmark.

ci-extra: all


stellaraccident (approved):
I have done an AI-assited deep dive on this code and endorse it as well constructed with a comprehensive test suite that surpasses the norm for this class of infrastructure. Spot checking key abstractions and coding choices reveal no problems worth blocking on.

krzysz00:
1. Hold on, do we *really* need our own Unicode handling library?
2. Well, if it passes fuzzers that's probably close enough to correct.
3. While I can hesitantly endorse not having a good handle on the code, how confident are you, personally, that the fuzzer that validated this work is correct?

krzysz00:
Oh, and
4. Do you have those benchmark numbers lying around somewhere in a form where you can publish them?

benvanik:
Poked the team to refresh their numbers and decided to do some more micro-optimization :)
Performance report from the agents: https://gist.github.com/benvanik/9090b155a3102b333896c1d92c451cbc

For those not familiar with how complex general-purpose tokenizers can get, here's some background:

### Code size comparison: IREE tokenizer vs HuggingFace tokenizers

**Direct LOC** (implementation + headers, excluding tests/tools/benchmarks):

| | IREE | HuggingFace |
|---|---|---|
| Core implementation | 44.9k C | 15k Rust (23k with inline tests) |

IREE is ~3x larger by direct LOC. But HuggingFace externalizes substantial functionality into crate dependencies:

| Dependency | What it does | Approximate LOC |
|---|---|---|
| `regex` + `regex-syntax` | Regular expression engine | ~30-40k |
| `fancy-regex` / `onig` | Backref-capable regex (optional) | ~5-15k |
| `aho-corasick` | Multi-pattern string matching | ~10k |
| `unicode-normalization-alignments` | NFD/NFC/NFKD/NFKC with alignment | ~5-8k |
| `unicode-segmentation` | Grapheme/word/sentence boundaries | ~3-5k |
| `unicode_categories` | Unicode category lookups | ~1-2k |
| `serde` + `serde_json` | JSON deserialization | ~20k+ |
| `spm_precompiled` | SentencePiece precompiled regex | ~1-2k |
| `esaxx-rs` | Suffix array (unigram training) | ~1-2k |
| `ahash` / `dary_heap` / `compact_str` | Data structures | ~5-10k |
| `rayon` | Parallel iteration | ~10k |
| **Estimated transitive total** | | **~90-130k** |

Including transitive dependencies, the code that actually executes when HuggingFace tokenizes is likely **105-155k lines** — 2-3x larger than IREE's 45k self-contained C implementation. This doesn't include the python side of tokenizers which is used for certain models that the Rust library does not support.

IREE implements all of this in-tree with zero external dependencies: regex engine, Unicode normalization with alignment tracking, multi-pattern matching, JSON parsing, BPE/WordPiece/Unigram models, and streaming encode/decode. This is part of why it's fast — no abstraction boundaries between tokenizer stages means the compiler sees everything, and purpose-built data structures avoid the generality tax of library code.

### Performance summary

**Performance — GPT-2 r50k_base (50K vocab, encode MiB/s):**

| | IREE (C) | Rust bpe | HF (Rust) | tiktoken-rs |
|---|---|---|---|---|
| ASCII | **44.7** | 39.9 | 4.0 | 15.0 |
| CJK | **37.9** | 37.6 | 4.2 | 15.4 |
| Code | **52.1** | 30.6 | 3.9 | 12.7 |

**Performance — GPT-2 r50k_base (50K vocab, decode MiB/s):**

| | IREE (C) | Rust bpe | HF (Rust) | tiktoken-rs |
|---|---|---|---|---|
| ASCII | **655** | 481 | 47.6 | 310 |
| CJK | **1,558** | 320 | 30.5 | 231 |
| Code | **2,428** | 490 | 36.4 | 338 |

**Performance — LLaMA 3.1 (128K vocab, encode MiB/s):**

The Rust `bpe` crate cannot load LLaMA 3's vocabulary (its rank-based invariant check fails on LLaMA 3's 280K merge rules for 128K tokens).

| | IREE (C) | HF (Rust) | Speedup |
|---|---|---|---|
| ASCII | **42.8** | 4.4 | **9.7x** |
| CJK | **23.0** | 5.8 | **4.0x** |
| Code | **57.1** | 4.5 | **12.7x** |

**Performance — LLaMA 3.1 (128K vocab, decode MiB/s):**

| | IREE (C) | HF (Rust) | Speedup |
|---|---|---|---|
| ASCII | **655** | 46.9 | **14.0x** |
| CJK | **1,558** | 44.6 | **34.9x** |
| Code | **2,428** | 49.4 | **49.1x** |

IREE leads on all GPT-2 corpora, is 1.1-1.7x faster than the best Rust implementation on encode, 1.4-5.0x on decode, and 4-49x faster than HuggingFace's Rust backend — in ~1/3 the transitive code, with no dependencies, streaming support, and bounded memory usage.

PR #23530, opened 2026-02-20
[VectorDistribute] Simplify DistributeStep pattern

Groverkss:
Rewrite the DistributeStep pattern to preserve contiguity on the element tile. The new implementation works in 2D [B*O, E] shape, keeping vector.step for the inner element tile so that downstream passes can infer contiguity for gather-like loads.

This replaces the previous ~185-line implementation (DimInfo struct, getRemainingDims, getLens, getPackedStrides, generateSlicedStep, and the getStrides helper) with a direct ~85-line version.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -1790,152 +1790,17 @@ struct DistributeBatchOuterToLayoutConversions final
   }
 };
 
-// This is a helper to extract strides from a given shape
-// E.g. : a shape of 2x3x4 will return strides [12, 4, 1]
-static SmallVector<int64_t> getStrides(ArrayRef<int64_t> shape) {
-  int64_t elementCount = ShapedType::getNumElements(shape);
-  SmallVector<int64_t> strides;
-  int64_t currStride = elementCount;
-  for (int64_t len : shape) {
-    currStride = currStride / len;
-    strides.push_back(currStride);
-  }
-  return strides;
-}
-
+// Distributes vector.step across threads/subgroups according to a nested
+// layout. For packed shape [S, B, O, T, E] with row-major strides, each
+// thread gets a vector of B*O*E offsets computed from the batch/outer/element
+// positions, plus runtime contributions from subgroup and thread indices.
 struct DistributeStep final : OpDistributionPattern<vector::StepOp> {
   using OpDistributionPattern::OpDistributionPattern;
-
-  // This is a helper aggregate
-  // to hold the information about
-  // a dimension.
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1955,34 +1820,68 @@ struct DistributeStep final : OpDistributionPattern<vector::StepOp> {
...
+    int64_t T = packed[3], E = packed[4];
+
+    // Row-major strides of [S, B, O, T, E].
+    int64_t sgStride = B * O * T * E;
+    int64_t batchStride = O * T * E;
+    int64_t outerStride = T * E;
+    int64_t threadStride = E;
+
+    int64_t BO = B * O;
+    auto indexType = rewriter.getIndexType();
```
kuhar: nit
```suggestion
    Type indexType = rewriter.getIndexType();
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1955,34 +1820,68 @@ struct DistributeStep final : OpDistributionPattern<vector::StepOp> {
...
+    // Combine: base = outer_offsets + element_step.
+    Value val = arith::AddIOp::create(rewriter, loc, outerBcast, stepBcast);
+
+    // Add runtime subgroup and thread contributions.
+    struct IdStride {
+      Value id;
+      int64_t stride;
+    };
+    for (auto [id, stride] : {IdStride{subgroupIndices[0], sgStride},
+                              IdStride{threadIndices[0], threadStride}}) {
```
kuhar: why not pair if you are not using the named fields anyway?

```suggestion
    for (auto [id, stride] : {std::pair{subgroupIndices[0], sgStride},
                              std::pair{threadIndices[0], threadStride}}) {
```

kuhar (approved):

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: Will add it, but the reason i didn't expand it was it was mostly clear from getIndexType()

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: fixed

Groverkss:
The failures look unrelated to the pr

PR #23531, opened 2026-02-20
Integrate llvm/llvm-project@c6cb8f3

efric:
Other points of interest:

- https://github.com/llvm/llvm-project/commit/b82c7fc65229 adds PCH infrastructure to LLVM with LLVMSupport PCH reuse. External projects (e.g., StableHLO) live in a separate binary dir from LLVM, so CMake cannot resolve the LLVMSupport PCH path for targets that try to reuse it. Fix by disabling PCH in the scope of iree_llvm_add_external_project(); LLVM/MLIR's own PCH is unaffected.

### DIFF
--- a/build_tools/cmake/iree_llvm.cmake
+++ b/build_tools/cmake/iree_llvm.cmake
@@ -301,6 +301,12 @@ function(iree_llvm_add_external_project name location)
     endif()
   endif()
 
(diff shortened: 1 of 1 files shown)


kuhar (commented):
PCH change looks reasonable -- we don't have that much code built as external projects anyway

kuhar (approved):

PR #23533, opened 2026-02-20
[Codegen] Move tuner attr flag to codegen options

kuhar:
Do not rely on an inline variable for flags. This follows how we handle other flags shared across GPU and CPU
pipelines.

Rename the flag but add alias with the old spelling. The new flag name is `--iree-codegen-add-tuner-attributes`.

Assisted-by: claude


bangtianliu (approved):
Thanks! I'll update the tuner README (also boo/fusilli tuner) on my end to reflect the new flag name.

Issue #23535, opened 2026-02-20
[Tuner] Compiler-side constraint generation


kuhar:
Tracking issue for https://github.com/iree-org/iree/discussions/23521

The completion criteria is to move all constraint generation from amdsharktuner to the iree compiler.

Tasks:

- [x] https://github.com/iree-org/iree/pull/23533 @kuhar
- [x] https://github.com/iree-org/iree/pull/23536 @kuhar
- [x] https://github.com/iree-org/iree/pull/23687 @kuhar
- [x] Add a hook to query a list of constraints from a module @RattataKing
   * https://github.com/llvm/llvm-project/pull/186131
   * https://github.com/iree-org/iree/pull/23832
- [x] Add a pass to convert an iree_codegen.constraints op to smt module @RattataKing
   * https://github.com/iree-org/iree/pull/23801/
- [x] Add a hook to convert a constraints op to smt and serialize as an smtlib string @RattataKing 
   * https://github.com/llvm/llvm-project/pull/187366
   * https://github.com/iree-org/iree/pull/23840
- [x] Support custom knob attrs / ops @kuhar
   * https://github.com/iree-org/iree/pull/23742
   * https://github.com/iree-org/iree/pull/23743
   * https://github.com/iree-org/iree/pull/23780
- [ ] Convert model form z3 + knobs ==> compilation_info attr @RattataKing  
   * https://github.com/nod-ai/amd-shark-ai/pull/2896
- [x] Add pipeline interface to emit constraint ops @kuhar 
   * https://github.com/iree-org/iree/pull/23795
   * https://github.com/iree-org/iree/pull/23816
   * https://github.com/iree-org/iree/pull/23851
   * https://github.com/iree-org/iree/pull/23864
- [x] Add lowering_config verification based on constraint ops
   * https://github.com/iree-org/iree/pull/23878
- [ ] Port pipeline constraints to the compiler side @RattataKing

kuhar:
All the work has landed and we switched amdsharktuner to use the new compiler-side constraints in https://github.com/nod-ai/amd-shark-ai/pull/2915

PR #23536, opened 2026-02-20
[Tuner][Codegen] Add set identifier to root op attributes

kuhar:
This is so that we can tell which root ops needs to be configured together. At the moment, the codegen still defaults to setting only one root op set 0, but we can be more fine-grained in the future.

Also add bindings.

Assisted-by: claude


compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
```diff
@@ -685,4 +685,21 @@ def WorkgroupScopeAttr : AttrDef<IREECodegen_Dialect, "WorkgroupScope", []> {
...
+  let mnemonic = "root_op";
+  let summary = "Marks an operation as a tuner root op in a given set";
+  let description = [{
+    Attribute identifying a root operation. The `set` parameter groups root ops
+    into numbered sets: all root ops in the given set share the same
+    `lowering_config`. Codegen does not rely on the `root_op` attribute; it is
+    only used for constraint generation when tuning.
+  }];
+  let parameters = (ins "int64_t":$set);
+  let assemblyFormat = "`<` `set` `=` $set `>`";
```
bangtianliu, line 702: My understanding is that the $set should not be negative numbers, should we add validation?

compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
```diff
@@ -67,6 +67,15 @@ MLIR_CAPI_EXPORTED MlirAttribute ireeCodegenCompilationInfoAttrGet(
...
 ireeCodegenCompilationInfoAttrGetParameters(MlirAttribute attr);
 
+MLIR_CAPI_EXPORTED bool ireeAttributeIsACodegenRootOpAttr(MlirAttribute attr);
+
+MLIR_CAPI_EXPORTED MlirTypeID ireeCodegenRootOpAttrGetTypeID(void);
+
+MLIR_CAPI_EXPORTED MlirAttribute ireeCodegenRootOpAttrGet(MlirContext mlirCtx,
+                                                          int64_t set);
+
+MLIR_CAPI_EXPORTED int64_t ireeCodegenRootOpAttrGetSet(MlirAttribute attr);
```
bangtianliu, line 77: Consider `ireeCodegenRootOpAttrGetSet` to `ireeCodegenRootOpAttrGetSetValue` or `ireeCodegenRootOpAttrGetSetNumber`

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
kuhar: It can be any number, the exact values don't matter

compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
kuhar: This follows the standard naming conventions for attribute parameters

bangtianliu (approved):

PR #23541, opened 2026-02-21
[Python] Bump pip requirements

kuhar:
Set the minimum version to the newest versions that still support python 3.9.

Fix new lint warnings.


docs/.markdownlint.yml
```diff
@@ -102,3 +102,9 @@ table-pipe-style: false
 # Allow table rows to have irregular numbers of columns
 # (markdownlint has false positives for `|` characters in code blocks)
 table-column-count: false
+
+# Allow inconsistent table column spacing styles
+table-column-style: false
+
+# Allow non-descriptive link text (e.g. "[here]")
+descriptive-link-text: false
```
kuhar, line 110: I added these because otherwise would have to reformat 50+ files

krzysz00: How bad's the descriptive link lint?

krzysz00 (approved):
Looks NFC to me

docs/.markdownlint.yml
kuhar: I can open fresh PR to opt onto it and reformat

PR #23543, opened 2026-02-21
[GPU][LDS] Fix promote_operands/promotion_types size mismatch

Yu-Zhewen:
Otherwise, enabling LDS DMA on fused matmul+elementwise ops, where matmul C operand is promoted, triggers an assertion at

https://github.com/iree-org/iree/blob/c23abe741b847f1db2ba742b22b0485cc19bc633/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPULoweringConfigUtils.cpp#L110-L111

because operands has 3 entries (A, B and C) but promotion_types only has 2.


lialan (approved):
+1！

Issue #16956, opened 2024-04-03
Missing support in scf.for bufferization


hanhanW:
The issue if forked from https://github.com/openxla/iree/issues/16421. I had a long discussion with @matthias-springer about the bufferization issue. We are not able to bufferize scf.for ops when it returns three tensors. There is a write to `%arg4`, but then there is a read from the original `%arg4` later. Therefore, `%24` must bufferize out-of-place. Meaning a new copy of `%arg4`. Meaning that a different buffer (newly allocated) buffer is passed to the next iteration. A workaround is adding `%y2 = bufferization.materialize_in_destination %24 in %arg4 : (tensor<64xf32>, tensor<64xf32>) -> tensor<64xf32>` and yielding `%y2` instead of `%24`. The `bufferization.materialize_in_destination` op is not needed for other values. @matthias-springer is it something we can implement in the bufferization? If so, would you help on that?

To repro: `iree-opt --iree-codegen-llvmcpu-bufferization-pipeline ~/repro.mlir`

```mlir
func.func @main_dispatch_0_attention_20x4096x64xf16() {
  %c4096 = arith.constant 4096 : index
  %c20 = arith.constant 20 : index
  %c0 = arith.constant 0 : index
  %0 = hal.interface.binding.subspan set(0) binding(0) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : !flow.dispatch.tensor<readonly:tensor<20x4096x64xf16>>
  %1 = hal.interface.binding.subspan set(0) binding(1) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : !flow.dispatch.tensor<readonly:tensor<20x4096x64xf16>>
  %2 = hal.interface.binding.subspan set(0) binding(2) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : !flow.dispatch.tensor<readonly:tensor<20x4096x64xf16>>
  %3 = hal.interface.binding.subspan set(0) binding(3) type(storage_buffer) alignment(64) offset(%c0) : !flow.dispatch.tensor<writeonly:tensor<20x4096x64xf16>>
  %workgroup_id_x = hal.interface.workgroup.id[0] : index
  %workgroup_count_x = hal.interface.workgroup.count[0] : index
  %workgroup_id_y = hal.interface.workgroup.id[1] : index
  %workgroup_count_y = hal.interface.workgroup.count[1] : index
[...]

PR #23544, opened 2026-02-22
Reapply "[CPU] Support dynamic attention by tiling K1 when needed." 

### LINKED ISSUE #23277: Compiler crash in LLVMCPUSelectLoweringStrategy with dynamic-shape iree_linalg_ext.attention
### Description

The compiler crashes during the `LLVMCPUSelectLoweringStrategy` pass when compiling an `iree_linalg_ext.attention` op with fully dynamic 3D input shapes (`tensor<?x?x?xf32>`).

The crash occurs in `setRootConfig` for `AttentionOp` when calling `limitVectorTileSizes`.

### Reproducer

Save as `attention_crash.mlir`:

```mlir
module @attention_crash {
  util.func public @attention(
      %query: tensor<?x?x?xf32>,
      %key: tensor<?x?x?xf32>,
      %value: tensor<?x?x?xf32>,
      %scale: f32
  ) -> tensor<?x?x?xf32> {

    %batch = tensor.dim %query, %c0 : tensor<?x?x?xf32>
    %seq_len = tensor.dim %query, %c1 : tensor<?x?x?xf32>
    %head_dim = tensor.dim %query, %c2 : tensor<?x?x?xf32>

    %output_init = tensor.empty(%batch, %seq_len, %head_dim) : tensor<?x?x?xf32>
    %output = iree_linalg_ext.attention {
      indexing_maps = [
        affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2)>,
        affine_map<(d0, d1, d2, d3, d4) -> (d0, d3, d2)>,
        affine_map<(d0, d1, d2, d3, d4) -> (d0, d3, d4)>,
        affine_map<(d0, d1, d2, d3, d4) -> ()>,
        affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d4)>
      ]
    } ins(%query, %key, %value, %scale : tensor<?x?x?xf32>, tensor<?x?x?xf32>, tensor<?x?x?xf32>, f32)
      outs(%output_init : tensor<?x?x?xf32>) {
    ^bb0(%arg0: f32):
      iree_linalg_ext.yield %arg0 : f32
    } -> tensor<?x?x?xf32>

    util.return %output : tensor<?x?x?xf32>
  }
}
```

Compile with:
```bash
iree-compile attention_crash.mlir --iree-hal-target-backends=llvm-cpu -o /dev/null
```

### Stack trace (key frames)

```
#4 limitVectorTileSizes(...) KernelDispatch.cpp:747
#8 setRootConfig(..., AttentionOp) KernelDispatch.cpp:2344
#14 LLVMCPUSelectLoweringStrategyPass::runOnOperation() LLVMCPUSelectLoweringStrategy.cpp:238
```

### Environment

- IREE: pip package (current release)
- OS: Fedora 43, Linux 6.17
- Target: `llvm-cpu`

### Notes

Static shapes compile successfully. The crash appears to occur during tile size computation when shapes are unknown at compile time.

---

*This issue was authored by Claude (Opus 4.5) and reviewed by @stellaraccident.*

hanhanW:
The K1 dimension (head_dim) in attention was unconditionally left untiled, which leads large stack allocation when the dimension is dynamic.

K1 is typically small (64/128 per AttentionOpDetail docs), so the original heuristic to leave it untiled was reasonable. The revision sets the tile sizes if the dimension is dynamic or it is not within typical range (<= 128).

E2E tests are added, and they have the same inputs and expected outputs like attention.mlir (which is a static version). Some backends, e.g., AMDGPU, does not support dynamic attention, so we create a new file. The test is enabled on CPU and VMVX backends in the revision.

Fixes https://github.com/iree-org/iree/issues/23277

Previously, it triggers a bug on android build. The root cause is that vectorization is not enabled (because masking is not natively supported), and it leads to non-trivial buffer allocation. Although forcing the emulation is not ideal, but it enables the functionality. The performance has not been prioritized for at least two years, so we accept the emulation as a workaround. The evidence is that the e2e tests are not enabled for at least two years for ARM CPU.

https://github.com/iree-org/iree/blob/243fe33c21cc52ca9283c88460ad5524fb9f1b4a/tests/e2e/attention/CMakeLists.txt#L1-L5

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -2402,13 +2402,18 @@ static LogicalResult setRootConfig(mlir::FunctionOpInterface entryPointFn,
     vecTileSizeBounds[i] = distTileSizes[i] ? distTileSizes[i] : ubs[i];
   }
 
+  // K1 dimensions (head_dim) are typically small. Per AttentionOpDetail docs
+  // (IndexingUtils.h), K1 is generally 64 or 128. When K1 is static and within
+  // this typical range, we leave it untiled. However, for dynamic K1 or
+  // unusually large K1, we must tile to avoid unbounded allocations.
+  constexpr int64_t kTypicalK1Threshold = 128;
   SmallVector<int64_t> vecTileSizes(vecTileSizeBounds.size(), 1);
-  // Due to the way attention works, K1 dimensions cannot be tiled. Mark k1
-  // reduction dimensions not to distribute.
   for (int i : opInfo.getK1Dims()) {
-    vecTileSizes[i] = 0;
+    int64_t k1Size = ubs[i];
+    if (ShapedType::isStatic(k1Size) && k1Size <= kTypicalK1Threshold) {
+      vecTileSizes[i] = 0;
+    }
   }
-
   for (auto i : llvm::seq<int64_t>(0, vecTileSizeBounds.size())) {
     if (vecTileSizes[i] == 0) {
       continue;
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
@@ -146,6 +146,21 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   pipelineOpts.enableVectorMasking =
       isX86(targetConfig) || isRISCV(targetConfig) ||
       (isAArch64(targetConfig) && hasAnySVEFeature(targetConfig));
+  // TODO(#16956): The decomposition of attention op leads to complex control
+  // flow, which leads to non-trivial stack allocation. Enforcing masking for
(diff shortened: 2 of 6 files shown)


hanhanW:
cc @banach-space @egebeysel it is a pure improvement as it enables the e2e compilation and execution.

hanhanW:
https://github.com/iree-org/iree/pull/23318 removes the workaround, but there are more things to investigate. So I end up with having this workaround for now.

Groverkss:
I don't have context on the cpu side changes, so LGTM because I already approved the attention changes

hanhanW:
> I don't have context on the cpu side changes, so LGTM because I already approved the attention changes

Can you help approve the change? Thanks.

hanhanW:
> I don't have context on the cpu side changes, so LGTM because I already approved the attention changes

For more context, this is not CPU specific issue. This is the bufferization when you can't vectorize the ops after you convert it to OnlineAttention and lower them to loops like the old issue: https://github.com/iree-org/iree/issues/16956

It requires further analysis for bufferization part. We did not fix it in the past, and my other PR improves the analysis. However, I'll need further investigation for the change: https://github.com/iree-org/iree/pull/23318

Groverkss (approved):

PR #23550, opened 2026-02-23
Delete LinalgExt::PackOp and LinalgExt::UnPackOp

### LINKED ISSUE #20030: Retire LinalgExt::PackOp and LinalgExt::UnPackOp
The main reason of LinalgExt pack/unpack ops exist is that we need the memref version for backends that do not vectorize them on tensors. E.g., there are some pack/unpack ops that VMVX backend can not vectorize. The other example is that a downstream project needs pack/unpack ops but they are not the tensor approach codegen, e.g., iree-amd-aie project.

With the recent tensor.pack/unpack -> linalg.pack/unpack change, we are able to retire the linalg_ext version. Here are the steps for the deprecation:

1. Upstream the [generateScalarImplementation](https://github.com/iree-org/iree/blob/6c07258c2e53595eeec95188a7c4ce0bf4d9e316/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td#L942) implementation to MLIR repro, so all the linalg.pack/unpack ops on memrefs can be lowered to scalar codes.
2. Check in if the tiling implementation for memref types is available or not. If not, fill the gap in the upstream. IIRC, the main difference is whether passing result types or not.
3. Update all the linalg_ext uses in IREE to linalg version.
4. Delete all the linalg_ext pack/unpack codes.

hanhanW:
All the transformations are upstreamed, so we can retire the ops.

Fixes https://github.com/iree-org/iree/issues/20030

ci-extra: test_torch


hanhanW:
okay, it looks like we can delete the IREE's pack/unpack ops after we land the upstream changes:

- https://github.com/llvm/llvm-project/pull/182837
- https://github.com/llvm/llvm-project/pull/182838

I'm working on the IREE change that retires the ops.

hanhanW:
Oh, I thought that my changes are propagated to IREE. It seems like the last one is not yet landed to IREE.

MaheshRavishankar:
ok nice!

MaheshRavishankar (approved):
Looks good if tests pass!

PR #23564, opened 2026-02-24
Revert "[Codegen][VectorExt] Support multi-batch gather vectorization to transfer_gather"

Groverkss:
Reverts iree-org/iree#23552

Looks like this broke ci builds


Groverkss:
This is not the offending pr. The issue is because of another pr, but this pr added a test that caught it. Going to fix it forward.

PR #23568, opened 2026-02-24
[Codegen][GPU] Clean up prefetch pipeline stages flag to support integer values

Yu-Zhewen:
Replaces the boolean `--iree-llvmgpu-enable-prefetch` flag with an integer `--iree-llvmgpu-prefetch-num-stages` flag backed by std::optional<uint64_t>. When unset (default), each code path uses its own heuristic default. 0 or 1 disables pipelining, and 2+ enables pipelining with the specified number of stages.

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -52,7 +52,8 @@ constexpr int64_t kPreferredCopyNumBits = 128;
 
 LogicalResult setDataTiledMmaInnerTiledLoweringConfig(
     IREE::GPU::TargetAttr target, mlir::FunctionOpInterface entryPoint,
-    Operation *op, IREE::Codegen::UKernelDescriptorAttr ukernelConfig) {
+    Operation *op, IREE::Codegen::UKernelDescriptorAttr ukernelConfig,
+    int64_t prefetchNumStages) {
   auto multiMmaOp = dyn_cast<IREE::Codegen::InnerTiledOp>(op);
   if (!multiMmaOp) {
     return failure();
@@ -109,11 +110,12 @@ LogicalResult setDataTiledMmaInnerTiledLoweringConfig(
   DictionaryAttr configDict = b.getDictionaryAttr(attrs);
   auto loweringConfig = IREE::GPU::LoweringConfigAttr::get(context, configDict);
 
-  // Don't add any special padding or prefetching, since the data-tiled layout
-  // is already what we want.
+  // By default, don't add any special padding or prefetching, since the
+  // data-tiled layout is already what we want.
   SmallVector<NamedAttribute, 1> pipelineAttrs;
+  int64_t prefetchStages = prefetchNumStages >= 0 ? prefetchNumStages : 0;
   auto pipelineOptions = IREE::GPU::GPUPipelineOptionsAttr::get(
-      context, /*prefetchNumStages=*/0,
+      context, /*prefetchNumStages=*/prefetchStages,
       /*no_reduce_shared_memory_bank_conflicts=*/true,
       /*use_igemm_convolution=*/false,
       /*reorder_workgroups_strategy=*/std::nullopt);
@@ -1012,9 +1014,11 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   return std::pair{loweringConfig, flatWorkgroupSize};
 }
 
-LogicalResult setIGEMMConvolutionLoweringConfig(
-    IREE::GPU::TargetAttr target, mlir::FunctionOpInterface entryPoint,
-    Operation *op, bool useDirectLoad, bool padConv) {
+LogicalResult
+setIGEMMConvolutionLoweringConfig(IREE::GPU::TargetAttr target,
+                                  mlir::FunctionOpInterface entryPoint,
+                                  Operation *op, bool useDirectLoad,
+                                  bool padConv, int64_t prefetchNumStages) {
   auto linalgOp = dyn_cast<linalg::LinalgOp>(op);
   if (!linalgOp || !linalg::isaConvolutionOpInterface(linalgOp)) {
     return failure();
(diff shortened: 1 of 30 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -109,11 +110,12 @@ LogicalResult setDataTiledMmaInnerTiledLoweringConfig(
   DictionaryAttr configDict = b.getDictionaryAttr(attrs);
   auto loweringConfig = IREE::GPU::LoweringConfigAttr::get(context, configDict);
 
-  // Don't add any special padding or prefetching, since the data-tiled layout
-  // is already what we want.
+  // By default, don't add any special padding or prefetching, since the
+  // data-tiled layout is already what we want.
   SmallVector<NamedAttribute, 1> pipelineAttrs;
+  int64_t prefetchStages = prefetchNumStages >= 0 ? prefetchNumStages : 0;
```
lialan:
```suggestion
  int64_t prefetchStages = std::max<int64_t>(prefetchNumStages, 0);
```

lialan: I would personally use std::optional<int64_t> to avoid -1 as default value.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1499,8 +1498,11 @@ static LogicalResult setContractConfig(IREE::GPU::TargetAttr target,
       auto loweringConfig =
           IREE::GPU::LoweringConfigAttr::get(context, configDict);
       SmallVector<NamedAttribute, 1> pipelineAttrs;
+      // Default to no prefetching if not specified.
+      int64_t prefetchStages =
```
lialan: seen a few places that has this, so preferably a refactor, or if we use std::optional<int64>:
```
prefetchStages = gpuOpts.prefertchNumStages.value_or(0);
```

Yu-Zhewen: Done. Use `std::optional` now

compiler/src/iree/compiler/Tools/iree_compile_lib.cc
```diff
@@ -46,21 +47,6 @@ enum class CompileMode {
   precompile,
 };
 
-struct BytecodeVersionParser : public llvm::cl::parser<std::optional<int64_t>> {
```
lialan, line 49: i am not sure if this needs to remove?

kuhar (dismissed):

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -2399,8 +2412,7 @@ propagateLoweringConfig(Operation *rootOperation,
 //===----------------------------------------------------------------------===//
 // Entry Point
 //===----------------------------------------------------------------------===//
-LogicalResult initGPULaunchConfig(FunctionOpInterface funcOp,
-                                  const GPUCodegenOptions &gpuOpts) {
+LogicalResult initGPULaunchConfig(FunctionOpInterface funcOp) {
```
jerryyin, line 2431: Can you add a guard in this routine, where all config setting logic start from, to enforce that the pipeline number of state is non-negative? This comment is motivated from @lialan 's previous comment.

A negative stage count won't do any damage because we'd guard it later in the lowering pipeline. But still I believe the current state as-is a negative value will be silently accepted and materialized in IR like (prefetch_num_stages=-1), which is odd.

compiler/src/iree/compiler/Tools/iree_compile_lib.cc
Yu-Zhewen: Code changed. `BytecodeVersionParser` is now left untouched in its original location

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
Yu-Zhewen: I am now using `std::optional<int64_t>`, as @lialan previously suggested.

jerryyin: I assume using std::optional<int64_t> does not solve the negative value problem? The type int64_t is signed, std::optional doesn't constrain the range. You can still pass `--iree-llvmgpu-prefetch-num-stages=-3` and it will happily parse into std::optional<int64_t>(-3) and materialize as prefetch_num_stages = -3 in the IR.

Yu-Zhewen: Good point. Negative values are now rejected at parse time.

compiler/src/iree/compiler/Utils/OptionUtils.h
```diff
@@ -84,16 +84,16 @@ struct Deprecated {
...
+// non-negative integers. Unset on the command line means std::nullopt.
+struct NonNegativeOptionalInt64Parser
+    : public llvm::cl::parser<std::optional<int64_t>> {
+  NonNegativeOptionalInt64Parser(llvm::cl::Option &O)
       : llvm::cl::parser<std::optional<int64_t>>(O) {}
   bool parse(llvm::cl::Option &O, llvm::StringRef, llvm::StringRef arg,
              std::optional<int64_t> &v) {
     long long w;
-    if (llvm::getAsSignedInteger(arg, 10, w)) {
+    if (llvm::getAsSignedInteger(arg, 10, w) || w < 0) {
```
kuhar: Why not parse it as `std::optional<uint64_t>` if you want non-negative?

Yu-Zhewen: Do you mean parse it as `uint64_t` and then cast to `int64_t` internally? I think `int64_t` is more preferable even though the value is always non-negative. Happy to change if you feel it's worth it though.

kuhar: no, always keep uint64 -- during and after parsing

Yu-Zhewen: Done.

kuhar (approved):

PR #23569, opened 2026-02-24
[CI] Disable internal linkage clang-tidy checks

kuhar:
These generate too many false positives.

### DIFF
--- a/compiler/.clang-tidy
+++ b/compiler/.clang-tidy
@@ -14,6 +14,7 @@ Checks: >
   -misc-non-private-member-variables-in-classes,
   -misc-unused-parameters,
   -misc-use-anonymous-namespace,
+  -misc-use-internal-linkage,
   -misc-override-with-different-visibility,
   bugprone-*,
   -bugprone-assignment-in-if-condition,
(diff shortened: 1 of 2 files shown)


amd-eochoalo (approved):

bangtianliu (approved):

compiler/.clang-tidy
```diff
@@ -14,6 +14,7 @@ Checks: >
   -misc-non-private-member-variables-in-classes,
   -misc-unused-parameters,
   -misc-use-anonymous-namespace,
+  -misc-use-internal-linkage,
```
amd-eochoalo, line 17: I know I approved already and also this change is just limited to clang-tidy checks. I'm kind of regretting it a little bit. I do prefer internal linkage where possible. If other members of the team are ok with removing this check, then that's fine. :)

kuhar: Same but it's too noisy in PR checks

kuhar: Most of the time it suggests to make public functions declared in headers internal 🤷

PR #23572, opened 2026-02-24
[CODEGEN] Pattern to hoist expand_shape & collapse_shape from scf.for loop

### LINKED ISSUE #23534: FuseAndHoist problem in direct convolution
Background of Direct convolution: Currently direct convolution works by tiling the filter reduction dimensions to 1 which makes the convolution into a GEMM and then we pack to intrinsic and follow the regular path. We dont even bother to unroll the scf.for's generated by the filter reduction dimensions as we have found that the backend does a good job at the following sctructure after fuse and hoist
```
scf.forall // workgroup
  scf.forall // thread
   scf.for // filter H
      scf.for // filter K
       scf.for // actual reduction
```

However, a step that happens between the filter dimension tiling and the GEMM reduction tiling is the `ExpandUndistributedInnerTilesPass` which can add collapse and expands which then break the hoisting in the fuse and hoist pass [here](https://gist.github.com/nirvedhmeshram/ffa2907422b69a55c5426c7c66aae07c) is the full IR dump

The simplest solution seems like we must unroll the filter scf.for's I wanted to check with @qedawkins @Max191 if there is some other propagation pattern that makes sense instead or some addition to the hoisiting logic in fuse-and-hoist?

Practically, what this means is that without fixing this we are blocked from using certain intrinsics for which `ExpandUndistributedInnerTilesPass` is not a no-op for example to me it seems on MI355 for `MFMA_F32_16x16x32_F16` this issue doesnt happen but for `MFMA_F32_32x32x16_F16` we have this problem.
FYI @YashDeshpande25 @yzhang93 @MaheshRavishankar

YashDeshpande25:
This PR adds a new rewrite pattern that hoists matching tensor.expand_shape and tensor.collapse_shape pairs out of scf.for loops. This change is intended to help the GPUFuseAndHoistParallelLoopsPass hoist scf.forall loops properly in case of Direct Convolution when using specific MFMA instructions. 

Fixes : https://github.com/iree-org/iree/issues/23534

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
@@ -323,6 +323,114 @@ struct ExpandDestinationForallOp final
   }
 };
 
+/// This pattern hoists expand_shape & collapse_shape ops out of scf.for loops.
+struct ExpandDestinationForOp final : OpRewritePattern<scf::YieldOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(scf::YieldOp yieldOp,
+                                PatternRewriter &rewriter) const override {
+    Location loc = yieldOp.getLoc();
+    auto forOp = dyn_cast<scf::ForOp>(yieldOp->getParentOp());
+    if (!forOp) {
+      return failure();
+    }
+    tensor::CollapseShapeOp collapseOp;
+    tensor::ExpandShapeOp expandOp;
+    int64_t tiedResultIdx = 0;
+
+    for (auto [idx, operand] : llvm::enumerate(yieldOp.getOperands())) {
+      collapseOp = operand.getDefiningOp<tensor::CollapseShapeOp>();
+      if (!collapseOp) {
+        continue;
+      }
+      if (collapseOp.getSrcType().getRank() ==
+          collapseOp.getResultType().getRank()) {
+        continue;
+      }
+
+      // Get the corresponding expandOp.
+      auto iterArg = forOp.getRegionIterArgs()[idx];
+      for (auto user : iterArg.getUsers()) {
+        expandOp = dyn_cast<tensor::ExpandShapeOp>(user);
+        if (expandOp &&
+            (expandOp.getReassociationIndices() ==
+             collapseOp.getReassociationIndices()) &&
+            (expandOp.getResultType() == collapseOp.getSrcType())) {
+          break;
+        } else {
+          expandOp = nullptr;
+        }
+      }
+
+      if (expandOp && collapseOp) {
+        bool hasOtherUsers = false;
+        for (auto user : iterArg.getUsers()) {
+          if (user != expandOp) {
+            hasOtherUsers = true;
+            expandOp = nullptr;
+            collapseOp = nullptr;
+            break;
+          }
+        }
+        if (!hasOtherUsers) {
+          tiedResultIdx = idx;
+          break;
+        }
+      }
+    }
+    if (!expandOp || !collapseOp) {
+      return failure();
+    }
+
+    // Create the expand -> new scf.for -> collapse chain.
+    rewriter.setInsertionPoint(forOp);
+
+    Value initArg = forOp.getInitArgs()[tiedResultIdx];
+    auto expandedDest = tensor::ExpandShapeOp::create(
+        rewriter, loc, expandOp.getResultType(), initArg,
+        expandOp.getReassociationIndices());
+
+    auto expandedInitArgs = llvm::to_vector_of<Value>(forOp.getInitArgs());
+    expandedInitArgs[tiedResultIdx] = expandedDest.getResult();
+
+    scf::ForOp newForOp = scf::ForOp::create(
+        rewriter, loc, forOp.getLowerBound(), forOp.getUpperBound(),
+        forOp.getStep(), expandedInitArgs);
+
+    auto collapsedOutput = tensor::CollapseShapeOp::create(
+        rewriter, loc, collapseOp.getResultType(),
+        newForOp.getResults()[tiedResultIdx],
+        collapseOp.getReassociationIndices());
+
+    // Users of the result of collapseOp must use the input to the collapseOp.
+    collapseOp->getResult(0).replaceAllUsesWith(collapseOp.getOperand());
+
+    // Users of the result of expandOp must use the iter_arg of the new forOp.
+    for (auto user : forOp.getRegionIterArgs()[tiedResultIdx].getUsers()) {
+      if (user->getNumResults() > 0) {
+        user->getResult(0).replaceAllUsesWith(
+            newForOp.getRegionIterArgs()[tiedResultIdx]);
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
```diff
@@ -446,6 +568,7 @@ void PropagateReshapesByExpansionPass::runOnOperation() {
   bubbleExpandShapePatterns
       .add<ExpandDestinationForallOp, SwapInnerBitcastWithExtractSlice>(
           context);
+  bubbleExpandShapePatterns.add<PropagateReshapeDestinationForOp>(context);
```
nirvedhmeshram: nit: you can just add the pattern to the existing patterns on the line above.

compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
```diff
@@ -323,6 +323,128 @@ struct ExpandDestinationForallOp final
...
+          expandOp = nullptr;
+        }
+      }
+
+      if (expandOp && collapseOp) {
+        // Apply the pattern only if expandOp & collapseOp are the only 2 users
+        // of the regionIterArg.
+        bool hasOtherUsers = false;
+        for (auto user : iterArg.getUsers()) {
+          if (user != expandOp && user != collapseOp) {
```
nirvedhmeshram: I dont follow why the collapse shape would be using iter arg, I dont think you need to check for that?

compiler/src/iree/compiler/Codegen/Common/test/propagate_reshapes_by_expansion.mlir
```diff
@@ -661,3 +661,32 @@ func.func @propagate_expand_through_inner_tiled_dynamic(
...
 // CHECK-SAME:        : tensor<?x3x4x16x16xf16>, tensor<4x2x16x16xf16> into tensor<?x3x2x16x16xf32>
 // CHECK:         return %[[INNER_TILED]]
+
+// -----
+
+func.func @propagate_reshape_dest_for(%init: tensor<4x16xf32>) -> tensor<4x16xf32> {
+  %c0 = arith.constant 0 : index
+  %c1 = arith.constant 1 : index
+  %c4 = arith.constant 4 : index
+  %result = scf.for %iv = %c0 to %c4 step %c1 iter_args(%arg = %init) -> tensor<4x16xf32> {
```
nirvedhmeshram, line 671: add a negative case where %arg does indeed have another user.

nirvedhmeshram (commented):
Looking good, just have some minor comments. you can mark this ready for review after addressing those.

compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
```diff
@@ -323,6 +323,128 @@ struct ExpandDestinationForallOp final
   }
 };
 
+/// This pattern hoists expand_shape & collapse_shape ops out of scf.for loops.
+struct PropagateReshapeDestinationForOp final : OpRewritePattern<scf::YieldOp> {
```
nirvedhmeshram: nit: maybe match the name of the pattern to the forall pattern and call it `ExpandDestinationForOp` ?

krzysz00: 1. This is over commented in a very LLM way
2. Creating a for loop like you do here feels overcomplicated - you can just take the unexpended argument, setType() it to its new type, and then RAUW it in before erasing the expand/collapse pair

krzysz00 (commented):
Some notes

compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
nirvedhmeshram: I think it would help reviewing if you see the pattern above that does the same thing for scf.forall . In that one we found that doing the erasing was causing crashes as there are other patterns in the rewriter that have the erased ops in there queue and then they crash, In general i think erasing ops like that runs into trouble with the greedy rewriter. We have found its better to just make a new op like this. That being said not opposed to trying it here to see if it works.

Also, dont see more comments in this pattern here compared to the pattern above, would be good to be specific about which ones you think are excessive.

compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
YashDeshpande25: Yes, no such case that I know of, so removing the check for collapseOp in the if-statement.

compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
YashDeshpande25: Hey, I have removed some of the smaller comments & will try the alternative approach as well. If there’s a previous similar implementation that I can refer to,  it would be great to better understand the pattern.

Marking this for review.

krzysz00: For the crashes, were you using modifyOpInPlace() and the other methods that notify the rewriter of the change?

nirvedhmeshram: I copied this strategy from some patterns from @Max191, this was a while back though so cant be sure but I do think the scf.for (in this case) was modifyinPlace() the problem happened with the expand/collapse ops. I dont remember why we have to erase them though, maybe can just leave them dead but vaguely remember that not being a decent option either.

compiler/src/iree/compiler/Codegen/Common/PropagateReshapesByExpansion.cpp
```diff
@@ -323,6 +323,115 @@ struct ExpandDestinationForallOp final
...
 };
 
+/// This pattern hoists expand_shape & collapse_shape ops out of scf.for loops.
+struct ExpandDestinationForOp final : OpRewritePattern<scf::YieldOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(scf::YieldOp yieldOp,
+                                PatternRewriter &rewriter) const override {
+    Location loc = yieldOp.getLoc();
+    auto parentOp = yieldOp->getParentOp();
+    auto forOp = dyn_cast<scf::ForOp>(parentOp);
```
krzysz00:
```suggestion
    auto forOp = dyn_cast<scf::ForOp>(yieldOp->getParentOp());
```

krzysz00 (commented):
On further reading, the comments seem like somewhat reasonable section headers, and I suspect that the code here is basically fine perf-wise (since I don't think it involves big clones)

nirvedhmeshram (approved):
LGTM, I dont see the need to do the in place modification / erase thing as we are just inlining the region of the old scf.for and not cloning and this approach looks simple enough to me. However lets give other reviewers some time before landing this.

krzysz00 (approved):
lgtm here

PR #23579, opened 2026-02-25
fix cpu_count comparison preventing specifying 64 core_ids.

bjacob:


krzysz00 (commented):
Seems reasonable to me?

krzysz00 (approved):

PR #23581, opened 2026-02-25
[ROCM][DT] Add gfx950 f8e4m3fn ukernel

Yu-Zhewen:
This PR ports the gfx942 data-tiled large ukernel to gfx950, with tuning applied to the specialization range to  address the performance regression observed when directly migrating the gfx942 configs (see https://github.com/iree-org/iree/issues/22121#issuecomment-3563023732).

After experimenting with intrinsic selection, unroll factors, and size constraint, the size constraint was identified as the most impactful tuning knob. And only the large ukernel proved beneficial, when N>=8192.

Results: Llama 8B f8e4m3fn, gfx950, prefill length 2048:
|                                     | Time (us) |
|-------------------------------------|-----------|
| codegen                             | 279       |
| codegen+ukernel                     | 245       |
| data_tiling                         | 253       |
| data_tiling+ukernel (before tuning) | 270       |
| data_tiling+ukernel (after tuning)  | **242**       |


bjacob (approved):

jtuyls (approved):
LGTM

PR #23587, opened 2026-02-26
[CPU] Make pipeline transformation work with single module.

hanhanW:
There are three passes working on variantOp scope, which makes the setup of pipeline tests weird. E.g., currently it requires either two level of nested module or invoking specific passes.

The revision adds an option to skip variantOp scope passes (like distribution and export specialization) and rename it to `LLVMCPULoweringPipeline`, because it is mainly for lowering. It also introduces an option to stop the lowering before LLVM conversion which makes IR checks much easier, which also aligns with the result of current specific pass invocation.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
@@ -685,62 +685,40 @@ void buildLLVMCPUCodegenConfigurationPassPipelineImpl(
 }
 
 void buildLLVMCPUCodegenConfigurationPassPipeline(
-    OpPassManager &variantPassManager, const CPUCodegenOptions &cpuOpts,
-    bool pipelineOnModule) {
-  if (!pipelineOnModule) {
-    variantPassManager.addPass(createSpecializeExportsPass());
-  }
-  OpPassManager &modulePassManager = pipelineOnModule
-                                         ? variantPassManager
-                                         : variantPassManager.nest<ModuleOp>();
+    OpPassManager &modulePassManager, const CPUCodegenOptions &cpuOpts) {
+  modulePassManager.addPass(createSpecializeExportsPass());
   buildLLVMCPUCodegenConfigurationPassPipelineImpl(modulePassManager, cpuOpts);
 }
 
-void buildLLVMCPUCodegenPassPipeline(OpPassManager &variantPassManager,
+void buildLLVMCPUCodegenPassPipeline(OpPassManager &modulePassManager,
                                      const CPUCodegenOptions &cpuOpts,
                                      bool enableAArch64SME,
-                                     bool pipelineOnModule,
                                      bool includeLLVMLowering) {
-  {
-    OpPassManager &passManager = pipelineOnModule
-                                     ? variantPassManager
-                                     : variantPassManager.nest<ModuleOp>();
-    passManager.addPass(createLowerExecutableUsingTransformDialectPass());
-    FunctionLikeNest(passManager)
-        .addPass([&]() {
-          return createLLVMCPULowerExecutableTargetPass(
-              LLVMCPULowerExecutableTargetPassOptions{cpuOpts});
-        })
-        .addPredicatedPass(includeLLVMLowering,
-                           createVerifyWorkgroupDistributionPass);
-    if (clPatchFuncOps) {
-      passManager.addPass(createPatchFuncOpsPass());
-    }
+  modulePassManager.addPass(createLowerExecutableUsingTransformDialectPass());
+  FunctionLikeNest(modulePassManager)
+      .addPass([&]() {
+        return createLLVMCPULowerExecutableTargetPass(
+            LLVMCPULowerExecutableTargetPassOptions{cpuOpts});
+      })
+      .addPredicatedPass(includeLLVMLowering,
+                         createVerifyWorkgroupDistributionPass);
+  if (clPatchFuncOps) {
+    modulePassManager.addPass(createPatchFuncOpsPass());
   }
 
   if (!includeLLVMLowering) {
     return;
   }
 
-  // Variant-level passes that require ExecutableVariantOp context. Skipped in
-  // pipelineOnModule mode (no exports to reconcile or resolve).
-  if (!pipelineOnModule) {
-    variantPassManager.addPass(createReconcileTranslationInfoPass());
-    variantPassManager.addPass(createResolveWorkgroupCountHintsPass());
-  }
-  variantPassManager.addPass(createIREECodegenLowerAffinePass());
-  variantPassManager.addPass(IREE::Util::createDropCompilerHintsPass());
+  modulePassManager.addPass(createReconcileTranslationInfoPass());
+  modulePassManager.addPass(createResolveWorkgroupCountHintsPass());
+  modulePassManager.addPass(createIREECodegenLowerAffinePass());
+  modulePassManager.addPass(IREE::Util::createDropCompilerHintsPass());
 
-  // Run conversion to LLVM at `ModuleOp` granularity.
-  {
-    OpPassManager &passManager = pipelineOnModule
-                                     ? variantPassManager
-                                     : variantPassManager.nest<ModuleOp>();
-    addLowerToLLVMPasses(passManager, enableAArch64SME, cpuOpts);
-  }
+  addLowerToLLVMPasses(modulePassManager, enableAArch64SME, cpuOpts);
   LLVM_DEBUG({
     llvm::dbgs() << "LLVMCPU codegen pass pipeline:\n";
-    variantPassManager.printAsTextualPipeline(llvm::dbgs());
+    modulePassManager.printAsTextualPipeline(llvm::dbgs());
     llvm::dbgs() << "\n";
   });
 }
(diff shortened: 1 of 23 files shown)


hanhanW:
@qedawkins I think you were going to improve the transformation pipeline, so I want to ask you for review. You also have context about how we set up distribution including `ReconcileTranslationInfoPass`, so I think it is better to have your eyes on it.

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
```diff
@@ -685,38 +685,58 @@ void buildLLVMCPUCodegenConfigurationPassPipelineImpl(
...
+  }
+
+  // Variant-level passes that require ExecutableVariantOp context. Skipped in
+  // pipelineOnModule mode (no exports to reconcile or resolve).
+  if (!pipelineOnModule) {
+    variantPassManager.addPass(createReconcileTranslationInfoPass());
+    variantPassManager.addPass(createResolveWorkgroupCountHintsPass());
+  }
   variantPassManager.addPass(createIREECodegenLowerAffinePass());
   variantPassManager.addPass(IREE::Util::createDropCompilerHintsPass());
```
qedawkins: These are still running on `variantPassManager`, shouldn't they be on `passManager`?

qedawkins (commented):
To me nest switches like this are a sign the pipeline is spanning too many stages. Instead it would make more sense to me to add a dedicated `LowerToLLVM` pass pipeline and a dedicated pass pipeline for the pre-variant steps. Then the current one stays as is.

In general I'd like us to trend towards more "logically distinct" stages in codegen that allows injecting partially lowered IR.

I won't block this as an intermediate step though.

hanhanW:
> To me nest switches like this are a sign the pipeline is spanning too many stages. Instead it would make more sense to me to add a dedicated `LowerToLLVM` pass pipeline and a dedicated pass pipeline for the pre-variant steps. Then the current one stays as is.
> 
> In general I'd like us to trend towards more "logically distinct" stages in codegen that allows injecting partially lowered IR.
> 
> I won't block this as an intermediate step though.

I have a question about why don't we run `createReconcileTranslationInfoPass` and `createResolveWorkgroupCountHintsPass` right after distribution. IIUC, we have enough information to lower ops within export op right after distribution. We can make it the nested structure easier if we move them earilier.

The other approach is forcing the passes work on ModuleOp instead of VariantOp and seek for the export op from parant if exists. This will update the tablegen, though. But it looks cleaner to me.

I'm feeling that they are mostly because of legac. If people who have fresh context agree with that, I'm happy to give the approach a shot.

The current nested structure is

```
  buildLLVMCPUCodegenConfigurationPassPipeline:

  Operates on variantPassManager (HAL ExecutableVariantOp level):

  variantPassManager                              // ExecutableVariantOp
  ├── createSpecializeExportsPass()
  └── nest<ModuleOp>  →  modulePassManager        // ModuleOp
      ├── FunctionLikeNest {
      │   └── addCommonTargetExecutablePreprocessingPasses(...)
      │ }
      ├── createMaterializeTuningSpecsPass(...)
      ├── createMaterializeUserConfigsPass()
      ├── FunctionLikeNest {
      │   ├── createMaterializeDeviceEncodingPass
      │   ├── createCPUPropagateDataLayoutPass
      │   ├── createRematerializeParallelOpsPass
      │   ├── createExpandF16OpToF32Pass
      │   ├── createConvertAccGEMMToGEMMPass
      │   └── createEraseHALDescriptorTypeFromMemRefPass
      │ }
      └── createLLVMCPUSelectLoweringStrategyPass()

  buildLLVMCPUCodegenPassPipeline                                                                                                           07:14:11 [46/88]

  Operates on variantPassManager (HAL ExecutableVariantOp level):

  variantPassManager                              // ExecutableVariantOp
  ├── nest<ModuleOp>  →  modulePassManager        // ModuleOp
  │   ├── createLowerExecutableUsingTransformDialectPass()
  │   ├── FunctionLikeNest {
  │   │   ├── createLLVMCPULowerExecutableTargetPass(...)
  │   │   └── createVerifyWorkgroupDistributionPass
  │   │ }
  │   └── [if clPatchFuncOps] createPatchFuncOpsPass()
  │
  ├── createReconcileTranslationInfoPass()        // back at variant level
  ├── createResolveWorkgroupCountHintsPass()
  ├── createIREECodegenLowerAffinePass()
  ├── IREE::Util::createDropCompilerHintsPass()
  │
  └── nest<ModuleOp>  →  modulePassManager        // ModuleOp (addLowerToLLVMPasses)
      ├── FunctionLikeNest { createEraseHALDescriptorTypeFromMemRefPass }
      ├── createLowerUKernelOpsToCallsPass()
      ├── FunctionLikeNest {
      │   ├── LinalgExt::createLinalgExtToLoopsPass
      │   ├── createMemrefCopyToLinalgPass
      │   ├── [predicated] createLLVMCPUEmitVectorizationRemarksPass
      │   ├── createConvertLinalgToLoopsPass
      │   ├── createConvertBf16ArithToF32Pass
      │   ├── createConvertUnsupportedFloatToIntBuffersPass(bf16 only)
      │   ├── createCanonicalizerPass
      │   └── createCSEPass
      │ }
      ├── createIREEBufferizeConstantsPass()
  ...
```

The issue is that the two passes are making it harder.

hanhanW:
> The issue is that the two passes are making it harder.

The pass order does not matter if we look for parent module.

qedawkins:
> The other approach is forcing the passes work on ModuleOp instead of VariantOp and seek for the export op from parant if exists. This will update the tablegen, though. But it looks cleaner to me.

Ok I thought about it a bit more and this is probably a good direction. Ultimately what we probably want is rather than reaching directly up and changing the export op, ReconcileTranslationInfo should be doing nothing more than populating a function that hints a different pass that's responsible for populating the workgroup count region. So something like
```
hal.executable.export @foo(%device, %workload) {
  // placeholder op for the workgroup count
}
module {
  func.func @foo() {
    %0 = hal.interface.id
    ...
  }
  iree_codegen.workgroup_count @foo {
    // workgroup count computation for function foo
  }
}
```

Then in codegen all passes are relegated to module or function scopes. There's no reaching up, instead we're just populating functions that are being implicitly called by the export region. Separate from lowering of the module itself, we have another pass that extracts the `iree_codegen.workgroup_count` function and uses it to populate the workgroup count region. The reason I think we want to do it this way is because once we start actually setting the workgroup count based on device queries it helps separate the stuff we can compute on device (*what* the problem shape is) from stuff we can only know on the host (things like GPU mode or CU count or something).

> I have a question about why don't we run createReconcileTranslationInfoPass and createResolveWorkgroupCountHintsPass right after distribution.

So actually I'm thinking the opposite. They should run after LLVM conversion if they can as the absolute last step (but before serialization/linking). The reason they can't right now is because the workgroup count computation is inline with the rest of the function.

hanhanW:
> > I have a question about why don't we run createReconcileTranslationInfoPass and createResolveWorkgroupCountHintsPass right after distribution.
> 
> So actually I'm thinking the opposite. They should run after LLVM conversion if they can as the absolute last step (but before serialization/linking). The reason they can't right now is because the workgroup count computation is inline with the rest of the function.

Why do we want to run them after LLVM conversion? One of an issue that I hit yesterday is about VMVX + data-tiling: https://github.com/iree-org/iree/issues/23588, although this path is not as active as others and we may need to revisit it.

> With data tiling, encoding dispatches contain linalg.pack ops whose outer dimensions depend on iree_codegen.query_tile_sizes (a Pure op that queries VMVX runtime tile sizes). When forall-based distribution tiles the pack, the scf.forall upper bounds carry this dependency.
>
> ReconcileTranslationInfoPass creates a workgroup_count_hint from the forall's upper bounds. ResolveWorkgroupCountHintsPass then tries to trace the hint back to workload ordinals, but the backward slice hits call @vmvx.query_tile_sizes.2d (the lowered form of iree_codegen.query_tile_sizes) and rejects it because func.call is not memory-effect-free (ResolveWorkgroupCountHints.cpp:158).
>
> The core issue is that iree_codegen.query_tile_sizes is Pure, but by the time ResolveWorkgroupCountHintsPass runs, it has already been lowered to a func.call by LowerUKernelOpsToCallsPass, losing the purity information.

Honestly, I don't follow why we have multiple passes and run some in later stage, but can't do it within a single distribution pass. Why do we split the passes?

hanhanW:
> > > I have a question about why don't we run createReconcileTranslationInfoPass and createResolveWorkgroupCountHintsPass right after distribution.
> > 
> > 
> > So actually I'm thinking the opposite. They should run after LLVM conversion if they can as the absolute last step (but before serialization/linking). The reason they can't right now is because the workgroup count computation is inline with the rest of the function.
> 
> Why do we want to run them after LLVM conversion? One of an issue that I hit yesterday is about VMVX + data-tiling: #23588, although this path is not as active as others and we may need to revisit it.
> 
> > With data tiling, encoding dispatches contain linalg.pack ops whose outer dimensions depend on iree_codegen.query_tile_sizes (a Pure op that queries VMVX runtime tile sizes). When forall-based distribution tiles the pack, the scf.forall upper bounds carry this dependency.
> > ReconcileTranslationInfoPass creates a workgroup_count_hint from the forall's upper bounds. ResolveWorkgroupCountHintsPass then tries to trace the hint back to workload ordinals, but the backward slice hits call @vmvx.query_tile_sizes.2d (the lowered form of iree_codegen.query_tile_sizes) and rejects it because func.call is not memory-effect-free (ResolveWorkgroupCountHints.cpp:158).
> > The core issue is that iree_codegen.query_tile_sizes is Pure, but by the time ResolveWorkgroupCountHintsPass runs, it has already been lowered to a func.call by LowerUKernelOpsToCallsPass, losing the purity information.
> 
> Honestly, I don't follow why we have multiple passes and run some in later stage, but can't do it within a single distribution pass. Why do we split the passes?

I think I need to study the passes more, I'll come back later.

qedawkins:
> Why do we want to run them after LLVM conversion? One of an issue that I hit yesterday is about VMVX + data-tiling: https://github.com/iree-org/iree/issues/23588, although this path is not as active as others and we may need to revisit it.

This might be an issue with the `query_tile_sizes` op rather than the pass ordering, but I'm unfamiliar with the VMVX + data-tiling path.

> Why do we want to run them after LLVM conversion? One of an issue that I hit yesterday is about VMVX + data-tiling: https://github.com/iree-org/iree/issues/23588, although this path is not as active as others and we may need to revisit it.

I'm thinking less about when the LLVM conversion happens and more about separating concerns. The point at which we introduce the hints and setup `scf.forall` ops we're local to an individual function. There may be other functions being translated in parallel and maybe have not yet had the hints set, and those functions could themselves be called from a shared entry point, which means we can't set the workgroup count immediately after tiling. What we can do though is separate the computation related to workgroup count, creating a distinct boundary between the IR that needs to go to the dispatch and the IR that runs on the workgroup. This distinction is important because certain values are only available on the host and certain values are only available on the device. Today the only reason we have to run LLVM conversion after ReconcileTranslationInfo is because the split doesn't happen until then and we can't do the split after converting to LLVM. So my reasoning then is that if TileAndDistributeToWorkgroups + other passes properly did the split in advance, ReconcileTranslationInfo could happen whenever and the problem of setting workgroup counts becomes self contained and easy to extend (adding in device queries, setting it based on available parallelism, etc.).

At least that's my view on it :P
The pass structure could use some love for sure so I really appreciate you looking at it.

hanhanW:
Starting an RFC here: https://github.com/iree-org/iree/discussions/23642

hanhanW:
https://github.com/iree-org/iree/pull/23953 is the outcome, which looks better

PR #23590, opened 2026-02-26
[Codegen] Add PipelineAttrInterface and PassPipelineAttr

qedawkins:
This change adds an attribute interface for representing pass pipelines and a single basic attribute that uses the string based pass interpreter to populate a pipeline. The intent of this change is NOT to induce a refactor of all the pass pipelines, instead it's primarily to make testing structural pipeline changes with partially lowered inputs much easier. Today if you want to work on a change that affects later stages of a pass pipeline but will also require changes to earlier steps, it's hard to stage those changes since there isn't a convenient way to jump into the middle of a codegen pass pipeline (unlike the rest of the compiler which offers distinct stages).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
@@ -19,6 +19,50 @@
 #include "mlir/Dialect/Utils/StructuredOpsUtils.h"
 #include "mlir/IR/DialectImplementation.h"
 #include "mlir/IR/StorageUniquerSupport.h"
+#include "mlir/Pass/PassRegistry.h"
+
+// Custom parse/print directives for TranslationInfoAttr's pipeline field.
+// These must be defined before the generated .cpp.inc is included because
+// the ODS-generated parse/print methods call them.
+namespace mlir::iree_compiler::IREE::Codegen {
+
+/// Parses either a DispatchLoweringPassPipeline enum keyword (e.g.,
+/// `CPUDefault`) or a generic attribute implementing PipelineAttrInterface
+/// (e.g., `#iree_codegen.pass_pipeline<"canonicalize">`).
+static ParseResult parsePipelineAttr(AsmParser &parser, Attribute &result) {
+  StringRef keyword;
+  SMLoc loc = parser.getCurrentLocation();
+  if (succeeded(parser.parseOptionalKeyword(&keyword))) {
+    std::optional<DispatchLoweringPassPipeline> pipeline =
+        symbolizeDispatchLoweringPassPipeline(keyword);
+    if (!pipeline) {
+      parser.emitError(loc, "unknown pipeline keyword: ") << keyword;
+      return failure();
+    }
+    result =
+        DispatchLoweringPassPipelineAttr::get(parser.getContext(), *pipeline);
+    return success();
+  }
+  Attribute attr;
+  if (parser.parseAttribute(attr)) {
+    return failure();
+  }
+  result = attr;
+  return success();
+}
+
+/// Prints DispatchLoweringPassPipelineAttr as a bare keyword and other
+/// attributes (e.g., PipelineAttrInterface impls) via the generic printer.
+static void printPipelineAttr(AsmPrinter &printer, Attribute pipelineAttr) {
+  if (auto enumAttr =
+          dyn_cast<DispatchLoweringPassPipelineAttr>(pipelineAttr)) {
+    printer << stringifyEnum(enumAttr.getValue());
+    return;
+  }
+  printer.printAttribute(pipelineAttr);
+}
+
+} // namespace mlir::iree_compiler::IREE::Codegen
 
 #define GET_ATTRDEF_CLASSES
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp.inc"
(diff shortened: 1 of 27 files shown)


compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
```diff
@@ -269,22 +295,15 @@ def IREECodegen_TranslationInfoAttr :
...
-  let assemblyFormat = [{
-    `<` `pipeline` `=` `` $passPipeline
-    (`codegen_spec` `=` $codegenSpec^)?
-    (`workgroup_size` `=` `[` $workgroupSize^ `]`)?
-    (`subgroup_size` `=` $subgroupSize^)?
-    (`,` $configuration^)? `>`
+    The `passPipeline` field can be either:
+    - A `DispatchLoweringPassPipelineAttr` (enum keyword like `CPUDefault`).
+    - Any attribute implementing `PipelineAttrInterface` (e.g.,
+      `#iree_codegen.pass_pipeline<"...">`).
```
Max191, line 301: Why not just make a new codegen attr that implements `PipelineAttrInterface` and carries the enum on it, rather than allowing both enums and PipelineAttrInterface? Should basically just be moving some code around, and then all pipeline population happens through the same interface.

I know it would break the tuner, but it would be a simple fix AFAICT.

qedawkins: Not breaking the tuner helps, but the main reason is that different backends are allowed to interpret different enums however they like (None and TransformDialect are shared for example).

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
```diff
@@ -19,6 +19,50 @@
...
+  return success();
+}
+
+/// Prints DispatchLoweringPassPipelineAttr as a bare keyword and other
+/// attributes (e.g., PipelineAttrInterface impls) via the generic printer.
+static void printPipelineAttr(AsmPrinter &printer, Attribute pipelineAttr) {
+  if (auto enumAttr =
+          dyn_cast<DispatchLoweringPassPipelineAttr>(pipelineAttr)) {
+    printer << stringifyEnum(enumAttr.getValue());
+  } else {
```
kuhar: nit: exit early and no else

kuhar (approved):
I'd probably go even a step further and say that every dispatch pipeline has to be its own attribute that implements an interface, instead of relying on enum values. This would require updating a whole bunch of test updates.

But anyway, this refactor does seem like a good incremental step.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
```diff
@@ -64,6 +108,28 @@ ArrayAttr ExportConfigAttr::getWorkgroupSizeIndexArray() {
...
 
+//===----------------------------------------------------------------------===//
+// iree_codegen.pass_pipeline
+//===----------------------------------------------------------------------===//
+
+LogicalResult PassPipelineAttr::buildPipeline(OpPassManager &pm) const {
+  if (failed(parsePassPipeline(getPipeline(), pm))) {
+    return failure();
+  }
+  return success();
```
krzysz00, line 119:
```suggestion
  return success(succeeded(parsePassPipeline(getPipeline(), pm)));
```

krzysz00 (approved):
Looks like a good first step to me

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
kuhar: I prefer what Quinn has because you can put breakpoints on each path

qedawkins:
> I'd probably go even a step further and say that every dispatch pipeline has to be its own attribute that implements an interface, instead of relying on enum values. This would require updating a whole bunch of test updates.
> 
> But anyway, this refactor doesn't seems like a good incremental step.

Yeah I'm thinking of something one step further. Adding support for new hardware/strategies/inputs is hitting scaling limits. It requires either extending the lowering config, new pipeline options, or new pass pipelines, all of which have disadvantages (hard to control attribute dsl, redundant/no-op passes, or duplicative code). So what I want to do is break up pipelines into smaller portions that cover a smaller space of the lowering pipeline. This has the benefit of giving us a notion of stages similar to the rest of the compiler. Imagine compile-from and compile-to for the codegen backends.

I also want to reframe the stuff for Codegen in the iree/src directory as a set of tools for building a compiler like LLVM and move the orchestration into the plugins.

kuhar:
> But anyway, this refactor doesn't seem like a good incremental step.

FYI, the negation here was an unintentional typo

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
```diff
@@ -170,45 +169,57 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
...
+  // Check for a custom pipeline via PipelineAttrInterface.
+  Attribute pipelineAttr = translationInfo.getPassPipeline();
+  if (auto customPipeline =
+          dyn_cast<IREE::Codegen::PipelineAttrInterface>(pipelineAttr)) {
+    if (failed(customPipeline.buildPipeline(passManager))) {
+      funcOp.emitOpError("failed to build custom pass pipeline");
+      return signalPassFailure();
+    }
+  } else {
+    switch (pipeline) {
```
hanhanW: Where is the variable declared now?

hanhanW (approved):

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
qedawkins: Oops, I messed up the rebase. Thanks!

PR #23598, opened 2026-02-27
Fix the ordering of the external stream.

### LINKED ISSUE #23575: HIP external stream: set_external_stream_data_completed fires before semaphore infrastructure is ready
## Summary

When using the HIP driver with an external stream (`iree_hal_hip_device_params_t::external_stream`), there is a race in `hip_device.c` where `iree_hal_hip_set_external_stream_data_completed` unblocks the caller thread **before** `iree_hal_hip_device_stream_signal_semaphores_and_add_cleanup` has recorded the HIP event and updated the semaphore's `max_value_to_be_signaled`. This creates a window where `iree_hal_semaphore_wait` can observe an inconsistent semaphore state and return before the GPU work has completed.

## Observed failures

We are using IREE as a runtime backend (via the HAL C-API) with external HIP streams. Under CI load (~1 in 100 runs or more), `iree_hal_device_transfer_d2h` returns stale data after `iree_hal_buffer_view_allocate_buffer_copy` or `iree_vm_invoke`. Some CI runs that failed are captured in https://gist.github.com/sjain-stanford/ff250dbadef95c45187dd42de3f927cf. 

The failures fall into three categories:

| Category | Example | Interpretation |
|----------|---------|----------------|
| **Zeros** | Read 0 instead of computed value | Output buffer never written — kernel or H2D hasn't completed |
| **Partial/stale values** | Read -100 instead of -37 (= ADD(-50,-50) instead of ADD(-50,13)) | Input buffer H2D incomplete — pool-reused memory still holds stale data |
| **NaN** | Read `nanf` instead of `16.0f` | Uninitialized device memory interpreted as float |

The partial-value case is diagnostic: it shows the H2D for the second input buffer returned before the GPU copy completed, so the kernel read stale data from a pool-reused allocation.

These are not reproducible on a quiet machine but appear consistently on loaded CI runners (GitHub Actions shared runners), consistent with a scheduling-dependent race.

### The race

Three dispatch callbacks in `hip_device.c` share the same problematic ordering:

**`iree_hal_hip_device_execute_now`** (lines ~2467-2480):
```c
// GPU work enqueued on external stream (hipGraphLaunch / hipMemcpyAsync / …)

// (A) Caller unblocked — but semaphore event not yet recorded
if (device->uses_external_stream) {
    iree_hal_hip_set_external_stream_data_completed(
        data->base.external_stream_dispatch_data);       // line 2468
}

// (B) HIP event recorded, max_value_to_be_signaled updated
status = iree_hal_hip_device_stream_signal_semaphores_and_add_cleanup(…);  // line 2477
```

The same pattern exists in:
- `iree_hal_hip_device_perform_queue_read_now` (lines ~2124-2126)
- The queue-write equivalent

### Why this causes data corruption

1. **Caller calls `iree_hal_device_transfer_and_wait`** → creates a fence semaphore → calls `iree_hal_device_queue_execute` → blocks on `iree_hal_hip_wait_for_dispatch`.

2. **Dispatch thread runs `execute_now`**: enqueues GPU work, then calls `set_external_stream_data_completed` **(A)** — the caller wakes up from `wait_for_dispatch` and returns from `queue_execute`.

3. **Caller enters `iree_hal_semaphore_wait`**: checks `max_value_to_be_signaled` and the event tree. If the dispatch thread hasn't reached **(B)** yet, `max_value_to_be_signaled` is stale (0) and the event tree is empty.

4. **The semaphore wait has a notification-based fallback** (`event_semaphore.c:961`) that re-checks after `state_notification` is posted. Under normal timing this catches up. But under scheduling pressure (loaded CI, context switches between steps A and B), the window is wide enough for the wait to observe an inconsistent state.

### Why only external streams are affected

With internal streams, `iree_hal_hip_wait_for_dispatch` is not called — the dispatch thread handles synchronization internally via GPU callbacks. With external streams, the caller thread is explicitly unblocked at step (A), creating the race with step (B).

## Proposed fix

Move `set_external_stream_data_completed` **after** `stream_signal_semaphores_and_add_cleanup` so that the semaphore event is recorded and `max_value_to_be_signaled` is updated before the caller is unblocked:

```c
// BEFORE (current):
if (device->uses_external_stream) {
    iree_hal_hip_set_external_stream_data_completed(…);  // caller wakes
}
status = iree_hal_hip_device_stream_signal_semaphores_and_add_cleanup(…);

// AFTER (proposed):
status = iree_hal_hip_device_stream_signal_semaphores_and_add_cleanup(…);
if (device->uses_external_stream) {
    iree_hal_hip_set_external_stream_data_completed(…);  // caller wakes
}
```

Three locations need this reorder:
1. `iree_hal_hip_device_execute_now` — lines ~2467-2480
2. `iree_hal_hip_device_perform_queue_read_now` — lines ~2124-2126
3. Queue-write equivalent (same pattern)

This ensures that when the caller's `iree_hal_semaphore_wait` runs, the HIP event is already in the semaphore's event tree and `max_value_to_be_signaled` has been advanced, so the `hipEventSynchronize` fast-path is taken immediately.

## Additional note

There is currently no CTS test coverage for the external stream code path. All existing HIP CTS variants use internally-managed streams.

## Environment

- IREE runtime built from source (pinned nightly)
- AMD Instinct MI300X (gfx942)
- ROCm / HIP 6.x
- Failures observed on GitHub Actions shared runners (Ubuntu, AMD GPU)

AWoloszyn:
There was an issue with the previous fix that we would not retain the data properly.

### DIFF
--- a/runtime/src/iree/hal/drivers/hip/hip_device.c
+++ b/runtime/src/iree/hal/drivers/hip/hip_device.c
@@ -1558,10 +1558,10 @@ static iree_status_t iree_hal_hip_device_perform_buffer_operation_now(
     }
   }
   IREE_TRACE_ZONE_END(z3);
-
+  iree_hal_hip_dispatch_completed_data_t* external_stream_dispatch_data = NULL;
   if (device->uses_external_stream) {
-    iree_hal_hip_set_external_stream_data_completed(
-        data->base.external_stream_dispatch_data);
+    external_stream_dispatch_data = data->base.external_stream_dispatch_data;
+    iree_hal_resource_retain(external_stream_dispatch_data);
   }
 
   const iree_hal_hip_dynamic_symbols_t* symbols = device->hip_symbols;
@@ -1579,6 +1579,12 @@ static iree_status_t iree_hal_hip_device_perform_buffer_operation_now(
                               iree_status_clone(status));
     }
     iree_hal_hip_device_destroy_buffer_callback_data(data);
+  }
+
+  if (external_stream_dispatch_data) {
+    iree_hal_hip_set_external_stream_data_completed(
+        external_stream_dispatch_data);
+    iree_hal_resource_release(external_stream_dispatch_data);
   }
 
   IREE_TRACE_ZONE_END(z0);
@@ -2053,6 +2059,11 @@ static iree_status_t iree_hal_hip_device_perform_queue_read_now(
         device, device->devices[device_ordinal].hip_dispatch_stream,
         data->base.wait_semaphore_list, device_ordinal);
   }
+  iree_hal_hip_dispatch_completed_data_t* external_stream_dispatch_data = NULL;
+  if (device->uses_external_stream) {
+    external_stream_dispatch_data = data->base.external_stream_dispatch_data;
+    iree_hal_resource_retain(external_stream_dispatch_data);
+  }
 
   const iree_hal_hip_dynamic_symbols_t* symbols = device->hip_symbols;
   iree_device_size_t amount_left = data->length;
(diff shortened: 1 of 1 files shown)


sjain-stanford (approved):
Thanks @AWoloszyn . I'm just stamping this one as the main change to reorder `iree_hal_hip_set_external_stream_data_completed` was reviewed/approved by @benvanik in https://github.com/iree-org/iree/pull/23583, and this just has a few extra fixes to keep ASAN happy.

PR #23600, opened 2026-03-01
[Python] Add Stable ABI (abi3) support for Python 3.12+

kuhar:
Build Python bindings against the Limited C API (`Py_LIMITED_API`, PEP 384) so that a single wheel built with Python 3.12 works on all future CPython versions. This reduces the total release artifact size.

The Limited API forbids direct access to CPython struct layouts and most convenience macros. The main impact is replacing the NumPy C API (whose headers use non-limited internals) with Python-level equivalents like `numpy.frombuffer()` and `numpy.dtype()`, and using nanobind's `type_slots()` to register buffer/sequence/mapping protocols instead of casting to `PyHeapTypeObject`.

Limitations:
- Only CPython 3.12+. Per-version wheels are still built for 3.10-3.11.
- Free-threaded builds (3.13t) are excluded: the free-threaded ABI is not yet stable.
- `Development.SABIModule` requires CMake 3.26+ (IREE minimum is 3.21), so `IREE_ENABLE_PYTHON_STABLE_ABI` is OFF by default. setup.py auto-enables it when building with CPython 3.12+. May want to bump it in the future.

The existing release workflow already builds with Python 3.12 on all platforms, so abi3 wheels are produced with no workflow changes. pkgci only builds cp311 by default (no abi3).

The LLVM submodule includes cherry-picks of llvm/llvm-project PRs for MLIR-side abi3 support.

Assisted-by: claude

### DIFF
--- a/runtime/bindings/python/binding.h
+++ b/runtime/bindings/python/binding.h
@@ -58,7 +58,15 @@ class ApiRefCounted {
   }
   void operator=(const ApiRefCounted&) = delete;
 
-  ~ApiRefCounted() { Release(); }
+  ~ApiRefCounted() {
+    // In stable ABI (abi3) mode, types are heap-allocated via PyType_FromSpec
+    // and instances may be destroyed during Py_FinalizeEx after IREE's type
+    // registry is no longer valid. Skip release when nanobind's internals
+    // are being torn down since the process is exiting anyway.
+    if (instance_ && py::is_alive()) {
+      Release();
+    }
+  }
 
   // Steals the reference to the object referenced by the given raw pointer and
   // returns a wrapper (transfers ownership).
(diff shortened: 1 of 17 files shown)


runtime/bindings/python/binding.h
```diff
@@ -130,37 +138,39 @@ inline py::object create_empty_tuple() {
...
+py::type_slots buffer_protocol_slots() {
+  // It is not legal to raise exceptions from buffer protocol callbacks.
+  static const PyType_Slot slots[] = {
+      {Py_bf_getbuffer,
+       reinterpret_cast<void*>(
+           +[](PyObject* raw_self, Py_buffer* view, int flags) -> int {
+             if (!view) {
+               PyErr_SetString(PyExc_ValueError, "NULL view in getbuffer");
+               return -1;
+             }
```
amd-eochoalo, line 155: Small question: It says it is not legal to raise exceptions from these callbacks. But [PyErr_SetString](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_SetString) is raising an exception, right? Can you clarify a little bit more the comment? At the moment it gives me the impression that something wrong is being done here but maybe I am misreading it.

runtime/bindings/python/invoke.cc
```diff
@@ -21,6 +21,23 @@ namespace python {
 
 namespace {
 
+// nanobind::python_error::what() is noexcept but under the limited API
+// (Py_LIMITED_API / abi3) it uses a code path that can internally throw,
```
amd-eochoalo, line 25: This seems like a bug in nanobind. Should this be raised upstream?

amd-eochoalo (approved):
Thanks! I think this looks good. Just a couple of questions.

> Free-threaded builds (3.13t) are excluded: the free-threaded ABI is not yet stable.

Do we have ifdef pragmas targetting 3.13t directly? 

> The LLVM submodule includes cherry-picks of llvm/llvm-project PRs
for MLIR-side abi3 support.

Probably worth reaching out to this week's gardener to coordinate.

runtime/bindings/python/binding.h
kuhar: This is a pre-existing comment

kuhar:
> Probably worth reaching out to this week's gardener to coordinate.

The next integrate will have to pick up both (either directly or cherry-pick). cc @RattataKing (this week's gardener)

kuhar:
> Do we have ifdef pragmas targetting 3.13t directly?

These don't need special ifdefs. From the API perspective, it's either limited or not, controlled by the cmake flag. In the future, we will opt into stable abi for 3.15t+

kuhar:
I'm going to merge now -- I want to see if anything breaks and I'd rather know that well ahead of the next release. If something breaks, we can revert.

PR #23606, opened 2026-03-01
[CI] Add typos pre-commit hook and dictionary. (6/6)

kuhar:
Configure crate-ci/typos as a pre-commit spell checker. Add a dictionary of domain-specific terms (HSA, SME, OLT, OFR, etc.) to suppress false positives. Exclude tokenizer test files that contain intentional subword fragments.

### DIFF
--- a/build_tools/linters/typos.toml
+++ b/build_tools/linters/typos.toml
@@ -1,5 +1,8 @@
-# Configuration file for the typos-rs spell checker used by pre-commit.
-# See https://crates.io/crates/typos for tool usage details.
+# Configuration file for the typos spell checker used by pre-commit.
+# https://github.com/crate-ci/typos
+#
+# To run manually from the repo root:
+#   typos --config build_tools/linters/typos.toml
 
 [files]
 extend-exclude = [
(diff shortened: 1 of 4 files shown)


kuhar:
This needs to be rebased and landed after the 5 previous PRs.

bangtianliu (approved):
Thanks

amd-eochoalo (approved):

build_tools/linters/typos.toml
ScottTodd, line 1: 1. Can you add a comment explaining what tool is using this file, with some syntax reference?
2. Does the file need to go in the repository root, or could it go in a subdirectory like `build_tools/`? (seems general enough to go in the root, but I usually ask the same question for every new file here)

kuhar: 1. Done
2. Done, moved to `build_tools/linters/typos.toml`

ScottTodd (approved):
LGTM once pre-commit runs succeed with no typos. Thanks for addressing my comments.

kuhar:
> LGTM once pre-commit runs succeed with no typos. Thanks for addressing my comments.

I think we will have to undo the move to build_tools -- it doesn't properly support `../` or `**/` when placed in a subdir

kuhar:
OK, I think `--force-excludes` makes this work when invoked through pre-commit

PR #23607, opened 2026-03-01
[CMake][Python] Bump minimum CMake version to 3.26

kuhar:
CMake 3.26 added FindPython's `Development.SABIModule` component, which is required for building Stable ABI (abi3). With this minimum version, we can unconditionally request `SABIModule` upfront in the Python `find_package` calls, before deciding whether to enable abi3. This is appealing because the SABI module must be found early and cannot be deferred to after the abi3 option is resolved; otherwise we need the user to decide through cmake flags.

CMake 3.26 availability:
  pip install cmake: 4.2.3
  Homebrew: 4.2.3
  Chocolatey (Windows CI): 4.2.3
  Ubuntu 24.04 (apt): 3.28.3
  Ubuntu 22.04 (apt): 3.22.1 (use pip or Kitware APT repo)
  Debian 12 bookworm (apt): 3.25.1 (3.31.6 in backports)
  
ci-extra: all

### DIFF
--- a/.github/workflows/build_clang_tidy.yml
+++ b/.github/workflows/build_clang_tidy.yml
@@ -42,6 +42,8 @@ jobs:
         with:
           submodules: true
 
+      - name: Install Python requirements
+        run: python3 -m pip install -r ./build_tools/github_actions/ci_requirements.txt
       - name: "Setting up CMake"
         run: |
           source ./build_tools/cmake/setup_build.sh
--- a/.github/workflows/ci_linux_arm64_clang.yml
+++ b/.github/workflows/ci_linux_arm64_clang.yml
@@ -32,7 +32,9 @@ jobs:
         with:
           submodules: true
       - name: Install Python requirements
-        run: python3 -m pip install -r ./runtime/bindings/python/iree/runtime/build_requirements.txt
+        run: |
+          python3 -m pip install -r ./runtime/bindings/python/iree/runtime/build_requirements.txt
+          python3 -m pip install -r ./build_tools/github_actions/ci_requirements.txt
       # TODO(#18557): enable sccache (watching for network bandwidth charges
       #               between the cache server and AWS VMs)
       - name: CMake - configure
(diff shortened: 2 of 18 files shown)


amd-eochoalo (approved):
Thanks! 

[CMake 3.31.2 fixed a bug related to SABIModule](https://discourse.cmake.org/t/cmake-3-31-2-available-for-download/13135)

> FindPython: Ensure SABIModule is found in all cases

I looked a bit further and I think it may [be this one](https://gitlab.kitware.com/cmake/cmake/-/issues/26493) which relates to cross compiling on Windows. I don't believe we need this patch, just making a note that this issue exist and was solved in version 3.31.2 .

kuhar:
> I looked a bit further and I think it may [be this one](https://gitlab.kitware.com/cmake/cmake/-/issues/26493) which relates to cross compiling on Windows. I don't believe we need this patch, just making a note that this issue exist and was solved in version 3.31.2 .

> Hi, I'm having issues using FindPython on Windows. When searching for both the Development.Module and Development.SABIModule components (without the Interpreter component), FindPython fails. When searching for just one of the two, it works correctly. Please find a minimal example below.

I think we should be fine because we always request the interpreter component: https://github.com/iree-org/iree/blob/a21fa38cf60715e8d9ed53a55876726555b58450/CMakeLists.txt#L748

Thanks for a thorough review BTW, I a wasn't aware of these potential teething issues.

kuhar:
We need to update cmake on `'shark10-ci-2'`

kuhar:
> We need to update cmake on `'shark10-ci-2'`

Done, as simple as:
```shell
sudo apt remove cmake -yqq
pip install --user cmake
```

kuhar:
Extra data points:
* Pytorch requires cmake 3.27 https://github.com/pytorch/pytorch/blob/main/requirements-build.txt
* TheRock requires cmake 3.27 https://github.com/ROCm/TheRock/blob/23a5517fa8feb5f09b4506c66fedf593cfc62197/dockerfiles/build_manylinux_x86_64.Dockerfile#L37-L41

kuhar:
Nobody borough up concerns and the CI passed, merging

PR #23610, opened 2026-03-02
[Codegen][VectorExt] Add VectorExt::TransferScatter

keshavvinayak01:
Adds `iree_vector_ext.transfer_scatter`, the write counterpart to `transfer_gather`, using the same unified `indexing_maps` attribute. For tensor operands, the operation returns the modified tensor; for memref operands, it has no result.

Example — scatter write: writing values into scattered rows of a 2D dest:
```
// dest[indices[i], j] = vector[i, j]
  %result = iree_vector_ext.transfer_scatter %vector into %dest[%c0, %c0]
    [%indices : vector<16xindex>] {
      indexing_maps = [
        affine_map<(d0, d1)[s0] -> (s0, d1)>,
        affine_map<(d0, d1)[s0] -> (d0)>
      ]
    } : vector<16x8xf16>, tensor<4096x8xf16> -> tensor<4096x8xf16>
```

Adds a lot of templatizing for shared code logic of TransferGather/TransferScatter, and missing tests in `vector_lowering.mlir`


hanhanW (changes requested):
It's good to have a single PR that demonstrate the impact, but you should really break it into a few small PRs. I did not review any details, and I think we can at least break it into something like

- Introduce the operation including roundtrip/invalid tests.
- A PR per interface implementation.
- Unrolling support

Please read below links for guidance:
- https://google.github.io/eng-practices/review/developer/small-cls.html
- https://google.github.io/eng-practices/review/developer/cl-descriptions.html

Groverkss:
Nice! I think splitting PR as Hanhan suggested is a good idea.

PR #23611, opened 2026-03-02
[VectorDistribute] Refactor VectorLayoutAnalysis into 2-phase forward/backward design

Groverkss:
Restructure the layout analysis from an interleaved forward+backward worklist into a clean two-phase design:

Phase 1 (forward): Multi-candidate propagation from ToLayoutOp anchors through uses. No IR mutation.

Resolve: Pick first candidate per value. This is a placeholder cost model for now which matches the old analysis. Eventually, we will consider coalescing, compute ops, mma layout, etc.

Phase 2 (backward fixup): Walk operations in reverse program order via recursive fixupRegion/fixupOp. For each op, derive operand layouts from resolved result layouts. Assign missing layouts, clone cheap ops (constants, create_mask, step), or insert to_layout conversions on conflict.

This naturally handles conflicts better in a predictable manner. // The forward analysis is the main driver of the analysis. The reason for this is that for a program to be well-formed for vector distribution, there must be some way for the final store/return to get a layout. Otherwise, there is not enough information in the program to determine how distribution should be done. The forward analysis ensures that the final return/store gets a layout in a well-formed program. The rest of the program can get their layouts from backward propagation, everything in the program must eventually reach the store/return.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -46,13 +46,13 @@ static constexpr int kMaxCandidatesPerValue = 4;
 //   conversions.
 //
 // The forward analysis is the main driver of the analysis. The reason for this
-// is that generally on GPUs you will always have something that will
-// assign a layout to the loads. If you don't, then your program is
-// ill-formed for vector distribute anyway. Then either the rest of the
-// program gets it's layout from the load layout, or it gets it's layout
-// from whatever backpropagated the layout to the load. We don't want to
-// handle other kinds of programs, because they are ill-formed and should
-// have had better layout setting on them.
+// is that for a program to be well-formed for vector distribution, there must
+// be some way for the final store/return to get a layout. Otherwise, there
+// is not enough information in the program to determine how distribution should
+// be done. The forward analysis ensures that the final return/store gets a
+// layout in a well-formed program. The rest of the program can get their
+// layouts from backward propagation, everything in the program must eventually
+// reach the store/return.
 //===----------------------------------------------------------------------===//
 
 namespace {
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -237,9 +237,6 @@ void LayoutAnalysis::propagateOneForward(Value val,
         continue;
       }
       if (contract.getLhs() == val || contract.getRhs() == val) {
-        if (contract->hasAttr("iree.gpu.mma")) {
-          continue;
-        }
         if (auto maskOp = dyn_cast<vector::MaskOp>(contract->getParentOp())) {
           AffineMap map = contract.getMatchingIndexingMap(&use);
           if (map.isPermutation()) {
@@ -315,7 +312,7 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // transfer_write: vector operand layout -> derive mask layout.
   if (auto write = dyn_cast<vector::TransferWriteOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(write.getVector());
-    if (!layout || !write.getMask()) {
+    if (!write.getMask()) {
       return;
     }
     AffineMap maskMap =
@@ -327,7 +324,7 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // transfer_read: result layout -> derive mask layout.
   if (auto read = dyn_cast<vector::TransferReadOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(read.getResult());
-    if (!layout || !read.getMask()) {
+    if (!read.getMask()) {
       return;
     }
     AffineMap maskMap =
@@ -339,9 +336,6 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // elementwise: result layout -> all operands get same layout.
   if (OpTrait::hasElementwiseMappableTraits(op)) {
     VectorLayoutInterface layout = getResolvedLayout(op->getResult(0));
-    if (!layout) {
-      return;
-    }
     for (OpOperand &operand : op->getOpOperands()) {
       setLayoutOrClone(&operand, layout);
     }
@@ -351,9 +345,6 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // to_layout: result layout -> input gets same layout.
   if (auto toLayout = dyn_cast<ToLayoutOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(toLayout.getResult());
-    if (!layout) {
-      return;
-    }
     setLayoutOrClone(&toLayout.getInputMutable(), layout);
     return;
   }
@@ -361,9 +352,6 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // multi_dim_reduction: result layout -> acc gets same layout.
   if (auto multiReduce = dyn_cast<vector::MultiDimReductionOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(multiReduce.getResult());
-    if (!layout) {
-      return;
-    }
     setLayoutOrClone(&multiReduce.getAccMutable(), layout);
     return;
   }
@@ -371,9 +359,6 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // transpose: result layout -> input gets inverse-permuted layout.
   if (auto transpose = dyn_cast<vector::TransposeOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(transpose.getResult());
-    if (!layout) {
-      return;
-    }
     setLayoutOrClone(
         &transpose.getVectorMutable(),
         layout.permute(invertPermutationVector(transpose.getPermutation())));
@@ -383,7 +368,7 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // broadcast: result layout -> source gets projected layout.
   if (auto broadcast = dyn_cast<vector::BroadcastOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(broadcast.getResult());
-    if (!layout || !isa<VectorType>(broadcast.getSourceType())) {
+    if (!isa<VectorType>(broadcast.getSourceType())) {
       return;
     }
     assert(broadcast.computeBroadcastedUnitDims().empty() &&
@@ -402,9 +387,6 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // contract: result layout -> acc gets same layout.
   if (auto contract = dyn_cast<vector::ContractionOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(contract.getResult());
-    if (!layout) {
-      return;
-    }
     setLayoutOrClone(&contract.getAccMutable(), layout);
     return;
   }
@@ -412,9 +394,6 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // gather: result layout -> indices, mask, passthru get same layout.
   if (auto gather = dyn_cast<vector::GatherOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(gather.getResult());
-    if (!layout) {
-      return;
-    }
     setLayoutOrClone(&gather.getIndicesMutable(), layout);
     setLayoutOrClone(&gather.getMaskMutable(), layout);
     setLayoutOrClone(&gather.getPassThruMutable(), layout);
@@ -424,9 +403,6 @@ void LayoutAnalysis::fixupOp(Operation *op) {
   // transfer_gather: result layout -> index vecs + mask get projected layouts.
   if (auto gather = dyn_cast<TransferGatherOp>(op)) {
     VectorLayoutInterface layout = getResolvedLayout(gather.getResult());
-    if (!layout) {
-      return;
-    }
     SmallVector<AffineMap> maps = gather.getIndexingMapsArray();
     int64_t numIndexVecs = gather.getIndexVecs().size();
     for (auto [i, operand] : llvm::enumerate(gather.getIndexVecsMutable())) {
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -25,50 +26,149 @@ namespace mlir::iree_compiler {
...
-      return;
-    }
-    layouts[val] = layout;
-    forward.push(val);
-    backward.push(val);
+/// Maximum number of candidate layouts tracked per value. Kept small to bound
+/// analysis cost; most values see 1-2 candidates in practice.
+static constexpr int kMaxCandidatesPerValue = 4;
+
+//===----------------------------------------------------------------------===//
```
sommerlukas, line 33: Should we merge this documentation with the existing one in `Transforms.h` or replace the old one entirely?

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -25,50 +26,149 @@ namespace mlir::iree_compiler {
...
+// Layout Analysis
+//
+// Phase 1: Forward propagation with multi-candidate tracking. Seeds from
```
sommerlukas, line 36: Are we using a specific topological order or similar here? Even if we don't, might be worth documenting this fact here.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -25,50 +26,149 @@ namespace mlir::iree_compiler {
...
+//
+// Phase 2: Backward fixup. Walks operations in reverse program order. For each
+//   op, determines operand layouts from resolved result/operand layouts.
+//   Assigns missing layouts, clones cheap ops, or inserts to_layout
+//   conversions.
+//
+// The forward analysis is the main driver of the analysis. The reason for this
+// is that generally on GPUs you will always have something that will
+// assign a layout to the loads. If you don't, then your program is
+// ill-formed for vector distribute anyway. Then either the rest of the
```
sommerlukas: Is such a program necessarily ill-formed? For example, an element-wise operation with a layout on one operand isn't necessarily ill-formed, as it can be trivial to determine the layout for the other operand/load.

This is probably a question of how much logic we want to put into lowering config selection and how much we want to propagate later on.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -25,50 +26,149 @@ namespace mlir::iree_compiler {
...
+
+//===----------------------------------------------------------------------===//
+// Phase 1: Forward Propagation (no IR mutation)
+//===----------------------------------------------------------------------===//
+
+/// Add a candidate layout for a value. Returns true if a new candidate was
+/// added. Schedules the value for forward propagation.
+bool LayoutAnalysis::addCandidate(Value val, VectorLayoutInterface layout) {
+  if (!layout) {
+    return false;
```
sommerlukas, line 104: Under which circumstances could we end up here without a layout? Should we document that here?

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -110,139 +210,184 @@ void LayoutInfo::propagateLayoutForward(Value val) {
...
-        setLayoutIfUnset(contract.getResult(), layout);
+        addCandidate(contract.getResult(), layout);
         continue;
       }
       if (contract.getLhs() == val || contract.getRhs() == val) {
         if (contract->hasAttr("iree.gpu.mma")) {
-          // Intrinsic ops have fixed layouts, do not try to infer them through
-          // maps.
-          // TODO: Move to iree_gpu.multi_mma ops.
           continue;
```
sommerlukas: Returning early here means that we never set a layout for the mask if this `vector.contract` is wrapped inside a `vector.mask`, but also has the `iree.gpu.mma` attribute. 

I had changed that in this commit on my prototype branch (but ended up not needing it, because I unwrap `vector.mask` before layout analysis now): https://github.com/sommerlukas/iree/commit/f2e434a9997ee5d6d18e5028c358b784bbe86098

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -110,139 +210,184 @@ void LayoutInfo::propagateLayoutForward(Value val) {
...
-      OpOperand *yielded = forOp.getTiedLoopYieldedValue(blockArg);
-      OpOperand *init = forOp.getTiedLoopInit(blockArg);
-      setLayoutOrClone(yielded, layout);
-      setLayoutOrClone(init, layout);
+/// Fix up operand layouts for a single operation. Result layouts are fixed
+/// (from resolve); this determines what operand layouts should be.
+void LayoutAnalysis::fixupOp(Operation *op) {
+  // transfer_write: vector operand layout -> derive mask layout.
+  if (auto write = dyn_cast<vector::TransferWriteOp>(op)) {
+    VectorLayoutInterface layout = getResolvedLayout(write.getVector());
```
sommerlukas, line 314: Why do we derive the mask layout here in the backward part, whereas we derive mask layouts for other operations in the forward part?

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -110,139 +210,184 @@ void LayoutInfo::propagateLayoutForward(Value val) {
...
 
-  Operation *defOp = val.getDefiningOp();
-  if (OpTrait::hasElementwiseMappableTraits(defOp)) {
-    for (OpOperand &operand : defOp->getOpOperands()) {
+  // elementwise: result layout -> all operands get same layout.
+  if (OpTrait::hasElementwiseMappableTraits(op)) {
+    VectorLayoutInterface layout = getResolvedLayout(op->getResult(0));
+    if (!layout) {
+      return;
+    }
```
sommerlukas: We have this check quite often; would it make sense to pull it into `setLayoutOrClone` or some helper wrapping it?

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -110,139 +210,184 @@ void LayoutInfo::propagateLayoutForward(Value val) {
...
     return;
   }
 
-  if (auto toLayout = dyn_cast<ToLayoutOp>(defOp)) {
+  // to_layout: result layout -> input gets same layout.
+  if (auto toLayout = dyn_cast<ToLayoutOp>(op)) {
+    VectorLayoutInterface layout = getResolvedLayout(toLayout.getResult());
+    if (!layout) {
+      return;
+    }
```
sommerlukas: Should we stop on shared memory conversions here? Something like this commit: https://github.com/sommerlukas/iree/commit/9225a655fe7e86edbbdca046f6fa391cf01446c1

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -304,28 +446,66 @@ void LayoutInfo::propagateLayoutBackward(Value val) {
...
+  if (getResolvedLayout(val->get()) == layout) {
+    return;
+  }
+
+  // Different layout -- clone cheap ops or insert to_layout conversion.
   OpBuilder b(val->getOwner());
   if (Operation *defOp = val->get().getDefiningOp()) {
+    // Clone constant-like and duplicatable ops per use site.
     bool isConstantLike = defOp->hasTrait<OpTrait::ConstantLike>();
     bool isDuplicatable =
```
sommerlukas, line 480: Should we handle `tensor.empty` here, too?

compiler/src/iree/compiler/Codegen/Common/test/vector_layout_analysis.mlir
```diff
@@ -809,3 +812,122 @@ func.func @propagate_1D_reshape_contract(%arg0: memref<4x32xf16>) -> vector<128x
...
+>
+
+func.func @clone_shared_mask_on_layout_conflict(
+    %arr: memref<16x16xf16>,
+    %a: vector<16x16xf16>,
+    %b: vector<16x16xf16>) {
+  %c0 = arith.constant 0 : index
+  %c12 = arith.constant 12 : index
+  // The mask gets cloned per transfer_write use site, each with the
+  // correct layout derived from the written value's layout.
```
sommerlukas, line 924: Should we add `FileCheck` to this test to make sure the `create_mask` is actually duplicated?

sommerlukas (commented):
Mainly questions and small comments.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: I think I'm going to remove some of the redundant documentation comments from the Transforms.h file. We should keep the implementation details in VectorLayoutAnalysis though.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: Since we collect all layouts, the order doesn't matter right?

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: If the layout on the elementwise op doesn't back propagate to a load, then it is ill formed

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: we just unconditionally do an addCandidate on ops even if they don't have a layout. This is just a matter of cleaner code so we don't have to guard everywhere in propagation.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: That's true, we actually don't need to return on this anymore. I  deleted all code that relied on this path, so I can just remove this if condition as well :P

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
sommerlukas: Yeah, that was my point: I wouldn't call a program ill-formed if we can infer layouts through forward and backward propagation.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
sommerlukas: Yeah, walk-order is fine here. I didn't mean to say the order is wrong, but as we're explicitly stating the order for the backward direction, I thought it might be worth documenting that we're simply using walk-order for future readers of the documentation.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: done

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: When we have a vector.mask wrapped around an operation, we are essentially forwarding it to the vector.yield operation inside that region, so it is real forwarding. vector.transfer_write / vector.transfer_read aren't wrapped in a region, and we need to propagate it back to the operand. Which is why this is slightly different.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: Good idea

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: I have this for my next commit where I will do it. Doesn't make sense to do it as part of this commit because it's never used with it, but next commit will do it.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: I don't think we should do that today honestly. We are currently only operating on vectors. Eventually, yes.

compiler/src/iree/compiler/Codegen/Common/test/vector_layout_analysis.mlir
Groverkss: I think the diagnostic should be enough, there cannot be any other operation that can produce the remark above here.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: For forward propagation, we do the propagation on values till fixed point. The order doesn't matter, because it is doing it till fixed point. You should be able to walk it in any order.

For reverse order, we are not doing it till fixed point, we are walking the operations backward and fixing up the layouts their operands needs. The order here actually matters.

This is why we specify the order for backward but not for forward, because one is value based other is operation based.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
Groverkss: Let me make it more clear in the docs.

Groverkss: done

keshavvinayak01 (approved):
LGTM, concerns addressed by Lukas.

sommerlukas (approved):
LGTM, thanks for the fixes!

PR #23613, opened 2026-03-02
[VectorDistribute] Overhaul mask distribution

sommerlukas:
So far, vector distribution of mask operations (`create_mask/constant_mask`) created a selection tree.

This PR overhauls the distribution of mask operations. Instead of the selection tree, a combination of `vector.step` and comparison with bounds provided in the masks is used to calculate the mask. For multi-dimensional vectors, the masks are calculated per dimension and then combined with boolean operations.

The actual distribution happens through distribution of the operations generated as part of the new pattern (`vector.step, ...).

A detailed explanation of the approach can be found on the documentation of the pattern.

Assisted-by: Claude Code

ci-extra: test_torch

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -1937,6 +1937,15 @@ SmallVector<Value> createDistributedMaskBounds(PatternRewriter &rewriter,
     auto distrUpperBound =
         arith::AddIOp::create(rewriter, loc, linearizedLastValidIdx, one);
 
+    // For threads past the boundary thread (tid > u3), all iterations in
+    // batches before the boundary batch are still valid. The count of such
+    // elements is linearize(u1, u2, 0).
+    auto distrUpperBoundPostThreads = affine::AffineLinearizeIndexOp::create(
+        rewriter, loc,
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1949,7 +1958,7 @@ SmallVector<Value> createDistributedMaskBounds(PatternRewriter &rewriter,
     //   if tid < u3:
     //     [u1][u2][max]
     //   if tid > u3:
-    //     all invalid.
+    //     [u1][u2][0]
```
efric: personal nit: can we update the surrounding comment chain to expound briefly on how `u0, u1, u2, .. == subgroup_tile, batch_tile, outer_tile, etc ` (iiuc)? I think it would be useful context for readers

efric (approved):
lgtm

sommerlukas:
After feedback from @Groverkss, I've completely overhauled how mask operations are distributed instead of fixing the existing selection tree approach. The tile and PR description have been updated accordingly.

sommerlukas:
Converted this to draft for now, as I've integrated the decomposition patterns from this PR into the new pass in https://github.com/iree-org/iree/pull/23679 to run earlier in the pipeline.

sommerlukas:
The decomposition of `create_mask/constant_mask` into `vector.step` + comparison was landed as an early masking materialization pattern as part of https://github.com/iree-org/iree/pull/23679. Closing this PR.

PR #13038, opened 2023-04-12
Add support for computing dispatch workgroup count using program slices.

MaheshRavishankar:
This PR changes the way the workgroup count calculation for dispatches are handled by default.

Previous approaches required to have context of what is put into a dispatch and what values are captured as workload. Managing this implicit link gets tricky. The approach in this PR assumes that all the information required for computing the number of workgroups is captured within the dispatch after the `flow.dispatch.workgroups` op is formed. In the backend, during tile and distribute a program slice is used to capture the computation that determines the number of workgroups, and is cloned into the workgroup count region.

One complication is that the signature of the `flow.dispatch.workgroups` operation is not preserved through the compilation pipeline. So in the backends it is not possible to connect the leaves of the slice back to the operands of the original `flow.dispatch.workgroups` operation that were captured as workload. To get around this, these operands are annotated within the dispatch with their position in the workload list.

Another side-effect of this change is that there is no need to instantiate `tensor.dim` operations to compute the workload. In dynamic shape cases, this leads to repeated instantiations of the same `tensor.dim` value leading to compile-time explosion even though these get CSE-ed away later.

To address the transform dialect path a new op `transform.iree.populate_workgroup_count_region_using_num_threads_slice` is s added that handles the workgroup count materialization for the transform dialect path. As a consequence
- `transform.iree.tile_to_forall_and_workgroup_count_region` is deprecated in favor of using `transform.structured.tile_to_forall`      and the newly added transform dialect operation above.
- `transform.iree.convert_conv2d_to_img2col_and_adjust_workgroup_count_region` is also deprecated.
    

Fixes #11608

PR #23615, opened 2026-03-02
[VMVX] Enable scf.forall distribution for VMVX pipelines.

### LINKED ISSUE #23588: VMVX forall distribution fails with data tiling + ukernels
## Summary

Enabling `scf.forall`-based workgroup distribution for VMVX fails when `--iree-opt-data-tiling` and `--iree-vmvx-enable-microkernels` are used together.

## Repro

```mlir
// /tmp/vmvx_matmul.mlir
util.func @matmul_dyn(%lhs: tensor<?x?xf32>, %rhs: tensor<?x?xf32>, %acc: tensor<?x?xf32>) -> tensor<?x?xf32> {
  %result = linalg.matmul ins(%lhs, %rhs: tensor<?x?xf32>, tensor<?x?xf32>) outs(%acc: tensor<?x?xf32>) -> tensor<?x?xf32>
  util.return %result: tensor<?x?xf32>
}
```

Apply the patch:

```patch
diff --git a/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp b/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
index c91569f043..b699e5bb8e 100644
--- a/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
@@ -34,11 +34,24 @@ static llvm::cl::opt<bool> clEnableUKernelsDecomposeLinalgGeneric(
                    "ukernels are enabled (experimental)"),
     llvm::cl::init(true));
 
+static llvm::cl::opt<bool> clTileDispatchUsingForall(
+    "iree-vmvx-tile-dispatch-using-forall",
+    llvm::cl::desc(
+        "Enable tile and distribute to workgroups using scf.forall"),
+    llvm::cl::init(true));
+
 static void addTileAndDistributePasses(OpPassManager &funcPassManager) {
-  funcPassManager.addPass(createTileAndDistributeToWorkgroupsPass());
-  funcPassManager.addPass(createCSEPass());
-  funcPassManager.addPass(createConvertToDestinationPassingStylePass());
-  funcPassManager.addPass(createFoldAffineMinInDistributedLoopsPass());
+  if (clTileDispatchUsingForall) {
+    funcPassManager.addPass(
+        createTileAndDistributeToWorkgroupsUsingForallOpPass());
+    funcPassManager.addPass(createBufferizeDispatchTensorLoadStorePass());
+    funcPassManager.addPass(createCombineResultLayoutTransformationPass());
+  } else {
+    funcPassManager.addPass(createTileAndDistributeToWorkgroupsPass());
+    funcPassManager.addPass(createCSEPass());
+    funcPassManager.addPass(createConvertToDestinationPassingStylePass());
+    funcPassManager.addPass(createFoldAffineMinInDistributedLoopsPass());
+  }
   funcPassManager.addPass(createConfigTrackingCanonicalizerPass());
   funcPassManager.addPass(createCSEPass());
   funcPassManager.addPass(createFuseTensorPadWithConsumerPass());
```

And run

```bash
iree-compile /tmp/vmvx_matmul.mlir \
  --output-format=vm-bytecode \
  --iree-hal-target-backends=vmvx \
  --iree-vmvx-enable-microkernels \
  --iree-opt-data-tiling \
  -o /dev/null
```

Error: `'iree_codegen.workgroup_count_hint' op failed to resolve workgroup count hint in terms of workload ordinals`

Without data tiling, or without ukernels, it works fine. Elementwise ops also work fine.

## Root Cause

With data tiling, encoding dispatches contain `linalg.pack` ops whose outer dimensions depend on `iree_codegen.query_tile_sizes` (a `Pure` op that queries VMVX runtime tile sizes). When forall-based distribution tiles the pack, the `scf.forall` upper bounds carry this dependency.

`ReconcileTranslationInfoPass` creates a `workgroup_count_hint` from the forall's upper bounds. `ResolveWorkgroupCountHintsPass` then tries to trace the hint back to workload ordinals, but the backward slice hits `call @vmvx.query_tile_sizes.2d` (the lowered form of `iree_codegen.query_tile_sizes`) and rejects it because `func.call` is not memory-effect-free (`ResolveWorkgroupCountHints.cpp:158`).

The core issue is that `iree_codegen.query_tile_sizes` is `Pure`, but by the time `ResolveWorkgroupCountHintsPass` runs, it has already been lowered to a `func.call` by `LowerUKernelOpsToCallsPass`, losing the purity information.

The old `scf.for` path avoids this entirely because `TileAndDistributeToWorkgroupsPass` materializes the workgroup count directly in the export's count region at tiling time.

hanhanW:
The iree_codegen.query_tile_sizes op was resolved during scf.for distribution. If we go with scf.forall distribution, it becomes `iree_codegen.ukernel.generic` before reconcile+resolve runs. Thus, it requires a workaround similar to the existing one.

Further more, LowerUKernelOpsToCallsPass would lower it to a func.call which is memory-effecting and would be rejected by backward slice analysis. So we put the reconcile+resolve in between, which is the stage that we have enough information.

- Move reconcile+resolve before LowerUKernelOpsToCallsPass in the VMVX variant pipeline.
- Extend the WAR hack in materializeSliceFromOrdinals (TODO #13038) to also handle the lowered UKernelGenericOp form of query_tile_sizes, replacing it with a constant 16 in the workgroup count region.
- Disable the distribution if it identifies that the distribution is already applied.

Fixes https://github.com/iree-org/iree/issues/23588

### DIFF
--- a/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
@@ -34,23 +34,11 @@ static llvm::cl::opt<bool> clEnableUKernelsDecomposeLinalgGeneric(
                    "ukernels are enabled (experimental)"),
     llvm::cl::init(true));
 
-static llvm::cl::opt<bool> clTileDispatchUsingForall(
-    "iree-vmvx-tile-dispatch-using-forall",
-    llvm::cl::desc("Enable tile and distribute to workgroups using scf.forall"),
-    llvm::cl::init(true));
-
 static void addTileAndDistributePasses(OpPassManager &funcPassManager) {
-  if (clTileDispatchUsingForall) {
-    funcPassManager.addPass(
-        createTileAndDistributeToWorkgroupsUsingForallOpPass());
-    funcPassManager.addPass(createBufferizeDispatchTensorLoadStorePass());
-    funcPassManager.addPass(createCombineResultLayoutTransformationPass());
-  } else {
-    funcPassManager.addPass(createTileAndDistributeToWorkgroupsPass());
-    funcPassManager.addPass(createCSEPass());
-    funcPassManager.addPass(createConvertToDestinationPassingStylePass());
-    funcPassManager.addPass(createFoldAffineMinInDistributedLoopsPass());
-  }
+  funcPassManager.addPass(
+      createTileAndDistributeToWorkgroupsUsingForallOpPass());
+  funcPassManager.addPass(createBufferizeDispatchTensorLoadStorePass());
+  funcPassManager.addPass(createCombineResultLayoutTransformationPass());
   funcPassManager.addPass(createConfigTrackingCanonicalizerPass());
   funcPassManager.addPass(createCSEPass());
   funcPassManager.addPass(createFuseTensorPadWithConsumerPass());
--- a/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
@@ -34,11 +34,24 @@ static llvm::cl::opt<bool> clEnableUKernelsDecomposeLinalgGeneric(
                    "ukernels are enabled (experimental)"),
     llvm::cl::init(true));
 
+static llvm::cl::opt<bool> clTileDispatchUsingForall(
+    "iree-vmvx-tile-dispatch-using-forall",
+    llvm::cl::desc(
+        "Enable tile and distribute to workgroups using scf.forall"),
+    llvm::cl::init(true));
+
 static void addTileAndDistributePasses(OpPassManager &funcPassManager) {
-  funcPassManager.addPass(createTileAndDistributeToWorkgroupsPass());
-  funcPassManager.addPass(createCSEPass());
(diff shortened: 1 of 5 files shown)


qedawkins (approved):

compiler/src/iree/compiler/Codegen/VMVX/Passes.cpp
```diff
@@ -34,11 +34,24 @@ static llvm::cl::opt<bool> clEnableUKernelsDecomposeLinalgGeneric(
                    "ukernels are enabled (experimental)"),
     llvm::cl::init(true));
 
+static llvm::cl::opt<bool> clTileDispatchUsingForall(
+    "iree-vmvx-tile-dispatch-using-forall",
```
qedawkins: Actually wait, do we need this flag?

hanhanW: We dont need this. It is just in case if it breaks something downstream on CI, so we can run with the flag disabled and fix issues afterwards.

Let me check if CI is happy. If so, I'll remove it.

FYI, I'm removing the CPU's one in https://github.com/iree-org/iree/pull/23614

PR #23619, opened 2026-03-03
[GPUHeuristics] Add min-based tile distribution for imbalanced M/N problems

yzhang93:
For imbalanced problems (4:1+ M/N tile ratio), GCD-based distribution fails when the dominant dimension has non-power-of-2 tile counts (e.g., 149 tiles for F=2376). This PR adds a min-based distribution fallback that redirects remaining tiles to the starved dimension, and uses min-based per-dimension assignment to handle odd tile counts.

Backward weight conv benchmarks on Mi355:

| Shape | Speedup | Baseline (μs) | This PR (μs) |
|-------|---------|---------------|--------------|
| convfp16 -n 32 -c 256 -H 100 -W 100 -k 2376 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -F 4 | 2.36x | 20776.2 | 8817.2 |
| convfp16 -n 32 -c 256 -H 13 -W 13 -k 2376 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -F 4 | 2.15x | 377.9 | 176.1 |
| convfp16 -n 32 -c 256 -H 25 -W 25 -k 2376 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -F 4 | 2.14x | 1866.8 | 872.4 |
| convfp16 -n 32 -c 256 -H 50 -W 50 -k 2376 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -F 4 | 1.66x | 3279.5 | 1981.1 |
| convfp16 -n 32 -c 256 -H 7 -W 7 -k 2376 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -F 4 | 1.54x | 129.7 | 83.9 |

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -533,8 +533,9 @@ static GPUMMASchedule getOptimalMMASchedule(const GPUMatmulShapeType &problem,
   // Redirect remaining tiles to the starved (dominant) dimension.
   auto redirectRemainingTiles = [&](bool condition, int64_t totalTiles,
                                     int64_t &tileSizeDistributed) {
-    if (!condition || remainingTiles <= 1)
+    if (!condition || remainingTiles <= 1) {
       return;
+    }
     int64_t newTile = std::min(remainingTiles, totalTiles);
     if (newTile > tileSizeDistributed) {
       remainingTiles /= (newTile / tileSizeDistributed);
@@ -553,7 +554,7 @@ static GPUMMASchedule getOptimalMMASchedule(const GPUMatmulShapeType &problem,
   LDBG() << "Collapsed tile sizes: M: " << mTileSizeDistributed
          << ", N: " << nTileSizeDistributed;
 
-  // Distribute collapsed counts to per-dimension M and N (inner → outer).
+  // Distribute collapsed counts to per-dimension M and N (inner -> outer).
   // Use min-based distribution for the dominant dimension in imbalanced
   // problems, since GCD fails for non-power-of-2 tile counts.
   auto distributeToDims = [](MutableArrayRef<int64_t> tileCounts,
@@ -561,8 +562,8 @@ static GPUMMASchedule getOptimalMMASchedule(const GPUMatmulShapeType &problem,
                              MutableArrayRef<int64_t> tileSizes,
                              int64_t &subgroupBudget, int64_t &tileBudget,
                              bool useMin) {
-    auto distribute = useMin ? distributeTilesUsingMin
-                             : distributeTilesUsingGCD;
+    int64_t distribute =
+        useMin ? distributeTilesUsingMin : distributeTilesUsingGCD;
     for (size_t e = tileCounts.size(), i = e - 1; i < e; --i) {
       subgroupCounts[i] = distribute(tileCounts[i], subgroupBudget);
       tileSizes[i] = distribute(tileCounts[i], tileBudget);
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_igemm_tile_and_fuse.mlir
```diff
@@ -389,3 +389,32 @@ func.func @conv_with_dps_init_producer(
 
 //     CHECK-LABEL: conv_with_dps_init_producer
 //           CHECK: promote_operands = [0, 1, 2]
+
+// -----
+
+// Backward weight conv where M=F=2376 has a prime tile count (149),
```
kuhar, line 395: Should we also have another testcase for N? Maybe a plain gemm?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -505,38 +515,69 @@ static GPUMMASchedule getOptimalMMASchedule(const GPUMatmulShapeType &problem,
...
+                         nTileSizeDistributed);
+
   LDBG() << "Leftover factors: subgroups: " << remainingSubgroups
          << ", tiles: " << remainingTiles;
   LDBG() << "Collapsed subgroup counts: M: " << mSubgroupDistributed
          << ", N: " << nSubgroupDistributed;
   LDBG() << "Collapsed tile sizes: M: " << mTileSizeDistributed
          << ", N: " << nTileSizeDistributed;
 
+  // Distribute collapsed counts to per-dimension M and N (inner → outer).
```
kuhar:
```suggestion
  // Distribute collapsed counts to per-dimension M and N (inner -> outer).
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -505,38 +515,69 @@ static GPUMMASchedule getOptimalMMASchedule(const GPUMatmulShapeType &problem,
...
+  // Use min-based distribution for the dominant dimension in imbalanced
+  // problems, since GCD fails for non-power-of-2 tile counts.
+  auto distributeToDims = [](MutableArrayRef<int64_t> tileCounts,
+                             MutableArrayRef<int64_t> subgroupCounts,
+                             MutableArrayRef<int64_t> tileSizes,
+                             int64_t &subgroupBudget, int64_t &tileBudget,
+                             bool useMin) {
+    auto distribute = useMin ? distributeTilesUsingMin
```
kuhar: nit: type

kuhar:
> Backward weight conv benchmarks:

What target did you use for this comparison? Did you observe the same pattern across CDNA and RDNA?

yzhang93:
> > Backward weight conv benchmarks:
> 
> What target did you use for this comparison? Did you observe the same pattern across CDNA and RDNA?

The above benchmark results are using Mi355. However, I also observe similar improvements on RDNA4, because on both systems the intrinsic for M dimension is 16, so there is non-power-of-2 tile count along that dimension.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_igemm_tile_and_fuse.mlir
yzhang93: I tried adding a plain GEMM test for the N-heavy case, but both the baseline and this PR produce identical schedules. The reason is that for single-dimension M and N, the collapsed budget gets assigned directly to that one dimension — so min-based and GCD-based distribution give the same result. The override specifically helps backward weight convolutions where M is a single dimension with a prime tile count and N is multi-dimensional and cleanly absorbs the sqrt. That combination starves M under GCD, which is what this PR fixes.

kuhar (approved):

PR #23620, opened 2026-03-03
[Utils] Add MultiPipelineNest and revise MultiOpNest for cross-type parallelism

qedawkins:
Add MultiPipelineNest, a builder for conditional pipeline dispatch over child operations. Each sub-pipeline is guarded by a condition predicate; the first matching condition wins at runtime. When MLIR multithreading is enabled, matched operations are processed in parallel via parallelForEach.

Revise MultiOpNest to use MultiPipelineNest internally instead of creating separate OpToOpPassAdaptor instances per op type. This enables parallel dispatch across all op types (e.g., func::FuncOp, InitializerOp, Util::FuncOp) through a single OpPipelineAdaptorPass, rather than processing each type sequentially.

Along the way this required changing the public API for MultiOpNest. The reason is that the implementation of MultiPipelineNest had to use deferred pass construction to enable post-addition merging of nests. The existing upstream OpPassManager will attempt to merge adjacent OpToOpPassAdaptor passes with identical nests, saving global syncs when users write:

passManager.nest<OpTy>(PassA);
passManager.nest<OpTy>(PassB);

In local cases this is an anti-pattern, however the place where this is an important feature is when considering pass composition, contributing to up to ~4% of total compilation time on a few of the CI run models (based on my local testing). This merge is internal to OpPassManager and relies on comparing the nest scope of adjacent passes, which is not possible until the pass pipeline has been fully realized. So this leaves us with the following options:

1. Eager pass insertion (create the pass right away and keep a reference to the nested pass managers to populate). This on its own is mutually exclusive with the aforementioned pass merging behavior.
2. Eager pass insertion with merging by walking the parent pass manager on MultiPipelineNest destruction. This almost works but leaves empty no-op passes in the pipeline.
3. Deferred pass introduction (what this does). The consequence of this is surprising pass ordering in some cases, such as:

```
void addPasses(OpPassManager &pm) {
  MultiPipelineNest funcNest(pm)
  funcNest.addPass<func::FuncOp>(PassA);
  pm.addPass(PassB);
}
```

In this example PassA is actually scheduled AFTER PassB. To get the expected ordering one would have to either wrap the MultiPipelineNest in an unnamed scope or call `funcNest.commit()` to commit the pass to the pass manager. This is in some ways unfortunate but I could not come up with another way around this without going in and changing the upstream pass manager. Note that the existing approach is not free of its own forms of footguns, such as:

```
void addPasses(OpPassManager &pm) {
  MultiOpNest funcNest(pm)
  funcNest.addPass<func::FuncOp>(PassA);
  pm.addPass(PassB);
  funcNest.addPass<func::FuncOp>(PassC);
}
```

Where in this example the actual pass order is `PassA, PassC, PassB` because `PassC` gets handed off to the OpToOpPassAdaptor backing the MultiOpNest while the concrete adaptor pass is scheduled before `PassB`.

After this there are a few things this enables directly on the Codegen side. Currently codegen uses a fixed pass to dynamically dispatch pipelines to annotated ops. This has the unfortunate effect of bubbling dialect dependencies up to the dispatching pass (the same way that target backends need to populate all dialect deps that the backend might generate). MultiPipelineNest allows us to directly populate a set of static pass pipelines based on matchable conditions. This also means that we'll be able to directly use target features from ExecutableTargetAttr when constructing the pipeline instead of having to thread it through the `LowerTargetExecutable` pass + pipeline options the way we do today.

After this the plan is to cleanup the way that pipeline options are specified across codegen backends:
 - Global cl options stay the same.
 - Target dependent options will be moved out of any pipeline options. buildTranslationPassPipeline already provides the target attribute in the backend but we do not use it and instead rediscover target information by analyzing the IR.
 - IR dependent pass options will be moved to the IR and removed from pipeline options.

The overall goal is to start configuring nested pass pipelines based on target information the way we typically do in the compiler rather than matching IR. Then after that we can start formalizing the accepted input(s) to codegen and adding multi-function support.

One last advantage of this is that custom pass pipelines become super easy with this approach since we can return the MultiPipelineNest directly to the target backend that calls it.

I also ran a quick local compile time benchmark on the PkgCI models to make sure this doesn't horribly regress and the results look ok.

Compile-Time Benchmarks (median of 3 runs)
```
┌─────────────────┬───────────┬──────────┬──────────────────┐
│      Model      │ Reference │ Feature  │      Delta       │
├─────────────────┼───────────┼──────────┼──────────────────┤
│ SDXL CLIP (O1)  │ 12,533ms  │ 12,553ms │ +20ms (+0.1%)    │
├─────────────────┼───────────┼──────────┼──────────────────┤
│ SDXL UNet (O3)  │ 20,521ms  │ 20,932ms │ +411ms (+2.0%)   │
├─────────────────┼───────────┼──────────┼──────────────────┤
│ SDXL punet (O3) │ 51,894ms  │ 50,566ms │ -1,328ms (-2.5%) │
├─────────────────┼───────────┼──────────┼──────────────────┤
│ LLaMA 8B (O3)   │ 5,408ms   │ 5,430ms  │ +22ms (+0.4%)    │
├─────────────────┼───────────┼──────────┼──────────────────┤
│ Total           │ 90,356ms  │ 89,481ms │ -875ms (-1.0%)   │
└─────────────────┴───────────┴──────────┴──────────────────┘
```

punet might be doing slightly better because it's more initializer heavy meaning we're parallelizing more there than the rest. This current change is pretty close to NFC with the way our pipelines are currently setup (this doesn't include any of the codegen changes).

### DIFF
--- a/compiler/src/iree/compiler/Bindings/Native/Transforms/Passes.cpp
+++ b/compiler/src/iree/compiler/Bindings/Native/Transforms/Passes.cpp
@@ -31,8 +31,9 @@ void buildTransformPassPipeline(OpPassManager &passManager,
 
   // Cleanup the IR after manipulating it.
   passManager.addPass(createInlinerPass());
-  FunctionLikeNest(passManager).addPass(createCanonicalizerPass);
-  FunctionLikeNest(passManager).addPass(createCSEPass);
+  FunctionLikeNest(passManager)
+      .addPass(createCanonicalizerPass)
+      .addPass(createCSEPass);
   passManager.addPass(createSymbolDCEPass());
 }
 
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -1060,7 +1060,16 @@ static void addLowerToLLVMGPUPasses(OpPassManager &modulePassManager,
                         : createEmulateNarrowTypePass();
       })
       .addPass(createIREECodegenAffineExpandIndexOpsPass)
-      .addPass(createIREECodegenLowerAffinePass);
+      .addPass(createIREECodegenLowerAffinePass)
+      // Software emulation for small float types (fp4/fp8).
+      .addPredicatedPass(forROCDL, [] {
+        return createConvertUnsupportedFloatArithPass(
+            ConvertUnsupportedFloatArithPassOptions{
+                clLLVMGPUEnableSmallFloatEmulation});
+      });
+
+  // Commit the func-level adaptor before adding module-level passes.
+  funcPassManager.commitPass();
 
   if (!preserveDebugInfo) {
     modulePassManager.addPass(createStripDebugInfoPass());
@@ -1071,20 +1080,12 @@ static void addLowerToLLVMGPUPasses(OpPassManager &modulePassManager,
       IREE::Util::DropCompilerHintsPassOptions{/*keepAssumeInt=*/true}));
 
   if (forROCDL) {
-    // convert to ROCDL.
-    // Software emulation for small float types (fp4/fp8) is controlled by
-    // --iree-llvmgpu-enable-small-float-emulation. When disabled (default),
-    // ConvertToROCDL will error on unsupported types.
-    funcPassManager.addPass([] {
-      return createConvertUnsupportedFloatArithPass(
-          ConvertUnsupportedFloatArithPassOptions{
-              clLLVMGPUEnableSmallFloatEmulation});
-    });
+    // Convert to ROCDL.
     modulePassManager.addPass(createConvertToROCDLPass());
     modulePassManager.addNestedPass<LLVM::LLVMFuncOp>(
         createROCDLAnnotateKernelForTranslationPass());
   } else {
-    // convert to NVVM.
+    // Convert to NVVM.
     modulePassManager.addPass(createConvertToNVVMPass());
   }
 }
--- a/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
@@ -222,6 +222,9 @@ static void addMemRefLoweringPasses(OpPassManager &modulePassManager) {
       .addPass(createCSEPass)
       .addPass([&]() { return createOptimizeVectorTransferPass(); });
 
+  // Commit the func-level adaptor before adding module-level passes.
+  funcPassManager.commitPass();
+
   // Turn multi-dimension memref into one-dimension. This is needed for
   // SPIR-V because we don't use upstream memref descriptors.
   modulePassManager.addPass(createFlattenMemRefSubspanPass());
--- a/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
@@ -181,49 +181,49 @@ static void addLoopMaterializationPasses(OpPassManager &funcPassManager) {
 /// cross loop nest optimizations. This should be invoked after structured op
 /// lowering and before final SPIR-V conversion.
 static void addMemRefLoweringPasses(OpPassManager &modulePassManager) {
-  FunctionLikeNest funcPassManager(modulePassManager);
-
-  funcPassManager.addPass(createCanonicalizerPass)
-      .addPass(createCSEPass)
-      .addPass(createConvertComplexToStandardPass)
-      // Math dialect ops rewrites, approximations, casts.
-      .addPass(createMathTransformPass)
-      .addPass(createPadDynamicAllocPass);
-
   // TODO: query this from the target.
   auto getIndexBitwidth = [](mlir::FunctionOpInterface) { return 32; };
-  funcPassManager
-      .addPass(
-          [&]() { return createGPUCheckResourceUsagePass(getIndexBitwidth); })
 
-      // Fold load/store from/to subview ops into the original memref when
-      // possible. In SPIR-V we don't use memref descriptor so it's not possible
-      // to handle subview ops.
-      .addPass(memref::createFoldMemRefAliasOpsPass)
-      .addPass(createConvertUnsupportedFloatArithPass)
-      .addPass(createEmulateNarrowTypePass)
-      .addPass(createCanonicalizerPass)
-      .addPass(createCSEPass)
+  {
+    FunctionLikeNest funcPassManager(modulePassManager);
+    funcPassManager.addPass(createCanonicalizerPass)
+        .addPass(createCSEPass)
+        .addPass(createConvertComplexToStandardPass)
+        // Math dialect ops rewrites, approximations, casts.
+        .addPass(createMathTransformPass)
+        .addPass(createPadDynamicAllocPass)
+        .addPass([&]() {
+          return createGPUCheckResourceUsagePass(getIndexBitwidth);
+        })
 
-      // Turn scalar load/store from memrefs into vectorized ones if possible.
-      // This gives better memory access patterns, which is very important for
-      // perf.
-      .addPass(createSPIRVVectorizeLoadStorePass)
-      // Perform optimizations that need to across the scf.for region boundary.
-      .addPass(createForOpCanonicalizationPass)
-      // Perform various vector-level cross-op optimizations like load-store
-      // forwarding, shape casting and casting op cancelling.
-      .addPass([&]() { return createOptimizeVectorTransferPass(); })
-      .addPass(createSPIRVBreakDownLargeVectorPass)
+        // Fold load/store from/to subview ops into the original memref when
+        // possible. In SPIR-V we don't use memref descriptor so it's not
+        // possible to handle subview ops.
+        .addPass(memref::createFoldMemRefAliasOpsPass)
+        .addPass(createConvertUnsupportedFloatArithPass)
+        .addPass(createEmulateNarrowTypePass)
+        .addPass(createCanonicalizerPass)
+        .addPass(createCSEPass)
 
-      // Perform optimizations that need to across the scf.for region boundary.
-      .addPass(createForOpCanonicalizationPass)
-      .addPass(createCanonicalizerPass)
-      .addPass(createCSEPass)
-      .addPass([&]() { return createOptimizeVectorTransferPass(); });
+        // Turn scalar load/store from memrefs into vectorized ones if possible.
+        // This gives better memory access patterns, which is very important for
+        // perf.
+        .addPass(createSPIRVVectorizeLoadStorePass)
+        // Perform optimizations that need to across the scf.for region
+        // boundary.
+        .addPass(createForOpCanonicalizationPass)
+        // Perform various vector-level cross-op optimizations like load-store
+        // forwarding, shape casting and casting op cancelling.
+        .addPass([&]() { return createOptimizeVectorTransferPass(); })
+        .addPass(createSPIRVBreakDownLargeVectorPass)
 
-  // Commit the func-level adaptor before adding module-level passes.
-  funcPassManager.commitPass();
+        // Perform optimizations that need to across the scf.for region
+        // boundary.
+        .addPass(createForOpCanonicalizationPass)
+        .addPass(createCanonicalizerPass)
+        .addPass(createCSEPass)
+        .addPass([&]() { return createOptimizeVectorTransferPass(); });
+  }
 
   // Turn multi-dimension memref into one-dimension. This is needed for
   // SPIR-V because we don't use upstream memref descriptors.
--- a/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
@@ -192,9 +192,8 @@ static void addMemRefLoweringPasses(OpPassManager &modulePassManager) {
         // Math dialect ops rewrites, approximations, casts.
         .addPass(createMathTransformPass)
         .addPass(createPadDynamicAllocPass)
-        .addPass([&]() {
-          return createGPUCheckResourceUsagePass(getIndexBitwidth);
-        })
+        .addPass(
+            [&]() { return createGPUCheckResourceUsagePass(getIndexBitwidth); })
 
         // Fold load/store from/to subview ops into the original memref when
         // possible. In SPIR-V we don't use memref descriptor so it's not
--- a/compiler/src/iree/compiler/Dialect/VM/Transforms/Passes.cpp
+++ b/compiler/src/iree/compiler/Dialect/VM/Transforms/Passes.cpp
@@ -94,26 +94,30 @@ void buildVMTransformPassPipeline(OpPassManager &passManager,
   // together.
   passManager.addPass(IREE::Util::createCombineInitializersPass());
 
-  FunctionLikeNest(passManager)
-      .addPass(mlir::createSCFForLoopCanonicalizationPass);
+  {
+    MultiPipelineNest nest(passManager);
+    nest.nest<func::FuncOp, IREE::Util::InitializerOp, IREE::Util::FuncOp>();
 
-  // This pass is sketchy as it can pessimize tight loops due to affine
-  // treating all indices as signed and the unsigned conversion pass not being
-  // able to handle that. The scf.for canonicalization does a decent job of
-  // removing trivial loops above and this catches the rest. It inserts nasty
-  // rem/div ops that we can never safely remove inside of the hot inner loop
-  // and that sucks. We still have this here for now as the cost of the rem/div
-  // are less than the cost of an additional loop that this could remove.
-  passManager.addNestedPass<func::FuncOp>(affine::createLoopCoalescingPass());
+    nest.addPass(mlir::createSCFForLoopCanonicalizationPass);
 
-  FunctionLikeNest(passManager)
-      .addPass(mlir::createLoopInvariantCodeMotionPass)
-      .addPass(mlir::createSCFToControlFlowPass)
-      // TODO: Maybe this should be a part of Affine lowering pass.
-      // Remove if it is added there.
-      // https://github.com/llvm/llvm-project/issues/78458
-      .addPass(createIREECodegenAffineExpandIndexOpsPass)
-      .addPass(createIREECodegenLowerAffinePass);
+    // This pass is sketchy as it can pessimize tight loops due to affine
+    // treating all indices as signed and the unsigned conversion pass not being
+    // able to handle that. The scf.for canonicalization does a decent job of
+    // removing trivial loops above and this catches the rest. It inserts nasty
+    // rem/div ops that we can never safely remove inside of the hot inner loop
+    // and that sucks. We still have this here for now as the cost of the
+    // rem/div are less than the cost of an additional loop that this could
+    // remove.
+    nest.addPassFor<func::FuncOp>(affine::createLoopCoalescingPass);
+
+    nest.addPass(mlir::createLoopInvariantCodeMotionPass)
+        .addPass(mlir::createSCFToControlFlowPass)
+        // TODO: Maybe this should be a part of Affine lowering pass.
+        // Remove if it is added there.
+        // https://github.com/llvm/llvm-project/issues/78458
+        .addPass(createIREECodegenAffineExpandIndexOpsPass)
+        .addPass(createIREECodegenLowerAffinePass);
+  }
 
   passManager.addPass(mlir::arith::createArithUnsignedWhenEquivalentPass());
 
--- a/compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
@@ -106,13 +106,10 @@ void buildGlobalOptimizationPassPipeline(
         importParametersOptions));
   }
 
-  if (clWarnOnUninitializedValues) {
-    FunctionLikeNest(mainPassManager)
-        .addPass(createWarnOnUninitializedValuesPass);
-  }
-
   // Preprocessing passes to get the program into a canonical state.
   FunctionLikeNest(mainPassManager)
+      .addPredicatedPass(clWarnOnUninitializedValues,
+                         createWarnOnUninitializedValuesPass)
       .addPredicatedPass(transformOptions.stripAssertions,
                          IREE::Util::createStripDebugOpsPass)
       .addPass(IREE::Util::createOptimizeIntArithmeticPass)
@@ -154,9 +151,7 @@ void buildGlobalOptimizationPassPipeline(
         GeneralizeLinalgNamedOpsPassOptions opt;
         opt.enableGeneralizeMatmul = transformOptions.generalizeMatmul;
         return createGeneralizeLinalgNamedOpsPass(opt);
-      });
-
-  FunctionLikeNest(mainPassManager)
+      })
       .addPredicatedPass(!clEnableEdgeReshapePropagation,
                          DispatchCreation::createInsertTensorBarriersPass);
   mainPassManager.addPass(DispatchCreation::createFoldUnitExtentDimsPass());
@@ -214,10 +209,9 @@ void buildGlobalOptimizationPassPipeline(
   }
   // Generalize transposes and any other remaining named linalg ops that can
   // now be represented as generics.
-  FunctionLikeNest(mainPassManager).addPass(createGeneralizeLinalgNamedOpsPass);
-
   // Hoist loop invariants (e.g. from scf loops) with zero-trip-check.
   FunctionLikeNest(mainPassManager)
+      .addPass(createGeneralizeLinalgNamedOpsPass)
       .addPass(createGlobalLoopInvariantCodeMotionPass)
       .addPass(IREE::Flow::createCanonicalizePass)
       .addPass(mlir::createCSEPass)
--- a/compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
@@ -207,11 +207,12 @@ void buildGlobalOptimizationPassPipeline(
     mainPassManager.addPass(createSimplifyPackUnpackPass());
     FunctionLikeNest(mainPassManager).addPass(createDataLayoutPropagationPass);
   }
-  // Generalize transposes and any other remaining named linalg ops that can
-  // now be represented as generics.
-  // Hoist loop invariants (e.g. from scf loops) with zero-trip-check.
+
   FunctionLikeNest(mainPassManager)
+      // Generalize transposes and any other remaining named linalg ops that can
+      // now be represented as generics.
       .addPass(createGeneralizeLinalgNamedOpsPass)
+      // Hoist loop invariants (e.g. from scf loops) with zero-trip-check.
       .addPass(createGlobalLoopInvariantCodeMotionPass)
       .addPass(IREE::Flow::createCanonicalizePass)
       .addPass(mlir::createCSEPass)
--- a/compiler/src/iree/compiler/Utils/PassUtils.cpp
+++ b/compiler/src/iree/compiler/Utils/PassUtils.cpp
@@ -7,8 +7,10 @@
 #include "iree/compiler/Utils/PassUtils.h"
 
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/DebugLog.h"
+#include "mlir/IR/Threading.h"
 
-#define DEBUG_TYPE "iree-utils"
+#define DEBUG_TYPE "iree-pass-utils"
 
 namespace mlir::iree_compiler {
 
@@ -22,6 +24,315 @@ void signalFixedPointModified(Operation *rootOp) {
 
   LLVM_DEBUG(llvm::dbgs() << "Signalling fixed-point iterator modification");
   rootOp->setAttr("iree.fixedpoint.modified", UnitAttr::get(context));
+}
+
+//===----------------------------------------------------------------------===//
+// OpPipelineAdaptorPass
+//===----------------------------------------------------------------------===//
+
+namespace detail {
+
+OpPipelineAdaptorPass::OpPipelineAdaptorPass(const OpPipelineAdaptorPass &other)
+    : PassWrapper(other) {
+  entries.reserve(other.entries.size());
+  for (const Entry &e : other.entries) {
+    entries.emplace_back(e.condition, OpPassManager(e.pipeline), e.opTypeID,
+                         e.batchIndex);
+  }
+}
+
+OpPassManager &OpPipelineAdaptorPass::addEntry(ConditionFn condition,
+                                               StringRef anchorOpName,
+                                               std::optional<TypeID> opTypeID) {
+  if (anchorOpName.empty()) {
+    entries.emplace_back(std::move(condition), OpPassManager(), opTypeID);
+  } else {
+    entries.emplace_back(std::move(condition), OpPassManager(anchorOpName),
+                         opTypeID);
+  }
+  return entries.back().pipeline;
+}
+
+void OpPipelineAdaptorPass::mergeFrom(OpPipelineAdaptorPass &other) {
+  // Determine the next batch index for incoming entries.
+  unsigned newBatch = 0;
+  for (const Entry &e : entries) {
+    newBatch = std::max(newBatch, e.batchIndex + 1);
+  }
+  for (Entry &e : other.entries) {
+    e.batchIndex = newBatch;
+    entries.push_back(std::move(e));
+  }
+  other.entries.clear();
+  // Invalidate async executors since entries changed.
+  asyncExecutors.clear();
+}
+
+void OpPipelineAdaptorPass::getDependentDialects(
+    DialectRegistry &registry) const {
+  for (const Entry &e : entries) {
+    e.pipeline.getDependentDialects(registry);
+  }
+}
+
+void OpPipelineAdaptorPass::runOnOperation() {
+  if (getContext().isMultithreadingEnabled()) {
+    runOnOperationAsync();
+  } else {
+    runOnOperationSync();
+  }
+}
+
+void OpPipelineAdaptorPass::runOnOperationSync() {
+  Operation *parentOp = getOperation();
+  LDBG() << "Running op pipeline adaptor synchronously on '"
+         << parentOp->getName() << "'";
+
+  for (Region &region : parentOp->getRegions()) {
+    for (Operation &op : region.getOps()) {
+      // Within a batch, first match wins. Across batches (from merged
+      // adaptors), all matching batches run.
+      std::optional<unsigned> lastMatchedBatch;
+      for (Entry &e : entries) {
+        if (lastMatchedBatch && e.batchIndex == *lastMatchedBatch) {
+          continue;
+        }
+        if (e.condition(&op)) {
+          LDBG() << "  Dispatching '" << op.getName() << "' to pipeline"
+                 << " (batch " << e.batchIndex << ")";
+          if (failed(runPipeline(e.pipeline, &op))) {
+            return signalPassFailure();
+          }
+          lastMatchedBatch = e.batchIndex;
+        }
+      }
+    }
+  }
+}
+
+void OpPipelineAdaptorPass::runOnOperationAsync() {
+  Operation *parentOp = getOperation();
+  MLIRContext *context = &getContext();
+
+  LDBG() << "Running op pipeline adaptor asynchronously on '"
+         << parentOp->getName() << "'";
+
+  // --- Phase 1: Collect matched operations (sequential). ---
+
+  // Information for a single matched operation. An op may match multiple
+  // entries when batches from merged adaptors are present.
+  struct OpDispatchInfo {
+    SmallVector<unsigned, 2> entryIndices;
+    Operation *op;
+  };
+
+  SmallVector<OpDispatchInfo> opInfos;
+  for (Region &region : parentOp->getRegions()) {
+    for (Operation &op : region.getOps()) {
+      OpDispatchInfo info;
+      info.op = &op;
+      // Within a batch, first match wins. Across batches, all matching
+      // batches run.
+      std::optional<unsigned> lastMatchedBatch;
+      for (unsigned i = 0, e = entries.size(); i < e; ++i) {
+        if (lastMatchedBatch && entries[i].batchIndex == *lastMatchedBatch) {
+          continue;
+        }
+        if (entries[i].condition(&op)) {
+          info.entryIndices.push_back(i);
+          lastMatchedBatch = entries[i].batchIndex;
+        }
+      }
+      if (!info.entryIndices.empty()) {
+        opInfos.push_back(std::move(info));
+      }
+    }
+  }
+
+  if (opInfos.empty()) {
+    return;
+  }
+
+  // Pre-create nested analysis managers so that the dynamic pipeline
+  // callback (invoked from runPipeline) only performs lookups — not
+  // insertions — on the parent AnalysisManager's child map during the
+  // parallel section. This is required because DenseMap insertion is not
+  // thread-safe.
+  AnalysisManager am = getAnalysisManager();
+  for (OpDispatchInfo &info : opInfos) {
+    (void)am.nest(info.op);
+  }
+
+  // --- Phase 2: Ensure per-thread entry clones exist. ---
+
+  unsigned numThreads = context->getThreadPool().getMaxConcurrency();
+  // Entries are immutable after pipeline construction, so size comparison
+  // suffices for staleness detection.
+  if (asyncExecutors.empty() ||
+      asyncExecutors.front().size() != entries.size()) {
+    LDBG() << "  Creating " << numThreads << " async executors with "
+           << entries.size() << " entries each";
+    asyncExecutors.clear();
+    asyncExecutors.resize(numThreads);
+    for (SmallVector<Entry> &executor : asyncExecutors) {
+      executor.reserve(entries.size());
+      for (const Entry &e : entries) {
+        executor.emplace_back(e.condition, OpPassManager(e.pipeline),
+                              e.opTypeID, e.batchIndex);
+      }
+    }
+  }
+
+  // --- Phase 3: Parallel dispatch. ---
+
+  // Track which executors are in use (one per thread).
+  std::vector<std::atomic<bool>> activeExecutors(asyncExecutors.size());
+  for (std::atomic<bool> &active : activeExecutors) {
+    active.store(false);
+  }
+  std::atomic<bool> hasFailure(false);
+
+  // NOTE: Using the dynamic pipeline API (Pass::runPipeline) rather than
+  // the internal OpToOpPassAdaptor::runPipeline means instrumentation events
+  // will report the parent thread ID rather than the actual worker thread.
+  // This is acceptable since OpPipelineAdaptorPass is an internal
+  // implementation detail.
+  parallelForEach(context, opInfos, [&](OpDispatchInfo &info) {
+    // Claim an inactive executor via atomic compare-and-swap.
+    auto it = llvm::find_if(activeExecutors, [](std::atomic<bool> &isActive) {
+      bool expected = false;
+      return isActive.compare_exchange_strong(expected, true);
+    });
+    assert(it != activeExecutors.end() &&
+           "more concurrent tasks than available executor slots");
+    unsigned executorIdx = it - activeExecutors.begin();
+
+    // Run all matched pipelines from this executor's clone. Multiple entries
+    // can match when batches from merged adaptors are present.
+    for (unsigned entryIdx : info.entryIndices) {
+      OpPassManager &pm = asyncExecutors[executorIdx][entryIdx].pipeline;
+      if (failed(runPipeline(pm, info.op))) {
+        hasFailure.store(true);
+        break;
+      }
+    }
+
+    // Release the executor.
+    activeExecutors[executorIdx].store(false);
+  });
+
+  if (hasFailure) {
+    signalPassFailure();
+  }
+}
+
+} // namespace detail
+
+//===----------------------------------------------------------------------===//
+// MultiPipelineNest
+//===----------------------------------------------------------------------===//
+
+MultiPipelineNest::MultiPipelineNest(OpPassManager &parentPm)
+    : parentPm(&parentPm),
+      ownedPass(std::make_unique<detail::OpPipelineAdaptorPass>()),
+      adaptorPass(ownedPass.get()) {}
+
+MultiPipelineNest::~MultiPipelineNest() {
+  if (!adaptorPass) {
+    return;
+  }
+  // Already committed to the PM — nothing to do.
+  if (!ownedPass) {
+    return;
+  }
+  // No entries were added — discard the pass silently.
+  if (adaptorPass->getEntries().empty()) {
+    return;
+  }
+  // Try to merge into the predecessor. On success the entries are moved
+  // and ownedPass is destroyed (now empty) when we return.
+  if (tryMergeIntoPredecessor()) {
+    return;
+  }
+  // No merge — insert the pass into the parent PM.
+  parentPm->addPass(std::move(ownedPass));
+}
+
+MultiPipelineNest::MultiPipelineNest(MultiPipelineNest &&other) noexcept
+    : parentPm(other.parentPm), ownedPass(std::move(other.ownedPass)),
+      adaptorPass(other.adaptorPass) {
+  other.parentPm = nullptr;
+  other.adaptorPass = nullptr;
+}
+
+MultiPipelineNest &
+MultiPipelineNest::operator=(MultiPipelineNest &&other) noexcept {
+  if (this != &other) {
+    // Flush the current pass before replacing.
+    if (adaptorPass && ownedPass && !adaptorPass->getEntries().empty()) {
+      if (!tryMergeIntoPredecessor()) {
+        parentPm->addPass(std::move(ownedPass));
+      }
+    }
+    parentPm = other.parentPm;
+    ownedPass = std::move(other.ownedPass);
+    adaptorPass = other.adaptorPass;
+    other.parentPm = nullptr;
+    other.adaptorPass = nullptr;
+  }
+  return *this;
+}
+
+void MultiPipelineNest::commitPass() {
+  if (!ownedPass) {
+    return; // Already committed.
+  }
+  parentPm->addPass(std::move(ownedPass));
+  // adaptorPass remains valid — it now points into the PM.
+}
+
+bool MultiPipelineNest::tryMergeIntoPredecessor() {
+  // Can't merge if we have no entries.
+  if (adaptorPass->getEntries().empty()) {
+    return false;
+  }
+
+  // Can only merge if all our entries have TypeIDs.
+  for (const auto &e : adaptorPass->getEntries()) {
+    if (!e.opTypeID) {
+      return false;
+    }
+  }
+
+  // Our pass is NOT in the PM (deferred). Check the last pass in the PM.
+  if (parentPm->empty()) {
+    return false;
+  }
+  Pass &lastPass = *std::prev(parentPm->end());
+  if (lastPass.getTypeID() != TypeID::get<detail::OpPipelineAdaptorPass>()) {
+    return false;
+  }
+  auto &predAdaptor = static_cast<detail::OpPipelineAdaptorPass &>(lastPass);
+  // Predecessor must have entries and all must have TypeIDs.
+  if (predAdaptor.getEntries().empty()) {
+    return false;
+  }
+  for (const auto &e : predAdaptor.getEntries()) {
+    if (!e.opTypeID) {
+      return false;
+    }
+  }
+  LDBG() << "Merging adaptor (" << adaptorPass->getEntries().size()
+         << " entries) into predecessor (" << predAdaptor.getEntries().size()
+         << " entries)";
+  predAdaptor.mergeFrom(*adaptorPass);
+  return true;
+}
+
+OpPassManager &MultiPipelineNest::nestIf(ConditionFn condition,
+                                         StringRef anchorOpName,
+                                         std::optional<TypeID> opTypeID) {
+  return adaptorPass->addEntry(std::move(condition), anchorOpName, opTypeID);
 }
 
 } // namespace mlir::iree_compiler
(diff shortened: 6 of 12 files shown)


compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -7,14 +7,258 @@
...
+namespace detail {
+
+/// An unregistered pass that adapts a parent pass manager to run sub-pipelines
+/// on child operations. The pass walks top-level operations in all regions of
+/// the parent operation, evaluates conditions in order for each child, and runs
+/// the first matching sub-pipeline. Children that match no condition are
+/// skipped.
+///
+/// When MLIR multithreading is enabled, child operations are dispatched in
+/// parallel using the same pattern as OpToOpPassAdaptor: sub-pipelines are
```
IanWood1, line 75: There seems to be some special logic for `OpToOpPassAdaptor` in the upstream infra. Do we need/want something similar?

e.g. 

https://github.com/llvm/llvm-project/blob/fbc3a312d69d120d0e594eb8a32bd50222e855cb/mlir/lib/Pass/IRPrinting.cpp#L72
https://github.com/llvm/llvm-project/blob/fbc3a312d69d120d0e594eb8a32bd50222e855cb/mlir/lib/Pass/PassCrashRecovery.cpp#L360
https://github.com/llvm/llvm-project/blob/fbc3a312d69d120d0e594eb8a32bd50222e855cb/mlir/lib/Pass/PassTiming.cpp#L93


Since this is an unregistered pass it won't be possible to parse it from the textual pass pipeline (but it gets printed).  `--mlir-pass-pipeline-crash-reproducer` will serialize `iree-op-pipeline-adaptor` but `iree-opt --run-reproducer` can't parse it back

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -7,14 +7,258 @@
...
+/// cloned per thread, and operations are processed via parallelForEach.
+///
+/// This pass is used internally by MultiPipelineNest and is not intended for
+/// direct use.
+class OpPipelineAdaptorPass final
+    : public PassWrapper<OpPipelineAdaptorPass, OperationPass<>> {
+public:
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(OpPipelineAdaptorPass)
+
+  using ConditionFn = std::function<bool(Operation *)>;
```
IanWood1, line 90: Do we need the condition lambdas? From the description it seems like these might be needed for later use. Every caller uses only `nest<T>()` (except for tests), which generates `[](Operation *op) { return isa<T>(op); }`. This is equivalent to the op-name check that `OpPassManager` already stores via its anchor. Also, the `ConditionFn` lambda makes entries non-serializable.

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,313 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  }
+}
+
+void OpPipelineAdaptorPass::runOnOperation() {
+  if (getContext().isMultithreadingEnabled()) {
+    runOnOperationAsync();
+  } else {
+    runOnOperationSync();
+  }
+}
```
IanWood1, line 89: One quirk of having the generic `nestIf` is that running with/without multithreading may lead to different results. `runOnOperationSync` does lazy evaluation of the `ConditionFn` while `runOnOperationAsync` does not. When running in `runOnOperationAsync`, there might be a pass that mutates the IR in a way that the lazy condition produces a different result vs if it were run at the beginning of the pipeline.

This isn't a problem currently because all the use cases of `nestIf` just look for op type which is immutable. However, it might be worth noting that the condition should only consider immutable IR state (e.g. op type).

compiler/src/iree/compiler/Bindings/Native/Transforms/Passes.cpp
hanhanW, line 1: This belong to a separate cleanup PR, IMO.

Same comment for few other files like:

- compiler/src/iree/compiler/Bindings/TFLite/Transforms/Passes.cpp
- compiler/src/iree/compiler/Dialect/Stream/Transforms/Passes.cpp
- compiler/src/iree/compiler/GlobalOptimization/Passes.cpp

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
```diff
@@ -212,10 +207,9 @@ void buildGlobalOptimizationPassPipeline(
   }
   // Generalize transposes and any other remaining named linalg ops that can
   // now be represented as generics.
-  FunctionLikeNest(mainPassManager).addPass(createGeneralizeLinalgNamedOpsPass);
```
hanhanW, line 217: The comment is not moved.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -1060,7 +1063,16 @@ static void addLowerToLLVMGPUPasses(OpPassManager &modulePassManager,
...
       })
       .addPass(createIREECodegenAffineExpandIndexOpsPass)
-      .addPass(createIREECodegenLowerAffinePass);
+      .addPass(createIREECodegenLowerAffinePass)
+      // Software emulation for small float types (fp4/fp8).
+      .addPredicatedPass(forROCDL, [] {
+        return createConvertUnsupportedFloatArithPass(
+            ConvertUnsupportedFloatArithPassOptions{
+                clLLVMGPUEnableSmallFloatEmulation});
+      });
```
hanhanW, line 1069: Is this move intended? I think we'll want this for NVVM path as well, though I don't know how to test it atm. It is fine to make the change to me.

compiler/src/iree/compiler/Dialect/VM/Transforms/Passes.cpp
```diff
@@ -94,26 +94,30 @@ void buildVMTransformPassPipeline(OpPassManager &passManager,
...
-      .addPass(mlir::createLoopInvariantCodeMotionPass)
-      .addPass(mlir::createSCFToControlFlowPass)
-      // TODO: Maybe this should be a part of Affine lowering pass.
-      // Remove if it is added there.
-      // https://github.com/llvm/llvm-project/issues/78458
-      .addPass(createIREECodegenAffineExpandIndexOpsPass)
-      .addPass(createIREECodegenLowerAffinePass);
+  {
+    MultiPipelineNest nest(passManager);
+    nest.nest<func::FuncOp, IREE::Util::InitializerOp, IREE::Util::FuncOp>();
```
IanWood1, line 99: Could this use FunctionLikeNest?

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,313 @@ void signalFixedPointModified(Operation *rootOp) {
...
+//===----------------------------------------------------------------------===//
+// MultiPipelineNest
+//===----------------------------------------------------------------------===//
+
+MultiPipelineNest::MultiPipelineNest(OpPassManager &parentPm)
+    : parentPm(&parentPm),
+      ownedPass(std::make_unique<detail::OpPipelineAdaptorPass>()),
+      adaptorPass(ownedPass.get()) {}
+
+MultiPipelineNest::~MultiPipelineNest() {
```
IanWood1, line 252: Could we add an assert here to check that no passes were added to the pass manager between construction & destruction?


Also, I think gtest supports checking for crashes (`EXPECT_DEATH`).

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -7,14 +7,258 @@
...
+  MultiPipelineNest &operator=(MultiPipelineNest &&other) noexcept;
+  MultiPipelineNest(const MultiPipelineNest &) = delete;
+  MultiPipelineNest &operator=(const MultiPipelineNest &) = delete;
+
+  /// Immediately insert the pass into the parent PM at the current position.
+  /// After this call, the pass is owned by the PM and no merge will be
+  /// attempted on destruction. Use this when interleaving parent-level and
+  /// nested passes with a named MultiPipelineNest variable.
+  void commitPass();
+
```
IanWood1, line 243: Suggestion: we could make `MultiOpNest` methods &&-qualified to force temporary only usage and then remove `commitPass()`. This eliminates the deferred-insertion ordering footgun. 

For functions that currently take `FunctionLikeNest &` (e.g., `addCommonTargetExecutablePreprocessingPasses`, `addEncodingToNopPasses`, `buildLoopOptimizationVMVXTransformPassPipeline`), they can be refactored to take `OpPassManager &` and construct their own `FunctionLikeNest` temporary internally. The auto-merge logic handles merging adjacent temporaries into a single adaptor pass, so the behavior is identical.

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -7,14 +7,258 @@
...
+/// This pass is used internally by MultiPipelineNest and is not intended for
+/// direct use.
+class OpPipelineAdaptorPass final
+    : public PassWrapper<OpPipelineAdaptorPass, OperationPass<>> {
+public:
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(OpPipelineAdaptorPass)
+
+  using ConditionFn = std::function<bool(Operation *)>;
+
+  struct Entry {
```
hanhanW, line 94: Maybe we can add a struct comment like:

```
  /// A condition predicate paired with the pass pipeline to run when the
  /// condition holds.
```

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -7,14 +7,258 @@
...
+    /// semantics apply. Across batches (from merged adaptors), all matching
+    /// batches run.
+    unsigned batchIndex = 0;
+
+    Entry(ConditionFn condition, OpPassManager pipeline,
+          std::optional<TypeID> opTypeID = std::nullopt,
+          unsigned batchIndex = 0)
+        : condition(std::move(condition)), pipeline(std::move(pipeline)),
+          opTypeID(opTypeID), batchIndex(batchIndex) {}
+  };
```
hanhanW, line 111: Check clang-tidy.

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -26,57 +270,40 @@ namespace mlir::iree_compiler {
...
-    std::get<index>(nestedPassManagers)->addPass(constructor());
-    addPassRecurse<F, index + 1, Rest...>(constructor);
+private:
+  template <typename T, typename... Rest>
+  void initNests() {
+    nest.template nest<T>();
+    if constexpr (sizeof...(Rest) > 0) {
+      initNests<Rest...>();
+    }
   }
```
hanhanW: This is what I learned from Claude (c++ is not easy):

MultiPipelineNest already has variadic nest<T1, T2, Rest...>() at line 193. The constructor could just call nest.template nest<OpTys...>() directly.

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -7,14 +7,258 @@
...
+                          std::optional<TypeID> opTypeID = std::nullopt);
+
+  /// Merge entries from another adaptor into this one. Entries from |other|
+  /// are assigned a new batch index so that dispatch preserves
+  /// first-match-wins within each original nest scope while running all
+  /// matching batches. |other| is left with empty entries after the merge.
+  void mergeFrom(OpPipelineAdaptorPass &other);
+
+  /// Access entries for iteration (e.g., to add passes to all pipelines).
+  MutableArrayRef<Entry> getEntries() { return entries; }
```
hanhanW: Instead of exposing internal details, do we add one or two methods to keep entries private?

compiler/src/iree/compiler/Utils/unittests/PassUtilsTest.cpp
hanhanW, line 1: Do we add a test case for `addPassFor` method?

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
qedawkins: Yeah this and some of the other NFC pipeline changes are an attempt to avoid relying on the pipeline merging optimization. I was surprised to learn that's how pipeline nests work, but essentially it tries to merge adjacent pass nests into one if they are nested on the same op. With the multi-nest approach it's harder to actually do the merging so I tried to remove it wherever I could (but still added the merging thing).

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: Oh good point. This code was unchanged so I didn't think about trying to change it now that the underlying implementation is different.

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: I can try but I'm not sure how to make that work coming from downstream. Breaking reproducers is certainly one of the most compelling reasons to NOT make this change, but looking at that code do the current reproducers even handle OpToOpPassAdaptors properly?

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: Current usage doesn't do anything with it yes, but getting these conditions was the main reason for me to make this change. Codegen pipelines condition on the IR (via attributes) so this will let us continue doing that but without the need to jump through a custom pass.

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: The reason I added the ability to commit was so that I could return MultiOpNests from pipeline constructors allowing users to populate their own custom condition in the middle of a pre-set pipeline (basically return the nest from `build<Backend>CodegenPassPipeline`).

compiler/src/iree/compiler/Utils/PassUtils.h
IanWood1: Adding `--mlir-pass-pipeline-local-reproducer` seems to work with these changes because it only prints the single pass that failed. It also seems like there are unrelated issues in IREE with the full pipeline reproducer. For printing/timing this will just introduce a bit of noise but that shouldn't be a problem.

compiler/src/iree/compiler/Utils/PassUtils.cpp
qedawkins: Good point, I didn't think about that. I can't make the Operation* const since almost no operation methods are const, so I don't know how to enforce that. I'll leave a comment but if you have a better suggestion I'll take it.

compiler/src/iree/compiler/Bindings/Native/Transforms/Passes.cpp
qedawkins: I did not change this because of the earlier comment, but if you insist I can remove them now I think.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
qedawkins: Oh, yeah this pass was already ROCDL only for some reason. Makes sense to me to have it on both but I don't want to change that here.

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: For a little more context, I deliberated for quite a while before deciding that I couldn't get around the commit function *and* not regress the compiler by not merging adjacent identical nests

Unless I went and did this upstream. Upstream they do this by finalizing the pass pipeline at the end (i.e. do the merging). Doing it downstream means I need to do it before handing it over to the PassManager, which means I need to wait until I've set all of the conditions up (so that I know the merge is valid). Maybe there is a way to do it without the commit function by doing something similar to what you're saying, but I kinda gave up at a certain point and decided this is explicit enough about what's happening to be usable.

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: Oh nice, thanks for confirming. That's a relief, I wasn't sure how I'd fix that.

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: Clang-tidy seems to be giving a false positive here. All callers pass temporaries meaning this is happening as a move. OpPassManager's copy constructor does a deep copy so we'd get the temporary + deep copy

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: Done!

compiler/src/iree/compiler/Dialect/VM/Transforms/Passes.cpp
qedawkins: MultiOpNest (and by proxy FunctionLikeNest) adds all passes to all sub-pipelines and for whatever reason (don't ask me why) the `createLoopCoalescingPass` here is only being applied to `func::FuncOp` so this has to use `MultiPipelineNest` directly.

compiler/src/iree/compiler/Utils/PassUtils.cpp
qedawkins: Done. If I get failures I might wait and do the cleanups in a follow up though.

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,317 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  other.adaptorPass = nullptr;
+}
+
+MultiPipelineNest &
+MultiPipelineNest::operator=(MultiPipelineNest &&other) noexcept {
+  if (this != &other) {
+    // Flush the current pass before replacing.
+    if (adaptorPass && ownedPass && !adaptorPass->empty()) {
+      if (!tryMergeIntoPredecessor()) {
+        parentPm->addPass(std::move(ownedPass));
```
IanWood1: You might want to add the same assert here `assert(parentPm->size() == parentPmSizeAtConstruction)`.

compiler/src/iree/compiler/Utils/PassUtils.h
IanWood1: That makes sense, I'm fine with the commit function as long as there is an `assert` in the destructor.

kuhar:
(I plan to review by tomorrow, currently stuck on other things...)

compiler/src/iree/compiler/Utils/PassUtils.cpp
qedawkins: Done

compiler/src/iree/compiler/Utils/unittests/PassUtilsTest.cpp
```diff
@@ -0,0 +1,784 @@
...
+/// An OperationPass<> that records its execution order via a shared atomic
+/// counter and stores the counter value as an integer attribute. Used to
+/// verify pass ordering across module-level and dispatched function-level
+/// passes.
+struct CounterPass final : public PassWrapper<CounterPass, OperationPass<>> {
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(CounterPass)
+
+  CounterPass(std::string attrName, std::atomic<int> *counter)
+      : attrName(std::move(attrName)), counter(counter) {}
+  CounterPass(const CounterPass &other)
```
krzysz00: Clang nits here

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -26,57 +294,32 @@ namespace mlir::iree_compiler {
 template <typename... OpTys>
 struct MultiOpNest {
 public:
```
kuhar: nit: redundant public

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  return entries.back().pipeline;
+}
+
+bool OpPipelineAdaptorPass::allEntriesHaveTypeID() const {
+  for (const Entry &e : entries) {
+    if (!e.opTypeID) {
+      return false;
+    }
+  }
+  return true;
```
kuhar:
```suggestion
  return llvm::all_of(entries,
                      [](const Entry &e) { return e.opTypeID.has_value(); });

```

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+      }
+    }
+  }
+
+  if (opInfos.empty()) {
+    return;
+  }
+
+  // Pre-create nested analysis managers so that the dynamic pipeline
+  // callback (invoked from runPipeline) only performs lookups — not
```
kuhar: ubernit: I generally avoid using non-ascii characters unless necessary (here em-dash). Also elsewhere.

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+        executor.emplace_back(e.condition, OpPassManager(e.pipeline),
+                              e.opTypeID, e.batchIndex);
+      }
+    }
+  }
+
+  // --- Phase 3: Parallel dispatch. ---
+
+  // Track which executors are in use (one per thread).
+  std::vector<std::atomic<bool>> activeExecutors(asyncExecutors.size());
```
kuhar: This may not work on all standard library implementations / after we update to future C++ standard -- atomic is neither copyable nor moveable. I know we don't actually have to resize, but this is the contract of `std::vector`.

compiler/src/iree/compiler/Bindings/Native/Transforms/Passes.cpp
```diff
@@ -31,8 +31,9 @@ void buildTransformPassPipeline(OpPassManager &passManager,
 
   // Cleanup the IR after manipulating it.
   passManager.addPass(createInlinerPass());
-  FunctionLikeNest(passManager).addPass(createCanonicalizerPass);
-  FunctionLikeNest(passManager).addPass(createCSEPass);
+  FunctionLikeNest(passManager)
+      .addPass(createCanonicalizerPass)
+      .addPass(createCSEPass);
   passManager.addPass(createSymbolDCEPass());
```
kuhar, line 37: High-level question / comment: I wonder if there's some alternative API that doesn't have to worry about commits / destruction order, such as:

```c++
  passManager.addPass(createInlinerPass());
  nestWith(passManager, [] (FunctionLikeNest& nest) {
    nest.addPass(createCanonicalizerPass)
        .addPass(createCSEPass);
  }).addPass(createSymbolDCEPass());
```

compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
```diff
@@ -222,6 +221,9 @@ static void addMemRefLoweringPasses(OpPassManager &modulePassManager) {
       .addPass(createCSEPass)
       .addPass([&]() { return createOptimizeVectorTransferPass(); });
 
+  // Commit the func-level adaptor before adding module-level passes.
+  funcPassManager.commitPass();
```
kuhar: Alternatively, use an unnamed function nest (temporary)?

compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
kuhar: Why not merge these two and move the lambda higher up?

compiler/src/iree/compiler/Utils/PassUtils.cpp
kuhar: You can use `make_unique<T[]>` instead

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+
+/// An unregistered pass that adapts a parent pass manager to run sub-pipelines
+/// on child operations. The pass walks top-level operations in all regions of
+/// the parent operation, evaluates conditions in order for each child, and runs
+/// the first matching sub-pipeline. Children that match no condition are
+/// skipped.
+///
+/// When MLIR multithreading is enabled, child operations are dispatched in
+/// parallel using the same pattern as OpToOpPassAdaptor: sub-pipelines are
+/// cloned per thread, and operations are processed via parallelForEach.
```
kuhar, line 76: I assume that this requires siblings to be isolated from above?

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+        : condition(std::move(condition)), pipeline(std::move(pipeline)),
+          opTypeID(opTypeID), batchIndex(batchIndex) {}
+  };
+
+  /// Construct with no entries. Entries are added later via addEntry().
+  OpPipelineAdaptorPass() = default;
+
+  /// Copy for thread-safe cloning. OpPassManagers are deep-copied;
+  /// ConditionFn objects are value-copied (safe because they are read-only).
+  OpPipelineAdaptorPass(const OpPipelineAdaptorPass &other);
```
kuhar, line 118: either define copy assignment or delete it

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+  /// automatic merging).
+  bool allEntriesHaveTypeID() const;
+
+  /// Add a pass to all existing sub-pipelines.
+  template <typename F>
+  void addPassToAll(F constructor) {
+    for (Entry &entry : entries) {
+      entry.pipeline.addPass(constructor());
+    }
+  }
```
kuhar:
```suggestion
  template <typename Fn>
  void addPassToAll(Fn constructor) {
    for (Entry &entry : entries) {
      entry.pipeline.addPass(constructor());
    }
  }
```

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+
+  /// Add a pass only to sub-pipelines targeting the given op type.
+  template <typename F>
+  void addPassToEntriesWithTypeID(TypeID targetID, F constructor) {
```
kuhar:
```suggestion
  template <typename Fn>
  void addPassToEntriesWithTypeID(TypeID targetID, Fn constructor) {
```

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+/// Usage:
+///   // Typical: temporary destroyed at semicolon — pass inserted here.
+///   MultiPipelineNest(pm).nest<FuncOp>().addPass(createMyPass);
+///
+///   // Named variable with explicit commit point:
+///   MultiPipelineNest nest(pm);
+///   nest.nest<FuncOp>();
+///   nest.commitPass();           // insert now, before subsequent passes
+///   pm.addPass(createOtherPass); // guaranteed to be after nest's pass
+class MultiPipelineNest {
```
kuhar: `final` or do you expect users to derive from this?

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+  /// (possibly merged with the predecessor) or when commitPass() is called.
+  explicit MultiPipelineNest(OpPassManager &parentPm);
+
+  /// On destruction: if the pass has not been committed, attempt to merge
+  /// with the predecessor adaptor. If merge fails, insert the pass.
+  ~MultiPipelineNest();
+
+  // Movable but not copyable. The moved-from object is left in a null state.
+  MultiPipelineNest(MultiPipelineNest &&other) noexcept;
+  MultiPipelineNest &operator=(MultiPipelineNest &&other) noexcept;
```
kuhar: We don't use noexcept in LLVM-based projects

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+  /// safe per-type pass addition after all entries are created.
+  OpPassManager &nestIf(ConditionFn condition, StringRef anchorOpName = "",
+                        std::optional<TypeID> opTypeID = std::nullopt);
+
+  /// Convenience: add a sub-pipeline for a specific op type. The pipeline
+  /// is anchored to OpT so that typed passes can be added directly.
+  /// Records the TypeID for automatic merging.
+  template <typename OpT>
+  OpPassManager &nest() {
+    return nestIf([](Operation *op) { return isa<OpT>(op); },
```
kuhar:
```suggestion
    return nestIf(llvm::IsaPred<OpT>,
```

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+                  OpT::getOperationName(), TypeID::get<OpT>());
+  }
+
+  /// Variadic convenience: add sub-pipelines for multiple op types at once.
+  /// Equivalent to calling nest<T>() for each type individually.
+  template <typename T1, typename T2, typename... Rest>
+  void nest() {
+    nest<T1>();
+    nest<T2>();
+    (nest<Rest>(), ...);
```
kuhar:
```suggestion
    (void)(nest<Rest>(), ...);
```
To suppress overzealous compilers / linters.

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+  }
+
+  /// Add a pass to ALL existing sub-pipelines.
+  template <typename F = std::unique_ptr<Pass> (*)()>
+  MultiPipelineNest &addPass(F constructor) {
```
kuhar: How does this default argument type work? If you pass in the parameter, it will be disregarded anyway, no? And there's not default value.

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+    adaptorPass->addPassToAll(constructor);
+    return *this;
+  }
+
+  /// Add a pass to ALL existing sub-pipelines if the predicate is true.
+  template <typename F = std::unique_ptr<Pass> (*)()>
```
kuhar: same here and below seems weird

kuhar: And if you actually need this, use `std::function`, since it already does small object optimization and is unlikely to allocate

compiler/src/iree/compiler/Utils/PassUtils.h
kuhar: Yeah I don't think it does anything because addPassToAll would copy and you can't copy unique_ptr

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -62,57 +334,32 @@ struct PipelineCache {
...
+  MultiOpNest(OpPassManager &parentPm) : nest(parentPm) {
+    nest.template nest<OpTys...>();
   }
 
   // We give the template param a default to support passing overload
   // constructors (i.e. createCanonicalizerPass).
   template <typename F = std::unique_ptr<Pass> (*)()>
   MultiOpNest &addPass(F constructor) {
-    addPassInternal(constructor);
+    nest.addPass(constructor);
```
kuhar: same weird default template arguments here and below

kuhar (commented):
Batch 1

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  LDBG() << "Running op pipeline adaptor synchronously on '"
+         << parentOp->getName() << "'";
+
+  for (Region &region : parentOp->getRegions()) {
+    for (Operation &op : region.getOps()) {
+      // Within a batch, first match wins. Across batches (from merged
+      // adaptors), all matching batches run.
+      std::optional<unsigned> lastMatchedBatch;
+      for (Entry &e : entries) {
+        if (lastMatchedBatch && e.batchIndex == *lastMatchedBatch) {
```
kuhar: nit/optional: the optional operator== already handles this
```suggestion
        if (lastMatchedBatch == e.batchIndex) {
```

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  unsigned numThreads = context->getThreadPool().getMaxConcurrency();
+  // Entries are immutable after pipeline construction, so size comparison
+  // suffices for staleness detection.
+  if (asyncExecutors.empty() ||
+      asyncExecutors.front().size() != entries.size()) {
+    LDBG() << "  Creating " << numThreads << " async executors with "
+           << entries.size() << " entries each";
+    asyncExecutors.clear();
+    asyncExecutors.resize(numThreads);
+    for (SmallVector<Entry> &executor : asyncExecutors) {
```
kuhar: nit/optional: one fewer place to update in case the small size changes in the future
```suggestion
    for (SmallVectorImpl<Entry> &executor : asyncExecutors) {
```

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+
+  SmallVector<OpDispatchInfo> opInfos;
+  for (Region &region : parentOp->getRegions()) {
+    for (Operation &op : region.getOps()) {
+      OpDispatchInfo info;
+      info.op = &op;
+      // Within a batch, first match wins. Across batches, all matching
+      // batches run.
+      std::optional<unsigned> lastMatchedBatch;
+      for (unsigned i = 0, e = entries.size(); i < e; ++i) {
```
kuhar: nit/optional: use enumerate?

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  LDBG() << "Running op pipeline adaptor asynchronously on '"
+         << parentOp->getName() << "'";
+
+  // --- Phase 1: Collect matched operations (sequential). ---
+
+  // Information for a single matched operation. An op may match multiple
+  // entries when batches from merged adaptors are present.
+  struct OpDispatchInfo {
+    SmallVector<unsigned, 2> entryIndices;
+    Operation *op;
```
kuhar:
```suggestion
    SmallVector<unsigned, 2> entryIndices;
    Operation *op = nullptr;
```

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+    active.store(false);
+  }
+  std::atomic<bool> hasFailure(false);
+
+  // NOTE: Using the dynamic pipeline API (Pass::runPipeline) rather than
+  // the internal OpToOpPassAdaptor::runPipeline means instrumentation events
+  // will report the parent thread ID rather than the actual worker thread.
+  // This is acceptable since OpPipelineAdaptorPass is an internal
+  // implementation detail.
+  parallelForEach(context, opInfos, [&](OpDispatchInfo &info) {
```
kuhar, line 211: Maybe worth adding a comment that we assume that this will never exceed `numThreads`?

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: Probably not. If we wanted to derive something we'd want a more generic base.

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+void MultiPipelineNest::commitPass() {
+  if (!ownedPass) {
+    return; // Already committed.
+  }
+  assert(parentPm->size() == parentPmSizeAtConstruction &&
+         "passes were added to the parent PM between MultiPipelineNest "
+         "construction and commitPass(); call commitPass() before adding "
+         "parent passes");
+  parentPm->addPass(std::move(ownedPass));
+  // adaptorPass remains valid — it now points into the PM.
```
kuhar: Do we need to guard against inserting empty adaptors like in the destructor?

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  if (!adaptorPass) {
+    return;
+  }
+  // Already committed to the PM — nothing to do.
+  if (!ownedPass) {
+    return;
+  }
+  // No entries were added — discard the pass silently.
+  if (adaptorPass->empty()) {
+    return;
```
kuhar: combine this with the very first `if`?

kuhar (commented):
Batch 2

compiler/src/iree/compiler/Utils/unittests/PassUtilsTest.cpp
```diff
@@ -0,0 +1,783 @@
...
+
+//===----------------------------------------------------------------------===//
+// Helpers
+//===----------------------------------------------------------------------===//
+
+/// Create a module containing private functions with the given names.
+static OwningOpRef<ModuleOp> createModule(MLIRContext &context,
+                                          ArrayRef<StringRef> funcNames) {
+  Builder builder(&context);
+  OwningOpRef<ModuleOp> module(ModuleOp::create(UnknownLoc::get(&context)));
```
kuhar: don't use `module` as a variable name -- it's going to become a keyword in C++20

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,321 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  predAdaptor.mergeFrom(*adaptorPass);
+  return true;
+}
+
+OpPassManager &MultiPipelineNest::nestIf(ConditionFn condition,
+                                         StringRef anchorOpName,
+                                         std::optional<TypeID> opTypeID) {
+  return adaptorPass->addEntry(std::move(condition), anchorOpName, opTypeID);
+}
+
```
kuhar, line 351: Should we add a test for multiple op types in `MultiOpNest<T0, T1>`?

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,7 +59,270 @@ struct PipelineCache {
...
+///
+/// Usage:
+///   // Typical: temporary destroyed at semicolon — pass inserted here.
+///   MultiPipelineNest(pm).nest<FuncOp>().addPass(createMyPass);
+///
+///   // Named variable with explicit commit point:
+///   MultiPipelineNest nest(pm);
+///   nest.nest<FuncOp>();
+///   nest.commitPass();           // insert now, before subsequent passes
+///   pm.addPass(createOtherPass); // guaranteed to be after nest's pass
```
kuhar:
```suggestion
///   nest.commitPass();           // Insert now, before subsequent passes.
///   pm.addPass(createOtherPass); // Guaranteed to be after nest's pass.
```

compiler/src/iree/compiler/Utils/PassUtils.cpp
kuhar: Also move semantics -- the logic seems nontrivial

kuhar (commented):
Batch 3

kuhar:
Overall, I'm worried this approach comes with footguns around non-trivial destructions / commit semantics. I wonder if there' some simpler (but possibly more verbose) alternative like https://github.com/iree-org/iree/pull/23620#discussion_r2908857615

qedawkins:
> Overall, I'm worried this approach comes with footguns around non-trivial destructions / commit semantics. I wonder if there' some simpler (but possibly more verbose) alternative like [#23620 (comment)](https://github.com/iree-org/iree/pull/23620#discussion_r2908857615)

I tried really hard to avoid it but without the ability to do the finalization step that PassManager does to merge adjacent nests there's no way around it. Consider two adjacent pipeline nests
```
nest<FunctionOpInterface>(passA);
nest<FunctionOpInterface>(passB);
```

These two passes may be combined into the same nest because the condition under which an operation will run a pass are the same (i.e. the module level barrier can be dropped). With multi-pipeline nests it works the same way, but we need to prove that the set of conditions for adjacent nests are identical. If we want to do that when we create the nest, we need to either merge passes immediately before running the pipeline (what `finalize` does in the PassManager with OpToOpAdaptor) or we need to try to do the merge as soon as the pass is inserted in the pipeline by looking at the immediate predecessor to the pass. If we insert the pass on construction of the MultiPipelineNest because we haven't setup any of the nesting conditions yet we can't do the merge. So either we need the API to require users to setup all nesting conditions up front or to manually commit the pass to the pipeline. I have a use case that requires returning the pipeline nest so I don't want to remove the commit, but I can add a version similar to what you're asking for, just know that it is its own form of footgun (losing out on adjacent pipeline nest merging).

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: I'm actually unsure. OpToOpPassAdaptor doesn't verify this for named nests so I didn't here. I don't immediately know what could go wrong with non-IsolatedFromAbove nests other than it being a massive footgun.

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: `std::unique_ptr<Pass> (*)()` is a function that returns a unique_ptr, not a unique_ptr directly, unless I'm misunderstanding your comment. So the unique_ptr isn't created until we invoke the constructor which happens only once we actually add the pass to the pipeline.

And I ran into issues trying to use `std::function` since `createCanonicalizerPass` among others are overloaded and couldn't resolve to a single function.

kuhar: I'm just really confused why this default type is needed at all

qedawkins: Ah it's because `createCanonicalizerPass` is overloaded with either no args or `GreedyRewriteConfig` and without the default type type deduction fails.

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,68 +59,310 @@ struct PipelineCache {
...
+  template <typename T1, typename T2, typename... Rest>
+  void nest() {
+    nest<T1>();
+    nest<T2>();
+    (void)(nest<Rest>(), ...);
   }
 
-  // We give the template param a default to support passing overload
-  // constructors (i.e. createCanonicalizerPass).
-  template <typename F = std::unique_ptr<Pass> (*)()>
```
qedawkins, line 71: @kuhar this is how MultiOpNest is currently implemented also, so that default type isn't a new thing.

compiler/src/iree/compiler/Utils/PassUtils.cpp
IanWood1: (feel free to ignore this. `make_unique<T[]>` is fine with me. I'm just interested in what's actually allowed for `std::vector`)

I also felt like something was wrong when looking at this. It seems that the implementations of SmallVector and std::vector differ, allowing std::vector to be initialized with types that are default construct-able but not copy assignable or move constructible. However, I think this behavior is explicitly allowed by the spec.

From [cppreference](https://en.cppreference.com/w/cpp/container/vector.html#:~:text=17%29-,The,requirements%2E,-Feature):
> The requirements that are imposed on the elements depend on the actual operations performed on the container. Generally, it is required that element type meets the requirements of [Erasable](https://en.cppreference.com/w/cpp/named_req/Erasable.html), but many member functions impose stricter requirements. This container (but not its members) can be instantiated with an incomplete element type if the allocator satisfies the [allocator completeness requirements](https://en.cppreference.com/w/cpp/named_req/Allocator.html#Allocator_completeness_requirements).

Looks like the restrictions were loosened in C++ 11:
> T must meet the requirements of CopyAssignable and CopyConstructible. | (until C++11)

compiler/src/iree/compiler/Utils/PassUtils.h
kuhar: Ack. I'd much rather have a helper function that disambiguates the canonicalizer pass instead by calling the correct overload. This seems like one big hack.

qedawkins: I'd want to do that separately then so that this PR isn't going and changing the canonicalizer passes everywhere.

kuhar: Maybe pre-commit this? (Should be easy with claude, hopefully, and quick to stamp)

qedawkins: Ok actually it's not just `canonicalizer` passes. It's every pass that has an options struct. Adding disambiguating helpers for everything is super ugly...

kuhar: OK, so leave as-is but add a comment explaining what this is for?

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,318 @@ void signalFixedPointModified(Operation *rootOp) {
...
+MultiPipelineNest::MultiPipelineNest(MultiPipelineNest &&other)
+    : parentPm(other.parentPm),
+      parentPmSizeAtConstruction(other.parentPmSizeAtConstruction),
+      ownedPass(std::move(other.ownedPass)), adaptorPass(other.adaptorPass) {
+  other.parentPm = nullptr;
+  other.adaptorPass = nullptr;
+}
+
+MultiPipelineNest &MultiPipelineNest::operator=(MultiPipelineNest &&other) {
+  if (this != &other) {
```
kuhar: nit: flip this and return early instead

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,318 @@ void signalFixedPointModified(Operation *rootOp) {
...
+               "adaptor at the desired position before adding parent passes");
+        parentPm->addPass(std::move(ownedPass));
+      }
+    }
+    parentPm = other.parentPm;
+    parentPmSizeAtConstruction = other.parentPmSizeAtConstruction;
+    ownedPass = std::move(other.ownedPass);
+    adaptorPass = other.adaptorPass;
+    other.parentPm = nullptr;
+    other.adaptorPass = nullptr;
```
kuhar: also reset `other.parentPmSizeAtConstruction` ?

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,318 @@ void signalFixedPointModified(Operation *rootOp) {
...
+      if (!tryMergeIntoPredecessor()) {
+        assert(parentPm->size() == parentPmSizeAtConstruction &&
+               "passes were added to the parent PM between MultiPipelineNest "
+               "construction and destruction; use commitPass() to insert the "
+               "adaptor at the desired position before adding parent passes");
+        parentPm->addPass(std::move(ownedPass));
+      }
+    }
+    parentPm = other.parentPm;
+    parentPmSizeAtConstruction = other.parentPmSizeAtConstruction;
```
kuhar: Don't we have to track the delta after the possible flush above?

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,318 @@ void signalFixedPointModified(Operation *rootOp) {
...
+    }
+  }
+
+  // --- Phase 3: Parallel dispatch. ---
+
+  // Track which executors are in use (one per thread). Using unique_ptr<T[]>
+  // because std::atomic is neither copyable nor moveable, which is required
+  // by std::vector's contract even if we never resize.
+  auto activeExecutors =
+      std::make_unique<std::atomic<bool>[]>(asyncExecutors.size());
```
kuhar, line 197: this is not guaranteed to zero-initialize atomics, you still need the for loop: https://cppreference.com/w/cpp/atomic/atomic/atomic.html

compiler/src/iree/compiler/Utils/PassUtils.cpp
```diff
@@ -24,4 +26,318 @@ void signalFixedPointModified(Operation *rootOp) {
...
+  return *this;
+}
+
+void MultiPipelineNest::commitPass() {
+  if (!ownedPass) {
+    return; // Already committed.
+  }
+  // No entries were added -- discard the pass silently.
+  if (adaptorPass->empty()) {
+    ownedPass.reset();
```
kuhar, line 312: do we also want to invalidate the adaptorPass?

compiler/src/iree/compiler/Utils/unittests/PassUtilsTest.cpp
```diff
@@ -0,0 +1,876 @@
...
+  // Both passes must have run, in order.
+  int firstOrder = getFuncAttrInt(*moduleOp, "alpha", "test.first");
+  int secondOrder = getFuncAttrInt(*moduleOp, "alpha", "test.second");
+  EXPECT_EQ(firstOrder, 0);
+  EXPECT_EQ(secondOrder, 1);
+}
+
+TEST(MultiOpNestTest, MergeSubsetTypes) {
+  // A 2-type nest followed by a 3-type superset nest should merge. The
+  // extra type in the second nest becomes a new entry in the combined pass.
```
kuhar: Wait, how is testing more than one-type nests? I don't get see the `3-type superset`.

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,68 +59,310 @@ struct PipelineCache {
...
+  /// automatic merging).
+  bool allEntriesHaveTypeID() const;
+
+  /// Add a pass to all existing sub-pipelines. Templated to support both
+  /// function pointers (createCanonicalizerPass) and lambdas. A non-template
+  /// std::function signature cannot resolve overloaded pass constructors, and
+  /// adding a function-pointer overload creates ambiguity with lambdas.
+  template <typename Fn>
+  void addPassToAll(Fn constructor) {
+    for (Entry &entry : entries) {
```
kuhar, line 152: would it be worth asserting that there's at least one entry?

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,68 +59,310 @@ struct PipelineCache {
...
+      entry.pipeline.addPass(constructor());
+    }
+  }
+
+  /// Add a pass only to sub-pipelines targeting the given op type.
+  template <typename Fn>
+  void addPassToEntriesWithTypeID(TypeID targetID, Fn constructor) {
+    for (Entry &entry : entries) {
```
kuhar, line 160: similar here

compiler/src/iree/compiler/Utils/PassUtils.h
```diff
@@ -50,68 +59,310 @@ struct PipelineCache {
...
+/// Usage:
+///   using FunctionLikeNest = MultiOpNest<IREE::Util::InitializerOp,
+///                                        IREE::Util::FuncOp>;
+///
+///   FunctionLikeNest(passManager)
+///     .addPass(createMyPass)
+///     .addPredicatedPass(enable, createMyOtherPass);
+template <typename... OpTys>
+struct MultiOpNest {
+  MultiOpNest(OpPassManager &parentPm) : nest(parentPm) {
```
kuhar:
```suggestion
  explicit MultiOpNest(OpPassManager &parentPm) : nest(parentPm) {
```

compiler/src/iree/compiler/Utils/PassUtils.h
kuhar: we can have a `using` declaration for this function pointer type and put the comment there

kuhar (approved):
LGTM % some small issues / suggestions. I don't love some of the potential footguns, but at least we have the documented, and I don't see an easy way around them...

compiler/src/iree/compiler/Utils/unittests/PassUtilsTest.cpp
```diff
@@ -0,0 +1,876 @@
...
+  }
+
+  ASSERT_TRUE(succeeded(pm.run(moduleOp.get())));
+  // Both passes should have run.
+  int firstOrder = getFuncAttrInt(*moduleOp, "alpha", "test.first");
+  int secondOrder = getFuncAttrInt(*moduleOp, "alpha", "test.second");
+  EXPECT_EQ(firstOrder, 0);
+  EXPECT_EQ(secondOrder, 1);
+}
+
```
kuhar, line 892: Can we also test this:
```c++
  MultiPipelineNest nest1(pm); 
  MultiPipelineNest nest2(pm);
  nest1.commitPass();
```
to exercise parent size updates / checks.

compiler/src/iree/compiler/Utils/PassUtils.h
qedawkins: Sure, makes sense to me, will do.

compiler/src/iree/compiler/Utils/PassUtils.cpp
qedawkins: Done by nulling the pass managers.

compiler/src/iree/compiler/Utils/unittests/PassUtilsTest.cpp
qedawkins: done

compiler/src/iree/compiler/Utils/PassUtils.cpp
kuhar: this not resolved yet

compiler/src/iree/compiler/Utils/PassUtils.cpp
kuhar: this not resolved yet

qedawkins: what, how did it get undone. Sorry about that.

kuhar:
https://github.com/iree-org/iree/actions/runs/22935188854/job/66565243186?pr=23620#step:5:10072

OK, I think we need to push it into the `,` operator since older compilers complain:
```c++
(((void)nest<Rest>()), ...);
```

or alternatively make sure it's never empty:
```c++
(void)(nest<Rest>(), ..., false);
```

kuhar:
eh, seems like some compilers are being obtuse here, just drop the void then maybe -- if we get warnings, we can address in the future

PR #23621, opened 2026-03-03
Add F16 support to contraction input demotion pass

HanKuanChen:


compiler/src/iree/compiler/GlobalOptimization/DemoteContractionInputs.cpp
```diff
@@ -135,30 +141,52 @@ struct DemoteContractionInputsToBF16Pattern
   DemotionOption demoteOption;
 };
 
-class DemoteContractionInputsToBF16Pass
-    : public impl::DemoteContractionInputsToBF16PassBase<
-          DemoteContractionInputsToBF16Pass> {
+template <typename PassBase, typename PatternType>
+class DemoteContractionInputsPassImpl : public PassBase {
```
MaheshRavishankar: This is too much templating. I dont have a strong opinion here, but instead of two separate passes, can we just have a single pass with pass options that will tell you which type to demote to?

HanKuanChen: The option already has optional value (`all|conv|matmul|none`).
```
#define DEMOTE_CONTRACTION_INPUTS_CL_OPT(TYPE, TYPE_UPPER)                     \
  static llvm::cl::opt<DemotionOption>                                         \
      clDemoteContractionInputsTo##TYPE_UPPER##Strategy(                       \
          "iree-global-opt-enable-demote-contraction-inputs-to-" #TYPE,        \
          llvm::cl::desc(                                                      \
              "Demotes inputs (LHS, RHS) of contraction ops to " #TYPE_UPPER   \
              ". "                                                             \
              "Selects types of contraction ops to demote."),                  \
          llvm::cl::values(clEnumValN(DemotionOption::All, "all",              \
                                      "Demote all contraction ops."),          \
                           clEnumValN(DemotionOption::Conv, "conv",            \
                                      "Only demote convolution ops."),         \
                           clEnumValN(DemotionOption::Matmul, "matmul",        \
                                      "Only demote matmul ops."),              \
                           clEnumValN(DemotionOption::None, "none",            \
                                      "Demote no contraction ops.")),          \
          llvm::cl::init(DemotionOption::None))

DEMOTE_CONTRACTION_INPUTS_CL_OPT(bf16, BF16);
DEMOTE_CONTRACTION_INPUTS_CL_OPT(f16, F16);
```
If we want to combine this, we can use something like
```
--iree-global-opt-demote-contraction-inputs-type=(bf16|f16)
--iree-global-opt-demote-contraction-inputs-operations=(all|conv|matmul|none)
```
Do you want this version?

HanKuanChen: I submit a new version. I think more types will be supported in the future (e.g., fp8).

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
```diff
@@ -44,17 +44,25 @@ static llvm::cl::opt<bool> clEnableEarlyMaterialization(
...
-    llvm::cl::desc("Demotes inputs (LHS, RHS) of contraction ops to BF16. "
-                   "Selects types of contraction ops to demote."),
-    llvm::cl::values(
-        clEnumValN(DemotionOption::All, "all", "Demote all contraction ops."),
-        clEnumValN(DemotionOption::Conv, "conv",
-                   "Only demote convolution ops."),
-        clEnumValN(DemotionOption::Matmul, "matmul", "Only demote matmul ops."),
-        clEnumValN(DemotionOption::None, "none", "Demote no contraction ops.")),
-    llvm::cl::init(DemotionOption::None));
+static llvm::cl::opt<DemoteType> clDemoteContractionInputsType(
```
MaheshRavishankar, line 47: Would be good to add a default value for this as we../

MaheshRavishankar (approved):

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
HanKuanChen: Setting f16 or bf16 as a default feels a bit awkward. The alternative would be adding an all option, but since this pass only executes once, it can't handle demoting both types simultaneously.

HanKuanChen:
ping

HanKuanChen:
solve merge conflict due to 2a0b797f5435c2123dc9674eddaa569c20acea88

HanKuanChen:
@MaheshRavishankar @hanhanW Hello, since I don't have merge permission, can you merge this?

kuhar:
Seems like this broke the MSVC builds: https://github.com/iree-org/iree/actions/runs/23209911746/job/67455603181

PR #23625, opened 2026-03-03
[VectorDistribution] Relax layout size constraint

sommerlukas:
Relax invariant on the nested layout attribute to allow the overall size of the layout to exceed the size of the underlying tensor.

By allowing the layout to exceed the size of the tensor, we can select tile sizes friendly to hardware even if the tensor itself has an odd shape, i.e., a size not divisible by HW-friend tile sizes.

The additional elements will be masked out by code generation.

This change also makes sure that the vectorization of `to_layout` operations inserts masks on `transfer_read/write`.

It also modifies the tensor layout configuration pass to use ceil-division in these cases to ensure full coverage of the tensor.

Assisted-by: Claude Code

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
@@ -416,6 +416,9 @@ LogicalResult NestedLayoutAttr::isValidLayout(ShapedType shapeTy,
            << shape.size() << ") does not match rank of layout (" << rank
            << ").";
   }
+  if (isa<RankedTensorType>(shapeTy)) {
+    return success();
+  }
   // Multiply all shapes in the layout.
   for (int i = 0, e = rank; i < e; ++i) {
     int64_t expectedShape = getSubgroupTile()[i] * getBatchTile()[i] *
@@ -423,7 +426,7 @@ LogicalResult NestedLayoutAttr::isValidLayout(ShapedType shapeTy,
                             getElementTile()[i];
     // The layout size exceeding the shape is allowed, because it will be
     // handled through padding/masking.
-    if (ShapedType::isStatic(shape[i]) && expectedShape < shape[i]) {
+    if (ShapedType::isStatic(shape[i]) && expectedShape != shape[i]) {
       std::string layoutStr;
       llvm::raw_string_ostream layoutOs(layoutStr);
       printStripped(layoutOs);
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
```diff
@@ -421,7 +421,9 @@ LogicalResult NestedLayoutAttr::isValidLayout(ShapedType shapeTy,
     int64_t expectedShape = getSubgroupTile()[i] * getBatchTile()[i] *
                             getOuterTile()[i] * getThreadTile()[i] *
                             getElementTile()[i];
-    if (ShapedType::isStatic(shape[i]) && expectedShape != shape[i]) {
+    // The layout size exceeding the shape is allowed, because it will be
+    // handled through padding/masking.
+    if (ShapedType::isStatic(shape[i]) && expectedShape < shape[i]) {
       std::string layoutStr;
       llvm::raw_string_ostream layoutOs(layoutStr);
       printStripped(layoutOs);
```
Groverkss, line 432: Honestly, I'm ok with deleting this layout size verification. Maybe we can move it to just checking on vectors? And allow anything on tensors? wdyt?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConfigureTensorLayouts.cpp
```diff
@@ -573,8 +574,12 @@ static LogicalResult setGPULoweringConfigLayout(
   // Use thread tile sizes as the vector width for each thread.
   SmallVector<int64_t> threadTileSizes = config.getStaticTilingLevelSizes(
       llvm::to_underlying(IREE::GPU::TilingLevel::Thread), candidate);
+  // Use ceil division here to make sure that we cover the entire tensor with
+  // the layout. If the tensor size isn't divisible by the chosen tile sizes, we
+  // want to have the vector larger than the tensor and mask out the excess
+  // elements.
   FailureOr<SmallVector<int64_t>> elementTile =
-      divideTile(bounds, threadTileSizes);
+      divideTile(bounds, threadTileSizes, /*ceilDiv=*/true);
```
Groverkss: Shouldn't we just switch everything to using ceilDiv? It should be okay to over allocate I think?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtAttrs.cpp
sommerlukas: I had experimented with it earlier, it works for the masking use case. I've changed it now to skip size verification entirely for tensors and keep the strict equality check for vectors.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConfigureTensorLayouts.cpp
sommerlukas: Done, it probably makes sense to accumulate it in the batch dimension if size doesn't exactly match the subgroup and thread tile sizes.

Groverkss (approved):

PR #23627, opened 2026-03-03
[Build scripts] Exclude async platform on unsupported targets

amd-eochoalo:
The build script assumed the left over platform was macOS, but we also have emscripten. This meant that iree::async::platform::async got linked with emscripten causing a CMake configuration error.

Assisted-by: Claude


amd-eochoalo:
Note: there should be a test_emscripten CI tag.

amd-eochoalo:
Sample run: https://github.com/iree-org/iree/actions/runs/22628918325

amd-eochoalo:
Sample run is succeeding.

kuhar (approved):
Makes sense but let's wait for for @benvanik

amd-eochoalo:
ping @benvanik

PR #23630, opened 2026-03-03
[Codegen] Fix layout analysis fixup crashes

Groverkss:
Earlier this would crash if something didn't get a layout, which doesn't happen today, but was introduced in a future patch.


sommerlukas (approved):
LGTM.

PR #23631, opened 2026-03-03
[GPU] Make GPUVectorAlloc allocate shared memory based on layout analysis

Groverkss:
Before this patch, we were doing in hack in layout configuration where we checked if there would be a layout conflict and try to promote the operand for attention. This hack was fragile as well as doesn't really scale beyond simple stuff.

This patch runs vector layout analysis to check if there would be any conflicts and uses shared memory to allocate shared memory for them.

Also removes some dead code from the alloc pass.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
@@ -4,6 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include "iree/compiler/Codegen/Common/GPU/GPUPatterns.h"
 #include "iree/compiler/Codegen/Common/GPU/Passes.h"
 #include "iree/compiler/Codegen/Common/Transforms.h"
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.h"
@@ -20,8 +21,8 @@
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/Interfaces/FunctionInterfaces.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Transforms/Passes.h"
+#include "mlir/Transforms/WalkPatternRewriteDriver.h"
 
 namespace mlir::iree_compiler {
 
@@ -29,48 +30,6 @@ namespace mlir::iree_compiler {
 #include "iree/compiler/Codegen/Common/GPU/Passes.h.inc"
 
 namespace {
-
-/// Convert stretching broadcasts (broadcasting a non-unit dim from 1) into
-/// broadcast + transpose so the layout analysis can handle them.
-/// Copied from LLVMGPUVectorDistribute.cpp::RemoveUnitDimStrechingBroadcast.
-struct RemoveStretchingBroadcast final : OpRewritePattern<vector::BroadcastOp> {
-  using Base::Base;
-
-  LogicalResult matchAndRewrite(vector::BroadcastOp broadcastOp,
-                                PatternRewriter &rewriter) const override {
-    SetVector<int64_t> stretchedDims = broadcastOp.computeBroadcastedUnitDims();
-    if (stretchedDims.empty()) {
-      return failure();
-    }
-
-    VectorType srcTy = cast<VectorType>(broadcastOp.getSource().getType());
-    VectorType dstTy = broadcastOp.getResultVectorType();
-    int64_t numLeadingBroadcastDims = dstTy.getRank() - srcTy.getRank();
-    SmallVector<int64_t> broadcastedUnitDims;
-    for (auto i : llvm::seq<int64_t>(numLeadingBroadcastDims)) {
-      if (dstTy.getShape()[i] == 1) {
-        broadcastedUnitDims.push_back(i);
-      }
-    }
-    if (stretchedDims.size() > broadcastedUnitDims.size()) {
-      return failure();
-    }
-
-    auto perm = llvm::to_vector(llvm::seq<int64_t>(dstTy.getRank()));
-    for (auto [stretchedDim, broadcastedUnitDim] :
-         llvm::zip(stretchedDims.getArrayRef(), broadcastedUnitDims)) {
-      std::swap(perm[stretchedDim], perm[broadcastedUnitDim]);
-    }
-    VectorType permutedBroadcastTy = VectorType::get(
-        applyPermutation(dstTy.getShape(), perm), dstTy.getElementType());
-    Value permutedBroadcast = vector::BroadcastOp::create(
-        rewriter, broadcastOp.getLoc(), permutedBroadcastTy,
-        broadcastOp.getSource());
-    rewriter.replaceOpWithNewOp<vector::TransposeOp>(broadcastOp,
-                                                     permutedBroadcast, perm);
-    return success();
-  }
-};
 
 // Allocates a tensor to copy the vector into a la bufferization.alloc_tensor.
 // This allocation is always static as vectors are currently always static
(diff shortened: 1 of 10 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -27,33 +30,47 @@ namespace mlir::iree_compiler {
 
 namespace {
 
-// For optimal performance we always want to copy 128 bits.
-constexpr int copyVectorNumBits = 128;
+/// Convert stretching broadcasts (broadcasting a non-unit dim from 1) into
+/// broadcast + transpose so the layout analysis can handle them.
+/// Copied from LLVMGPUVectorDistribute.cpp::RemoveUnitDimStrechingBroadcast.
+struct RemoveStretchingBroadcast final : OpRewritePattern<vector::BroadcastOp> {
```
Groverkss: This could probably go in a single place, but I decided to just copy it for now... Happy to move it to some other place. Not sure where.

krzysz00: I'm sure there's a Transforms.cpp running around somewhere?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -96,51 +113,88 @@ static Value readVectorFromTensor(OpBuilder &b, VectorType vectorType,
...
+    OpOperand &operand = op.getInputMutable();
+    FailureOr<Value> ret =
+        allocateTensorForVector(builder, op->getLoc(), operand.get());
+    if (failed(ret)) {
+      return failure();
+    }
+
+    // Synchronize after the write to shared memory before we read from it.
+    auto synced =
+        IREE::GPU::ValueBarrierOp::create(builder, op->getLoc(), *ret);
```
krzysz00, line 109: @qedawkins Is this the sort of thing that would justify memory spaces on value barrier, or will that get handled by the tensor getting the workgroup address space during bufferization?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -96,51 +113,88 @@ static Value readVectorFromTensor(OpBuilder &b, VectorType vectorType,
...
+materializeSharedMemoryConversions(FunctionOpInterface funcOp) {
+  SmallVector<IREE::VectorExt::ToLayoutOp> opsToPromote;
+  funcOp.walk([&](IREE::VectorExt::ToLayoutOp op) {
+    if (op.getSharedMemoryConversion()) {
+      opsToPromote.push_back(op);
+    }
+  });
+
+  for (IREE::VectorExt::ToLayoutOp op : opsToPromote) {
+    OpBuilder builder(op);
```
krzysz00: IIRC, OpBuilders are fairly expensive - you might want to allocate an IRRewriter outside the walk and `setInsertionPoint` it in here.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -96,51 +113,88 @@ static Value readVectorFromTensor(OpBuilder &b, VectorType vectorType,
...
+    // reads in the previous iteration of a loop. We set this barrier
+    // at the start of this block.
+    builder.setInsertionPointToStart(op->getBlock());
+    gpu::BarrierOp::create(builder, op->getLoc(), gpu::AddressSpace::Workgroup);
+
+    // Promote both of the input operands, excluding the accumulator.
+    builder.setInsertionPoint(op);
+    OpOperand &operand = op.getInputMutable();
+    FailureOr<Value> ret =
+        allocateTensorForVector(builder, op->getLoc(), operand.get());
```
krzysz00, line 102: Long-term note: once we stop having bank conflict avoidance be all-or-nothing, where will the logic that knows whether we need to pad or swizzle this tensor for bank conflicts go?

krzysz00 (approved):
I've got some high-level questions, but don't see any problems here as it stands.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
Groverkss: We can explicitly say here what kind of swizzling we want. We know the exact way the input is written/read. I want to plumb swizzling here as well, just need more exploration.

But based on the layouts here, we can determine optimal way to avoid bank conflicts.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -96,51 +113,88 @@ static Value readVectorFromTensor(OpBuilder &b, VectorType vectorType,
...
+    // HACK: Until proper barrier placement is handled later we have to
+    // synchronize explicitly in this pass.
+
+    // Synchronize before the write to shared memory to avoid stepping over
+    // reads in the previous iteration of a loop. We set this barrier
+    // at the start of this block.
+    builder.setInsertionPointToStart(op->getBlock());
+    gpu::BarrierOp::create(builder, op->getLoc(), gpu::AddressSpace::Workgroup);
+
+    // Promote both of the input operands, excluding the accumulator.
```
sommerlukas: What are "both of the input operands" here? Maybe I'm overlooking something, but it looks like only one operand is handled here.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -96,51 +113,88 @@ static Value readVectorFromTensor(OpBuilder &b, VectorType vectorType,
...
-      OpOperand &operand = op.getInputMutable();
-      FailureOr<Value> ret =
-          allocateTensorForVector(builder, op->getLoc(), operand.get());
-      if (failed(ret)) {
+    // Remove stretching broadcasts before layout analysis — the analysis
+    // asserts that broadcasts don't stretch.
+    {
+      RewritePatternSet patterns(funcOp.getContext());
+      patterns.add<RemoveStretchingBroadcast>(funcOp.getContext());
+      if (failed(applyPatternsGreedily(funcOp, std::move(patterns)))) {
```
sommerlukas: Do we need a greedy driver here or could we use `walkAndApply`?

sommerlukas (approved):
LGTM, just minor stuff.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
Groverkss: moved builder out of for loop, there is no reason for a rewriter though.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
Groverkss: oops, this is an old comment.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
Groverkss: added a todo comment about this.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
Groverkss: good catch, thanks

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
Groverkss: Moved to GPUPatterns.h for now as VectorLayout canonicalizations.

PR #23632, opened 2026-03-03
[Torch] Add flag to enable shape refinement

IanWood1:
Allows enabling shape refinement for the torch pipeline which is useful when the input torch MLIR is not as static as it could be. For example, Fusilli's custom ops (https://github.com/iree-org/fusilli/pull/188) use dynamic shapes to prevent having to find and replace each dim size.

### DIFF
--- a/compiler/plugins/input/Torch/InputConversion/Passes.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/Passes.cpp
@@ -50,6 +50,9 @@ void createTorchToIREEPipeline(
       torch::Torch::createReduceOpVariantsPass(llvm::StringRef()));
   pm.addNestedPass<func::FuncOp>(
       mlir::torch::TorchConversion::createConvertCustomQuantOpPass());
+  if (options.enableShapeRefinement) {
+    torch::Torch::createTorchShapeRefinementPipeline(pm, /*options=*/{});
+  }
   if (options.decompose) {
     pm.addNestedPass<func::FuncOp>(
         torch::Torch::createDecomposeComplexOpsPass(BackendLegalOps::get()));
@@ -57,9 +60,6 @@ void createTorchToIREEPipeline(
   pm.addNestedPass<func::FuncOp>(torch::Torch::createFuseQuantizedOpsPass());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(torch::Torch::createScalarizeShapesPass());
-  if (options.enableShapeRefinement) {
-    torch::Torch::createTorchShapeRefinementPipeline(pm, /*options=*/{});
-  }
   pm.addNestedPass<func::FuncOp>(torch::createConvertTorchToTMTensorPass());
   pm.addNestedPass<func::FuncOp>(
       TorchInput::createConvertTMTensorToLinalgExtPass());
--- a/compiler/plugins/input/Torch/InputConversion/Passes.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/Passes.cpp
@@ -57,6 +57,9 @@ void createTorchToIREEPipeline(
   pm.addNestedPass<func::FuncOp>(torch::Torch::createFuseQuantizedOpsPass());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(torch::Torch::createScalarizeShapesPass());
+  if (options.enableShapeRefinement) {
+    torch::Torch::createTorchShapeRefinementPipeline(pm, /*options=*/{});
+  }
   pm.addNestedPass<func::FuncOp>(torch::createConvertTorchToTMTensorPass());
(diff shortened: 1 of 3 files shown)


compiler/plugins/input/Torch/InputConversion/Passes.cpp
```diff
@@ -57,6 +57,9 @@ void createTorchToIREEPipeline(
   pm.addNestedPass<func::FuncOp>(torch::Torch::createFuseQuantizedOpsPass());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(torch::Torch::createScalarizeShapesPass());
+  if (options.enableShapeRefinement) {
```
zjgarvey: This should happen before decompositions (many check for static dims) We have an agreement in the torch-mlir community that the decompose complex ops patterns should propagate and emit correct types for all created ops.

zjgarvey (commented):
I'm fine with adding this option. I assume fusilli is emitting torch-ir but can't get intermediate shapes because they don't have the torch utilities to run shape prop on an fx graph? I'm trying to understand the intended use-case here.

sjain-stanford:
@zjgarvey I could help answer your first question on the use-case. The idea is to enable a fast path in Fusilli to add custom ops which (contrary to the built-in ops) embed the MLIR asm directly, so bypassing fusilli's validation, shape/type inference and asm emitter entirely. 

For this to work properly, the custom MLIR asms bake the ops but not the shape/types, to allow for portability. 

Here's an example of an SDPA custom op (we have explicit types but it doesn't need to):
```mlir
  func.func private @sdpa(
      %arg0: !torch.vtensor<[?,?,?,?],f16>,
      %arg1: !torch.vtensor<[?,?,?,?],f16>,
      %arg2: !torch.vtensor<[?,?,?,?],f16>)
      -> !torch.vtensor<[?,?,?,?],f16> {
    %none_mask = torch.constant.none
    %dropout = torch.constant.float 0.000000e+00
    %is_causal = torch.constant.bool false
    %scale = torch.constant.float 1.250000e-01
    %enable_gqa = torch.constant.bool false
    %0 = torch.aten.scaled_dot_product_attention %arg0, %arg1, %arg2,
        %none_mask, %dropout, %is_causal, %scale, %enable_gqa :
        !torch.vtensor<[?,?,?,?],f16>, !torch.vtensor<[?,?,?,?],f16>,
        !torch.vtensor<[?,?,?,?],f16>, !torch.none, !torch.float, !torch.bool,
        !torch.float, !torch.bool -> !torch.vtensor<[?,?,?,?],f16>
    return %0 : !torch.vtensor<[?,?,?,?],f16>
  }
```

The templated shapes and types here get materialized based on user-specified inputs/output tensor metadata (https://gist.github.com/sjain-stanford/8cabe6436e3ba5af30b0878d6926c053#file-iree-compile-input-mlir).

```mlir
  func.func @main(%sdpa_OUT_0_: !torch.tensor<[1,8,64,64],f16>, %k: !torch.vtensor<[1,8,64,64],f16>, %q: !torch.vtensor<[1,8,64,64],f16>, %v: !torch.vtensor<[1,8,64,64],f16>) attributes {torch.assume_strict_symbolic_shapes} {
  
    ...

    %q_sdpa_i0_dyn = torch.tensor_static_info_cast %q_sdpa_i0_perm : !torch.vtensor<[1,8,64,64],f16> to !torch.vtensor<[?,?,?,?],f16>

 ...
  
    %k_sdpa_i1_dyn = torch.tensor_static_info_cast %k_sdpa_i1_perm : !torch.vtensor<[1,8,64,64],f16> to !torch.vtensor<[?,?,?,?],f16>

  ...
  
    %v_sdpa_i2_dyn = torch.tensor_static_info_cast %v_sdpa_i2_perm : !torch.vtensor<[1,8,64,64],f16> to !torch.vtensor<[?,?,?,?],f16>

    %sdpa_OUT_0_sdpa_dyn = func.call @sdpa(%q_sdpa_i0_dyn, %k_sdpa_i1_dyn, %v_sdpa_i2_dyn) : (!torch.vtensor<[?,?,?,?],f16>, !torch.vtensor<[?,?,?,?],f16>, !torch.vtensor<[?,?,?,?],f16>) -> !torch.vtensor<[?,?,?,?],f16>
    %sdpa_OUT_0_sdpa_perm = torch.tensor_static_info_cast %sdpa_OUT_0_sdpa_dyn : !torch.vtensor<[?,?,?,?],f16> to !torch.vtensor<[1,8,64,64],f16>

...

    return
  }
```

zjgarvey (approved):
I'm fine with this change. It will probably be useful elsewhere, too.

If you care to hear it, I think that this approach (for Fusilli CustomOp IR generation) might be a mistake.

If you want, we can discuss this elsewhere, since I still don't mind this IREE change in isolation and don't plan on blocking it.

keshavvinayak01 (commented):
I missed this change, looks like a good step!

PR #23633, opened 2026-03-03
[Codegen] Remove attention transpose intrinsic hacks

Groverkss:
After https://github.com/iree-org/iree/pull/23631 we will be using the analysis to determine conflicts, which allows us to undo a bunch of hacks we did earlier to not have conflicts between the chained mma in attention.

This patch simply sets the intrinsic to be col major in KernelConfig and rest of the pipeline automatically picks up if there needs to be a intermediate conflict using shared memory or not.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
@@ -954,25 +954,56 @@ static LogicalResult setAttentionIntrinsicBasedVectorDistributionConfig(
       NamedAttribute("reduction", b.getI64ArrayAttr(reductionTileSizes))};
   IREE::GPU::appendPromotedOperandsList(context, attrs, {0, 1, 2});
 
+  // Check if transposing both intrinsics eliminates the layout conflict
+  // between QK output and PV LHS input.
+  auto matchLayout = [](IREE::GPU::MMASingleSubgroupLayout a,
+                        IREE::GPU::MMASingleSubgroupLayout b) -> bool {
+    return (a.element == b.element) && (a.thread == b.thread) &&
+           (a.tstrides == b.tstrides);
+  };
+  IREE::GPU::MMASingleSubgroupLayout qkOutLayout =
+      IREE::GPU::getSingleSubgroupLayout(qkSchedule.mmaKind,
+                                         IREE::GPU::kMMAOperandAcc);
+  IREE::GPU::MMASingleSubgroupLayout pvRhsLayout =
+      IREE::GPU::getSingleSubgroupLayout(pvSchedule.mmaKind,
+                                         IREE::GPU::kMMAOperandRhs);
+  bool useColMajor = matchLayout(qkOutLayout, pvRhsLayout);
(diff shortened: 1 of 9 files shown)


Groverkss:
Depends on https://github.com/iree-org/iree/pull/23631

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -954,25 +954,56 @@ static LogicalResult setAttentionIntrinsicBasedVectorDistributionConfig(
...
+      -> IREE::Codegen::InnerTileDescAttrInterface {
+    if (auto mma = dyn_cast<IREE::GPU::MMAAttr>(mmaKind)) {
+      return IREE::GPU::MMAAttr::get(context, mma.getIntrinsic(),
+                                     /*colMajor=*/true);
+    }
+    if (auto vmma = dyn_cast<IREE::GPU::VirtualMMAAttr>(mmaKind)) {
+      return IREE::GPU::VirtualMMAAttr::get(context, vmma.getIntrinsic(),
+                                            /*colMajor=*/true);
+    }
+    return mmaKind;
```
sommerlukas, line 985: Is it safe to just return the original attribute if it wasn't changed to column-major?

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -954,25 +954,56 @@ static LogicalResult setAttentionIntrinsicBasedVectorDistributionConfig(
...
+
   SmallVector<NamedAttribute, 2> qkConfig;
   SmallVector<NamedAttribute, 2> pvConfig;
 
   // Configuring for qk matmul.
   IREE::GPU::appendPromotedOperandsList(context, qkConfig, {0, 1});
-  IREE::GPU::setMmaKind(context, qkConfig, qkSchedule.mmaKind);
+  IREE::GPU::setMmaKind(context, qkConfig,
+                        useColMajor ? getColMajorIntrinsic(qkSchedule.mmaKind)
+                                    : qkSchedule.mmaKind);
```
sommerlukas: Why don't we inline the ternary operator into `getColMajorIntrinsic`? We could pass `useColMajor` as an additional parameter to the lambda or capture it. If it's true, we get the column-major version of the attribute, otherwise we just return the original attribute.

sommerlukas (approved):
Just small stuff. 

Do we still have a e2e or pipeline test that doesn't use `col_major = true` to ensure that path still works?

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
Groverkss: added a comment. it is safe.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
Groverkss: done

PR #23634, opened 2026-03-03
[Python][CI] Fix release validation installing wrong package versions

kuhar:
The validate workflow was silently installing stable releases from PyPI
instead of the just-built release candidate wheels. This happened because
`pip install -f file://... iree-base-compiler` ignores pre-release
versions (e.g., `3.11.0rc20260222`) by default, falling back to the latest
stable (`3.10.0`) on PyPI.

Fix by pinning exact versions (`==`) in all install commands, which also
implicitly allows pre-releases. Thread `compiler_package_version`,
`runtime_package_version`, and `legacy_package_version` from the build
workflow to the validate workflow.

Also add `--no-deps` to `pip wheel` in all platform build scripts so
that only IREE wheels are produced in the output directory, not bundled
copies of `numpy`/`sympy`/`ml_dtypes`. The validate workflow now fetches
these dependencies from PyPI at install time.

Discovered by Scott.

### DIFF
--- a/.github/workflows/build_package.yml
+++ b/.github/workflows/build_package.yml
@@ -346,8 +346,6 @@ jobs:
       - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: wheels-${{ matrix.build-family }}-${{ matrix.build-package }}
-          # We upload all wheels (which includes deps so that subsequent
-          # steps can run without further fetching).
           path: ./bindist/*
           retention-days: 5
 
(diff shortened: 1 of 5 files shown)


amd-eochoalo (approved):

ScottTodd (approved):
Looks reasonable. Did you test this somehow?

kuhar:
I built wheels locally (build_package_linux.sh in docker) and ran the setup steps in a fresh venv, but I haven’t run the full workflow because github actions are not exactly testable without running the whole thing

PR #23641, opened 2026-03-03
[GPUHeuristics] Prefer larger MMA intrinsics for very large compute-bound GEMMs

lialan:
* add a new category `VeryLargeGemms` for heuristic.
* in `VeryLargeGemms`, prefer intrinsics with higher compute throughput, and lower VGPR pressure from operand staging.
*  The cutoff is set at 22x the compute-memory ratio (perfTflops / memoryBandwidthTbps). On mi355x , this gives ~7000 compute intensity, corresponding to square matmuls around 10k-11k. This is empirical number determined by experiments done on mi355x.
* This change starts to beat the baseline around ~10.5k square matmul, and can do ~14% better for 16k square matmuls with minimal changes to the code. The number continues to grow with the size.
* Reuse heuristic seeds with `LargeGemms`, as the experiments did not show improvements over a dedicated seed.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
@@ -299,3 +299,20 @@ func.func @scaled_matmul_accumulate(
 // CHECK-REMARKS: [Analysis] SharedMemoryUsage
 // CHECK-REMARKS-SAME: Category:deduceMMASchedule
 // CHECK-REMARKS-SAME: Remark=157184
+
+// -----
+
+// Large f16 matmul — compute-bound, so picks 32x32x16 (higher compute per
+// instruction, lower VGPR pressure than 16x16x32).
+func.func @matmul_f16_compute_bound(
+    %arg0: tensor<4096x4096xf16>,
+    %arg1: tensor<4096x4096xf16>,
+    %arg2: tensor<4096x4096xf32>) -> tensor<4096x4096xf32> {
+  // CHECK-LABEL: func.func @matmul_f16_compute_bound
+  // CHECK-SAME:   #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+  // CHECK:   lowering_config = #iree_gpu.lowering_config
+  // CHECK-SAME: mma_kind = #iree_gpu.mma_layout<MFMA_F32_32x32x16_F16>
+  %0 = linalg.matmul ins(%arg0, %arg1 : tensor<4096x4096xf16>, tensor<4096x4096xf16>)
+                      outs(%arg2 : tensor<4096x4096xf32>) -> tensor<4096x4096xf32>
+  return %0 : tensor<4096x4096xf32>
+}
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
@@ -308,11 +308,11 @@ func.func @matmul_f16_compute_bound(
     %arg0: tensor<16384x16384xf16>,
     %arg1: tensor<16384x16384xf16>,
     %arg2: tensor<16384x16384xf32>) -> tensor<16384x16384xf32> {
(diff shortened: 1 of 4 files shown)


lialan:
I think with DMA + pipelining, we will be able to push the limit down to much lower compute-intensity scenarios.

yzhang93:
FYI, I am working on a similar heuristics that choosing 32x32x16 instead of 16x16x32 intrinsics for the large M/N size. However, what blocks me is the regressions on some convolution shapes in the tracking sheet. Maybe the condition I set was too aggressive, but I would suggest a similar full run of benchmarks on all GEMM and conv shapes before this PR is merged.

lialan:
> FYI, I am working on a similar heuristics that choosing 32x32x16 instead of 16x16x32 intrinsics for the large M/N size. However, what blocks me is the regressions on some convolution shapes in the tracking sheet. Maybe the condition I set was too aggressive, but I would suggest a similar full run of benchmarks on all GEMM and conv shapes before this PR is merged.

@yzhang93 True, I will get some numbers to make sure it do not regress other fields.

My experiment shows that the cutoff should be very big. Take MNK=8k,8k,8k for example, I could not find a configuration that uses 32x32x16 that beats the baseline which uses 16x16x32.

The compute intensity of 8k square matmul is ~5.5k, and we need 7k to break even.

lialan:
Okay I confirm no impact on the benchmark numbers ... because no entries are categorized as `VeryLargeGemms`.

yzhang93:
> Okay I confirm no impact on the benchmark numbers ... because no entries are categorized as `VeryLargeGemms`.

That's good, but I feel like the cutoff is setting too high. During my experiments, I found that a bunch of shapes in the tracking sheet actually benefit from using larger intrinsics, e.g., m = 150000 shapes ([150000, 16384], [16384, 4096] or [150000, 4096], [4096, 16384]). But with your cutoff, these shapes are still using 16x16x32 intrinsics, which are suboptimal.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
```diff
@@ -299,3 +299,20 @@ func.func @scaled_matmul_accumulate(
...
+// Very large f16 matmul — compute-bound, so picks 32x32x16 (higher compute per
+// instruction, lower VGPR pressure than 16x16x32).
+func.func @matmul_f16_compute_bound(
+    %arg0: tensor<16384x16384xf16>,
+    %arg1: tensor<16384x16384xf16>,
+    %arg2: tensor<16384x16384xf32>) -> tensor<16384x16384xf32> {
+  // CHECK-LABEL: func.func @matmul_f16_compute_bound
+  // CHECK-SAME:   #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+  // CHECK:   lowering_config = #iree_gpu.lowering_config
+  // CHECK-SAME: mma_kind = #iree_gpu.mma_layout<MFMA_F32_32x32x16_F16>
```
yzhang93: To be consistent with other tests, can you move the checks after the IR?

yzhang93 (approved):
I'm approving this PR because it sets a very strict condition on `VeryLargeGemm` and have improvements on large square shaped GEMMs. My above concern can be addressed separately as next steps. I will continue looking into how to optimize the imbalance shaped GEMMs.

lialan:
> e.g., m = 150000 shapes ([150000, 16384], [16384, 4096] or [150000, 4096], [4096, 16384]). But with your cutoff, these shapes are still using 16x16x32 intrinsics, which are suboptimal.

@yzhang93 I see. We should tune the bounds a bit, its compute intensity is very close to the cutoff (6413 vs 7000).

Let's plan on making some progress this front, particularly, I want to see if we can find lower criteria to make large intrinsics happen.

yzhang93:
> > e.g., m = 150000 shapes ([150000, 16384], [16384, 4096] or [150000, 4096], [4096, 16384]). But with your cutoff, these shapes are still using 16x16x32 intrinsics, which are suboptimal.
> 
> @yzhang93 I see. We should tune the bounds a bit, its compute intensity is very close to the cutoff (6413 vs 7000).
> 
> Let's plan on making some progress this front, particularly, I want to see if we can find lower criteria to make large intrinsics happen.

SG. There are a bunch of m=150000 shapes in the tracking sheet that can be improved. Could you take those shapes and see if by simply tuning the bounds can improve them all?

nirvedhmeshram (approved):
LGTM, also +1 to using this more effectively as @yzhang93 pointed out.

PR #23652, opened 2026-03-04
[GPUHeuristics] Further tune `LargeGemm` perf

lialan:
* Adds additional, optional utilization rate as a guard in the heuristic seeds.
* Tunes `LargeGemm` MMA schedule heuristic seeds from sg=4/MNT=16 to sg=8/MNT=32 and upgrades the workgroup count adjustment guard to be utilization-aware.
* Inflate LargeGemm heuristic seeds for gfx950 (sg=8, MNT=32) to better utilize CDNA4 compute resources. This is based on tuner results.
* Benchmarked on 75 LargeGemm bf16 shapes from mi355x: +7.0% geomean improvement with 28 shapes improved (up to 30%) and no regressions >5%.
    
Here are some `LargeGemm`s representative results:
| M×N×K | Change |
  |---|---|
  | 150000×4096×16384 | **+11%** |
  | 150000×16384×4096 | **+10%** |
  | 150000×2048×8192 | **+12%** |
  | 150000×4096×2268 | **+10%** |
  | 21760×3840×3840 | **+30%** |
  | 16640×3840×3840 | **+26%** |
  | 11520×3840×3840 | **+17%** |
  | 6400×3840×3840 | **+13%** |
  | 24576×2048×1536 | **+10%** |
  | 4096×8192×2048 | **+6%** |
  | 3840×3840×4352 | neutral |
  | 4096×16384×150000 (large-K) | neutral |
  | 2268×4096×150000 | neutral |


yzhang93:
I like the idea of using utilization-aware guard when increasing the workgroup counts. The main concern about this PR is the generalizability. It would be good to have a thorough benchmarking of all convolution and gemm shapes on CDNA4. It would be also nice to check the performance impact on other GPUs, e.g. CDNA3, RDNA series.

lialan:
> I like the idea of using utilization-aware guard when increasing the workgroup counts. The main concern about this PR is the generalizability. It would be good to have a thorough benchmarking of all convolution and gemm shapes on CDNA4. It would be also nice to check the performance impact on other GPUs, e.g. CDNA3, RDNA series.

Simply applying this idea to all archs will inevitably have regressions on some. So this patch:
* only tweaks on CDNA4/gfx950 at the moment, and nothing else. The new adjuster function can be implemented for other architectures. Before that, we do not change any heuristic results on CDNA3 and RDNA series at all, so their performance are the same before and after.
* Only tweaks GFX950's `LargeGemm`s, so small gemms will not change at all. I have included the large gemm results in the message. There is a regression case among general improvements.

However, I haven't tried its impact on convolutions maybe we should team up and try that out on conv because I was focusing on gemms. @yzhang93

yzhang93:
@yzhang93 This patch should only change large gemms, so I think the convs you saw are noises.

As for the regression in GEMM, I think it is hit or miss. We saw improvements and a few regressions.

lialan:
> @yzhang93 This patch should only change large gemms, so I think the convs you saw are noises.
> 
> As for the regression in GEMM, I think it is hit or miss. We saw improvements and a few regressions.

@yzhang93 Oops ... I am sorry it seems like I accidentally modified your comment instead of replying to it ....

yzhang93:
> @yzhang93 This patch should only change large gemms, so I think the convs you saw are noises.
> 
> As for the regression in GEMM, I think it is hit or miss. We saw improvements and a few regressions.

What I shared are 1x1 convs and they should fall into large GEMMs category. Could you double check?

yzhang93:
> > @yzhang93 This patch should only change large gemms, so I think the convs you saw are noises.
> > As for the regression in GEMM, I think it is hit or miss. We saw improvements and a few regressions.
> 
> @yzhang93 Oops ... I am sorry it seems like I accidentally modified your comment instead of replying to it ....

There should be my original comments in your email. Could you check it?

lialan:
> > @yzhang93 Oops ... I am sorry it seems like I accidentally modified your comment instead of replying to it ....
> 
> There should be my original comments in your email. Could you check it?

Ahh I turned off email notifications for github as there are too many...

Anyway, I have updated the code to make sure it will not regress your conv case.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -680,65 +680,150 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+  std::optional<int64_t> wgpCount = getWgpCount(target);
+  if (!wgpCount.has_value()) {
+    return;
+  }
+
+  int64_t numWorkgroups =
+      computeWorkgroupCount(seeds, problem, intrinsic, splitReductionTripCnt);
+
+  // Utilization-aware MNT reduction for very low utilization scenarios.
+  constexpr double kMinUtilizationThreshold = 0.50;
```
lialan: this is empirical number.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -680,65 +680,149 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
     LDBG() << "Decreasing bestMNTileCountPerSubgroup to "
-           << bestMNTileCountPerSubgroup;
-    numWorkgroups = computeWorkgroupCount();
+           << seeds.bestMNTileCountPerSubgroup;
+    numWorkgroups =
+        computeWorkgroupCount(seeds, problem, intrinsic, splitReductionTripCnt);
+  }
+}
+
+void adjustSeedsForCDNA4(GPUMMAHeuristicSeeds &seeds,
```
yzhang93: I don't see why this should only works for CDNA4. Maybe some of the constant derived from the experiments on Mi355? I think this function should be at least tested on CDNA3 as well.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -680,65 +680,149 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+                             problem.gemmSize == GemmSize::VeryLargeGemm);
+
+  // For large gemms on CDNA4, boost MNT when the K dimension is balanced
+  // and the output is large enough for 2+ waves after the boost.
+  if (isLargeOrVeryLarge) {
+    int64_t mSize = ShapedType::getNumElements(problem.mSizes);
+    int64_t nSize = ShapedType::getNumElements(problem.nSizes);
+    int64_t kSize = ShapedType::getNumElements(problem.kSizes);
+    std::optional<int64_t> wgpCount = getWgpCount(target);
+    int64_t targetMNT = 32;
```
yzhang93: How do you decide this value? Is it a constant?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -680,65 +680,149 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+    // Only boost when K is not excessively large relative to M and N,
+    // and M*N is large enough for 2+ waves with the boosted tile size.
+    int64_t boostedWGSize = targetMNT * intrinsic.mSizes[0] *
+                            intrinsic.nSizes[0] *
+                            seeds.bestSubgroupCountPerWorkgroup;
+    bool kBalanced = kSize <= std::max(mSize, nSize);
+    bool enoughOutput = !wgpCount.has_value() ||
```
yzhang93: if !wgpCount.has_value(), shouldn't it skip the seed adjustment?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -680,65 +680,149 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+  adjustSeeds(seeds, problem, intrinsic, target, splitReductionTripCnt);
+
+  if (!isLargeOrVeryLarge) {
+    return;
+  }
+
+  std::optional<int64_t> wgpCount = getWgpCount(target);
+  if (!wgpCount.has_value()) {
+    return;
+  }
```
yzhang93: Shouldn't these checks move to the front of this function?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -259,43 +260,54 @@ static GemmCutoff computeGemmCutoffsForAI(IREE::GPU::TargetAttr target,
...
+  }
   switch (gemmSize) {
   case GemmSize::SmallGemm:
     return GPUMMAHeuristicSeeds(
         {/*bestSubgroupCountPerWorkgroup=*/2,
          /*bestMNTileCountPerSubgroup=*/2,
          /*bestKTileCountPerSubgroup=*/4,
-         /*bestKElementCountPerSubgroup=*/2 * kCacheLineSizeBits / inBitWidth});
+         /*bestKElementCountPerSubgroup=*/2 * kCacheLineSizeBits / inBitWidth,
+         adjustFn});
```
yzhang93: Please rebase once this PR is merged https://github.com/iree-org/iree/pull/23717.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -680,65 +680,149 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+  if (splitReductionTripCnt > 1) {
+    numWorkgroups *= splitReductionTripCnt;
+  }
+  return numWorkgroups;
+}
+
+/// Helper to get wgpCount from TargetAttr. Returns std::nullopt if
+/// unavailable.
+static std::optional<int64_t> getWgpCount(IREE::GPU::TargetAttr target) {
+  if (target) {
```
kuhar: flip this and exit early instead: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -755,9 +839,8 @@ FailureOr<GPUMMASchedule> deduceMMASchedule(
     // more than once in a row, and we want to keep the original seeds intact
     // for the next call.
     GPUMMAHeuristicSeeds localSeeds = seeds;
-    localSeeds.bestMNTileCountPerSubgroup = adjustSeedsForWgpCount(
-        problem, intrinsic, wgpCount, seeds.bestSubgroupCountPerWorkgroup,
-        seeds.bestMNTileCountPerSubgroup, splitReductionTripCnt);
+    seeds.seedAdjustFn(localSeeds, problem, intrinsic, target,
+                       splitReductionTripCnt);
```
Max191: Why not just adjust the seeds before calling deduceMMASchedule? Passing the lambda to adjust a parameter that is set by the caller seems unnecessary.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -72,6 +103,9 @@ struct GPUMMAHeuristicSeeds {
   // equivalent to `bestKTileCountPerSubgroup * bestIntrinsic.kSize`, for
   // some chosen intrinsic `bestIntrinsic`.
   int64_t bestKElementCountPerSubgroup = 0;
+  // Target-specific seed adjustment callback. Defaults to the baseline
+  // MNT-only reduction (adjustSeeds).
+  SeedAdjustFn seedAdjustFn = adjustSeeds;
```
kuhar: At this point, why not make it a virtual function?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -259,43 +260,54 @@ static GemmCutoff computeGemmCutoffsForAI(IREE::GPU::TargetAttr target,
...
 
 static std::optional<GPUMMAHeuristicSeeds>
-getGemmHeuristicSeeds(GemmSize gemmSize, int64_t inBitWidth, bool scaled) {
+getGemmHeuristicSeeds(GemmSize gemmSize, int64_t inBitWidth, bool scaled,
+                      IREE::GPU::TargetAttr target = nullptr) {
+  SeedAdjustFn adjustFn = adjustSeeds;
+  if (target) {
+    auto maybeChipset = amdgpu::Chipset::parse(target.getArch());
+    if (succeeded(maybeChipset) && *maybeChipset == amdgpu::Chipset(9, 5, 0)) {
+      adjustFn = adjustSeedsForCDNA4;
```
Max191: This is very strange to me. We define the adjustSeeds functions in GPUHeuristics.h, but passing them as a lambda in ConfigUtils.cpp to deduceMMASchedule, which is also defined in GPUHeuristics.h. Everything could just be defined in GPUHeuristics.cpp without any need to pass a lambda.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Max191: +1

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Max191: +1

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Max191: +1

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Max191: For constants like this, can you link to a discussion or describe how this was derived? Once this PR merges, the context about how you came up with this value will be lost.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -148,13 +182,14 @@ struct GPUMMASchedule {
...
 /// input |problem|. Returns std::nullopt if we cannot find such a schedule.
+/// Seed adjustment is performed using seeds.seedAdjustFn.
 /// When |doCPromotion| is true, the accumulator uses shared memory. This can be
 /// due to padding requirements or because the operation has an existing
 /// accumulator that needs to be loaded from global memory (matmul_accumulate).
 FailureOr<GPUMMASchedule> deduceMMASchedule(
     const GPUMatmulShapeType &problem, ArrayRef<GPUIntrinsicType> intrinsics,
     const GPUMMAHeuristicSeeds &seeds, int64_t sharedMemLimitInBytes,
-    int64_t subgroupSize, std::optional<int64_t> cuCount, Location loc,
+    int64_t subgroupSize, IREE::GPU::TargetAttr target, Location loc,
```
Max191, line 189: IMO using `std::optional<IREE::GPU::TargetAttr>` is better, since it makes it clear that the field is optional.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
lialan: hmm, do you mean we would want to make `GPUMMAHeuristicSeeds` inheritable?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -60,6 +64,33 @@ struct GPUIntrinsicType : public GPUMatmulShapeType {
...
+/// estimated workgroup count fills all CUs. Matches the baseline behavior
+/// for all architectures.
+void adjustSeeds(GPUMMAHeuristicSeeds &seeds, const GPUMatmulShapeType &problem,
+                 const GPUIntrinsicType &intrinsic,
+                 IREE::GPU::TargetAttr target, int64_t splitReductionTripCnt);
+
+/// CDNA4-specific seed adjustment. Inherits the default adjustSeeds behavior,
+/// then applies utilization-aware joint sg+MNT reduction for
+/// LargeGemm/VeryLargeGemm.
+void adjustSeedsForCDNA4(GPUMMAHeuristicSeeds &seeds,
```
Max191: I'd prefer to have a single function for adjusting the seeds, which takes the target as a parameter, and derives some key constants/parameters from the target. Having separate entire functions per architecture can lead to a big divergence in how we decide seeds for different architectures, which would be really hard to undo. Unless there is a very compelling reason to do it this way, I'd prefer there to be a single path that is parameterized based on architecture-specific constants.

Btw, I'm not asking to derive everything from hardware details. I'm just saying that if there are empirical constants we are using, then we should pass those as parameters, not the entire function.

Max191 (changes requested):
I would avoid having functions like `adjustSeedsForCDNA4`. I would be surprised if we really need completely different logic for different architectures. It's better to have shared heuristic logic, with specific parameters that are chosen based on the target architecture.

lialan:
> I would avoid having functions like `adjustSeedsForCDNA4`. I would be surprised if we really need completely different logic for different architectures. It's better to have shared heuristic logic, with specific parameters that are chosen based on the target architecture.

@Max191 This is not a patch to have multiple heuristic logics.

This is just a way to reduce seed explosion by extending existing `adjustSeedsForWgpCount` to consider more target arch and problem parameters.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
lialan: You either do this or do that: https://github.com/iree-org/iree/pull/23717

inevitably we will have many more parameters. So either we cascade them into a big spreadsheet and then pick one, or choose from some predefined ones and then adjust according to other parameters.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
lialan: I assume in the future we will get a `adjustSeedsForCDNA3`. 

But if we don't, we will just use the default `adjustSeeds`

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
kuhar: yes

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
yzhang93: I think Max's suggestion is also what I was asking. The adjustSeed function shouldn't be architecture-specific, e.g., adjusting seeds based on utilization should be applicable to all kinds of hardware. However, the parameters can be specific to different architectures.

lialan: @yzhang93 @Max191 I have made some updates to this patch. Specifically:
* initial CDNA4 seeds are now categorized into gemm, scaledGemm and conv;
* `adjustSeedsForTarget` function now is generic to all architectures, but there is a target-specific callback which can further tune for specific target. I think this is the best we can do there to keep generality in current framework, as we would need the matmul shape problem info to further tune the value.

Let me know what you think, or if you have better ideas let me know.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
lialan: This has been updated.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
lialan: TargetAttr itself is already a nullable

Max191: My comment is more just about code style. It is just a nit, but I think it is slightly more clear to make it std::optional.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -689,64 +690,137 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
-           << bestMNTileCountPerSubgroup;
-    numWorkgroups = computeWorkgroupCount();
+           << seeds.bestMNTileCountPerSubgroup;
+    numWorkgroups = computeEstimatedWorkgroupCount(seeds, problem, intrinsic,
+                                                   splitReductionTripCnt);
+  }
+
+  if (seeds.targetSeedAdjustmentCallback) {
+    seeds.targetSeedAdjustmentCallback(seeds, problem, intrinsic, target,
+                                       wgpCount, splitReductionTripCnt);
```
Max191: This is still providing an arbitrary callback for the adjustment. What I'm looking for is for the seed adjustment logic to be shared by all targets (i.e., no `targetSeedAdjustmentCallback` field, just a generic function in GPUHeuristics.cpp that does the adjustment for all architectures), but the parameters to the seed adjustment logic can be given based on the target architecture.

The future that we ultimately would love to arrive at is one where we can compute these seeds from specific hardware parameters (like shared mem size, VGPR count, num WGPs, etc.), independent of the specific hardware architecture. Each architecture will define these parameters, which are inherent to the architecture or SKU itself, and then the seeds are automatically derived. There is no thinking about it, just pasting in the correct hardware specs.

Today, we can't really do that, so we compromise by computing some "meta-parameters" statistically/empirically. These meta-parameters can be set per architecture, but we hope that the meta-parameters could one day be derived by hardware features.

If you give a generic callback to the architecture, then this goes way beyond any meta-parameters, and each architecture could come up with their own arbitrary ways of determining the seeds. This is in the opposite direction of what we want, because we no longer have a nicely contained set of meta-parameters, and instead we have entire callback functions for each architecture that we need to somehow eventually derive from a common set of hardware parameters.

It would be much better if you could implement this via some common `kMinUtilizationThreshold` parameter, which gets set by each architecture, and then we share the logic for adjusting the seed based on `kMinUtilizationThreshold`.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -689,64 +690,128 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+
+  int64_t mSize = ShapedType::getNumElements(problem.mSizes);
+  int64_t nSize = ShapedType::getNumElements(problem.nSizes);
+  int64_t kSize = ShapedType::getNumElements(problem.kSizes);
+
+  // For large GEMMs with balanced K dimensions, boost MNT to improve
+  // per-workgroup compute density. The value 32 was empirically determined
+  // to balance register pressure and occupancy for typical large GEMM
+  // shapes (e.g. 4096x4096x4096).
+  constexpr int64_t kTargetMNT = 32;
```
Max191: `kTargetMNT` seems like something that should also be parameterized based on the architecture.

Also, do you have more information about how you derived the value 32? It would be helpful to dump your process for how you found this value in an iree discussion, and link it in the code.

Max191: Please also explain what this parameter means in the docs. I don't fully understand how to interpret without studying the code.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1284,6 +1284,28 @@ static constexpr ArchSeedSet kDefaultSeeds = {
     },
 };
 
+/// CDNA4 seeds use the default values plus utilization-aware MNT tuning.
+static constexpr ArchSeedSet kCDNA4Seeds = {
```
Max191, line 1292: Should we be setting minUtilizationThreshold in here? I don'e see where this gets set for CDNA4.

Max191 (commented):
Nice, this looks much better! Still a few comments, but this is the direction that I was thinking.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
lialan: it is already set in the kCDNA4Seeds.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
lialan: Done. `kTargetMNT` has been replaced with boostMNTileCountPerSubgroup, a new `std::optional<int64_t>` field in GPUMMAHeuristicSeeds. It's set per-architecture in the seed tables in KnownTargets.cpp (currently only for CDNA4 LargeGemm/VeryLargeGemm entries, value 32). The generic adjustSeedsForTarget logic uses it when set.

Added documentation in the struct explaining the parameter: it's the MN tile count boost target for large GEMMs with balanced K. When set, bestMNTileCountPerSubgroup is boosted to at least this value, increasing per-workgroup compute density (more output elements per workgroup), provided the output tensor is large enough to keep the GPU busy at the boosted tile size.

RE: derivation of value 32 — will create a discussion with benchmarking methodology and link it. Added a TODO in the meantime.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1318,8 +1340,13 @@ const ArchSeedSet &getArchSeedSet(TargetAttr target) {
   // RDNA4 is gfx1200/gfx1201 (major=12, minor=0). Note: gfx1250 (minor=5)
   // is a separate experimental target and should not use RDNA4 seeds.
```
Yu-Zhewen: Nit: the comment here needs to be moved down, as it belongs to `bool isRDNA4 = ...`

lialan: done.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -689,64 +689,127 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+
+  bool isLargeOrVeryLarge = (problem.gemmSize == GemmSizeKind::LargeGemm ||
+                             problem.gemmSize == GemmSizeKind::VeryLargeGemm);
+
+  // For large GEMMs with balanced K dimensions, boost MNT to the
+  // architecture-specific target to improve per-workgroup compute density
+  // (more output elements per workgroup). "Balanced K" means K does not
+  // dominate M or N, so the workload benefits from wider M*N tiles rather
+  // than deeper K unrolling.
+  if (isLargeOrVeryLarge && seeds.boostMNTileCountPerSubgroup) {
```
Max191: Do we need to restrict this to `isLargeOrVeryLarge`? The parameter is optional, so each architecture can control which types of GEMM use this field.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -83,6 +87,21 @@ struct GPUMMAHeuristicSeeds {
...
+  // ceil(numWorkgroups / numCUs). A threshold of 0.50 means at least 50% of
+  // CU-slots in the last wave must be occupied.
+  std::optional<double> minUtilizationThreshold = std::nullopt;
+  // Optional MN tile count boost target for large GEMMs with balanced K.
+  // When set, adjustSeedsForTarget will boost bestMNTileCountPerSubgroup to at
+  // least this value for large/very-large GEMMs whose K dimension does not
+  // dominate M or N, provided the output tensor is large enough to keep the GPU
+  // busy at the boosted tile size. A higher value increases per-workgroup
+  // compute density (more output elements per workgroup), which can improve
+  // performance when the GPU has enough work to stay saturated.
```
Max191: nit: Can you define specifically what K balanced means (i.e., `K < max(M, N)`)?

Max191 (approved):
Looks good! Just a couple nits.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -689,64 +689,127 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+      mnTileSizePerSubgroup * seeds.bestSubgroupCountPerWorkgroup;
+  assert(workgroupSize > 0 && "workgroup size must be positive");
+  int64_t numWorkgroups = mSize * nSize / workgroupSize;
+  if (splitReductionTripCnt > 1) {
+    numWorkgroups *= splitReductionTripCnt;
+  }
+  return numWorkgroups;
+}
+
+/// Adjust MNT seeds based on target hardware and problem characteristics.
```
yzhang93: Nit: Expand `MNT` seeds since first time reader may not get the meaning.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -689,64 +689,127 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+static void adjustSeedsForTarget(GPUMMAHeuristicSeeds &seeds,
+                                 const GPUMatmulShapeType &problem,
+                                 const GPUIntrinsicType &intrinsic,
+                                 IREE::GPU::TargetAttr target,
+                                 int64_t splitReductionTripCnt) {
+  IREE::GPU::TargetChipAttr chip = target ? target.getChip() : nullptr;
+  int64_t wgpCount = chip ? chip.getWgpCount() : 0;
+  if (wgpCount == 0) {
+    return;
   }
```
yzhang93, line 729: Nit: Add debug info for the early return.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -689,64 +689,127 @@ sortMMAIntrinsics(GPUMatmulShapeType problem,
...
+  // than deeper K unrolling.
+  if (isLargeOrVeryLarge && seeds.boostMNTileCountPerSubgroup) {
+    int64_t boostMNT = *seeds.boostMNTileCountPerSubgroup;
+    int64_t mSize = ShapedType::getNumElements(problem.mSizes);
+    int64_t nSize = ShapedType::getNumElements(problem.nSizes);
+    int64_t kSize = ShapedType::getNumElements(problem.kSizes);
+    int64_t boostedWGSize = boostMNT * intrinsic.mSizes[0] *
+                            intrinsic.nSizes[0] *
+                            seeds.bestSubgroupCountPerWorkgroup;
+    bool kBalanced = kSize <= std::max(mSize, nSize);
```
yzhang93: I think it's more clear to name the variable as `kDominated` and flip the check.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1284,6 +1284,50 @@ static constexpr ArchSeedSet kDefaultSeeds = {
...
 
+/// CDNA4 seeds use the default values plus utilization-aware MNT tuning.
+/// The boostMNTileCountPerSubgroup of 32 was empirically determined by
+/// benchmarking large GEMM shapes (e.g. 4096x4096x4096) on MI350 to balance
+/// per-workgroup compute density against register pressure and occupancy.
+/// TODO: Link to iree-org/iree discussion with full benchmarking methodology.
+static constexpr ArchSeedSet kCDNA4Seeds = {
+    /*gemm=*/{
+        /*SmallGemm=*/     {2, 2,  4, 2 * kCacheLineSizeBits,
+                            /*minUtilizationThreshold=*/0.50},
```
yzhang93: Does the benchmark results show using `/*minUtilizationThreshold=*/0.50` is good for small and medium gemms?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1284,6 +1284,50 @@ static constexpr ArchSeedSet kDefaultSeeds = {
...
+        /*LargeGemm=*/     {8, 32, 2, kCacheLineSizeBits / 2,
+                            /*minUtilizationThreshold=*/0.50,
+                            /*boostMNTileCountPerSubgroup=*/32},
+        /*VeryLargeGemm=*/ {8, 32, 2, kCacheLineSizeBits / 2,
+                            /*minUtilizationThreshold=*/0.50,
+                            /*boostMNTileCountPerSubgroup=*/32},
+    },
+    /*conv=*/{
+        /*SmallGemm=*/     {2, 2,  4, kCacheLineSizeBits,
+                            /*minUtilizationThreshold=*/0.50},
```
yzhang93: Same questions for conv, could you check the benchmark results?

yzhang93 (changes requested):
This refactor looks much better and aligns with my mental model. Could you double check the benchmark results, especially for small and medium GEMMs and Convs? 

Also it's good to add some tests for expected lowering config after change.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
lialan: No, they don't show performance improvements (in fact regressed a bit) for small and medium gemms.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
lialan: conv seeds are left untouched, so i think the numbers will not change.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
lialan: updated

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
lialan: done.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
lialan: Done. Opted to use "M*N tile-count (bestMNTileCountPerSubgroup)" in the doc.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
yzhang93: Then why do you add `/*minUtilizationThreshold=*/0.50` for these shapes? Why not leave it not assigned?

lialan: Good point, my bad. I removed them. so only large/very large gemms, and no changes to scaled gemms as well.

yzhang93:
LGTM. My last request is to add some large GEMM tests for expected lowering config after the change.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
lialan: @yzhang93 any other issues I need to address?

yzhang93: @lialan You missed my comments about adding some tests.

lialan: @yzhang93 done, test added: https://github.com/iree-org/iree/pull/23652/changes#diff-71790d2daff09394b4f33f03a88dd0ffad2d8d65267f4e4b82b985c9c7236e3d

yzhang93: Thanks @lialan. I think the tests can be added to the existing test `config_tile_and_fuse_gfx950.mlir`, and you can  use check-prefixes similar to https://github.com/iree-org/iree/blob/d9fc453ea46a6340ca74709eb95e74e863628aa5/compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_igemm_tile_and_fuse.mlir#L4.

lialan: @yzhang93 good suggestion. combined both.

yzhang93 (approved):

PR #23653, opened 2026-03-05
[Codegen] Introduce VectorizableOpInterface and migrate LinalgExt::GatherOp

hanhanW:
Introduce a VectorizableOpInterface that provides an unified extension point for vectorization. Operations implement this interface to declare, whether it is vectorizable, how they should be vectorized, and the driver dispatches through it.

- Implement external model for IREE::LinalgExt::GatherOp, inlining the vectorization logic from vectorizeLinalgExtGatherToTransferGather.
- Update GenericVectorization.cpp to dispatch through the interface for ops that implement it, with TypeSwitch fallback for unmigrated ops.

No new tests because it is an NFC in terms of functionality. It just follows different mechanism for vectorization.

Assisted-by: Claude


compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -0,0 +1,168 @@
...
+    return true;
+  }
+
+  FailureOr<SmallVector<Value>> vectorize(Operation *op, RewriterBase &rewriter,
+                                          ArrayRef<int64_t> vectorSizes,
+                                          ArrayRef<bool> scalableDims,
+                                          DictionaryAttr options) const {
+    auto gatherOp = cast<IREE::LinalgExt::GatherOp>(op);
+    int64_t batchRank = gatherOp.getBatchRank();
+    Location loc = gatherOp.getLoc();
```
sommerlukas, line 61: Might be worth creating an `ImplicitLocOpBuilder` here and use it in the rest of this function.

sommerlukas (approved):
Just a small nit, but please also wait for @Groverkss to review.

Groverkss (approved):

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -0,0 +1,168 @@
...
+    VectorType indicesMaskType = indicesVecTy.clone(rewriter.getI1Type());
+    SmallVector<OpFoldResult> gatherDims =
+        tensor::getMixedSizes(rewriter, loc, gatherOp.getOutput());
+    Value indicesMask = vector::CreateMaskOp::create(
+        rewriter, loc, indicesMaskType,
+        ArrayRef(gatherDims).take_front(gatherOp.getBatchRank()));
+    auto indicesVecRead = vector::TransferReadOp::create(
+        rewriter, loc, indicesVecTy.clone(indicesTy.getElementType()),
+        gatherOp.getIndices(), SmallVector<Value>(indicesTy.getRank(), zero),
+        std::nullopt);
```
Groverkss, line 90: Note that eventually I would like to move this to an interface. Using transfer_read/transfer_write to vectorize something is not a very nice way of vectorization. We should be doing dialect conversion imo. We can do that after the VectorTile size analysis lands.

hanhanW: Is there a plan for VectorTile size analysis yet? Is it something similar to what we implemented: 

https://github.com/iree-org/iree/blob/4e3da929fdc741c1011bcbb434276c7bc6fb44f9/compiler/src/iree/compiler/Codegen/Utils/Utils.h#L363-L372

sommerlukas: There is a PR with a first attempt at the analysis here: https://github.com/iree-org/iree/pull/23668

The idea is that we take the tile sizes from `to_layout` operations (that come from `lowering_config`) and propagate that to the other compute ops. It's then annotated as a discardable attribute, so we can use it in `GenericVectorization`. There, we use three sources of information in this order:
1. `lowering_config`
2. Result of the VectorTileSizeAnalysis
3. `inferSizesFromIR`

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: I tried `ImplicitLocOpBuilder` when it was created, but I did not like the idea much. It was useful if you create operations in a single line, though I still don't like the idea.

The other point is that it extends OpBuilder, not RewriterBase, so methods like `modifyOpInPlace` don't work.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: The quick comment is about `getVectorScalableFlags` method. The `lowering_config` can implement the method to provide vector tile sizes. It's a little fragile because lowering config can be messed up during the transformation, so it is default off on x86 backends.

https://github.com/iree-org/iree/blob/4e3da929fdc741c1011bcbb434276c7bc6fb44f9/compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp#L41-L43

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: I'm not sure what you mean by 
> operations in a single line

If we create an `ImplicitLocOpBuilder` once in the beginning of `vectorize`, there is nine places where we pass `rewriter, loc` and could be passing that builder instead.

For things like `modifyOpInPlace`, you can keep using the `rewriter`, I think.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: So, you mean we need to propagate the scalable flags in the analysis?

hanhanW: I meant the design idea was using the interface method, if the lowering config has the information. Scalable flags is just something similar to vector size and only some CPU backends use it.

PR #23661, opened 2026-03-05
[Codegen] Do not vectorize with invalid vector sizes from lowering config

hanhanW:
Vectorization with zero sizes is incorrect, and we should just bail out. It is much better than vectorizing with incorrect tile sizes that may lead to correctness issue.


nirvedhmeshram (approved):
LGTM

PR #23658, opened 2026-03-05
[Codegen] Migrate ToLayoutOp to VectorizableOpInterface

hanhanW:
No new tests because it is an NFC in terms of functionality. It just follows different mechanism for vectorization.

It is a step towards https://lists.lfaidata.foundation/g/iree-technical-discussion/message/15

Assisted-by: Claude

PR #23656, opened 2026-03-05
[Codegen] Migrate ArgCompareOp to VectorizableOpInterface

hanhanW:
- Add external model for IREE::LinalgExt::ArgCompareOp, inlining the vectorization logic from vectorizeLinalgExtArgCompare.
- Remove ArgCompareOp from GenericVectorization TypeSwitch and candidate isa<> check (now dispatched through interface).

No new tests because it is an NFC in terms of functionality. It just follows different mechanism for vectorization.

It is a step towards https://lists.lfaidata.foundation/g/iree-technical-discussion/message/15

Assisted-by: Claude

PR #23662, opened 2026-03-05
[Codegen] Migrate MapStoreOp to VectorizableOpInterface

hanhanW:
The revision also deletes VectorizeIREELinalgExtOps, because it is already covered in GenericVectorization pass. The tests from `vectorize_iree_linalg_ext_ops.mlir` are migrated to `generic_vectorization.mlir`.

The map_store vectorization is only handled on LLVMGPU, which requires DecomposeMapStoreOp pass, so the vectorization is only enabled on LLVMGPU for now -- which matches the current behavior in IREE.

Assisted-by: Claude

ci-extra: test_torch


hanhanW:
It depends on

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -299,6 +299,7 @@ static void addGPUVectorizationPasses(OpPassManager &funcPassManager,
   options.enableCleanup = false;
   options.foldCastIntoContract = true;
   options.enableVectorMasking = enableMasking;
+  options.vectorizeMapStore = true;
```
Max191, line 302: FYI, we probably don't *need* the pass option, but it may still be good to keep to avoid untested code paths. The main reason this wasn't enabled elsewhere is that we don't see map_store anywhere else, so it should be fine to have on in other pipelines, but just not really used/tested. We would also need to add the decomposition pass to the other pipelines for it to work e2e.

To be clear, no action item here. Just providing some context. Maybe you can add a comment in case someone wants to enable it on other pipelines in the future.

Max191 (approved):
Nice! I'm glad we're getting rid of all the ad-hoc vectorization passes

Groverkss (approved):

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
hanhanW: Thanks for the context, I tried that before (including adding `DecomposeMapStoreOp` to CPU pipeline). The performance is the same for one of the tracking models but the binary size regresses. Given that the op is not fully developed in other backend, I end up with having the flag for now; we can do the flip when other backends catch up.

PR #23665, opened 2026-03-05
[Stream] Prevent CSE from aliasing dispatch-allocated buffers

jtuyls:
AsyncDispatchOp was declared as Stream_PureOp (NoMemoryEffect), which let CSE merge identical dispatches that produce non-tied resource results (new allocations). When two consumers each need their own independent buffer (e.g. two ScatterElements writing into separate zero-initialized buffers), merging makes them alias the same buffer and corrupt each other.

This PR fixes it by implementing MemoryEffectsOpInterface on AsyncDispatchOp. The new getEffects() reports MemoryEffects::Allocate when any result is a non-tied resource (i.e. a fresh buffer allocation). This is conservative and prevents CSE from merging such dispatches while still allowing CSE for dispatches where all results are tied.


tests/e2e/linalg_ext_ops/scatter_multi_consumer.mlir
```diff
@@ -0,0 +1,143 @@
+// Tests that two scatter ops consuming the same fill-initialized buffer each
```
MaheshRavishankar, line 1: Can you move this to tests/e2e/regression/  since this is not an op test but rather testing general handling of a program.

MaheshRavishankar (approved):
This makes sense to me. 

@hanhanW  or @benvanik  do you guys have any feedback here?

hanhanW (commented):
This needs @benvanik's eyes because we made the flip 6 months ago:

- https://github.com/iree-org/iree/commit/2d1ae1e84ef0f112c7fdd2e5c8f94c18b06da61e
- https://github.com/iree-org/iree/commit/b30ec6085cc5238c882cd51ac93f443a75599a9b

They were for fixing duplicated globals, so I'll need to test again. We don't have in-tree tests because it was hard to add. Perhaps I can come up with a pipeline test today.

PR #23667, opened 2026-03-05
Set exact flag on index_castui during iree-util-optimize-int-arithmetic

amd-eochoalo:
Since https://github.com/llvm/llvm-project/pull/183395 was merged upstream,
optimization of index_cast(index_cast(x)) -> x requires the exact flag.
This pattern without the exact flag is unsound since the size of index is assumed to be 64 bits
but that is not always the case.

IREE applies this pattern when running --iree-util-optimize-int-arithmetic. 
This pass uses the util dialect to prove the correctness of transforming signed operations into unsigned operations.
We take advantage of this to set the exact flag which allows the pattern to get applied soundly.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
+++ b/compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
@@ -104,6 +104,8 @@ struct ConvertOpToUnsigned : public OpRewritePattern<Signed> {
     }
     auto newOp = rewriter.replaceOpWithNewOp<Unsigned>(
         op, op->getResultTypes(), op->getOperands(), op->getAttrs());
+    // This is safe to do because staticallyLegalToConvert makes sure
+    // that op's operand fits in the newOp's type.
     if constexpr (std::is_same_v<Unsigned, arith::IndexCastUIOp>) {
       newOp.setExact(true);
     }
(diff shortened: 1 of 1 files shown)


amd-eochoalo:
Converting back to draft. I'll add exact instead on one of the passes like `ConvertUnsignedI64IndexCastProducerToIndex`.

compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
```diff
@@ -102,8 +102,11 @@ struct ConvertOpToUnsigned : public OpRewritePattern<Signed> {
     if (failed(staticallyLegalToConvertToUnsignedOp(solver, op))) {
       return failure();
     }
-    rewriter.replaceOpWithNewOp<Unsigned>(op, op->getResultTypes(),
-                                          op->getOperands(), op->getAttrs());
+    auto newOp = rewriter.replaceOpWithNewOp<Unsigned>(
+        op, op->getResultTypes(), op->getOperands(), op->getAttrs());
+    if constexpr (std::is_same_v<Unsigned, arith::IndexCastUIOp>) {
+      newOp.setExact(true);
+    }
```
kuhar, line 111: Can you add a comment explaining why this is safe to set here?

amd-eochoalo: https://github.com/iree-org/iree/pull/23667/commits/138f855ed6af7564fc61907e94247ec640aa2322

kuhar (approved):

amd-eochoalo:
@RattataKing I'll merge after your LLVM integration, so no need to worry about this PR. I'll also point to the new LLVM branch.

krzysz00:
I'm going to flag that these patterns also exist upstream, and that there are a few other spots in IREE where we generate index_cast ops.

amd-eochoalo:
I'm closing this without merging due to it being reverted upstream. See https://github.com/llvm/llvm-project/pull/184876

PR #23668, opened 2026-03-05
[VectorDistribute] Add VectorTileSizeAnalysis

sommerlukas:
Motivation for this change is to better support masking to make up for differences between actual tensor shape and tile sizes chosen to match the HW. For example, lowering config selection would be able to choose a tile size of 64, even if the input has size 127. To ensure correct computation, proper masking needs to be introduced in this case during `GenericVectorization`. To that end, `GenericVectorization` needs reliable vector tile size information.

The goal of the `VectorTileSizeAnalysis` is to provide such information. The analysis is seeded from `to_layout` operations, which in turn have been inserted based on `lowering_config`. The `VectorTileSizeAnalysis` propagates the information through the graph. It uses the upstream MLIR dataflow analysis framework and combines a sparse forward and sparse backward analysis in a single solver with a shared lattice.

Conflicts are resolved by an overdefined/"top" state. Consumers such as `GenericVectorization` can then choose a candidate, e.g. based on heuristics.

To make the vector tile size readily available in `GenericVectorization`, the `MaterializeVectorTileSizePass` materializes the analysis results as discardable attribute on compute ops.

Assisted-by: Claude Code


compiler/src/iree/compiler/Codegen/Common/test/materialize_vector_tile_sizes.mlir
```diff
@@ -0,0 +1,196 @@
...
+    linalg.yield %add : f16
+  } -> tensor<63xf16>
+  %1 = iree_vector_ext.to_layout %0 to layout(#layout) : tensor<63xf16>
+  // CHECK: linalg.generic
+  // CHECK-SAME: iree_codegen.vector_tile_sizes = [array<i64: 64>]
+  %2 = linalg.generic {
+    indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>],
+    iterator_types = ["parallel"]
+  } ins(%1 : tensor<63xf16>) outs(%empty : tensor<63xf16>) {
+  ^bb0(%in: f16, %out: f16):
```
Groverkss, line 30: Can you add a test with dynamic shapes?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,521 @@
...
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/SymbolTable.h"
+
+#include "llvm/ADT/SmallSet.h"
+
+#define DEBUG_TYPE "iree-codegen-vector-tile-size-analysis"
+
+// The purpose of this analysis is to propagate information about the
+// undistributed vector tile size across the operation graph. The vector tile
```
Groverkss: I don't think "undistributed" should be part of this description. We have some ops that would like a particular vector size, and we are catering towards that.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,521 @@
...
+// for that dimension. The bottom (uninitialized) state of the lattice is simply
+// empty. The join/merge operation for two lattice elements is the per-dimension
+// set-union of candidates. For example, in the 2D case:
+// ({2, 4}, {16}) U ({8}, {32}) = ({2, 4, 8}, {16, 32})
+//
+// As the sets can only grow, the join/meet operator is by definition monotonic.
+// As the set union can not result in a conflict, no lattice state for top
+// (overdefined) is required in this lattice.
+//
+// The lattice is initialized from `to_layout` operations.
```
Groverkss: This is an implementation detail. to_layout requests a particular vector size. It doesn't mean that this analysis only works on to_layout.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,521 @@
...
+          MaterializeVectorTileSizesPass> {
+public:
+  void runOnOperation() override {
+    auto funcOp = getOperation();
+
+    DataFlowSolver solver;
+    dataflow::loadBaselineAnalyses(solver);
+    solver.load<TileSizeForwardAnalysis>();
+    SymbolTableCollection symbolTable;
+    solver.load<TileSizeBackwardAnalysis>(symbolTable);
```
Groverkss: Why do we need the symbol table for this?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,521 @@
...
+      if (init.getDefiningOp<tensor::EmptyOp>()) {
+        return true;
+      }
+    }
+  }
+  if (auto fillOp = dyn_cast<linalg::FillOp>(defOp)) {
+    if (fillOp.getOutputs()[0].getDefiningOp<tensor::EmptyOp>()) {
+      return true;
+    }
+  }
```
Groverkss: tbf, we can duplicate any operation that is not a "load". Maybe we should start with "generator like" operations.

We can duplicate any operation which has no tensor inputs (either captured in region, or through Inputs).

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,521 @@
...
+    // seeds are set even for to_layout ops in regions that DeadCodeAnalysis
+    // hasn't yet marked as live during init.
+    top->walk([&](ToLayoutOp toLayout) {
+      LDBG() << "Anchor: " << toLayout;
+      auto candidates = TileSizeCandidates::fromSizes(
+          toLayout.getLayout().getUndistributedShape());
+      auto *lattice = getLatticeElement(toLayout.getResult());
+      propagateIfChanged(lattice, lattice->join(candidates));
+    });
+    return SparseForwardDataFlowAnalysis::initialize(top);
```
Groverkss: I don't think we care about DCE. If a to_layout op was dead, it would get removed. to_layout operations are Pure.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: The constructor of the `mlir::dataflow::SparseBackwardDataFlowAnalysis` base class requires it. It uses the symbol table to resolve call operands. We don't use that (our `visitCallOperand` is no-op), but the constructor still requires it.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: This wasn't about DCE for `to_layout` specifically. The forward analysis only processes blocks after they have been marked as live by DCE. But we should be fine by discovering the `to_layout` as blocks are marked alive, so I've removed this bit and moved initialization from `to_layout` into the forward analysis `visitOperation`.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: I've rephrased this bit to talk more generically about "anchor operations" and just mention `to_layout` as one example.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Removed.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,506 @@
...
+//
+// For some other operations, no propagation rules are defined on purpose. For
+// example, `extract_slice` and `insert_slice` operations are natural boundaries
+// of tiling/padding, therefore no information is propagated across them.
+//
+// After the dataflow solver reaches a fixpoint, the
+// MaterializeVectorTileSizesPass materializes the result as a discardable
+// attribute. At this point, the result is a set of candidate vector tile sizes
+// per iteration dimension. It is up to the users of the analysis how to select
+// a tile size from the set of candidates.
```
hanhanW: Why does a dimension can have a set of candidate vector tile sizes? When does that happen? Doesn't it look like a bug in IR?

hanhanW (changes requested):
Thanks for the patch! I can see the value of this pass, which is an improved version of `useConfiguredSizes` in generic vectorization. I'm thinking to extend the pass (in a follow-up) which moves the `useConfiguredSizes` logic into the anslysis, which can work well with the `getVectorSize` lowering config interface method.

The main concern on my side is about multiple candidates. Isn't it a bug if it happens?

sommerlukas:
Thanks for the feedback!

> The main concern on my side is about multiple candidates. Isn't it a bug if it happens?

I don't think it's necessarily a bug in the IR, consider the scenario below:
```mlir
%empty = tensor.empty ...
%a = linalg.generic ins(...) outs(%empty) ...
%al = to_layout %a, layout with tile size 64
%b = linalg.generic ins(...) outs(%empty)
%c = linalg.generic ins(%b) ...
%cl = to_layout %c, layout with tile size 32
...
```
Here, we would propagate tile size 64 from %al backward to `%a` and then `%empty`. Then we propagate that tile size from `%empty` forward through to `%b`, and `%b` also has tile size 64. If we then propagate the tile size 32 backward from `%cl` through `%c` to `%b`, we have two tile sizes for `%b` and need to resolve that somehow. If, instead of collecting a set of possible tile sizes for each candidate (and dimension), we would move to an overdefined top element in the lattice, we would lose tile size information for `%b`.

CC @Groverkss who might have other scenarios in mind.

Groverkss (commented):
LGTM modulo hanhan's comments. I think it's okay to just bail out on duplicate sizes for now and we can improve if needed.

Groverkss (approved):

sommerlukas:
I've updated the analysis to only track one size per dimension. In case of conflict, we now go to a top/overdefined state. Propagation from duplicatable operations is now completely disabled, which avoids conflict in the scenario I outlined above.

The tracking of one size per dimension also facilitates adding the scalable flag to the analysis for CPU later on.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+      TileSizes combined;
+      for (auto [operandLattice, operandVal] :
+           llvm::zip(operands, op->getOperands())) {
+        combined.merge(getTileSizesFor(operandVal, operandLattice));
+      }
+      for (TileSizeLattice *result : results) {
+        propagateIfChanged(result, result->join(combined));
+      }
+      return success();
+    }
```
hanhanW: Why do we have this? Are there any ops vectorized before the analysis? linalg ops do not have the trait. IIUC, it is for arith ops and vector ops.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+      for (unsigned i = 0; i < genericOp.getNumDpsInits(); ++i) {
+        OpOperand *init = genericOp.getDpsInitOperand(i);
+        AffineMap map = genericOp.getMatchingIndexingMap(init);
+        auto resultTileSizes = iterTileSizes.mapFromIterationSpace(map);
+        if (!resultTileSizes.empty()) {
+          propagateIfChanged(results[i], results[i]->join(resultTileSizes));
+        }
+      }
+      return success();
+    }
```
hanhanW: Although we are tending to generic op form, there are still some named ops. Given that we only care linalg structured properties, we should use `linalg::LinalgOp` here.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
hanhanW, line 1: I think this file should be renamed as `MaterializeVectorTileSizes.cpp`. It is not an analysis or a class/util. It is the pass implementation.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+    return kOverdefined;
+  }
+  if (a == kUninitialized) {
+    return b;
+  }
+  if (b == kUninitialized) {
+    return a;
+  }
+  return a == b ? a : kOverdefined;
+}
```
hanhanW: They seem to be local to `TileSizes`. Should we move them into the class?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+public:
+  TileSizes() = default;
+  explicit TileSizes(unsigned rank) : dims(rank, kUninitialized) {}
+
+  /// Construct from a single concrete tile size (one value per dimension).
+  static TileSizes fromSizes(ArrayRef<int64_t> sizes) {
+    TileSizes result;
+    result.dims.assign(sizes.begin(), sizes.end());
+    return result;
+  }
```
hanhanW: Do we check if the number of elements in `sizes` matches `result` size? If it is used as initializer, we should declare a constructor instead?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+  // fill-like linalg.generic broadcasting a scalar) is a generator and
+  // duplicatable.
+  if (auto linalgOp = dyn_cast<linalg::LinalgOp>(defOp)) {
+    if (llvm::none_of(linalgOp->getOpOperands(), [&](OpOperand &operand) {
+          return isa<ShapedType>(operand.get().getType()) &&
+                 linalgOp.payloadUsesValueFromOperand(&operand);
+        })) {
+      return true;
+    }
+  }
```
hanhanW, line 234: Are you looking for [verifyFillInterface](https://github.com/llvm/llvm-project/blob/227ced1b1eb442fed2196c4a9e80ee00cb97efe3/mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.h#L216-L217)?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+  auto &tileSizes = lattice->getValue();
+  if (tileSizes.empty()) {
+    return empty;
+  }
+  if (isDuplicatable(val)) {
+    return empty;
+  }
+  return tileSizes;
+}
```
hanhanW, line 261: We want to avoid static variable here, because it is not trivially destructible, see https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

I think returning values is fine, because `TileSizes` is a wrapper of `SmallVector<>` which is small enough. It usually only has few elements.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+  using SparseBackwardDataFlowAnalysis::SparseBackwardDataFlowAnalysis;
+
+  void setToExitState(TileSizeLattice *lattice) override {
+    // Exit state is uninitialized (identity for meet).
+  }
+
+  LogicalResult
+  visitOperation(Operation *op, ArrayRef<TileSizeLattice *> operands,
+                 ArrayRef<const TileSizeLattice *> results) override {
+    // to_layout: propagate result tile sizes backward to input.
```
hanhanW: I'd phrase it like `ToLayout op is always an anchor op. Propagate tile sizes backward to input.`.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+        return {};
+      }
+      unsigned iterDim = dimExpr.getPosition();
+      if (iterDim >= rank() || dims[iterDim] == kUninitialized) {
+        return {};
+      }
+      result.dims[i] = dims[iterDim];
+    }
+    return result;
+  }
```
hanhanW, line 158: Do we assert if the given indexing map is a projected permutation? Looking at the users of `mapToIterationSpace`, do you really need to pass in `numLoops`? It can be `indexingMap.getNumDims()`, if IIUC. We also need to check if the ranks match or not, IIUC.

I'm thinking if we rename them to `scatterToIterationSpace` and `gatherFromIterationSpace`. The original names are okay as well. I just feel scatter and gather sounds better, and it may be just my preference. What do you think?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+      }
+      for (auto [operandLattice, operandVal] :
+           llvm::zip(operands, op->getOperands())) {
+        if (!isa<ShapedType>(operandVal.getType())) {
+          continue;
+        }
+        propagateIfChanged(operandLattice, operandLattice->meet(combined));
+      }
+      return success();
+    }
```
hanhanW: ditto: deadcode?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+      auto &ts = getTileSizesFor(toLayout.getResult(), results[0]);
+      if (!ts.empty()) {
+        TileSizeLattice *inputLattice = operands[0];
+        propagateIfChanged(inputLattice, inputLattice->meet(ts));
+      }
+      return success();
+    }
+
+    // linalg.generic: propagate through indexing maps.
+    if (auto genericOp = dyn_cast<linalg::GenericOp>(op)) {
```
hanhanW: ditto: use `linalg::LinalgOp`.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+  }
+  return iterTileSizes;
+}
+
+/// Given a linalg op and the solver, compute per-dimension tile sizes.
+/// Returns a vector of one tile size per iteration dimension, or nullopt if
+/// any dimension is uninitialized or overdefined.
+static std::optional<SmallVector<int64_t>>
+getPerDimTileSizes(linalg::LinalgOp linalgOp, const DataFlowSolver &solver) {
+  auto tileSizes = getIterationSpaceTileSizes(linalgOp, solver);
```
hanhanW:
```suggestion
  TileSizes tileSizes = getIterationSpaceTileSizes(linalgOp, solver);
```

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+  if (!tileSizes.isDefined()) {
+    return std::nullopt;
+  }
+  unsigned numLoops = linalgOp.getNumLoops();
+
+  SmallVector<int64_t> results;
+  for (unsigned i = 0; i < numLoops; ++i) {
+    results.push_back(tileSizes[i]);
+  }
+  return results;
```
hanhanW: Should we have `dims()` accessor which returns ArrayRef, and we do `return llvm::to_vector(tileSizes.dims().take_front(numLoops))` here?

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+    solver.load<TileSizeForwardAnalysis>();
+    SymbolTableCollection symbolTable;
+    solver.load<TileSizeBackwardAnalysis>(symbolTable);
+
+    if (failed(solver.initializeAndRun(funcOp))) {
+      return signalPassFailure();
+    }
+
+    funcOp->walk([&](linalg::LinalgOp linalgOp) {
+      auto perDimSizes = getPerDimTileSizes(linalgOp, solver);
```
hanhanW:
```suggestion
      std::optional<SmallVector<int64_t>> perDimSizes = getPerDimTileSizes(linalgOp, solver);
```

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+
+  bool operator==(const TileSizes &rhs) const { return dims == rhs.dims; }
+
+  void print(raw_ostream &os) const {
+    os << "[";
+    llvm::interleaveComma(dims, os, [&](int64_t v) {
+      if (v == kUninitialized) {
+        os << "?";
+      } else if (v == kOverdefined) {
+        os << "T";
```
hanhanW: Do we use `<overdefined>` here?

Learning from LLM, the conventional symbol for top (overdefined) in lattice theory is ⊤, but that's a unicode character. It is not a well-known symbol (or I may be an outlier), so it is better to spell it out from IR's perspective. This will flow down to other passes.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+  }
+
+  unsigned rank() const { return dims.size(); }
+  bool empty() const { return dims.empty(); }
+
+  int64_t operator[](unsigned i) const { return dims[i]; }
+
+  /// Returns true if all dimensions have a defined (positive) tile size.
+  bool isDefined() const {
+    return !empty() && llvm::all_of(dims, [](int64_t v) { return v > 0; });
```
hanhanW: nit: avoid magic number. We should check they are neither kUninitailized nor kOverdefined.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
```diff
@@ -0,0 +1,505 @@
...
+
+    if (failed(solver.initializeAndRun(funcOp))) {
+      return signalPassFailure();
+    }
+
+    funcOp->walk([&](linalg::LinalgOp linalgOp) {
+      auto perDimSizes = getPerDimTileSizes(linalgOp, solver);
+      if (!perDimSizes) {
+        return;
+      }
```
hanhanW: I think we should either emit a message or signal a failure when it happens.

hanhanW (commented):
> Thanks for the feedback!
> 
> > The main concern on my side is about multiple candidates. Isn't it a bug if it happens?
> 
> I don't think it's necessarily a bug in the IR, consider the scenario below:
> 
> ```mlir
> %empty = tensor.empty ...
> %a = linalg.generic ins(...) outs(%empty) ...
> %al = to_layout %a, layout with tile size 64
> %b = linalg.generic ins(...) outs(%empty)
> %c = linalg.generic ins(%b) ...
> %cl = to_layout %c, layout with tile size 32
> ...
> ```
> 
> Here, we would propagate tile size 64 from %al backward to `%a` and then `%empty`. Then we propagate that tile size from `%empty` forward through to `%b`, and `%b` also has tile size 64. If we then propagate the tile size 32 backward from `%cl` through `%c` to `%b`, we have two tile sizes for `%b` and need to resolve that somehow. If, instead of collecting a set of possible tile sizes for each candidate (and dimension), we would move to an overdefined top element in the lattice, we would lose tile size information for `%b`.
> 
> CC @Groverkss who might have other scenarios in mind.

I still think that it is a bug if it happens. It means that the config generator has a bug. The example you showed is valid, and you have a good fix for it. We should stop propagation if it is `tensor.empty()` op. It is describing the shape of the result, but it is not a connection between two `linalg.generic` ops.

Anyway, the current revision mostly looks good. I left few comments, please take a look. Thanks!

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Unfortunately, `verifyFillInterface` emits diagnostics, even if we discarded the `failure` itself, so I don't think it's the right tool for this job.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: > Do we assert if the given indexing map is a projected permutation?

I don't think we need to assert that here. If a result is not a dim expression, that dimension is skipped already and `TileSizes` that are not fully defined are never materialized.

> Looking at the users of mapToIterationSpace, do you really need to pass in numLoops? It can be indexingMap.getNumDims(), if IIUC. We also need to check if the ranks match or not, IIUC.

I've removed `numLoops` and added an assert in all places using it to assert the correct rank.

> I'm thinking if we rename them to scatterToIterationSpace and gatherFromIterationSpace. The original names are okay as well. I just feel scatter and gather sounds better, and it may be just my preference. What do you think?

For me personally, gather/scatter is always linked to data movement (gathering/scattering data elements into/from a bigger vector/matrix/...). What we are doing here is using an AffineMap to map information from operand to iteration space and vice versa. Therefore, IMHO, the current name is a good description.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: I've made a constructor.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: I've changed it to `<overdefined>`.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: I've switched to `linalg::LinalgOp`.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: At the current stage of the pipeline, we probably don't have elementwise-mappable operations, removed this.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Removed.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done, I've added an assert and then return `getDims()` directly.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: I think there's legitimate cases where operations don't get a tile size assigned (e.g., the anchor operation is inside a tiled loop), so emitting a warning/error here or even failing the pass isn't appropriate IMHO.

sommerlukas (commented):
Thanks for the feedback @hanhanW! I've addressed the comments and put some more replies inline.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
hanhanW: This file should be deleted?

hanhanW (commented):
I'll take a look at other code, but I think you forgot to delete old file (or make git mv).

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
sommerlukas: Yeah, absolutely, something must have gone wrong. Should now be deleted, sorry.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+//===----------------------------------------------------------------------===//
+
+class TileSizeLattice : public dataflow::Lattice<TileSizes> {
+public:
+  using Lattice::Lattice;
+};
+
+/// Read the TileSizes from a lattice, returning empty tile sizes if the lattice
+/// value is from a duplicatable operation.
+static const TileSizes getTileSizesFor(Value val,
```
hanhanW: Delete the `const`: https://clang.llvm.org/extra/clang-tidy/checks/readability/const-return-type.html

When you return by value, the caller gets a new object. The const on the return type is saying "the caller can't modify this temporary." But temporaries are already about to be either moved or copied into the caller's variable:

```
  // At the call site:
  auto ts = getTileSizesFor(val, lattice);  // const is irrelevant here
```


The caller's `ts` is a new independent copy. Whether the returned temporary was const or not doesn't affect what the caller can do with `ts`.

It is worse, because it prevents move semantics.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+public:
+  TileSizes() = default;
+  explicit TileSizes(unsigned rank) : dims(rank, kUninitialized) {}
+
+  /// Construct from concrete tile sizes (one value per dimension).
+  TileSizes(ArrayRef<int64_t> sizes) : dims(sizes) {}
+
+  unsigned rank() const { return dims.size(); }
+  bool empty() const { return dims.empty(); }
+  const llvm::SmallVector<int64_t> &getDims() const { return dims; }
```
hanhanW: I thought that you should return `ArrayRef<int64_t>`?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+  int64_t operator[](unsigned i) const { return dims[i]; }
+
+  /// Returns true if all dimensions have a defined (positive) tile size.
```
hanhanW: I think it is better to call out `empty`. A caller might reasonably think `zero dimensions, all zero of them are defined, so it's defined.`

```suggestion
  /// Returns true if the tile sizes are non-empty and every dimension has a concrete tile size (not uninitialized or overdefined).
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+      return success();
+    }
+
+    // Linalg ops: propagate through indexing maps.
+    if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
+      unsigned numLoops = linalgOp.getNumLoops();
+      TileSizes iterTileSizes(numLoops);
+      for (OpOperand &operand : linalgOp->getOpOperands()) {
+        auto &ts = getTileSizesFor(operand.get(),
+                                   operands[operand.getOperandNumber()]);
```
hanhanW: It should just be `TileSizes`, not `auto &`. Please also update other places.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+          continue;
+        }
+        AffineMap map = linalgOp.getMatchingIndexingMap(&operand);
+        assert(map.getNumDims() == numLoops);
+        iterTileSizes.merge(ts.mapToIterationSpace(map));
+      }
+      for (unsigned i = 0; i < linalgOp.getNumDpsInits(); ++i) {
+        OpOperand *init = linalgOp.getDpsInitOperand(i);
+        AffineMap map = linalgOp.getMatchingIndexingMap(init);
+        auto resultTileSizes = iterTileSizes.mapFromIterationSpace(map);
```
hanhanW: Spell out the type, same for other places.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+        if (!resultTileSizes.empty()) {
+          propagateIfChanged(results[i], results[i]->join(resultTileSizes));
+        }
```
hanhanW: This is a redundant check because merging with empty is a no-op.

```
  void merge(const TileSizes &other) {
      ...
      if (other.empty()) {
        return;
      }
      ...
  }
```

So `join(existingLattice, emptyTileSizes)` returns `NoChange`, and `propagateIfChanged` does nothing. The `if (!resultTileSizes.empty())` guard just duplicates logic that merge already handles.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+  visitOperation(Operation *op, ArrayRef<TileSizeLattice *> operands,
+                 ArrayRef<const TileSizeLattice *> results) override {
+    // to_layout is always an anchor op; Propagate tile sizes backward to the
+    // input.
+    if (auto toLayout = dyn_cast<ToLayoutOp>(op)) {
+      auto &ts = getTileSizesFor(toLayout.getResult(), results[0]);
+      if (!ts.empty()) {
+        TileSizeLattice *inputLattice = operands[0];
+        propagateIfChanged(inputLattice, inputLattice->meet(ts));
+      }
```
hanhanW: ditto: redundant check.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+      return success();
+    }
+
+    // Linalg ops: propagate through indexing maps.
+    if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
+      unsigned numLoops = linalgOp.getNumLoops();
+      TileSizes iterTileSizes(numLoops);
+      // Gather result tile sizes into iteration space via DPS init maps.
+      for (auto [result, resultLattice] :
+           llvm::zip(linalgOp.getOperation()->getResults(), results)) {
```
hanhanW: Use `zip_equal` when possible. `zip` trims the elements if the length does not match. Also, you don't need `getOperation()` I believe.

```suggestion
           llvm::zip_equal(linalgOp->getResults(), results)) {
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+};
+
+/// Read the TileSizes from a lattice, returning empty tile sizes if the lattice
+/// value is from a duplicatable operation.
+static const TileSizes getTileSizesFor(Value val,
+                                       const TileSizeLattice *lattice) {
+  if (!lattice) {
+    return {};
+  }
+  auto &tileSizes = lattice->getValue();
```
hanhanW: spell out the type. It is not trivial to know the type. Is it `const TileSizes &`?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+/// any dimension is uninitialized or overdefined.
+static std::optional<SmallVector<int64_t>>
+getPerDimTileSizes(linalg::LinalgOp linalgOp, const DataFlowSolver &solver) {
+  TileSizes tileSizes = getIterationSpaceTileSizes(linalgOp, solver);
+  if (!tileSizes.isDefined()) {
+    return std::nullopt;
+  }
+  assert(tileSizes.rank() == linalgOp.getNumLoops());
+  return tileSizes.getDims();
+}
```
hanhanW: Do you really need this static method? If we inline it, it is just around a few lines?

```
if (!tileSizes.isdefined() {
  LDBG() << ...
}
// continue...
```

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
hanhanW: I see, I was looking for isaFillInterface, but the upstream API sucks. How does it only accept `linalg::GenericOp`. Anyway, we can leave it as what it is.

compiler/src/iree/compiler/Codegen/Common/VectorTileSizeAnalysis.cpp
hanhanW: I think we should emit errors or at least warnings if there are overdefined tile sizes. It means that there is a bug in the analysis or input program.

It is okay if there are uninitialized tile sizes, because some dimension may not be tiled due to odd reasons.

The point is over-defined is a bug to me.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+        propagateIfChanged(inputLattice, inputLattice->meet(ts));
+      }
+      return success();
+    }
+
+    // Linalg ops: propagate through indexing maps.
+    if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
+      unsigned numLoops = linalgOp.getNumLoops();
+      TileSizes iterTileSizes(numLoops);
+      // Gather result tile sizes into iteration space via DPS init maps.
```
hanhanW: Out of curious, I wonder if we really need to iterate on `inits`? They usually tie to result already. I feel that all the operands from inits all belong to "isDuplicable` category?

btw, this is a redundant loop, because `getOpOperands` includes both `inputs` and `inits`.

---

`Init` is quite well-known in linalg ops, so I think we don't need `DPS` in the comment.

```suggestion
      // Gather result tile sizes into iteration space via init maps.
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+  assert(tileSizes.rank() == linalgOp.getNumLoops());
+  return tileSizes.getDims();
+}
+
+//===----------------------------------------------------------------------===//
+// MaterializeVectorTileSizesPass
+//===----------------------------------------------------------------------===//
+
+#define GEN_PASS_DEF_MATERIALIZEVECTORTILESIZESPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
```
hanhanW: We usually have this at the top of the file.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+#include "llvm/Support/DebugLog.h"
+#include "mlir/Analysis/DataFlow/SparseAnalysis.h"
+#include "mlir/Analysis/DataFlow/Utils.h"
+#include "mlir/Analysis/DataFlowFramework.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/SymbolTable.h"
+
+#define DEBUG_TYPE "iree-codegen-vector-tile-size-analysis"
```
hanhanW: It should match to the pass name or filename.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+    if (failed(solver.initializeAndRun(funcOp))) {
+      return signalPassFailure();
+    }
+
+    funcOp->walk([&](linalg::LinalgOp linalgOp) {
+      std::optional<SmallVector<int64_t>> perDimSizes =
+          getPerDimTileSizes(linalgOp, solver);
+      if (!perDimSizes) {
+        LDBG() << "Analysis did not determine tile size for" << *linalgOp;
```
hanhanW:
```suggestion
        LDBG() << "Analysis did not determine tile size for " << *linalgOp;
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+        return;
+      }
+
+      LDBG() << "Materializing tile size on " << *linalgOp;
```
hanhanW: This log is redundant to me, because it is easy to spot in the output IR. The above one is more useful to me, because you can grab failed cases quickly with the `debug-only` flag.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,469 @@
...
+  bool isDefined() const {
+    return !empty() && llvm::all_of(dims, [](int64_t v) {
+      return v != kUninitialized && v != kOverdefined;
+    });
+  }
+
+  /// Returns true if any dimension is overdefined.
+  bool isOverdefined() const {
+    return llvm::any_of(dims, [](int64_t v) { return v == kOverdefined; });
+  }
```
hanhanW, line 106: FYI, this is not used at all.

compiler/src/iree/compiler/Codegen/Common/test/materialize_vector_tile_sizes.mlir
```diff
@@ -0,0 +1,281 @@
+// RUN: iree-opt --pass-pipeline='builtin.module(any(iree-codegen-materialize-vector-tile-sizes))' --split-input-file %s | FileCheck %s
```
hanhanW: This is my first time seeing `any`, and I learned it. I wonder if we should just use `func.func` because we usually see `func.func` in codegen IRs.

Allowing `any` may introduce inconsistency when people add tests. E.g., they may use `util.func`.

IIRC, IREE only has one `any` use and they are all new code (introduced by @Groverkss ). I feel that `Claude` uses it once and the rest just copy-over this.

hanhanW (changes requested):
Two high-level comments:

- I think most of `empty` checks in `tileSizes` are redundant, see https://github.com/iree-org/iree/pull/23668/changes#r2966973979 for the reason
- We should be consistent in terms of naming. I'm +1 on replacing all the `ts` with `tileSizes`.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: This is the backward propagation. In this case, we (potentially) have tile size information for the results of the `LinalgOp` and want to propagate that to its operands. This loop is not iterating over (DPS) init operands, but over the results, we're just using the indexing map for the corresponding init operand.

In this case, we're also not just propagating to the init operands (which are mostly duplicatable), but also to the other operands, if we can fully define their tile sizes.

I've removed the "DPS" from the comment.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: If it's OK for you, I'd like to keep it here. The part above is mostly the analysis, and we have a comment separating the pass bit of this file from the rest just above.

sommerlukas (commented):
Thanks for the review, I've addressed the comments. Two answers inline.

compiler/src/iree/compiler/Codegen/Common/test/materialize_vector_tile_sizes.mlir
Groverkss: I used this intentionally. It's mainly following the idea that any operation implementing the FunctionOpInterface is a valid input.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
hanhanW: I still like it better if we move it to the top, because it is more like a declaration to me. I'll leave the decision to you.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,456 @@
...
+public:
+  TileSizes() = default;
+  explicit TileSizes(unsigned rank) : dims(rank, kUninitialized) {}
+
+  /// Construct from concrete tile sizes (one value per dimension).
+  TileSizes(ArrayRef<int64_t> sizes) : dims(sizes) {}
+
+  unsigned rank() const { return dims.size(); }
+  bool empty() const { return dims.empty(); }
+  const ArrayRef<int64_t> getDims() const { return dims; }
```
hanhanW: `ArrayRef` already has const semantics, I think we should drop `const` like other implementations. E.g., https://github.com/llvm/llvm-project/blob/b670265b588f51eac9f7cc08edb3787361022af8/mlir/include/mlir/IR/BuiltinTypes.h#L65-L66

```
  /// Returns the shape of this tensor type.
  ArrayRef<int64_t> getShape() const;
```



```suggestion
 ArrayRef<int64_t> getDims() const { return dims; }
```

compiler/src/iree/compiler/Codegen/Common/test/materialize_vector_tile_sizes.mlir
hanhanW, line 228: Does the `linalg.fill` op have a hint? Can you add the check or check-not?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
hanhanW: Okay, I misread it. It is getting the values from `result` not `init` operands. Thanks for the details!

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,456 @@
...
+    // Linalg ops: propagate through indexing maps.
+    if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
+      unsigned numLoops = linalgOp.getNumLoops();
+      TileSizes iterTileSizes(numLoops);
+      for (OpOperand &operand : linalgOp->getOpOperands()) {
+        TileSizes tileSizes = getTileSizesFor(
+            operand.get(), operands[operand.getOperandNumber()]);
+        if (tileSizes.empty()) {
+          continue;
+        }
```
hanhanW: This is a redundant check to me, per the previous high-level comments.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,456 @@
...
+        assert(map.getNumDims() == numLoops);
+        iterTileSizes.merge(tileSizes.mapToIterationSpace(map));
+      }
+      // Gather operand tile sizes into iteration space.
+      for (OpOperand &operand : linalgOp->getOpOperands()) {
+        TileSizes tileSizes = getTileSizesFor(
+            operand.get(), operands[operand.getOperandNumber()]);
+        if (tileSizes.empty()) {
+          continue;
+        }
```
hanhanW: ditto

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,456 @@
...
+        assert(map.getNumDims() == numLoops);
+        iterTileSizes.merge(tileSizes.mapToIterationSpace(map));
+      }
+      // Map iteration space tile sizes back to each operand.
+      for (OpOperand &operand : linalgOp->getOpOperands()) {
+        AffineMap map = linalgOp.getMatchingIndexingMap(&operand);
+        TileSizes operandTileSizes = iterTileSizes.mapFromIterationSpace(map);
+        if (operandTileSizes.empty()) {
+          continue;
+        }
```
hanhanW: ditto

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,456 @@
...
+    if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
+      unsigned numLoops = linalgOp.getNumLoops();
+      TileSizes iterTileSizes(numLoops);
+      // Gather result tile sizes into iteration space via init maps.
+      for (auto [result, resultLattice] :
+           llvm::zip_equal(linalgOp->getResults(), results)) {
+        TileSizes tileSizes = getTileSizesFor(result, resultLattice);
+        if (tileSizes.empty()) {
+          continue;
+        }
```
hanhanW: ditto

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -0,0 +1,456 @@
...
+                                            const DataFlowSolver &solver) {
+  unsigned numLoops = linalgOp.getNumLoops();
+  TileSizes iterTileSizes(numLoops);
+  for (OpOperand &operand : linalgOp->getOpOperands()) {
+    Value val = operand.get();
+    const TileSizeLattice *lattice = solver.lookupState<TileSizeLattice>(val);
+    TileSizes tileSize = getTileSizesFor(val, lattice);
+    if (tileSize.empty()) {
+      continue;
+    }
```
hanhanW, line 398: ditto

hanhanW (approved):
LGTM, just minor comments.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: I removed the other checks after your previous comments but kept the ones that continue the loop early. While it's true that merge would handle this, we perform non-trivial operations before we eventually call `merge`, so an early continuation of the loop might make sense.

It's also an application of [this](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code) rule in the LLVM Coding Standards. 

What do you think, should we keep it?

hanhanW: Just for future reference, let me post my thought here, and thanks for addressing all the comments!

My feeling is that the rest of computation is very cheap, so I lean to remove the early-exit. The LLVM rule fits multiple conditions better to me.

It is a simple early-exit, and I can think about few downside (or I can explain what I was thinking when I saw the change).
- If it shows up several times, I'd need to look at other places that don't have the logic. If we require it, then it spams.
- If I understand the idea correctly and based on my linalg knowledge, it is handling "scalar operand", which indicates the indexing map is trivial and it is not a special case.

> When you have very, very small loops, this sort of structure is fine.

I don't expect the loop becoming large, and it makes code cleaner.

PR #23669, opened 2026-03-05
[NFC][Codegen] Move distribution pattern infrastructure to VectorExt

Groverkss:
Move DistributionPattern, VectorLayoutOptions, DistributionSignature, and related signature utility functions from Codegen/Common/GPU/ into Codegen/Dialect/VectorExt/Transforms/DistributionPatterns.{h,cpp}.

The concrete NestedLayout distribution patterns remain in Common/GPU/ to avoid pulling Codegen-level dependencies into the VectorExt dialect layer.

This change allows us to define distribution patterns outside of Common/GPU without having a dependency on Common/GPU. This allows us to define new patterns by only taking a dependency on VectorExtTransforms.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/BUILD.bazel
@@ -118,6 +118,7 @@ iree_compiler_cc_library(
         "//compiler/src/iree/compiler/Codegen/Dialect/PCF/IR",
         "//compiler/src/iree/compiler/Codegen/Dialect/PCF/Transforms",
         "//compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR:IREEVectorExtDialect",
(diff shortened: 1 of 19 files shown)


MaheshRavishankar (approved):

AGindinson (approved):
LGTM

Do I understand correctly that the original infra in [GPUVectorDistribution.cpp](https://github.com/iree-org/iree/pull/23669/changes#diff-279bc812c5a931e7af48dc5473c4afc7cc999deb9fb6fb50c42a829313a0d643) has not been unit-tested so far? Or am I missing some LITs that could also be converted into `VectorExt` dialect tests?

Groverkss:
> LGTM
> 
> Do I understand correctly that the original infra in [GPUVectorDistribution.cpp](https://github.com/iree-org/iree/pull/23669/changes#diff-279bc812c5a931e7af48dc5473c4afc7cc999deb9fb6fb50c42a829313a0d643) has not been unit-tested so far? Or am I missing some LITs that could also be converted into `VectorExt` dialect tests?

There are unit tests, they live in Common/GPU/test/gpu_nested...

PR #23671, opened 2026-03-05
[Codegen] Add iree_map dialect with PackMapAttr

Groverkss:
Introduces the `iree_map` dialect and `PackMapAttr`, a layout map defined by a (Shape, Stride) pair of hierarchical IntTuples. Provides layout algebra operations (compose, complement, logicalDivide, logicalProduct, filter, rightInverse, etc.) for use in vector distribution passes.


Groverkss:
I will send more information on discord or an RFC

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapBase.td
```diff
@@ -0,0 +1,45 @@
...
+
+include "mlir/IR/OpBase.td"
+include "mlir/IR/AttrTypeBase.td"
+
+//===----------------------------------------------------------------------===//
+// IREE Map Dialect
+//===----------------------------------------------------------------------===//
+
+def IREEMap_Dialect : Dialect {
+  let name = "iree_map";
```
MaheshRavishankar, line 18: `iree_map` is too general a term. Maybe `iree_coordinate_map`?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.td
```diff
@@ -0,0 +1,166 @@
...
+
+  let hasCustomAssemblyFormat = 1;
+  let genVerifyDecl = 1;
+
+  let extraClassDeclaration = [{
+    // --- Properties ---
+    int32_t getRank();
+    int32_t getDepth();
+    /// Product of all leaf shape values. May overflow for large layouts.
+    int32_t getSize();
```
MaheshRavishankar: Why are you chosing int32_t. Saving those bits is not going to give us much in the compiler :)

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Parse an IntTuple: either a bare integer or `(` IntTuple (`,` IntTuple)*
+/// `)`.
+static FailureOr<Attribute> parseIntTuple(AsmParser &parser) {
+  MLIRContext *ctx = parser.getContext();
+
+  if (succeeded(parser.parseOptionalLParen())) {
+    SmallVector<Attribute> elements;
+    auto result = parseIntTuple(parser);
```
MaheshRavishankar: Nit: You could use a `do... while` loop?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+using namespace mlir;
+using namespace mlir::iree_compiler::IREE::Map;
+
+//===----------------------------------------------------------------------===//
+// IntTuple parsing/printing helpers
+//===----------------------------------------------------------------------===//
+
+/// Parse an IntTuple: either a bare integer or `(` IntTuple (`,` IntTuple)*
+/// `)`.
+static FailureOr<Attribute> parseIntTuple(AsmParser &parser) {
```
MaheshRavishankar: Couldnt we just use `assemblyFormat` for this?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+      }
+    }
+    return success();
+  };
+  if (failed(checkPositive(shape, "shape"))) {
+    return failure();
+  }
+
+  SmallVector<int32_t> strideLeaves = getLeaves(stride);
+  for (int32_t v : strideLeaves) {
```
MaheshRavishankar: Nit: If you are using lambda above, use the same lamda here. If not then lambda above seems unnecessary

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+                         << v;
+    }
+  }
+
+  if (!isCongruent(shape, stride)) {
+    return emitError() << "shape and stride must be congruent (identical tree "
+                          "structure)";
+  }
+
+  if (isLeaf(shape)) {
```
MaheshRavishankar: Why this?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+// PackMapAttr — property methods
+//===----------------------------------------------------------------------===//
+
+int32_t PackMapAttr::getRank() { return Map::getRank(getShape()); }
+
+int32_t PackMapAttr::getDepth() { return Map::getDepth(getShape()); }
+
+int32_t PackMapAttr::getSize() { return Map::getSize(getShape()); }
+
+/// Maximum index + 1: cosize = sum((s-1)*d for each leaf) + 1.
```
MaheshRavishankar: I think this needs to go on the declaration (in the TD file)

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.td
```diff
@@ -0,0 +1,166 @@
...
+    int32_t getCosize();
```
MaheshRavishankar: I think this and the methods below need some comments.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+  } else if (coord.size() == shapes.size()) {
+    natCoord.assign(coord.begin(), coord.end());
+  } else {
+    // Partial coordinate: fewer coords than leaves but more than 1.
+    // Each coord indexes a top-level mode; encode as a flat index
+    // using Horner's method left-to-right (lexicographic / row-major).
+    int32_t flatIdx = 0;
+    SmallVector<int32_t> modeShapes;
+    for (int32_t i = 0; i < getRank(); ++i) {
+      modeShapes.push_back(Map::getSize(getShapeMode(i)));
```
MaheshRavishankar: I personally find this indirection hard to read.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+  SmallVector<int32_t> shapes = getLeaves(getShape());
+  assert(shapes.size() == strides.size());
+
+  SmallVector<int32_t> natCoord;
+  if (coord.size() == 1 && shapes.size() > 1) {
+    natCoord = idx2crd(coord[0], getShape());
+  } else if (coord.size() == shapes.size()) {
+    natCoord.assign(coord.begin(), coord.end());
+  } else {
+    // Partial coordinate: fewer coords than leaves but more than 1.
```
MaheshRavishankar: What if you have more coords than leaves?

MaheshRavishankar: Do we have a case where we want to support this partial thing? Seems like a bit of a footgun.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+      accShape = si * accShape;
+      continue;
+    }
+    result.push_back({si, di});
+  }
+
+  std::reverse(result.begin(), result.end());
+
+  SmallVector<Attribute> newShape, newStride;
+  for (auto [s, d] : result) {
```
MaheshRavishankar: Instead of reverse above, you can just do llvm::reverse in the loop, which should avoid the rewriter from above?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+
+  SmallVector<std::pair<int32_t, int32_t>> result;
+  result.push_back({shapes.back(), strides.back()});
+
+  for (int i = static_cast<int>(shapes.size()) - 2; i >= 0; --i) {
+    int32_t si = shapes[i];
+    int32_t di = strides[i];
+    auto &[accShape, accStride] = result.back();
+
+    if (si == 1) {
```
MaheshRavishankar, line 214: I think you still need to check the stride?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+///   - si == 1: trivial, contributes nothing -- skip.
+///   - accShape == 1: accumulator is trivial -- replace it with (si, di).
+///   - di == accShape * accStride: the new leaf starts exactly where the
+///     accumulator ends, so they are contiguous -- merge into
+///     (si * accShape, accStride).
+///   - otherwise: non-contiguous -- push the accumulator and start fresh.
+///
+/// Example: (2, 4) : (4, 1) -> di=4 == 4*1 -> contiguous -> (8) : (1).
+///          (4, 8) : (1, 4) -> di=1 != 8*4 -> not contiguous -> unchanged.
+PackMapAttr PackMapAttr::coalesce() {
```
MaheshRavishankar: Hmm, just reading this, I am not so sure about the logic here. Will look at tests to figure out how "correct" this is.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+/// `accumulated` -- the contiguous block [0, accumulated) covered so far.
+/// If the next mode has stride d > accumulated, the indices
+/// [accumulated, d) are unreached -- a gap. We fill it with a complement
+/// mode of shape d/accumulated and stride accumulated: this places
+/// d/accumulated copies of the already-covered block at regular offsets,
+/// exactly tiling the gap. After processing a mode (s, d), the frontier
+/// advances to d * s. A final mode extends from the last frontier to
+/// cotarget.
+///
+/// Example: A = (4):(1) covers [0,4), accumulated=4, cotarget=16.
```
MaheshRavishankar: I dont think I understand this example at all. Maybe a more representative example helps? Maybe I still dont undestand the representation, but shape is `(4)` and stride is `(1)` . How is the cotarget `16` ?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+
+int32_t PackMapAttr::getRank() { return Map::getRank(getShape()); }
+
+int32_t PackMapAttr::getDepth() { return Map::getDepth(getShape()); }
+
+int32_t PackMapAttr::getSize() { return Map::getSize(getShape()); }
+
+/// Maximum index + 1: cosize = sum((s-1)*d for each leaf) + 1.
+int32_t PackMapAttr::getCosize() {
+  SmallVector<int32_t> shapes = getLeaves(getShape());
```
MaheshRavishankar: What is the difference between `size` and `cosize`?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+}
+
+TEST_F(PackMapTest, CoalesceRemovesUnitModes) {
+  // (1, 8) : (0, 1) -> coalesced to (8) : (1)
+  auto layout = make({1, 8}, {0, 1});
+  auto coalesced = layout.coalesce();
+  EXPECT_EQ(coalesced.getRank(), 1);
+  EXPECT_EQ(getLeafValue(coalesced.getShapeMode(0)), 8);
+}
+
```
MaheshRavishankar, line 201: I am not sure what happens when there is a layout with holes?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
MaheshRavishankar: I got claude to give me the answer, but would be good to document. (Man Claude knows this shit well!)

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.td
```diff
@@ -0,0 +1,166 @@
...
+// PackMapAttr
+//===----------------------------------------------------------------------===//
+
+def IREEMap_PackMapAttr : IREEMap_Attr<"PackMap"> {
+  let mnemonic = "pack_map";
+
+  let summary = [{A layout map: a (Shape, Stride) pair mapping coordinates to indices.}];
+
+  let description = [{
+    A layout map is a function from coordinates to 1-D indices, defined by a
```
MaheshRavishankar, line 22: You should mention somewhere here that 0 stride is broadcast.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+TEST_F(PackMapTest, EvaluateColumnMajor) {
+  // (4, 8) : (1, 4) — column-major
+  auto layout = make({4, 8}, {1, 4});
+  EXPECT_EQ(layout.evaluate({0, 0}), 0);
+  EXPECT_EQ(layout.evaluate({1, 0}), 1);
+  EXPECT_EQ(layout.evaluate({0, 1}), 4);
+  EXPECT_EQ(layout.evaluate({3, 7}), 31);
+}
+
```
MaheshRavishankar, line 138: Missing a bunch of test cases.

  1. Case 1 (flat index): No tests for evaluate({flatIdx}) on multi-leaf layouts
  2. Case 3 (partial mode coords): No tests for hierarchical layouts with mode-level coords like evaluate({5, 3}) on a 3-leaf layout
  3. Strided/non-contiguous: No tests for layouts like (4) : (3) or (4, 8) : (2, 1) where cosize > size
  4. Stride-0 (broadcast): No tests for (4, 8) : (0, 1) — evaluate should ignore the broadcast dim
  5. Edge case: evaluate({0}) on any layout (should always return 0)

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+//===----------------------------------------------------------------------===//
+// No auto-coalesce (key difference from PackLayout)
+//===----------------------------------------------------------------------===//
+
+TEST_F(PackMapTest, NoAutoCoalesce) {
+  // Create a layout with redundant modes: (2, 4) : (4, 1) is equivalent
+  // to (8) : (1) after coalescing, but PackMap should NOT auto-coalesce.
+  auto layout = make({2, 4}, {4, 1});
```
MaheshRavishankar: Where is this checking coalesce? Is there a canonicalization or something cause this isnt calling coalesce?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
MaheshRavishankar: I am actually confused about the strides. For example
(4, 1, 8) : (4, 0, 1) and (4, 1, 1) : (4, 99, 1) seem like two different maps. The former seems like it is meant for "broadcasting" while the latter is the slicing. Last time I looked at cute layouts this distinction really drove me nuts, and I am still having issues understanding the consistency here.

MaheshRavishankar (commented):
I have lots of comments :) . I got through only few of the algebra methods here cause I am missing some understanding (and I think the tests are missing some cases that help make this understanding better). Left some initial comments. I can come back and review after that.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/test/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
+# Copyright 2026 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+load("//build_tools/bazel:iree_lit_test.bzl", "iree_lit_test_suite")
+
+package(
+    features = ["layering_check"],
```
sommerlukas, line 10: I'm not a Bazel expert, but other Bazel files for dialects specify a `default_visibility`. Do we need the same here?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+TEST_F(PackMapTest, Rank) {
+  EXPECT_EQ(make({8}, {1}).getRank(), 1);
+  EXPECT_EQ(make({4, 8}, {8, 1}).getRank(), 2);
+  EXPECT_EQ(make({2, 3, 4}, {12, 4, 1}).getRank(), 3);
+}
+
+TEST_F(PackMapTest, Depth) {
+  EXPECT_EQ(make({8}, {1}).getDepth(), 1);
+  EXPECT_EQ(make({4, 8}, {8, 1}).getDepth(), 1);
```
sommerlukas, line 73: Should we add a case with `getDepth() > 1`?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+  EXPECT_EQ(make({4, 8}, {1, 4}).getCosize(), 32);
+}
+
+//===----------------------------------------------------------------------===//
+// Mode access
+//===----------------------------------------------------------------------===//
+
+TEST_F(PackMapTest, ModeAccess) {
+  auto layout = make({4, 8}, {8, 1});
+  EXPECT_EQ(getLeafValue(layout.getShapeMode(0)), 4);
```
sommerlukas, line 102: What would `getLeafValue` do on a more complex example?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+  Attribute outerD = makeLeaf(c, 4);
+  auto layout = PackMapAttr::get(c, makeTuple(c, {innerS, outerS}),
+                                 makeTuple(c, {innerD, outerD}));
+  auto flat = layout.flatten();
+  EXPECT_EQ(flat.getRank(), 3);
+  SmallVector<int32_t> shapes = getLeaves(flat.getShape());
+  EXPECT_EQ(shapes[0], 2);
+  EXPECT_EQ(shapes[1], 4);
+  EXPECT_EQ(shapes[2], 8);
+}
```
sommerlukas, line 245: What are the expected strides after flattening? Should they be tested here, too?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+  // (4) : (1) with cotarget=16 -> complement covers indices 4..15
+  auto layout = make({4}, {1});
+  auto comp = layout.complement(16);
+  // The complement should cover the "rest" of the index space.
+  EXPECT_EQ(comp.getSize(), 4); // 16 / 4 = 4
+}
+
+TEST_F(PackMapTest, ComplementWithGap) {
+  // (4) : (2) with cotarget=8.
+  // Original covers {0, 2, 4, 6}. Complement fills the interleaving gap.
```
sommerlukas, line 314: I may be misunderstanding the complement operation, but wouldn't it need to cover `{1, 3, 5, 7}`?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+TEST_F(PackMapTest, LogicalProductBasic) {
+  // (4):(1) product with (8):(1) -> ((4),(8)):((1),(4)).
+  auto layout = make({4}, {1});
+  auto tiler = make({8}, {1});
+  auto result = layout.logicalProduct(tiler);
+  EXPECT_EQ(result.getRank(), 2);
+  EXPECT_EQ(result.getSize(), 32);
+  // Mode 0 is the original layout.
+  EXPECT_EQ(getSize(result.getShapeMode(0)), 4);
+  EXPECT_EQ(getLeaves(result.getStrideMode(0))[0], 1);
```
sommerlukas: Can we use `getLeafValue` here?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,496 @@
...
+  EXPECT_EQ(getLeafValue(projected.getShapeMode(1)), 2);
+}
+
+//===----------------------------------------------------------------------===//
+// MakeIdentity
+//===----------------------------------------------------------------------===//
+
+TEST_F(PackMapTest, MakeIdentity) {
+  auto id = PackMapAttr::makeIdentity(getContext(), {4, 8});
+  EXPECT_EQ(id.getRank(), 2);
```
sommerlukas: Should we also test `getSize()` in this and the test below?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,387 @@
...
+  int32_t acc = 1;
+  for (int i = n - 1; i >= 0; --i) {
+    result[i] = acc;
+    acc *= leaves[i];
+  }
+  SmallVector<Attribute> attrs;
+  attrs.reserve(n);
+  for (int32_t v : result) {
+    attrs.push_back(makeLeaf(ctx, v));
+  }
```
sommerlukas: You could use `llvm::map_to_vector` here.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,387 @@
...
+  SmallVector<int32_t> coords(n);
+  int32_t remaining = idx;
+  for (int i = n - 1; i >= 0; --i) {
+    coords[i] = remaining % leaves[i];
+    remaining /= leaves[i];
+  }
+  return coords;
+}
+
+int32_t crd2idx(ArrayRef<int32_t> coord, Attribute shape, Attribute stride) {
```
sommerlukas: Why does this function need `shape`?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,387 @@
...
+  assert(shapes.size() == strides.size());
+
+  if (shapes.empty()) {
+    return {shape, stride};
+  }
+
+  SmallVector<std::pair<int32_t, int32_t>> result;
+  result.push_back({shapes.back(), strides.back()});
+
+  for (int i = static_cast<int>(shapes.size()) - 2; i >= 0; --i) {
```
sommerlukas: Why does `i` have to be `int` here?

sommerlukas: Can we use a for-each loop with reverse and dropFront/Back?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+  SmallVector<Attribute> newShape, newStride;
+  for (int64_t i : perm) {
+    newShape.push_back(getShapeMode(i));
+    newStride.push_back(getStrideMode(i));
+  }
+  return PackMapAttr::get(ctx, makeTuple(ctx, newShape),
+                          makeTuple(ctx, newStride));
+}
+
+/// Drop modes where droppedDims[i] is true.
```
sommerlukas: Is the semantic that `true` means drop the dimension in line with other projects (e.g., affine maps) in MLIR?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+/// The original layout covers [0, size). To create copies, we enlarge the
+/// space to target = size * tiler.cosize and take this.complement(target).
+/// The complement covers exactly the indices between copies of `this` --
+/// the "gaps" where new copies can be placed without overlap. Composing
+/// the complement with the tiler parameterizes which copy to access using
+/// the tiler's coordinate system. Mode 0 is the original, mode 1 selects
+/// which copy.
+///
+/// Example: (4):(1) product with (8):(1)
+///          -> ((4), (8)) : ((1), (4)), total size 32.
```
sommerlukas: Does this mean that we make eight copies of our original `(4) : (1)` to cover the entire 32 elements?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+///
+/// The rightInverse pairs each of A's digits with the corresponding
+/// natural stride: same shapes (same digit sizes), but natural strides
+/// instead of A's strides. This creates a layout that converts from A's
+/// output ordering back to natural ordering -- when A is applied on top,
+/// the two reorderings cancel, yielding the original index x.
+///
+/// We collect contiguously from stride=1 because a gap means A skips
+/// some indices (non-injective), limiting the inverse to the contiguous
+/// prefix where A is a bijection.
```
sommerlukas: An example would be nice here.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,716 @@
...
+      .coalesce();
+}
+
+/// Find L such that L(A(x)) = x for all x in [0, size).
+///
+/// If A is not surjective, its range has gaps, so rightInverse alone cannot
+/// cover the full output domain. The fix: complement fills those gaps,
+/// creating a combined layout (comp, A) that IS injective over [0, size).
+/// The rightInverse of this combined layout then works for all outputs,
+/// and when restricted to A's actual outputs, it recovers x.
```
sommerlukas: An example would be nice here.

sommerlukas (commented):
I think there are quite a few opportunities in this PR to use things like `llvm::map_to_vector` instead of for-loops etc, I didn't comment on each of them. 

The documentation on the attribute is a decent start, but I think we should have a page explaining all the algebra in detail and with examples (not necessarily in this PR).

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapBase.td
Groverkss: It's too long and also just a name. I don't think any name we give it will be more descriptive. If we eventually need the name, we can rename it, but i'd prefer keeping the name.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.td
Groverkss: I'll make it int64_t. I think I initially chose int32_t to signify that when generating indexing for this map, it's a hint that you can use int32 instead of int64. But I think the compiler can figure it out.

Groverkss: I'll make it int64_t. I think I initially chose int32_t to signify that when generating indexing for this map, it's a hint that you can use int32 instead of int64. But I think the compiler can figure it out.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: I don't think we can do the full thing as `assemblyFormat`, but I changed it to:

```
`<` custom<IntTuple>($shape) `:` custom<IntTuple>($stride) `>`
```

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: I will have to write a stack based implementation. I changed it, but not sure if there is a big benefit to it.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: removed.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: removed

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: It's also the reason why I want to use these. Claude can really understand what is happening with these layouts.

I moved some comments from .cpp file to .td file where this is explained.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: good question. I'll just remove it, although i know it could be useful, but it's okay to remove.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: removed.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: We treat all size 1 shapes as "broadcasting". There is a check below that converts any shape = 1 to stride = 0. I don't plan on having them as seperate. We don't really use them to represent slicing.

Groverkss: We don't need to check the stride btw. shape == 1 means there is only one coordinate there, 0. 0 * stride = 0, the stride doesn't matter. We canonicalize the stride to 0 here. (We had the same problem with NestedLayoutAttr and did the same).

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: done

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: The cotarget is an input. You specify that the entire range is (0, 16) and the complement is all elements that doen't exist in that range in this layout. I'll update the documentation.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: yes

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: yes, i will write this in the doccumentation to explain better.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: added

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: added

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/test/BUILD.bazel
Groverkss: I don't think test files need that.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: done

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: You can't really have a more complicated example, it needs to be a leaf.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: I can remove that test. I think it's an old artifact where I had PackMap coalesce automatically, but I moved it to PackLayout instead.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: added

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: it doesn't coalesce. Added a test.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: added

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: The complement would need to be a layout that you can do a product with to get (0, 8]. So it just needs to be (2, 1) : (2, 1), which would basically say (4) : (2) starting with 0 offset, and then (4) : (2) starting from offset 1, which would covert both.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: I moved everything to just build the resulting layout.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: I started using `make` builder utility to test these.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
Groverkss: done

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
Groverkss: good question, it probably doesn't.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
Groverkss: i removed this implementation, it got merged with coalesceImpl.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.td
Groverkss: added

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.td
Groverkss: done

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+bool isIntTuple(Attribute attr) {
+  if (!attr) {
+    return false;
+  }
+  if (auto intAttr = dyn_cast<IntegerAttr>(attr)) {
+    return intAttr.getType().isInteger(64);
+  }
+  if (auto arrAttr = dyn_cast<ArrayAttr>(attr)) {
+    return llvm::all_of(arrAttr,
+                        [](Attribute elem) { return isIntTuple(elem); });
```
kuhar:
```suggestion
    return llvm::all_of(arrAttr, isIntTuple);
```

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+Attribute simplify(Attribute attr) {
+  if (isLeaf(attr)) {
+    return attr;
+  }
+  auto arr = cast<ArrayAttr>(attr);
+  if (arr.size() == 1) {
+    return simplify(arr[0]);
+  }
+  SmallVector<Attribute> result;
+  for (auto elem : arr) {
```
kuhar: nit: type

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+  }
+  return makeTuple(attr.getContext(), result);
+}
+
+Attribute flatten(MLIRContext *ctx, Attribute tuple) {
+  SmallVector<int64_t> leaves = getLeaves(tuple);
+  SmallVector<Attribute> leafAttrs;
+  leafAttrs.reserve(leaves.size());
+  for (int64_t v : leaves) {
+    leafAttrs.push_back(makeLeaf(ctx, v));
```
kuhar: map_to_vector?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+// --- Arithmetic ---
+
+int64_t innerProduct(Attribute coord, Attribute stride) {
+  if (isLeaf(coord)) {
+    assert(isLeaf(stride));
+    return getLeafValue(coord) * getLeafValue(stride);
+  }
+  auto coordArr = cast<ArrayAttr>(coord);
+  auto strideArr = cast<ArrayAttr>(stride);
+  assert(coordArr.size() == strideArr.size());
```
kuhar: zip_equal already asserts this

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+  for (int i = n - 1; i >= 0; --i) {
+    coords[i] = remaining % leaves[i];
+    remaining /= leaves[i];
+  }
+  return coords;
+}
+
+int64_t crd2idx(ArrayRef<int64_t> coord, Attribute stride) {
+  SmallVector<int64_t> strides = getLeaves(stride);
+  assert(coord.size() == strides.size());
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+}
+
+SmallVector<int64_t> getLeaves(Attribute attr) {
+  SmallVector<int64_t> result;
+  if (isLeaf(attr)) {
+    result.push_back(getLeafValue(attr));
+    return result;
+  }
+  for (Attribute child : cast<ArrayAttr>(attr)) {
+    auto childLeaves = getLeaves(child);
```
kuhar: nit: type

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+    result.push_back({shapes[i], strides[i], dataStrides[i]});
+  }
+  return result;
+}
+
+SmallVector<LeafInfo>
+filterLeafInfos(Attribute shape, Attribute stride,
+                llvm::function_ref<bool(const LeafInfo &)> pred) {
+  SmallVector<LeafInfo> result;
+  for (auto &leaf : getLeafInfos(shape, stride)) {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,313 @@
...
+    }
+  }
+  return result;
+}
+
+int64_t
+foldLeafInfos(Attribute shape, Attribute stride, int64_t init,
+              llvm::function_ref<int64_t(int64_t, const LeafInfo &)> fn) {
+  int64_t acc = init;
+  for (auto &leaf : getLeafInfos(shape, stride)) {
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.h
```diff
@@ -0,0 +1,125 @@
...
+
+// --- Leaf info ---
+
+/// Info about a single leaf in a (shape, stride) mode pair.
+/// `stride` is the layout stride (0 = broadcast).
+/// `dataStride` is the lex data stride (product of all subsequent leaf sizes).
+struct LeafInfo {
+  int64_t size;
+  int64_t stride;
+  int64_t dataStride;
```
kuhar:
```suggestion
  int64_t size = {};
  int64_t stride = {};
  int64_t dataStride = {};
```

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+        stack.push_back({});
+      } else {
+        int64_t val;
+        if (failed(parser.parseInteger(val))) {
+          return failure();
+        }
+        result = makeLeaf(ctx, val);
+        startItem = false;
+      }
+    } else {
```
kuhar: can we continue early to reduce nesting?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+        }
+        result = makeTuple(ctx, stack.back());
+        stack.pop_back();
+      }
+    }
+  }
+}
+
+static ParseResult parseIntTuple(AsmParser &parser, Attribute &result) {
+  auto parsed = parseIntTupleImpl(parser);
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+
+int64_t PackMapAttr::getRank() { return Map::getRank(getShape()); }
+
+int64_t PackMapAttr::getDepth() { return Map::getDepth(getShape()); }
+
+int64_t PackMapAttr::getSize() { return Map::getSize(getShape()); }
+
+int64_t PackMapAttr::getCosize() {
+  int64_t result = 0;
+  for (const auto &leaf : getLeafInfos(getShape(), getStride())) {
```
kuhar: type

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+  return PackMapAttr::get(ctx, makeTuple(ctx, newShape),
+                          makeTuple(ctx, newStride));
+}
+
+/// Coalesce within each top-level mode independently.
+/// Unlike coalesce(), leaves are never merged across mode boundaries.
+PackMapAttr PackMapAttr::coalesceModes() {
+  MLIRContext *ctx = getContext();
+  SmallVector<Attribute> newShape, newStride;
+  for (int64_t i = 0; i < getRank(); ++i) {
```
kuhar: nit: cache end value

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+    Attribute mShape = getShapeMode(i), mStride = getStrideMode(i);
+    if (isLeaf(mShape)) {
+      newShape.push_back(mShape);
+      newStride.push_back(getLeafValue(mShape) == 1 ? makeLeaf(ctx, 0)
+                                                    : mStride);
+      continue;
+    }
+    auto merged = coalesceImpl(getLeaves(mShape), getLeaves(mStride));
```
kuhar: ntype

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+
+  auto [filtShape, filtStride] = filterZeros(ctx, getShape(), getStride());
+  SmallVector<int64_t> shapes = getLeaves(filtShape);
+  SmallVector<int64_t> strides = getLeaves(filtStride);
+
+  SmallVector<std::pair<int64_t, int64_t>> modes;
+  for (auto [s, d] : llvm::zip_equal(shapes, strides)) {
+    modes.push_back({s, d});
+  }
+  llvm::sort(modes, [](auto &a, auto &b) { return a.second < b.second; });
```
kuhar: use `llvm::less_second`

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+    newStride.push_back(getStrideMode(i));
+  }
+  return PackMapAttr::get(ctx, makeTuple(ctx, newShape),
+                          makeTuple(ctx, newStride));
+}
+
+PackMapAttr PackMapAttr::project(ArrayRef<bool> droppedDims) {
+  MLIRContext *ctx = getContext();
+  SmallVector<Attribute> newShape, newStride;
+  for (size_t i = 0; i < droppedDims.size(); ++i) {
```
kuhar: enumerate

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+      acc *= shapes[i];
+    }
+  }
+
+  SmallVector<std::tuple<int64_t, int64_t, int64_t>> sorted;
+  for (int i = 0; i < n; ++i) {
+    sorted.push_back({strides[i], shapes[i], rStrides[i]});
+  }
+  llvm::sort(sorted,
+             [](auto &a, auto &b) { return std::get<0>(a) < std::get<0>(b); });
```
kuhar: less_first

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapDialect.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "iree/compiler/Codegen/Dialect/Map/IR/IREEMapDialect.h"
+
+#include "iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.h"
+#include "mlir/IR/DialectImplementation.h"
+
+#include "iree/compiler/Codegen/Dialect/Map/IR/IREEMapDialect.cpp.inc"
+
+namespace mlir::iree_compiler::IREE::Map {
+
+struct IREEMapDialectOpAsmInterface final : public OpAsmDialectInterface {
```
kuhar:
```suggestion
struct IREEMapDialectOpAsmInterface final : OpAsmDialectInterface {
```

kuhar (commented):
just drive-by-nits, I only skimmed the PR

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,495 @@
...
+  T(std::initializer_list<T> children) : val(std::vector<T>(children)) {}
+
+  Attribute toAttr(MLIRContext *c) const {
+    if (auto *v = std::get_if<int64_t>(&val)) {
+      return makeLeaf(c, *v);
+    }
+    SmallVector<Attribute> attrs;
+    for (const T &child : std::get<std::vector<T>>(val)) {
+      attrs.push_back(child.toAttr(c));
+    }
```
sommerlukas: I think we can use `map_to_vector` here.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,495 @@
...
+  EXPECT_EQ(make({{2, 3}, 4}, {{12, 1}, 4}).getSize(), 24);
+}
+
+TEST_F(PackMapTest, Cosize) {
+  // (8) : (1) -> cosize = (8-1)*1 + 1 = 8
+  EXPECT_EQ(make({8}, {1}).getCosize(), 8);
+  // (4, 8) : (8, 1) -> cosize = (4-1)*8 + (8-1)*1 + 1 = 32
+  EXPECT_EQ(make({4, 8}, {8, 1}).getCosize(), 32);
+  // Nested ((4, 8)) : ((1, 4)) -> same leaves, cosize = (4-1)*1 + (8-1)*4 + 1 =
+  // 32
```
sommerlukas: The result looks correct, but are the leaves really the same if the strides are different?
```suggestion
  // Nested ((4, 8)) : ((1, 4)) -> cosize = (4-1)*1 + (8-1)*4 + 1 = 32
```

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,495 @@
...
+  // strides.
+  auto layout = make({4, 8}, {1, 4});
+  EXPECT_EQ(layout.evaluate({0}), 0);  // (0,0) -> 0
+  EXPECT_EQ(layout.evaluate({1}), 4);  // (0,1) -> 0*1+1*4 = 4
+  EXPECT_EQ(layout.evaluate({5}), 20); // (0,5) -> 0*1+5*4 = 20
+  EXPECT_EQ(layout.evaluate({8}), 1);  // (1,0) -> 1*1+0*4 = 1
+}
+
+TEST_F(PackMapTest, EvaluateStrided) {
+  // (4):(3) — stride > 1, cosize (10) > size (4).
```
sommerlukas, line 151: Should we test `cosize > size` here as well?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+SmallVector<int64_t> getLeaves(Attribute attr) {
+  SmallVector<int64_t> result;
+  if (isLeaf(attr)) {
+    result.push_back(getLeafValue(attr));
+    return result;
+  }
+  for (Attribute child : cast<ArrayAttr>(attr)) {
+    SmallVector<int64_t> childLeaves = getLeaves(child);
+    result.append(childLeaves.begin(), childLeaves.end());
```
kuhar: append_range

kuhar (commented):
(don't wait for my approval on this, I don't have cycles to meaningfully review unfortunately, but overall +1 for using cute layouts)

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
sommerlukas: Ah, so by repeating the the `(4) : (2)` layout twice, offset by 1 = `(2) : (1)`, we cover the entire range? 

(Just a question for my understanding, no comment on the code itself.)

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
```diff
@@ -0,0 +1,495 @@
...
+  EXPECT_EQ(PackMapAttr::makeIdentity(getContext(), {2, 3, 4}),
+            make({2, 3, 4}, {12, 4, 1}));
+}
+
+//===----------------------------------------------------------------------===//
+// IntTuple: filterLeafInfos and foldLeafInfos
+//===----------------------------------------------------------------------===//
+
+TEST_F(PackMapTest, FilterLeafInfosZeroStride) {
+  // (4, 8) : (0, 1) -> zero-stride leaves: {4, 0, _}
```
sommerlukas: What does `_` signify here?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IntTuple.cpp
```diff
@@ -0,0 +1,306 @@
...
+  if (auto arrAttr = dyn_cast<ArrayAttr>(attr)) {
+    return llvm::all_of(arrAttr, isIntTuple);
+  }
+  return false;
+}
+
+bool isLeaf(Attribute attr) { return isa<IntegerAttr>(attr); }
+
+int64_t getLeafValue(Attribute attr) {
+  return static_cast<int64_t>(cast<IntegerAttr>(attr).getInt());
```
sommerlukas: Doesn't `getInt()` already return `int64_t`?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+
+using namespace mlir;
+using namespace mlir::iree_compiler::IREE::Map;
+
+//===----------------------------------------------------------------------===//
+// IntTuple parsing/printing helpers
+//===----------------------------------------------------------------------===//
+
+/// Parse an IntTuple: either a bare integer or `(` IntTuple (`,` IntTuple)*
+/// `)`. Uses an iterative shift-reduce approach with an explicit stack to
```
sommerlukas: The line break in the middle of the "grammar" is a bit unfortunate.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+
+int64_t PackMapAttr::getDepth() { return Map::getDepth(getShape()); }
+
+int64_t PackMapAttr::getSize() { return Map::getSize(getShape()); }
+
+int64_t PackMapAttr::getCosize() {
+  int64_t result = 0;
+  for (const LeafInfo &leaf : getLeafInfos(getShape(), getStride())) {
+    result += (leaf.size - 1) * leaf.stride;
+  }
```
sommerlukas: Can we use `foldLeafInfos` here?

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+/// Two adjacent leaves can be merged into one when they cover a contiguous
+/// range of indices. Scanning right-to-left (innermost first), we accumulate
+/// into (accShape, accStride) and check each new leaf (si, di):
+///
+///   - si == 1: trivial, contributes nothing -- skip.
+///   - accShape == 1: accumulator is trivial -- replace it with (si, di).
+///   - di == accShape * accStride: the new leaf starts exactly where the
+///     accumulator ends, so they are contiguous -- merge into
+///     (si * accShape, accStride).
+///   - otherwise: non-contiguous -- push the accumulator and start fresh.
```
sommerlukas, line 196: An example would be nice here.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
```diff
@@ -0,0 +1,664 @@
...
+       coalesceImpl(getLeaves(getShape()), getLeaves(getStride()))) {
+    newShape.push_back(makeLeaf(ctx, s));
+    newStride.push_back(makeLeaf(ctx, s == 1 ? 0 : d));
+  }
+  return PackMapAttr::get(ctx, makeTuple(ctx, newShape),
+                          makeTuple(ctx, newStride));
+}
+
+/// Coalesce within each top-level mode independently.
+/// Unlike coalesce(), leaves are never merged across mode boundaries.
```
sommerlukas: An example would be nice. Ideally, it would be the same input as for the `coalesce` example, so the difference becomes clear.

sommerlukas (approved):
Just minor stuff.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
MaheshRavishankar: There are not tests for coalesceModes either

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
MaheshRavishankar: Ok, I think I might have figured out source of my confusion, but want to verify. I tried to see what will happen with a packmap attr of `(4, 1, 8) : (8, 99, 1)` . Asking claude to see if coalesce works here it gave me this

```
  Start with accumulator: (8, 1)
  i=1: si=1 → skip
  i=0: si=4, di=8. accShape*accStride = 8*1 = 8 == 8 → merge: (32, 1)
  Result: (32) : (1)

  Original: maps to {0..7, 8..15, 16..23, 24..31} — contiguous [0, 32). The stride 99 on the size-1 dim is invisible.
  Coalesced: (32):(1) — also contiguous [0, 32). ✓`
```

Now it is saying the original maps to a contiguous set of points `[0, 32)`. This is true if you take the size and delinearize it. But I was thinking of this in terms of memory access, so seeing this as a slice of `[0, 0, 0] [4, 1, 8], [8, 99, 1]` . But these are not equivalent. The "co-ordinates" are contiguous, but in the slice view of this, it is not a contiguous slice. So I shouldnt really think of the stride as "stride in memory". It is just "gaps" in the co-ordinate space that is bounded by the size. Is this correct?

MaheshRavishankar (approved):
I left some more comments, but someone needs to work through the algebra here . Its definitely interesting, but it might be better for someone with hands on keyboard to use this. @sommerlukas maybe to verify all the algebra here. Otherwise I think the algebra methods could be tested better, but I dont think it is a blocker. Could lead to subtle bugs though.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: true, removed

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: I don't think it really matters... evaluate is just a weighted sum over the leaves really. Also, that's is a different test.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/unittests/PackMapTest.cpp
Groverkss: It's the LeafInfo struct, added a value for it.

compiler/src/iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.cpp
Groverkss: Right, yes these are not like stride in memory, each mode can be thought of independently. Yes it is gaps in coordinate space bounded by the size.

PR #23673, opened 2026-03-05
[GPU] Add coalescing to reduction tiling

### LINKED ISSUE #23557: Need pipling support for direct convolutions
See some background information about direct convolutions described here https://github.com/iree-org/iree/issues/23534

The loop structure doesnt lend itself well to prefetching as it has three nested loops, just prefetching the inner loop is not the most effective thing

Here is an example of performance for this shape (with tuning)
```
convfp16 -n 32 -c 256 -H 50 -W 50 -k 256 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC -m conv -g 1 -F 1 -t 1
```
Method | With Prefetch | Without Prefetch
-- | -- | --
IGEMM | 159.77 us* | 181.27 us
Direct Conv | 505.36 us | 168.99 us*

*current default

[Here ](https://gist.github.com/nirvedhmeshram/84547a035d8f6dac88124a28a0e29abf) are the tuned configurations to reproduce this with with `--compile-from=executable-configurations`, (I usually over write the cached boo vmfb and re-run for easy performance assessment)

nirvedhmeshram:
This is useful for doing further optimizations like prefetching see details in https://github.com/iree-org/iree/issues/23557

Regrading the implementation, we check for a parent loop because we may do reduction tiling multiple times e.g in direct convolution we tile the filter reduction dims, do pack to intrinsics and reshape patterns and then we later tile the channel dimension, so we want to be able to coalesce all these loops in one loop. It is assumed that all loops (parent or tiling) are from tiling of the same root op, see discussion in the PR why we didnt end up adding additional checks to verify this.

I check the effect of this on direct convolution and found no difference in the performance on 336 convolution shapes (excluded any filter 1 shapes as they go down GEMM path) and found no performance difference. I believe the true impact of this can only be seen with tuning.

Fixes : https://github.com/iree-org/iree/issues/23557

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_apply_tiling_level.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_apply_tiling_level.mlir
@@ -112,33 +112,150 @@ func.func @matmul_transpose_b(%5: tensor<64x64xf32>, %6: tensor<64x1280xf16>, %7
 
 // -----
 
-#config = #iree_gpu.lowering_config<{reduction = [0, 8]}>
-#map1 = affine_map<(d0, d1) -> (d0, d1)>
-#map2 = affine_map<(d0, d1) -> (d0)>
-func.func @reduction(%3: tensor<128x384xf32>) -> tensor<128xf32> {
+#config = #iree_gpu.lowering_config<{reduction = [0, 8, 4]}>
+#map1 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+#map2 = affine_map<(d0, d1, d2) -> (d0)>
+func.func @reduction(%arg0: tensor<128x384x256xf32>) -> tensor<128xf32> {
+  %c0 = arith.constant 0 : index
+  %c1 = arith.constant 1 : index
+  %c3 = arith.constant 3 : index
   %cst = arith.constant 0.000000e+00 : f32
   %empty = tensor.empty() : tensor<128xf32>
-  %4 = linalg.fill ins(%cst : f32) outs(%empty : tensor<128xf32>) -> tensor<128xf32>
-  %5 = linalg.generic {
-    indexing_maps = [#map1, #map2],
-    iterator_types = ["parallel", "reduction"]
-    } ins(%3 : tensor<128x384xf32>) outs(%4 : tensor<128xf32>) attrs =  {lowering_config = #config} {
-  ^bb0(%in: f32, %out: f32):
-    %7 = arith.addf %in, %out : f32
-    linalg.yield %7 : f32
-  } -> tensor<128xf32>
-  return %5 : tensor<128xf32>
+  %init = linalg.fill ins(%cst : f32) outs(%empty : tensor<128xf32>) -> tensor<128xf32>
+
+  // Parent scf.for loop that will be coalesced with reduction tiling loops.
+  %result = scf.for %iv = %c0 to %c3 step %c1 iter_args(%arg1 = %init) -> (tensor<128xf32>) {
+    %slice = tensor.extract_slice %arg0[0, 0, 0] [128, 384, 256] [1, 1, 1] : tensor<128x384x256xf32> to tensor<128x384x256xf32>
+    %reduced = linalg.generic {
+      indexing_maps = [#map1, #map2],
+      iterator_types = ["parallel", "reduction", "reduction"]
+      } ins(%slice : tensor<128x384x256xf32>) outs(%arg1 : tensor<128xf32>) attrs =  {lowering_config = #config} {
+    ^bb0(%in: f32, %out: f32):
+      %add = arith.addf %in, %out : f32
+      linalg.yield %add : f32
+    } -> tensor<128xf32>
+    scf.yield %reduced : tensor<128xf32>
+  }
+  return %result : tensor<128xf32>
 }
 
 // CHECK-LABEL: func.func @reduction
-//       CHECK:   %[[FILL:.+]] = linalg.fill {{.*}} tensor<128xf32>
-//       CHECK:   scf.for %{{.*}} = %c0 to %c384 step %c8 iter_args(%{{.*}} = %[[FILL]])
-//       CHECK:     linalg.generic {{.*}} ins(%{{.*}} : tensor<128x8xf32>)
+//  CHECK-SAME:   %[[ARG0:[A-Za-z0-9]+]]: tensor<128x384x256xf32>
+//   CHECK-DAG:   %[[C9216:.+]] = arith.constant 9216 : index
+//   CHECK-DAG:   %[[C0:.+]] = arith.constant 0 : index
+//   CHECK-DAG:   %[[C1:.+]] = arith.constant 1 : index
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_apply_padding_online_attention.mlir
```diff
@@ -210,9 +210,9 @@ func.func @online_attention_tile_then_pad_7(%n_batches: index, %query: tensor<?x
 
   // CHECK-COUNT-7: tensor.pad
   //         CHECK: iree_linalg_ext.online_attention {{.*}} ins(%{{[0-9a-z_]*}}, %{{[0-9a-z_]*}}, %{{[0-9a-z_]*}}, %{{[0-9a-z_]*}}, %{{[0-9a-z_]*}}
-  //    CHECK-SAME:   : tensor<4x8x64xf32>, tensor<4x32x64xf32>, tensor<4x32x64xf32>, f32, tensor<4x8xf32>)
+  //    CHECK-SAME:   : tensor<4x?x64xf32>, tensor<4x?x64xf32>, tensor<4x?x64xf32>, f32, tensor<4x?xf32>)
   //    CHECK-SAME:     outs(%{{[0-9a-z_]*}}, %{{[0-9a-z_]*}}, %{{[0-9a-z_]*}}
-  //    CHECK-SAME:   : tensor<4x8x64xf32>, tensor<4x8xf32>, tensor<4x8xf32>)
+  //    CHECK-SAME:   : tensor<4x?x64xf32>, tensor<4x?xf32>, tensor<4x?xf32>)
```
nirvedhmeshram: looks like a missing canonicalization that this dim becomes dynamic, hopefully is not impactful to performance and gets cleaned up later on.

nirvedhmeshram: I investigated this further and the reason is that the delinearize adds floor divswhich the the bound analysis cant support, I think it would be best to not do this coalescing for loops with dynamic UB. I will change that.

nirvedhmeshram: Added guard to not coalesce here in the PR

compiler/src/iree/compiler/Codegen/Common/TileAndFuseUtils.cpp
```diff
@@ -482,6 +485,94 @@ LogicalResult applyTileAndFuseToEachRoot(
...
+      for (LoopLikeOpInterface loop : tiledResults->loops) {
+        if (auto forOp = dyn_cast<scf::ForOp>(loop.getOperation())) {
+          forLoops.push_back(forOp);
+        }
+      }
+
+      // Only include parent if it forms a proper iter_args chain with the
+      // tiled loops. This follows the same validation as upstream's
+      // coalescePerfectlyNestedSCFForLoops.
+      if (parentForOp && !forLoops.empty()) {
```
MaheshRavishankar: I am confused why you need this logic? You should be able to just take the loops directly cause they should already satisfy all the constraints your checking by construction?

Max191:
> I check the effect of this on direct convolution and found no difference in the performance on 336 convolution shapes

There was no difference at all? Doesn't this cause prefetching to happen where it otherwise wouldn't? Were you testing the default heuristic or using the tuned configurations that were generated from when we had non-coalesced loops?

compiler/src/iree/compiler/Codegen/Common/TileAndFuseUtils.cpp
nirvedhmeshram: Without this guard this test that I have added `reduction_nochain_iter_args` in the PR will hit this assert

```
iree/third_party/llvm-project/mlir/lib/Dialect/SCF/Utils/Utils.cpp:975: llvm::LogicalResult mlir::coalesceLoops(mlir::RewriterBase &, MutableArrayRef<scf::ForOp>): Assertion `llvm::equal(outerLoop.getRegionIterArgs(), innerLoop.getInitArgs())' failed.
```

While that test is contrived, I wanted to make sure this transform is safe if such a situation ever happens, the actual useful case is where the parent loop has come from reduction tiling of the same root op from a different stage, but if there is already a parent loop op not from that we want to not include that in the coalescing.

nirvedhmeshram:
> > I check the effect of this on direct convolution and found no difference in the performance on 336 convolution shapes
> 
> There was no difference at all? Doesn't this cause prefetching to happen where it otherwise wouldn't? Were you testing the default heuristic or using the tuned configurations that were generated from when we had non-coalesced loops?

By default prefetching is off for direct conv, and i didnt use a tuning spec when doing this comparison, just wanted to show in context of this PR that this transform is not harmful to performance. With tuning I am able to see perf improvement with prefetching when doing coalescing that was not possible without it.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_apply_tiling_level.mlir
```diff
@@ -142,6 +156,109 @@ func.func @reduction(%3: tensor<128x384xf32>) -> tensor<128xf32> {
...
+  return %result : tensor<128xf32>
+}
+
+// CHECK-LABEL: func.func @reduction_dynamic_trip_count
+//  CHECK-SAME:   %[[ARG0:[A-Za-z0-9]+]]: tensor<128x384x256xf32>
+//  CHECK-SAME:   %[[DYN_UB:[A-Za-z0-9]+]]: index
+//   CHECK-DAG:   %[[C0:.+]] = arith.constant 0 : index
+//   CHECK-DAG:   %[[C1:.+]] = arith.constant 1 : index
+//       CHECK:   %[[INIT:.+]] = linalg.fill {{.*}} tensor<128xf32>
+//       CHECK:   scf.for %{{.*}} = %[[C0]] to %[[DYN_UB]] step %[[C1]] iter_args(%[[ARG1:.+]] = %[[INIT]])
```
MaheshRavishankar, line 193: Why does not get coalesced?

compiler/src/iree/compiler/Codegen/Common/TileAndFuseUtils.cpp
MaheshRavishankar: That assert to that example does not match up to me. This logic seems unnecessary to me. The generated loops should already by all `scf.for` if one of them is `scf.for` and the init_args should already setup this way from tiling. So we dont need this logic.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_apply_tiling_level.mlir
MaheshRavishankar: Saw the comment below.

MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/Common/TileAndFuseUtils.cpp
nirvedhmeshram: ok, I have removed the logic and the related test.

PR #23677, opened 2026-03-06
[GPU][Codegen] Implement virtual dense mfma (VDMFMA)

efric:
This PR introduces virtual dense MFMAs (VDMFMA), which implement the sparse trick for skinny GEMM (M=8) workloads inspired by the hugging face article [Creating custom kernels for the AMD MI300](https://huggingface.co/blog/mi300kernels). This patch implements VDMFMA for all CDNA3 variants of the sparse mfma intrinsic. Further plumbing and support for CDNA4 variants will be future PRs.

Sparse MFMA (V_SMFMAC) instructions perform MMA on an imbalanced pair of operands: a 4:2 structured-sparse matrix A and a dense matrix B. The instruction also takes a sparsity index that encodes which 2 of every 4 elements along K are non-zero within the sparse matrix A. The trick exploits this by pairing even/odd lanes to jointly describe a full dense row. 

The benefit over the current approach of padding M=8 to M=16 for dense MFMA is that sparse MFMA processes twice the K-depth of the equivalent dense MFMA in the same number of cycles.

Performance comparison for FP16/FP8 (coupled with optimizing out redundant accumulator processing):

```
shape              vdmfma  baseline  %
-----------------  ------  --------  -----
f16_8x13312x16384  214 us  217 us    +1.4%
f16_8x13312x8192   123 us  122 us    -
f16_8x2304x16384   137 us  145 us    +5.5%
f16_8x2304x8192    114 us  115 us    -
f16_8x6656x16384   144 us  145 us    -
f16_8x6656x8192    116 us  123 us    +5.7%
fp8_8x13312x16384  135 us  133 us    -
fp8_8x13312x8192   113 us  106 us    -6.6%
fp8_8x2304x16384   113 us  126 us    +10.3%
fp8_8x2304x8192    92.6 us 98.6 us   +6.1%
fp8_8x6656x16384   114 us  133 us    +14.3%
fp8_8x6656x8192    97.8 us 108 us    +9.4%
```
Assisted by: Claude


compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -1327,21 +1356,64 @@ void VirtualMMAAttr::getUndistributedTileTypes(
...
-                 getThreadVectorType(context, intrinsic, kMMAOperandAcc)});
+  // VDMFMA layouts pair adjacent lanes to emulate a wider tile, so
+  // threadProduct < subgroupSize (physicalLanesPerThread > 1). We need
+  // getPerLaneElements to divide out the lane grouping and compute the
+  // correct per-lane element count.
+  auto lhsLayout = getSingleSubgroupLayout(intrinsic, kMMAOperandLhs);
+  int64_t subgroupSize = getSubgroupSize();
+  int64_t physicalLanesPerThread =
+      subgroupSize / llvm::product_of(lhsLayout.thread);
+  if (isVDMFMAIntrinsic(intrinsic) && physicalLanesPerThread > 1) {
```
efric, line 1676: this `isVDMFMAIntrinsic` is actually redundant; it was a suggestion to make it explicit in the code that this part is only relevant for VDMFMA

efric:
@krzysz00 updated with `util.hoistable_conversion`

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -1741,6 +2010,48 @@ LogicalResult VirtualMMAAttr::buildUnderlyingOperations(
...
+        /*n=*/16,
+        /*nativeK=*/64,
+        /*unrollFactor=*/getIntrinsicsK(),
+        /*sparseIndexVectorType=*/
+        VectorType::get({2}, builder.getIntegerType(16)),
+        /*evenSparseIndex=*/0x4444,
+        /*oddSparseIndex=*/0xEEEE,
+        /*aSliceWidth=*/8,
+        /*bInterleaveIndices=*/
+        {{0, 1, 16, 17, 2, 3, 18, 19, 4, 5, 20, 21, 6, 7, 22, 23},
```
krzysz00: I assume there wasn't a better way to generate this than writing the constant out?

krzysz00 (approved):
I think this design makes a lot of sense and I don't see any real issues with it. One question (aka can we abstract out some magic constants into a function) and then let's ship it!

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
efric: Thanks! factored into a `getVDMFMABInterleaveIndices`

PR #23679, opened 2026-03-06
[Codegen] Materialize vector masking

sommerlukas:
This introduces a pass to materialize vector masking earlier in the pipeline through two main transformations:
- Unwrap `vector.mask { vector.* }` patterns by applying masking to the operands of `vector.X` with `arith.select` and the corresponding identity value for the `vector.*` operation instead. Currently supported operations for unwrapping are `vector.contract`, `vector.multi_reduction` and `vector.reduction`.
- Decompose `vector.create_mask` and `vector.constant_mask` operations into a combination of `vector.step` and comparison with the mask bounds instead.

Materializing the vector masking through these transformations earlier in the pipeline removes the need to handle `vector.mask` in later passes and reduces their complexity. 

Assisted-by: Claude Code

ci-extra: test_torch


sommerlukas:
The motivation for this pass is that the presence of `vector.mask` operations caused issues in multiple passes later in the pipeline. This is partially due to the design of `vector.mask`, where the operands of the wrapped operation are not captured by the `vector.mask` operation. This causes issues for things like forward/backward slices, leading to bugs in barrier elimination and other passes. It also complicates dataflow propagation for things like `VectorLayoutAnalysis` or `LLVMGPUCastTypeToFitMMA`.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/simplify_vector_masking.mlir
```diff
@@ -0,0 +1,439 @@
+// RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(func.func(iree-llvmgpu-simplify-vector-masking))" %s | FileCheck %s --implicit-check-not="vector.mask"
```
Groverkss: Maybe adding canonoicalize/cse would simplify the checks a bit.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUSimplifyVectorMasking.cpp
```diff
@@ -0,0 +1,418 @@
...
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_LLVMGPUSIMPLIFYVECTORMASKINGPASS
```
Groverkss: materializeVectorMasking would probably be a better name for the pass

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUSimplifyVectorMasking.cpp
```diff
@@ -0,0 +1,418 @@
...
+    AffineMap rhsMap = indexingMaps[1];
+
+    auto lhsMask = projectMaskToOperand(builder, mask, lhsType, lhsMap);
+    if (failed(lhsMask)) {
+      return failure();
+    }
+    auto rhsMask = projectMaskToOperand(builder, mask, rhsType, rhsMap);
+    if (failed(rhsMask)) {
+      return failure();
+    }
```
Groverkss: Note that you only need to mask out the reduction dimensions. You don't need to materialize the mask for any other dimension.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,418 @@
...
+                            vector::CombiningKind kind) const {
+    auto sourceType = cast<VectorType>(reductionOp.getVector().getType());
+    auto maskType = cast<VectorType>(mask.getType());
+    if (maskType.getShape() != sourceType.getShape()) {
+      return failure();
+    }
+    Value identitySource =
+        getCombiningIdentityValue(builder.getLoc(), builder, kind, sourceType);
+    Value sourceMasked = arith::SelectOp::create(
+        builder, mask, reductionOp.getVector(), identitySource);
```
Groverkss, line 294: We only need to mask the reduction dimensions. Hopefully we are only doing that. we should add a comment if we are.

Groverkss (changes requested):
We should only mask out reduction dimensions, parallel dimensions don't really care about masking.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
sommerlukas: I've changed the logic to only mask reduction dimensions for the other cases. I think for the case of `vector.reduction`, it doesn't make a difference because that operation is limited to 1D.

sommerlukas (commented):
I've made the change to only mask reduction dimensions.

As discussed on Discord, I've also added flag, so the decomposition of `create_mask/constant_mask` doesn't happen in the TileAndFuse pipeline for now.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+  return mask;
+}
+
+// Extracts the iterator-dimension positions referenced by an indexing map.
+// Returns failure if any result expression is not a simple AffineDimExpr.
+static FailureOr<SmallVector<int64_t>>
+getIteratorDimsForMap(AffineMap indexingMap) {
+  SmallVector<int64_t> dims;
+  for (int64_t i = 0; i < indexingMap.getNumResults(); ++i) {
```
kuhar: nit: don't recalculate the end index
```suggestion
  for (int64_t i = 0, e = indexingMap.getNumResults(); i < e; ++i) {
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+  SmallVector<int64_t> extractPos(numExtract, 0);
+  Value reductionMask =
+      vector::ExtractOp::create(builder, transposed, extractPos);
+
+  // Broadcast the reduction-only mask to the full operand shape.
+  // Put parallel operand dims first (broadcast dims), reduction dims last
+  // (source dims), then transpose to the correct operand layout.
+  SmallVector<int64_t> bcastPerm;
+  bcastPerm.append(parallelOperandDims.begin(), parallelOperandDims.end());
+  bcastPerm.append(reductionOperandDims.begin(), reductionOperandDims.end());
```
kuhar: use llvm::append_range

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(vector::MaskOp maskOp,
+                                PatternRewriter &rewriter) const override {
+    auto innerOp = dyn_cast<InnerOpType>(maskOp.getMaskableOp());
+    if (!innerOp) {
+      return failure();
+    }
+
+    auto mask = validateMaskOp(maskOp, rewriter);
```
kuhar: type

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+    : UnwrapMaskedOpPattern<UnwrapMaskedContractPattern,
+                            vector::ContractionOp> {
+  using UnwrapMaskedOpPattern::UnwrapMaskedOpPattern;
+
+  FailureOr<vector::ContractionOp>
+  createUnmaskedReplacement(ImplicitLocOpBuilder &builder,
+                            vector::ContractionOp contractOp, Value mask,
+                            vector::CombiningKind kind) const {
+    auto lhsType = contractOp.getLhsType();
+    auto rhsType = contractOp.getRhsType();
```
kuhar: Type / VectorType -- I don't know which one is returned here

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+    Value identityLhs =
+        getCombiningIdentityValue(builder.getLoc(), builder, kind, lhsType);
+    Value identityRhs =
+        getCombiningIdentityValue(builder.getLoc(), builder, kind, rhsType);
+
+    AffineMap lhsMap = indexingMaps[0];
+    AffineMap rhsMap = indexingMaps[1];
+
+    auto lhsMask =
```
kuhar: type

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+        projectMaskToOperand(builder, mask, lhsType, lhsMap, iteratorTypes);
+    if (failed(lhsMask)) {
+      return failure();
+    }
+    auto rhsMask =
```
kuhar: type

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+    SmallVector<vector::IteratorType> iteratorTypes(
+        rank, vector::IteratorType::parallel);
+    for (int64_t dim : multiReduceOp.getReductionDims()) {
+      iteratorTypes[dim] = vector::IteratorType::reduction;
+    }
+
+    AffineMap identityMap =
+        AffineMap::getMultiDimIdentityMap(rank, builder.getContext());
+
+    auto sourceMask = projectMaskToOperand(builder, mask, sourceType,
```
kuhar: type

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+//
+// ConcreteType must implement:
+//   FailureOr<InnerOpType> createUnmaskedReplacement(
+//       ImplicitLocOpBuilder &builder, InnerOpType innerOp,
+//       Value mask, vector::CombiningKind kind) const;
+//
+template <typename ConcreteType, typename InnerOpType>
+struct UnwrapMaskedOpPattern : OpRewritePattern<vector::MaskOp> {
+  using OpRewritePattern::OpRewritePattern;
+
```
kuhar, line 162: you can do
```suggestion
struct UnwrapMaskedOpPattern : OpRewritePattern<vector::MaskOp> {
  using OpRewritePattern::OpRewritePattern;
  using Base = UnwrapMaskedOpPattern;

```

and then have `using Base::Base` in all the patterns that derive from it

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+          builder, DenseElementsAttr::get(maskType, builder.getBoolAttr(true)));
+      rewriter.replaceOp(maskOp, allTrue);
+      return success();
+    }
+
+    // Combine per-dimension masks with AND.
+    Value combined = dimMasks[0];
+    for (size_t i = 1; i < dimMasks.size(); ++i) {
+      combined = arith::AndIOp::create(builder, combined, dimMasks[i]);
+    }
```
kuhar: Use a range-based for loop, you don't need the index

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+    SmallVector<int64_t> invPerm = invertPermutationVector(permToTrailing);
+    return vector::TransposeOp::create(builder, bcast, invPerm);
+  }
+
+  static SmallVector<Value> getUpperBounds(ImplicitLocOpBuilder &builder,
+                                           MaskOpTy maskOp) {
+    if constexpr (std::is_same_v<MaskOpTy, vector::CreateMaskOp>) {
+      return SmallVector<Value>(maskOp.getOperands());
+    } else {
+      SmallVector<Value> bounds;
```
kuhar: nit: reserve the elements upfront or use `map_to_vector`

kuhar (commented):
Just drive-by nits, I didn't review the logic

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
krzysz00: I'm not sure this needs to be a for loop anyway - it could just be iterating over the map

compiler/src/iree/compiler/Codegen/LLVMGPU/test/materialize_vector_masking.mlir
```diff
@@ -0,0 +1,430 @@
...
+      kind = #vector.kind<mul>
+    } %lhs, %rhs, %acc : vector<8x16xf32>, vector<16x8xf32> into vector<8x8xf32>
+  } : vector<8x8x16xi1> -> vector<8x8xf32>
+  return %result : vector<8x8xf32>
+}
+
+// -----
+
+// CHECK-LABEL: func.func @preserve_attributes
+func.func @preserve_attributes(%lhs: vector<8x16xf32>, %rhs: vector<16x8xf32>, %acc: vector<8x8xf32>, %m: index, %n: index, %k: index) -> vector<8x8xf32> {
```
Groverkss: i don't think we need this test

compiler/src/iree/compiler/Codegen/LLVMGPU/test/materialize_vector_masking.mlir
```diff
@@ -0,0 +1,430 @@
+// RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(func.func(iree-llvmgpu-materialize-vector-masking,cse,canonicalize))" %s | FileCheck %s --implicit-check-not="vector.mask"
+
+// CHECK-LABEL: func.func @unwrap_masked_matmul_add(
+// CHECK-SAME: %[[LHS:[a-zA-Z0-9]+]]: vector<8x16xf32>
+// CHECK-SAME: %[[RHS:[a-zA-Z0-9]+]]: vector<16x8xf32>
+// CHECK-SAME: %[[ACC:[a-zA-Z0-9]+]]: vector<8x8xf32>
+// CHECK-SAME: %[[M:[a-zA-Z0-9]+]]: index
+// CHECK-SAME: %[[N:[a-zA-Z0-9]+]]: index
+// CHECK-SAME: %[[K:[a-zA-Z0-9]+]]: index
```
Groverkss: A lot of these tests for contract seem redundant. We only need two tests imo:

1. vector.contract with transposed rhs
2. vector.contract with multiple k dimensions

compiler/src/iree/compiler/Codegen/LLVMGPU/test/materialize_vector_masking.mlir
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// -----
+
+// CHECK-LABEL: func.func @unwrap_non_create_mask(
+// CHECK-SAME: %[[LHS:[a-zA-Z0-9]+]]: vector<8x16xf32>
+// CHECK-SAME: %[[RHS:[a-zA-Z0-9]+]]: vector<16x8xf32>
+// CHECK-SAME: %[[ACC:[a-zA-Z0-9]+]]: vector<8x8xf32>
+// CHECK-SAME: %[[MASK:[a-zA-Z0-9]+]]: vector<8x8x16xi1>
+func.func @unwrap_non_create_mask(
```
Groverkss: we don't need tests for where mask comes from. These are redundant. Can you keep only useful tests?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
+// Copyright 2026 The IREE Authors
```
Groverkss, line 1: This pass doesn't need to live in LLVMGPU. It should like in Codegen/Common

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+      } else {
+        operandDimSizes.push_back(allDimSizes[iterDim]);
+      }
+    }
+    return vector::ConstantMaskOp::create(builder, maskType, operandDimSizes)
+        .getResult();
+  }
+
+  // Generic fallback: extract only reduction dimensions from the iterator mask,
+  // then broadcast to the full operand shape.
```
Groverkss, line 100: Honestly, this is the only path we should maintain. I don't think other paths are useful. They should fold to this. If they don't, we need to fix something somewhere.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUMaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,490 @@
...
+    for (auto [i, iterDim] : llvm::enumerate(*iterDims)) {
+      if (operandIterTypes[i] == vector::IteratorType::parallel) {
+        operandDimSizes.push_back(operandType.getDimSize(i));
+      } else {
+        operandDimSizes.push_back(allDimSizes[iterDim]);
+      }
+    }
+    return vector::ConstantMaskOp::create(builder, maskType, operandDimSizes)
+        .getResult();
+  }
```
Groverkss: we don't need these paths. If something is forcing us to use these paths, we should fix the folders.

Groverkss (changes requested):

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
sommerlukas: I've removed the path for `constant_mask`, as it will be folded. 

However, I think we should keep the path for `create_mask`. 

Assume an example like this:
```mlir
%mask = vector.create_mask %n, %m, %k
vector.mask %mask {
  vector.contract (parallel, parallel, reduction)
}
```

If we don't have the special path for `create_mask`, unwrapping the `vector.mask` will turn this into:
```mlir
%mask = vector.create_mask %n, %m, %k
%lhs_mask = vector.extract %mask [0, 0]
%lhs = arith.select %lhs_mask ...
vector.contract %lhs
```
There is an existing folder for `extract(create_mask)`, but it doesn't handle this case. Folding to `%lhs_mask = vector.create_mask %k` is not correct in the general case, because the mask will be all-false if `%n == 0` or `%m == 0`. 

If we then decompose the `create_mask`, we end up with:
```mlir
%k_cmp = %k < K
%m_cmp = %m < M
%n_cmp = %n < N
%and1 = andi %k_cmp, %m_cmp
%and2 = andi %n_cmp, %and1
%lhs_mask = vector.extract %mask [0, 0]
%lhs = arith.select %lhs_mask ...
vector.contract %lhs
```
This again can't be folded, because `%n == 0` or `%m == 0` could make the mask all-false, so this is the final IR we end up with. 

On the other hand, if we keep the special path for `create_mask`, we can leverage some domain-specific knowledge we have here: (1) we know we're not interested in `%m` for the LHS operand and (2) `%n` is parallel dimension, so we don't care about it for unwrapping the `vector.mask` (it will be handled by the `transfer_read/write`).

So, we can unwrap the mask to:
```mlir
%lhs_mask = vector.create_mask %k
%lhs = arith.select %mask
vector.contract %lhs
```

After decomposing the `create_mask`, we get:
```mlir
%k_cmp = %k < K
%lhs = arith.select %k_cmp
vector.contract %lhs
```

This much more concise and not achievable with a general folder, therefore I believe we should keep the special path for `create_mask`.

sommerlukas (commented):
Thanks for the feedback everyone! 

I've addressed the code nits, moved the pass to `Codegen/Common` and reduced the testing.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
Groverkss: Ugh, this is super annoying how vector.contract is masked... Giving a 3D mask for 2D inputs is very weird.

It's okay to keep the create_mask handeling then... and add a TODO comment that this is because the upstream generation of create_mask is just bad. We shouldn't be generating a 3-D mask for something that is 2-D at all.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,477 @@
...
+
+namespace {
+
+// Validates that maskOp has no passthru and has a vector mask type.
+// Returns the mask value on success, failure otherwise.
+static FailureOr<Value> validateMaskOp(vector::MaskOp maskOp,
+                                       PatternRewriter &rewriter) {
+  if (maskOp.getPassthru()) {
+    return rewriter.notifyMatchFailure(maskOp, "passthru not supported");
+  }
```
Groverkss, line 32: Wouldn't multi_reduction and vector.contract have a passthru in practice? Do they not?

Groverkss (approved):

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
sommerlukas: No, AFAICT `GenericVectorization` doesn't use the passthru. 

We end up [here](https://github.com/llvm/llvm-project/blob/8a9c6a346c53f310c310526e0a5ba2b89dd569a0/mlir/lib/Dialect/Linalg/Transforms/Vectorization.cpp#L568-L569), which calls `mlir::vector::maskOperation` without a passthru argument (defaults to `Value()`), so we don't create `MaskOp` with passthru.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,481 @@
...
+  int64_t numIterDims = iteratorTypes.size();
+  llvm::SmallDenseSet<int64_t> reductionIterDimSet(reductionIterDims.begin(),
+                                                   reductionIterDims.end());
+  SmallVector<int64_t> transposePerm;
+  for (int64_t i = 0; i < numIterDims; ++i) {
+    if (!reductionIterDimSet.contains(i)) {
+      transposePerm.push_back(i);
+    }
+  }
+  transposePerm.append(reductionIterDims.begin(), reductionIterDims.end());
```
kuhar: nit: use append_range

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,481 @@
...
+    } else {
+      parallelOperandDims.push_back(i);
+    }
+  }
+
+  // Transpose the iterator mask to move reduction dims (used by this operand)
+  // to the back, then extract them.
+  int64_t numIterDims = iteratorTypes.size();
+  llvm::SmallDenseSet<int64_t> reductionIterDimSet(reductionIterDims.begin(),
+                                                   reductionIterDims.end());
```
kuhar:
```suggestion
  llvm::SmallDenseSet<int64_t> reductionIterDimSet(llvm::from_range,
                                                   reductionIterDims);
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -0,0 +1,481 @@
...
+
+  int64_t numExtract = numIterDims - reductionIterDims.size();
+  SmallVector<int64_t> extractPos(numExtract, 0);
+  Value reductionMask =
+      vector::ExtractOp::create(builder, transposed, extractPos);
+
+  // Broadcast the reduction-only mask to the full operand shape.
+  // Put parallel operand dims first (broadcast dims), reduction dims last
+  // (source dims), then transpose to the correct operand layout.
+  SmallVector<int64_t> bcastPerm;
```
kuhar: nit: reserve the final size

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorMasking.cpp
```diff
@@ -133,13 +133,12 @@ projectMaskToOperand(ImplicitLocOpBuilder &builder, Value iterMask,
   // Put parallel operand dims first (broadcast dims), reduction dims last
   // (source dims), then transpose to the correct operand layout.
   SmallVector<int64_t> bcastPerm;
+  bcastPerm.reserve(parallelOperandDims.size() + reductionOperandDims.size());
   llvm::append_range(bcastPerm, parallelOperandDims);
   llvm::append_range(bcastPerm, reductionOperandDims);
```
kuhar: actually, you can do this:
```c++
auto bcastPerm = llvm::to_vector(llvm::concat<int64_t>(parallelOperandDims, reductionOperandDim));

kuhar:
(don't wait for my approval on this, I only skimmed the code)

keshavvinayak01 (approved):

PR #23636, opened 2026-03-03
[Torch] Handle dynamic head dimensions for attention

IanWood1:
Uses createOrFold to handle the dynamic case by creating a `tensor.dim` + `math.rsqrt` op.

PR #23680, opened 2026-03-06
 [Torch] Support dynamic head_dim in attention conversion

keshavvinayak01:
When `head_dim` is dynamic, scale is computed at runtime via `index_cast` + `sitofp` + `rsqrt` instead of being folded to a constant.

Static `head_dim` behavior is unchanged.


keshavvinayak01:
Closing, duplicate of https://github.com/iree-org/iree/pull/23636

PR #23684, opened 2026-03-06
[SPIRV] Enable scf.forall-based workgroup distribution

hanhanW:
The forall path also requires `ConvertAccGEMMToGEMM` before distribution to avoid accumulating matmul's shared_outs being bufferized to workgroup memory with a copy-out outside the forall, which violates the workgroup
distribution verifier.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
@@ -53,11 +53,6 @@ static llvm::cl::opt<int> clSPIRVIndexingBits(
     llvm::cl::desc("Set the bit width of indices in SPIR-V."),
     llvm::cl::init(32));
 
-static llvm::cl::opt<bool> clSPIRVTileDispatchUsingForall(
-    "iree-spirv-tile-dispatch-using-forall",
-    llvm::cl::desc("Enable tile and distribute to workgroups using scf.forall"),
-    llvm::cl::init(true));
-
 //===----------------------------------------------------------------------===//
 // Bufferization Configuration
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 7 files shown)


hanhanW:
It depends on

- https://github.com/iree-org/iree/pull/23685
- https://github.com/iree-org/iree/pull/23686

compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
```diff
@@ -53,6 +53,11 @@ static llvm::cl::opt<int> clSPIRVIndexingBits(
     llvm::cl::desc("Set the bit width of indices in SPIR-V."),
     llvm::cl::init(32));
 
+static llvm::cl::opt<bool> clSPIRVTileDispatchUsingForall(
+    "iree-spirv-tile-dispatch-using-forall",
+    llvm::cl::desc("Enable tile and distribute to workgroups using scf.forall"),
+    llvm::cl::init(true));
```
kuhar: should this be hidden? I don't realistically expect anyone to flip this flag

compiler/src/iree/compiler/Codegen/SPIRV/test/pipeline_matmul_cooperative_ops.mlir
```diff
@@ -149,39 +148,28 @@ hal.executable public @matmul_256x1024x128_div_exp {
...
-//         CHECK:     %[[AC:.+]] = spirv.AccessChain %[[C_MEM]]
-//         CHECK:     spirv.KHR.CooperativeMatrixStore %[[AC]], %[[MA5]], %[[C9]], <RowMajor>
-//         CHECK:     %[[AC:.+]] = spirv.AccessChain %[[C_MEM]]
-//         CHECK:     spirv.KHR.CooperativeMatrixStore %[[AC]], %[[MA3]], %[[C9]], <RowMajor>
-//         CHECK:     %[[AC:.+]] = spirv.AccessChain %[[C_MEM]]
-//         CHECK:     spirv.KHR.CooperativeMatrixStore %[[AC]], %[[MA1]], %[[C9]], <RowMajor>
+// CHECK-COUNT-8:     %{{.+}} = spirv.KHR.CooperativeMatrixMulAdd %{{.+}}, %{{.+}}, %{{.+}}
+
+// Matmul results are stored to workgroup memory for the epilogue.
+// CHECK-COUNT-4:     spirv.KHR.CooperativeMatrixStore %{{.+}}, %{{.+}}, %[[C9]], <RowMajor>
```
kuhar: Can we restore checks for the order loaded values?

compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp
hanhanW: I forgot to mention that I'll remove the option if CI is happy. And yes, it should be hidden if we preserve the flag.

kuhar (approved):

krzysz00 (commented):
Seems reasonable to me

PR #23690, opened 2026-03-07
Cherry-pick ODS double-space fix and update tests

kuhar:
Cherry-pick https://github.com/llvm/llvm-project/pull/184253 into third_party/llvm-project and fix double spaces in tests for `gpu.thread_id`, `gpu.subgroup_reduce`, `gpu.shuffle`, `gpu.block_dim`, `nvvm.shfl.sync`, and `transform.print`.


kuhar:
@qedawkins I wanted to prefetch this PR because it can break LIT tests that previously relied on leading spaces, e.g., with `[ foo, ...]`, but turns out all updates were trivial on the IREE side.

qedawkins (approved):

PR #23694, opened 2026-03-09
Replace libc printf with eyalroz/printf and add streaming status formatting.

benvanik:
Integrates [eyalroz/printf](https://github.com/eyalroz/printf) as a vendored dependency to replace all libc snprintf/vsnprintf calls across the runtime. This gives us a consistent, portable printf implementation across all platforms (including embedded and bare-metal targets) with no platform-specific quirks.

Beyond the drop-in replacement, this extends the status formatting API with streaming support:

- Add iree_status_format_to: callback-based streaming formatter that avoids contiguous buffers and pre-measurement passes. The callback returns bool for flow control, enabling early termination on buffer full or allocation failure.

- Add iree_string_builder_append_status: single-pass status-to-builder formatting via iree_status_format_to, resolving the status/string_builder dependency cycle that previously required a local copy in libhsa.c.

- Rewrite iree_status_fprint to stream directly to FILE* with zero heap allocation (previously required a contiguous buffer).

- Rewrite iree_status_to_string as single-pass using string builder instead of two-pass measure-then-format.

- Fix va_copy usage in status allocation: use C99 va_copy for the two-pass measure-then-format pattern instead of calling va_start twice on the same parameter (implementation-defined behavior).

- Convert Status::ToString (C++) and ApiStatusToString (Python bindings) from two-pass iree_status_format to single-pass iree_status_format_to.

- Replace HIP dynamic_symbols.c allocate-format-free pattern with direct iree_string_builder_append_status (eliminates intermediate heap allocation).

- RISC-V toolchain fix: strip debug info from libgcc.a during CMake configure. eyalroz/printf's long double support in %Lf/%Le compiles `va_arg(args, long double)` which on RISC-V (128-bit quad precision) generates references to `__extenddftf2`/`__trunctfdf2` soft-float builtins. These are resolved from libgcc.a, but the GCC 12.2.0-compiled objects in that archive contain DWARF debug info with relocation types 60/61 (in `.debug_rnglists` and `.debug_loclists` sections) that LLD cannot parse. The pre-built RISC-V toolchain lacks compiler-rt builtins as an alternative, so we strip the debug metadata from libgcc.a at configure time using `llvm-objcopy --strip-debug`. This preserves all code and data sections and is idempotent. The strip runs inside the `RISCV_TOOLCHAIN_ROOT` guard so users with their own toolchains are unaffected. Verified locally with the same toolchain used in CI (`toolchain_iree_manylinux_2_28_20231012`).

ci-extra: all

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,3 +37,6 @@
 [submodule "third_party/hsa-runtime-headers"]
 	path = third_party/hsa-runtime-headers
 	url = https://github.com/iree-org/hsa-runtime-headers.git
+[submodule "third_party/printf"]
+	path = third_party/printf
+	url = https://github.com/eyalroz/printf.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1042,6 +1042,9 @@ if(IREE_ENABLE_CPUINFO)
   endif()
 endif()
 
+# Platform-independent printf implementation (eyalroz/printf).
+add_subdirectory(build_tools/third_party/printf EXCLUDE_FROM_ALL)
+
 # This defines the iree-flatcc-cli target, so we don't use EXCLUDE_FROM_ALL.
 add_subdirectory(build_tools/third_party/flatcc)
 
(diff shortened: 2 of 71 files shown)


benvanik:
mac failures are disk space, linux arm64 is unrelated and have a pending a fix.

PR #23696, opened 2026-03-09
[VectorExt] Add TransferScatterOp folds and canonicalization

keshavvinayak01:
Add fold and canonicalization support for transfer_scatter. Templatize the existing transfer_gather fold infrastructure (foldTransferIndexVecs,FoldSingleElementIndexVec, FoldIndexVecAddBroadcast, broadcast/transpose/step folds, all-true mask fold) to work with both ops via TransferOpAdaptor.

Add scatter-specific patterns: FoldAllFalseMaskTransferScatter (returns original base) and FoldContiguousScatterToTransferWrite (lowers to vector.transfer_write).

Part 2/4 from https://github.com/iree-org/iree/pull/23610


compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -190,9 +190,51 @@ struct IndexingMapFoldResult {
...
+      llvm::append_range(operands, newIndexVecs);
+      if (mask) {
+        operands.push_back(mask);
+      }
+      op.getProperties().setOperandSegmentSizes(
+          {1, 1, static_cast<int32_t>(op.getOffsets().size()),
+           static_cast<int32_t>(newIndexVecs.size()),
+           static_cast<int32_t>(mask ? 1 : 0)});
+    }
+  }
```
Groverkss: There is no point of reuse if we are going to if-else them anyway. Please use different structs for each.

Groverkss (changes requested):

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
keshavvinayak01: Done.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -593,14 +654,13 @@ struct FoldIndexVecAddBroadcast final : OpRewritePattern<TransferGatherOp> {
...
-/// Replace an all-false masked transfer_gather with a broadcast of the padding.
-/// Only handles splat constants; non-splat all-false constants are expected to
-/// be canonicalized to splat form beforehand.
-struct FoldAllFalseMaskTransferGather final
-    : OpRewritePattern<TransferGatherOp> {
-  using Base::Base;
+/// Replace an all-false masked transfer_gather/scatter with a no-op.
+/// Gather: broadcast of the padding value. Scatter: return original base.
+template <typename OpTy>
+struct FoldAllFalseMask final : OpRewritePattern<OpTy> {
```
Groverkss: Can we just have a seperate pattern for gather/scatter here?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -616,59 +676,81 @@ struct FoldAllFalseMaskTransferGather final
...
-      // First, apply the mask indexing map if it's not identity.
       AffineMap maskMap = op.getIndexingMapsArray().back();
-      ArrayRef<int64_t> targetShape = op.getType().getShape();
+      ArrayRef<int64_t> targetShape = vectorType.getShape();
       if (!maskMap.isIdentity()) {
         mask = applyTransformMapToVector(rewriter, op.getLoc(), mask, maskMap,
                                          targetShape);
       }
-      // Then, compress the mask to match transfer_read's expected mask type
-      // (which drops broadcast dims from the permutation map).
```
Groverkss, line 657: Please don't drop comments.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -616,59 +676,81 @@ struct FoldAllFalseMaskTransferGather final
...
     if (!op.getIndexVecs().empty()) {
       return failure();
     }
 
     AffineMap permutationMap = op.getPermutationMap();
+    VectorType vectorType = op.getVectorType();
 
     Value mask = op.getMask();
     if (mask) {
-      // First, apply the mask indexing map if it's not identity.
```
Groverkss, line 649: Please don't drop comments.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
```diff
@@ -190,9 +190,67 @@ struct IndexingMapFoldResult {
...
+  static OperandRange getIndexVecs(TransferGatherOp op) {
+    return op.getIndexVecs();
+  }
+  static Value getMask(TransferGatherOp op) { return op.getMask(); }
+  static Value getResult(TransferGatherOp op) { return op.getResult(); }
+
+  static void rebuildOperands(TransferGatherOp op,
+                              SmallVectorImpl<Value> &operands,
+                              SmallVectorImpl<Value> &newIndexVecs,
+                              Value mask) {
```
Groverkss: Can we have a IndexedVecOpInterface for these instead? I think we can reuse most of this code. The interface would have following methods:

// Get the base value that is being indexed 
getBase 
// Get index vecs
getIndexVecs
// Get mask
getMask
// Get indexing map in format: {baseMap, indexVecMaps, optional(maskMap))
getIndexingMaps
rebuildOperands

Then, have a rebuildOperands helper which takes the interface, dyn_casts and rebuilds correctly.

Groverkss (changes requested):
Most of the code could be cleaned up with an interface. The templating is trying to replicate interfaces.

Groverkss (approved):

PR #23703, opened 2026-03-09
[Codegen][VectorExt] Add TransferScatterOp interface implementations

keshavvinayak01:
Add bufferization, GPU nested layout distribution, and vector layout analysis support for TransferScatterOp.

Part 3/4 from https://github.com/iree-org/iree/pull/23610


hanhanW (commented):
I think you should break it into 2-3 PRs:

- Bufferization
- Vector layout analysis (maybe with NestedLayoutDistribution if they are highly related)

keshavvinayak01 (commented):
Will do.

PR #23711, opened 2026-03-09
[NFC] Make TileSwizzle members private

bjacob:
Reviewer note: this is on top of https://github.com/iree-org/iree/pull/23710.  Look only at https://github.com/iree-org/iree/pull/23711/changes/a24ec6b7d5dd9751e106ad1984d52c7c94103b19.

I want to make further changes both immediately after this PR, and potentially in the future.  Having a proper class with private stuff will help with that.


hanhanW (approved):
I reviewed the last commit and it looks good to me. Can we add `nfc` to the PR title?

PR #23712, opened 2026-03-09
[Codegen] Migrate PadOp/PackOp/UnPackOp to VectorizableOpInterface

hanhanW:
No new tests because it is an NFC in terms of functionality. It just follows different mechanism for vectorization.

Assisted-by: Claude

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
+++ b/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
@@ -13,6 +13,8 @@
 #include "iree/compiler/Utils/Indexing.h"
 #include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/UB/IR/UBOps.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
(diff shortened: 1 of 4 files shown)


hanhanW:
It depends on

- https://github.com/iree-org/iree/pull/23662

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -454,6 +456,61 @@ struct MapStoreOpVectorizationModel
...
+    if (failed(result)) {
+      return failure();
+    }
+    return result->replacements;
+  }
+};
+
+/// External model for tensor::PadOp. Different dialect, goes through
+/// linalg::vectorize.
+struct PadOpVectorizationModel
```
sommerlukas, line 488: The two external models look almost identical to me. Can't we use the templated version from above for `PadOp` as well?

Groverkss (approved):

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: It is intended because they are from different dialect. IMHO, the some upstream linalg ops are not completed because they don't implement LinalgOp interface. There may be extentions for such ops, while tensor.pad is a very special case in vectorization.

sommerlukas: But the dialect isn't part of the template, right? IMHO, we shouldn't duplicate code (DRY). If there is an extension of the upstream linalg ops in the future, we can always make a change here.

hanhanW: I think you're over-abstracting the things. The duplication is okay to me if they are small. I've seen so many times about split -> merge -> split ->etc, and my thought is that if we already know it, we should do the split. It is clear to me that they have very different development attention, and the pack/unpack ops evolve much faster than pad op. Thus, splitting them in the first place is reasonable to me.

We also have good reasoning here, but not just "different dialect". As I mentioned, they should implement LinalgOp interface IMO. The other different thing is that tensor.pad op does not implement DPS interface, which can fall into different category easily. It is not happening now, but I'd like to make the isolation in the first place as I can see many difference between them.

sommerlukas: Ok, if you expect a split is likely to happen because of the evolution of pack/unpack, we can keep it as is.

sommerlukas (approved):

PR #23713, opened 2026-03-09
[Codegen] Migrate all ops to VectorizableOpInterface and remove TypeSwitch

hanhanW:
The revision inlines the implementation from VectorExt transforms to avoid circular dependency. Now all the ops are vectorized through the interface, so TypeSwitch is removed from driver.

It also updates `vectorizeGatherLikeGenericToTransferGather` to not rewrite the result internally. It's expected to be replaced by drivers.

It is the last step towards https://lists.lfaidata.foundation/g/iree-technical-discussion/message/15

Assisted-by: Claude


hanhanW:
It depends on

- https://github.com/iree-org/iree/pull/23662
- https://github.com/iree-org/iree/pull/23712

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
```diff
@@ -212,43 +224,17 @@ void GenericVectorizationPass::runOnOperation() {
...
-          if (succeeded(result)) {
-            rewriter.replaceOp(op, result->replacements);
-          }
-        });
-  };
+    FailureOr<SmallVector<Value>> result = vectorizableOp.vectorize(
+        rewriter, vectorSizes, scalableVecDims, linalgOptions);
+    if (succeeded(result)) {
+      rewriter.replaceOp(op, *result);
+    }
```
sommerlukas: Should we signal pass failure here if the `result` is `failure`? The interface documentation says returning `failure` after `isVectorizable` returned `true` is a bug, so maybe we should signal this here. Otherwise, we will have silent failures that we need to track back to here later on.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/VectorizeIREEVectorExtOps.cpp
```diff
@@ -6,15 +6,9 @@
 
 #include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
 #include "iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.h"
-#include "iree/compiler/Codegen/Dialect/VectorExt/Transforms/Transforms.h"
 #include "iree/compiler/Codegen/Interfaces/VectorizableOpInterface.h"
-#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
-#include "iree/compiler/Dialect/LinalgExt/Utils/Utils.h"
-#include "mlir/Dialect/Linalg/IR/Linalg.h"
-#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
```
sommerlukas: Why do we need a new header if we only delete code?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Transforms.h
```diff
@@ -7,19 +7,10 @@
 #ifndef IREE_COMPILER_CODEGEN_DIALECT_VECTOR_EXT_TRANSFORMS_TRANSFORMS_H_
 #define IREE_COMPILER_CODEGEN_DIALECT_VECTOR_EXT_TRANSFORMS_TRANSFORMS_H_
 
-#include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.h"
-#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
-#include "mlir/Dialect/Linalg/IR/Linalg.h"
-#include "mlir/IR/Builders.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/IR/PatternMatch.h"
```
sommerlukas, line 10: Why do we need a new header if we only delete code?

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -456,6 +456,337 @@ struct MapStoreOpVectorizationModel
...
+  AffineMap ident = rewriter.getMultiDimIdentityMap(fullOp.getNumLoops());
+
+  for (Value val : newInputs) {
+    auto [in, map] = tmap[val];
+    ins.push_back(in);
+    indexingMaps.push_back(map);
+  }
+
+  for (Value val : newOutputs) {
+    Value out = tensor::EmptyOp::create(rewriter, fullOp.getLoc(), vectorSizes,
```
sommerlukas, line 505: Nit: We could modernize this code with an `ImplictLocOpBuilder`.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -456,6 +456,337 @@ struct MapStoreOpVectorizationModel
...
+    AffineMap readMap = inverseAndBroadcastProjectedPermutation(map);
+    VectorType readType = VectorType::get(canonicalVectorSizes, elemType);
+
+    SmallVector<Value> operandIndices(map.getNumResults(), zero);
+
+    // TODO: Mask the operation here. It's really hard to do that here though
+    // because we don't have access to the vectorization infra, but maybe there
+    // are easier ways to do it here.
+    auto read = vector::TransferReadOp::create(
+        rewriter, loc, readType, tensor, operandIndices,
```
sommerlukas, line 666: Nit: Use an `ImplictLocOpBuilder`.

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
```diff
@@ -160,32 +159,44 @@ void GenericVectorizationPass::runOnOperation() {
...
+  SmallVector<VectorizableOpInterface> candidates;
+  funcOp.walk([&](VectorizableOpInterface op) {
+    Operation *operation = op.getOperation();
+    // Filter out CopyOp based on pass option.
+    if (isa<linalg::CopyOp>(operation) && !vectorizeCopies) {
+      return;
     }
+    // Filter out PadOp/PackOp/UnPackOp when masking is disabled.
+    // TODO(hanchung): Enable the vectorization without masking. This is mostly
+    // legacy code because it used to not working without masking.
```
sommerlukas, line 188: Probably not as relevant here, as the dependency on masking should go away, but when should such a check happen in the driver vs. when should we pass the option to `isVectorizable` and decide there?

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -456,6 +456,337 @@ struct MapStoreOpVectorizationModel
...
+        }
+        // Fall through to normal vectorization if the gather path did not
+        // apply.
+      }
+    }
+
+    FailureOr<linalg::VectorizationResult> result = linalg::vectorize(
+        rewriter, op, vectorSizes, scalableDims, vectorizeNDExtract,
+        flatten1DDepthwiseConv, /*assumeDynamicDimsMatchVecSizes=*/false,
+        createNamedContraction);
```
Groverkss, line 782: I would eventually like to move away from upstream vectorization, but nothing actionable for now.

Groverkss (approved):

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
hanhanW: Passing InputVectorSize implies masking has been the default behavior for a long time. This option is used to triggering whether a vector size inference is required or not (see below). Reason: some targets do not support masking and it will require emulation in this context. It provides a path to not vectorize the op for some targets like old aarch64 targets.

Just for more context: the code is legacy code that the option was added because the non-masking implementation was done after masking support.

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
hanhanW: Good point, let me give it a shot. It aims to match current behavior, but let's see if it triggers silent bugs or not.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Transforms.h
hanhanW: IWYU, it is for `RewritePatternSet`, see [PatternMatch.h](https://github.com/llvm/llvm-project/blob/e564ce97b1149be53018e4e2e1ffc272a3408f0e/mlir/include/mlir/IR/PatternMatch.h#L818-L822)

```cpp
// RewritePatternSet

class RewritePatternSet {
  using NativePatternListT = std::vector<std::unique_ptr<RewritePattern>>;
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/VectorizeIREEVectorExtOps.cpp
hanhanW: Good catch! I'll remove it and the below `VectorOps.h`.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: I'm not convinced that `ImplictLocOpBuilder` is the modern builder. It's out for a long time, and we don't use it. The past usage that I've seen is mostly in stablehlo project, FYI, but that seems to be someone's preference.

I agree that it may be a proper util for op creation that requires body though, like generic op.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: That would be tricky as we may need to dup the code for other ops like elementwise-op, contraction, reduction, etc.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -530,12 +875,16 @@ void registerVectorizableOpInterfaceExternalModels(DialectRegistry &registry) {
         *ctx);
   });
 
-  // Upstream linalg ops (PackOp, UnPackOp).
+  // Upstream linalg structured ops.
```
hanhanW: I forgot to update the comment when I reviewed the code myself. PackOp and UnPackOp are not structured ops.

```suggestion
  // Upstream linalg ops.
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/VectorizeIREEVectorExtOps.cpp
hanhanW: They were required because we have dialect deps in `Passes.td`. I remove all of them.

hanhanW: (Previously, it relied on indirect include to solve the issue. That's why it's added by Claude.)

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
```diff
@@ -482,6 +482,8 @@ static void addLowerToLLVMPasses(OpPassManager &modulePassManager,
   modulePassManager.addPass(createLowerUKernelOpsToCallsPass());
 
   FunctionLikeNest(modulePassManager)
+      // Decompose vectorized+bufferized map_store ops before lowering to loops.
+      .addPass(IREE::LinalgExt::createDecomposeMapStorePass)
```
hanhanW: I need to drop this commit, it was an experimental change.

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
```diff
@@ -165,32 +164,44 @@ void GenericVectorizationPass::runOnOperation() {
...
-      // mostly legacy code because it used to not working without masking.
-      if (!enableVectorMasking &&
-          isa<tensor::PadOp, linalg::PackOp, linalg::UnPackOp>(op)) {
-        return;
-      }
-      candidates.push_back(op);
+
+  // Build DictionaryAttr options from pass options. These are forwarded to
+  // upstream linalg::vectorize().
+  SmallVector<NamedAttribute> linalgOptionsList;
```
kuhar: nit: We know there are at most 2 elements
```suggestion
  SmallVector<NamedAttribute, 2> linalgOptionsList;
```

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
```diff
@@ -165,32 +164,44 @@ void GenericVectorizationPass::runOnOperation() {
...
+  linalgOptionsList.push_back(
+      rewriter.getNamedAttr("vectorizeNDExtract", rewriter.getBoolAttr(true)));
+  if (vectorizeToTransferGather) {
+    linalgOptionsList.push_back(rewriter.getNamedAttr(
+        "vectorizeToTransferGather", rewriter.getBoolAttr(true)));
+  }
```
kuhar: nit: Don't use `Builder::getNamedAttr`, you can just pass in a StringRef to the NamedAttribute constructor:

```suggestion
  SmallVector<NamedAttribute> linalgOptionsList = {{"vectorizeNDExtract", rewriter.getBoolAttr(true)}};
  if (vectorizeToTransferGather) {
    linalgOptionsList.emplace_back(
        "vectorizeToTransferGather", rewriter.getBoolAttr(true));
  }
```

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
```diff
@@ -165,32 +164,44 @@ void GenericVectorizationPass::runOnOperation() {
...
+  DictionaryAttr linalgOptions =
```
kuhar: ubernit/optional: could use auto here

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -455,6 +455,337 @@ struct MapStoreOpVectorizationModel
   }
 };
 
+static linalg::GenericOp
+buildPartialGenericOp(RewriterBase &rewriter, linalg::GenericOp fullOp,
+                      ArrayRef<int64_t> vectorSizes,
+                      SmallVector<Operation *> partial,
+                      DenseMap<Value, std::pair<Value, AffineMap>> &tmap) {
```
kuhar: What does tmap stand for? I think this would be worth expanding or adding a docstring for.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -455,6 +455,337 @@ struct MapStoreOpVectorizationModel
   }
 };
 
+static linalg::GenericOp
+buildPartialGenericOp(RewriterBase &rewriter, linalg::GenericOp fullOp,
+                      ArrayRef<int64_t> vectorSizes,
+                      SmallVector<Operation *> partial,
```
kuhar: This causes copies

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -455,6 +455,337 @@ struct MapStoreOpVectorizationModel
...
+
+      if (tmap.contains(operand)) {
+        newInputs.insert(operand);
+      }
+    }
+
+    // If a user of the operation is not in partial, it needs to be a result.
+    for (Value result : op->getResults()) {
+      for (Operation *user : result.getUsers()) {
+        if (llvm::find(partial, user) == partial.end()) {
```
kuhar: use `llvm::is_contained`

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -455,6 +455,337 @@ struct MapStoreOpVectorizationModel
...
+  // gather vectorization.
+  SmallVector<Operation *> preExtract;
+  tensor::ExtractOp extractOp;
+  SmallVector<Operation *> postExtract;
+
+  for (Operation &op : linalgOp.getBody()->getOperations()) {
+    if (extractOp) {
+      // Already found extract, add to postExtract.
+      postExtract.push_back(&op);
+    } else {
```
kuhar: nit: continue and no else

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -455,6 +455,337 @@ struct MapStoreOpVectorizationModel
...
+    int64_t symIdx = numSymbols++;
+    sourceMapExprs.push_back(getAffineSymbolExpr(symIdx, ctx));
+    // The index vec map: the read result has shape canonicalVectorSizes, so
+    // it's indexed by all vector dims (identity map).
+    indexVecMaps.push_back(
+        rewriter.getMultiDimIdentityMap(canonicalVectorSizes.size()));
+    indexVecs.push_back(read.getResult());
+  }
+
+  AffineMap sourceMap = AffineMap::get(
```
kuhar: ubernit/optional: could use auto here

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -455,6 +455,337 @@ struct MapStoreOpVectorizationModel
...
+  }
+
+  return *postResult;
+}
+
+/// External model for all linalg structured ops. Wraps upstream
+/// linalg::vectorizeOpPrecondition and linalg::vectorize.
+template <typename OpTy>
+struct LinalgStructuredOpVectorizationModel
+    : public VectorizableOpInterface::ExternalModel<
```
kuhar: nit: `public` is the default for structs, so this is redundant
```suggestion
struct LinalgStructuredOpVectorizationModel
    : VectorizableOpInterface::ExternalModel<
```

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -455,6 +455,337 @@ struct MapStoreOpVectorizationModel
...
+    bool vectorizeNDExtract = getBoolOption(options, "vectorizeNDExtract");
+    bool flatten1DDepthwiseConv =
+        getBoolOption(options, "flatten1DDepthwiseConv");
+    bool createNamedContraction =
+        getBoolOption(options, "createNamedContraction");
+
+    // Handle gather-like generic vectorization via TransferGather.
+    if (auto genericOp = dyn_cast<linalg::GenericOp>(op)) {
+      if (getBoolOption(options, "vectorizeToTransferGather")) {
+        auto gatherResult = vectorizeGatherLikeGenericToTransferGather(
```
kuhar: nit: type

kuhar (commented):
just drive-by nits

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
hanhanW: Is it your coding preference? I don't see an issue of using `Builder::getNamedAttr`. If we already construct things from builder, I'd lean to being consistent.

In terms of LSP intergration, the function signature is provided in this case. Using brace initializers does not provide the signature easily to the tools (at least it does not work for my tools).

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: Good catch -- note that it is an pre-existing issue, and I'm happy to fix it!

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
hanhanW: Note that it is only true in this snapshot, but it is not true for the existing upstream API: https://github.com/llvm/llvm-project/blob/3f65a03e8abb3e6fb3372cf4c254d6c9f090e2e0/mlir/lib/Dialect/Linalg/Transforms/Vectorization.cpp#L2728-L2732

```
FailureOr<VectorizationResult> mlir::linalg::vectorize(
    RewriterBase &rewriter, Operation *op, ArrayRef<int64_t> inputVectorSizes,
    ArrayRef<bool> inputScalableVecDims, bool vectorizeNDExtract,
    bool flatten1DDepthwiseConv, bool assumeDynamicDimsMatchVecSizes,
    bool createNamedContraction) {
```

I'm happy to make the change for this snapshot, but I'd like to point out that it's like asking future contributors to update the size when they add new options.

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
hanhanW: It triggers a bug in CUDA backend: https://github.com/iree-org/iree/issues/23750

I turned it back with additional debug message. I think it is better to bypass the upstream bug (if exists) so the compiler can be more stable.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: It is explained in the below comment. Let me add a function comment and move the documentation there.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.td
```diff
@@ -13,8 +13,6 @@ def VectorizeIREEVectorExtOpsPass :
     Pass<"iree-vector-ext-vectorize-ops", ""> {
   let summary = "Vectorizes then lowers a few iree_vector_ext ops before vectorization.";
   let dependentDialects = [
-    "::mlir::vector::VectorDialect",
```
sommerlukas, line 16: Do we really not depend on these dialects anymore?

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
```diff
@@ -217,43 +228,18 @@ void GenericVectorizationPass::runOnOperation() {
...
-          if (succeeded(result)) {
-            rewriter.replaceOp(op, result->replacements);
-          }
-        });
-  };
+    FailureOr<SmallVector<Value>> result = vectorizableOp.vectorize(
+        rewriter, vectorSizes, scalableVecDims, linalgOptions);
+    if (failed(result)) {
+      return signalPassFailure();
+    }
```
sommerlukas, line 242: I'm a bit confused between this code change and the reply to my previous comment. Does the bug https://github.com/iree-org/iree/issues/23750 currently mean we can't signal pass failure here or can we just keep this change?

hanhanW: I'm sorry about the confusion. Apparently, I made changes without commiting. 🤦‍♂️

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.td
hanhanW: The dependentDialects are mainly for context -- the tool requires the context to generate the new ops. Somehow the context is loaded in the pass. I'm not going to investigate the reason because I'm going to deprecate the pass in a follow-up.

This pass can be deleted because the only use is right before GenericVectorizationPass which is done within this PR. Do you prefer me removing the pass within the PR? I'm okay with both.

https://github.com/iree-org/iree/blob/7649e20ce8d3e8aaaa74d5a91270a28179c6a8a7/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L293-L303

sommerlukas: What happens with the vectorization of `to_layout` that's still in this pass?

I guess the dialects are already loaded into the context because they're used in the input or marked as dependent by other passes. IMHO, it's not great style to rely on that. 

I think it's fine to remove the pass in a separate PR, but then we can probably keep the dependent dialects for now, if we delete the entire definition of the pass in the next step anyway.

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
sommerlukas: What happens with operations that remain unvectorized because their vectorization failed? Will they be handled by some alternative path or will they just fail later in the pipeline?

hanhanW: It depends on pipeline setup, but the basic lowering should handle it. They should be lowered to scalar codes. There may be bufferization (like large stack buffer or shared memory) though.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.td
hanhanW: It's already migrated and the GenericVectorization pass can handle it: https://github.com/iree-org/iree/commit/28ee31a5069c1d64ef0a1ad35f656747880579df

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
sommerlukas: I find it difficult to say what the best strategy is here. I agree with your point that failing the pass here could mean that upstream changes/bugs cause instability. On the other hand, silently falling back to scalar code could either be inefficient (slow, which we would only catch in benchmarks, potentially with some delay that requires bisecting) or mis-compilation later on in the pipeline (stack buffer too large etc.). 

The right choice probably depends on how bad the fallback can become. As there is an existing bug, maybe we can stick with this approach for now and add a TODO to add the pass failure later.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.td
sommerlukas: Ok, so the deletion of the pass is imminent, should be fine then.

sommerlukas (approved):

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
hanhanW: We can add an option to that escalates silent failure to an actual failure for such case, so it may be easier to discover when we see bnehcmark regressions. This will require more work because there may be other similar "bugs" when we look at all the pipelines.

https://github.com/iree-org/iree/blob/ccebc03048180f3d4b409071add6431d12c3d6b1/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.h#L14-L29

PR #23717, opened 2026-03-10
[GPUHeuristics] Refactor MMA heuristic seeds to be architecture-specific

yzhang93:
Each GPU architecture now defines a constexpr ArchSeedSet containing gemm, scaled gemm, and convolution seed arrays indexed by GemmSize. The architecture is determined from the target and the corresponding seed set is returned (e.g. RDNA4 uses tuned seeds from RX 9070 XT benchmarking, others use the default). GemmSize enum values are changed to start at 0 for direct array indexing.

Add RDNA4 (gfx1201) config tests covering small, medium, and large matmul and convolution shapes.

RDNA4 benchmark results (RX 9070 XT, vs default seeds):
- GEMM shapes: ~40% of shapes improved, 15.9% geometric mean speedup.
- Prod Conv shapes: ~25% of shapes improved, 6% geometric mean speedup.
- Proxy Conv shapes: ~30% of shapes improved, 14.5% geometric mean speedup.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
@@ -10,7 +10,13 @@
 
 namespace mlir::iree_compiler {
 
-enum class GemmSize { NotSet, SmallGemm, MediumGemm, LargeGemm, VeryLargeGemm };
+enum class GemmSize : int {
+  NotSet = -1,
+  SmallGemm = 0,
+  MediumGemm = 1,
+  LargeGemm = 2,
+  VeryLargeGemm = 3,
+};
 
 llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const GemmSize &gemmSize);
 
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
@@ -5,20 +5,22 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include <cstdint>
+#include <optional>
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUInterfaces.h"
 #include "mlir/IR/Types.h"
 
 namespace mlir::iree_compiler {
 
-enum class GemmSize : int {
-  NotSet = -1,
-  SmallGemm = 0,
-  MediumGemm = 1,
-  LargeGemm = 2,
-  VeryLargeGemm = 3,
+enum class GemmSizeKind : int {
+  SmallGemm,
+  MediumGemm,
+  LargeGemm,
+  VeryLargeGemm,
+  Count, // Must be last — used for static array sizes.
 };
 
-llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const GemmSize &gemmSize);
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                              const GemmSizeKind &gemmSize);
 
 /// Struct containing information about a matmul's shape and type.
 struct GPUMatmulShapeType {
@@ -33,7 +35,7 @@ struct GPUMatmulShapeType {
   Type aScaleType;
   Type bScaleType;
 
-  GemmSize gemmSize = GemmSize::NotSet;
+  std::optional<GemmSizeKind> gemmSize;
 
   // Number of horizontally fused operations.
   // Horizontal fusion: C1,C2 = fused_matmul(A, B1, B2) where A is shared.
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
@@ -4,6 +4,9 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#ifndef IREE_COMPILER_CODEGEN_COMMON_GPU_GPUHEURISTICS_H_
+#define IREE_COMPILER_CODEGEN_COMMON_GPU_GPUHEURISTICS_H_
+
 #include <cstdint>
 #include <optional>
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUInterfaces.h"
@@ -182,3 +185,5 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
                               const GPUMMASchedule &schedule);
 
 } // namespace mlir::iree_compiler
+
+#endif // IREE_COMPILER_CODEGEN_COMMON_GPU_GPUHEURISTICS_H_
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -258,89 +258,110 @@ static GemmCutoff computeGemmCutoffsForAI(IREE::GPU::TargetAttr target,
   return {smallGemmCutoff, largeGemmCutoff, veryLargeGemmCutoff};
 }
 
-static std::optional<GPUMMAHeuristicSeeds>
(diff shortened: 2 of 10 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -258,89 +258,110 @@ static GemmCutoff computeGemmCutoffsForAI(IREE::GPU::TargetAttr target,
...
-  if (isGemm) {
-    return getGemmHeuristicSeeds(gemmSize, inBitWidth, scaled);
-  }
-  return getConvolutionHeuristicSeeds(gemmSize, inBitWidth);
+  const ArchSeedSet &archSeeds = getArchSeedSet(target);
+  assert(problem.gemmSize != GemmSize::NotSet && "GemmSize must be set");
+
+  // Pick the right category, index by GemmSize, then convert the stored
+  // K-element bits to an actual element count.
+  const auto *table = !isGemm
```
Yu-Zhewen:
```suggestion
  const GPUMMAHeuristicSeeds *table = !isGemm
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -258,89 +258,110 @@ static GemmCutoff computeGemmCutoffsForAI(IREE::GPU::TargetAttr target,
...
+    },
+};
+
+// clang-format on
+
+/// Look up the seed set for the given target architecture.
+static const ArchSeedSet &getArchSeedSet(IREE::GPU::TargetAttr target) {
+  if (target) {
+    StringRef arch = target.getArch();
+    if (arch == "rdna4" || arch.starts_with("gfx120")) {
```
Yu-Zhewen: Optional: consider using `amdgpu::Chipset::parse` + major version check ([example](https://github.com/iree-org/iree/blob/c3ed12e4203bed18674ed220f401f5da3173e51a/compiler/plugins/target/ROCM/ROCMTarget.cpp#L735)) instead of string prefix matching, in case we have gfx12001 in the future :)

Yu-Zhewen (approved):
LGTM, thanks!

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -10,7 +10,13 @@
 
 namespace mlir::iree_compiler {
 
-enum class GemmSize { NotSet, SmallGemm, MediumGemm, LargeGemm, VeryLargeGemm };
+enum class GemmSize : int {
```
lialan: Since you are using enum numbers for array index in subsequent code, do not use negative values in `NotSet`.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
kuhar: +1, I don't think this should be optional

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -10,7 +10,13 @@
...
+  NotSet = -1,
+  SmallGemm = 0,
+  MediumGemm = 1,
+  LargeGemm = 2,
+  VeryLargeGemm = 3,
```
kuhar: The usual pattern for indexed enums is to have an entry for `Last` or `Count` -- then you can use that for static array sizes. llvm also provides `enum_seq` for iterating over enums, if you end up having to do that in the future.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
kuhar: Could we call this something less generic like `GemmSizeKind`?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -258,89 +258,110 @@ static GemmCutoff computeGemmCutoffsForAI(IREE::GPU::TargetAttr target,
...
-         /*bestKElementCountPerSubgroup=*/kCacheLineSizeBits / 2 / inBitWidth});
-  default:
-    assert(false && "Unhandled convolution gemm size");
-    return std::nullopt;
+/// Complete seed set for a GPU architecture, with separate tables for
+/// gemm, scaled gemm, and convolution — each indexed by GemmSize.
+struct ArchSeedSet {
+  GPUMMAHeuristicSeeds gemm[4];
+  GPUMMAHeuristicSeeds scaledGemm[4];
+  GPUMMAHeuristicSeeds conv[4];
```
```suggestion
  static constexpr int numKinds = static_cast<int>(GemmSize::Count);
  GPUMMAHeuristicSeeds gemm[numKinds] = {};
  GPUMMAHeuristicSeeds scaledGemm[numKinds] = {};
  GPUMMAHeuristicSeeds conv[numKinds] = {};
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -258,89 +258,110 @@ static GemmCutoff computeGemmCutoffsForAI(IREE::GPU::TargetAttr target,
...
+    },
+    /*conv=*/{
+        /*SmallGemm=*/     {2, 2,  4, kCacheLineSizeBits},
+        /*MediumGemm=*/    {8, 4,  4, 2 * kCacheLineSizeBits},
+        /*LargeGemm=*/     {8, 8,  2, kCacheLineSizeBits / 2},
+        /*VeryLargeGemm=*/ {8, 8,  2, kCacheLineSizeBits / 2},
+    },
+};
+
+/// RDNA4 seeds (tuned based on RX 9070 XT benchmarking data).
```
kuhar: I guess the question is whether this should be arch or SKU-specific. With CDNA3, we saw that different SKUs like MI300X vs. MI308 needed different heuristics. We have that supported for default tuning specs.

kuhar (commented):
Encoding target-specific features in generic compiler code is an anti-pattern... Is there any way to move it to the rocm target?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
MaheshRavishankar: Maybe, but it also depends on what we have tried. I think this is something that we will have to evolve and keep in mind to manage the different SKUs. But without measuring we wont know what these should be.

MaheshRavishankar (approved):

yzhang93:
> Encoding target-specific features in generic compiler code is an anti-pattern... Is there any way to move it to the rocm target?

Good point. Moved `ArchSeedSet`, the seed tables, and `getArchSeedSet()` to `KnownTargets.cpp/.h`. Could you review it again? @kuhar

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
yzhang93: Done.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1252,4 +1253,74 @@ TargetAttr getFullTarget(StringRef targetAPI, StringRef aliasTarget,
...
+        /*LargeGemm=*/     {4, 8,  4, kCacheLineSizeBits},
+        /*VeryLargeGemm=*/ {4, 8,  4, kCacheLineSizeBits},
+    },
+};
+
+// clang-format on
+
+/// Look up the seed set for the given target architecture.
+const ArchSeedSet &getArchSeedSet(TargetAttr target) {
+  if (target) {
```
kuhar: nit: flip this and return early instead

kuhar (approved):
Thanks for the changes. I think this is a pragmatic compromise -- we can land in the current form and decide how to extend later when we learn what it takes to generalize it to other architectures / SKUs.

PR #23721, opened 2026-03-10
TileSwizzle: `distributionFactor`, `symbolicMultiplier`, serialization fix, documentation

bjacob:
A number of things in this PR so we don't spend the rest of the week landing a chain of PRs. It's still divided in two separate commits for review.
* `distributionSize` is replaced by `distributionFactor`, acting as a multiplier on the size.
  * `distributionSize` was never properly supported in serialization/deserialization and equality operators. Now it's supported and tested.
  * The new `distributionFactor` is only defined for CrossThread dims. Previously, `distributionSize` was set to 1 for other dims, and some code (`getSwizzledDistributionShape`) was relying on that.  That logic has been moved to that code, so now TileSwizzle itself has simpler semantics.
* A new `symbolicMultiplier` field is added, for Internal dimensions specifically.
  * It enables using TileSwizzle on scalable vector ISAs.
  * As it is mutually exclusive with `distributionFactor`, it coexists in a `union` with it.
* Brand new class documentation for TileSwizzle.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
@@ -325,8 +325,6 @@ getRowMajorTilesMNKShape(MMAIntrinsic intrinsic) {
 Codegen::TileSwizzle getIntrinsicSwizzle(IREE::CPU::MMAIntrinsic mma,
                                          int operandIdx) {
   using TileSwizzle = Codegen::TileSwizzle;
-  using Kind = TileSwizzle::Dim::Kind;
-
   auto maybeMnkTuple = getRowMajorTilesMNKShape(mma);
   if (!maybeMnkTuple) {
     // Whenever one adds support for a new intrinsic that doesn't have a
@@ -339,7 +337,7 @@ Codegen::TileSwizzle getIntrinsicSwizzle(IREE::CPU::MMAIntrinsic mma,
   swizzle.expandShape().resize(2);
   auto expandIfNonUnit = [](TileSwizzle &swizzle, int dim, int size) {
     if (size > 1) {
-      Codegen::expand(swizzle, dim, TileSwizzle::Dim{Kind::Internal, size});
+      Codegen::expand(swizzle, dim, TileSwizzle::Dim::internal(size));
     }
   };
 
@@ -362,11 +360,13 @@ Codegen::TileSwizzle getIntrinsicSwizzle(IREE::CPU::MMAIntrinsic mma,
 Codegen::TileSwizzle getSwizzle(IREE::CPU::DataTiledMMAAttr mma,
                                 int operandIdx) {
   using TileSwizzle = Codegen::TileSwizzle;
-  using Kind = TileSwizzle::Dim::Kind;
   TileSwizzle swizzle = getIntrinsicSwizzle(mma.getIntrinsic(), operandIdx);
-  TileSwizzle::Dim intrinsicsM = {Kind::CrossIntrinsic, mma.getIntrinsicsM()};
-  TileSwizzle::Dim intrinsicsN = {Kind::CrossIntrinsic, mma.getIntrinsicsN()};
-  TileSwizzle::Dim intrinsicsK = {Kind::CrossIntrinsic, mma.getIntrinsicsK()};
+  TileSwizzle::Dim intrinsicsM =
+      TileSwizzle::Dim::crossIntrinsic(mma.getIntrinsicsM());
+  TileSwizzle::Dim intrinsicsN =
+      TileSwizzle::Dim::crossIntrinsic(mma.getIntrinsicsN());
+  TileSwizzle::Dim intrinsicsK =
+      TileSwizzle::Dim::crossIntrinsic(mma.getIntrinsicsK());
   // LHS: (M, K); RHS: (K, N); Acc: (M, N).
   if (operandIdx == 0) {
     constexpr int M = 0, K = 1;
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
@@ -7,6 +7,7 @@
 #ifndef IREE_COMPILER_CODEGEN_DIALECT_CODEGEN_IR_IREECODEGENTYPES_H_
 #define IREE_COMPILER_CODEGEN_DIALECT_CODEGEN_IR_IREECODEGENTYPES_H_
 
+#include <cassert>
 #include <cstdint>
 
 #include "llvm/ADT/SmallVector.h"
@@ -58,36 +59,39 @@ public:
       CrossIntrinsic
     };
 
-    // Support constructing from any size type.
-    template <typename T>
-    Dim(Kind kind, T size) : kind_(kind), size_(size) {
-      if (kind_ == Kind::CrossThread) {
-        distributionSize_ = size;
-      }
+    Dim(Kind kind, int64_t size) : kind_(kind), size_(size) {}
+
+    static Dim crossThread(int64_t size, int64_t distributionFactor = 1) {
+      Dim dim(Kind::CrossThread, size);
+      dim.distributionFactor_ = distributionFactor;
+      return dim;
     }
-    template <typename T>
-    Dim(Kind kind, T size, T distributionSize)
-        : kind_(kind), size_(size), distributionSize_(distributionSize) {}
 
     Kind kind() const { return kind_; }
     int size() const { return size_; }
-    int distributionSize() const { return distributionSize_; }
+    int distributionFactor() const {
+      assert(kind() == Kind::CrossThread &&
+             "distributionFactor() only defined for CrossThread dims");
+      return distributionFactor_;
+    }
 
   private:
     Kind kind_ = Kind::Internal;
 
-    // The size of the dimension.
-    int16_t size_ = 0;
+    // The distribution multiplier on the size of the dimension, for
+    // distribution purposes. This applies only to CrossThread dimensions and
+    // describes the situation where multiple threads see the same data.
+    // `distributionFactor_` is the number of threads sharing the same data. In
+    // that case, the size of the dimension becomes  `distributionFactor_` times
+    // larger for distribution purposes. `distributionFactor_` consecutive
+    // positions along the dimension are the same data, seen by
+    // `distributionFactor_` threads.
+    int8_t distributionFactor_ = 1;
 
-    // The size of the dimension for distribution. This is used for CrossThread
-    // dimensions, because we may want to distribute more than `size` threads to
-    // this dimension. The `distributionSize` is expected to be greater than or
-    // equal to `size`, and the mapping of the delinearized (by the distribution
-    // sizes) thread ID index to the offset into the Dim is
-    // `delinearized_tid / (distributionSize / size)`. The `distributionSize`
-    // for non-CrossThread dimensions should always be 1, since there is no
-    // distribution for these dimensions.
-    int16_t distributionSize_ = 1;
+    // The size of the dimension.
+    // For CrossThread dimensions, this may get multiplied by
+    // distributionFactor.
+    int16_t size_ = 0;
   };
 
   using ExpandShapeDimVectorType = SmallVector<Dim, 4>;
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
@@ -104,12 +104,9 @@ public:
       // introduce a separate enum value for each streaming mode.
       ArmSveVLIn128bitUnits,
       // The multiplier is the VLEN parameter in the RISC-V ISA, expressed in
-      // multiples of 128 bits. This is just a placeholder for future use. I
-      // have no idea if 128 bits is the right granularity for this. Note
-      // however that we can only multiply, not divide, so the unit better not
-      // be too small. If no one unit satisfies all use cases, we can introduce
-      // separate enum values for different units.
-      RiscvVlenIn128bitUnits
+      // multiples of 64 bits. If no one unit satisfies all use cases, we can
+      // introduce a separate enum value for each unit.
+      RiscvVlenIn64bitUnits
     };
 
     //
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
@@ -96,14 +96,13 @@ public:
       // The multiplier is the constant value 1. This is the common case, used
       // for all but the scalable dimensions.
       One,
-      // The multiplier is the `vscale` parameter of the Arm ISA. By definition,
-      // it is vector length divided by 128 bits, e.g. vscale=2 means 256 bits.
-      // Note that with SME, the value of vscale depends on the streaming mode.
+      // The SVE vector length in 128-bit units.
+      // Note that with SME, this value depends on the streaming mode.
       // The current semantics is that we just allow that dependenced on the
       // streaming mode to exist. Whenever we get to implementing data-tiling
       // with SME, we will find out if this works in practice or if we need to
       // introduce a separate enum value for each streaming mode.
-      ArmVscale,
+      ArmSveVLIn128bitUnits,
       // The multiplier is the VLEN parameter in the RISC-V ISA, expressed in
       // multiples of 128 bits. This is just a placeholder for future use. I
       // have no idea if 128 bits is the right granularity for this. Note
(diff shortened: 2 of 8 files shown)


compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
```diff
@@ -171,7 +211,47 @@ static std::optional<TileSwizzle::Dim> arrayAttrToSwizzleDim(Attribute attr) {
   if (!maybeKind) {
     return std::nullopt;
   }
-  return TileSwizzle::Dim(maybeKind.value(), sizeAttr.getInt());
+  const int64_t size = sizeAttr.getInt();
+  if (size <= 0 || size > std::numeric_limits<int16_t>::max()) {
+    return std::nullopt;
+  }
+  if (maybeKind.value() == TileSwizzle::Dim::Kind::Internal) {
```
hanhanW, line 218: Do we use switch-case here?

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
```diff
@@ -24,7 +24,7 @@ namespace mlir::iree_compiler::IREE::Codegen {
 //===----------------------------------------------------------------------===//
 
 bool operator==(TileSwizzle::Dim lhs, TileSwizzle::Dim rhs) {
-  return lhs.kind() == rhs.kind() && lhs.size() == rhs.size();
+  return !memcmp(&lhs, &rhs, sizeof lhs);
```
hanhanW, line 27: I learned it from Claude, and it makes sense to me. This may be wrong for two instance created by `crossIntrinsic(sameSize)`, because the union byte is indeterminate.

```cpp
  static Dim crossIntrinsic(int64_t size) {
    return Dim(Kind::CrossIntrinsic, size);  // union byte is indeterminate
  }
```

The private constructor `Dim(Kind kind, int64_t size) : kind_(kind), size_(size) {}` leaves the union (byte 1 of 4) uninitialized. Two separately-constructed `CrossIntrinsic` dims with the same size can have different junk in the union byte, causing `memcmp` to return non-zero. This is also technically undefined behavior (reading indeterminate memory).

Claude suggests to fix it by adding a default member initializer to the union (see below), but I think it may be fragile. If future contributors add more fields without initialization, we'll silently hit this bug. How about explicitly check all the fields?

```cpp
  union {
    int8_t distributionFactor_ = 0;  // zero-initialize by default
    SymbolicMultiplier symbolicMultiplier_;
  };
```

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
```diff
@@ -171,7 +211,47 @@ static std::optional<TileSwizzle::Dim> arrayAttrToSwizzleDim(Attribute attr) {
...
+    if (attrs.size() > 2) {
+      auto symbolicMultiplierAttr = dyn_cast<StringAttr>(attrs[2]);
+      if (!symbolicMultiplierAttr) {
+        return std::nullopt;
+      }
+      if (auto maybeSymbolicMultiplier = convertStringToSymbolicMultiplier(
+              symbolicMultiplierAttr.getValue())) {
+        if (!maybeSymbolicMultiplier) {
+          return std::nullopt;
+        }
```
hanhanW: This is deadcode because it's checked in the previous if statement.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/unittests/UtilsTest.cpp
```diff
@@ -39,24 +41,52 @@ TEST(TileSwizzle, DimKindToString) {
...
+      convertStringToSymbolicMultiplier("RiscvVlenIn128bitUnits");
+  EXPECT_THAT(maybeSymbolicMultiplier,
+              Optional(SymbolicMultiplier::RiscvVlenIn128bitUnits));
+  maybeSymbolicMultiplier = convertStringToSymbolicMultiplier("deadbeef");
+  EXPECT_FALSE(maybeSymbolicMultiplier.has_value());
+}
+
+TEST(TileSwizzle, SerializationDeserialization) {
+  using Dim = Dim;
+  using SymbolicMultiplier = SymbolicMultiplier;
```
hanhanW: Do we remove them? We already have the alias at namespace scope (lines 19-21).

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
```diff
@@ -58,36 +88,92 @@ class TileSwizzle {
...
+      // with SME, we will find out  if this works in practice or if we need to
+      // introduce a separate enum value for each streaming mode.
+      ArmVscale,
+      // The multiplier is the VLEN parameter in the RISC-V ISA, expressed in
+      // multiples of 128 bits. This is just a placeholder for future use. I
+      // have no idea if 128 bits is the right granularity for this. Note
+      // however that  we can only multiply, not divide, so the unit better not
+      // be too small. If no one unit satisfies all use cases, we can introduce
+      // separate enum values for different units.
+      RiscvVlenIn128bitUnits
```
hanhanW: nit: remove double whitespaces.

```suggestion
    enum class SymbolicMultiplier : int8_t {
      // The multiplier is the constant value 1. This is the common case, used
      // for all but the scalable dimensions.
      One,
      // The multiplier is the `vscale` parameter of the Arm ISA. By definition,
      // it is vector length divided by 128 bits, e.g. vscale=2 means 256 bits.
      // Note that with SME, the value of vscale depends on the streaming mode.
      // The current semantics is that we just allow that dependenced on the
      // streaming mode to exist. Whenever we get to implementing data-tiling
      // with SME, we will find out if this works in practice or if we need to
      // introduce a separate enum value for each streaming mode.
      ArmVscale,
      // The multiplier is the VLEN parameter in the RISC-V ISA, expressed in
      // multiples of 128 bits. This is just a placeholder for future use. I
      // have no idea if 128 bits is the right granularity for this. Note
      // however that we can only multiply, not divide, so the unit better not
      // be too small. If no one unit satisfies all use cases, we can introduce
      // separate enum values for different units.
      RiscvVlenIn128bitUnits
```

compiler/src/iree/compiler/Codegen/ExternalInterfaces/Utils.cpp
```diff
@@ -177,8 +177,20 @@ void adjustTileSizesForBitcast(RankedTensorType type,
     int64_t scaledExpandSize = innermostDim.size() * originalBits;
     assert(scaledExpandSize % storageBits == 0 &&
            "scaled expand size must be divisible by storage bits");
-    innermostExpandDims.back() = Codegen::TileSwizzle::Dim(
-        innermostDim.kind(), scaledExpandSize / storageBits);
+    int64_t newSize = scaledExpandSize / storageBits;
+    switch (innermostDim.kind()) {
+    case Codegen::TileSwizzle::Dim::Kind::Internal:
+      innermostExpandDims.back() = Codegen::TileSwizzle::Dim::internal(newSize);
```
hanhanW: optional: this path is not exercised today, but I wonder if we preserve the `symbolicMultiplier`.

(We can add a TODO or revisit it when scalable vectors are ready to this path.)

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
bjacob: I believe that switch-case is only preferable if each of the cases is short.  Here, they were becoming too heaving, started requiring their own {...} scopes to scope local variables, etc.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
bjacob: Ha, good point.  I will add the zero-initialization.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/Utils.cpp
bjacob: Good catch!!

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
```diff
@@ -70,6 +70,7 @@ class TileSwizzle {
       // one intrinsic on one thread. That dimension of size 4 is 'Internal'.
       Internal,
       // This dimension is internal to one intrinsic, but is across threads.
+      // By definition, this only happens on SIMT architectures (GPUs).
```
hanhanW, line 73: Thanks for adding this!

hanhanW (approved):
LGTM, thanks!

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
```diff
@@ -58,36 +89,97 @@ class TileSwizzle {
...
+      // The multiplier is the constant value 1. This is the common case, used
+      // for all but the scalable dimensions.
+      One,
+      // The multiplier is the `vscale` parameter of the Arm ISA. By definition,
+      // it is vector length divided by 128 bits, e.g. vscale=2 means 256 bits.
+      // Note that with SME, the value of vscale depends on the streaming mode.
+      // The current semantics is that we just allow that dependenced on the
+      // streaming mode to exist. Whenever we get to implementing data-tiling
+      // with SME, we will find out if this works in practice or if we need to
+      // introduce a separate enum value for each streaming mode.
```
banach-space: Thanks! All in all makes sense to me, but it might be a bit unclear to folks less familiar with these concepts. Below I am sharing some high-level comments. 

---
`vscale` isn't really a concept specific to Arm (in fact, there is no mention of `vscale` in Arm ISAs). Instead, it's a concept from LLVM to model scalable vectors in general:
* https://llvm.org/docs/LangRef.html#llvm-vscale-intrinsic

AFAIK, it is also used for RISC-V. As for Arm ISAs, this is pretty much all that the [docs](https://developer.arm.com/documentation/102476/0101/Introducing-SVE) say:

> SVE allows flexible vector length implementations with a range of possible values in CPU implementations. The vector length can vary from a minimum of 128 bits up to a maximum of 2048 bits, but **must be a power of two**. Valid vector length implementations are therefore 128, 256, 512, 1024, and 2048 bits.

In essence, SVE Vector Length is `N x 128bits`, where `N = {1, 2. 4. 8. 16}`. It just happened that in LLVM (and MLIR), `vscale` is used to model `N`.

---
As for SME, the streaming mode is introduced e.g. here: https://developer.arm.com/documentation/109974/0100/What-is-SME-SME2-?lang=en.

Hopefully this example will help clarify the difference between VL and Streaming VL.
```asm
// x0 = VL outside streaming mode, in bytes
// x1 = VL inside  streaming mode, in bytes

cntb    x0              // non-streaming SVE VL in bytes

smstart sm              // enter streaming mode
cntb    x1              // streaming SVE VL in bytes
smstop  sm              // leave streaming mode
```
(This should be corect, but I haven't run it to verify).

Documentation:
* [cntb](https://developer.arm.com/documentation/ddi0602/2025-12/SVE-Instructions/CNTB--CNTD--CNTH--CNTW--Set-scalar-to-multiple-of-predicate-constraint-element-count-?lang=en)
* [smstart](https://developer.arm.com/documentation/ddi0602/2025-12/Base-Instructions/SMSTART--Enables-access-to-Streaming-SVE-mode-and-SME-architectural-state--an-alias-of-MSR--immediate--?lang=en)
* [smstop](https://developer.arm.com/documentation/ddi0602/2025-12/Base-Instructions/SMSTOP--Disables-access-to-Streaming-SVE-mode-and-SME-architectural-state--an-alias-of-MSR--immediate--?lang=en)
---

Hopefully this is helpful. This can probably be summarised as: 
> Perhaps rename ArmVscale as ArmRuntimeVecMultiplier 

🤷🏻 :)

bjacob: Thanks for the comment! Pushed a renaming commit. Hope the new descriptive name is OK.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
```diff
@@ -58,36 +89,96 @@ class TileSwizzle {
...
+      ArmSveVLIn128bitUnits,
+      // The multiplier is the VLEN parameter in the RISC-V ISA, expressed in
+      // multiples of 128 bits. This is just a placeholder for future use. I
+      // have no idea if 128 bits is the right granularity for this. Note
+      // however that we can only multiply, not divide, so the unit better not
+      // be too small. If no one unit satisfies all use cases, we can introduce
+      // separate enum values for different units.
+      RiscvVlenIn128bitUnits
```
egebeysel: this would be in 64 bit units: https://llvm.org/docs/RISCV/RISCVVectorExtension.html#mapping-to-llvm-ir-types

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -339,7 +337,7 @@ Codegen::TileSwizzle getIntrinsicSwizzle(IREE::CPU::MMAIntrinsic mma,
   swizzle.expandShape().resize(2);
   auto expandIfNonUnit = [](TileSwizzle &swizzle, int dim, int size) {
     if (size > 1) {
-      Codegen::expand(swizzle, dim, TileSwizzle::Dim{Kind::Internal, size});
+      Codegen::expand(swizzle, dim, TileSwizzle::Dim::internal(size));
```
egebeysel, line 340: question: what if we have an intrinsic with a symbolic (scalable) multiplier here? Should that be queried in a similar fashion to `getRowMajorTilesMNKShape` and handled here as well? Or rather elsewhere?

egebeysel (approved):
LGTM % the RVV symbolic multiplier name. Thanks for taking the time and implementing this!

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
bjacob: Like the comment says, people can add more enum values as they need to. I understand that VLEN may be less than 128 bits, which could require adding such a new enum value, but the current one may still be needed, since a smaller unit means a higher numerical value and we can only represent integral values.

Example: suppose that RiscvVlenIn128bitUnits=2,  RiscvVlenIn64bitUnits=4. Suppose that in that case we want a size to be 6.  We can represent that as 3 * RiscvVlenIn128bitUnits, but if we only have RiscvVlenIn64bitUnits then we can't represent it.

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
bjacob: getRowMajorTilesMNKShape  is just a local utility function. Its purpose is to give a simple way of specifying simple, non-scalable cases.  If someone populates scalable cases in the future, they would be free to design their own local helpers for that. That might take the form of a scalable variant of getRowMajorTilesMNKShape   or it could be completely different.  As long as `getIntrinsicSwizzle` returns the expected TileSwizzle.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h
egebeysel: I'm not sure I understand the comment, this multiplier is supposed to just symbolize the multiplier of the base vector size to get the hardware (runtime) vector length. In the case of SVE, it is defined as (as in, if you lower `vector.vscale` operation to [CNT* instructions](https://developer.arm.com/documentation/ddi0602/2022-09/SVE-Instructions/CNTB--CNTD--CNTH--CNTW--Set-scalar-to-multiple-of-predicate-constraint-element-count-) ) the actual vector length in bits divided by 128 bits, all the time. Similar applies to RVV, this gives us the actual hardware vector length divided by 64 bits.

egebeysel: In this PR, this is merely a naming issue, but when we implement actual logic to materialize these on top of it, and if I want, say enough FP32 elements to fit in 4 vector registers on one dimension, the number of FP32 elements always corresponds to `(#VL * #vector_registers) / 32 bits = ((64 bits * vscale) * 4) / 32 bits = 8 * vscale` for RVV. That means this dimension would be represented as `TileSwizzle::Dim::internal(8, SymbolicMultiplier::RiscvVlenIn*64*bitUnits)`.

egebeysel: > In essence, SVE Vector Length is N x 128bits, where N = {1, 2. 4. 8. 16}. It just happened that in LLVM (and MLIR), vscale is used to model N.

Basically RVV Vector Length is N x 64bits, where N = {1*, 2, 4, ..., 1024} and vscale is used to model N.

bjacob: Thanks. Changed to 64 bit units.

PR #23725, opened 2026-03-10
Remove 64-operation batch size limit from io_uring submit.

benvanik:
The submit path used a uint64_t bitmask to track which batch positions contained software operations (semaphore signal/wait), with an assert capping batches at 64 entries. This caused debug-mode crashes for tests submitting larger batches (e.g. MessageTest.MultipleInFlight with 100 messages). In release builds the assert was compiled out and the shift UB happened to be benign for all-kernel-op batches, but batches with software ops at indices >= 64 would silently misidentify operations.

The bitmask was motivated by a concern that concurrent submitters could flush SQEs to the kernel, allowing fast-completing kernel ops to be recycled before Phase 3 reads their type. This race cannot occur: the operations array is caller-local and software ops are never submitted to the kernel, so their structs cannot be recycled by concurrent completion paths. Replace the bitmask with a direct iree_async_proactor_io_uring_is_software_op() type check in Phase 3.

### DIFF
--- a/runtime/src/iree/async/cts/socket/socket_benchmark.cc
+++ b/runtime/src/iree/async/cts/socket/socket_benchmark.cc
@@ -128,7 +128,7 @@ struct LoopbackContext {
 
   void ResetZCTracking() { zc_achieved_count = 0; }
 
-  // Polls until completed or timeout.
+  // Polls until the completed flag is set or timeout.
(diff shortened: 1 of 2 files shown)


kuhar (approved):
This fixed the assertion failures for me

PR #23734, opened 2026-03-11
[LLVMCPU][RISCV] Add RVV int8 vector-contract custom kernels for widening i8*i8->i32 matmul

### LINKED ISSUE #23730: [RISCV] Add RVV int8 custom vector-contract kernels for widening i8*i8->i32 matmul
### Request description

Scope:
- add RVV VLEN-aware inner-tile selection for widening i8*i8->i32 matmul
- add RVV vector-size selection in LLVMCPU/KernelDispatch
- add RVV int8 custom vector-contract kernels in LLVMCPU/VectorContractCustomKernels

Out of scope:
- vendor-specific extensions (e.g. xsmtvdot / SpacemiT)
- FP8 / FP16 / BF16 custom-kernel changes
- convolution changes
- K>1 / loop-unroll variants

Motivation:
- enable RVV int8 widening matmul lowering through the existing custom-kernel path
- keep the implementation aligned with the existing RVV fp32/f16 infrastructure

Compiler

### Additional context

_No response_

copparihollmann:
Fixes iree-org/iree#23730

This PR adds RVV custom vector-contract kernel support for widening i8*i8->i32 matmul on RISC-V.

It updates:
- CPU encoding tile enumeration for RVV int8->i32 matmul
- LLVMCPU kernel dispatch RVV vector sizes
- VectorContractCustomKernels with an RVV intrinsic-based int8 path
- lit tests for vector-contract custom kernels and mmt4d vector lowering

Something I would like some feedback on is what would be your preferred way to test this e2e. So far I compile it and run it on the Banana-pi SpacemiT X60.

Assisted-by: ChatGPT

### DIFF
--- a/runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_tiles.inl
+++ b/runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_tiles.inl
@@ -13,6 +13,12 @@ IREE_UK_MMT4D_TILE(riscv_64, f32, f32, f32, 2, 1, _v)
 IREE_UK_MMT4D_TILE(riscv_64, f32, f32, f32, 4, 1, _v)
 IREE_UK_MMT4D_TILE(riscv_64, f32, f32, f32, 7, 1, _v)
 
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 1, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 2, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 4, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 7, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 8, 1, _v)
+
 IREE_UK_MMT4D_TILE(riscv_64, f16, f16, f32, 1, 1, _zvfhmin)
 IREE_UK_MMT4D_TILE(riscv_64, f16, f16, f32, 2, 1, _zvfhmin)
 IREE_UK_MMT4D_TILE(riscv_64, f16, f16, f32, 4, 1, _zvfhmin)
--- a/runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_tiles.inl
+++ b/runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_tiles.inl
@@ -17,7 +17,6 @@ IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 1, 1, _v)
 IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 2, 1, _v)
 IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 4, 1, _v)
 IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 7, 1, _v)
-IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 8, 1, _v)
 
 IREE_UK_MMT4D_TILE(riscv_64, f16, f16, f32, 1, 1, _zvfhmin)
 IREE_UK_MMT4D_TILE(riscv_64, f16, f16, f32, 2, 1, _zvfhmin)
--- a/runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
+++ b/runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
@@ -10,23 +10,22 @@
 #include "iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_internal.h"
 
 IREE_UK_ATTRIBUTE_ALWAYS_INLINE static inline void
-iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v(
+iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_7xXXx1_riscv_64_v(
     void* IREE_UK_RESTRICT out_tile, const void* IREE_UK_RESTRICT lhs_panel,
     const void* IREE_UK_RESTRICT rhs_panel,
     const iree_uk_mmt4d_params_t* params, int M0) {
-  IREE_UK_ASSERT(M0 >= 1 && M0 <= 8);
+  IREE_UK_ASSERT(M0 >= 1 && M0 <= 7);
   const iree_uk_int8_t* IREE_UK_RESTRICT lhs_ptr = lhs_panel;
   const iree_uk_int8_t* IREE_UK_RESTRICT rhs_ptr = rhs_panel;
   iree_uk_int32_t* IREE_UK_RESTRICT out_ptr = out_tile;
 
-  vint32m4_t acc0, acc1, acc2, acc3, acc4, acc5, acc6, acc7;
+  vint32m4_t acc0, acc1, acc2, acc3, acc4, acc5, acc6;
   int N0 = params->N0;
   size_t vl = N0;
 
-#define IREE_UK_RISCV_64_VWACC_I8(ACC, LHS, RHS, VL)      \
-  do {                                                     \
-    vint16m2_t prod = __riscv_vwmul_vx_i16m2(RHS, LHS, VL); \
-    ACC = __riscv_vwadd_wv_i32m4(ACC, prod, VL);          \
+#define IREE_UK_RISCV_64_VWMACC_I8(ACC, LHS, RHS, VL)                    \
+  do {                                                                   \
+    ACC = __riscv_vwmacc_vx_i32m4(ACC, (iree_uk_int16_t)(LHS), RHS, VL); \
   } while (0)
 
   if (M0 == 1) {
@@ -37,9 +36,10 @@ iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v(
     }
     for (int k = 0; k < params->K; ++k) {
       vint8m1_t rhs = __riscv_vle8_v_i8m1(rhs_ptr, vl);
+      vint16m2_t rhs_wide = __riscv_vsext_vf2_i16m2(rhs, vl);
       rhs_ptr += N0;
       iree_uk_int8_t lhs = *lhs_ptr++;
-      IREE_UK_RISCV_64_VWACC_I8(acc0, lhs, rhs, vl);
+      IREE_UK_RISCV_64_VWMACC_I8(acc0, lhs, rhs_wide, vl);
     }
     __riscv_vse32_v_i32m4(out_ptr, acc0, vl);
   } else if (M0 == 2) {
@@ -52,11 +52,12 @@ iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v(
     }
     for (int k = 0; k < params->K; ++k) {
       vint8m1_t rhs = __riscv_vle8_v_i8m1(rhs_ptr, vl);
+      vint16m2_t rhs_wide = __riscv_vsext_vf2_i16m2(rhs, vl);
       rhs_ptr += N0;
       iree_uk_int8_t lhs0 = *lhs_ptr++;
       iree_uk_int8_t lhs1 = *lhs_ptr++;
-      IREE_UK_RISCV_64_VWACC_I8(acc0, lhs0, rhs, vl);
-      IREE_UK_RISCV_64_VWACC_I8(acc1, lhs1, rhs, vl);
+      IREE_UK_RISCV_64_VWMACC_I8(acc0, lhs0, rhs_wide, vl);
+      IREE_UK_RISCV_64_VWMACC_I8(acc1, lhs1, rhs_wide, vl);
     }
     __riscv_vse32_v_i32m4(out_ptr, acc0, vl);
     __riscv_vse32_v_i32m4(out_ptr + N0, acc1, vl);
@@ -74,15 +75,16 @@ iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v(
     }
     for (int k = 0; k < params->K; ++k) {
       vint8m1_t rhs = __riscv_vle8_v_i8m1(rhs_ptr, vl);
+      vint16m2_t rhs_wide = __riscv_vsext_vf2_i16m2(rhs, vl);
       rhs_ptr += N0;
       iree_uk_int8_t lhs0 = *lhs_ptr++;
       iree_uk_int8_t lhs1 = *lhs_ptr++;
       iree_uk_int8_t lhs2 = *lhs_ptr++;
       iree_uk_int8_t lhs3 = *lhs_ptr++;
-      IREE_UK_RISCV_64_VWACC_I8(acc0, lhs0, rhs, vl);
-      IREE_UK_RISCV_64_VWACC_I8(acc1, lhs1, rhs, vl);
-      IREE_UK_RISCV_64_VWACC_I8(acc2, lhs2, rhs, vl);
-      IREE_UK_RISCV_64_VWACC_I8(acc3, lhs3, rhs, vl);
+      IREE_UK_RISCV_64_VWMACC_I8(acc0, lhs0, rhs_wide, vl);
+      IREE_UK_RISCV_64_VWMACC_I8(acc1, lhs1, rhs_wide, vl);
+      IREE_UK_RISCV_64_VWMACC_I8(acc2, lhs2, rhs_wide, vl);
+      IREE_UK_RISCV_64_VWMACC_I8(acc3, lhs3, rhs_wide, vl);
     }
     __riscv_vse32_v_i32m4(out_ptr, acc0, vl);
     __riscv_vse32_v_i32m4(out_ptr + N0, acc1, vl);
@@ -108,6 +110,7 @@ iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v(
     }
     for (int k = 0; k < params->K; ++k) {
       vint8m1_t rhs = __riscv_vle8_v_i8m1(rhs_ptr, vl);
+      vint16m2_t rhs_wide = __riscv_vsext_vf2_i16m2(rhs, vl);
       rhs_ptr += N0;
       iree_uk_int8_t lhs0 = *lhs_ptr++;
       iree_uk_int8_t lhs1 = *lhs_ptr++;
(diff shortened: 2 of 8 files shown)


bjacob:
Thanks for the PR. Just some context as I will be OOO for a week (March 12-22).

VectorContractCustomKernels is the most ancient form of microkernels, dating back to 2021. Handling only a vector.contract op and not a whole loop, it is inherently limited in what it can do. In particular, it has to operate on the linear layout vectors, it has nowhere to hide layout changes and that tends to severely limit its ability to achieve peak performance outside of simple cases. This is mentioned here as the original motivation for whole-loop microkernels (which became the current mmt4d microkernels):
https://github.com/iree-org/iree/blob/00b956a50ce8192f65dc24abba57b7f506c75e57/compiler/src/iree/compiler/Codegen/LLVMCPU/VectorContractCustomKernels.cpp#L491-L517

It's only incidental that VectorContractCustomKernels  hasn't been removed yet.

But now the timing is that even mmt4d microkernels are becoming legacy as I'm trying to import more modern techniques that we developed on the GPU side, into the CPU side: https://hackmd.io/ti1_aQYbQK2rn0YUM1lFKQ

So I don't really want to tell you to do mmt4d microkernels instead now, as that's what I'm trying to make legacy.

So I'm OK with merging this, seeing that it's not very big. Just know that this is assuming that this is a one-off: we wouldn't want too much more of this.

Regarding testing:
* The current form of this PR already does the right thing in terms of lit-tests.
* Regarding e2e testing, this would be handled by `tests/e2e/matmul/`.  Edit the BUILD.bazel file there and run bazel_to_cmake.py.  This is how we create tests with specific CPU features enabled: https://github.com/iree-org/iree/blob/00b956a50ce8192f65dc24abba57b7f506c75e57/tests/e2e/matmul/BUILD.bazel#L138-L205

bjacob (approved):
Approving to unblock as I'll be OOO but:
* Wait for @egebeysel 's review on RISC-V aspects
* Read my other comment.
* e2e testing for this is optional.

copparihollmann:
Actually, based on what you pointed me to, I would be open to putting in the effort and taking inspiration from what you just linked. Ideally, I would like to follow your latest practices. 

I will have a read to what you just sent. I actually ended up here after what we discussed regarding the mmt4d ukernel and mlir ukernels, when you mentioned about moving things to the compiler and doing proper data tiling. 

We can still merge it if you think is good. But in any case, I will definitely do my best to move things to how you are trying to implement now things moving forward @bjacob

bjacob:
So the problem though is that we are in the early stages of a transition where the existing framework (mmt4d) will soon be legacy and the new framework (inner_tiled) isn't available (for CPU) just yet. That's why I'm OK if you just want to merge this. Up to you.

copparihollmann:
Okay I see, then in that case it would be lovely to merge it if possible :)

Is there any way to help on the inner tile approach?

Or parallelly while you design it start getting ready some RVV approach for it?

bjacob:
> Okay I see, then in that case it would be lovely to merge it if possible :)

Retryign the bazel build to see if the failure was temporary.

> 
> Is there any way to help on the inner tile approach?

Probably not at first as I need to lay the foundation. Yes in later stages I'll work with you and other stakeholders to port your existing kernels to the new framework.

> 
> Or parallelly while you design it start getting ready some RVV approach for it?

Yes, I'm talking with @egebeysel and others to ensure alignment.

copparihollmann:
I did a test on the Banana-Pi SpacemiTX60

For running this though I need to change the following:


[runtime/src/iree/base/internal/cpu.c‎](https://github.com/ucb-bar/iree/commit/d81ebc6a65068b1fa56839b61e09a061fbbcf847#diff-c6ed6e2ceb0a79b34bd6a2e55781b3f56a6c601d466de9d641a691069a45781f)
```
+21-10Lines changed: 21 additions & 10 deletions
Original file line number	Diff line number	Diff line change
@@ -381,6 +381,7 @@ static void iree_cpu_initialize_from_platform_x86_64(uint64_t* out_fields) {
#include <sched.h>
#include <sys/syscall.h>
#include <unistd.h>
#include <sys/auxv.h>
#ifdef __has_include
#if __has_include(<asm/hwprobe.h>)
#include <asm/hwprobe.h>
@@ -418,17 +419,27 @@ long riscv_hwprobe(struct riscv_hwprobe* pairs, size_t pair_count,

static void iree_cpu_initialize_from_platform_riscv_64(uint64_t* out_fields) {
  long rc = riscv_hwprobe(&kv_pairs[0], 1, 0, NULL, 0);
  if (rc != 0) {
    fprintf(stderr, "riscv_hwprobe syscall failed");
    exit(1);
  if (rc == 0) {
    unsigned long long hwprobe = kv_pairs[0].value;
    IREE_COPY_BITS(out_fields[0], IREE_CPU_DATA0_RISCV_64_V, hwprobe,
                  IREE_RISCV_HWPROBE_IMA_V);
    IREE_COPY_BITS(out_fields[0], IREE_CPU_DATA0_RISCV_64_ZVFHMIN, hwprobe,
                  IREE_RISCV_HWPROBE_EXT_ZVFHMIN);
    IREE_COPY_BITS(out_fields[0], IREE_CPU_DATA0_RISCV_64_ZVFH, hwprobe,
                  IREE_RISCV_HWPROBE_EXT_ZVFH);
    return;
  }
  
  unsigned long hwcap = getauxval(AT_HWCAP);
  if (hwcap != 0) {
    unsigned long hwcap_v = (1 << ('V' - 'A'));
    IREE_COPY_BITS(out_fields[0], IREE_CPU_DATA0_RISCV_64_V, hwcap, hwcap_v);
    return;
  }
  unsigned long long hwprobe = kv_pairs[0].value;
  IREE_COPY_BITS(out_fields[0], IREE_CPU_DATA0_RISCV_64_V, hwprobe,
                 IREE_RISCV_HWPROBE_IMA_V);
  IREE_COPY_BITS(out_fields[0], IREE_CPU_DATA0_RISCV_64_ZVFHMIN, hwprobe,
                 IREE_RISCV_HWPROBE_EXT_ZVFHMIN);
  IREE_COPY_BITS(out_fields[0], IREE_CPU_DATA0_RISCV_64_ZVFH, hwprobe,
                 IREE_RISCV_HWPROBE_EXT_ZVFH);
  
  fprintf(stderr, "riscv_hwprobe syscall failed");
  fprintf(stderr, "Unified RISC-V CPU probe failed (no hwprobe, no HWCAP).\n");
}

#else
```

## Benchmark Results

Both paths compiled with `+m,+a,+f,+d,+c,+v,+zvl256b`. The only difference is data-tiling + ukernel enablement.

| Matmul Size | With mmt4d Ukernel | Without Ukernel | Speedup |
|---|---|---|---|
| 256x256 i8 | 4.43 ms | 4.41 ms | ~1.0x (parity) |
| 512x512 i8 | 8.82 ms | 35.3 ms | **4.0x faster** |

### Tile configuration

- **Tile shape:** 7x32x1 (M0=7, N0=VLEN/8=32, K0=1)
- **Widening chain:** i8(m1) → i16(m2) → i32(m4) using `vwmul.vx` + `vwadd.wv`
- **Register budget:** 7×LMUL4(acc) + LMUL2(product) + LMUL1(rhs) = 31/32 vector registers
- Matches the existing f32 RVV path (M0=7, N0=VLEN/8, LMUL=4)

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_tiles.inl
```diff
@@ -13,6 +13,12 @@ IREE_UK_MMT4D_TILE(riscv_64, f32, f32, f32, 2, 1, _v)
 IREE_UK_MMT4D_TILE(riscv_64, f32, f32, f32, 4, 1, _v)
 IREE_UK_MMT4D_TILE(riscv_64, f32, f32, f32, 7, 1, _v)
 
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 1, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 2, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 4, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 7, 1, _v)
+IREE_UK_MMT4D_TILE(riscv_64, s8, s8, s32, 8, 1, _v)
```
egebeysel: is this intentional?

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
```diff
@@ -0,0 +1,201 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <riscv_vector.h>
+
+#include "iree/builtins/ukernel/arch/riscv_64/common_riscv_64.h"
+#include "iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_internal.h"
+
+IREE_UK_ATTRIBUTE_ALWAYS_INLINE static inline void
+iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v(
```
egebeysel: just to be clear, you choose the inner tile size M to be 7 max, but implement the ukernel for 8 as well?

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
```diff
@@ -0,0 +1,201 @@
...
+      IREE_UK_RISCV_64_VWACC_I8(acc6, lhs6, rhs, vl);
+    }
+    __riscv_vse32_v_i32m4(out_ptr, acc0, vl);
+    __riscv_vse32_v_i32m4(out_ptr + N0, acc1, vl);
+    __riscv_vse32_v_i32m4(out_ptr + N0 * 2, acc2, vl);
+    __riscv_vse32_v_i32m4(out_ptr + N0 * 3, acc3, vl);
+    __riscv_vse32_v_i32m4(out_ptr + N0 * 4, acc4, vl);
+    __riscv_vse32_v_i32m4(out_ptr + N0 * 5, acc5, vl);
+    __riscv_vse32_v_i32m4(out_ptr + N0 * 6, acc6, vl);
+  } else if (M0 == 8) {
```
egebeysel: as previously mentioned, I guess in this case you wouldn't have enough registers and would spill (8 * 4 for acc and you're out for the prods and rhs), so if I'm not missing something, there's no need to keep this here. If this is intended and I'm missing something, please add the layout to the external models file, and a  comment explaining the kernel :)

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
```diff
@@ -0,0 +1,201 @@
...
+
+  vint32m4_t acc0, acc1, acc2, acc3, acc4, acc5, acc6, acc7;
+  int N0 = params->N0;
+  size_t vl = N0;
+
+#define IREE_UK_RISCV_64_VWACC_I8(ACC, LHS, RHS, VL)      \
+  do {                                                     \
+    vint16m2_t prod = __riscv_vwmul_vx_i16m2(RHS, LHS, VL); \
+    ACC = __riscv_vwadd_wv_i32m4(ACC, prod, VL);          \
+  } while (0)
```
egebeysel: Do we have to use a two-step multiply-add here? What if we sign-extend the operands to `i16` and use vwmacc `i16 x i16 -> i32` instead? How would that look like in means of performance? Is there a specific reason you do it this way? I think @adeel10x also pointed this out in Discord.

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
```diff
@@ -0,0 +1,201 @@
...
+    iree_uk_mmt4d_tile_s8s8s32_2xXXx1_riscv_64_v, 2)
+IREE_UK_MMT4D_TILE_FUNC_IMPL_FOR_M0(
+    iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v,
+    iree_uk_mmt4d_tile_s8s8s32_4xXXx1_riscv_64_v, 4)
+IREE_UK_MMT4D_TILE_FUNC_IMPL_FOR_M0(
+    iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v,
+    iree_uk_mmt4d_tile_s8s8s32_7xXXx1_riscv_64_v, 7)
+IREE_UK_MMT4D_TILE_FUNC_IMPL_FOR_M0(
+    iree_uk_mmt4d_tile_s8s8s32_1xXXx1_to_8xXXx1_riscv_64_v,
+    iree_uk_mmt4d_tile_s8s8s32_8xXXx1_riscv_64_v, 8)
```
egebeysel: same as above

runtime/src/iree/builtins/ukernel/arch/riscv_64/query_tile_sizes_riscv_64_entry_point.c
```diff
@@ -19,6 +19,18 @@ iree_uk_query_matmul_tile_sizes_riscv_64_f32f32f32(
   return (iree_uk_matmul_tile_sizes_t){.M = 8, .K = 1, .N = 8};
 }
 
+static iree_uk_matmul_tile_sizes_t
+iree_uk_query_matmul_tile_sizes_riscv_64_i8i8i32(
+    const iree_uk_query_tile_sizes_2d_params_t* params) {
+#if defined(IREE_UK_BUILD_RISCV_64_V)
+  if (iree_uk_cpu_riscv_64_v(params->cpu_data)) {
+    return (iree_uk_matmul_tile_sizes_t){.M = 7, .K = 1, .N = 32};
```
egebeysel, line 27: I'm not sure how important this actually is, but the N parameter passed here has the assumption that we have `LMUL=4` and `VLEN=256bits` here, whereas the above f32 one has `LMUL=4` but `VLEN=128bits`. Not sure if they have to be aligned or not...

egebeysel (commented):
Thanks for the updated patch!

Generally looks good to me, the main point is about the use of `vwmul + vwadd` vs `sext + vwmacc`. I think we could use some numbers there to judge :)

egebeysel:
cc @adeel10x

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_tiles.inl
bjacob: Indeed, the logic in compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp does not produce the case 8 so this tile function is likely unused.

runtime/src/iree/builtins/ukernel/arch/riscv_64/query_tile_sizes_riscv_64_entry_point.c
bjacob: query_tile_sizes is a historical thing, from an era when we considered supporting fully dynamic tile sizes with a runtime query based on the device CPU. hasn't actually been used.  So, I don't care much what you do here; I don't remember how much the test harness forces you to implement, but do stick to the bare minimum required to pass tests.

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_tiles.inl
copparihollmann: Good catch, this was from a previous test of mine in that patch that I forgot to remove. Will do so tomorrow!

copparihollmann: Also I will do an implementation with `sext + vwmacc` and get back to you on that :)

runtime/src/iree/builtins/ukernel/arch/riscv_64/query_tile_sizes_riscv_64_entry_point.c
copparihollmann: Okay then I will just focus on keeping it to a bare minimum to get the tests passing

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
copparihollmann: To be honest I focused on getting out the patch I had before so that I could get some feedback first an not overengineer it, but that is a very valid point that I forgot to test :)

Tomorrow I will get to it and get you some metrics back

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
copparihollmann: Will change it accordingly, you are definitely right there and it was something I got leftover

copparihollmann:
Will update tomorrow addressing your comments

copparihollmann:
I went through the review points:

- removed the stray `M0=8` tile registration and implementation path
- kept only `M0 = 1, 2, 4, 7`
- switched the final widening path to `vsext.vf2 + vwmacc.vx`

On Banana Pi SpacemiT X60:

- Arithmetic comparison on `matmul_q_i8_1024.mlir`:
  - `vwmul.vx + vwadd.wv`: `68.9 ms`
  - `vsext.vf2 + vwmacc.vx`: `42.8 ms`
- End-to-end `512x512` i8:
  - with ukernel: `5.61 ms`
  - without ukernel: `32.6 ms`

So the final implementation is both better inner sequence and better e2e :) 

Let me know if this is sufficient or I should change something

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
adeel10x: Yes, In my experience sign-extending the operands to i16 and using vwmacc i16 x i16 -> i32 instead of the 2-step multiply-add results in better perf. I see that you've already made the required changes :).

runtime/src/iree/builtins/ukernel/arch/riscv_64/mmt4d_riscv_64_v_i8.c
```diff
@@ -0,0 +1,153 @@
...
+  iree_uk_int32_t* IREE_UK_RESTRICT out_ptr = out_tile;
+
+  vint32m4_t acc0, acc1, acc2, acc3, acc4, acc5, acc6;
+  int N0 = params->N0;
+  size_t vl = N0;
+
+#define IREE_UK_RISCV_64_VWMACC_I8(ACC, LHS, RHS, VL)                    \
+  do {                                                                   \
+    ACC = __riscv_vwmacc_vx_i32m4(ACC, (iree_uk_int16_t)(LHS), RHS, VL); \
+  } while (0)
```
adeel10x: Is there a specific reason to use a macro here? Wouldn't it be a bit cleaner to use the vwmacc built-in directly in the ukernel?

copparihollmann: Fair enough, let me fix that :)

copparihollmann: Should be fixed now, sorry for the delay!

egebeysel:
Can you fix the DCO error? @copparihollmann 

Also, do you need [this patch](https://github.com/ucb-bar/iree/commit/d81ebc6a65068b1fa56839b61e09a061fbbcf847#diff-c6ed6e2ceb0a79b34bd6a2e55781b3f56a6c601d466de9d641a691069a45781f) to run this? If so, someone else has to review that.

Other than that, it looks good to me, thanks for updating this and testing :) I'll approve once the DCO is fixed and according to the above patch situation 👍 Would also be nice if you could update the title and the description.

If you have time to do so, you can also get some measurements with more regular/irregular shapes, e.g. pow2s from 128 to 4096 instead of just 512, and non-pow2s, with possibly wide M/N/K dims. But that's not necessary for the PR, you can defer that to later.

copparihollmann:
Hi all!

Sorry I was a bit by busy last week, I will get down with it toda/tomorrow :)

copparihollmann:
> Also, do you need [this patch](https://github.com/ucb-bar/iree/commit/d81ebc6a65068b1fa56839b61e09a061fbbcf847#diff-c6ed6e2ceb0a79b34bd6a2e55781b3f56a6c601d466de9d641a691069a45781f) to run this? If so, someone else has to review that.

The patch is only if you want to run it on the banana-pi because of the old linux distro

> If you have time to do so, you can also get some measurements with more regular/irregular shapes, e.g. pow2s from 128 to 4096 instead of just 512, and non-pow2s, with possibly wide M/N/K dims. But that's not necessary for the PR, you can defer that to later.

Working on this right now :)

egebeysel:
@copparihollmann Just a quick ping here. If you could fix the DCO issue and the CI failures we can land this PR as is prior to the more extensive measurements.

copparihollmann:
Done! 

Sorry for the delay, our board died (forgot to give a heads-up), so I was waiting for the new one to arrive and do the tests. 

DCO should be fixed now @egebeysel

egebeysel:
Merging this as is, will fix in a follow-up PR if anything pops up because of something that changed since the quite oldish CI run.

Issue #22119, opened 2025-09-26
Scaled matmul fails to compile when using global load DMA promotion


Max191:
https://github.com/iree-org/iree/pull/22094 turns off DMA for scaled matmul, because it causes compilation failures. We eventually want to turn it back on once the issue is resolved.

### Reproducing ###

1. Revert https://github.com/iree-org/iree/pull/22094
2. Run the scaled matmul test harness in https://github.com/nod-ai/shark-ai/issues/2252#issuecomment-3330094967
   - This seems to happen for K sizes greater than 128: `./run_and_compare.sh 128 128 256`

### Debug Notes ###

Here is the IR dump for the failure: https://gist.github.com/Max191/70a7e767952fc54198964d76acda8f1f

From a quick look, the first place I notice something wrong is after GPUFuseAndHoistParallelLoops:
```mlir
// -----// IR Dump After GPUFuseAndHoistParallelLoopsPass (iree-codegen-gpu-fuse-and-hoist-parallel-loops) //----- //
func.func @scaled_mfma_16x16x128_dispatch_0_reduction_128x128x256x32_f4E2M1FNxf4E2M1FNxf8E8M0FNUxf8E8M0FNUxf32() attributes {translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse workgroup_size = [256, 1, 1] subgroup_size = 64, {gpu_pipeline_options = #iree_gpu.pipeline_options<prefetch_shared_memory = true, no_reduce_shared_memory_bank_conflicts = true, use_igemm_convolution = false>}>} {
  %c32 = arith.constant 32 : index
  %c64 = arith.constant 64 : index
  %cst = arith.constant 0.000000e+00 : f32
  %c0 = arith.constant 0 : index
  %0 = hal.interface.binding.subspan layout(<bindings = [#hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, Indirect>], flags = Indirect>) binding(0) alignment(64) offset(%c0) flags("ReadOnly|Indirect") : memref<128x256x32xf4E2M1FN, #hal.descriptor_type<storage_buffer>>
[...]

PR #23736, opened 2026-03-11
[Codegen][GPU] Fix shared memory estimation for multi-buffering 

Yu-Zhewen:
- Pass `useDirectLoad` and `prefetchNumStages` through `calculateOperandsSharedMemoryUsedInBytes`, so that it can account
  for multi-buffering when using direct loads with prefetching.
- Properly guard direct load flag for scaled matmuls as forced off (#22119) to avoid inconsistency, by overriding the flag with a
  warning emitted.
- As such, for regular matmul the shared memory usage scales with `prefetchNumStages` when `useDirectLoad` is enabled. For scaled matmul, shared memory usage is unchanged since direct load is forced off.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -69,7 +69,8 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const GemmSize &gemmSize) {
 static int64_t calculateOperandsSharedMemoryUsedInBytes(
     const GPUMMASchedule &schedule, int64_t lhsBitwidth, int64_t rhsBitwidth,
     int64_t lhsScaleBitwidth = 0, int64_t rhsScaleBitwidth = 0,
-    int64_t numRhs = 1) {
+    int64_t numRhs = 1, bool useDirectLoad = false,
+    int64_t prefetchNumStages = 0) {
   int64_t tileM = schedule.getTotalMSize() * schedule.getTotalMTileSize() *
                   schedule.getTotalMSubgroupCount();
   int64_t tileN = schedule.getTotalNSize() * schedule.getTotalNTileSize() *
@@ -83,10 +84,17 @@ static int64_t calculateOperandsSharedMemoryUsedInBytes(
   int64_t tileKb = schedule.kSizes.back() * schedule.kTileSizes.back();
   int64_t tileKo = tileK / tileKb;
 
-  int64_t lhsSharedMemoryUsed = tileM * tileK * lhsBitwidth;
-  int64_t rhsSharedMemoryUsed = numRhs * tileN * tileK * rhsBitwidth;
-  int64_t aScaleSharedMemoryUsed = tileM * tileKo * lhsScaleBitwidth;
-  int64_t bScaleSharedMemoryUsed = numRhs * tileN * tileKo * rhsScaleBitwidth;
+  // Account for multi-buffering when using direct loads.
+  int64_t numBuffers =
+      (useDirectLoad && prefetchNumStages > 0) ? prefetchNumStages : 1;
+
+  int64_t lhsSharedMemoryUsed = numBuffers * tileM * tileK * lhsBitwidth;
+  int64_t rhsSharedMemoryUsed =
+      numBuffers * numRhs * tileN * tileK * rhsBitwidth;
+  int64_t aScaleSharedMemoryUsed =
+      numBuffers * tileM * tileKo * lhsScaleBitwidth;
+  int64_t bScaleSharedMemoryUsed =
+      numBuffers * numRhs * tileN * tileKo * rhsScaleBitwidth;
 
   return (lhsSharedMemoryUsed + rhsSharedMemoryUsed + aScaleSharedMemoryUsed +
           bScaleSharedMemoryUsed) /
@@ -740,7 +748,8 @@ FailureOr<GPUMMASchedule> deduceMMASchedule(
     const GPUMMAHeuristicSeeds &seeds, int64_t sharedMemLimitInBytes,
     int64_t subgroupSize, std::optional<int64_t> wgpCount, Location loc,
     bool transposedLhs, bool transposedRhs, bool canUpcastAcc,
-    bool mustBeAligned, bool doCPromotion, int64_t splitReductionTripCnt) {
+    bool useDirectLoad, int64_t prefetchNumStages, bool mustBeAligned,
+    bool doCPromotion, int64_t splitReductionTripCnt) {
 
   SmallVector<GPUIntrinsicType> sortedIntrinsics =
       sortMMAIntrinsics(problem, intrinsics);
@@ -776,7 +785,8 @@ FailureOr<GPUMMASchedule> deduceMMASchedule(
                              transposedLhs, transposedRhs);
       int64_t sharedMemoryUsed = calculateOperandsSharedMemoryUsedInBytes(
           schedule, lhsBitwidth, rhsBitwidth, lhsScaleBitwidth,
-          rhsScaleBitwidth, problem.numHorizontallyFusedOps);
+          rhsScaleBitwidth, problem.numHorizontallyFusedOps, useDirectLoad,
+          prefetchNumStages);
       // Add accumulator/result memory when it uses shared memory (LDS):
       // - Result needs padding in shared memory, OR
       // - matmul_accumulate loads accumulator from global memory via shared mem
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
@@ -84,21 +84,21 @@ static int64_t calculateOperandsSharedMemoryUsedInBytes(
   int64_t tileKb = schedule.kSizes.back() * schedule.kTileSizes.back();
   int64_t tileKo = tileK / tileKb;
 
-  // Account for multi-buffering when using direct loads.
-  int64_t numBuffers =
-      (useDirectLoad && prefetchNumStages > 0) ? prefetchNumStages : 1;
+  int64_t lhsSharedMemoryUsed = tileM * tileK * lhsBitwidth;
+  int64_t rhsSharedMemoryUsed = numRhs * tileN * tileK * rhsBitwidth;
+  int64_t aScaleSharedMemoryUsed = tileM * tileKo * lhsScaleBitwidth;
+  int64_t bScaleSharedMemoryUsed = numRhs * tileN * tileKo * rhsScaleBitwidth;
 
-  int64_t lhsSharedMemoryUsed = numBuffers * tileM * tileK * lhsBitwidth;
-  int64_t rhsSharedMemoryUsed =
-      numBuffers * numRhs * tileN * tileK * rhsBitwidth;
-  int64_t aScaleSharedMemoryUsed =
-      numBuffers * tileM * tileKo * lhsScaleBitwidth;
-  int64_t bScaleSharedMemoryUsed =
-      numBuffers * numRhs * tileN * tileKo * rhsScaleBitwidth;
+  int64_t totalBits = lhsSharedMemoryUsed + rhsSharedMemoryUsed +
+                      aScaleSharedMemoryUsed + bScaleSharedMemoryUsed;
 
-  return (lhsSharedMemoryUsed + rhsSharedMemoryUsed + aScaleSharedMemoryUsed +
-          bScaleSharedMemoryUsed) /
-         8;
+  // In direct load mode, ROCDLPrefetchSharedMemoryPass multi-buffers shared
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1192,12 +1206,10 @@ setMatmulLoweringConfig(IREE::GPU::TargetAttr target,
   LoweringConfigAttr loweringConfig = configAndWgSize->first;
 
   SmallVector<NamedAttribute, 1> pipelineAttrs;
-  // Default to 2 stages if not specified.
-  int64_t prefetchStages = prefetchNumStages.value_or(2);
   auto pipelineOptions = IREE::GPU::GPUPipelineOptionsAttr::get(
       linalgOp->getContext(),
       /*prefetchNumStages=*/prefetchStages,
-      /*no_reduce_shared_memory_bank_conflicts=*/useDirectLoad,
+      /*no_reduce_shared_memory_bank_conflicts=*/useDirectLoad || isScaled,
```
jerryyin, line 1212: This is not in previous diff, why do we need to change this?

kuhar (approved):

kuhar:
@bangtianliu @RattataKing we should mirror this in the tuner (eventually)

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -350,9 +350,9 @@ getContractionHeuristicSeeds(GPUMatmulShapeType problem, bool isGemm,
...
     IREE::GPU::TargetAttr target, GPUMatmulShapeType problem, Location loc,
-    bool transposedLhs, bool transposedRhs, bool isGemm,
-    bool mustBeAligned = true, bool doCPromotion = false, bool scaled = false,
-    int64_t splitReductionTripCnt = 0) {
+    bool transposedLhs, bool transposedRhs, bool isGemm, bool scaled,
+    bool useDirectLoad, int64_t prefetchNumStages, bool mustBeAligned = true,
+    bool doCPromotion = false, int64_t splitReductionTripCnt = 0) {
   const int64_t targetSubgroupSize = target.getPreferredSubgroupSize();
   SmallVector<GPUIntrinsicType> intrinsics;
   if (scaled) {
```
lialan, line 358: @Yu-Zhewen just a small reminder about this clang-tidy check by copilot. It does not show up in the comment list, but it is in the `Files changed` tab.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
```diff
@@ -45,6 +57,14 @@ func.func @scaled_matmul(
 // CHECK-REMARKS-SAME: Category:deduceMMASchedule
 // CHECK-REMARKS-SAME: Remark=34816
 
+// CHECK-REMARKS-DIRECT-LOAD-2: [Analysis] SharedMemoryUsage
+// CHECK-REMARKS-DIRECT-LOAD-2-SAME: Category:deduceMMASchedule
+// CHECK-REMARKS-DIRECT-LOAD-2-SAME: Remark=34816
+
+// CHECK-REMARKS-DIRECT-LOAD-3: [Analysis] SharedMemoryUsage
+// CHECK-REMARKS-DIRECT-LOAD-3-SAME: Category:deduceMMASchedule
+// CHECK-REMARKS-DIRECT-LOAD-3-SAME: Remark=34816
```
lialan, line 66: same value as 2, so either we don't need to check it or just keep 1.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
```diff
@@ -83,6 +103,14 @@ func.func @scaled_matmul_with_batch(
...
```
lialan, line 112: same here, find if we only need 1, or something is not right.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
```diff
@@ -316,3 +368,15 @@ func.func @matmul_f16_compute_bound(
...
+// CHECK-REMARKS-SAME: Category:deduceMMASchedule
+// CHECK-REMARKS-SAME: Remark=32768
+
+// CHECK-REMARKS-DIRECT-LOAD-2: [Analysis] SharedMemoryUsage
+// CHECK-REMARKS-DIRECT-LOAD-2-SAME: Category:deduceMMASchedule
+// CHECK-REMARKS-DIRECT-LOAD-2-SAME: Remark=65536
+
+// CHECK-REMARKS-DIRECT-LOAD-3: [Analysis] SharedMemoryUsage
+// CHECK-REMARKS-DIRECT-LOAD-3-SAME: Category:deduceMMASchedule
+// CHECK-REMARKS-DIRECT-LOAD-3-SAME: Remark=98304
```
lialan, line 382: oh oh okay ... this one is different, so maybe it is worth checking ...

Yu-Zhewen: Yeah, direct load is currently forced off for scaled matmul, so we can only observe the difference on regular matmul.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -83,10 +84,17 @@ static int64_t calculateOperandsSharedMemoryUsedInBytes(
...
+  int64_t numBuffers =
+      (useDirectLoad && prefetchNumStages > 0) ? prefetchNumStages : 1;
+
+  int64_t lhsSharedMemoryUsed = numBuffers * tileM * tileK * lhsBitwidth;
+  int64_t rhsSharedMemoryUsed =
+      numBuffers * numRhs * tileN * tileK * rhsBitwidth;
+  int64_t aScaleSharedMemoryUsed =
+      numBuffers * tileM * tileKo * lhsScaleBitwidth;
+  int64_t bScaleSharedMemoryUsed =
+      numBuffers * numRhs * tileN * tileKo * rhsScaleBitwidth;
```
jerryyin: I think it'd be cleaner if we just multiply numBuffers at the very end before return if we are scaling `numBuffers *` uniformly.

Also, add a explicit comment about the ROCDLPrefetchSharedMemoryPass that the current decisions of multiple buffering is the num of multi-buffer equals to numStages in direct load mode. 

Help me inspect another scenario too: At this point is the `useDirectLoad` a done deal or will it be overridden in some way later? I don't want us to be overly aggressive in our estimation if that's the case.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
```diff
@@ -9,6 +9,18 @@
...
+// RUN: --iree-codegen-llvmgpu-use-tile-and-fuse-matmul=true --iree-codegen-llvmgpu-test-tile-and-fuse-vectorize=true \
+// RUN: --iree-codegen-llvmgpu-use-igemm=false --iree-llvmgpu-use-direct-load=true --iree-llvmgpu-prefetch-num-stages=2 \
+// RUN: --pass-pipeline="builtin.module(iree-llvmgpu-select-lowering-strategy)" \
+// RUN: --remarks-filter=".*" %s 2>&1 | FileCheck %s --check-prefix=CHECK-REMARKS-DIRECT-LOAD-2
+
+// RUN: iree-opt --mlir-print-local-scope --split-input-file --iree-gpu-test-target=gfx950 \
+// RUN: --iree-codegen-llvmgpu-use-tile-and-fuse-matmul=true --iree-codegen-llvmgpu-test-tile-and-fuse-vectorize=true \
+// RUN: --iree-codegen-llvmgpu-use-igemm=false --iree-llvmgpu-use-direct-load=true --iree-llvmgpu-prefetch-num-stages=3 \
+// RUN: --pass-pipeline="builtin.module(iree-llvmgpu-select-lowering-strategy)" \
+// RUN: --remarks-filter=".*" %s 2>&1 | FileCheck %s --check-prefix=CHECK-REMARKS-DIRECT-LOAD-3
```
jerryyin, line 22: I see a lot of the `CHECK-REMARKS-DIRECT-LOAD-2` and 3 are duplicate of each other. You can add another common prefix like `CHECK-DIRECT-LOAD-ALL` for example that will check both when they are expected to be similar.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Yu-Zhewen: In the old code, `useDirectLoad` was wrongly set to `true` for scaled matmul to disable bank conflict reduction: 

https://github.com/iree-org/iree/blob/e002a09daff78de57af924c98926c7d0b2e5f912/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp#L1179-L1182

However, direct load is not actually supported for scaled matmul (#22119), as noted in the existing TODO:
https://github.com/iree-org/iree/blob/e002a09daff78de57af924c98926c7d0b2e5f912/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp#L959-L961

This PR fixes the inconsistency by overriding useDirectLoad = false for scaled matmul and using || isScaled here instead to preserve the original bank conflict reduction behavior.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
Yu-Zhewen: The only place useDirectLoad gets overridden is the scaled matmul case discussed above (@lialan is looking into this as an orthogonal issue). By the time we reach `calculateOperandsSharedMemoryUsedInByte`s, the flag reflects the final decision, so there's no risk of over-estimation.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_gfx950.mlir
Yu-Zhewen: Those identical values are still from scaled matmul tests where direct load is currently forced off (#22119). Once that issue is resolved and direct load is enabled, the values will diverge. I'd prefer to keep the separate prefixes since the fix is expected soon.

jerryyin (approved):
LGTM

lialan (approved):
+1

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -156,7 +156,8 @@ FailureOr<GPUMMASchedule> deduceMMASchedule(
     const GPUMMAHeuristicSeeds &seeds, int64_t sharedMemLimitInBytes,
     int64_t subgroupSize, std::optional<int64_t> cuCount, Location loc,
     bool transposedLhs = false, bool transposedRhs = false,
-    bool canUpcastAcc = false, bool mustBeAligned = true,
+    bool canUpcastAcc = false, bool useDirectLoad = false,
+    int64_t prefetchNumStages = 0, bool mustBeAligned = true,
```
Max191, line 160: Please update the function docs to explain what the new params do.

Also, I think the naming here is a bit confusing. The decision to use direct load is independent to the decision to use multi-buffering AFAIU. It just happens that the decision is linked today. Perhaps instead of these params we could just have a single int64_t param called `numMultiBufferingStages` (defaulting to 1), and then we just multiply by that value in the shared mem computation.

Max191 (commented):
Sorry, I just missed the merge, but could you address my comment as a followup?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
Yu-Zhewen: > Please update the function docs to explain what the new params do.

Thanks, will do.

> Also, I think the naming here is a bit confusing. The decision to use direct load is independent to the decision to use multi-buffering AFAIU. It just happens that the decision is linked today. Perhaps instead of these params we could just have a single int64_t param called numMultiBufferingStages (defaulting to 1), and then we just multiply by that value in the shared mem computation.

I'd prefer to keep the two separate parameters, as they correspond to two user-facing flags. Also, `prefetchNumStages` has different resource implications (multi-buffering via LDS or VGPRs) depending on the load mode.

PR #23738, opened 2026-03-11
[Integrate] Integrate llvm-project@ae920e2636d3

qedawkins:
Integrate up to
https://github.com/llvm/llvm-project/commit/ae920e2636d3

Changes:
- Bump third_party/llvm-project to ae920e2636d3 (234 commits)
- Fix DIDerivedTypeAttr::get calls after MLIR added file, line, scope parameters (e55805076e05)
- Update tests for arith.extui+uitofp canonicalization fold (5368b8163d50) which now folds extui+uitofp into direct uitofp, and removes the corresponding BitwiseAnd with 255 in SPIRV lowering


kuhar (approved):

kuhar:
Oh wait, the vulkan onnx failure looks like an actual issue

qedawkins:
Dang, so close. Taking a look

qedawkins:
Testing an upstream fix and if it passes I'll send it upstream and keep it as a cherry-pick.

qedawkins:
This is the upstream fix: https://github.com/llvm/llvm-project/pull/186136

I can either land this now (it has the change added locally) or wait for the above to make it in.

kuhar:
I'd land now, the next integrate can pick up the fix

PR #23746, opened 2026-03-12
[DataTiling] Adds data tiling support for 2D convolutions (on CPU targets)

jschuhmacher:
End-to-end data tiling for 2D convolutions following the structure in https://iree.dev/community/blog/2025-08-25-data-tiling-walkthrough. 

Encoding materialization (`getEncodingInfoForConv`) computes packing parameters (inner tile sizes, outerDimsPerm) for input, filter, and output tensors. Tiles along IC and OC dimensions using CPU-specific tile sizes (aarch64/x86_64).

All layouts are normalized into a canonical form (in line with https://hackmd.io/@phemashekar/conv-dt-layout):

- Input    [N, IC/16, H, W, ic=16]
- Filter   [IC/16, KH, KW, OC/16, ic=16, oc=16]
- Output   [N, OC/16, OH, OW, oc=16]

The `inferConvDimsFromMaps` helper should be replaced by a future upstream `inferConvolutionDims(ArrayRef<AffineMap>)` overload.

Limitations
 - Limited set of convolutions supported, grouped, depthwise, 1D, 3D, all remain to be done.
 - Only f32 support at the moment.
 - Tile sizes are hardcoded

Assisted-by: Claude Code


compiler/src/iree/compiler/DispatchCreation/AnnotateDataTilingHints.cpp
```diff
@@ -137,6 +137,35 @@ static bool isSupportedContractionOp(linalg::LinalgOp linalgOp) {
...
+  const bool isConv2D = cDims->outputImage.size() == 2;
+  if (!isConv2D) {
+    return false;
+  }
+
+  // Reject grouped convolutions (depth dims) until the packing/lowering
+  // supports them.
+  if (!cDims->depth.empty()) {
+    return false;
+  }
```
Max191, line 164: I think it would be better for this logic to be in the resolver. Then, if another downstream user wanted to implement their own resolver for the convolution encodings, they wouldn't have this constraint of only 2D non-grouped convs.

When the layout is resolved, the resolver can just pick the identity layout if the conv type isn't supported.

compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
```diff
@@ -597,3 +597,738 @@ func.func @matmul_lowering_i8i4i32_aarch64_i8mm(
...
+// CHECK-SAME:    %[[OUTPUT:.+]]: tensor<1x1x14x14x16xf32>
+//
+// Direct-access 9D tiled computation (input accessed via tensor.extract):
+// CHECK:         %[[RESULT:.+]] = linalg.generic
+// CHECK-SAME:      indexing_maps = [#[[$M_FLT]], #[[$M_OUT]]]
+// CHECK-SAME:      iterator_types = ["parallel", "parallel", "parallel", "parallel",
+// CHECK-SAME:                        "reduction", "reduction", "reduction",
+// CHECK-SAME:                        "parallel", "reduction"]
+// CHECK-SAME:      ins(%[[FILTER]]
+// CHECK-SAME:      outs(%[[OUTPUT]]
```
Max191, line 753: Does the image operand get removed and accessed via implicit capture? What is the purpose of doing this? It seems like it's losing some semantic information, which makes the matcher for the data tiled conv more complex/fragile.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
```diff
@@ -526,4 +526,176 @@ getEncodingInfoForMatmul(Encoding::EncodingAttr encoding,
...
+}
+
+/// Returns true if dim `d` participates in any result expression of `map`
+/// (bare or compound).
+static bool isUsedInMap(AffineMap map, unsigned d) {
+  return map.isFunctionOfDim(d);
+}
+
+FailureOr<ConvDimClassification>
+inferConvDimsFromMaps(ArrayRef<AffineMap> maps) {
```
Max191, line 543: Could we just use linalg::inferConvolutionDims?

compiler/src/iree/compiler/GlobalOptimization/GeneralizeLinalgNamedOps.cpp
```diff
@@ -55,11 +55,11 @@ void GeneralizeLinalgNamedOpsPass::runOnOperation() {
     bool generalizeConvOps = linalg::isaConvolutionOpInterface(linalgOp);
     if (clDisableConvGeneralization &&
         isa<linalg::Conv2DNhwcHwcfOp, linalg::Conv2DNchwFchwOp,
-            linalg::PoolingNhwcSumOp, linalg::PoolingNhwcMaxOp,
-            linalg::PoolingNhwcMaxUnsignedOp, linalg::PoolingNhwcMinOp,
-            linalg::PoolingNhwcMinUnsignedOp, linalg::PoolingNchwSumOp,
-            linalg::PoolingNchwMaxOp, linalg::DepthwiseConv2DNhwcHwcOp>(
-            linalgOp)) {
+            linalg::Conv2DNhwcFhwcOp, linalg::PoolingNhwcSumOp,
```
Max191, line 58: Maybe this change can be its own small PR. Seems like an harmless addition to handle another case. Perhaps we should even just use `linalg::isaConvolutionOpInterface` to capture all conv types, but may need some additional context for what the flag was for. Would be easier to discuss in a separate isolated PR.

Max191 (commented):
Overall, the encoding structure looks pretty good! I just have some comments/questions about the materialization part and about splitting up the PR:

I think this could be split into a few PRs. Particularly the encoding changes and the CPU external models changes. I think it could be something like:

 - Encoding infra
 - Encoding materialization
 - Codegen changes (i.e., KernelDispatch.cpp)
 - Enable setting data tiling hints (AnnotateDataTilingHints.cpp)
 - Small PR for named op generalization change

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
jschuhmacher: Unfortunately no. Unlike [inferContractionDims](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.h#L68) upstream does not have the same overload on maps for [inferConvolutionDims](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.h#L115)  (yet), and I don't have access to the op where I need this.

compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
jschuhmacher: Yes, `tensor.extract` is used to work around the input map not being a projected permutation (`(d0..d8) -> (d0, d4, d2*stride + d5*dilation, d3*stride +
d6*dilation, d8)`). This way the map can be dropped from the generic, and it does not get rejected by the upstream vectorization code later. The upstream vectorization code doesn't seem to expose any way to hook into it.

compiler/src/iree/compiler/DispatchCreation/AnnotateDataTilingHints.cpp
jschuhmacher: Good point, I'll move it.

jschuhmacher:
@Max191 Thanks for the review! I'll split this into the parts you suggested, and open the smaller PRs.

hanhanW:
> Overall, the encoding structure looks pretty good! I just have some comments/questions about the materialization part and about splitting up the PR:
> 
> I think this could be split into a few PRs. Particularly the encoding changes and the CPU external models changes. I think it could be something like:
> 
> * Encoding infra
> * Encoding materialization
> * Codegen changes (i.e., KernelDispatch.cpp)
> * Enable setting data tiling hints (AnnotateDataTilingHints.cpp)
> * Small PR for named op generalization change

Thanks Max for the review and the suggestion! I'll wait and review the split PRs.

compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
Max191: I see. Would it work for your use case if you materialized into a regular conv (with non-projected perm maps) during encoding materialization, and then added a pass before vectorization that decomposes convolutions into this tensor.extract form? That way we can keep the high level semantic information that this is a convolution (i.e., it passes linalg::isaConvolutionOpInterface) during higher level codegen phases and configuration (e.g., SelectLoweringStrategy), but the vectorizer will still be happy later once it is decomposed.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
Max191: I see. We should be able to add the overload for convolution upstream as well, and then use that here. If you send a PR upstream, I'd be happy to review it and help get it merged. It should be a straightforward addition so I'm sure nobody upstream will have and issues with it being added.

ai-mannamalai:
could be useful for memory constrained systems; would like to see it integrated sooner

jschuhmacher:
> could be useful for memory constrained systems; would like to see it integrated sooner

Definitely, that was the use case for it. Note that @phemashekar has taken over integrating the PR, and did many improvements on top. Most of it is already integrated, and useable today. See the tracking issue at https://github.com/iree-org/iree/issues/24710.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.cpp
jschuhmacher: @phemashekar added the overload upstream

PR #23749, opened 2026-03-12
[GPU] [CUDA]  Fix sm_89 target resolution and select a compatible default PTX version

slobodaapl:
## Summary

Fix CUDA target handling for `sm_89` by centralizing target resolution and choosing a compatible default PTX version.

## Problem

CUDA target handling for `sm_89` was inconsistent causing some GPUs to not be properly resolved:
- target normalization already mapped NVIDIA aliases to `sm_89`
- target-details lookup did not recognize `sm_89`
- default CUDA target features still effectively assumed `+ptx76`

This caused a failure on attempting to compile for CUDA target on `sm_89` GPUs because `getCUDATargetDetails(...)` could return null for `sm_89`, so the executable target was created without `iree_codegen.target_info`, which later failed during kernel configuration.

Even once target metadata was resolved, LLVM NVPTX rejects `sm_89` with PTX 7.6 because `sm_89` requires at least PTX 7.8.

## Fix

This change fixed CUDA target resolution in target utils:
- normalize NVIDIA CUDA targets before target-details lookup
- recognize `sm_89` / `ada` in the centralized NVIDIA target table
- select `+ptx78` for `sm_89` and thus move the hardcoded default ptx selection into a shared resolver with default fallback to `+ptx76`

## Testing
- added CUDA smoketest coverage for `sm_89`
- verified CUDA compilation succeeds for `sm_89` GPU
- verified the emitted PTX uses a compatible ISA level for `sm_89`

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
@@ -22,6 +22,8 @@ namespace mlir::iree_compiler::IREE::GPU {
 
 namespace {
 
+StringRef normalizeNVIDIAGPUTarget(StringRef target);
+
 //===----------------------------------------------------------------------===//
 // Target details structs
 //===----------------------------------------------------------------------===//
@@ -937,6 +939,7 @@ const WgpDetails *getPascalWgpDetails() {
 }
 
 std::optional<TargetDetails> getNVIDIAGPUTargetDetails(StringRef target) {
+  target = normalizeNVIDIAGPUTarget(target);
   const WgpDetails *ampereWgp = getAmpereWgpDetails();
   const WgpDetails *turingWgp = getTuringWgpDetails();
   const WgpDetails *voltaWgp = getVoltaWgpDetails();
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1134,12 +1138,26 @@ TargetAttr getMetalTargetDetails(MLIRContext *context) {
                           /*features=*/"spirv:v1.3,cap:Shader", context);
 }
 
+std::string getCUDATargetFeatures(StringRef target, StringRef features) {
```
Muzammiluddin-Syed-ECE: No need to have this return std::string, a `StringRef` would suffice.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.h
```diff
@@ -7,6 +7,8 @@
 #ifndef IREE_COMPILER_CODEGEN_DIALECT_GPU_TARGETUTILS_KNOWNTARGETS_H_
 #define IREE_COMPILER_CODEGEN_DIALECT_GPU_TARGETUTILS_KNOWNTARGETS_H_
 
+#include <string>
```
Muzammiluddin-Syed-ECE: We shouldn't need `std::string`

Issue #23094, opened 2026-01-12
[Codegen] Improve Vector lowerings to Backends (LLVM/SPIRV)


Groverkss:
## Problem

N-D vectors are a very good abstraction to work with when coming from higher tensor languages. For example:

```
vector<4x8xf16>
```

can directly represent a super vector load from a `64x64xf16` buffer.

However, backends don't generally have support for N-d vectors and need some way to lower these:

LLVM: Doesn't support N-D vectors -> 1-D vectors 
SPIR-V: Doesn't support N-D vectors + more restrictions

There are actually multiple ways we can lower these N-D vectors:

```
4 * vector<8xf16>
8 * vector<4xf16>
vector<32xf16>
```

each of these have their own trade-offs and reasons why we might prefer one over another.

## How do we solve/lower this today?

LLVM: N-D vector -> array<4 x vector<8>>
SPIR-V: Break down vector<4x8xf16> -> vector<1xvecsizexf16>
        using `vector.extract_strided_slice`

I will focus on our LLVMGPU lowering because I'm more familiar with that

### LLVMGPU Lowering

For some ops, we go to 1-d vectors, for some ops we don't do anything and for some ops we flatten:

- arith.select --> nothing? (this has been fixed in a recent UnrollElementwise patch)
- vector.transfer ops --> unrolling
- vector.contract -> flattening/unrolling ?
- vector.multi_reduction -> flattening/unrolling ?

This random pattern of patterns (pun intended) leads to IR that doesn't exactly compose well

```mlir
transfer_read %mask : vector<4x8xf16>

// ->

%c0 = arith.constant : vector<4x8xi1>
%mask = arith.select ... : vector<4x8xi1>
%mask0 = extract %mask : vector<8xi1> from vector<4x8xi1>
vector.maskedload %mask0 : vector<8xf16>
vector.maskedload %mask1 : vector<8xf16>
vector.maskedload %mask2 : vector<8xf16>
vector.maskedload %mask3 : vector<8xf16>
```

The masked transfer read here should have really represented a:

```mlir
if (mask) {
  load
}
```

which would have been visible if we unrolled the arith.select, but we choose to not do it today.

In the same spirit, we today flatten vector.contract operands when there are multiple parallel/reduction iterators to have 1 parallel and 1 reduction iterator:

```mlir
vector<4x8xf16> // p, p
vector.contract, iterator_types = [p, p, r]

vector<32xf16> // p, p --> p * p
[...]

PR #23752, opened 2026-03-12
Add vector.contract canonicalization.

amd-eochoalo:
This commit adds a canonicalization pattern that brings vector.contract into BMNK form in a single rewrite step.

Multi-dimensional vector.contract will be unrolled (with patterns yet to be added in https://github.com/iree-org/iree/pull/23753) . This unrolling requires that reduction dimensions appear in the innermost dimensions and parallel dimensions in the outermost dimensions.

Canonicalization of vector.contract is done in a single step. Alternative designs were explored but required running multiple patterns sequentially.

This canonicalization and unrolling are motivated by ticket https://github.com/iree-org/iree/issues/23094, but will remain unused until more components of the pipeline are available.

Note to reviewers: this is my first time submitting a bigger PR to IREE. Please feel free to point thing out anything that you'd like me to learn.

Assisted-by: Claude


kuhar (commented):
Where in the pipeline do you plan to run it? Whenever possible, we'd want to use specialized intrinsics for contract, and these often have strong preference on the exact form. For example, you can check how we handle contract -> i8 dot prod codegen in the spirv pipeline.

amd-eochoalo:
> Where in the pipeline do you plan to run it? Whenever possible, we'd want to use specialized intrinsics for contract, and these often have strong preference on the exact form. For example, you can check how we handle contract -> i8 dot prod codegen in the spirv pipeline.

Let me look at the SPIR-V pipeline in more detail. Thanks!

amd-eochoalo:
EDIT: Sorry, I was typing and pressed enter and it closed the PR instead of editing this comment.

I need to take a look at all possible targets to truly answer this question. I'll answer below, but I think there's some value in starting with one pass that lowers from n-d to 1-d vectors and then adapt it to multiple cases.

For the case of SPIR-V's "contract -> i8 dot prod", I think you are talking about this piece of code, right?

```
    // Prepare for SPIR-V integer dot product lowering.
    if (emitIntegerDotProdOps) {
      vector::populateVectorContractCanonicalizeMatmulToMMT(
          patterns, detectI8ToI32Matmul);
      if (failed(applyPatternsGreedily(funcOp, std::move(patterns)))) {
        return signalPassFailure();
      }

      debugPrint(funcOp, "after preparing for SPIR-V dot product lowering");
    }
```

So MMT:

```
/// Canonicalization of a `vector.contract %a, %b, %c` with row-major matmul
/// semantics to a contraction with MMT semantics (matrix matrix multiplication
/// with the RHS transposed). This specific form is meant to have the vector
/// operands are organized such that the reduction dimension is contiguous.
/// Example:
/// ```
/// vector.contract {indexing_maps = [affine_map<(m, n, k) -> (m, k)>,
///                                   affine_map<(m, n, k) -> (n, k)>,
///                                   affine_map<(m, n, k) -> (m, n)>],
///                  iterator_types = ["parallel", "parallel", "reduction"],
///                  kind = #vector.kind<add>} %a, %b, %c : ...
/// ```
```

The canonicalization pattern in this PR is a bit more general and also allows vector.contract operations to have batch dimensions (which will be ordered to be the outermost dimension). It then orders the iteration space to be [batch, batch ..., left, left, ... right, right..., reduction, reduction]. For the operand layouts:

* for lhs: [batch, ..., lhs-free, ... reduction]
* for rhs: [batch, ..., rhs-free, ... reduction]
* for acc [batch, ..., lhs-free, ..., rhs-free, ... reduction]

But let's continue to unrolling...

At the moment, the proposed unrolling will unroll vector.contract operations until only reduction dimensions remain. My intuition tells me that this should be lowered to either vector.multi_reduction and then unroll vector.multi_reduction until we get to the end cases where there are either:

* 1-d vector.reduction operations
* n-d arith elementwise operations (which can be unrolled with ElementwiseUnroll. See 

```
struct UnrollElementwiseOps final : public RewritePattern {
  UnrollElementwiseOps(MLIRContext *context, PatternBenefit benefit = 1)
      : RewritePattern(MatchAnyOpTypeTag(), benefit, context) {}
```

in compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUVectorLowering.cpp.

Let's compare this to SPIR-V's "contract -> i8 dot prod" pipeline.

```
    // Then unroll vectors to native vector size. We try to use 128-bit
    // vectors for memory access and 4/2/1 vector sizes for computation.
    {
      populateVectorUnrollPatterns(patterns, emitIntegerDotProdOps);
      if (failed(applyPatternsGreedily(funcOp, std::move(patterns)))) {
        return signalPassFailure();
      }
    }
```

which uses a different unrolling strategy than the one proposed. This one tiles all dimensions at once, while the one proposed in the following PR uses a rank reduction unrolling. We both achieve 1-d vector.contract ops but in different ways. I'll still need to make sure that I get them to a native size. (I have some intuition but I don't want to go through it just now.)

Finally after these are unrolled:

```
    // Lower reduction-unrolled vector contract ops. Such contract ops have
    // their reduction dimensions all be one, so we can convert them into
    // elementwise ops.
    {
      auto options =
          vector::VectorTransformsOptions().setVectorTransformsOptions(
              vector::VectorContractLowering::ParallelArith);
      vector::populateVectorContractLoweringPatterns(
          patterns, options.vectorContractLowering);
```

we could use this same lowerings for the 1-D vectors. Let me think about it.

krzysz00:
Without having thoroughly internalized the context, I'm not sure I agree that vector.contract should be unrolled until it's only reduction dimensions - matrix multiply intrinsics that get matched from `vector.contract` exist.

amd-eochoalo:
> Without having thoroughly internalized the context, I'm not sure I agree that vector.contract should be unrolled until it's only reduction dimensions - matrix multiply intrinsics that get matched from vector.contract exist.

Yes, that's a fair point that I failed to see initially as I just went with what the issue description stated.

> How we ideally want to structure conversion to backends is:
>
> Set of patterns to do unrolling from N-D vectors to 1-D vectors

But I think we could possibly tweak the pattern to unroll up to a given size maybe? I'll have to think about it. Thanks for pointing this out!

amd-eochoalo:
@krzysz00 do you think you have time to chat today? I'd be interested to hear if there are other operations not being fully unrolled would lower to optimized intrinsics (or similar).

amd-eochoalo:
> Without having thoroughly internalized the context, I'm not sure I agree that vector.contract should be unrolled until it's only reduction dimensions - matrix multiply intrinsics that get matched from vector.contract exist.

> do you think you have time to chat today?

Something that we agreed from this discussion is that targets / pipelines would benefit from being able to customize somewhat the unrolling and layout. The specific example we discussed was matrix multiplication (parallel-parallel-reduction). I mentioned that we could have a flattening pass which runs after the canonicalization which would flatten trailing reduction dimensions (which would be guaranteed by the current canonicalization pass here). (As a side note, we could customize this canonicalization to allow for different layouts, but I think at the moment that is beyond the scope of this PR and could be added later.)

There was another concrete request from Krzysztof where things (paraphrasing) "should not be unrolled naively". This meant, for example, a vector with the trailing dimensions being 1 would unroll naively to scalar code.

I mentioned that this does not directly contradict the description in https://github.com/iree-org/iree/issues/23094 which also allows for other optimizations to run before unrolling (which can remove the trailing dimension for example). 

Note: please do not read "canonicalization" in the sense of running during the greedy rewriter and more in the sense of "a pass that establishes some invariants for other passes".

compiler/src/iree/compiler/Codegen/Common/ContractCanonicalizationPatterns.cpp
```diff
@@ -0,0 +1,270 @@
...
+    populateContractLayoutCanonicalizationPatterns(patterns, /*benefit=*/1);
+    if (failed(applyPatternsGreedily(getOperation(), std::move(patterns)))) {
+      return signalPassFailure();
+    }
+  }
+};
+
+} // namespace
+
+void populateContractLayoutCanonicalizationPatterns(RewritePatternSet &patterns,
```
MaheshRavishankar: Just make the name more specific to what the pattern is doing instead of just calling it `Canonicalization`

compiler/src/iree/compiler/Codegen/Common/ContractCanonicalizationPatterns.h
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef IREE_COMPILER_CODEGEN_COMMON_CONTRACTCANONICALIZATIONPATTERNS_H_
+#define IREE_COMPILER_CODEGEN_COMMON_CONTRACTCANONICALIZATIONPATTERNS_H_
+
+#include "mlir/IR/PatternMatch.h"
+
+namespace mlir::iree_compiler {
+
+void populateContractLayoutCanonicalizationPatterns(RewritePatternSet &patterns,
```
MaheshRavishankar: Dont need a separate header file for this. Add it to existing Transforms header files.

amd-eochoalo: https://github.com/iree-org/iree/pull/23752/commits/4bb81fa29a03fd99bdfa0ff4fe0ebf37e8f13708

compiler/src/iree/compiler/Codegen/Common/ContractCanonicalizationPatterns.cpp
amd-eochoalo: https://github.com/iree-org/iree/pull/23752/commits/7b0c774db76e8cb9710c610691ed15217a124d7d https://github.com/iree-org/iree/pull/23752/commits/8b37fe43c07775429bb5ed23a5b94b2d6ca04899

amd-eochoalo:
> Whenever possible, we'd want to use specialized intrinsics for contract, and these often have strong preference on the exact form. For example, you can check how we handle contract -> i8 dot prod codegen in the spirv pipeline.

I think I can rewrite this pass to allow more flexibility. I think technically if change the layout to BMNK here, it is the same form (modulo an inversion or switching lhs and rhs operands...). But for the unrolling, we may be able to switch this pattern applicator into a partial conversion and specify that i8 matmuls are legal.

PR #23756, opened 2026-03-12
[Codegen] Delete TileAndDistributeToWorkgroups pass.

hanhanW:
After all the backends migrate to scf.forall distribution, the scf.for distribution no longer has users. The revision deletes the pass and the relevant tests.


hanhanW:
It depends on a cleanup PR: https://github.com/iree-org/iree/pull/23751

MaheshRavishankar (commented):
Nice!

MaheshRavishankar (approved):

PR #23757, opened 2026-03-12
[Codegen][Tuner] Add col_major parameter to MMAAttr/VirtualMMAAttr Python binding

bangtianliu:
PR #23633  removes the `attention_qk_matmul` and `attention_pv_matmul` marker attributes from attention decomposition configs, replacing them with a new `col_major = true` parameter on MMA intrinsics.

This PR updates the Python bindings to support the `col_major `parameter, enabling the tuner to generate attention configs compatible with the new approach.

After this PR, I will also add the required changes to the tuner side. 


kuhar (commented):
Can every mma intrinsic be made col_major? I thought that this wasn't supported for some wmmas. cc @krzysz00

bangtianliu:
> Can every mma intrinsic be made col_major? I thought that this wasn't supported for some wmmas. cc @krzysz00

This is really a valid point, cc @Groverkss here too.

krzysz00:
I'm _pretty_ sure all the WMMAs are symmetric - they all have 16x16xK shapes and take [M/N]xK input, so they're all allowed to do the `tr(C) = tr(B) * tr(A)` trick trivially (that is, by just std::swap-ing the left and right hand arguments)

The main thing is that the handling for WMMAs is missing that swap. I was fixing to have Claude add it and run some benchmarks around enabling it by default - I think there might be something there in terms of having a more contiguous writeback.

Groverkss:
> Can every mma intrinsic be made col_major? I thought that this wasn't supported for some wmmas. cc @krzysz00

yes, if handled correctly in compiler. This is a property of matmul, not intrinsics.

kuhar (approved):

PR #23758, opened 2026-03-12
[Codegen] Emulate `gather_to_lds` when it has narrow element types 

lialan:
First step to support DMA for scaled GEMMs.

* Add `ConvertGatherToLDS` pattern to `AMDGPUEmulateNarrowType pass`.
* In the pass focusing on `gather_to_lds` op, adjust subbyte element type to i8. e.g. `vector<32xf4E2M1FN>` -> `vector<16xi8>`.
* Semantically the same before and after.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
@@ -61,8 +61,8 @@ struct ConvertGatherToLDS final : OpConversionPattern<amdgpu::GatherToLDSOp> {
   LogicalResult
   matchAndRewrite(amdgpu::GatherToLDSOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    auto origSrcType = cast<MemRefType>(op.getSrc().getType());
-    auto origDstType = cast<MemRefType>(op.getDst().getType());
+    MemRefType origSrcType = op.getSrc().getType();
+    MemRefType origDstType = op.getDst().getType();
     auto newSrcType = cast<MemRefType>(adaptor.getSrc().getType());
     auto newDstType = cast<MemRefType>(adaptor.getDst().getType());
 
@@ -119,12 +119,9 @@ struct ConvertGatherToLDS final : OpConversionPattern<amdgpu::GatherToLDSOp> {
     Type newTransferType = convertTransferType(
         rewriter.getContext(), op.getTransferType(), origSrcBits, newSrcBits);
 
-    auto newOp = amdgpu::GatherToLDSOp::create(
+    amdgpu::GatherToLDSOp::create(
         rewriter, loc, adaptor.getSrc(), ValueRange{srcIdx}, adaptor.getDst(),
-        ValueRange{dstIdx}, TypeAttr::get(newTransferType));
-    if (op.getAsync()) {
-      newOp.setAsync(true);
-    }
+        ValueRange{dstIdx}, TypeAttr::get(newTransferType), op.getAsyncAttr());
 
     rewriter.eraseOp(op);
     return success();
@@ -141,7 +138,8 @@ private:
                                 int newBits) {
     auto [strides, offset] = origType.getStridesAndOffset();
 
-    // Fail if the offset or any stride is dynamic.
+    // Fail if the offset or any stride is dynamic. Dynamic offsets could be
+    // supported if guaranteed byte-aligned, but that hasn't been needed yet.
     if (ShapedType::isDynamic(offset)) {
       return nullptr;
     }
@@ -178,12 +176,16 @@ private:
   }
 
   // Converts the transfer type from sub-byte elements to byte-sized elements,
-  // preserving the total transfer size in bits.
+  // preserving the total transfer size in bits. The caller must ensure
+  // totalBits is a multiple of newBits (the op verifier enforces that
+  // transfer sizes are 8, 16, 32, 96, or 128 bits, all multiples of 8).
   static Type convertTransferType(MLIRContext *context, Type origType,
                                   int origBits, int newBits) {
     if (auto vecType = dyn_cast<VectorType>(origType)) {
       int64_t totalBits =
           vecType.getNumElements() * vecType.getElementTypeBitWidth();
+      assert(totalBits % newBits == 0 &&
+             "transfer size must be a multiple of the new element bit width");
       int64_t newElems = totalBits / newBits;
       return VectorType::get({newElems}, IntegerType::get(context, newBits));
     }
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
@@ -78,10 +78,10 @@ struct ConvertGatherToLDS final : OpConversionPattern<amdgpu::GatherToLDSOp> {
     }
 
     Location loc = op.getLoc();
-    int origSrcBits = origSrcType.getElementTypeBitWidth();
-    int newSrcBits = newSrcType.getElementTypeBitWidth();
-    int origDstBits = origDstType.getElementTypeBitWidth();
-    int newDstBits = newDstType.getElementTypeBitWidth();
+    int64_t origSrcBits = origSrcType.getElementTypeBitWidth();
+    int64_t newSrcBits = newSrcType.getElementTypeBitWidth();
+    int64_t origDstBits = origDstType.getElementTypeBitWidth();
+    int64_t newDstBits = newDstType.getElementTypeBitWidth();
 
     // Only convert when the transfer vector's total bits are a multiple of
     // the new element bit width. E.g. vector<3xf4E2M1FN> (12 bits) cannot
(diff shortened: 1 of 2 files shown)


lialan:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/iree-org/iree/pull/23758#pullrequestreview-3939474642)

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
```diff
@@ -50,6 +51,146 @@ struct ConvertRawBufferCast final
...
+/// (e.g. f4E2M1FN) to byte-sized types (i8) during narrow type emulation.
+/// The pattern linearizes multi-dimensional indices into the converted 1D
+/// memref space and adjusts the transfer type accordingly.
+struct ConvertGatherToLDS final : OpConversionPattern<amdgpu::GatherToLDSOp> {
+  using Base::Base;
+
+  LogicalResult
+  matchAndRewrite(amdgpu::GatherToLDSOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto origSrcType = cast<MemRefType>(op.getSrc().getType());
```
krzysz00: Do these need the casts?

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
```diff
@@ -50,6 +51,146 @@ struct ConvertRawBufferCast final
...
+    }
+
+    // Adjust transfer type to use the new element type.
+    Type newTransferType = convertTransferType(
+        rewriter.getContext(), op.getTransferType(), origSrcBits, newSrcBits);
+
+    auto newOp = amdgpu::GatherToLDSOp::create(
+        rewriter, loc, adaptor.getSrc(), ValueRange{srcIdx}, adaptor.getDst(),
+        ValueRange{dstIdx}, TypeAttr::get(newTransferType));
+    if (op.getAsync()) {
```
krzysz00: ... this should be an argument to the builder?

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
```diff
@@ -50,6 +51,146 @@ struct ConvertRawBufferCast final
...
+  //   linearIdx = offset + sum(idx[i] * stride[i])
+  //   packedIdx = linearIdx / (newBits / origBits)
+  static Value linearizeAndPack(ConversionPatternRewriter &rewriter,
+                                Location loc, ValueRange indices,
+                                MemRefType origType, int origBits,
+                                int newBits) {
+    auto [strides, offset] = origType.getStridesAndOffset();
+
+    // Fail if the offset or any stride is dynamic.
+    if (ShapedType::isDynamic(offset)) {
```
krzysz00, line 139: Does this actually need to be a failure? You can look up the offsets.

(I think the only constraint is if the eventual address is *odd* - that is, if you're trying to land the first element of the DMA into the middle of a byte) but I can't see how that'd come up.

... Ok, in theory, it could come up, but I think the solution here is to require byte alignment when dealing with sub-byte DMAs, which feels fairly easy to maintain.

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
```diff
@@ -50,6 +51,146 @@ struct ConvertRawBufferCast final
...
+    }
+
+    return linearIdx;
+  }
+
+  // Converts the transfer type from sub-byte elements to byte-sized elements,
+  // preserving the total transfer size in bits.
+  static Type convertTransferType(MLIRContext *context, Type origType,
+                                  int origBits, int newBits) {
+    if (auto vecType = dyn_cast<VectorType>(origType)) {
```
krzysz00, line 196: I assume a non-multiple-of-8-bits transfer size is a can't happen?

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
lialan: Done. Removed the unnecessary `cast<MemRefType>` on `op.getSrc().getType()` and `op.getDst().getType()` since the op accessors already return `TypedValue<MemRefType>`. Kept the casts on the adaptor accessors where they're needed.

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
lialan: Done. Passing `op.getAsyncAttr()` directly as a builder argument instead of setting it after creation.

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
lialan: Added a comment explaining the rationale: dynamic offsets could be supported if guaranteed byte-aligned, but that hasn't been needed yet. Left it as a failure for now since we don't have a use case requiring it.

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
lialan: Yes, the op verifier enforces transfer sizes of 8, 16, 32, 96, or 128 bits — all multiples of 8. Added an assert and a comment documenting this invariant.

lialan:
@krzysz00 no offence, I was testing claude automation, and it replied your review comments all by itself.

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
```diff
@@ -50,6 +51,148 @@ struct ConvertRawBufferCast final
...
+      return failure();
+    }
+
+    // If types didn't change, nothing to do.
+    if (newSrcType == origSrcType && newDstType == origDstType) {
+      return failure();
+    }
+
+    Location loc = op.getLoc();
+    int origSrcBits = origSrcType.getElementTypeBitWidth();
```
krzysz00:
```suggestion
    int64_t origSrcBits = origSrcType.getElementTypeBitWidth();
```
and below

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
```diff
@@ -50,6 +51,148 @@ struct ConvertRawBufferCast final
...
+    // Fail if the number of indices doesn't match the rank.
+    if (indices.size() != strides.size()) {
+      return nullptr;
+    }
+
+    // Linearize: offset + sum(idx[i] * stride[i]).
+    Value linearIdx = arith::ConstantIndexOp::create(rewriter, loc, offset);
+    for (auto [idx, stride] : llvm::zip(indices, strides)) {
+      Value strideVal = arith::ConstantIndexOp::create(rewriter, loc, stride);
+      Value product = arith::MulIOp::create(rewriter, loc, idx, strideVal);
```
krzysz00: Stick some nsw and nuw on these

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
```diff
@@ -50,6 +51,148 @@ struct ConvertRawBufferCast final
...
+    if (!srcIdx) {
+      return rewriter.notifyMatchFailure(
+          op, "failed to linearize source indices (dynamic or mismatched "
+              "strides/offset, or invalid bit-width ratio)");
+    }
+
+    // Linearize destination indices.
+    Value dstIdx = linearizeAndPack(rewriter, loc, op.getDstIndices(),
+                                    origDstType, origDstBits, newDstBits);
+    if (!dstIdx) {
```
krzysz00: This'll violate the "failing patterns don't modify IR" rule

Check both the source and destination memrefs for convertabiliity before matching either.

lialan: Good catch, moved the check into `canLinearizeAndPack` to guard it.

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
lialan: done.

compiler/src/iree/compiler/Codegen/LLVMGPU/AMDGPUEmulateNarrowType.cpp
lialan: done.

krzysz00 (approved):
This is approved in that it's correct

*However*, I think it will have issues in practice.

I think you can get away with the static-everything restriction for the LDS half of the address.

*However*, for global memory, we'll need to handle matrices like `memref<?x?x32xf4E2M1FN>` or the `offset: ?` you often get in whole-model compilation.

For these, in practice, I'd like to loosen up the condition to "the last stride is 1 and the last index has the correct divisibility". I think, as an initial matter, just checking for an even constant (probably 0) in that final field will work for now.

PR #23760, opened 2026-03-12
[DMA][Codegen] Use DMA to promote scaled matmul operands

lialan:
Step 2 of enabling DMA for scaled GEMMs. Lands the codegen/config plumbing so DMA can be opted into for scaled matmul, while keeping it off by default until follow-ups land.

  * Drop the blanket guard that rejected DMA for all scaled matmuls.
  * Route LHS/RHS *and* `scale` operands through `UseGlobalLoadDMAAttr`. Scale tiles are required to be DMA-aligned; misaligned tiles fall back to the existing stream-copy path.
  * Disable XOR swizzle when DMA is active for scaled GEMMs (the two paths are mutually exclusive for now; XOR + DMA will come in a follow-up).
  * Gate sub-byte scaled DMA off until narrow-type emulation lands (mxfp4 lowering bug).
  * DMA-side fixes needed to make scale operands work end-to-end:
    - Single-segment DWORD-alignment exemption on `tensor.pad` rows so
      small (≤ one segment) scale copies are not rejected.
    - Per-warp tile alignment cap so the workgroup-aligned-but-warp-
      sub-aligned scale case (#24139 sibling) lowers cleanly.
    - Identity per-warp tiling for that case.
    - Drop the warp guard for single-subgroup DMA copies.
  * New e2e + pipeline tests: `e2e_matmul_cdna4_mxfp4_dma`, ROCDL
    pipeline_scaled_matmul_dma lit.


lialan:
Working on DMA with XOR Swizzle support in a subsequent PR.

tests/e2e/matmul/CMakeLists.txt
```diff
@@ -2624,6 +2624,37 @@ iree_generated_e2e_runner_test(
...
+  TEST_TYPE
+    matmul
+  GENERATOR
+    "generate_e2e_matmul_tests.py"
+  GENERATOR_ARGS
+    "--lhs_rhs_type=f4E2M1FN"
+    "--acc_type=f32"
+    "--mx_scale_type=f8E8M0FNU"
+    "--mx_block_size=32"
+    "--shapes=easy_large_static"
```
krzysz00: Do we need to use large shapes here?

krzysz00 (commented):
This looks broadly reasonable to me, but @Muzammiluddin-Syed-ECE .

tests/e2e/matmul/CMakeLists.txt
lialan: right, changed to small.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -830,42 +830,32 @@ struct GPUConvertToCoalescedDMAPass final
...
-    // derived_thread_config). If ALL are DMA-convertible, upgrade them all to
-    // use_global_load_dma. If ANY fails, downgrade them all to
-    // derived_thread_config.
+    // Pre-check: verify that all copies marked with use_global_load_dma are
+    // actually DMA-convertible. If any DMA-marked copy fails the check,
+    // downgrade ALL DMA-marked copies to derived_thread_config.
+    // Copies already marked with derived_thread_config are left unchanged —
+    // they should not be upgraded to use_global_load_dma because they may
+    // have shapes (e.g. scale operands) that are too small for DMA after
+    // per-warp tiling, leading to incorrect thread distribution.
```
Muzammiluddin-Syed-ECE: This seems like a good thing to track in an issue. As a follow up itd be nice if we had a util function to compute what tiling configs make operands suitable or unsuitable for DMA so that we can eventually enable async copies of scales by selecting amenable tiling configs.

Muzammiluddin-Syed-ECE (commented):
Excited for this set of changes to go through, downgrading the scales makes sense for now

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
lialan: Done: https://github.com/iree-org/iree/issues/23797

kuhar (commented):
Have you checked what the performance impact of this change is?

lialan:
> Have you checked what the performance impact of this change is?

I haven't. It will be bad or not impressive at most. 
Because at this transition stage, it disables XOR swizzling when turned on.

We need to turn on both features and then benchmark

bjacob (approved):
approving tests/e2e

krzysz00 (commented):
Seems correct, though same note as from the other PR about not landing dramatic regressions

lialan:
Updated to support `scale` as well, though the DMA loading can be quite small, it is needed to enable async multi-buffering.

Next step is to enable and test XOR swizzling + DMA for scaled GEMMs.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_scaled_matmul_dma.mlir
hanhanW, line 1: I'd appreciate if someone can review https://github.com/iree-org/iree/pull/24080

Then we can use `--iree-codegen-llvmgpu-nvvm-lowering-pipeline='include-llvm-lowering=false'` here

hanhanW: Many thanks!

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -435,9 +476,13 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
     // After tiling, the input is typically:
     //   tensor.extract_slice %padded[...] [...] [1, 1]
     // We need to trace through extract_slice to find if source is tensor.pad.
+    // There may be nested pads (e.g., from GPUPromoteMatmulOperands padding
```
Yu-Zhewen, line 479: The comment here makes sense, but I’m not fully following how the code handles nested padding. It looks like we only unwrap one level, am I missing something?

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_scaled_matmul_dma.mlir
```diff
@@ -0,0 +1,94 @@
...
+// Verify pipeline completes and produces scaled MFMA compute ops.
+// LHS/RHS are promoted to workgroup shared memory and scales use thread-based
+// copies. The compute uses 16x16x128 scaled MFMA instructions.
+
+// CHECK-LABEL: func.func @scaled_matmul_dma
+//   CHECK-DAG:   memref.alloc() : memref<{{.*}}xf8E8M0FNU, #gpu.address_space<workgroup>>
+//   CHECK-DAG:   memref.alloc() : memref<{{.*}}xf4E2M1FN, #gpu.address_space<workgroup>>
+//       CHECK:   gpu.thread_id
+//       CHECK:   scf.for
+// TODO: The DMA config is set but the pipeline currently lowers LHS/RHS copies
```
Yu-Zhewen, line 90: I’m not sure why this is marked as a TODO. Can we enable DMA for this test?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -472,8 +515,36 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
...
+              ShapedType::getNumElements(sourceType.getShape());
+          singleSegment = totalSrcElements <= *minAligned;
+        }
+        if (rowBytes % 4 != 0 && !singleSegment) {
+          LLVM_DEBUG(llvm::dbgs()
+                     << "Skipping DMA: row size " << rowBytes
+                     << " bytes not DWORD-aligned (slow path)\n");
+          return failure();
+        }
+      }
```
Yu-Zhewen: I think the whole check should be moved inside `hasDWORDAlignedRows`.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -191,6 +192,14 @@ static void resolveHintOp(RewriterBase &rewriter,
     if (isa<amdgpu::GatherToLDSOp, ViewLikeOpInterface>(user)) {
       continue;
     }
+    // Treat reshape/subview ops as transparent users. When a swizzle_hint is
```
Yu-Zhewen: This looks like a merging issue, since the `ViewLikeOpInterface` above already covers these cases.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -35,6 +34,7 @@
 
 namespace mlir::iree_compiler {
 
+
```
Yu-Zhewen:
```suggestion
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUDistributeForall.cpp
```diff
@@ -21,6 +21,9 @@
 
 namespace mlir::iree_compiler {
 
+static constexpr llvm::StringLiteral kRedundantOnDistribute =
```
Yu-Zhewen: We should probably revert all changes in this file. We no longer need to use `redundant_on_distribute`, since `scf.if` with multiple warps has been confirmed to provide better performance.

compiler/src/iree/compiler/Codegen/Common/NormalizeLoopBounds.cpp
```diff
@@ -158,6 +158,9 @@ LogicalResult normalizeLoopBounds(RewriterBase &rewriter,
       rewriter, rewriter.getUnknownLoc(), newLoopParams->lowerBounds,
       newLoopParams->upperBounds, newLoopParams->steps, forallOp.getOutputs(),
       forallOp.getMapping());
+  // Preserve discardable attributes (e.g. pipeline hints set by earlier
```
Yu-Zhewen: Same, no longer needed

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -10,6 +10,7 @@
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Arith/Utils/Utils.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
```
Yu-Zhewen: Looks like a stale change because of merge.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -187,7 +188,8 @@ static void resolveHintOp(RewriterBase &rewriter,
     // Gather_to_lds destination-side swizzle is handled by
     // AMDGPULowerCoalescedDMAToGatherLDS, which applies the inverse swizzle
     // to source indices. Treat gather_to_lds and view-like ops as transparent
-    // users that pass through the swizzled allocation.
+    // users that pass through the swizzled allocation. ViewLikeOpInterface
```
Yu-Zhewen: Stale change

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -787,7 +787,10 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
                              rhsScaleType};
 
   Location loc = operands[0].getLoc();
-  if (useDirectLoad &&
+  // Scaled matmul enables DMA via the promotionArray path below (mxfp4 data
+  // operands), so the general rejection (which also filters non-f16/bf16) is
+  // skipped here.
+  if (useDirectLoad && !scaled &&
```
Yu-Zhewen: Instead of having `!scaled` here, maybe move it inside `shouldRejectDirectLoadDMA`?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -936,18 +939,53 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
+    if (useDirectLoad) {
+      Attribute useGlobalDma = IREE::GPU::UseGlobalLoadDMAAttr::get(context);
+      // Check if per-workgroup scale shape is DMA-aligned. Scale operands
+      // have shape [M/N, Ko] where Ko = K-reduction-tile *
+      // koBlocksPerIntrinsic. reductionTileSizes[contractionK] counts MFMA
+      // invocations, not Ko elements; multiply by koBlocksPerIntrinsic to get
+      // the element count.
+      int64_t scaleMDim = workgroupTileSizes[contractionM.back()];
+      auto smmaKind = cast<IREE::GPU::ScaledMMAAttr>(kind);
+      int64_t koBlocksPerIntrinsic = std::get<2>(smmaKind.getScaledMNKShape());
```
Yu-Zhewen: Can we also move this part inside `shouldRejectDirectLoadDMA`? Something like

```
static bool shouldRejectDirectLoadDMA() {
  // Case 1: DMA requires hardware support (gfx950+ / CDNA4+).
  if (!targetSupportsGlobalLoadDMA(target)) {
    return true;
  }

  // Case 2: Only GEMM are supported currently.
  if (!isGemm) {
    return true;
  }

  // Case 3: LHS transposed, RHS not transposed show regressions with DMA.
  if (transposedLhs && !transposedRhs) {
    return true;
  }

  if (scaled) {
    // Case 4: Scale not aligned.
    if (!isScaleAligned(scaleElements, dmaSizes)) {
      return true;
    }
  }
  else {
    // Case 5: Only f16/bf16 are supported currently.
    if (!isF16OrBF16(lhsElemType) || !isF16OrBF16(rhsElemType)) {
      return true;
    }
  }
 return false;
}
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp
```diff
@@ -58,6 +58,8 @@ static constexpr llvm::StringLiteral kUnrollAccDistribute =
     "unroll_acc_distribute";
 static constexpr llvm::StringLiteral kUnrollAccReassemble =
     "unroll_acc_reassemble";
+static constexpr llvm::StringLiteral kRedundantOnDistribute =
```
Yu-Zhewen: Same here, the file can be reverted

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -850,45 +1046,26 @@ struct GPUConvertToCoalescedDMAPass final
...
 
-    // Pre-check: decide whether all linalg.copy ops should be DMA-converted.
-    // Only activate when at least one copy already has use_global_load_dma
-    // (indicating DMA intent from upstream config, e.g. --iree-llvmgpu-use-
-    // direct-load). Collect all promoted copies (use_global_load_dma or
-    // derived_thread_config). If ALL are DMA-convertible, upgrade them all to
-    // use_global_load_dma. If ANY fails, downgrade them all to
-    // derived_thread_config.
-    // Note: GatherOps are excluded — they come from input IR (not from
+    // Pre-check: verify that each copy marked with use_global_load_dma is
```
Yu-Zhewen: The refactor here was introduced to handle cases where only the LHS/RHS are DMA, while the scales are stream copies. However, we’ve already decided to drop that strategy. We should consider reverting the changes in lines 1049–1068.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -593,9 +689,135 @@ struct ConvertToCoalescedDMABase : OpRewritePattern<OpTy> {
                                                             OpTy op) const = 0;
 };
 
+/// Pad a sub-aligned copy's source and destination to DMA-aligned sizes.
```
Yu-Zhewen: I think this is a useful feature to have. However, after the scale alignment guard in ConfigUtils, we should no longer hit this situation? If that’s the case, it might be better to move all `padCopyForDMA` related changes into a separate PR

Yu-Zhewen (commented):
Thanks @lialan.  Just a bit of final cleanup and we should be good to land this.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -317,31 +349,45 @@ static bool isCopyDMAConvertible(linalg::CopyOp copyOp) {
   }
 
   auto outputType = cast<RankedTensorType>(copyOp.getOutputs()[0].getType());
-  int64_t rank = outputType.getRank();
   ArrayRef<int64_t> shape = outputType.getShape();
-  int64_t innermostDim = shape[rank - 1];
-  if (ShapedType::isDynamic(innermostDim)) {
+  if (llvm::any_of(shape, ShapedType::isDynamic)) {
```
Yu-Zhewen: I'm not sure quite sure why we want to tighten the constraint here from `ShapedType::isDynamic(innermostDim)` to `llvm::any_of(shape, ShapedType::isDynamic)`

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -317,31 +349,45 @@ static bool isCopyDMAConvertible(linalg::CopyOp copyOp) {
...
+  // Padding requires replacing the destination with tensor.empty, so it's
+  // only viable when the original output is already tensor.empty.
+  if (!outputIsEmpty) {
     return false;
   }
 
-  return getDMAAlignedSubgroupSize(funcOp, outputType.getElementType(),
-                                   availableElements)
-      .has_value();
+  // Check if padding can bring it to alignment.
```
Yu-Zhewen: I wonder what this check is doing here? I suspect it was related to `padCopyForDMA`, which has now been removed from this PR, so it might no longer be needed. Maybe we can just revert the whole `isCopyDMAConvertible` to previous state?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -554,10 +600,14 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
...
+    // Eligibility (including the DWORD / single-segment check) was already
+    // vetted by isCopyDMAConvertible in the pre-check, so isValidPadForDMA
+    // here is a consistency guard rather than a fresh decision.
     tensor::PadOp pad = traceToTensorPad(input);
-    if (pad && isValidPadForDMA(pad)) {
-      source =
-          createClampedSourceSliceFromPad(rewriter, loc, pad, inParallelOp);
+    auto parentFuncOp =
+        innerOp->template getParentOfType<FunctionOpInterface>();
+    if (pad && isValidPadForDMA(pad, parentFuncOp)) {
```
Yu-Zhewen, line 617: Looks like a merge issue. Why we are dropping `createClampedSourceSliceFromPad` (introduced in #24116)?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -696,6 +747,28 @@ struct ConvertToCoalescedDMABase : OpRewritePattern<OpTy> {
 struct ConvertCopyToCoalescedDMA : ConvertToCoalescedDMABase<linalg::CopyOp> {
   using ConvertToCoalescedDMABase::ConvertToCoalescedDMABase;
 
+  LogicalResult matchAndRewrite(linalg::CopyOp op,
```
Yu-Zhewen: This override looks redundant now

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -1106,92 +1153,88 @@ struct GPUConvertToCoalescedDMAPass final
...
+    // If available elements are not aligned to transfer size, skip.
+    if (availableElements % *minAligned != 0) {
       return failure();
     }
 
-    // In the linearized-availability case the per-warp tile shrinks with more
-    // warps and can fall below the minimum DMA transfer, leaving an orphan
-    // `linalg.copy {use_global_load_dma}` inside a warp-mapped forall
-    // (issue #24139). Halve `totalWarps` until each warp's tile is aligned.
+    // Check if this is a tensor.pad fusion case.
```
Yu-Zhewen: I think there are also some merging issues between line 1180–1237:
1. The `maxSegments` calculation duplicates `computeMaxFeasibleNumWarps` from #24116
2. The `if (isPadFusion) { single-iteration }` branch reintroduces the workaround #24116 explicitly removed

Would reverting the whole block between line 1180–1237 work?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -999,9 +1044,31 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
+    // invocations, not Ko elements; multiply by koBlocksPerIntrinsic.
+    int64_t scaleMDim = workgroupTileSizes[contractionM.back()];
+    auto smmaKind = cast<IREE::GPU::ScaledMMAAttr>(kind);
+    int64_t koBlocksPerIntrinsic = std::get<2>(smmaKind.getScaledMNKShape());
+    int64_t scaleKoDim =
+        reductionTileSizes[contractionK.back()] * koBlocksPerIntrinsic;
+    int64_t scaleElements = scaleMDim * scaleKoDim;
+    constexpr int64_t kScaleBits = 8; // f8E8M0FNU is 8-bit
+    ArrayRef<int64_t> dmaSizes;
+    if (auto dmaSizesAttr = target.getWgp().getDmaSizes()) {
```
Yu-Zhewen: Nit:
```suggestion
    if (DenseI64ArrayAttr dmaSizesAttr = target.getWgp().getDmaSizes()) {
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -1106,92 +1153,88 @@ struct GPUConvertToCoalescedDMAPass final
...
-    }
-
-    // Get the element type bit width.
     Type elementType = outputType.getElementType();
-    int64_t elementBits = elementType.getIntOrFloatBitWidth();
-
-    // Get DMA sizes from target to compute minimum transfer size.
-    IREE::GPU::TargetAttr target = getGPUTargetAttr(funcOp);
-    if (!target) {
+    auto minAligned = getMinDMAAlignedElements(funcOp, elementType);
```
Yu-Zhewen: Nit:

```suggestion
    std::optional<int64_t> minAligned = getMinDMAAlignedElements(funcOp, elementType);
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -206,32 +221,23 @@ getDMAAlignedSubgroupSize(FunctionOpInterface funcOp, Type elementType,
...
-    if (perWarp >= minElementsPerTransfer &&
-        perWarp % minElementsPerTransfer == 0) {
-      return n;
-    }
+/// Returns the subgroup size if the available elements are aligned to DMA
+/// transfer sizes, std::nullopt otherwise.
+static std::optional<int64_t>
+getDMAAlignedSubgroupSize(FunctionOpInterface funcOp, Type elementType,
+                          int64_t availableElements) {
+  auto minAligned = getMinDMAAlignedElements(funcOp, elementType);
```
Yu-Zhewen:
```suggestion
  std::optional<int64_t> minAligned = getMinDMAAlignedElements(funcOp, elementType);
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -277,35 +283,61 @@ computeThreadNumThreadsImpl(OpBuilder &builder, Operation *op,
...
+  }
+  // Single-segment exemption: safe only when the innermost dimension has no
+  // padding. When the inner dim is padded, output row width != source row
+  // width, so linearized DMA reads wrong source elements for padded positions
+  // instead of returning zeros.
+  SmallVector<OpFoldResult> highPad = pad.getMixedHighPad();
+  if (highPad.empty() || !isConstantIntValue(highPad.back(), 0)) {
+    return false;
+  }
+  auto minAligned = getMinDMAAlignedElements(funcOp, elemType);
```
Yu-Zhewen:
```suggestion
  std::optional<int64_t> minAligned = getMinDMAAlignedElements(funcOp, elemType);
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -317,31 +349,45 @@ static bool isCopyDMAConvertible(linalg::CopyOp copyOp) {
...
     availableElements = ShapedType::getNumElements(shape);
   }
 
   tensor::PadOp pad = traceToTensorPad(copyOp.getInputs()[0]);
-  if (pad && !isValidPadForDMA(pad)) {
+  if (pad && !isValidPadForDMA(pad, funcOp)) {
+    return false;
+  }
+
+  auto minAligned =
```
Yu-Zhewen:
```suggestion
  std::optional<int64_t> minAligned =
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -803,52 +876,26 @@ struct ConvertGatherToCoalescedDMA
...
     if (ShapedType::isDynamic(innermostDim)) {
       return failure();
     }
 
-    Type elementType = outputType.getElementType();
-    int64_t elementBits = elementType.getIntOrFloatBitWidth();
-
-    IREE::GPU::TargetAttr target = getGPUTargetAttr(funcOp);
-    if (!target || !targetSupportsGlobalLoadDMA(target)) {
+    auto minAligned =
```
Yu-Zhewen:
```suggestion
    std::optional<int64_t> minAligned =
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_convert_to_coalesced_dma.mlir
```diff
@@ -483,15 +483,15 @@ func.func @copy_with_tensor_pad_fusion(%source: tensor<121x64xf32>, %init: tenso
...
     outs(%init : tensor<4x64xf32>) -> tensor<4x64xf32>
 
-  // Key check: tensor.pad is fused. The DMA source is a clamped sub-slice of
-  // the pre-pad source. in_bounds = [false, true] because M dim has padding.
-  // CHECK-DAG: %[[EXTRACTED:.+]] = tensor.extract_slice %[[SRC]]
-  // CHECK: scf.forall
-  // CHECK:   scf.forall (%[[LANE:.+]]) in (64)
-  // CHECK:     arith.minsi
-  // CHECK:     %[[WARP_SRC:.+]] = tensor.extract_slice %[[EXTRACTED]]
+  // Key check: tensor.pad is fused — DMA source is the pre-pad extract_slice
```
Yu-Zhewen: I think we only need to add one new test (`copy_scale_capped_warps`). Any other changes or deletions in the file might no longer be needed (I guess the CI will tell).

Yu-Zhewen (approved):
LGTM, thanks @lialan. Only one concern on lit test

PR #23767, opened 2026-03-12
[Codegen] Migrate InnerTiles to VectorizableOpInterface.

hanhanW:
The revision moves the implementation to interface and retires the VectorizeIREEGPUOps pass. The implementation of `ValueBoundsOpInterface` is done within the PR, because it is used for vector size inference.

The pass and the location of the implementation were not moved during IREEGPU -> IREECodegen move, so they looks wrong.


compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -843,6 +845,73 @@ struct PadOpVectorizationModel
...
+                                          ArrayRef<bool> scalableDims,
+                                          DictionaryAttr options) const {
+    auto tiledOp = cast<IREE::Codegen::InnerTiledOp>(op);
+    OpBuilder::InsertionGuard g(rewriter);
+    rewriter.setInsertionPoint(tiledOp);
+    Location loc = tiledOp.getLoc();
+
+    SmallVector<ShapedType> argTypes = tiledOp.getOperandShapedTypes();
+
+    // Construct the (never used) zero padding value for each operand.
```
qedawkins: I don't think this is right. If the outer dimensions were dynamic I think the padding value would be used (inner dimensions are forced static). I suspect to be exceedingly correct we'd need the InnerTile interface to return the padding value to use, but I'm guessing this is a pre-existing issue so just updating the comment is fine.

qedawkins (approved):

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: yeah, it's a pre-existing issue. Let me update the comment.

https://github.com/iree-org/iree/blob/b2f73e337ac3cb54872fc834b781f398d8ad0be9/compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp#L2010-L2015

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization.mlir
```diff
@@ -1216,3 +1216,108 @@ func.func @map_store_f4_mask_depends_on_inner_index(
...
+    semantics = #iree_gpu.mma_semantics<distributed = true, opaque = false>
+  } : tensor<2x3x4xf16>, tensor<3x5x4xf16> into tensor<2x5x4xf32>
+  return %0 : tensor<2x5x4xf32>
+}
+
+//      CHECK-LABEL: func @tensor_multi_mma
+// CHECK-MASK-LABEL: func @tensor_multi_mma
+// CHECK-FOLD-LABEL: func @tensor_multi_mma
+// CHECK-GATHER-LABEL: func @tensor_multi_mma
+// CHECK-MAP-STORE-LABEL: func @tensor_multi_mma
```
sommerlukas, line 1241: What's the purpose of this? Avoid false matches for check lines above?

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -843,6 +845,76 @@ struct PadOpVectorizationModel
...
+
+    auto zero = arith::ConstantIndexOp::create(rewriter, loc, 0);
+    SmallVector<Value> results;
+    for (auto [result, tensorAcc] :
+         llvm::zip_equal(newTiledOp.getResults(), tiledOp.getOutputs())) {
+      int64_t rank = cast<RankedTensorType>(tensorAcc.getType()).getRank();
+      auto write = vector::TransferWriteOp::create(
+          rewriter, loc,
+          /*vector=*/result,
+          /*source=*/tensorAcc,
```
sommerlukas: Nit:
```suggestion
          /*valueToStore=*/result,
          /*base=*/tensorAcc,
```

sommerlukas (approved):

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
hanhanW: I'd just remove the comments, they are clear to me.

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization.mlir
hanhanW: Yes; I'm going to split the test file in a follow-up. I'm waiting for few in-flight PR landed, then I'll start the work.

hanhanW:
Investigating the failure, may be missing implementation for ValueBoundsOpInterface

PR #23768, opened 2026-03-12
[Codegen] Apply bounds to subgroup_id

krzysz00:
Since we're fixing to start using subgroup_id more often with PCF, apply bounds to it based on the subgroup size (or sizes) and the number of threads in the workgroup. Also extend subgroup_size handling to account for known subgroup sizes instead of giving up completely when there isn't a fixed choice made yet.

Also fix up some double-spaces in test attributes.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
@@ -208,32 +208,20 @@ struct PropagateDispatchSizeBoundsPass final
     }
 
     // Compute the subgroup ID bound: max total threads / min subgroup size.
+    std::optional<int64_t> maxFlatWorkgroupSize;
     std::optional<int64_t> subgroupIdBound;
-    if (minSubgroupSize) {
-      int64_t maxTotalThreads = 1;
-      bool allSizesKnown = true;
-      for (std::optional<int64_t> size : workgroupSizes) {
-        if (size) {
-          maxTotalThreads *= *size;
-          // Cap at the hardware thread-per-workgroup limit inside the loop
-          // to avoid overflow from multiplying per-dimension maximums.
-          if (target) {
-            maxTotalThreads =
-                std::min(maxTotalThreads,
-                         static_cast<int64_t>(
-                             target.getWgp().getMaxThreadCountPerWorkgroup()));
-          }
-        } else {
-          allSizesKnown = false;
-          break;
-        }
-      }
-      if (!allSizesKnown && target) {
-        maxTotalThreads = target.getWgp().getMaxThreadCountPerWorkgroup();
-      }
-      if (allSizesKnown || target) {
-        subgroupIdBound = llvm::divideCeil(maxTotalThreads, *minSubgroupSize);
-      }
+    if (staticWorkgroupSize) {
+      maxFlatWorkgroupSize = llvm::product_of(*staticWorkgroupSize);
+    }
+    if (target) {
+      maxFlatWorkgroupSize = std::min(
+          maxFlatWorkgroupSize.value_or(std::numeric_limits<int64_t>::max()),
+          static_cast<int64_t>(
+              target.getWgp().getMaxThreadCountPerWorkgroup()));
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
```diff
@@ -179,9 +207,45 @@ struct PropagateDispatchSizeBoundsPass final
...
+          if (target) {
+            maxTotalThreads =
+                std::min(maxTotalThreads,
+                         static_cast<int64_t>(
+                             target.getWgp().getMaxThreadCountPerWorkgroup()));
+          }
+        } else {
+          allSizesKnown = false;
+          break;
+        }
```
amd-eochoalo: Nit: I think this is a little bit more difficult to parse than it should be. Could we remove one level of nesting by moving the break earlier and then the else should not be necessary, right? Maybe changing the logic slightly too for calculating the maxTotalThreads when target is known and avoid overflow somehow or adding variables?


// pseudocode

```suggestion
      // if target is available
      std::optional<int64_t> maxThreadCountPerWg = static_cast<int64_t> (target.getWgp().getMaxThreadCountPerWorkgroup()));
        
      for (std::optional<int64_t> size : workgroupSizes) {
        if (!size) {
          allSizesKnown = false;
          break;
        }
        
        maxTotalThreads *= *size;
        // Cap at the hardware thread-per-workgroup limit inside the loop
        // to avoid overflow from multiplying per-dimension maximums.
        if (maxThreadCountPerWg) {
          maxTotalThreads = std::min(maxTotalThreads, maxThreadCountPerWg)
        }
      }
```

amd-eochoalo (approved):
I think this looks good, but you may want to wait for other reviewers. Just one comment about the for-loop.

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
krzysz00: I refactored all that code - do you like the new version?

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
```diff
@@ -157,10 +171,24 @@ struct PropagateDispatchSizeBoundsPass final
 
       if (std::optional<uint64_t> exportSubgroupSize =
               exportOp->getSubgroupSizeAsUInt()) {
-        subgroupSize = exportSubgroupSize;
+        staticSubgroupSize = *exportSubgroupSize;
```
efric: nit: prefer explicit cast

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
```diff
@@ -179,9 +207,33 @@ struct PropagateDispatchSizeBoundsPass final
...
+    if (maxFlatWorkgroupSize && minSubgroupSize) {
+      subgroupIdBound =
+          llvm::divideCeil(*maxFlatWorkgroupSize, *minSubgroupSize);
+    }
+
+    std::optional<int64_t> constantSubgroupSize;
+    if (minSubgroupSize && maxSubgroupSize &&
+        *minSubgroupSize == *maxSubgroupSize) {
+      constantSubgroupSize = *minSubgroupSize;
+    }
```
efric: nit: is there a test for this path?

efric (approved):
lgtm overall % nits, thanks

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
krzysz00: Yeah, I'm pretty sure existing tests cover this

krzysz00: ... Oh, no, you're right, there should be a test for this, I'll go add one (and move this to the right spot)

PR #23769, opened 2026-03-12
[plugins][nfc] Drop redundant `public` from struct inheritance. 1/5

hanhanW:
`public` is the default access specifier for struct inheritance in C++, so `struct Foo : public Bar` is equivalent to `struct Foo : Bar`.


kuhar (approved):
Most of these, if not all, can be made final. I wish we had a clang-tidy check for this.

PR #23774, opened 2026-03-13
[Codegen][Map] Add PackLayout distribution patterns for generic vector ops

Groverkss:
Add distribution patterns for vector.shape_cast, vector.broadcast,
vector.transpose, and vector.step under PackLayoutAttr. These are
simple enough to be clubbed together in a single change.


Groverkss:
The core patterns are just 200 lines + 150 lines of helpers. Rest is mostly setup and tests.

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef IREE_COMPILER_CODEGEN_DIALECT_MAP_TRANSFORMS_MAPDISTRIBUTIONUTILS_H_
+#define IREE_COMPILER_CODEGEN_DIALECT_MAP_TRANSFORMS_MAPDISTRIBUTIONUTILS_H_
+
+#include "iree/compiler/Codegen/Dialect/Map/IR/IREEMapAttrs.h"
+#include "llvm/ADT/SmallVector.h"
+#include "mlir/IR/Builders.h"
+
+namespace mlir::iree_compiler {
+
+/// Maps a distributed vector dim to its original TV layout dim and leaf stride.
```
sommerlukas: Nit: I think we should spell out what `TV` means here, as we aren't using that abbreviation elsewhere.

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.cpp
```diff
@@ -0,0 +1,148 @@
...
+/// Compute per-dimension thread offsets from a linearized thread ID.
+///
+/// Each dimension (mode) of a PackLayoutAttr has interleaved thread leaves
+/// (stride > 0) and broadcast leaves (stride == 0). The goal is to convert a
+/// flat thread ID into a data-space offset for each dimension.
+///
+/// Example: layout <((4, 2), (4, 8)) : ((1, 0), (0, 4))>
+///   Dim 0 mode: sizes (4, 2), strides (1, 0) -> thread leaf size=4, bcast
+///   Dim 1 mode: sizes (4, 8), strides (0, 4) -> bcast, thread leaf size=8
+///   Result: offset_0 = (tid % 4) * 2,  offset_1 = (tid / 4) % 8
```
sommerlukas: I think it would make sense to add an explanation _why_ this is the correct result. 

Just to check my understanding of this example layout (and the map layouts in general): What we have here is a layout where every thread holds 2 values in dim0 and 4 values in dim1. The threads are arranged column-first, so something like:
```
t0 t4 t8  t12 ...
t1 t5 t9  t13 ...
t2 t6 t10 t14 ...
t3 t7 t11 t15 ...
```

The equivalent nested layout would be something like this (ignoring batch and outer for now):
```
thread_tile = [4, 8],
element_tile = [2, 4],

thread_stride = [1, 4]
```

Is that understanding correct?

How does this compare to the previous nested layout? Are the broadcast leaves what used to be the element (or batch x outer x element) tile in nested layout?

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.cpp
```diff
@@ -0,0 +1,148 @@
...
+  int64_t rank = layout.getRank();
+  for (int64_t d = 0; d < rank; ++d) {
+    SmallVector<LeafInfo> valLeaves =
+        filterLeafInfos(layout.getShapeMode(d), layout.getStrideMode(d),
+                        [](const LeafInfo &l) { return l.stride == 0; });
+    if (valLeaves.empty()) {
+      map.push_back({d, 1});
+      continue;
+    }
+    for (const LeafInfo &leaf : valLeaves) {
```
sommerlukas, line 32: Is this still a map if we have multiple entries per dimension? Or do we only expect a single entry here?

sommerlukas (commented):
Implementation mostly looks good to me.

I'm not entirely sure about the naming: Analog to `GenericVectorization`, based on PR title and file names, I initially had expected these files to handle `linalg.generic`, so maybe we can find a better name.

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.h
Groverkss: that's an artifact of the old definition. Removed.

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.cpp
Groverkss: It is still a map. We are just finding leaves and assigning them the top level dimension.

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.cpp
Groverkss: The broadcast leaves are used to represent batch x outer x element in nested layout. The strided leaves are used to represent thread/subgroup stuff.

Yes that would be the equivalent layout.

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.cpp
```diff
@@ -0,0 +1,137 @@
...
+///
+///   This works because linearize_index computes:
+///     result = coord_0 * (S1*S2*...*Sn) + coord_1 * (S2*...*Sn) + ... +
+///     coord_n
+///
+///   The effective stride for position i is product(S_{i+1}...S_n), which is
+///   exactly how getLeafInfos defines dataStride[i]. Broadcast zeros drop out,
+///   so only thread leaves contribute: each scaled by its correct data stride.
+///
+///   For the example above, dim 0 produces:
```
sommerlukas: Nit: Which example above (either I'm overlooking it or this comment is outdated)?

sommerlukas (approved):
Just one minor nit left, otherwise LGTM.

compiler/src/iree/compiler/Codegen/Dialect/Map/Transforms/MapDistributionUtils.cpp
Groverkss: oof let me remove this... this is old

PR #23778, opened 2026-03-13
[Codegen][GPU] Add XOR swizzle hints to GPUVectorAlloc for bank conflict elimination

keshavvinayak01:
This PR adds an opt-in path for `GPUVectorAlloc` to use XOR swizzling for promoted GPU shared-memory allocations.

The current padding-based bank-conflict reduction changes the physical LDS shape by padding the innermost dimension. This patch keeps the logical allocation shape visible to the vector distribution path, attaches an `iree_codegen.swizzle_hint` when the propagated vector layouts indicate that an XOR shuffle is useful, and resolves that hint later into physical swizzled accesses. Allocations that are not hinted still flow through the existing padding pass.

## Logical changes

* `GPUVectorAlloc` now inspects the propagated `NestedLayoutAttr` for the writer and reader layouts of a promoted shared-memory allocation.
* From those layouts it derives the XOR swizzle parameters instead of hard-coding shape-specific behavior:
  * `innerDim` comes from the promoted vector type.
  * `accessWidth` comes from the writer/reader element tiles.
  * the number of shared-memory banks comes from the GPU target.
  * swizzling is skipped when the row stride is already coprime with the bank count, because consecutive rows already distribute across banks.
* For element types that are at least one full LDS bank wide, the f32/f64-style scalar-reader case caps the effective access width instead of using the writer width directly. This keeps the swizzle bijective for wide stores followed by narrower reads.
* For sub-bank element types such as f16/bf16, swizzling is now used only when the writer and reader access widths match. If the access widths differ and the chosen swizzle width is wider than scalar, the allocation is left unhinted so `GPUReduceBankConflicts` can pad it instead. This avoids paying per-element XOR offset computation on the scalar/sub-access side in `ResolveSwizzleHints`; in attention this keeps Q/K-style vector-vector allocations swizzled and leaves V-style vector-store/scalar-load allocations to padding.
* When swizzling is selected, the allocation is wrapped in `iree_codegen.swizzle_hint` with an `#iree_codegen.xor_shuffle` attribute.
* `FlattenSwizzleHintAllocs` converts the hinted multi-dimensional allocation into the flat 1D form expected by `ResolveSwizzleHints`, while preserving the logical expanded shape for users.
* `ResolveSwizzleHints` materializes the hint by rewriting vector loads/stores to use swizzled flat offsets. Wide vector accesses are split into `accessWidth` chunks and each chunk is swizzled independently. If any direct load/store access for a hint cannot be safely swizzled, the optimization is dropped for that entire hint instead of partially swizzling the allocation.
* `GPUReduceBankConflicts` remains enabled. It skips only shared-memory allocs directly protected by a swizzle hint, so unrelated allocs in the same dispatch can still use padding.

## How the XOR swizzle works

For a flat LDS buffer representing rows of `row_width` elements and contiguous access groups of `access_width` elements, the swizzle operates at access-group granularity:

```
N = row_width / access_width
logical group (row i, column j) -> physical group (row i, column ((i / per_phase) % N) ^ j)
```

`ResolveSwizzleHints` aligns an offset down to the containing `access_width` group, computes the XOR-swizzled group column, and then adds the original intra-group remainder back. This preserves element-level addressing while changing which LDS bank each logical access group maps to.

## Benchmark summary

The latest f16/f32 benchmark CSVs and the exact benchmark script are in this gist:

https://gist.github.com/keshavvinayak01/a936c0e65a5a6ae7dd6fd7239039d443

The benchmark is non-causal attention (`is_causal = false`): the generated IR uses `iree_linalg_ext.attention` without a causal mask and yields the score unchanged.

F16: 

| hd | shapes | median speedup | mean speedup | exact win/same/loss | +/-1% win/same/loss | bank-conflict result |
| ---: | ---: | ---: | ---: | --- | --- | --- |
| all | 110 | 0.9951x | 0.99442x | 21 / 1 / 88 | 4 / 84 / 22 | padding BC=0 on 55/110, swizzle BC=0 on 110/110 |
| 64 | 55 | 0.9971x | 0.99628x | 17 / 0 / 38 | 3 / 43 / 9 | swizzle BC=0 on 55/55 |
| 128 | 55 | 0.9932x | 0.99255x | 4 / 1 / 50 | 1 / 41 / 13 | swizzle BC=0 on 55/55 |

F32:
| hd | shapes | median speedup | mean speedup | swizzle faster | bank-conflict result |
| ---: | ---: | ---: | ---: | ---: | --- |
| all | 110 | 1.0824x | 1.0703x | 109/110 | BC counter unchanged |
| 64 | 55 | 1.0362x | 1.0518x | 54/55 | BC counter unchanged |
| 128 | 55 | 1.0892x | 1.0889x | 55/55 | BC counter unchanged |

The f32 result is consistently positive. The latest f16 guarded policy eliminates the measured bank conflicts for the full sweep and reduces LDS size, but it is performance-neutral to slightly negative overall; the guard avoids applying swizzle to the scalar/sub-access f16 allocation pattern that was responsible for the worst slowdown.

Benchmarking done on MI300X


Groverkss (changes requested):
We need data to see if it works or not.

kuhar:
> We need data to see if it works or not.

+1. Specifically, we should get lds bank conflict counts from rocprof, not just raw timing.

keshavvinayak01:
Yeah I need to have backing data and also compare with the padding based solution before we review this.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -31,11 +34,56 @@ namespace mlir::iree_compiler {
...
+  int64_t elemBytes = (elementType.getIntOrFloatBitWidth() + 7) / 8;
+
+  int64_t rowBytes = innerDimSize * elemBytes;
+  int64_t rowStrideBanks = (rowBytes / kSharedMemoryBankWidthBytes) % numBanks;
+
+  // If stride is 0 every row hits the same banks (worst case).
+  // If gcd(stride, numBanks) == 1, no conflicts.
+  if (rowStrideBanks != 0 && std::gcd(rowStrideBanks, numBanks) == 1) {
+    return std::nullopt;
+  }
```
Muzammiluddin-Syed-ECE: This would be great to have as a utility function in Codegen/Utils/GPUUtils.cpp

Muzammiluddin-Syed-ECE (commented):
Looks like the right direction! However, I think this is a good opportunity to improve the XOR related utility functions we have in Codegen/Utils/GPUUtils.cpp and we can also integrate some of them into this change, such as `isXORShuffleValid`.

keshavvinayak01:
The benchmark shapes were taken from https://github.com/nod-ai/iree-kernel-benchmark/blob/b2351fb4af6f5061f595b06321c47d39fd341a1d/iree_kernel_benchmark/attentionbench/problems.py#L12

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -105,10 +221,28 @@ materializeSharedMemoryConversions(FunctionOpInterface funcOp) {
...
+    std::optional<IREE::Codegen::XORShuffleAttr> swizzle;
+    if (clEnableVectorAllocSwizzle) {
+      auto inputLayout = layouts.lookup(op.getInput());
+      auto outputLayout = layouts.lookup(op.getResult());
+      // If the input value doesn't have a propagated layout, use the
+      // to_layout's own layout as the output and treat the input as having
+      // the same element tile (conservative: accessWidth = reader's width).
+      if (!inputLayout && outputLayout) {
+        inputLayout = outputLayout;
+      }
```
sommerlukas: I don't think that's the right approach here. If we don't have a layout on the read from global memory, we should take the opportunity and use a layout that maximizes coalescing etc instead of using a specialized MMA layout for global data access.

compiler/src/iree/compiler/Codegen/Common/FlattenSwizzleHintAllocs.cpp
```diff
@@ -44,9 +44,26 @@ struct FlattenSwizzleHintAllocsPass final
...
                                      IREE::Codegen::SwizzleHintOp hintOp) {
   auto allocOp = hintOp.getOperand().getDefiningOp<memref::AllocOp>();
-  if (!allocOp || !allocOp->hasOneUse()) {
+  if (!allocOp) {
     return;
   }
+  // The alloc may have one direct use (the hint) or two uses (the hint and a
+  // matching memref.dealloc inserted by bufferization). Anything else (e.g.
+  // an aliased subview) means we can't safely rewrite the alloc shape.
+  memref::DeallocOp deallocOp;
```
sommerlukas, line 53: Why do we need handling of `dealloc` while the existing paths through this pass don't need it?

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -73,24 +111,34 @@ static void swizzleLoad(RewriterBase &rewriter, vector::LoadOp load,
   Location hintLoc = hintOp.getLoc();
   int64_t accessWidth = hintOp.getSwizzle().getAccessElementCount();
   VectorType type = load.getVectorType();
+  SwizzledAccessKind accessKind = getSwizzledAccessKind(load, accessWidth);
+  assert(accessKind != SwizzledAccessKind::Unsupported &&
+         "expected verified swizzled vector load");
+  if (accessKind == SwizzledAccessKind::Unsupported) {
+    return;
+  }
```
sommerlukas: Why do we have an assert and an early return here?

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -51,6 +51,44 @@ static Value createOrFoldNewStaticAdd(RewriterBase &rewriter, Value v,
   return arith::AddIOp::create(rewriter, v.getLoc(), v, offsetVal);
 }
 
+enum class SwizzledAccessKind {
+  Unsupported,
+  SingleElement,
+  Chunked,
+};
+
+static SwizzledAccessKind getSwizzledAccessKind(VectorType type,
```
sommerlukas, line 60: Why do we need these changes? Why were they not necessary on the other paths so far? Wouldn't it be better to avoid this scenario in `GPUVectorAlloc` by not inserting the hint?

sommerlukas (commented):
Some high-level comments in the first iteration. IMHO, this should also be rebased on https://github.com/iree-org/iree/pull/24408 because we can then freely select the writer layout to fit our needs for swizzling. 

Also, how can the speedup on `f32` be explained if the bank conflict count didn't change with swizzling?

keshavvinayak01:
@sommerlukas I suspect the f32 performance increase is due to IR size + lower LDS usage as a side effect (I did a diff to notice these things), but that's empirical. I'll put up more data to pinpoint it.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
keshavvinayak01: Removed the redundant early return. The verifier/classifier decides whether the hint is resolvable before rewriting; once swizzleLoad/swizzleStore are called, unsupported access is an invariant violation, so the assert is enough.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
keshavvinayak01: GPUVectorAlloc now avoids emitting hints for the problematic sub-bank mismatched-access cases, so f16/bf16 fall back to padding there. ResolveSwizzleHints still supports the single-element access case because the f32 promoted-read path writes wider vectors but the MFMA consumer reads scalar elements; dropping that path removes the f32 swizzle in the final ROCDL lowering. The resolver now classifies accesses once and only accepts full chunks or vector<1>, with lit coverage for scalar load/store and unsupported wider sub-accesses.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
keshavvinayak01: Merged the promoted-read layout changes and changed swizzle selection to use the derived coalesced global-read layout as the writer layout. We no longer synthesize a writer layout from the MMA consumer layout when the global read itself had no propagated layout.

compiler/src/iree/compiler/Codegen/Common/FlattenSwizzleHintAllocs.cpp
keshavvinayak01: The dealloc case is needed for the bufferized GPUVectorAlloc path: bufferization leaves a memref.dealloc as a direct user of the alloc before FlattenSwizzleHintAllocs runs, so without rewriting it the rank-2 alloc cannot be flattened and the swizzle is dropped later. I constrained this to only one memref.dealloc as the sole non-hint use, rewrote it to the new flat alloc, and added lit coverage for both the accepted case and the extra-use bailout.

keshavvinayak01:
@sommerlukas 
I went back to profiling some of the f32 shapes, and I consistently observed the following:
```
| metric, 1x1024x128xf32 | padding | swizzle |
|------------------------|---------|----------|
| LdsLatency | 48.21 | 43.65 |
| SQ_LDS_BANK_CONFLICT | 532480 | 532480 |
| SQ_WAIT_INST_ANY | 2.04M | 1.62M |
| SQ_WAVE_CYCLES | 7.88M | 7.67M |
| SQ_INSTS | 1.87M | 2.40M |
| SQ_INSTS_LDS | 298K | 514K |
| SQ_INSTS_MFMA | 393K | 393K |
| SQ_INSTS_VMEM | 27648 | 27648 |
```

For this ~8\% perf gain: 
- `LdsLatency` drops about 9.5%: 48.21 -> 43.65 cycles.
- `SQ_WAIT_INST_ANY` drops about 21%: 2.04M -> 1.62M.
- `SQ_WAVE_CYCLES` drops about 2.7%: 7.88M -> 7.67M.

Swizzling essentially replaces padded 2D LDS layouts with flat power-of-two LDS layouts and XOR-remapped indices.

Speedup therefore appears to come from improved LDS critical-path behavior, not from reduced global-memory traffic, MFMA work, or aggregate bank-conflict count.

keshavvinayak01:
I'll also wait for https://github.com/iree-org/iree/pull/24408 to go in first, like you said.

PR #23780, opened 2026-03-13
[Tuner] Add iree_codegen.smt.assert op

kuhar:
Add a named SMT assertion for use inside iree_codegen.smt.constraints regions. Even though the SMT dialect already comes with an assert op, I found constraints to be much more readable when supplemented with the explanation of what is being checked. The new assert op carries a format string with `{}` placeholders and !smt.int args for diagnostics in the evaluator/verifier, so that we can produce proper diagnostics later on.

This is the last op I need before we can start emitting pipeline constraints.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
@@ -566,3 +566,18 @@ LogicalResult LookupOp::verify() {
   }
   return success();
 }
+
+LogicalResult AssertOp::verify() {
+  size_t placeholderCount = 0;
+  StringRef fmt = getMsg();
+  for (size_t pos = 0; (pos = fmt.find("{}", pos)) != StringRef::npos;
+       pos += 2) {
+    ++placeholderCount;
+  }
+  if (placeholderCount != getPrintArgs().size()) {
+    return emitOpError("format string has ")
+           << placeholderCount << " placeholder(s) but got "
+           << getPrintArgs().size() << " arg(s)";
+  }
+  return success();
+}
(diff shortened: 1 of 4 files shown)


bangtianliu (approved):

amd-eochoalo (changes requested):
Why is LLVM updated here?

kuhar:
> Why is LLVM updated here?

gah, good catch, fixed

third_party/llvm-project
amd-eochoalo, line 1: It is still here! :)

kuhar: it's gone

amd-eochoalo: Now it is gone. Thanks!

amd-eochoalo (approved):

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
```diff
@@ -566,3 +566,18 @@ LogicalResult LookupOp::verify() {
   }
   return success();
 }
+
+LogicalResult AssertOp::verify() {
+  size_t placeholderCount = 0;
+  StringRef fmt = getMsg();
+  for (size_t pos = 0; (pos = fmt.find("{}", pos)) != StringRef::npos;
```
amd-eochoalo, line 573: I think one comment I have is that aren't int64_t preferred over size_t? Or is that just in LLVM.
Also, this loop is a bit strange, but I don't think it is wrong.

kuhar: find/npos are size_t

amd-eochoalo: size_t is find since that is the type of the parameter. Never mind. Thanks!

kuhar: > I think one comment I have is that aren't int64_t preferred over size_t?

ask any 3 people and you will get 10 different answers :) my personal heuristic is to align with the upstream and use unsigned types for counting

PR #23781, opened 2026-03-13
[Util] Add InferIntDivisibilityOpInterface for affine.delinearize_index

Max191:
Teaches the integer divisibility analysis to propagate through affine.delinearize_index ops by building per-result affine expressions that mirror the delinearize semantics (floordiv + mod of the linear input by stride products).

Also implements visitModExpr for constant RHS expressions, which previously returned minimum divisibility (1,1) for all mod expressions. Now, it computes gcd(lhs_divisibility, rhs_constant), and returns divisibility 0 (lattice top) when the LHS divisibility is a multiple of the mod constant, since the result is always zero in that case.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
+++ b/compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
@@ -71,7 +71,46 @@ util.func @affine_apply_floordiv_non_exact(%arg0 : index) -> index {
 util.func @affine_apply_mod(%arg0 : index) -> index {
   %0 = util.assume.int %arg0<udiv = 16> : index
   %1 = affine.apply affine_map<(d0) -> (d0 mod 16)>(%0)
+  // 16 % 16 == 0, so x mod 16 is always 0 -> divisibility 0 (lattice top).
+  // CHECK: divisibility = "udiv = 0, sdiv = 0"
+  %2 = "iree_unregistered.test_int_divisibility"(%1) : (index) -> index
+  util.return %2 : index
+}
+
+// -----
+
+// CHECK-LABEL: @affine_apply_mod_always_zero_multiple
+// If x has divisibility 32 and we compute x mod 16, then 32 % 16 == 0 so
+// x is always a multiple of 16, meaning x mod 16 is always 0.
+util.func @affine_apply_mod_always_zero_multiple(%arg0 : index) -> index {
+  %0 = util.assume.int %arg0<udiv = 32> : index
+  %1 = affine.apply affine_map<(d0) -> (d0 mod 16)>(%0)
+  // CHECK: divisibility = "udiv = 0, sdiv = 0"
+  %2 = "iree_unregistered.test_int_divisibility"(%1) : (index) -> index
+  util.return %2 : index
+}
+
+// -----
+
+// CHECK-LABEL: @affine_apply_mod_non_constant_rhs
+// Non-constant RHS in mod expression -> bail to divisibility 1.
+util.func @affine_apply_mod_non_constant_rhs(%arg0 : index, %arg1 : index) -> index {
+  %0 = util.assume.int %arg0<udiv = 16> : index
+  %1 = affine.apply affine_map<(d0)[s0] -> (d0 mod s0)>(%0)[%arg1]
   // CHECK: divisibility = "udiv = 1, sdiv = 1"
+  %2 = "iree_unregistered.test_int_divisibility"(%1) : (index) -> index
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -278,6 +298,81 @@ struct AffineMaxInferIntDivisibilityOpInterface
...
+    // Operands are: [linear_index, dynamic_basis_values...]
+    IREE::Util::ConstantIntDivisibility linearDiv =
+        getDivisibilityOfOperand(delinOp.getLinearIndex(), argDivs[0]);
+
+    ArrayRef<int64_t> staticBasis = delinOp.getStaticBasis();
+    int64_t numResults = delinOp.getNumResults();
+
+    // Build affine expressions for each result.
+    // Dim 0 = linear index, symbols = dynamic basis values.
+    AffineExpr linearExpr = getAffineDimExpr(0, ctx);
```
krzysz00, line 321: I _think_ there might be code for this upstream, but it might be coupled to a lowering and so not very reusable.

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -278,6 +298,81 @@ struct AffineMaxInferIntDivisibilityOpInterface
...
+
+    // Collect operand divisibilities: [linear_index_div, dynamic_basis_divs...]
+    SmallVector<IREE::Util::ConstantIntDivisibility> operandDivs;
+    operandDivs.push_back(linearDiv);
+
+    // Map static/dynamic basis values to affine expressions.
+    int dynIdx = 0;
+    SmallVector<AffineExpr> basisExprs;
+    for (int64_t i = 0; i < (int64_t)staticBasis.size(); ++i) {
```
krzysz00: You can get the basis as `OpFoldResult`s

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
```diff
@@ -278,6 +298,81 @@ struct AffineMaxInferIntDivisibilityOpInterface
...
+    // Effective basis B[k] = basisExprs[basisStart + k], for k = 0..N-2.
+    // Stride s[i] = product of B[i..N-2] = product of
+    //               basisExprs[basisStart+i .. end].
+    //
+    // result[0]   = x floordiv s[0]
+    // result[i>0] = (x floordiv s[i]) mod B[i-1]
+    // For i=N-1, s[N-1]=1, so result[N-1] = x mod B[N-2].
+    for (int64_t i = 0; i < numResults; ++i) {
+      AffineExpr stride = getAffineConstantExpr(1, ctx);
+      for (int64_t j = basisStart + i; j < (int64_t)basisExprs.size(); ++j) {
```
krzysz00: Can we lose the O(n^2) here by computing the stride as we go up the basis / iterating from the end?

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
krzysz00: ... Ah, I see why we need this, never mind

compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp
krzysz00: Also the usual nit about `var = init, e = size; var < e` for for loops

krzysz00 (commented):
Concept looks correct, minor implementation nits

compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
```diff
@@ -276,3 +315,60 @@ util.func @scf_forall_multiple_ivs(%arg0 : index, %arg1 : index, %arg2 : index,
...
   util.return
 }
+
+// -----
+
+// CHECK-LABEL: @delinearize_static_no_outer
+util.func @delinearize_static_no_outer(%arg0 : index) {
+  %input = util.assume.int %arg0<udiv = 8> : index
+  %0:3 = affine.delinearize_index %input into (3, 16) : index, index, index
+  // result[0] = input floordiv 48 -> udiv = 1 (8 does not divide 48 evenly)
```
nirvedhmeshram: nit: should the comment say 48 does not divide 8 evenly?

krzysz00 (approved):
lgtm here

PR #23783, opened 2026-03-13
[Codegen] Delete ConvertToDestinationPassingStylePass pass.

hanhanW:
The pass is used by base pipeline in SPIRV and LLVMGPU. They were added for scf.for fusion and better analysis in bufferization, but there is no tiling before the bufferization in the pipeline. Thus, the pass can be removed from the pipeline.

Then it no longer has users, so we can retire the pass.

There are few onnx tests passing after removing the pass, so they are enabled:
- test_tfidfvectorizer_tf_only_bigrams_skip0
- test_tfidfvectorizer_tf_onlybigrams_levelempty
- test_tfidfvectorizer_tf_onlybigrams_skip5
- test_tfidfvectorizer_tf_uniandbigrams_skip5

ci-extra: test_torch


hanhanW:
MI355 seems to be flaky. They go different pipeline, i.e., they don't use base pipeline.

MaheshRavishankar (approved):

PR #23789, opened 2026-03-14
[SPIRV] Fix ub.poison legalization failure for 2d vectors

kuhar:
This is to fix a compilation failure with a proprietary model before the v3.11 release; the proper fix will land later.

`conv_1d` with unit filter size produces `vector.contract` results with trailing unit dimensions (e.g., `vector<1x16x1xf32>`). After stripping the leading unit dim, `vector<16x1xf32>` remains. Vector unrolling creates a `ub.poison` + `vector.insert` chain for this type, and the subsequent `vector.extract` ops read from intermediate inserts. SPIR-V lowering chokes on this because of 2d vectors.

The upstream extract fold (`ExtractFromInsertTransposeChainState`) bails on this case and, AFAICT, no other composition of canon patterns can handle it today. Add `FoldExtractFromInsertUnitDim` to `SPIRVInitialVectorLowering` as a local workaround that handles this specific case.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
@@ -41,6 +41,60 @@ namespace mlir::iree_compiler {
 
 namespace {
 
+/// Folds vector.extract from vector.insert when the extract position is a
+/// prefix of the insert position and the remaining (un-indexed) dimensions
+/// of the extracted sub-vector are all size 1. In that case the extracted
+/// value is fully determined by the inserted value.
+///
+/// Example:
+///   %ins = vector.insert %s, %v [3, 0] : f32 into vector<16x1xf32>
+///   %ext = vector.extract %ins [3] : vector<1xf32> from vector<16x1xf32>
+/// folds to:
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
```diff
@@ -41,6 +41,60 @@ namespace mlir::iree_compiler {
...
+    }
+
+    ArrayRef<int64_t> extractPos = extractOp.getStaticPosition();
+    ArrayRef<int64_t> insertPos = insertOp.getStaticPosition();
+
+    // The extract position must be a strict prefix of the insert position.
+    if (extractPos.size() >= insertPos.size()) {
+      return failure();
+    }
+    if (extractPos != insertPos.take_front(extractPos.size())) {
```
qedawkins, line 76: Combine these checks or add distinguishing match failures?

qedawkins (approved):

compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
kuhar: Ack but that's minor, I'm going to land as is and fix when moving this to mlir

PR #23791, opened 2026-03-15
[Codgen][ROCm] Fix vector distribution for transposed outputs

kuhar:
Layer norm-style dispatches with a multi-output generic that has a transposed output used to crash with `failed to distribute` on a proprietary model.

Teach `shouldAttachLoweringConfig` to recognize non-identity output indexing maps so the op gets a `lowering_config` and proper `to_layout` anchors.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
@@ -428,6 +428,15 @@ populateConfigInfo(const llvm::SetVector<linalg::LinalgOp> &computeOps,
   // LinalgOp with only parallel dims. This is needed if the op cannot be fused
   // with a reduction or introduces new loop dimensions.
   auto shouldAttachLoweringConfig = [&](linalg::LinalgOp linalgOp) -> bool {
+    // If any output has a non-identity indexing map, the op needs its own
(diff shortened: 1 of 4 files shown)


kuhar:
I don’t think this is a fundamental limitation of vector distribute, but I don’t have the cycles to dig into the layout propagation code now before v3.11. I think this surfaced a gap in forward layout propagation.
cc: @Groverkss

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -662,6 +662,31 @@ LogicalResult setReductionConfig(IREE::GPU::TargetAttr target,
...
+        LDBG() << "Bailing: non-projected-permutation output map " << map;
+        return failure();
+      }
+      // isProjectedPermutation guarantees all results are AffineDimExprs.
+      // Reject maps where the result dimensions are not in ascending order,
+      // which indicates a transposed output.
+      auto dimPositions =
+          llvm::map_to_vector(map.getResults(), [](AffineExpr expr) {
+            return cast<AffineDimExpr>(expr).getPosition();
+          });
```
qedawkins: There is a helper on `AffineMap` for `isIdentity` I think

kuhar:
wait, I think I found a proper fix…

qedawkins (approved):

PR #23792, opened 2026-03-15
[Python] Fix MappedMemory refcount leak in to_host(); add import_host_allocation

jtuyls:
`SimpleNewFromData` previously used `PyBuffer_FillInfo` to associate MappedMemory with the memoryview, but `PyMemoryView_FromBuffer` handles the obj reference differently across Python versions (3.11 double-
decrefs on dealloc, 3.12 does not), so no single cleanup strategy works portably.  Replace with py::keep_alive<0, 1>() on the asarray binding so nanobind manages the lifetime correctly on all versions.

 Also add `HalAllocator.import_host_allocation()` which wraps a Python buffer-protocol object as a HAL buffer via `iree_hal_allocator_import_buffer (zero-copy)`. Rewrite `_copy_to_host()` to use import when the allocator supports it (HIP/CUDA, 1.3-13x faster on HIP) and fall back to a staging buffer otherwise (e.g. Vulkan).

### DIFF
--- a/runtime/bindings/python/hal.cc
+++ b/runtime/bindings/python/hal.cc
@@ -1595,7 +1595,38 @@ void SetupHalBindings(nanobind::module_ m) {
            "object. The buffer is configured as optimal for use on the device "
            "as a transfer buffer. For buffers of unknown providence, this is a "
            "last resort method for making them compatible for transfer to "
-           "arbitrary devices.");
+           "arbitrary devices.")
+      .def(
+          "import_host_allocation",
+          [](HalAllocator& self, py::object buffer_obj, int memory_type,
+             int allowed_usage) -> HalBuffer {
+            Py_buffer view;
+            if (PyObject_GetBuffer(buffer_obj.ptr(), &view,
+                                   PyBUF_WRITABLE | PyBUF_C_CONTIGUOUS) < 0) {
+              throw py::python_error();
+            }
+            PyBufferReleaser view_releaser(view);
+            iree_hal_buffer_params_t params = {0};
+            params.type = (iree_hal_memory_type_t)memory_type;
+            params.usage = (iree_hal_buffer_usage_t)allowed_usage;
+            params.access =
+                IREE_HAL_MEMORY_ACCESS_ALL | IREE_HAL_MEMORY_ACCESS_UNALIGNED;
+            iree_hal_external_buffer_t ext = {};
+            ext.type = IREE_HAL_EXTERNAL_BUFFER_TYPE_HOST_ALLOCATION;
+            ext.size = (iree_device_size_t)view.len;
+            ext.handle.host_allocation.ptr = view.buf;
+            iree_hal_buffer_t* buffer = nullptr;
+            iree_status_t status = iree_hal_allocator_import_buffer(
+                self.raw_ptr(), params, &ext,
+                iree_hal_buffer_release_callback_null(), &buffer);
+            CheckApiStatus(status, "could not import host allocation");
+            return HalBuffer::StealFromRawPtr(buffer);
+          },
+          py::arg("buffer"), py::arg("memory_type"), py::arg("allowed_usage"),
+          py::keep_alive<0, 2>(),
+          "Imports a Python buffer-protocol object (e.g. numpy array) as a HAL "
+          "buffer without copying. The buffer object is kept alive for the "
+          "lifetime of the returned HalBuffer.");
 
   auto hal_buffer = py::class_<HalBuffer>(m, "HalBuffer");
   VmRef::BindRefProtocol(hal_buffer, iree_hal_buffer_type,
@@ -1877,18 +1908,17 @@ void SetupHalBindings(nanobind::module_ m) {
       .def(
           "asarray",
           [](HalMappedMemory* self, py::handle shape, py::object dtype_descr) {
-            py::object py_mapped_memory = py::cast(self);
             size_t rank = py::len(shape);
             intptr_t* dims =
                 static_cast<intptr_t*>(alloca(sizeof(intptr_t) * rank));
             for (size_t i = 0; i < rank; ++i) {
               dims[i] = py::cast<intptr_t>(shape[i]);
             }
-            return numpy::SimpleNewFromData(rank, dims, dtype_descr,
-                                            self->mapped_memory().contents.data,
-                                            py_mapped_memory);
+            return numpy::SimpleNewFromData(
+                rank, dims, dtype_descr, self->mapped_memory().contents.data);
           },
-          py::arg("shape"), py::arg("numpy_dtype_descr"));
+          py::arg("shape"), py::arg("numpy_dtype_descr"),
+          py::keep_alive<0, 1>());
 
   py::class_<HalExternalTimepoint>(m, "HalExternalTimepoint")
       .def(
--- a/runtime/bindings/python/iree/runtime/array_interop.py
+++ b/runtime/bindings/python/iree/runtime/array_interop.py
@@ -12,6 +12,7 @@ import numpy as np
 import numpy.lib.mixins
 
 from ._binding import (
+    BufferCompatibility,
     BufferUsage,
     HalBufferView,
     HalDevice,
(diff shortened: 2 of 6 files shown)


runtime/bindings/python/iree/runtime/array_interop.py
```diff
@@ -152,32 +152,31 @@ def _map_to_host(self) -> Tuple[MappedMemory, np.ndarray]:
...
-        source_buffer = self._buffer_view.get_buffer()
-        host_buffer = self._device.allocator.allocate_buffer(
-            memory_type=(MemoryType.HOST_LOCAL | MemoryType.DEVICE_VISIBLE),
-            allowed_usage=(BufferUsage.TRANSFER_TARGET | BufferUsage.MAPPING_SCOPED),
-            allocation_size=source_buffer.byte_length(),
+        raw_dtype = self._get_raw_dtype()
+        # Allocate numpy-owned memory and import it as a HAL buffer so the
+        # device can DMA directly into it (no staging buffer needed).
+        result = np.empty(self._buffer_view.shape, dtype=raw_dtype)
+        host_hal_buffer = self._device.allocator.import_host_allocation(
```
kuhar: I don't think this is universally supported. For example, it fails on vulkan with something like:
```py
import numpy as np
import iree.runtime as rt

config = rt.Config("vulkan")
device = config.device

data = np.arange(12, dtype=np.float32).reshape(3, 4)
da = rt.asdevicearray(device, data)

result = da._copy_to_host()
np.testing.assert_array_equal(result, data)
print("OK")
```

```
  File "/home/jakub/iree/review/issue.py", line 16, in <module>
    result = da._copy_to_host()
             ^^^^^^^^^^^^^^^^^^
  File "/home/jakub/iree/review/build/runtime/bindings/python/iree/runtime/array_interop.py", line 159, in _copy_to_host
    host_hal_buffer = self._device.allocator.import_host_allocation(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not import host allocation: ../iree/runtime/src/iree/hal/drivers/vulkan/native_allocator.cc:644: INVALID_ARGUMENT; cannot import mapped memory; stack:
```

kuhar: You should be able to query this with `self._device.allocator.query_buffer_compatibility(...)`

jtuyls: Thanks for catching! I updated it to use `alloc.query_buffer_compatibility` now and fallback to an IREE allocated buffer if importing the numpy allocated buffer isn't supported.

runtime/bindings/python/hal.cc
```diff
@@ -1595,7 +1595,38 @@ void SetupHalBindings(nanobind::module_ m) {
...
+                IREE_HAL_MEMORY_ACCESS_ALL | IREE_HAL_MEMORY_ACCESS_UNALIGNED;
+            iree_hal_external_buffer_t ext = {};
+            ext.type = IREE_HAL_EXTERNAL_BUFFER_TYPE_HOST_ALLOCATION;
+            ext.size = (iree_device_size_t)view.len;
+            ext.handle.host_allocation.ptr = view.buf;
+            iree_hal_buffer_t* buffer = nullptr;
+            iree_status_t status = iree_hal_allocator_import_buffer(
+                self.raw_ptr(), params, &ext,
+                iree_hal_buffer_release_callback_null(), &buffer);
+            PyBuffer_Release(&view);
```
kuhar: Could we use `PyBufferReleaser` (RAII-based)?

runtime/bindings/python/iree/runtime/array_interop.py
```diff
@@ -152,32 +153,52 @@ def _map_to_host(self) -> Tuple[MappedMemory, np.ndarray]:
...
+            allocation_size=source.byte_length(),
+        )
+        if compat & int(BufferCompatibility.IMPORTABLE):
+            result = np.empty(self._buffer_view.shape, dtype=raw_dtype)
+            host_buf = alloc.import_host_allocation(
+                result,
+                memory_type=mem_type,
+                allowed_usage=usage,
+            )
+            dma_and_wait(host_buf)
```
kuhar, line 188: This doesn't really need `MAPING_SCOPED`, no?

runtime/bindings/python/tests/hal_test.py
```diff
@@ -204,6 +205,52 @@ def testAllocateHostStagingBufferCopy(self):
...
+        del arr
+        after = sys.getrefcount(mapped)
+
+        self.assertEqual(
+            after,
+            before,
+            msg=f"MappedMemory refcount changed by {after - before} after "
+            f"asarray()+del; likely missing PyBuffer_Release in numpy_interop.cc",
+        )
+
```
kuhar, line 284: Should we also have a testcase that checks that zero copy is actually zero copy? E.g., DMA a known pattern through the imported HAL buffer and check that the numpy array sees it.

runtime/bindings/python/tests/hal_test.py
```diff
@@ -204,6 +205,52 @@ def testAllocateHostStagingBufferCopy(self):
...
+        self.assertEqual(buffer.byte_length(), data.nbytes)
+        del buffer
+        np.testing.assert_array_equal(data, np.zeros(16, dtype=np.float32))
+
+    def testAsarrayRefcountBalanced(self):
+        """MappedMemory.asarray() must not leak a reference on MappedMemory.
+
+        Without PyBuffer_Release after PyMemoryView_FromBuffer in
+        numpy_interop.cc, PyBuffer_FillInfo's Py_INCREF is never balanced,
+        leaking MappedMemory until interpreter shutdown.
```
kuhar: I don't think the code uses FillInfo anymore, why do we reference it here?

runtime/bindings/python/tests/hal_test.py
```diff
@@ -204,6 +205,52 @@ def testAllocateHostStagingBufferCopy(self):
...
+        buf_view = iree.runtime.asdevicearray(self.device, data)
+
+        mapped = buf_view._buffer_view.map()
+        before = sys.getrefcount(mapped)
+        arr = mapped.asarray(buf_view.shape, np.dtype("float32"))
+        # Delete arr so the memoryview and its managed buffer are fully
+        # deallocated.  Only a leaked PyBuffer_FillInfo ref (the bug)
+        # would survive.  Measuring after del makes the assertion
+        # independent of Python-version differences in how
+        # PyMemoryView_FromBuffer manages internal references.
```
kuhar, line 274: Don't use double spaces between sentences, most style guides discourage it

runtime/bindings/python/iree/runtime/array_interop.py
AWoloszyn: In fact, if we never use iree_hal_buffer_map_range/iree_hal_buffer_unmap_range on this buffer we are not allowed reading it from host at all. If we expect this memory to be up to date on the host as soon as the write its done, IREE_HAL_BUFFER_USAGE_MAPPING_PERSISTENT is what we are looking for.

AWoloszyn:
Overall, the C++/Hal side seems correct (minus the mapping question). I am less familiar with the python side, so I will rely on @kuhar's comments there.

runtime/bindings/python/hal.cc
jtuyls: Yes, updated

runtime/bindings/python/iree/runtime/array_interop.py
jtuyls: Thanks for catching! Both should be addressed now.

runtime/bindings/python/tests/hal_test.py
jtuyls: Yes, added it here: https://github.com/iree-org/iree/pull/23792/changes#diff-d15b11fe855d3c8f3a470a2d51da7505663283325c5474c53f71e1b55828cdabR226

runtime/bindings/python/tests/hal_test.py
jtuyls: It was a stale comment, updated it.

kuhar (approved):

PR #23793, opened 2026-03-16
[Codegen][GPU] Add DistributeArgCompare pattern

bangtianliu:
This PR adds the `DistributeArgCompare `pattern to distribute `iree_vector_ext.arg_compare` operations across GPU threads and subgroups.
 
For supported comparators, we use a ballot-based approach that leverages `gpu.subgroup_reduce` + `gpu.ballot` for reduction.
Supported comparators include:
1. Direct comparison on values (e.g., arith.cmpf ogt for argmax)
2. Same unary op applied to both arguments before comparison (e.g., math.absf for argmax of absolute values)

Unsupported comparators fall back to the portable butterfly shuffle approach. Currently, this is mainly used for argmax/argmin operations, but we can extend support for additional comparators as needed.

Issue: #23005

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -21,7 +21,9 @@
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/Utils.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/GPU/Utils/GPUUtils.h"
+#include "mlir/Dialect/Math/IR/Math.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/Dialect/Vector/Utils/VectorUtils.h"
@@ -936,6 +938,233 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
   return layout.getThreadTile()[dim];
 }
 
+/// Computes the distributed shape after partial reduction, with reduction
+/// dimensions set to 1 across all tile groups (batch, outer, element).
+static SmallVector<int64_t>
+getPartialReducedDistributedShape(NestedLayoutAttr srcLayout,
+                                  ArrayRef<int64_t> reductionDims) {
+  int64_t rank = srcLayout.getRank();
+  SmallVector<int64_t> shape = srcLayout.getDistributedShape();
+  // Iterate over 3 tile groups: batch, outer, element.
+  for (int64_t tileGroupIdx : llvm::seq<int64_t>(3)) {
+    int64_t tileGroupOffset = tileGroupIdx * rank;
+    for (int64_t rDim : reductionDims) {
+      shape[tileGroupOffset + rDim] = 1;
+    }
+  }
+  return shape;
+}
+
+/// Computes the undistributed shape for partial reduction results, where
+/// all tiles except subgroup are reduced.
+static SmallVector<int64_t>
+getPartialReductionShape(NestedLayoutAttr srcLayout, ArrayRef<int64_t> srcShape,
+                         ArrayRef<int64_t> reductionDims) {
+  SmallVector<int64_t> preDistrShape = srcLayout.getUndistributedPackedShape();
+  SmallVector<int64_t> shape = llvm::to_vector(srcShape);
+  for (int64_t rDim : reductionDims) {
+    shape[rDim] = preDistrShape[rDim];
+  }
+  return shape;
+}
+
+/// Reshapes flat 1-d vectors back to target types. For 0-d vectors,
+/// uses extract+broadcast since shape_cast to 0-d is not supported.
+static std::pair<VectorValue, VectorValue>
+reshapeFlatToTarget(RewriterBase &rewriter, Location loc, VectorValue flatValue,
+                    VectorValue flatIndex, VectorType targetValueType,
+                    VectorType targetIndexType) {
+  if (targetValueType.getRank() == 0) {
+    Value scalarVal = vector::ExtractOp::create(rewriter, loc, flatValue,
+                                                ArrayRef<int64_t>{0});
+    VectorValue reshapedValue =
+        vector::BroadcastOp::create(rewriter, loc, targetValueType, scalarVal);
+    Value scalarIdx = vector::ExtractOp::create(rewriter, loc, flatIndex,
+                                                ArrayRef<int64_t>{0});
+    VectorValue reshapedIndex =
+        vector::BroadcastOp::create(rewriter, loc, targetIndexType, scalarIdx);
+    return {reshapedValue, reshapedIndex};
+  }
+  VectorValue reshapedValue =
+      vector::ShapeCastOp::create(rewriter, loc, targetValueType, flatValue);
+  VectorValue reshapedIndex =
+      vector::ShapeCastOp::create(rewriter, loc, targetIndexType, flatIndex);
+  return {reshapedValue, reshapedIndex};
+}
+
+static LogicalResult checkBitwidthForShuffle(Operation *op, Type type,
+                                             int64_t maxBitsPerShuffle,
+                                             StringRef typeName,
+                                             PatternRewriter &rewriter) {
+  unsigned bitwidth = type.getIntOrFloatBitWidth();
+  if (bitwidth > maxBitsPerShuffle) {
+    return rewriter.notifyMatchFailure(
+        op, llvm::formatv("{0} bitwidth {1} greater than maxBitsPerShuffle {2}",
+                          typeName, bitwidth, maxBitsPerShuffle));
+  }
+  return success();
+}
+
+/// Creates an equality comparison operation for the given values.
+/// Returns arith.cmpf for floating-point types and arith.cmpi for integers.
+static Value createEqualityComparison(RewriterBase &rewriter, Location loc,
+                                      Value lhs, Value rhs) {
+  if (isa<FloatType>(lhs.getType())) {
+    return arith::CmpFOp::create(rewriter, loc, arith::CmpFPredicate::OEQ, lhs,
+                                 rhs);
+  }
+  return arith::CmpIOp::create(rewriter, loc, arith::CmpIPredicate::eq, lhs,
+                               rhs);
+}
+
+/// Result of analyzing a comparator region for ballot-based reduction.
+struct ComparatorAnalysis {
+  gpu::AllReduceOperation reduceOp;
+  /// The unary transformation operation applied to both comparison operands,
+  /// or nullptr if comparing block arguments directly.
+  Operation *transformOp = nullptr;
+};
+
+/// Maps a floating-point comparison predicate to the corresponding
+/// gpu::AllReduceOperation for reduction.
+static std::optional<gpu::AllReduceOperation>
+mapFCmpPredicateToReduceOp(arith::CmpFPredicate pred) {
+  switch (pred) {
+  case arith::CmpFPredicate::OGT:
+  case arith::CmpFPredicate::OGE:
+  case arith::CmpFPredicate::UGT:
+  case arith::CmpFPredicate::UGE:
+    return gpu::AllReduceOperation::MAXNUMF;
+  case arith::CmpFPredicate::OLT:
+  case arith::CmpFPredicate::OLE:
+  case arith::CmpFPredicate::ULT:
+  case arith::CmpFPredicate::ULE:
+    return gpu::AllReduceOperation::MINNUMF;
+  default:
+    return std::nullopt;
+  }
+}
+
+/// Maps an integer comparison predicate to the corresponding
+/// gpu::AllReduceOperation for reduction.
+static std::optional<gpu::AllReduceOperation>
+mapICmpPredicateToReduceOp(arith::CmpIPredicate pred) {
+  switch (pred) {
+  case arith::CmpIPredicate::sgt:
+  case arith::CmpIPredicate::sge:
+    return gpu::AllReduceOperation::MAXSI;
+  case arith::CmpIPredicate::slt:
+  case arith::CmpIPredicate::sle:
+    return gpu::AllReduceOperation::MINSI;
+  case arith::CmpIPredicate::ugt:
+  case arith::CmpIPredicate::uge:
+    return gpu::AllReduceOperation::MAXUI;
+  case arith::CmpIPredicate::ult:
+  case arith::CmpIPredicate::ule:
+    return gpu::AllReduceOperation::MINUI;
+  default:
+    return std::nullopt;
+  }
+}
+
+/// Analyzes the comparator region of an arg_compare operation to determine
+/// if it can use the efficient ballot-based reduction approach.
+///
+/// This function detects two patterns:
+/// 1. Simple comparison: direct comparison on block arguments
+///    Example: arith.cmpf ogt, %lhs, %rhs
+///
+/// 2. Transformed comparison: same unary transformation applied to both
+///    arguments before comparison
+///    Example: arith.cmpf ogt, (math.absf %lhs), (math.absf %rhs)
+///
+/// Returns the ComparatorAnalysis if the pattern is detected, or std::nullopt
+/// for comparators that require the shuffle-based fallback.
+static std::optional<ComparatorAnalysis>
+analyzeComparatorForSubgroupReduce(Region &comparatorRegion) {
+  Block &block = comparatorRegion.front();
+
+  // Get the yield operation. The verifier guarantees:
+  // - Terminator is YieldOp (SingleBlockImplicitTerminator trait)
+  // - YieldOp has exactly 1 operand of type i1
+  auto yieldOp = cast<IREE::VectorExt::YieldOp>(block.getTerminator());
+  Value yieldedValue = yieldOp.getValues()[0];
+
+  // The yielded value must have a defining op (not a block argument).
+  Operation *cmpOp = yieldedValue.getDefiningOp();
+  if (!cmpOp) {
+    return std::nullopt;
+  }
+
+  Value blockArg0 = block.getArgument(0);
+  Value blockArg1 = block.getArgument(1);
+
+  // Helper lambda to analyze comparison operands and build ComparatorAnalysis.
+  // Works for both floating-point and integer comparisons.
+  auto analyzeComparisonOperands =
+      [&](Value lhs, Value rhs, std::optional<gpu::AllReduceOperation> reduceOp)
+      -> std::optional<ComparatorAnalysis> {
+    if (!reduceOp) {
+      return std::nullopt;
+    }
+
+    // Case 1: Direct comparison on block arguments.
+    if (lhs == blockArg0 && rhs == blockArg1) {
+      return ComparatorAnalysis{*reduceOp, /*transformOp=*/nullptr};
+    }
+
+    // Case 2: Same unary transformation applied to both arguments.
+    // Check if both operands come from the same type of unary op applied
+    // to the block arguments.
+    Operation *lhsOp = lhs.getDefiningOp();
+    Operation *rhsOp = rhs.getDefiningOp();
+    if (lhsOp && rhsOp && lhsOp->getName() == rhsOp->getName() &&
+        lhsOp->getNumOperands() == 1 && lhsOp->getNumResults() == 1 &&
+        rhsOp->getNumOperands() == 1 && rhsOp->getNumResults() == 1 &&
+        lhsOp->getOperand(0) == blockArg0 &&
+        rhsOp->getOperand(0) == blockArg1) {
+      return ComparatorAnalysis{*reduceOp, /*transformOp=*/lhsOp};
+    }
+
+    return std::nullopt;
+  };
+
+  // Check for floating-point comparison.
+  if (auto cmpfOp = dyn_cast<arith::CmpFOp>(cmpOp)) {
+    auto reduceOp = mapFCmpPredicateToReduceOp(cmpfOp.getPredicate());
+    return analyzeComparisonOperands(cmpfOp.getLhs(), cmpfOp.getRhs(),
+                                     reduceOp);
+  }
+
+  // Check for integer comparison.
+  if (auto cmpiOp = dyn_cast<arith::CmpIOp>(cmpOp)) {
+    auto reduceOp = mapICmpPredicateToReduceOp(cmpiOp.getPredicate());
+    return analyzeComparisonOperands(cmpiOp.getLhs(), cmpiOp.getRhs(),
+                                     reduceOp);
+  }
+
+  return std::nullopt;
+}
+
+/// Clones the body of a comparator region, mapping block arguments to the
+/// given lhs/rhs values. Returns the yielded i1 comparison result.
+static Value cloneComparatorRegion(RewriterBase &rewriter, Region &region,
+                                   Value lhs, Value rhs) {
+  Block &block = region.front();
+  IRMapping mapper;
+  mapper.map(block.getArgument(0), lhs);
+  mapper.map(block.getArgument(1), rhs);
+  for (Operation &op : block.without_terminator()) {
+    Operation *clonedOp = rewriter.clone(op, mapper);
+    for (const auto &[origResult, clonedResult] :
+         llvm::zip_equal(op.getResults(), clonedOp->getResults())) {
+      mapper.map(origResult, clonedResult);
+    }
+  }
+  auto yieldOp = cast<IREE::VectorExt::YieldOp>(block.getTerminator());
+  return mapper.lookup(yieldOp.getValues()[0]);
+}
+
 /// The lowering for multi_reduction is done in two steps:
 ///   1. Local Reduce: Each thread reduces all elements carried by it along
 ///      the reduction dimensions. This is the batch, outer and element dims.
@@ -946,6 +1175,93 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
 ///   4. Subgroup reduce : each subgroup will store the partial reductions
 ///      to shared memory and will be reloaded into a layout where partial
 ///      reductions will be placed inside threads.
+
+/// Computes the layout for reading reduction results from shared memory.
+/// Shared by DistributeMultiReduction and DistributeArgCompare.
+static NestedLayoutAttr
+computeLayoutForReductionFromBuffer(NestedLayoutAttr srcLayout,
+                                    ArrayRef<int64_t> reductionDims) {
+  auto subgroupTileLens =
+      llvm::to_vector_of<int64_t>(srcLayout.getSubgroupTile());
+  auto batchTileLens = llvm::to_vector_of<int64_t>(srcLayout.getBatchTile());
+  auto outerTileLens = llvm::to_vector_of<int64_t>(srcLayout.getOuterTile());
+  auto threadTileLens = llvm::to_vector_of<int64_t>(srcLayout.getThreadTile());
+  auto elementTileLens =
+      llvm::to_vector_of<int64_t>(srcLayout.getElementTile());
+  auto subgroupStrides =
+      llvm::to_vector_of<int64_t>(srcLayout.getSubgroupStrides());
+  auto threadStrides =
+      llvm::to_vector_of<int64_t>(srcLayout.getThreadStrides());
+
+  int64_t threadsRequired = 1;
+  for (int64_t rDim : reductionDims) {
+    threadsRequired *= llvm::PowerOf2Ceil(subgroupTileLens[rDim]);
+  }
+
+  std::optional<int64_t> availableThreads;
+  int64_t threadStride = 0;
+  for (int64_t rDim : reductionDims) {
+    if (threadTileLens[rDim] >= threadsRequired) {
+      availableThreads = threadTileLens[rDim];
+      threadStride = threadStrides[rDim];
+      break;
+    }
+  }
+
+  for (int64_t rDim : reductionDims) {
+    batchTileLens[rDim] = 1;
+    outerTileLens[rDim] = 1;
+    elementTileLens[rDim] = 1;
+    if (availableThreads.has_value()) {
+      int64_t used = llvm::PowerOf2Ceil(subgroupTileLens[rDim]);
+      threadStrides[rDim] = threadStride;
+      threadTileLens[rDim] = used;
+      availableThreads.value() /= used;
+      threadStride *= used;
+    } else {
+      threadStrides[rDim] = 0;
+      threadTileLens[rDim] = 1;
+    }
+    subgroupTileLens[rDim] = 1;
+    subgroupStrides[rDim] = 0;
+  }
+
+  return IREE::VectorExt::NestedLayoutAttr::get(
+      srcLayout.getContext(), subgroupTileLens, batchTileLens, outerTileLens,
+      threadTileLens, elementTileLens, subgroupStrides, threadStrides);
+}
+
+/// Computes the inter-subgroup write layout by replacing reduced tiles with
+/// unit dimensions. Shared by DistributeMultiReduction and
+/// DistributeArgCompare.
+static NestedLayoutAttr
+computeInterSubgroupWriteLayout(NestedLayoutAttr srcLayout,
+                                ArrayRef<int64_t> reductionDims) {
+  auto subgroupTileLens =
+      llvm::to_vector_of<int64_t>(srcLayout.getSubgroupTile());
+  auto batchTileLens = llvm::to_vector_of<int64_t>(srcLayout.getBatchTile());
+  auto outerTileLens = llvm::to_vector_of<int64_t>(srcLayout.getOuterTile());
+  auto threadTileLens = llvm::to_vector_of<int64_t>(srcLayout.getThreadTile());
+  auto elementTileLens =
+      llvm::to_vector_of<int64_t>(srcLayout.getElementTile());
+  auto subgroupStrides =
+      llvm::to_vector_of<int64_t>(srcLayout.getSubgroupStrides());
+  auto threadStrides =
+      llvm::to_vector_of<int64_t>(srcLayout.getThreadStrides());
+
+  for (int64_t rDim : reductionDims) {
+    batchTileLens[rDim] = 1;
+    outerTileLens[rDim] = 1;
+    threadTileLens[rDim] = 1;
+    elementTileLens[rDim] = 1;
+    threadStrides[rDim] = 0;
+  }
+
+  return IREE::VectorExt::NestedLayoutAttr::get(
+      srcLayout.getContext(), subgroupTileLens, batchTileLens, outerTileLens,
+      threadTileLens, elementTileLens, subgroupStrides, threadStrides);
+}
+
 struct DistributeMultiReduction final
     : MaskedOpDistributionPattern<vector::MultiDimReductionOp> {
   using MaskedOpDistributionPattern::MaskedOpDistributionPattern;
@@ -975,12 +1291,9 @@ struct DistributeMultiReduction final
     }
 
     Type elemTy = srcVector.getType().getElementType();
-    unsigned elemBitwidth = elemTy.getIntOrFloatBitWidth();
-    if (elemBitwidth > maxBitsPerShuffle) {
-      return rewriter.notifyMatchFailure(
-          multiReduceOp,
-          llvm::formatv("element bitwidth greater than maxBitsPerShuffle",
-                        elemBitwidth, maxBitsPerShuffle));
+    if (failed(checkBitwidthForShuffle(multiReduceOp, elemTy, maxBitsPerShuffle,
+                                       "element", rewriter))) {
+      return failure();
     }
 
     VectorValue disSrc =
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+          argCompareOp, "expected nested layout attr for input value");
+    }
+
+    Type elemTy = inputValue.getType().getElementType();
+    if (failed(checkBitwidthForShuffle(argCompareOp, elemTy, maxBitsPerShuffle,
+                                       "element", rewriter))) {
+      return failure();
+    }
+
+    Type indexElemTy = initIndex.getType().getElementType();
```
sommerlukas: Nit: I think it would be nicer to keep the same order of checks for `initIndex` (first layout, then bitwidth).

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+    constexpr int64_t kNumTileGroups = 3; // Batch, outer, element.
+    int64_t distributedReductionDim = 2 * rank + reductionDim;
+
+    // Distribute init values with their natural layout (gives rank 3 for 1D).
+    // Then insert unit dimensions to expand to rank 5.
+    //
+    // For a 2D input with reduction on dim 1:
+    //   Distributed input: [batch0, batch1, outer0, outer1, elem0, elem1] =
+    //   rank 6 Distributed init (1D): [batch0, outer0, elem0] = rank 3 Need to
+    //   expand to: [batch0, 1, outer0, 1, elem0] = rank 5 (inserting unit dims
```
sommerlukas: Why do we only need to expand to rank 5 and not rank 6 here?

sommerlukas: Is this comment duplicated with the one below?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+          if (tile < 2) {
+            // Insert unit dim for shape cast, then broadcast to tile size.
+            intermediateShape.push_back(1);
+            int64_t tileSize = (tile == 0)
+                                   ? valueLayout.getBatchTile()[reductionDim]
+                                   : valueLayout.getOuterTile()[reductionDim];
+            broadcastShape.push_back(tileSize);
+          }
+          // Skip element[reductionDim] - this is the dimension being reduced.
+        } else {
```
sommerlukas: Use `continue` instead of `else`.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+    // TODO(Bangtian): Implement masked arg_compare distribution.
+    // This requires:
+    // 1. Distributing the mask with the same layout as the input value
+    // 2. Applying mask to select between distributed input and init values
+    // 3. Ensuring masked lanes don't participate in thread/subgroup reductions
+    // 4. Proper handling of mask in ballot optimization (exclude masked lanes)
+    if (maskOp) {
+      return rewriter.notifyMatchFailure(
+          argCompareOp, "masked arg_compare distribution not yet implemented");
+    }
```
sommerlukas: I think we should check invariants for this pattern before creating IR in the step before.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+    // element) it holds along the reduction dimension.
+    // The distributed reduction mask is the same mask appended thrice (for
+    // batch, outer, element tiles).
+    SmallVector<bool> reductionMask(rank, false);
+    reductionMask[reductionDim] = true;
+    SmallVector<bool> distributedReductionMask;
+    distributedReductionMask.reserve(kNumTileGroups * rank);
+    for (int64_t tileIdx = 0; tileIdx < kNumTileGroups; ++tileIdx) {
+      distributedReductionMask.append(reductionMask.begin(),
+                                      reductionMask.end());
```
sommerlukas: Nit: Use `llvm::append_range`.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+                                       localMapper);
+
+    VectorValue localValueResult =
+        cast<VectorValue>(localArgCompare.getResultValue());
+    VectorValue localIndexResult =
+        cast<VectorValue>(localArgCompare.getResultIndex());
+
+    // Collapse unit dimensions inserted for reduction.
+    // Example: [batch0, batch1, outer0, outer1, elem0] -> [batch0, outer0,
+    // elem0]
```
sommerlukas: Is this comment accurate? Shouldn't this be `1` instead of `batch1` and `outer1`?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+      //   before comparison (e.g., argmax of absolute values)
+      auto analysis =
+          analyzeComparatorForSubgroupReduce(argCompareOp.getRegion());
+      if (analysis.has_value()) {
+        // Use gpu.subgroup_reduce + gpu.ballot for efficient reduction.
+        result = doThreadReductionWithBallot(rewriter, loc, valueLayout,
+                                             localValueResult, localIndexResult,
+                                             *analysis, reductionDim);
+      }
+      if (!analysis.has_value()) {
```
sommerlukas:
```suggestion
      } else {
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,775 @@ struct DistributeMultiReduction final
...
+    rewriter.replaceOp(argCompareOp,
+                       {subgroupReduced.first, subgroupReduced.second});
+    return success();
+  }
+
+private:
+  /// Performs thread-level arg_compare reduction using portable butterfly
+  /// shuffles that work across all GPU targets.
+  FailureOr<std::pair<VectorValue, VectorValue>>
+  doThreadReduction(RewriterBase &rewriter, Location loc,
```
sommerlukas: Nit/personal taste: `doCrossThreadReduction` might be more descriptive for the purpose of this function.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution_argcompare.mlir
sommerlukas, line 1: The tests don't capture the dataflow between the operations, which could lead to them producing false matches and also makes it hard to follow the expected output.

sommerlukas (commented):
Some comments on style and tests. For the distribution logic, it would be good to also get @Groverkss's eyes on this.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: Because element[reductionDim] is the dimension we're about to reduce, I will update the comment to make it clear.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: The current name `doThreadReduction `is consistent with the naming in `DistributeMultiReduction `which also uses similar terminology.

bangtianliu:
> Some comments on style and tests. For the distribution logic, it would be good to also get @Groverkss's eyes on this.

Sure, thanks for your time reviewing this PR.

bangtianliu:
cc @Groverkss for review

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -936,6 +938,160 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
...
+                          typeName, bitwidth, maxBitsPerShuffle));
+  }
+  return success();
+}
+
+/// Creates an equality comparison operation for the given values.
+/// Returns arith.cmpf for floating-point types and arith.cmpi for integers.
+static Value createEqualityComparison(RewriterBase &rewriter, Location loc,
+                                      Value lhs, Value rhs) {
+  if (isa<FloatType>(lhs.getType())) {
```
sommerlukas, line 1002: Should we assert here that `lhs` and `rhs` have the same type?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -936,6 +938,160 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
...
+analyzeComparatorForSubgroupReduce(Region &comparatorRegion) {
+  Block &block = comparatorRegion.front();
+
+  // Get the yield operation. The verifier guarantees:
+  // - Terminator is YieldOp (SingleBlockImplicitTerminator trait)
+  // - YieldOp has exactly 1 operand of type i1
+  auto yieldOp = cast<IREE::VectorExt::YieldOp>(block.getTerminator());
+  Value yieldedValue = yieldOp.getValues()[0];
+
+  // The yielded value should come from a comparison operation.
```
sommerlukas: Nit: The comment isn't quite accurate, as we don't check for a comparison, we only check that there is a defining operation.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,751 @@ struct DistributeMultiReduction final
...
+    // 1. Shape cast to insert unit dims: [2,1,1] -> [2,1,1,1,1]
+    // 2. Broadcast to expand unit dims: [2,1,1,1,1] -> [2,2,1,1,1]
+    VectorValue disInitValue3 =
+        getDistributed(rewriter, initValue, initValueLayout);
+    VectorValue disInitIndex3 =
+        getDistributed(rewriter, initIndex, initIndexLayout);
+
+    ArrayRef<int64_t> initValueShape = disInitValue3.getType().getShape();
+    SmallVector<int64_t> intermediateShape; // With unit dims inserted.
+    SmallVector<int64_t> broadcastShape;    // Final shape for broadcast.
```
sommerlukas: Do we need to construct this or can we just take it from the other operands distributed shape (and drop the reduction dim if necessary)?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,751 @@ struct DistributeMultiReduction final
...
+                                    disInitValue3)
+            .getResult());
+    VectorValue shapeCastIndex = cast<VectorValue>(
+        vector::ShapeCastOp::create(rewriter, loc, intermediateIndexType,
+                                    disInitIndex3)
+            .getResult());
+
+    // Step 2: Broadcast to final shape (expands unit dims to tile sizes).
+    VectorType broadcastValueType = VectorType::get(
+        broadcastShape, disInitValue3.getType().getElementType());
```
sommerlukas: Nit: We repeatedly use `disInitValue3.getType().getElementType()`, maybe makes sense to store it in some variable.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,751 @@ struct DistributeMultiReduction final
...
+      if (analysis.transformOp) {
+        // Must broadcast original value from winning lane.
+        resultValue =
+            gpu::ShuffleOp::create(rewriter, loc, localValue, winningLane,
+                                   subgroupSizeVal, gpu::ShuffleMode::IDX)
+                .getShuffleResult();
+      } else {
+        // For simple comparators, reducedValue is already the correct result.
+        resultValue = reducedValue;
+      }
```
sommerlukas:
```suggestion
      // For simple comparators, reducedValue is already the correct result.
      Value resultValue = reducedValue;
      if (analysis.transformOp) {
        // Must broadcast original value from winning lane.
        resultValue =
            gpu::ShuffleOp::create(rewriter, loc, localValue, winningLane,
                                   subgroupSizeVal, gpu::ShuffleMode::IDX)
                .getShuffleResult();
      } 
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1530,751 @@ struct DistributeMultiReduction final
...
+  doSubgroupReduction(RewriterBase &rewriter, Location loc,
+                      VectorValue srcVector, NestedLayoutAttr srcLayout,
+                      int64_t reductionDim, VectorValue threadReducedValue,
+                      VectorValue threadReducedIndex, Region &comparatorRegion,
+                      VectorValue initValue, VectorValue initIndex,
+                      VectorLayoutInterface resLayout) const {
+    int64_t rank = srcLayout.getRank();
+
+    // Re-insert reduction dimensions to maintain rank (similar to
+    // multi_reduction).
```
sommerlukas: If it's similar to `multi_reduction`, can we share any helpers?

sommerlukas (commented):
Some more nits and questions.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: the `arith.cmpf` and `arith.cmpi` verifiers will catch any type mismatch.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -938,6 +938,70 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
...
+}
+
+/// Unwraps a ToSIMD operation if present. For 0-d vectors (scalars), ToSIMD
+/// ops are identity operations that would otherwise leave orphaned ops in IR.
+static VectorValue unwrapToSIMD(VectorValue val) {
+  if (auto toSIMD = val.getDefiningOp<IREE::VectorExt::ToSIMDOp>()) {
+    return cast<VectorValue>(toSIMD.getInput());
+  }
+  return val;
+}
```
sommerlukas: Is this the best approach to solve this issue? If the `ToSIMD` is no-op/identity in those cases, can't we have a pattern that removes them instead of putting the burden on potentially multiple different distribution patterns?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -938,6 +938,70 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
...
+  if (targetValueType.getRank() == 0) {
+    Value scalarVal = vector::ExtractOp::create(rewriter, loc, flatValue,
+                                                ArrayRef<int64_t>{0});
+    reshapedValue =
+        vector::BroadcastOp::create(rewriter, loc, targetValueType, scalarVal);
+    Value scalarIdx = vector::ExtractOp::create(rewriter, loc, flatIndex,
+                                                ArrayRef<int64_t>{0});
+    reshapedIndex =
+        vector::BroadcastOp::create(rewriter, loc, targetIndexType, scalarIdx);
+  } else {
```
sommerlukas: Use early return instead of `else`.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1616,8 +1663,38 @@ struct DistributeArgCompare final
         return rewriter.notifyMatchFailure(
             argCompareOp, "expected nested layout attr for input index");
       }
-      disIndex =
-          getDistributed(rewriter, inputIndexVec, signature[inputIndexVec]);
+      disIndex = cast<VectorValue>(
+          getDistributed(rewriter, inputIndexVec, signature[inputIndexVec]));
+    } else {
+      // Implicit-index mode: generate an iota index vector for the distributed
```
sommerlukas: Could we insert a `vector.step` operation here instead and distribute that afterwards?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1578,935 @@ struct DistributeMultiReduction final
...
+      // Collapse unit dimensions inserted for reduction.
+      localValueResult = cast<VectorValue>(
+          vector::ShapeCastOp::create(rewriter, loc, resultValueDistType,
+                                      localReduced->first)
+              .getResult());
+      localIndexResult = cast<VectorValue>(
+          vector::ShapeCastOp::create(rewriter, loc, resultIndexDistType,
+                                      localReduced->second)
+              .getResult());
+    }
```
sommerlukas:
```suggestion
    VectorValue localValueResult = disValue;
    VectorValue localIndexResult = disIndex;
    if(elementTile > 1) {
      auto localReduced = doLocalArgCompareReduction(
          rewriter, loc, disValue, cast<VectorValue>(disIndex), disInitValue,
          disInitIndex, argCompareOp.getRegion(), distributedReductionDim);
      if (failed(localReduced)) {
        return rewriter.notifyMatchFailure(
            argCompareOp,
            "failed to perform local element reduction for arg_compare");
      }
      localValueResult = localReduced->first;
      localIndexResult = localReduced->second;
    }
    // Collapse unit dimensions inserted for reduction.
    localValueResult = cast<VectorValue>(
        vector::ShapeCastOp::create(rewriter, loc, resultValueDistType,
                                    localValueResult)
            .getResult());
    localIndexResult = cast<VectorValue>(
        vector::ShapeCastOp::create(rewriter, loc, resultIndexDistType,
                                    localIndexResult)
            .getResult());
```

sommerlukas: You could also move the special treatment for `elementTile == 1` into `doLocalArgCompareReduction` instead.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -2119,18 +2301,48 @@ struct DistributeArgCompare final
...
-    vector::TransferWriteOp::create(rewriter, loc, valueToWrite, valueBuffer,
-                                    indices, inBounds);
-    vector::TransferWriteOp::create(rewriter, loc, indexToWrite, indexBuffer,
-                                    indices, inBounds);
+    auto valueWrite = vector::TransferWriteOp::create(
+        rewriter, loc, valueToWrite, valueBuffer, indices, inBounds);
+    auto indexWrite = vector::TransferWriteOp::create(
+        rewriter, loc, indexToWrite, indexBuffer, indices, inBounds);
+
+    // Compute inter-subgroup write layout (same as DistributeMultiReduction).
```
sommerlukas: It it's the same as in `DistributeMultiReduction`, can we have a shared helper?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1578,935 @@ struct DistributeMultiReduction final
...
+    }
+    setSignatureForRedistribution(rewriter, secondArgCompare.getOperation(),
+                                  inputLayouts, resultLayouts);
+
+    return {cast<VectorValue>(secondArgCompare.getResultValue()),
+            cast<VectorValue>(secondArgCompare.getResultIndex())};
+  }
+
+  /// Get layout for reading reduction results from shared memory.
+  /// Reuses the logic from DistributeMultiReduction.
```
sommerlukas: It it's the same as in `DistributeMultiReduction`, can we have a shared helper?

sommerlukas (commented):
Some more code style comments.

This review is also a bit of moving target as new functionality/new approaches get added between reviews. @Groverkss can you please review this, so we can land it and add more functionality in separate, smaller PRs.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: +1 something seems wrong if you have to do this.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: +1, also does not having an index vector actually give us something? If not, can we just remove the "implicit" index vector behavior from arg_compare?

To start, using vector.step and setting redistribute on it would be okay. But we should add a TODO that we should remove the implicit index vector behavior if it doesn't give us anything.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1578,935 @@ struct DistributeMultiReduction final
...
+      // Just collapse the unit reduction dimensions.
+      localValueResult =
+          cast<VectorValue>(vector::ShapeCastOp::create(
+                                rewriter, loc, resultValueDistType, disValue)
+                                .getResult());
+      localIndexResult =
+          cast<VectorValue>(vector::ShapeCastOp::create(
+                                rewriter, loc, resultIndexDistType, disIndex)
+                                .getResult());
+    } else {
```
Groverkss: Don't special case for these cases. Canonicalization should just fold these later on. It just adds more pathways we have to test and maintain.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: We should remove special treatments like this.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1578,935 @@ struct DistributeMultiReduction final
...
+        vector::InsertOp::create(rewriter, loc, accIdx, zeroIdx, insertPos);
+
+    return std::make_pair(cast<VectorValue>(resultVal),
+                          cast<VectorValue>(resultIdx));
+  }
+
+  /// Performs thread-level arg_compare reduction using portable butterfly
+  /// shuffles that work across all GPU targets.
+  FailureOr<std::pair<VectorValue, VectorValue>>
+  doThreadReduction(RewriterBase &rewriter, Location loc,
```
Groverkss: Can we share this implementation with vector.multi_reduction thread distribution? My understanding is that the only thing arg_compare and multi_reduction should differ in is the "combiner" function. We could have some free functions that take the combiner function as a lambda and just perform the thread reduce and subgroup reduce. For local reduction, you just apply the combiner function directly.

To be more concrete, I'm thinking of something like a parent pattern DistributeReduction or something that takes a lambda combiner as an input to the constructor and uses it to do reductions.

Groverkss (changes requested):
I think.a lot of the code could be reused with DistributeMultiReduction. The only thing they differ is the reduction combiner function they use.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: Yes, it is only about the explicit index model.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: > Can we share this implementation with vector.multi_reduction thread distribution? 

multi_reduction does makeArithReduction(kind, val, shuffledVal) ==> single Value. arg_compare does compare(val, shuffledVal) then select on both value and index ==> (Value, Value).

We have two versions of implementations to switch: 1) shuffle version for custom unsupported comparator region, 2) multi_reduction + gpu.ballot to handle argmax/argmin.  

> To be more concrete, I'm thinking of something like a parent pattern DistributeReduction

It's doable, but let's explore one separate PR: perhaps a DistributeReductionBase that parameterizes over the combiner and the number of tracked quantities. I've left a TODO comment on `doThreadReduction `for this.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -936,6 +938,233 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
   return layout.getThreadTile()[dim];
 }
 
+/// Computes the distributed shape after partial reduction, with reduction
+/// dimensions set to 1 across all tile groups (batch, outer, element).
+static SmallVector<int64_t>
+getPartialReducedDistributedShape(NestedLayoutAttr srcLayout,
```
sommerlukas: Nit: The name is quite similar to the function below, but they do different things. Maybe we can change both names to incorporate what they do.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -936,6 +938,233 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
...
+  for (int64_t rDim : reductionDims) {
+    shape[rDim] = preDistrShape[rDim];
+  }
+  return shape;
+}
+
+/// Reshapes flat 1-d vectors back to target types. For 0-d vectors,
+/// uses extract+broadcast since shape_cast to 0-d is not supported.
+static std::pair<VectorValue, VectorValue>
+reshapeFlatToTarget(RewriterBase &rewriter, Location loc, VectorValue flatValue,
```
sommerlukas: Nit: The processing of value and index seems quite independent here. Wouldn't it be simpler to have this function more generic to process only one `VectorValue` and instead call it twice for value and index separately?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -946,6 +1175,93 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
...
 ///      to shared memory and will be reloaded into a layout where partial
 ///      reductions will be placed inside threads.
+
+/// Computes the layout for reading reduction results from shared memory.
+/// Shared by DistributeMultiReduction and DistributeArgCompare.
+static NestedLayoutAttr
+computeLayoutForReductionFromBuffer(NestedLayoutAttr srcLayout,
+                                    ArrayRef<int64_t> reductionDims) {
+  auto subgroupTileLens =
+      llvm::to_vector_of<int64_t>(srcLayout.getSubgroupTile());
```
sommerlukas: For what do we need the `to_vector_of` here and in the following lines?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -946,6 +1175,93 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
...
+}
+
+/// Computes the inter-subgroup write layout by replacing reduced tiles with
+/// unit dimensions. Shared by DistributeMultiReduction and
+/// DistributeArgCompare.
+static NestedLayoutAttr
+computeInterSubgroupWriteLayout(NestedLayoutAttr srcLayout,
+                                ArrayRef<int64_t> reductionDims) {
+  auto subgroupTileLens =
+      llvm::to_vector_of<int64_t>(srcLayout.getSubgroupTile());
```
sommerlukas: Same question as above about `to_vector_of`.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1593,669 @@ struct DistributeMultiReduction final
...
+    VectorValue disValue =
+        getDistributed(rewriter, inputValue, signature[inputValue]);
+
+    auto inputIndexVec = cast<VectorValue>(inputIndex);
+    auto indexLayout =
+        dyn_cast_if_present<NestedLayoutAttr>(signature[inputIndexVec]);
+    if (!indexLayout) {
+      return rewriter.notifyMatchFailure(
+          argCompareOp, "expected nested layout attr for input index");
+    }
```
sommerlukas, line 1651: Does `getDistributed` above create IR? If so, I'd prefer to move this check & failure before that.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1593,669 @@ struct DistributeMultiReduction final
...
+    VectorValue disIndex = cast<VectorValue>(
+        getDistributed(rewriter, inputIndexVec, signature[inputIndexVec]));
+
+    // Unwrap identity ToSIMDOp on 0-d inits. MultiDimReductionOp avoids this
+    // because its acc can be a plain scalar, but ArgCompareOp inits are always
+    // VectorValue so the framework wraps even 0-d vectors.
+    VectorValue disInitValue = initValue;
+    if (isNonZeroRank(initValue)) {
+      disInitValue = getDistributed(rewriter, initValue, initValueLayout);
+    } else if (auto toSIMD =
```
sommerlukas, line 1667: Nit, same below.
```suggestion
    } 
    if (auto toSIMD =
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1593,669 @@ struct DistributeMultiReduction final
...
+      replaceOpWithDistributedValues(rewriter, argCompareOp,
+                                     {localValueResult, localIndexResult});
+      return success();
+    }
+
+    std::pair<VectorValue, VectorValue> threadReduced = {localValueResult,
+                                                         localIndexResult};
+    if (hasThreadReductions) {
+      std::optional<ComparatorAnalysis> analysis =
+          analyzeComparatorForSubgroupReduce(argCompareOp.getRegion());
```
sommerlukas: Nit or personal taste: The function is called `analyzeComparatorForSubgroupReduce` and then we call `doThreadReduction`. Maybe `analyzeComparatorForThreadReduction` would be better?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1593,669 @@ struct DistributeMultiReduction final
...
+  FailureOr<std::pair<VectorValue, VectorValue>> doLocalArgCompareReduction(
+      RewriterBase &rewriter, Location loc, VectorValue inputVal,
+      VectorValue inputIdx, VectorValue initVal, VectorValue initIdx,
+      Region &comparatorRegion, int64_t reductionDim, int64_t rank) const {
+    VectorType valType = inputVal.getType();
+    VectorType idxType = inputIdx.getType();
+    int64_t distRank = valType.getRank();
+    if (distRank != 3 * rank || idxType.getRank() != distRank) {
+      return failure();
+    }
```
sommerlukas: Can this really happen in a well-formed IR? Should this be an assertion instead?

sommerlukas (commented):
A few more nits, thanks for all the improvements so far!

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: The else if is intentional here:  without it, a non-zero-rank value that also happens to be defined by ToSIMDOp would hit both branches, with the second overwriting the getDistributed result. Added code comment for it.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1575,674 @@ struct DistributeMultiReduction final
...
+    auto indexMask = vector::CreateMaskOp::create(
+        rewriter, loc, indexReadTy.clone(rewriter.getI1Type()),
+        memref::getMixedSizes(rewriter, loc, indexBuffer));
+
+    Value valuePad = arith::ConstantOp::create(
+                         rewriter, loc, rewriter.getZeroAttr(valueElemType))
+                         .getResult();
+    Value indexPad = arith::ConstantOp::create(
+                         rewriter, loc, rewriter.getZeroAttr(indexElemType))
+                         .getResult();
```
sommerlukas: Is zero the correct neutral element for the reduction in all cases?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1575,674 @@ struct DistributeMultiReduction final
...
+    auto zero = arith::ConstantIndexOp::create(rewriter, loc, 0);
+    auto inBounds = rewriter.getBoolArrayAttr(
+        SmallVector<bool>(readLayout.getRank(), true));
+
+    auto valueMask = vector::CreateMaskOp::create(
+        rewriter, loc, valueReadTy.clone(rewriter.getI1Type()),
+        memref::getMixedSizes(rewriter, loc, valueBuffer));
+    auto indexMask = vector::CreateMaskOp::create(
+        rewriter, loc, indexReadTy.clone(rewriter.getI1Type()),
+        memref::getMixedSizes(rewriter, loc, indexBuffer));
```
sommerlukas: @Groverkss I wonder if we should land https://github.com/iree-org/iree/pull/23613 after all if we create new `create_mask` operations during distribution that have to be distributed, too.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1377,6 +1575,674 @@ struct DistributeMultiReduction final
...
+          vector::InsertOp::create(rewriter, loc, accIdx, outIdxVec, outIndices)
+              .getResult();
+    }
+
+    return std::make_pair(cast<VectorValue>(outValVec),
+                          cast<VectorValue>(outIdxVec));
+  }
+
+  /// Flattened reduction state shared between ballot and shuffle impls.
+  struct FlattenedReductionState {
```
sommerlukas, line 1843: We could avoid passing around the `FlattenedReductionState` struct if we make `doThreadReductionWith[Shuffles|Ballot]` members of that struct, so they can access the members. 

It's probably still small enough that it doesn't matter much and the current implementation is fine, so just an idea.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -946,6 +1165,85 @@ static int64_t getShuffleWidth(NestedLayoutAttr layout, int64_t dim) {
 ///   4. Subgroup reduce : each subgroup will store the partial reductions
 ///      to shared memory and will be reloaded into a layout where partial
 ///      reductions will be placed inside threads.
```
sommerlukas: I think this comment should be attached to `struct DistributeMultiReduction`.

sommerlukas (approved):
LGTM apart from a few last small things & questions. Please also wait for @Groverkss to approve.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: No, I kept this in mind but just forgot to make the changes

bangtianliu: Thanks for catching it

bangtianliu:
cc @Groverkss here

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
Groverkss: Wait why do we do this. We should remove this.

bangtianliu: Removed. FYI, this mimicked the pattern in DistributeMultiReduction::doSubgroupReductionFromBuffer, which also creates a create_mask with inBounds all-true. Is the mask redundant there as well? If so, I'd be happy to clean that up in a follow-up.

Groverkss (approved):

bangtianliu:
I'll go ahead and land this PR. If any issues come up, I'll send follow-up PRs to address them.

PR #23795, opened 2026-03-16
[Tuner][Codegen] Add pipeline constraint interfaces and insert pass

kuhar:
Allow each pipeline attr to generate its own SMT constraints via
`PipelineAttrInterface::emitConstraints` (external model). The insert pass
is backend-agnostic -- it enumerates pipelines from the target and
lets each implementation self-filter. This keeps constraint logic
co-located with the pipeline it describes and extensible to new
backends without modifying the pass.

Only workgroup tile divisibility for `VectorDistribute` so far to keep this small,
full constraints will land incrementally.

Common utilities (`createConstraintsOpShell`, `mkIntConst`, `mkKnob`)
live in `SMTConstraintUtils` for reuse across pipeline generators.


kuhar:
I rebased this and simplified a bit

compiler/src/iree/compiler/Codegen/Common/test/insert_smt_constraints.mlir
```diff
@@ -0,0 +1,223 @@
+// RUN: iree-opt --split-input-file \
+// RUN:   --iree-codegen-add-tuner-attributes \
+// RUN:   --pass-pipeline='builtin.module(hal.executable(hal.executable.variant(builtin.module(func.func(iree-codegen-insert-smt-constraints)))))' %s \
+// RUN:   | FileCheck %s
+
+// RUN: iree-opt --split-input-file \
+// RUN:   --pass-pipeline='builtin.module(hal.executable(hal.executable.variant(builtin.module(func.func(iree-codegen-insert-smt-constraints)))))' %s \
+// RUN:   | FileCheck %s --check-prefix=NO_TUNER_ATTRS
+
+// -----
```
qedawkins:
```suggestion
```

compiler/src/iree/compiler/Codegen/Common/test/insert_smt_constraints.mlir
```diff
@@ -0,0 +1,223 @@
+// RUN: iree-opt --split-input-file \
+// RUN:   --iree-codegen-add-tuner-attributes \
+// RUN:   --pass-pipeline='builtin.module(hal.executable(hal.executable.variant(builtin.module(func.func(iree-codegen-insert-smt-constraints)))))' %s \
+// RUN:   | FileCheck %s
+
+// RUN: iree-opt --split-input-file \
+// RUN:   --pass-pipeline='builtin.module(hal.executable(hal.executable.variant(builtin.module(func.func(iree-codegen-insert-smt-constraints)))))' %s \
+// RUN:   | FileCheck %s --check-prefix=NO_TUNER_ATTRS
```
qedawkins: This check is strange since it's only checking one thing yet running on all current and future tests here. I would split it into its own test.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
```diff
@@ -776,6 +776,41 @@ def IREECodegen_PipelineAttrInterface :
   ];
 }
 
+def IREECodegen_PipelineConstraintAttrInterface :
```
qedawkins: I came to the conclusion that it should be fine to roll this into the pipeline interface. Given pipeline + applicable op -> emitting a constraint set for different symbolic knobs is a pretty generic idea and fewer interfaces is better.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
```diff
@@ -776,6 +776,41 @@ def IREECodegen_PipelineAttrInterface :
...
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/[{
+        Emit constraint ops for a set of root ops in a dispatch function.
+        The implementation examines the root ops and decides which pipelines
+        and strategies apply. Each root op carries a RootOpAttr identifying
+        its set number; all root ops in the array have the same set number.
+        Returns success even if it skips all ops (not an error). Returns
+        failure only on unrecoverable error.
```
qedawkins: Can you document where these constraint ops are supposed to be emitted? It's not immediately clear to me based on the signature.

qedawkins (commented):
Sorry took a while to understand the interfaces added here. This makes sense but we can simplify the boilerplate before landing.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.td
kuhar: Good point, the default implementation can be a noop without too much harm

kuhar:
@qedawkins thanks for the idea with merging both interfaces, should be simpler now

qedawkins (approved):

PR #23798, opened 2026-03-16
[Codegen][nfc] Restructure and improve the vectorization tests.

hanhanW:
The revision breaks the large file into a few categories, based on the options:

- `generic_vectorization_unmasked`: static cases only, the vector input size is empty.
- `generic_vectorization_masked_configured`: use configured vector size for masked vectorization.
- `generic_vectorization_masked_inferred`: infer the vector sizes from IR for masked vectorization.
- `generic_vectorization_using_transfer_gather`: test cases with `transfer_gather` option is on.
- `generic_vectorization_map_store`: test suite for map_store op, which is guarded under an option.
- `generic_vectorization_fold_cast`: test that the cast op is folded into contract op, which is guarded under an option.

Additionally, the revision adds a couple dynamic cases to `generic_vectorization_using_transfer_gather`, which ensures that the vector inference is working for the case.


hanhanW:
For reviewers:
- The first commit is pure-copy commits, with a little duplication for a TODO item.
- The rest of commits are improvements for tests.

bangtianliu (approved):

PR #23803, opened 2026-03-17
[compiler][nfc] Drop `Pass` suffix from filenames.

hanhanW:
The convention that IREE follows is
- Having `Pass` in declaration.
- Not having `Pass` suffix in filenames.


kuhar (approved):
Maybe worth adding to the style guide on iree.dev?

bangtianliu (approved):

hanhanW:
> Maybe worth adding to the style guide on iree.dev?

It seems to be a hidden convention since MLIR was started. But yeah, I'm happy to call it out.

PR #23805, opened 2026-03-17
[Tokenizer] Fix rstrip flag: remove match rejection, add whitespace consumption

jtuyls:
## Summary

`flags_allow_match()` treated `rstrip=true` as a match condition, rejecting tokens not followed by whitespace or end of input. Per [HuggingFace's implementation(https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs), `rstrip` is a post-processing directive that consumes trailing whitespace after the match, not a match gate. This caused special tokens like `<|user|>` in Phi-4-mini to fall through to BPE subword tokenization when followed by text.

**Example** (Phi-4-mini-instruct, token `<|user|>` has `rstrip=true`):

Input: "<|user|>Hello"
Before fix:  [27, 91, 1428, 91, 29, 13225]  ← <|user|> split into subwords
After fix:   [200021, 13225]                  ← matches HuggingFace output

## Changes

- Remove the rstrip rejection branch from `flags_allow_match()` in `special_tokens.c`
- Add rstrip whitespace consumption in `tokenizer.c` after a successful match
- Add `matched_flags` field to encode state to propagate flags to the caller

### HuggingFace smoketest (`huggingface_smoketest.py`)

**1667/1667** tokenization comparisons pass across ~80 HuggingFace models (0 mismatches).

76 additional tests fail to **load** (not tokenization mismatches) — these are tiktoken models listed in the HF smoketest that aren't valid HuggingFace model identifiers. They're tested by the dedicated tiktoken smoketest instead.

### Tiktoken smoketest (`tiktoken_smoketest.py`)

**72/76** tokenization comparisons pass across 4 tiktoken encodings (cl100k_base, o200k_base, r50k_base, p50k_base).

4 failures — **identical between upstream and this PR** (pre-existing):

| Encoding | Failing test | Cause |
|----------|-------------|-------|
| cl100k_base | `special_token_endoftext` | IREE matches `<\|endoftext\|>` as special token; tiktoken's `encode_ordinary` treats it as literal text |
| o200k_base | `special_token_endoftext` | Same |
| r50k_base | `special_token_endoftext` | Same |
| p50k_base | `special_token_endoftext` | Same |

Root cause: the IREE tokenizer has no "encode ordinary" mode (equivalent to tiktoken's `disallowed_special=()`).

### DIFF
--- a/runtime/src/iree/tokenizer/tokenizer.c
+++ b/runtime/src/iree/tokenizer/tokenizer.c
@@ -1639,10 +1639,22 @@ static iree_host_size_t iree_tokenizer_strip_lstrip_whitespace(
     flags = special_tokens->flags[check_state->partial_token_index];
   }
 
-  // If LSTRIP is set, strip trailing whitespace from the safe prefix.
+  // If LSTRIP is set, strip trailing Unicode whitespace from the safe prefix.
   if (iree_any_bit_set(flags, IREE_TOKENIZER_SPECIAL_TOKEN_FLAG_LSTRIP)) {
-    while (safe > 0 && (uint8_t)input.data[safe - 1] <= 0x20) {
-      --safe;
+    while (safe > 0) {
+      // Walk back over UTF-8 continuation bytes (10xxxxxx) to find the
+      // leading byte. ASCII bytes (0xxxxxxx) don't match the continuation
+      // pattern, so the loop is a no-op for single-byte characters.
+      iree_host_size_t start = safe - 1;
+      while (start > 0 && ((uint8_t)input.data[start] & 0xC0) == 0x80) {
+        --start;
+      }
+      iree_host_size_t pos = 0;
+      iree_string_view_t tail =
+          iree_make_string_view(input.data + start, safe - start);
+      uint32_t codepoint = iree_unicode_utf8_decode(tail, &pos);
+      if (!iree_unicode_is_whitespace(codepoint)) break;
+      safe = start;
     }
   }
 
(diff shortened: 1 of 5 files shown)


runtime/src/iree/tokenizer/tokenizer.c
```diff
@@ -2325,6 +2325,16 @@ static iree_status_t iree_tokenizer_encode_state_pump(
         chunk->size -= match_length;
         *out_made_progress = true;
 
+        // rstrip: consume trailing whitespace (bytes <= 0x20: space and
+        // control chars) after the matched token.
+        if (iree_any_bit_set(state->special_token_match.matched_flags,
+                             IREE_TOKENIZER_SPECIAL_TOKEN_FLAG_RSTRIP)) {
+          while (chunk->size > 0 && (uint8_t)chunk->data[0] <= 0x20) {
```
benvanik: I think you need to use iree_unicode_is_whitespace here, unless RSTRIP *exclusively* applies to ASCII text (which would feel wrong)

benvanik (changes requested):

runtime/src/iree/tokenizer/tokenizer.c
jtuyls: Thanks for catching! I updated it here and in the existing lstrip case as well: https://github.com/iree-org/iree/pull/23805/changes#diff-16663523ce43be07425c9dfad27d6f2c518e09094df60a668002a28af3300373L1644

benvanik (approved):

PR #23806, opened 2026-03-17
[GPU] Fix RDNA3 WMMA f16/bf16 accumulator layout

jtuyls:
The WMMAR3_F16_16x16x16_F16 (and BF16) accumulator layout on gfx1100 was incorrect: outer={16,1} assumed 16 valid elements per lane, but v_wmma_f16_16x16x16_f16 only produces 8 valid results at even indices. Fix the layout to outer={8,1}, thread={2,16} and add interleave/ deinterleave shuffles in createMmaOp to widen to the hardware's native vector<16xf16> type.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
@@ -453,7 +453,10 @@ MMASingleSubgroupLayout getSingleSubgroupLayout(MMAIntrinsic intrinsic,
       return {/*outer=*/{1, 1}, /*thread=*/{1, 16}, /*tstrides=*/{0, 1},
               /*element=*/{16, 1}};
     case kMMAOperandAcc:
-      return {/*outer=*/{16, 1}, /*thread=*/{1, 16}, /*tstrides=*/{0, 1},
+      // RDNA3 WMMA f16/bf16 output uses vector<16xf16> but only 8 values are
+      // valid (at even indices with subwordOffset=0). The actual lane mapping
+      // is the same as the f32 accumulator: 2 thread groups in M, 16 in N.
+      return {/*outer=*/{8, 1}, /*thread=*/{2, 16}, /*tstrides=*/{16, 1},
               /*element=*/{1, 1}};
     }
   case MMAIntrinsic::WMMAR4_F32_16x16x16_F16:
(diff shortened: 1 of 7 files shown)


krzysz00:
So, one thing I'd like to flag is that these instructions let you pick whether or not to write the even or odd elements of the result vector, which means we'll probably want to make a virtual intrinsic that uses pairs of such WMMAs to get a bigger tile. Not necessarily relevant to this PR but I figured I'd comment

krzysz00:
On the other hand, implementing the intrinsic as just producing the even results makes sense, maybe.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -796,6 +798,32 @@ static Value createMmaOp(OpBuilder &builder, Location loc,
...
+          builder, loc,
+          SplatElementsAttr::get(halfVecTy, builder.getZeroAttr(elemTy)));
+      Value hwAcc = vector::InterleaveOp::create(builder, loc, acc, zero);
+      Value hwResult =
+          amdgpu::WMMAOp::create(builder, loc, hwVecTy, layout.mSize,
+                                 layout.nSize, layout.kSize, lhs, rhs, hwAcc)
+              .getResult();
+      // Deinterleave 16 → 8: extract even indices (the valid results).
+      auto deinterleave =
+          vector::DeinterleaveOp::create(builder, loc, hwResult);
```
efric, line 841: I suspect this interleave/deinterleave pattern might add nontrivial overhead if the heuristics decide to tile multiple intrinsics along K. But I'm working on addressing a very similar issue for vdmfma accumulator. i think it can be extended to support this too

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -796,6 +798,32 @@ static Value createMmaOp(OpBuilder &builder, Location loc,
...
+    // the accumulator to vector<16xf16> here so the LLVM backend gets the
+    // native type and generates correct code in reduction loops.
+    bool isRDNA3HalfAcc =
+        (intrinsic == MMAIntrinsic::WMMAR3_F16_16x16x16_F16 ||
+         intrinsic == MMAIntrinsic::WMMAR3_BF16_16x16x16_BF16);
+    if (isRDNA3HalfAcc) {
+      Type elemTy = cast<VectorType>(acc.getType()).getElementType();
+      auto halfVecTy = VectorType::get({8}, elemTy);
+      auto hwVecTy = VectorType::get({16}, elemTy);
+      // Interleave 8 → 16: place elements at even indices, zeros at odd.
```
efric: I think we prefer to stick with ascii character (i.e, `->`)

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_vector_distribute_gfx1100.mlir
```diff
@@ -83,13 +83,13 @@ hal.executable.variant @rocm target(<"rocm", "rocm-hsaco-fb">) {
 }
 
 //    CHECK-LABEL: func.func @matmul_256x256x256_f16_f16
-//          CHECK:   scf.for {{.*}} = %c0 to %c256 step %c128 iter_args({{.*}}) -> (vector<2x2x16x1x1x1xf16>)
+//          CHECK:   scf.for {{.*}} = %c0 to %c256 step %c128 iter_args({{.*}}) -> (vector<2x2x8x1x1x1xf16>)
 // Each subgroup handles 2 * 2 tiles, and for each tile we accumulate 8 times
 // along the K dimension. So in total 32 wmma ops.
 // CHECK-COUNT-32:     amdgpu.wmma {{.*}} : vector<16xf16>, vector<16xf16>, vector<16xf16>
```
efric: I think we should add the new (32?) interleaves and deinterleaves to the lit test (same to above).

efric (commented):
Overall looks good to me, but i am not as familiar with these intrinsics as others may be

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_vector_distribute_gfx1100.mlir
jtuyls: Added them

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
krzysz00: More accurately, can we confirm that these cancel within loops?

... Ok, actually, this leads me to a more general thought, see #23806

jtuyls: They don't seem to cancel and result in additional v_mov_b16 instructions. I think it's worthwhile to get this in already as it fixes a correctness issue, wdyt? I added a TODO referring to these options you mentioned though.

krzysz00 (approved):
Let's go ahead and land this so we don't have a correctness issue, I'll poke at the general case.

bjacob (approved):
Agree with the two-step plan (collecting above comments) to:
1. Land this to fix the correctness problem as suggested by @jtuyls.
2. Introduce a virtual intrinsic pairing two even/odd real intrinsics to compute a 2x bigger tile without having to create additional interleaving ops as suggested by @krzysz00 .

This layer of codegen is meant to be a 1:1 reflection of target intrinsics and not introduce any layout-change op. That you were forced to introduce such ops (the interleaving ops) just means that RDNA3 is improperly supported by our design.  If this is a one-off for RDNA3 then at this point the above two-step plan might be enough.  If this is going to keep happening, then we need to decide: are we OK with having to solve this again in the future with virtual intrinsics, or do we need to generalize our abstractions here? IIUC this is another instance of GPU instructions dissociating the dimensions of the accumulator vs LHS/RHS tile, just in a different way than the ones we already knew about.

PR #23807, opened 2026-03-17
[DMA][Swizzle] Enable LDS DMA with swizzling

lialan:
* Add swizzle detection helper that traces destination memref through `expand_shape/collapse_shape/subview` to find `SwizzleHintOp`.
* Apply the swizzle attribute's offset transformation to source linear offsets in the `gather_to_lds` lowering.
* XOR swizzle is self-inverse, so applying it to source addresses produces the correct swizzled layout in LDS without violating `gather_to_lds`'s uniform-destination constraint.
* Add pipeline tests and E2E tests.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -148,8 +148,7 @@ computeTransferSegments(int64_t totalElements, int64_t elementBits,
 ///
 /// The gather_to_lds instruction requires:
 ///   - Source indices: per-lane divergent (each lane reads from different loc)
-///   - Destination indices: per-lane divergent (each lane writes to distinct
-///     LDS location)
+///   - Destination indices: subgroup-uniform (all lanes write to same LDS base)
 ///
 /// Index computation rules for each dimension:
 ///
@@ -160,7 +159,7 @@ computeTransferSegments(int64_t totalElements, int64_t elementBits,
 ///
 /// Where:
 ///   - srcDimOffset: position with lane offset (divergent per lane)
-///   - dstDimOffset: position with lane offset (divergent per lane)
+///   - dstDimOffset: position without lane offset (uniform across subgroup)
 ///   - indices[dim]: index memref mapping dest positions to source positions
 static std::pair<SmallVector<Value>, SmallVector<Value>>
 generateGatherIndices(OpBuilder &rewriter, Location loc,
@@ -408,10 +407,9 @@ private:
           SmallVector<Value> srcDimOffsets(outerDimOffsets);
           llvm::append_range(srcDimOffsets, srcDelinearize.getResults());
 
-          // Destination indices: include lane offset (divergent per lane) so
-          // each lane writes to its own distinct LDS location.
+          // Destination indices: no lane offset (subgroup-uniform).
           auto dstDelinearize = affine::AffineDelinearizeIndexOp::create(
-              rewriter, loc, srcLinearOffset, basis, /*hasOuterBound=*/true);
+              rewriter, loc, linearOffsetVal, basis, /*hasOuterBound=*/true);
 
           SmallVector<Value> dstDimOffsets(outerDimOffsets);
           llvm::append_range(dstDimOffsets, dstDelinearize.getResults());
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -9,6 +9,7 @@
 #include <optional>
 #include "iree/compiler/Codegen/Common/GPU/Passes.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.h"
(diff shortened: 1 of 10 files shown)


lialan:
Initial benchmark results on MI355X without tuning:

| Shape | Branch | Time (mean) | TFLOPS | vs main |
|-------|--------|-------------|--------|---------|
| 8192³ | PR (XOR swizzle + DMA) | 0.522 ms | 2106 | -17% |
| 8192³ | main (XOR swizzle only) | 0.434 ms | 2533 | baseline |
| 16384³ | PR (XOR swizzle + DMA) | 4.02 ms | 2188 | -23% |
| 16384³ | main (XOR swizzle only) | 3.09 ms | 2847 | baseline |
| 32768³ | PR (XOR swizzle + DMA) | 33.3 ms | 2113 | -22% |
| 32768³ | main (XOR swizzle only) | 27.3 ms | 2578 | baseline |

lialan:
looking into the reason of regression.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -144,6 +145,31 @@ computeTransferSegments(int64_t totalElements, int64_t elementBits,
   return segments;
 }
 
+/// Trace a memref value through reshape/subview ops to find a SwizzleHintOp.
+/// Returns the swizzle attribute if found, std::nullopt otherwise.
+static std::optional<IREE::Codegen::SwizzleAttrInterface>
+getDestSwizzleAttr(Value dest) {
+  while (true) {
+    if (auto expandOp = dest.getDefiningOp<memref::ExpandShapeOp>()) {
```
Yu-Zhewen: Can we just use `ViewLikeOpInterface` here?

Yu-Zhewen:
> looking into the reason of regression.

For BF16, I noticed that multi-buffering does not work with swizzling (https://github.com/iree-org/iree/issues/23919). This might be the same issue here.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -911,36 +911,33 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
...
+    FailureOr<Attribute> lhsSwizzleAttr = getXorShuffleAttr(
+        context, baseAttr, target, kind, schedule->kTileSizes, kMMAOperandLhs);
+    FailureOr<Attribute> rhsSwizzleAttr = getXorShuffleAttr(
+        context, baseAttr, target, kind, schedule->kTileSizes, kMMAOperandRhs);
+    if (failed(lhsSwizzleAttr) || failed(rhsSwizzleAttr)) {
+      promotionArray =
+          useDirectLoad
+              ? SmallVector<Attribute>{useGlobalDma, useGlobalDma,
+                                       defaultConfigAttr, defaultConfigAttr}
+              : SmallVector<Attribute>{};
```
Muzammiluddin-Syed-ECE:
```suggestion
      promotionArray = SmallVector<Attribute>{baseAttr, baseAttr, defaultConfigAttr, defaultConfigAttr};
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1581,3 +1581,73 @@ func.func @no_lower_oob_without_fat_raw_buffer(
...
   return
 }
+
+// -----
+
+// Test: Inverse source swizzle when destination has swizzle_hint.
+// The destination traces through expand_shape -> swizzle_hint -> alloc.
+// Source indices should be swizzled, destination indices should be linear.
+//
+// Shape: 4x128 f32 dest. With 32 lanes, 128-bit DMA = 4 elements/lane.
```
Muzammiluddin-Syed-ECE: nit: can we change this to be a test with subgroup size = 64? That feels more relevant for gfx950 although it doesn't really matter. This is sufficient for showing the swizzling on src indices.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
Muzammiluddin-Syed-ECE, line 347: Can we update this function documentation to mention the inverse swizzling.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
krzysz00: I'm not sure it's appropriate here, as other ViewLikeOphinterface implementations are things like subview that break the contiguity requirements

krzysz00: ... Oh, if we're allowing subview, yeah, this should just be ViewLikeOpInterface, I think

krzysz00 (commented):
I don't have correctness concerns, but I don't think I'd want to land this with those perf regressions. Maybe we should investigate that multibuffering issue?

Yu-Zhewen:
> I don't have correctness concerns, but I don't think I'd want to land this with those perf regressions. Maybe we should investigate that multibuffering issue?

Agree. I’m currently looking into it.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
lialan: done, updated to use `ViewLikeOpInterface`

lialan:
The plan is to split this PR into smaller, ones. 

* https://github.com/iree-org/iree/pull/24094 only handles reverse swizzling, will go in first.

PR #23808, opened 2026-03-17
[LLVMGPU] Remove special bufferization pipeline for VectorDistribute

Groverkss:
Reuse the TileAndFuse pipeline. VectorDistribute always allocates it's own workgroup memory in GPUVectorAlloc pass. Any memref.copy created by bufferization is either a no-op or a bug. Usually bufferization will generate noop copies from scf.forall parallel insert_slice stuff, but we shouldn't barrier on them as they are noops.

ci-extra: test_torch


qedawkins (approved):
:o

PR #23810, opened 2026-03-17
[Codegen] Remove dead ExtractAddressComputation passes

Groverkss:
The ExtractAddressComputation pass in the LLVMGPU pipeline is immediately undone by the FoldMemRefAliasOps pass that follows it, making it a no-op round-trip. Remove both the GPU-specific and common ExtractAddressComputation passes since neither is used effectively.

ci-extra: test_torch


krzysz00 (approved):
Lgtm here

kuhar (approved):

PR #23815, opened 2026-03-17
[Codegen] Drop legacy split-reduction handling from ReconcileTranslationInfo pass

hanhanW:
The revision removes `foldSplitReductionLoopIntoWorkgroupMappingLoop` option, because the option has been default on since https://github.com/iree-org/iree/commit/8c6194fe39d39a9b397a1eb15011004b67c587b7, i.e., ~8 months.


MaheshRavishankar (approved):
Ok, hopefully this doesnt have issues.

PR #23816, opened 2026-03-17
[Codegen] Move LLVMGPU pipelines to `#iree_gpu.pipeline<...>` attr

kuhar:
Replace LLVMGPU entries in the global `DispatchLoweringPassPipeline` enum with a dedicated `#iree_gpu.pipeline<...>` attribute backed by a new `IREE::GPU::LoweringPipeline` enum.

The overall goal is to enable pipeline-specific constraint generation -- I want to be able to attach an (external) interface to gpu pipelines.

The global enum carried target backend-specific pipelines that belong in their respective dialect. This moves the seven LLVMGPU pipelines into `iree_gpu`, with `PipelineAttrInterface` so they integrate with `TranslationInfoAttr` and `ConstraintsOp`.

I plan to move the remaining pipelines in follow up PRs, to keep each PR as small as feasible.

### DIFF
--- a/compiler/plugins/target/ROCM/builtins/mlir_ukernel/ukernel_patterns_gfx942.mlir
+++ b/compiler/plugins/target/ROCM/builtins/mlir_ukernel/ukernel_patterns_gfx942.mlir
@@ -67,7 +67,7 @@ pdl.pattern @annotate_matmul_like_f8E4M3FNUZ_medium_expanded : benefit(1) {
         workgroup = [1, 128, 256, 0],
         workgroup_reordering_strategy = #iree_gpu.conditional_transpose<8, 38>
       }>,
-      translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+      translation_info = #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
         workgroup_size = [512, 1, 1] subgroup_size = 64,
         // This strategy manually prefetches and eliminates bank conflicts on LDS
         // by using swizzling (rotate_rows). Therefore, we disable prefetch_shared_memory
@@ -158,7 +158,7 @@ pdl.pattern @annotate_matmul_like_f8E4M3FNUZ_large_expanded : benefit(2) {
         workgroup = [1, 256, 256, 0],
         workgroup_reordering_strategy = #iree_gpu.conditional_transpose<8, 38>
       }>,
-      translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+      translation_info = #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
         workgroup_size = [512, 1, 1] subgroup_size = 64,
         // This strategy manually prefetches and eliminates bank conflicts on LDS
         // by using swizzling (rotate_rows). Therefore, we disable prefetch_shared_memory
@@ -242,7 +242,7 @@ pdl.pattern @annotate_matmul_like_f16_large : benefit(1) {
     %config = pdl.attribute = #iree_codegen.compilation_info<
       lowering_config = #iree_gpu.lowering_config<{workgroup = [256, 256, 0],
                                                    workgroup_reordering_strategy = #iree_gpu.conditional_transpose<8, 38>}>,
-      translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+      translation_info = #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
         workgroup_size = [512, 1, 1] subgroup_size = 64,
         // This strategy manually prefetches and eliminates bank conflicts on LDS
         // by using swizzling (rotate_rows). Therefore, we disable prefetch_shared_memory
@@ -330,7 +330,7 @@ pdl.pattern @annotate_matmul_like_f16_medium_expanded : benefit(1) {
         workgroup = [1, 128, 256, 0],
         workgroup_reordering_strategy = #iree_gpu.conditional_transpose<8, 38>
         }>,
-      translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+      translation_info = #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
         workgroup_size = [512, 1, 1] subgroup_size = 64,
         // This strategy manually prefetches and eliminates bank conflicts on LDS
         // by using swizzling (rotate_rows). Therefore, we disable prefetch_shared_memory
@@ -418,7 +418,7 @@ pdl.pattern @annotate_matmul_like_f16_large_expanded : benefit(2) {
     %config = pdl.attribute = #iree_codegen.compilation_info<
       lowering_config = #iree_gpu.lowering_config<{workgroup = [1, 256, 256, 0],
                                                    workgroup_reordering_strategy = #iree_gpu.conditional_transpose<8, 38>}>,
-      translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+      translation_info = #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
         workgroup_size = [512, 1, 1] subgroup_size = 64,
         // This strategy manually prefetches and eliminates bank conflicts on LDS
         // by using swizzling (rotate_rows). Therefore, we disable prefetch_shared_memory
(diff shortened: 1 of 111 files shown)


bangtianliu (approved):

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUEnums.td
```diff
@@ -462,4 +462,24 @@ def IREEGPU_ReorderWorkgroupsStrategy : IREEGPU_I32PipelineEnumAttr<"ReorderWork
...
+//===----------------------------------------------------------------------===//
+
+// LLVMGPU pipelines.
+def IREEGPU_PP_Default : I32EnumAttrCase<"Default", 0>;
+def IREEGPU_PP_BaseLowering : I32EnumAttrCase<"BaseLowering", 1>;
+def IREEGPU_PP_Distribute : I32EnumAttrCase<"Distribute", 2>;
+def IREEGPU_PP_Vectorize : I32EnumAttrCase<"Vectorize", 3>;
+def IREEGPU_PP_VectorDistribute : I32EnumAttrCase<"VectorDistribute", 4>;
+def IREEGPU_PP_WinogradVectorize : I32EnumAttrCase<"WinogradVectorize", 5>;
+def IREEGPU_PP_TileAndFuse : I32EnumAttrCase<"TileAndFuse", 6>;
```
hanhanW: What does PP stand for?

hanhanW (approved):
The direction looks okay to me, but please wait for an approval from @qedawkins since he has more context.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUEnums.td
kuhar: pass pipeline

kuhar: actually, maybe let me change it to LP to match the name just below

hanhanW: LP sounds better to me, thanks!

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUEnums.td
```diff
@@ -462,4 +462,24 @@ def IREEGPU_ReorderWorkgroupsStrategy : IREEGPU_I32PipelineEnumAttr<"ReorderWork
     ]> {
 }
 
+//===----------------------------------------------------------------------===//
+// GPU Lowering Pipelines
+//===----------------------------------------------------------------------===//
+
+// LLVMGPU pipelines.
+def IREEGPU_LP_Default : I32EnumAttrCase<"Default", 0>;
```
MaheshRavishankar, line 470: I am assuming `_LP_` is just for differentiating from existing enums

compiler/plugins/target/ROCM/builtins/mlir_ukernel/ukernel_patterns_gfx942.mlir
```diff
@@ -67,7 +67,7 @@ pdl.pattern @annotate_matmul_like_f8E4M3FNUZ_medium_expanded : benefit(1) {
         workgroup = [1, 128, 256, 0],
         workgroup_reordering_strategy = #iree_gpu.conditional_transpose<8, 38>
       }>,
-      translation_info = #iree_codegen.translation_info<pipeline = LLVMGPUTileAndFuse
+      translation_info = #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
```
qedawkins, line 70: Once you're done with the refactor `pipeline =` will probably become redundant in all cases since the attribute names are likely to specify, so probably worth removing.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
```diff
@@ -32,21 +32,6 @@ def CPU_DataTiling
...
-def LLVMGPU_SimpleDistribute
-    : I32EnumAttrCase<"LLVMGPUDistribute", 102>;
-def LLVMGPU_Vectorize
-    : I32EnumAttrCase<"LLVMGPUVectorize", 103>;
-def LLVMGPU_VectorDistribute
-    : I32EnumAttrCase<"LLVMGPUVectorDistribute", 104>;
-def LLVMGPU_WinogradVectorize
-    : I32EnumAttrCase<"LLVMGPUWinogradVectorize", 105>;
-def LLVMGPU_TileAndFuse
-    : I32EnumAttrCase<"LLVMGPUTileAndFuse", 106>;
```
qedawkins, line 48: Enums skip 100 now, looks odd (I understand it's transient)

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -2465,6 +2466,18 @@ bool LaneIdAttr::isLinearMapping() const { return true; }
...
+//===----------------------------------------------------------------------===//
+// GPU Pipeline Attribute
+//===----------------------------------------------------------------------===//
+
+LogicalResult PipelineAttr::buildPipeline(OpPassManager &pm) const {
+  // GPU pipeline construction requires target-specific context (pipeline
+  // options, forROCDL flag) that PipelineAttrInterface::buildPipeline does not
+  // provide. LLVMGPULowerExecutableTarget dispatches on GPU::PipelineAttr
+  // directly before the generic PipelineAttrInterface path.
+  return failure();
```
qedawkins: Hmmm this looks very odd, why even use the interface if we're doing this? The thinking was we can put the options on the attribute, but if not that at least constructing a version of the pipeline with default options would be preferred to this IMO.

qedawkins: Also to avoid circular deps trying to construct the pass pipeline here I'd recommend implementing the interface as an external model.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
kuhar: I didn't want to touch the actual values; I'm not sure anyone relies on these, but from the compiler side they just need to be unique

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
kuhar: The tricky part is that buildPipeline only takes an OpPassManager -- it doesn't have access to the function op for pipelineOptions or forROCDL. So we'd either need to extend the interface signature or store those
on the attribute. We talked about the latter and I think you didn't like the idea of extending attributes. WDYT?

MaheshRavishankar (commented):
I am confused how the verifier allows a mix of the earlier enums and this attribute

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUEnums.td
kuhar: it's just a name based on the (future) enum attr name: dialect + enum_attr_name + value_name

kuhar:
> I am confused how the verifier allows a mix of the earlier enums and this attribute

we check the interface:
```c++
  } else if (!passPipeline
                  .hasPromiseOrImplementsInterface<PipelineAttrInterface>()) 
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -2465,6 +2466,14 @@ bool LaneIdAttr::isLinearMapping() const { return true; }
 
 int64_t LaneIdAttr::getRelativeIndex() const { return getDim(); }
 
+//===----------------------------------------------------------------------===//
+// GPU Pipeline Attribute
+//===----------------------------------------------------------------------===//
+
+// PipelineAttrInterface is implemented via an external model registered
+// from GPUPipelineExternalModels.cpp. The builder callback is set by the
+// LLVMGPU backend in registerCodegenLLVMGPUPasses().
```
qedawkins: I would put this in the tablegen rather than here

qedawkins (approved):

PR #23824, opened 2026-03-18
[Preprocessing] Add PromoteLinalgOutputsPass

HanKuanChen:
This pass promotes contraction operation outputs from lower-precision floating-point types (f16/bf16) to higher-precision types (f32).

This is the counterpart of DemoteContractionInputsPass, which handles input type demotion.


compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,195 @@
...
+          .insert<PromoteContractionOutputsPattern<Float16Type, Float32Type>>(
+              context, ops);
+      break;
+    case PromoteType::BF16:
+      patterns
+          .insert<PromoteContractionOutputsPattern<BFloat16Type, Float32Type>>(
+              context, ops);
+      break;
+    default:
+      llvm_unreachable("Unsupported premotion type");
```
IanWood1:
```suggestion
      llvm_unreachable("Unsupported promotion type");
```

compiler/src/iree/compiler/GlobalOptimization/Passes.h
```diff
@@ -172,6 +172,17 @@ enum class DemoteOperation { All, Conv, Matmul, None };
 std::unique_ptr<Pass>
 createDemoteContractionInputsPass(DemoteType type, DemoteOperation operation);
 
+// Enum for selecting the target type for promotion
+enum class PromoteType { F16, BF16 };
+
+// Enum for selecting which operations to promote
+enum class PromoteOperation { All, Conv, Matmul, None };
+
+// Unified pass for demoting contraction inputs
```
IanWood1:
```suggestion
// Unified pass for promoting contraction inputs
```

compiler/src/iree/compiler/GlobalOptimization/test/promote_contraction_outputs_from_bf16.mlir
```diff
@@ -0,0 +1,349 @@
+// RUN: iree-opt --split-input-file -iree-global-opt-promote-contraction-outputs="type=bf16 operation=matmul" %s | FileCheck %s --check-prefix=F16-MATMUL
+// RUN: iree-opt --split-input-file -iree-global-opt-promote-contraction-outputs="type=bf16 operation=conv" %s | FileCheck %s --check-prefix=F16-CONV
+
+util.func public @matmul_bf16bf16bf16(%arg0 : tensor<100x250xbf16>, %arg1 : tensor<250x500xbf16>,
+    %arg2 : tensor<100x500xbf16>) -> tensor<100x500xbf16> {
+  %0 = linalg.matmul ins(%arg0, %arg1 : tensor<100x250xbf16>, tensor<250x500xbf16>)
+      outs(%arg2 : tensor<100x500xbf16>) -> tensor<100x500xbf16>
+  util.return %0 : tensor<100x500xbf16>
+}
+// F16-MATMUL-LABEL: util.func public @matmul_bf16bf16bf16
```
IanWood1: nit:
```suggestion
// BF16-MATMUL-LABEL: util.func public @matmul_bf16bf16bf16
```

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
```diff
@@ -64,6 +64,26 @@ static llvm::cl::opt<DemoteOperation> clDemoteContractionInputsOperations(
...
+    "iree-global-opt-promote-contraction-outputs-operations",
+    llvm::cl::desc("Select the type of contraction ops to promote."),
+    llvm::cl::values(clEnumValN(PromoteOperation::All, "all",
+                                "Promote all contraction ops."),
+                     clEnumValN(PromoteOperation::Conv, "conv",
+                                "Only promote convolution ops."),
+                     clEnumValN(PromoteOperation::Matmul, "matmul",
+                                "Only promote matmul ops."),
+                     clEnumValN(PromoteOperation::None, "none",
+                                "Promote no contraction ops.")),
```
IanWood1: Do we want/need a test for "all" and "none"?

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,189 @@
...
+    } else if (promoteConv && isa<linalg::Conv2DNhwcFhwcOp>(op)) {
+      replaceOpOutputs(static_cast<linalg::Conv2DNhwcFhwcOp *>(nullptr));
+    } else if (promoteConv && isa<linalg::Conv2DNgchwFgchwOp>(op)) {
+      replaceOpOutputs(static_cast<linalg::Conv2DNgchwFgchwOp *>(nullptr));
+    } else if (promoteConv && isa<linalg::Conv2DNgchwGfchwOp>(op)) {
+      replaceOpOutputs(static_cast<linalg::Conv2DNgchwGfchwOp *>(nullptr));
+    } else {
+      return failure();
+    }
+
```
IanWood1: Non-blocking: this might not be feasible, but it would be nice if we had a way to either mutate the operation or create a new operation with promoted output type through the `linalg::LinalgOp` interface so we don't need to list all types. This should be possible for the demotion pass but might be problematic here because we'd need to change the result type.

compiler/src/iree/compiler/GlobalOptimization/test/promote_contraction_outputs_from_bf16.mlir
```diff
@@ -0,0 +1,349 @@
+// RUN: iree-opt --split-input-file -iree-global-opt-promote-contraction-outputs="type=bf16 operation=matmul" %s | FileCheck %s --check-prefix=F16-MATMUL
```
IanWood1: Also, if this was done through the interface we probably wouldn't need as many tests.

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
HanKuanChen: I don't think we need explicit tests for "all" and "none":

- "all" is just the combination of matmul + conv, already covered by existing tests
- "none" would be a no-op (no changes to the IR)

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: Thanks for the suggestion! I agree that a unified interface would be cleaner. 
I'll keep the current approach for now since it's non-blocking, but this is good to keep in mind for future improvements.

HanKuanChen:
ping

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
```diff
@@ -44,17 +44,45 @@ static llvm::cl::opt<bool> clEnableEarlyMaterialization(
...
+                                "Demote all contraction ops."),
+                     clEnumValN(DemoteOperation::Conv, "conv",
+                                "Only demote convolution ops."),
+                     clEnumValN(DemoteOperation::Matmul, "matmul",
+                                "Only demote matmul ops."),
+                     clEnumValN(DemoteOperation::None, "none",
+                                "Demote no contraction ops.")),
+    llvm::cl::init(DemoteOperation::None));
+
+static llvm::cl::opt<PromoteType> clPromoteContractionOutputsType(
```
MaheshRavishankar: Just FYI, we treat these kind of flags as "developer" flags, and these are not expected to be supported at all. Also they only work with command line interface and do not work with the library interface. So if you are relying on it for a "deployment" use case, then that is inherently unstable.
We obviously wont pull the rug underneath from you, but just letting you know the general expectation of these kind of flags. I have no issues with actually having these flags.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,189 @@
...
+      Value promoteOutput =
+          linalg::GenericOp::create(
+              rewriter, loc, TypeRange{promoteOutputType}, ValueRange{output},
+              ValueRange{promoteEmpty}, maps, iteratorTypes,
+              [&](OpBuilder &b, Location loc, ValueRange args) {
+                Value result = arith::ExtFOp::create(b, loc, destType, args[0]);
+                linalg::YieldOp::create(b, loc, result);
+              })
+              ->getResult(0);
+      using LinalgOpTy = std::remove_pointer_t<decltype(typePtr)>;
```
MaheshRavishankar: I dont think these are expected ways to handle this.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    Type srcType = SrcType::get(rewriter.getContext());
+    if (!llvm::all_of(linalgOp->getOperands(), [&](auto operand) {
+          auto operandType = dyn_cast<RankedTensorType>(operand.getType());
+          return operandType && operandType.getElementType() == srcType;
+        })) {
+      return failure();
+    }
+
+    auto replaceOpOutputs = [&](auto *typePtr) {
```
MaheshRavishankar: Having this as a lambda seems to be just confusing. Just make it a template method defined separately.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,189 @@
...
+    };
+
+    bool promoteMatmul = (promoteOperation == PromoteOperation::All) ||
+                         (promoteOperation == PromoteOperation::Matmul);
+
+    bool promoteConv = (promoteOperation == PromoteOperation::All) ||
+                       (promoteOperation == PromoteOperation::Conv);
+
+    Operation *op = linalgOp.getOperation();
+    if (promoteMatmul && IREE::LinalgExt::isPureMatmul(op)) {
```
MaheshRavishankar: I think you can do this better with TypeSwitch.

```
TypeSwitch<Operation *, void>(op)
   .Case<linalg::MatmulOp, linalg::MatVecOp, ...>(auto typedOp) {
   replaceOutputs(typedOp);
 };
```

MaheshRavishankar (changes requested):

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
HanKuanChen: Thanks for the heads-up. I appreciate the clarification about these being developer flags.
We may use this in deployment. What would be the recommended stable approach?

HanKuanChen:
ping

HanKuanChen:
ping

HanKuanChen:
ping
cc @hanhanW

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,187 @@
...
+                                                destType);
+        return success();
+      }
+    }
+    return llvm::TypeSwitch<Operation *, LogicalResult>(op)
+        .template Case<linalg::MatvecOp, linalg::VecmatOp,
+                       linalg::BatchMatvecOp, linalg::BatchVecmatOp,
+                       linalg::MatmulTransposeAOp, linalg::MatmulTransposeBOp,
+                       linalg::BatchMatmulTransposeAOp,
+                       linalg::BatchMatmulTransposeBOp>([&](auto T) {
```
IanWood1:
```suggestion
                       linalg::BatchMatmulTransposeBOp>([&](auto op) {
```

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,187 @@
...
+          if (!promoteMatmul) {
+            return failure();
+          }
+          replaceOpOutputs<decltype(T)>(linalgOp, rewriter, srcType, destType);
```
IanWood1: Can we just pass the op instead of `linalgOp` so we don't need `decltype(T)`?

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  LogicalResult matchAndRewrite(linalg::LinalgOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (promoteOperation == PromoteOperation::None) {
+      return failure();
+    }
+    if (!isa<linalg::ContractionOpInterface, linalg::ConvolutionOpInterface>(
+            linalgOp.getOperation())) {
+      return failure();
+    }
```
IanWood1: Can you move the `promoteMatmul` and `promoteConv` checks here?

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,187 @@
...
+          ValueRange{promoteEmpty}, maps, iteratorTypes,
+          [&](OpBuilder &b, Location loc, ValueRange args) {
+            Value result = arith::ExtFOp::create(b, loc, destType, args[0]);
+            linalg::YieldOp::create(b, loc, result);
+          })
+          ->getResult(0);
+  auto namedOp = cast<T>(linalgOp.getOperation());
+  auto newLinalgOp = T::create(rewriter, loc, linalgOp.getDpsInputs(),
+                               ValueRange{promoteOutput},
+                               linalg::getPrunedAttributeList(namedOp));
```
IanWood1: I think this removes the non-unit conv stride/dilation.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,187 @@
...
+        replaceOpOutputs<linalg::MatmulOp>(linalgOp, rewriter, srcType,
+                                           destType);
+        return success();
+      }
+      if (IREE::LinalgExt::isPureBatchMatmul(op)) {
+        replaceOpOutputs<linalg::BatchMatmulOp>(linalgOp, rewriter, srcType,
+                                                destType);
+        return success();
+      }
+    }
```
IanWood1: What's the purpose of the isPureMatmul checks? Are these needed?

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
hanhanW: Are we able to use `mlir::clone`?

https://github.com/llvm/llvm-project/blob/b73ce3e53fcb9f72c759139e93db7e16a813c3b2/mlir/lib/Dialect/Utils/StructuredOpsUtils.cpp#L197-L207

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
```diff
@@ -0,0 +1,187 @@
...
+        linalg::YieldOp::create(b, loc, trunc);
+      });
+}
+
+// Template pattern for promoting contraction outputs from a narrower
+// floating-point type.
+// SrcType: The source floating-point type (e.g., BFloat16Type, Float16Type)
+// DestType: The destination floating-point type (e.g., Float32Type)
+template <typename SrcType, typename DestType>
+struct PromoteContractionOutputsPattern
```
hanhanW: Conv is not a contraction op by the definition of linalg. Maybe you want to use `namedOp`? Out of curious, do you want to support generic form as well?

compiler/src/iree/compiler/GlobalOptimization/test/promote_contraction_outputs_from_f16.mlir
hanhanW, line 1: Why two files? Can we have them a single file?

[optional] Some checks can be shared by using CHECK-ALL trick, e.g., https://github.com/iree-org/iree/blob/b0149471b41ac365001d36ef0c9a248b2ecd5311/compiler/src/iree/compiler/Codegen/Common/test/decompose_pack_unpack_ops.mlir#L1-L21

compiler/src/iree/compiler/GlobalOptimization/Passes.h
```diff
@@ -162,11 +162,26 @@ void buildGlobalOptimizationPassPipeline(
...
+enum class PromoteType { F16, BF16 };
+
+// Enum for selecting which operations to promote
+enum class PromoteOperation { All, Conv, Matmul, None };
+
+// Unified pass for promoting contraction inputs
 std::unique_ptr<Pass>
-createDemoteContractionInputsToBF16Pass(DemotionOption option);
+createPromoteContractionOutputsPass(PromoteType type,
+                                    PromoteOperation operation);
```
hanhanW: What is the issue of using the builtin pass option (which is generated by tablegen)?

hanhanW (commented):
I wonder if we can move it to preprocessing. I don't maintain this phase actively now, but adding more experimental flags has the caveat that we could drop them without notifications.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: I think this is required. Please see https://github.com/iree-org/iree/commit/3df06b73d17bb327204e9ebfa5f18c091da82fc2.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: solved in https://github.com/iree-org/iree/pull/23824/changes/8a6bd35dd679027dde542fc5e0f924a7794183be

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: solved in https://github.com/iree-org/iree/pull/23824/changes/bf0780bbc3bbfd808ddeb6296a9cd88b679ab290

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: solved in https://github.com/iree-org/iree/pull/23824/commits/1923ec873086077dfacdd9fbf74487767c8d937e

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: solved in https://github.com/iree-org/iree/pull/23824/commits/ec550d1ccdbacd9aacdc0449b01310f3c0568cf5 and https://github.com/iree-org/iree/pull/23824/commits/7538ae99e130aeeb86cb5ab1c6f37e69d1633188

compiler/src/iree/compiler/GlobalOptimization/test/promote_contraction_outputs_from_f16.mlir
HanKuanChen: solved in https://github.com/iree-org/iree/pull/23824/commits/e68d6a6f747a1ef66bab165e9b9d384194d938e9

compiler/src/iree/compiler/GlobalOptimization/Passes.h
HanKuanChen: No issue. But options in compiler/src/iree/compiler/GlobalOptimization/Passes.cpp is required for iree-compile.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
hanhanW: My question is not answered. Using `mlir::clone` allows us to get rid of template, which looks better to me.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
hanhanW: This is not answered

compiler/src/iree/compiler/GlobalOptimization/Passes.h
hanhanW: Let's use the option in Passes.cpp. Maintaining these wrappers is a burden, IMO.

hanhanW (commented):
I did not do a review because some questions are not answered.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: 1. I followed the naming from the existing `DemoteContractionInputs` pass. Renaming only this pass would create inconsistency. Should I rename both passes?
2. Supporting generic operations is not in my plan.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
HanKuanChen: I cannot use `mlir::clone` here. The issue is that `mlir::clone` copies the region from the original operation, but the region's body contains yield operations with the old type (e.g., `linalg.yield bf16`). When I change the output type to `f32`, the cloned region still yields `bf16`, which creates a type mismatch and fails verification.

I cannot find a way to automatically update the region when changing the operation's result type. The template approach with `T::create` is the standard way to regenerate the region with the correct types.

I did explore using `mlir::cloneWithoutRegions` + `getRegionBuilder()`, but that requires significantly more code (manual block creation, type computation, etc.) without providing any real benefit over the simpler template approach.

compiler/src/iree/compiler/GlobalOptimization/Passes.h
HanKuanChen: solved in https://github.com/iree-org/iree/pull/23824/changes/3e2dd183f2771b4196688ed328b086ffb580c5fb

hanhanW:
> I wonder if we can move it to preprocessing. I don't maintain this phase actively now, but adding more experimental flags has the caveat that we could drop them without notifications.

I have a quick chat with @bjacob on discord: https://discord.com/channels/689900678990135345/1456679894812590326/1504186688216502302

I think we should move them to preprocessing, because we will be able to handle it in data-tiling. The demotion is useful independently of data-tiling, so it is better to keep it in IREE. IMO, preprocessing is a better home for them; it will reduce the "review noise", because the bar of preprocessing pass is lower versus experiemental pass in other phases IMO.

I still can help review the code. @IanWood1 I'll take over the review, so you can save some time (unless you're interested in reviewing the patch). @HanKuanChen can you move it to https://github.com/iree-org/iree/tree/main/compiler/src/iree/compiler/Preprocessing/Common and update the pass/pattern name? It is not contraction only. I understand your point about consistency, and I'll make the rename when I move the pass. It is easier to me if we can make it correct in the first place. Then the PR should be ready to go.

compiler/src/iree/compiler/GlobalOptimization/PromoteContractionOutputs.cpp
hanhanW: Thanks for the try, then `T::create` is good! (It is easier for reviewers if you can reply what you've done or why we don't apply the review comment.)

HanKuanChen:
ping

bjacob:
> ping

Thanks for moving to Preprocessing like previously requested. I have triggered CI, I'll review once it's passed.

compiler/src/iree/compiler/Preprocessing/Common/PromoteLinalgOutputs.cpp
```diff
@@ -0,0 +1,164 @@
...
+                       linalg::MatmulTransposeAOp, linalg::MatmulTransposeBOp,
+                       linalg::BatchMatmulTransposeAOp,
+                       linalg::BatchMatmulTransposeBOp, linalg::Conv2DOp,
+                       linalg::Conv2DNchwFchwOp, linalg::Conv2DNhwcHwcfOp,
+                       linalg::Conv2DNhwcFhwcOp, linalg::Conv2DNgchwFgchwOp,
+                       linalg::Conv2DNgchwGfchwOp>([&](auto op) {
+          replaceOpOutputs(op, rewriter, srcType, destType);
+          return success();
+        })
+        .Default(failure);
```
bjacob, line 126: This handles this finite list of Linalg named ops, but doesn't handle linalg.generic (or any future named op not listed here, but the bigger problem is not handling generic).

But I just checked and DemoteContractionInputs seems to be similar, so at least it's consistent with that.

OK to merge! Let me know if you need me to do so. Let's also wait a few days to see if another reviewer chimes in.

bjacob (approved):

compiler/src/iree/compiler/Preprocessing/Common/PromoteLinalgOutputs.cpp
bjacob: There is a real Bazel build error: https://github.com/iree-org/iree/actions/runs/25905555752/job/77091908873?pr=23824#step:5:1053

```
compiler/src/iree/compiler/Preprocessing/Common/PromoteLinalgOutputs.cpp:7:10: error: module //compiler/src/iree/compiler/Preprocessing/Common:Transforms does not depend on a module exporting 'iree/compiler/Dialect/LinalgExt/Utils/Utils.h'
#include "iree/compiler/Dialect/LinalgExt/Utils/Utils.h"
         ^
compiler/src/iree/compiler/Preprocessing/Common/PromoteLinalgOutputs.cpp:13:10: error: module //compiler/src/iree/compiler/Preprocessing/Common:Transforms does not depend on a module exporting 'mlir/Dialect/Linalg/Utils/Utils.h'
#include "mlir/Dialect/Linalg/Utils/Utils.h"
```

It's saying that you need to edit the `deps` attribute of the `Transforms` target here:
https://github.com/iree-org/iree/blob/2098bbef0425906ad53860aea469e14528466912/compiler/src/iree/compiler/Preprocessing/Common/BUILD.bazel#L58

and in that `deps` list, you need to add (at the right position to preserve alphabetic sorting) whichever target that exports (in its `hdrs` attribute) the file iree/compiler/Dialect/LinalgExt/Utils/Utils.h. Looking up... seems to be: `"//compiler/src/iree/compiler/Dialect/LinalgExt/Utils"`.

bjacob (approved):

hanhanW (approved):
Hey, sorry for the delayed. Honestly I don't review the final code, because I'm ooo. Removing the blocker from myself and thanks for Benoit for the final review!

HanKuanChen:
ping

bjacob:
Do you want us to merge this?  The CI failure is just an infrastructure issue not related to your PR.

HanKuanChen:
Thanks. I don't have the permission to merge PRs.

bjacob:
It is merged now :-)

PR #23828, opened 2026-03-18
[Tokenizer] Fix silent output truncation in encode/decode finalize

jtuyls:
## Summary

Return RESOURCE_EXHAUSTED instead of silently truncating when the output buffer is too small, matching the documented API contract in tokenizer.h.

### HuggingFace smoketest (`huggingface_smoketest.py`)

**1667/1667** tokenization comparisons pass across ~80 HuggingFace models (0 mismatches).

76 additional tests fail to **load** (not tokenization mismatches) — these are tiktoken models listed in the HF smoketest that aren't valid HuggingFace model identifiers. They're tested by the dedicated tiktoken smoketest instead. These tests are removed from the HF smoketest in https://github.com/iree-org/iree/pull/23830.

### Tiktoken smoketest (`tiktoken_smoketest.py`)

**72/76** tokenization comparisons pass across 4 tiktoken encodings (cl100k_base, o200k_base, r50k_base, p50k_base).

4 failures — **identical between upstream and this PR** (pre-existing, same as `fix-tokenizer-added-tokens` branch):

| Encoding | Failing test | Cause |
|----------|-------------|-------|
| cl100k_base | `special_token_endoftext` | IREE matches `<\|endoftext\|>` as special token; tiktoken's `encode_ordinary` treats it as literal text |
| o200k_base | `special_token_endoftext` | Same |
| r50k_base | `special_token_endoftext` | Same |
| p50k_base | `special_token_endoftext` | Same |

Root cause: the IREE tokenizer has no "encode ordinary" mode (equivalent to tiktoken's `disallowed_special=()`). Fixed in https://github.com/iree-org/iree/pull/23830.

### DIFF
--- a/runtime/src/iree/tokenizer/tokenizer_decode_test.cc
+++ b/runtime/src/iree/tokenizer/tokenizer_decode_test.cc
@@ -1250,6 +1250,81 @@ TEST_F(TokenizerDecodeTest, ByteFallbackInvalidContinuation) {
   iree_tokenizer_free(tokenizer);
 }
 
+//===----------------------------------------------------------------------===//
+// Output Buffer Overflow Detection
+//
+// Verifies that decode functions return IREE_STATUS_RESOURCE_EXHAUSTED when the
+// output buffer is too small, rather than silently truncating.
+//===----------------------------------------------------------------------===//
+
+class TokenizerDecodeOverflowTest : public ::testing::Test {};
+
+// Batch decode with output buffer too small for the decoded text.
+// "Hello" = 5 bytes, but buffer only holds 1.
+TEST_F(TokenizerDecodeOverflowTest, BatchDecodeOutputBufferTooSmall) {
+  ScopedVocabBuilder vocab_builder;
+  vocab_builder.AddToken(0, "Hello");
+  vocab_builder.AddToken(1, "world");
+  ScopedVocab vocab = vocab_builder.Build();
+
+  iree_tokenizer_t* tokenizer = BuildByteLevelTokenizer(vocab.release());
+  ASSERT_NE(tokenizer, nullptr);
+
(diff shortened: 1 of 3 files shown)


runtime/src/iree/tokenizer/tokenizer_decode_test.cc
```diff
@@ -1250,6 +1250,85 @@ TEST_F(TokenizerDecodeTest, ByteFallbackInvalidContinuation) {
...
+  // Finalize with 5 bytes — fits one U+FFFD (3 bytes) but not two (6 bytes).
+  char finalize_output[5];
+  iree_host_size_t finalize_written = 0;
+  iree_status_t status = iree_tokenizer_decode_state_finalize(
+      state, iree_make_mutable_string_view(finalize_output, 5),
+      &finalize_written);
+  EXPECT_TRUE(iree_status_is_resource_exhausted(status))
+      << "Expected RESOURCE_EXHAUSTED when only partial flush fits";
+  EXPECT_EQ(finalize_written, 3u) << "Should have flushed one U+FFFD (3 bytes)";
+  iree_status_free(status);
```
benvanik: use IREE_EXPECT_STATUS_IS - it let's you do this in one line (it consumes status, so no need to free). here and elsewhere.
e.g.
  IREE_EXPECT_STATUS_IS(IREE_STATUS_INVALID_ARGUMENT,
                        loom_tokenizer_consume_status(t.get()));

benvanik (changes requested):

runtime/src/iree/tokenizer/tokenizer_decode_test.cc
jtuyls: Thanks, updated it.

benvanik (approved):

PR #23832, opened 2026-03-18
[Tuner][Codegen] Expose iree codegen ops to py bindings

RattataKing:
Registered iree codegen ops to python bindings, to allow usage of
`op = ir.get_ops_by_type(input_module, iree_codegen.OpType)`

### DIFF
--- a/compiler/bindings/python/test/api/tuner_api_test.py
+++ b/compiler/bindings/python/test/api/tuner_api_test.py
@@ -627,17 +627,8 @@ def test_get_iree_constraints_op():
             iree_codegen.smt.constraints
                 target = <set = 0>,
                 pipeline = LLVMGPUVectorDistribute,
-                knobs = {wg_m = #iree_codegen.smt.int_knob<"wg_m">,
-                         mma_idx = #iree_codegen.smt.int_knob<"mma_idx">}
+                knobs = {}
                 dims() {
-                ^bb0:
-                    %wg_m = iree_codegen.smt.knob "wg_m" : !smt.int
-                    %idx = iree_codegen.smt.knob "mma_idx" : !smt.int
-                    %mma_m = iree_codegen.smt.lookup %idx [0, 1] -> [16, 32] : !smt.int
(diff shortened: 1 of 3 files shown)


compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -618,3 +618,56 @@ def test_one_of_knob_attr():
...
+                    pipeline = LLVMGPUVectorDistribute,
+                    knobs = {}
+                    dims() {
+                    }
+                return
+            }
+        }
+    """
+    input_module = ir.Module.parse(module_str)
+    # Test if IREE Codegen Op (eg. ree_codegen.ConstraintsOp) types are exposed
```
kuhar:
```suggestion
    # Test if IREE Codegen Op (eg., `iree_codegen.ConstraintsOp`) types are exposed
```

compiler/bindings/python/test/api/tuner_api_test.py
```diff
@@ -618,3 +618,56 @@ def test_one_of_knob_attr():
...
+                         mma_idx = #iree_codegen.smt.int_knob<"mma_idx">}
+                dims() {
+                ^bb0:
+                    %wg_m = iree_codegen.smt.knob "wg_m" : !smt.int
+                    %idx = iree_codegen.smt.knob "mma_idx" : !smt.int
+                    %mma_m = iree_codegen.smt.lookup %idx [0, 1] -> [16, 32] : !smt.int
+                    %cond = smt.int.cmp le %wg_m, %wg_m
+                    %cond_mma = smt.int.cmp le %mma_m, %wg_m
+                    iree_codegen.smt.assert %cond, "wg_m <= wg_m" : !smt.bool
+                    iree_codegen.smt.assert %cond_mma, "mma_m <= wg_m" : !smt.bool
```
kuhar: in this testcase, we don't really need to have a full constraints op -- we can use one with just one or two ops and it should be just fine

kuhar (approved):

PR #23833, opened 2026-03-18
Revert "[GPUHeuristics] Further tune `LargeGemm` perf (#23652)"

lialan:
This reverts commit 954eddcf1ec08a415a1a0d0749c180b2bed49e25.

Need to look into gfx942 regression.


efric (approved):

lialan:
Maybe we should hold off from merging it. I need to double check to make sure the CDNA3 path is not altered by this PR.

PR #23834, opened 2026-03-18
[GPUHeuristics] Improve MMA intrinsic selection and tiling for small-channel group backward convolutions

yzhang93:
Three changes to improve performance of grouped convolutions with small per-group channel counts (e.g., 32 groups with 8 input/output channels):

1. MMA intrinsic selection: Add MN utilization check as the highest-priority rule. When one intrinsic has >= 2x better MN utilization, prefer it regardless of K-alignment. This avoids selecting 32x32 intrinsics (6.25% util) over 16x16 intrinsics (25% util) for problems with small M/N dimensions. For same MN intrinsics differing only in K (e.g., 16x16x16 vs 16x16x32), use a 10% utilization threshold: at moderate util prefer smaller compute; at very low util let later rules pick the larger-K intrinsic for better K throughput.

2. Degenerate schedule fallback: When GCD-based distribution produces all-1 subgroup/tile counts (because tile counts are small odd numbers like 3x3 with GCD=1 against power-of-2 seeds), fall back to min-based distribution. This ensures the schedule assigns meaningful work to subgroups for small filter spatial dimensions.

3. Batch dimension tiling: When the inner M and N dimensions both require padding (smaller than the intrinsic tile), increase the batch tile size from 1 to up to 4. This gives each workgroup more useful work to amortize dispatch and memory access overhead, which is critical for grouped convolutions where the per-group computation is very small.

Benchmark results on MI355X:
  - Top improvements (group backward weight convolutions, g=32): 
  3.94x (2466 -> 627 us)  g=32, 256, 200x200, 3x3, stride 2 
  3.51x (1929 -> 550 us)  g=32, 256, 100x100, 3x3
  1.94x (370 -> 191 us)   g=32, 512, 50x50, 3x3
  - No significant regressions.
 
Similar performance improvements are also observed on RDNA4 and Mi300x.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -842,9 +842,24 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   SmallVector<int64_t> workgroupTileSizes(bounds.size(), 0);
   SmallVector<int64_t> reductionTileSizes(bounds.size(), 0);
   SmallVector<int64_t> subgroupTileSizes(bounds.size(), 0);
-  // Tile all batch dimensions with unit size.
+
+  // Tile batch dimensions. Normally each workgroup handles one batch element.
+  // When both M and N need padding (e.g., grouped convolutions with small
+  // per-group channels), tile batch up to 4 to give each workgroup more
+  // useful work and amortize dispatch overhead.
+  bool needsMNPadding = problem.mSizes.back() < schedule->getTotalMSize() &&
+                        problem.nSizes.back() < schedule->getTotalNSize();
   for (int64_t batch : contractionB) {
-    workgroupTileSizes[batch] = 1;
+    int64_t batchTile = 1;
+    if (needsMNPadding && !ShapedType::isDynamic(bounds[batch])) {
+      for (int64_t t = 4; t >= 2; t /= 2) {
+        if (bounds[batch] % t == 0) {
+          batchTile = t;
+          break;
+        }
+      }
+    }
+    workgroupTileSizes[batch] = batchTile;
   }
 
   // Tile all m, n, k and k_b dimensions to 1 except the innermost. Unit dims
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -842,9 +842,24 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   SmallVector<int64_t> workgroupTileSizes(bounds.size(), 0);
   SmallVector<int64_t> reductionTileSizes(bounds.size(), 0);
   SmallVector<int64_t> subgroupTileSizes(bounds.size(), 0);
-  // Tile all batch dimensions with unit size.
+
+  // Tile batch dimensions. Normally each workgroup handles one batch element.
```
nirvedhmeshram: This looks dangerous to do outside the heuristic code, because we are changing the shared memory usage implicitly? if it was already near threshold it would exceed the capacity? Would it be possible to do this inside the heuristic logic where it finds the tile sizes.

yzhang93: Good point. I moved it to `deduceMMASchedule` function.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -786,10 +786,14 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
   // Intentionally padded GEMM proved to be beneficial for performance for
   // the following layouts: 1) [M, K] x [K, N] 2) [M, K] x [N, K]
   // Therefore we disallow padding only when LHS is transposed.
+  // Include all batch dims (static and dynamic) so the heuristic can compute
+  // per-dimension batch tile sizes aligned with contractionB indices.
+  auto allBatchBounds = llvm::map_to_vector(
```
nirvedhmeshram: This is diverging from what we do for other dimensions which is fine, I just want to make sure we are doing it for a good reason? For other dimensions we will always tile outer dimensions to 1 and only the inner dimensions get something meaningful thats why we didnt care about skipping the dynamic dims, but the logic here is setup in a way that you can have multiple non unit batch dims, is that something you have found necessary?

nirvedhmeshram (approved):
LGTM, I do have one question but the answer is not a blocker, just a good to know.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
yzhang93: No, I haven't found cases where multiple non-unit batch dims are needed (since there is no multiple batch dims in real applications). However, to keep consistent with other dimensions, I updated the codes to pass only static batch dims (using existing `getDimBoundsNoPad(batchDims)`) and only tile the innermost dim if possible.

PR #23835, opened 2026-03-18
[Codegen] Fix the reshape propagation in InsertBatchDimForBatchlessConv

### LINKED ISSUE #23812: ShuffleNet | Conv + reshape + Batchnorm + MapStore Dispatch
With #23377 we see that due to conv generalization -> aggressive fusion takes place and we have a dispatch with the following structure :-
```
  %a = fill
  %b = conv %a
  %c = reshape ops %b
  %d = batchnorm %c
  %e = pack op (later folded into MapStore)
```
[Here](https://gist.github.com/Abhishek-Varma/7bcb10cb9696e9f098b44fc1487896a2) is the IR.

This currently gets assigned CPUDoubleTilingExpert and leads to :-
```
error: write affecting operations on global resources are restricted to workgroup distributed contexts.
```
This is because `map_store` is found to be outside the workgroup scf.forall.

And I tried to make it go through the `CPUConvTileAndDecomposeExpert` like other generic conv op support. Therefore the above IR will shove `CPUConvTileAndDecomposeExpert`.

Now, this lead to large vector sizes. The state of the IR after tile and fuse is [here](https://gist.github.com/Abhishek-Varma/01776a4a9c0853a85777360bccfa3b6b).

I tried looking into MMT4D's pipeline to make use of :-
```
funcPassManager.addPass(createLLVMCPUTileLastOpAndFuseProducerConsumerPass(
      IREE::CPU::TilingLevel::VectorInnerParallelTiles));
```
But the map_store wasn't being solved because it wasn't still inside the workgroup scf.forall as the generic conv.

For this I tried implementing a stopgap tile+fuse pass which will tile through the reshapes -> batchnorm -> map_store but it introduced lot of non-trivialities.

@hanhanW @Max191 - can I please get some pointers for this ?

hanhanW:
The phase 2 propagation should only allow `collapse_shape`. Otherwise, the `expand_shape` op can be bubbled up first. Then it becomes a blocker for sinking down `collapse_shape`.

```
collapse_shape
linalg.generic
expand_shape
```

Fixes https://github.com/iree-org/iree/issues/23812

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/InsertBatchDimForBatchlessConv.cpp
@@ -245,11 +245,15 @@ struct InsertBatchDimForBatchlessConvPass final
       populateReshapeToInterfaceTensorPatterns(reshapePatterns);
 
       linalg::ControlFusionFn controlFn = [&](OpOperand *fusedOperand) {
-        Operation *op = fusedOperand->getOwner();
-        if (op->getBlock() != newConvOp->getBlock()) {
+        auto collapseOp =
+            fusedOperand->get().getDefiningOp<tensor::CollapseShapeOp>();
+        if (!collapseOp) {
           return false;
         }
-        return newConvOp->isBeforeInBlock(op);
+        if (collapseOp->getBlock() != newConvOp->getBlock()) {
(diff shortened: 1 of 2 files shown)


Abhishek-Varma (approved):
Thanks for this @hanhanW ! It worked.

PR #23837, opened 2026-03-18
Move unrolling to common and use in CPU pipeline

amd-eochoalo:
UnrollElementwiseOps is an pattern that we currently apply exclusively to the GPU pipeline.
This pattern takes arith elementwise operations with vector operands and unroll them in a rank reduction way until 1-D vector operations remain.

This PR partly addresses https://github.com/iree-org/iree/issues/23094.
By unrolling arith elementwise operations, the LLVM backend will not translate them into nested arrays of vectors.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 The IREE Authors
+// Copyright 2026 The IREE Authors
 //
 // Licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
--- a/compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
@@ -1,0 +1,72 @@
+// Copyright 2025 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/Transforms.h"
+#include "mlir/Dialect/UB/IR/UBOps.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/PatternMatch.h"
+
+namespace mlir::iree_compiler {
+
+namespace {
+
+// Unrolls N-D elementwise vector operations to 1-D by iterating over all
+// leading dimensions and extracting/inserting 1-D subvectors.
+struct UnrollElementwiseOps final : RewritePattern {
+  UnrollElementwiseOps(MLIRContext *context, PatternBenefit benefit = 1)
+      : RewritePattern(MatchAnyOpTypeTag(), benefit, context) {}
+
+  LogicalResult matchAndRewrite(Operation *op,
+                                PatternRewriter &rewriter) const override {
+    if (!OpTrait::hasElementwiseMappableTraits(op) ||
+        op->getNumResults() != 1) {
+      return failure();
+    }
+
+    Location loc = op->getLoc();
+    VectorType dstVecTy = dyn_cast<VectorType>(op->getResult(0).getType());
+    if (!dstVecTy || dstVecTy.getRank() <= 1) {
+      return failure();
+    }
+    ArrayRef<int64_t> originalSize = dstVecTy.getShape();
+
+    Value result = ub::PoisonOp::create(rewriter, loc, dstVecTy);
+    auto subVecTy =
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+  LogicalResult matchAndRewrite(Operation *op,
+                                PatternRewriter &rewriter) const override {
+    if (!OpTrait::hasElementwiseMappableTraits(op) ||
+        op->getNumResults() != 1) {
+      return failure();
+    }
+
+    Location loc = op->getLoc();
+    VectorType dstVecTy = dyn_cast<VectorType>(op->getResult(0).getType());
```
hanhanW, line 31: style nit: use auto

compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
hanhanW, line 1: It's better to emit mismatching message on failure, i.e., use notifyMatchFailure.

compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
```diff
@@ -0,0 +1,72 @@
+// Copyright 2025 The IREE Authors
```
hanhanW: 2026

compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace mlir::iree_compiler {
+
+namespace {
+
+// Unrolls N-D elementwise vector operations to 1-D by iterating over all
+// leading dimensions and extracting/inserting 1-D subvectors.
+struct UnrollElementwiseOps final : RewritePattern {
+  UnrollElementwiseOps(MLIRContext *context, PatternBenefit benefit = 1)
+      : RewritePattern(MatchAnyOpTypeTag(), benefit, context) {}
```
hanhanW, line 21: Can't it just be `using Base::Base;`?

compiler/src/iree/compiler/Codegen/LLVMCPU/ConvertToLLVM.cpp
```diff
@@ -1027,6 +1027,7 @@ void ConvertToLLVMPass::runOnOperation() {
     RewritePatternSet vectorToLoopsPatterns(&getContext());
     populateVectorToSCFConversionPatterns(
         vectorToLoopsPatterns, VectorTransferToSCFOptions().enableFullUnroll());
+    populateUnrollElementwiseOpsPatterns(vectorToLoopsPatterns);
```
hanhanW, line 1030: We can keep this, because there are some patterns may generate n-D vectors after vector lowering. However, I'd expect a new pass added to below. It likely should happen after virtual vector lowering.

https://github.com/iree-org/iree/blob/30c3eaa2e59a7b72fbb814539337be6f3cd416b2/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L108-L127

or we populate the patterns within the pass:

https://github.com/iree-org/iree/blob/f69c15c35d030b4359b1eb74836a83a6aeb136be/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUVirtualVectorLowering.cpp#L74-L101


btw, you should move this line to the above snippet. It belongs to "Run Vector -> Vector transformations ahead of conversion to LLVM.".

compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
kuhar:
```suggestion
// Copyright 2026 The IREE Authors
```

kuhar (approved):

hanhanW (changes requested):

compiler/src/iree/compiler/Codegen/Common/UnrollElementwiseOps.cpp
kuhar: this is pre-existing code -- I'd limit this PR to simple code motion

hanhanW: I think you left similar comments in my other PR, like https://github.com/iree-org/iree/pull/23713#discussion_r2918169413 ?

It is a small change, and I think improving the code quality when moving the code is okay. If it is a big change, I'd ask for doing it in a follow-up. I'm open to both though.

kuhar: oh, I didn't realize that was also doing code motion

Groverkss (changes requested):
Let's discuss before landing this today.

hanhanW:
> Let's discuss before landing this today.

Can someone put a summary here? There was not an agenda last night, so I missed the meeting...

amd-eochoalo:
> Can someone put a summary here?

This will not be merged, instead I'll take over https://github.com/iree-org/iree/pull/23853 .

PR #23838, opened 2026-03-18
Integrate llvm/llvm-project@eabcfcee0803

efric:
Carrying reverts:

- https://github.com/llvm/llvm-project/commit/429e9717e232 (adds MLIRAsmParser dep to ArithUtils
in CMake but missing Bazel overlay port, breaks linux_x64_bazel)
- llvm/llvm-project@a1a714b8b87e (Make getMutableSuccessorOperands overridable on ReturnLike ops, breaks torch-mlir build — three ops define custom getMutableSuccessorOperands but DeclareOpInterfaceMethods no longer generates the declaration since the method now has a default). Fix requires change in torch-mlir as well; temporary revert for now and will add back later.

Other points of interest:

- llvm/llvm-project@4f6379069eb2 added a createTosaAttachTarget pass inside addTosaToLinalgPasses with a default that only declares the "doubleround" extension. This overwrites IREE's earlier createTosaAttachTarget that declared the "dynamic" extension, causing tosa.table with variable table values to be rejected by validation.

- Fix by passing IREE's target options (with both "dynamic" and "doubleround" extensions) through the new attachTargetOptions parameter and removing the now-redundant standalone createTosaAttachTarget call.

assisted by:claude

### DIFF
--- a/compiler/plugins/input/TOSA/InputConversion/Passes.cpp
+++ b/compiler/plugins/input/TOSA/InputConversion/Passes.cpp
@@ -54,11 +54,11 @@ void buildTOSAInputConversionPassPipeline(OpPassManager &passManager) {
   tosaToLinalgNamedOptions.preferConv2DKernelLayoutHWCF = true;
   tosa::TosaValidationOptions tosaValidationOptions;
   tosa::TosaAttachTargetOptions tosaTargetOptions;
-  tosaTargetOptions.extensions = {"dynamic"};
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):
lgtm though I wonder if we still need that arith revert - Bazel tends to get fixed up pretty quick

efric:
> lgtm though I wonder if we still need that arith revert - Bazel tends to get fixed up pretty quick

we still needed it at the time of this integrate but maybe not anymore today. we'll see soon!

PR #23841, opened 2026-03-18
[GPUHeuristics] Fix RDNA4 conv heuristic seeds to resolve regressions

yzhang93:
After the previous RDNA4 seed tuning commit, several medium/large-sized convolution shapes showed significant regressions. To fix the issue, increase conv subgroupCount seed from 4 to 8. This change will bring back the performance upto 2x faster, while not introduce new regressions.

Also update the convolution tests to use IGEMM path with use_igemm=true.


nirvedhmeshram (approved):
LGTM

PR #23842, opened 2026-03-18
Revert "Add F16 support to contraction input demotion pass"

hanhanW:
Reverts iree-org/iree#23621

It breaks MSVC builds: https://github.com/iree-org/iree/actions/runs/23209911746/job/67455603181


hanhanW:
@HanKuanChen  we can add `ci-extra:windows_x64_msvc` to the end of PR description, which verifies the fix, when you re-apply the change + the fix.

kuhar (approved):
cc: @efric

efric:
@HanKuanChen fyi this seemed to do the trick https://github.com/iree-org/iree/pull/23846 , though you may want to double check

PR #23843, opened 2026-03-18
[CI] Move windows_x64_msvc to presubmit

efric:
Moves `windows_x64_msvc` to presubmit (and fix a typo and update doc while we're here). This will help catch occasional msvc specific bugs that surface atm on postsubmit.


build_tools/github_actions/configure_ci.py
```diff
@@ -188,7 +187,7 @@ def contains(cls, val):
             "*win32*",
             "*windows*",
             "*msvc*",
-            ".github/worklflows/ci_windows_x64_msvc.yml",
```
kuhar, line 191: yikes

kuhar (approved):
Thanks, let's land after the build is green again

PR #23846, opened 2026-03-18
[GlobalOptimization] Fix MSVC issue in DemoteContractionInputs#23845

efric:
Remove the explicit constructor taking `const DemoteContractionInputsPassOptions &` which was redundant with the inherited Base constructor (via `using Base::Base`).

The tablegen-generated PassBase already provides a constructor taking `DemoteContractionInputsPassOptions` by value that initializes all option fields. The explicit by-const-ref overload created an ambiguity (C2668) on MSVC when `std::make_unique` forwarded an rvalue — both the inherited by-value constructor and the explicit by-const-ref constructor were equally viable candidates.

Assisted by: Claude

ci-extra: windows_x64_msvc


hanhanW:
I'll ask @HanKuanChen to reland it with the fix, thanks for taking care it.

PR #23847, opened 2026-03-18
[Codegen][GPU] Add iree_gpu.arg_compare operation

bangtianliu:
This PR adds the `iree_gpu.arg_compare` operation to represent the distributed form of `iree_vector_ext.arg_compare`.

The next step is to implement lowering for `iree_gpu.arg_compare` using DPP and rocdl::ballot instructions.

Note: I will continue extending arg_compare under the GPU dialect. For now, this change focuses on enabling it to flow through the vector distribute pipeline for ArgCompare.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
@@ -172,7 +172,8 @@ def IREEGPU_ValueBarrierOp : Op<IREEGPU_Dialect, "value_barrier", [
 
 def IREEGPU_YieldOp : Op<IREEGPU_Dialect, "yield", [
     Pure, ReturnLike, Terminator,
-    HasParent<"::mlir::iree_compiler::IREE::GPU::BarrierRegionOp">]> {
+    ParentOneOf<["::mlir::iree_compiler::IREE::GPU::BarrierRegionOp",
+                 "::mlir::iree_compiler::IREE::GPU::ArgCompareOp"]>]> {
   let summary = [{Yield values from a iree_gpu region.}];
   let description = [{
      This operation is used to yield values from a within a region.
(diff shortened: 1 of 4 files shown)


kuhar (commented):
Can you explain why we need it in addition to vector_ext.arrg_compare? Both operate on vector operands.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -384,4 +385,114 @@ def IREEGPU_GlobalSubgroupBarrierOp : Op<IREEGPU_Dialect,
...
+    AnyVectorOfAnyRank:$init_value,
+    AnyVectorOfAnyRank:$init_index,
+    Optional<AnyType>:$index_base,
+    I64Attr:$dimension
+  );
+
+  let results = (outs
+    AnyVectorOfAnyRank:$result_value,
+    AnyVectorOfAnyRank:$result_index
+  );
```
kuhar, line 449: This is identical to https://github.com/iree-org/iree/blob/6f5a868644daf3ae22215732168ff500ef5b5545/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.td#L442-L454

bangtianliu:
> Can you explain why we need it in addition to vector_ext.arrg_compare? Both operate on vector operands.

When distributing `iree_vector_ext.arg_compare` after vectorization, we can't use `gpu.subgroup_reduce` because it doesn't track indices. `gpu.subgroup_reduce` lowers to DPP instructions, but we need to track both the reduced value and its index.                                                                                                                  
                                                                                                                                                                                            
Ideally, we'd define `gpu.subgroup_argreduction` upstream (sth like that) and add proper lowering. For now, we're defining the op in iree_gpu and lowering directly to DPP + rocdl::ballot. Long-term, we should propose gpu.subgroup_argreduction upstream.

kuhar:
But why can't you first distribute `iree_vector_ext.arg_compare` (to smaller `arg_compare`s) and then lower it to subgroup operations?

bangtianliu:
> But why can't you first distribute `iree_vector_ext.arg_compare` (to smaller `arg_compare`s) and then lower it to subgroup operations?

Good point. We could lower to smaller arg_compare ops, but we need a way to indicate that the op represents a cross-thread subgroup reduction (tracking both value and index), which is semantically different from the pre-distribution `iree_vector_ext.arg_compare`.

kuhar:
You could stick an attribute on it if you can't tell otherwise.

bangtianliu:
> You could stick an attribute on it if you can't tell otherwise.

Yes, that could work since the types are the same. I'll update this PR to use an attribute on `iree_vector_ext.arg_compare` instead.

kuhar:
Discussed in a meeting, we probably want to go directly to gpu.shuffle and possibly introduce gpu.ballot if required for performance.

PR #23848, opened 2026-03-18
[CI] Add auto run generate_exports.py to pre-commit

RattataKing:
It is easy to forget updating C API exports when adding new Python bindings.
This change adds a pre-commit hook that runs `generate_exports.py` whenever header files under compiler/bindings/c/iree/compiler/ are modified.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Checking out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Initializing llvm-project submodule
+        run: git submodule update --init --depth=1 third_party/llvm-project
       - name: Setting up python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
       - name: Running pre-commit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -126,6 +126,6 @@ repos:
         name: Regenerate CAPI export files
         entry: compiler/src/iree/compiler/API/generate_exports.py
         language: python
-        files: '^compiler/bindings/python/'
+        files: '^compiler/bindings/c/iree/compiler/.*\.h$'
 
     # TODO(scotttodd): mypy type checking for Python (https://mypy-lang.org/)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -122,4 +122,10 @@ repos:
         language: fail
         files: "BUILD$"
 
+      - id: generate_capi_exports
+        name: Regenerate CAPI export files
+        entry: compiler/src/iree/compiler/API/generate_exports.py
+        language: python
+        files: '^compiler/bindings/python/'
+
     # TODO(scotttodd): mypy type checking for Python (https://mypy-lang.org/)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -127,5 +127,6 @@ repos:
         entry: compiler/src/iree/compiler/API/generate_exports.py
         language: python
         files: '^compiler/bindings/c/iree/compiler/.*\.h$'
+        pass_filenames: false
 
     # TODO(scotttodd): mypy type checking for Python (https://mypy-lang.org/)


bangtianliu (approved):
Thanks, this is very helpful.

kuhar (commented):
Looks great, thanks for automating this.

Have you tried adding a new CAPI to see what the output looks like? Could you show an example?

hanhanW (commented):
Does it print out how to fix the failure (e.g., running the script) when it fails?

.github/workflows/lint.yml
```diff
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Checking out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Initializing llvm-project submodule
+        run: git submodule update --init --depth=1 third_party/llvm-project
```
ScottTodd, line 21: This added an extra 1 minute to the lint action, so it now takes closer to 2 minutes total (https://github.com/iree-org/iree/actions/runs/23275039150/job/67676136414?pr=23848). A sparse checkout of the submodule just for the `mlir-c` subdirectory may be worth it here. https://github.com/iree-org/iree/blob/6f5a868644daf3ae22215732168ff500ef5b5545/compiler/src/iree/compiler/API/generate_exports.py#L122-L123

.pre-commit-config.yaml
```diff
@@ -122,4 +122,11 @@ repos:
         language: fail
         files: "BUILD$"
 
+      - id: generate_capi_exports
+        name: Regenerate CAPI export files
+        entry: compiler/src/iree/compiler/API/generate_exports.py
+        language: python
+        files: '^compiler/bindings/c/iree/compiler/.*\.h$'
+        pass_filenames: false
```
ScottTodd, line 130: Note that this will fail if the LLVM submodule is not initialized (I think you found that out based on the changes to lint.yml). There may be an edge case for developers who also skip initializing that submodule and then try to run `pre-commit run --all-files` without skipping this hook. For most cases I think the `files:` condition here should be sufficient though.

ScottTodd (commented):
This does looks appropriate for a pre-commit hook, thanks.

RattataKing:
> Does it print out how to fix the failure (e.g., running the script) when it fails?

It's included in the pre-commit test, CI will print ```If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.```

RattataKing:
> Have you tried adding a new CAPI to see what the output looks like? Could you show an example?


Declare a test sample in `/compiler/bindings/c/iree/compiler/dialects/iree_codegen.h`
```
MLIR_CAPI_EXPORTED bool TestPreCommitForCAPI();
```
Run `pre-commit run --all-files`, and get terminal output:
```shell
...
Regenerate CAPI export files.............................................Failed
- hook id: generate_capi_exports
- files were modified by this hook
```

.github/workflows/lint.yml
```diff
@@ -17,6 +17,10 @@ jobs:
     steps:
       - name: Checking out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Sparse checkout llvm-project mlir-c headers
+        run: # Sparse checkout mlir-c subdirectory
```
amd-eochoalo:
```suggestion
        run: |
          # Sparse checkout mlir-c subdirectory
```

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsrun

.github/workflows/lint.yml
```diff
@@ -17,6 +17,10 @@ jobs:
     steps:
       - name: Checking out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Sparse checkout llvm-project mlir-c headers
```
kuhar: Does the checkout action support sparse checkout?

kuhar: (it wouldn't surprise me if it does not)

ScottTodd: https://github.com/actions/checkout supports sparse checkout but I don't know if sparse checkout _of a submodule_ is possible through it

ScottTodd: https://github.com/iree-org/iree/actions/runs/23316948865/job/67818793591?pr=23848 oof this took 6 minutes? seems worse than without the sparse checkout??

RattataKing: Yeah I don't know why it takes this long on remote setup, I tested locally, sparse checkout works for submodule

kuhar: I think an extra minute is acceptable for a lint -- developers should run pre-commit locally anyway, so not a big deal.

hanhanW: (Honestly, I did not know `pre-commit` until I saw this PR. I always manually run few tools like clang-format, buildifier, etc. Good to know this tool, anyway.)

kuhar: you can add it to your git commit hook with `pip install pre-commit && pre-commit install`

RattataKing: Reverted to checkout entire submodule

kuhar (approved):

PR #23849, opened 2026-03-19
[nfc] Simplify op rewrite pattern inheriting constructor definitions.

hanhanW:
The revision uses the `Base` type alias and deletes custom constructor that is only about `benefit`. The caller should decide the `benefit` when they add patterns.

Note that `using Base::Base` doesn't work for template classes with dependent base classes (e.g., OpRewritePattern<T> where T is a template parameter) due to C++ dependent name lookup rules.


kuhar (approved):
Thanks. Would be nice to also make these final.

This could probably be turned into a clang tidy check -- they are not hard to write and we build it from source.

PR #23851, opened 2026-03-19
[Codegen] Move SPIRV pipelines to `#iree_gpu.spirv_pipeline<...>` attr

kuhar:
Follow-up to the LLVMGPU pipeline migration in #23816. Same goal: enable pipeline-specific constraint generation by getting backend pipelines out of the global `DispatchLoweringPassPipeline` enum and behind `PipelineAttrInterface`.

Same mechanics as #23816; I homed the new attribute in `iree_gpu` since we don't have a separate dialect for spirv. Added a CRTP helper `CodegenPipelineOptionsBase` to deduplicate the `TypeID`/`classof` boilerplate shared with `GPUCodegenPipelineOptions` and future pipeline option classes.


compiler/src/iree/compiler/Codegen/Dialect/GPU/ExternalInterfaces/SPIRVPipelineExternalModels.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace {
+
+struct SPIRVPipelineExternalModel final
+    : Codegen::PipelineAttrInterface::ExternalModel<SPIRVPipelineExternalModel,
+                                                    SPIRVPipelineAttr> {
+  LogicalResult buildPipeline(Attribute attr, OpPassManager &pm,
+                              const CodegenPipelineOptions *options) const {
+    SPIRVPipelineBuilder builder = getSPIRVPipelineBuilder();
+    assert(builder && "no SPIRV pipeline builder registered; ensure "
```
bangtianliu, line 35: Not sure if assert will be stripped out or not for release build

Maybe better to do 


```suggestion
   if (!builder) {
    return emitError(...) << "xxx";
  }
```

bangtianliu (approved):

compiler/src/iree/compiler/Codegen/Dialect/GPU/ExternalInterfaces/SPIRVPipelineExternalModels.cpp
kuhar: We use asserts to guard against programming errors -- the compiler shouldn't assert on any valid input IR, it may only error out. Here, not having a builder could only happen if the registration code somehow broke (i.e, a compiler bug), and this is captured in the assertion message.

PR #23853, opened 2026-03-19
[LLVMGPU] Legalize ND vectors to 1D vectors

Groverkss:
Just testing if there are any regressions

ci-extra: test_torch


Groverkss:
Some notes on missing test coverage:

- No pipeline tests for mxfp4 cdna4 matmuls. Only caught by e2e tests and not by LLVMGPU pipeline tests
- No CUDA ld_matrix/mma.sync LLVMGPU pipeline tests

When I say pipeline, I mean we have nothing that tests if we lower correctly to nvvm/rocdl.

PR #23854, opened 2026-03-19
[VectorExt] IndexedVectorOpInterface for transfer_gather/scatter

Groverkss:
These ops are very similar to each other, this patches moves their shared infrastructure to an interface.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
@@ -68,12 +68,15 @@ void TransferGatherOp::getEffects(
   }
 }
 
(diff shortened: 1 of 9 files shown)


compiler/src/iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.cpp
sommerlukas: Nit: I think we can use `getIndexVecIndexingMaps` here.

sommerlukas: Nit: The way we build the expected shape is the same for mask and index vectors. We can put that in a shared helper (lamdba or function).

sommerlukas (approved):
LGTM, just two nits.

PR #23855, opened 2026-03-19
[LinalgExt] Implement direct vectorization for im2col op

Max191:
Implements direct vectorization for im2col, and completes the support for padding on the im2col op. The padding attributes are used to compute the read mask when vectorizing. In the old path, we would have separate padding on the input and the result of the im2col, and we try to compose those pads into a single masked read. This is fragile and difficult for cases where the im2col result dims don't map well to the input dims. With this direct vectorization approach, we can compute the mask based on the input and result padding simultaneously. This will make flattening of the spatial dimensions of convolutions possible.

### Performance results: ###

 - Run 1: https://github.com/nod-ai/amd-shark-ai-reports/tree/main/boo/boo-custom-runs/2026-03-27_04-36_d1b822f45ac693a8593232a7d3fc5d67b1087f7e/comparison
 - Run 2: https://github.com/nod-ai/amd-shark-ai-reports/tree/main/boo/boo-custom-runs/2026-03-27_20-34_cf0d758bb199713b96a84a67245bc8b24ba7b74a/comparison

These runs were taken on different commits, but they are functionally the same (just some cleanup differences). I am only able to reproduce 3 of the regressions locally, and most of the improvers (~35 of them with 10-40% speedup) are real. There seems to have been some noise in the runs, but overall there is a good perf improvement.

ci-extra: test_torch


Max191:
This PR will depend on https://github.com/iree-org/iree/pull/23859 landing in order to avoid performance regressions, but the implementation in this PR will stay the same.

I will share benchmark results once I have them. My first run failed for some reason.

EDIT: That PR didn't merge, but the issue was fixed in #23947

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -283,6 +282,7 @@ static void addGPUVectorizationPasses(OpPassManager &funcPassManager,
   options.enableVectorMasking = enableMasking;
   options.vectorizeMapStore = true;
   funcPassManager.addPass(createGenericVectorizationPass(options));
+  funcPassManager.addPass(IREE::LinalgExt::createDecomposeIm2colPass());
```
yzhang93, line 294: It would be good to add some comments of why this pass is moved. Maybe better to rename this pass as `DecomposeAndVectorizeIm2colPass` (it doesn't have to be addressed in this PR).

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/canonicalize.mlir
```diff
@@ -249,3 +249,137 @@ func.func public @no_convert_dynamic_identity_map_load_into_copy(
...
+      tensor.yield %cst_pad : f32
+  } : tensor<2x34x34x640xf32> to tensor<2x36x36x640xf32>
+  %empty = tensor.empty() : tensor<2x1024x5760xf32>
+  %result = iree_linalg_ext.im2col
+      strides = [1, 1] dilations = [1, 1] kernel_size = [3, 3]
+      offsets = [0, 0, 0] output_sizes = [[2], [32, 32], [3, 3, 640]]
+      batch_pos = [0] m_pos = [1, 2] k_pos = [3]
+      input_k_perm = [0, 1, 2]
+      output_perm = [0, 1, 2]
+      input_pad_low = [0, 0, 0, 0] input_pad_high = [0, 0, 0, 0] pad_value(%cst_im2col : f32)
```
yzhang93, line 335: If the pad values are compatible, e.g., `input_pad_low = [1, 0, 0, 1] input_pad_high = [1, 0, 0, 1] pad_value(%cst_pad : f32)`. Will it fold another pad for input?

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_im2col.mlir
```diff
@@ -91,32 +83,26 @@ module {
 //       CHECK:     %[[mLOOP:.+]] = scf.for %[[m:.+]] = %[[C0]] to %[[mSIZE]] step %[[C1]] iter_args(%[[OUT1:.+]] = %[[OUT0]]) -> (tensor<2x?x?xf32>)
 //       CHECK:       %[[kLOOP:.+]] = scf.for %[[k:.+]] = %[[C0]] to %[[kSIZE]] step %[[C1]] iter_args(%[[OUT2:.+]] = %[[OUT1]]) -> (tensor<2x?x?xf32>)
 //   CHECK-DAG:         %[[kIDX:.+]] = affine.apply #[[$MAP]](%[[k]])[%[[kOFF]]]
-//   CHECK-DAG:         %[[kParts:.+]]:3 = affine.delinearize_index %[[kIDX]] into (640, 5, 2)
+//   CHECK-DAG:         %[[kParts:.+]]:3 = affine.delinearize_index %[[kIDX]] into (5, 2) : index, index, index
```
yzhang93: Why does it change from 3 dimensions to 2 now?

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1124,6 +1130,148 @@ static void registerInterfaceForLinalgOps(MLIRContext *ctx) {
...
+      b, loc, im2colOp, srcIndices, innerTileSize, outputIVs, vecOutputDim);
+  return vector::CreateMaskOp::create(b, loc, vecI1Type, validSize);
+}
+
+struct Im2colOpVectorizationModel
+    : public VectorizableOpInterface::ExternalModel<Im2colOpVectorizationModel,
+                                                    IREE::LinalgExt::Im2colOp> {
+
+  bool isVectorizable(Operation *op, ArrayRef<int64_t> vectorSizes,
+                      ArrayRef<bool> scalableDims,
```
yzhang93, line 1227: `vectorSizes` and `scalableDims` are not used in this function.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1124,6 +1130,148 @@ static void registerInterfaceForLinalgOps(MLIRContext *ctx) {
...
+    if (!outputType.hasStaticShape()) {
+      return false;
+    }
+
+    return true;
+  }
+
+  FailureOr<SmallVector<Value>> vectorize(Operation *op, RewriterBase &rewriter,
+                                          ArrayRef<int64_t> vectorSizes,
+                                          ArrayRef<bool> scalableDims,
```
yzhang93, line 1235: `vectorSizes` and `scalableDims` are not used in this function.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -2564,9 +2815,272 @@ LogicalResult Im2colOp::verify() {
...
+      padValue = im2colOp.getPadValue();
+    }
+
+    Location loc = padOp.getLoc();
+    SmallVector<OpFoldResult> lowPad = padOp.getMixedLowPad();
+    SmallVector<OpFoldResult> highPad = padOp.getMixedHighPad();
+
+    // The fold replaces pad(im2col(...)) with im2col(bigger_output), writing
+    // from offset 0. This is only valid when lowPad is all-zero; non-zero
+    // lowPad would leave the low-padded region uninitialized.
```
yzhang93: I couldn't get why non-zero lowPad would leave the low-padded region uninitialized.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -2564,9 +2815,272 @@ LogicalResult Im2colOp::verify() {
...
+
+    // If the im2col already has padding, the pad values must be compatible
+    // (same constant value) for the fold to be valid, since we can only have
+    // one pad_value on the im2col.
+    if (im2colOp.hasPadding()) {
+      auto existingConst =
+          im2colOp.getPadValue().getDefiningOp<arith::ConstantOp>();
+      auto newConst = padValue.getDefiningOp<arith::ConstantOp>();
+      if (!existingConst || !newConst ||
+          existingConst.getValue() != newConst.getValue()) {
```
yzhang93, line 2761: Comparing attribute values via `!=` seems to be fragile for float representations (e.g., 1.0 : f32 vs 1.0 : f16). Should it fold or not in such cases?

Max191: The element type must match for type consistency, so having im2col(pad) with different element types shouldn't be possible. Also, we need the padding value to be bitwise identical for the folding to be valid anyway.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
Max191: Ah, this is an old restriction from a previous implementation that I missed in the review. Thanks for the catch, I'll remove it!

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
Max191: Sure, I can add a comment. Although, I think the decompose im2col pass actually doesn't belong in addGPUVectorizationPasses anymore, since it is not actually doing vectorization. It is just a fallback in case we ended up with an im2col with dynamic result sizes, which shouldn't happen in practice. I'll move the pass and add a comment.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/canonicalize.mlir
Max191: Yes, it should. The test below was actually supposed to be testing that but I didn't check it carefully enough. Let me fix that.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/decompose_im2col.mlir
Max191: I believe the difference is whether or not the outer-most delinearized index is clamped to 640 or not (i.e., adding a `mod 640` expr at the end). In the new implementation, it is left unclamped for bounds checking in case there is padding. For the non-padded case, the index should remain in-bounds anyway, so there is no need to clamp it.

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_im2col.mlir
hanhanW, line 1: It is not clear how the vector sizes are derived for this naming. I categorize them a few, and I wonder if we move the tests into one of them?

Image

hanhanW (dismissed):
This is a very large PR. I wonder if we break it into at least two PRs or not:

- padding semangics
- vectorization

Max191:
> This is a very large PR. I wonder if we break it into at least two PRs or not:
> 
> * padding semangics
> * vectorization

This is already split up into as small of a chunk as seems reasonable to me. This is part of a larger refactor to the IGEMM path in order to enable additional flattening, and each piece needs to land without performance regressions. Going through the process of splitting out parts of the implementation, re-reviewing the split up changes, and testing/debugging on our full suite of convolutions is time consuming, and I've already spent a great deal of time on that to get to the split that I have so far. The issue is that the vectorization requires the padding semantics to be meaningful, and adding padding semantics without the vectorization and making that work with purely the decomposition without causing any performance regressions will take more time to do. One option would be to just add the fields to the im2col op but not use them, although I think that doesn't provide much benefit in terms of review (the padding PR would be very small, and it provides useful context when looking at it together with this PR anyway), and will more likely just slow down the process.

I agree that we should try to keep well-scoped PRs to help make review easier, but sometimes splitting up implementation can just slow down the process, which I believe is the case here. I feel that I've already spent too much time spinning on debugging/fixing random performance regressions that happen at intermediate stages of the split but not in the final state, and I don't want to keep wasting time on things like that.

If folks really need this PR to be split up further, then I can do it, but IMO it isn't really worth it.

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_im2col.mlir
Max191: I'm not sure I follow exactly what you mean. Is your point that the op-specific tests (like map_store transfer_gather, etc.) should be moved into masked_configured/masked_inferred/unmasked?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -1380,6 +1430,11 @@ def IREELinalgExt_Im2colOp : IREELinalgExt_Op<"im2col",
     `k_pos` `=` $k_pos
     `input_k_perm` `=` $input_k_perm
     `output_perm` `=` $output_perm
+    custom<OptionalPadding>($input_pad_low, $static_input_pad_low,
+                            $input_pad_high, $static_input_pad_high,
+                            $output_pad_low, $static_output_pad_low,
+                            $output_pad_high, $static_output_pad_high,
+                            $pad_value, type($pad_value))
```
hanhanW: Can we use `custom<DynamicIndexList>` for each pair? And the `pad_value` can be wrapped within `(^)` I think. E.g., https://github.com/llvm/llvm-project/blob/c703ea52be7976e68208585da4edd242c61d5bc5/mlir/include/mlir/Dialect/Tensor/IR/TensorOps.td#L1373-L1380

```
  // TODO: Remove custom<InferType> when AllTypesMatch supports opt. operands.
  let assemblyFormat = [{
    $source
    (`nofold` $nofold^)?
    `low` `` custom<DynamicIndexList>($low, $static_low)
    `high` `` custom<DynamicIndexList>($high, $static_high)
    $region attr-dict `:` type($source) `to` type($result)
  }];

```

hanhanW (commented):
> > This is a very large PR. I wonder if we break it into at least two PRs or not:
> > 
> > * padding semangics
> > * vectorization
> 
> This is already split up into as small of a chunk as seems reasonable to me. This is part of a larger refactor to the IGEMM path in order to enable additional flattening, and each piece needs to land without performance regressions. Going through the process of splitting out parts of the implementation, re-reviewing the split up changes, and testing/debugging on our full suite of convolutions is time consuming, and I've already spent a great deal of time on that to get to the split that I have so far. The issue is that the vectorization requires the padding semantics to be meaningful, and adding padding semantics without the vectorization and making that work with purely the decomposition without causing any performance regressions will take more time to do. One option would be to just add the fields to the im2col op but not use them, although I think that doesn't provide much benefit in terms of review (the padding PR would be very small, and it provides useful context when looking at it together with this PR anyway), and will more likely just slow down the process.
> 
> I agree that we should try to keep well-scoped PRs to help make review easier, but sometimes splitting up implementation can just slow down the process, which I believe is the case here. I feel that I've already spent too much time spinning on debugging/fixing random performance regressions that happen at intermediate stages of the split but not in the final state, and I don't want to keep wasting time on things like that.
> 
> If folks really need this PR to be split up further, then I can do it, but IMO it isn't really worth it.

My "skimming through read" is that there are changes in both tiling and vectorization. The vectorization is new and the tiling implementation already exists today. So my intuition is that you can add padding semantics with tiling changes. Then we can add vectorization support later on. I thought that you can always isolate the vectorization changes from this PR, and the rest can go with a single PR. Do I miss something?

> The issue is that the vectorization requires the padding semantics to be meaningful, and adding padding semantics without the vectorization and making that work with purely the decomposition without causing any performance regressions will take more time to do.

I don't follow the performance regression issue. You add padding semantics to the op, but it does not change the input and the folding does not exist? In this case, adding the semantics (with existing interface implementation) does not hurt performance, right?

Max191:
> My "skimming through read" is that there are changes in both tiling and vectorization. The vectorization is new and the tiling implementation already exists today. So my intuition is that you can add padding semantics with tiling changes. Then we can add vectorization support later on. I thought that you can always isolate the vectorization changes from this PR, and the rest can go with a single PR. Do I miss something?

Let me try to split it up. I think that should be fine, since it will effectively be a non-functional change. My concern is about adding the im2col(pad) canonicalizers, which would require me to actually support decomposition with padding in the intermediate state, which would probably just cause performance regressions. I can leave decomposition with padding unimplemented in the intermediate state, though.

On a more meta-side-note: I think my view on PR reviews has changed somewhat with AI assisted coding and review, so that's probably contributing to my feeling here. I think having the changes together makes it easier to navigate the full picture, and AI helps us filter through large changes more easily. It makes it easier to review large PRs, and we typically do end up implementing large, unsplit prototypes before splitting them up into reviewable pieces. I think my opinion is becoming more and more that the splitting up part of the process is less important because having all the context in one place/PR can actually make it easier to use agents to understand and review large changes. But I feel that we are in a transition phase now, and it takes time to adapt to new workflows, so I still understand the need to split things up when the reviewers request it. I will try splitting this one up further.

hanhanW:
> On a more meta-side-note: I think my view on PR reviews has changed somewhat with AI assisted coding and review, so that's probably contributing to my feeling here. I think having the changes together makes it easier to navigate the full picture, and AI helps us filter through large changes more easily. It makes it easier to review large PRs, and we typically do end up implementing large, unsplit prototypes before splitting them up into reviewable pieces. I think my opinion is becoming more and more that the splitting up part of the process is less important because having all the context in one place/PR can actually make it easier to use agents to understand and review large changes. But I feel that we are in a transition phase now, and it takes time to adapt to new workflows, so I still understand the need to split things up when the reviewers request it. I will try splitting this one up further.

I agree that it is good to have a single PR for full picture as long as they are splited into several commit. If we are going to do that, we may want to flip the "merge guidance" from "squash and merge" into "merge a chain of commit`. You still want "smaller piecies" (e.g., several commits) when you want people to review the code.

The other benefits of having several PRs is that you can easily revert a problematic one if it breaks something. You can argue that they'd be caught in a single big PR, but that's not 100% because some failures only happenned on post-submit or downstream projects. It is much easier to manage the states if they are several commits.

AI is a good tool, and it may change our mental model. I'm actually happy that we are growing with these tools.

---

I was going to offer that I can review a single PR or part of it, but I was in a meeting. I won't insist here, as it does not bother me much. But I'd like to share the pros of breaking PRs/commits above.

Max191:
Padding + tiling implementation PR is here: https://github.com/iree-org/iree/pull/23950

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
Max191: Fixed, thanks for the suggestion!

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_im2col.mlir
hanhanW: Each file tests with different configuration. E.g., `_unmasked` tests are mainly for cases that `input_vector_size` is empty. `masked_configured` is mainly for cases that have `input_vector_size` where it is derived from configurations, etc. This one belongs to `_unmasked` IMO.

Max191: I see, thanks for the context. I'll move it. I think map_store should also be moved there, which I can do in another PR.

Max191: I actually opted to implement the vector size analysis for im2col, and put the tests in the `_inferred` file. I want to later use divisibility analysis to determine vector sizes for im2col, and we have access to a dataflow solver in the analysis, so this is a better place for it.

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_inferred.mlir
```diff
@@ -873,3 +873,243 @@ func.func @scaled_tensor_multi_mma(%arg0: tensor<3x5x1x32xf4E2M1FN>, %arg1: tens
...
+}
+// CHECK-LABEL: func.func @im2col_no_vectorize_dynamic
+//       CHECK:   iree_linalg_ext.im2col
+//   CHECK-NOT:   vector.transfer_read
+//   CHECK-NOT:   vector.transfer_write
+
+// -----
+
+// Source padding (conv padding folded into im2col). NHWC layout.
+// Vectorizes along K with masked transfer_read.
```
yzhang93, line 980: Nit: since it's masked transfer_read, it would be good to add checks for the mask.

yzhang93 (approved):
Thanks for the changes. LGTM.

Max191:
VAE failures are pre-existing ever since the test was added to torch tests: https://github.com/iree-org/iree/actions/runs/23915109640/job/69748051690

PR #23836, opened 2026-03-18
Reapply "[GPUHeuristics] Further tune `LargeGemm` perf (#23652)"

lialan:
Here is previous commit message.
---
* Adds additional, optional utilization rate as a guard in the heuristic seeds.
* Tunes `LargeGemm` MMA schedule heuristic seeds from sg=4/MNT=16 to sg=8/MNT=32 and upgrades the workgroup count adjustment guard to be utilization-aware.
* Inflate LargeGemm heuristic seeds for gfx950 (sg=8, MNT=32) to better utilize CDNA4 compute resources. This is based on tuner results.
* Benchmarked on 75 LargeGemm bf16 shapes from mi355x: +7.0% geomean improvement with 28 shapes improved (up to 30%) and no regressions >5%.

Here are some `LargeGemm`s representative results: | M×N×K | Change |
  |---|---|
  | 150000×4096×16384 | **+11%** |
  | 150000×16384×4096 | **+10%** |
  | 150000×2048×8192 | **+12%** |
  | 150000×4096×2268 | **+10%** |
  | 21760×3840×3840 | **+30%** |
  | 16640×3840×3840 | **+26%** |
  | 11520×3840×3840 | **+17%** |
  | 6400×3840×3840 | **+13%** |
  | 24576×2048×1536 | **+10%** |
  | 4096×8192×2048 | **+6%** |
  | 3840×3840×4352 | neutral |
  | 4096×16384×150000 (large-K) | neutral |
  | 2268×4096×150000 | neutral |

---------

PR #23856, opened 2026-03-19
Fix conflict tests when reapply commit ea39ece52

yzhang93:
Fixes https://github.com/iree-org/iree/pull/23836 that has conflicts with main branch before it merged.


lialan (approved):
+1

PR #23859, opened 2026-03-19
[LLVMGPU] Handle decomposed masks in ROCDLBufferInstructionsOptimization

Max191:
The tile-and-fuse pipeline now decomposes masks (`decomposeMasks=true`), producing step+broadcast+cmpi+andi IR instead of create_mask ops. This commit adapts the ROCDL buffer optimization passes to work with the new IR shape.

**ROCDLBufferInstructionsOptimization** is simplified to pattern-match `vector.broadcast(%scalar_i1)` as the mask on `vector.transfer_read` and `vector.maskedload`, replacing with an unmasked operation + `arith.select` (or just the unmasked operation if the mask is constant true). Moved from post-bufferize to after vector lowering in the pipeline.

**OptimizeComparisonOps** is a new pass that simplifies vector `arith.cmpi` where one operand is a broadcast of a scalar with known divisibility. Uses IREE's `IntegerDivisibilityAnalysis` to determine if the comparison result is uniform across all vector lanes, and rewrites to a scalar comparison + broadcast.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
@@ -6,12 +6,7 @@
 
 #include "iree/compiler/Codegen/LLVMGPU/Passes.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
-#include "iree/compiler/Codegen/Utils/Utils.h"
-#include "iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.h"
-#include "llvm/Support/DebugLog.h"
-#include "mlir/Analysis/DataFlow/ConstantPropagationAnalysis.h"
-#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
-#include "mlir/Analysis/DataFlowFramework.h"
+#include "mlir/IR/Matchers.h"
 
 #define DEBUG_TYPE "iree-codegen-rocdl-buffer-instructions-optimization"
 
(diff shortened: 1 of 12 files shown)


Max191:
I still need to review properly this myself, so probably don't review yet, but making this draft to provide context in another PR.

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLBufferInstructionsOptimization.cpp
```diff
@@ -22,207 +23,293 @@ namespace mlir::iree_compiler {
...
+/// Try to simplify a vector arith.cmpi where one operand is a broadcast of a
+/// scalar. If the broadcast scalar's divisibility guarantees that the
+/// comparison result is the same for all vector lanes, rewrite to a scalar
+/// comparison broadcast.
+static bool simplifyUniformCmpI(IRRewriter &rewriter, arith::CmpIOp cmpOp,
+                                DataFlowSolver &solver) {
+  // Only handle slt predicate — the decomposed mask pattern is always
+  // "step < broadcast(bound)". Other predicates have different always-true/
+  // always-false conditions and rewrite semantics.
+  if (cmpOp.getPredicate() != arith::CmpIPredicate::slt) {
```
krzysz00: Get ult in here too

Max191: It now handles all the signed predicate types, but the unsigned comparisons aren't really valid due to the divisibility analysis being inherently signed. Unsigned integers break any divisibility that isn't a Po2, because underflow/overflow acts as a modulo.

Max191:
Converted to draft because this is causing some performance regressions that are difficult to deal with. I will see about coming back to this in the near future.

PR #23862, opened 2026-03-19
[CI] Cache Bazel BUILD in Azure

amd-eochoalo:
* Reduces bazel CI action from[ 47 minutes ](https://github.com/iree-org/iree/actions/runs/23314145479/job/67809032499)down to [8 minutes.](https://github.com/iree-org/iree/actions/runs/23316104302/job/67815947305)
* Updates to Bazel version 7.4.0 to take advantage of experimental_disk_cache_gc_max_size
* The current cache size is 1.3 GB (after compression)
* The current cache uncompressed is 6.6 GB 
* Sets maximum cache size to 10GB (before compression)

Limitations:
* It is unclear why the run for this PR took [38 minutes](https://github.com/iree-org/iree/actions/runs/23318543943/job/67824645719?pr=23862) and[ 27 minutes](https://github.com/iree-org/iree/actions/runs/23323331160/job/67839155691), when previously we measured 8 minutes? I'm wondering if noisy tenants is more of an issue than the cache itself. Ah! It might be because we are no longer on top of main, which means LLVM possibly changed and we are testing these changes on the merged branch (i.e., no cache hits)

Assisted-by: Claude


build_tools/bazel/build_test_all.sh
```diff
@@ -170,6 +171,11 @@ declare -a BAZEL_TEST_CMD=(
 
 if [[ ! -z "${SANDBOX_BASE}" ]]; then
   BAZEL_TEST_CMD+=(--sandbox_base="${SANDBOX_BASE}")
+  BAZEL_TEST_CMD+=(--experimental_disk_cache_gc_max_size=10G)
+fi
+
+if [[ -n "${BAZEL_DISK_CACHE_DIR}" ]]; then
+  BAZEL_TEST_CMD+=(--disk_cache="${BAZEL_DISK_CACHE_DIR}")
```
amd-eochoalo:
```suggestion
fi

if [[ -n "${BAZEL_DISK_CACHE_DIR}" ]]; then
  BAZEL_TEST_CMD+=(--disk_cache="${BAZEL_DISK_CACHE_DIR}")
```

build_tools/bazel/build_test_all.sh
```diff
@@ -172,6 +173,10 @@ if [[ ! -z "${SANDBOX_BASE}" ]]; then
   BAZEL_TEST_CMD+=(--sandbox_base="${SANDBOX_BASE}")
 fi
 
+if [[ -n "${BAZEL_DISK_CACHE_DIR}" ]]; then
+  BAZEL_TEST_CMD+=(--disk_cache="${BAZEL_DISK_CACHE_DIR}")
```
amd-eochoalo, line 177:
```suggestion
  BAZEL_TEST_CMD+=(--disk_cache="${BAZEL_DISK_CACHE_DIR}")
  BAZEL_TEST_CMD+=(--experimental_disk_cache_gc_max_size=10G)
```

Ok, I don't think the disk cache gc max size flag decreased performance. It is likely that there was another tenant competing for resources. Best case, we are down to 8 minutes. Worse case, no worse than we are now.

kuhar:
I think the biggest productivity win would be to make the azure care public to download on PRs opened from forks-- otherwise we are only speeding up post-submit jobs, which are generally not a bottleneck

amd-eochoalo:
> I think the biggest productivity win would be to make the azure care public to download on PRs opened from forks-- otherwise we are only speeding up post-submit jobs

I understand, I wanted to highlight that even sccache builds are not taking advantage of this. For example: https://github.com/iree-org/iree/actions/runs/23327547876/job/67851940559?pr=23872

```
  env:
    BUILD_DIR: build
    SCCACHE_AZURE_CONNECTION_STRING: 
    SCCACHE_AZURE_BLOB_CONTAINER: ccache-container
    SCCACHE_CACHE_ZSTD_LEVEL: 10
    SCCACHE_AZURE_KEY_PREFIX: ci_linux_x64_clang
Connection string _not_ set, skipping sccache setup
```

kuhar:
> I understand, I wanted to highlight that even sccache builds are not taking advantage of this. For example: https://github.com/iree-org/iree/actions/runs/23327547876/job/67851940559?pr=23872

This is a very good find, we should fix that too

amd-eochoalo:
> This is a very good find, we should fix that too

[I found out that that isn't possible at the moment, Scott Todd already opened an issue about this:](https://github.com/mozilla/sccache/issues/2258)

>  [To support public access to Azure without credentials, we need efforts from both the opendal and sccache sides](https://github.com/mozilla/sccache/issues/2258#issuecomment-23448451598)

The alternative is to essentially do the same that we did here for bazel, but instead for sccache. I.e., download cache from azure, use it, upload on merge. But at that point, maybe the GitHub cache would be better (if we don't have to worry about the cache limits, if we do, then maybe we should keep the current approach and wait for that issue to close).

Regarding Bazel, let me know if you have any more comments. I noticed a slowdown but I think this is because the LLVM integrate was merged yesterday which means less cache hits.

kuhar:
> > This is a very good find, we should fix that too
> 
> [I found out that that isn't possible at the moment, Scott Todd already opened an issue about this:](https://github.com/mozilla/sccache/issues/2258)
> 
> > [To support public access to Azure without credentials, we need efforts from both the opendal and sccache sides](https://github.com/mozilla/sccache/issues/2258#issuecomment-23448451598)
> 
> The alternative is to essentially do the same that we did here for bazel, but instead for sccache. I.e., download cache from azure, use it, upload on merge. But at that point, maybe the GitHub cache would be better (if we don't have to worry about the cache limits, if we do, then maybe we should keep the current approach and wait for that issue to close).
> 
> Regarding Bazel, let me know if you have any more comments. I noticed a slowdown but I think this is because the LLVM integrate was merged yesterday which means less cache hits.

upload can happen in post-submit though, no? 

If we can switch to github cache that's also fine, but not sure if we'd have to buy more storage: https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#increasing-cache-size

amd-eochoalo:
> upload can happen in post-submit though, no?

Yes. At the moment, it is only uploaded on post-submit on the main branch. (That's what I meant by it happens on merge)

amd-eochoalo:
> but not sure if we'd have to buy more storage:

I think it will be necessary to avoid disrupting the other caches.

amd-eochoalo:
@kuhar is there anything else missing from this PR? Just in case you missed it, PRs from forks can access the cache after this commit: https://github.com/iree-org/iree/pull/23862/commits/e2d159395d5e9e238b419e4ad7d79c154f623cc6

kuhar (approved):

PR #23863, opened 2026-03-19
[Codegen] Set upstream GPU attributes to set bounds on subgroup ID

krzysz00:
1. This removes the call to the subgroup ID expansion patterns in ConvertToROCDL, matching all the other targets (which don't use this pass), which will enable subgroup ID to correctly lower to rocdl.wave.id on gfx12!
2. To make that pattern work efficiently, set the gpu.known_block_size attribute on functions when propagating dispatch size bounds so that those OCKL calls become constants pre-gfx12 and so that, when thread_id gets emitted, we get the correct ranges in place.

(This will hopefully make Wave's job a bit easier)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
@@ -232,6 +232,17 @@ struct PropagateDispatchSizeBoundsPass final
                        staticSubgroupSize);
     applyBounds(funcOp, workgroupSizes, workgroupCounts, maxSubgroupSize,
                 subgroupIdBound);
+
+    if (auto *gpuDialect = getContext().getLoadedDialect<gpu::GPUDialect>()) {
+      if (staticWorkgroupSize) {
+        SmallVector<int32_t, 3> blockSize(3, 1);
+        for (auto [idx, s] : llvm::enumerate(*staticWorkgroupSize)) {
+          blockSize[idx] = static_cast<int32_t>(s);
+        }
+        gpuDialect->getKnownBlockSizeAttrHelper().setAttr(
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
```diff
@@ -232,6 +232,17 @@ struct PropagateDispatchSizeBoundsPass final
                        staticSubgroupSize);
     applyBounds(funcOp, workgroupSizes, workgroupCounts, maxSubgroupSize,
                 subgroupIdBound);
+
+    if (auto *gpuDialect = getContext().getLoadedDialect<gpu::GPUDialect>()) {
+      if (staticWorkgroupSize) {
+        SmallVector<int32_t, 3> blockSize(3, 1);
```
kuhar: nit:
```suggestion
        std::array<int32_t, 3> blockSize = {1, 1, 1};
```

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchSizeBounds.cpp
```diff
@@ -232,6 +232,17 @@ struct PropagateDispatchSizeBoundsPass final
...
+        for (auto [idx, s] : llvm::enumerate(*staticWorkgroupSize)) {
+          blockSize[idx] = static_cast<int32_t>(s);
```
kuhar: or maybe you can `map_to_vector<3>(*staticWorkgroupSize, llvm::StaticCastTo<int32_t>)` this?

krzysz00: I think in theory the list of workgroup sizes can be less than 3 long? I don't know if I've seen that, but I know this code tries hard to avoid relying on that fact

kuhar (approved):

PR #23864, opened 2026-03-19
[Codegen] Move LLVMCPU pipelines to `#iree_cpu.pipeline<...>` attr

kuhar:
Same motivation as the GPU/SPIRV pipeline migrations (#23816, #23851): per-backend pipeline attrs decouple the global
`DispatchLoweringPassPipeline` enum from individual backends, letting each backend own its pipeline identifiers and evolve independently. This also unblocks constraint generation for CPU pipelines via `PipelineAttrInterface`.


compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.h
```diff
@@ -87,6 +88,19 @@ struct LLVMCPUPipelineOptions {
...
+/// PipelineAttrInterface::buildPipeline.
+struct CPUCodegenPipelineOptions final
+    : CodegenPipelineOptionsBase<CPUCodegenPipelineOptions> {
+  CPUCodegenPipelineOptions(
+      const LLVMCPUPipelineOptions &options,
+      IREE::Codegen::LoweringConfigAttrInterface loweringConfig)
+      : options(options), loweringConfig(loweringConfig) {}
+
+  LLVMCPUPipelineOptions options;
+  IREE::Codegen::LoweringConfigAttrInterface loweringConfig;
```
kuhar, line 101: FYI, on the GPU side this isn't needed because the GPU pipelines read the config from the IR directly. Here the CPU `DoubleTilingExpert` pipeline takes it as an explicit argument instead...

hanhanW: I think it means that GPU pipeline is "fixed" while CPU pipeline allows multi-level tiling expert which inserts the passes based on lowering_config. I.e., it accepts "distribution-only", "two-level of tiling", "three-level of tiling including cache-level", etc. This is why the method is named with `addMultiTilingExpertPassPipeline`.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -58,6 +58,19 @@
 
 namespace mlir::iree_compiler {
 
+using CPUPipeline = IREE::CPU::LoweringPipeline;
+
+/// Helper to build a TranslationInfoAttr with a CPU pipeline.
+static IREE::Codegen::TranslationInfoAttr
+getCPUTranslationInfo(MLIRContext *ctx, CPUPipeline pipeline,
+                      ArrayRef<int64_t> workgroupSize = {},
+                      std::optional<int64_t> subgroupSize = std::nullopt,
```
hanhanW: These two feilds can be dropped, because CPU don't set them at all.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td
```diff
@@ -17,20 +17,8 @@ include "mlir/IR/EnumAttr.td"
...
-def CPU_Mmt4dTilingExpert
-    : I32EnumAttrCase<"Mmt4dTilingExpert", 3>;
-def CPU_BufferOpsTileAndVectorize
-    : I32EnumAttrCase<"CPUBufferOpsTileAndVectorize", 4>;
-def CPU_DataTiling
-    : I32EnumAttrCase<"CPUDataTiling", 5>;
-def CPU_LinalgExtTileAndVectorize
-    : I32EnumAttrCase<"CPULinalgExtTileAndVectorize", 6>;
+// Note: CPU pipelines have been moved to #iree_cpu.pipeline<...> attr
+// (IREE::CPU::PipelineAttr in IREECPUAttrs.td).
```
hanhanW: The comment can be deleted, IMO. People will see them in the IR and just search in the codebase. People wouldn't come here to read this comment.

compiler/src/iree/compiler/Codegen/Dialect/CPU/ExternalInterfaces/CPUPipelineExternalModels.cpp
```diff
@@ -0,0 +1,49 @@
...
+  static CPUPipelineBuilder builder = nullptr;
+  return builder;
+}
+
+void registerCPUPipelineBuilder(CPUPipelineBuilder builder) {
+  CPUPipelineBuilder &current = getCPUPipelineBuilder();
+  assert((!current || current == builder) &&
+         "CPU pipeline builder already registered with a different callback");
+  current = builder;
+}
```
hanhanW: I wonder why `CPUPipelineBuilder` is a static object, but not a member of dialect; the method can also belong to the dialect?

Here is an exmple in Codegen dialect, where `mutex` is a variable and it has several methods.

https://github.com/iree-org/iree/blob/e8853aa2794e427a3a0f41b045ccc39873ee62ab/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenDialect.td#L88-L105

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -4090,8 +4122,8 @@ lowerUsingDefaultPipeline(mlir::FunctionOpInterface entryPointFn) {
     return success();
   }
   // Otherwise lower using default pipeline.
-  auto translationInfo = IREE::Codegen::TranslationInfoAttr::get(
-      entryPointFn->getContext(), DispatchLoweringPassPipeline::CPUDefault);
+  auto translationInfo = getCPUTranslationInfo(entryPointFn->getContext(),
+                                                CPUPipeline::Default);
```
hanhanW: nit: spell out the type. It is no longer `xxx::create`.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -3977,33 +4006,36 @@ adjustTileSizesForRootUnPackOp(mlir::FunctionOpInterface entryPointFn,
...
+    DictionaryAttr pipelineConfig = tInfo.getConfiguration();
     auto enableLoopPeelingAttrName =
         getEnableLoopPeelingAttrName(rootOp->getContext());
     auto newPipelineConfigEntries = llvm::filter_to_vector(
         pipelineConfig.getValue(), [&](NamedAttribute entry) {
           return entry.getName() != enableLoopPeelingAttrName;
         });
 
-    pipelineConfig =
+    DictionaryAttr newPipelineConfig =
```
hanhanW: use auto

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -3977,33 +4006,36 @@ adjustTileSizesForRootUnPackOp(mlir::FunctionOpInterface entryPointFn,
   }
 
   IREE::Codegen::TranslationInfoAttr tInfo = getTranslationInfo(entryPointFn);
-  DispatchLoweringPassPipeline pipeline =
-      tInfo.getDispatchLoweringPassPipeline();
-  DictionaryAttr pipelineConfig = tInfo.getConfiguration();
+  IREE::Codegen::TranslationInfoAttr translationInfo = tInfo;
```
hanhanW: Why do we need this variable?

hanhanW (changes requested):
I don't follow why external model mechanism is used in this case (and sorry that I did not ask such question in GPU PR).

There are mainly two reasons when I use external model mechanism:
- Avoid circular dependency.
- Do not want to relocate domain-specific logic to the dialect implementation.

This interface naturely sit into the op implementation, IMO. We create the op for pipeline, and it implements the interface.

kuhar:
> I don't follow why external model mechanism is used in this case (and sorry that I did not ask such question in GPU PR).
> 
> There are mainly two reasons when I use external model mechanism:
> 
> * Avoid circular dependency.
> * Do not want to relocate domain-specific logic to the dialect implementation.
> 
> This interface naturely sit into the op implementation, IMO. We create the op for pipeline, and it implements the interface.

It's to avoid circular dependencies. I think it goes like this: LLVMCPU -> iree_cpu dialect -> iree_codegen dialect.
Because buildPipeline needs to call into addMultiTilingExpertPassPipeline, addCPUDefaultPassPipeline, etc., which live in the LLVMCPU backend, we can't have the model on the iree_cpu side directly.

compiler/src/iree/compiler/Codegen/Dialect/CPU/ExternalInterfaces/CPUPipelineExternalModels.cpp
kuhar: It gets initialized only once during dialect registration and doesn't depend on MLIRContext -- IMO no reason to have a per-context copy.

hanhanW:
> > I don't follow why external model mechanism is used in this case (and sorry that I did not ask such question in GPU PR).
> > There are mainly two reasons when I use external model mechanism:
> > 
> > * Avoid circular dependency.
> > * Do not want to relocate domain-specific logic to the dialect implementation.
> > 
> > This interface naturely sit into the op implementation, IMO. We create the op for pipeline, and it implements the interface.
> 
> It's to avoid circular dependencies. I think it goes like this: LLVMCPU -> iree_cpu dialect -> iree_codegen dialect. Because buildPipeline needs to call into addMultiTilingExpertPassPipeline, addCPUDefaultPassPipeline, etc., which live in the LLVMCPU backend, we can't have the model on the iree_cpu side directly.

I may be missing something, but `buildPipeline` is defined in `compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp`?

kuhar:
codegen/llvmcppu depends on the iree_cpu dialect, so if we move it to iree_cpu we wouldn't be able to take the reverse dependency?

hanhanW:
> codegen/llvmcppu depends on the iree_cpu dialect, so if we move it to iree_cpu we wouldn't be able to take the reverse dependency?

I was thinking that you expose `registerCPUPipelineBuilder` method in `CPUDialect.h` (or `CPUTypes.h`), and you still provide callback in `Codegen/LLVMCPU/Passes.cpp`.

(I'm still thinking about moving them into context, though. I need to think more about it.)

kuhar:
> I was thinking that you expose `registerCPUPipelineBuilder` method in `CPUDialect.h` (or `CPUTypes.h`), and you still provide callback in `Codegen/LLVMCPU/Passes.cpp`.

hmm, let me try that

compiler/src/iree/compiler/Codegen/Dialect/CPU/ExternalInterfaces/CPUPipelineExternalModels.cpp
kuhar: Registration is global, so it can't possibly run concurrently. I added a comment to clarify this. I think it's good as-is.

kuhar:
> > I was thinking that you expose `registerCPUPipelineBuilder` method in `CPUDialect.h` (or `CPUTypes.h`), and you still provide callback in `Codegen/LLVMCPU/Passes.cpp`.
> 
> hmm, let me try that

@hanhanW thanks for the suggestion, you were right, we can avoid the circular dep without an external model

Groverkss (approved):

compiler/bindings/python/test/ir/dialects_test.py
hanhanW, line 1: I've seen this from switching GPU pipeline to CPU pipeline, then now is VMVX pipeline. What happens to this test when we modernize VMVX pipelines?

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.td
```diff
@@ -102,6 +102,25 @@ def IREECPU_DataTiledMMAAttr
...
+//===----------------------------------------------------------------------===//
+
+def IREECPU_PipelineAttr : AttrDef<IREECPU_Dialect, "Pipeline", [
+    DeclareAttrInterfaceMethods<IREECodegen_PipelineAttrInterface>]> {
+  let mnemonic = "pipeline";
+  let summary = "LLVMCPU lowering pipeline identifier.";
+  let description = [{
+    Identifies an LLVMCPU lowering pipeline. The builder callback is
+    registered by registerCodegenLLVMCPUPasses() via
+    registerCPUPipelineBuilder().
```
hanhanW: The comment viloates layering to me. `registerCodegenLLVMCPUPasses` is IREE's implementation details. We should document `registerCPUPipelineBuilder` must be called by users, and we should have better comments. E.g., the user should configure the lowering pipeline for each enum member in `CPU Lowering Pipelines` section. I.e., the builder must handle all LoweringPipeline enum values.

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -22,6 +22,30 @@
...
+static CPUPipelineBuilder &getCPUPipelineBuilderStorage() {
+  static CPUPipelineBuilder builder = nullptr;
+  return builder;
+}
+
+void registerCPUPipelineBuilder(CPUPipelineBuilder builder) {
+  // Called exactly once during global pass registration
+  // (ireeCompilerGlobalInitialize -> registerCodegenLLVMCPUPasses).
+  getCPUPipelineBuilderStorage() = builder;
+}
```
hanhanW, line 41: Continue the discussion from https://github.com/iree-org/iree/pull/23864#discussion_r2963268066. The comment is not moved after the change.

From @kuhar :

> It gets initialized only once during dialect registration and doesn't depend on MLIRContext -- IMO no reason to have a per-context copy.

> Registration is global, so it can't possibly run concurrently. I added a comment to clarify this. I think it's good as-is.

I agree that it can be registered once and we dont need a copy for each context. TLDR: I think using `std::call_once` and static members for `std::once_flag ` and `CPUPipelineBuilder`.

Few benefits:

- We can have better documentation in `Dialect.td` about registeration. I think it is the entry point when you look at a dialect.
- Visible ownership: it belongs to IREE CPU Dialect. It is declared on the dialect, not hidden in a .cpp file.
- It is still global, so we dont have a per-context copy.
- `std::call_once` has the semantics: it can only be registered once. Although my Claude told me that there are no values for using `std::call_once` in practice because registeration should happen way before loading the context. But it gives us the semantics.

What do you think?

hanhanW (commented):
The major comment is about the location of the `pipelineBuilder`, see inline comments. Thanks!

compiler/bindings/python/test/ir/dialects_test.py
kuhar: when the attribute is empty it will go away

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.td
kuhar: good point, should be addressed now

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
kuhar: I added a plain assert to make sure it's not registered twice, but none of the registration code is very defensive AFAIK. I think call once is not worth it.

compiler/bindings/python/test/ir/dialects_test.py
hanhanW: sg, thanks!

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUTypes.h
```diff
@@ -40,8 +40,23 @@ struct LoweringConfigLevelInfo {
...
+} // namespace mlir::iree_compiler
+
 namespace mlir::iree_compiler::IREE::CPU {
 
+/// Callback type for CPU pipeline builders. The callback receives a
+/// PipelineAttr and must handle all LoweringPipeline enum values.
+/// Returns success if the pipeline was built.
+using CPUPipelineBuilder =
+    LogicalResult (*)(Attribute pipelineAttr, OpPassManager &pm,
+                      const CodegenPipelineOptions *options);
```
hanhanW, line 54: Do we use `std::function`?

hanhanW (approved):
LGTM, thanks for addressing the comments and having good conversation! I only have a final question which is optional.

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUTypes.h
kuhar: we could but this is not an extension point -- it's just to break a circular dependency, so we know that a function pointer is all we ever need

hanhanW: I see, thanks

PR #23865, opened 2026-03-19
[LLVMGPU] Move address computation optimization after vector lowering

Groverkss:
We should be lowering vectors before optimizing for addresses. We had a hack where we lowered transfer_reads before to make up for it. After the address optimizations, we have i32 vectors which are harder to reason than index vectors for transfer_gather/transfer_scatter indices vectors.

ci-extra: test_torch


krzysz00 (approved):
I suspect that there wasn't a reason this order happened, probably some amount of inertia, and the possibility that there was address computation logic (maybe swizzle related?) that was matching on transfer_read.

If we're feeling paranoid, I'd go check the passes that make up address computation lowering and check if they're doing logic on `transfer_*`s.

I'd also go check the other pipelines to see if we have the same mistake elsewhere.

PR #23867, opened 2026-03-19
[HAL] Support sub-byte element types in buffer view size calculation

IanWood1:
Compute packed size `ceil(elements * bits / 8)` instead of rejecting sub-byte types. Add i4 e2e tests.

### DIFF
--- a/runtime/src/iree/hal/buffer_view.c
+++ b/runtime/src/iree/hal/buffer_view.c
@@ -52,13 +52,14 @@ IREE_API_EXPORT iree_status_t iree_hal_buffer_view_create(
     iree_hal_buffer_retain(buffer_view->buffer);
     buffer_view->element_type = element_type;
     buffer_view->encoding_type = encoding_type;
-    buffer_view->byte_length =
-        iree_hal_element_dense_byte_count(buffer_view->element_type);
     buffer_view->shape_rank = shape_rank;
+    iree_device_size_t element_count = 1;
     for (iree_host_size_t i = 0; i < shape_rank; ++i) {
       buffer_view->shape[i] = shape[i];
-      buffer_view->byte_length *= shape[i];
+      element_count *= shape[i];
     }
+    buffer_view->byte_length = iree_hal_element_packed_byte_count(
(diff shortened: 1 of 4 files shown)


MaheshRavishankar (approved):
I think this makes sense.

@bjacob  or @benvanik  for any historical context on this.

runtime/src/iree/hal/buffer_view.c
```diff
@@ -52,13 +52,15 @@ IREE_API_EXPORT iree_status_t iree_hal_buffer_view_create(
...
     buffer_view->shape_rank = shape_rank;
+    iree_host_size_t bit_count =
+        iree_hal_element_bit_count(buffer_view->element_type);
+    iree_device_size_t element_count = 1;
     for (iree_host_size_t i = 0; i < shape_rank; ++i) {
       buffer_view->shape[i] = shape[i];
-      buffer_view->byte_length *= shape[i];
+      element_count *= shape[i];
     }
+    buffer_view->byte_length = (element_count * bit_count + 7) / 8;
```
bjacob: This not-quite-trivial logic is repeated below in buffer_view_util.c.  Could you introduce a helper so you only need to write it once?

runtime/src/iree/hal/buffer_view_util.c
```diff
@@ -31,16 +31,16 @@ IREE_API_EXPORT iree_status_t iree_hal_buffer_compute_view_size(
...
+                                "opaque element types cannot be indexed");
       }
-      byte_length = iree_hal_element_dense_byte_count(element_type);
+      iree_host_size_t bit_count = iree_hal_element_bit_count(element_type);
+      iree_device_size_t element_count = 1;
       for (iree_host_size_t i = 0; i < shape_rank; ++i) {
-        byte_length *= shape[i];
+        element_count *= shape[i];
       }
+      byte_length = (element_count * bit_count + 7) / 8;
```
bjacob: The thing to share in a helper.

bjacob (commented):
Pretty sure that historical context is simply "sub-byte types didn't exist when this code was designed".

PR #23870, opened 2026-03-20
[SPIRV] Fix crash in getMemoryVectorSize on non-int/float types

dbrll:
`getMemoryVectorSize()` crashes on SPIRV on `index` element types because `getIntOrFloatBitWidth()` only handles int/float.

This two-lines fix adds a type guard to prevent this, as the llvm-cpu backend does.
Test included.

Assisted-by: claude
ci-extra: macos_arm64_clang

### DIFF
--- a/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
@@ -113,6 +113,11 @@ int getComputeVectorSize(int64_t size) {
 }
 
 int getMemoryVectorSize(Value source, Type scalarType, int64_t size) {
+  // Only optimize memory access for int/float types with known bitwidths.
+  // Types like index or i1 cannot be bitcast to 32-bit element vectors.
+  if (!scalarType.isIntOrFloat() || scalarType.isInteger(1)) {
+    return 1;
+  }
   int bitwidth = scalarType.getIntOrFloatBitWidth();
   while (auto sliceOp = source.getDefiningOp<tensor::ExtractSliceOp>()) {
     source = sliceOp.getSource();
--- a/compiler/src/iree/compiler/Codegen/SPIRV/test/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/test/BUILD.bazel
@@ -48,6 +48,7 @@ iree_lit_test_suite(
             "illegal_configuration.mlir",
             "initial_vector_lowering_0d.mlir",
             "initial_vector_lowering_extract_insert_unit_dim.mlir",
+            "initial_vector_lowering_index_type.mlir",
             "link_executables.mlir",
             "lower_masks.mlir",
             "lowering_matmul_fusion.mlir",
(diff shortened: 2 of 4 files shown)


kuhar (changes requested):
This PR doesn't appear to comply with our policy on tool-generated content,
and requires additional justification for why it is valuable enough to the
project for us to review it. Please see our developer policy on
AI-generated contributions: https://iree.dev/developers/general/contributing/#ai-tool-use

dbrll:
Sure, sorry for the misunderstanding about the AI policy.

I started investigating this issue after having segfaults on SPIRV on valid StableHLO generated by a compiler I'm working on. However this is not specific to it and JAX / torch-mlir could trigger this on the SPIRV backend also.

Right now, `getMemoryVectorSize()` has no guard for `index` element types, which can appear in any dispatch containing vectorized `linalg.index`. The triggering pattern in my case was `stablehlo.iota` fused with a reduction.

The fix is just 2 lines to align the SPIRV backend behavior to the local-task backend behavior.

This fix also reveals an `vector.step` legalization error which was previously unreachable due to the earlier segfault. It seems to happen because SPIRV has a maximum vector width, whereas LLVM IR supports arbitrary vector widths. I see that `SPIRVInitialVectorLowering` is supposed to break large vectors down to native sizes before conversion, but it doesn't cover `vector.step`. This will probably be for another issue/PR.

kuhar:
I flagged this because the PR came from a new contributor, with an LLM-generated PR description that seemed overly long and no tests -- all these being the usual hallmarks of a contribution that would violate the 'human in the loop' rule. We should either rewrite the PR description and add tests, or convert this into an issue that someone could investigate and fix.
I'm not going to review or read the PR description in the current form.

compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
```diff
@@ -113,6 +113,10 @@ int getComputeVectorSize(int64_t size) {
 }
 
 int getMemoryVectorSize(Value source, Type scalarType, int64_t size) {
+  // Only optimize memory access for int/float types with known bitwidths.
+  // Types like index or i1 cannot be bitcast to 32-bit element vectors.
+  if (!scalarType.isIntOrFloat() || scalarType.isInteger(1))
+    return 1;
```
kuhar: Why is 1 the correct number? Would it make sense to return `failure()` instead?

kuhar (commented):
Thanks for cleaning up the PR

dbrll:
Thanks for your feedback. I rewrote the description and added a lit test.
The test is manually verified and the PR description now fully hand written.

compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
dbrll: On line 140, this function already returns 1 as its fallback.

Since this functions returns an int, I didn't want to change its signature to add `LogicalResult` as it would mean also changing `getNativeVectorShapeImpl` and others, which feels disproportionate to me.

kuhar:
CI failed

kuhar (approved):
LGTM, thanks for contributing

PR #23871, opened 2026-03-20
[VectorDistribute] Pack partial reduction instead of using expand_dims

Groverkss:
Replace the GPUExpandDimensions pass with GPUPackPartialReductions that uses linalg::pack to split reduction dimensions instead of manual expand_shape/collapse_shape pairs with optimization barriers.

- Simplify config generation: remove expand_dims attribute emission and hasExpandCompatibleIndexing check from ReductionConfigUtils.cpp
- New GPUPackPartialReductions pass: for reduction dims where partial_reduction[d] > 1 and thread[d] > 1, pack with inner_tile = thread[d], then update the lowering config to split tile sizes between outer and inner dims
- Remove DimensionExpansionAttr from IREEGPUAttrs (ODS, C++, utils)
- Delete GPUExpandDimensions.cpp
- Update all config and pipeline tests

ci-extra: test_torch


krzysz00 (approved):
I can't see anything wrong here and it's probably a nice simplification, lgtm though if you're worried I'd wait for a second approver

kuhar (commented):
Do we need to update https://iree.dev/developers/lowering-config/#dimension-expansion-expand_dims ?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUPackPartialReductions.cpp
```diff
@@ -0,0 +1,230 @@
+// Copyright 2025 The IREE Authors
```
efric: 2026

compiler/src/iree/compiler/Codegen/Common/GPU/GPUPackPartialReductions.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+  linalg::LinalgOp packedOp = maybeResult->packedLinalgOp;
+
+  // Now build the updated lowering config for the packed op.
+  // The packed op has numLoops + dimsToPackVec.size() dimensions.
+  // For each packed dim d, it splits into [outer=d, inner=newDim].
+  //
+  // Tile size splitting:
+  //   partial_reduction[d] = P, thread[d] = T
+  //   → outer: partial_reduction = P/T, thread = 1
```
efric: i think we prefer ascii chars throughout i.e, `->`

efric (approved):
nice! :smile: 

only nits are to address minor llm leftovers

efric:
> Do we need to update https://iree.dev/developers/lowering-config/#dimension-expansion-expand_dims ?

i can update this after this gets merged to note that this is basically implicit now

Groverkss:
> > Do we need to update https://iree.dev/developers/lowering-config/#dimension-expansion-expand_dims ?
> 
> i can update this after this gets merged to note that this is basically implicit now

Thanks. I think we can just remove that section, we shouldn't need to expose it through lowering_config. These were developer docs anyway so we don't have to worry about changes there.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUPackPartialReductions.cpp
Groverkss: man claude

Groverkss:
This pass is wrong because the upstream linalg::pack actually doesn't work for anything which doesn't have a reduction combiner which doesn't have a zero identity. It also doesn't work for any region containing a linalg.index.

Groverkss:
First fix: https://github.com/llvm/llvm-project/pull/188261

Groverkss:
I got things to work out in the end. The pass documents the problems with linalg::pack

keshavvinayak01 (approved):

PR #23874, opened 2026-03-20
Add default value for IREE_HOST_BIN_DIR

HanKuanChen:
The `build_riscv.sh` script requires the `IREE_HOST_BIN_DIR` environment variable to locate `iree-compile`. By default, `build_all.sh` (when run without additional arguments) installs to `build/install`, meaning `IREE_HOST_BIN_DIR` should point to `build/install/bin`. By setting this as the default path directly within `build_riscv.sh`, we can eliminate the need for manual configuration when running the script.

### DIFF
--- a/build_tools/cmake/build_riscv.sh
+++ b/build_tools/cmake/build_riscv.sh
@@ -8,8 +8,8 @@
 
 # Cross-compile the runtime using CMake targeting RISC-V
 #
-# The required IREE_HOST_BIN_DIR environment variable indicates the location
-# of the precompiled IREE binaries.
+# The IREE_HOST_BIN_DIR environment variable indicates the location of the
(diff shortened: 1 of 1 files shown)


kuhar (commented):
Can you update the PR description and explain why we need this?

HanKuanChen:
> Can you update the PR description and explain why we need this?

The `build_riscv.sh` script requires the `IREE_HOST_BIN_DIR` environment variable to locate `iree-compile`. By default, `build_all.sh` (when run without additional arguments) installs to `build/install`, meaning `IREE_HOST_BIN_DIR` should point to `build/install/bin`. By setting this as the default path directly within `build_riscv.sh`, we can eliminate the need for manual configuration when running the script.

kuhar:
OK. Can you put this in the PR description?

kuhar (approved):

PR #23877, opened 2026-03-20
[Util] Implement int divisibility inference for Add/Sub/Min/Max arith ops

Max191:
Implements the IntegerDivsibilityAnalysis for basic arith ops:
 - arith.addi/subi
 - arith.minsi/maxsi
 - arith.minui/maxui

The divisibility analysis is the same for each of these ops, just returning the GCD of each operand's divisibility.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
+++ b/compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
@@ -331,6 +331,102 @@ util.func @scf_forall_multiple_ivs(%arg0 : index, %arg1 : index, %arg2 : index,
 
 // -----
 
+// CHECK-LABEL: @arith_addi_gcd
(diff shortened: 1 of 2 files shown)


krzysz00 (approved):
Looks correct.

If you'll take a few further notes on operations I figure should be here:
- `divsi` doesn't seem implemented? (Or floordivsi / ceildiv, ...]
- I'm pretty sure this pattern can apply (and should) apply to `arith.select`, even though it's not a binop so you'll have to write it separately

Max191:
> Looks correct.
> 
> If you'll take a few further notes on operations I figure should be here:
> 
>     * `divsi` doesn't seem implemented? (Or floordivsi / ceildiv, ...]
> 
>     * I'm pretty sure this pattern can apply (and should) apply to `arith.select`, even though it's not a binop so you'll have to write it separately

Yep, sounds good! I won't get to it right away, but I left an issue for myself so I can come back to it later: https://github.com/iree-org/iree/issues/23880

PR #23878, opened 2026-03-20
[Codegen] Add VerifyPipelineConstraints pass for compilation info

kuhar:
The pass evaluates pipeline constraints against the actual lowering config to verify that the tuner constraints are consistent with configs the compiler produces. In the future, full tuner constraints can also validate the configs picked by the configuration selection heuristics.

Gated behind `--iree-codegen-experimental-verify-pipeline-constraints` (off by default, implies `--iree-codegen-add-tuner-attributes`). Phase ordering: `SelectLoweringStrategy` -> `InsertSMTConstraints` -> `Verify`, since we need the root ops to be decided already. Constraint ops are erased regardless of outcome since they only serve verification and tuning.

Implemented with a simple evaluator; I considered using folders or dataflow instead, but I like the simplicity of the evaluator.

This is not fully generic yet and only works for llvmgpu constraints. In the future, we'd need extra interfaces to make it fully generic without assuming the exact dialects used for compilation info attrs.


compiler/src/iree/compiler/Codegen/Common/VerifyPipelineConstraints.cpp
```diff
@@ -0,0 +1,503 @@
...
+/// walks each op, tracks known int/bool values, and collects assertion
+/// violations with formatted diagnostic messages.
+///
+/// Ops whose inputs are unknown propagate unknown (std::nullopt) through
+/// the dataflow, so assertions involving unresolved knobs are silently
+/// skipped rather than producing false positives.
+struct ConstraintEvaluator {
+  DenseMap<Value, std::optional<int64_t>> intValues;
+  DenseMap<Value, std::optional<bool>> boolValues;
+  SmallVector<std::pair<Location, std::string>> violations;
```
qedawkins: Are these supposed to be public?

compiler/src/iree/compiler/Codegen/Common/VerifyPipelineConstraints.cpp
```diff
@@ -0,0 +1,503 @@
...
+      if (failed(result)) {
+        return failure();
+      }
+    }
+    return success();
+  }
+
+private:
+  std::optional<int64_t>
+  evalVariadicInt(ValueRange inputs, int64_t identity,
```
qedawkins: You can drop the identity if you just initialize the accumulator with the first value. I assume all ops with summation semantics require at least a single operand?

compiler/src/iree/compiler/Codegen/Common/VerifyPipelineConstraints.cpp
```diff
@@ -0,0 +1,503 @@
...
+  }
+
+  LogicalResult eval(smt::IntConstantOp constOp,
+                     const DenseMap<StringAttr, int64_t> &) {
+    intValues[constOp.getResult()] = constOp.getValue().getSExtValue();
+    return success();
+  }
+
+  LogicalResult eval(IREE::Codegen::KnobOp knobOp,
+                     const DenseMap<StringAttr, int64_t> &knobValues) {
```
qedawkins: This is the only case that actually uses `knobValues`. Can we remove the operand from all other cases and split this one out in the above?

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -1109,6 +1109,11 @@ static void buildLLVMGPUCodegenConfigurationPassPipelineImpl(
   modulePassManager.addPass(createMaterializeTuningSpecsPass());
   modulePassManager.addPass(createMaterializeUserConfigsPass());
   modulePassManager.addPass(createLLVMGPUSelectLoweringStrategyPass());
+  // After strategy selection, root ops are annotated and lowering configs set.
+  // Emit constraints and verify them against the chosen config.
+  FunctionLikeNest(modulePassManager)
+      .addPass(createInsertSMTConstraintsPass)
+      .addPass(createVerifySMTConstraintsPass);
```
qedawkins: If the passes are gated on a compiler level flag I would expect them to be gated here rather than within the pass itself actually. I missed this in the previous change.

qedawkins (approved):
Sorry, got to this late. The evaluator is neat, I like the expression trees in the error messages :)

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
kuhar: gah, right, that's a much better way to go about it

compiler/src/iree/compiler/Codegen/Common/VerifyPipelineConstraints.cpp
kuhar: I made it a member of this struct instead, since it's an immutable input

PR #23879, opened 2026-03-20
[Util] Add NarrowVectorBroadcast pattern to OptimizeIntArithmeticPass

Max191:
When doing integer type narrowing, index casts can be leftover because broadcast ops don't get narrowed, so we get left with sequences like:
```mlir
%add = arith.addi %lhs, %rhs : i32
%cast0 = arith.index_cast %add : i32 to index
%broadcast = vector.broadcast %cast0 : index to vector<4xindex>
%cast1 = arith.index_cast %broadcast : vector<4xindex> to vector<4xi32>
```

This PR adds a cleanup pattern that narrows the broadcast op so that the index casts can all fold.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Transforms/BUILD.bazel
+++ b/compiler/src/iree/compiler/Dialect/Util/Transforms/BUILD.bazel
@@ -88,6 +88,7 @@ iree_compiler_cc_library(
         "@llvm-project//mlir:TransformUtils",
         "@llvm-project//mlir:Transforms",
         "@llvm-project//mlir:UBDialect",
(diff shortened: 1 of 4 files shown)


krzysz00 (approved):
1. I think this can also be put upstream in their copies of this code
2. You might also want broadcast(ext[su]i(x)) => ext[su]i(broadcast(x)) by the same logic?
3. Logic seems sound if somewhat narrow, approved.

krzysz00:
Oh, and, 4. Get `shape_cast` too.

Max191:
Made an issue for this one too: https://github.com/iree-org/iree/issues/23882

Maybe I'll see what Claude can do with just the issue link

PR #23881, opened 2026-03-20
Add 256-alignment padding tier for IGEMM tile size heuristics

Max191:
Adds a padding tier for dims > 256 to align to 256, complementing the existing 128 and 32 tiers.


Max191:
Waiting on benchmarks to run, and will add to the description when ready.

Max191:
This has some regressions for some convolution shapes when we don't flatten, so this will have to come as part of the larger flattening PR. Closing this PR for now.

PR #23883, opened 2026-03-20
[Base] fix printf issue in gcc

efric:
`unsigned int abs_n = (unsigned int)(-n);` did not provide the “ironclad range proof” we were looking for gcc 11 because the cast happens after `-n` is evaluated, so we can still get errors (https://github.com/iree-org/iree/actions/runs/23336776640/job/67880398515) complaining about it. fix by just casting first to get the intended behavior. While we’re here, made things a bit more explicit and simplify the lookup a bit. 

working gcc run: https://github.com/iree-org/iree/actions/runs/23359589322

### DIFF
--- a/runtime/src/iree/base/printf.c
+++ b/runtime/src/iree/base/printf.c
@@ -775,13 +775,12 @@ static const double kPowersOf10[] = {
 };
 
 static double iree_printf_pow10(int n) {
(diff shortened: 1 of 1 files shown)


benvanik (approved):
*shakes fist at GCC*

PR #23875, opened 2026-03-20
[CI] Enable sccache for linux_x64_clang post-submit jobs 

amd-eochoalo:
./setup_ccache clobbers over sccache which leads to zero hits when building linux_x64_clang. 
This change allows sccache to be used correctly (still only from this repo's PRs and not forks).

* First run took [18 minutes](https://github.com/iree-org/iree/actions/runs/23348078464/job/67918880748)
* Second run took [6 minutes](https://github.com/iree-org/iree/actions/runs/23348894907/job/67922550630?pr=23875)

```
+ sccache --show-stats
Compile requests                   6774
Compile requests executed          6774
Cache hits                         6756
Cache hits (C/C++)                 6756
Cache misses                          6
Cache misses (C/C++)                  6
```

PR #23886, opened 2026-03-20
Fix sccache on windows

amd-eochoalo:
Similar to #23875 but for windows.

According to claude: there may be some compatibility issues between sccache and msvc. Leaving as draft for now.


amd-eochoalo:
Closed in favour of #24096

PR #23889, opened 2026-03-20
[CPU] Add vector level tiling sizes to LinalgExt::FftOp.

### LINKED ISSUE #22776: Getting "op exceeded stack allocation limit" from linalg_ext.fft
### What happened?

The `linalg_ext.fft` operation outputs 2 tensors: the real part and the imaginary part of the operation. If the tensors are large enough and one of the output tensors is not used later in the program, the compile will fail with `error: 'func.func' op exceeded stack allocation limit of 32768 bytes for function.`. This compile error is surprising for users because the exact same program works with smaller tensor sizes, then fails to compile once the tensors are large enough, even if the size of the 1D FFT is the same.

For example, here is the last part of an IRFFT (takes a complex input and returns a real output).
```
%5:2 = iree_linalg_ext.fft ins(%c1, %cst_8, %cst_7 : index, tensor<1xf32>, tensor<1xf32>) outs(%4#0, %4#1 : tensor<32xf32>, tensor<32xf32>) : tensor<32xf32>, tensor<32xf32>
%6:2 = iree_linalg_ext.fft ins(%c2, %cst_6, %cst_5 : index, tensor<2xf32>, tensor<2xf32>) outs(%5#0, %5#1 : tensor<32xf32>, tensor<32xf32>) : tensor<32xf32>, tensor<32xf32>
%7:2 = iree_linalg_ext.fft ins(%c3, %cst_4, %cst_3 : index, tensor<4xf32>, tensor<4xf32>) outs(%6#0, %6#1 : tensor<32xf32>, tensor<32xf32>) : tensor<32xf32>, tensor<32xf32>
%8:2 = iree_linalg_ext.fft ins(%c4, %cst_2, %cst_1 : index, tensor<8xf32>, tensor<8xf32>) outs(%7#0, %7#1 : tensor<32xf32>, tensor<32xf32>) : tensor<32xf32>, tensor<32xf32>
%9:2 = iree_linalg_ext.fft ins(%c5, %cst_0, %cst : index, tensor<16xf32>, tensor<16xf32>) outs(%8#0, %8#1 : tensor<32xf32>, tensor<32xf32>) : tensor<32xf32>, tensor<32xf32>
util.return %9#0 : tensor<32xf32>
```
Notice that it returns `%9#0` and ignores `%9#1` because the imaginary part is not needed. The problem is that, once the input sizes get large enough, this errors out with `error: 'func.func' op exceeded stack allocation limit`. The reason is that the imaginary part of that last FFT got marked as readonly. Here are what the tensors to that last FFT get turned into
```
%0 = hal.interface.binding.subspan layout(#pipeline_layout3) binding(0) alignment(64) offset(%c0) flags(Indirect) : !iree_tensor_ext.dispatch.tensor<readwrite:tensor<32xf32>>
%1 = hal.interface.binding.subspan layout(#pipeline_layout3) binding(1) alignment(64) offset(%c256) flags("ReadOnly|Indirect") : !iree_tensor_ext.dispatch.tensor<readonly:tensor<32xf32>>
```

We've been working around this by using the imaginary output in some non-trivial way to essentially fool the compiler into not marking it readonly. I discovered that using `util.optimization_barrier %9#1` will also work around the issue.

@MaheshRavishankar suggested in issue https://github.com/iree-org/iree/issues/22695 that the right fix would be vectorize the FFT operation.

@hanhanW Moving the conversation here from https://github.com/iree-org/iree/issues/22473

Repro case is attached here: [irfft.zip](https://github.com/user-attachments/files/23779602/irfft.zip)

iree-compile --iree-hal-target-device=local --iree-hal-local-target-device-backends=llvm-cpu --iree-llvmcpu-target-cpu=host -o /dev/null irfft.mlir

Compiler

### Version information

3.9.0 (I also tried 3.8.0 and 3.5.0)

### Additional context

_No response_

hanhanW:
Fixes https://github.com/iree-org/iree/issues/22776

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -2514,22 +2514,36 @@ static LogicalResult setRootConfig(mlir::FunctionOpInterface entryPointFn,
   SmallVector<int64_t> distTileSizes =
       getDefaultDistributedLevelTileSizes(fftOp, DistributionHeuristicConfig{});
   auto rank = fftOp.getOperandRank();
-  if (distTileSizes.size() >= rank && distTileSizes[rank - 1] != 0) {
-    APInt value;
-    if (matchPattern(fftOp.getStage(), m_ConstantInt(&value))) {
-      distTileSizes[rank - 1] = 1ll << value.getSExtValue();
-      distTileSizes[rank - 1] = std::max(
-          distTileSizes[rank - 1], static_cast<int64_t>(clDefaultDistTileSize));
-    } else {
-      return fftOp.emitOpError("non-constant stage might not work for fft op");
-    }
+  std::optional<int64_t> stageValue = getConstantIntValue(fftOp.getStage());
+  if (!stageValue) {
+    return fftOp.emitOpError("non-constant stage might not work for fft op");
   }
-  // Append vector level tiling sizes using zero values, which means no tiling
-  // in the pipeline.
+  if (distTileSizes.size() >= rank && distTileSizes[rank - 1] != 0) {
+    distTileSizes[rank - 1] = 1ll << *stageValue;
+    distTileSizes[rank - 1] = std::max(
+        distTileSizes[rank - 1], static_cast<int64_t>(clDefaultDistTileSize));
+  }
+  // Set vector tile sizes. The FFT dimension is set to the full butterfly
+  // stride (= 2^stage = 2*halfSize). Each butterfly group reads lhs at
+  // [0..halfSize-1] and rhs at [halfSize..2*halfSize-1], so the tile must
+  // span the complete group. The innermost batch dimension is scaled inversely
+  // so the total vector width stays close to the native vector size, keeping
+  // SIMD utilization high across all stages.
+  int64_t halfSize = 1ll << (*stageValue - 1);
+  int64_t fftTileSize = 2 * halfSize;
+  int64_t nativeVectorSize =
+      getNativeVectorSizeInBytes(entryPointFn) /
+      (fftOp.getOperandType().getElementTypeBitWidth() / 8);
+  // Scale the innermost batch dim to fill the native vector width.
+  int64_t batchTile = std::max<int64_t>(1, nativeVectorSize / halfSize);
   LoweringConfigGenerator generator(fftOp);
   generator.setDistributionTileSizes(distTileSizes);
-  SmallVector<int64_t> zeros(rank, 0);
-  generator.setVectorTileSizes(zeros);
+  SmallVector<int64_t> vecTiles(rank, 1);
+  if (rank >= 2) {
+    vecTiles[rank - 2] = batchTile;
+  }
+  vecTiles.back() = fftTileSize;
+  generator.setVectorTileSizes(vecTiles);
   IREE::CPU::LoweringConfigAttr loweringConfig =
       generator.generateCPULoweringConfig();
   return setOpConfigAndEntryPointFnTranslation(
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_x86_64_lowering_strategy.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_x86_64_lowering_strategy.mlir
@@ -205,7 +205,7 @@ func.func @static_1d_fft_stage2(%2: tensor<32xf32>, %3: tensor<32xf32>) -> tenso
   %4:2 = iree_linalg_ext.fft ins(%c2, %cst, %cst_0 : index, tensor<2xf32>, tensor<2xf32>) outs(%2, %3 : tensor<32xf32>, tensor<32xf32>) : tensor<32xf32>, tensor<32xf32>
   return %4 : tensor<32xf32>
 }
-//   CHECK-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64]>
+//   CHECK-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64], vector_common_parallel = [4]>
 //   CHECK-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<LinalgExtTileAndVectorize>>
 //       CHECK: func.func @static_1d_fft_stage2(
 //  CHECK-SAME:     translation_info = #[[TRANSLATION]]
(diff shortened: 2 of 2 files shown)


hanhanW:
@pstarkcdpr can you help review? Thanks

pstarkcdpr:
I've never looked at IREE vectorizing code before, so this is new to me. It passes my internal FFT tests, so it looks good.

Forgive me for asking dumb questions, but since I'm new to this, I figured I'd just ask if everything makes sense. Take the following simple IRFFT
```
module @jit_irfft {
  func.func public @main(%arg0: tensor<128x153x17xcomplex<f32>>) -> tensor<128x153x32xf32> {
    %0 = stablehlo.fft %arg0, type =  IRFFT, length = [32] : (tensor<128x153x17xcomplex<f32>>) -> tensor<128x153x32xf32>
    return %0 : tensor<128x153x32xf32>
  }
}
```
 Looking at `executable-configurations`, the big change compared to previous versions of IREE is:
```
#config3 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 16, 2]>
#config4 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 8, 4]>
#config5 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 4, 8]>
#config6 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 2, 16]>
#config7 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 1, 32]>
```
As the middle dimension scales down, the innermost dimension scales up to keep things at 32, right? Why 32? I don't actually know what the units are here, but I have `native_vector_size = 64`, and these are `f32`s (4 bytes), so I intuitively expected either 16 (since `4*16=64` bytes) or 64 (since that's the native vector size).

hanhanW:
> I've never looked at IREE vectorizing code before, so this is new to me. It passes my internal FFT tests, so it looks good.

It is not the vectorization support, it is more like tiling which produces smaller ranges in the innermost loops. The dimensions will be bounded by `vector_common_parallel` dims, so the stack allocation (which may be promoted to register in LLVM backend) is bounded by tiling sizes.

> Forgive me for asking dumb questions, but since I'm new to this, I figured I'd just ask if everything makes sense. Take the following simple IRFFT
> 
> ```
> module @jit_irfft {
>   func.func public @main(%arg0: tensor<128x153x17xcomplex<f32>>) -> tensor<128x153x32xf32> {
>     %0 = stablehlo.fft %arg0, type =  IRFFT, length = [32] : (tensor<128x153x17xcomplex<f32>>) -> tensor<128x153x32xf32>
>     return %0 : tensor<128x153x32xf32>
>   }
> }
> ```
> 
> Looking at `executable-configurations`, the big change compared to previous versions of IREE is:
> 
> ```
> #config3 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 16, 2]>
> #config4 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 8, 4]>
> #config5 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 4, 8]>
> #config6 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 2, 16]>
> #config7 = #iree_cpu.lowering_config<distribution = [32, 51, 64], vector_common_parallel = [1, 1, 32]>
> ```
> 
> As the middle dimension scales down, the innermost dimension scales up to keep things at 32, right? Why 32? I don't actually know what the units are here, but I have `native_vector_size = 64`, and these are `f32`s (4 bytes), so I intuitively expected either 16 (since `4*16=64` bytes) or 64 (since that's the native vector size).

Any questions are welcome, and they are not dumb!

The op is designed base on [Cooley–Tukey FFT algorithm
](https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm), where the algorithm is:

```
algorithm iterative-fft is
    input: Array a of n complex values where n is a power of 2.
    output: Array A the DFT of a.
 
    bit-reverse-copy(a, A)
    n ← a.length 
    for s = 1 to log(n) do
        m ← 2s
        ωm ← exp(−2πi/m) 
        for k = 0 to n-1 by m do
            ω ← 1
            for j = 0 to m/2 – 1 do
                t ← ω A[k + j + m/2]
                u ← A[k + j]
                A[k + j] ← u + t
                A[k + j + m/2] ← u – t
                ω ← ω ωm
   
    return A
```

(copy-paste does not work well, so you may want to check the wiki page.......)

Looking at the heuristic and comment in the code snippet below. It mainly set the innermost tile size to `2^s`, which matches a "butterfly pair". Then it scales the batch dimension with the given `native_vector_size` factor and fftTileSize factor, which sticks to `32` in this case.

```cpp
  // Set vector tile sizes. The FFT dimension is set to the full butterfly
  // stride (= 2^stage = 2*halfSize). Each butterfly group reads lhs at
  // [0..halfSize-1] and rhs at [halfSize..2*halfSize-1], so the tile must
  // span the complete group. The innermost batch dimension is scaled inversely
  // so the total vector width stays close to the native vector size, keeping
  // SIMD utilization high across all stages.
  int64_t halfSize = 1ll << (*stageValue - 1);
  int64_t fftTileSize = 2 * halfSize;
  int64_t nativeVectorSize =
      getNativeVectorSizeInBytes(entryPointFn) /
      (fftOp.getOperandType().getElementTypeBitWidth() / 8);
  // Scale the innermost batch dim to fill the native vector width.
  int64_t batchTile = std::max<int64_t>(1, nativeVectorSize / halfSize);
  LoweringConfigGenerator generator(fftOp);
  generator.setDistributionTileSizes(distTileSizes);
  SmallVector<int64_t> vecTiles(rank, 1);
  if (rank >= 2) {
    vecTiles[rank - 2] = batchTile;
  }
  vecTiles.back() = fftTileSize;
```

pstarkcdpr:
Ah, thanks. That makes sense. Since my FFT size is reasonable (32), with the new tiling, the arbitrarily large outer dimensions can't cause large stack allocations anymore.

hanhanW:
> Ah, thanks. That makes sense. Since my FFT size is reasonable (32), with the new tiling, the arbitrarily large outer dimensions can't cause large stack allocations anymore.

yep, that's right! FYI, I have a vectorization PR coming which improves ~15% perf, if the coeff is not constant. For constant, which is more pratical, it does not matter because it is memory bound. The generated asm code will look better, indeed.

nirvedhmeshram (approved):
LGTM

PR #23893, opened 2026-03-20
[Codegen] Define DispatchConfigOp and YieldOp to IREE::Codegen.

hanhanW:
- `dispatch_config`: module-level op holding workgroup count computation and dispatch metadata (workgroup_size, subgroup_size) for a function. Uses IsolatedFromAbove with a single-block region whose block args are workload values (index type). The regions returns exactly 3 index values (workgroup count x, y, z), which is as the same as HAK::ExportOp that requires exactly three yielded values in `verifyWorkgroupCountRegion`.
- `yield`: terminator for IREE::Codegen op's regions.

The dispatch_config op uses a plain StrAttr (printed as @name) rather than the Symbol trait, because it must coexist with a func.func of the same name in the same builtin.module symbol table. It can't be symbol, because two ops can't define the same symbol. The dispatch_config op is expected to live along the corresponding func op.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
@@ -26,31 +26,46 @@
 #include "mlir/Interfaces/ValueBoundsOpInterface.h"
 #include "mlir/Support/LLVM.h"
 
+using namespace mlir;
+using namespace mlir::iree_compiler::IREE::Codegen;
+namespace IREE = mlir::iree_compiler::IREE;
+
+//===----------------------------------------------------------------------===//
+// Custom parsers/printers (must precede generated .cpp.inc)
+//===----------------------------------------------------------------------===//
+
 // Custom parse/print helper for the knobs dictionary in constraints op.
 // Prints `knobs = { ... }` on its own line with newlines before and after.
-static mlir::ParseResult parseKnobsDictionary(mlir::OpAsmParser &parser,
-                                              mlir::DictionaryAttr &attr) {
+static ParseResult parseKnobsDictionary(OpAsmParser &parser,
+                                        DictionaryAttr &attr) {
   if (parser.parseKeyword("knobs") || parser.parseEqual()) {
-    return mlir::failure();
+    return failure();
   }
   return parser.parseAttribute(attr);
 }
-static void printKnobsDictionary(mlir::OpAsmPrinter &p, mlir::Operation *,
-                                 mlir::DictionaryAttr attr) {
+static void printKnobsDictionary(OpAsmPrinter &p, Operation *,
+                                 DictionaryAttr attr) {
   p.printNewline();
   p << " knobs = ";
   p.printAttributeWithoutType(attr);
   p.printNewline();
 }
 
+/// Parses a string attribute printed in `@name` syntax.
+static ParseResult parseFunctionRef(OpAsmParser &parser, StringAttr &attr) {
+  return parser.parseSymbolName(attr);
+}
+
+/// Prints a string attribute in `@name` syntax.
+static void printFunctionRef(OpAsmPrinter &printer, Operation *,
+                             StringAttr attr) {
+  printer.printSymbolName(attr.getValue());
+}
+
 // clang-format off
 #define GET_OP_CLASSES
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp.inc" // IWYU pragma: keep
 // clang-format on
-
-using namespace mlir;
-using namespace mlir::iree_compiler::IREE::Codegen;
-namespace IREE = mlir::iree_compiler::IREE;
 
 //===----------------------------------------------------------------------===//
 // ExtractStridedMetadataOp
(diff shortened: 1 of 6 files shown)


krzysz00:
... High-level comment - couldn't you, instead of using a StringAttr, use a SymbolRefAttr to make the dispatch config reference the function it describes? Then it'd show up as a "caller" of the function, effectively.

krzysz00:
Or couldn't we stick a discardable attribute on the function that has a reference to the dispatch config symbol?

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -727,6 +728,61 @@ def IREECodegen_AssertOp :
...
+    A module-level op that captures the workgroup count computation and
+    dispatch metadata for a function. The region computes the workgroup
+    count (x, y, z) from workload values passed as block arguments.
+    The terminator must yield exactly 3 index values, matching the
+    contract of `hal.executable.export`'s workgroup count region.
+
+    The `function_ref` names the corresponding `func.func`. The op is
+    not a Symbol because it lives alongside the `func.func` of the same
+    name in a `builtin.module` symbol table, and two ops cannot define
+    the same symbol.
```
qedawkins: I haven't looked at the later patches yet but I would expect this to be a symbol user so that symbol name updates can catch this op. This also gives us verification that the symbol exists.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
```diff
@@ -26,32 +26,47 @@
 #include "mlir/Interfaces/ValueBoundsOpInterface.h"
 #include "mlir/Support/LLVM.h"
 
+using namespace mlir;
+using namespace mlir::iree_compiler::IREE::Codegen;
+namespace IREE = mlir::iree_compiler::IREE;
```
qedawkins: clang-tidy saying the last line is unused?

qedawkins (commented):
Couple quick comments then this LGTM

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/test/roundtrip.mlir
```diff
@@ -316,3 +316,49 @@ func.func @smt_lookup_sparse(%arg0: index) {
...
+  %c2 = arith.constant 2 : index
+  %c1 = arith.constant 1 : index
+  iree_codegen.yield %c4, %c2, %c1 : index, index, index
+}
+// CHECK-LABEL: iree_codegen.dispatch_config @static_count
+// CHECK-SAME:    workgroup_size = [64]
+// CHECK:         %[[C4:.+]] = arith.constant 4 : index
+// CHECK:         %[[C2:.+]] = arith.constant 2 : index
+// CHECK:         %[[C1:.+]] = arith.constant 1 : index
+// CHECK:         iree_codegen.yield %[[C4]], %[[C2]], %[[C1]] : index, index, index
```
qedawkins, line 364: What does parsing/printing look like with no workgroup_size (and no subgroup_size)? I don't see a test for it unless I missed it.

qedawkins (commented):
Oops didn't submit this one.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -727,6 +728,61 @@ def IREECodegen_AssertOp :
...
+      ^bb0(%w0: index, %w1: index, %w2: index, %w3: index):
+        %0 = affine.apply affine_map<()[s0] -> (s0 ceildiv 256)>()[%w2]
+        %c1 = arith.constant 1 : index
+        iree_codegen.yield %0, %c1, %c1 : index, index, index
+    }
+    ```
+  }];
+
+  let arguments = (ins
+    StrAttr:$function_ref,
```
MaheshRavishankar: Is there a reason for this to be a `StringRef` instead of a `SymbolRef` ?

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
```diff
@@ -727,6 +728,61 @@ def IREECodegen_AssertOp :
...
+    OptionalAttr<I64Attr>:$subgroup_size
+  );
+  let regions = (region SizedRegion<1>:$body);
+
+  let builders = [
+    OpBuilder<(ins "::llvm::StringRef":$functionRef)>,
+  ];
+
+  let assemblyFormat = [{
+    custom<FunctionRef>($function_ref)
```
MaheshRavishankar: Because this is not a `SymbolRef` now this is "cheating" to make it look like a `SymbolRef`.

MaheshRavishankar (changes requested):

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.td
hanhanW: I have a follow-up commit coming which makes it `FlatSymbolRefAttr`. I think they belong the same scope, so `FlatSymbolRefAttr` is better than `SymbolRef`.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
hanhanW: I restore the changes, it is a pre-existing issue. I can help fix in a different PR

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
```diff
@@ -472,6 +472,59 @@ void WorkgroupCountHintOp::print(OpAsmPrinter &printer) {
                                 /*elidedAttrs=*/{"static_sizes"});
 }
 
+//===----------------------------------------------------------------------===//
+// DispatchConfigOp
+//===----------------------------------------------------------------------===//
+
+void DispatchConfigOp::build(OpBuilder &odsBuilder, OperationState &odsState,
+                             FlatSymbolRefAttr functionRef) {
+  odsState.addAttribute("function_ref", functionRef);
```
MaheshRavishankar: Its strange that you would need to do this. You should look at the generated class to see if there is a class static method you can use to get this string name.

MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.cpp
hanhanW: I think I should just use the builder from the base builder and just pass nullptr.

PR #23894, opened 2026-03-20
[Codegen] Introduce CreateDispatchConfig pass.

hanhanW:
Clones export count region bodies into dispatch_config ops placed after the corresponding function in the inner module.

If the body does not exist, creates the body that returns `{1, 1, 1}`.

The revision also relax the constraint of block argument, so it can accept whatever types. See the discussion on disocrd: https://discord.com/channels/689900678990135345/690274711523164166/1486476323001208993

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
@@ -8,9 +8,7 @@
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
 #include "iree/compiler/Dialect/HAL/IR/HALOps.h"
 #include "iree/compiler/Dialect/HAL/IR/HALTypes.h"
-#include "llvm/ADT/SmallVectorExtras.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
-#include "mlir/IR/IRMapping.h"
 
 namespace mlir::iree_compiler {
 
@@ -66,39 +64,20 @@ void CreateDispatchConfigPass::runOnOperation() {
       return signalPassFailure();
     }
 
-    // Dispatch config block args: (index, index, ...) — skip !hal.device.
-    unsigned configArity = exportBlock->getNumArguments() - 1;
-
-    // Create dispatch_config right after the function.
+    // Create dispatch_config right after the function, clone the export region
+    // and drop the first block argument (!hal.device).
     builder.setInsertionPointAfter(funcOp);
-    auto configOp =
-        IREE::Codegen::DispatchConfigOp::create(builder, loc, FlatSymbolRefAttr::get(funcOp.getNameAttr()));
-
-    // Create block with workload args (index type).
-    Block *configBlock = builder.createBlock(&configOp.getBody());
-    for (unsigned i = 0; i < configArity; ++i) {
-      configBlock->addArgument(builder.getIndexType(), loc);
-    }
-
-    // Clone ops from export count region into dispatch_config body.
-    // Map export block args (offset by 1) to config block args.
-    IRMapping mapping;
-    for (unsigned i = 0; i < configArity; ++i) {
-      mapping.map(exportBlock->getArgument(i + 1), configBlock->getArgument(i));
-    }
-
-    builder.setInsertionPointToEnd(configBlock);
-    for (Operation &op : exportBlock->without_terminator()) {
-      builder.clone(op, mapping);
-    }
-
-    // Replace hal.return with iree_codegen.yield.
-    auto returnOp = cast<IREE::HAL::ReturnOp>(exportBlock->getTerminator());
-    auto yieldValues =
-        llvm::map_to_vector(returnOp.getOperands(), [&](Value v) {
-          return mapping.lookupOrDefault(v);
-        });
-    IREE::Codegen::YieldOp::create(builder, returnOp.getLoc(), yieldValues);
+    auto configOp = IREE::Codegen::DispatchConfigOp::create(
+        builder, loc, FlatSymbolRefAttr::get(funcOp.getNameAttr()));
+    builder.cloneRegionBefore(exportOp.getWorkgroupCount(), configOp.getBody(),
+                              configOp.getBody().end());
+    Block *configBlock = &configOp.getBody().front();
+    configBlock->eraseArgument(0);
+    auto returnOp = cast<IREE::HAL::ReturnOp>(configBlock->getTerminator());
+    builder.setInsertionPoint(returnOp);
+    IREE::Codegen::YieldOp::create(builder, returnOp.getLoc(),
+                                   returnOp.getOperands());
+    returnOp.erase();
   }
 }
 
--- a/compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
@@ -1,0 +1,106 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/Passes.h"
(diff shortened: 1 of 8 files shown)


hanhanW:

compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    // Create dispatch_config right after the function.
+    builder.setInsertionPointAfter(funcOp);
+    auto configOp =
+        IREE::Codegen::DispatchConfigOp::create(builder, loc, funcOp.getName());
+
+    // Create block with workload args (index type).
+    Block *configBlock = builder.createBlock(&configOp.getBody());
+    for (unsigned i = 0; i < configArity; ++i) {
+      configBlock->addArgument(builder.getIndexType(), loc);
```
MaheshRavishankar: We dont need to make this `Index` type, you can just use the same index type as the exportop.

compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // Clone ops from export count region into dispatch_config body.
+    // Map export block args (offset by 1) to config block args.
+    IRMapping mapping;
+    for (unsigned i = 0; i < configArity; ++i) {
+      mapping.map(exportBlock->getArgument(i + 1), configBlock->getArgument(i));
+    }
+
+    builder.setInsertionPointToEnd(configBlock);
```
MaheshRavishankar: You should be able to use [cloneRegionBefore](https://github.com/llvm/llvm-project/blob/4cef1c13416d020cc2f8a67a374ff34b444f1b5b/mlir/include/mlir/IR/Builders.h#L611)

compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    // Create dispatch_config right after the function, clone the export region
+    // and drop the first block argument (!hal.device).
+    builder.setInsertionPointAfter(funcOp);
+    auto configOp = IREE::Codegen::DispatchConfigOp::create(
+        builder, loc, FlatSymbolRefAttr::get(funcOp.getNameAttr()));
+    builder.cloneRegionBefore(exportOp.getWorkgroupCount(), configOp.getBody(),
+                              configOp.getBody().end());
+    Block *configBlock = &configOp.getBody().front();
```
MaheshRavishankar, line 66: Doing this without the rewriter is probably going to create a undefined state if a listener is using this. I think it is better to use IRMapping and just map the first argument of the source block to nullptr and then clone it.

hanhanW: I'm not sure if you are right or not because the [cloneRegionBefore](https://github.com/llvm/llvm-project/blob/4cef1c13416d020cc2f8a67a374ff34b444f1b5b/mlir/include/mlir/IR/Builders.h#L611) does not use builder/rewriter at all.

hanhanW: I think it does not matter because the OpBuilder is a plain builder with no listener attached. There is no listener to notify.

MaheshRavishankar: What do you mean it doesnt use the builder? It is part of the `OpBuilder` class? Also `OpBuilder` has a Listener https://github.com/llvm/llvm-project/blob/4cef1c13416d020cc2f8a67a374ff34b444f1b5b/mlir/include/mlir/IR/Builders.h#L318

MaheshRavishankar: Why are you not using the cloneRegionBefore with IRMapping. You should check for any uses of `!hal.device` argument and bail (cause that isnt supported in your flow), and then you can map that argument to nullptr and use `cloneRegionBefore` with the `IRMapping` to clone from the previous region into the new region.

MaheshRavishankar (changes requested):

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -464,6 +464,26 @@ def ResolveWorkgroupCountHintsPass
...
+def CreateDispatchConfigPass
+    : Pass<"iree-codegen-create-dispatch-config", "IREE::HAL::ExecutableVariantOp"> {
+  let summary = "Create dispatch_config ops by cloning export count regions.";
+  let description = [{
+    For each `hal.executable.export`, creates an `iree_codegen.dispatch_config`
+    op in the inner module by cloning the export's count region body. The
+    dispatch_config body inherits the placeholder ops (`from_slice`,
+    `split_reduction_modifier`) with block args mapped from the export's
+    workload args (skipping `!hal.device` at position 0). The terminator is
+    replaced with `iree_codegen.yield`.
```
qedawkins: We need a plan for how to handle the `!hal.device`. My thinking was that the dispatch_config op wouldn't clone the whole region and instead just represents what `iree_tensor_ext.dispatch.workgroup_count_from_slice` returns, then later (potentially per backend) passes are responsible for setting counts based on device queries separately, but the clone everything -> overwrite everything strategy doesn't seem to leave an opportunity to do device queries. There are other approaches we can take, I just gave one idea.

Not going to block since it has no immediate impact on any actual use cases.

qedawkins (approved):
LGTM % Mahesh's comments

compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
hanhanW: It does not have a listener because the constructor uses `nullptr` by default: https://github.com/llvm/llvm-project/blob/4cef1c13416d020cc2f8a67a374ff34b444f1b5b/mlir/include/mlir/IR/Builders.h#L214-L216

Anyway, I think we don't need `IRMapping` here because it does nothing. The `cloneRegionBefore` method inline all the blocks including the block arguments. I took a look at what `IRMapping` could do, and it seems that it's for when the region being cloned references values defined outside that region. E.g.,

```
  %threshold = arith.constant 42 : index
  scf.for %i = ... {
    %cmp = arith.cmpi ult, %i, %threshold  // references %threshold from outside
  }
```

In this case, we don't need `IRMapping`.

compiler/src/iree/compiler/Codegen/Common/Passes.td
MaheshRavishankar: For this change it is fine to clone the region. 

But going forward, I am not sure it works if you dont clone the whole region...

compiler/src/iree/compiler/Codegen/Common/CreateDispatchConfig.cpp
MaheshRavishankar: > It does not have a listener because the constructor uses nullptr by default: https://github.com/llvm/llvm-project/blob/4cef1c13416d020cc2f8a67a374ff34b444f1b5b/mlir/include/mlir/IR/Builders.h#L214-L216

You can always set a listener later outside of the constuctor. We might not use the listener anywhere, but its a kind of footgun that is hard to find if we dont use the builder.

Regarding IRMapping, you are right we probably dont need it, but we need to use the builder method.

hanhanW: It is using the `cloneRegionBefore` method from the builder, does it look good now?

compiler/src/iree/compiler/Codegen/Common/Passes.td
hanhanW: My original idea was about matching without IREE's ops. After thinking more, I realized that we already use IREE's ops during codegen, so maybe uncondionally creating and propagating DispatchConfigOp is fine, if the block arguments match.

On the device-codegen-only side (that @qedawkins may not have context), the op can be created with `iree_tensor_ext.dispatch.workgroup_count_from_slice` ops; it's aligned with the idea: the export op is provided by frontend and the goal of the pass is localizing it to module scope for Codegen.

So how about we make it a pure clone?

MaheshRavishankar (approved):
Sorry for being pedantic about it. Just something seemed off, and wanted to understand the reasoning a bit more.

PR #23895, opened 2026-03-20
[Codegen] Introduce PropagateDispatchConfig pass.

hanhanW:
The pass clones the dispatch metadata and slice computation from IREE::Codegen::DispatchConfig op to HAL::ExportOp. It works at `IREE::HAL::ExecutableVariantOp` scope because it needs to access the export op and configs within the inner module.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
@@ -1,0 +1,121 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "iree/compiler/Dialect/HAL/IR/HALOps.h"
+#include "llvm/ADT/SmallVectorExtras.h"
+#include "llvm/Support/DebugLog.h"
+#include "mlir/IR/IRMapping.h"
+
+#define DEBUG_TYPE "iree-codegen-propagate-dispatch-config"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_PROPAGATEDISPATCHCONFIGPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+
+class PropagateDispatchConfigPass final
+    : public impl::PropagateDispatchConfigPassBase<
+          PropagateDispatchConfigPass> {
+public:
+  using Base::Base;
+  void runOnOperation() override;
+};
+
+void PropagateDispatchConfigPass::runOnOperation() {
+  IREE::HAL::ExecutableVariantOp variantOp = getOperation();
+  ModuleOp innerModule = variantOp.getInnerModule();
+  if (!innerModule) {
+    return;
+  }
+
+  // Collect all dispatch_config ops.
+  SmallVector<IREE::Codegen::DispatchConfigOp> configOps;
+  innerModule->walk(
+      [&](IREE::Codegen::DispatchConfigOp op) { configOps.push_back(op); });
+  if (configOps.empty()) {
+    return;
+  }
+
+  // Build a map from export name to export op.
+  DenseMap<StringRef, IREE::HAL::ExecutableExportOp> exportMap;
+  for (auto exportOp : variantOp.getExportOps()) {
+    exportMap[exportOp.getSymName()] = exportOp;
+  }
+
+  for (IREE::Codegen::DispatchConfigOp configOp : configOps) {
+    StringRef funcRef = configOp.getFunctionRef();
+
+    if (!exportMap.contains(funcRef)) {
+      // No export for this function (e.g. a helper that is not an entry
+      // point).  Erase the dispatch_config and move on.
+      configOp.erase();
+      continue;
+    }
+    IREE::HAL::ExecutableExportOp exportOp = exportMap[funcRef];
+
+    // Replace the export count region body with the dispatch_config body.
+    // Map dispatch_config block args to export block args (offset by 1
+    // for !hal.device at position 0).
+    Region &countRegion = exportOp.getWorkgroupCount();
+    OpBuilder builder(&getContext());
+
+    if (!countRegion.empty()) {
+      Block &configBlock = configOp.getBody().front();
+      Block *exportBlock = exportOp.getWorkgroupCountBody();
+      unsigned configArity = configBlock.getNumArguments();
+      unsigned exportArity = exportBlock->getNumArguments();
+      // Export count region has !hal.device as block arg 0, then workloads.
+      if (configArity + 1 > exportArity) {
+        configOp.emitError("workload arity mismatch: dispatch_config has ")
+            << configArity << " args but export count region has "
+            << exportArity << " (expected >= " << configArity + 1
+            << " = config args + !hal.device)";
+        return signalPassFailure();
+      }
+      exportBlock->clear();
+      IRMapping mapping;
+      for (unsigned i = 0; i < configArity; ++i) {
+        mapping.map(configBlock.getArgument(i),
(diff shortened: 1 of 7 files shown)


hanhanW:

- https://github.com/iree-org/iree/pull/23894

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
```diff
@@ -0,0 +1,121 @@
...
+            << " = config args + !hal.device)";
+        return signalPassFailure();
+      }
+      exportBlock->clear();
+      IRMapping mapping;
+      for (unsigned i = 0; i < configArity; ++i) {
+        mapping.map(configBlock.getArgument(i),
+                    exportBlock->getArgument(i + 1));
+      }
+      builder.setInsertionPointToEnd(exportBlock);
```
MaheshRavishankar: Same as earlier PR. Use `cloneRegionBefore` with the mapper?

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
```diff
@@ -0,0 +1,121 @@
...
+      auto yieldOp = cast<IREE::Codegen::YieldOp>(configBlock.getTerminator());
+      auto returnValues =
+          llvm::map_to_vector(yieldOp.getOperands(), [&](Value v) {
+            return mapping.lookupOrDefault(v);
+          });
+      IREE::HAL::ReturnOp::create(builder, yieldOp.getLoc(), returnValues);
+    }
+
+    // Set workgroup_size and subgroup_size on the export.
+    auto wgSize = configOp.getWorkgroupSize();
```
qedawkins: nit: spell out type

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
```diff
@@ -0,0 +1,121 @@
...
+  IREE::HAL::ExecutableVariantOp variantOp = getOperation();
+  ModuleOp innerModule = variantOp.getInnerModule();
+  if (!innerModule) {
+    return;
+  }
+
+  // Collect all dispatch_config ops.
+  SmallVector<IREE::Codegen::DispatchConfigOp> configOps;
+  innerModule->walk(
+      [&](IREE::Codegen::DispatchConfigOp op) { configOps.push_back(op); });
```
qedawkins: I don't think we need a walk here. Symbol tables (i.e. the variant's inner module) require all symbol definitions to be defined within the direct body of the module. As a result we can just look for symbol definitions within the direct body rather than walking the whole thing. It's true that this op isn't directly a symbol itself per-se, but IMO we can treat it like a symbol, kind of like an add-on to the function.

qedawkins (approved):
LGTM

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
hanhanW: I think `innerModule.getOps<IREE::Codegen::DispatchConfigOp>())` is better and cleane, because we don't need nested walk.

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
hanhanW: I'm feeling that `cloneRegionBefore` is not a good fit here, because it creates a new block. We'll need the mapping between block arguments anyway, so `builder.clone` looks better to me.

We can clear the region and use `cloneRegionBefore`, and insert `!hal.device` as arg 0 though, but it seems that we don't respect the block argument.

EDIT: if we go with pure-clone, we can just clear the region and use `takeBody`.

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
```diff
@@ -0,0 +1,100 @@
...
+      Block *exportBlock = exportOp.getWorkgroupCountBody();
+      TypeRange configArgTypes = configBlock.getArgumentTypes();
+      TypeRange exportArgTypes = exportBlock->getArgumentTypes();
+      if (configArgTypes != exportArgTypes) {
+        configOp.emitError("block argument mismatch: dispatch_config has (")
+            << configArgTypes << ") but export count region has ("
+            << exportArgTypes << ")";
+        return signalPassFailure();
+      }
+      countRegion.takeBody(configOp.getBody());
```
MaheshRavishankar: Could we use builder interfaces here too? I have seen it way too often with subtle bugs being introduced when we dont consistently use the builder interface. It is just better to use those instead of using the direct op based methods during rewrites.

compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp
MaheshRavishankar: I didnt see this comment. These kind of things  introduce very subtle bugs that point back to not going through builder interface. It then will fall upon some one later to track down. I think it is good hygiene to just keep things using builder methods as much as we can.

hanhanW: I'm not aware of such issue in this context; it is actually pointless in this kind of rewrite. We could use `cloneRegionInto`, but it is pointless because listener is `nullptr` and there are no drivers at all to notify. It basically does nothing even if we use the method. https://github.com/llvm/llvm-project/blob/0d3e51479228b637f66a1db8727eadc2e7367be7/mlir/lib/IR/Builders.cpp#L593-L599

The `mapping` is null, so the later loop is also nop when `listener` is set.

```cpp
void OpBuilder::cloneRegionBefore(Region &region, Region &parent,
                                  Region::iterator before, IRMapping &mapping) {
  region.cloneInto(&parent, before, mapping);

  // Fast path: If no listener is attached, there is no more work to do.
  if (!listener)
    return;

  // Notify about op/block insertion.
  for (auto it = mapping.lookup(&region.front())->getIterator(); it != before;
       ++it) {
    listener->notifyBlockInserted(&*it, /*previous=*/nullptr,
                                  /*previousIt=*/{});
    it->walk<WalkOrder::PreOrder>([&](Operation *walkedOp) {
      listener->notifyOperationInserted(walkedOp, /*previous=*/{});
      notifyBlockInsertions(walkedOp, listener);
    });
  }
}
```

I can use this method if you insist, but I'd like to point out that this does not cause any issue from my understanding. There are no changes in the region, because the region is taken over without any modification.

I think listener is important when you use upstream drivers or maybe in transform ops, but it is meaningless if those drivers are not involved.

MaheshRavishankar: You are indexing too much on the current state of things. These change silently/behind the scenes, and not using the builder causes issues when things change upstream. Forget the mapping if you dont need it (cause you are just cloning the entire region). Bit todays state is that the listener does not provide any hooks for anything to do here, but it very easily could be extended. This is the pattern I typically see

1. The method her gets moved into a utility that can be called from other passes/transformations
2. Then the code upstream is changed to use the listener for some use cases.

Then some calls this utility method using a builder (with an attached listener) and some transformation infra fails cause there was some notification that did not connect. 

All the bugs we have hit of this sort is because we hadnt used the builder to update some value, and those are the bugs that are hardest to triage cause they just segfault somewhere else. The expected hygiene from downstream users is to always use the builder for IR modifications. Mixing builder and non-builder methods is a footgun someone eventually going to step on.

So yeah, I kind of do insist :P

hanhanW: > You are indexing too much on the current state of things.

My theory is that there are no ops got modified except the `export op`. The state is the same. For whatever reason that it become a pattern (which is def. wrong because we are not looking for fixed-point convergence), the code will be updated correctly unless they create the export op themselves -- it becomes legal if that's the case. Touching the same existing op is fundamentally not correct and it's nondeterministic , if multiple patterns can rewrite it in different form.

> The method her gets moved into a utility that can be called from other passes/transformations

I may have high bar for people who made the move, as they should understand what they are moving when they make change. But I understand that not everyone "read" the code carefully.

Anyway, that's my theory and understanding, and I don't really care the actual implementation because I have the context. So I'm making the change. Thanks for the discussion!

hanhanW:
Landing the PR because @MaheshRavishankar's main concern is addressed.

PR #23898, opened 2026-03-20
[CI] Revert Bazel's cache

amd-eochoalo:
Revert Bazel's cache PRs. The garbage collection flag appears to be modifying the content after the build is finished, even though there is more space than needed. Turning off the server does not fix the issue. Turning off garbage collection is not a good idea since LLVM integrates happen daily and one can imagine the cache increasing every day. Let's revert this for now.


amd-eochoalo:
@kuhar updated the description.

kuhar (approved):

PR #23899, opened 2026-03-20
[Codegen] Plumb through dispatch_config for workgroup count computation.

hanhanW:
Change ReconcileTranslationInfo and ResolveWorkgroupCountHints to run on ModuleOp and operate on dispatch_config ops. Wire CreateDispatchConfig and PropagateDispatchConfig into all backend pipelines.


hanhanW:

- https://github.com/iree-org/iree/pull/23894
- https://github.com/iree-org/iree/pull/23895

I'll send the cleanup for tests in a follow-up.

compiler/src/iree/compiler/Codegen/Common/TransformExtensions/CommonExtensions.cpp
```diff
@@ -673,11 +674,42 @@ transform_dialect::PopulateWorkgroupCountRegionUsingNumThreadsSliceOp::
   }
 
   auto funcOp = forAllOp->getParentOfType<mlir::FunctionOpInterface>();
-  if (failed(
-          lowerWorkgroupCountFromSliceOp(rewriter, funcOp, workgroupCount))) {
+
+  // Resolve the workgroup_count_from_slice in the dispatch_config op.
```
MaheshRavishankar: Can we just change the implementation of `lowerWorkgroupCountrFromSliceOp` to do this instead of removing that and inlining it?

hanhanW: It only has a single user, so it is inlined. I restored the implementation to touch on dispatch_config op now,

MaheshRavishankar (approved):

PR #23903, opened 2026-03-23
[GPU] Add in-tree e2e fixed intrinisic tests with TileAndFuse

nirvedhmeshram:
Currently we only have VectorDistribute tests so this adds TileAndFuse tests in addition to those.

### DIFF
--- a/tests/e2e/matmul/CMakeLists.txt
+++ b/tests/e2e/matmul/CMakeLists.txt
@@ -1655,6 +1655,34 @@ iree_generated_e2e_runner_test(
 
 iree_generated_e2e_runner_test(
   NAME
+    e2e_matmul_${_CDNA_ARCH}_tileandfusemfma_f16
+  TEST_TYPE
+    matmul
+  GENERATOR
+    "generate_e2e_matmul_tests.py"
+  GENERATOR_ARGS
+    "--lhs_rhs_type=f16"
+    "--acc_type=f32"
+    "--shapes=easy_large_static"
+    "--compilation_info=LLVMGPUTileAndFuseMFMA"
+  TEST_RUNNER
+    iree_tools_testing_e2e_iree-e2e-matmul-test
+  TARGET_BACKENDS
+    "rocm"
+  DRIVERS
+    "hip"
+  COMPILER_FLAGS
+    ${IREE_HIP_TEST_COMPILER_FLAGS}
+  LABELS
+    "noasan"
+    "nomsan"
+    "notsan"
+    "noubsan"
+    "requires-gpu-${_CDNA_ARCH}"
+)
+
+iree_generated_e2e_runner_test(
+  NAME
     e2e_matmul_${_CDNA_ARCH}_vecdistmfma_f32
   TEST_TYPE
     matmul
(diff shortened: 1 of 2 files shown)


nirvedhmeshram:
~Looks like the CUDA tests are also using this and they dont work with TileANDFuse as you would expect~, I will convert to draft while I investigate.

Groverkss (changes requested):
why not both

nirvedhmeshram:
> why not both

I am guessing you mean why not do TileAndFuse for the CUDA tests too? Yeah I was thinking about it but I havent ever tried running these tests so didnt touch them. I can quick try it and see if the CI is happy.

Groverkss:
> > why not both
> 
> I am guessing you mean why not do TileAndFuse for the CUDA tests too? Yeah I was thinking about it but I havent ever tried running these tests so didnt touch them. I can quick try it and see if the CI is happy.

Sorry, I mean why not test both VectorDistribute and TileAndFuse. Why "change" and not "add"

nirvedhmeshram:
oh wait maybe you mean why not have both VectorDistribute and TileAndFuse? I didnt see the value in having both as its just twice the runtime, the way I see it, the point of these tests is to mostly make sure the intrinsics are working e2e so might as well track with default pipeline. WDYT?

nirvedhmeshram:
@Groverkss This would add 18 more tests to my count, that seems a lot to have, perhaps this is the reason we dont have both in the first place, CUDA has fewer SKUs and hence fewer tests and therefore we have both TileAndFuse and VectorDistribute in that one.

bjacob:
Cost is a worthy argument to consider in E2E tests, but if that's the argument, then you could consider partial coverage with either pipeline.

nirvedhmeshram:
My real motivation was to simplify this testing before adding some new block intrinsic GEMM tests that I will check e2e for TileAndFuse. I wont be testing how they work for VectorDistribute as thats not in my current scope at least, so I can just add those to TileAndFuse and the existing test can stay with VectorDistribute, as TileAndFuse is already tested by default in the other tests even though the current intrinsic specific ones use VectorDistribute.

Groverkss:
> My real motivation was to simplify this testing before adding some new block intrinsic GEMM tests that I will check e2e for TileAndFuse. I wont be testing how they work for VectorDistribute as thats not in my current scope at least, so I can just add those to TileAndFuse and the existing test can stay with VectorDistribute, as TileAndFuse is already tested by default in the other tests even though the current intrinsic specific ones use VectorDistribute.

It's a bit weird that we aren't handeling block intrinsics in VectorDistribute. I'm not sure if I agree with this. Both pipelines should be handeling this. We shouldn't be plumbing these things partially. There is nothing special that TileAndFuse is doing that VectorDistribute shouldn't be able to handle. VectorDistribute uses inner_tiled operation as well.

bjacob:
I recall we had a similar debate a few months ago and the conclusion then was that we do want, like @Groverkss says, to keep fully supporting both pipelines for matmuls, regardless of what is the current default for one particular root op. I believe that means: err on the side of testing both pipelines here even if that means a higher cost of e2e tests. After all, the cost bump here is small in the grand scheme of things, and usefully buys time to allow the best pipelines to emerge :-)

nirvedhmeshram:
Oh I see, I didnt realize that was a choice we made, yeah I can look into supporting them with VectorDistribute as well then.

nirvedhmeshram:
@bjacob @Groverkss PTAL, I have added TileAndFuse tests and also retained the VectorDistribute tests.

Groverkss (approved):
Looks fine to me, can you update the pr description.

PR #23904, opened 2026-03-23
[Codegen] Skip bf16 expansion for targets with native conversion

Muzammiluddin-Syed-ECE:
Don't expand arith.truncf to integer bit manipulation for bf16 dtype. On targets like gfx950 and gfx12+, which have v_cvt_pk_bf16_f32 for hardware f32→bf16 conversion, this expansion costs VGPRs. On targets that don't have an explicit intrinsic to perform this conversion, llvm backend will handle the conversion for us. In either case this expansion is unnecessary.

On gfx950, This frees 12 VGPRs in scaled gemm workloads across 50 shapes (profiled with rocprofv3).

Made-with: Cursor

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
@@ -332,7 +332,6 @@ struct ConvertToROCDLPass final
       vector::populateVectorTransposeLoweringPatterns(
           patterns, options.vectorTransposeLowering);
       vector::populateVectorTransferLoweringPatterns(patterns);
-      arith::populateExpandBFloat16Patterns(patterns);
       if (failed(applyPatternsGreedily(m, std::move(patterns), config))) {
         return signalPassFailure();
       }
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/test/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/test/BUILD.bazel
@@ -36,6 +36,7 @@ iree_lit_test_suite(
             "convert_to_nvvm.mlir",
             "convert_to_rocdl.mlir",
             "convert_to_rocdl_gfx1201.mlir",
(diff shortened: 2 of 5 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp
```diff
@@ -332,7 +332,11 @@ struct ConvertToROCDLPass final
       vector::populateVectorTransposeLoweringPatterns(
           patterns, options.vectorTransposeLowering);
       vector::populateVectorTransferLoweringPatterns(patterns);
-      arith::populateExpandBFloat16Patterns(patterns);
+      // gfx950 and gfx12+ have native bf16 conversion (v_cvt_pk_bf16_f32).
```
krzysz00: Can we check what happens if we just drop this call entirely? Last I checked, this'll rewrite to a software version of that instruction over in LLVM on pre-gfx942 targets (though you'll want to check that)

krzysz00 (commented):
I think we should check if we can do simpler than this

Muzammiluddin-Syed-ECE:
so no compilation failures on older hw but instead a correctness issue, thats surprising. Also an issue on MI355 with data tiled microkernels.

krzysz00:
Huh, yeah, that is surprising - and probably interesting to look at

krzysz00:
... the MI-355 correctness issue is in an fp16 test that shouldn't care about this unless those are also bf16 tests in disguise? And the MI-300 thing seems to be machine flakiness.

I'd try rebasing

Muzammiluddin-Syed-ECE:
> ... the MI-355 correctness issue is in an fp16 test that shouldn't care about this unless those are also bf16 tests in disguise? And the MI-300 thing seems to be machine flakiness.
> 
> I'd try rebasing

worked!

krzysz00 (approved):
Approved pending that you fix the commit message to reflect the new nature of the PR / note that LLVM performs the expansion for old targets

PR #23907, opened 2026-03-23
[DMA] Support DMA in convolution lowering

lialan:
This patch implements the lowering to DMA from the `iree_linalg_ext.im2col` op for unpadded scenarios.

* `Im2colOp` adds `Im2colDecomposeModeAttr` enum and `decompose_mode` optional attribute.
* Adding `DecomposeIm2col` to lower to DMA async copying


compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -846,6 +780,393 @@ struct ConvertGatherToCoalescedDMA
...
+/// For batch position i:
+///   (b, m) = delinearize(i, [batch_tile, M_tile])
+///   (oh, ow, ...) = delinearize(m_off + m, output_sizes_M)
+///   (kh, kw, ...) = delinearize(k_off / C, window_sizes)
+///   spatial[j] = m_coord[j] * stride[j] + window[j] * dilation[j]
+///   n = batch_off + b
+///   lin = n * dim[0] * dim[1] * ... + spatial[0] * dim[1] * ... + ...
+static Value buildIm2colIndexTensor(PatternRewriter &rewriter, Location loc,
+                                    IREE::LinalgExt::Im2colOp im2colOp,
+                                    int64_t batchSize) {
```
jerryyin: I think @Max191 knows more about this so he can probably give better feedback than I do. But on the high level, I think having `GPUConvertToCoalescedDMA` own this logic looks to be the wrong abstraction to me. 

All this index computation logic looks familiar to me, so two points:
 1) The convert to DMA pass shouldn't own this, im2col decomposition should
 2) The common logic between im2col decomposition (existing, stream copy mode) should be factored out with the im2col decomposition (new, async copy mode) so majority of the work is reused, not reimplemented
 3) conver to DMA pass should be trivially just let the im2col op decompose itself. 

This way the two components are strictly decoupled from each other.

To be concrete, I am referring to this function:
https://github.com/iree-org/iree/blob/007b1eef6f31c490666e00130e55c3e3e871dc9d/compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp#L732

All this existing logic should be moved as a different mode into that existing decompose function. And then entire step 2 logic can be shared between stream copy mode implementation and async copy mode:

https://github.com/iree-org/iree/blob/007b1eef6f31c490666e00130e55c3e3e871dc9d/compiler/src/iree/compiler/Dialect/LinalgExt/IR/AggregatedOpInterfaceImpl.cpp#L791

Then all async copy differs would be to populate `GatherOp` instead.

lialan: Love the feedback, and it makes sense. So here are changes:
* Decomposition ownership moved to `Im2colOp`.
* `GPUConvertToCoalescedDMA` trivially calls the new `im2colOp.decomposeOperationAsyncCopy` function to decompose it self.

lialan:
WIP padded im2col support: https://github.com/iree-org/iree/pull/24142

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -636,50 +641,6 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
         return failure();
       }
     }
-
-  } else if constexpr (std::is_same_v<OpTy, IREE::LinalgExt::GatherOp>) {
```
Yu-Zhewen, line 640: Optional: but does this mean we can de-template the whole `createDMAInForall` function

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
Yu-Zhewen, line 159:
```suggestion
/// Traces through extract_slice, pad, collapse_shape ops to find the originating op.
```

Yu-Zhewen (approved):
LGTM, only two Nits

PR #23908, opened 2026-03-23
[codegen] Update sharktank models tuning spec to use #iree_gpu.pipeline<...> attr

efric:
These tuning specs were not updated in https://github.com/iree-org/iree/commit/f69c15c35d030b4359b1eb74836a83a6aeb136be so we fell back to default heuristics where relevant. This resulted in sharktank tests failing to pass. Fix by updating these to use the new `#iree_gpu.pipeline<VectorDistribute>` attr instead of `LLVMGPUVectorDistribute` enum.


MaheshRavishankar (approved):

bangtianliu (approved):
Thanks!

PR #23911, opened 2026-03-23
[GPU] Fix WMMA col_major LHS/RHS swap in createMmaOp

yzhang93:
The `col_major` flag on MMAAttr triggers an LHS/RHS operand swap in createMmaOp() to produce a column-major result. This swap was implemented for MFMA but missing for WMMA, so col_major = true had no effect on any WMMA intrinsic.

Add the swap for the `is_AMD_WMMA` branch with tests for `WMMAR3` and `WMMAR4`. Benchmarking standalone matmuls on RX 9070 XT (gfx1201) showed no measurable performance difference between row-major and col-major. Further testing on attention and other workloads is still needed to evaluate the full performance impact.

This PR is to solve https://github.com/nod-ai/amd-shark-ai/issues/2872.


krzysz00 (approved):
LGTM, thanks!

PR #23915, opened 2026-03-24
Add byte_offset to hal.tensor.import for subspan alignment

stellaraccident:
Adds an optional byte_offset operand to HAL_TensorImportOp so that non-zero offsets into parent buffers survive into the stream dialect. The HAL-to-Stream conversion creates a stream.resource.subview at the offset, making it visible to AnnotateDispatchArguments which computes correct alignment via GCD (e.g. gcd(64, 16) = 16).

Syntax: hal.tensor.import %bv offset(%off) : !hal.buffer_view -> tensor<...>

### DIFF
--- a/compiler/src/iree/compiler/Bindings/TFLite/Transforms/WrapEntryPoints.cpp
+++ b/compiler/src/iree/compiler/Bindings/TFLite/Transforms/WrapEntryPoints.cpp
@@ -221,6 +221,7 @@ private:
           recalculateBuilder, loc, inputValue.getType(), inputPlaceholder,
           inputValue.getType(), dynamicDims, /*consume=*/false,
           /*wait_fence=*/Value{},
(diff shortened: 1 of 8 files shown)


benvanik (approved):

benvanik:
This _may_ block optimizations if it were anywhere else in the IR, as we don't expect subviews on input - but I don't think it'll matter in practice. Another potential issue would be that alias analysis is probably not going to work through the subviews. But since these are on I/O, that's probably fine.

stellaraccident:
> This _may_ block optimizations if it were anywhere else in the IR, as we don't expect subviews on input - but I don't think it'll matter in practice. Another potential issue would be that alias analysis is probably not going to work through the subviews. But since these are on I/O, that's probably fine.

Yeah, that was my thought. I actually went deeper on this and then concluded that "IO is special [like it always has been]"

PR #23869, opened 2026-03-20
[DispatchCreation] Dynamic selection of split reduction target tile size for outer reductions 

YashDeshpande25:
This PR updates the logic to select the target tile size of split reduction for outer reductions proportional to the total reduction work to be carried out.  This should help select tiles dynamically based on the problem size to improve parallelism. The proposed decision formula is derived from experiments on Mi355.

PR #23916, opened 2026-03-24
[DispatchCreation] Dynamic selection of split reduction target tile size for outer reductions with fix

bangtianliu:
Reland PR https://github.com/iree-org/iree/pull/23869 with the fix, which skips outer reduction split entirely when `totalReductionWork < splitReductionTargetSize`

ci-extra: test_torch


YashDeshpande25 (approved):

yzhang93 (approved):
LGTM. Thanks!

Issue #23924, opened 2026-03-25
Make --iree-stream-resource-max-allocation-size also limit size of allocations for constants and result reservations


schuermans-roofline:
### Request description

Currently, the `--iree-stream-resource-max-allocation-size` option to `iree-compile` does not limit the size of the buffers created for constants and for result reservations in the `ScheduleAllocation` pass.

Certain custom accelerator hardwares have a limit on the size of device allocations imposed by either the hardware itself or by the software API provided by the vendor. Compiling larger models for those accelerators produces device buffers bigger than the limit, even when `--iree-stream-resource-max-allocation-size` is set.

In order to enable support for those devices, I'd like to suggest to extend the `ScheduleAllocation` pass to limit the size of the allocations for constants and for resource reservations to `--iree-stream-resource-max-allocation-size`.

### What component(s) does this issue relate to?

Compiler

### Additional context

I have a change ready for this and have opened a PR: https://github.com/iree-org/iree/pull/23925

PR #23925, opened 2026-03-25
feat(ScheduleAllocation): limit max allocation size in constants and result reservations

schuermans-roofline:
Extend the ScheduleAllocationPass to limit the size of the allocations for constants and result reservations to the maximum set by the --iree-stream-resource-max-allocation-size parameter when joining multiple ones into a single buffer.
If a single constant or result is already bigger than the maximum allocation size, it will still be allocated, but in a separate buffer.

This is required for custom devices that have a strict upper limit on the size of each device memory allocation. When compiling bigger models, the joining of constants and results produced buffers larger than this limit. This change makes those join operations respect this limit.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/Transforms/ScheduleAllocation.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/Transforms/ScheduleAllocation.cpp
@@ -1280,6 +1280,38 @@ struct ConstantAllocation {
   SmallVector<ConstantReservation> reservations;
 };
 
+// A bucket for collecting objects (e.g. AsyncConstantOps, ResourceReservation)
+// while tracking the total (static) size of the objects in the bucket.
+template <typename ObjT>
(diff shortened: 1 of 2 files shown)


schuermans-roofline:
This PR is about https://github.com/iree-org/iree/issues/23924

compiler/src/iree/compiler/Dialect/Stream/Transforms/ScheduleAllocation.cpp
```diff
@@ -1280,6 +1280,38 @@ struct ConstantAllocation {
   SmallVector<ConstantReservation> reservations;
 };
 
+// A bucket for collecting objects (e.g. AsyncConstantOps, ResourceReservation)
```
schuermans-roofline, line 1283: A lot has changed since I opened this PR. Some models that triggered this work here are now passing without this change. I need to double check if there are still models that need this change. I will resolve this comment once this is done.

Issue #23926, opened 2026-03-25
PackConstants can produce zero-sized buffers


schuermans-roofline:
### What happened?

Under certain circumstances, the `PackConstants` pass can produce buffers with zero size. Those zero-sized buffers can cause downstream passes or the runtime to fail.

### Steps to reproduce your issue

1.  Create a MLIR file `small_constant_for_reproducer.mlir` with the following contents:
```
util.func public @smallConstantForReproducer() -> (!stream.resource<constant>, !stream.timepoint) {
  %c256 = arith.constant 256 : index
  %0:2 = stream.resource.constants :
    !stream.resource<constant>{%c256} = dense<42> : tensor<64xi32>
    => !stream.timepoint
  util.return %0#0, %0#1 : !stream.resource<constant>, !stream.timepoint
}
```
2. Run
```
iree-compile --iree-hal-target-backends=llvm-cpu --iree-llvmcpu-target-triple=x86_64-linux-gnu --iree-llvmcpu-target-cpu=host --iree-stream-resource-max-allocation-size=128 -o small_constant_for_reproducer.vmfb -dump-compilation-phases-to=dump small_constant_for_reproducer.mlir
```
The important point is here that the limit (128) provided on the command line is less than the size of the first constant in the input MLIR.
3. Check first line of `dump/small_constant_for_reproducer.7.stream.mlir`:
```
#composite_of_0b = #util.composite<0xi8, []>
```

### What component(s) does this issue relate to?

Compiler

### Version information

7703cd6d1f85ff1674c79a1aa75fa85eaaeaba06

### Additional context

I have a suggested fix in a PR: https://github.com/iree-org/iree/pull/23927.

PR #23927, opened 2026-03-25
fix(PackConstants): don't start new storage buffer if current one is still empty

schuermans-roofline:
The PackConstants pass introduces a new buffer when a constant cannot be packed into the current buffer without exceeding the maximum size. However, if a constant is itself larger than the limit, it gets still added, but in a separate buffer.
This change fixes a bug in the packing logic when a constant bigger than the limit is encountered when the current buffer is still empty. In this case, the current buffer stayed empty and the big constant got added to a new buffer. The empty buffer caused errors of size 0 buffers not being supported in following passes or in the runtime. This change avoids adding a new buffer is the current one is empty and thus avoids these errors.


schuermans-roofline:
This PR is a suggested fix for https://github.com/iree-org/iree/pull/23926

Manewing (approved):

PR #23928, opened 2026-03-25
[Util] Add int narrowing pattern for arith.select

Max191:
Narrow arith.select through index_cast ops when both true/false operands are either index_cast from a narrower type or arith.constant. For example:

```mlir
  %t = arith.index_cast %a : i32 to index
  %f = arith.constant 0 : index
  %s = arith.select %cond, %t, %f : index
  ->
  %f_narrow = arith.constant 0 : i32
  %s_narrow = arith.select %cond, %a, %f_narrow : i32
  %s = arith.index_cast %s_narrow : i32 to index
```

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
+++ b/compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
@@ -319,6 +319,122 @@ struct NarrowVectorBroadcast : OpRewritePattern<vector::BroadcastOp> {
   }
 };
 
+/// Narrow an arith.select through index_cast ops. Each operand may be either
+/// an index_cast from a narrower type or an arith.constant (which can be
+/// rematerialized in the narrow type). At least one operand must be an
+/// index_cast to determine the target narrow type.
+///
+/// Rewrites e.g.:
+///   %t = arith.index_cast %a : i32 to index
+///   %f = arith.constant 0 : index
+///   %s = arith.select %cond, %t, %f : index
+/// to:
+///   %f_narrow = arith.constant 0 : i32
+///   %s_narrow = arith.select %cond, %a, %f_narrow : i32
+///   %s = arith.index_cast %s_narrow : i32 to index
+struct NarrowArithSelect : OpRewritePattern<arith::SelectOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(arith::SelectOp op,
+                                PatternRewriter &rewriter) const override {
+    // The select must operate on index type.
+    if (!isa<IndexType>(getElementTypeOrSelf(op.getResult().getType()))) {
+      return failure();
+    }
+
+    // Find the index_cast that determines the narrow type. Try true first,
+    // then false.
+    Operation *trueDef = op.getTrueValue().getDefiningOp();
+    Operation *falseDef = op.getFalseValue().getDefiningOp();
+    if (!trueDef || !falseDef) {
+      return failure();
+    }
+    Operation *castOp =
+        dyn_cast<arith::IndexCastOp>(trueDef);
+    if (!castOp) {
+      castOp = dyn_cast<arith::IndexCastUIOp>(trueDef);
+    }
+    if (!castOp) {
+      castOp = dyn_cast<arith::IndexCastOp>(falseDef);
+    }
+    if (!castOp) {
+      castOp = dyn_cast<arith::IndexCastUIOp>(falseDef);
+    }
+    if (!castOp) {
+      return failure();
+    }
+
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
```diff
@@ -319,6 +319,122 @@ struct NarrowVectorBroadcast : OpRewritePattern<vector::BroadcastOp> {
...
+      // Same kind of index_cast with the same narrow type.
+      if (defOp->getName() == castOp->getName()) {
+        return getElementTypeOrSelf(defOp->getOperand(0).getType()) ==
+               narrowElemTy;
+      }
+      // Constant with a value that fits in the narrow type.
+      auto constOp = dyn_cast<arith::ConstantOp>(defOp);
+      if (!constOp) {
+        return false;
+      }
```
Groverkss, line 386: Can we use ConstantLikeInterface maybe?

Max191: I don't think an interface exists for it, but I think I can use the ConstantLike trait. I'll update the implementation.

krzysz00: There are `m_Constant`-type matchers.

compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
```diff
@@ -319,6 +319,122 @@ struct NarrowVectorBroadcast : OpRewritePattern<vector::BroadcastOp> {
...
+    }
+
+    // Find the index_cast that determines the narrow type. Try true first,
+    // then false.
+    Operation *trueDef = op.getTrueValue().getDefiningOp();
+    Operation *falseDef = op.getFalseValue().getDefiningOp();
+    if (!trueDef || !falseDef) {
+      return failure();
+    }
+    Operation *castOp =
```
krzysz00, line 352: This could probably work on extsi/extui also, and I colud've sworn there's a nice helper for this.

Also, what you're looking for is `isa<IndexCastOp, IndexCastUIOp>(trueDef)` and similar for falseDef - if it `isa` one of these, you've found the `castOp`.

compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
```diff
@@ -319,6 +319,122 @@ struct NarrowVectorBroadcast : OpRewritePattern<vector::BroadcastOp> {
...
+    // Validate that each operand is either a matching index_cast or a
+    // constant that fits in the narrow type.
+    auto isValidOperand = [&](Operation *defOp) {
+      // Same kind of index_cast with the same narrow type.
+      if (defOp->getName() == castOp->getName()) {
+        return getElementTypeOrSelf(defOp->getOperand(0).getType()) ==
+               narrowElemTy;
+      }
+      // Constant with a value that fits in the narrow type.
+      auto constOp = dyn_cast<arith::ConstantOp>(defOp);
```
krzysz00, line 383: You could check for the range being constant here?

krzysz00: Yeah, I think the check is if the min == max of this value fits in the narrower type.

Also, I think there's code elsewhere around here that handles similar tests for for lops.

compiler/src/iree/compiler/Dialect/Util/Transforms/OptimizeIntArithmetic.cpp
```diff
@@ -319,6 +319,122 @@ struct NarrowVectorBroadcast : OpRewritePattern<vector::BroadcastOp> {
...
+      narrowResultType = VectorType::get(vecType.getShape(), narrowElemTy);
+    } else {
+      narrowResultType = narrowElemTy;
+    }
+
+    auto getNarrowOperand = [&](Operation *defOp) -> Value {
+      if (defOp->getName() == castOp->getName()) {
+        return defOp->getOperand(0);
+      }
+      auto intAttr = cast<IntegerAttr>(cast<arith::ConstantOp>(defOp).getValue());
```
krzysz00, line 412: I think for this case I'd either follow existing logic for materializing constants in a generic way or I'd just `createOrFold` the relevant truncation to get the new constant.

krzysz00 (commented):
I've got some implementation comments

Groverkss:
(my comment is drive-by, non need to wait for me)

krzysz00:
Wondering about PR state here

PR #23929, opened 2026-03-25
[Util] Add integer divisibility inference for arith.select

Max191:
The result divisibility of arith.select is the GCD of the true and false operands' divisibilities (the i1 condition is irrelevant).

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
+++ b/compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
@@ -498,3 +498,45 @@ util.func @delinearize_dynamic_basis(%arg0 : index, %arg1 : index) {
   %probe1 = "iree_unregistered.test_int_divisibility"(%0#1) : (index) -> index
   util.return
 }
+
+// -----
+
+// arith.select: divisibility is the GCD of true and false operands.
+// Both operands are divisible by 16, so the result is divisible by 16.
+// CHECK-LABEL: @select_divisibility_both_divisible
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Dialect/Util/Transforms/test/test_integer_divisibility_analysis.mlir
```diff
@@ -498,3 +498,45 @@ util.func @delinearize_dynamic_basis(%arg0 : index, %arg1 : index) {
...
+
+// arith.select with constant false value 0: divisibility of 0 is 0
+// (divides everything), so result takes the true operand's divisibility.
+// CHECK-LABEL: @select_divisibility_with_constant_zero
+util.func @select_divisibility_with_constant_zero(%arg0 : index, %cond : i1) {
+  %c0 = arith.constant 0 : index
+  %0 = util.assume.int %arg0<udiv = 8> : index
+  %1 = arith.select %cond, %0, %c0 : index
+  // CHECK: divisibility = "udiv = 8, sdiv = 8"
+  %probe = "iree_unregistered.test_int_divisibility"(%1) : (index) -> index
```
Groverkss, line 540: do we support vector operands? If so can we add a test?

Max191: I don't think vector tests add value here. The interface implementation just works with Value's, so there is no difference with vector/scalar/tensor/etc operands. The divisibility analysis infra is what's responsible for providing the divisibilities of the operands.

krzysz00 (approved):

Groverkss (approved):

PR #23930, opened 2026-03-25
[Codegen] Add pattern to hoist expand_shape out of forall

Max191:
Adds a pattern to collapse scf.forall destinations by hoisting expand_shape ops out to the forall's result. This enables further propagation by collapsing during the IGEMM transformation when we have pre-existing split-reduction loops.

The inverse pattern for hoisting collapse_shape and expanding the forall is moved to Codegen/Common/Transforms.cpp because some utilities from that pattern are reused in this one.


nirvedhmeshram (approved):
LGTM

krzysz00 (approved):
lgtm here too

PR #23933, opened 2026-03-26
Unify RISC-V CMake variables for RV32 and RV64

HanKuanChen:
Merge RISCV32_TEST_DEFAULT_LLVM_FLAGS and
RISCV64_TEST_DEFAULT_LLVM_FLAGS into a single
RISCV_TEST_DEFAULT_LLVM_FLAGS variable to simplify condition checks in iree_macros.cmake.

### DIFF
--- a/build_tools/cmake/iree_macros.cmake
+++ b/build_tools/cmake/iree_macros.cmake
@@ -681,20 +681,14 @@ function(iree_compile_flags_for_platform OUT_FLAGS IN_FLAGS)
     list(APPEND _FLAGS "--iree-llvmcpu-target-triple=${_TARGET_TRIPLE}")
   endif()
 
-  if(IREE_ARCH STREQUAL "riscv_64" AND
+  if((IREE_ARCH STREQUAL "riscv_64" OR
+      IREE_ARCH STREQUAL "riscv_32") AND
      CMAKE_SYSTEM_NAME STREQUAL "Linux" AND
      NOT IN_FLAGS MATCHES "iree-llvmcpu-target-triple")
-    # RV64 Linux crosscompile toolchain can support iree-compile with
-    # specific CPU flags. Add the llvm flags to support RV64 RVV codegen if
+    # RISC-V Linux crosscompile toolchain can support iree-compile with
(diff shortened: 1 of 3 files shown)


HanKuanChen:
ping

HanKuanChen:
ping

HanKuanChen:
ping @hanhanW @ScottTodd

ScottTodd:
(I haven't been an active contributor/maintainer to this repository for over a year now, pinging me won't help much)

hanhanW:
I think there is a disconnected point between CI and cmake files. It looks like you've been refactoring the cmake files (and thanks for that), but we never have riscv-32 job in IREE's CI. I wonder if SiFive folks (according to your profile) have CI jobs internally for such target?

If so, are we able to bring it or whatever riscv-32 job to public CI? Otherwise, I could blindly approve the PRs which makes me feel bad.

HanKuanChen:
Thanks for the feedback. This PR is part of my effort to separate RV32 and RV64 logic by moving configurations into toolchain files. The current mixed logic makes it harder to maintain RV64 support.

For context, internally at SiFive we only use RV64 (both Linux and bare-metal environments). I'm not aware of any active RV32 users in the IREE ecosystem, but I could be wrong.

This brings up a question: **Should we consider removing RV32 support if there's no CI coverage and no known active users?** Without CI, the RV32 code may already be broken, and it adds maintenance burden that affects RV64 development.

hanhanW:
> Thanks for the feedback. This PR is part of my effort to separate RV32 and RV64 logic by moving configurations into toolchain files. The current mixed logic makes it harder to maintain RV64 support.
> 
> For context, internally at SiFive we only use RV64 (both Linux and bare-metal environments). I'm not aware of any active RV32 users in the IREE ecosystem, but I could be wrong.
> 
> This brings up a question: **Should we consider removing RV32 support if there's no CI coverage and no known active users?** Without CI, the RV32 code may already be broken, and it adds maintenance burden that affects RV64 development.

I started a discussion on discord: https://discord.com/channels/689900678990135345/788499363739336716/1493649191044648962

It'd be good if you folks can chime in, thanks!

PR #23935, opened 2026-03-26
[Codegen] Masked vectorization of inner_tiled

sommerlukas:
So far, vectorization of `inner_tiled` did not support dynamic tensor shapes. This PR enables vectorization of `inner_tiled` with dynamic tensor shapes in the outer dimensions through masking for the case that vector sizes are provided.

The two main changes in this PR are:
- Add support for `inner_tiled`, `linalg.pack` and `linalg.unpack` operations to the vector tile size analysis. The vector tile sizes computed by this analysis can serve as vector sizes for the vectorization.
- Vectorization of `inner_tiled` with dynamic tensor sizes, assuming the provided vector sizes and using masked reads/writes to mask out OOB values.

Assisted-by: Claude Code

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_inferred.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_inferred.mlir
@@ -873,3 +873,45 @@ func.func @scaled_tensor_multi_mma(%arg0: tensor<3x5x1x32xf4E2M1FN>, %arg1: tens
 //       CHECK:   %[[MMA:.+]] = iree_codegen.inner_tiled ins(%[[LHS]], %[[RHS]], %[[LHS_SCALE]], %[[RHS_SCALE]]) outs(%[[ACC]])
 //  CHECK-SAME: : vector<3x5x1x32xf4E2M1FN>, vector<5x1x7x32xf8E4M3FN>, vector<3x5x1xf8E8M0FNU>, vector<5x7x1xf8E8M0FNU> into vector<3x7x4xf32>
 //       CHECK:   vector.transfer_write %[[MMA]], %arg4[%c0, %c0, %c0] {{.*}} : vector<3x7x4xf32>, tensor<3x7x4xf32>
+
+// -----
+
+// Masked vectorization of inner_tiled with dynamic outer dimensions.
+// Vector sizes come from the iree_codegen.vector_tile_sizes attribute.
+// The LHS has shape <?x?x4xf16> (outer dims dynamic), iteration space is
+// [i=2, j=5, k=3], so reads should be masked.
+#contraction_accesses = [
+ affine_map<(i, j, k) -> (i, k)>,
+ affine_map<(i, j, k) -> (k, j)>,
+ affine_map<(i, j, k) -> (i, j)>
+]
+func.func @masked_tensor_multi_mma(%lhs: tensor<?x?x4xf16>, %rhs: tensor<?x?x4xf16>, %acc: tensor<?x?x4xf32>) -> tensor<?x?x4xf32> {
+  %0 = iree_codegen.inner_tiled ins(%lhs, %rhs) outs(%acc) {
+    iree_codegen.vector_tile_sizes = array<i64: 2, 5, 3>,
+    indexing_maps = #contraction_accesses,
+    iterator_types = [#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>],
+    kind = #iree_gpu.mma_layout<MFMA_F32_16x16x16_F16>,
+    semantics = #iree_gpu.mma_semantics<distributed = true, opaque = false>
+  } : tensor<?x?x4xf16>, tensor<?x?x4xf16> into tensor<?x?x4xf32>
+  return %0 : tensor<?x?x4xf32>
+}
+
+// CHECK-LABEL: func @masked_tensor_multi_mma
+
+// With vectorSizes, reads use create_mask for dynamic dims.
+// LHS: outer (i=2, k=3) + inner (4) → vector<2x3x4xf16>
+// RHS: outer (k=3, j=5) + inner (4) → vector<3x5x4xf16>
+// ACC: outer (i=2, j=5) + inner (4) → vector<2x5x4xf32>
+//   CHECK-DAG:   %[[CSTF16:.+]] = arith.constant 0.000000e+00 : f16
+//   CHECK-DAG:   %[[CSTF32:.+]] = arith.constant 0.000000e+00 : f32
+//       CHECK:   %[[LHS_MASK:.+]] = vector.create_mask {{.*}} : vector<2x3x4xi1>
+//       CHECK:   %[[LHS:.+]] = vector.transfer_read %arg0{{.*}}, %[[CSTF16]], %[[LHS_MASK]]{{.*}} : tensor<?x?x4xf16>, vector<2x3x4xf16>
+//       CHECK:   %[[RHS_MASK:.+]] = vector.create_mask {{.*}} : vector<3x5x4xi1>
+//       CHECK:   %[[RHS:.+]] = vector.transfer_read %arg1{{.*}}, %[[CSTF16]], %[[RHS_MASK]]{{.*}} : tensor<?x?x4xf16>, vector<3x5x4xf16>
+//       CHECK:   %[[ACC_MASK:.+]] = vector.create_mask {{.*}} : vector<2x5x4xi1>
+//       CHECK:   %[[ACC:.+]] = vector.transfer_read %arg2{{.*}}, %[[CSTF32]], %[[ACC_MASK]]{{.*}} : tensor<?x?x4xf32>, vector<2x5x4xf32>
+//       CHECK:   %[[MMA:.+]] = iree_codegen.inner_tiled ins(%[[LHS]], %[[RHS]]) outs(%[[ACC]])
+//  CHECK-SAME:     : vector<2x3x4xf16>, vector<3x5x4xf16> into vector<2x5x4xf32>
+//       CHECK:   %[[WRITE_MASK:.+]] = vector.create_mask {{.*}} : vector<2x5x4xi1>
+//       CHECK:   vector.transfer_write %[[MMA]], %arg2{{.*}}, %[[WRITE_MASK]] {in_bounds = [false, false, true]}
+//  CHECK-SAME:     : vector<2x5x4xf32>, tensor<?x?x4xf32>
--- a/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
+++ b/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
@@ -20,6 +20,7 @@
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/UB/IR/UBOps.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/Dialect/Vector/Utils/VectorUtils.h"
 #include "mlir/IR/AffineMap.h"
 #include "mlir/IR/IRMapping.h"
 
@@ -1042,6 +1043,28 @@ struct PadOpVectorizationModel
   }
 };
 
+/// Compute the outer iteration space sizes for an InnerTiledOp.
+static SmallVector<int64_t>
+getStaticOuterLoopRanges(IREE::Codegen::InnerTiledOp tiledOp) {
+  SmallVector<AffineMap> indexingMaps = tiledOp.getIndexingMapsArray();
+  SmallVector<ShapedType> argTypes = tiledOp.getOperandShapedTypes();
+  int64_t numLoops = indexingMaps.front().getNumDims();
+  SmallVector<int64_t> staticSizes(numLoops, ShapedType::kDynamic);
+  for (auto [i, argType] : llvm::enumerate(argTypes)) {
+    AffineMap map = indexingMaps[i];
+    ArrayRef<int64_t> outerShape =
+        argType.getShape().take_front(map.getNumResults());
+    for (auto [mapResult, dimSize] :
+         llvm::zip_equal(map.getResults(), outerShape)) {
+      auto dimExpr = cast<AffineDimExpr>(mapResult);
+      if (!ShapedType::isDynamic(dimSize)) {
+        staticSizes[dimExpr.getPosition()] = dimSize;
+      }
+    }
+  }
+  return staticSizes;
+}
+
 /// External model for IREE::Codegen::InnerTiledOp. Reads tensor operands into
 /// vectors, creates a vector-semantic InnerTiledOp, and writes results back.
 struct InnerTiledOpVectorizationModel
@@ -1055,15 +1078,16 @@ struct InnerTiledOpVectorizationModel
     if (!tiledOp.hasTensorSemantics()) {
       return false;
     }
+    SmallVector<int64_t> loopRanges = getStaticOuterLoopRanges(tiledOp);
     // If vector sizes are provided (from tile size analysis or config),
-    // dynamic outer shapes are fine — they'll be masked during vectorization.
+    // dynamic outer shapes are fine - they'll be masked during vectorization.
+    // However, vector sizes must be >= the static outer dimension sizes.
     if (!vectorSizes.empty()) {
-      return true;
+      return succeeded(
+          vector::isValidMaskedInputVector(loopRanges, vectorSizes));
     }
-    // Without vector sizes, require static shapes.
-    SmallVector<ShapedType> argTypes = tiledOp.getOperandShapedTypes();
-    return llvm::all_of(argTypes,
-                        [](ShapedType st) { return st.hasStaticShape(); });
+    // Without vector sizes, require static outer shapes.
+    return !ShapedType::isDynamicShape(loopRanges);
   }
 
   FailureOr<SmallVector<Value>> vectorize(Operation *op, RewriterBase &rewriter,
@@ -1078,12 +1102,17 @@ struct InnerTiledOpVectorizationModel
     SmallVector<ShapedType> argTypes = tiledOp.getOperandShapedTypes();
     SmallVector<AffineMap> indexingMaps = tiledOp.getIndexingMapsArray();
 
-    // Determine whether we need masking: vectorSizes present and any operand
-    // has dynamic outer dimensions.
-    bool needsMasking =
-        !vectorSizes.empty() && llvm::any_of(argTypes, [](ShapedType st) {
-          return !st.hasStaticShape();
-        });
+    // If no vector sizes are provided, use static loop ranges and the inBounds
+    // attribute instead of masking.
+    bool needsMasking = true;
+    if (vectorSizes.empty()) {
+      SmallVector<int64_t> loopRanges = getStaticOuterLoopRanges(tiledOp);
+      if (ShapedType::isDynamicShape(loopRanges)) {
+        return rewriter.notifyMatchFailure(op, "unable to infer vector sizes");
+      }
+      vectorSizes = loopRanges;
+      needsMasking = false;
+    }
 
     // Construct the zero padding value for each operand. Ideally, we'd need the
     // InnerTile interface to return the padding value to use. If it is not
@@ -1102,8 +1131,8 @@ struct InnerTiledOpVectorizationModel
         readShapes.push_back(llvm::to_vector(argType.getShape()));
         continue;
       }
-      // In case we need masking, outer dimensions com from vector sizes via the
-      // indexing map, the inner dimensions are static.
+      // Outer dimensions come from vector sizes via the indexing map, inner
+      // dimensions are static.
       SmallVector<int64_t> readShape;
       AffineMap map = indexingMaps[i];
       for (AffineExpr expr : map.getResults()) {
@@ -1152,8 +1181,7 @@ struct InnerTiledOpVectorizationModel
 
       auto write = vector::TransferWriteOp::create(
           rewriter, loc, result, tensorAcc, indices, inBounds);
-
-      if (needsMasking && !tensorType.hasStaticShape()) {
+      if (llvm::is_contained(inBounds, false)) {
         auto vecType = cast<VectorType>(result.getType());
         auto maskType = vecType.cloneWith({}, rewriter.getI1Type());
         SmallVector<OpFoldResult> mixedSizes =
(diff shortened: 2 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -426,22 +629,73 @@ class MaterializeVectorTileSizesPass final
...
       }
-      assert(tileSizes.rank() == linalgOp.getNumLoops());
 
-      linalgOp->setAttr(
-          kVectorTileSizesAttrName,
-          DenseI64ArrayAttr::get(linalgOp->getContext(), tileSizes.getDims()));
+      // Pack/unpack: look up tile sizes in the unpacked domain and transform
+      // to the vectorization domain via mapPackSourceToDest.
+      // For pack, the source lattice holds unpacked tile sizes.
+      // For unpack, the result lattice holds unpacked tile sizes.
```
Groverkss: I think this needs a bit more context. What is the iteration space of pack/unpack?

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1081,29 +1095,75 @@ struct InnerTiledOpVectorizationModel
...
+
+    // Read each operand into a vector, with masking if needed.
+    SmallVector<Value> newOperands(tiledOp->getOperands());
+    for (auto [operand, readShape, padValue] :
+         llvm::zip_equal(newOperands, readShapes, padValues)) {
       operand = vector::createReadOrMaskedRead(
-          rewriter, loc, operand, type.getShape(), padValue,
-          /*useInBoundsInsteadOfMasking=*/true);
+          rewriter, loc, operand, readShape, padValue,
+          /*useInBoundsInsteadOfMasking=*/!needsMasking);
```
Groverkss, line 1132: Where is the padValue coming from btw? Using a padValue of zero is not always correct. If needed, we should plumb padding value through the intrinsic first, otherwise we will special case for matmul (my understanding is inner_tiled is more general).

Groverkss:
I think it's a bit unfortunate that we aren't using linalg.pack to seed tile sizes as well. Are we still taking the max of possible tile sizes in the analysis? Ideally, linalg.pack/linalg.unpack also give information that you should use some specific tile sizes and the padding can be done just via selects.

sommerlukas:
> I think it's a bit unfortunate that we aren't using linalg.pack to seed tile sizes as well. Are we still taking the max of possible tile sizes in the analysis? 

No, we aren't taking the maximum of multiple values, we're only tracking a single tile size value per dimension and go to an overdefined/top state if different tile sizes meet/join.

I had considered seeding the vector tile size analysis from `linalg.pack`, but they don't really provide much additional information. In our masked/dynamic case, they look something like this:
```mlir
%pack_7 = linalg.pack %35 padding_value(%cst : f16) inner_dims_pos = [2, 1] inner_tiles = [32, 16] into %38 : 
  tensor<1x?x64xf16> -> tensor<1x?x2x32x16xf16>
```

While the `inner_tiles` tell us that the tensor is at least `16x32`, they don't tell us the overall size. On the other hand, the `to_layout` operation on the operand tells us the exact size of the overall tensor (`1x4x2x16x32` in this case).

In this case, seeding the analysis from `linalg.pack` would be actively harmful. If we backpropagated the `inner_tiles` sizes `16x32` from the `linalg.pack` to its operands, it would meet with the `64x64` information from the `to_layout` information, leading to the overdefined state and we would lose the tile size altogether.

> Ideally, linalg.pack/linalg.unpack also give information that you should use some specific tile sizes and the padding can be done just via selects.

The vectorization treats those operations individually, so we have to insert masked `transfer_read/write` for each of the operations. 
```mlir
// Vectorization of linalg.pack
%64 = vector.create_mask %c1, %35, %c64 : vector<1x64x64xi1>
%65 = vector.transfer_read %57[%c0, %c0, %c0], %cst_4, %64 ...
%66 = vector.shape_cast %65 : vector<1x64x64xf16> to vector<1x4x16x2x32xf16>
%67 = vector.transpose %66, [0, 1, 3, 4, 2] : vector<1x4x16x2x32xf16> to vector<1x4x2x32x16xf16>
%68 = vector.create_mask %c1, %62, %c2, %c32, %c16 : vector<1x4x2x32x16xi1>
%69 = vector.transfer_write %67, %63[%c0, %c0, %c0, %c0, %c0], %68 ...
// Vectorization of inner_tiled
%78 = vector.create_mask %c1, %62, %c2, %c32, %c16 : vector<1x4x2x32x16xi1>
%79 = vector.transfer_read %69[%c0, %c0, %c0, %c0, %c0], %cst_4, %78 ...
...
%82 = iree_codegen.inner_tiled ins(%25, %79) outs(%81)
```

The `OptimizeTensorInsertExtractSlicesPass` however has patterns to eliminate the `transfer_read -> transfer_write` pairs and to materialize as `arith.select`. After running that pass, we end up with this IR:
```mlir
%cst_3 = arith.constant dense<0.000000e+00> : vector<1x4x2x32x16xf16>
%cst_4 = arith.constant dense<0.000000e+00> : vector<1x64x64xf16>
...
// linalg.pack
%37 = arith.select %28, %34, %cst_4 : vector<1x64x64xi1>, vector<1x64x64xf16>
%38 = vector.shape_cast %37 : vector<1x64x64xf16> to vector<1x4x16x2x32xf16>
%39 = vector.transpose %38, [0, 1, 3, 4, 2] : vector<1x4x16x2x32xf16> to vector<1x4x2x32x16xf16>
// inner_tiled
%40 = vector.create_mask %c1, %36, %c2, %c32, %c16 : vector<1x4x2x32x16xi1>
%45 = arith.select %40, %39, %cst_3 : vector<1x4x2x32x16xi1>, vector<1x4x2x32x16xf16>
%47 = iree_codegen.inner_tiled ins(%19, %45) outs(%46) ...
```

If we wanted to vectorize `inner_tiled` together with the packing/unpacking, we would need to change the pattern to handle `linalg.pack` on its operands (and `linalg.unpack` on the result) or switch GenericVectorization to be a dialect conversion, where we could use the remapped operands provided by the `ConversionPattern`.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: There is a long-standing comment about this where `padValue` is constructed: https://github.com/iree-org/iree/blob/ffcbc3b6e91e115a9dcdcdf53a548780ccb19ee3/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L1074-L1077

I agree with you (and the comment) that this isn't ideal but considered it out-of-scope for this PR to change this. Happy to address this in a follow up, though.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: I've simplified the code and clarified the comment.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1055,6 +1055,12 @@ struct InnerTiledOpVectorizationModel
     if (!tiledOp.hasTensorSemantics()) {
       return false;
     }
+    // If vector sizes are provided (from tile size analysis or config),
+    // dynamic outer shapes are fine — they'll be masked during vectorization.
+    if (!vectorSizes.empty()) {
+      return true;
+    }
```
hanhanW, line 1067: Similar to the upstream method, you should check the vector sizes are greator than dim sizes if they are static.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1070,6 +1076,14 @@ struct InnerTiledOpVectorizationModel
...
 
     SmallVector<ShapedType> argTypes = tiledOp.getOperandShapedTypes();
+    SmallVector<AffineMap> indexingMaps = tiledOp.getIndexingMapsArray();
+
+    // Determine whether we need masking: vectorSizes present and any operand
+    // has dynamic outer dimensions.
+    bool needsMasking =
+        !vectorSizes.empty() && llvm::any_of(argTypes, [](ShapedType st) {
+          return !st.hasStaticShape();
+        });
```
hanhanW: I thought needMasking should just check whether `vectorSizes` is empty or not. Static shape should be maskable. Think that you have `3x5` tensor, and you may want to mask it with `4x8`. The `vectorSizes` control the behavior, IMO.

IMO, you should construct the `vectorSizes` when it is empty. Then the read/write ops get the shape from `vectorSizes`. It is how it's done in upstream when we worked on vectorization. E.g.,

https://github.com/llvm/llvm-project/blob/00aebbff71ff4e348538708064ba2e033ccd6b2a/mlir/lib/Dialect/Linalg/Transforms/Vectorization.cpp#L1883-L1893

hanhanW (changes requested):
It looks like two PRs? One for inner_tiled vectorization improvement and the other is pack/unpack vector size propagation? (I haven't reviewed the latter part yet)

sommerlukas:
> It looks like two PRs? One for inner_tiled vectorization improvement and the other is pack/unpack vector size propagation? (I haven't reviewed the latter part yet)

The two changes belong together. The `to_layout` that provides the seed for the tile size analysis is applied to the operands/result of pack/unpack, so we need the vector tile size propagation for pack/unpack to determine the vector tile sizes used for vectorization of `inner_tiled`.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: I'm now using the upstream `isValidMaskedInputVector` to check that.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: Thanks, I've made the code similar to upstream, so that vector tile sizes greater than the static shapes also trigger masking.

Groverkss:
> > It looks like two PRs? One for inner_tiled vectorization improvement and the other is pack/unpack vector size propagation? (I haven't reviewed the latter part yet)
> 
> The two changes belong together. The `to_layout` that provides the seed for the tile size analysis is applied to the operands/result of pack/unpack, so we need the vector tile size propagation for pack/unpack to determine the vector tile sizes used for vectorization of `inner_tiled`.

I think you can still split out the vectorization of inner_tiled as a seperate pr and tile size analysis stuff as a seperate pr.

sommerlukas:
> > > It looks like two PRs? One for inner_tiled vectorization improvement and the other is pack/unpack vector size propagation? (I haven't reviewed the latter part yet)
> > 
> > 
> > The two changes belong together. The `to_layout` that provides the seed for the tile size analysis is applied to the operands/result of pack/unpack, so we need the vector tile size propagation for pack/unpack to determine the vector tile sizes used for vectorization of `inner_tiled`.
> 
> I think you can still split out the vectorization of inner_tiled as a seperate pr and tile size analysis stuff as a seperate pr.



> > > It looks like two PRs? One for inner_tiled vectorization improvement and the other is pack/unpack vector size propagation? (I haven't reviewed the latter part yet)
> > 
> > 
> > The two changes belong together. The `to_layout` that provides the seed for the tile size analysis is applied to the operands/result of pack/unpack, so we need the vector tile size propagation for pack/unpack to determine the vector tile sizes used for vectorization of `inner_tiled`.
> 
> I think you can still split out the vectorization of inner_tiled as a seperate pr and tile size analysis stuff as a seperate pr.

@hanhanW @Groverkss I've split this up into this PR with masked vectorization of `inner_tiled` and the extension of the vector tile size analysis in https://github.com/iree-org/iree/pull/23971.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1042,6 +1043,28 @@ struct PadOpVectorizationModel
...
+    for (auto [mapResult, dimSize] :
+         llvm::zip_equal(map.getResults(), outerShape)) {
+      auto dimExpr = cast<AffineDimExpr>(mapResult);
+      if (!ShapedType::isDynamic(dimSize)) {
+        staticSizes[dimExpr.getPosition()] = dimSize;
+      }
+    }
+  }
+  return staticSizes;
+}
```
hanhanW: I think you can use `tiledOp.getIterationBounds()` method? (and inline the implemenation)

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1055,9 +1078,16 @@ struct InnerTiledOpVectorizationModel
...
+    SmallVector<int64_t> loopRanges = getStaticOuterLoopRanges(tiledOp);
+    // If vector sizes are provided (from tile size analysis or config),
+    // dynamic outer shapes are fine - they'll be masked during vectorization.
+    // However, vector sizes must be >= the static outer dimension sizes.
+    if (!vectorSizes.empty()) {
+      return succeeded(
+          vector::isValidMaskedInputVector(loopRanges, vectorSizes));
+    }
+    // Without vector sizes, require static outer shapes.
+    return !ShapedType::isDynamicShape(loopRanges);
```
hanhanW: nit: avoid double negation

```suggestion
    return ShapedType::isStaticShape(loopRanges);
```

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1070,6 +1100,19 @@ struct InnerTiledOpVectorizationModel
...
+    SmallVector<AffineMap> indexingMaps = tiledOp.getIndexingMapsArray();
+
+    // If no vector sizes are provided, use static loop ranges and the inBounds
+    // attribute instead of masking.
+    bool needsMasking = true;
+    if (vectorSizes.empty()) {
+      SmallVector<int64_t> loopRanges = getStaticOuterLoopRanges(tiledOp);
+      if (ShapedType::isDynamicShape(loopRanges)) {
+        return rewriter.notifyMatchFailure(op, "unable to infer vector sizes");
+      }
```
hanhanW: I think we prefer an assertion. We have full control for inner_tile vectorization, so this should never happen.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1081,29 +1124,74 @@ struct InnerTiledOpVectorizationModel
...
-         llvm::zip_equal(newTiledOp.getResults(), tiledOp.getOutputs())) {
-      int64_t rank = cast<RankedTensorType>(tensorAcc.getType()).getRank();
+    unsigned numInputs = tiledOp.getNumInputs();
+    for (auto [i, result, tensorAcc] :
+         llvm::enumerate(newTiledOp.getResults(), tiledOp.getOutputs())) {
+      auto tensorType = cast<RankedTensorType>(tensorAcc.getType());
+      int64_t rank = tensorType.getRank();
+      SmallVector<Value> indices(rank, zero);
+
+      SmallVector<int64_t> &writeShape = readShapes[numInputs + i];
```
hanhanW: Why do you use `SmallVector<int64>&`? I think what you're looking for is `ArrayRef`.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1081,29 +1124,74 @@ struct InnerTiledOpVectorizationModel
...
     auto newTiledOp = IREE::Codegen::InnerTiledOp::create(
         rewriter, loc,
         ValueRange{newOperands}.take_front(tiledOp.getNumInputs()),
         ValueRange{newOperands}.take_back(tiledOp.getNumOutputs()),
         tiledOp.getIndexingMaps(), tiledOp.getIteratorTypes(),
         tiledOp.getKind(), tiledOp.getSemantics());
 
+    // Write results back to tensor, with masking if needed.
+    // TODO: Use createWriteOrMaskedWrite once it is promoted to a public
+    // utility in mlir/Dialect/Vector/Utils/VectorUtils.h.
```
hanhanW, line 1144: Are you working on this?

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1081,29 +1124,74 @@ struct InnerTiledOpVectorizationModel
...
+         llvm::enumerate(newTiledOp.getResults(), tiledOp.getOutputs())) {
+      auto tensorType = cast<RankedTensorType>(tensorAcc.getType());
+      int64_t rank = tensorType.getRank();
+      SmallVector<Value> indices(rank, zero);
+
+      SmallVector<int64_t> &writeShape = readShapes[numInputs + i];
+      SmallVector<bool> inBounds(rank);
+      for (int64_t d = 0; d < rank; ++d) {
+        inBounds[d] = !tensorType.isDynamicDim(d) &&
+                      tensorType.getDimSize(d) >= writeShape[d];
```
hanhanW:
```suggestion
        inBounds[d] = tensorType.isStaticDim(d) &&
                      tensorType.getDimSize(d) >= writeShape[d];
```

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1081,29 +1124,74 @@ struct InnerTiledOpVectorizationModel
...
+    // utility in mlir/Dialect/Vector/Utils/VectorUtils.h.
     auto zero = arith::ConstantIndexOp::create(rewriter, loc, 0);
     SmallVector<Value> results;
-    for (auto [result, tensorAcc] :
-         llvm::zip_equal(newTiledOp.getResults(), tiledOp.getOutputs())) {
-      int64_t rank = cast<RankedTensorType>(tensorAcc.getType()).getRank();
+    unsigned numInputs = tiledOp.getNumInputs();
+    for (auto [i, result, tensorAcc] :
+         llvm::enumerate(newTiledOp.getResults(), tiledOp.getOutputs())) {
+      auto tensorType = cast<RankedTensorType>(tensorAcc.getType());
```
hanhanW: I'd rename `tesnorAcc` as `dest`, which is clearer top me, and `tensorType` can be `destType`.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1081,29 +1124,74 @@ struct InnerTiledOpVectorizationModel
...
+      }
+
       auto write = vector::TransferWriteOp::create(
-          rewriter, loc, result, tensorAcc,
-          /*indices=*/SmallVector<Value>(rank, zero),
-          /*inBounds=*/SmallVector<bool>(rank, true));
+          rewriter, loc, result, tensorAcc, indices, inBounds);
+      if (llvm::is_contained(inBounds, false)) {
+        auto vecType = cast<VectorType>(result.getType());
+        auto maskType = vecType.cloneWith({}, rewriter.getI1Type());
```
hanhanW: `std::nullopt` is clearer.

```suggestion
        auto maskType = vecType.cloneWith(/*shape=*/std::nullopt, rewriter.getI1Type());
```

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1081,29 +1124,74 @@ struct InnerTiledOpVectorizationModel
...
+        SmallVector<OpFoldResult> mixedSizes =
+            tensor::getMixedSizes(rewriter, loc, tensorAcc);
+        Value mask =
+            vector::CreateMaskOp::create(rewriter, loc, maskType, mixedSizes);
+        results.push_back(
+            mlir::vector::maskOperation(rewriter, write, mask)->getResult(0));
+        continue;
+      }
       results.push_back(write.getResults().front());
```
hanhanW: Why not just use the existing variable and flip the condition, i.e.,`if (!needsMasking)`? It makes the nested block shorter, right? And we do early-exit for simple cases.

hanhanW (changes requested):

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_inferred.mlir
```diff
@@ -873,3 +873,45 @@ func.func @scaled_tensor_multi_mma(%arg0: tensor<3x5x1x32xf4E2M1FN>, %arg1: tens
...
+// The LHS has shape <?x?x4xf16> (outer dims dynamic), iteration space is
+// [i=2, j=5, k=3], so reads should be masked.
+#contraction_accesses = [
+ affine_map<(i, j, k) -> (i, k)>,
+ affine_map<(i, j, k) -> (k, j)>,
+ affine_map<(i, j, k) -> (i, j)>
+]
+func.func @masked_tensor_multi_mma(%lhs: tensor<?x?x4xf16>, %rhs: tensor<?x?x4xf16>, %acc: tensor<?x?x4xf32>) -> tensor<?x?x4xf32> {
+  %0 = iree_codegen.inner_tiled ins(%lhs, %rhs) outs(%acc) {
+    iree_codegen.vector_tile_sizes = array<i64: 2, 5, 3>,
```
hanhanW, line 890: We will have to make it carry inner tile sizes in the future, but it is okay with what it is for now.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: Yes, I'm planning to bring an upstream PR once this PR is landed and update this with the upstream utility once the upstream PR is done.

hanhanW (approved):

Groverkss (approved):

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: @hanhanW The follow-up upstream PR is here: https://github.com/llvm/llvm-project/pull/190967

sommerlukas: @hanhanW Follow-up IREE PR after the upstream PR landed: https://github.com/iree-org/iree/pull/24069

PR #23937, opened 2026-03-26
[VectorDistribute] Allow duplication of operation chains

sommerlukas:
If a value gets assigned two different layouts, the VectorLayoutAnalysis tries to duplicate that value if it is easily duplicatable to avoid layout transformation via LDS.

So far, this duplication was limited to single operations. This PR extends this to bounded chains of easy to duplicate/recompute values.

The motivation for this change is masks. CSE will deduplicate `create_mask` operations. After early materialization of `create_mask` operations, the mask is transformed to a chain of operations:
```mlir
%step = vector.step : vector<64xindex>
%limit = vector.broadcast %n : index to vector<64xindex>
%mask_1d = arith.cmpi slt, %step, %limit : vector<64xindex>
%mask = vector.broadcast %mask_1d : vector<64xi1> to vector<16x64xi1>
```
As the mask is used for different operations, it will receive multiple different layouts.  Therefore, we would materialize these masks to LDS for layout transformation. 

By allowing chains of cheap operations to be duplicated, we avoid that materialization in LDS.

Assisted-by: Claude Code

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -16,7 +16,6 @@
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
 #include "mlir/Analysis/TopologicalSortUtils.h"
-#include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
@@ -33,6 +32,10 @@ using namespace IREE::VectorExt;
 /// Maximum number of candidate layouts tracked per value. Kept small to bound
 /// analysis cost; most values see 1-2 candidates in practice.
 static constexpr int kMaxCandidatesPerValue = 4;
+
+/// Maximum length of chains of cheap-to-compute operations that get duplicated
+/// for layout conflict resolution.
+static constexpr unsigned kMaxChainLength = 8;
 
 //===----------------------------------------------------------------------===//
 // Layout Analysis
@@ -482,10 +485,10 @@ static bool isCheapToClone(Operation *op) {
   if (isDuplicatableLeaf(op)) {
     return true;
   }
-  return op->getNumResults() == 1 && isPure(op) &&
-         isa<vector::BroadcastOp, vector::TransposeOp, vector::ShapeCastOp,
-             arith::CmpIOp, arith::SelectOp, arith::ExtSIOp, arith::ExtUIOp,
-             arith::TruncIOp, arith::IndexCastOp>(op);
+  return isPure(op) &&
+         (isa<vector::BroadcastOp, vector::TransposeOp, vector::ShapeCastOp>(
+              op) ||
+          OpTrait::hasElementwiseMappableTraits(op));
 }
 
 /// Collect a chain of ops that can be cloned together. Starting from `op`,
@@ -496,7 +499,6 @@ static bool isCheapToClone(Operation *op) {
 static bool collectDuplicatableChain(Operation *op,
                                      SmallVectorImpl<Operation *> &chain) {
   // The chain is built bottom-up (from consumer toward producers).
-  constexpr unsigned kMaxChainLength = 8;
   Block *block = op->getBlock();
   std::queue<Operation *> worklist;
   llvm::SmallPtrSet<Operation *, 8> visited;
@@ -569,8 +571,9 @@ void LayoutAnalysis::setLayoutOrClone(OpOperand *val,
         for (Operation *op : chain) {
           b.clone(*op, mapping);
         }
-        val->set(mapping.lookup(defOp->getResult(0)));
-        resolved[mapping.lookup(defOp->getResult(0))] = layout;
+        Value cloned = mapping.lookup(val->get());
+        val->set(cloned);
+        resolved[cloned] = layout;
         // Propagate layouts through the cloned chain. The cloned ops are
         // not visited by the outer fixupRegion walk (which collects ops
         // upfront), so we must fix them up here. Walk in reverse program
--- a/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
@@ -35,7 +35,7 @@ static constexpr int kMaxCandidatesPerValue = 4;
 
 /// Maximum length of chains of cheap-to-compute operations that get duplicated
 /// for layout conflict resolution.
-static constexpr unsigned kMaxChainLength = 8;
+static constexpr size_t kMaxChainLength = 8;
 
 //===----------------------------------------------------------------------===//
 // Layout Analysis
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/test/vector_layout_analysis_chain_cloning.mlir
```diff
@@ -0,0 +1,97 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-test-vector-layout-analysis))" --split-input-file %s | FileCheck %s
```
Muzammiluddin-Syed-ECE, line 1: nit: can we have a negative test?

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -464,6 +468,71 @@ void LayoutAnalysis::fixupOp(Operation *op) {
...
+/// cloned as part of a duplicatable chain. These ops must be pure and have
+/// exactly one result.
+static bool isCheapToClone(Operation *op) {
+  if (isDuplicatableLeaf(op)) {
+    return true;
+  }
+  return op->getNumResults() == 1 && isPure(op) &&
+         isa<vector::BroadcastOp, vector::TransposeOp, vector::ShapeCastOp,
+             arith::CmpIOp, arith::SelectOp, arith::ExtSIOp, arith::ExtUIOp,
+             arith::TruncIOp, arith::IndexCastOp>(op);
```
Groverkss: use elementwise op traits. all elementwise ops are cheap.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -464,6 +468,71 @@ void LayoutAnalysis::fixupOp(Operation *op) {
...
+}
+
+/// Returns true if the operation is a cheap single-result op that can be
+/// cloned as part of a duplicatable chain. These ops must be pure and have
+/// exactly one result.
+static bool isCheapToClone(Operation *op) {
+  if (isDuplicatableLeaf(op)) {
+    return true;
+  }
+  return op->getNumResults() == 1 && isPure(op) &&
```
Groverkss: i dont think we should check if num elements == 1. Even if it isn't, these things are cheap.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -464,6 +468,71 @@ void LayoutAnalysis::fixupOp(Operation *op) {
...
+
+/// Collect a chain of ops that can be cloned together. Starting from `op`,
+/// walk backward through single-result, cheap-to-clone ops until we reach
+/// duplicatable leaves, constants, or non-vector operands. Returns true if
+/// the entire chain is safe to clone. Shared intermediates (with multiple
+/// uses) are allowed because all ops in the chain are cheap to duplicate.
+static bool collectDuplicatableChain(Operation *op,
+                                     SmallVectorImpl<Operation *> &chain) {
+  // The chain is built bottom-up (from consumer toward producers).
+  constexpr unsigned kMaxChainLength = 8;
```
Groverkss: promote this outside the function.

compiler/src/iree/compiler/Codegen/Common/test/vector_layout_analysis_chain_cloning.mlir
sommerlukas: I've added a negative test where chain cloning is aborted because a non-cheap operation.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -30,6 +33,10 @@ using namespace IREE::VectorExt;
 /// analysis cost; most values see 1-2 candidates in practice.
 static constexpr int kMaxCandidatesPerValue = 4;
 
+/// Maximum length of chains of cheap-to-compute operations that get duplicated
+/// for layout conflict resolution.
+static constexpr unsigned kMaxChainLength = 8;
```
Groverkss: any reason for unsigned? Can we just use int? (No need to reply, just keep it if you have a reason)

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -464,6 +471,70 @@ void LayoutAnalysis::fixupOp(Operation *op) {
...
+         (isa<vector::BroadcastOp, vector::TransposeOp, vector::ShapeCastOp>(
+              op) ||
+          OpTrait::hasElementwiseMappableTraits(op));
+}
+
+/// Collect a chain of ops that can be cloned together. Starting from `op`,
+/// walk backward through single-result, cheap-to-clone ops until we reach
+/// duplicatable leaves, constants, or non-vector operands. Returns true if
+/// the entire chain is safe to clone. Shared intermediates (with multiple
+/// uses) are allowed because all ops in the chain are cheap to duplicate.
```
Groverkss: Not sure why single result is important here. We happen to have most things as single result today, but just curious why.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -464,6 +471,70 @@ void LayoutAnalysis::fixupOp(Operation *op) {
...
+      return false;
+    }
+    if (isDuplicatableLeaf(current)) {
+      continue;
+    }
+    for (Value operand : current->getOperands()) {
+      // Non-vector operands (scalars, indices) don't need cloning.
+      if (!isa<VectorType>(operand.getType())) {
+        continue;
+      }
```
Groverkss, line 527: We could probably say this for any vector with 1 element as well.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
```diff
@@ -489,17 +560,29 @@ void LayoutAnalysis::setLayoutOrClone(OpOperand *val,
...
+        Value cloned = mapping.lookup(val->get());
+        val->set(cloned);
+        resolved[cloned] = layout;
+        // Propagate layouts through the cloned chain. The cloned ops are
+        // not visited by the outer fixupRegion walk (which collects ops
+        // upfront), so we must fix them up here. Walk in reverse program
+        // order so that result layouts propagate to operands.
+        for (Operation *op : llvm::reverse(chain)) {
+          fixupOp(mapping.lookup(op->getResult(0)).getDefiningOp());
+        }
```
Groverkss, line 589: Can we add a comment on why this will not recursively loop?

Groverkss (approved):
LGTM, feel free to land after addressing comments

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
sommerlukas: Done, I've added an early `continue` for single-element vectors.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
sommerlukas: Added.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
sommerlukas: The comment was outdated, the code didn't check for a single result anymore, but I forgot to update the comment.

compiler/src/iree/compiler/Codegen/Common/VectorLayoutAnalysis.cpp
sommerlukas: The only place we use it is for a comparison with `size()`, so I now decided to make it `size_t` to match the return type.

PR #23922, opened 2026-03-24
[CPU] Sink down reshapes across packing ops.

hanhanW:
The reshape ops are fused because of data-tiling fusion. We should sink down the ops and fuse them into bindings, if possible. If it is not possible, they become MapStoreOp so the sequence of the ops all implement tiling interface.

Both solutions are required, and the revision adds the former optimization to the pass.

Fixes https://github.com/iree-org/iree/issues/23920

PR #23943, opened 2026-03-26
Revert "[CPU] Sink down reshapes across packing ops."

hanhanW:
Reverts iree-org/iree#23922

It breaks shark_tank cpu tests that I did not notice when I landed the PR: https://github.com/iree-org/iree/actions/runs/23508790791/job/68424982838


bangtianliu (approved):
Thanks

PR #23945, opened 2026-03-27
[Codegen] Absorb SwizzleHintOp into alloc attribute for pipelining

### LINKED ISSUE #23919: SwizzleHintOp blocks double-buffering in ROCDLPrefetchSharedMemoryPass
When XOR swizzle is enabled alongside DMA, ROCDLPrefetchSharedMemoryPass currently fails to apply double-buffering.

With swizzle enabled, the alloc becomes flat 1D and swizzle_hint + expand_shape sit between the alloc and the K-loop:
```
%alloc = memref.alloc() : memref<4096xbf16, #gpu.address_space<workgroup>>
// swizzle_hint + expand_shape sit OUTSIDE the K-loop
%hint = iree_codegen.swizzle_hint %alloc[#iree_codegen.xor_shuffle<128, 8>]
  : memref<4096xbf16, #gpu.address_space<workgroup>>
%expand = memref.expand_shape %hint [[0, 1]] output_shape [128, 32]
  : memref<4096xbf16, ...> into memref<128x32xbf16, ...>
scf.for %iv = %c0 to %c16 step %c4 {           // K-loop
  %sub = memref.subview %expand[%off, 0] ...
  amdgpu.gather_to_lds ..., %sub               // DMA write
  vector.transfer_read %expand ...              // MMA read
}
```

`memref::multiBuffer` produces `memref<2x4096xbf16>` with per-iteration subviews (`memref<4096xbf16, strided<[1], offset: ?>>`), but swizzle_hint requires a 1D non-strided memref. Double-buffering silently fails.

Should we:
(a) Clone SwizzleHintOp into the K-loop so it operates on per-iteration subviews, or
(b) Change multi-buffering to produce a flat 1D alloc (memref<8192xbf16> instead of memref<2x4096xbf16>) so swizzle can stay outside the loop?

Yu-Zhewen:
Absorb `SwizzleHintOp` into an attribute on `memref.alloc` before pipelining, then re-insert hints at leaf users afterward. This decouples swizzle from both `memref::multiBuffer` and `scf::pipelineForLoop`, as neither upstream utility handles `SwizzleHintOp`.

- **AbsorbSwizzleHintToAllocPass** (before pipeliner): moves swizzle info from the SSA chain to an `iree_codegen.swizzle` attribute on the alloc, erases SwizzleHintOp. Multi-buffering and pipelining proceed on clean IR.
- **ROCDLPrefetchSharedMemoryCopy**: copies the swizzle attribute to the new alloc after `memref::multiBuffer`.
- **ReinsertSwizzleHintsPass** (after vector lowering): traces back to attributed allocs and inserts `collapse_shape -> swizzle_hint -> expand_shape` at each use site.

Fixes: #23919

Assisted-by: Cursor (Claude)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
@@ -1,0 +1,142 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Interfaces/ViewLikeInterface.h"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_REINSERTSWIZZLEHINTSPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+struct ReinsertSwizzleHintsPass final
+    : impl::ReinsertSwizzleHintsPassBase<ReinsertSwizzleHintsPass> {
+  using Base::Base;
+  void runOnOperation() override;
+};
+} // namespace
+
+/// Traces a memref value backward through view-like ops, memref.cast, and
+/// scf.for iter_args/results to find the root memref.alloc.
+static memref::AllocOp traceToAllocation(Value val) {
+  DenseSet<Value> visited;
+  SmallVector<Value> worklist = {val};
+  while (!worklist.empty()) {
+    Value current = worklist.pop_back_val();
+    if (!visited.insert(current).second) {
+      continue;
+    }
+    Operation *defOp = current.getDefiningOp();
+    if (!defOp) {
+      auto blockArg = cast<BlockArgument>(current);
+      auto forOp = dyn_cast<scf::ForOp>(blockArg.getOwner()->getParentOp());
+      unsigned argIdx = blockArg.getArgNumber();
+      if (!forOp || argIdx == 0) {
+        continue;
+      }
+      worklist.push_back(forOp.getInitArgs()[argIdx - 1]);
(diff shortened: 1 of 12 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -289,7 +399,7 @@ static LogicalResult multiBufferLDSAllocations(scf::ForOp forOp,
 
   // First, clone view ops inside the loop for each allocation
```
lialan: nit: no more `view`.

Yu-Zhewen:
After discussing with @jerryyin, we feel the current approach is quite brittle. I'm going to explore alternative options. One possibility is to have `SwizzleHintOp` handle multi-buffering directly (currently it only handles flattening), and leave it outside the pipelining loop.

Converting back to draft while I investigate.

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
```diff
@@ -0,0 +1,142 @@
...
+  IRRewriter rewriter(funcOp->getContext());
+  DenseMap<Operation *, IREE::Codegen::SwizzleAttrInterface> swizzleCache;
+
+  // For each vector.load/store whose base traces to a swizzled alloc, wrap the
+  // base with collapse_shape -> swizzle_hint -> expand_shape.
+  funcOp.walk([&](Operation *op) {
+    Value base;
+    if (auto loadOp = dyn_cast<vector::LoadOp>(op)) {
+      base = loadOp.getBase();
+    } else if (auto storeOp = dyn_cast<vector::StoreOp>(op)) {
```
Yu-Zhewen, line 117: Handles `StoreOp` for non-DMA cases.

`gather_to_lds` swizzle is not handled here, as it is done separately in `AMDGPULowerCoalescedDMAToGatherLDS`(#23807), which applies an inverse swizzle on the source indices. This cannot use `swizzle_hint`  op because the source is the strided global memory.

jerryyin:
Given that direct DMA load hasn't become default yet, can you break down this PR into multiple smaller ones so it is easier to review? Like:
 - PR 1: temporarily flip swizzle hint enablement and absorb pass
 - PR 2: reinsert pass
 - PR 3: prefetch pass and swizzle hint re-enablement

The current PR is a bit context heavy and touched a lot of orthogonal components.

Yu-Zhewen:
> Given that direct DMA load hasn't become default yet, can you break down this PR into multiple smaller ones so it is easier to review? Like:
> 
> * PR 1: temporarily flip swizzle hint enablement and absorb pass
> * PR 2: reinsert pass
> * PR 3: prefetch pass and swizzle hint re-enablement
> 
> The current PR is a bit context heavy and touched a lot of orthogonal components.

Thanks, it's now split into #24000, #24001, #24002

PR #23947, opened 2026-03-27
[LLVMGPU] Support minor identity permutation maps in ROCDLBufferInstructionsOptimization

Max191:
The pass was bailing out on vector.transfer_read ops with non-identity permutation maps (e.g., 1D reads from a 4D memref). After https://github.com/iree-org/iree/pull/23855, we will frequently see 1D reads, which need to be supported here. Ideally, we will do something like what is done in https://github.com/iree-org/iree/pull/23859, but that approach is causing performance regressions that are difficult to deal with. For now, this provides a solution for the new mask types we will be seeing.


krzysz00 (approved):
Looks like an obvious generalization, lgtm

Issue #23941, opened 2026-03-26
[GPU] Add CDNA block intrinsics


nirvedhmeshram:
Block intrinsics allow subgroups to work on muliple blocks (batch dimenion) in parallel. They often have smaller sizes which make them a good fit for skinny shapes. This issue will track various steps to have them added. These tasks are sharded from this test PR which verified that we have e2e correctness https://github.com/iree-org/iree/pull/23934

- [x] [Add intrinsics and layout](https://github.com/iree-org/iree/pull/23949)
- [x] [Add Pack to intrinisic support](https://github.com/iree-org/iree/pull/23963)
- [x] [Add ConfigureTensorLayout support (required by VectorDistribute)](https://github.com/iree-org/iree/pull/23973)
- [x] [Fix inner tiled to amdgpu.mfma lowering for single element result seen in block intrinisics](https://github.com/iree-org/iree/pull/23977)
- [x]  [Add e2e batch matmul tests for both TileAndFuse and VectorDistribute](https://github.com/iree-org/iree/pull/23987)
- [x] [Add blockIntrinisics in the target details and update heuristic to use it in appropriate situations with correct lowering configs.](https://github.com/iree-org/iree/pull/24020)

PR #23949, opened 2026-03-27
[GPU] Implement block intrinsics introduced in CDNA3 (1/6)

nirvedhmeshram:
Add block intrinsic enum definitions and corresponding layout structures. These layouts have been e2e verified in a test PR
https://github.com/iree-org/iree/pull/23934. and also locally tested on cdna4.
They will not be added to the target details until we have correct heuristic support and selection policy for them. The plan for this is described in https://github.com/iree-org/iree/issues/23941


nirvedhmeshram:
@krzysz00 could you please review this PR, I suspect you already looked at the changes in https://github.com/iree-org/iree/pull/23963

krzysz00 (approved):
I trust that you've got the layouts correct and don't want to go digging in the manual for them, appronved

PR #23955, opened 2026-03-29
samples: Refactor input handling with data-driven approach

HanKuanChen:
- Introduce `input_buffer_descriptor_t` struct to describe input buffer metadata (shape, element type, data pointer, size).
- Add `push_back_buffer_view()` helper function to encapsulate buffer view allocation and input pushing logic, eliminating code duplication.
- Use array of descriptors with a loop to process inputs, making it easier to add or modify inputs without repeating code.

This refactoring improves code maintainability and makes it simpler to work with different element types (e.g., float32, float16, uint8).

### DIFF
--- a/samples/static_library/static_library_demo.c
+++ b/samples/static_library/static_library_demo.c
@@ -24,6 +24,40 @@ extern iree_status_t create_module(iree_vm_instance_t* instance,
 
 extern void print_success();
 
+typedef struct {
+  iree_host_size_t shape_rank;
+  const iree_hal_dim_t* shape;
+  iree_hal_element_type_t element_type;
+  const void* data;
+  iree_host_size_t data_length;
+} input_buffer_descriptor_t;
+
+// Helper function to allocate a buffer view and push it as an input to a
+// runtime call.
+// This function performs the following operations:
+// 1. Allocates a device-local buffer view with the provided data
+// 2. Pushes the buffer view to the runtime call's input list
+// 3. Releases the buffer view reference (the call retains its own reference)
+static iree_status_t push_back_buffer_view(
+    iree_hal_device_t* device, iree_host_size_t shape_rank,
+    const iree_hal_dim_t* shape, iree_hal_element_type_t element_type,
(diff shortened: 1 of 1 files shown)


HanKuanChen:
can you take a look of this? @benvanik 
cc @hanhanW @ScottTodd

HanKuanChen:
ping

samples/static_library/static_library_demo.c
```diff
@@ -24,6 +24,40 @@ extern iree_status_t create_module(iree_vm_instance_t* instance,
...
+// Helper function to allocate a buffer view and push it as an input to a
+// runtime call.
+// This function performs the following operations:
+// 1. Allocates a device-local buffer view with the provided data
+// 2. Pushes the buffer view to the runtime call's input list
+// 3. Releases the buffer view reference (the call retains its own reference)
+static iree_status_t push_back_buffer_view(
+    iree_hal_device_t* device, iree_host_size_t shape_rank,
+    const iree_hal_dim_t* shape, iree_hal_element_type_t element_type,
+    const void* data, iree_host_size_t data_length, iree_runtime_call_t* call) {
```
hanhanW: You should probably pass in `input_buffer_descriptor_t`. If we're introducing a descriptor struct, the helper should accept it, which is a significant readability improvement

hanhanW (commented):
The refactoring is fine, but I wonder why do we need the change? Isn't it just a sample about how IREE can be used?

samples/static_library/static_library_demo.c
HanKuanChen: fix

HanKuanChen:
> The refactoring is fine, but I wonder why do we need the change? Isn't it just a sample about how IREE can be used?

Since samples are often used as starting points for downstream users to copy and modify, I think providing cleaner patterns helps them adapt the code more easily.

With this refactoring, adding new inputs or supporting different element types (float16, uint8, etc.) just requires adding entries to the descriptor array, rather than duplicating the buffer allocation code. This makes the sample more practical as a reference for real-world use cases.

hanhanW (approved):

PR #23960, opened 2026-03-30
[LLVMGPU] Only set reduction lowering_config for ops that need it

Groverkss:
An operation only needs a reduction lowering_config if it cannot infer it from a consumer compute op. Before this patch, we were adding lowering_configs for some elementwise operations that didn't need them and caused extra loops if we did not add the rematerialize elementwise operations pass.

Before this patch we would end up with something like:

```
elementwise {lowering_config}
reduce {lowering_config}
```

The rematerialize pass would fuse the elementwise with the reduction keeping:

```
reduce {lowering_config}
```

if we didn't do this, the elementwise and reduce would end up in their own scf.for loop. This was a workkaround of us setting lowering_config on the elementwise sometimes accidently. This patch makes it so that we only set:

```
elementwise
reduce {lowering_config}
```

This patch fixes the lowering_config setting and removes the rematerialize elementwise operations pass from VectorDistribute pipeline.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
@@ -158,14 +158,51 @@ static LogicalResult checkSingleCombiner(linalg::LinalgOp op) {
   return success();
 }
 
-static llvm::FailureOr<IREE::GPU::LoweringConfigAttr>
-getVectorDistributeReductionConfig(
+/// Stores per-op tile sizes without building the lowering config attribute.
+/// The workgroup tile sizes are tracked separately in sharedWgpTiles and only
+/// applied to the last compute operation.
+struct TileSizesConfig {
+  SmallVector<int64_t> tileSizes; // Tile sizes for "serial" or
+                                  // "partial_reduction" level.
+  SmallVector<int64_t> threadTileSizes;
+  SmallVector<int64_t> threadCounts;
+  SmallVector<int64_t> subgroupCounts;
+  bool isReduction = false;
+
+  LoweringConfigAttr
+  buildConfig(MLIRContext *context,
+              std::optional<ArrayRef<int64_t>> workgroupTileSizes) const {
+    Builder b(context);
+    SmallVector<int64_t> mapping(tileSizes.size());
+    std::iota(mapping.begin(), mapping.end(), 0);
+
+    ArrayAttr subgroupBasisAttr = b.getArrayAttr(
+        {b.getI64ArrayAttr(subgroupCounts), b.getI64ArrayAttr(mapping)});
+    ArrayAttr laneBasisAttr = b.getArrayAttr(
+        {b.getI64ArrayAttr(threadCounts), b.getI64ArrayAttr(mapping)});
+
+    SmallVector<NamedAttribute> configAttrs;
+    if (workgroupTileSizes) {
+      configAttrs.push_back(
+          b.getNamedAttr("workgroup", b.getI64ArrayAttr(*workgroupTileSizes)));
+    }
+    configAttrs.push_back(
+        b.getNamedAttr(isReduction ? "partial_reduction" : "serial",
+                       b.getI64ArrayAttr(tileSizes)));
+    configAttrs.push_back(
+        b.getNamedAttr("thread", b.getI64ArrayAttr(threadTileSizes)));
+    configAttrs.push_back(b.getNamedAttr("lane_basis", laneBasisAttr));
+    configAttrs.push_back(b.getNamedAttr("subgroup_basis", subgroupBasisAttr));
+
+    auto configDict = b.getDictionaryAttr(configAttrs);
+    return IREE::GPU::LoweringConfigAttr::get(context, configDict);
+  }
+};
+
+static FailureOr<TileSizesConfig> getVectorDistributeReductionConfig(
     linalg::LinalgOp op, IREE::GPU::TargetAttr target,
     llvm::SmallDenseMap<unsigned, unsigned> &sharedWgpTiles,
     int64_t workgroupSize, int64_t subgroupSize, int64_t threadLoads) {
-  MLIRContext *context = op.getContext();
-  Builder b(context);
-
   SmallVector<unsigned> parallelDims;
   SmallVector<unsigned> reductionDims;
   op.getParallelDims(parallelDims);
@@ -173,12 +210,9 @@ getVectorDistributeReductionConfig(
 
   SmallVector<int64_t> bounds = op.getStaticLoopRanges();
 
-  SmallVector<int64_t> workgroupTileSizes(op.getNumLoops(), 0);
   SmallVector<int64_t> threadTileSizes(op.getNumLoops(), 0);
   SmallVector<int64_t> threadCounts(op.getNumLoops(), 1);
-  SmallVector<int64_t> subGroupCounts(op.getNumLoops(), 1);
-  SmallVector<int64_t> mapping(op.getNumLoops());
-  std::iota(mapping.begin(), mapping.end(), 0);
+  SmallVector<int64_t> subgroupCounts(op.getNumLoops(), 1);
 
   // Set the configuration for the operation with no reduction dims.
   // The workgroup tile sizes are set by the reduction operation.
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -158,38 +158,69 @@ static LogicalResult checkSingleCombiner(linalg::LinalgOp op) {
   return success();
 }
 
-static llvm::FailureOr<IREE::GPU::LoweringConfigAttr>
-getVectorDistributeReductionConfig(
+/// Storage for tracking tile size configuration. Workgroup tile sizes are
+/// tracked separate. "tileSizes" materializes as serial level for parallel
+/// iterators and reduction level for reduction iterators.
+struct TileSizesConfig {
+  SmallVector<int64_t> tileSizes;
```
sommerlukas, line 165: Is this the equivalent of `batch x outer x element` in nested layouts?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -158,38 +158,69 @@ static LogicalResult checkSingleCombiner(linalg::LinalgOp op) {
...
+  SmallVector<int64_t> threadCounts;
+  SmallVector<int64_t> subgroupCounts;
+  bool isReduction = false;
+
+  LoweringConfigAttr
+  buildConfig(MLIRContext *context,
+              std::optional<ArrayRef<int64_t>> workgroupTileSizes) const {
+    Builder b(context);
+    SmallVector<int64_t> mapping(tileSizes.size());
+    std::iota(mapping.begin(), mapping.end(), 0);
```
sommerlukas:
```suggestion
    SmallVector<int64_t> mapping = llvm::to_vector(llvm::seq<int64_t>(tileSizes.size()));
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -158,38 +158,69 @@ static LogicalResult checkSingleCombiner(linalg::LinalgOp op) {
...
+    ArrayAttr laneBasisAttr = b.getArrayAttr(
+        {b.getI64ArrayAttr(threadCounts), b.getI64ArrayAttr(mapping)});
+
+    SmallVector<NamedAttribute> configAttrs;
+    if (workgroupTileSizes) {
+      configAttrs.push_back(
+          b.getNamedAttr("workgroup", b.getI64ArrayAttr(*workgroupTileSizes)));
+    }
+    configAttrs.push_back(
+        b.getNamedAttr(isReduction ? "partial_reduction" : "serial",
```
sommerlukas, line 188: Could there be cases where we want `reduction` instead of `partial_reduction` here?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -371,69 +367,74 @@ populateConfigInfo(const llvm::SetVector<linalg::LinalgOp> &computeOps,
...
+    tileSizeConfigs.push_back({linalgOp, *config});
   }
-  // Only set lowering configs once we've successfully determined them for all
-  // operations, to avoid leaving the IR in an inconsistent state on failure.
-  for (auto &[linalgOp, loweringConfig] : loweringConfigs) {
-    setLoweringConfig(linalgOp, loweringConfig);
+
+  // Build and set lowering configurations. Only the first op in
+  // tileSizeConfigs (the last compute op in topological order) gets the
+  // workgroup tile sizes.
```
sommerlukas: What is the rationale for this? Can we document that rationale here?

compiler/src/iree/compiler/Codegen/LLVMGPU/test/reduction_pipeline_cuda.mlir
```diff
@@ -519,9 +519,9 @@ hal.executable private @i4_dequant_matvec {
 //         CHECK:   scf.for %{{.+}} = %[[C0]] to %[[C32]] step %[[C1]] iter_args(%{{.*}} = %[[CST]]) -> (vector<1x1x1x1x1x1x1x1x1xf16>)
 //         CHECK:     vector.transfer_read {{.*}} : memref<32x128xf16, {{.*}}>, vector<1x4xf16>
 //         CHECK:     vector.transfer_read {{.*}} : memref<4096x32x128xi4, {{.*}}>, vector<1x1x4xi4>
-//         CHECK:     arith.uitofp %{{.*}} : vector<1x1x1x1x1x1x1x1x1x1x1x4xi4> to vector<1x1x1x1x1x1x1x1x1x1x1x4xf16>
-//         CHECK:     arith.subf %{{.*}}, %{{.*}} : vector<1x1x1x1x1x1x1x1x1x1x1x4xf16>
-//         CHECK:     arith.mulf %{{.*}}, %{{.*}} : vector<1x1x1x1x1x1x1x1x1x1x1x4xf16>
+//         CHECK:     arith.uitofp %{{.*}} : vector<1x1x1x1x1x1x1x1x4xi4> to vector<1x1x1x1x1x1x1x1x4xf16>
+//         CHECK:     arith.subf %{{.*}}, %{{.*}} : vector<1x1x1x1x1x1x1x1x4xf16>
+//         CHECK:     arith.mulf %{{.*}}, %{{.*}} : vector<1x1x1x1x1x1x1x1x4xf16>
```
sommerlukas, line 524: What's the background of this change? Is it because we avoid the additional loops now or dropped the pass?

sommerlukas (commented):
Can you elaborate a bit on why setting fewer lowering configs allows us to drop the pass that fuses elementwise `linalg.generic`? Maybe with an example?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
Groverkss: Not really. This is the tiling configuration. This has no relation to batch x outer x element.

Groverkss: Not really. This is the tiling configuration. This has no relation to batch x outer x element.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
Groverkss: Not in the current IR. Note that the configuration change is NFC.

Groverkss:
> Can you elaborate a bit on why setting fewer lowering configs allows us to drop the pass that fuses elementwise `linalg.generic`? Maybe with an example?

added description with example

compiler/src/iree/compiler/Codegen/LLVMGPU/test/reduction_pipeline_cuda.mlir
Groverkss: we were fusing the elementwise with the reduce before, now we don't and it becomes an elementwise op by itself so fewer 1s.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -371,69 +366,83 @@ populateConfigInfo(const llvm::SetVector<linalg::LinalgOp> &computeOps,
...
+    if (failed(config)) {
+      return failure();
     }
+    tileSizeConfigs.push_back({linalgOp, *config});
+  }
+
+  // Build and set lowering configurations. Only the last operation in the
+  // dispatch gets the workgroup tile sizes. This is because workgroup tile
+  // sizes can only be set on dimensions that are shared on all operations, so
+  // we just need to set it on one operation. We just choose the last operation.
```
sommerlukas, line 430: Nit: I think it would make sense to have the comment after this unrelated early return closer to the logic it describes.

sommerlukas (approved):
LGTM, just a really minor nit.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
Groverkss: I can fix it up as part of the next pr, i'm going to move some things around here anyway.

PR #23961, opened 2026-03-30
[compiler][nfc] Move general passes from Codegen/Common/ to compiler/Transforms/

hanhanW:
These passes often serve as wrappers around upstream MLIR passes or implement IREE-specific but backend-agnostic transformations. Making them in Codegen/Common enforce the deps for phases that use the passes. The revision moves them to `compiler/Transforms/` and add a `README.md` to avoid the deps, as they can be used by any phases.


krzysz00 (approved):
No objections

nirvedhmeshram (approved):
Thanks!

AGindinson (approved):
LGTM

PR #23969, opened 2026-03-31
[Ccodegen] Add util.hoistable_conversion, use it to clean up accumulators

krzysz00:
Firstly, this commit introduces the util.hoistable_conversion operation. This operation lets you wrap a block of operations, giving them a tag and the tag of their inverse. This operation is removed by the new `eliminateHoistableConversions` set of pattern when either
1. A block of conversions flows directly into its inverse or
2. The cancellation is loop-carried: you convert a loop arg, then, right before yielding in the loop, you perform the inverse conversion. (in that case, the conversions are moved to the loop inits and the final results, respectively)
3. If none of these can happen, the operations get inlined.

This infrastructure is introduced to allow us to pull the conversions that get placed around inner_tiled intrinsics out of loops, though it may have other future uses. For example, after this change, the interleaves and deinterleaves needed to make RDNA3's f16/bf16 intrinsics behave correctly are pulled out of the loop. Less performance-sensitively, the varous unrollings and shape_casts we introduce on parallel inner_tiled accumulators now get pulled out of their loops, simplifying types and making it clearer that we're dealing with independent values (and getting rid of a bunch of shuffling) which should make it clearer that we're generating independent tiles.

These hoistable_conversion ops are currently meant to be introduced and removed within a single pipeline step, given their nature as opaque black boxes that you shouldn't pull operations into or out of. However, if a usecase is found for keeeping these around longer, we might want to leave them around in the future.

This PR puts hoistable_conversion ops anywhere they make sense in the back part of inner_tiled lowering, since it was the motivating example, though it can be used in more places than that going forward.

Note that we add an `iree-codegen-hoist-inner-tiled-acc-swizzles` pass to be run after vector distribution to fix up the transpose + shape_cast chains that vector distribution inserts, so that we can hoist the results of various unrollings and lowerings to intrinsics later in the pipeline. This is a hack, but doing otherwise seems to require getting deep into the guts of vector distribution or adding special cases for hoistable conversions.


compiler/src/iree/compiler/Codegen/Common/HoistInnerTiledAccSwizzles.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Interfaces/LoopLikeInterface.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_HOISTINNERTILEDACCSWIZZLESPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+static bool isSwizzleOp(Operation *op) {
```
efric: nit: I find `swizzle` to be quite an overloaded term. can we add a brief comment explaining what constitutes a swizzle here?

compiler/src/iree/compiler/Codegen/Common/HoistInnerTiledAccSwizzles.cpp
```diff
@@ -0,0 +1,142 @@
+// Copyright 2026 The IREE Authors
```
efric: I can see this getting the job done but the right approach is probably to teach vector distribute about this op. I wonder what @Groverkss or @sommerlukas thoughts are.

compiler/src/iree/compiler/Dialect/Util/Transforms/EliminateHoistableConversions.cpp
```diff
@@ -0,0 +1,345 @@
...
+  parentBlock->getOperations().splice(op->getIterator(), body.getOperations());
+
+  rewriter.replaceOp(op, returnValues);
+}
+
+//===----------------------------------------------------------------------===//
+// Directly cancelling hoistable conversions
+//===----------------------------------------------------------------------===//
+
+struct CancelInversePairPattern : OpRewritePattern<HoistableConversionOp> {
```
efric:
```suggestion
struct CancelInversePairPattern final : OpRewritePattern<HoistableConversionOp> {
```

compiler/src/iree/compiler/Dialect/Util/Transforms/EliminateHoistableConversions.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+//===----------------------------------------------------------------------===//
+// Hoisting conversions at loop edges
+//===----------------------------------------------------------------------===//
+
+/// Matches a HoistableConversionOp G whose results are yielded back to loop
+/// iter_args, finds the matching F that consumes those iter_args, and hoists
+/// both conversions out of the loop. F may consume a subset of the loop's
+/// iter_args (partial hoisting).
+struct HoistConversionFromLoopPattern
```
efric:
```suggestion
struct HoistConversionFromLoopPattern final
```

compiler/src/iree/compiler/Dialect/Util/Transforms/EliminateHoistableConversions.cpp
```diff
@@ -0,0 +1,345 @@
...
+    return failure();
+  }
+
+  SmallVector<HoistableConversionOp> toInline;
+  op->walk([&](HoistableConversionOp hc) { toInline.push_back(hc); });
+  IRRewriter rewriter(context);
+  for (auto hc : toInline) {
+    if (!hc->hasAttr(kWasHoistedAttr) && isPlausiblyHoistable(hc)) {
+      hc->emitRemark(
+          "hoistable_conversion was not hoisted or cancelled; inlining in "
```
efric, line 344: is there a test for this?

compiler/src/iree/compiler/Dialect/Util/Transforms/DropCompilerHints.cpp
```diff
@@ -28,6 +28,22 @@ struct DropCompilerHintsPass
       if (auto op = dyn_cast<IREE::Util::OptimizationBarrierOp>(genericOp)) {
         op.replaceAllUsesWith(op.getOperands());
         op.erase();
+      } else if (auto op =
+                     dyn_cast<IREE::Util::HoistableConversionOp>(genericOp)) {
+        op->emitWarning("hoistable_conversion should have been eliminated "
```
efric, line 33: is there a test for this?

compiler/src/iree/compiler/Dialect/Util/TransformOps/test/eliminate_hoistable_conversions.mlir
```diff
@@ -0,0 +1,247 @@
+// RUN: iree-opt --transform-interpreter --split-input-file %s | FileCheck %s
```
efric: IIUC this should also work with nested loops? If so, can we have a test showing that?

compiler/src/iree/compiler/Codegen/Common/HoistInnerTiledAccSwizzles.cpp
```diff
@@ -0,0 +1,142 @@
...
+  LogicalResult matchAndRewrite(IREE::Codegen::InnerTiledOp tiledOp,
+                                PatternRewriter &rewriter) const override {
+    auto loopLike = dyn_cast<LoopLikeOpInterface>(tiledOp->getParentOp());
+    if (!loopLike || loopLike.getRegionIterArgs().empty()) {
+      return rewriter.notifyMatchFailure(tiledOp,
+                                         "not inside a loop with iter_args");
+    }
+
+    bool anyOutputMatched = false;
+    for (int64_t outputIdx = 0, numOutputs = tiledOp.getOutputs().size();
```
efric:
```suggestion
    for (size_t outputIdx = 0, numOutputs = tiledOp.getOutputs().size();
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/Transforms.cpp
```diff
@@ -1614,35 +1653,61 @@ struct DropInnerTiledUnitDimsPattern
...
     auto newTiledOp = IREE::Codegen::InnerTiledOp::create(
-        rewriter, loc,
-        ValueRange{newOperands}.take_front(tiledOp.getNumInputs()),
-        ValueRange{newOperands}.drop_front(tiledOp.getNumInputs()),
+        rewriter, loc, ValueRange{newOperands}.take_front(numInputs),
+        ValueRange{newOperands}.drop_front(numInputs),
         rewriter.getAffineMapArrayAttr(emptyMaps), rewriter.getArrayAttr({}),
         tiledOp.getKind(), tiledOp.getSemantics());
 
+    // Pull extract / broadcast pairs out of the loop so the hardware
```
efric, line 1712: I think we should have a brief comment on the motivation of `HoistableConversionOp` in each use, like you do here.

efric (changes requested):
Fundamentally, I think this is well motivated but a few notes:

- The tags just being strings are convenient but imo fragile. I don’t have a good solution for this but perhaps at a minimum we can create a table of all pairs of tags/their inverses in one place. It will be helpful in preventing overlooked typos at least. If we really must stick with the way things are currently, then at least we should comment `/*tag*/` and `/*inversetag*/` or something like it when passing in the string args. 
- Regarding `HoistInnerTiledAccSwizzle.cpp`; I can see this getting the job done but the right approach might be to teach vector distribute about this. I wonder what @Groverkss or @sommerlukas thoughts are.
- I think we may be missing some tests when we stray off the happy path, or they are not shown with `--verify-diagnostics`. I identified a few places below but may be missing some.

compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.td
```diff
@@ -512,6 +512,94 @@ def Util_AssumeIntOp : Util_PureOp<"assume.int", [
...
+    OpBuilder<(ins
+      "StringRef":$tag,
+      "StringRef":$inverseTag,
+      "TypeRange":$resultTypes,
+      "ValueRange":$inputs,
+      "function_ref<SmallVector<Value>(OpBuilder &, Location, ValueRange)>":$bodyBuilder
+    )>,
+    OpBuilder<(ins
+      "StringRef":$tag,
+      "StringRef":$inverseTag,
```
qedawkins, line 594: String based tagging won't work. Here's an example of a valid sequence of transformations that can cause correctness issues:

Start with nested for loops:

```
%0 = ...
%3 = scf.for iter_arg(%outer = %0) {
  %1 = scf.for iter_arg(%inner_1 = %outer) {
    %inner_1_from = util.hoistable_conversion "from" inverts("to") %inner_1
    %inner_1_to = util.hoistable_conversion "to" inverts("from") %inner_1_from
    scf.yield %inner_1_to
  }
  %2 = scf.for iter_arg(%inner_2 = %1) {
    %inner_2_from = util.hoistable_conversion "from" inverts("to") %inner_2
    %inner_2_to = util.hoistable_conversion "to" inverts("from") %inner_2_from
    scf.yield %inner_2_to
  }
  scf.yield %2
}
```

Now if both inner for loops get canonicalized away in the event they are single-trip, we'd end up with:

```
%0 = ...
%3 = scf.for iter_arg(%outer = %0) {
  %inner_1_from = util.hoistable_conversion "from" inverts("to") %outer
  %inner_1_to = util.hoistable_conversion "to" inverts("from") %inner_1_from
  %inner_2_from = util.hoistable_conversion "from" inverts("to") %inner_1_to
  %inner_2_to = util.hoistable_conversion "to" inverts("from") %inner_2_from
  scf.yield %2
}
```

At which point the tags for the first loop's `from` and the second loop's `to` now match up, but they are completely unrelated per their construction. If you want to do this you have to either tie the two invertible operations together symbolically or via SSA.

krzysz00: ... I'd actually expect the "to" from the first loop to cancel with the "from" from the second loop, so I'd call that works-as-intended (assuming the types all line up etc.)

Heck, even if those weren't single iteration loops, if you hoist the conversions, the "to" from loop 1 will meet the "from" from loop two and those should cancel out.

We may want a better mechanism to prevent collisions between different passes that decide to use the same string ... but given how ephemeral these are right now, it's probably fine.

krzysz00: (though I'm all ears for systems that won't produce accidental collisions - but if you're a pass author and, say, "from" is unrolling a vector and "to" is rolling it back up, I'd want that to cancel between loops like shown)

compiler/src/iree/compiler/Dialect/Util/TransformOps/test/eliminate_hoistable_conversions.mlir
krzysz00: Added

compiler/src/iree/compiler/Dialect/Util/Transforms/DropCompilerHints.cpp
krzysz00: There is now

compiler/src/iree/compiler/Dialect/Util/Transforms/EliminateHoistableConversions.cpp
krzysz00: Done

compiler/src/iree/compiler/Codegen/Common/HoistInnerTiledAccSwizzles.cpp
krzysz00: I've gone and swapped this to "reshape" - I'm less unhappy with that name, and I hope you are too

compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.td
krzysz00: I've added a test that shows this behavior and make it explicit that this is meant to work

compiler/src/iree/compiler/Dialect/Util/TransformOps/test/eliminate_hoistable_conversions.mlir
```diff
@@ -0,0 +1,385 @@
...
+// CHECK: %[[SC0:.+]] = vector.shape_cast %[[INIT]]
+// CHECK: %[[RESULT1:.+]]:2 = scf.for {{.*}} %[[SC0]]
+// CHECK-NOT: vector.shape_castw
+// CHECK:   arith.addf
+// CHECK-NOT:   vector.shape_castw
+// CHECK:   scf.yield
+// CHECK: %[[RESULT2:.+]]:2 = scf.for {{.*}} %[[RESULT1]]#1
+// CHECK-NOT:   vector.shape_castw
+// CHECK:   arith.addf
+// CHECK-NOT:   vector.shape_castw
```
efric:
```suggestion
// CHECK-NOT: vector.shape_cast
// CHECK:   arith.addf
// CHECK-NOT:   vector.shape_cast
// CHECK:   scf.yield
// CHECK: %[[RESULT2:.+]]:2 = scf.for {{.*}} %[[RESULT1]]#1
// CHECK-NOT:   vector.shape_cast
// CHECK:   arith.addf
// CHECK-NOT:   vector.shape_cast
```

compiler/src/iree/compiler/Codegen/Common/HoistInnerTiledAccReshapes.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+static constexpr llvm::StringLiteral kAccReshapeTo = "acc_reshape_to_intrinsic";
+static constexpr llvm::StringLiteral kAccReshapeFrom =
+    "acc_reshape_from_intrinsic";
+
+namespace {
+
+struct WrapAccReshapesPattern final
+    : OpRewritePattern<IREE::Codegen::InnerTiledOp> {
+  using Base = OpRewritePattern<IREE::Codegen::InnerTiledOp>;
```
efric:
```suggestion

```

compiler/src/iree/compiler/Dialect/Util/Transforms/EliminateHoistableConversions.cpp
```diff
@@ -0,0 +1,354 @@
...
+      SmallVector<BlockArgument> oldIterArgsInNewLoop;
+      for (unsigned yi : yieldIndices) {
+        oldIterArgsInNewLoop.push_back(allNewLoopIterArgs[yi]);
+      }
+      for (auto [oldArg, converted] :
+           llvm::zip_equal(oldIterArgsInNewLoop, convertedBack)) {
+        rewriter.replaceAllUsesExcept(oldArg, converted, f);
+      }
+    }
+    rewriter.replaceOp(f, newIterArgs);
```
efric, line 289: not sure if we would get here atm; but is there a test for this?

efric (approved):
LGTM after nits are addressed. Thanks!

hanhanW (commented):
I'd appreciate if you can break this type of PR into smaller PRs in the future. It took me some time to navigate what I'm looking for in this big PR. If I read it correctly, it contains:
- Define new ops.
- Implement a core transformation (`eliminateHoistableConversions`) with the op.
- Use it in inner tiled op lowering. The lowering should not locate at IREEGPU dialect. I think we've been missing the code move after you move the InnerTiledOp to Codegen dialect.
- Add HoistInnerTiledAccReshapes pass and use it in GPU.

krzysz00:
Re where the lowering was, it stayed in IREEGPU because there wasn't anything requiring us to move it

(and yeah, this could maybe have been smaller PRs, you're right, but also then people get on you for adding ops without tests / a usecase)

hanhanW:
> Re where the lowering was, it stayed in IREEGPU because there wasn't anything requiring us to move it

If the op is moved to Codegen dialect, the transformations that do not have GPU specifics should be moved, IMO.
 
> (and yeah, this could maybe have been smaller PRs, you're right, but also then people get on you for adding ops without tests / a usecase)

You can add the op with the transformation; the GPU part can be a different PR. The other approach is using multiple commits, it has more burden so I never ask for such request in my review.

hanhanW:
(It is a good feature btw, and thanks for pushing on this.)

PR #23971, opened 2026-03-31
[Codegen] Support more operations in tile size analysis

sommerlukas:
Add support for `inner_tiled`, `linalg.pack` and `linalg.unpack` operations to the vector tile size analysis.

Assisted-by: Claude Code

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
@@ -161,6 +161,7 @@ public:
   /// Map tile sizes from pack source space (rank N) to pack dest space
   /// (rank N+K). Divides packed dims by inner tile sizes, applies
   /// outer_dims_perm, and appends inner tile sizes as new dimensions.
+  /// The static inner tiles are assumed to not be dynamic values.
   TileSizes mapPackSourceToDest(ArrayRef<int64_t> innerDimsPos,
                                 ArrayRef<int64_t> staticInnerTiles,
                                 ArrayRef<int64_t> outerDimsPerm) const {
@@ -170,6 +171,8 @@ public:
     TileSizes result = *this;
     for (auto [dimPos, tileSize] :
          llvm::zip_equal(innerDimsPos, staticInnerTiles)) {
+      assert(ShapedType::isStatic(tileSize) &&
+             "expecting static inner tile size");
       if (result.dims[dimPos] == kUninitialized ||
           result.dims[dimPos] == kOverdefined) {
         continue;
@@ -369,7 +372,8 @@ public:
       return success();
     }
 
-    // InnerTiledOp: propagate through indexing maps (outer dims only).
+    // InnerTiledOp: Propagate the outer dimensions through the indexing maps
+    // and then append the static inner dimensions.
     if (auto innerTiledOp = dyn_cast<IREE::Codegen::InnerTiledOp>(op)) {
       SmallVector<AffineMap> indexingMaps = innerTiledOp.getIndexingMapsArray();
       unsigned numLoops = indexingMaps[0].getNumDims();
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
@@ -6,6 +6,7 @@
 
 #include "iree/compiler/Codegen/Common/Passes.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
 #include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
 
 #include "llvm/Support/DebugLog.h"
@@ -157,6 +158,76 @@ public:
     return result;
   }
 
+  /// Map tile sizes from pack source space (rank N) to pack dest space
+  /// (rank N+K). Divides packed dims by inner tile sizes, applies
+  /// outer_dims_perm, and appends inner tile sizes as new dimensions.
+  TileSizes mapPackSourceToDest(ArrayRef<int64_t> innerDimsPos,
+                                ArrayRef<int64_t> staticInnerTiles,
+                                ArrayRef<int64_t> outerDimsPerm) const {
+    if (empty()) {
+      return {};
+    }
+    TileSizes result = *this;
+    for (auto [dimPos, tileSize] :
+         llvm::zip_equal(innerDimsPos, staticInnerTiles)) {
+      if (result.dims[dimPos] == kUninitialized ||
+          result.dims[dimPos] == kOverdefined) {
+        return {};
+      }
+      if (result.dims[dimPos] % tileSize != 0) {
+        return {};
+      }
+      result.dims[dimPos] /= tileSize;
+    }
+    if (!outerDimsPerm.empty()) {
+      applyPermutationToVector(result.dims, outerDimsPerm);
+    }
+    for (int64_t t : staticInnerTiles) {
+      result.dims.push_back(t);
+    }
+    return result;
+  }
+
+  /// Append dimensions from `suffix` to produce a higher-rank TileSizes.
+  TileSizes extend(ArrayRef<int64_t> suffix) const {
+    SmallVector<int64_t> fullDims(dims);
+    fullDims.append(suffix.begin(), suffix.end());
+    return TileSizes(fullDims);
+  }
+
+  /// Map tile sizes from pack dest space (rank N+K) back to source space
+  /// (rank N). Truncates to outer dims, applies inverse permutation, and
+  /// multiplies packed dims by inner tile sizes.
+  TileSizes mapPackDestToSource(unsigned sourceRank,
+                                ArrayRef<int64_t> innerDimsPos,
+                                ArrayRef<int64_t> staticInnerTiles,
+                                ArrayRef<int64_t> outerDimsPerm) const {
+    if (empty()) {
+      return {};
+    }
+    assert(sourceRank <= rank() && "sourceRank exceeds dest tile size rank");
+    TileSizes result(sourceRank);
+    for (unsigned i = 0; i < sourceRank; ++i) {
+      result.dims[i] = dims[i];
+    }
+    if (!outerDimsPerm.empty()) {
+      applyPermutationToVector(result.dims,
+                               invertPermutationVector(outerDimsPerm));
+    }
+    for (auto [dimPos, tileSize] :
+         llvm::zip_equal(innerDimsPos, staticInnerTiles)) {
+      if (result.dims[dimPos] == kUninitialized ||
+          result.dims[dimPos] == kOverdefined) {
+        // Uninitialized (0) stays 0 after multiply; overdefined is preserved
+        // by skipping. Safe because we only multiply here, unlike
+        // mapPackSourceToDest which must divide (and can't divide by zero).
+        continue;
+      }
+      result.dims[dimPos] *= tileSize;
+    }
+    return result;
+  }
+
   /// Lattice join: per-dimension merge. Uninitialized is identity.
   static TileSizes join(const TileSizes &lhs, const TileSizes &rhs) {
     TileSizes result = lhs;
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/test/materialize_vector_tile_sizes.mlir
hanhanW, line 1: I think we should have tests with dynamic shapes.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -157,6 +158,76 @@ class TileSizes {
...
+        return {};
+      }
+      if (result.dims[dimPos] % tileSize != 0) {
+        return {};
+      }
+      result.dims[dimPos] /= tileSize;
+    }
+    if (!outerDimsPerm.empty()) {
+      applyPermutationToVector(result.dims, outerDimsPerm);
+    }
```
hanhanW, line 185: We need a test for `outer_dims_perm`. Please use a permutation that is not identical to the inversed permutation. e.g., do not use `[1, 0]`.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -157,6 +158,76 @@ class TileSizes {
...
+    for (int64_t t : staticInnerTiles) {
+      result.dims.push_back(t);
+    }
```
hanhanW: `result.dims.append(staticInnerTiles)`?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -157,6 +158,76 @@ class TileSizes {
...
+         llvm::zip_equal(innerDimsPos, staticInnerTiles)) {
+      if (result.dims[dimPos] == kUninitialized ||
+          result.dims[dimPos] == kOverdefined) {
+        return {};
+      }
+      if (result.dims[dimPos] % tileSize != 0) {
+        return {};
+      }
+      result.dims[dimPos] /= tileSize;
+    }
```
hanhanW, line 182: Q1: Don't we need to check if `tileSize` is static or not?
Q2: don't we preserve `kUninitialized` and `kOverdefined` when it happens? Why do you return `{}`?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -157,6 +158,76 @@ class TileSizes {
...
+    return result;
+  }
+
+  /// Append dimensions from `suffix` to produce a higher-rank TileSizes.
+  TileSizes extend(ArrayRef<int64_t> suffix) const {
+    SmallVector<int64_t> fullDims(dims);
+    fullDims.append(suffix.begin(), suffix.end());
+    return TileSizes(fullDims);
+  }
```
hanhanW: How about naming it as `append` which matches the comment and methods from SmallVector?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -157,6 +158,76 @@ class TileSizes {
...
+                               invertPermutationVector(outerDimsPerm));
+    }
+    for (auto [dimPos, tileSize] :
+         llvm::zip_equal(innerDimsPos, staticInnerTiles)) {
+      if (result.dims[dimPos] == kUninitialized ||
+          result.dims[dimPos] == kOverdefined) {
+        // Uninitialized (0) stays 0 after multiply; overdefined is preserved
+        // by skipping. Safe because we only multiply here, unlike
+        // mapPackSourceToDest which must divide (and can't divide by zero).
+        continue;
```
hanhanW: > Uninitialized (0) stays 0 after multiply

This does not match the code... How about just saying that they are preserved? The last comment is not useful to me.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -304,6 +375,66 @@ class TileSizeForwardAnalysis
       return success();
     }
 
+    // InnerTiledOp: propagate through indexing maps (outer dims only).
```
hanhanW: Why do we exclude inner dims? We will need it when people switch CPU data-tiling to inner_tiles and that is required by scalable vectors and future RVV work.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -426,22 +629,61 @@ class MaterializeVectorTileSizesPass final
       return signalPassFailure();
     }
 
-    funcOp->walk([&](linalg::LinalgOp linalgOp) {
-      TileSizes tileSizes = getIterationSpaceTileSizes(linalgOp, solver);
-      if (tileSizes.isOverdefined()) {
-        linalgOp.emitOpError()
-            << "tile size analysis did not determine a valid tile size";
+    auto materialize =
+        [&](Operation *op, TileSizes tileSizes) {
```
hanhanW: Do you need to capture anything?

```suggestion
        [](Operation *op, TileSizes tileSizes) {
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -383,20 +586,20 @@ class TileSizeBackwardAnalysis
 // Result querying
 //===----------------------------------------------------------------------===//
 
-/// Gather tile sizes into the iteration space of a linalg op by looking up each
+/// Gather tile sizes into the iteration space of an op by looking up each
 /// operand's lattice state in the solver.
-static TileSizes getIterationSpaceTileSizes(linalg::LinalgOp linalgOp,
+static TileSizes getIterationSpaceTileSizes(Operation *op, unsigned numLoops,
+                                            ArrayRef<AffineMap> indexingMaps,
                                             const DataFlowSolver &solver) {
```
hanhanW, line 596: I feel that you want to pass in `TilingInterface` op which has the additional information.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -426,22 +629,61 @@ class MaterializeVectorTileSizesPass final
...
+
+      if (auto innerTiledOp = dyn_cast<IREE::Codegen::InnerTiledOp>(op)) {
+        SmallVector<AffineMap> indexingMaps =
+            innerTiledOp.getIndexingMapsArray();
+        unsigned numLoops = indexingMaps[0].getNumDims();
+        TileSizes tileSizes =
+            getIterationSpaceTileSizes(op, numLoops, indexingMaps, solver);
+        materialize(op, tileSizes);
         return;
       }
```
hanhanW, line 686: Like the other comment, I feel that we should structure these like:

```
// Special cases like pack/unpack
// TilingInterface op. Or isa<linalg::LinalgOp, InnerTiledOp>
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -426,22 +629,61 @@ class MaterializeVectorTileSizesPass final
...
-      TileSizes tileSizes = getIterationSpaceTileSizes(linalgOp, solver);
-      if (tileSizes.isOverdefined()) {
-        linalgOp.emitOpError()
-            << "tile size analysis did not determine a valid tile size";
+    auto materialize =
+        [&](Operation *op, TileSizes tileSizes) {
+          if (tileSizes.isOverdefined()) {
+            op->emitOpError()
+                << "tile size analysis did not determine a valid tile size";
+            return;
```
hanhanW: I missed this when you landed the previous PR. I was not clear enough. I think the pass should signal failure if it is overdefined. Emitting error messages is bundled with pass failure to me, and I did not explicitly spell it out in my other comment.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -157,6 +158,76 @@ class TileSizes {
...
+    for (auto [dimPos, tileSize] :
+         llvm::zip_equal(innerDimsPos, staticInnerTiles)) {
+      if (result.dims[dimPos] == kUninitialized ||
+          result.dims[dimPos] == kOverdefined) {
+        return {};
+      }
+      if (result.dims[dimPos] % tileSize != 0) {
+        return {};
+      }
+      result.dims[dimPos] /= tileSize;
```
hanhanW: I thought that it is always `ceilDiv` in this case, because pack have padding semantics?

hanhanW (changes requested):
I feel that dynamic cases are not exercised enough, because I'm seeing weird assumption in the implementation. E.g., the inlined comment in `mapPackSourceToDest`.

Not saying dynamic inner tile sizes, the inner tile sizes are static in the scope. The outer dimension can be dynamic and the op can carry padding value.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -157,6 +158,76 @@ class TileSizes {
     return result;
   }
 
+  /// Map tile sizes from pack source space (rank N) to pack dest space
+  /// (rank N+K). Divides packed dims by inner tile sizes, applies
+  /// outer_dims_perm, and appends inner tile sizes as new dimensions.
+  TileSizes mapPackSourceToDest(ArrayRef<int64_t> innerDimsPos,
```
keshavvinayak01, line 165: There doesn't seem to be a lit test that exercises the `outer_dims_perm` logic with `applyPermutationToVector`

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: I've switched to using ceil division, but in the same course also deactivated backward propagation for `pack` with padding.

That propagation would in the padding case would otherwise propagate too large tile sizes (or lead to overdefined).

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: The comment isn't phrased very well. We don't exclude the inner dims (we append them before propagating, see below), but they don't participate in the propagation through indexing maps (because they don't need to). I've clarified the comment. We also don't materialize the inner dims in the attribute on `inner_tiled`, because it would duplicate the information.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: Clarified.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: Q1: This is already checked at the call-sites.
Q2: Changed to preserve.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: Added proper pass failure.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: IIUC, `TilingInterface` doesn't give us access to the indexing maps, so we won't gain much here.

sommerlukas (commented):
Thanks for the feedback! I've added tests for dynamic shapes and `outer_dims_perm`, some more comments inline.

keshavvinayak01 (approved):
LG

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
hanhanW: RE Q1: If we look the code from class's perspective, it is better to document the assumption in method comment, or we should just handle it here.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
hanhanW: > I've clarified the comment. 

I don't see the update on the comment. Do you miss the change?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -540,7 +534,12 @@ class TileSizeBackwardAnalysis
 
     // Pack ops: result tile sizes → source tile sizes (backward).
     if (auto packOp = dyn_cast<linalg::PackOp>(op)) {
-      if (llvm::any_of(packOp.getStaticInnerTiles(), ShapedType::isDynamic)) {
+      if (ShapedType::isDynamicShape(packOp.getStaticInnerTiles())) {
+        return success();
+      }
+      if (packOp.getPaddingValue()) {
```
hanhanW, line 544: Can we have a test for padding value?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
hanhanW: Oh, I missed that. I was going to suggest `IndexingMapOpInterface`, then I found that InnerTileOp haven't implemented the interface, so it is okay for now.

hanhanW (approved):

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: I've documented the assumption in the method comment and also added an `assert`.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: Should be updated now, must have missed pushing the update.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: I've added a test.

Groverkss (approved):
lgtm

PR #23973, opened 2026-03-31
[GPU] Implement CDNA block intrinsics (3/6)

nirvedhmeshram:
Extend LLVMGPUConfigureTensorLayouts to handle CDNA block intrinsics, enabling VectorDistribute support.

Part of iree-org/iree#23941


krzysz00 (approved):
I can't help but wonder if we shouldn't just give everything a "block dimension" of 1 but that's not important and can be fixed later, looks correct.

PR #23974, opened 2026-03-31
Supporting i64 in cf.switch lowering (from llvm/llvm-project#189230).

### LINKED ISSUE #23972: Index size mismatch in VM branch table
https://github.com/llvm/llvm-project/pull/189230 changed the bit-width of the `cf.index_switch` index from `i32` to `i64`, which causes a mismatch with the IREE VM branch table. 

Example error message: 
```
vm.br_table' op operand #0 must be 32-bit signless integer, but got 'i64'
```

Example CI run with the error: https://github.com/iree-org/iree/actions/runs/23795532833/job/69341730033?pr=23970

I've dropped the responsible upstream commit from `third-party/llvm-project` in https://github.com/iree-org/llvm-project/commit/798b31f3ae6d1f9ea01f4c6240ae80ac373c6f64 for now. After this issue is fixed, we should drop this revert in LLVM integration.

benvanik:
Fixes #23972.


sommerlukas (approved):
LGTM.

PR #23978, opened 2026-03-31
[Codegen][GPU] Use gpu::AddressSpace::Global for constants

bangtianliu:
This PR fixes crashes on ROCM when constants are accessed with dynamic indexing, which is triggered by split reduction.  

Below the MLIR is what we found with crashes: bufferized constants placed in the default flat address space (0) cause crashes during code generation when accessed with dynamic indexing. This occurs because AMDGPU's flat vector load codegen cannot handle dynamic indices on constants in flat memory.

```mlir
// Constant bufferized to default address space (0).
memref.global "private" constant @__constant_4x15xi32 : memref<4x15xi32> = dense<...>

func.func @split_reduction(%arg0: ...) {
  %cst = memref.get_global @__constant_4x15xi32 : memref<4x15xi32>

  // Split reduction: parallel workgroups process different slices.
  scf.forall (%wg_id) in (%num_wgs) {
    // Dynamic index computed from workgroup ID (trigger crash).
    %val = memref.load %cst[%wg_id, %i] : memref<4x15xi32>
    ...
  }
}
```

After this PR

```mlir
// Constant bufferized to global address space for ROCM.
memref.global "private" constant @__constant_4x15xi32 : memref<4x15xi32, #gpu.address_space<global>> = dense<...>

func.func @split_reduction(%arg0: ...) {
  %cst = memref.get_global @__constant_4x15xi32 : memref<4x15xi32, #gpu.address_space<global>>

  // Split reduction with global address space.
  scf.forall (%wg_id) in (%num_wgs) {
    %val = memref.load %cst[%wg_id, %i] : memref<4x15xi32, #gpu.address_space<global>>
    ...
  }
}
```

Refer to gist for detailed repro mlir: https://gist.github.com/bangtianliu/ef9eb622998237e63fe25f5770cda5da, and also you need to comment out below code: https://github.com/iree-org/iree/blob/c0ec02880cae915d63547c1eb491d024a555fad1/compiler/src/iree/compiler/DispatchCreation/SetSplitReductionSizes.cpp#L104-L108 to enable split reduction for the <4x15xi32> case. 

Note: The PR extends the fix to all GPU backends.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/IREEComprehensiveBufferize.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/IREEComprehensiveBufferize.cpp
@@ -278,6 +278,18 @@ void IREEComprehensiveBufferizePass::runOnOperation() {
   // data races on GPU.
   options.checkParallelRegions = false;
 
+  // ROCM needs global address space for constants to avoid invalid opcodes
+  // when accessed with dynamic indexing (e.g., split reduction).
+  if (auto target = IREE::HAL::ExecutableTargetAttr::lookup(funcOp)) {
+    if (target.getBackend() == "rocm") {
+      options.defaultMemorySpaceFn =
+          [](TensorType t) -> std::optional<Attribute> {
+        return gpu::AddressSpaceAttr::get(t.getContext(),
+                                          gpu::AddressSpace::Global);
+      };
+    }
+  }
+
   bufferization::BufferizationState bufferizationState;
   if (failed(runIREEOneShotBufferize(funcOp, options, bufferizationState))) {
     return signalPassFailure();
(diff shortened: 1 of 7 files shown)


compiler/src/iree/compiler/Codegen/Common/IREEComprehensiveBufferize.cpp
```diff
@@ -294,9 +294,29 @@ void IREEComprehensiveBufferizePass::runOnOperation() {
...
+  // default flat address space (0).
+  auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(op);
+  bool isROCM = isROCMBackend(targetAttr);
+
   mlir::bufferization::OneShotBufferizationOptions opt;
   opt.copyBeforeWrite = true;
   opt.opFilter.allowOperation(arith::ConstantOp::getOperationName());
+
+  // For ROCM, use gpu::AddressSpace::Global for constants.
+  opt.defaultMemorySpaceFn = [ctx,
```
krzysz00: Tempted to suggest that we go to the generality of teaching `ExecutableTargetAttr` to `getConstantMemorySpace()` but that can be future work

krzysz00 (approved):
Ok, so, this _should_ work, but I do want to observe the existence of address space 4 - see https://llvm.org/docs/AMDGPUUsage.html#address-spaces - which is meant to be used specifically for constants. We may want to duck upstream and add either `#gpu.address_space<constant>` (mapping to global for platforms that don't have one) or maybe a `#amdgpu.address_space<constant>`.

But there's nothing wrong with just making them globals, I think, so let's roll with it for now?

krzysz00:
And to be fair to the backend, a constant in flat memory is ambiguous as to whethere it's on the stack, in LDS, in global, etc etc.

krzysz00:
... I'd suggest we probably want the same logic for other GPU targets

bangtianliu:
> Ok, so, this _should_ work, but I do want to observe the existence of address space 4 - see https://llvm.org/docs/AMDGPUUsage.html#address-spaces - which is meant to be used specifically for constants. We may want to duck upstream and add either `#gpu.address_space<constant>` (mapping to global for platforms that don't have one) or maybe a `#amdgpu.address_space<constant>`.
> 
> But there's nothing wrong with just making them globals, I think, so let's roll with it for now?

Yeah, my main concern is the size of the constant memory. That is why I went for the current implementation.

krzysz00:
> Yeah, my main concern is the size of the constant memory.

I'm pretty sure this'll have the same size as a global?

bangtianliu:
> > Yeah, my main concern is the size of the constant memory.
> 
> I'm pretty sure this'll have the same size as a global?

I mean the GPU specification. 
Constant memory (address space 4) is much more limited. On AMDGPU, it uses a dedicated hardware constant cache that's typically limited to ~64KB per kernel, while global memory uses full device VRAM (GBs).

krzysz00:
Huh. I didn't know about the constant cache. Best to stick the things in global theni

bangtianliu:
Our fix changes constants to gpu::AddressSpace::Global, but downstream passes don't know how to use constants with global address space. They expect flat address space. This causes memref.copy + memref.alloc() to be inserted to copy data from global to flat address space, and memref.alloc() on GPU becomes malloc. I will still need to see an appropriate fix for this problem, and I will convert this PR to draft now.

krzysz00 (approved):
Approved

bangtianliu:
cc @qedawkins for a look

qedawkins (approved):

Issue #23976, opened 2026-03-31
Improve backward data convolution with non-unit strides


yzhang93:
## Problem

Backward-data convolutions with non-unit strides (stride 2, stride 3) produce a strided `tensor.insert_slice` into a zero-filled buffer, followed by a convolution that reads the scattered result. On the current main branch this lowers to:

1. **Memset** — zero-fill the output buffer via DMA
2. **slow_memcpy** — copy the source elements at strided positions via DMA
3. **Conv dispatch** — run the convolution on the padded buffer

The `slow_memcpy` dispatch uses hardware DMA copy engines, which have high bandwidth for moderate tensor sizes (~100–200 MB) but degrade significantly for large tensors (>1 GB). The `tensor.insert_slice` cannot be fused into the convolution dispatch, so the scatter always runs as a separate dispatch before the convolution. For 1x1 backward convolutions this is particularly wasteful because the scatter commutes with the contraction and could be deferred to after it, operating on a much smaller result tensor.

### Current IR after dispatch formation

**1x1 backward conv** (`convfp16 -n 32 -c 2048 -k 1024 -y 1 -x 1 -u 2 -v 2 -g 1`):
```mlir
// IR after FormDispatchRegionsPass — insert_slice stays outside the dispatch
util.func public @...(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer_view, %arg3: !hal.fence, %arg4: !hal.fence) -> !hal.buffer_view {
  %cst = arith.constant 0.000000e+00 : f32
  %cst_0 = arith.constant dense<0.000000e+00> : tensor<32x50x50x2048xf16>
  %0 = hal.tensor.import wait(%arg3) => %arg0 : !hal.buffer_view -> tensor<32x25x25x2048xf16>
  %1 = iree_tensor_ext.compute_barrier.start %0 : tensor<32x25x25x2048xf16> -> tensor<32x25x25x2048xf16>
  %2 = hal.tensor.import wait(%arg3) => %arg2 : !hal.buffer_view -> tensor<2048x1x1x1024xf16>
  // insert_slice becomes Memset + slow_memcpy — not fused with conv dispatch
  %inserted_slice = tensor.insert_slice %1 into %cst_0[0, 0, 0, 0] [32, 25, 25, 2048] [1, 2, 2, 1] : tensor<32x25x25x2048xf16> into tensor<32x50x50x2048xf16>
[...]

PR #23979, opened 2026-03-31
[Preprocessing] Add pass to swap strided insert_slice with contraction

yzhang93:
Add `iree-preprocessing-swap-strided-insert-slice-with-contraction` pass that swaps a strided `tensor.insert_slice` into zeros with its consumer contraction for 1x1 backward convolutions.

For 1x1 backward data convolutions the strided insert_slice commutes with the contraction because the reduction dimensions have loop bound 1. This pass reorders the computation: the contraction operates on the small un-scattered source (e.g., 2x8x8 instead of 2x18x18 for stride 2), then the result is scattered into the output. This reduces contraction FLOPs by ~4x.

This PR is a partial implementation for https://github.com/iree-org/iree/issues/23976.

### DIFF
--- a/compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
+++ b/compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
@@ -126,40 +126,45 @@ public:
     unsigned rank = scatteredTy.getRank();
     Location loc = insertOp.getLoc();
 
-    // For strided dims, the indexing map must reference exactly one parallel
-    // loop dim; any other dims must be unit-range reductions. This accepts
-    // 1x1 convs (d_spatial + d_kernel with kernel=1) while rejecting 3x3.
+    // For strided dims, the indexing map expression must be equivalent to a
+    // single parallel loop dim (after zeroing out unit-range reduction dims).
+    // This accepts 1x1 convs (d_spatial + d_kernel with kernel=1, which
+    // simplifies to d_spatial) while rejecting 3x3, scaled dims (3*d0), or
+    // sums of parallel dims (d0 + d1).
+    MLIRContext *ctx = rewriter.getContext();
+    DenseMap<AffineExpr, AffineExpr> unitReductionMap;
+    for (unsigned i = 0, e = iterTypes.size(); i < e; ++i) {
+      if (iterTypes[i] == utils::IteratorType::reduction &&
+          i < loopRanges.size() && loopRanges[i] == 1) {
+        unitReductionMap[getAffineDimExpr(i, ctx)] =
+            getAffineConstantExpr(0, ctx);
+      }
+    }
+
     SmallVector<int64_t> newResultShape(resultTy.getShape());
     for (unsigned d = 0; d < rank; d++) {
       if (strides[d] == 1) {
         continue;
       }
-      int parallelDim = -1;
-      bool valid = true;
-      scatterMap.getResult(d).walk([&](AffineExpr e) {
-        if (auto dim = dyn_cast<AffineDimExpr>(e)) {
-          unsigned pos = dim.getPosition();
-          if (iterTypes[pos] == utils::IteratorType::parallel) {
-            parallelDim = pos;
-          } else if (pos >= loopRanges.size() || loopRanges[pos] != 1) {
-            valid = false;
-          }
-        }
-      });
-      if (!valid || parallelDim < 0) {
+      AffineExpr expr = scatterMap.getResult(d).replace(unitReductionMap);
+      auto dimExpr = dyn_cast<AffineDimExpr>(expr);
+      if (!dimExpr) {
+        return failure();
+      }
+      unsigned parallelDim = dimExpr.getPosition();
+      if (parallelDim >= iterTypes.size() ||
+          iterTypes[parallelDim] != utils::IteratorType::parallel) {
         return failure();
       }
       // Compute new result shape by mapping parallel dim through the result
       // map.
       for (auto [resIdx, resExpr] : llvm::enumerate(resultMap.getResults())) {
         auto resDimExpr = dyn_cast<AffineDimExpr>(resExpr);
-        if (resDimExpr &&
-            resDimExpr.getPosition() == static_cast<unsigned>(parallelDim)) {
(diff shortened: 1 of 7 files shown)


MaheshRavishankar:
@zjgarvey  can you please review this.

compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+    // Clone the chain of single-input elementwise consumers (e.g., truncf)
+    // for the smaller result shape.
+    SmallVector<Operation *> consumerChain;
+    Operation *lastOp = genericOp;
+    while (lastOp->hasOneUse()) {
+      auto genericUser = dyn_cast<linalg::GenericOp>(*lastOp->user_begin());
+      if (!genericUser || genericUser.getNumDpsInputs() != 1 ||
+          genericUser.getNumDpsInits() != 1 ||
+          genericUser.getNumReductionLoops() != 0) {
```
zjgarvey, line 229: Would a transpose satisfy this? You might need to check that the indexing maps for the unary ops are the identity here for the shape logic to be valid below (or handle permutations).

compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    // The insert_slice result must feed a contraction with single use.
+    if (!insertOp->hasOneUse()) {
+      return failure();
+    }
+    Value insertResult = insertOp.getResult();
+    auto genericOp = dyn_cast<linalg::GenericOp>(*insertOp->user_begin());
+    if (!genericOp || genericOp.getNumDpsInputs() != 2 ||
+        genericOp.getNumDpsInits() != 1 ||
```
zjgarvey, line 90: I think it's necessary for the generic op to be a contraction (mul with sum accumulate/reduction). Anything else could be incorrect.

zjgarvey (commented):
Two possible edge cases I noticed (must check it's a contraction, and should verify consumer elementwise chain isn't permuting dims) otherwise the core logic seems right.

compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
yzhang93: Good point. I have added checks for these corner cases.

compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
```diff
@@ -0,0 +1,302 @@
...
+          if (iterTypes[pos] == utils::IteratorType::parallel) {
+            parallelDim = pos;
+          } else if (pos >= loopRanges.size() || loopRanges[pos] != 1) {
+            valid = false;
+          }
+        }
+      });
+      if (!valid || parallelDim < 0) {
+        return failure();
+      }
```
zjgarvey: I'm trying to parse the logic here.

Let's say we have an expression `E` for one of these strided scatter dims:

If I understand correctly, `E` needs to be very specific for this commute logic to be valid. 

E.g., if the insert slice was by a stride of 2, and `E = 3*d0`, this should not commute, but if I parse the logic here correctly, I think this is allowed (I forget how the walk works, but I'm pretty sure it would get down to `d0` eventually and be like "yup that's parallel")? 

What about `E = d0 + d1` where both `d0` and `d1` are parallel dims? This doesn't commute with insert slice either, but passes the check?

I think the only valid indexing map is something which is actually equivalent to `E = <parallel dim expr>` (e.g., adding a reduction loop of iteration size 1 is essentially doing `parallel_dim + 0`. I'm not sure the best way to do the check, but functionally, it's necessary for these ops to commute.

compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
```diff
@@ -0,0 +1,302 @@
...
+    for (OpOperand *input : genericOp.getDpsInputOperands()) {
+      newInputs.push_back(input == scatterOperand ? source : input->get());
+    }
+    Value currentResult =
+        cloneGenericWithShape(genericOp, newInputs, ValueRange{newFill},
+                              newResultTy)
+            .getResult(0);
+
+    // Clone the chain of single-input elementwise consumers (e.g., truncf)
+    // for the smaller result shape.
```
zjgarvey, line 222: Only being overly cautious here because there are legitimately awful linalg generic ops being generated out there. 

E.g. a generic op that does something like a flip + elementwise add could look like a unary op.

```
external_tensor (same shape as *scattered* tensor)
ins (lastOp) outs (init) {
    for i < rank
         index_i = size_external_tensor_dim_i - 1 - linalg.index(i) (flipped indices)
    yield arg0 + tensor.extract(external_tensor, [index_0, index_1, ...])
}
```

Banning `linalg.index` in the payload is probably a good idea here in general, since that kind of logic is no-longer valid on the smaller result shape.

zjgarvey (commented):
A few more things I've noticed. I'll finish reviewing after an appointment.

compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
yzhang93: Agreed, added a check that bails out of the consumer chain if the body contains any linalg::IndexOp.

compiler/src/iree/compiler/Preprocessing/Common/SwapStridedInsertSliceWithContraction.cpp
yzhang93: Good catch! The walk was indeed too permissive. Replaced it with AffineExpr::replace that substitutes unit-range reduction dims with 0, then checks the result is a bare AffineDimExpr.

zjgarvey (approved):
LGTM

PR #23980, opened 2026-03-31
[LLVMGPU] Switch LLVMGPU to module-scope pipeline.

hanhanW:
Similar to https://github.com/iree-org/iree/commit/a867e93fb1e4bc90ebdad998c3e0fceca68e6c06, the revision moves the variant-scope passes to the boundary of pipelines, so the LLVMGPU codegen pipeline becomes a pure module-scope pipeline.

Additionally, the `HoistExecutableObjectsPass` is not needed anymore after we revisit the pipeline. Because the attribute is preserved since https://github.com/iree-org/iree/commit/03c158ee3c427d9724882d8468c63fdbd1238d68, so `llvm.call` op has the attribute and the HAL pipeline's HoistExecutableObjectsPass runs right between linking and serialization finds them and moves them to the variant.

To achieve the module-scope pipeline, some addtional changes are required:
- Update ROCDLAnnotateKernelForTranslationPass to work on DispatchConfigOp instead of HAL::ExecutableExportOp.
- Add a mode to use DispatchConfigOp in PropagateDispatchSizeBounds. The option is needed because the pass is also used by SPIR-V and it will be retired once SPIR-V migrate to module-scope pipeline.


hanhanW:
@bjacob I just found that `HoistExecutableObjectsPass` is no longer needed, see the PR description, FYI.

qedawkins (approved):

hanhanW:
Looking at the ukernel stuff, there is a failure: https://github.com/iree-org/iree/actions/runs/23824661146/job/69445510431?pr=23980

Maybe I was wrong on the `HoistExecutableObjectsPass`

PR #23983, opened 2026-04-01
Add vae tests to torch models

yash-amd:
Add vae tests to torch model tests( as part of moving sharktank model tests to torch)

### DIFF
--- a/tests/external/iree-test-suites/torch_models/sdxl/modules/vae_cpu.json
+++ b/tests/external/iree-test-suites/torch_models/sdxl/modules/vae_cpu.json
@@ -1,0 +1,10 @@
+{
+    "mlir": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-vae-decode/model.mlir",
+    "compiler_flags": [
+        "--iree-hal-local-target-device-backends=llvm-cpu",
+        "--iree-hal-target-device=local",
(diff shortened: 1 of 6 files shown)


amd-eochoalo (approved):
I think this is ok.

For some context, I believe this PR is motivated by @MaheshRavishankar's [comment on discord:](https://discord.com/channels/689900678990135345/689957613152239638/1468300246282010807)

>  how many models do we have in sharktank... i think we should try to move them to just use torch tests....

yash-amd:
in the [torch model tests](https://github.com/iree-org/iree/tree/main/tests/external/iree-test-suites/torch_models), llama 8b fp16, llama 8b fp8 and sdxl models have been added for numerics and benchmarking, do we need to add any other tests?
cc @MaheshRavishankar

MaheshRavishankar (commented):
Could we change the place where the expected inputs/outputs are stored to be consistent with the other tests.

Groverkss:
LGTM, after Mahesh's comment is addressed

yash-amd:
> Could we change the place where the expected inputs/outputs are stored to be consistent with the other tests.

moved all the artifacts to iree-test-suites/torch-models/ in sharkpublic. 
I have tested it locally, can anyone please retrigger the CI workflows on this pr(as i don't have access to it).

MaheshRavishankar (approved):

PR #23986, opened 2026-04-01
[LinalgExt] Add optional logsumexp output to AttentionOp

keshavvinayak01:
Add an optional logsumexp DPS init operand to iree_linalg_ext.attention for training backward pass support. When present, the lowering computes logsumexp = max * ln(2) + log(sum) from online_attention's intermediate max and sum statistics.

This is needed for SDPA generate_stats support in hipDNN/fusilli (iree-org/fusilli#275).

Both `mergeReductions` and `decomposeAttention` perform the `logsumexp` from the partial results, but the GPU path eliminates the dead code from `decomposeAttention`  since `mergeReductions` also computes the same from all the partial results.


keshavvinayak01:
Alternative strategy to remove `AttentionOp` and just use the `logsumexp` generated from `OnlineAttentionOp`

PR #23988, opened 2026-04-01
[LLVMGPU] Outer Reductions should use TileAndFuse Pipeline for better performance

YashDeshpande25:
This PR adds a check to select LLVMGPUTileAndFuse Pipeline when dealing with an outer reduction. This change is based on observed performance improvement when using TileAndFuse against VectorDistribute for outer reductions. 

### On MI300X
- 21 shapes improved (by avg 105us)
- 0 regressions
- 10 shapes showed minimal change

### On MI355

- 20 shapes improved (by avg 101us ~43%)
- 0 regressions
- 11 shapes showed minimal change

All comparisons were made against the VectorDistribute timings on the respective machines.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
@@ -122,6 +122,12 @@ static llvm::cl::opt<bool> clDirectConvolution(
     llvm::cl::desc("Use direct convolution in tile and fuse pipeline"),
     llvm::cl::init(false));
 
+static llvm::cl::opt<bool> clGPUEnableOuterReductionTileAndFuse(
+    "iree-codegen-llvmgpu-enable-outer-reduction-tile-and-fuse",
+    llvm::cl::desc(
+        "Enable usage of tile and fuse pipeline for outer reduction"),
+    llvm::cl::init(true));
+
 // Custom parser for llvm::cl::opt<std::optional<uint64_t>>. Allows a flag to
 // be truly optional: unset on the command line means std::nullopt, while a
 // user-provided non-negative integer is stored in the optional.
@@ -2338,6 +2344,32 @@ static LogicalResult setConvolutionConfig(
 // Pipeline Configuration
 //====---------------------------------------------------------------------===//
 
+static bool isOuterReduction(linalg::LinalgOp op) {
+  if (!op.getNumReductionLoops()) {
+    return false;
+  }
+  SmallVector<unsigned> reductionDims;
+  op.getReductionDims(reductionDims);
+  for (OpOperand *input : op.getDpsInputOperands()) {
+    AffineMap map = op.getMatchingIndexingMap(input);
+    if (map.getNumResults() == 0) {
+      continue;
+    }
+    if (!map.isPermutation()) {
+      continue;
+    }
+    auto firstResult = dyn_cast<AffineDimExpr>(map.getResult(0));
+    if (!firstResult) {
+      continue;
+    }
+    unsigned dimPos = firstResult.getPosition();
+    if (llvm::is_contained(reductionDims, dimPos)) {
+      return true;
+    }
+  }
+  return false;
+}
+
 static LogicalResult setRootConfig(IREE::GPU::TargetAttr target,
                                    mlir::FunctionOpInterface entryPointFn,
                                    Operation *computeOp) {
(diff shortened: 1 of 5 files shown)


Groverkss:
I think we need to talk about this a bit more

Groverkss:
> On MI355
20 shapes improved (by avg 101us ~43%)
0 regressions
11 shapes showed minimal change
All comparisons were made against the default TileAndFuse timings on the respective machines.

Can you post specific inputs?

YashDeshpande25:
> > On MI355
> > 20 shapes improved (by avg 101us ~43%)
> > 0 regressions
> > 11 shapes showed minimal change
> > All comparisons were made against the default TileAndFuse timings on the respective machines.
> 
> Can you post specific inputs?

Discussed in meeting.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -2392,6 +2418,13 @@ static LogicalResult setRootConfig(IREE::GPU::TargetAttr target,
       LDBG() << "Contract Config";
       return success();
     }
+    if (isOuterReduction(linalgOp)) {
```
nirvedhmeshram: Lets put this behind a default true flag like `clGPUEnableOuterReductionTileAndFuse`

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -2338,6 +2338,32 @@ static LogicalResult setConvolutionConfig(
 // Pipeline Configuration
 //====---------------------------------------------------------------------===//
 
+static bool isOuterReduction(linalg::LinalgOp op) {
```
nirvedhmeshram, line 2347: Please add a positive and a negative config tests for this logic, since you are allowing multiple reduction dims, I would test something like this: an operand with dims  (reduction, parallel, reduction). I believe with this logic, as long as every operand has a outer reduction dim, we are going with TileAndFuse, which (to me) is a fine choice but lets make a note of that with a config test. And the negative case can be one operand has a outer reduction dim but another has a outer parallel dim so we bail out for that one.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_outer_reduction.mlir
```diff
@@ -0,0 +1,76 @@
...
+    ^bb0(%arg0: f32, %arg1: f32):
+      %add = arith.addf %arg0, %arg1 : f32
+      linalg.yield %add : f32
+  } -> tensor<16384xf32>
+  iree_tensor_ext.dispatch.tensor.store %result, %1, offsets = [0], sizes = [16384], strides = [1]
+      : tensor<16384xf32> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<16384xf32>>
+  return
+}
+// CHECK-LABEL: func.func @inner_reduction
+// CHECK-SAME:    #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<VectorDistribute> workgroup_size = [64, 1, 1] subgroup_size = 64
```
nirvedhmeshram: we prefer to minimize tests where possible, therefore here you can just do the CHECK-LABEL followed by
// CHECK-NOT: TileAndFuse

nirvedhmeshram (approved):
Just a small nit, looks good overall.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
YashDeshpande25: Yes, as a note. Currently, it should trigger a TileAndFuse for any one of the operand having Outer Reductions. The tests that I have added are one for outer reduction & one for inner reduction. 
Currently I haven't added any case with both within a single op but due to the presence of the outer reduction, such a case should also trigger TileAndFuse.

nirvedhmeshram: Yeah I think the current tests are fine

Groverkss:
> > > On MI355
> > > 20 shapes improved (by avg 101us ~43%)
> > > 0 regressions
> > > 11 shapes showed minimal change
> > > All comparisons were made against the default TileAndFuse timings on the respective machines.
> > 
> > 
> > Can you post specific inputs?
> 
> Discussed in meeting.

Can you post specific shapes on this issue and post benchmarks on what the original perf was and the new perf? It makes it easier for me to track it.

YashDeshpande25:
> > > > On MI355
> > > > 20 shapes improved (by avg 101us ~43%)
> > > > 0 regressions
> > > > 11 shapes showed minimal change
> > > > All comparisons were made against the default TileAndFuse timings on the respective machines.
> > > 
> > > 
> > > Can you post specific inputs?
> > 
> > 
> > Discussed in meeting.
> 
> Can you post specific shapes on this issue and post benchmarks on what the original perf was and the new perf? It makes it easier for me to track it.

Sure, I have shared the performance numbers with the shapes on the private discord communication as the shapes are not public.

PR #23989, opened 2026-04-01
[TensorExt] Add resolveRange and getDistributionInfo to SparseCastOpInterface 

MaheshRavishankar:
This is PR 2/N of landing block sparse changes that are captured in this
branch :
https://github.com/MaheshRavishankar/iree/tree/users/MaheshRavishankar/blockSparseCommits

Extends SparseCastOpInterface with two new methods:                                                         
  - `resolveRange`: Resolves a range from sparse result space to source                                       
    tensor space by computing linearized offsets via column_lengths.                                          
    Accepts an inBounds bit vector and optional padding value for future                                      
    out-of-bounds guard generation.                                                                           
  - `getDistributionInfoForSparseDimensions`: Returns which sparse                                            
    dimensions can be distributed across workgroups. For ragged tensors,                                      
    only the outer sparse dimension is distributable since the inner                                          
    dimension bounds depend on the outer IV.                                                                  
                                                                                                              
Also refactors `lowerLoopRange` to support lowering individual sparse                                       
dimensions independently (outer-only or both), and switches the inner                                       
loop to iterate in relative index space (0..columnLength) rather than                                       
absolute linearized space.
                                                                    
Signed-off-by: MaheshRavishankar <mahesh.ravishankar@gmail.com>

### DIFF
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtAttrs.h
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtAttrs.h
@@ -17,8 +17,11 @@
 
 namespace mlir::iree_compiler::IREE::TensorExt {
 
-// Get the attribute that specifies sparse iteration dimensions on operations.
+// Get/Set the attribute that specifies sparse iteration dimensions on
+// operations.
 SparseIterationDimsAttr getSparseIterationDimsAttr(Operation *op);
+void setSparseIterationDimsAttr(Operation *op,
+                                ArrayRef<int64_t> sparseIterationDims);
 
 } // namespace mlir::iree_compiler::IREE::TensorExt
 
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtInterfaces.td
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtInterfaces.td
@@ -96,6 +96,39 @@ def IREETensorExt_SparseCastOpInterface :
           "::mlir::RewriterBase &":$rewriter,
           "::llvm::ArrayRef<int64_t>":$sparseDims,
           "::llvm::ArrayRef<::mlir::Range>":$givenRanges)
+    >,
+    InterfaceMethod<
+      /*description =*/[{Resolve a range in terms of the source tensor.
+
+      Given a range that represents a slice of a sparse tensor (all dimensions
+      of the result), return an equivalent range in terms of the source tensor.
+
+      The inBounds bit vector (one bit per range entry) specifies whether each dimension's
+      access is guaranteed to be in-bounds. If false for any dimension, the implementation
+      should fail, as out-of-bounds handling is not yet supported.
+
+      The optional paddingValue parameter specifies the value to use for out-of-bounds
+      accesses. This is reserved for future use when guard generation is implemented.
+      }],
+      /*retTy =*/"::llvm::FailureOr<::llvm::SmallVector<::mlir::Range>>",
+      /*methodName =*/"resolveRange",
+      /*arguments =*/(ins
+          "::mlir::RewriterBase &":$rewriter,
+          "::llvm::ArrayRef<::mlir::Range>":$givenRanges,
+          "const ::llvm::BitVector &":$inBounds,
+          "::std::optional<::mlir::Value>":$paddingValue)
+    >,
+    InterfaceMethod<
+      /*description =*/[{
+        Returns which sparse dimensions can be distributed across workgroups.
+
+        For sparse dimensions where bounds depend on outer dimensions, those
+        dimensions cannot be distributed across workgroups. This method returns
+        a BitVector where bit i is set if sparse dimension i (from
+        SparseShapeAttrInterface::getSparseDimensions()) can be distributed.
+      }],
+      /*retTy =*/"::llvm::BitVector",
+      /*methodName =*/"getDistributionInfoForSparseDimensions"
     >
   ];
 }
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtInterfaces.td
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtInterfaces.td
@@ -98,7 +98,8 @@ def IREETensorExt_SparseCastOpInterface :
           "::llvm::ArrayRef<::mlir::Range>":$givenRanges)
     >,
     InterfaceMethod<
-      /*description =*/[{Resolve a range in terms of the source tensor.
+      /*description =*/[{
+      Resolve a range in terms of the source tensor.
 
       Given a range that represents a slice of a sparse tensor (all dimensions
       of the result), return an equivalent range in terms of the source tensor.
--- a/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
@@ -14,6 +14,7 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/BuiltinAttributeInterfaces.h"
 #include "mlir/IR/Dominance.h"
 
@@ -561,6 +562,16 @@ OpFoldResult CastToRaggedShapeOp::getNumRaggedRowsAsOfr() {
   return IntegerAttr::get(IndexType::get(getContext()), resultShape[raggedDim]);
 }
 
+// Cast `v` to index type if it's not already index type.
+static Value castToIndexTypeOrSelf(RewriterBase &rewriter, Location loc,
+                                   Value v) {
+  if (v.getType() != rewriter.getIndexType()) {
+    return arith::IndexCastOp::create(rewriter, loc, rewriter.getIndexType(),
+                                      v);
+  }
+  return v;
+}
+
 // Starting at `v`, compute a backward slice dominated by the given `sparseOp`
 // and clone the operations in the slice, replacing any occurrence of
 // `dim(%sparseOp, givenDim)` with the value returned by
@@ -584,9 +595,9 @@ static FailureOr<SmallVector<Value>> cloneAndReplaceDimInBackwardSlice(
   IRMapping mapping;
   for (auto op : slice) {
     IntegerAttr attr;
-    if (!matchPattern(
-            op, m_Op<memref::DimOp>(matchers::m_Val(sparseOp->getResult(0)),
-                                    m_Constant(&attr)))) {
+    Value dimSource;
+    if (!matchDimOp(op, dimSource, attr) ||
+        dimSource != sparseOp->getResult(0)) {
       rewriter.clone(*op, mapping);
       continue;
     }
@@ -658,10 +669,16 @@ CastToRaggedShapeOp::getEstimatedLoopRange(RewriterBase &rewriter,
       cloneAndReplaceDimInBackwardSlice(
           rewriter, loc, dominanceInfo, innerUb, *this, expectedSparseDims[1],
           [&](RewriterBase &rewriter, Location loc) {
-            OpFoldResult sourceDim =
-                memref::DimOp::create(rewriter, loc, getSource(),
-                                      expectedSparseDims[0])
-                    .getResult();
+            OpFoldResult sourceDim;
+            if (isa<MemRefType>(getSource().getType())) {
+              sourceDim = memref::DimOp::create(rewriter, loc, getSource(),
+                                                expectedSparseDims[0])
+                              .getResult();
+            } else {
+              sourceDim = tensor::DimOp::create(rewriter, loc, getSource(),
+                                                expectedSparseDims[0])
+                              .getResult();
+            }
             OpFoldResult numRaggedRows = getNumRaggedRowsAsOfr();
             AffineExpr s0, s1;
             bindSymbols(rewriter.getContext(), s0, s1);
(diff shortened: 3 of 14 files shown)


MaheshRavishankar:
cc @Groverkss  if you want to take a look. I got a basic CPU test working e2e. (Hit some error on GPU side that I will triage later), but I am landing all the pieces needed for a simple e2e in a way that is not disruptive. If you have any suggestions based on having looked at it, I am happy to take that into consideration before landing something. 

Just for context, this is just a "first solution". I know there are a lot of simplifying assumptions I used here. I am not tied to the full solution, but more just landing whatever "worked for now"

hanhanW:
I was heads-down for CPU benchmarks, and I'll review it on Monday.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +691,277 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
          "invalid specification of op with more than two sparse dimensions");
-  if (expectedSparseDims != sparseDims) {
-    return emitOpError(
-        "cannot lower the loop range for given sparse dimensions");
+
+  // Check that all requested sparse dims are valid.
+  for (int64_t sparseDim : sparseDims) {
+    if (!llvm::is_contained(expectedSparseDims, sparseDim)) {
+      return emitOpError(
+          "cannot lower the loop range for given sparse dimensions");
```
hanhanW: Instead of emitting op error (which prints out to stdout), maybe we should use `rewriter.notifyMatchFailure`?

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +691,277 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
+  // Verify constraints on sparse dimension ranges.
+  // For the outer sparse dimension: size must be 1, stride must be 1.
+  auto isOne = [](OpFoldResult ofr) -> bool {
+    if (auto attr = dyn_cast<Attribute>(ofr)) {
+      if (auto intAttr = dyn_cast<IntegerAttr>(attr)) {
+        return intAttr.getInt() == 1;
+      }
+    }
+    return false;
+  };
```
hanhanW: you can use `isOneInteger` method: https://github.com/llvm/llvm-project/blob/be072baa33d25dbad9d2cf53c2604b19aa293315/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L37-L38

```
/// Return true if `v` is an IntegerAttr with value `1`.
bool isOneInteger(OpFoldResult v);
```

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +691,277 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
+    }
+    if (resultDim == innerSparseDim) {
+      // Skip the inner sparse dimension - it's already handled.
+      continue;
+    }
+    // This is the outer sparse dimension - compute the range for the
+    // corresponding source dimension.
+    Value columnLengths = getColumnLengths();
+
+    // column_lengths[offset_0]
```
hanhanW: Complete the comment.

```suggestion
    // Extract the column_lengths[offset_0] value.
```

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +691,277 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
+    Value offset0 = getValueOrCreateConstantIndexOp(
+        rewriter, loc, givenRanges[outerSparseDim].offset);
+    Value columnLengthsCurrent;
+    if (isa<MemRefType>(columnLengths.getType())) {
+      columnLengthsCurrent =
+          memref::LoadOp::create(rewriter, loc, columnLengths, offset0);
+    } else {
+      // Tensor type - use tensor.extract
```
hanhanW: delete it, it's redundant comment, because we already have the above mega comment before the `if-else` statement.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +691,277 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
+                  arith::AddIOp::create(rewriter, loc, outerIv, one);
+              Value columnEnd =
+                  memref::LoadOp::create(rewriter, loc, columnLengths, plusOne);
+              Value columnStart =
+                  memref::LoadOp::create(rewriter, loc, columnLengths, outerIv);
+              // Cast to index type if needed before affine operations.
+              if (columnEnd.getType() != rewriter.getIndexType()) {
+                columnEnd = arith::IndexCastOp::create(
+                    rewriter, loc, rewriter.getIndexType(), columnEnd);
+              }
```
hanhanW: You may want to have a `castToIndexTypeOrSelf()` static method, so you can reuse it for this files. I've seen this pattern at least three times.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +691,277 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
+      columnLengthsCurrent = arith::IndexCastOp::create(
+          rewriter, loc, rewriter.getIndexType(), columnLengthsCurrent);
+    }
+
+    // Since we verified all accesses are in-bounds, we can use the requested
+    // size directly without clamping.
+    OpFoldResult sizeVal = givenRanges[innerSparseDim].size;
+
+    // stride = step_1
+    OpFoldResult strideVal = givenRanges[innerSparseDim].stride;
```
hanhanW, line 936: Complete the comment or just remove it. I think removing it is fine in this context..

```suggestion
    // Since we verified all accesses are in-bounds, we can use the requested
    // size directly without clamping.
    OpFoldResult sizeVal = givenRanges[innerSparseDim].size;
    OpFoldResult strideVal = givenRanges[innerSparseDim].stride;
```

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +691,277 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
+      getResultSparseEncoding().getSparseDimensions();
+
+  // For ragged tensors with 2 sparse dimensions:
+  // - First sparse dimension (raggedRow) CAN be distributed
+  // - Second sparse dimension (raggedRow + 1) CANNOT be distributed
+  //   because its bounds depend on the outer dimension.
+  llvm::BitVector result(sparseDims.size(), false);
+  if (!sparseDims.empty()) {
+    result.set(0); // Only first sparse dimension is distributable
+  }
```
hanhanW, line 967:
```suggestion
  // For ragged tensors with 2 sparse dimensions:
  // - First sparse dimension (raggedRow) CAN be distributed.
  // - Second sparse dimension (raggedRow + 1) CANNOT be distributed
  //   because its bounds depend on the outer dimension.
  llvm::BitVector result(sparseDims.size(), false);
  if (!sparseDims.empty()) {
    result.set(0); // Only first sparse dimension is distributable.
  }
```

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtAttrs.h
```diff
@@ -17,8 +17,11 @@
 
 namespace mlir::iree_compiler::IREE::TensorExt {
 
-// Get the attribute that specifies sparse iteration dimensions on operations.
+// Get/Set the attribute that specifies sparse iteration dimensions on
+// operations.
 SparseIterationDimsAttr getSparseIterationDimsAttr(Operation *op);
+void setSparseIterationDimsAttr(Operation *op,
+                                ArrayRef<int64_t> sparseIterationDims);
```
hanhanW, line 24: This is unused. Does it belong to the other PR? It's fine to leave it here as they are getter and setter pair.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtInterfaces.td
```diff
@@ -96,6 +96,39 @@ def IREETensorExt_SparseCastOpInterface :
...
+
+      The inBounds bit vector (one bit per range entry) specifies whether each dimension's
+      access is guaranteed to be in-bounds. If false for any dimension, the implementation
+      should fail, as out-of-bounds handling is not yet supported.
+
+      The optional paddingValue parameter specifies the value to use for out-of-bounds
+      accesses. This is reserved for future use when guard generation is implemented.
+      }],
+      /*retTy =*/"::llvm::FailureOr<::llvm::SmallVector<::mlir::Range>>",
+      /*methodName =*/"resolveRange",
```
hanhanW, line 115: We also need to add test for resolveRange — either by extending the test pass or by a dedicated test harness. The test pass `iree-tensor-ext-test-sparse-op-interface-methods` (TestSparseInterfaceMethods.cpp) only has options test-lower-loop-range and test-get-estimated-loop-range. There is no option or code path that calls resolveRange. 120 lines of new logic go in untested.

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtInterfaces.td
```diff
@@ -96,6 +96,39 @@ def IREETensorExt_SparseCastOpInterface :
...
+      /*description =*/[{
+        Returns which sparse dimensions can be distributed across workgroups.
+
+        For sparse dimensions where bounds depend on outer dimensions, those
+        dimensions cannot be distributed across workgroups. This method returns
+        a BitVector where bit i is set if sparse dimension i (from
+        SparseShapeAttrInterface::getSparseDimensions()) can be distributed.
+      }],
+      /*retTy =*/"::llvm::BitVector",
+      /*methodName =*/"getDistributionInfoForSparseDimensions"
```
hanhanW, line 132: ditto, this looks unused?

hanhanW (changes requested):
I don't see tensor test cases in the test file. Are we missing the test coverage for tensor cases? Comments from my agent wrt tensor cases:
  1. Fix lowerLoopRange to handle tensor-typed columnLengths (crash bug)
  2. Fix cloneAndReplaceDimInBackwardSlice to handle tensor.dim (silent miscompile)

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtAttrs.h
MaheshRavishankar: It probably does. I think the split of commits isnt as clean, and I am adding somethings that is not worth having a separate PR for.

MaheshRavishankar (commented):
(pushing comments)

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
```diff
@@ -691,103 +708,264 @@ CastToRaggedShapeOp::lowerLoopRange(RewriterBase &rewriter,
...
+      getResultSparseEncoding().getSparseDimensions();
+
+  // For ragged tensors with 2 sparse dimensions:
+  // - First sparse dimension (raggedRow) CAN be distributed.
+  // - Second sparse dimension (raggedRow + 1) CANNOT be distributed
+  //   because its bounds depend on the outer dimension.
+  llvm::BitVector result(sparseDims.size(), false);
+  if (!sparseDims.empty()) {
+    result.set(0); // Only first sparse dimension is distributable.
+  }
```
hanhanW, line 967: Do we need this check? The interface verifier already has the check. Can we just assume this?

https://github.com/iree-org/iree/blob/f737a3e084df1c98bbcb012b278bacbc95f3a064/compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOpInterfaces.cpp#L50-L54

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtInterfaces.td
```diff
@@ -96,6 +96,39 @@ def IREETensorExt_SparseCastOpInterface :
...
+      Given a range that represents a slice of a sparse tensor (all dimensions
+      of the result), return an equivalent range in terms of the source tensor.
+
+      The inBounds bit vector (one bit per range entry) specifies whether each dimension's
+      access is guaranteed to be in-bounds. If false for any dimension, the implementation
+      should fail, as out-of-bounds handling is not yet supported.
+
+      The optional paddingValue parameter specifies the value to use for out-of-bounds
+      accesses. This is reserved for future use when guard generation is implemented.
+      }],
```
hanhanW: Let's move the first sentence to the second line for consistency? I just found that I missed this line when I read the description..

```suggestion
      /*description =*/[{
      Resolve a range in terms of the source tensor.

      Given a range that represents a slice of a sparse tensor (all dimensions
      of the result), return an equivalent range in terms of the source tensor.

      The inBounds bit vector (one bit per range entry) specifies whether each dimension's
      access is guaranteed to be in-bounds. If false for any dimension, the implementation
      should fail, as out-of-bounds handling is not yet supported.

      The optional paddingValue parameter specifies the value to use for out-of-bounds
      accesses. This is reserved for future use when guard generation is implemented.
      }],
```

hanhanW (approved):
Just two nits

compiler/src/iree/compiler/Dialect/TensorExt/IR/TensorExtOps.cpp
MaheshRavishankar: We could but because it is doing `.set` it;ll segfault if it is empty.  So didnt have on opinion on the code that was generated by Claude here. Simple check avoids the segfault.

PR #20220, opened 2025-03-12
[DispatchCreation] Allow collapsing generic ops with index semantics

Groverkss:
We don't have any evidence right now that this regresses performance. If it does after this PR, it's an issue somewhere else in the compiler and it's better to not hide the issue.

This patch also moves a extract_slice test which was in reality a better test for linalg indexing operations. This patch moves this test to tests/e2e/linalg/index.mlir and XFAILs it for RISCV targets because of a bug in the llvm backend.

PR #23990, opened 2026-04-01
[Preprocessing] Add pass to convert strided insert_slice to linalg.generic

yzhang93:
Add `iree-preprocessing-convert-strided-insert-slice-to-generic` pass that converts strided `tensor.insert_slice` into zero-constant destinations to a `linalg.generic` with index arithmetic. This targets backward data convolution patterns where the upstream gradient is scattered into a zero buffer at strided positions.

The replacement generic computes the strided scatter in a single dispatch, replacing the `Memset + slow_memcpy` pair. Power-of-2 strides use bitwise ops instead of expensive integer division. The result is selected via `arith.select` for branchless GPU execution.

This PR is a partial implementation for https://github.com/iree-org/iree/issues/23976.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
@@ -221,6 +221,27 @@ static bool isEligibleForCollapse(Operation *op) {
     return false;
   }
 
+  // Skip collapse for scatter-like generics that use tensor.extract with
+  // linalg.index-based addressing. These are strided scatter patterns where
+  // 1D collapse introduces expensive delinearization (div/mod chains) that
+  // dominates execution. Keeping the multi-dimensional iteration space
+  // allows direct workgroup tiling without delinearization.
+  {
+    bool hasTensorExtract = false;
+    bool hasLinalgIndex = false;
+    genericOp.getBlock()->walk([&](Operation *inner) {
+      if (isa<tensor::ExtractOp>(inner)) {
+        hasTensorExtract = true;
+      }
+      if (isa<linalg::IndexOp>(inner)) {
+        hasLinalgIndex = true;
+      }
+    });
+    if (hasTensorExtract && hasLinalgIndex) {
+      return false;
+    }
+  }
+
   auto hasEncoding = [](Type type) -> bool {
     auto rankedTensorType = dyn_cast<RankedTensorType>(type);
     if (!rankedTensorType || !rankedTensorType.getEncoding()) {
--- a/compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
@@ -221,27 +221,6 @@ static bool isEligibleForCollapse(Operation *op) {
     return false;
   }
 
-  // Skip collapse for scatter-like generics that use tensor.extract with
-  // linalg.index-based addressing. These are strided scatter patterns where
-  // 1D collapse introduces expensive delinearization (div/mod chains) that
-  // dominates execution. Keeping the multi-dimensional iteration space
-  // allows direct workgroup tiling without delinearization.
-  {
-    bool hasTensorExtract = false;
-    bool hasLinalgIndex = false;
-    genericOp.getBlock()->walk([&](Operation *inner) {
-      if (isa<tensor::ExtractOp>(inner)) {
-        hasTensorExtract = true;
-      }
-      if (isa<linalg::IndexOp>(inner)) {
-        hasLinalgIndex = true;
-      }
-    });
-    if (hasTensorExtract && hasLinalgIndex) {
-      return false;
-    }
-  }
-
   auto hasEncoding = [](Type type) -> bool {
     auto rankedTensorType = dyn_cast<RankedTensorType>(type);
     if (!rankedTensorType || !rankedTensorType.getEncoding()) {
--- a/compiler/src/iree/compiler/Preprocessing/Common/ConvertStridedInsertSliceToGeneric.cpp
+++ b/compiler/src/iree/compiler/Preprocessing/Common/ConvertStridedInsertSliceToGeneric.cpp
@@ -1,0 +1,256 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Preprocessing/Common/Passes.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/WalkPatternRewriteDriver.h"
+
+namespace mlir::iree_compiler::Preprocessing {
+
+#define GEN_PASS_DEF_CONVERTSTRIDEDINSERTSLICETOGENERICPASS
+#include "iree/compiler/Preprocessing/Common/Passes.h.inc"
+
+namespace {
+
+/// Convert a strided `tensor.insert_slice` into a zero-constant destination
+/// to a `linalg.generic` with index arithmetic.
+///
+/// This targets the backward data convolution pattern where the upstream
+/// gradient is scattered into a zero buffer at strided positions, producing
+/// a separate Memset + slow_memcpy dispatch pair. The replacement generic
+/// computes the strided scatter in a single pass and is potentially
+/// fusable with consumer ops in later dispatch formation passes.
+///
+/// For each output position, the generic checks whether the position maps
+/// to a valid source element (i.e., (pos - offset) is non-negative,
+/// divisible by stride, and the quotient is in-bounds). Source indices are
+/// clamped to valid range so the extract is always safe, and arith.select
+/// chooses between the extracted value and zero.
+class ConvertStridedInsertSliceToGeneric
+    : public OpRewritePattern<tensor::InsertSliceOp> {
+public:
+  using Base::Base;
+  LogicalResult matchAndRewrite(tensor::InsertSliceOp op,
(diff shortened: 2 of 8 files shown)


compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
```diff
@@ -221,6 +221,27 @@ static bool isEligibleForCollapse(Operation *op) {
...
+      }
+      if (isa<linalg::IndexOp>(inner)) {
+        hasLinalgIndex = true;
+      }
+    });
+    if (hasTensorExtract && hasLinalgIndex) {
+      return false;
+    }
+  }
+
```
IanWood1: BTW a similar, but more broad, check used to exist: https://github.com/iree-org/iree/pull/20220

compiler/src/iree/compiler/Preprocessing/Common/ConvertStridedInsertSliceToGeneric.cpp
```diff
@@ -0,0 +1,256 @@
...
+    // A dim is "passthrough" if it has stride 1 and offset 0 — the scatter
+    // generic just copies it without any index arithmetic.
+    auto isPassthrough = [&](unsigned d) {
+      return strides[d] == 1 && offsets[d] == 0;
+    };
+
+    // Skip conversion when too many non-batch passthrough elements exist
+    // per spatial position. In grouped backward-data convolutions the
+    // group+channel dims are passthrough; when their product is large the
+    // scatter generic is slower than hardware DMA (Memset + slow_memcpy).
```
IanWood1, line 95: We probably dont' want to be adding backend heuristics (kMaxPassthroughElems) in Preprocessing.

compiler/src/iree/compiler/Preprocessing/Common/ConvertStridedInsertSliceToGeneric.cpp
```diff
@@ -0,0 +1,256 @@
...
+            Value rem, srcIdx;
+            if (llvm::isPowerOf2_64(stride)) {
+              rem = arith::AndIOp::create(b, loc, shifted, cstIdx(stride - 1));
+              srcIdx = arith::ShRSIOp::create(b, loc, shifted,
+                                              cstIdx(llvm::Log2_64(stride)));
+            } else {
+              Value strVal = cstIdx(stride);
+              rem = arith::RemSIOp::create(b, loc, shifted, strVal);
+              srcIdx = arith::DivSIOp::create(b, loc, shifted, strVal);
+            }
```
IanWood1: It seems a bit early to be doing this. Also, the rhs is a constant, so I'd assume backends should be able to handle this. It might make sense to always use remsi/divsi?

IanWood1 (commented):
Would you be able to explain the reasoning behind this a bit more?

I saw this from https://github.com/iree-org/iree/issues/23976 which seems to suggest that this might aid fusion:
>The tensor.insert_slice cannot be fused into the convolution dispatch, so the scatter always runs as a separate dispatch before the convolution

But the conv indexing map for the insert_slice will never be a projected permutation, so this won't fuse even if it's a `linalg.generic`. I'm wondering if it makes more sense to keep the `tensor.insert_slice` (+ find a way to fuse insert_slice in dispatch creation) then handle the conversion to a linalg.generic in codegen. This comment seems to suggest that the transformation belongs in codegen:

> when their product is large the scatter generic is slower than hardware DMA (Memset + slow_memcpy).

yzhang93:
> Would you be able to explain the reasoning behind this a bit more?
> 
> I saw this from #23976 which seems to suggest that this might aid fusion:
> 
> > The tensor.insert_slice cannot be fused into the convolution dispatch, so the scatter always runs as a separate dispatch before the convolution
> 
> But the conv indexing map for the insert_slice will never be a projected permutation, so this won't fuse even if it's a `linalg.generic`. I'm wondering if it makes more sense to keep the `tensor.insert_slice` (+ find a way to fuse insert_slice in dispatch creation) then handle the conversion to a linalg.generic in codegen. This comment seems to suggest that the transformation belongs in codegen:
> 
> > when their product is large the scatter generic is slower than hardware DMA (Memset + slow_memcpy).

You're right that the conv's indexing map for the scattered input won't be a projected permutation, so the scatter generic won't fuse with the conv dispatch even as a linalg.generic. I did experiment with putting the scatter and conv in the same dispatch (relaxing isFusableWithProducer in FormDispatchRegions to bypass the projected-permutation check). The two ops co-located in one dispatch, but the TileAndFuse codegen pipeline couldn't efficiently fuse the scatter as a tiled producer and the result was either compilation failures or severe performance regressions.

IMO, the actual benefit of converting to a linalg.generic is not fusion — it's replacing the Memset + slow_memcpy DMA pair with a single compute dispatch. For very large tensors (>1 GB), slow_memcpy bandwidth degrades significantly and the compute-based scatter is faster. For moderate tensors, DMA seems to be more efficient.

compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
yzhang93: I see. But this guard is narrower than the old one. It requires both `tensor.extract` AND `linalg.index`, targeting specifically scatter-like patterns where collapsing introduces delinearization on the tensor.extract indices (not just any index op). Without it, the scatter generic regresses 3.8x (38 μs → 144 μs) because the collapsed 1D indices must be delinearized back to multi-dim for the tensor.extract.

compiler/src/iree/compiler/Preprocessing/Common/ConvertStridedInsertSliceToGeneric.cpp
yzhang93: Agreed and confirmed there is no performance impact between two forms. I've simplified to always use remsi/divsi.

compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
```diff
@@ -221,6 +221,27 @@ static bool isEligibleForCollapse(Operation *op) {
     return false;
   }
 
+  // Skip collapse for scatter-like generics that use tensor.extract with
+  // linalg.index-based addressing. These are strided scatter patterns where
+  // 1D collapse introduces expensive delinearization (div/mod chains) that
+  // dominates execution. Keeping the multi-dimensional iteration space
+  // allows direct workgroup tiling without delinearization.
```
IanWood1: Could you add a small test for this?

compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
IanWood1: I'm a bit hesitant about the CollapseDimensions restriction since codegen sometimes relies on collapsed dims for good tiling configs, but because there's no evidence of perf regressions on other workloads it shouldn't block this PR.


> Without it, the scatter generic regresses 3.8x (38 μs → 144 μs) because the collapsed 1D indices must be delinearized back to multi-dim for the tensor.extract.

I'm not 100% sure, but I think another problem is that when it gets collapsed codegen loses information about the loop structure. For example, for `tensor.insert_slice %src into %cst[0, 1, 1, 0, 0] [32, 100, 100, 32, 8] [1, 2, 2, 1, 1]`, the innermost dim (size 256 after the pass collapses [32, 8]) is contiguous in memory, but after CollapseDimensions flattens to 1D, codegen can't prove that and falls back to `vector.gather` instead of contiguous vector loads.

IanWood1 (approved):
One last idea: what if we enable fusing (or cloning) the `arith.constant`/`linalg.fill` with the `tensor.insert_slice`. Then each backend can handle the fill + strided copy pattern how it sees fit. This would remove the backend-specific heuristics (kMaxPassthroughElems) from preprocessing and avoid the CollapseDimensions restriction.

I realize fusion might not be trivial since tensor.insert_slice isn't a linalg op. So, this is okay for now at least.

Other than that, this PR looks good to me.

compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
yzhang93: Okay, I need to double check the effect of this part. I think it makes sense to split this part as a separate PR.

compiler/src/iree/compiler/Preprocessing/Common/test/convert_strided_insert_slice_to_generic.mlir
```diff
@@ -0,0 +1,132 @@
...
+// CHECK-SAME:      tensor<1x25x25x4x8xf16> into tensor<1x25x25x32xf16>
+// Generic at reduced rank 4.
+// CHECK:       %[[GENERIC:.*]] = linalg.generic
+// CHECK-SAME:      iterator_types = ["parallel", "parallel", "parallel", "parallel"]
+// CHECK-SAME:      outs({{.*}} : tensor<1x52x52x32xf16>)
+// Dim 0 (batch): passthrough.
+// CHECK:         linalg.index 0
+// Dim 1 (H): strided.
+// CHECK:         linalg.index 1
+// CHECK:         arith.remsi
```
zjgarvey, line 81: Nit: maybe add a subi check here, since there is non-zero offset for the strided dims.

zjgarvey (commented):
I don't have any blocking comments.

Related to Ian's concerns, it does seem like the purpose in this pass is to force the source IR into a form that will trick the lower level GPU-specific compiler pipeline into doing something specific, but I don't have enough context to know why this should be done here vs. handled in the lower level pipeline.

yzhang93:
> One last idea: what if we enable fusing (or cloning) the `arith.constant`/`linalg.fill` with the `tensor.insert_slice`. Then each backend can handle the fill + strided copy pattern how it sees fit. This would remove the backend-specific heuristics (kMaxPassthroughElems) from preprocessing and avoid the CollapseDimensions restriction.
> 
> I realize fusion might not be trivial since tensor.insert_slice isn't a linalg op. So, this is okay for now at least.
> 
> Other than that, this PR looks good to me.

Thanks for the suggestion. Fusing the fill + strided insert_slice and letting each backend handle it could be a cleaner long-term approach. As you noted, it's not trivial since `tensor.insert_slice` isn't a linalg op, so the existing fusion infrastructure doesn't handle it directly.

For now, the preprocessing pipeline approach is safe — it will be gated behind an explicit flag so it only runs when needed. The `kMaxPassthroughElems` heuristic and `CollapseDimensions` guard can be revisited when a backend-aware insert_slice lowering is available.

On that note, I've moved the `CollapseDimensions` guard to a separate PR https://github.com/iree-org/iree/pull/24034 for independent CI testing and review.

PR #23992, opened 2026-04-02
[LLVMGPU] Modernize GPU pipeline tests.

hanhanW:
- Rename pipelines: `iree-codegen-linalg-to-nvvm-pipeline` → `iree-codegen-llvmgpu-nvvm-lowering-pipeline` `iree-codegen-linalg-to-rocdl-pipeline` → `iree-codegen-llvmgpu-rocdl-lowering-pipeline` with `LLVMGPULoweringPipelineOptions`.
- Remove `hal.executable` / `hal.executable.variant` wrapping from pipeline test IR and use `hal.executable.target` attribute on `func.func` instead.
- Update RUN lines to use direct pipeline flags instead of nested `--pass-pipeline` syntax.
- Update CHECK lines to match flattened output (llvm.func labels).
- Fix two exposed issue.
  * The MathTransformPass should not fail if executable target is not found. It means that the device library function or hardware intrinsic may not exist, so it is safer to enable approximation.
  * ROCDLAnnotateKernelForTranslation should check the configuration before access.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MathTransform.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MathTransform.cpp
@@ -161,9 +161,6 @@ public:
   void runOnOperation() override {
     RewritePatternSet patterns(&getContext());
     auto target = IREE::HAL::ExecutableTargetAttr::lookup(getOperation());
-    if (!target) {
-      return signalPassFailure();
-    }
     populateMathFunctionsRewritePatterns(patterns, [target](StringRef name) {
       return predicateRewrite(name, target);
     });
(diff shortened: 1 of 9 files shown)


hanhanW:
To reviewers, it is easier to review if you enable "hide whitespaces" feature.

image

compiler/src/iree/compiler/Codegen/Common/MathTransform.cpp
```diff
@@ -161,9 +161,6 @@ class MathTransformPass final
   void runOnOperation() override {
     RewritePatternSet patterns(&getContext());
     auto target = IREE::HAL::ExecutableTargetAttr::lookup(getOperation());
-    if (!target) {
```
MaheshRavishankar, line 164: Why remove this?

MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/Common/MathTransform.cpp
hanhanW: Because it is causing pipeline tests crash when executable target is not set. Many GPU pipeline tests do not require executable target. They only need GPUTarget attribute, which is derived from cli flag.

This variable is used to determine what math ops for approximantion. If target is not found, it defaults to all of them.

PR #23995, opened 2026-04-02
Skip dlclose under ASAN to fix LSAN module resolution

IanWood1:
When ASAN is enabled, keep dynamically loaded libraries mapped at process exit instead of closing them during teardown. This allows LeakSanitizer to resolve module names and trace pointer reachability through library globals, enabling library-name suppressions (e.g. `leak:libhsa-runtime64.so`) to work correctly.

Without this, IREE's dlclose of HIP/ROCm libraries causes LSAN to report all leaked addresses as `<unknown module>`. This was causing issues with Fusilli's gpu-asan CI iree-org/fusilli#292 (using LD_PRELOAD as a workaround).

Fusilli PR that encountered this: https://github.com/iree-org/fusilli/pull/287

### DIFF
--- a/runtime/src/iree/base/internal/dynamic_library_posix.c
+++ b/runtime/src/iree/base/internal/dynamic_library_posix.c
@@ -275,8 +275,10 @@ static void iree_dynamic_library_delete(iree_dynamic_library_t* library) {
   iree_allocator_t allocator = library->allocator;
   IREE_TRACE_ZONE_BEGIN(z0);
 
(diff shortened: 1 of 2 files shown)


MaheshRavishankar (approved):
Seems ok to me.

@rsuderman  any thoughts?

sjain-stanford (approved):

sjain-stanford (approved):
Seems inline with another dynamic lib handling in IREE: https://github.com/iree-org/iree/blob/4fb283371312ca9f85943f10da1dd7e2c3b925cd/runtime/src/iree/base/internal/dynamic_library_posix.c#L278-L295

PR #23996, opened 2026-04-02
[DispatchCreation] Improve broadcast consumer fusion check

Groverkss:


Groverkss:
I discussed this with Mahesh and IanWood and we don't want to do this.

PR #23997, opened 2026-04-02
[SPIRV] Scope SPIRVConvertGPUTarget on module op.

hanhanW:
The revision gathers the executable target from func ops, instead of assuming hal.executab.variant op exists. It also improves the pass to accept CLI GPUTarget testing flags and avoid crashes when target is not available.


efric (approved):
LGTM, thank you!

PR #23998, opened 2026-04-02
[SPIRV] Switch SPIR-V to module-scope pipeline.

hanhanW:
The revision moves the variant-scope passes to the boundary of pipelines, and the actual codegen works on module-scope after the change. The pipeline tests are adapted, so there are no hal.variant op involved.

It also fixes a long-due TODO in HALInterfaceLoadConstantConverter.


hanhanW:
Similar to other patches, it is much easier to review the lit tests part, if you enable "hide whitespace" option.

image

efric (approved):

PR #23999, opened 2026-04-02
[TMTensor][LinalgExt] Fuse causal mask into attention decomposition

keshavvinayak01:
Add `is_causal` attribute to `AttentionOp` and `OnlineAttentionOp`. When set, the causal condition `(k2 > m)` is computed inline from loop indices during decomposition, fused into the max-reduction and exp steps. This eliminates the need for a materialized N×N mask tensor.


Groverkss (changes requested):
We can just support iree_linalg_ext.index in the online_attention region and teach tiling how to work with it. I don't think we should be doing this.

keshavvinayak01:
> We can just support iree_linalg_ext.index in the online_attention region and teach tiling how to work with it. I don't think we should be doing this.

I think adding the `is_causal` optional attribute is still beneficial, since we won't have to detect downstream (iree here) if the mask is causal. 

https://github.com/llvm/torch-mlir/pull/4520/

keshavvinayak01:
I'll delay this work, pending more changes; since we're going to use `online_attention` and not attention from linalg moving forward.

PR #24000, opened 2026-04-02
[Codegen] Add pass to absorb swizzle hints into alloc attributes

Yu-Zhewen:
Part 1/3 of enabling XOR swizzle with pipelining (#23919).

**This PR:** Adds `AbsorbSwizzleHintToAllocPass`, which moves swizzle info from `SwizzleHintOp` to an `iree_codegen.swizzle` attribute on the defining `memref.alloc` and erases the hint op. The pass is not yet wired into any pipeline.

Assisted-by: Cursor (Claude)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/AbsorbSwizzleHintToAlloc.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/AbsorbSwizzleHintToAlloc.cpp
@@ -1,0 +1,56 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/PatternMatch.h"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_ABSORBSWIZZLEHINTTOALLOCPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+struct AbsorbSwizzleHintToAllocPass final
+    : impl::AbsorbSwizzleHintToAllocPassBase<AbsorbSwizzleHintToAllocPass> {
+  using Base::Base;
+  void runOnOperation() override;
+};
+} // namespace
+
+/// Absorbs `iree_codegen.swizzle_hint` ops into an attribute on the defining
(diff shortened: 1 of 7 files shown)


compiler/src/iree/compiler/Codegen/Common/AbsorbSwizzleHintToAlloc.cpp
```diff
@@ -0,0 +1,56 @@
...
+};
+} // namespace
+
+/// Absorbs `iree_codegen.swizzle_hint` ops into an attribute on the defining
+/// `memref.alloc`, then erases the hint. This allows downstream passes
+/// (multi-buffering, pipelining) to operate properly.
+static LogicalResult
+absorbSwizzleHintToAlloc(RewriterBase &rewriter,
+                         IREE::Codegen::SwizzleHintOp hintOp) {
+  auto allocOp = hintOp.getOperand().getDefiningOp<memref::AllocOp>();
```
jerryyin, line 31: [Question] Is this guaranteed for allocation to be its immediate parent?

jerryyin (approved):
LGTM with one question

compiler/src/iree/compiler/Codegen/Common/AbsorbSwizzleHintToAlloc.cpp
lialan: So have we made a decision where to resolve the swizzling? do we resolve it at the memory spot, or at the loading/storing operations?

Yu-Zhewen: > [Question] Is this guaranteed for allocation to be its immediate parent?

Currently yes, `SwizzleHintOp` is placed immediately after `tensor.empty` in [GPUPromoteMatmulOperandsPass](https://github.com/iree-org/iree/blob/44ef72e7b92ab4e1da4fcaf7857dac6380fd8260/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/PromotionImpls.cpp#L49). Plus other passes like [FlattenSwizzleHintAlloc ](https://github.com/iree-org/iree/blob/44ef72e7b92ab4e1da4fcaf7857dac6380fd8260/compiler/src/iree/compiler/Codegen/Common/FlattenSwizzleHintAllocs.cpp#L46) also make the same assumption.

Yu-Zhewen: > So have we made a decision where to resolve the swizzling? do we resolve it at the memory spot, or at the loading/storing operations?

Based on the discussion with @qedawkins , we're keeping the resolution at loading/storing operations (i.e. keep the current behavior in [ResolveSwizzleHints](https://github.com/iree-org/iree/blob/44ef72e7b92ab4e1da4fcaf7857dac6380fd8260/compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp#L180)). This way we minimize code changes to the existing resolution logic.

jerryyin (approved):

PR #24001, opened 2026-04-02
[Codegen] Add pass to reinsert swizzle hints from alloc attributes

Yu-Zhewen:
Part 2/3 of enabling XOR swizzle with software pipelining (#23919).

**This PR:** Adds `ReinsertSwizzleHintsPass`, which traces `vector.load` and `vector.store` ops back to attributed `memref.alloc`s and inserts `collapse_shape -> swizzle_hint -> expand_shape` at each use site. `gather_to_lds` swizzle is handled separately in `AMDGPULowerCoalescedDMAToGatherLDS`, which applies an inverse swizzle on the source indices directly -- `swizzle_hint` cannot be used there because the source is strided global memory. The pass is not yet wired into any pipeline.

Assisted-by: Cursor (Claude)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/Passes.td
+++ b/compiler/src/iree/compiler/Codegen/Common/Passes.td
@@ -1005,6 +1005,11 @@ def PropagateReshapesByExpansionPass :
   }];
 }
 
+def ReinsertSwizzleHintsPass :
+    InterfacePass<"iree-codegen-reinsert-swizzle-hints", "mlir::FunctionOpInterface"> {
+  let summary = "Re-inserts swizzle_hint ops from alloc attributes at vector.load/store sites";
+}
+
 def RematerializeParallelOpsPass :
     InterfacePass<"iree-codegen-rematerialize-parallel-ops", "mlir::FunctionOpInterface"> {
   let summary = "Pass to rematerialize and merge parallel ops into consumers.";
--- a/compiler/src/iree/compiler/Codegen/Common/Passes.td
+++ b/compiler/src/iree/compiler/Codegen/Common/Passes.td
@@ -1008,6 +1008,10 @@ def PropagateReshapesByExpansionPass :
 def ReinsertSwizzleHintsPass :
     InterfacePass<"iree-codegen-reinsert-swizzle-hints", "mlir::FunctionOpInterface"> {
   let summary = "Re-inserts swizzle_hint ops from alloc attributes at vector.load/store sites";
+  let dependentDialects = [
+    "IREE::Codegen::IREECodegenDialect",
+    "memref::MemRefDialect",
+  ];
 }
 
 def RematerializeParallelOpsPass :
--- a/compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
@@ -1,0 +1,142 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Interfaces/ViewLikeInterface.h"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_REINSERTSWIZZLEHINTSPASS
+#include "iree/compiler/Codegen/Common/Passes.h.inc"
+
+namespace {
+struct ReinsertSwizzleHintsPass final
+    : impl::ReinsertSwizzleHintsPassBase<ReinsertSwizzleHintsPass> {
+  using Base::Base;
+  void runOnOperation() override;
+};
+} // namespace
+
+/// Traces a memref value backward through view-like ops, memref.cast, and
+/// scf.for iter_args/results to find the root memref.alloc.
+static memref::AllocOp traceToAllocation(Value val) {
+  DenseSet<Value> visited;
+  SmallVector<Value> worklist = {val};
+  while (!worklist.empty()) {
+    Value current = worklist.pop_back_val();
+    if (!visited.insert(current).second) {
+      continue;
+    }
+    Operation *defOp = current.getDefiningOp();
+    if (!defOp) {
+      auto blockArg = cast<BlockArgument>(current);
+      auto forOp = dyn_cast<scf::ForOp>(blockArg.getOwner()->getParentOp());
+      unsigned argIdx = blockArg.getArgNumber();
+      if (!forOp || argIdx == 0) {
+        continue;
+      }
+      worklist.push_back(forOp.getInitArgs()[argIdx - 1]);
+    } else if (auto allocOp = dyn_cast<memref::AllocOp>(defOp)) {
+      return allocOp;
+    } else if (auto forOp = dyn_cast<scf::ForOp>(defOp)) {
+      unsigned resultIdx = cast<OpResult>(current).getResultNumber();
+      worklist.push_back(forOp.getInitArgs()[resultIdx]);
+    } else if (isa<ViewLikeOpInterface>(defOp)) {
+      worklist.push_back(defOp->getOperand(0));
+    } else if (auto castOp = dyn_cast<memref::CastOp>(defOp)) {
+      worklist.push_back(castOp.getSource());
+    }
+  }
+  return nullptr;
+}
+
+/// Returns the swizzle attribute on the alloc that val traces to, using
+/// cache to avoid repeated tracing.
+static IREE::Codegen::SwizzleAttrInterface lookupSwizzleAttr(
+    Value val,
+    DenseMap<Operation *, IREE::Codegen::SwizzleAttrInterface> &cache) {
+  memref::AllocOp allocOp = traceToAllocation(val);
+  if (!allocOp) {
+    return nullptr;
+  }
+  auto it = cache.find(allocOp.getOperation());
+  if (it != cache.end()) {
+    return it->second;
+  }
+  auto swizzle = allocOp->getAttrOfType<IREE::Codegen::SwizzleAttrInterface>(
+      "iree_codegen.swizzle");
+  cache[allocOp.getOperation()] = swizzle;
+  return swizzle;
+}
+
+/// Wraps |source| with collapse_shape -> swizzle_hint -> expand_shape so that
+/// downstream ResolveSwizzleHints can apply the XOR transform. For 1D memrefs,
+/// only the swizzle_hint is inserted.
+static Value insertSwizzleHint(IRRewriter &rewriter, Location loc, Value source,
+                               IREE::Codegen::SwizzleAttrInterface swizzle) {
+  auto sourceType = cast<MemRefType>(source.getType());
+
+  Value hintInput = source;
+  SmallVector<ReassociationIndices> reassoc;
+
+  if (sourceType.getRank() > 1) {
+    reassoc.push_back(
+        llvm::to_vector(llvm::seq<int64_t>(0, sourceType.getRank())));
+    hintInput = memref::CollapseShapeOp::create(rewriter, loc, source, reassoc);
+  }
+
+  auto hintOp =
+      IREE::Codegen::SwizzleHintOp::create(rewriter, loc, hintInput, swizzle);
+
+  if (sourceType.getRank() > 1) {
+    return memref::ExpandShapeOp::create(rewriter, loc, sourceType.getShape(),
+                                         hintOp.getResult(), reassoc);
+  }
+  return hintOp.getResult();
+}
+
+void ReinsertSwizzleHintsPass::runOnOperation() {
+  FunctionOpInterface funcOp = getOperation();
+  IRRewriter rewriter(funcOp->getContext());
+  DenseMap<Operation *, IREE::Codegen::SwizzleAttrInterface> swizzleCache;
+
+  // For each vector.load/store whose base traces to a swizzled alloc, wrap the
+  // base with collapse_shape -> swizzle_hint -> expand_shape.
+  funcOp.walk([&](Operation *op) {
+    Value base;
+    if (auto loadOp = dyn_cast<vector::LoadOp>(op)) {
+      base = loadOp.getBase();
+    } else if (auto storeOp = dyn_cast<vector::StoreOp>(op)) {
+      base = storeOp.getBase();
+    } else {
+      return;
+    }
+    IREE::Codegen::SwizzleAttrInterface swizzle =
+        lookupSwizzleAttr(base, swizzleCache);
+    if (!swizzle) {
+      return;
+    }
+    rewriter.setInsertionPoint(op);
+    Value wrapped = insertSwizzleHint(rewriter, op->getLoc(), base, swizzle);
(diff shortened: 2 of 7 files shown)


lialan:
So I guess the reason we want to do pushing it to `memref.alloc` and then pushing down again, is because we have many loading/storing ops that we want to handle?

i still think it is better to push it to the load/store op (where we consume/resolve the hint) instead of the mem op from the beginning.

Yu-Zhewen:
> So I guess the reason we want to do pushing it to `memref.alloc` and then pushing down again, is because we have many loading/storing ops that we want to handle?
> 
> i still think it is better to push it to the load/store op (where we consume/resolve the hint) instead of the mem op from the beginning.

It isn't about the number of load/store ops. The reason for "absorb-then-reinsert" is that SwizzleHintOp in the SSA chain between memref.alloc and its users breaks both `memref::multiBuffer` and `scf::pipelineForLoop` (see #23919 for details). After discussing with @jerryyin @krzysz00 @qedawkins, we think it is a reasonable short-term approach, as other options like encoding swizzle into the memref layout or teaching the pipeliner to understand SwizzleHintOp may require significantly larger changes.

jerryyin:
Do we really need a new pass for this? I do remember @qedawkins mention that if he is to implement this, he'd re-insert the hint first then resolve those hint. But I thought this is all still done in the same resolve swizzle hint pass, not a separate one.

Yu-Zhewen:
> Do we really need a new pass for this? I do remember @qedawkins mention that if he is to implement this, he'd re-insert the hint first then resolve those hint. But I thought this is all still done in the same resolve swizzle hint pass, not a separate one.

I was actually following your original [suggestion](https://github.com/iree-org/iree/issues/23919#issuecomment-4165755626) for a standalone pass at the "Candidate" position. I also think keeping it separate is cleaner. Happy to merge them if you feel strongly though.

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
```diff
@@ -0,0 +1,142 @@
...
+    } else if (auto allocOp = dyn_cast<memref::AllocOp>(defOp)) {
+      return allocOp;
+    } else if (auto forOp = dyn_cast<scf::ForOp>(defOp)) {
+      unsigned resultIdx = cast<OpResult>(current).getResultNumber();
+      worklist.push_back(forOp.getInitArgs()[resultIdx]);
+    } else if (isa<ViewLikeOpInterface>(defOp)) {
+      worklist.push_back(defOp->getOperand(0));
+    } else if (auto castOp = dyn_cast<memref::CastOp>(defOp)) {
+      worklist.push_back(castOp.getSource());
+    }
```
jerryyin: Can we unconditionally walk through all operand regardless of what current op is? The current approach will work but it might be tailored too closely to the IR we know right now.

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
```diff
@@ -0,0 +1,142 @@
...
+  }
+  return nullptr;
+}
+
+/// Returns the swizzle attribute on the alloc that val traces to, using
+/// cache to avoid repeated tracing.
+static IREE::Codegen::SwizzleAttrInterface lookupSwizzleAttr(
+    Value val,
+    DenseMap<Operation *, IREE::Codegen::SwizzleAttrInterface> &cache) {
+  memref::AllocOp allocOp = traceToAllocation(val);
```
jerryyin: Correct me if I'm wrong. It looks like this step is the most expensive but not cached. Once an allocation is found, then obtaining its swizzle attribute is trivial. Why isn't allocation tracing cached but attribute cached instead?

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
```diff
@@ -0,0 +1,142 @@
...
+  SmallVector<Value> worklist = {val};
+  while (!worklist.empty()) {
+    Value current = worklist.pop_back_val();
+    if (!visited.insert(current).second) {
+      continue;
+    }
+    Operation *defOp = current.getDefiningOp();
+    if (!defOp) {
+      auto blockArg = cast<BlockArgument>(current);
+      auto forOp = dyn_cast<scf::ForOp>(blockArg.getOwner()->getParentOp());
```
krzysz00: We could make this on LoopLikeOpInterface and check if it's an iteration argument, but that's maybe overkill?

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
```diff
@@ -0,0 +1,142 @@
...
+  if (it != cache.end()) {
+    return it->second;
+  }
+  auto swizzle = allocOp->getAttrOfType<IREE::Codegen::SwizzleAttrInterface>(
+      "iree_codegen.swizzle");
+  cache[allocOp.getOperation()] = swizzle;
+  return swizzle;
+}
+
+/// Wraps |source| with collapse_shape -> swizzle_hint -> expand_shape so that
```
krzysz00, line 84: To check, the only values of `source` this'll come up for are allocations and block arguments?

If so, I think we can rewrite the alloc() to collapse its shape and then just have expand_shape, but for the block arguments we'll need what this function does.

... though, for block arguments, could you actually do [swizzle, expand_shape] ... operations ... [collapse_shap, yield], changing the type of the loop argument? I think `hoistable_conversion` might be helpful here? Possibly before the swizzle hint gets inserted?

krzysz00 (commented):
Minor structural nitpicks (and perhaps wondering if we want a `FlattenLoopCarriedMemrefs` somewhere?

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
Yu-Zhewen: Done. Updated `traceToAllocation` to generically push all memref typed operands onto the worklist, with `scf::ForOp` as the only explicit case for correct index mapping

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
Yu-Zhewen: Good catch. Code updated.

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
Yu-Zhewen: Yeah we could, but this would be a more "invasive" change. Any particular benefit to do so?

krzysz00: Mainly wanting to be able to `fold-memref-alias-ops` that `expand_shape` into all the loads/stores so that they see the SwizzleHintOp directly

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
Yu-Zhewen: Shouldn't matter in practice since `scf.for` is the only loop here, but done. Switched to LoopLikeOpInterface

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
Yu-Zhewen: But even with the current structure, the `expand_shape` ops are already folded by `FoldMemRefAliasOps` before `ResolveSwizzleHints` runs, so loads/stores see the `SwizzleHintOp` directly either way. The only difference I see is the location of collapse_shape, unless I'm missing something:

Current: [collapse_shape, swizzle_hint, expand_shape] ... ops ... [yield]
Your suggestion: [swizzle_hint, expand_shape] ... ops ... [collapse_shape, yield]

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
```diff
@@ -0,0 +1,146 @@
...
+struct ReinsertSwizzleHintsPass final
+    : impl::ReinsertSwizzleHintsPassBase<ReinsertSwizzleHintsPass> {
+  using Base::Base;
+  void runOnOperation() override;
+};
+} // namespace
+
+/// Traces a memref value backward through defining ops and loop
+/// iter_args/results to find the root memref.alloc.
+static memref::AllocOp traceToAllocation(Value val) {
```
jerryyin, line 29: This implementation looks somewhat familiar to me now... Can this be replaced with getBackwardSlice() call? Like two steps: 1) we do a get backward slice call and then 2) attempt to find allocation from the backward slice? What different does that make compared to current implementation?

Yu-Zhewen: I don't think `getBackwardSlice` would handle correctly and efferently for the pipelined forOp? With multiple memref iter_args it could trace to the wrong alloc.

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
jerryyin: I don't think it is wise to optimize the hint re-insertion mechanism further. If we skip re-inserting SwizzleHintOp entirely and instead teach ResolveSwizzleHints to do the backward trace + index rewriting directly (trace load → alloc attribute → linearize + swizzle indices → emit 1D load), the hint never needs to be re-materialized and the insertion mechanism becomes irrelevant.

The only benefit of a separate re-insertion pass is making the IR change incremental and visible between passes. That's a reasonable tradeoff for now, but amending loop types to optimize it seems like the opposite direction from where we want to go — which is toward not having a re-insertion step at all.

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
jerryyin: Ah right, fair enough.

compiler/src/iree/compiler/Codegen/Common/test/reinsert_swizzle_hints.mlir
```diff
@@ -0,0 +1,89 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-reinsert-swizzle-hints))" \
+// RUN:   --split-input-file --mlir-print-local-scope %s | FileCheck %s
+
```
jerryyin, line 3: An orthorganal comment to this review: I remember multi-buffered memref looks like:
  > memref<8192xbf16, strided<[1], off: ?>>

Can you help me confirm if the dynamic offset matters swizzle hint resolution later.

jerryyin, line 3: I think one test case on more complex situations (pipelined case) might be missing. For example, 
```mlir
 %alloc = memref.alloc {swizzle_hint}
 ...
 %res = scf.for iter_args(%arg = %alloc) {
   vector.load %arg
   scf.yield %arg
 }
 vector.load %res
```
 1) The scf.for carrying the alloc as an argument
 2) The %res in the pipelined epilogue that use the swizzled result

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -1020,6 +1005,11 @@ def PropagateReshapesByExpansionPass :
   }];
 }
 
+def ReinsertSwizzleHintsPass :
+    InterfacePass<"iree-codegen-reinsert-swizzle-hints", "mlir::FunctionOpInterface"> {
+  let summary = "Re-inserts swizzle_hint ops from alloc attributes at vector.load/store sites";
+}
```
jerryyin, line 1015: Should probably list memref as a dependent dialect.

compiler/src/iree/compiler/Codegen/Common/test/reinsert_swizzle_hints.mlir
Yu-Zhewen: Thanks. Done

compiler/src/iree/compiler/Codegen/Common/Passes.td
Yu-Zhewen: Added

compiler/src/iree/compiler/Codegen/Common/test/reinsert_swizzle_hints.mlir
Yu-Zhewen: Currently there is a `isIdentity()` check in [hint resolution](https://github.com/iree-org/iree/blob/007b1eef6f31c490666e00130e55c3e3e871dc9d/compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp#L166), which fails for any non-zero offset. I was removing the check during experiments and numerical is still correct. I believe that check is overly strict, since the final address is computed as `base_offset + swizzle(local_index)`, and the swizzle doesn't touch the base offset.

@qedawkins @Muzammiluddin-Syed-ECE correct me if I'm wrong on this.

compiler/src/iree/compiler/Codegen/Common/ReinsertSwizzleHints.cpp
krzysz00: Yeah, fair, I think this approach works fine

compiler/src/iree/compiler/Codegen/Common/test/reinsert_swizzle_hints.mlir
krzysz00: The issue is that that offset could be within the range being swizzled in general, and it's hard to know that it isn't, so I can't think of a good way to loosen that check.

edit: short of some bit on the swizzle hint that has "trust me" semantics - `unsafe_is_swizzleable_subview`, maybe?

Yu-Zhewen: Are you worried that after swizzling, we could access outside the range `[base_offset, base_offset + size)`? That concern exists regardless of whether the base offset is zero or not. I think it depends on `SwizzleAttrInterface::swizzleOffset` being a valid permutation within the local index range. The `isIdentity()` check itself doesn't add any safety needed.

jerryyin: The identity() call does serve for some purpose, just the offset part is irrelevant - Since the swizzled offset is guaranteed to be within a row, adding a base dynamic offset wouldn't matter.

I think what matters are:
 - non unit stride - swizzled address may push destination address to a gap, which isn't valid
 - Arbitrary affine map - When memref carries an arbitrary access pattern expressed by the affine map attribute, then this is no longer linearized indexing pattern so swizzled result can be wrong.

I just asked cursor what would be best guard against that two and get:

```cpp
SmallVector<int64_t> strides;
int64_t offset;
if (failed(memrefType.getStridesAndOffset(strides, offset)))
  return failure();  // Non-strided affine layout — reject
if (!llvm::all_of(strides, [](int64_t s) { return s == 1; }))
  return failure();  // Non-unit stride — reject
// Dynamic offset is fine
```

So conclusion seems to be that `isIdentity()` is indeed too strict, and we should replace the check with above instead.

jerryyin (approved):
LGTM

compiler/src/iree/compiler/Codegen/Common/test/reinsert_swizzle_hints.mlir
Yu-Zhewen: Agreed. Looking at the existing guard:

https://github.com/iree-org/iree/blob/007b1eef6f31c490666e00130e55c3e3e871dc9d/compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp#L161-L175

`isStaticShapeAndContiguousRowMajor` already rejects both cases (non unit stride, arbitrary affine map). So we can simply remove `isIdentity()` from the condition.

krzysz00: Good point - we can probably just drop the isIdentity()

PR #24002, opened 2026-04-02
[LLVMGPU] Enable swizzle absorb/reinsert passes

Yu-Zhewen:
Part 3/3 of enabling XOR swizzle with software pipelining (#23919).

**This PR:** Wires `AbsorbSwizzleHintToAllocPass` before and `ReinsertSwizzleHintsPass` after `ROCDLPrefetchSharedMemoryPass`.

Assisted-by: Cursor (Claude)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -271,7 +271,7 @@ static LogicalResult multiBufferLDSAllocations(scf::ForOp forOp,
                                                unsigned numBuffers) {
   SetVector<memref::AllocOp> sharedAllocs;
 
-  // Find all LDS allocations used by gather_to_lds
+  // Find all LDS allocations used by gather_to_lds.
   forOp->walk([&](amdgpu::GatherToLDSOp gatherOp) {
     if (auto alloc = traceToAllocation(gatherOp.getDst())) {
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -287,22 +287,28 @@ static LogicalResult multiBufferLDSAllocations(scf::ForOp forOp,
...
   }
 
-  // Now apply multi-buffering
+  // Now apply multi-buffering, and copy the swizzle attribute if present.
   for (memref::AllocOp alloc : sharedAllocs) {
     Location loc = alloc.getLoc();
-    if (failed(memref::multiBuffer(alloc, numBuffers,
-                                   /*skipOverrideAnalysis=*/true))) {
+    auto swizzleAttr = alloc->getAttr("iree_codegen.swizzle");
+    FailureOr<memref::AllocOp> newAlloc =
```
jerryyin: Is the upstream multi-buffer call isn't able to create a new buffer with same attribute? I think either approach will make it more robust:
 - Make upstream multi-buffer to take the same attribute as original allocation
 - Or, clone the attribute from original allocation to new allocation, regardless what it is

I think it is useful to not be as specific as cloning just the particular attribute.

Yu-Zhewen: Ah my bad. Upstream already handles that

jerryyin (approved):
LGTM

PR #24004, opened 2026-04-03
[LLVMCPU] Implement conv heuristics in getVectorPreProcStrategy

josephbak:
## Summary

Named conv ops previously returned `VectorPreProcStrategy::None`unconditionally in `getVectorPreProcStrategy`, bypassing all target-aware heuristics. This caused two problems:

1. The peeling pipeline was never attached to conv ops.
2. `allowIncompleteTile` stayed false, forcing the tiler to find exact divisors and producing suboptimal tile sizes for odd kernel shapes (e.g. KH=3, KW=3).

Remove the early exit so named convs fall through to the existing target-aware heuristics. In `setConvRootConfig`, clamp `Masking` to `None` (not yet wired for convs) and connect `vecPreProcStrategy` to `distConfig.allowIncompleteTile`, mirroring the matmul path in `setContractionRootConfig`.

## Testing

Tested locally on AArch64 NEON. x86 (no AVX-512) and RISC-V are expected to follow the same Peeling path but rely on CI for validation. SVE (Masking) is explicitly clamped to None pending future wiring.

## Notes

- Pre-existing test failure in `select_aarch64_lowering_strategy.mlir` due to pipeline format mismatch (`CPUDoubleTilingExpert` vs `#iree_cpu.pipeline<DoubleTilingExpert>`) unrelated to this fix.
- Masking support for convs is deferred, tracked by the TODO added in `setConvRootConfig`.

Assisted-by: Claude (Anthropic)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -241,11 +241,6 @@ getVectorPreProcStrategy(linalg::LinalgOp linalgOp) {
     return clPProcStrategy;
   }
 
-  // TODO: Implement heuristics for Convs
-  if (isa<linalg::ConvolutionOpInterface>(linalgOp.getOperation())) {
-    return VectorPreProcStrategy::None;
-  }
-
   // Select a strategy based on heuristics.
   if (linalgOp.hasPureBufferSemantics()) {
     return VectorPreProcStrategy::None;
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -3030,10 +3025,18 @@ setConvRootConfig(mlir::FunctionOpInterface entryPointFn,
   }
   assert(!getLoweringConfig(convOp) && "expected lowering_config is not set");
 
+  // Masking is not yet wired for convs (no pipelineConfig branch exists).
+  // TODO: wire Masking support for convs.
```
hanhanW, line 3029: I dont follow this. Masking support should be available. It is not flipped because we don't look at conv perf for a while on CPU. What do you mean by this comment?

On the other hand, we should not silently fallback here. It is better to have the logic in `getVectorPreProcStrategy`, so we wont spread logic everywhere.

hanhanW (changes requested):
Thanks for the patch, I wonder if you did any benchmarking for the change? What is the outcome of this PR?

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
josephbak: Thanks for the review!

**Benchmarks**: Ran `iree-benchmark-module` on M-series AArch64 NEON (local-task) — inconclusive so far on tested shapes. Hunting odd-dim convs (KH=3) next; will add numbers soon.

**Masking**: Got it — moving conv strategy logic to `getVectorPreProcStrategy`, removing `setConvRootConfig` fallback. Update coming shortly.

PR #24005, opened 2026-04-03
[IO] Add FILE_SHARE_READ for read-only opens on Windows

jtuyls:
On Windows, files opened for reading cannot be opened again by another reader — the second open fails with ERROR_SHARING_VIOLATION. This breaks workflows where multiple IREE sessions load the same parameter file (.irpa).

The first commit adds a test that reproduces the failure — confirmed on Windows CI: https://github.com/iree-org/iree/actions/runs/23940576218/job/69825870972

The second commit enables the fix: add FILE_SHARE_READ to read-only opens, matching POSIX behavior where multiple readers are always allowed.


jtuyls:
Closing as I should just use IREE_IO_FILE_MODE_SHARE_READ in call site.

PR #24016, opened 2026-04-06
Add pattern to lower linalg.conv_2d_nchw_fchw_q to linalg.conv_2d_nchw_fchw

Tameem-10xE:
This PR extends LinalgQuantizedConvToConvPass `(iree-global-opt-quantized-conv-to-conv)` to support the **NCHW/FCHW** quantized conv2d op by lowering `linalg.conv_2d_nchw_fchw_q` to `linalg.conv_2d_nchw_fchw`. It also adds MLIR tests mirroring the existing NHWC tests.

The reason for this pattern is that if quantized_nchw_conv was not converted by this pass, it will not be lowered when the `img2col` pass runs, resulting in unoptimized MLIR generation.

Compilation Command (x86):
```
./iree_build/tools/iree-compile --iree-preprocessing-pass-pipeline="builtin.module(util.func(iree-global-opt-quantized-conv-to-conv, iree-global-opt-conv2d-to-img2col))"  \
  --iree-hal-target-device=local     --iree-hal-local-target-device-backends=llvm-cpu  \
  --iree-llvmcpu-target-cpu=host --iree-llvmcpu-target-cpu-features=host  \
  conv2d_nchw.mlir -o conv2d_nchw.vmfb
```

Test Run:
```
ctest -R iree/compiler/GlobalOptimization/test/linalg_quantized_conv_to_conv.mlir.test --output-on-failure
```

### DIFF
--- a/compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
@@ -124,7 +124,8 @@ Value multiplyDims(ImplicitLocOpBuilder &builder, Value value,
 //
 // This is implementing the math explained in Section 2.3 of
 // https://arxiv.org/abs/1712.05877.
-struct QuantizedConvToConv : OpRewritePattern<linalg::Conv2DNhwcHwcfQOp> {
+struct QuantizedConvNhwcToConvNhwc
+    : public OpRewritePattern<linalg::Conv2DNhwcHwcfQOp> {
   using Base::Base;
 
   LogicalResult matchAndRewrite(linalg::Conv2DNhwcHwcfQOp op,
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
```diff
@@ -239,6 +240,125 @@ struct QuantizedConvToConv : OpRewritePattern<linalg::Conv2DNhwcHwcfQOp> {
...
 
+// Pattern lowering conv_2d_nchw_fchw_q to conv_2d_nchw_fchw.
+struct QuantizedConvNchwToConvNchw
+    : public OpRewritePattern<linalg::Conv2DNchwFchwQOp> {
+  using Base::Base;
+
+  LogicalResult matchAndRewrite(linalg::Conv2DNchwFchwQOp op,
+                                PatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder builder(op.getLoc(), rewriter);
+    Value input = op.getInputs()[0];
```
rsuderman, line 251: These have named getters right? We should use them for readability. Also the `getType` ops should just call them on the values instead of the ops.

compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
```diff
@@ -239,6 +240,125 @@ struct QuantizedConvToConv : OpRewritePattern<linalg::Conv2DNhwcHwcfQOp> {
...
+      auto expandTy =
+          RankedTensorType::get({inputTy.getDimSize(0), 1,
+                                 inputTy.getDimSize(2), inputTy.getDimSize(3)},
+                                accETy);
+      inputSum = tensor::ExpandShapeOp::create(builder, expandTy, inputSum,
+                                               reassociationMap);
+
+      SmallVector<int64_t> poolDims;
+      SmallVector<Value> poolDynDims;
+      GetDynamicDym(builder, poolDims, poolDynDims, inputSum, 0);
```
rsuderman: Comment indicating what the numeric value is for.

compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
```diff
@@ -239,6 +240,125 @@ struct QuantizedConvToConv : OpRewritePattern<linalg::Conv2DNhwcHwcfQOp> {
...
+          resultShape[0], // N
+          resultShape[2], // OH
+          resultShape[3]  // OW
+      };
+      auto collapseTy = RankedTensorType::get(collapseShape, accETy);
+      inputSum = tensor::CollapseShapeOp::create(builder, collapseTy, inputSum,
+                                                 reassociationMap);
+
+      // Apply the zero-point update based on the input sum.
+      newConv = applyZeroPoint(builder, newConv, inputSum, fZp, {0, 2, 3});
```
rsuderman, line 347: Comment on what the set argument is

rsuderman (changes requested):

compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
Tameem-10xE: (1) I searched for the named operand for this op, but could not able to find any corresponding getters, so keeping this unchanged. The other patterns in the file also use the op.getInputs(...)
(2) Updated the code to use the output value directly instead of op.getType(0)

compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
Tameem-10xE: Done

compiler/src/iree/compiler/GlobalOptimization/QuantizedConvToConv.cpp
Tameem-10xE: Added comment

Tameem-10xE:
@rsuderman I’ve made the changes. Could you please take another look and let me know if anything else needs to be change/modify?
Thanks!

Tameem-10xE:
Hi @hanhanW, could you please review this PR and enable CI whenever you get a chance?

Thanks!

hanhanW:
> Hi @hanhanW, could you please review this PR and enable CI whenever you get a chance?
> 
> Thanks!

I'd ask for generalizing the pattern, i.e., use `inferConvolutionDims` to get the dims, so you can make a template pattern and register all the ops we care.

However, I found that there is a bug in upstream. These convolution all implement LinalgConvolutionOpInterface, which are declared in yaml. The implementation of [inferConvolutionDims](https://github.com/llvm/llvm-project/blob/c501f3747a46c4716fcf76e31075e725b3660502/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp#L887-L888) is not aligned with the [verification](https://github.com/llvm/llvm-project/blob/c501f3747a46c4716fcf76e31075e725b3660502/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp#L923-L924), which is a bug to me. We should correct the `inferConvolutionDims` in upstream, then you'll be able to generalize the pattern in IREE.

Can you give it a shot?

hanhanW:
I can approve the change if you need a stop-gap support here; you can work on the upstream fix then generalize the pattern later. However, it seems not that urgent since the PR is floating for ~1 month. So I'd appreciate if you can give it a shot.

Tameem-10xE:
Okay, agreed. Thanks for the review.

I’ll first try to correct `inferConvolutionDims` this week, then follow up with generalizing the pattern.

dimp-pl:
Checking in as another person interested in this PR getting merged. I've been doing some benchmarking of IREE vs ExecuTorch and other inference runtimes and this PR helps a LOT with INT8 performance (at least on Resnet18 and Resnet50, probably on more models).

I see that this is blocked because llvm/llvm-project#198462 still hasn't been merged. I'll see if i can help attract some attention to it.

NoumanAmir657:
cc @egebeysel

egebeysel:
> I’ll first try to correct inferConvolutionDims this week, then follow up with generalizing the pattern.

We'll try and take a look at both of these :) Thanks for the patches! I'll also ping here once that gets into LLVM, and comes to IREE through an integrate.

Tameem-10xE:
Since, the LLVM patch is still under review so for now I have pushed the generalized implementation (using `inferConvolutionDims`) to a separate branch: [code link](https://github.com/iree-org/iree/compare/main...10x-Engineers:iree-rise:tameem/quantized-conv-generalize)

The current implementation in this PR does not depend on the pending LLVM patch and can be merged independently. However, since the previous reviewer requested generalizing the patterns, I've prepared the generalized version too, but I haven't replaced this PR with it yet and I wanted to confirm before proceeding.

About the generalize pattern branch:

1. It uses a single templated rewrite pattern for the quantized convolutions, deriving the batch, channel, spatial, and filter dimensions via inferConvolutionDims instead of hardcoding them per layout.
2. It depends on the pending LLVM change that lets `inferConvolutionDims` accept convolutions with more than two DPS inputs.
3. It also folds the existing depthwise lowering into the same generic pattern. If you'd prefer to keep depthwise separate, I can leave that pattern as is.

The generalized version passes all existing tests + I have also tested it with `YOLOv8 (int8)` inference as well. If the approach looks good, I'll update this PR branch with it once the LLVM patch merge.

Thanks for taking the time to review this PR.

> Note: Generalization code change is assisted by Claude Code

PR #24018, opened 2026-04-06
[LinalgExt] Reduce im2col multiple k_pos test sizes to fix TSAN timeout

Max191:
The `@im2col_multiple_k_pos` test case used large K dimension factors (16, 48, 32) = 24576 total iterations. Under TSAN instrumentation with unroll=true, loop unrolling generates massive IR that takes ~38s to process (on my machine), pushing the lit test close to the 60s CTest timeout and causing flaky CI failures when running on weaker machines.

Reduce K factors to (2, 2, 4) = 16 total iterations while preserving the same structural coverage (multiple k_pos entries, both used in extract_slice offsets).

Successful CI run: https://github.com/iree-org/iree/actions/runs/24038339206


yzhang93 (approved):
Thanks for fixing the issue!

PR #24024, opened 2026-04-07
Avoid flatcc POSIX dependency in bare-metal runtime builds

HanKuanChen:
Background:
Commit 08b6af6b773558aaca14f7c8b783eb6d1f3622d0 assumed that the runtime could use flatcc when cross-compiling, which pulled in flatcc components that depend on POSIX APIs. However, bare-metal targets do not provide a POSIX environment, and only the host-side iree-flatcc-cli is expected to run with POSIX support.

Solution:
Introduce IREE_ENABLE_POSIX (default ON) to control building tools that
require POSIX APIs. When OFF, skip building iree-flatcc-cli and import
the host binary instead.


HanKuanChen:
ping

build_tools/third_party/flatcc/CMakeLists.txt
```diff
@@ -86,6 +86,13 @@ external_cc_library(
   PUBLIC
 )
 
+if(CMAKE_CROSSCOMPILING AND CMAKE_SYSTEM_NAME STREQUAL "Generic")
+  if(IREE_HOST_BIN_DIR)
+    iree_import_binary(NAME iree-flatcc-cli)
+  endif()
+  return()
+endif()
```
ScottTodd: This doesn't look quite right, or at least it should be accompanied by a comment.

In https://github.com/iree-org/iree/pull/22755 the goal was to _build_ `iree-flatcc-cli` and `iree-c-embed-data` when crosscompiling but continue to use the host versions within the build system. This chooses one specific configuration ("Generic" cmake system) to skip this one build.

Some suggestions:
1. Make that clear with a comment
2. Filter using something more specific than `CMAKE_SYSTEM_NAME STREQUAL "Generic"`
3. Filter using an explicit option and set that in bare metal builds, e.g. with a toolchain file

HanKuanChen: Thanks for the feedback! I think there might be a misunderstanding about the nature of this issue.

As mentioned in the PR description:
> Commit 08b6af6b773558aaca14f7c8b783eb6d1f3622d0 assumed that the runtime could use flatcc when cross-compiling, which pulled in flatcc components that depend on POSIX APIs. However, bare-metal targets do not provide a POSIX environment, and only the host-side iree-flatcc-cli is expected to run with POSIX support.

The iree-flatcc-cli source code uses POSIX file I/O and other POSIX APIs that are not available in bare-metal environments (e.g., RISC-V Generic targets). When cross-compiling to bare-metal, the build currently fails because of these missing dependencies.

The goal of PR #22755 was to build target binaries for cross-compilation scenarios, which makes sense for targets that can run these tools. However, bare-metal targets cannot build or run iree-flatcc-cli at all.

ScottTodd: That was already clear, but the comment definitely helps explain for future readers.

I'm skeptical about treating `CMAKE_SYSTEM_NAME STREQUAL "Generic"` and "is unable to build `iree-flatcc-cli`" as direct equivalents. From looking through known system names at https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html, there are a few that may also be missing POSIX APIs, and I could imagine someone picking "Generic" as the system name in a toolchain file where building _is_ possible.

I'm wondering if checking an option like `IREE_ENABLE_THREADING` would be a better fit: https://github.com/iree-org/iree/blob/d87d49e5f948f63e1b5a88aa053c39b7d1240b69/CMakeLists.txt#L50

---

Is there a CI system showing the build errors? These sorts of issues are much easier to fix (and keep fixed) once there is CI coverage for them.

We have https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci_test_riscv64.yml that builds using https://github.com/iree-org/iree/blob/main/build_tools/cmake/linux_riscv64.cmake then tests using qemu. Maybe a similar build workflow job could be added using https://github.com/iree-org/iree/blob/main/build_tools/cmake/generic_riscv64.cmake here: https://github.com/iree-org/iree/blob/d87d49e5f948f63e1b5a88aa053c39b7d1240b69/.github/workflows/ci.yml#L52-L82

HanKuanChen: I'm not sure `IREE_ENABLE_THREADING` is the right check here. Threading and POSIX are not equivalent:
- Some platforms have threading but not POSIX (e.g., Windows, some RTOS)
- The actual issue is that iree-flatcc-cli requires POSIX file I/O APIs (open/read/write/close), not threading

If you prefer a more explicit option, I could introduce `IREE_ENABLE_POSIX` (default ON) that can be set to OFF in bare-metal toolchain files. Would that be clearer?

Regarding CI: I'd like to add generic riscv64 coverage. However, the build currently has issues (this PR fixes part of it). Would you prefer:
1. Fix first, then add CI: Wait until the build is fully working, then add CI coverage
2. Add CI now, allow failures temporarily: Add the CI job now (as non-blocking) while I work on fixing the remaining issues

ScottTodd: If it's due to file APIs, maybe we could add a CMake option that drives `IREE_FILE_IO_ENABLE`, similar to how `IREE_ENABLE_THREADING` is used? https://github.com/iree-org/iree/blob/7ca09a2fc250c4c129d36fdf08c4069521371d2b/build_tools/cmake/iree_copts.cmake#L361-L366 https://github.com/iree-org/iree/blob/7ca09a2fc250c4c129d36fdf08c4069521371d2b/runtime/src/iree/base/config.h#L144-L160 

Could have `IREE_ENABLE_POSIX` too, which sets the finer-grained `IREE_ENABLE_FILE_IO` option and possibly others?

---

RE: CI, probably easier to get the build working then add the coverage, so people aren't distracted by the failing (even if skipped) signal. Hopefully there isn't too much churn while there is no automated coverage.

HanKuanChen: It seems like IREE_ENABLE_POSIX is the only solution. The actual error is:

```
riscv64-unknown-elf-ld.lld: error: undefined symbol: posix_memalign
>>> referenced by builder.c:442 (third_party/flatcc/src/runtime/builder.c:442)
>>>               build_tools/third_party/flatcc/CMakeFiles/iree-flatcc-cli.dir/__/__/__/third_party/flatcc/src/runtime/builder.c.obj:(flatcc_builder_finalize_aligned_buffer)
>>> referenced by builder.c:480 (third_party/flatcc/src/runtime/builder.c:480)
>>>               build_tools/third_party/flatcc/CMakeFiles/iree-flatcc-cli.dir/__/__/__/third_party/flatcc/src/runtime/builder.c.obj:(flatcc_builder_aligned_alloc)
```

This is neither threading nor file I/O related. It's the POSIX `posix_memalign` function.
I'll introduce an `IREE_ENABLE_POSIX` option (default ON) similar to `IREE_ENABLE_THREADING`, and use it to skip building iree-flatcc-cli when set to OFF. Does that sound good?

ScottTodd: Sure, sounds good to me. Thanks!

HanKuanChen: fix

HanKuanChen:
ping

ScottTodd (approved):

ScottTodd:
We may need to trigger CI a few times while https://www.githubstatus.com/incidents/myrbk7jvvs6p is ongoing.

PR #24026, opened 2026-04-07
Add strnlen fallback for newlib in thread.c

HanKuanChen:
newlib does not provide the strnlen() function, which is used in iree_strncpy_s(). Provide a simple fallback implementation for newlib-based toolchains (e.g., bare-metal RISC-V).


HanKuanChen:
ping
cc @hanhanW @ScottTodd

hanhanW (approved):
The code looks okay to me, but I don't have context of NEWLIB support

HanKuanChen:
> The code looks okay to me, but I don't have context of NEWLIB support

The build failed because the RISC-V bare-metal toolchain relies on newlib.

PR #24027, opened 2026-04-07
[Codegen] Make CreateDispatchConfig idempotent

Groverkss:
Recompiling a dumped configured hal.executable reruns the executable configuration pipeline. CreateDispatchConfigPass was blindly creating a new iree_codegen.dispatch_config even when one already existed, which left duplicate configs for the same entry point.

That later caused PropagateDispatchConfigPass to fail on the stale config without workgroup_size, even though reconciliation had populated the other one.

Make CreateDispatchConfigPass skip functions that already have a matching dispatch_config, and add a regression test covering rerunning the pass on an already-configured executable.

Assisted-by: Codex

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/create_dispatch_config.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/create_dispatch_config.mlir
@@ -116,3 +116,31 @@ hal.executable private @interleaving {
 //       CHECK:   iree_codegen.dispatch_config @fn1
 //       CHECK:   func.func @fn2
 //       CHECK:   iree_codegen.dispatch_config @fn2
+
+// -----
+
+// Rerunning the pass on an already-configured executable should be a no-op.
+#pipeline_layout = #hal.pipeline.layout<bindings = [
+  #hal.pipeline.binding<storage_buffer>
+]>
+hal.executable private @already_configured {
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/test/create_dispatch_config.mlir
```diff
@@ -116,3 +116,31 @@ hal.executable private @interleaving {
...
+      iree_codegen.dispatch_config @entry_point {
+      ^bb0(%device: !hal.device, %arg0: index):
+        %x, %y, %z = iree_tensor_ext.dispatch.workgroup_count_from_slice(%arg0)
+        iree_codegen.yield %x, %y, %z : index, index, index
+      }
+    }
+  }
+}
+// CHECK-LABEL: hal.executable private @already_configured
+// CHECK-COUNT-1: iree_codegen.dispatch_config @entry_point
```
hanhanW, line 146: We should probably do `CHECK + CHECK-NOT`? Otherwise, the test is still passing without your change, IIUC?

hanhanW (approved):

compiler/src/iree/compiler/Codegen/Common/test/create_dispatch_config.mlir
Groverkss: done

PR #24028, opened 2026-04-07
[LLVMGPU] Legalize ND vectors to 1D vectors

amd-eochoalo:
Adds a pass to legalize ND vectors to 1-D vectors and another pass to canonicalize 1-D vectors.


amd-eochoalo:
@Groverkss I added the missing tests and corrected one test. I will move the transfer_gather pattern here tomorrow. From the meeting I believe it was mentioned that the vector.contract and vector.multi_reduction can go as a last resort, but at the moment it looks like all tests are passing. I'll try to find out if there's a pipeline that allows multidimensional vector.contract and vector.multi_reduction to reach this legalization pass. Anything else you'd like me to look into before looking into flattening?

krzysz00:
High-level question before I start reading - do all the known patterns that care about N-D vectors in some special way (vector.contract lowerings, etc) run before this?

Also, we should put some of this upstream once it lands here?

amd-eochoalo:
> High-level question before I start reading - do all the known patterns that care about N-D vectors in some special way (vector.contract lowerings, etc) run before this?

Anything that is target specific should run before this legalization. Anything that is not target specific will be unrolled here. 

I'm moving iree_vector_ext.transfer_{scatter,gather} into N-D vector legalization. @Groverkss says that some trivial unrolling for vector.multi_reduction and contraction can be added in N-D vector legalization but again everything target specific should run before this pass.

krzysz00:
I'm basically calling for a check of the massive pile of stuff that gets called inside of LLVGPUToX (for various X) to make sure none of that needs to be moved before vector unrolling.

amd-eochoalo:
@krzysz00 thanks for making it clearer! I'll take a look (after I fix the rebase here).

amd-eochoalo:
Also, if others already have ideas for things that should be moved before this legalization, happy to learn about it, thanks!

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -994,13 +994,19 @@ static void addLowerToLLVMGPUPasses(OpPassManager &modulePassManager,
       .addPass(createCanonicalizerPass)
       .addPass(createCSEPass);
 
-  // This pass needs to run before SCF -> CF.
```
krzysz00, line 997: That comment's important, please put it where it belongs

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
+// Copyright 2025 The IREE Authors
```
krzysz00, line 1: High-level question: What's "GPU" about this? Can this be a pass all the way out in Codegen/ and usable in CPU etc.?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
...
+        types.push_back(innerDimVector);
+      }
+      return success();
+    });
+
+    // Some ops (e.g. nvgpu.ldmatrix, nvgpu.mma.sync) abuse n-D vector types
+    // to represent a "struct of vectors". When these ops interact with the
+    // converted 1-D values, we need materializations to bridge the gap.
+    // This is unfortunate — ops that want to represent structs should use
+    // struct types, not n-D vectors.
```
krzysz00, line 54: MLIR doesn't really have struct types so

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
...
+    addSourceMaterialization([](OpBuilder &builder, VectorType targetType,
```
krzysz00, line 55: Ok, this bit is GPU-specific, but we could handle that by having a general pass and making this pass extend the type converter.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
...
+/// type converter's 1:N mapping.
+struct UnrollElementwiseOps final
+    : public OpTraitConversionPattern<OpTrait::Elementwise> {
+  using Base::Base;
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<ValueRange> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (!OpTrait::hasElementwiseMappableTraits(op) ||
+        op->getNumResults() != 1) {
```
krzysz00: Why the 1-result restriction?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
...
+    if (failed(getTypeConverter()->convertType(op->getResult(0).getType(),
+                                               convertedTypes))) {
+      return failure();
+    }
+
+    // Invert operands: from list-of-ValueRanges to N lists of operands.
+    SmallVector<SmallVector<Value>> opOperands;
+    for (auto i : llvm::seq<int64_t>(convertedTypes.size())) {
+      SmallVector<Value> newOperands;
+      for (ValueRange operandList : operands) {
```
krzysz00, line 112: Does this work for something like `affine.delinearize_index` where there are scalar operands? Add a test?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
...
+    SmallVector<int64_t> outerStrides = computeStrides(outerShape);
+
+    int64_t numOuter = outerShape.size();
+    int64_t numOuterIndices =
+        std::min(static_cast<int64_t>(staticPos.size()), numOuter);
+    bool hasInnerIndex = static_cast<int64_t>(staticPos.size()) > numOuter;
+
+    int64_t offset = 0;
+    for (int64_t i = 0; i < numOuterIndices; i++) {
+      offset += staticPos[i] * outerStrides[i];
```
krzysz00, line 164: ... dynamic positions are allowed, handle them?

krzysz00: Ok, more accurately, a dynamic position outside of the final position is a `switch` over its possible values, that should be doable.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
...
+    rewriter.replaceOpWithMultiple(op, {results});
+    return success();
+  }
+};
+
+/// Convert vector.shape_cast on n-D vectors. If the inner dimension is the
+/// same, just pass through the 1-D vectors. Otherwise, flatten all scalars
+/// via vector.to_elements and regroup into new 1-D vectors via
+/// vector.from_elements.
+struct ConvertVectorShapeCast final
```
krzysz00, line 337: Is scalarizing the right approach here? Probably, but ...

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -0,0 +1,761 @@
...
+        ConvertVectorInsertStridedSlice, ConvertArithConstant, ConvertUBPoison,
+        ConvertVectorToElements, ConvertVectorFromElements,
+        ConvertVectorBroadcast, ConvertVectorBitcast>(typeConverter, ctx);
+
+    // Some nvgpu ops abuse n-D vector types to represent a "struct of
+    // vectors". These ops are legal despite having n-D vectors — the
+    // materializations above bridge the gap. This is unfortunate; ops that
+    // want to represent structs should use struct types, not n-D vectors.
+    target.addLegalOp<nvgpu::LdMatrixOp, nvgpu::MmaSyncOp>();
+    target.addDynamicallyLegalOp<func::FuncOp>([&](func::FuncOp op) {
```
krzysz00: This could be any function-like op so this works on `util.func` too

krzysz00 (commented):
Overall idea seems sound, just poking at missed cases

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
amd-eochoalo: Yes, I agree that this pass can technically be in CPU as well. (I actually started looking into CPU just to make it a bit easier to begin with but @Groverkss told me to focus first on GPU). I wouldn't oppose moving this to common.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
amd-eochoalo: Yes, I can rephrase the comment, but it is my understanding that this pass is motivated by needing to remove struct of vectors during the VectorToLLVM lowering upstream that happens in the type converter. By having only 1-D vectors, we guarantee that no multidimensional vector will be converted to struct of vectors during vector to LLVM lowering upstream.

krzysz00: Fair

And I think we *can* try to fix up nvgpu if it'd make sense to - go talk to some of the people who work on those ops and see if they'd mind a different representation

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
amd-eochoalo: Mmmm... `affine.delinearize_index` wouldn't be matched here because `OpTrait::hasElementwiseMappableTraits(op)` would filter it out. But if `affine.delinearize_index` reaches this pass, the pass would fail because it cannot convert it. I'll take this to mean that a conversion for `affine.delinearize_index` pattern needs to be written along with a test. Thanks @krzysz00!

krzysz00: ... Ok, hold on, how is delinearize not elementwise mappable?

And same for linearize?

They *should* be, no?

amd-eochoalo: Yes, I think they could be elementwise mappable, but the current ops don't allow for tensors operands/results. 

```
bool OpTrait::hasElementwiseMappableTraits(Operation *op) {
  return op->hasTrait<Elementwise>() && op->hasTrait<Scalarizable>() &&
         op->hasTrait<Vectorizable>() && op->hasTrait<Tensorizable>();
}
```

```
// AffineLinearizeIndexOp

def AffineLinearizeIndexOp : Affine_Op<"linearize_index",
    [Pure, AttrSizedOperandSegments, Elementwise,
     // Infer multi_index types from result type during parsing.
     TypesMatchWith<"multi_index types must match result type",
                    "linear_index", "multi_index", "$_self",
                    "[](::mlir::Type a, ::mlir::TypeRange b) { "
                    "return llvm::all_of(b, [a](::mlir::Type t) { "
                    "return t == a; }); }">]>

  let arguments = (ins Variadic<Affine_IndexOrVectorOfIndex>:$multi_index,
    Variadic<Index>:$dynamic_basis,
    DenseI64ArrayAttr:$static_basis,
    UnitProp:$disjoint);
  let results = (outs Affine_IndexOrVectorOfIndex:$linear_index);

// AffineDelinearizeIndexOp

def AffineDelinearizeIndexOp : Affine_Op<"delinearize_index",
    [Pure, Elementwise,
     // Infer linear_index type from the first result type during parsing.
     TypesMatchWith<"linear_index type must match result types",
                    "multi_index", "linear_index", "$_self[0]">
    ]>

  let arguments = (ins Affine_IndexOrVectorOfIndex:$linear_index,
    Variadic<Index>:$dynamic_basis,
    DenseI64ArrayAttr:$static_basis);
  let results = (outs Variadic<Affine_IndexOrVectorOfIndex>:$multi_index);
```

```
// Type constraint for index-like types: index or vector of index.
def Affine_IndexOrVectorOfIndex :
    Type<Or<[Index.predicate,
             VectorOfAnyRankOf<[Index]>.predicate]>,
         "index or vector of index">; 
```

I'm not too familiar with the affine dialect to say whether it makes sense for these operations to have tensor operands/results. But maybe a different solution would be to change the implementation here in LegalizeNDVectors and have patterns that match these ops.

Groverkss: We should just check for Elementwise. Elementwise mappable traits is too big of a catch all. If something is elementwise and has vector operands we are good to go.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
Groverkss: This is intentionally disabled. If any vector.extract with dynamic offsets reaches here, we should fail. Keeping them further on from here is an anti-pattern.

It's okay to use it, but before stuff like vector unrolling, loops should have the correct loop structure (they should be unrolled properly).

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
Groverkss: It is. This pass shouldn't do anything fancy. If someone cares about this, it should be done before this pass.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
Groverkss: It could be put in Codegen/Common yes. I want to start with LLVMGPU though. For other backends, we will need better benchmark tracking to enable something like this.

Also, this pulls in nvgpu like dialects so not easy to put in Common/ . Maybe in future when we have an interface.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
amd-eochoalo: https://github.com/iree-org/iree/pull/24028/commits/26f764eabadd98731d75b00ca273290422201b7b https://github.com/iree-org/iree/pull/24028/commits/5361654d9efe40e0cadffb2a672bced4f63279a3

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
amd-eochoalo: https://github.com/iree-org/iree/pull/24028/commits/9fa57355126d94a2815415e8d219d767a7b82e83 https://github.com/iree-org/iree/pull/24028/commits/36bcf336ac277fa9686d0d876fafefd49fc94493

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
amd-eochoalo: https://github.com/iree-org/iree/pull/24028/commits/27c79e887e2d4883dd34df25a1c6bfd2438efb32

amd-eochoalo:
@Groverkss I moved the transfer_gather and transfer_scatter unrolling patterns to ND legalization and then called their lowerings after the unrolling.

Groverkss:
I think maybe keep transfer_gather/transfer_scatter out of this patch for now. The patterns you removed are also needed in CPU pipelines, so we can't just move them into LLVMGPU pipeline just yet. Otherwise lgtm, please wait for @krzysz00

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
krzysz00: Then we should be printing an optimization remark to make the failure loud

But also, stuff like the VDMFMA interleaves had (I don't know if it survived into the final version) have extracts at indices that were a function of thread ID

amd-eochoalo: I added a remark here: https://github.com/iree-org/iree/pull/24028/commits/696363fce8e8c18c1eaf633453b7ea04b129ed7c (if you want I can change it to a  notifyMatchFailure instead)?

krzysz00:
Ok, so I still have a high-level flag.

https://github.com/iree-org/iree/blob/7c9ec25c291c006b89e5d3a88df1e38244fae11f/compiler/src/iree/compiler/Codegen/LLVMGPU/ConvertToROCDL.cpp#L294-L315 - (and probably the analogous files) *still* have a bunch of lowering patterns for high-level vector ops running on what is probably the wrong side of this 1-D unrolling.

amd-eochoalo:
@krzysz00: @Groverkss and I thought it was better to leave this PR small, but I can move these other patterns in this PR if you prefer.

krzysz00:
@amd-eochoalo What I think you could do is make a PR that goes *before* this one and moves [some of] those patterns from where they are now to a pass that runs before ConvertToROCDL (and you can also grab Nvidia's patterns there). That change should be NFC.

Then, you can insert unrolling between those rewrites and conversion to LLVM. You might end up breaking up those patterns into two groups (those that run before and after 1-D unrolling) so you can insert 1-D unrolling into the middle.

Groverkss (commented):
LGTM, Just wait for review from @krzysz00 before merging

Groverkss (approved):

krzysz00 (approved):
No fundamental objections, though I'll have my ear out for math code that scalarizes because there's a stray `x 1 x T` in the vectors

PR #24030, opened 2026-04-07
Add optional CSPRNG stub for generic/bare-metal builds

HanKuanChen:
For some embedded/bare-metal environments we never need a CSPRNG, but we still want the runtime to build. In those cases it is better to compile successfully and have CSPRNG calls return an explicit UNIMPLEMENTED status than to fail the build at compile time.

This change adds an opt-in CMake option IREE_CSPRNG_USE_STUB (default OFF) that enables a stub iree_csprng_fill returning IREE_STATUS_UNIMPLEMENTED. The default behavior remains strict: if the option is not enabled, unsupported platforms still hit the existing


HanKuanChen:
ping

runtime/src/iree/base/internal/csprng.c
```diff
@@ -93,6 +93,18 @@ offset += chunk_size;
...
+
+// This is intended for builds where no operating system CSPRNG is available and
+// callers can tolerate IREE_STATUS_UNIMPLEMENTED. On platforms that can provide
+// a real CSPRNG, prefer adding a proper implementation above instead of
+// enabling this stub.
+IREE_API_EXPORT iree_status_t iree_csprng_fill(iree_byte_span_t buffer) {
+  if (buffer.data_length == 0) return iree_ok_status();
+  return iree_make_status(IREE_STATUS_UNIMPLEMENTED,
+                          "CSPRNG not implemented for this platform");
+}
```
hanhanW, line 106: I think we need a dedicated stub test that asserts iree_csprng_fill returns IREE_STATUS_UNIMPLEMENTED, so the behavior is actually covered:

https://github.com/iree-org/iree/blob/main/runtime/src/iree/base/internal/csprng_test.cc

CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ option(IREE_ENABLE_COMPILER_TRACING "Enables instrumented compiler tracing." OFF
 option(IREE_ENABLE_RENDERDOC_PROFILING "Enables profiling HAL devices with the RenderDoc tool." OFF)
 option(IREE_ENABLE_THREADING "Builds IREE in with thread library support." ON)
 option(IREE_SYNCHRONIZATION_DISABLE_UNSAFE "Disables all synchronization primitives (for single-threaded bare-metal targets only)." OFF)
+
+# Allow opting into a stub CSPRNG implementation on generic/bare-metal targets.
+option(IREE_CSPRNG_USE_STUB
+  "Allow building a stub CSPRNG implementation on generic/bare-metal targets (returns IREE_STATUS_UNIMPLEMENTED)."
```
hanhanW, line 55: Should we have a warning, like `Builds a stub iree_csprng_fill that returns UNIMPLEMENTED — DO NOT enable in builds where cryptographic randomness is required.`?

hanhanW (commented):
I'm not the expert in this area, so I just skimmed through the code and left few comments.

The doc is not updated, it says that `IREE_STATUS_INTERNAL` is returned on failures. 

https://github.com/iree-org/iree/blob/f3220d8f29bcdc4ad7067052a14a0d51a81f5458/runtime/src/iree/base/internal/csprng.h#L20-L44

runtime/src/iree/base/internal/csprng.c
HanKuanChen: bare-metal targets have `IREE_BUILD_TESTS=OFF`, so tests won't run on the actual target platform. Should I add a host-side test that compiles with `-DIREE_CSPRNG_USE_STUB` to verify the stub implementation (but it may make the test more complicated), or is the stub simple enough that we can skip dedicated test coverage (I prefer this one)?

PR #24038, opened 2026-04-08
[Codegen] Deprecate the old ExportOp path from PropagateDispatchSizeBounds

hanhanW:
After the module-scope pipeline migration, all the GPU backends do not touch export op during codegen. Thus, this option can be deprecated.

The tests are updated to exclude hal ops.


hanhanW:
Using "hide whitespace" feature will make review easier.

image

hanhanW:
It depends on https://github.com/iree-org/iree/pull/24037

qedawkins (approved):

PR #24041, opened 2026-04-08
Fix generic RISC-V build: disable async sockets/signals on IREE_PLATFORM_GENERIC

HanKuanChen:
Commit 7802eb5dad309333118a9d552c5f2f98ac302de7 added async socket and signal support assuming a host platform with POSIX (or Windows) facilities available. The generic RISC-V toolchain used by build_tools/cmake/generic_riscv64.cmake defines IREE_PLATFORM_GENERIC and targets a bare-metal environment without <sys/socket.h>, <signal.h>, or real OS signals. This caused the generic RISC-V runtime build to fail.

Fix this by making async sockets and signals explicitly unsupported on IREE_PLATFORM_GENERIC:
- Provide stub implementations of iree_async_socket_* on IREE_PLATFORM_GENERIC that always return IREE_STATUS_UNIMPLEMENTED and avoid including POSIX socket headers.
- Restrict inclusion of <signal.h> and sigset_t-based helpers to non-generic platforms, and make the generic bare-metal signal helpers no-ops.
Bare-metal targets are not expected to expose a POSIX socket API, so this matches the existing pattern used for file I/O and time on IREE_PLATFORM_GENERIC. Behavior on existing host platforms (Linux/Android/Windows/BSD/Emscripten) is unchanged.

### DIFF
--- a/runtime/src/iree/async/socket.c
+++ b/runtime/src/iree/async/socket.c
@@ -9,6 +9,86 @@
 #include "iree/async/proactor.h"
 
 //===----------------------------------------------------------------------===//
+// Platforms without socket support (generic bare-metal).
+//===----------------------------------------------------------------------===//
+
+#if defined(IREE_PLATFORM_GENERIC)
+
+IREE_API_EXPORT iree_status_t iree_async_socket_create(
+    iree_async_proactor_t* proactor, iree_async_socket_type_t type,
+    iree_async_socket_options_t options, iree_async_socket_t** out_socket) {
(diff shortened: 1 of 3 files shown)


HanKuanChen:
ping

hanhanW:
@AWoloszyn @benvanik can you help review? I'm not familiar with these code and wonder if we should have a test or not.

HanKuanChen:
ping

HanKuanChen:
ping

PR #24043, opened 2026-04-08
[Codegen] Fix transfer_scatter failing tests

Groverkss:
The commit didn't account for canonicalizations kicking in (which were landed as part of another pr).


sommerlukas (approved):
LGTM.

PR #24045, opened 2026-04-08
Fix Generic RISC-V bare-metal build by avoiding libdl and stubbing threading

HanKuanChen:
CMake's GNU platform logic was causing RISC-V Generic bare-metal builds (riscv*-unknown-elf) to link against libdl even though the sysroot does not provide it, and new async/threading code paths were pulling in POSIX and thread APIs that are not available on bare-metal targets.

Fix this in three parts:
- Clear CMAKE_DL_LIBS for Generic platforms in iree_setup_toolchain.cmake. Setting CMAKE_DL_LIBS only from the generic_riscv64.cmake toolchain file is not sufficient: CMake's Platform modules (e.g. Platform/GNU.cmake for GNU-style compilers) run after the toolchain file and unconditionally set CMAKE_DL_LIBS back to "dl". Overriding it in iree_setup_toolchain.cmake ensures that Generic builds do not pick up -ldl on their link lines.
- Add a Generic-only threading stub implementation in runtime/src/iree/base/threading/thread_generic.c that provides the iree_thread_* API but returns IREE_STATUS_UNAVAILABLE or no-ops on IREE_PLATFORM_GENERIC. This allows code that depends on the thread API (such as proactor_thread) to link on bare-metal targets without dragging in a real OS thread implementation.
- Exclude IREE_PLATFORM_GENERIC from the POSIX proactor path in runtime/src/iree/async/proactor_platform.c. Generic bare-metal platforms do not have the necessary POSIX APIs, so attempting to create a POSIX proactor results in unresolved symbols. For Generic we leave the proactor unavailable instead.

### DIFF
--- a/build_tools/cmake/iree_setup_toolchain.cmake
+++ b/build_tools/cmake/iree_setup_toolchain.cmake
@@ -69,6 +69,19 @@ macro(iree_setup_toolchain)
   # Supports dynamic library loading.
   #-----------------------------------------------------------------------------
 
+  # Bare-metal "Generic" targets (such as riscv*-unknown-elf) typically don't
+  # ship libdl in their sysroots. In those environments we don't support
+  # dynamic library loading anyway, and attempting to link against -ldl causes
+  # link failures. We cannot rely on setting CMAKE_DL_LIBS only from a
+  # toolchain file, as CMake's Platform modules (for example Platform/GNU.cmake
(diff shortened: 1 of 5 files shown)


HanKuanChen:
ping

HanKuanChen:
ping

HanKuanChen:
ping

PR #24046, opened 2026-04-08
[SPIRV] Add native shapes for step op

### LINKED ISSUE #24035: `stabehlo.select` fails to compile with SPIRV due to using `vector.step : vector<Nxindex>`
### What happened?

We have a model that works okay with the `llvm-cpu` backend but fails to compile for `vulkan` giving the compile error:
```
error: failed to legalize operation 'vector.step': %44 = "vector.step"() : () -> vector<5xindex>
```

I got Claude to make a minimal reproducer, and this was its conclusion:
>Here's a summary of what was found:                                                                                                                                                                                      
>                                                                                                                                                                                                                                                                                 
>  Root cause: stablehlo.gather + stablehlo.reduce + stablehlo.select get fused into a single dispatch. The vectorized gather uses vector.step : vector<5xindex> for index computation, which cannot be lowered to SPIRV (SPIRV has no index type).                               
>                                                                                                                                                                                                                                                                                 
>  Required pattern to trigger the bug:                                                                                                                                                                                                                                           
>  1. stablehlo.gather → bool mask                                                                                                                                                                                                                                              
>  2. Same bool mask → convert → reduce (this forces gather+reduce into the same dispatch)                                                                                                                                                                                        
>  3. Reduce result (broadcasted back) used as the select's true-value                                                                                                                                                                                                            
>  4. stablehlo.select uses the same gather bool mask as condition                                                                                                                                                                                                                
>                                                                                                                                                                                                                                                                                 
>  If any of these elements is missing (no reduce, different bool in select, or reduce result not used in select), the bug doesn't trigger.                                                                                                                                                                                                                                                                                   

Here's the minimal repro case. The comments are from Claude, so I can't neither confirm nor deny if the conclusion is correct.
```
// Minimal reproducer for iree-compile error:
// "failed to legalize operation 'vector.step': ... -> vector<5xindex>"
//
// Root cause: stablehlo.gather fused with stablehlo.reduce and stablehlo.select
// in the same dispatch. The vectorized gather uses vector.step : vector<Nxindex>
// for index computation, which cannot be lowered to SPIRV (SPIRV has no 'index' type).
//
// Required pattern:
//   1. gather -> bool mask
//   2. bool mask -> reduce (forces gather+reduce into same dispatch)
//   3. reduce result (broadcasted) is select's true-value
//   4. select uses the SAME bool mask from the gather as condition
//
// Compile with:
//   iree-compile --iree-hal-target-device=vulkan --iree-vulkan-target=sm_75 \
//     -o /dev/null model12_minimal.mlir
module {
  func.func public @main(
      %arg0: tensor<2x5x5xui8>,       // gather source
      %arg1: tensor<2x32x5x2xi32>     // gather indices
  ) -> tensor<2x32x5xf32> {
    %cst = stablehlo.constant dense<0.000000e+00> : tensor<f32>
    %c_0 = stablehlo.constant dense<0> : tensor<ui8>

    // Gather -> compare -> bool mask
    %gathered = "stablehlo.gather"(%arg0, %arg1) <{dimension_numbers = #stablehlo.gather<collapsed_slice_dims = [1, 2], operand_batching_dims = [0], start_indices_batching_dims = [0], start_index_map = [1, 2], index_vector_dim = 3>, slice_sizes = array<i64: 1, 1, 1>}> : (tensor<2x5x5xui8>, tensor<2x32x5x2xi32>) -> tensor<2x32x5xui8>
    %c_zero = stablehlo.broadcast_in_dim %c_0, dims = [] : (tensor<ui8>) -> tensor<2x32x5xui8>
    %mask = stablehlo.compare GT, %gathered, %c_zero, UNSIGNED : (tensor<2x32x5xui8>, tensor<2x32x5xui8>) -> tensor<2x32x5xi1>

    // Reduce over dim 2 using the same mask (fuses gather+reduce into same dispatch)
    %mask_f = stablehlo.convert %mask : (tensor<2x32x5xi1>) -> tensor<2x32x5xf32>
    %reduce = stablehlo.reduce(%mask_f init: %cst) applies stablehlo.add across dimensions = [2] : (tensor<2x32x5xf32>, tensor<f32>) -> tensor<2x32xf32>
    %reduce_bcast = stablehlo.broadcast_in_dim %reduce, dims = [0, 1] : (tensor<2x32xf32>) -> tensor<2x32x5xf32>

    // Select using SAME mask: true-value depends on reduce result
    // This is the operation where vector.step fails to legalize
    %zero = stablehlo.broadcast_in_dim %cst, dims = [] : (tensor<f32>) -> tensor<2x32x5xf32>
    %result = stablehlo.select %mask, %reduce_bcast, %zero : tensor<2x32x5xi1>, tensor<2x32x5xf32>
    return %result : tensor<2x32x5xf32>
  }
}
```

`iree-compile --iree-hal-target-device=vulkan --iree-vulkan-target=sm_75 -o /dev/null repro.mlir` 


Compiler

### Version information

I saw the same result from 3.11 through to latest.

### Additional context

_No response_

amd-eochoalo:
SPIRV's TypeConverter failed with the following debug message:

>  illegal: not a valid composite type

which for vectors and composite types means that the vector had an invalid size.
`SPIRVInitialVectorLoweringPass` includes the upstream VectorUnroll patterns which themselves can unroll vector.step.
However, vector.step was skipped from unrolling since it lacked a target shaped defined in the VectorUnroll options.

```
[vector-unroll VectorUnroll.cpp:92 1] Get unroll shape for op vector.step
[vector-unroll VectorUnroll.cpp:110 1] --vector op shape: 5
[vector-unroll VectorUnroll.cpp:114 1] --no unrolling target shape defined 
%20 = vector.step : vector<5xindex>-> SKIP
```

To fix this we simply define a target shape for vector.step following the established conventions for other operations.

Fixes #24035

### DIFF
--- a/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
@@ -284,7 +284,7 @@ SmallVector<int64_t> getNativeVectorShapeImpl(vector::ReductionOp op) {
 }
 
 SmallVector<int64_t> getNativeVectorShapeImpl(vector::StepOp op) {
-  VectorType srcVectorType = static_cast<VectorType>(op.getType());
+  VectorType srcVectorType = op.getType();
   assert(srcVectorType.getRank() == 1); // Guaranteed by semantics
   int64_t vectorSize = getComputeVectorSize(srcVectorType.getDimSize(0));
   return {vectorSize};
--- a/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
+++ b/compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
@@ -283,6 +283,13 @@ SmallVector<int64_t> getNativeVectorShapeImpl(vector::ReductionOp op) {
   return {vectorSize};
 }
 
+SmallVector<int64_t> getNativeVectorShapeImpl(vector::StepOp op) {
+  VectorType srcVectorType = static_cast<VectorType>(op.getType());
+  assert(srcVectorType.getRank() == 1); // Guaranteed by semantics
+  int64_t vectorSize = getComputeVectorSize(srcVectorType.getDimSize(0));
+  return {vectorSize};
+}
+
 SmallVector<int64_t> getNativeVectorShapeImpl(vector::TransposeOp op) {
   VectorType vectorType = op.getResultVectorType();
   SmallVector<int64_t> nativeSize(vectorType.getRank(), 1);
@@ -312,7 +319,8 @@ getNativeVectorShape(Operation *op, bool targetSupportsDotProd) {
 
   return TypeSwitch<Operation *, std::optional<SmallVector<int64_t>>>(op)
       .Case<VectorTransferOpInterface, vector::MultiDimReductionOp,
-            vector::ReductionOp, vector::TransposeOp, vector::GatherOp>(
+            vector::ReductionOp, vector::TransposeOp, vector::GatherOp,
+            vector::StepOp>(
           [](auto typedOp) { return getNativeVectorShapeImpl(typedOp); })
       .Case([=](vector::ContractionOp contract) {
         return getNativeVectorShapeImpl(contract, targetSupportsDotProd);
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/SPIRV/SPIRVInitialVectorLowering.cpp
```diff
@@ -283,6 +283,13 @@ SmallVector<int64_t> getNativeVectorShapeImpl(vector::ReductionOp op) {
   return {vectorSize};
 }
 
+SmallVector<int64_t> getNativeVectorShapeImpl(vector::StepOp op) {
+  VectorType srcVectorType = static_cast<VectorType>(op.getType());
```
efric: nit; do we need this cast? 

```suggestion
  VectorType srcVectorType = op.getType();
```

efric (approved):
Nice!

PR #24048, opened 2026-04-08
[Codegen] Fix BufferizeDispatchTensorLoadStore to handle nested ops

Yu-Zhewen:
The original logic only iterates over top-level operations in the function body. This missed ops nested inside loops such as `scf.forall`, which are created by split reduction for matmuls with large K dimensions.

Assisted-by: Cursor (Claude)


Yu-Zhewen:
CC: @lialan @jerryyin The bug was causing GPUConvertToCoalescedDMAPass to silently skip DMA conversion for split K.

jerryyin (approved):
LGTM, nice catch!

PR #24049, opened 2026-04-08
Add HIP native executable support, device symbol APIs, and HAL

AWoloszyn:
This squashes a series of incremental changes that extend the IREE HAL runtime to support native HIP executable loading, device global symbol management, and raw device memory transfers. These additions enable the pyre-runtime HIP binding to load and dispatch native GPU code and interact with device-resident global variables.

HAL API extensions:

- iree_hal_device_query_string: query device configuration values as strings (e.g. architecture name).
- iree_hal_device_transfer_h2d_raw / iree_hal_device_transfer_d2h_raw: synchronous transfers to/from raw device addresses not backed by HAL buffers, used for device global symbol access.
- iree_hal_executable_lookup_global: look up a global variable symbol by name in a loaded executable, returning its device address and size.
- All HAL device driver vtables updated with the new entry points (amdgpu, cuda, hip, local_sync, local_task, metal, null, vulkan).

HIP driver — native executable loading (native_executable_hipf.c/h):

- Full parser for AMD GPU HIP fat binary format (.hipfb), supporting both compressed (zstd) and uncompressed offload bundles.
- Extracts code object blobs for the target architecture from bundled multi-architecture fat binaries.
- Integrates with the existing native_executable infrastructure for export lookup, kernel dispatch, and parameter introspection.
- Expanded executable_export_parameter_t.size from uint8_t to uint32_t to support large kernel argument structures.

HIP driver — allocator and command buffer enhancements:

- hip_allocator: added memory info tracking (total/free), pitched 2D allocation support, and passthrough buffer wrapping for externally managed device pointers.
- stream_command_buffer: extended dispatch to handle hidden kernel arguments (block count, group/thread IDs), warp size reporting, and symbol-addressed global memory copies.
- hip_device: added query_string and raw transfer vtable implementations, total memory tracking via device properties.

Bug fixes:

- mimalloc allocator: replaced mi_heap_free(mi_prim_get_default_heap()) with mi_free() to fix cross-thread free assertions in debug builds.
- tree utility: fixed iteration over trees with gaps from deleted nodes.
- HIP driver: changed device index type from uint32_t to int32_t for consistency with HIP API conventions.


runtime/src/iree/base/allocator_mimalloc.c
```diff
@@ -123,7 +123,7 @@ static iree_status_t iree_allocator_mimalloc_free(void** inout_ptr) {
   void* ptr = *inout_ptr;
   if (IREE_LIKELY(ptr != NULL)) {
     IREE_TRACE_FREE(ptr);
-    mi_heap_free(mi_prim_get_default_heap(), ptr);
+    mi_free(ptr);
```
benvanik: please remove unless required - we consistently do not use the heap-less functions in here

runtime/src/iree/hal/drivers/amdgpu/logical_device.c
```diff
@@ -1174,6 +1206,7 @@ static const iree_hal_device_vtable_t iree_hal_amdgpu_logical_device_vtable = {
     .topology_info = iree_hal_amdgpu_logical_device_topology_info,
     .refine_topology_edge = iree_hal_amdgpu_logical_device_refine_topology_edge,
     .assign_topology_info = iree_hal_amdgpu_logical_device_assign_topology_info,
+    .query_string = iree_hal_amdgpu_logical_device_query_string,
```
benvanik: this needs to be up after query_i64 (in all files)

runtime/src/iree/hal/queue.h
```diff
@@ -30,7 +30,10 @@ extern "C" {
 // This enables just-in-time selection as a command buffer could be made
 // available to some set of queues when recorded and then AND'ed with an actual
 // set of queues to execute on during submission.
+#ifndef IREE_HAL_QUEUE_AFFINITY_T_DEFINED
```
benvanik: remove this

runtime/src/iree/hal/device.h
```diff
@@ -536,6 +536,21 @@ IREE_API_EXPORT iree_status_t iree_hal_device_assign_topology_info(
     iree_hal_device_t* device,
     const iree_hal_device_topology_info_t* topology_info);
 
+// Queries a configuration value as string data.
```
benvanik: same order issues

runtime/src/iree/hal/executable.h
```diff
@@ -19,6 +19,12 @@ extern "C" {
 
 typedef struct iree_hal_device_t iree_hal_device_t;
 
+// Forward declaration for queue affinity (defined in queue.h).
```
benvanik: remove

runtime/src/iree/hal/executable.h
```diff
@@ -97,14 +103,15 @@ typedef uint16_t iree_hal_executable_export_parameter_flags_t;
...
-  // Size of the parameter in bytes. Does not contain padding.
-  uint8_t size;
   // Flags indicating parameter behavior.
   iree_hal_executable_export_parameter_flags_t flags;
-  // Parameter name if available, otherwise empty.
-  iree_string_view_t name;
   // Offset of the parameter in bytes or binding ordinal, depending on type.
   uint16_t offset;
+  // Size of the parameter in bytes. Does not contain padding.
+  // Expanded to 32 bits to support large structures (e.g., TensorIterator).
```
benvanik: no transient comments ("expanded" makes no sense outside of this diff, as future readers only see the final change - this is an idiot claude-ism)

runtime/src/iree/hal/executable.h
```diff
@@ -154,6 +161,16 @@ IREE_API_EXPORT iree_status_t iree_hal_executable_lookup_export_by_name(
     iree_hal_executable_t* executable, iree_string_view_t name,
     iree_hal_executable_export_ordinal_t* out_export_ordinal);
 
+// Looks up a global variable symbol by name in the executable.
+// Returns the device address and size of the global variable.
+// |queue_affinity| specifies which device's memory space to query.
+// Returns IREE_STATUS_NOT_FOUND if the symbol doesn't exist.
+// Returns IREE_STATUS_UNIMPLEMENTED if the backend doesn't support globals.
+IREE_API_EXPORT iree_status_t iree_hal_executable_lookup_global(
```
benvanik: this would be _by_name for consistency.

we should return an iree_hal_buffer_t - then we don't need the raw transfer functions - just regular copies. you can use the heap buffer wrapper or a thinner one

runtime/src/iree/hal/device.h
```diff
@@ -1041,6 +1061,20 @@ typedef struct iree_hal_device_vtable_t {
       const iree_hal_device_profiling_options_t* options);
   iree_status_t(IREE_API_PTR* profiling_flush)(iree_hal_device_t* device);
   iree_status_t(IREE_API_PTR* profiling_end)(iree_hal_device_t* device);
+
+  // Optional: transfers data to/from raw device addresses.
+  // Used for global symbols in loaded modules.
+  // May be NULL if the device doesn't support raw transfers.
+  iree_status_t(IREE_API_PTR* transfer_h2d_raw)(iree_hal_device_t* device,
```
benvanik: we don't do optional vtables - see the note in executable.h - I think we can just make globals returned as buffers and then use normal queue_copy or mapping (also makes it possible to use them in command buffers/etc)

benvanik (changes requested):

runtime/src/iree/hal/executable.c
```diff
@@ -58,3 +58,28 @@ IREE_API_EXPORT iree_status_t iree_hal_executable_lookup_export_by_name(
...
+  IREE_ASSERT_ARGUMENT(executable);
+  IREE_ASSERT_ARGUMENT(out_buffer);
+  *out_buffer = NULL;
+  IREE_TRACE_ZONE_BEGIN(z0);
+
+  const iree_hal_executable_vtable_t* vtable =
+      (const iree_hal_executable_vtable_t*)((const iree_hal_resource_t*)
+                                                executable)
+          ->vtable;
+  if (!vtable->lookup_global_by_name) {
```
benvanik: similar to the device vtable, we don't want to support NULLs - the intent is that _VTABLE_DISPATCH can be turned into a direct call in compilation where only a single implementation exists, and that's not possible if we require it to be a pointer poked at via the pointer math.

runtime/src/iree/hal/drivers/vulkan/vulkan_device.cc
```diff
@@ -1604,6 +1604,23 @@ static iree_status_t iree_hal_vulkan_device_assign_topology_info(
   return iree_ok_status();
 }
 
+static iree_status_t iree_hal_vulkan_device_query_string(
```
benvanik: needs to be moved up after query_i64 (consistent orders in vtable and function definition)

runtime/src/iree/hal/drivers/local_sync/sync_device.c
```diff
@@ -300,6 +300,23 @@ static iree_status_t iree_hal_sync_device_assign_topology_info(
   return iree_ok_status();
 }
 
+static iree_status_t iree_hal_sync_device_query_string(
```
benvanik: same here on ordering (please check all files)

benvanik (changes requested):

runtime/src/iree/hal/executable.c
```diff
@@ -59,6 +59,14 @@ IREE_API_EXPORT iree_status_t iree_hal_executable_lookup_export_by_name(
   return status;
 }
 
+IREE_API_EXPORT iree_status_t
+iree_hal_executable_lookup_global_by_name_not_supported(
```
benvanik: exporting on the public API for implementations to return a status isn't good - we try to keep the API small and relevant to users, not implementers - just have them return unimplemented - that also keeps the source location and relevant: by centralizing this we see just executable.c:66 unimplemented, not hip/native_executable.c:123 unimplemented which tells us exactly which one it is

runtime/src/iree/hal/drivers/local_task/task_device.c
```diff
@@ -325,6 +325,23 @@ static iree_status_t iree_hal_task_device_query_i64(
       (int)category.size, category.data, (int)key.size, key.data);
 }
 
+static iree_status_t iree_hal_task_device_query_string(
+  iree_hal_device_t* base_device, iree_string_view_t category,
+  iree_string_view_t key, iree_host_size_t out_string_size,
+  char* out_string) {
+(void)base_device;
```
benvanik: something biffed

benvanik (changes requested):

AWoloszyn:
That should be all,  but I am relying on the presubmits to make sure I didn't mess up on most backends 
(I suspect the formatting will need fixing, but the content should be right)

runtime/src/iree/hal/device.h
```diff
@@ -505,6 +505,21 @@ IREE_API_EXPORT iree_status_t iree_hal_device_query_i64(
...
+// the key is not convertible, or the size exceeds that of out_string.
+//
+// Well-known queries (category :: key):
+//   hal.device.architecture :: some-pattern-*
+//
+// Returned values must remain the same for the lifetime of the device as
+// callers may cache them to avoid redundant calls.
+IREE_API_EXPORT iree_status_t iree_hal_device_query_string(
+    iree_hal_device_t* device, iree_string_view_t category,
+    iree_string_view_t key, iree_host_size_t out_string_size, char* out_string);
```
benvanik: the lifetime here is mismatched. the comment about lifetime is wrong as the caller is passing in a buffer.

we don't do nul terminated strings in general; an example signature:
```
iree_host_size_t out_string_capacity, char* out_string, iree_host_size_t* out_string_length
```
caller then allocates or stack-hosts a buffer, passes in `IREE_ARRAYSIZE(string_buffer), string_buffer, &string_length`, or passes in `0, NULL, &required_length` and heap allocates for it then passes in `string_length, buffer, &string_length`.

(I missed out_string_capacity being wrong there in review, but it's in the codebase so you can copy that to see the behavior)

if you do need nul terminators, you can +1 the required capacity and then add the NUL yourself e.g. `buffer[string_length] = 0;` without losing the precision/pre-calculation we use elsewhere to be able to put in iree_string_view_t.

benvanik (changes requested):

benvanik (approved):

benvanik:
Note that we will be taking a lighter touch on the AMDGPU side - we'll be expecting HSA-compatible ELFs to be provided (instead of offloading bundles/etc). I'm ok with allowing that here because HIP (:shrug:) but we'll be sticking to HSA-level on the AMDGPU side. The HIP compatibility shim should take on doing decompression/etc as needed to get into a form hsa_code_object_reader_create_from_memory can accept. I'm also going to propose a ROCR enhancement to give us the arg info so we don't need to do the messagepack stuff - e.g.:
```
    typedef struct hsa_ven_amd_kernel_arg_info_t {
      uint32_t offset;
      uint32_t size;
      uint32_t align;
      const char* value_kind;   // "global_buffer", "by_value", "hidden_*", etc.
      const char* address_space;
      const char* access;
    } hsa_ven_amd_kernel_arg_info_t;
  hsa_status_t hsa_ven_amd_loader_executable_symbol_iterate_kernel_args(
      hsa_executable_symbol_t symbol,
      hsa_status_t (*callback)(
          hsa_executable_symbol_t symbol,
          const hsa_ven_amd_kernel_arg_info_t* arg,
          void* user_data),
      void* user_data);
```
(but would be satisfied with a hsa_ven_amd_loader_code_object_reader_get_note so we didn't have to walk the ELF :)

benvanik:
Filed request: https://github.com/ROCm/TheRock/issues/4514

PR #24050, opened 2026-04-08
[DispatchCreation] Fuse producers with multiple users

### LINKED ISSUE #22528: Layernorm compiles to 3 dispatches instead of 1

### LINKED ISSUE #24033: [DispatchCreation] Layernorm training should fuse as a single dispatch
IR: https://gist.github.com/Groverkss/adc6c192d8fe0ec302faed3f2738612a

Repro: `iree-compile --iree-hal-target-device=hip --iree-rocm-target=gfx1201 --iree-opt-level=O3 layernorm_train_scale_bias_nhwc.mlir --compile-to=dispatch-creation --iree-opt-level=O3`

Running the command produces 2 dispatches.

Expected: The IR should fuse into a single dispatch

IanWood1:
Improve `FormDispatchRegions` fusion to allow fusing producers with multiple users when external users can be safely moved after the new dispatch. This enables patterns like layernorm to stay in a single dispatch.

Changes:
- Replace `hasUsesBetweenProducerAndRoot` + `getFusableUses` with `hasUnmovableUse`, which allows fusion when external uses of a producer can be safely moved after the dispatch
- Add cross-block fusion guard in `fuseRootsWithProducers` (was handled by `getFusableUses()`)
- Move external users of fused producers after the dispatch region in `movePrecedingOpsIntoDispatchRegion` (RegionOpUtils.cpp).

These changes can be disabled with `--iree-dispatch-creation-fuse-multi-use=false`

Closes https://github.com/iree-org/iree/issues/24033 
Closes https://github.com/iree-org/iree/issues/22528

ci-extra: test_torch


IanWood1:
One xfailed test is now passing and there are two new failures. I created https://github.com/iree-org/iree/issues/24058 to track the new failures. Both seem codegen and encoding related.

tests/external/iree-test-suites/onnx_ops/onnx_ops_cpu_llvm_sync_O2.json
```diff
@@ -117,6 +116,8 @@
     "onnx/node/generated/test_pow_types_int32_int32",
     "onnx/node/generated/test_pow_types_int64_float32",
     "onnx/node/generated/test_pow_types_int64_int64",
+    "onnx/node/generated/test_qlinearmatmul_2D_uint8_float16",
+    "onnx/node/generated/test_qlinearmatmul_2D_uint8_float32",
```
hanhanW: This will be fixed by https://github.com/iree-org/iree/pull/24066 and https://github.com/iree-org/iree/pull/24098

Why two PRs? Because the former one exposes a bug and the latter one fixes the issue.

compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
```diff
@@ -303,12 +306,88 @@ clonePrecedingOpIntoDispatchRegion(RewriterBase &rewriter, Operation *target,
...
 movePrecedingOpsIntoDispatchRegion(RewriterBase &rewriter,
                                    ArrayRef<Operation *> targets,
                                    IREE::Flow::DispatchRegionOp regionOp) {
+  llvm::SetVector<Operation *> targetSet;
+  targetSet.insert(targets.begin(), targets.end());
+
+  // Gather values that would need to be rewritten to dispatch results and
+  // ensure their external users can move below the dispatch.
+  SmallVector<Value> precheckReplacedValues;
+  for (Operation *target : llvm::reverse(targets)) {
```
MaheshRavishankar: Why llvm::reverse ?

compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
```diff
@@ -303,12 +306,88 @@ clonePrecedingOpIntoDispatchRegion(RewriterBase &rewriter, Operation *target,
   return newTargetOp;
 }
 
+bool hasUnmovableUse(Block *block, Operation *insertionPoint,
```
MaheshRavishankar: I am confused by how this is called from the caller. The `insertionPoint` and the `sliceBoundary` are both the `dispatchOp`. Then the slice boundary for the backward slice basically makes it such that the slice is always empty?

compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
```diff
@@ -303,12 +306,88 @@ clonePrecedingOpIntoDispatchRegion(RewriterBase &rewriter, Operation *target,
...
+  for (Value val : values) {
+    for (OpOperand &use : val.getUses()) {
+      Operation *user = block->findAncestorOpInBlock(*use.getOwner());
+      if (!user || isInGroup(user)) {
+        continue;
+      }
+      if (!user->isBeforeInBlock(insertionPoint)) {
+        continue;
+      }
+      if (isa<IREE::Flow::DispatchRegionOp>(user)) {
```
MaheshRavishankar, line 347: So the `user` cannot be a dispatch region into which the producer (whose results are `values`) is being fused? Thats not entirely clear to me.

MaheshRavishankar (approved):
Just had a couple of questions, but looks good to me.

IanWood1:
TODO: guard with a flag & run against torch_models tests.

compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
IanWood1: removed

compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
IanWood1: I've removed this check entirely. This utility already assumed all producers were movable so i'll keep that assumption and not check it twice.

tests/external/iree-test-suites/onnx_ops/onnx_ops_cpu_llvm_sync_O2.json
hanhanW: landed, please rebase before landing the change, thanks!

compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
IanWood1: This shouldn't matter now because its only used pre-dispatch formation. But this worked previously as the ` if (!user->isBeforeInBlock(insertionPoint))` would continue because insertionPoint is the same as user.

Issue #24053, opened 2026-04-08
[Codegen][AMDGPU] Efficient Top-K via Bitonic Sorting


bangtianliu:
## Motivation

Top-K Op is about finding the highest (or lowest) k of n tuples according to a ranking function. This is a well-studied problem in computer science, and now shines again in the LLM inference and MoE routing (selecting top-K experts). 

Within IREE, I am working on an efficient parallel top-K implementation on AMDGPU along the VectorDistribute pipeline. The existing LinalgExt `TopkOp` is used by StableHLO conversion (entry point), SplitReduction (the only optimization), and the generic tiling infrastructure. **No GPU-specific lowering exists for it**: TopkOp has no dedicated AMD GPU codegen path; it relies entirely on generic TilingInterface scalar lowering.

## Op Design: `topk_v2`
See the [Discord discussion](https://discord.com/channels/689900678990135345/1491812950451814540) for the design choice.

Here, use IR examples to illustrate how op looks like

```mlir
// Top-K with index tracking (K=8)
%0:2 = iree_linalg_ext.topk_v2 dimension(1)
  ins(%values : tensor<1024x1024xf32>)
  outs(%out_values, %out_indices : tensor<1024x8xf32>, tensor<1024x8xi32>) {
  ^bb0(%lhs: f32, %rhs: f32):
    %cmp = arith.cmpf ogt, %lhs, %rhs : f32
    iree_linalg_ext.yield %cmp : i1
} -> tensor<1024x8xf32>, tensor<1024x8xi32>
``` 
```mlir
// Top-K without index tracking (K=8)
%0 = iree_linalg_ext.topk_v2 dimension(1)
  ins(%values : tensor<1024x1024xf32>)
  outs(%out_values : tensor<1024x8xf32>) {
  ^bb0(%lhs: f32, %rhs: f32):
    %cmp = arith.cmpf ogt, %lhs, %rhs : f32
    iree_linalg_ext.yield %cmp : i1
} -> tensor<1024x8xf32>
```

```mlir
// Sorted Top-K
%0:2 = iree_linalg_ext.topk_v2 dimension(1) is_sorted
  ins(%values : tensor<1024x1024xf32>)
  outs(%out_values, %out_indices : tensor<1024x8xf32>, tensor<1024x8xi32>) {
  ^bb0(%lhs: f32, %rhs: f32):
    %cmp = arith.cmpf ogt, %lhs, %rhs : f32
    iree_linalg_ext.yield %cmp : i1
[...]

PR #24054, opened 2026-04-08
[LinalgExt] Add topk_v2 op with roundtrip and invalid mlir test

bangtianliu:
**Context**:  Refer to [Discord discussion](https://discord.com/channels/689900678990135345/1491812950451814540) for the design choice.  We aim to add a top-K op with bitonic sort vectorization along the VectorDistribute pipeline. Refer to Issue: https://github.com/iree-org/iree/issues/24053 for more details.   

This PR, as the first step, defines the` iree_linalg_ext.topk_v2` operation and defines its verifier along with corresponding roundtrip and invalid tests (`roundtrip.mlir` and `invalid.mlir`).

Support for topk_v2 will be added incrementally across multiple PRs: 

- Add the `iree_linalg_ext.topk_v2` op definition and verification logic, along with roundtrip and invalid IR tests (this PR)
- Add tiling support
- Add convert_to_loops lowering support
- Add e2e support along the VectorDistribute pipeline (vectorization, layout analysis, and distribution, etc)

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -1289,6 +1289,105 @@ MutableOperandRange TopkOp::getDpsInitsMutable() {
 }
 
 //===----------------------------------------------------------------------===//
+// TopkV2Op
+//===----------------------------------------------------------------------===//
+
+LogicalResult TopkV2Op::verify() {
+  Operation *op = getOperation();
+  auto inputValuesType = getInputType();
+  auto outputValuesType = cast<ShapedType>(getOutputValues().getType());
+
+  if (getDimension() >= static_cast<uint64_t>(getInputRank())) {
+    return op->emitOpError("dimension exceeds rank");
+  }
+  if (inputValuesType.getElementType() != outputValuesType.getElementType()) {
+    return op->emitOpError("expected input/output value types to be identical");
+  }
+  if (inputValuesType.getRank() != outputValuesType.getRank()) {
+    return op->emitOpError("expected input/output to have the same rank");
+  }
+
+  if (Value inputIndices = getInputIndices()) {
+    auto inputIndicesType = cast<ShapedType>(inputIndices.getType());
+    if (!isa<IntegerType>(inputIndicesType.getElementType())) {
+      return op->emitOpError("expected input indices to be integer type");
+    }
+    if (failed(verifyCompatibleShape(inputValuesType, inputIndicesType))) {
+      return op->emitOpError("input values/indices shape must match");
+    }
+  }
+
+  if (Value outputIndices = getOutputIndices()) {
+    auto outputIndicesType = cast<ShapedType>(outputIndices.getType());
+    if (!isa<IntegerType>(outputIndicesType.getElementType())) {
+      return op->emitOpError("expected output indices to be integer type");
+    }
+    if (failed(verifyCompatibleShape(outputValuesType, outputIndicesType))) {
+      return op->emitOpError("output values/indices shape must match");
+    }
+  }
+
+  // All dimensions except the sort dimension must match.
+  uint64_t dim = getDimension();
+  if (!llvm::all_of(
+          llvm::enumerate(llvm::zip_equal(inputValuesType.getShape(),
+                                          outputValuesType.getShape())),
+          [dim](auto e) {
+            if (e.index() == dim) {
+              return true;
+            }
+            std::tuple<int64_t, int64_t> s = e.value();
+            return succeeded(
+                verifyCompatibleShape(std::get<0>(s), std::get<1>(s)));
+          })) {
+    return op->emitOpError("incompatible input/output shapes");
+  }
+
+  // Validate that output K does not exceed input along the sort dimension.
+  int64_t inputDimSize = inputValuesType.getDimSize(dim);
+  int64_t outputDimSize = outputValuesType.getDimSize(dim);
+  if (!ShapedType::isDynamic(inputDimSize) &&
+      !ShapedType::isDynamic(outputDimSize)) {
+    if (outputDimSize == 0) {
+      return op->emitOpError("output dimension must be positive");
+    }
+    if (outputDimSize > inputDimSize) {
+      return op->emitOpError("output dimension must not exceed input, got ")
+             << outputDimSize << " > " << inputDimSize;
+    }
+  }
+
+  Block &block = getRegion().front();
+  if (block.getNumArguments() != 2) {
+    return op->emitOpError("region block should have 2 arguments");
+  }
+  if (block.getArgument(0).getType() != inputValuesType.getElementType() ||
+      block.getArgument(1).getType() != inputValuesType.getElementType()) {
+    return op->emitOpError("region block types must match input value type");
+  }
+  auto terminatorOp = dyn_cast<YieldOp>(block.getTerminator());
+  if (!terminatorOp || terminatorOp.getNumOperands() != 1 ||
+      !terminatorOp.getOperand(0).getType().isInteger(1)) {
+    return op->emitOpError("region block must end with a linalg_ext.yield i1");
+  }
+  return success();
+}
+
+LogicalResult
+TopkV2Op::reifyResultShapes(OpBuilder &b,
+                            ReifiedRankedShapedTypeDims &reifiedReturnShapes) {
+  return cast<LinalgExtOp>(getOperation())
+      .reifyResultShapes(b, reifiedReturnShapes);
+}
+
+MutableOperandRange TopkV2Op::getDpsInitsMutable() {
+  // Operands order: values, [input_indices], output_values, [output_indices]
+  unsigned numInputs = 1 + (getInputIndices() ? 1 : 0);
+  unsigned numInits = 1 + (getOutputIndices() ? 1 : 0);
+  return MutableOperandRange(*this, numInputs, numInits);
+}
+
+//===----------------------------------------------------------------------===//
 // ArgCompareOp
 //===----------------------------------------------------------------------===//
 
@@ -2944,6 +3043,7 @@ DEFINE_OP_GET_EFFECTS(SortOp)
 DEFINE_OP_GET_EFFECTS(FftOp)
 DEFINE_OP_GET_EFFECTS(ScanOp)
 DEFINE_OP_GET_EFFECTS(TopkOp)
+DEFINE_OP_GET_EFFECTS(TopkV2Op)
 DEFINE_OP_GET_EFFECTS(ArgCompareOp)
 DEFINE_OP_GET_EFFECTS(WinogradInputTransformOp)
 DEFINE_OP_GET_EFFECTS(WinogradFilterTransformOp)
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
@@ -1293,82 +1293,88 @@ MutableOperandRange TopkOp::getDpsInitsMutable() {
 //===----------------------------------------------------------------------===//
 
 LogicalResult TopkV2Op::verify() {
-  Operation *op = getOperation();
   auto inputValuesType = getInputType();
   auto outputValuesType = cast<ShapedType>(getOutputValues().getType());
+  uint64_t dim = getDimension();
 
-  if (getDimension() >= static_cast<uint64_t>(getInputRank())) {
-    return op->emitOpError("dimension exceeds rank");
+  if (dim >= static_cast<uint64_t>(getInputRank())) {
+    return emitOpError("dimension exceeds rank");
   }
   if (inputValuesType.getElementType() != outputValuesType.getElementType()) {
-    return op->emitOpError("expected input/output value types to be identical");
+    return emitOpError("expected input/output value types to be identical");
   }
   if (inputValuesType.getRank() != outputValuesType.getRank()) {
-    return op->emitOpError("expected input/output to have the same rank");
+    return emitOpError("expected input/output to have the same rank");
   }
 
   if (Value inputIndices = getInputIndices()) {
+    if (!getOutputIndices()) {
+      return emitOpError(
+          "input indices require output indices to carry provenance");
+    }
     auto inputIndicesType = cast<ShapedType>(inputIndices.getType());
     if (!isa<IntegerType>(inputIndicesType.getElementType())) {
-      return op->emitOpError("expected input indices to be integer type");
+      return emitOpError("expected input indices to be integer type");
     }
     if (failed(verifyCompatibleShape(inputValuesType, inputIndicesType))) {
-      return op->emitOpError("input values/indices shape must match");
+      return emitOpError("input values/indices shape must match");
     }
   }
 
   if (Value outputIndices = getOutputIndices()) {
     auto outputIndicesType = cast<ShapedType>(outputIndices.getType());
     if (!isa<IntegerType>(outputIndicesType.getElementType())) {
-      return op->emitOpError("expected output indices to be integer type");
+      return emitOpError("expected output indices to be integer type");
     }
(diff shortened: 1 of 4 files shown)


hanhanW:
I'm not a fan of having xxx_ext op under LinalgExt... What is the issue of having the support for `LinalgExt::SortOp`?

bangtianliu:
> I'm not a fan of having xxx_ext op under LinalgExt... What is the issue of having the support for `LinalgExt::SortOp`?

Added an issue to the PR description.   This is the first small PR relevant to the task.

The main motivation is to add top-K support with bitonic sorting, providing an efficient parallel implementation on AMDGPU along the VectorDistribute pipeline.                              
                                                                                                                                                                                               
Changing `SortOp` in place would require significant changes across multiple consumers. Once complete, we can consider to retire the original SortOp and then go back to use SortOp name.  

Any suggestions about a better name? I am open to renaming it.

hanhanW:
> > I'm not a fan of having xxx_ext op under LinalgExt... What is the issue of having the support for `LinalgExt::SortOp`?
> 
> Added an issue to the PR description. This is the first small PR relevant to the task.
> 
> The main motivation is to add top-K support with bitonic sorting, providing an efficient parallel implementation on AMDGPU along the VectorDistribute pipeline.
> 
> Changing `SortOp` in place would require significant changes across multiple consumers. Once complete, we can consider to retire the original SortOp and then go back to use SortOp name.
> 
> Any suggestions about a better name? I am open to renaming it.

Naming is not a big concern here, the concern is that what you're going to do is unclear even I read the issue and did further research myself.

E.g., I don't follow what the IR would look like after each core transformation like tiling, distribution. IIUC, the algorithm itself creates two sequential loops in the outermost, then there are oppotunities to parallelize inner computation. How are they connected? 

The implementation of current SortOp is minimal. It only support tiling on parallel dimensions and only have convertToLoop support which is scalar code. How does changing SortOp in place require significant changes in this context? Is replacing the old sort op with the new one the end goal? Or they will essentially do different things?

I can't see where the trick (and what they are) is kicked in, and what makes `SortOp` and your `SortExtOp` different in terms of transformation. They are not clear to me, and maybe it is just because I don't have enough context. (And honestly, it is a bad sign for the issue that you linked as it does not provide much values to people who has partial context.) Anyway, I'll leave the reviews and all the changes to you as you're the experts in this area. I'm just sharing what I think after I read the issue you shared, and many thanks for the context.

bangtianliu:
> Naming is not a big concern here, the concern is that what you're going to do is unclear even I read the issue and did further research myself.
> 
> E.g., I don't follow what the IR would look like after each core transformation like tiling, distribution. IIUC, the algorithm itself creates two sequential loops in the outermost, then there are oppotunities to parallelize inner computation. How are they connected?
> 
> The implementation of current SortOp is minimal. It only support tiling on parallel dimensions and only have convertToLoop support which is scalar code. How does changing SortOp in place require significant changes in this context? Is replacing the old sort op with the new one the end goal? Or they will essentially do different things?
> 
> I can't see where the trick (and what they are) is kicked in, and what makes `SortOp` and your `SortExtOp` different in terms of transformation. They are not clear to me, and maybe it is just because I don't have enough context. (And honestly, it is a bad sign for the issue that you linked as it does not provide much values to people who has partial context.) Anyway, I'll leave the reviews and all the changes to you as you're the experts in this area. I'm just sharing what I think after I read the issue you shared, and many thanks for the context.

Thanks for the detailed feedback. These are very fair points, and I appreciate them very much. Apologies for my earlier reply, since it was written without enough thought.

So the core question from your feedback is: **Why not modify SortOp in place?** 

The two ops have fundamentally different structures, at least according to the current PR changes. 
I will illustrate with IR examples. 

SortOp does **in-place** sorting, in which the data is passed as `outs` with no separate `ins` (the verifier enforces `getNumDpsInputs() == 0`).

```mlir
// Sort values and indices together, 4 block args (2 per output)
%0:2 = iree_linalg_ext.sort dimension(1)
  outs(%values, %indices : tensor<?x10xf32>, tensor<?x10xi64>) {
  ^bb0(%arg0: f32, %arg1: f32, %arg2: i64, %arg3: i64):
    %cmp = arith.cmpf oge, %arg0, %arg1 : f32
} -> tensor<?x10xf32>, tensor<?x10xi64>
```

SortExtOp has separate `ins` and `outs`, always takes exactly 2 block arguments, with index tracking and top-k as first-class optional operands:

```mlir
// Full sort with index tracking, still only 2 block args
%0:2 = iree_linalg_ext.sort_ext dimension(1)
  ins(%values : tensor<?x10xf32>)
  outs(%out_values, %out_indices : tensor<?x10xf32>, tensor<?x10xi64>) {
  ^bb0(%lhs: f32, %rhs: f32):
    %cmp = arith.cmpf oge, %lhs, %rhs : f32
} -> tensor<?x10xf32>, tensor<?x10xi64>

// Top-K mode,  output sort dim (5) < input sort dim (10)
%1:2 = iree_linalg_ext.sort_ext dimension(1)
  top_k(%k : index)
  ins(%values : tensor<?x10xf32>)
  outs(%out_values, %out_indices : tensor<?x5xf32>, tensor<?x5xi64>) {
  ^bb0(%lhs: f32, %rhs: f32):
    %cmp = arith.cmpf oge, %lhs, %rhs : f32
} -> tensor<?x5xf32>, tensor<?x5xi64>
```

I used AI to summarize one table about the differences: 

| | SortOp | SortExtOp |
|---|---|---|
| **Input/Output style** | In-place: data passed as `outs` only, no `ins` | Separate `ins` and `outs` |
| **Outputs** | Variadic (multi-tensor co-sort) | Single output values + optional output indices |
| **Block args** | `2 * numOutputs` (a pair per co-sorted tensor) | Always exactly 2 |
| **Index tracking** | Via co-sorting a second tensor as output | First-class optional operand with integer type constraint |
| **Top-K** | Not supported; verifier requires all outputs have identical shapes | Optional; output sort dim can be smaller than input sort dim |
| **Lowering** | Bubble sort via `generateScalarImplementation` | Intended: bitonic sort network for GPU parallelism |

My previous statement about "significant changes" is wrong.  The cost of changing SortOp isn't really the point. The point is they serve different purposes. SortOp handles frontend lowering (StableHLO/Torch), SortExtOp is designed for efficient GPU sort/top-k lowering. 

The exact design of the new op does need feedback to finalize, but I think the goals I want to support are clear:
- Top-K support 
- Efficient lowering via bitonic sort on the AMDGPU backend along the VectorDistribute pipeline

The criticism about the transformation pipeline clarity is very fair. I will update the issue (previously written in a rush). 

cc @Groverkss here, since you motivated this task, please correct any of the above if you think I've misstated something. Also would appreciate your feedback on the current op design.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -680,6 +680,73 @@ def IREELinalgExt_TopkOp : IREELinalgExt_Op<"topk",[
...
+    `comparator`. Optionally selects only the top-K elements when `top_k` is
+    specified.
+
+    Accepts a single N-D input tensor of values and an optional N-D tensor of
+    input indices (any integer type). If input indices aren't provided, the
+    index mapping is inferred based on the sort dimension. Produces sorted
+    output values and optionally output indices (integer type) tracking
+    original positions.
+
+    When `top_k` is specified, only the top K elements along the sort dimension
```
MaheshRavishankar: Are we doing two operations in one here? It seems like the basic algorithm for top_k and full sort should be very different? I know top_k == input dimension effectively gives you sort, but top_k just needs top_k values. Do those need to be sorted too?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -680,6 +680,73 @@ def IREELinalgExt_TopkOp : IREELinalgExt_Op<"topk",[
...
+  DeclareOpInterfaceMethods<LinalgExtInterface>
+]>{
+  let summary = [{Unified sort and top-k operator.}];
+  let description = [{
+    Sorts the input tensor along the given `dimension` using the given
+    `comparator`. Optionally selects only the top-K elements when `top_k` is
+    specified.
+
+    Accepts a single N-D input tensor of values and an optional N-D tensor of
+    input indices (any integer type). If input indices aren't provided, the
```
MaheshRavishankar, line 701: I am not sure I follow what `input indices` are?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -680,6 +680,73 @@ def IREELinalgExt_TopkOp : IREELinalgExt_Op<"topk",[
...
+    type and yields an i1 indicating whether the first argument should be
+    ordered before the second.
+  }];
+
+  let arguments = (ins AnyShaped:$values,
+                       Optional<AnyShaped>:$input_indices,
+                       Optional<Index>:$top_k,
+                       AnyShaped:$output_values,
+                       Optional<AnyShaped>:$output_indices,
+                       I64Attr:$dimension
```
MaheshRavishankar: Nit: can we keep the order of operands consistent with the order they appear in the assembly format below? Change the assembly format if you need to.

MaheshRavishankar:
Some points about the previous discussion on difference from SortOp. To clarify , I am not saying we should not add a new op. Its fine if the old op doesnt work and you want a new operation.

1. I think the op your defining is really top-k and not really a sort. In the extreme top-k becomes a sort, but I think the algorithm for top-k and algorithm for sort are totally different. Fundamentally, in top-k do you need the "k" to be sorted as well? I think you dont need that, so then the complexity of top-k is much smaller than the complexity of sort.
2. If you take sort, the output shape is always the input shape. then passing input values separately means that you cant really make it destination passing style. If the op has to have "similar" semantics for tensor and memref, then you cant really have an `ins` operand and an `outs` operand and support in-place sort. You can break an in-place sort into an out-of-place sort by adding a explicit out-of-place sort operation, but that was not what the original sort-operation did (cause it really wanted to model in-place sort).

I do think here we are conflating top-k and sort. For top-k obviously you cant have "in-place top-k". You need a separate ins operand and a separate `outs` operand.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
bangtianliu: It is needed for partial reduction (split reduction). Same semantics as the existing TopkOp's optional indices operand.

bangtianliu:
This PR adds the `topk_v2 `op to the LinalgExt dialect, replacing the earlier sort_ext approach. See the [Discord discussion](https://discord.com/channels/689900678990135345/1491812950451814540) for the design context.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -680,6 +680,73 @@ def IREELinalgExt_TopkOp : IREELinalgExt_Op<"topk",[
...
+    output indices are optional instead of mandatory, input indices of any
+    integer type are supported for pre-permuted inputs, and the comparator
+    region uses a pure ordering predicate decoupled from the swap semantics
+    of the legacy op.
+
+    Accepts a single N-D input tensor of values and an optional N-D tensor of
+    input indices (any integer type). If input indices aren't provided, the
+    index mapping is inferred based on the sort dimension. Produces output
+    values and optionally output indices (integer type) tracking original
+    positions.
```
hanhanW:
```suggestion
    Accepts a single N-D input tensor of values and an optional N-D tensor of
    input indices (any integer type). If input indices aren't provided, the
    index mapping is inferred based on the sort dimension (i.e., `[0, N)`). Produces output
    values and optionally output indices (integer type) tracking original
    positions.
```

A couple questions:

1. What does it mena when input indices is passed but not output indices? Should we deny the case?
2. If input indices aren't provided, `[0, N)` is inferred, right? How about adding it to the doc?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1289,6 +1288,105 @@ MutableOperandRange TopkOp::getDpsInitsMutable() {
...
+                verifyCompatibleShape(std::get<0>(s), std::get<1>(s)));
+          })) {
+    return op->emitOpError("incompatible input/output shapes");
+  }
+
+  // Validate that output K does not exceed input along the sort dimension.
+  int64_t inputDimSize = inputValuesType.getDimSize(dim);
+  int64_t outputDimSize = outputValuesType.getDimSize(dim);
+  if (!ShapedType::isDynamic(inputDimSize) &&
+      !ShapedType::isDynamic(outputDimSize)) {
```
hanhanW: nit: avoid double negation.

```suggestion
  if (ShapedType::isStatic(inputDimSize) &&
      ShapedType::isStatic(outputDimSize)) {
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1289,6 +1288,105 @@ MutableOperandRange TopkOp::getDpsInitsMutable() {
...
+  }
+  if (block.getArgument(0).getType() != inputValuesType.getElementType() ||
+      block.getArgument(1).getType() != inputValuesType.getElementType()) {
+    return op->emitOpError("region block types must match input value type");
+  }
+  auto terminatorOp = dyn_cast<YieldOp>(block.getTerminator());
+  if (!terminatorOp || terminatorOp.getNumOperands() != 1 ||
+      !terminatorOp.getOperand(0).getType().isInteger(1)) {
+    return op->emitOpError("region block must end with a linalg_ext.yield i1");
+  }
```
hanhanW, line 1378: This seems to belong `verifyRegions` check, then you can use `cast` instead of `dyn_cast`. I learned it from @sommerlukas early this year: https://github.com/iree-org/iree/commit/7553f058e042080ae050d34cc42848d8a13c5c7c

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -1289,6 +1288,105 @@ MutableOperandRange TopkOp::getDpsInitsMutable() {
...
+          [dim](auto e) {
+            if (e.index() == dim) {
+              return true;
+            }
+            std::tuple<int64_t, int64_t> s = e.value();
+            return succeeded(
+                verifyCompatibleShape(std::get<0>(s), std::get<1>(s)));
+          })) {
+    return op->emitOpError("incompatible input/output shapes");
+  }
```
hanhanW: Having complex lambda in if statement is usually hard to read the code. And the code will be mutated easily when people update the code, because of formatting. How about just converting it to a loop? E.g.,

```cpp
  uint64_t dim = getDimension();
  for (auto [idx, inDim, outDim] : llvm::enumerate(
           inputValuesType.getShape(), outputValuesType.getShape())) {
    if (idx == dim) {
      continue;
    }
    if (failed(verifyCompatibleShape(inDim, outDim))) {
      return op->emitOpError("incompatible input/output shapes");
    }
  }
```

In this case, you can also signal which dimension is problematic.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
bangtianliu: Good catch! Say yes to both the two Qs, and will apply the changes.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.cpp
bangtianliu: TIL

hanhanW (approved):
LGTM. Please land it next week. The final version is done on late Friday (US time), so I think maybe others may want to take a look on Monday.

bangtianliu:
cc @MaheshRavishankar @Groverkss

Groverkss (approved):
LGTM.

Issue #24051, opened 2026-04-08
iree-compile crashes on iree_linalg_ext.attention with broadcasting


rkayaith:
`iree-compile` segfaults when compiling an SDPA op with broadcasted batch and head dims, where Q head dim size differs from V head dim size (d_qk=128, d_v=64).

Input IR:

<details>
<summary>input.mlir</summary>

```mlir
module @module {

  func.func private @sdpa_fprop(
      %arg0: !torch.vtensor<[8,2,64,128],bf16>,
      %arg1: !torch.vtensor<[8,2,16,128],bf16>,
      %arg2: !torch.vtensor<[8,2,16,64],bf16>)
      -> !torch.vtensor<[8,2,64,64],bf16> {
    %none_mask = torch.constant.none
    %dropout = torch.constant.float 0.000000e+00
    %is_causal = torch.constant.bool true
    %scale = torch.constant.float 8.838835e-02
    %enable_gqa = torch.constant.bool false
    %0 = torch.aten.scaled_dot_product_attention %arg0, %arg1, %arg2,
        %none_mask, %dropout, %is_causal, %scale, %enable_gqa :
        !torch.vtensor<[8,2,64,128],bf16>, !torch.vtensor<[8,2,16,128],bf16>,
        !torch.vtensor<[8,2,16,64],bf16>, !torch.none, !torch.float, !torch.bool,
        !torch.float, !torch.bool -> !torch.vtensor<[8,2,64,64],bf16>
    return %0 : !torch.vtensor<[8,2,64,64],bf16>
  }

  func.func @main(%o_3_: !torch.tensor<[2,8,64,64],bf16>, %k_1: !torch.vtensor<[1,1,16,128],bf16>, %q_0: !torch.vtensor<[1,1,64,128],bf16>, %v_2: !torch.vtensor<[1,1,16,64],bf16>) attributes {torch.assume_strict_symbolic_shapes} {

    %permute_IN_0_val_0_sdpa_fprop_i0 = torch.constant.int 0
    %permute_IN_0_val_1_sdpa_fprop_i0 = torch.constant.int 1
    %permute_IN_0_val_2_sdpa_fprop_i0 = torch.constant.int 2
    %permute_IN_0_val_3_sdpa_fprop_i0 = torch.constant.int 3
[...]

PR #24059, opened 2026-04-09
[LinalgExt] Fix FuseTransposeWithAttentionOp skipping valid operands

IanWood1:
Move the projected permutation check for the producer's indexing maps inside the operand search loop. Previously, the loop selected the first input with a single-yield generic producer, then validated its indexing maps outside the loop. If validation failed, the pattern returned failure without considering subsequent inputs that may have had valid producers.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/test/elementwise_op_fusion.mlir
+++ b/compiler/src/iree/compiler/DispatchCreation/test/elementwise_op_fusion.mlir
@@ -316,6 +316,63 @@ util.func public @fuse_attention_with_broadcast_transpose(%arg0: tensor<4x?x8x12
 
 // -----
 
+// Verify that when Q's producer has a non-projected-permutation input map
+// (constant 0), the pattern skips it and fuses K's transpose instead.
+util.func public @skip_non_projected_permutation_producer(
+    %q: tensor<1x32x64x128xf16>,
+    %k: tensor<4x64x32x128xf16>,
+    %v: tensor<4x32x64x128xf16>,
+    %scale: f16) -> tensor<4x32x64x128xf16> {
+  %empty_q = tensor.empty() : tensor<4x32x64x128xf16>
+  %q_prod = linalg.generic {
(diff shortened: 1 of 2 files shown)


Groverkss (commented):
can you add a test

Groverkss (commented):
can you add a test

Groverkss (commented):
can you add a test

compiler/src/iree/compiler/DispatchCreation/test/elementwise_op_fusion.mlir
```diff
@@ -316,6 +316,63 @@ util.func public @fuse_attention_with_broadcast_transpose(%arg0: tensor<4x?x8x12
...
+// Verify that when Q's producer has a non-projected-permutation input map
+// (constant 0), the pattern skips it and fuses K's transpose instead.
+util.func public @skip_non_projected_permutation_producer(
+    %q: tensor<1x32x64x128xf16>,
+    %k: tensor<4x64x32x128xf16>,
+    %v: tensor<4x32x64x128xf16>,
+    %scale: f16) -> tensor<4x32x64x128xf16> {
+  %empty_q = tensor.empty() : tensor<4x32x64x128xf16>
+  %q_prod = linalg.generic {
+      indexing_maps = [affine_map<(d0, d1, d2, d3) -> (0, d1, d2, d3)>,
```
MaheshRavishankar, line 328: EEEESH! this is a broadcast :D . Where did we get the original issue from?

MaheshRavishankar (approved):

compiler/src/iree/compiler/DispatchCreation/test/elementwise_op_fusion.mlir
IanWood1: I just made up the input IR. The `0` in the indexing map should be simplified before this pass was ever run, but this was the first example I thought of that would trigger the unexpected behavior. I found this purely by looking at the code when investigating https://github.com/iree-org/iree/issues/24051, so I don't have a real life example.

PR #24060, opened 2026-04-09
[Encoding] Encoding infra for convolution data tiling

jschuhmacher:
Part 1 of introducing data tiling support for convolutions.

Split off from https://github.com/iree-org/iree/pull/23746, see there for full context.


jschuhmacher:
Merging with https://github.com/iree-org/iree/pull/24061#pullrequestreview-4084353962

PR #24061, opened 2026-04-09
Add data tiling hints for convolutions

jschuhmacher:
Parts 1 and 2 of introducing data tiling support for convolutions.

- Adds encoding infra for convolution data tiling.
- Adds data tiling hints for convolutions.
- Updates AnnotateDispatches pass for new encodings.

Split off from https://github.com/iree-org/iree/pull/23746, see there for full context.

Assisted by: Claude

### DIFF
--- a/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
+++ b/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
@@ -760,14 +760,6 @@ struct CPUEncodingResolverMaterializerAttr final
       return lowerFillOpWithResolvedLayouts(b, fillOp, convertedResTypes,
                                             convertedOperands);
     }
-    // Convolution is not yet supported, so we drop the encoding and clone the
-    // op as-is.
-    if (linalg::isaConvolutionOpInterface(linalgOp)) {
-      int64_t numInputs = linalgOp.getNumDpsInputs();
-      return dropEncodingAndCloneOp(b, linalgOp,
-                                    convertedOperands.take_front(numInputs),
-                                    convertedOperands.drop_front(numInputs));
-    }
     // Scaled contraction (MX matmul) is not yet supported on CPU, so we drop
     // the encoding and clone the op as-is.
     if (IREE::LinalgExt::isaScaledContractionOpInterface(linalgOp)) {
@@ -943,14 +935,6 @@ struct VMVXEncodingResolverMaterializerAttr final
     if (auto fillOp = dyn_cast<linalg::FillOp>(op)) {
       return lowerFillOpWithResolvedLayouts(b, fillOp, convertedResTypes,
                                             convertedOperands);
-    }
-    // Convolution is not yet supported, so we drop the encoding and clone the
-    // op as-is.
-    if (linalg::isaConvolutionOpInterface(linalgOp)) {
-      int64_t numInputs = linalgOp.getNumDpsInputs();
-      return dropEncodingAndCloneOp(b, linalgOp,
-                                    convertedOperands.take_front(numInputs),
-                                    convertedOperands.drop_front(numInputs));
     }
     if (linalg::isaContractionOpInterface(linalgOp)) {
       return lowerContractionOpWithEncoding(
--- a/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
+++ b/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
@@ -760,6 +760,14 @@ struct CPUEncodingResolverMaterializerAttr final
       return lowerFillOpWithResolvedLayouts(b, fillOp, convertedResTypes,
                                             convertedOperands);
     }
+    // Convolution is not yet supported, so we drop the encoding and clone the
+    // op as-is.
+    if (linalg::isaConvolutionOpInterface(linalgOp)) {
+      int64_t numInputs = linalgOp.getNumDpsInputs();
+      return dropEncodingAndCloneOp(b, linalgOp,
+                                    convertedOperands.take_front(numInputs),
+                                    convertedOperands.drop_front(numInputs));
+    }
     // Scaled contraction (MX matmul) is not yet supported on CPU, so we drop
     // the encoding and clone the op as-is.
     if (IREE::LinalgExt::isaScaledContractionOpInterface(linalgOp)) {
@@ -935,6 +943,14 @@ struct VMVXEncodingResolverMaterializerAttr final
     if (auto fillOp = dyn_cast<linalg::FillOp>(op)) {
       return lowerFillOpWithResolvedLayouts(b, fillOp, convertedResTypes,
                                             convertedOperands);
+    }
+    // Convolution is not yet supported, so we drop the encoding and clone the
+    // op as-is.
+    if (linalg::isaConvolutionOpInterface(linalgOp)) {
+      int64_t numInputs = linalgOp.getNumDpsInputs();
+      return dropEncodingAndCloneOp(b, linalgOp,
+                                    convertedOperands.take_front(numInputs),
+                                    convertedOperands.drop_front(numInputs));
     }
     if (linalg::isaContractionOpInterface(linalgOp)) {
       return lowerContractionOpWithEncoding(
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
@@ -148,10 +148,6 @@ SerializableAttr::getEncodingProperties(Operation *op) {
 
   // Return encoding properties for convolutions operations.
   if (linalg::isaConvolutionOpInterface(linalgOp)) {
-    auto cDims = linalg::inferConvolutionDims(linalgOp);
-    if (failed(cDims) || cDims->outputImage.size() != 2) {
-      return failure();
-    }
     Type inputElemType =
         getElementTypeOrSelf(linalgOp.getDpsInputOperand(0)->get().getType());
     Type filterElemType =
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
@@ -6,7 +6,6 @@
 
 #include "iree/compiler/Dialect/Encoding/IR/EncodingTypes.h"
 
-#include "EncodingTypes.h"
 #include "iree/compiler/Dialect/LinalgExt/Utils/MatchUtils.h"
 #include "llvm/ADT/SmallVector.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
@@ -205,11 +205,11 @@ static std::string stringifyConvOperandIndex(IntegerAttr valueAttr) {
   uint64_t value = valueAttr.getValue().getZExtValue();
   switch (value) {
   case CONV_IN:
-    return "IN";
+    return "CONV_IN";
   case CONV_FILTER:
-    return "FILTER";
+    return "CONV_FILTER";
   case CONV_OUT:
-    return "OUT";
+    return "CONV_OUT";
   default:
     assert(false && "invalid index");
     return "";
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
@@ -225,7 +225,7 @@ std::string stringifyOperandIndex(EncodingOpType opType,
   case EncodingOpType::conv:
     return stringifyConvOperandIndex(valueAttr);
   default:
-    assert(false && "invalid index");
+    assert(false && "invalid opType");
     return "";
   }
 }
--- a/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
+++ b/compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
@@ -146,6 +146,26 @@ SerializableAttr::getEncodingProperties(Operation *op) {
     return props;
   }
 
+  // Return encoding properties for convolutions operations.
+  if (linalg::isaConvolutionOpInterface(linalgOp)) {
+    auto cDims = linalg::inferConvolutionDims(linalgOp);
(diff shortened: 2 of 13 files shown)


compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
```diff
@@ -160,8 +180,7 @@ std::string stringifyOperandIndex(IntegerAttr valueAttr) {
   case MATMUL_RESULT:
     return "RESULT";
   default:
-    assert(false && "invalid index");
-    return "";
+    return "OPERAND" + std::to_string(value);
```
hanhanW: What happens to this? I think we'll need to updat ethe method to take OpType as well, so we can have accurate naming in `[Flow/Transforms/AnnotateDispatches.cpp](https://github.com/iree-org/iree/blob/b518b35c39adf84a44275a8a2fbde2cf49126f6e/compiler/src/iree/compiler/Dialect/Flow/Transforms/AnnotateDispatches.cpp#L454)`.

I wonder why you need the change, and we may need to separate the change with a test. If it is causing issue and we have to land it this way, I'll send a PR for the change in a follow-up.

hanhanW (changes requested):
We need a `ListOption` flag for the pass and `DispatchCreation/Passes`. The option only enables matmul and scaled_matmul on by default, which matches the current behavior.

The main reason is that it will regress model performance if both conv and matmul show up in the model. Because there may be additional dispatches or some dispatches may result in additional results (one with encodings and the other without encodings). Since we just start the work for convolution, I think it is better to guard it under a flag during development. We can enable it once we have more confident.

For `ListOption`, I wonder if we can use `enum class`. If it is getting tricky, string is also fine.

Note, that parent PR can be collapsed into this one. Otherwise, that code is deadcode in a single PR.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
jschuhmacher: I had the assert going off during development, but in the current state it really shouldn't trigger. I'll remove this, and add the OpType change too.

jschuhmacher:
> We need a `ListOption` flag for the pass and `DispatchCreation/Passes`. The option only enables matmul and scaled_matmul on by default, which matches the current behavior.
> 
> The main reason is that it will regress model performance if both conv and matmul show up in the model. Because there may be additional dispatches or some dispatches may result in additional results (one with encodings and the other without encodings). Since we just start the work for convolution, I think it is better to guard it under a flag during development. We can enable it once we have more confident.
> 
> For `ListOption`, I wonder if we can use `enum class`. If it is getting tricky, string is also fine.

Thanks, I've added the list option. Using an enum class took a bit of trying, but it is works.

> Note, that parent PR can be collapsed into this one. Otherwise, that code is deadcode in a single PR.

Ok, I'll close the parent PR, it's contained in here as well.

compiler/src/iree/compiler/DispatchCreation/Passes.h
```diff
@@ -19,6 +21,45 @@ namespace mlir::iree_compiler::DispatchCreation {
...
 enum class EncodingOptions { Padding, Generic };
 
+// Custom cl::parser for IREE::Encoding::EncodingOpType. Enables use of enum
+// class in cl::list.
+struct EncodingOpTypeParser
+    : public llvm::cl::parser<IREE::Encoding::EncodingOpType> {
+  using llvm::cl::parser<IREE::Encoding::EncodingOpType>::parser;
+
+  void initialize() {
+    using OpType = IREE::Encoding::EncodingOpType;
```
hanhanW: Is is possible to move it to the global of the struct? E.g., line 29.

compiler/src/iree/compiler/DispatchCreation/Passes.h
```diff
@@ -19,6 +21,45 @@ namespace mlir::iree_compiler::DispatchCreation {
...
+    case OpType::conv:
+      os << "convolution";
+      break;
+    }
+  }
+};
+
+struct AnnotateDataTilingHintsPassOptions {
+  llvm::SmallVector<IREE::Encoding::EncodingOpType> opTypes;
+};
```
hanhanW: What happens if we have the option in `Passes.td`, so we can use the one generated by tablegen?

compiler/src/iree/compiler/DispatchCreation/Passes.h
```diff
@@ -42,6 +83,18 @@ struct TransformOptions : PassPipelineOptions<TransformOptions> {
...
+          *this,
+          "data-tiling-op-types",
+          llvm::cl::desc(
+              "Op families eligible for data-tiling annotation. "
+              "Defaults to {matmul, scaled_matmul}; add 'convolution' to "
+              "enable convolution annotation."),
+          llvm::cl::list_init<IREE::Encoding::EncodingOpType>(
+              {IREE::Encoding::EncodingOpType::matmul,
+               IREE::Encoding::EncodingOpType::scaled_matmul}),
+      };
```
hanhanW: I was thinking that we start with a global `llvm::cl` option and mark it experimental. This way you don't have to deal with the plumb through correctly during the development.

I think for the current snippet, you would be suffer when you use `iree-compile`.

compiler/src/iree/compiler/DispatchCreation/AnnotateDataTilingHints.cpp
```diff
@@ -167,17 +211,42 @@ static bool isSupportedScaledContractionOp(linalg::LinalgOp linalgOp) {
...
-                     isSupportedScaledContractionOp(linalgOp))) {
-      candidates.push_back(op);
+    if (!linalgOp) {
       return WalkResult::advance();
     }
+    const bool matches =
+        (enableMatmul && isSupportedContractionOp(linalgOp)) ||
+        (enableScaledMatmul && isSupportedScaledContractionOp(linalgOp)) ||
+        (enableConvolution && isSupportedConvolutionOp(linalgOp));
+    if (matches) {
```
hanhanW: maybe just inline the check?

hanhanW (commented):
The changes in `compiler/src/iree/compiler/Codegen/ExteranlInterfaces` can be dropped. This PR focuses more on SetEncoding part.

We should either isolate the `AnnoatateDispatches.cpp` changes to a separate PR or add tests to [annotate_dispatches.mlir](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/Flow/Transforms/test/annotate_dispatches.mlir). Both work for me since I already reviewed the code.

compiler/src/iree/compiler/DispatchCreation/Passes.h
jschuhmacher: Unfortunately, that doesn't work out of the box for ListOptions containing enum classes (yet). I'll try to move as much as possible to tablegen.

compiler/src/iree/compiler/DispatchCreation/Passes.h
jschuhmacher: I added a separate experimental global flag that adds convolutions so that it can be used from iree-compile too. Does that work?

jschuhmacher:
> The changes in `compiler/src/iree/compiler/Codegen/ExteranlInterfaces` can be dropped. This PR focuses more on SetEncoding part.
> 
> We should either isolate the `AnnoatateDispatches.cpp` changes to a separate PR or add tests to [annotate_dispatches.mlir](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/Flow/Transforms/test/annotate_dispatches.mlir). Both work for me since I already reviewed the code.

I've now added the test cases for it.

compiler/src/iree/compiler/DispatchCreation/Passes.h
jschuhmacher: Aside from this addition https://github.com/iree-org/iree/pull/24061/changes#diff-0a6c1157ccc694f7596b1e6ba5139fc61da5ede36a26391cd2e5094983ae84f5R20 most of it could be moved to tablegen

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
hanhanW, line 1: What happens to these Codegen files? See https://github.com/iree-org/iree/pull/24061#pullrequestreview-4108144215

compiler/src/iree/compiler/Dialect/Flow/Transforms/test/annotate_dispatches.mlir
```diff
@@ -150,6 +150,169 @@ flow.executable private @ex1 {
...
+      %3 = iree_encoding.unset_encoding %2 : tensor<?x?xf32, #encoding_sm_result> -> tensor<?x?xf32>{%arg1, %arg2}
+      iree_tensor_ext.dispatch.tensor.store %3, %1, offsets = [0, 0], sizes = [%arg1, %arg2], strides = [1, 1] : tensor<?x?xf32> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<?x?xf32>>{%arg1, %arg2}
+      return
+    }
+  }
+}
+
+// -----
+
+// Encoding ops for convolutions
```
hanhanW:
```suggestion
// Encoding ops for convolutions.
```

compiler/src/iree/compiler/Dialect/Flow/Transforms/test/annotate_dispatches.mlir
```diff
@@ -150,6 +150,169 @@ flow.executable private @ex1 {
...
+      %3 = iree_encoding.unset_encoding %2 : tensor<?x?xf32, #encoding_result> -> tensor<?x?xf32>{%arg1, %arg2}
+      iree_tensor_ext.dispatch.tensor.store %3, %1, offsets = [0, 0], sizes = [%arg1, %arg2], strides = [1, 1] : tensor<?x?xf32> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<?x?xf32>>{%arg1, %arg2}
+      return
+    }
+  }
+}
+
+// -----
+
+// Encoding ops for scaled matmul
```
hanhanW:
```suggestion
// Encoding ops for scaled matmul.
```

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
```diff
@@ -202,7 +202,10 @@ void buildGlobalOptimizationPassPipeline(
   // Enable data tiling after they are in a canonical form.
   if (transformOptions.dataTiling) {
     FunctionLikeNest(mainPassManager)
-        .addPass(DispatchCreation::createAnnotateDataTilingHintsPass)
+        .addPass([]() {
+          return DispatchCreation::createAnnotateDataTilingHintsPass(
+              DispatchCreation::AnnotateDataTilingHintsPassOptions{});
+        })
```
hanhanW: I think we no longer need this change?

compiler/src/iree/compiler/DispatchCreation/Passes.h
```diff
@@ -42,6 +74,17 @@ struct TransformOptions : PassPipelineOptions<TransformOptions> {
...
+      *this,
+      "data-tiling-op-types",
+      llvm::cl::desc(
+          "Op families eligible for data-tiling annotation. "
+          "Defaults to {matmul, scaled_matmul}; add 'convolution' to "
+          "enable convolution annotation."),
+      llvm::cl::list_init<IREE::Encoding::EncodingOpType>(
+          {IREE::Encoding::EncodingOpType::matmul,
+           IREE::Encoding::EncodingOpType::scaled_matmul}),
+  };
```
hanhanW: I'm thinking something simpler. We don't need the pipeline option, so they can be dropped.

The global experiemental flag lives in `Passes.cpp` and we overwrite the `*PassOption` with the value when we add the pass.

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.h
```diff
@@ -20,6 +20,8 @@
 
 namespace mlir::iree_compiler::IREE::Encoding {
 
+enum class EncodingOpType : uint32_t;
```
hanhanW: I think this is needed because we include `Attrs.h.inc` at the end. I was going to ask you if we can move it above like `StreamTypes.h`. It may run into rabbit holes, so I can take it over (in a follow-up) if it is tricky.

https://github.com/iree-org/iree/blob/f3220d8f29bcdc4ad7067052a14a0d51a81f5458/compiler/src/iree/compiler/Dialect/Stream/IR/StreamTypes.h#L34-L37

compiler/src/iree/compiler/DispatchCreation/Passes.h
hanhanW: (Feel free to reach out on discord if some comments are not clear. I don't want to waste your time.)

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
jschuhmacher: They were here to remove unsupported convolution encoding from the IR until their support landed in the next PRs. Now with the added experimental flag I'll remove just it.

jschuhmacher: Removed

compiler/src/iree/compiler/GlobalOptimization/Passes.cpp
jschuhmacher: Removed it.

compiler/src/iree/compiler/DispatchCreation/Passes.h
jschuhmacher: Much appreciated, we don't have too much timezone overlap though. I think I got what you meant.

jschuhmacher:
Thanks for taking the time @hanhanW

compiler/src/iree/compiler/Dialect/Encoding/IR/EncodingTypes.cpp
```diff
@@ -165,6 +181,55 @@ std::string stringifyOperandIndex(IntegerAttr valueAttr) {
...
+                                  IntegerAttr valueAttr) {
+  switch (opType) {
+  case EncodingOpType::matmul:
+    return stringifyMatmulOperandIndex(valueAttr);
+  case EncodingOpType::scaled_matmul:
+    return stringifyScaledMatmulOperandIndex(valueAttr);
+  case EncodingOpType::conv:
+    return stringifyConvOperandIndex(valueAttr);
+  default:
+    assert(false && "invalid index");
```
hanhanW:
```suggestion
    assert(false && "invalid opType");
```

hanhanW (approved):
Thanks, just final nit that I missed in the previous reviews. LGTM, nice patch!

hanhanW:
Let me know if you need me to help land the PR

jschuhmacher:
> Let me know if you need me to help land the PR

Thanks, that would be great, I don’t think I have the required permissions.

PR #24065, opened 2026-04-09
[LinalgExt] Fix getIterationDomain() for broadcast batch dims

IanWood1:
`getIterationDomain` only used Q/K/V to determine iteration bounds. After broadcast fusion removes batch dims from Q/K/V (leaving them only in mask/output maps), the batch dim bounds were never populated, causing a null dereference during tiling. Include all operands when computing iteration bounds for both `AttentionOp` and `OnlineAttentionOp`.

Fixes crash reported in https://github.com/iree-org/iree/issues/24051

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
@@ -2757,12 +2757,10 @@ static Operation *getPermutedSlice(OpBuilder &b, Location loc, Value val,
 //===----------------------------------------------------------------------===//
 
 SmallVector<Range> AttentionOp::getIterationDomain(OpBuilder &b) {
-  // Attention shape can be determined from Q, K, V alone.
-  SmallVector<Value> shapedValues = {getQuery(), getKey(), getValue()};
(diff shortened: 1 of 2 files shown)


Groverkss (changes requested):
This should be a verifier error. AttentionOp shouldn't do broadcasts like these. If there are inputs/outputs that have dimensions other than those that Q, K, V have, then it's written wrong. I don't think we should support this.

IanWood1:
> This should be a verifier error. AttentionOp shouldn't do broadcasts like these. If there are inputs/outputs that have dimensions other than those that Q, K, V have, then it's written wrong. I don't think we should support this.

Would you be okay with a case where there are no batch dims at all?

MaheshRavishankar:
> This should be a verifier error. AttentionOp shouldn't do broadcasts like these. If there are inputs/outputs that have dimensions other than those that Q, K, V have, then it's written wrong. I don't think we should support this.

In the general case the batch dimension missing from all the inputs and outputs shouldnt matter right? The backend code-generation should be able to handle it seemlessly? If there is no batch dimension in inputs and outputs, it is like a batch dimension of 1.

But, if all the inputs are broadcasts as it is in this example, then maybe we should not fuse the broadcasts to the output this way. I dont fully remember the historical context for this. We might need to go rediscover that.

PR #24067, opened 2026-04-09
[CPU] Improve splitReduction for bounded-dynamic cases.

### LINKED ISSUE #24010: LLVMCPUSplitReduction does not handle bounded dynamic reductions
# Task: Enable LLVMCPUSplitReduction for bounded dynamic reductions

Prototype: https://github.com/hanhanW/iree/commit/626bf42c28144852e6b07060bf5d5b566bf4118f

## Problem

`LLVMCPUSplitReduction` (in `compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp`)
fails on dynamic `tensor<?xf32>` inputs because the upstream `linalg::splitReduction`
requires static shapes. However, the dynamic dimension is often bounded at compile time.

### Concrete example

sum 1M f32 with split-reduction tile_size=124992 (8 chunks):

```mlir
// The affine.min tells us dim is bounded by 124992 (which is 7812 × 16)
%4 = affine.min affine_map<(d0) -> (-d0 + 1000000, 124992)>(%arg0)
%5 = iree_tensor_ext.dispatch.tensor.load %0, offsets = [%arg0], sizes = [%4], strides = [1]
    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<1000000xf32>> -> tensor<?xf32>

// This linalg.generic has a single reduction dim over tensor<?xf32>
%13 = linalg.generic {
    indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> ()>],
    iterator_types = ["reduction"]
} ins(%5 : tensor<?xf32>) outs(%12 : tensor<f32>)
  attrs = {lowering_config = #iree_cpu.lowering_config<distribution = [0], vector_reduction = [16]>}
```

The `vector_reduction = [16]` tells us the vectorizer wants 16-wide reductions (AVX-512 f32).
The bound 124992 is a multiple of 16. So we can reshape `tensor<?xf32>` → `tensor<?x16xf32>`
and turn the inner dim into a parallel dimension that vectorizes cleanly.

### What happens today

- **Static shapes** (e.g., `tensor<500000xf32>`): `LLVMCPUSplitReduction` succeeds →
  inner reduction gets split into parallel + reduction → vectorizer produces `vaddps zmm`
  accumulator loop → good codegen.
- **Dynamic shapes** (`tensor<?xf32>`): `linalg::splitReduction` returns failure →
  pass skips → vectorizer produces scalar horizontal reduction (`vaddss`) per loop iteration →
  **~16x worse throughput**.

### Assembly comparison

**Good (static, split fires):** 1 `vaddps zmm` per iteration, horizontal reduce after loop.
**Bad (dynamic, split fails):** 16 sequential `vaddss` per iteration inside the loop.

## Proposed fix

Implement a new version of the split-reduction logic in `LLVMCPUSplitReduction.cpp` that
handles bounded dynamic dimensions. Do NOT modify upstream `linalg::splitReduction` — mirror
the needed logic locally.

### Approach

1. **Detect bounded dynamic dims**: Use `ValueBoundsConstraintSet::computeConstantBound`
   (from `mlir/Interfaces/ValueBoundsOpInterface.h`) to get the upper bound of the dynamic
   reduction dimension.

2. **Check alignment**: Verify the bound is a multiple of the vector reduction size
   (from `lowering_config.vector_reduction`, typically 16 for AVX-512 f32).

3. **Reshape**: Insert `tensor.expand_shape` to reshape `tensor<?xf32>` → `tensor<?x16xf32>`.
   `expand_shape` supports one dynamic dim. The outer `?` becomes `ceildiv(original_dim, 16)`,
   the inner `16` is static.

4. **Rewrite the linalg.generic**: After the reshape, the reduction has two dims:
   - Outer `?` dim → becomes the reduction loop (tiled by the vectorizer)
   - Inner `16` dim → becomes a parallel dim → vectorizes to `zmm` accumulator

   The rewritten generic looks like:
   ```mlir
   // Phase 1: partial reduction (vectorized, 16-wide accumulator)
   %expanded = tensor.expand_shape %input [[0, 1]]
       : tensor<?xf32> into tensor<?x16xf32>
   %partial = linalg.generic {
                        affine_map<(d0, d1) -> (d1)>],
       iterator_types = ["reduction", "parallel"]
   } ins(%expanded : tensor<?x16xf32>) outs(%init16 : tensor<16xf32>)

   // Phase 2: final horizontal reduction (trivial, 16 → scalar)
   %result = linalg.generic {
       indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> ()>],
       iterator_types = ["reduction"]
   } ins(%partial : tensor<16xf32>) outs(%init_scalar : tensor<f32>)
   ```

5. **Handle remainder**: If the original dim isn't exactly a multiple of 16 at runtime,
   the last outer tile will be partial. The masking in the vectorizer handles this.

### Key files

- **`compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp`** — The pass to modify.
  Current `splitReductionPrecondition` and `splitReductionImpl` need a new path for
  bounded-dynamic inputs.
- **`compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp`** — Pipeline wiring (line ~198).
  No changes needed here.
- **`compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp`** — Sets `lowering_config`.
  Read the `vector_reduction` tile size from here. No changes needed.

### ValueBoundsConstraintSet usage

```cpp
#include "mlir/Interfaces/ValueBoundsOpInterface.h"

// Get the upper bound of a dynamic dim
FailureOr<int64_t> ub = ValueBoundsConstraintSet::computeConstantBound(
    presburger::BoundType::UB,
    /*value=*/dynamicDimValue,
    /*stopCondition=*/nullptr);
```

### How to verify

It should introduce a new scf.for op around the reduction op after we run the
pass with `enable-fp-reduction-reordering` flag. E.g., `iree-opt --pass-pipeline="builtin.module(func.func(iree-llvmcpu-split-reduction{enable-fp-reduction-reordering=true},cse,canonicalize))"`

```mlir
func.func @sum_dispatch_0_reduction_D_f32() attributes {gpu.known_block_size = array<i32: 1, 1, 1>, translation_info = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>} {
  %c-1 = arith.constant -1 : index
  %c124992 = arith.constant 124992 : index
  %0 = hal.interface.binding.subspan layout(<bindings = [#hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, Indirect>], flags = Indirect>) binding(0) alignment(64) offset(%c0) flags("ReadOnly|Indirect") : !iree_tensor_ext.dispatch.tensor<readonly:tensor<1000000xf32>>
  %1 = hal.interface.binding.subspan layout(<bindings = [#hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, Indirect>], flags = Indirect>) binding(1) alignment(64) offset(%c0) flags(Indirect) : memref<9xf32, #hal.descriptor_type<storage_buffer>>
  %2 = tensor.empty() : tensor<9xf32>
  %3 = scf.forall (%arg0) = (0) to (1000000) step (124992) shared_outs(%arg1 = %2) -> (tensor<9xf32>) {
    %4 = affine.min affine_map<(d0) -> (-d0 + 1000000, 124992)>(%arg0)
    %5 = iree_tensor_ext.dispatch.tensor.load %0, offsets = [%arg0], sizes = [%4], strides = [1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<1000000xf32>> -> tensor<?xf32>
    %6 = arith.cmpi slt, %arg0, %c0 : index
    %7 = arith.subi %c-1, %arg0 : index
    %8 = arith.select %6, %7, %arg0 : index
    %9 = arith.divsi %8, %c124992 : index
    %10 = arith.subi %c-1, %9 : index
    %11 = arith.select %6, %10, %9 : index
    %extracted_slice = tensor.extract_slice %arg1[%11] [1] [1] : tensor<9xf32> to tensor<f32>
    %12 = linalg.fill ins(%cst : f32) outs(%extracted_slice : tensor<f32>) -> tensor<f32>
    %13 = linalg.generic {indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> ()>], iterator_types = ["reduction"]} ins(%5 : tensor<?xf32>) outs(%12 : tensor<f32>) attrs =  {iree_linalg_ext.split_reduction = [124992 : index], lowering_config = #iree_cpu.lowering_config<distribution = [0], vector_reduction = [16]>} {
    ^bb0(%in: f32, %out: f32):
      %14 = arith.addf %out, %in : f32
      linalg.yield %14 : f32
    } -> tensor<f32>
    scf.forall.in_parallel {
      tensor.parallel_insert_slice %13 into %arg1[%11] [1] [1] : tensor<f32> into tensor<9xf32>
    }
  } {mapping = [#iree_linalg_ext.split_reduction_mapping<0>]}
  iree_codegen.store_to_buffer %3, %1 : tensor<9xf32> into memref<9xf32, #hal.descriptor_type<storage_buffer>>
  return
}
```

hanhanW:
The revision reworks the upstream `splitReduction` within IREE and supports the bounded-dynamic cases.

The structure of splitReduction function mirrors the inner-parallel path of upstream `mlir::linalg::splitReduction` as closely as possible to keep the port easy to maintain. The only intentional deltas are:
  1. Uses `OpFoldResult` + `tensor::getMixedSize(s)` so non-reduction dims and the outer reduction dim can be dynamic.
  2. Computes the outer reduction size with `affine.apply floorDiv splitSize` instead of dividing a static shape at compile time.
  3. Passes `output_shape` to `tensor.expand_shape` so the outer dim can be a dynamic value.
  4. Single-input / single-output only (no outer loop over input operands).
  5. Inner-parallel only (no outer-parallel branch, no `useAlloc`).

Fixes https://github.com/iree-org/iree/issues/24010

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
@@ -108,6 +108,11 @@ LogicalResult splitReductionPrecondition(Operation *op,
 
   SmallVector<unsigned> dims;
   linalgOp.getReductionDims(dims);
+  if (dims.size() != 1) {
+    LDBG() << "expected exactly 1 reduction dimension for targets";
+    return failure();
+  }
+
   AffineMap map =
       linalgOp.getMatchingIndexingMap(linalgOp.getDpsInputOperand(0));
   unsigned lastIdx = map.getNumResults() - 1;
@@ -137,13 +142,13 @@ LogicalResult splitReductionPrecondition(Operation *op,
 /// corresponds to the single reduction iterator dimension of `op`. Fails if
 /// `op` does not have exactly one reduction dimension or if that dimension is
 /// not referenced by the input indexing map.
+/// The method assumes that the preconditions verified by
+/// `splitReductionPrecondition`.
 static FailureOr<unsigned> getReductionOperandDimPosition(linalg::LinalgOp op) {
   SmallVector<unsigned> reductionDims;
   op.getReductionDims(reductionDims);
-  if (reductionDims.size() != 1) {
-    return failure();
-  }
-
+  assert(reductionDims.size() == 1 &&
+         "precondition guarantees exactly one reduction dimension");
   AffineMap inputMap = op.getMatchingIndexingMap(op.getDpsInputOperand(0));
   for (auto [idx, dimExpr] : llvm::enumerate(inputMap.getResults())) {
     if (cast<AffineDimExpr>(dimExpr).getPosition() == reductionDims.front()) {
@@ -154,7 +159,8 @@ static FailureOr<unsigned> getReductionOperandDimPosition(linalg::LinalgOp op) {
 }
 
 /// Returns the size of the input operand along its reduction dimension. Prefers
-/// the static tensor shape when available.
+/// the static tensor shape when available. This method does not create any IR,
+/// i.e., it is safe to be used in precondition checks.
 static FailureOr<OpFoldResult> getReductionSize(linalg::LinalgOp op) {
   FailureOr<unsigned> reductionOperandDimPos =
       getReductionOperandDimPosition(op);
@@ -196,7 +202,7 @@ static FailureOr<Value> createIdentityValue(OpBuilder &builder,
 }
 
 /// Returns true if the reduction dimension size of `op` is provably a multiple
-/// of `splitSize`. For dynamic cases, returns false when  `solver` is null or
+/// of `splitSize`. For dynamic cases, returns false when `solver` is null or
 /// has no divisibility information.
 static bool canSplitReductionWithSize(linalg::LinalgOp op, int64_t splitSize,
                                       DataFlowSolver *solver) {
@@ -420,10 +426,8 @@ LogicalResult splitReductionImpl(Operation *op, int64_t size,
   // splitReduction already replaces the op.
   auto tiledOp = cast<linalg::LinalgOp>(tileResFirst->tiledOps.back());
   FailureOr<OpFoldResult> reductionSize = getReductionSize(tiledOp);
-  if (failed(reductionSize)) {
-    LDBG() << "failed to determine tiled reduction size";
-    return success();
-  }
+  assert(succeeded(reductionSize) &&
+         "expected to be able to determine reduction size after tiling");
   linalg::SplitReductionResult splitRes =
       splitReduction(tiledOp, size, lastIdx, *reductionSize, rewriter);
 
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
@@ -7,13 +7,24 @@
 #include "iree/compiler/Codegen/Dialect/CPU/IR/IREECPUTypes.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
 #include "iree/compiler/Codegen/LLVMCPU/Passes.h"
+#include "iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.h"
+#include "iree/compiler/Dialect/Util/IR/UtilTypes.h"
 #include "llvm/Support/DebugLog.h"
+#include "mlir/Analysis/DataFlow/ConstantPropagationAnalysis.h"
+#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
+#include "mlir/Analysis/DataFlowFramework.h"
+#include "mlir/Analysis/SliceAnalysis.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/Affine/Utils.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/SCF/Transforms/TileUsingInterface.h"
 #include "mlir/Dialect/SCF/Transforms/Transforms.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/Pass/Pass.h"
 
 #define DEBUG_TYPE "iree-llvmcpu-split-reduction"
@@ -25,6 +36,19 @@ namespace mlir::iree_compiler {
 
 namespace {
 
+/// Returns the single combiner operation of the reduction body for the
+/// `initIndex`-th output of `op`. Returns nullptr if the body does not reduce
+/// with a single operation.
+static Operation *matchSingleCombinerOp(linalg::LinalgOp op,
+                                        unsigned initIndex) {
+  SmallVector<Operation *, 4> combinerOps;
+  if (!matchReduction(op.getRegionOutputArgs(), initIndex, combinerOps) ||
+      combinerOps.size() != 1) {
+    return nullptr;
+  }
+  return combinerOps.front();
+}
+
 /// Make sure that
 /// - the pass has not been applied before
 /// - has tensor semantics
@@ -35,6 +59,7 @@ namespace {
 /// - has exactly 1 input
 /// - if enableReductionReordering is not set, then operand is an int
 /// - innermost dimension of the input operand is reduction
+/// - the reduction body is a single combiner op with a known neutral element
 /// TODO: support named ops, numInputs > 1, and modify lastDim check below
 /// accordingly. If fpReductionReordering is not enabled by default, it must
 /// be an integer or index type to proceed to allow associative reordering.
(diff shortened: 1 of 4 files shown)


hanhanW:
@bangtianliu could I use your review on this change, since you worked on the splitReduction before? Thanks in advance, and take your time! It is not hurry.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
```diff
@@ -92,26 +117,291 @@ LogicalResult splitReductionPrecondition(Operation *op,
...
+static FailureOr<unsigned> getReductionOperandDimPosition(linalg::LinalgOp op) {
+  SmallVector<unsigned> reductionDims;
+  op.getReductionDims(reductionDims);
+  if (reductionDims.size() != 1) {
+    return failure();
+  }
+
+  AffineMap inputMap = op.getMatchingIndexingMap(op.getDpsInputOperand(0));
+  for (auto [idx, dimExpr] : llvm::enumerate(inputMap.getResults())) {
+    if (cast<AffineDimExpr>(dimExpr).getPosition() == reductionDims.front()) {
```
bangtianliu, line 149: I am not sure that using cast here is safe.

bangtianliu: Either document the implict the precondition or guard it

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
```diff
@@ -92,26 +117,291 @@ LogicalResult splitReductionPrecondition(Operation *op,
...
+  }
+  Type elementType =
+      getElementTypeOrSelf(op.getDpsInits()[initIndex].getType());
+  return arith::ConstantOp::create(builder, op.getLoc(), elementType,
+                                   *identityAttr)
+      .getResult();
+}
+
+/// Returns true if the reduction dimension size of `op` is provably a multiple
+/// of `splitSize`. For dynamic cases, returns false when  `solver` is null or
```
bangtianliu:
```suggestion
/// of `splitSize`. For dynamic cases, returns false when `solver` is null or
```

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
```diff
@@ -92,26 +117,291 @@ LogicalResult splitReductionPrecondition(Operation *op,
...
+                                                   RewriterBase &rewriter) {
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(linalgOp);
+
+  SmallVector<unsigned> dims;
+  linalgOp.getReductionDims(dims);
+  unsigned reductionDim = dims[0];
+  unsigned insertSplitDimension = reductionDim + 1;
+
+  Operation *reductionOp = matchSingleCombinerOp(linalgOp, /*initIndex=*/0);
```
bangtianliu, line 256: `matchSingleCombinerOp` is called with `initIndex=0` in `splitReductionPrecondition` and then again inside `splitReduction` + `createIdentityValue`. The precondition already guarantees success, so the two re-checks in `splitReduction` are redundant.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
```diff
@@ -92,26 +117,291 @@ LogicalResult splitReductionPrecondition(Operation *op,
...
+    return failure();
+  }
+  Value input = op.getDpsInputOperand(0)->get();
+  auto inputType = cast<RankedTensorType>(input.getType());
+  int64_t staticDim = inputType.getDimSize(*reductionOperandDimPos);
+  if (ShapedType::isStatic(staticDim)) {
+    return OpFoldResult(
+        IntegerAttr::get(IndexType::get(op.getContext()), staticDim));
+  }
+  if (auto viewOp = input.getDefiningOp<OffsetSizeAndStrideOpInterface>()) {
```
bangtianliu, line 172: Whether the dynamic input may be produced by an op that is not `OffsetSizeAndStrideOpInterface`.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
```diff
@@ -128,30 +418,32 @@ LogicalResult splitReductionImpl(Operation *op, int64_t size,
...
   // splitReduction already replaces the op.
-  FailureOr<linalg::SplitReductionResult> splitRes = splitReduction(
-      rewriter, cast<linalg::LinalgOp>(tileResFirst->tiledOps.back()), fn);
-  if (failed(splitRes)) {
-    LDBG() << "failed on step 2 (SplitReduction)";
+  auto tiledOp = cast<linalg::LinalgOp>(tileResFirst->tiledOps.back());
+  FailureOr<OpFoldResult> reductionSize = getReductionSize(tiledOp);
+  if (failed(reductionSize)) {
+    LDBG() << "failed to determine tiled reduction size";
     return success();
```
bangtianliu: not relevant to your change, but let's return failure() here

compiler/src/iree/compiler/Codegen/LLVMCPU/test/split_reduction.mlir
```diff
@@ -144,3 +144,131 @@ func.func @dont_reassociate(%arg0 : tensor<4096xi32>, %arg1 : tensor<f32>) -> te
...
+// CHECK-SAME:      iterator_types = ["reduction", "parallel"]
+// CHECK:             %[[MAX0:.+]] = arith.maximumf
+// CHECK:         %[[FINAL:.+]] = linalg.generic
+// CHECK-SAME:      iterator_types = ["reduction"]
+// CHECK:             %[[MAX1:.+]] = arith.maximumf
+
+// -----
+
+#config = #iree_cpu.lowering_config<vector_reduction = [16]>
+func.func @split_bounded_dynamic_with_linalg_index(%arg0 : tensor<256xf32>, %arg1 : tensor<i32>) -> tensor<i32> {
```
bangtianliu, line 248: All the nest tests use 1D, could we also add one 2D+ case for coverage?

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
hanhanW: Actually this should always succeed, because it is checked in precondition. Let's convert it to a cheap assertion.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
hanhanW: They are two sanity checks and the checks are very cheap, so I ended up with assertion here.

Note: they are used in this function, so we always need to declare it.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
hanhanW: We only support projected permutation cases. If it can't be casted to `AffineDimExpr`, something is wrong. Let me improve the document. Good catch!

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
hanhanW: You are right on the statement, and we can create a `DimOp` to query it. However, this is used for precondition checks as well, so I don't want to create IR in this phase. Let me update the comment to reflect this.

Side note, ideally we should implement the IREE::Util::ShapeAwareOp  interface for these operations, and we can have a wider coverage. I don't add a TODO because we could run into rabbit holes. I don't want the TODO floating here forever.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
```diff
@@ -92,26 +122,292 @@ LogicalResult splitReductionPrecondition(Operation *op,
...
+  }
+
   return success();
 }
 
+/// Returns the result position in the input operand's indexing map that
+/// corresponds to the single reduction iterator dimension of `op`. Fails if
+/// `op` does not have exactly one reduction dimension or if that dimension is
+/// not referenced by the input indexing map.
+/// The method assumes that the preconditions verified by
```
bangtianliu:
```suggestion
/// The method assumes that the preconditions are verified by
```

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
```diff
@@ -83,6 +108,11 @@ LogicalResult splitReductionPrecondition(Operation *op,
 
   SmallVector<unsigned> dims;
   linalgOp.getReductionDims(dims);
+  if (dims.size() != 1) {
```
bangtianliu: redundant? already covered by getNumReductionLoops() != 1 above

bangtianliu (approved):
LGTM % nits

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSplitReduction.cpp
hanhanW: My bad, I skimmed through code and did not find it. 🤦‍♂️

hanhanW:
> LGTM % nits

Thanks for the quick review, really appreciate!

PR #24069, opened 2026-04-10
[Codegen] Use createWriteOrMaskedWrite utility

sommerlukas:
https://github.com/llvm/llvm-project/pull/190967 made the `createWriteOrMaskedWrite` utility available for downstream use. 

This is a follow up to https://github.com/iree-org/iree/pull/23935 and replaces custom logic for masked write in the vectorization of `inner_tiled` introduced in that PR with a use of the upstream utility.

The change is mostly NFC with one small exception: The upstream utility sets all `in_bounds` attributes to `true` if it uses masking, therefore one test needed updating. 

Assisted-by: Claude Code

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_inferred.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_inferred.mlir
@@ -913,7 +913,7 @@ func.func @masked_tensor_multi_mma(%lhs: tensor<?x?x4xf16>, %rhs: tensor<?x?x4xf
 //       CHECK:   %[[MMA:.+]] = iree_codegen.inner_tiled ins(%[[LHS]], %[[RHS]]) outs(%[[ACC]])
 //  CHECK-SAME:     : vector<2x3x4xf16>, vector<3x5x4xf16> into vector<2x5x4xf32>
 //       CHECK:   %[[WRITE_MASK:.+]] = vector.create_mask {{.*}} : vector<2x5x4xi1>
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_inferred.mlir
```diff
@@ -913,7 +913,7 @@ func.func @masked_tensor_multi_mma(%lhs: tensor<?x?x4xf16>, %rhs: tensor<?x?x4xf
 //       CHECK:   %[[MMA:.+]] = iree_codegen.inner_tiled ins(%[[LHS]], %[[RHS]]) outs(%[[ACC]])
 //  CHECK-SAME:     : vector<2x3x4xf16>, vector<3x5x4xf16> into vector<2x5x4xf32>
 //       CHECK:   %[[WRITE_MASK:.+]] = vector.create_mask {{.*}} : vector<2x5x4xi1>
-//       CHECK:   vector.transfer_write %[[MMA]], %arg2{{.*}}, %[[WRITE_MASK]] {in_bounds = [false, false, true]}
+//       CHECK:   vector.transfer_write %[[MMA]], %arg2{{.*}}, %[[WRITE_MASK]] {in_bounds = [true, true, true]}
```
hanhanW, line 916: I was going to point this out that all the bounds should be true within a mask, but somehow I forgot. (I checked the upstream tests when I reviewed it.)

hanhanW (approved):
Nice! Thanks for following-up on this!

PR #24072, opened 2026-04-10
[Codegen][DMA] Defer no_reduce_shared_memory_bank_conflicts setting until DMA convertibility is verified

Yu-Zhewen:
Previously, `no_reduce_shared_memory_bank_conflicts` (disable padding) was set during lowering strategy selection based on `useDirectLoad`. This caused a performance gap: when DMA was requested but later demoted silently to stream copy `GPUConvertToCoalescedDMA`. Note swizzle is not always feasible (e.g., transposed LHS), so disabling padding too early left
some kernels with no bank conflict mitigation at all.

This PR defers the padding decision until DMA convertibility is verified. Additionally, `sourceIsFromFatRawBuffer` is now part of convertibility check.

Assisted-by: Cursor (Claude)


Yu-Zhewen:
https://github.com/iree-org/iree/pull/24208 is a better solution, by moving DMA convertibility check earlier instead

PR #24073, opened 2026-04-10
Fix ExpandDestinationForallOp rejecting non-store users

jtuyls:
The pattern only needs to validate store users for expandability. Non-store users (e.g. tensor.dim) are safe because the hoisted collapse_shape preserves the original result type.

Previously, any non-store user caused the pattern to bail out. With dynamic shapes on gfx1100, tensor.dim ops on the forall result blocked the pattern, preventing the WMMAR3 accumulator reshape from folding into the output buffer and causing a shared memory overflow (80KB > 65KB).


nirvedhmeshram (approved):
LGTM

PR #24077, opened 2026-04-10
Integrate llvm-project to 15e46e25f26c.

Muzammiluddin-Syed-ECE:
No reverts.

Clean bump.

Note (keep note alive):
- We are carrying a patch in iree-org/torch-mlir that needs to be 1) removed once they have an LLVM bump and 2) cherry-picked any time a torch-mlir bump is performed.
- https://github.com/iree-org/torch-mlir/commit/172a8b4d5caddbb3af9045f7329dd4121b590cbf

Made-with: Cursor


Yu-Zhewen (approved):
Thanks!

PR #24080, opened 2026-04-11
[LLVMGPU] Add `includeLLVMLowering` option to pipeline and migrate tests 1/n

hanhanW:
The revision introduces the flag and migrate some pipeline tests to module-scope tests. This includes distribution verification and removes the nesting HAL ops.

There is a failure in ROCDL/tile_and_fuse pipeline tests, and the change is significant big, so the revision only migrate part of pipeline tests.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
@@ -1138,7 +1138,8 @@ void buildLLVMGPUCodegenConfigurationPassPipeline(
 }
 
 void buildLLVMGPUCodegenPassPipeline(OpPassManager &modulePassManager,
-                                     bool useROCM, bool preserveDebugInfo) {
+                                     bool useROCM, bool preserveDebugInfo,
+                                     bool includeLLVMLowering) {
   modulePassManager.addPass(createLowerExecutableUsingTransformDialectPass());
   LLVMGPULowerExecutableTargetPassOptions options;
   options.forROCDL = useROCM;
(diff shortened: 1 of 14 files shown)


hanhanW (commented):
It's much easier to review if you enable "hide whitespaces" feature.

Image

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -1131,7 +1138,8 @@ void buildLLVMGPUCodegenConfigurationPassPipeline(
 }
 
 void buildLLVMGPUCodegenPassPipeline(OpPassManager &modulePassManager,
-                                     bool useROCM, bool preserveDebugInfo) {
+                                     bool useROCM, bool preserveDebugInfo,
```
MaheshRavishankar, line 1141: Maybe make this simpler and just pass the GPU options struct here directly?

MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
hanhanW: That is a good idea and what we've done for CPU side.

https://github.com/iree-org/iree/blob/cffd3099cf5cd4ab134a60ac3637826163b7623f/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L655-L658

I'll revisit it in a follow-up because the state is a little messy to me. E.g., `debugSymbols` belongs to `GPUCodegenOption` but now it is in `ROCMTargetOption`. It is passed in both configuration phase and translation phase.

PR #24085, opened 2026-04-13
[Codegen] Fix bufferization dropping negative strides on dispatch tensor load/store

ziliangzl:
The `equalTensorShape` check in the bufferization interface for `DispatchTensorLoadOp` and `DispatchTensorStoreOp` only compared result shapes and dynamic dims to determine if a "whole tensor" fast path could be taken. This missed the case where strides are non-unit (e.g. `strides = [-1]` for tensor reversal via `tensor.extract_slice`), causing the fast path to replace the op with the raw binding buffer and silently drop the stride info.

For example, the following IR:
```
  %0 = iree_tensor_ext.dispatch.tensor.load %src,
       offsets = [3], sizes = [4], strides = [-1]
       : !dispatch.tensor<readonly:tensor<4xf32>> -> tensor<4xf32>
```
was incorrectly bufferized to a plain `memref<4xf32>` identity copy instead of the correct `memref.subview` with `strided<[-1], offset: 3>`.

This resulted in `tensor.extract_slice` with negative strides (which should reverse/flip a tensor) being silently turned into a no-op by IREE, while upstream MLIR's `mlir-cpu-runner` produced correct results.

The fix adds a unit-stride check (`areAllConstantIntValue(strides, 1)`) before taking the fast "whole tensor" path. When strides are non-unit (e.g. negative), we fall through to the existing `memref.subview` code path which correctly forwards the mixed offsets, sizes, and strides.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/BufferizeDispatchTensorLoadStore.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/BufferizeDispatchTensorLoadStore.cpp
@@ -13,6 +13,7 @@
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/Debug.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/Interfaces/SubsetOpInterface.h"
 
 namespace mlir::iree_compiler {
@@ -32,7 +33,8 @@ static Value getBufferizedSubspanOrSubview(
   SmallVector<Value> dynamicSizes;
   std::tie(std::ignore, dynamicSizes) = decomposeMixedValues(sizes);
   if (equalTensorShape(tensorType, dynamicSizes, dispatchTensorType,
-                       subspanDynamicSizes)) {
+                       subspanDynamicSizes) &&
+      areAllConstantIntValue(strides, 1)) {
     return subspanMemref;
   }
   MemRefType subviewMemRefType = memref::SubViewOp::inferRankReducedResultType(
--- a/compiler/src/iree/compiler/Codegen/Common/test/bufferize_dispatch_tensor_load_store.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/bufferize_dispatch_tensor_load_store.mlir
@@ -190,3 +190,35 @@ func.func @nested_dispatch_tensor_load_in_forall() {
 // CHECK:           iree_codegen.load_from_buffer %[[SUBVIEW]]
 // CHECK-SAME:          : memref<3072x512xbf16, strided<[512, 1], offset: ?>, #hal.descriptor_type<storage_buffer>> -> tensor<3072x512xbf16>
 // CHECK:         iree_codegen.store_to_buffer %{{.+}}, %[[OUTPUT]]
+
+// -----
+
+// Verify that dispatch.tensor.load with negative strides is correctly
+// bufferized to a memref.subview with negative strides, not the raw buffer.
+
+#pipeline_layout = #hal.pipeline.layout<bindings = [
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>
+]>
+func.func @negative_stride_load_store() {
+  %0 = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<4xf32>>
+  %1 = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) : !iree_tensor_ext.dispatch.tensor<writeonly:tensor<4xf32>>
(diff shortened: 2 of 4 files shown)


ziliangzl:
## **Reproducing the Bug**

### **Test case**

`tensor.extract_slice` with `step = -1` should reverse/flip a tensor. When the input and output shapes are identical (e.g., both `tensor<4xf32>`), the negative stride is the only semantic difference from an identity slice.

**`flip_test_iree.mlir`**:

```bash
func.func @flip_tensor(%input: tensor<4xf32>) -> tensor<4xf32> {

%result = tensor.extract_slice %input[3] [4] [-1] : tensor<4xf32> to tensor<4xf32>

return %result : tensor<4xf32>

}
```

**`flip_test_mlir_runner.mlir`**:

```bash
func.func @main() {
%input = arith.constant dense<[1.0, 2.0, 3.0, 4.0]> : tensor<4xf32>
%result = tensor.extract_slice %input[3] [4] [-1] : tensor<4xf32> to tensor<4xf32>
%c3 = arith.constant 3 : index

%v0 = tensor.extract %result[%c0] : tensor<4xf32>
%v1 = tensor.extract %result[%c1] : tensor<4xf32>
%v2 = tensor.extract %result[%c2] : tensor<4xf32>
%v3 = tensor.extract %result[%c3] : tensor<4xf32>

vector.print %v0 : f32
vector.print %v1 : f32
vector.print %v2 : f32
vector.print %v3 : f32

return

}
```

### **Upstream MLIR (correct result)**

```bash
MLIR_OPT=/path/to/mlir-opt

MLIR_RUNNER=/path/to/mlir-cpu-runner

RUNNER_UTILS=/path/to/lib/libmlir_runner_utils.so

C_RUNNER_UTILS=/path/to/lib/libmlir_c_runner_utils.so

$MLIR_OPT flip_test_mlir_runner.mlir \

--one-shot-bufferize="bufferize-function-boundaries" \

--expand-strided-metadata \

--lower-affine \

--convert-scf-to-cf \

--convert-vector-to-llvm \

--finalize-memref-to-llvm \

--convert-func-to-llvm \

--convert-arith-to-llvm \

--reconcile-unrealized-casts | \

$MLIR_RUNNER -e main -entry-point-result=void \

--shared-libs=$RUNNER_UTILS,$C_RUNNER_UTILS
```

**Output (correct)**:

```bash
4

3

2

1
```

### **IREE (wrong result, before fix)**

```bash
IREE_COMPILE=/path/to/iree-compile

IREE_RUN=/path/to/iree-run-module

$IREE_COMPILE --iree-hal-target-backends=llvm-cpu \

flip_test_iree.mlir -o /tmp/flip_test.vmfb

$IREE_RUN --module=/tmp/flip_test.vmfb \

--function=flip_tensor \

--input="4xf32=1 2 3 4"
```

**Output (wrong -- tensor is not flipped)**:

```bash
EXEC @flip_tensor

result[0]: hal.buffer_view

4xf32=1 2 3 4
```

**Expected output**:

```bash
EXEC @flip_tensor

result[0]: hal.buffer_view

4xf32=4 3 2 1
```

### **Root cause**

During `IREEComprehensiveBufferizePass`, `dispatch.tensor.load` with `offsets=[3], sizes=[4], strides=[-1]` is checked by `equalTensorShape()` which only compares shapes and dynamic dims, ignoring offsets/strides. Since both input and output are `tensor<4xf32>`, it matches as "whole tensor" and the op is replaced with the raw binding buffer -- the negative stride is silently dropped, turning the flip into a no-op identity copy.

The same issue exists in `BufferizeDispatchTensorLoadStore.cpp`.

hanhanW (changes requested):
I'm okay with the patch as it actually fixes non-unit strides, thought the intention (e.g., negative strides) is not a good idea IMO.

Using `tensor.extract_slice` op to represent a reverse op is not ideal (and it is honestly bad) in IREE. The op does not have TilingInterface implementation and what you're doing is just a "copy" using scalars.

I'll likely approve the PR if you make the tests using positive strides like `2` for example; you'd want your program using a linalg.generic (or gather) like https://github.com/iree-org/iree/commit/76cad82974b652c4d82ef6d3e38ec394ee616e55

ziliangzl:
@hanhanW Thanks for suggestion. 
However, I just noticed that for any load with an offset that is not `0` or a stride that is not a `unit`, the shape will definitely be inconsistent. 
Therefore, we only need to check the shape, and do not need to check the stride. 
`offset = n - 1`, `stride = -1` is the only case that will cause a bug. 
Either we fix `stride = -1`, or we should throw an error when the stride is negative.

hanhanW:
> @hanhanW Thanks for suggestion. However, I just noticed that for any load with an offset that is not `0` or a stride that is not a `unit`, the shape will definitely be inconsistent. Therefore, we only need to check the shape, and do not need to check the stride. `offset = n - 1`, `stride = -1` is the only case that will cause a bug. Either we fix `stride = -1`, or we should throw an error when the stride is negative.

We should throw an error, and we can introduce a `VerifyCodegenInputPass` to `Codegen/Common`.

PR #24086, opened 2026-04-13
[GPU] Allow swizzle promotion for transpose linalg.generic producers (preshuffling prs 1/3)

Muzammiluddin-Syed-ECE:
This PR is the first in a series of PRs meant to enable preshuffling for scaled gemms.

Currently, operands to be swizzled are skipped when producing SwizzleHintOps if their producer is a linalg::LinalgOp. This means that transpose linalg.generic ops feeding into a matmul with SwizzleOperand promotion would never get XOR swizzle hints applied to their output buffer.

This change adds an exception: transpose linalg.generic producers (identified via linalg::isaTransposeOpInterface) no longer exit early. For a GEMM dispatch where the transpose is folded into memory access, this does not produce any extra allocations, bufferization correctly recognizes that the `tensor.empty` associated to the TransposeOp and the SwizzleHintOp are redundant and removes the copy.

Non-transpose linalg.generic producers continue to get the early return with lowering_config annotation as before.

Made-with: Cursor

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_promote_matmul_operands.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_promote_matmul_operands.mlir
@@ -394,6 +394,104 @@ func.func @swizzle_operand_no_promote_fill(%b: tensor<128x128xf32>) -> tensor<4x
 
 // -----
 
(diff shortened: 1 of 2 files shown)


amd-eochoalo (approved):
I think this makes sense.

keshavvinayak01 (approved):

PR #24088, opened 2026-04-13
[Codegen][MXFP4] Add PartialDataTiledScaledMMAAttr to enable scales preshuffling (preshuffling prs 2/3)

Muzammiluddin-Syed-ECE:
This PR is the second in a series of PRs meant to enable preshuffling for scaled gemms.

We introduce a new kind called `PartialDataTiledScaledMMAAttr` which is meant to be identical to `DataTiledScaledMMAAttr` in every way except three:
1) We do not use the default heuristic in `chooseDataTiledMMAAttr` to choose tiling config, instead use `getMmaScheduleFromProblemAndTarget`.
2) Perform shuffling on scale operands only; Do not hoist data operands to a separate dispatch and importantly, do not apply any permutations to these operands.
3) Add ability to handle row-major (non-permuted) data operands in `populateOperandOffsetsSizesStrides`.

These changes allow us to only create separate dispatches to preshuffle scale operands. This provides an immediate 31.5% geomean improvement in throughput across 50 shapes tested.

### DIFF
--- a/compiler/plugins/target/ROCM/Dialect/ROCM/IR/ROCMAttrs.cpp
+++ b/compiler/plugins/target/ROCM/Dialect/ROCM/IR/ROCMAttrs.cpp
@@ -211,13 +211,18 @@ Attribute TensorUKernelProviderAttr::getDataLayoutForUKernel(
         continue;
       }
     } else if (opType == IREE::Encoding::EncodingOpType::scaled_matmul) {
-      auto dtScaledMma = dyn_cast<GPU::DataTiledScaledMMAAttr>(mma);
-      if (!dtScaledMma) {
+      auto scaledIntrinsic =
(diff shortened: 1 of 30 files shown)


Muzammiluddin-Syed-ECE:
See baseline perf numbers here: [Link removed]()

See perf numbers after this change here: [Link removed]()

See graph [here](https://github.com/nod-ai/fp4-benchmark/blob/main/artifacts/20260413-PartialDTResults.png)
See raw data [here](https://github.com/nod-ai/fp4-benchmark/blob/main/artifacts/20260413-PartialDTResults.csv)

Muzammiluddin-Syed-ECE:
This PR can be further split into two PRs, one introducing the new kind and another with the data tiling changes.

Muzammiluddin-Syed-ECE:
Switching to draft while I investigate the source of performance degradation with swizzling but not hoisting data operands

PR #24091, opened 2026-04-13
 [Codegen] Relax identity layout check in ResolveSwizzleHints verification

Yu-Zhewen:
The swizzle operates on the local index, and the memref base offset doesn't affect it since the final address is `base_offset +
swizzle(local_index)`. Removing the isIdentity() check allows multi-buffered memrefs (e.g. `memref<Nxty, strided<[1], off: ?>>`) to pass through. The change is also more consistent with the existing comment ("For rank 1, allow dynamic memrefs or static contiguous row-major memrefs").

Discussion: https://github.com/iree-org/iree/pull/24001#discussion_r3051951926

Assisted-by: Cursor (Claude)


jerryyin (approved):
LGTM

PR #22249, opened 2025-10-09
[Codegen] Update the td spec using the contraction matcher op

bangtianliu:
This PR removes `attention_and_matmul_spec_flux_mi300.mlir` (it is not used at all) and updates `attention_and_matmul_spec_punet_mi300.mlir`, `iree_default_tuning_spec_gfx942.mlir` and `attention_and_matmul_spec_unet_fp16_mi308.mlir` using contraction matcher op.

PR #24093, opened 2026-04-13
[Codegen] Add VectorDistribute constraint generation scaffolding for matmul and conv

RattataKing:
Adds the foundational helpers and knobs template for VectorDistribute constraint generation on matmul and convolution ops.

The complete constraints are left for the next PR to keep this PR small.


compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,15 +36,194 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  return IntegerAttr::get(IntegerType::get(ctx, 64), value);
+}
+
+/// Helper to create a IntKnobAttr.
+static Attribute makeIntKnob(MLIRContext *ctx, StringRef name) {
+  return IREE::Codegen::IntKnobAttr::get(ctx, StringAttr::get(ctx, name));
+}
+
+/// Get compatible MMA attrs for the given target and element types.
+/// Uses the same filtering as KernelConfig.cpp: subgroup size and
```
qedawkins, line 52: nit: Don't reference KernelConfig.cpp here by name. Just describe what it does standalone.

If this is copying the compatibility logic, instead expose the logic elsewhere and reuse it rather than reimplementing it.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -67,24 +247,68 @@ static LogicalResult emitConstraints(OpBuilder &builder, Operation *rootOp,
...
+  }
+
+  MLIRContext *ctx = rootOp->getContext();
   OpBuilder builder(ctx);
-  DictionaryAttr knobs = buildKnobsDict(ctx, numLoops);
-  ConstraintsOpShell shell =
-      createConstraintsOpShell(builder, rootOp, getRootOpInfo(rootOp),
-                               pipelineAttr, knobs, numLoops, indexingMaps);
-  return emitConstraints(builder, rootOp, shell.smtDimArgs);
+  auto compatibleMMAs = getCompatibleMMAAttrsForLinalgOp(gpuTarget, linalgOp);
```
qedawkins, line 273: nit: spell out type

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -67,24 +247,68 @@ static LogicalResult emitConstraints(OpBuilder &builder, Operation *rootOp,
...
+  DictionaryAttr knobs =
+      buildVectorDistributeKnobsDict(ctx, *loopInfo, *dims, compatibleMMAs);
+  Attribute pipelineAttr = IREE::GPU::PipelineAttr::get(
+      ctx, IREE::GPU::LoweringPipeline::VectorDistribute);
+
+  auto shell =
```
qedawkins, line 279: nit: spell out type

qedawkins (changes requested):
I don't see any testing of the new constraint generation. I would expect to see some verification of `buildVectorDistributeKnobsDict` producing all of the knobs we expect for a couple different example problem shapes.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.h
```diff
@@ -7,11 +7,48 @@
...
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/IR/AffineMap.h"
 #include "mlir/IR/Attributes.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/Operation.h"
 
 namespace mlir::iree_compiler {
 
+/// Contraction-like dimension classification used by both matmul and conv.
+struct ContractionLikeDims {
```
bangtianliu, line 20: batch dimension?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -97,27 +307,13 @@ LogicalResult emitLLVMGPUConstraints(Attribute attr,
...
-  }
-  if (!mainRoot) {
+  // KernelConfig.cpp currently only labels one root op per set.
+  Operation *tunableOp = rootOps.front();
+  if (!tunableOp) {
     return success();
   }
-
-  return emitConstraintsForOp(mainRoot, pipelineAttr);
+  auto opAttr = tunableOp->getAttrOfType<IREE::Codegen::RootOpAttr>("root_op");
```
bangtianliu, line 302: `getAttrOfType` returns a null attribute if the operation doesn't carry `"root_op"`.

```suggestion
  auto opAttr = getRootOpInfo(tunableOp);
  if (!opAttr) {
    return success();
  }
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -67,24 +233,68 @@ static LogicalResult emitConstraints(OpBuilder &builder, Operation *rootOp,
...
-static LogicalResult emitConstraintsForOp(Operation *rootOp,
-                                          Attribute pipelineAttr) {
+/// Emit VectorDistribute constraints for contraction-like dims (matmul/conv).
+/// TODO(#23535): Complete real constraint logics here.
+static LogicalResult
+emitVectorDistributeConstraints(OpBuilder &builder, linalg::LinalgOp linalgOp,
+                                const ContractionLikeDims &dims,
+                                IREE::GPU::TargetAttr gpuTarget,
+                                ArrayRef<Value> smtDimArgs) {
+  Location loc = linalgOp.getLoc();
```
bangtianliu, line 223: `emitVectorDistributeConstraints` only constrains M dims, but knobs exist for M+N (Logic Inconsistency)

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,15 +36,180 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
-static DictionaryAttr buildKnobsDict(MLIRContext *ctx, unsigned numLoops) {
-  SmallVector<Attribute> workgroupEntries;
-  for (unsigned d = 0; d < numLoops; ++d) {
-    workgroupEntries.push_back(IntKnobAttr::get(ctx, ("wg_" + Twine(d)).str()));
+/// Helper to create an i64 IntegerAttr with a fixed value.
+static Attribute makeIntAttr(MLIRContext *ctx, int64_t value = 0) {
+  return IntegerAttr::get(IntegerType::get(ctx, 64), value);
+}
+
+/// Helper to create a IntKnobAttr.
```
bangtianliu, line 46:
```suggestion
/// Helper to create an IntKnobAttr.
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,15 +36,180 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+/// Get contraction-like (m,n,k) dims for a linalg op.
+/// Only supports contraction and convolution today.
+FailureOr<ContractionLikeDims>
+inferContractionLikeDims(linalg::LinalgOp linalgOp) {
+  if (linalg::isaContractionOpInterface(linalgOp)) {
+    auto contractionDims = linalg::inferContractionDims(linalgOp);
+    if (failed(contractionDims)) {
+      return failure();
+    }
+    return ContractionLikeDims{
```
bangtianliu, line 123: Use `llvm::to_vector` for cleaner range-to-vector conversion

```suggestion
    return ContractionLikeDims{
  llvm::to_vector(contractionDims->m),
  llvm::to_vector(contractionDims->n),
  llvm::to_vector(contractionDims->k)};
```

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,412 @@
...
+  }
+
+  ContractionLikeDims matmulDims() {
+    return ContractionLikeDims{/*m=*/{0}, /*n=*/{1}, /*k=*/{2}};
+  }
+
+  // conv_2d_nhwc_hwcf with input=1x18x18x4, filter=3x3x4x8, output=1x16x16x8.
+  // 7 loops: (n, oh, ow, oc, kh, kw, ic).
+  RootOpLoopInfo loopInfoForConv() {
+
```
bangtianliu, line 309: Stray blank line at start of `loopInfoForConv()`

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.h
RattataKing: I think it's not needed anywhere from constraint generator, at least for the VectorDistribute pipeline, I think TileAndFuse doesnt need it either for constraints generation.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
RattataKing: `emitLLVMGPUConstraints()` is invoked from: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/InsertSMTConstraints.cpp#L72-L84, where the passed `rootOps` are guarantee to be `root_op`

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
RattataKing: These constraints are not complete yet, this PR is to add some scaffolding and util functions.

RattataKing:
The existing `Codegen/Common/test/insert_smt_constraints.mlir` is the lit test for constraint generation.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.h
bangtianliu: I see what you mean. 

We need the batch info for the tuner to build the spec. See: https://github.com/iree-org/iree/pull/22249
I am not saying it should be included in the constraint generation phase. But this is one kind of knobs (constant?) we need to handle.

RattataKing: `ContractionLikeDims` here wont be passed to tuner. For info needed when building a TD file, e.g.`element types, indexing maps, ...`, tuner can get those by calling mlir py binding of `inferContractionDims`

bangtianliu: After moving the core part to IREE, the tuner should be only about calling z3-solver and then generating TD specs (This is my understanding, maybe wrong).  There is no need to parse the MLIR file again from the tuner side. 

But nvm, this is some Qs we can think about later.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -227,6 +223,12 @@ FailureOr<std::pair<GPUMMASchedule, GPUMMASchedule>> deduceAttentionSchedule(
 llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
                               const GPUMMASchedule &schedule);
 
+// checks if the given intrinsic can be used for the given problem.
```
bangtianliu:
```suggestion
/// Checks if the given intrinsic can be used for the given problem.
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,56 +37,237 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+    }
+    // Check if the mma intrinsic supports the problem.
+
+    if (failed(canTargetIntrinsic(problem, getIntrinsic(mma),
+                                  targetSubgroupSize, /*canUpcastAcc*/ true,
+                                  /*mustBeAligned*/ false))) {
+      continue;
+    }
+
+    Attribute mmaAttr = Attribute(mma);
```
bangtianliu: cast is redundant since IREE::GPU::MMAAttr already inherits from Attribute.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -97,27 +280,13 @@ LogicalResult emitLLVMGPUConstraints(Attribute attr,
...
-      mainRoot = op;
-      break;
-    }
-    if (!mainRoot) {
-      mainRoot = op;
-    }
-  }
-  if (!mainRoot) {
+  // Currently only labels one root op per set.
+  Operation *tunableOp = rootOps.front();
```
bangtianliu, line 285: Crash if rootOps is empty. 
```C++
if (rootOps.empty()) {                                                                                       
    return success();                                                                                          
}                                                                                                            
Operation *tunableOp = rootOps.front();
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
bangtianliu: I see, either way works.

``` C++
IREE::Codegen::RootOpAttr getRootOpInfo(Operation *op) {
  return op->getAttrOfType<IREE::Codegen::RootOpAttr>(kRootOpInfoAttrName);
} 
```

But this one is defined specifically for this purpose.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,56 +37,237 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  subgroupBasisEntries.emplace_back("mapping",
+                                    ArrayAttr::get(ctx, subgroupMapping));
+  knobsEntries.emplace_back("subgroup_basis",
+                            DictionaryAttr::get(ctx, subgroupBasisEntries));
+
+  // Add workgroup size and subgroup size at the top level.
+  SmallVector<Attribute> wgSizeKnobs;
+  wgSizeKnobs.push_back(makeIntKnob(ctx, "wg_x"));
+  wgSizeKnobs.push_back(makeIntKnob(ctx, "wg_y"));
+  wgSizeKnobs.push_back(makeIntKnob(ctx, "wg_z"));
```
bangtianliu:
```suggestion
  SmallVector<Attribute> wgSizeKnobs = {                                                                       
      makeIntKnob(ctx, "wg_x"),                                                                                
      makeIntKnob(ctx, "wg_y"),                                                                                
      makeIntKnob(ctx, "wg_z"),
  };   
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,56 +37,237 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
   }
+  if (linalg::isaConvolutionOpInterface(linalgOp)) {
+    FailureOr<mlir::linalg::ConvolutionDimensions> convolutionDims =
+        mlir::linalg::inferConvolutionDims(linalgOp);
+    if (failed(convolutionDims) || convolutionDims->outputImage.empty() ||
+        convolutionDims->outputChannel.empty() ||
+        convolutionDims->inputChannel.empty()) {
+      return failure();
+    }
+    // Maps outputImage→M, outputChannel→N, inputChannel→K.
```
bangtianliu: Avoid this kind of code comment.

> Aim to describe what the code is trying to do and why, not how it does it at a micro level

https://llvm.org/docs/CodingStandards.html#commenting

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,56 +37,237 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+                               const ContractionLikeDims &dims,
+                               ArrayRef<Attribute> compatibleMMAs) {
   SmallVector<NamedAttribute> knobsEntries;
+
+  // Build workgroup array: M and N dims get IntKnobAttr, others get 0 : i64.
+  SmallVector<Attribute> workgroupEntries(loopInfo.numLoops,
+                                          makeIntAttr(ctx, 0));
+  for (ArrayRef<unsigned> dimSet : {ArrayRef(dims.m), ArrayRef(dims.n)}) {
+    for (unsigned d : dimSet) {
+      workgroupEntries[d] = makeIntKnob(ctx, "wg_" + std::to_string(d));
```
bangtianliu: Some places use "wg_" + std::to_string(d)  while others use ("dim_" +     Twine(d) + ...) 

Consider: Consistently use Twine

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,56 +37,237 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  // Only innermost M and N dims get subgroup tiling, others stay 1.
+  SmallVector<Attribute> subgroupCounts(loopInfo.numLoops, makeIntAttr(ctx, 1));
+  // A successful call to linalg::inferContractionDims guarantees that dims.m
+  // and dims.n are non-empty.
+  subgroupCounts[dims.m.back()] = makeIntKnob(ctx, "sg_m_cnt");
+  subgroupCounts[dims.n.back()] = makeIntKnob(ctx, "sg_n_cnt");
+  subgroupBasisEntries.emplace_back("counts",
+                                    ArrayAttr::get(ctx, subgroupCounts));
+  SmallVector<Attribute> subgroupMapping;
+  for (size_t i = 0; i < loopInfo.numLoops; ++i) {
```
bangtianliu: Use `unsigned i ' to match the type of numLoops

bangtianliu (changes requested):

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
RattataKing: thanks!

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
RattataKing: thanks!

RattataKing:
Circular dependency issues triggered in this PR is solved here: https://github.com/iree-org/iree/pull/24203

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,465 @@
...
+  IREE::GPU::TargetAttr getGfx942Target() {
+    return IREE::GPU::getHIPTargetDetails("gfx942", "", &ctx);
+  }
+
+  IREE::GPU::TargetAttr getGfx1201Target() {
+    return IREE::GPU::getHIPTargetDetails("gfx1201", "", &ctx);
+  }
+
+  linalg::LinalgOp parseLinalgOp(StringRef mlirText) {
```
bangtianliu: All linalg ops are constructed by parsing MLIR text strings via `parseSourceString<ModuleOp>`. See examples here: https://github.com/iree-org/llvm-project/blob/5ea5b9eb8eec9c5ba7f053cb74327a3efd6b762d/mlir/unittests/Dialect/Linalg/InferConvolutionDimsTest.cpp.

bangtianliu: Then you can remove the MLIRParser dependency from both BUILD.bazel and CMakeLists.txt if it exists.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
bangtianliu: Still, opAttr could be null. Even though the caller currently guarantees `root_op` exists, the function should be defensive.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -35,60 +37,242 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
-  for (unsigned d = 0; d < numLoops; ++d) {
-    workgroupEntries.push_back(IntKnobAttr::get(ctx, ("wg_" + Twine(d)).str()));
+/// Helper to create an i64 IntegerAttr with a fixed value.
+static Attribute makeIntAttr(MLIRContext *ctx, int64_t value = 0) {
+  return IntegerAttr::get(IntegerType::get(ctx, 64), value);
+}
+
+/// Helper to create an IntKnobAttr.
+static Attribute makeIntKnob(MLIRContext *ctx, StringRef name) {
+  return IREE::Codegen::IntKnobAttr::get(ctx, StringAttr::get(ctx, name));
```
bangtianliu:
```suggestion
  return IntKnobAttr::get(ctx, StringAttr::get(ctx, name));
```
The alias is defined above, so you can use it

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -97,27 +281,13 @@ LogicalResult emitLLVMGPUConstraints(Attribute attr,
...
-  }
-  if (!mainRoot) {
+  // Currently only labels one root op per set.
+  Operation *tunableOp = rootOps.front();
+  if (!tunableOp) {
     return success();
   }
-
-  return emitConstraintsForOp(mainRoot, pipelineAttr);
+  auto opAttr = tunableOp->getAttrOfType<IREE::Codegen::RootOpAttr>("root_op");
```
bangtianliu: Also, another issue: spell out types instead of `auto`

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,465 @@
...
+       BuildVectorDistributeKnobs_Matmul_DictStr) {
+  MLIRContext *ctx = getContext();
+  DictionaryAttr matmulKnobDict = buildVectorDistributeKnobsDict(
+      ctx, loopInfoForMatmul(), matmulDims(), compatibleMMAs());
+
+  std::string result;
+  llvm::raw_string_ostream os(result);
+  matmulKnobDict.print(os);
+
+  StringRef expected =
```
bangtianliu: Test structurally,  verify keys exist, check array sizes, and verify specific knob names.

Here, compare the full printed string of a `DictionaryAttr`. If the printing format of any constituent attribute changes upstream, these tests break without semantic change.

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,465 @@
...
+      "workgroup = [#iree_codegen.smt.int_knob<\"wg_0\">, "
+      "#iree_codegen.smt.int_knob<\"wg_1\">, 0], "
+      "workgroup_size = [#iree_codegen.smt.int_knob<\"wg_x\">, "
+      "#iree_codegen.smt.int_knob<\"wg_y\">, "
+      "#iree_codegen.smt.int_knob<\"wg_z\">]"
+      "}";
+  EXPECT_EQ(result, expected);
+}
+
+TEST_F(BuildVectorDistributeKnobsDictTest,
```
bangtianliu: Also here

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
   Value rem = smt::IntModOp::create(builder, loc, lhs, rhs);
   Value eq = smt::EqOp::create(builder, loc, rem, zero);
   std::string fmtMsg = (msg + " ({} % {} == 0)").str();
-  IREE::Codegen::AssertOp::create(builder, loc, eq, fmtMsg,
-                                  ValueRange{lhs, rhs});
+  AssertOp::create(builder, loc, eq, fmtMsg, ValueRange{lhs, rhs});
 }
 
-using IntKnobAttr = IREE::Codegen::IntKnobAttr;
+std::string makeVarName(StringRef prefix, unsigned idx) {
```
bangtianliu: Should it be static?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.h
```diff
@@ -7,11 +7,76 @@
...
+constexpr StringLiteral kKnobWgSizeYName = "wg_size_y";
+constexpr StringLiteral kKnobWgSizeZName = "wg_size_z";
+
+// Knob variable name prefixes. The loop dim count varies
+// per problem, so names are built at runtime as prefix + dim idx.
+constexpr StringLiteral kKnobWgPrefix = "wg_";
+constexpr StringLiteral kKnobRedPrefix = "red_";
+
+/// Build a knob variable name from a prefix, e.g. ("wg_", 2) -> "wg_2".
+std::string makeVarName(StringRef prefix, unsigned idx);
```
bangtianliu: If this is intentionally public for reuse, document why in the header. If it's  only public for testing, consider testing `buildVectorDistributeKnobsDict` output instead (which the tests already do) and making `makeVarName `static.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+
+  for (IREE::GPU::MMAAttr mma : gpuTarget.getWgp().getMma()) {
+    if (mma.getSubgroupSize() != targetSubgroupSize) {
+      continue;
+    }
+    if (!mma.getDistributionMappingKind()) {
+      continue;
+    }
+    // Check if the mma intrinsic supports the problem.
+
```
bangtianliu: empty line?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  }
+  if (linalg::isaConvolutionOpInterface(linalgOp)) {
+    FailureOr<mlir::linalg::ConvolutionDimensions> convolutionDims =
+        mlir::linalg::inferConvolutionDims(linalgOp);
+    if (failed(convolutionDims) || convolutionDims->outputImage.empty() ||
+        convolutionDims->outputChannel.empty() ||
+        convolutionDims->inputChannel.empty()) {
+      return failure();
+    }
+    return ContractionLikeDims{llvm::to_vector(convolutionDims->outputImage),
```
bangtianliu: Add one comment, something like: "TODO: Not sure this is concrete enough for convs; revisit when plumbing
through conv constraint generation."

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,523 @@
...
+                          unsigned int oc = 1) {
+
+    OpBuilder builder = setupBuilder();
+    Location loc = builder.getUnknownLoc();
+    auto f32 = builder.getF32Type();
+
+    unsigned int oh = ih - kh + 1;
+    unsigned int ow = iw - kw + 1;
+
+    auto input_ty = RankedTensorType::get({b, ih, iw, ic}, f32);
```
bangtianliu: Solve these warnings

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,523 @@
...
+  MLIRContext *getContext() { return &ctx; }
+
+  IREE::GPU::TargetAttr getGfx942Target() {
+    return IREE::GPU::getHIPTargetDetails("gfx942", "", &ctx);
+  }
+
+  IREE::GPU::TargetAttr getGfx1201Target() {
+    return IREE::GPU::getHIPTargetDetails("gfx1201", "", &ctx);
+  }
+
```
bangtianliu: Unused test helper methods getGfx942Target() and getGfx1201Target()

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  // Only innermost M and N dims get subgroup tiling, others stay 1.
+  SmallVector<Attribute> subgroupCounts(loopInfo.numLoops, makeIntAttr(ctx, 1));
+  // A successful call to linalg::inferContractionDims guarantees that dims.m
+  // and dims.n are non-empty.
+  subgroupCounts[dims.m.back()] = makeIntKnobAttr(ctx, kKnobSgMCntName);
+  subgroupCounts[dims.n.back()] = makeIntKnobAttr(ctx, kKnobSgNCntName);
+  subgroupBasisEntries.emplace_back(kKnobCountsKey,
+                                    ArrayAttr::get(ctx, subgroupCounts));
+  SmallVector<Attribute> subgroupMapping;
+  for (unsigned i = 0; i < loopInfo.numLoops; ++i) {
```
bangtianliu, line 260: Could use llvm::seq

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,523 @@
...
+  ASSERT_TRUE(!!op);
+
+  auto loopInfo = getRootOpLoopInfo(op);
+  auto dims = inferContractionLikeDims(op);
+  auto result = getCompatibleMMAAttrs(op, target, *loopInfo, *dims);
+  EXPECT_TRUE(result.empty());
+}
+
+TEST_F(VectorDistributeKnobsTest, KnobDictForMatmul) {
+
```
bangtianliu: Also here, remove empty line

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,523 @@
...
+  IREE::GPU::TargetAttr getGfx942Target() {
+    return IREE::GPU::getHIPTargetDetails("gfx942", "", &ctx);
+  }
+
+  IREE::GPU::TargetAttr getGfx1201Target() {
+    return IREE::GPU::getHIPTargetDetails("gfx1201", "", &ctx);
+  }
+
+  linalg::MatmulOp getTestMatmulOp(Type inputType, Type accType,
```
bangtianliu: Two getTestMatmulOp functions, some redundancy here.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  SmallVector<NamedAttribute> subgroupBasisEntries;
+  // Only innermost M and N dims get subgroup tiling, others stay 1.
+  SmallVector<Attribute> subgroupCounts(loopInfo.numLoops, makeIntAttr(ctx, 1));
+  // A successful call to linalg::inferContractionDims guarantees that dims.m
+  // and dims.n are non-empty.
+  subgroupCounts[dims.m.back()] = makeIntKnobAttr(ctx, kKnobSgMCntName);
+  subgroupCounts[dims.n.back()] = makeIntKnobAttr(ctx, kKnobSgNCntName);
+  subgroupBasisEntries.emplace_back(kKnobCountsKey,
+                                    ArrayAttr::get(ctx, subgroupCounts));
+  SmallVector<Attribute> subgroupMapping;
```
bangtianliu, line 258: The subgroup mapping is always [0, 1, 2, ...] (identity), so just one placeholder here? But it is not always true. Maybe add one code comment here about it.

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
RattataKing: thanks for catching it

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
RattataKing: yes it's true for VD conv and matmul and not true for VectorDistribute atten. Thanks for catching this, actually this should be just dropped instead of emitted to the knob dict.

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,493 @@
...
+class RootOpUtilsTest : public LinalgTestBase {};
+
+TEST_F(RootOpUtilsTest, InferContractionDimsForFillOp) {
+  auto op = getTestFillOp();
+  ASSERT_TRUE(!!op);
+  EXPECT_TRUE(failed(inferContractionLikeDims(op)));
+}
+
+TEST_F(RootOpUtilsTest, InferContractionDimsForMatmulOp) {
+  auto op = getTestMatmulOp();
```
bangtianliu: Several auto usages where type is not obvious

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,493 @@
...
+  ASSERT_TRUE(!!op);
+
+  auto dims = inferContractionLikeDims(op);
```
bangtianliu: also here

bangtianliu: auto should only be used when the type is obvious from the RHS.

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
bangtianliu: Not solved?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
bangtianliu: not solved

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,493 @@
...
+  auto target = IREE::GPU::getHIPTargetDetails("gfx942", "", getContext());
+  ASSERT_TRUE(target);
+
+  Type i1 = IntegerType::get(getContext(), 1);
+  auto op = getTestMatmulOp(i1, i1);
+  ASSERT_TRUE(!!op);
+
+  auto loopInfo = getRootOpLoopInfo(op);
+  auto dims = inferContractionLikeDims(op);
+  auto result = getCompatibleMMAAttrs(op, target, *loopInfo, *dims);
```
bangtianliu: IncompatibleTypes test dereferences without assertion guards.

*loopInfo and *dims are dereferenced without first asserting that loopInfo.has_value()  
  and succeeded(dims).

do sth like this 

```C++
    auto loopInfo = getRootOpLoopInfo(op);                                                         
    ASSERT_TRUE(loopInfo.has_value());    
    auto dims = inferContractionLikeDims(op);                                                      
    ASSERT_TRUE(succeeded(dims));                                                                  
    auto result = getCompatibleMMAAttrs(op, target, *loopInfo, *dims);                             
    EXPECT_TRUE(result.empty());  
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+                                         shell.smtDimArgs);
 }
 
 LogicalResult emitLLVMGPUConstraints(Attribute attr,
                                      ArrayRef<Operation *> rootOps) {
+  if (rootOps.empty()) {
+    return success();
+  }
+  // Currently only labels one root op per set.
+  Operation *tunableOp = rootOps.front();
```
bangtianliu: The function takes ArrayRef<Operation *> rootOps, but only processes the first element. Maybe change the interface

bangtianliu (approved):
LGTM % nit

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
kuhar: This regresses the multi-root set case that the deleted test used to cover. InsertSMTConstraints preserves walk order within a set, so a fill marked with the same root set is passed before the matmul; with rootOps.front() the emitter sees the fill, returns success at the contraction gate, and emits no constraints for the matmul. I reproduced this by adding the root_op attr back to the fill in insert_smt_constraints.mlir: FileCheck then fails because no iree_codegen.smt.constraints is emitted. Please keep selecting the tunable contraction from the set, or add an explicit test showing that multi-root sets cannot reach this emitter anymore.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  SmallVector<Attribute> reductionEntries(loopInfo.numLoops,
+                                          makeIntAttr(ctx, 0));
+  for (unsigned d : dims.k) {
+    reductionEntries[d] = makeIntKnobAttr(ctx, makeVarName(kKnobRedPrefix, d));
+  }
+  knobsEntries.emplace_back(kKnobReductionKey,
+                            ArrayAttr::get(ctx, reductionEntries));
+
+  // Add mma_kind knob.
+  knobsEntries.emplace_back(
```
kuhar, line 246: compatibleMMAs can be empty when the target has no matching intrinsic for this contraction-like op. Emitting mma_kind with an empty options array makes the constraint system unsatisfiable for otherwise valid ops, so the pipeline cannot materialize a config. Please bail out before building the knobs dict or omit mma_kind when there is no candidate.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+}
+
+/// Helper to create an i64 IntegerAttr with a fixed value.
+static IntegerAttr makeIntAttr(MLIRContext *ctx, int64_t value = 0) {
+  return IntegerAttr::get(IntegerType::get(ctx, 64), value);
+}
+
+/// Helper to create an IntKnobAttr.
+static IntKnobAttr makeIntKnobAttr(MLIRContext *ctx, StringRef name) {
+  return IntKnobAttr::get(ctx, StringAttr::get(ctx, name));
```
kuhar: Don't we have a constructor (`::get` overload)that takes StringRef directly? If not, we should add one and drop this helper.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+}
+
+/// Get unique compatible MMA attrs for matmul and conv ops.
+SmallVector<Attribute> getCompatibleMMAAttrs(linalg::LinalgOp op,
```
kuhar: Could we make this and the other functions below static as well?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +37,259 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  for (unsigned d : dims.k) {
+    reductionEntries[d] = makeIntKnobAttr(ctx, makeVarName(kKnobRedPrefix, d));
+  }
+  knobsEntries.emplace_back(kKnobReductionKey,
+                            ArrayAttr::get(ctx, reductionEntries));
+
+  // Add mma_kind knob.
+  knobsEntries.emplace_back(
+      kKnobMmaKindKey,
+      OneOfKnobAttr::get(ctx, StringAttr::get(ctx, kKnobMmaIdxName),
```
kuhar: also here: would be nice to have a `::get` overload that takes plain StringRef

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.h
```diff
@@ -7,11 +7,73 @@
...
+constexpr StringLiteral kKnobSgNCntName = "sg_n_cnt";
+constexpr StringLiteral kKnobSgSizeName = "sg_size";
+constexpr StringLiteral kKnobWgSizeXName = "wg_size_x";
+constexpr StringLiteral kKnobWgSizeYName = "wg_size_y";
+constexpr StringLiteral kKnobWgSizeZName = "wg_size_z";
+
+// Knob variable name prefixes. The loop dim count varies
+// per problem, so names are built at runtime as prefix + dim idx.
+constexpr StringLiteral kKnobWgPrefix = "wg_";
+constexpr StringLiteral kKnobRedPrefix = "red_";
```
kuhar: Could we move this to a .cpp file so that we don't expose these in the public API?

compiler/src/iree/compiler/Codegen/LLVMGPU/unittests/LLVMGPUConstraintGeneratorTest.cpp
```diff
@@ -0,0 +1,498 @@
+// Copyright 2026 The IREE Authors
```
kuhar: We generally don't use unit tests (with gtest) for IR / compiler logic, and prefer LIT tests instead. We don't need to have all the helpers tested directly, which will in turn allow us to make them `static`.

kuhar:
@RattataKing this needs LIT tests for the generated constraints

kuhar (changes requested):
test comment

kuhar (commented):
Test coverage gap: the only negative path tested is "no compatible MMAs" via empty mma = []. Please add at least:

1. A negative test where subgroup_size_choices does not match any MMA's getSubgroupSize() (covers the mma.getSubgroupSize() != targetSubgroupSize skip).
2. A negative test where the element types do not match any MMA's a/b types (covers the aType/bType mismatch in canTargetIntrinsic).
3. A test with a batched contraction (non-empty batch dim) so that the batch handling in the M/N/K assignment is exercised; today only rank-2 / rank-4 generics and conv are tested.
4. A test where multiple MMAs are compatible to verify the de-duplication / ordering of the OneOfKnobAttr options.

These are the kinds of edge cases the constraint generator needs to be robust on before tuning relies on it.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +80,278 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+buildVectorDistributeKnobsDict(MLIRContext *ctx, const RootOpLoopInfo &loopInfo,
+                               const ContractionLikeDims &dims,
+                               ArrayRef<Attribute> compatibleMMAs) {
   SmallVector<NamedAttribute> knobsEntries;
-  knobsEntries.emplace_back("workgroup", ArrayAttr::get(ctx, workgroupEntries));
+
+  // Build workgroup array: M and N dims get IntKnobAttr, others get 0 : i64.
+  SmallVector<Attribute> workgroupEntries(loopInfo.numLoops,
+                                          makeIntAttr(ctx, 0));
+  for (ArrayRef<unsigned> dimSet : {ArrayRef(dims.m), ArrayRef(dims.n)}) {
```
kuhar: Current auto VectorDistribute config only materializes non-unit workgroup tiles for the innermost M/N/K dims: KernelConfig sets the dropped M/N/K dims to 1 before filling mDim/nDim/kDim. The constraint template is wider and makes every M/N dim a knob (`expanded_matmul` exposes `wg_0..wg_3`).

If that wider search space is intentional, please add a test or comment showing that non-1 outer M/N tiles are valid when supplied by a tuning spec. Otherwise keep the template aligned with the generated configs and encode those outer dims as fixed 1.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +80,278 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+                             bounds[dims.k.back()], lhsElemType,
+                             rhsElemType,           initElemType};
+
+  auto getIntrinsic = [](IREE::GPU::MMAAttr mma) -> GPUIntrinsicType {
+    auto [mSize, nSize, kSize] = mma.getMNKShape();
+    auto [aType, bType, cType] = mma.getABCElementTypes();
+    return GPUIntrinsicType{mSize, nSize, kSize, aType, bType, cType, mma};
+  };
+
+  for (IREE::GPU::MMAAttr mma : gpuTarget.getWgp().getMma()) {
```
kuhar, line 134: This loop admits block MMA intrinsics, but the VectorDistribute matmul and convolution config paths explicitly skip block intrinsics. Because getIntrinsic rebuilds every MMAAttr through the non-batch GPUIntrinsicType constructor, canTargetIntrinsic will not reject those block layouts, so mma_kind can contain choices the pipeline never selects. Please mirror the config filter here and skip block intrinsics. We should also have a test for this.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +80,278 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  SmallVector<NamedAttribute> subgroupBasisEntries;
+  // Only innermost M and N dims get subgroup tiling, others stay 1.
+  SmallVector<Attribute> subgroupCounts(loopInfo.numLoops, makeIntAttr(ctx, 1));
+  // A successful call to linalg::inferContractionDims guarantees that dims.m
+  // and dims.n are non-empty.
+  subgroupCounts[dims.m.back()] = IntKnobAttr::get(ctx, kKnobSgMCntName);
+  subgroupCounts[dims.n.back()] = IntKnobAttr::get(ctx, kKnobSgNCntName);
+  subgroupBasisEntries.emplace_back(kKnobCountsKey,
+                                    ArrayAttr::get(ctx, subgroupCounts));
+  knobsEntries.emplace_back(kKnobSubgroupBasisKey,
```
kuhar: This makes the generated constraints non-verifying for real VectorDistribute configs. The knobs template records `subgroup_basis` as a nested dict `{counts = [...]}`, but existing GPU lowering configs store `subgroup_basis` as `[[counts], [mapping]]` via `GPULoweringConfigUtils::setBasis`. `VerifyPipelineConstraints::extractKnobValues` requires the same container shape and returns failure on this mismatch, so the verifier skips the whole constraints region before checking `workgroup` or `reduction`. I reproduced that with a matmul whose lowering config had `workgroup = [7, 256, 0]`; the verify pass still exited 0. Since the current PR does not consume subgroup-basis knobs and the existing review thread already notes this should be dropped for now, please remove `subgroup_basis` from the knobs template until it can match the real lowering-config shape.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/config_llvm_gpu_constraints.mlir
```diff
@@ -0,0 +1,183 @@
...
+// CHECK-LABEL: func.func @expanded_matmul
+// CHECK:       linalg.generic
+// CHECK:       iree_codegen.smt.constraints target = <set = 0>, pipeline = #iree_gpu.pipeline<VectorDistribute>,
+// CHECK-NEXT:  knobs = {
+// CHECK-DAG:   mma_kind = #iree_codegen.smt.one_of_knob<"mma_idx", [#iree_gpu.mma_layout<MFMA_F32_16x16x4_F32>]>
+// CHECK-DAG:   reduction = [0, 0, 0, 0, #iree_codegen.smt.int_knob<"red_4">]
+// CHECK-DAG:   subgroup_basis = {counts = [1, 1, #iree_codegen.smt.int_knob<"sg_m_cnt">, #iree_codegen.smt.int_knob<"sg_n_cnt">, 1]}
+// CHECK-DAG:   subgroup_size = #iree_codegen.smt.int_knob<"sg_size">
+// CHECK-DAG:   workgroup = [#iree_codegen.smt.int_knob<"wg_0">, #iree_codegen.smt.int_knob<"wg_1">, #iree_codegen.smt.int_knob<"wg_2">, #iree_codegen.smt.int_knob<"wg_3">, 0]
+// CHECK-DAG:   workgroup_size = [#iree_codegen.smt.int_knob<"wg_size_x">, #iree_codegen.smt.int_knob<"wg_size_y">, #iree_codegen.smt.int_knob<"wg_size_z">]
```
kuhar, line 162: The new LIT test verifies the structure of the knobs dictionary but never checks any of the iree_codegen.smt.assert "... must be divisible by ..." lines that the constraint generator actually emits. The whole point of emitVectorDistributeConstraints is which loop dims get a divisibility check (M/N -> wg_*, K -> red_*, others -> nothing) and how the loop range argument is named in the diagnostic. None of that is locked in by these tests.

I ran the new file through iree-opt and confirmed the assertions are produced. Concretely, for @matmul_and_fill you get assertions on dim_0/wg_0, dim_1/wg_1, and dim_2/red_2. For @conv_2d_nhwc_hwcf only wg_1, wg_2, wg_3, red_6 get assertions (outer batch is 0 and skipped; spatial filter and inner channel dims are skipped). For @expanded_matmul you get wg_0..wg_3 plus red_4. None of this asymmetry between knob structure and emitted assertions is verified.

Please add CHECK lines for the assertions in at least the matmul, expanded_matmul, and conv cases. The "non-innermost-M/N divisibility" and "K-only reduction divisibility" behavior is exactly the part of this PR most likely to regress and is the most user-visible output of the pass.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -6,18 +6,68 @@
...
+// Knobs dict keys.
+constexpr StringLiteral kKnobWorkgroupKey = "workgroup";
+constexpr StringLiteral kKnobReductionKey = "reduction";
+constexpr StringLiteral kKnobMmaKindKey = "mma_kind";
+constexpr StringLiteral kKnobSubgroupBasisKey = "subgroup_basis";
+constexpr StringLiteral kKnobWorkgroupSizeKey = "workgroup_size";
+constexpr StringLiteral kKnobSubgroupSizeKey = "subgroup_size";
+// For subgroup basis subdict.
+constexpr StringLiteral kKnobCountsKey = "counts";
+constexpr StringLiteral kKnobMappingKey = "mapping";
```
kuhar: Dead constant: kKnobMappingKey is declared here ("mapping") but never referenced anywhere in this file or the rest of the compiler tree. The accompanying comment says "For subgroup basis subdict", but buildVectorDistributeKnobsDict only emits the kKnobCountsKey entry. Either drop kKnobMappingKey or add a TODO explaining when it will be used; right now it is just noise that will drift out of date.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +80,278 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+                               loopInfo->numLoops, loopInfo->indexingMaps);
+
+  return emitVectorDistributeConstraints(builder, linalgOp, *dims, gpuTarget,
+                                         shell.smtDimArgs);
+}
+
+/// Multiple root ops may be present in a set, e.g. <set = 0>:
+/// [linalg.fill, linalg.matmul]. This function will choose the matmul op
+/// over the fill op.
+static Operation *getTunableOp(Attribute attr, ArrayRef<Operation *> rootOps) {
```
kuhar: The Attribute attr parameter to getTunableOp is unused. Drop it from the signature; the caller can keep cast<IREE::GPU::PipelineAttr>(attr) where it is. (Removing the parameter also removes the implication that the selector is somehow pipeline-aware -- it currently is not.)

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +80,278 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
   }
+
+  SmallVector<Attribute> mmaAttrs;
+  const int64_t targetSubgroupSize = gpuTarget.getPreferredSubgroupSize();
+  SmallVector<int64_t> bounds = loopInfo.staticLoopRanges;
+  Type lhsElemType = getElementTypeOrSelf(op.getDpsInputOperand(0)->get());
+  Type rhsElemType = getElementTypeOrSelf(op.getDpsInputOperand(1)->get());
+  Type initElemType = getElementTypeOrSelf(op.getDpsInitOperand(0)->get());
+
+  GPUMatmulShapeType problem{bounds[dims.m.back()], bounds[dims.n.back()],
```
kuhar: Static-shape guard missing before passing problem dims to canTargetIntrinsic. We should bail out early on dynamic shapes since we can't quite handle them in the tuner as of today. If you need a testcase:

```
  func.func @dynamic_matmul(%lhs: tensor<?x?xf32>, %rhs: tensor<?x?xf32>, %empty: tensor<?x?xf32>)
      -> tensor<?x?xf32> attributes {hal.executable.target = #exec_target} {
    %r = linalg.matmul {root_op = #iree_codegen.root_op<set = 0>}
        ins(%lhs, %rhs : tensor<?x?xf32>, tensor<?x?xf32>)
        outs(%empty : tensor<?x?xf32>) -> tensor<?x?xf32>
    return %r : tensor<?x?xf32>
  }
```

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +80,278 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+getCompatibleMMAAttrs(linalg::LinalgOp op, IREE::GPU::TargetAttr gpuTarget,
+                      const RootOpLoopInfo &loopInfo,
+                      const ContractionLikeDims &dims) {
+  if (gpuTarget.getWgp().getMma().empty()) {
+    return {};
   }
+
+  SmallVector<Attribute> mmaAttrs;
+  const int64_t targetSubgroupSize = gpuTarget.getPreferredSubgroupSize();
+  SmallVector<int64_t> bounds = loopInfo.staticLoopRanges;
```
kuhar: nit: SmallVector<int64_t> bounds = loopInfo.staticLoopRanges; makes an unnecessary heap copy. Either use ArrayRef<int64_t> bounds = loopInfo.staticLoopRanges; or just index loopInfo.staticLoopRanges directly.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,65 +80,278 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  knobsEntries.emplace_back(
+      kKnobMmaKindKey,
+      OneOfKnobAttr::get(ctx, kKnobMmaIdxName, compatibleMMAs));
+
+  // Build subgroup basis with counts only, as mapping is always identity
+  // for VectorDistribute matmul and conv.
+  SmallVector<NamedAttribute> subgroupBasisEntries;
+  // Only innermost M and N dims get subgroup tiling, others stay 1.
+  SmallVector<Attribute> subgroupCounts(loopInfo.numLoops, makeIntAttr(ctx, 1));
+  // A successful call to linalg::inferContractionDims guarantees that dims.m
```
kuhar: The comment "A successful call to linalg::inferContractionDims guarantees that dims.m and dims.n are non-empty." is misleading because this code path is also reached for convolutions, where dims was populated from inferConvolutionDims (output image / output channel / input channel). The non-empty guarantee for the conv case is enforced separately by the explicit outputImage.empty() || outputChannel.empty() || inputChannel.empty() check in inferContractionLikeDims -- please update the comment so a future reader does not weaken the conv-side check thinking only the contraction guarantee matters. Suggestion: "Both inferContractionLikeDims branches guarantee dims.m and dims.n are non-empty (asserted/early-returned for conv)."

compiler/src/iree/compiler/Codegen/Common/test/insert_smt_constraints.mlir
```diff
@@ -31,8 +31,7 @@ hal.executable @matmul_f32_ex {
         %lhs = tensor.empty() : tensor<128x64xf32>
         %rhs = tensor.empty() : tensor<64x256xf32>
         %empty = tensor.empty() : tensor<128x256xf32>
-        %fill = linalg.fill {root_op = #iree_codegen.root_op<set = 0>}
-            ins(%cst : f32) outs(%empty : tensor<128x256xf32>) -> tensor<128x256xf32>
+        %fill = linalg.fill ins(%cst : f32) outs(%empty : tensor<128x256xf32>) -> tensor<128x256xf32>
```
kuhar: Inconsistent test fixtures: in insert_smt_constraints.mlir (this PR) the linalg.fill had its root_op = #iree_codegen.root_op<set = 0> annotation removed, while in the new config_llvm_gpu_constraints.mlir test matmul_and_fill retains the same root_op annotation on both linalg.fill and linalg.matmul. Pick one convention -- the realistic one (root_op on both, since DispatchCreation marks all roots in a set) is what getTunableOp is designed for, so I would keep the root_op on the fill in the existing test as well. Otherwise the existing test stops exercising the "fill is also a root, prefer matmul" selection logic that getTunableOp is the entire reason for.

compiler/src/iree/compiler/Codegen/LLVMGPU/CMakeLists.txt
```diff
@@ -102,6 +122,7 @@ iree_cc_library(
     "ROCDLPrefetching.cpp"
     "Verifiers.cpp"
   DEPS
+    ::LLVMGPUConstraintGenerator
```
kuhar, line 124: This new dependency makes the parent LLVMGPU library depend on the split LLVMGPUConstraintGenerator target, but the parent still lists LLVMGPUConstraintGenerator.cpp in its own SRCS. That keeps the source compiled in both targets and makes the split mostly cosmetic. Please remove LLVMGPUConstraintGenerator.cpp from the parent SRCS here and in the Bazel target.

compiler/src/iree/compiler/Codegen/Common/test/verify_smt_constraints_e2e.mlir
```diff
@@ -0,0 +1,72 @@
...
+    workgroup_size = [64, 1, 1] subgroup_size = 64>
+
+func.func @matmul_e2e_generated_violation(
+    %lhs: tensor<128x64xf32>, %rhs: tensor<64x256xf32>)
+    -> tensor<128x256xf32>
+    attributes {hal.executable.target = #exec_target,
+                translation_info = #translation} {
+  %cst = arith.constant 0.0 : f32
+  %init = tensor.empty() : tensor<128x256xf32>
+  %fill = linalg.fill ins(%cst : f32) outs(%init : tensor<128x256xf32>)
```
kuhar: Please keep this e2e failure case realistic by giving the fill the same `root_op` set. With both ops marked, the verifier currently finds the fill first, sees no lowering config, and skips verification instead of reaching the matmul. The verifier should pick the configured root op or keep walking.

compiler/src/iree/compiler/Codegen/Common/VerifyPipelineConstraints.cpp
```diff
@@ -23,70 +23,67 @@ namespace mlir::iree_compiler {
...
+        }
+        ArrayAttr options = knob.getOptions();
+        const auto *it = llvm::find(options, configAttr);
+        if (it == options.end()) {
+          return failure();
+        }
+        result[knob.getName()] = std::distance(options.begin(), it);
+        return success();
+      })
+      .Case([&](IntegerAttr) -> LogicalResult { return success(); })
```
kuhar: Fixed literals in the knob template should be checked against the config. Right now any `IntegerAttr` template entry succeeds, so a config with a non-1 outer tile can bypass the literal `1` entries emitted for unsupported outer M/N dims. Please require equality or fail extraction.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/config_llvm_gpu_constraints.mlir
```diff
@@ -0,0 +1,374 @@
...
+  max_load_instruction_bits = 128,
+  max_workgroup_sizes = [1024, 1024, 1024],
+  max_thread_count_per_workgroup = 1024,
+  max_workgroup_memory_bytes = 65536,
+  max_workgroup_counts = [2147483647, 2147483647, 2147483647]
+>>
+#exec_target_mismatch_mma_element_types = #hal.executable.target<"rocm", "rocm-hsaco-fb",
+    {iree_codegen.target_info = #gpu_target_mismatch_mma_element_types}>
+
+func.func @matmul_with_mismatch_mma_element_types()
```
kuhar: These two `@matmul_with_mismatch_mma_element_types` tests appear to exercise the same skip path. Please drop one, or rename/comment them so the distinct coverage is clear.

compiler/src/iree/compiler/Codegen/Common/test/verify_smt_constraints_e2e.mlir
```diff
@@ -0,0 +1,72 @@
...
+      -> tensor<128x256xf32>
+  %result = linalg.matmul
+      ins(%lhs, %rhs : tensor<128x64xf32>, tensor<64x256xf32>)
+      outs(%fill : tensor<128x256xf32>) -> tensor<128x256xf32>
+  return %result : tensor<128x256xf32>
+}
+
+// CHECK-LABEL: func.func @matmul_e2e_verify
```
kuhar: This e2e file still lacks a real success-path test. `@matmul_e2e_verify` has no `hal.executable.target`, so insertion returns early and no constraints are produced; the `CHECK-NOT` passes trivially. Please add a sibling with a HAL target and satisfying `lowering_config` so the test proves generated constraints are evaluated and erased.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,97 +82,311 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  SmallVector<Attribute> mmaAttrs;
+  const int64_t targetSubgroupSize = gpuTarget.getPreferredSubgroupSize();
+  Type lhsElemType = getElementTypeOrSelf(op.getDpsInputOperand(0)->get());
+  Type rhsElemType = getElementTypeOrSelf(op.getDpsInputOperand(1)->get());
+  Type initElemType = getElementTypeOrSelf(op.getDpsInitOperand(0)->get());
+  int64_t mSize = loopInfo.staticLoopRanges[dims.m.back()];
+  int64_t nSize = loopInfo.staticLoopRanges[dims.n.back()];
+  int64_t kSize = loopInfo.staticLoopRanges[dims.k.back()];
+
+  // Dynamic extents are represented as -1 and are not supported by tuner yet.
```
kuhar: This comment says dynamic extents are `-1` but that's an implementation detail subject to change; use `ShapedType::kDynamic` instead or `ShapedType::isDynamic(...)` as a predicate.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -30,97 +82,311 @@ static void assertDivisible(OpBuilder &builder, Location loc, Value lhs,
...
+  MLIRContext *ctx = rootOp->getContext();
   OpBuilder builder(ctx);
-  DictionaryAttr knobs = buildKnobsDict(ctx, numLoops);
-  ConstraintsOpShell shell =
-      createConstraintsOpShell(builder, rootOp, getRootOpInfo(rootOp),
-                               pipelineAttr, knobs, numLoops, indexingMaps);
-  return emitConstraints(builder, rootOp, shell.smtDimArgs);
+  DictionaryAttr knobs =
+      buildVectorDistributeKnobsDict(ctx, *loopInfo, *dims, compatibleMMAs);
+  auto pipelineIface = dyn_cast<IREE::Codegen::PipelineAttrInterface>(
```
kuhar: This `dyn_cast<PipelineAttrInterface>` plus null check is dead code: `IREE::GPU::PipelineAttr` declares the interface. Please use `cast<>` or pass the concrete `PipelineAttr` directly to `createConstraintsOpShell`.

kuhar (commented):
A few comments

kuhar:
Looks good overall. @qedawkins can you take a second look and unblock?

compiler/src/iree/compiler/Codegen/LLVMGPU/test/config_llvm_gpu_constraints.mlir
```diff
@@ -0,0 +1,374 @@
+// RUN: iree-opt --split-input-file \
+// RUN:   --iree-codegen-experimental-verify-pipeline-constraints \
+// RUN:   --pass-pipeline='builtin.module(func.func(iree-codegen-insert-smt-constraints))' %s \
+// RUN:   | FileCheck %s
```
qedawkins, line 4: nit: you have --split-input-file but no splits. Either add the splits or remove the flag (preferred to add splits)

qedawkins (approved):

compiler/src/iree/compiler/Codegen/LLVMGPU/test/config_llvm_gpu_constraints.mlir
RattataKing: There are splits at the bottom, the first test group is just a looong one.

kuhar (approved):
LGTM, thanks for all the fixes

PR #24094, opened 2026-04-13
[DMA][Swizzle] Apply inverse source swizzle in DMA lowering

lialan:
* Add `getDestSwizzleAttr` helper that traces destination memref through view-like ops to find SwizzleHintOp.
* Apply the swizzle attribute's offset transformation to source linear offsets in `gather_to_lds` lowering. XOR swizzle is self-inverse, so applying it to source addresses produces the correct swizzled layout in LDS without violating `gather_to_lds`'s uniform-destination constraint.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -9,6 +9,7 @@
 #include <optional>
 #include "iree/compiler/Codegen/Common/GPU/Passes.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.h"
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.h"
 #include "iree/compiler/Codegen/Utils/GPUUtils.h"
@@ -30,6 +31,7 @@
 #include "mlir/Dialect/Vector/Transforms/VectorTransforms.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/OpDefinition.h"
+#include "mlir/Interfaces/ViewLikeInterface.h"
 #include "mlir/Support/LLVM.h"
 #include "mlir/Transforms/WalkPatternRewriteDriver.h"
 
@@ -144,6 +146,19 @@ computeTransferSegments(int64_t totalElements, int64_t elementBits,
   return segments;
 }
 
+/// Trace a memref value through view-like ops to find a SwizzleHintOp.
+/// Returns the swizzle attribute if found, std::nullopt otherwise.
+static std::optional<IREE::Codegen::SwizzleAttrInterface>
+getDestSwizzleAttr(Value dest) {
+  while (auto viewOp = dest.getDefiningOp<mlir::ViewLikeOpInterface>()) {
+    dest = viewOp.getViewSource();
+  }
+  if (auto hintOp = dest.getDefiningOp<IREE::Codegen::SwizzleHintOp>()) {
+    return hintOp.getSwizzle();
+  }
+  return std::nullopt;
+}
+
 /// Generates source and destination indices for a GatherToLDS operation.
 ///
 /// The gather_to_lds instruction requires:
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -144,6 +146,19 @@ computeTransferSegments(int64_t totalElements, int64_t elementBits,
   return segments;
 }
 
+/// Trace a memref value through view-like ops to find a SwizzleHintOp.
+/// Returns the swizzle attribute if found, std::nullopt otherwise.
+static std::optional<IREE::Codegen::SwizzleAttrInterface>
+getDestSwizzleAttr(Value dest) {
```
Yu-Zhewen, line 153: Do we want to guarantee it's a XOR swizzle, as other swizzle types might not be self-inverse?

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -227,6 +228,14 @@ static void resolveHintOp(RewriterBase &rewriter,
       gatherToLDSOps.push_back(gatherToLDSOp);
       continue;
     }
```
Yu-Zhewen: Do we need to drop all `gatherToLDSOp` handlings from `ResolveSwizzleHints`, as they are now done in `AMDGPULowerCoalescedDMAToGatherLDS`.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -227,6 +228,14 @@ static void resolveHintOp(RewriterBase &rewriter,
...
+    // Treat reshape/subview ops as transparent users. When a swizzle_hint is
+    // used as the destination (via expand_shape/collapse_shape/subview) of a
+    // gather_to_lds, the swizzle is applied at the source-side in the DMA
+    // lowering pass, so these ops just pass through the swizzled allocation.
+    if (isa<memref::ExpandShapeOp, memref::CollapseShapeOp, memref::SubViewOp>(
```
Yu-Zhewen: We might not need this as well? When swizzle hint ops are inserted (https://github.com/iree-org/iree/pull/24001), I deliberately skip reinsertion for gather_to_lds, since it trusts AMDGPULowerCoalescedDMAToGatherLDS having already applied the inverse swizzling, and there’s nothing left to resolve further.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
lialan: Good call. Added an `isa<IREE::Codegen::XORShuffleAttr>` guard — `getDestSwizzleAttr` now returns `std::nullopt` for non-XOR swizzles (e.g. `rotate_rows`), since only XOR is self-inverse.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
lialan: Yes, dropped the source-side `gatherToLDSOp` swizzle handling (the `swizzleGatherToLDS` function and collection loop). `GatherToLDSOp` users are now just treated as transparent pass-through — the hint gets dropped without modifying source indices, since DMA lowering already applies the inverse swizzle before this pass runs.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
lialan: Kept the reshape/subview pass-through but updated the comment. These view-like ops are still needed as recognized transparent users: the DMA lowering creates `gather_to_lds` ops with dest chains like `subview(expand_shape(swizzle_hint(alloc)))`, so the `expand_shape`/`subview` show up as direct users of the `swizzle_hint` and would hit the "unsupported user" error without this.

Yu-Zhewen: I see. I was thinking we could just use `dest = subview(expand_shape(alloc))` after the inverse swizzle is inserted at the source, but this makes sense to me as well.

One more thing: Can we use `if (isa<amdgpu::GatherToLDSOp, ViewLikeOpInterface>(user))` here?

Yu-Zhewen (approved):
LGTM, thanks! Just one more nit above

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
lialan: Good idea, done. Combined into `isa<amdgpu::GatherToLDSOp, ViewLikeOpInterface>(user)` and replaced the `MemRef/IR/MemRef.h` include with `Interfaces/ViewLikeInterface.h`.

PR #24098, opened 2026-04-13
[CPU] Block root op chain fusion in non-root anchored tiling.

hanhanW:
When tiling from a non-root anchor op at VectorInnerParallelTiles level, there are two paths, because the intention is for tiling dimensions that not captured by rootOp:
(a) Start from an anchor op before the root op.
(b) Start from an anchor op after the root op.

In both cases, we should not fuse rootOp and the consumers/producers for (a)/(b).

The revision aims to fix the case that reduction dimension is not tiled on the root op which fused the root op unintended; it results in numeric issues as expected. This can happen when the reduction dimension is much less than native vector sizes.


Yu-Zhewen (approved):
LGTM

PR #24104, opened 2026-04-14
[Codegen] Add transfer_{gather/scatter} to vector.{gather/scatter} lowering

NoumanAmir657:
According to discussion with @Groverkss , this pass is supposed to run at the point where all N-D vectors have been legalized to 1-D so this pass should only handle the 1-D `transfer_{gather/scatter}` case.


krzysz00:
Week hold for I'm refactoring vector.gsther to make N-D indexing work in a more sensible way

Groverkss:
> Week hold for I'm refactoring vector.gsther to make N-D indexing work in a more sensible way

This only emits 1-D vector.gather. We don't have any problems with the N-D indexing stuff here.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // Check that all non-symbol source dims are constants or dim exprs.
+    for (unsigned i = 0; i < numResults - 1; ++i) {
+      AffineExpr expr = sourceMap.getResult(i);
+      if (!isa<AffineConstantExpr>(expr) && !isa<AffineDimExpr>(expr)) {
+        return rewriter.notifyMatchFailure(
+            op, "non-gathered source dims must be constants or dim exprs");
+      }
+    }
```
amd-eochoalo: Is this correct? Should all non-symbol source dims be dim exprs?

```suggestion
    // Check that the symbol appears only in the last source dimension.
    unsigned numResults = sourceMap.getNumResults();
    for (unsigned i = 0; i < numResults - 1; ++i) {
      if (sourceMap.getResult(i).isFunctionOfSymbol(0)) {
        return rewriter.notifyMatchFailure(
            op, "symbol must only appear in last source dim");
      }
    }
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_transfer_gather_to_vector_gather.mlir
```diff
@@ -0,0 +1,35 @@
+// RUN: iree-opt %s -pass-pipeline='builtin.module(func.func(iree-vector-ext-lower-transfer-gather-to-gather))' --split-input-file --mlir-print-local-scope | FileCheck %s
```
amd-eochoalo: The following test case, where %idx is an index type passes the 1-D legalization process as the 1-D legalization process (at the moment) only unrolls the result of vector.transfer_gather is 1 dimensional. (The current 1-D legalization process can be found here https://github.com/iree-org/iree/pull/24028 )

```
func.func @bug_scalar_index(%base: memref<8x16xf16>,
    %idx: index, %pad: f16) -> vector<8xf16> {
  %result = iree_vector_ext.transfer_gather %base[%c0, %c0]
    [%idx : index], %pad {
    indexing_maps = [
      affine_map<(d0)[s0] -> (0, s0)>,
      affine_map<(d0)[s0] -> ()>
    ]
  } : memref<8x16xf16>, vector<8xf16>
  return %result : vector<8xf16>
}
```

Passing this test case into this pass, gives the following error:

```
iree-opt --pass-pipeline="builtin.module(func.func(iree-llvmgpu-legalize-nd-vectors),util.func(iree-llvmgpu-legalize-nd-vectors))" bug3.mlir | iree-opt -pass-pipeline='builtin.module(func.func(iree-vector-ext-lower-transfer-gather-to-gather))'
<stdin>:6:10: error: 'vector.gather' op operand #3 must be vector of integer or index values, but got 'index'
    %0 = iree_vector_ext.transfer_gather %arg0[%c0, %c0] [%arg1 : index], %arg2 {indexing_maps = [#map, #map1]} : memref<8x16xf16>, vector<8xf16>
         ^
<stdin>:6:10: note: see current operation: %3 = "vector.gather"(%arg0, %0, %0, %arg1, %1, %2) : (memref<8x16xf16>, index, index, index, vector<8xi1>, vector<8xf16>) -> vector<8xf16>
```

When this operand is an index, this becomes a broadcast. This pattern is documented as lowering only gathers. I think you can modify your pattern to avoid matching operations where the only IndexVec is an IndexType and not a VectorType.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
```diff
@@ -0,0 +1,109 @@
+// Copyright 2025 The IREE Authors
```
amd-eochoalo:
```suggestion
// Copyright 2026 The IREE Authors
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
```diff
@@ -0,0 +1,109 @@
...
+    rewriter.replaceOpWithNewOp<vector::GatherOp>(op, resultType, op.getBase(),
+                                                  op.getOffsets(), indexVec,
+                                                  mask, passthru);
+    return success();
+  }
+};
+
+} // namespace
+
+namespace {
```
amd-eochoalo:
```suggestion
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_transfer_gather_scatter_to_vector.mlir
```diff
@@ -0,0 +1,35 @@
...
+// CHECK-LABEL: func.func @lower_transfer_gather_to_vector_gather
+// CHECK-SAME:  %[[SRC:.+]]: tensor<1x1x31xf32>
+// CHECK-DAG:     %[[PASS_THRU:.+]] = arith.constant dense<0.000000e+00> : vector<16xf32>
+// CHECK-DAG:     %[[MASK:.+]] = arith.constant dense<true> : vector<16xi1>
+// CHECK-DAG:     %[[STRIDE:.+]] = arith.constant dense<2> : vector<16xindex>
+// CHECK-DAG:     %[[C0:.+]] = arith.constant 0 : index
+// CHECK:         %[[STEP:.+]] = vector.step : vector<16xindex>
+// CHECK:         %[[INDICES:.+]] = arith.muli %[[STEP]], %[[STRIDE]] : vector<16xindex>
+// CHECK:         %[[GATHER:.+]] = vector.gather %[[SRC]][%[[C0]], %[[C0]], %[[C0]]] [%[[INDICES]]], %[[MASK]], %[[PASS_THRU]]
+// CHECK-SAME:      : tensor<1x1x31xf32>, vector<16xindex>, vector<16xi1>, vector<16xf32> into vector<16xf32>
```
amd-eochoalo, line 35: * Can you add negative tests for each pattern failure?
* Can you add a test for non-trivial leading dims?

amd-eochoalo (changes requested):

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
NoumanAmir657: Yeah, that's not correct. In the case of 1D transfer_gather, the non-symbol dims of the source map can only be constant. Did you mean that?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
NoumanAmir657: Fixed.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
NoumanAmir657: Fixed.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_transfer_gather_to_vector_gather.mlir
NoumanAmir657: Check added.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_transfer_gather_scatter_to_vector.mlir
NoumanAmir657: Added tests.

Groverkss (commented):
Can you also do vector.scatter in this patch? I would prefer having both of them together, it should be the same logic.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
amd-eochoalo: Correct, sorry my question was intended to be answer "non-symbol source dims should be constant", but my code snippet was wrong (which lead to some confusion)

```suggestion
    // Check that all non-symbol source dims are constants or dim exprs.
    for (unsigned i = 0; i < numResults - 1; ++i) {
      AffineExpr expr = sourceMap.getResult(i);
      if (!isa<AffineConstantExpr>(expr)) {
        return rewriter.notifyMatchFailure(
            op, "non-gathered source dims must be constants");
      }
    }
```

but I see that you already fixed that. Thanks!

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_transfer_gather_to_vector_gather.mlir
```diff
@@ -0,0 +1,150 @@
...
+}
+// CHECK-LABEL: @lower_gather_nontrivial_leading_dims
+// CHECK-DAG: %[[C2:.+]] = arith.constant 2 : index
+// CHECK-DAG: %[[C0:.+]] = arith.constant 0 : index
+// CHECK:     vector.gather %{{.+}}[%[[C2]], %[[C0]]]
+
+// -----
+
+// Multiple index vecs not lowered.
+func.func @no_lower_multiple_index_vecs(
```
amd-eochoalo: For negative test cases, can you prepend function's names with `negative` instead of `no`?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
```diff
@@ -0,0 +1,110 @@
...
+      if (!isa<AffineConstantExpr>(expr)) {
+        return rewriter.notifyMatchFailure(
+            op, "non-gathered source dims must be constants or dim exprs");
+      }
+    }
+
+    if (op.getMask()) {
+      return rewriter.notifyMatchFailure(
+          op, "masked transfer_gather not yet supported");
+    }
```
amd-eochoalo: I'm ok with incremental progress, but since @Groverkss already asked for scatter, I wonder if it would make sense to also support masked transfer_gather before merging.

amd-eochoalo (changes requested):
I think this looks good, just small nits. Please add a lowering for scatter case as requested by @Groverkss.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
NoumanAmir657: Just to pick your brain over this, in the case of 1D-vectors, the mask indexing map can only be this `(d0)[s0] -> (d0)`, because the verfier gurantees that mask indexing map cannot have symbols so handling mask here is just assigning the same mask to `vector.gather` if it satisfies the above indexing map.

Groverkss: Yes. The mask map will not have any symbols. You can directly use the same mask.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_transfer_gather_to_vector_gather.mlir
NoumanAmir657: Fixed.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherToGather.cpp
NoumanAmir657: Added the scatter lowering.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherScatterToVector.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.h.inc"
+
+namespace {
+
+static LogicalResult validateMaps(Operation *op, OperandRange indexVecs,
+                                  AffineMap sourceMap,
+                                  PatternRewriter &rewriter) {
+  if (indexVecs.size() != 1) {
+    return rewriter.notifyMatchFailure(op, "expected exactly one index vec");
+  }
```
Groverkss, line 34: Just add a TODO here that vector.gather has a restriction for this and we could flatten everything to 1-D to make it work.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherScatterToVector.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  for (unsigned i = 0; i < numResults - 1; ++i) {
+    if (!isa<AffineConstantExpr>(sourceMap.getResult(i))) {
+      return rewriter.notifyMatchFailure(op,
+                                         "non-gathered dims must be constants");
+    }
+  }
+
+  return success();
+}
```
Groverkss, line 53: Instead of 3, you can just do:

check all non last dims are affine constant expr.
check last dim is affine symbol expr.

Groverkss (approved):
LGTM, please wait for @amd-eochoalo to have a look as well

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherScatterToVector.cpp
NoumanAmir657: Added.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerTransferGatherScatterToVector.cpp
NoumanAmir657: Fixed.

amd-eochoalo (approved):
Thanks!

NoumanAmir657:
@Groverkss @amd-eochoalo Can one of you merge this for me?

Thanks for the review!!

NoumanAmir657:
Oh nvm you already did. Thanks!!

PR #24106, opened 2026-04-14
[runtime/python] Remove forced HOST_VISIBLE from AllocateBufferCopy

### LINKED ISSUE #24092: [rocm] Extern dispatch (hal.executable.source) receives stale data for function argument inputs on binding 0
When a function argument tensor is passed directly to a flow.dispatch targeting an extern HIP kernel (hal.executable.source), the kernel receives zeros/stale data at the corresponding binding. Tensors produced by IREE-internal dispatches (e.g. linalg.fill) arrive correctly at other bindings in the same dispatch.

Repro, see script below:

Input:    [11. 22. 33. 44.]   ← function arg, passed as binding 0
Output:   [ 0. 55. 777.  0.]
Expected: [11. 55. 777.  0.]
Binding 0 (function argument tensor) → zeros (wrong)
Binding 1 (linalg.fill result) → 55.0 (correct)
Binding 2 (output, kernel writes 777) → correct
The kernel provably executes (777 marker). The issue is that binding 0's buffer doesn't contain the host-uploaded function argument data.

Observed on: gfx1100 (RX 7900 XTX), ROCm 6.x, IREE at iree-org/iree HEAD (pip iree-base-compiler).

Python script:
```
import subprocess, tempfile, shutil, sys, os
from pathlib import Path
import numpy as np

SCRIPT_DIR = Path(__file__).parent
TARGET_ARCH = os.environ.get("TARGET_ARCH", "gfx1100")


def compile_kernel():
    co_path = SCRIPT_DIR / "probe_kernel.co"
    if co_path.exists():
        return str(co_path)
    cu_path = SCRIPT_DIR / "probe_kernel.cu"
    assert cu_path.exists(), f"Missing {cu_path}"
    subprocess.run([
        "/opt/rocm/llvm/bin/clang", "-x", "hip", "--offload-device-only",
        f"--offload-arch={TARGET_ARCH}", "--rocm-path=/opt/rocm", "-O3",
        str(cu_path), "-o", str(co_path),
    ], check=True)
    return str(co_path)


def compile_mlir(iree_compile, co_path):
    target = f'#hal.executable.target<"rocm", "rocm-hsaco-fb", {{target_arch = "{TARGET_ARCH}"}}>'

    mlir = f"""
#target = {target}
module @repro attributes {{hal.device.targets = [#hal.device.target<"hip", [#target]> : !hal.device]}} {{

  // Extern HIP kernel: reads binding 0 and 1, writes to binding 2.
  hal.executable.source private @k attributes {{
    objects = #hal.executable.objects<{{#target = [#hal.executable.object<{{path = "{co_path}"}}>]}}>
  }} {{
    hal.executable.export public @probe ordinal(0)
        layout(#hal.pipeline.layout<bindings = [
          #hal.pipeline.binding<storage_buffer, ReadOnly>,
          #hal.pipeline.binding<storage_buffer, ReadOnly>,
          #hal.pipeline.binding<storage_buffer>
        ]>) count(%d: !hal.device) -> (index, index, index) {{
      hal.return %c1, %c1, %c1 : index, index, index
    }} attributes {{workgroup_size = [64 : index, 1 : index, 1 : index]}}
  }}

  // Test function:
  //   binding 0 = %a (function argument)       → BUG: kernel sees zeros
  //   binding 1 = %b (linalg.fill result)      → OK:  kernel sees 55.0
  //   binding 2 = result (new allocation)       → OK:  kernel writes correctly
  util.func public @test(%a: tensor<4xf32>) -> tensor<4xf32> {{
    %cst = arith.constant 55.0 : f32
    %init = tensor.empty() : tensor<4xf32>
    %b = linalg.fill ins(%cst : f32) outs(%init : tensor<4xf32>) -> tensor<4xf32>
    %r = flow.dispatch @k::@probe(%a, %b) : (tensor<4xf32>, tensor<4xf32>) -> tensor<4xf32>
    util.return %r : tensor<4xf32>
  }}
}}
"""
    mlir_path = SCRIPT_DIR / "repro.mlir"
    mlir_path.write_text(mlir)

    vmfb_path = SCRIPT_DIR / "repro.vmfb"
    subprocess.run([
        iree_compile, str(mlir_path),
        "--iree-hal-target-backends=rocm",
        f"--iree-rocm-target={TARGET_ARCH}",
        "--iree-opt-const-eval=false",
        "-o", str(vmfb_path),
    ], check=True)
    return str(vmfb_path)


def main():
    iree_compile = shutil.which("iree-compile")
    assert iree_compile, "iree-compile not found in PATH"

    print(f"Target: {TARGET_ARCH}")
    co_path = compile_kernel()
    print(f"Kernel: {co_path}")
    vmfb_path = compile_mlir(iree_compile, co_path)
    print(f"VMFB:   {vmfb_path}")

    import iree.runtime as rt
    config = rt.Config(driver_name="hip")
    ctx = rt.SystemContext(config=config)
    with open(vmfb_path, "rb") as f:
        vm_module = rt.VmModule.from_flatbuffer(ctx.instance, f.read())
    ctx.add_vm_module(vm_module)
    test_fn = ctx.modules.repro["test"]

    a = np.array([11.0, 22.0, 33.0, 44.0], dtype=np.float32)
    out = np.asarray(test_fn(a))

    print()
    print(f"Input:    {a}")
    print(f"Output:   {out}")
    print(f"Expected: [11. 55. 777. 0.]")
    print()

    ok = out[0] == 11.0 and out[1] == 55.0 and out[2] == 777.0
    if ok:
        print("PASS")
    else:
        print("FAIL")
        if out[2] == 777.0:
            print("  ✓ Kernel executed (marker=777)")
        if out[1] == 55.0:
            print("  ✓ Binding 1 (linalg.fill) correct")
        if out[0] != 11.0:
            print(f"  ✗ Binding 0 (function arg) WRONG: {out[0]} (expected 11.0)")

    return 0 if ok else 1


if __name__ == "__main__":
    sys.exit(main())
```

probe_kernel.cu:

```
// Minimal HIP kernel for IREE extern dispatch repro.
// Copies first element of binding 0 and binding 1 to the output (binding 2).
// Writes 777.0 as a marker proving the kernel executed.
#include <hip/hip_runtime.h>

extern "C" __global__ void probe(float* a, float* b, float* out) {
    if (threadIdx.x == 0) {
        out[0] = a[0];    // binding 0: should be function arg
        out[1] = b[0];    // binding 1: should be fill-computed value
        out[2] = 777.0f;  // marker: kernel ran
    }
}
```

jtuyls:
Resolves: https://github.com/iree-org/iree/issues/24092

AllocateBufferCopy forced HOST_VISIBLE and callers requested MAPPING — leftovers from before the function was refactored to use transfer_h2d().

On HIP this causes hipMallocManaged allocation + memcpy fast path in transfer_h2d (buffer_transfer.c:212), which lacks synchronization with subsequent GPU dispatches. The GPU reads stale data (zeros).

Remove HOST_VISIBLE from AllocateBufferCopy and MAPPING from callers to match iree-run-module behavior (DEVICE_LOCAL + DEFAULT, function_io.c:448). transfer_h2d then takes the command buffer path with proper sync.

Benchmark (FunctionInvoker add_one, 500 iters, gfx1100 RX 7900 XTX):

| Size | KB | HIP before | HIP after | Vulkan before | Vulkan after |
|------|---:|-----------|----------|---------------|-------------|
| 4x4 | 0 | 3.44 ms (**WRONG**) | 0.11 ms | 0.13 ms | 0.16 ms |
| 64x64 | 16 | 3.31 ms (**WRONG**) | 0.11 ms | 0.13 ms | 0.16 ms |
| 256x256 | 256 | 3.71 ms (**WRONG**) | 0.61 ms | 0.19 ms | 0.28 ms |
| 512x512 | 1024 | 4.39 ms (**WRONG**) | 1.29 ms | 0.38 ms | 0.55 ms |

HIP: all correct=False before, correct=True after. 3-30x faster. Vulkan: correct both before and after. ~1.4x slower for large buffers due to command buffer path vs memcpy, but logic matching iree-run-module now.

I don't see a good way to add tests for this as python tests aren't enabled in hip CI runs?

### DIFF
--- a/runtime/src/iree/hal/cts/buffer/allocator_test.cc
+++ b/runtime/src/iree/hal/cts/buffer/allocator_test.cc
@@ -227,6 +227,23 @@ TEST_P(AllocatorTest, ImportHostAllocationWithCallback) {
   iree_allocator_free_aligned(iree_allocator_system(), host_ptr);
 }
 
+// Requesting DEVICE_LOCAL + MAPPING (without HOST_VISIBLE) must succeed.
+// The allocator should promote the memory type so that the buffer is
+// mappable — the Python bindings produce this combination via
+// asdevicearray(allowed_usage=DEFAULT|MAPPING).
+TEST_P(AllocatorTest, DeviceLocalMappingAllocationSucceeds) {
+  iree_hal_buffer_params_t params = {0};
+  params.type = IREE_HAL_MEMORY_TYPE_DEVICE_LOCAL;
+  params.usage = IREE_HAL_BUFFER_USAGE_DEFAULT | IREE_HAL_BUFFER_USAGE_MAPPING;
+
+  // Allocation must not fail with "mappable buffers require host pointers".
+  iree_hal_buffer_t* buffer = NULL;
+  IREE_ASSERT_OK(iree_hal_allocator_allocate_buffer(device_allocator_, params,
+                                                    kAllocationSize, &buffer));
+  ASSERT_NE(nullptr, buffer);
+  iree_hal_buffer_release(buffer);
+}
+
 CTS_REGISTER_TEST_SUITE(AllocatorTest);
 
 }  // namespace iree::hal::cts
--- a/runtime/bindings/python/hal.cc
+++ b/runtime/bindings/python/hal.cc
@@ -192,8 +192,7 @@ py::object HalAllocator::AllocateBufferCopy(
   PyBufferReleaser py_view_releaser(py_view);
 
   iree_hal_buffer_params_t params = {0};
-  // TODO: Should not require host visible :(
-  params.type = memory_type | IREE_HAL_MEMORY_TYPE_HOST_VISIBLE;
+  params.type = memory_type;
   params.usage = allowed_usage;
 
   iree_hal_buffer_t* hal_buffer = nullptr;
--- a/runtime/bindings/python/invoke.cc
+++ b/runtime/bindings/python/invoke.cc
@@ -216,8 +216,8 @@ class InvokeStatics {
 
             retained_bv = c.allocator().AllocateBufferCopy(
                 IREE_HAL_MEMORY_TYPE_DEVICE_LOCAL,
-                IREE_HAL_BUFFER_USAGE_DEFAULT | IREE_HAL_BUFFER_USAGE_MAPPING,
-                c.device(), host_array, hal_element_type);
+                IREE_HAL_BUFFER_USAGE_DEFAULT, c.device(), host_array,
+                hal_element_type);
             bv = py::cast<HalBufferView*>(retained_bv);
           }
 
@@ -420,8 +420,7 @@ class InvokeStatics {
 
       // Put it on the device.
       py::object retained_bv = c.allocator().AllocateBufferCopy(
-          IREE_HAL_MEMORY_TYPE_DEVICE_LOCAL,
-          IREE_HAL_BUFFER_USAGE_DEFAULT | IREE_HAL_BUFFER_USAGE_MAPPING,
+          IREE_HAL_MEMORY_TYPE_DEVICE_LOCAL, IREE_HAL_BUFFER_USAGE_DEFAULT,
           c.device(), host_array, hal_element_type);
       HalBufferView* bv = py::cast<HalBufferView*>(retained_bv);
 
--- a/runtime/src/iree/hal/drivers/cuda/cuda_allocator.c
+++ b/runtime/src/iree/hal/drivers/cuda/cuda_allocator.c
@@ -278,6 +278,18 @@ iree_hal_cuda_allocator_query_buffer_compatibility(
     }
   }
 
+  // If the caller requests mappable device-local memory but did not set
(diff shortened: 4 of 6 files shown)


runtime/src/iree/hal/cts/buffer/allocator_test.cc
```diff
@@ -227,6 +227,23 @@ TEST_P(AllocatorTest, ImportHostAllocationWithCallback) {
   iree_allocator_free_aligned(iree_allocator_system(), host_ptr);
 }
 
+// Requesting DEVICE_LOCAL + MAPPING (without HOST_VISIBLE) must succeed.
```
benvanik, line 230: this test can't be a part of the CTS - not all devices support device-local + mapping.
the hip one can stay, though

benvanik (changes requested):

runtime/src/iree/hal/cts/buffer/allocator_test.cc
jtuyls: Ok, deleted it now. Could you have another look?

benvanik (approved):

PR #24109, opened 2026-04-14
[CPU] Factor thread utilization into mmt4d lowering tile selection

bjacob:
Extend the mmt4d distribution cost model with a penalty when the M×N tile count is too small to use the configured runtime parallelism, so we prefer shapes that keep more workgroups active.

[Benchmark](https://docs.google.com/spreadsheets/d/1q0pY8uAI4ApGSUndJYAdSq7CKGRzTJgaWlEzMkvSL9A/edit?usp=sharing) (3 PRs together: #24107, #24108, #24109) showing no regression across many matmul shapes and slight improvement to scaling SDXL-clip to many threads.

The point is not the tiny perf improvement per se, but ensuring that existing parameters have the intended semantics.

Made-with: Cursor

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -1894,20 +1894,21 @@ getMmt4dLoweringConfig(linalg::LinalgOp op, DictionaryAttr targetConfig) {
   const SmallVector<std::pair<int64_t, int64_t>> divisorsOfM1 = getDivisors(M1);
   const SmallVector<std::pair<int64_t, int64_t>> divisorsOfN1 = getDivisors(N1);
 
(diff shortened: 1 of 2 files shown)


hanhanW (approved):
Is the link to `Benchmark` wrong? It links to self.

bjacob:
Thanks, fixed the benchmark link.

hanhanW:
> Thanks, fixed the benchmark link.

If the table is small, how about inling the benchmark results just in case if the doc becomes stale?

bjacob:
The linked spreadsheet is not small, and while one could extract some subset of it, it's not about "here is a small piece of data showing that it's a clear improvement". It's more about "here is a lot of data that shows it's not a regression".

hanhanW:
> The linked spreadsheet is not small, and while one could extract some subset of it, it's not about "here is a small piece of data showing that it's a clear improvement". It's more about "here is a lot of data that shows it's not a regression".

I missed the other tab, sorry about that. 🤦‍♂️

PR #24113, opened 2026-04-14
[DMA][Swizzle] Enforce access_width >= elementsPerLane in DMA lowering

lialan:
* When a `CoalescedGatherDMAOp`'s destination carries an XOR swizzle hint, the `gather_to_lds` lowering applies the inverse swizzle only to each lane's base offset.
* Per-lane contiguous reads of `elementsPerLane` elements must stay inside one `access_width`-sized swizzle chunk, requiring `elementsPerLane <= access_width` and `access_width % elementsPerLane == 0`.

This patch adds a guard for now to emit a clear op error when the selected DMA size violates the invariant, replacing the prior silent miscompile. Adds a negative lit test covering the error path; existing positive tests already cover compatible configs.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -166,6 +166,35 @@ getDestSwizzleAttr(Value dest) {
   return std::nullopt;
 }
 
+/// Verifies that every transfer segment is compatible with the destination
+/// XOR swizzle's |accessWidth|.
+///
+/// `gather_to_lds` writes |elementsPerLane| contiguous elements per lane at
+/// `dstBase + laneId * elementsPerLane`, while the inverse XOR swizzle is
+/// applied only to each lane's base offset. The lane therefore reads
+/// `src[swizzle(perLaneBase) + k]` for k in [0, elementsPerLane). For this
+/// to match the swizzled LDS layout we need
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -166,6 +166,35 @@ getDestSwizzleAttr(Value dest) {
...
+///   accessWidth % elementsPerLane == 0.
+///
+/// Returns the offending segment's elementsPerLane on failure, std::nullopt
+/// on success.
+static std::optional<int64_t>
+findSwizzleIncompatibleSegment(ArrayRef<TransferSegment> segments,
+                               int64_t accessWidth) {
+  for (const TransferSegment &seg : segments) {
+    if (seg.elementsPerLane > accessWidth ||
+        accessWidth % seg.elementsPerLane != 0) {
```
sommerlukas: Nit: `accessWidth % elementsPerLane` will return `accessWidth` (which is `!= 0`) if `elementsPerLane > accessWidth`, so the two conditions are redundant. 

But we can also keep both to clearly state intent.

sommerlukas (approved):

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
Yu-Zhewen: +1 on merging into a single divisibility check

Yu-Zhewen (approved):

PR #24114, opened 2026-04-14
[Codegen][GPU] Guard multi-buffering behind pipelining pre-flight check

jerryyin:
Move computeStageClassification before multiBufferLDSAllocations as a pre-flight validation. If stage classification fails (e.g. gather_to_lds inside scf.if, insufficient loop iterations), bail out before mutating the IR. This prevents doubled LDS allocations with no actual pipelining.

After multi-buffering succeeds, re-run classification on the mutated IR to capture the correct double buffered schedule.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -1254,10 +1254,6 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
       LDBG() << "Async copy mode supports at most 3 stages, got " << numStages;
       return failure();
     }
-    // Apply multi-buffering: numStages buffers for N-stage pipelining
-    if (failed(multiBufferLDSAllocations(forOp, /*numBuffers=*/numStages))) {
-      return failure();
-    }
   } else {
     // Stream copy: no buffering, just validate numStages
     // No prefetching needed for single-stage pipelining.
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -1274,7 +1270,25 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
...
+  if (mode == PipelineMode::AsyncCopy) {
+    // Apply multi-buffering: numStages buffers for N-stage pipelining.
+    if (failed(multiBufferLDSAllocations(forOp, /*numBuffers=*/numStages))) {
+      return failure();
+    }
+  }
+
+  // Re-run classification on the (potentially multi-buffered) IR to capture
+  // any new operations introduced by multi-buffering in the schedule.
   auto stagesOr = computeStageClassification(forOp, mode, numStages);
```
Yu-Zhewen, line 1292: I’m wondering if by any chance, multi-buffering itself could cause the classification to fail during a re-run? Or this can an assertion here.

Yu-Zhewen (approved):
LGTM, thanks!

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
jerryyin: Multi-buffering on its own has very little risk to pipeline because the only IR mutation is type + modulo compute which is nothing special to pipeline. Theoretically speaking there are several places that can return failure after multi-buffer, but the actual risk is quite low.

The alternative of this can be much more challenging: un-transform the IR when something goes wrong - I don't find any reason to do that because the failing cases are mostly straightforward and can easily be filtered off by this pre-flight check.

Yu-Zhewen: Yeah, I agree. Un-transforming the IR feels like overkill

PR #24115, opened 2026-04-14
[GPUHeuristics] Improve large GEMM intrinsic selection on CDNA4

yzhang93:
Extend the compute-throughput-first intrinsic preference to LargeGemm shapes, preferring MFMA_F32_32x32x16_F16 over MFMA_F32_16x16x32_F16 (4x more output per instruction). Add VGPR pressure cap to prevent spilling when MNT boost sets high tile counts with 32x32 intrinsics.

Top GEMM improvements on MI355X:
```
4096x1024x150000:  2112us -> 1538us (1.37x)
2268x4096x150000: 11359us -> 8529us (1.33x)
1024x4096x150000:  1982us -> 1573us (1.26x)
4096x2048x150000:  4015us -> 3307us (1.21x)
2048x8192x4096:     183us ->  154us (1.19x)
```

Top conv improvements on MI355X (NHWC, fp16):
```
n32 c256 H100xW100 k2376 3x3 wgrad: 7983us -> 6634us (1.20x)
n32 c256 H25xW25   k2376 3x3 wgrad:  777us ->  664us (1.17x)
n32 c256 H100xW100 k2376 3x3 fwd:   7042us -> 6122us (1.15x)
n32 c256 H25xW25   k2376 3x3 fwd:    452us ->  405us (1.12x)
n32 c256 H50xW50   k2376 3x3 fwd:   1711us -> 1541us (1.11x)
```

Overall GEMM benchmark: **+6.3%** geomean speedup.
Overall Proxy conv benchmark: **+2.5%** geomean speedup.

Some regressions exist in K-dominated wgrad shapes due to larger workgroup tiles, but overall improvements outweigh regressions across all benchmarks.


compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -898,6 +904,11 @@ static void adjustSeedsForTarget(GPUMMAHeuristicSeeds &seeds,
           std::max(seeds.bestMNTileCountPerSubgroup, boostMNT);
       LDBG() << "Boosting MNT to " << seeds.bestMNTileCountPerSubgroup
              << " for balanced large gemm";
+      // Halve subgroup count to offset the MNT boost, keeping the total
+      // workgroup resource footprint (threads, LDS) in check for occupancy.
+      seeds.bestSubgroupCountPerWorkgroup /= 2;
```
lialan: nit: i don't know, maybe we should have a minimal value guard?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -928,6 +939,28 @@ static void adjustSeedsForTarget(GPUMMAHeuristicSeeds &seeds,
...
+
+  // Cap per-subgroup MN tile count based on output VGPR pressure from the
+  // selected intrinsic. Only applies when the MNT boost (step 2) is
+  // configured, since the boost can push MN tile counts high enough to
+  // cause spilling with large-output intrinsics (32x32). Capping at 128
+  // output VGPRs per thread (8 MN tiles for 32x32, 32 for 16x16) prevents
+  // spilling while preserving the boost for intrinsics that can handle
+  // higher tile counts.
+  if (seeds.boostMNTileCountPerSubgroup) {
+    constexpr int64_t kMaxOutputVGPRsPerThread = 128;
```
lialan: we should move this k definition to header.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -775,7 +776,9 @@ static bool compareIntrinsics(const GPUMatmulShapeType &problem,
   // (compute=8192, area=512) because throughput matters more. Among
   // 16x16x32 and 32x32x16 (both area=1024), prefer smaller K (16 vs 32)
   // for less operand staging pressure.
-  if (problem.gemmSize == GemmSizeKind::VeryLargeGemm) {
+  if (problem.gemmSize == GemmSizeKind::VeryLargeGemm ||
```
lialan: so the new boolean is just make `LargeGemm` behave like `VeryLargeGemm`, not sure if this is "preferHighCompute"?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
```diff
@@ -938,8 +971,10 @@ FailureOr<GPUMMASchedule> deduceMMASchedule(
     bool useDirectLoad, int64_t prefetchNumStages, bool mustBeAligned,
     bool doCPromotion, int64_t splitReductionTripCnt) {
 
+  bool preferHighComputeIntrinsic =
+      !doCPromotion && seeds.boostMNTileCountPerSubgroup.has_value();
```
lialan: I don't know, maybe just change `problem.gemmSize` here to avoid all the hustles?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
yzhang93: Moved it to be a seed field `maxOutputVGPRsPerThread` on `GPUMMAHeuristicSeeds` instead. This way each architecture can set its own VGPR cap — currently only CDNA4 LargeGemm sets it to 128.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp
yzhang93: Moved the problem size decision (VeryLargeGemm / LargeGemm) into the `preferHighComputeIntrinsic` flag decision in `deduceMMASchedule`. Added comments explaining when each condition applies. This keeps intrinsic sorting logic clean and all the gating decisions in one place where the full context (problem, seeds, doCPromotion) is available.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1351,9 +1351,10 @@ static constexpr ArchSeedSet kCDNA4Seeds = {
     /*gemm=*/{
         /*SmallGemm=*/     {2, 2,  4, 2 * kCacheLineSizeBits},
         /*MediumGemm=*/    {4, 8,  4, 2 * kCacheLineSizeBits},
-        /*LargeGemm=*/     {4, 16, 2, kCacheLineSizeBits / 2,
+        /*LargeGemm=*/     {8, 16, 2, kCacheLineSizeBits / 2,
                             /*minUtilizationThreshold=*/0.50,
-                            /*boostMNTileCountPerSubgroup=*/32},
+                            /*boostMNTileCountPerSubgroup=*/32,
+                            /*maxOutputVGPRsPerThread=*/128},
```
lialan, line 1357: Curious about `maxOutputVGPRsPerThread`:
* can we deduce it from something else? if we want to achieve certain occupancy rate, we should be able to calculate how many VGPRs can be used per thread?

yzhang93: Yeah we could compute it from `vgprSpaceBits` and a target occupancy, but the "right" occupancy and how much budget to reserve for non-output registers are still empirical guesses. Felt simpler to just set it directly so it's easy to see and tweak. The 128 comes from: CDNA4 has 512 VGPRs/SIMD, targeting 2+ waves occupancy gives ~256 VGPRs/thread, half reserved for operand staging and loop overhead leaves 128 for output.

lialan (approved):

PR #24116, opened 2026-04-14
[Codegen][GPU] Remove the single-iteration workaround and distribute pad-fused copies

jerryyin:
Previously, pad-fused `linalg.copy` operations were wrapped in a single-iteration `scf.forall` (TODO #23365), restricting the DMA load to one warp. After single-iteration loop removal, this produced an `scf.if` predicated on `thread_id` which blocked software pipelining — the pipeliner cannot handle `gather_to_lds` inside conditional blocks. On a 1134×2048×150000 bf16 GEMM (LHS transposed), this caused a ~36% direct-load regression.

This PR does the following:
 1. Remove single-iteration workaround in `tileAtSubgroupLevel`: Pad-fused copies now use `computeSubgroupTileSizes` for multi-warp distribution, same as non-padded copies.
 2. Propagate source offsets in `createDMAInForall`: When a `tensor.pad` is detected, trace from the tiled `extract_slice` back to the original pad to recover warp offsets. Create a new `tensor.extract_slice` from the pre-pad source with offsets and sizes clamped to actual source bounds via `arith.minsi/maxsi/subi`. The DMA's `in_bounds` attribute signals which dimensions may read OOB, relying on hardware (`fat_raw_buffer`) to return zeros.
 3. Guard `ConvertPadFusionCopyToCoalescedDMA`: Skip copies already distributed into warp-mapped foralls, making this pattern a fallback for edge cases.

**Before** (single warp, blocks pipelining):
```
scf.forall (%warp) = (0, 0) to (4, 64) step (4, 64)   // 1 iteration
  coalesced_gather_dma %pre_pad_source into %full_init
```

**After** (all warps participate, pipelining enabled):
```
scf.forall (%warp) = (0, 0) to (4, 64) step (1, 64)   // 4 iterations
  %clamped = extract_slice %pre_pad_source[minsi(%warp, %src_dim), 0]
                                          [minsi(%remaining, 1), 64]
  coalesced_gather_dma %clamped into %warp_init in_bounds [false, true]
```

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -455,9 +455,88 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
       }
 
       if (validPad) {
-        // Use pad.getSource() directly as the DMA source.
-        // This is the tensor.extract_slice result (e.g., tensor<?x64xf32>).
-        source = pad.getSource();
+        Value preSource = pad.getSource();
+
+        // Find the extract_slice closest to the pad output. This is the
+        // subgroup-level tiling's extract_slice, whose offsets tell us
+        // where this warp's tile starts within the padded tensor.
+        tensor::ExtractSliceOp tilingES;
+        {
+          Value trace = input;
+          while (auto es = trace.getDefiningOp<tensor::ExtractSliceOp>()) {
+            trace = es.getSource();
+            if (trace.getDefiningOp<tensor::PadOp>() == pad) {
+              tilingES = es;
+              break;
+            }
+          }
+        }
+
+        if (tilingES) {
+          // Subgroup tiling applied — create a sub-slice of the pre-pad
+          // source at the warp's tiling offset with sizes clamped to
+          // source bounds. Since pad has low=[0,0], the coordinate
+          // systems of the padded output and pre-pad source are aligned.
+          // The DMA's in_bounds attribute handles the case where the
+          // clamped source is smaller than the warp's init tile (the
+          // fat_raw_buffer returns zero for OOB reads).
+          rewriter.setInsertionPoint(inParallelOp);
+
+          SmallVector<OpFoldResult> warpOffsets = tilingES.getMixedOffsets();
+          SmallVector<OpFoldResult> warpSizes = tilingES.getMixedSizes();
+          auto preSourceType = cast<RankedTensorType>(preSource.getType());
+          int64_t rank = preSourceType.getRank();
+
+          SmallVector<OpFoldResult> subOffsets, subSizes, subStrides;
+          for (int64_t i = 0; i < rank; i++) {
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -455,9 +455,88 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
...
+            } else {
+              subOffsets.push_back(warpOffsets[i]);
+              subSizes.push_back(warpSizes[i]);
+            }
+          }
+
+          source = tensor::ExtractSliceOp::create(
+              rewriter, loc, preSource, subOffsets, subSizes, subStrides);
+        } else {
+          // No subgroup tiling (single-warp case) — use full pre-pad
```
lialan, line 536: i am wondering if we will see a case where there are no subgroup tiling today.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -455,9 +455,88 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
...
-        source = pad.getSource();
+        Value preSource = pad.getSource();
+
+        // Find the extract_slice closest to the pad output. This is the
+        // subgroup-level tiling's extract_slice, whose offsets tell us
+        // where this warp's tile starts within the padded tensor.
+        tensor::ExtractSliceOp tilingES;
+        {
+          Value trace = input;
+          while (auto es = trace.getDefiningOp<tensor::ExtractSliceOp>()) {
```
Yu-Zhewen, line 466: Can we just look at the direct user of the pad op, rather than tracing backwards from copy input?

jerryyin: Yes, good point.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
jerryyin: You are right, such case doesn't exist. I get rid of the else condition.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -458,7 +493,79 @@ static LogicalResult createDMAInForall(scf::ForallOp threadForallOp,
     // We need to trace through extract_slice to find if source is tensor.pad.
     tensor::PadOp pad = traceToTensorPad(input);
     if (pad && isValidPadForDMA(pad)) {
-      source = pad.getSource();
+      Value preSource = pad.getSource();
```
Yu-Zhewen: NIt: Can we move line 496 to line 567 into a function for clarity

Yu-Zhewen (approved):
LGTM

PR #24121, opened 2026-04-15
compiler/plugins/input/TOSA: fix: TOSA arith lowering must handle apply scale introduced by linalg lowering

Manewing:
TOSA to linalg lowering may (re-)introduce some additional TOSA operations as not all TOSA operations can be lowering to linalg. These need a following run of TOSA to arith lowering for element wise operations. The current pass pipeline is missing the option to enable lowering of tosa.apply_scale, which may be introduced during the lowering to linalg. This causes errors in the later stage of the compilation flow such as vectorization.

### DIFF
--- a/compiler/plugins/input/TOSA/InputConversion/test/apply_scale_lowering.mlir
+++ b/compiler/plugins/input/TOSA/InputConversion/test/apply_scale_lowering.mlir
@@ -1,0 +1,29 @@
+// RUN: iree-compile --compile-to=input --split-input-file %s | FileCheck %s
+
+// Make sure that tosa.apply_scale operations generated during TOSA to
+// linalg lowering, are properly lowered during TOSA to arith lowering
+
+// tosa.mul with a non-zero integer shift lowers through tosa.apply_scale
+//
+// CHECK-LABEL: util.func public @shifted_mul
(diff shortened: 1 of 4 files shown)


lialan:
Yeah you need to update the lit tests as well.

Manewing:
@lialan sry for the confusion, I incorrectly assumed that `use32Bit = true` would also be required for proper lowering. However, after further investigation it turned out that it is not. Now the LIT tests are also not failing anymore!

compiler/plugins/input/TOSA/InputConversion/test/apply_scale_lowering.mlir
egebeysel, line 1: nit: I understand this is due to the apply rescale flag turned on and the tests here are testing them, but that might not be entirely clear without the context of the PR. I would leave a comment here.

egebeysel (approved):
Thanks, LGTM!

@lialan if you don't have anything to add here, I'll help @Manewing merge this in :)

compiler/plugins/input/TOSA/InputConversion/test/apply_scale_lowering.mlir
Manewing: I added a comment in the tests

PR #24123, opened 2026-04-15
[Codegen][LinalgExt] Port attention rewrites to OnlineAttentionOp

keshavvinayak01:
Rewriting `AttentionOp` patterns to work with `OnlineAttention` with templatization.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ReshapeFusion.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ReshapeFusion.cpp
@@ -1334,32 +1334,37 @@ SmallVector<unsigned> defaultControlDropUnitDims(Operation *op) {
   return llvm::to_vector(llvm::seq<unsigned>(0, fusionOp.getNumLoops()));
 }
 
-struct DropAttentionUnitDims final
-    : OpRewritePattern<IREE::LinalgExt::AttentionOp> {
-  DropAttentionUnitDims(MLIRContext *context,
-                        linalg::ControlDropUnitDims options,
-                        PatternBenefit benefit = 1)
-      : OpRewritePattern<AttentionOp>(context, benefit),
+template <typename AttentionOpType>
+struct DropAttentionLikeUnitDims final : OpRewritePattern<AttentionOpType> {
+  DropAttentionLikeUnitDims(MLIRContext *context,
+                            linalg::ControlDropUnitDims options,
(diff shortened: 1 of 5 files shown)


MaheshRavishankar (approved):

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ReshapeFusion.cpp
```diff
@@ -1334,32 +1334,37 @@ SmallVector<unsigned> defaultControlDropUnitDims(Operation *op) {
...
+      for (auto [oldOperand, newOperand] :
+           llvm::zip(attentionOp->getOperands(), newOperands)) {
+        mapping.map(oldOperand, newOperand);
+      }
+      auto *cloned = b.clone(*attentionOp, mapping);
+      auto newOp = cast<AttentionOpType>(cloned);
+      for (auto [result, type] : llvm::zip(newOp->getResults(), resultTypes)) {
+        result.setType(type);
+      }
+      newOp.setIndexingMapsAttr(b.getAffineMapArrayAttr(newIndexingMaps));
```
IanWood1, line 1367: Can we add a similar builder to OnlineAttentionOp so we can keep `::create()` instead of cloning? 

```
OpBuilder<(ins "TypeRange":$results, "ValueRange":$inputOperands,
                 "ValueRange":$initOperands, "ArrayAttr":$indexing_maps)>
```


Besides being simpler to go through the create method, cloning the op creates it in an invalid state which can be problematic for listeners. For example, debug listeners that ensure that no transitively invalid IR is generated.

IanWood1, line 1367: Do we need to tell the rewriter that we are modifying an op that we created? I'm actually not sure about this.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/TransposeFusion.cpp
```diff
@@ -178,15 +179,17 @@ struct BubbleTransposeVFromAttentionOp
 void populateFuseLinalgExtOpsWithTransposes(
```
IanWood1, line 179: Minor nit: the comments in Transforms.h for `populateFuseLinalgExtOpsWithTransposes()` and `populateBubbleTransposeFromLinalgExtOps()` say that only AttentionOp is supported. That comment could be updated/removed.

PR #24131, opened 2026-04-15
Some fixes to work around a bug in cmake, and isa naming

AWoloszyn:


build_tools/cmake/iree_amdgpu_binary.cmake
krzysz00, line 92: Can you expand on why you need this?

AWoloszyn: Sorry I thought I had a comment that seems to have gotten eaten. I will put up another comment later, but long story short. There is a bug in cmake that in CERTAIN circumstances (not sure exactly what the env difference is, since the same version works for Ben and not for me) having targets that have the same 
1) name
2) single source file
3) MAIN_DEPENDENCY 
cause a failure in iree_cc_library. 
This showed up in the amdgpu backend with
blit.c
blit (the bitcode)
blit (the library)
Where it asserted that no source files were being added to the library.

PR #24132, opened 2026-04-15
[DMA] Reject invalid pads early in DMA convertible check

Yu-Zhewen:
Extracts pad validation into a standalone `isValidPadForDMA` helper and calls it in `isCopyDMAConvertible`, so copies with invalid pads are rejected early and gracefully downgraded to `derived_thread_config` instead of failing deep in `createDMAInForall`.

As part of the refactoring, also fixes a bug where pads with dynamic innermost source dimensions were incorrectly assumed to be DWORD-aligned.

Assisted-by: Cursor (Claude)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -591,7 +591,17 @@ struct ConvertToCoalescedDMABase : OpRewritePattern<OpTy> {
       return failure();
     }
 
-    return createDMAInForall<OpTy>(threadForallOp, rewriter);
+    // createDMAInForall must not fail after tileToThreadLevel, because
+    // tileToThreadLevel already erased the original op via replaceOp.
+    // Failing here would leave a dangling reference (use-after-free).
+    // All eligibility checks must happen before this point (e.g., in
+    // isCopyDMAConvertible / computeThreadNumThreads).
+    [[maybe_unused]] LogicalResult result =
+        createDMAInForall<OpTy>(threadForallOp, rewriter);
+    assert(succeeded(result) &&
+           "createDMAInForall must not fail after tileToThreadLevel erased "
+           "the original op");
+    return success();
   }
 
 protected:
@@ -667,7 +677,12 @@ struct ConvertPadFusionCopyToCoalescedDMA : OpRewritePattern<linalg::CopyOp> {
       return failure();
     }
 
-    return createDMAInForall<linalg::CopyOp>(threadForallOp, rewriter);
+    [[maybe_unused]] LogicalResult result =
+        createDMAInForall<linalg::CopyOp>(threadForallOp, rewriter);
+    assert(succeeded(result) &&
+           "createDMAInForall must not fail after tileToThreadLevel erased "
+           "the original op");
+    return success();
   }
 };
 
@@ -756,9 +771,8 @@ struct ConvertGatherToCoalescedDMA
       return WalkResult::interrupt();
     });
 
-    if (!tiledGatherOp) {
-      return failure();
-    }
+    assert(tiledGatherOp &&
+           "tiled gather op must exist after tileToThreadLevel");
 
     Block *forallBody = threadForallOp.getBody();
     Value sharedOut = forallBody->getArguments().back();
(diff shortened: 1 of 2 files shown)


lialan (approved):
+1

jerryyin (changes requested):
I'm a bit worried that we'd no longer have visibility into what we actually rejected. Mentality wise, I think it is justified if rejection is for performance reasons, not correctness reasons. 
 1) We need to keep track of every correctness issues that got shielded off before this PR is merged Document them in tickets, whether that's related with #24139 or not.
 2) When all the correctness issues are fixed, we need to come back revert this and experiment between DMA and non-DMA pass to make proper decisions here.

For now I'd say document and merge. We can revisit this after bringup is done.

Yu-Zhewen:
> I'm a bit worried that we'd no longer have visibility into what we actually rejected. Mentality wise, I think it is justified if rejection is for performance reasons, not correctness reasons.
> 
> 1. We need to keep track of every correctness issues that got shielded off before this PR is merged Document them in tickets, whether that's related with #24139 or not.
> 2. When all the correctness issues are fixed, we need to come back revert this and experiment between DMA and non-DMA pass to make proper decisions here.
> 
> For now I'd say document and merge. We can revisit this after bringup is done.

Just to clarify, this PR fixes two bugs, not shields them.

1. On main with DMA enabled, [e2e_matmul_cdna4_pad_i8](https://github.com/iree-org/iree/blob/main/tests/e2e/matmul/CMakeLists.txt#L2348-L2370) segfaults during compilation. In `ConvertCopyToCoalescedDMA`, `tileToThreadLevel` erases the original op, then `createDMAInForall` rejects the pad source and fails the pattern match, causing a use-after-free. The fix moves pad validation earlier, and asserts that `createDMAInForall` must succeed after tiling.

2. After fixing the crash, the test was producing wrong results because the original DWORD alignment check didn't reject a dynamic innermost dim. Fixed now.

(For "1", ideally we can refactor `tileToThreadLevel` and `createDMAInForall` to make failure after op erase structurally impossible, but I'd prefer not to do that now to avoid a lot of merge conflicts with other ongoing DMA work. Plus potential performance gain for early reject anyway.)

Yu-Zhewen:
On a separate note, #24114 also fixes numerical errors for two different tests: [check_rocm_hip_stream_dot_general](https://github.com/iree-org/iree/blob/main/tests/e2e/stablehlo_ops/CMakeLists.txt#L486-L508) and [check_rocm-rocm_dot_general](https://github.com/iree-org/iree/blob/main/tests/e2e/stablehlo_ops/CMakeLists.txt#L708-L730). The failures were non-deterministic, so it might be barrier related. Were you expecting that, or should I file an issue?

krzysz00:
(haven't read because OOO but how will we change this for gfx1250?

jerryyin:
Ok I think I've misunderstood this PR with some of the context in #24139.

1. You are right that all `createDMAInForall()` cases that can return a failure must be hoisted to the validation earlier otherwise we'd be in trouble. We should not mutate IR in matchAndRewrite situations and that's a hard rule.

      My concerns are more in the failures in `createDMAInForall()` though: How many of the failures are structural that can't be supported, and how many of them are non-structural - We reject them just because it is more work to implement? If it is the latter, then just rejecting it earlier will harden the original mis-conception and make it hard for us to remember to fix those.

3. Ok got it. But in non-async buffer load situation, we'd issue a conditional check to see if it is actually dword aligned so the dynamic offset can be handled. Refer to https://github.com/llvm/llvm-project/pull/135014 and https://github.com/llvm/llvm-project/pull/135982. So it doesn't sounds like a situation we can't fix. This is an example that "because it is more work to implement and potentially less performant" reject situation that originally don't hold that well.

    This is a class of problems that in my opinion can be addressed if we are willing to inflate the `numRecords` of buffer load descriptor and then use a sentinel we know would OOB.

Overall I'm not disagreeing with you that this fixed real bugs but my concern is more like when the issues are not as glaring as "bugs" we'd have less incentives to fix them - Then it got forgotten what are the legit cases that DMA should support and which aren't. That's what I meant about document and move-on for now then re-visit later. We can merge this PR once failure cases from createDMA are document - If you can attach the particular test can that will introduce those failures, even better because then later it'd be much easier for us to reproduce and judge the legitimacy/immediacy of those cases one by one.

Yu-Zhewen:
@jerryyin Thanks, agreed. I think none of these rejection cases are structural and all should be solvable. I've documented them in #24156.

jerryyin (approved):
LGTM

PR #24140, opened 2026-04-16
Revert "[SPIRV] Fix ub.poison legalization failure for 2d vectors (#23789)"

Yu-Zhewen:
The fix has been upstreamed in LLVM: https://github.com/llvm/llvm-project/pull/192109


amd-eochoalo (approved):
Thanks! I ran the deleted test cases locally just to make sure.

PR #24145, opened 2026-04-16
[CI] Cache bazel builds

amd-eochoalo:
Caches Bazel build in Azure using same container as sccache.

Important to notes:

* bazel shutdown will return an error like the following, so we just avoid running bazel shutdown
```
WARNING: Waiting for server process to terminate (waited 10 seconds, waiting at most 10)
INFO: Waited 10 seconds for server process (pid=1661) to terminate.
FATAL: Attempted to kill stale server process (pid=1661) using SIGKILL, but it did not die in a timely fashion.
Error: Process completed with exit code 36.
```
* we manually compress the cache. this works because we set --experimental_disk_cache_gc_idle_delay to a sufficiently long time which guarantees that the cache is not being garbage collected while we manually compress the cache.
* this only works for PRs from this repo, not forks. 
* See https://github.com/iree-org/iree/actions/runs/24539542215/job/71742199747 for a successful run.


build_tools/bazel/build_test_all.sh
```diff
@@ -172,6 +173,11 @@ if [[ ! -z "${SANDBOX_BASE}" ]]; then
   BAZEL_TEST_CMD+=(--sandbox_base="${SANDBOX_BASE}")
 fi
 
+if [[ -n "${BAZEL_DISK_CACHE_DIR}" ]]; then
+  BAZEL_TEST_CMD+=(--disk_cache="${BAZEL_DISK_CACHE_DIR}")
+  BAZEL_TEST_CMD+=(--experimental_disk_cache_gc_idle_delay=86400s)
```
Muzammiluddin-Syed-ECE: i think you can write this as 
```--experimental_disk_cache_gc_idle_delay=1d```

File: .github/workflows/ci_linux_x64_bazel.yml [hunk 2]
```diff
@@ -22,13 +22,30 @@ jobs:
...
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           submodules: true
-      - name: Install Python requirements
-        run: python3 -m pip install -r ./runtime/bindings/python/iree/runtime/build_requirements.txt
+      - name: Install dependencies
+        run: |
+          apt-get update && apt-get install -y file zip
+          python3 -m pip install -r ./runtime/bindings/python/iree/runtime/build_requirements.txt
+          python3 -m pip install --force-reinstall --no-deps cffi
```
Muzammiluddin-Syed-ECE, line 39: Im curious, whats the reason for the special handling for the package `cffi`?

.github/workflows/ci_linux_x64_bazel.yml
```diff
@@ -37,5 +54,44 @@ jobs:
...
+          # WARNING: Waiting for server process to terminate (waited 10 seconds, waiting at most 10)
+          # INFO: Waited 10 seconds for server process (pid=1661) to terminate.
+          # FATAL: Attempted to kill stale server process (pid=1661) using SIGKILL, but it did not die in a timely fashion.
+          #
+          # Until then, do not shutdown manually.
+          # This could in theory be bad, but it is believed that with a high value for
+          # --experimental_disk_cache_gc_idle_delay the cache is not being compressed / collected
+          # by bazel.
+          #
+          # bazel --noworkspace_rc --bazelrc=build_tools/bazel/iree.bazelrc shutdown
```
Muzammiluddin-Syed-ECE, line 68: What is the motivation for shutting down manually, is there some advantage? According to the [user manual ](https://bazel.build/docs/user-manual#shutdown) there's an automatic shutdown after some amount of idle time.

.github/workflows/ci_linux_x64_bazel.yml
amd-eochoalo: It would bring some extra dependencies. I need cffi for the azure python script that I am running, but I don't want to bring the other dependencies in. (We don't need them).

.github/workflows/ci_linux_x64_bazel.yml
amd-eochoalo: The original motivation behind manual shutdown was that we can guarantee that the cache is not being modified by Bazel itself. However, doing a manual shutdown leads to around 3 minutes or more of just waiting for bazel to shutdown and it always shutdowns with a non-zero error code, which leads to CI failing at the last step.

amd-eochoalo:
@kuhar

.github/workflows/ci_linux_x64_bazel.yml
```diff
@@ -37,5 +54,44 @@ jobs:
...
+            # Delete oldest-accessed files first
+            find "$CACHE_DIR" -type f -printf '%A@ %s %p\n' | \
+              sort -n | \
+              awk -v target="$overshoot" '{
+                sum += $2;
+                print $3;
+                if (sum >= target) exit
+              }' | xargs rm
+          fi
+          tar -czf /tmp/cache.tar.gz -C "${BAZEL_DISK_CACHE_DIR}" .
```
kuhar: This looks like something that should be checked into the repo as a script so that we can test locally; actions are not reproducible otherwise. Maybe make it a part of your script? WDYT?

kuhar (approved):

RattataKing (approved):

PR #24129, opened 2026-04-15
[LinalgExt] Add TilingInterface support for TopkV2Op

bangtianliu:
On top of #24146 

This PR implements `getTiledImplementation `and `getResultTilePosition `for TopkV2Op, replacing the placeholder  
stubs from #24146.     

Assisted-by:  [Claude Code](https://claude.ai/code)

PR #24146, opened 2026-04-17
[LinalgExt] Add scalar implementation for TopkV2Op

bangtianliu:
This PR adds scalar lowering (`generateScalarImplementation`) for TopkV2Op, including bubble sort support for the `is_sorted `flag. This enables convert-to-loops for TopkV2Op. Also, this PR extracts a shared `emitBubbleSortSweep `helper and refactors SortOp to use it, replacing the inline `scf.if `swap logic with `arith.select`.                                                                  
  
`getTiledImplementation `and `getResultTilePosition `return failure() as placeholders; real tiling support is sent in PR #24129

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
@@ -913,12 +913,69 @@ LogicalResult SortOp::getResultTilePosition(
   return success();
 }
 
+/// Emit one bubble-sort sweep: compare adjacent elements along sortDim and
+/// conditionally swap. compareOperands are fed to the comparator region (as
+/// interleaved pairs); carryOperands are swapped in lockstep but not compared.
+/// Convention: comparator returning true means keep current order (no swap).
+static void emitBubbleSortSweep(OpBuilder &b, Location loc, Value zero,
+                                Value one, Value ub, int64_t sortDim,
+                                ValueRange ivs, ValueRange compareOperands,
+                                ValueRange carryOperands,
+                                Region &comparatorRegion) {
+  Block &cmpBlock = comparatorRegion.front();
+  scf::ForOp::create(
+      b, loc, zero, ub, one, /*iterArgs=*/ValueRange{},
+      [&](OpBuilder &b, Location loc, Value j, ValueRange) {
+        // Compare adjacent elements at positions j (lhs) and j+1 (rhs).
+        Value jPlusOne = arith::AddIOp::create(b, loc, j, one);
+        SmallVector<Value> lhsIndices(ivs);
+        lhsIndices[sortDim] = j;
+        SmallVector<Value> rhsIndices(ivs);
+        rhsIndices[sortDim] = jPlusOne;
+
+        SmallVector<Value> cmpArgs;
+        SmallVector<Value> lhsVals, rhsVals;
+        for (Value operand : compareOperands) {
+          Value lhs = memref::LoadOp::create(b, loc, operand, lhsIndices);
+          Value rhs = memref::LoadOp::create(b, loc, operand, rhsIndices);
+          cmpArgs.push_back(lhs);
+          cmpArgs.push_back(rhs);
+          lhsVals.push_back(lhs);
+          rhsVals.push_back(rhs);
+        }
+
+        for (Value operand : carryOperands) {
+          lhsVals.push_back(
+              memref::LoadOp::create(b, loc, operand, lhsIndices));
+          rhsVals.push_back(
+              memref::LoadOp::create(b, loc, operand, rhsIndices));
+        }
+
+        IRMapping sortMap;
+        sortMap.map(cmpBlock.getArguments(), cmpArgs);
+        for (auto &blockOp : cmpBlock.without_terminator()) {
+          b.clone(blockOp, sortMap);
+        }
+        Value keepOrder =
+            sortMap.lookup(cmpBlock.getTerminator()->getOperand(0));
+
+        SmallVector<Value> allOps;
+        allOps.append(compareOperands.begin(), compareOperands.end());
+        allOps.append(carryOperands.begin(), carryOperands.end());
+        for (auto [operand, lhs, rhs] :
+             llvm::zip_equal(allOps, lhsVals, rhsVals)) {
+          Value newLhs = arith::SelectOp::create(b, loc, keepOrder, lhs, rhs);
+          Value newRhs = arith::SelectOp::create(b, loc, keepOrder, rhs, lhs);
+          memref::StoreOp::create(b, loc, newLhs, operand, lhsIndices);
+          memref::StoreOp::create(b, loc, newRhs, operand, rhsIndices);
+        }
+        scf::YieldOp::create(b, loc, /*operands=*/ValueRange{});
+      });
+}
+
 LogicalResult SortOp::generateScalarImplementation(OpBuilder &b, Location loc,
                                                    ValueRange ivs) {
   auto sortDim = getDimension();
-  SmallVector<Value> indices, sortBlkArgs;
-  indices.append(ivs.begin(), ivs.end());
-  // Bubble sort innermost loop.
   Value zero = arith::ConstantIndexOp::create(b, loc, 0);
   Value one = arith::ConstantIndexOp::create(b, loc, 1);
   Value ub;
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -913,12 +913,68 @@ LogicalResult SortOp::getResultTilePosition(
...
+/// Emit one bubble-sort sweep: compare adjacent elements along sortDim and
+/// conditionally swap. compareOperands are fed to the comparator region (as
+/// interleaved pairs); carryOperands are swapped in lockstep but not compared.
+/// Convention: comparator returning true means keep current order (no swap).
+static void emitBubbleSortSweep(OpBuilder &b, Location loc, Value zero,
+                                Value one, Value ub, uint64_t sortDim,
+                                ValueRange ivs, ValueRange compareOperands,
+                                ValueRange carryOperands,
+                                Region &comparatorRegion) {
+  auto &cmpBlock = comparatorRegion.front();
```
hanhanW: Spell out the type

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -913,12 +913,68 @@ LogicalResult SortOp::getResultTilePosition(
   return success();
 }
 
+/// Emit one bubble-sort sweep: compare adjacent elements along sortDim and
+/// conditionally swap. compareOperands are fed to the comparator region (as
+/// interleaved pairs); carryOperands are swapped in lockstep but not compared.
+/// Convention: comparator returning true means keep current order (no swap).
+static void emitBubbleSortSweep(OpBuilder &b, Location loc, Value zero,
+                                Value one, Value ub, uint64_t sortDim,
```
hanhanW: nit: avoid `unsigned` unless you have a reason.

> Unsigned integers are good for representing bitfields and modular arithmetic. Because of historical accident, the C++ standard also uses unsigned integers to represent the size of containers - many members of the standards body believe this to be a mistake, but it is effectively impossible to fix at this point. The fact that unsigned arithmetic doesn't model the behavior of a simple integer, but is instead defined by the standard to model modular arithmetic (wrapping around on overflow/underflow), means that a significant class of bugs cannot be diagnosed by the compiler. In other cases, the defined behavior impedes optimization.

> That said, mixing signedness of integer types is responsible for an equally large class of problems. The best advice we can provide: try to use iterators and containers rather than pointers and sizes, try not to mix signedness, and try to avoid unsigned types (except for representing bitfields or modular arithmetic). Do not use an unsigned type merely to assert that a variable is non-negative.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -913,12 +913,68 @@ LogicalResult SortOp::getResultTilePosition(
...
+        allOps.append(compareOperands.begin(), compareOperands.end());
+        allOps.append(carryOperands.begin(), carryOperands.end());
+        for (auto [operand, vJ, vJ1] : llvm::zip_equal(allOps, valsJ, valsJ1)) {
+          Value newJ = arith::SelectOp::create(b, loc, keepOrder, vJ, vJ1);
+          Value newJ1 = arith::SelectOp::create(b, loc, keepOrder, vJ1, vJ);
+          memref::StoreOp::create(b, loc, newJ, operand, idxJ);
+          memref::StoreOp::create(b, loc, newJ1, operand, idxJ1);
+        }
+        scf::YieldOp::create(b, loc, /*operands=*/ValueRange{});
+      });
```
hanhanW, line 973: `J`, `J1`, `vJ`, `vJ1`, they are all bad names.. I think what we can do is documents a bit and use `lhs` and `rhs`. WDYT?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -913,12 +913,68 @@ LogicalResult SortOp::getResultTilePosition(
...
+        for (Value operand : carryOperands) {
+          valsJ.push_back(memref::LoadOp::create(b, loc, operand, idxJ));
+          valsJ1.push_back(memref::LoadOp::create(b, loc, operand, idxJ1));
+        }
+
+        IRMapping sortMap;
+        for (auto [arg, val] :
+             llvm::zip_equal(cmpBlock.getArguments(), cmpArgs)) {
+          sortMap.map(arg, val);
+        }
```
hanhanW: Can this just be `sortMap.map(cmpBlock.getArguments(), cmpArgs);`?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1593,6 +1598,186 @@ LogicalResult TopkOp::getResultTilePosition(
...
+  auto &srcBlock = getRegion().front();
+  IRMapping bvmF;
+  {
+    OpBuilder::InsertionGuard guard(b);
+    b.setInsertionPointToEnd(&scfFor.getRegion().front());
+    SmallVector<Value> forwardValues{loopCarryValues[0], kValue};
+    for (auto [arg, val] :
+         llvm::zip_equal(srcBlock.getArguments(), forwardValues)) {
+      bvmF.map(arg, val);
+    }
```
hanhanW: ditto, do you need the loop?

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1593,6 +1598,186 @@ LogicalResult TopkOp::getResultTilePosition(
...
+  indices[kDim] = scfFor.getInductionVar();
+  auto loopCarryValues = scfFor.getRegionIterArgs();
+
+  // Clone the comparator region with forward mapping f(x,y).
+  auto &srcBlock = getRegion().front();
+  IRMapping bvmF;
+  {
+    OpBuilder::InsertionGuard guard(b);
+    b.setInsertionPointToEnd(&scfFor.getRegion().front());
+    SmallVector<Value> forwardValues{loopCarryValues[0], kValue};
```
hanhanW: See https://abseil.io/tips/88 and https://llvm.org/docs/CodingStandards.html#do-not-use-braced-initializer-lists-to-call-a-constructor

```suggestion
      SmallVector<Value> reverseValues = {kValue, loopCarryValues[0]};
```

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
```diff
@@ -1593,6 +1598,186 @@ LogicalResult TopkOp::getResultTilePosition(
...
+      Value resultKIndex = arith::SelectOp::create(b, loc, indexCmpRes,
+                                                   loopCarryValues[1], kIndex);
+      memref::StoreOp::create(b, loc, resultKIndex, outputIndices, indices);
+      Value resultCarryIndex = arith::SelectOp::create(
+          b, loc, indexCmpRes, kIndex, loopCarryValues[1]);
+      yieldValues.push_back(resultCarryIndex);
+    }
+
+    scf::YieldOp::create(b, loc, yieldValues);
+  }
```
hanhanW: I'm wondering why this is outside the scf.for creation? Is there any reason?

hanhanW (changes requested):

hanhanW:
Just some nits and questions.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
bangtianliu: TIL

compiler/src/iree/compiler/Dialect/LinalgExt/IR/TilingInterfaceImpl.cpp
bangtianliu: Just followed the style of TopkOp previously, I will apply your suggestion.

hanhanW: I see, thanks!

hanhanW (approved):

PR #24150, opened 2026-04-17
[Tokenizer] Fix infinite loop when encountering a special token in partial segment mode

Finistere:
I've encountered an infinite loop when encoding a rendered template (Gemma4-31B-it). This PR fixes the underlying problem and adds a regression test for it. I've also tested that current code fails.

The problem occurs whenever a special token is encountered while we're still in a partial segment that exceeds the ring buffer size. The old code doesn't reset the `read_position` & `segmenter_view_start` position correctly and thus ends up restarting at the same place again.

The test is a bit more complex than that because it took me a while to figure out that this was effectively the problem. I was using the suggested max ring buffer size of 16KiB, so 8KiB in practice after halving. The failing case was roughly:

```
<|think|>
...system prompt...
...AGENTS.md...
<|tool>...
```

Both the system prompt and the content of the `AGENTS.md` were nothing special and did not even exceed 8 KiB. However, Gemma4, at least the `tokenizer.json` config on HuggingFace at the time of writing, normalizes spaces with "▁" (U+2581), taking 3 bytes instead of 1, and splits on whitespaces only. With this additional pre-processing, there are no segment boundaries anymore between the two special tokens. And with all the additional bytes, my text went from 8093 bytes to 9993, exceeding the ring buffer size.

### DIFF
--- a/runtime/src/iree/tokenizer/tokenizer.c
+++ b/runtime/src/iree/tokenizer/tokenizer.c
@@ -2606,10 +2606,15 @@ static iree_status_t iree_tokenizer_encode_state_pump(
         state->segmenter_view_start = state->read_position;
       }
     }
-    // When a forced-complete drains the ring, exit partial mode so fresh
-    // text after the special token goes through the normal segmenter path.
-    if (force_complete && segment_complete &&
-        state->read_position >= state->write_position) {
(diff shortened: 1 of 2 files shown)


benvanik (approved):
thanks!

PR #24155, opened 2026-04-17
[Codegen] Add transfer_{gather/scatter} unrolling and lowering to CPU backend

NoumanAmir657:


amd-eochoalo (commented):
I am ok with this, but wait until @Groverkss reviews. There's a small bazel error.

hanhanW (commented):
I think we are missing a pipeline test, can you add one to https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_tests.mlir

NoumanAmir657:
> I think we are missing a pipeline test, can you add one to https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_tests.mlir

Hi @hanhanW, currently no direct lowering exists from `linalg_ext.gather` which gathers from the last dimension. This is the vectorization code for `linalg_ext.gather` [here](https://github.com/iree-org/iree/blob/ce2de4628c3ffb543c6e2c74f927ab8e4f07ad89/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L53). 
For `linalg.generic`, there does exists a vectorization [here](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L863), which was added by me but that is guarded by the pass option `vectorize-to-transfer-gather`. So a full pipeline test won't work from linalg level. I can maybe add the `vectorize-to-transfer-gather` as a pipeline option?

hanhanW:
> > I think we are missing a pipeline test, can you add one to https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_tests.mlir
> 
> Hi @hanhanW, currently no direct lowering exists from `linalg_ext.gather` which gathers from the last dimension. This is the vectorization code for `linalg_ext.gather` [here](https://github.com/iree-org/iree/blob/ce2de4628c3ffb543c6e2c74f927ab8e4f07ad89/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L53). For `linalg.generic`, there does exists a vectorization [here](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L863), which was added by me but that is guarded by the pass option `vectorize-to-transfer-gather`. So a full pipeline test won't work from linalg level. I can maybe add the `vectorize-to-transfer-gather` as a pipeline option?

I see. I'm thinking if we should make that option on by default in this case, if the trend is using transfer gather.

https://github.com/iree-org/iree/blob/bbf54e4525322cd09bfa502da20b72ab31d942d1/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L246-L249

What is the current support for the transfer_gather op? Is it correct to say that using transfer_gather ops is always better?

hanhanW:
Maybe @Groverkss has the context

NoumanAmir657:
@hanhanW 
After some investigation on whether we can drop the flag `vectorize-to-transfer-gather`, listing down the findings below:
- Currently, two vectorizations are behind the flag `vectorize-to-transfer-gather`: https://github.com/iree-org/iree/blob/6f099e9abc947108aea841ad0cf4ef1a25c245e0/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L974-L990 For the `vectorizeImplicitGatherToTransferGather` we can drop the flag `vectorize-to-transfer-gather` and the compilation for a `implicitGatherLikeGeneric` should work e2e on the CPU backend.
- For the case of `gatherLikeGeneric` (which operates on having tensor.extract inside the generic body), the e2e compilation of test https://github.com/iree-org/iree/blob/6f099e9abc947108aea841ad0cf4ef1a25c245e0/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_using_transfer_gather.mlir#L234-L260 fails because it ends up making a stack allocation of `memref<8x8xindex>`, and when the pass `LLVMCPUCheckIRBeforeLLVMConversionPass` runs it crashes the compilation because of https://github.com/iree-org/iree/blob/6f099e9abc947108aea841ad0cf4ef1a25c245e0/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUCheckIRBeforeLLVMConversion.cpp#L98, as it can't reason about the bitwidth of `index`. Note that all the other test cases of `gatherLikeGeneric` pass e2e compilation so this above test case is the only one stopping use from dropping the flag.

hanhanW:
> @hanhanW After some investigation on whether we can drop the flag `vectorize-to-transfer-gather`, listing down the findings below:
> 
> * Currently, two vectorizations are behind the flag `vectorize-to-transfer-gather`: https://github.com/iree-org/iree/blob/6f099e9abc947108aea841ad0cf4ef1a25c245e0/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L974-L990
>    For the `vectorizeImplicitGatherToTransferGather` we can drop the flag `vectorize-to-transfer-gather` and the compilation for a `implicitGatherLikeGeneric` should work e2e on the CPU backend.
> * For the case of `gatherLikeGeneric` (which operates on having tensor.extract inside the generic body), the e2e compilation of test https://github.com/iree-org/iree/blob/6f099e9abc947108aea841ad0cf4ef1a25c245e0/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_using_transfer_gather.mlir#L234-L260
>    fails because it ends up making a stack allocation of `memref<8x8xindex>`, and when the pass `LLVMCPUCheckIRBeforeLLVMConversionPass` runs it crashes the compilation because of https://github.com/iree-org/iree/blob/6f099e9abc947108aea841ad0cf4ef1a25c245e0/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUCheckIRBeforeLLVMConversion.cpp#L98
>   , as it can't reason about the bitwidth of `index`. Note that all the other test cases of `gatherLikeGeneric` pass e2e compilation so this above test case is the only one stopping use from dropping the flag.

(I just want to say that this is on my list, and I'm trying something locally. I'll circle back later.)

hanhanW:
I got a patch, and I'll split one PR from it: https://gist.github.com/hanhanW/54157127831bede9b834be4aab5fd9bb

It's built on top of your PR, but the dependency should be flipped. So you'll be able to add pipeline tests. Note that the change in `Passes.cpp` is enabling the option by default except buffer pipeline. We should bailout memref types in this case, and I can send a separate PR for this.

Preparing the PR now.

hanhanW:
> Preparing the PR now.

https://github.com/iree-org/iree/pull/24201

NoumanAmir657:
@hanhanW since #24201 has been merged. Do you want me to enable `vectorize-to-transfer-gather` in CPU pipelines in this PR or a separate one after which I land this?

hanhanW:
> @hanhanW since #24201 has been merged. Do you want me to enable `vectorize-to-transfer-gather` in CPU pipelines in this PR or a separate one after which I land this?

I'm investigating a gather lowering performance, and unfortunately the option is not available for that case. It fails in similar compilation errors.

To unblock your work, can you add a flag (e.g., `iree-llvmcpu-experimental-vectorize-to-transfer-gather`) and only enables it when the flag is on? Then I think you should be able to add a pipeline test and it should unblock your use case.

NoumanAmir657:
@hanhanW I have added the experimental flag and the pipeline tests.

hanhanW (approved):

hanhanW:
Just a note that the lowering and vectorization of `transfer_gather` is not as mature as the upsteam vector code, so it may take sometime to enable the flag. I've been experiencing some limits yesterday, and I'll file some issues for it.

NoumanAmir657:
Failed on pre-commit. Fixed.

NoumanAmir657:
@hanhanW the CI failure seems irrelevant to this PR.

PR #23860, opened 2026-03-19
[Common] Update ComplexChain definition in CombineSourceLayoutTransform

Abhishek-Varma:
With the regression reported in https://github.com/iree-org/iree/pull/23814 we see the following IR snippet :-

```
%78 = scf.for %arg9 = %c0 to %c512 step %c512 iter_args(%arg10 = %arg8) -> (tensor<1x1x1x8x4x16x2x8xf8E4M3FNUZ>) {
  %79 = affine.apply affine_map<(d0, d1) -> (d0 + d1)>(%arg9, %arg4)
  %80:8 = affine.delinearize_index %79 into (1, 1, 1, 8, 4, 16, 1, 1) : index, index, index, index, index, index, index, index
  %extracted_slice_7 = tensor.extract_slice %51[%63, %arg6, %80#3, %80#4, %80#5, 0, 0] [1, 1, 1, 1, 1, 2, 8] [1, 1, 1, 1, 1, 1, 1] : tensor<?x64x8x4x16x2x8xf8E4M3FNUZ> to tensor<1x1x1x1x1x2x8xf8E4M3FNUZ>
  %extracted_slice_8 = tensor.extract_slice %arg10[0, 0, 0, %80#3, %80#4, %80#5, 0, 0] [1, 1, 1, 1, 1, 1, 2, 8] [1, 1, 1, 1, 1, 1, 1, 1] : tensor<1x1x1x8x4x16x2x8xf8E4M3FNUZ> to tensor<1x1x1x1x1x1x2x8xf8E4M3FNUZ>
  %expanded = tensor.expand_shape %extracted_slice_7 [[0, 1], [2], [3], [4], [5], [6], [7]] output_shape [1, 1, 1, 1, 1, 1, 2, 8] : tensor<1x1x1x1x1x2x8xf8E4M3FNUZ> into tensor<1x1x1x1x1x1x2x8xf8E4M3FNUZ>
  %81 = linalg.copy {lowering_config = #iree_gpu.derived_thread_config} ins(%expanded : tensor<1x1x1x1x1x1x2x8xf8E4M3FNUZ>) outs(%extracted_slice_8 : tensor<1x1x1x1x1x1x2x8xf8E4M3FNUZ>) -> tensor<1x1x1x1x1x1x2x8xf8E4M3FNUZ>
[...]

PR #23814, opened 2026-03-17
Revert "[LLVMGPU] Add CombineSourceLayoutTransformation to TileAndFus…

efric:
…e pipeline (#23745)"

This reverts commit 71af3a5e41a8e265330bc693194c708cf6df4724.

Substantial degradation in `torch_models::llama_8b_fp8/prefill_benchmark_seq2048_mi325_data_tiling.json` from 208.7711642185847 <= 187` after the introduction of this pass.

PR #24158, opened 2026-04-17
[LLVMGPU] Add CombineSourceLayoutTransformation to TileAndFuse pipeline

Abhishek-Varma:
-- Re-insert CombineSourceLayoutTransformation in the GPU TileAndFuse pipeline.
-- This is done in order to combine complex relayout op chains from load_from_buffer into a map_load op.
-- It was previously reverted as part of [regression](https://github.com/iree-org/iree/pull/23814) which is now solved by [updating complex chain definition](https://github.com/iree-org/iree/pull/23860).


hanhanW (approved):
Same PR like https://github.com/iree-org/iree/pull/23745, approved.

Issue #24071, opened 2026-04-10
Compiler crash when lowering torch.aten.index_put


sixcy:
### What happened

Lowering torch.aten.index_put leads to a compiler crash

#### Reduced MLIR

```mlir
module {
  func.func @main(%arg0: !torch.vtensor<[1,6],si64>, %arg1: !torch.vtensor<[151936,1024],f32>, %arg2: !torch.vtensor<[1,6],si64>, %arg3: !torch.vtensor<[64],f32>, %arg4: !torch.vtensor<[1024],f32>, %arg5: !torch.vtensor<[1024,1024],f32>, %arg6: !torch.vtensor<[128],f32>, %arg7: !torch.vtensor<[1],si64>) -> !torch.vtensor<[1,8,127,128],f32> {
    %int-1 = torch.constant.int -1
    %false = torch.constant.bool false
    %false_0 = torch.constant.bool false
    %0 = torch.aten.embedding %arg1, %arg0, %int-1, %false, %false_0 : !torch.vtensor<[151936,1024],f32>, !torch.vtensor<[1,6],si64>, !torch.int, !torch.bool, !torch.bool -> !torch.vtensor<[1,6,1024],f32>
    %int0 = torch.constant.int 0
    %1 = torch.aten.unsqueeze %arg3, %int0 : !torch.vtensor<[64],f32>, !torch.int -> !torch.vtensor<[1,64],f32>
    %int2 = torch.constant.int 2
    %2 = torch.aten.unsqueeze %1, %int2 : !torch.vtensor<[1,64],f32>, !torch.int -> !torch.vtensor<[1,64,1],f32>
    %int1 = torch.constant.int 1
    %int-1_1 = torch.constant.int -1
    %int1_2 = torch.constant.int 1
    %3 = torch.prim.ListConstruct %int1, %int-1_1, %int1_2 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %false_3 = torch.constant.bool false
    %4 = torch.aten.expand %2, %3, %false_3 : !torch.vtensor<[1,64,1],f32>, !torch.list<int>, !torch.bool -> !torch.vtensor<[1,64,1],f32>
    %int1_4 = torch.constant.int 1
    %5 = torch.aten.unsqueeze %arg2, %int1_4 : !torch.vtensor<[1,6],si64>, !torch.int -> !torch.vtensor<[1,1,6],si64>
    %int6 = torch.constant.int 6
[...]

PR #24159, opened 2026-04-17
Fix LinalgFusionOpInterface adaptor for linalg ops

IanWood1:
`getIndexingMapsForOperands` is documented as returning one map per operand, but the `linalg` and `linalg.softmax` adapters only returned maps for dps inputs and not dps outputs. This caused `getMatchingIndexingMap` to read OOB on init operands for softmax (the linalg adapter hid the bug by overriding `getMatchingIndexingMap` to forward to `linalg::LinalgOp`'s native one).

Hit this while working on a fix for https://github.com/iree-org/iree/issues/24071


MaheshRavishankar (approved):

MaheshRavishankar:
Do you want to add a test here too?

IanWood1:
> Do you want to add a test here too?

I don't have a good way to test this because the current passes that use the interface can't hit this crash. My fix for https://github.com/iree-org/iree/issues/24071 might make this testable so I can add a test there.

PR #24166, opened 2026-04-20
[DispatchCreation] Fuse scalar reductions with their parallel consumers

### LINKED ISSUE #24148: RMSNorm reduction doesn't fuse with its normalize+scale consumer
This comes up in Qwen MoE models.

Expected: 1 dispatch (fused reduce + normalize + scale)
Actual:   2 dispatches (reduction, normalize+scale)

Repro:
```
// Compile:
//   iree-compile repro.mlir --iree-hal-target-backends=rocm --iree-rocm-target=gfx1100 \
//     --compile-to=flow -o repro_flow.mlir
// Or with O3:
//   iree-compile repro.mlir --iree-hal-target-backends=rocm --iree-rocm-target=gfx1100 \
//     --compile-to=flow -o repro_flow.mlir

module @rmsnorm_fusion_repro {
  func.func @rmsnorm(
      %input: tensor<1x2048xf16>,
      %weight: tensor<2048xf16>
  ) -> tensor<1x2048xf16> {
    %cst_0 = arith.constant 0.000000e+00 : f32
    %cst_eps = arith.constant 1.000000e-06 : f32

    // Step 1: sum of squares (reduction)
    %empty_ss = tensor.empty() : tensor<1xf32>
    %fill_ss = linalg.fill ins(%cst_0 : f32) outs(%empty_ss : tensor<1xf32>) -> tensor<1xf32>
    %sumsq = linalg.generic {
                         affine_map<(d0, d1) -> (d0)>],
        iterator_types = ["parallel", "reduction"]
    } ins(%input : tensor<1x2048xf16>) outs(%fill_ss : tensor<1xf32>) {
    ^bb0(%in: f16, %acc: f32):
      %f = arith.extf %in : f16 to f32
      %sq = arith.mulf %f, %f : f32
      %add = arith.addf %sq, %acc : f32
      linalg.yield %add : f32
    } -> tensor<1xf32>

    // Step 2: normalize + scale (elementwise, consumes reduction scalar)
    %empty_out = tensor.empty() : tensor<1x2048xf16>
    %result = linalg.generic {
                         affine_map<(d0, d1) -> (d0)>,
                         affine_map<(d0, d1) -> (d1)>,
                         affine_map<(d0, d1) -> (d0, d1)>],
        iterator_types = ["parallel", "parallel"]
    } ins(%input, %sumsq, %weight : tensor<1x2048xf16>, tensor<1xf32>, tensor<2048xf16>)
      outs(%empty_out : tensor<1x2048xf16>) {
    ^bb0(%x: f16, %ss: f32, %w: f16, %out: f16):
      %xf = arith.extf %x : f16 to f32
      %n = arith.constant 2048.0 : f32
      %mean = arith.divf %ss, %n : f32
      %rms = arith.addf %mean, %cst_eps : f32
      %rsqrt = math.rsqrt %rms : f32
      %normed = arith.mulf %xf, %rsqrt : f32
      %wf = arith.extf %w : f16 to f32
      %scaled = arith.mulf %normed, %wf : f32
      %result = arith.truncf %scaled : f32 to f16
      linalg.yield %result : f16
    } -> tensor<1x2048xf16>

    return %result : tensor<1x2048xf16>
  }
}
```

Abhishek-Varma:
Patterns of the shape :-
```
  %s = reduce(%x)
  %res = elementwise(%x, %s)
```
were being emitted as two dispatches instead of one as observed [here](https://github.com/iree-org/iree/issues/24148).

The fusion check in `getRootParallelLoopToOpMap` rejected the consumer because the composed root-to-consumer map had all-zero results (e.g. () -> (0, 0) for a consumer broadcasting the reduction's scalar).
That check is meant to catch consumers that don't actually depend on any of the root's parallel loops.

But when the root is a full reduction to a scalar, it has no parallel loops to begin with - so the map comes out all zeros for any consumer, and the check ends up rejecting fusion.

This PR narrows the rejection to fire only when the root has at least one parallel dim, which lets the above pattern fuse into a single dispatch.

Fixes: https://github.com/iree-org/iree/issues/24148 (for `Qwen MoE`)


compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -566,6 +566,39 @@ isFusableWithConsumer(OpOperand &fusedOperand, const FusionTracker &tracker,
     return false;
   }
 
+  // If the root is a full reduction (no parallel loops) and the
+  // consumer has parallel work, only fuse when they share an input
+  // tensor. Without a shared input, fusing has no payoff.
```
Groverkss: I don't think this fix is correct. Even if they don't share an input, we usually fuse them.

Groverkss (changes requested):
@IanWood1 Can you check the part of the fix that I commented on?

Otherwise LGTM, good fix, thanks!

MaheshRavishankar (changes requested):
Dont want to add a extra condition this way. The existing logic probably does not handle this corner case. Maybe we need to take a look at the `getFusionGroup(producer).isFusable(consumer)` method to fix the logic more natively.

Abhishek-Varma:
Hi @Groverkss @MaheshRavishankar - I've updated the logic to be natively within `isFusable`. I understand we fuse even if the inputs aren't shared but in this particular edge since we have a pure scalar producer -> parallel consumer, the shared input case seems like the only valid combination to enable fusion. Else if the inputs weren't shared then in such cases we shouldn't fuse because the parallel work that would have run across many workgroups in its own dispatch would then execute serially inside the reduction's dispatch (as the reduction's workgroup config would be inherited by the parallel consumer).

Please correct me if I'm wrong and I'll adapt accordingly. :)

IanWood1 (commented):
I think the first commit was on the right track (https://github.com/iree-org/iree/pull/24166/changes/c771e1318956022f9acb234be5cd9f65093840e3). This looks like something that was missed from https://github.com/iree-org/iree/pull/23659. If the consumer broadcasts the scalar results, the "composed map" will have 0s in the results but no dims. 

The "composed map" maps the root's parallel loops to the consumers parallel loops. So, this should be changed to only check for dims and not the number of results because we are looking for cases where there are no root parallel loops to map to.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
IanWood1: I agree, this can be removed

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
```diff
@@ -2366,6 +2368,126 @@ util.func public @fuse_scalar_reduction_with_scalar_consumer(
...
+//  CHECK-SAME:         iterator_types = ["parallel"]
+//       CHECK:     %[[REDUCE:.+]] = linalg.generic
+//  CHECK-SAME:         iterator_types = ["reduction"]
+//  CHECK-SAME:         ins(%[[ELT]],
+//       CHECK:     flow.return %[[REDUCE]]
+//       CHECK:   util.return %[[DISPATCH]]
+
+// -----
+
+// Check that a parallel elementwise producer still fuses with a full-reduction
```
IanWood1: Do we still need the "no shared input" tests?

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
```diff
@@ -2366,6 +2368,126 @@ util.func public @fuse_scalar_reduction_with_scalar_consumer(
 
 // -----
 
+// Check that a full reduction producing a scalar fuses with a parallel
+// elementwise consumer when they share an input tensor.
+util.func public @fuse_scalar_reduction_with_parallel_consumer_shared_input(
```
IanWood1, line 2373: Doesn't this test the same thing as `fuse_no_shared_parallel_loops` above?

IanWood1 (approved):
LGTM!

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
Abhishek-Varma: No, `fuse_no_shared_parallel_loops` exercises no shared input case while `fuse_scalar_reduction_with_parallel_consumer_shared_input` is the shared input case.

I checked by reverting this patch's logic and saw that we will need to keep this lit test as it won't pass.

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
Abhishek-Varma: Removed. Good catch!

Abhishek-Varma:
Hi @Groverkss @MaheshRavishankar - I've updated the PR description and addressed all review comments. Can you please take a look now?

Groverkss (approved):

MaheshRavishankar (approved):

PR #24167, opened 2026-04-20
Enable building without google benchmark dependency 

AaronStGeorge:
When building in TheRock we need to [pull the third party google benchmark dependency](https://github.com/ROCm/TheRock/blob/497af57abf59b11237b55c4ebd553b89d00d488b/build_tools/fetch_sources.py#L661) even though the build in TheRock builds [without tests](https://github.com/ROCm/TheRock/blob/497af57abf59b11237b55c4ebd553b89d00d488b/iree-libs/CMakeLists.txt#L19-L38) and does not require `iree-benchmark-module`. 

This PR adds an option to build IREE without benchmark dependency, allowing us to remove the dep from TheRock. Removing the dep limits operational risk and allows us to sidestep a configure failure coming from benchmark CMake setup in ASan builds (a compiler invocation inside a capability check fails to find the ASan runtime library).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,7 @@ option(IREE_ENABLE_RUNTIME_TRACING "Enables instrumented runtime tracing." OFF)
 option(IREE_ENABLE_COMPILER_TRACING "Enables instrumented compiler tracing." OFF)
 option(IREE_ENABLE_RENDERDOC_PROFILING "Enables profiling HAL devices with the RenderDoc tool." OFF)
 option(IREE_ENABLE_THREADING "Builds IREE in with thread library support." ON)
+option(IREE_ENABLE_THIRD_PARTY_BENCHMARK "Enables building and linking the bundled google benchmark dependency." ON)
 option(IREE_SYNCHRONIZATION_DISABLE_UNSAFE "Disables all synchronization primitives (for single-threaded bare-metal targets only)." OFF)
 option(IREE_ENABLE_CLANG_TIDY "Builds IREE in with clang tidy enabled on IREE's libraries." OFF)
 
@@ -84,6 +85,15 @@ option(IREE_BUILD_TRACY "Enables building the 'iree-tracy-capture' CLI tool and 
 option(IREE_BUILD_BUNDLED_LLVM "Builds the bundled llvm-project (vs using installed)" ON)
 option(IREE_BUILD_CLANG_TOOLS_EXTRA "Enables building the 'clang-tools-extra' LLVM targets." OFF)
 option(IREE_USE_SYSTEM_DEPS "Uses certain system libraries instead of building downloaded dependencies" OFF)
+
+if(NOT IREE_ENABLE_THIRD_PARTY_BENCHMARK)
+  if(IREE_BUILD_TESTS)
+    message(FATAL_ERROR "IREE_ENABLE_THIRD_PARTY_BENCHMARK=OFF requires IREE_BUILD_TESTS=OFF.")
+  endif()
+  if(IREE_BUILD_PYTHON_BINDINGS)
+    message(FATAL_ERROR "IREE_ENABLE_THIRD_PARTY_BENCHMARK=OFF requires IREE_BUILD_PYTHON_BINDINGS=OFF.")
+  endif()
+endif()
 
 if(IREE_USE_SYSTEM_DEPS)
   message(WARNING "IREE_USE_SYSTEM_DEPS is on. This option is untested and may find incompatible dependencies.")
@@ -982,7 +992,7 @@ if(IREE_BUILD_TESTS)
   add_subdirectory(third_party/googletest EXCLUDE_FROM_ALL)
 endif()
 
-if(IREE_ENABLE_THREADING)
+if(IREE_ENABLE_THREADING AND IREE_ENABLE_THIRD_PARTY_BENCHMARK)
   # Benchmark.
   iree_set_benchmark_cmake_options()
   add_subdirectory(third_party/benchmark EXCLUDE_FROM_ALL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,11 @@ option(IREE_BUILD_BUNDLED_LLVM "Builds the bundled llvm-project (vs using instal
 option(IREE_BUILD_CLANG_TOOLS_EXTRA "Enables building the 'clang-tools-extra' LLVM targets." OFF)
 option(IREE_USE_SYSTEM_DEPS "Uses certain system libraries instead of building downloaded dependencies" OFF)
 
+# IREE_BUILD_BENCHMARKS is required when IREE_BUILD_TESTS=ON because test
+# targets transitively depend on the bundled google benchmark target.
+cmake_dependent_option(IREE_BUILD_BENCHMARKS "Builds IREE benchmarks." ON
+    "NOT IREE_BUILD_TESTS" ON)
+
 if(IREE_USE_SYSTEM_DEPS)
   message(WARNING "IREE_USE_SYSTEM_DEPS is on. This option is untested and may find incompatible dependencies.")
 endif()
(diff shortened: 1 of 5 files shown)


runtime/src/iree/testing/CMakeLists.txt
ScottTodd, line 36: You could instead mark these `TESTONLY`, as is already done for the libraries below.

I guess the question is "should it be possible to build benchmarks but _not_ tests?" If someone wants to build/distribute the benchmark tools without also building tests then yes. Otherwise, no.

runtime/bindings/python/CMakeLists.txt
```diff
@@ -152,6 +160,7 @@ iree_symlink_tool(
   TO_EXE_NAME iree/_runtime_libs/iree-c-embed-data
 )
 
+if(IREE_BUILD_BENCHMARKS)
 iree_symlink_tool(
```
benvanik: indent

runtime/bindings/python/CMakeLists.txt
```diff
@@ -405,9 +415,6 @@ cmake_language(DEFER DIRECTORY \"${IREE_SOURCE_DIR}\"
   CALL install
   TARGETS
     iree-cpuinfo
-    iree-benchmark-executable
```
benvanik, line 408: they still need to be installed when enabled

tools/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ if(NOT IREE_ENABLE_THREADING)
   return()
 endif()
 
+if(IREE_BUILD_BENCHMARKS)
 iree_cc_binary(
```
benvanik: indent (here and everywhere else)

benvanik (changes requested):

CMakeLists.txt
```diff
@@ -87,6 +87,11 @@ option(IREE_BUILD_BUNDLED_LLVM "Builds the bundled llvm-project (vs using instal
 option(IREE_BUILD_CLANG_TOOLS_EXTRA "Enables building the 'clang-tools-extra' LLVM targets." OFF)
 option(IREE_USE_SYSTEM_DEPS "Uses certain system libraries instead of building downloaded dependencies" OFF)
 
+# IREE_BUILD_BENCHMARKS is required when IREE_BUILD_TESTS=ON because test
+# targets transitively depend on the bundled google benchmark target.
+cmake_dependent_option(IREE_BUILD_BENCHMARKS "Builds IREE benchmarks." ON
+    "NOT IREE_BUILD_TESTS" ON)
```
ScottTodd: I'm confused here. Is this doing what you want?

* If `IREE_BUILD_TESTS` is `OFF`, then `IREE_BUILD_BENCHMARKS` will be forced off by `cmake_dependent_option`
* If `IREE_BUILD_TESTS` is `ON`, then `IREE_BUILD_BENCHMARKS` is available to set to either `ON` or `OFF`, but the comment says that it must always be `ON`?

Am I reading this right?

runtime/bindings/python/CMakeLists.txt
AaronStGeorge: They should be included in `_EXTRA_INSTALL_TOOL_TARGETS` when `IREE_BUILD_BENCHMARKS` is enabled that's included below                                                                         
  https://github.com/iree-org/iree/blob/835d1b88ab3f3a7cdd0fc952c28337143602c2d8/runtime/bindings/python/CMakeLists.txt#L418-L424

it's the same pattern as tracy

https://github.com/iree-org/iree/blob/835d1b88ab3f3a7cdd0fc952c28337143602c2d8/runtime/bindings/python/CMakeLists.txt#L10-L14

CMakeLists.txt
AaronStGeorge: ### From [cmake docs](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html): 
`cmake_dependent_option(<variable> <help> <value> <condition> <else-value>)`

- If `<condition>` evaluates to boolean true, a boolean cache variable named `<variable>` is created with default `<value>`, and option is shown in the GUI, allowing the user to enable or disable it.
- If `<condition>` evaluates to boolean false, option is hidden from the user in the GUI, and a local variable `<variable>` is set to `<else-value>`.

### Our case:

`cmake_dependent_option(IREE_BUILD_BENCHMARKS "Builds IREE benchmarks." ON "NOT IREE_BUILD_TESTS" ON)`
- if `IREE_BUILD_TESTS=OFF` `NOT IREE_BUILD_TESTS` is `TRUE` and `IREE_BUILD_BENCHMARKS` becomes an optional with the default value of `ON`. Defaulting to `ON` seems reasonable to me but `OFF` would also be fine.
- if `IREE_BUILD_TESTS=ON`  `NOT IREE_BUILD_TESTS` is `FLASE` and `IREE_BUILD_BENCHMARKS` is forced `ON`. Several test targets depend on the benchmarks library, so that's what we want in this case.

I'll make the comment clearer, I was pretty convinced Claude got this wrong when going over it again 😆

runtime/src/iree/testing/CMakeLists.txt
AaronStGeorge: > should it be possible to build benchmarks but not tests?

That's a good question. If `IREE_BUILD_TESTS=OFF` implied that benchmark module was not needed that would solve fusilli's current problem.

🤷‍♂️ my instinct would be to allow building benchmarks without tests but happy to go with whatever the consensuses opinion here is.

CMakeLists.txt
benvanik: yeah, this is odd. we ship iree-benchmark-module/etc that this PR is trying to disable and should be able to build those binaries independent from tests. tests != benchmarks.

AaronStGeorge: That's what's happening IIUC. When `IREE_BUILD_TESTS=ON`, `IREE_BUILD_BENCHMARKS` is forced `ON` but when `IREE_BUILD_TESTS=OFF` it `IREE_BUILD_BENCHMARKS` is configurable. Tests imply benchmarks because of the library dependency but test != benchmarks.

runtime/src/iree/testing/CMakeLists.txt
AaronStGeorge: https://github.com/iree-org/iree/pull/24167#discussion_r3164854195 sounds like we want to keep them independent.

runtime/bindings/python/CMakeLists.txt
AaronStGeorge: done

tools/CMakeLists.txt
AaronStGeorge: done

benvanik:
(I'll defer to @ScottTodd here as he'll probably be the one most bothered by what the end state is in TheRock :)

ScottTodd (approved):
> When building in TheRock we need to [pull the third party google benchmark dependency](https://github.com/ROCm/TheRock/blob/497af57abf59b11237b55c4ebd553b89d00d488b/build_tools/fetch_sources.py#L661) even though the build in TheRock builds [without tests](https://github.com/ROCm/TheRock/blob/497af57abf59b11237b55c4ebd553b89d00d488b/iree-libs/CMakeLists.txt#L19-L38) and does not require `iree-benchmark-module`.

Requirements may change there over time if we want to also run IREE tests against the latest ROCm code in TheRock / rocm-libraries / etc. If the issue is IREE using its own version of the dependency, we could find the package that the superproject (TheRock) already includes, maybe by using `IREE_USE_SYSTEM_DEPS` with a branch like what is done here: https://github.com/iree-org/iree/blob/2ba8b6f581c0f5eb9eef0f34c0aa1fc96794ae5a/CMakeLists.txt#L872-L881
applied here also? https://github.com/iree-org/iree/blob/2ba8b6f581c0f5eb9eef0f34c0aa1fc96794ae5a/CMakeLists.txt#L987-L990

AaronStGeorge:
>  If the issue is IREE using its own version of the dependency, we could find the package that the superproject (TheRock) already includes, maybe by using IREE_USE_SYSTEM_DEPS

👍 that's also what we did for the hip headers in https://github.com/iree-org/iree/pull/22715

PR #24068, opened 2026-04-10
[DispatchCreation][LinalgExt] Add OnlineAttentionOp support in dispatch formation and reshape fusion

keshavvinayak01:
This PR teaches dispatch creation and the LinalgExt reshape-fusion infrastructure about OnlineAttentionOp, which is missing from the paths currently specialized only for `AttentionOp`. Torch input conversion was split out to #24177, and rewrite porting  was split to #24123. Codegen support landed in #24110.

PR #24169, opened 2026-04-20
[DispatchCreation] Allow fusion of multi-result producers

keshavvinayak01:
Enables consumer fusion for multi-result producers like `iree_linalg_ext.online_attention` whose results flow into a single consumer via operands with different ranks (e.g. acc and sum of the normalization `linalg.generic`).

Supports https://github.com/iree-org/iree/pull/24068


keshavvinayak01 (commented):
> One concern that I have is that I don't think we will be currently able to fuse `iree_linalg_ext.online_attention` -> `linalg.generic` (1/sum + x) into a single dispatch. This will end up regressing a lot of our benchmarks.

@IanWood1

This PR relaxes that constraint, so online_attention should function like attention and allows the above fusion. At least for all the existing cases. 

Does this change make sense?

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -393,10 +393,20 @@ static bool isUnpackLikeOp(Operation *op) {
 static SmallVector<OpOperand *>
 getFusableUses(MLIRContext *context, Operation *op,
                DominanceInfo const &dominanceInfo, bool aggressiveFusion) {
-  if (!aggressiveFusion && llvm::count_if(op->getUses(), [](OpOperand &use) {
```
IanWood1, line 396: I think you can just change the original from `op->getUses()` to `op->getUsers()`

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
```diff
@@ -393,10 +393,20 @@ static bool isUnpackLikeOp(Operation *op) {
 static SmallVector<OpOperand *>
 getFusableUses(MLIRContext *context, Operation *op,
```
IanWood1, line 394: This may be an underlying bug but if this returns 2 OpOperands of the same user, we should only fuse with the user if both OpOperands are deemed fusible. `multi_result_conflict_no_fusion` passes due to the maps but their might be other reasons to not fuse with one operand.


Maybe this can just be made into an issue?

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
```diff
@@ -2756,3 +2757,100 @@ util.func public @scatter_chain_dispatch_ordering(
...
+//            DEFAULT:   util.return %[[D]]
+
+// -----
+
+// Multi-result producer whose outputs use differently-permuted indexing maps.
+// The two composed iteration-space maps land concrete (non-broadcast)
+// expressions at the same positions, so mergeComposedMaps must reject the
+// merge and producer fusion must not happen.
+
+util.func public @multi_result_conflict_no_fusion(%x: tensor<16x16xf32>) -> tensor<16x16xf32> {
```
IanWood1, line 2828: It might be worth adding a test this one since this will now fuse with "DEFAULT" pass flags:


```
util.func public @single_result_two_uses_same_consumer_default(
    %a : tensor<?x?xf32>, %b : tensor<?x?xf32>) -> tensor<?x?xf32> {
  %cst = arith.constant 0.0 : f32
  %d0 = tensor.dim %b, %c0 : tensor<?x?xf32>
  %d1 = tensor.dim %b, %c1 : tensor<?x?xf32>
  %empty = tensor.empty(%d0, %d1) : tensor<?x?xf32>
  %fill = linalg.fill ins(%cst : f32) outs(%empty : tensor<?x?xf32>) -> tensor<?x?xf32>
  %m = linalg.matmul ins(%a, %b : tensor<?x?xf32>, tensor<?x?xf32>)
         outs(%fill : tensor<?x?xf32>) -> tensor<?x?xf32>
  %r = linalg.generic {
                       affine_map<(d0, d1) -> (d0, d1)>,
                       affine_map<(d0, d1) -> (d0, d1)>],
      iterator_types = ["parallel", "parallel"]}
      ins(%m, %m : tensor<?x?xf32>, tensor<?x?xf32>)
      outs(%empty : tensor<?x?xf32>) {
    ^bb0(%b0 : f32, %b1 : f32, %out : f32) :
      %s = arith.addf %b0, %b1 : f32
      linalg.yield %s : f32
  } -> tensor<?x?xf32>
  util.return %r : tensor<?x?xf32>
}
```

compiler/src/iree/compiler/DispatchCreation/test/form_dispatch_regions.mlir
```diff
@@ -2756,3 +2757,100 @@ util.func public @scatter_chain_dispatch_ordering(
...
+    iree_linalg_ext.yield %score : f32
+  } -> tensor<20x4096x64xf32>, tensor<20x4096xf32>, tensor<20x4096xf32>
+  %norm = linalg.generic {
+    indexing_maps = [
+      affine_map<(b, m, n) -> (b, m, n)>,
+      affine_map<(b, m, n) -> (b, m)>,
+      affine_map<(b, m, n) -> (b, m, n)>
+    ],
+    iterator_types = ["parallel", "parallel", "parallel"]
+  } ins(%r#0, %r#2 : tensor<20x4096x64xf32>, tensor<20x4096xf32>)
```
IanWood1, line 2802: I think these changes will also allow fusion in a simlar case to this but where `%r#0` and `%r#2` aren't from the same op but instead are from two different ops that are both in the "fusion group". Maybe add a test for this?

IanWood1 (commented):
Looks good to me. I can't spot any issues with this change. However, this seems like a change that may have unintended side effects. You should probably add "ci-extra: test_torch" to the PR description to run the torch tests to check for regressions.

I'm holding off on approving since I want to give this another look once the extra tests are passing.

hanhanW:
> Looks good to me. I can't spot any issues with this change. However, this seems like a change that may have unintended side effects. You should probably add "ci-extra: test_torch" to the PR description to run the torch tests to check for regressions.
> 
> I'm holding off on approving since I want to give this another look once the extra tests are passing.

FYI, test_torch is default on in presubmit since https://github.com/iree-org/iree/commit/65f8952861e72ca833fef28fbbca06fe47d1439d

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
keshavvinayak01: Yes, I'm already using `getUsers()` directly.

compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
keshavvinayak01: Filed https://github.com/iree-org/iree/issues/24206

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/LoopMappingUtils.cpp
```diff
@@ -48,6 +48,34 @@ llvm::SmallBitVector getOuterParallelLoops(Operation *op) {
...
+    } else if (isZero(ea)) {
+      merged.push_back(eb);
+    } else {
+      return failure();
+    }
+  }
+  return AffineMap::get(a.getNumDims(), a.getNumSymbols(), merged,
+                        a.getContext());
+}
+
```
Groverkss: nit: can you use more descriptive names for a, b, ea, eb?

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/LoopMappingUtils.cpp
```diff
@@ -48,6 +48,34 @@ llvm::SmallBitVector getOuterParallelLoops(Operation *op) {
   return llvm::SmallBitVector{};
 }
 
+/// Combines two composed iteration-space maps result-wise. A constant `0`
+/// is a broadcast placeholder and yields to a concrete expression; two
+/// different concrete expressions at the same position are a conflict. A
+/// null `a` acts as the identity.
```
Groverkss: Can you add an example?

Groverkss (approved):
LGTM, but wait for @IanWood1 to stamp please

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/LoopMappingUtils.cpp
```diff
@@ -48,6 +48,46 @@ llvm::SmallBitVector getOuterParallelLoops(Operation *op) {
   return llvm::SmallBitVector{};
 }
 
+/// Combines two composed iteration-space maps result-wise. A constant `0`
+/// is a broadcast placeholder and yields to a concrete expression; two
+/// different concrete expressions at the same position are a conflict. A
+/// null `accumulated` acts as the identity.
```
IanWood1, line 54: Side note: `0` here is a bit more general than a broadcast placeholder: it means the loop has no mapping to the root op's outer parallel loop space. I wondered whether `mergeComposedMaps` could therefore allow fusion we do not want, but I reasoned through the cases I was concerned about and tried a few repros without finding bad behavior so I think this is fine.

IanWood1 (approved):
LGTM!

PR #24174, opened 2026-04-20
Split ContractToChainFMA

amd-eochoalo:
This PR contributes towards #23094 . We split up the pattern ContractToChainFMA into:

* optimization / transformation: transpose
* flattening
* unrolling vector.contract (next commit)
* lowering to FMA

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUVectorLowering.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUVectorLowering.cpp
@@ -442,19 +442,15 @@ private:
   }
 };
 
-// Rewrites vector.contracts that are already in normalized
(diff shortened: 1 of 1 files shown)


amd-eochoalo:
@Groverkss would it also make sense to split this pattern into flattening / unrolling / lowering?

amd-eochoalo:
@efric if this is very important for performance and for some reason this refactoring affects the performance, we can keep this pattern as is. Thanks for giving some context in today's meeting.

efric:
> @efric if this is very important for performance and for some reason this refactoring affects the performance, we can keep this pattern as is. Thanks for giving some context in today's meeting.

Thanks for letting me know and performing this refactor! Since this is still a draft, I looked only at the high level but the directions looks makes sense to me. Also performance shouldn't be affected as long as semantics are preserved.

amd-eochoalo:
I am removing this in favour of #24211 and #24212

Issue #24122, opened 2026-04-15
[LinalgExt] Deprecating LinalgExt::AttentionOp


keshavvinayak01:
## Description

We use `OnlineAttention` everywhere, `AttentionOp` is just used for dispatch creation, we can deprecate and use just `OnlineAttention` directly.

## Progress:
1. E2E compilation support for online_attention:
    - https://github.com/iree-org/iree/pull/24110
    - https://github.com/iree-org/iree/pull/24123
    - https://github.com/iree-org/iree/pull/24068
         - https://github.com/iree-org/iree/pull/24169
         - https://github.com/iree-org/iree/pull/24177

Once all of the above are in, a PR to replace all IR usages of `iree_linalg_ext.attention` <-> `iree_linalg_ext.online_attention` and subsequently all defintions/support for `iree_linalg_ext.attention` removed from the codeabase.

PR #24177, opened 2026-04-21
[TMTensor][LinalgExt] Lower TMTensor::AttentionOp directly to OnlineAttentionOp

keshavvinayak01:
Converts the torch input path to emit `iree_linalg_ext.online_attention` + a normalization `linalg.generic` (acc * 1/sum) instead of the legacy `iree_linalg_ext.attention`.

Supports https://github.com/iree-org/iree/issues/24122


ziereis:
I have some concerns regarding this PR, and especially regarding the entire plan of deprecating the `linalg_ext.attention` op.

Generally I would like to keep the `linalg_ext.attention` op and handle the conversion from attention to online_attention similarly to how we deal with im2col for example, because it is very much hardware- and codegen-pipeline-dependent whether keeping the attention op and decomposing it before dispatch creation, or converting it to online_attention, is the right call.

As a real-world example, we currently decompose the attention op as part of a preprocessing pipeline for all "cpu-like" targets because running it decomposed is just a lot faster for most cpus. Additionally, I'm not certain that this would be a net gain for other targets like SPIR-V either (does it even support this op at all??).

Unless we are sure that online_attention is always better than attention, I would prefer an alternative flow where we keep the `linalg_ext.attention` op at import and add some pass (maybe in global opt?) that converts it to online_attention at this point. This way other users still have the possibility of decomposing the `attention` op as part of a preprocessing step.

Groverkss:
> I have some concerns regarding this PR, and especially regarding the entire plan of deprecating the `linalg_ext.attention` op.
> 
> Generally I would like to keep the `linalg_ext.attention` op and handle the conversion from attention to online_attention similarly to how we deal with im2col for example, because it is very much hardware- and codegen-pipeline-dependent whether keeping the attention op and decomposing it before dispatch creation, or converting it to online_attention, is the right call.
> 
> As a real-world example, we currently decompose the attention op as part of a preprocessing pipeline for all "cpu-like" targets because running it decomposed is just a lot faster for most cpus. Additionally, I'm not certain that this would be a net gain for other targets like SPIR-V either (does it even support this op at all??).
> 
> Unless we are sure that online_attention is always better than attention, I would prefer an alternative flow where we keep the `linalg_ext.attention` op at import and add some pass (maybe in global opt?) that converts it to online_attention at this point. This way other users still have the possibility of decomposing the `attention` op as part of a preprocessing step.

If I was doing this on CPU, I would do the decomposition as:

kernel 1: QK matmul
kernel 2: elementwise + softmax-pv-matmul

you can represent the softmax-pv-matmul as exp_reduction

I agree it's not useful to do flash attention always (example: cpu), but I'm not too sure if not using the softmax-pv-matmul combination is a bad idea. This op usually comes from torch.sdpa ops which are almost always used to represent flash attention style compute, but it's ok to do it in 2 kernels for things like cpu where you have a smaller vector width for the first matmul reduction.

ziereis:
@Groverkss i think this sounds super interesting, however i still think that we should keep around the possibility to decompose to a regular matmul(QK) + softmax + matmul(PV) decomposition until the proposed change lands and we are sure that performance is actually on par this way. 

Currently what you can do is run something like:

```
--iree-preprocessing-pass-pipeline="builtin.module(util.func(iree-linalg-ext-decompose-aggregated-ops{filter-ops=iree_linalg_ext.attention}))"
```

With this change we would lose this and be forced to run online_attention, this is what i want to avoid.

PR #24178, opened 2026-04-21
[LinalgExt] Fix attention NaN for fully-masked rows

### LINKED ISSUE #24175: [SDPA] `iree_linalg_ext.attention` produces NaN for fully masked inputs, instead of zeros
Attention is outputting NaNs instead of zeroes for fully masked-out input on bf16.

`sdpa_mask_nan.mlir`:
```mlir
module @module {
  func.func @main(%attn_mask_4: !torch.vtensor<[1,1,1,1],bf16>, %k_1: !torch.vtensor<[1,1,1,2],bf16>, %q_0: !torch.vtensor<[1,1,1,2],bf16>, %v_2: !torch.vtensor<[1,1,1,2],bf16>) -> !torch.vtensor<[1,1,1,2],bf16> attributes {torch.assume_strict_symbolic_shapes} {
    %permute_Q_val_0_sdpa_fprop = torch.constant.int 0
    %permute_Q_val_1_sdpa_fprop = torch.constant.int 1
    %permute_Q_val_2_sdpa_fprop = torch.constant.int 2
    %permute_Q_val_3_sdpa_fprop = torch.constant.int 3
    %permute_Q_sdpa_fprop = torch.prim.ListConstruct %permute_Q_val_0_sdpa_fprop, %permute_Q_val_1_sdpa_fprop, %permute_Q_val_2_sdpa_fprop, %permute_Q_val_3_sdpa_fprop : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %q_0_sdpa_fprop_perm = torch.aten.permute %q_0, %permute_Q_sdpa_fprop : !torch.vtensor<[1,1,1,2],bf16>, !torch.list<int> -> !torch.vtensor<[1,1,1,2],bf16>

    %permute_K_val_0_sdpa_fprop = torch.constant.int 0
    %permute_K_val_1_sdpa_fprop = torch.constant.int 1
    %permute_K_val_2_sdpa_fprop = torch.constant.int 2
    %permute_K_val_3_sdpa_fprop = torch.constant.int 3
    %permute_K_sdpa_fprop = torch.prim.ListConstruct %permute_K_val_0_sdpa_fprop, %permute_K_val_1_sdpa_fprop, %permute_K_val_2_sdpa_fprop, %permute_K_val_3_sdpa_fprop : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %k_1_sdpa_fprop_perm = torch.aten.permute %k_1, %permute_K_sdpa_fprop : !torch.vtensor<[1,1,1,2],bf16>, !torch.list<int> -> !torch.vtensor<[1,1,1,2],bf16>

    %permute_V_val_0_sdpa_fprop = torch.constant.int 0
    %permute_V_val_1_sdpa_fprop = torch.constant.int 1
    %permute_V_val_2_sdpa_fprop = torch.constant.int 2
    %permute_V_val_3_sdpa_fprop = torch.constant.int 3
    %permute_V_sdpa_fprop = torch.prim.ListConstruct %permute_V_val_0_sdpa_fprop, %permute_V_val_1_sdpa_fprop, %permute_V_val_2_sdpa_fprop, %permute_V_val_3_sdpa_fprop : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %v_2_sdpa_fprop_perm = torch.aten.permute %v_2, %permute_V_sdpa_fprop : !torch.vtensor<[1,1,1,2],bf16>, !torch.list<int> -> !torch.vtensor<[1,1,1,2],bf16>

    %permute_mask_val_0_sdpa_fprop = torch.constant.int 0
    %permute_mask_val_1_sdpa_fprop = torch.constant.int 1
    %permute_mask_val_2_sdpa_fprop = torch.constant.int 2
    %permute_mask_val_3_sdpa_fprop = torch.constant.int 3
    %permute_mask_sdpa_fprop = torch.prim.ListConstruct %permute_mask_val_0_sdpa_fprop, %permute_mask_val_1_sdpa_fprop, %permute_mask_val_2_sdpa_fprop, %permute_mask_val_3_sdpa_fprop : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %attn_mask_4_sdpa_fprop_perm = torch.aten.permute %attn_mask_4, %permute_mask_sdpa_fprop : !torch.vtensor<[1,1,1,1],bf16>, !torch.list<int> -> !torch.vtensor<[1,1,1,1],bf16>

    %dropout_sdpa_fprop = torch.constant.float 0.000000e+00
    %is_causal_sdpa_fprop = torch.constant.bool false
    %scale_sdpa_fprop = torch.constant.float 7.071068e-01
    %enable_gqa_sdpa_fprop = torch.constant.bool false
    %o_3_sdpa_fprop_perm = torch.aten.scaled_dot_product_attention %q_0_sdpa_fprop_perm, %k_1_sdpa_fprop_perm, %v_2_sdpa_fprop_perm, %attn_mask_4_sdpa_fprop_perm, %dropout_sdpa_fprop, %is_causal_sdpa_fprop, %scale_sdpa_fprop, %enable_gqa_sdpa_fprop : !torch.vtensor<[1,1,1,2],bf16>, !torch.vtensor<[1,1,1,2],bf16>, !torch.vtensor<[1,1,1,2],bf16>, !torch.vtensor<[1,1,1,1],bf16>, !torch.float, !torch.bool, !torch.float, !torch.bool -> !torch.vtensor<[1,1,1,2],bf16>

    %permute_O_val_0_sdpa_fprop = torch.constant.int 0
    %permute_O_val_1_sdpa_fprop = torch.constant.int 1
    %permute_O_val_2_sdpa_fprop = torch.constant.int 2
    %permute_O_val_3_sdpa_fprop = torch.constant.int 3
    %permute_O_sdpa_fprop = torch.prim.ListConstruct %permute_O_val_0_sdpa_fprop, %permute_O_val_1_sdpa_fprop, %permute_O_val_2_sdpa_fprop, %permute_O_val_3_sdpa_fprop : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %o_3 = torch.aten.permute %o_3_sdpa_fprop_perm, %permute_O_sdpa_fprop : !torch.vtensor<[1,1,1,2],bf16>, !torch.list<int> -> !torch.vtensor<[1,1,1,2],bf16>

    return %o_3 : !torch.vtensor<[1,1,1,2],bf16>
  }
}
```

```bash
iree-compile sdpa_mask_nan.mlir \
  --iree-hal-target-backends=rocm --iree-rocm-target=mi300x \
  --iree-opt-level=O3 \
  -o sdpa_mask_nan.vmfb

iree-run-module --module=sdpa_mask_nan.vmfb --device=hip --function=main \
  --input="1x1x1x1xbf16=-Inf"    `# attn_mask (fully masked)` \
  --input="1x1x1x2xbf16=1,2"     `# k` \
  --input="1x1x1x2xbf16=3,4"     `# q` \
  --input="1x1x1x2xbf16=5,6"     `# v`
```

Output:
```
result[0]: hal.buffer_view
1x1x1x2xbf16=[[[NAN NAN]]]
```

Expected: `1x1x1x2xbf16=[[[0 0]]]`

Source pytorch graph this was generated from, which is minimized from this test: https://github.com/pytorch/pytorch/blob/8cb62a8b4caf7a473870064c05d5de5ae511a2c5/test/test_transformers.py#L3148
```python
import torch

q = torch.randn(1, 1, 1, 2, dtype=torch.bfloat16, device="cuda")
k = torch.randn(1, 1, 1, 2, dtype=torch.bfloat16, device="cuda")
v = torch.randn(1, 1, 1, 2, dtype=torch.bfloat16, device="cuda")
mask = torch.zeros(1, 1, dtype=torch.bool, device="cuda")  # all-False = all positions masked

out = torch.nn.functional.scaled_dot_product_attention(q, k, v, attn_mask=mask)
# out = [[[[nan, nan]]]]; expected zeros
```

## Environment

- `iree-base-compiler==3.12.0rc20260417`

keshavvinayak01:
Attention softmax normalization produced `NaN` for fully-masked rows.

  The failing case is:
  - masked scores are all `-inf`
  - softmax numerator becomes `0`
  - softmax denominator becomes `0`
  - final normalization computes `0 / 0`

PyTorch SDPA uses `_safe_softmax`, which explicitly zeroes fully-masked rows, so IREE should produce `0` here instead of `NaN`.

This PR handles that in both attention lowering paths:

  - Standalone `iree_linalg_ext.attention` decomposition clamps the row softmax denominator with `max(sum, 1)` before `P / sum`.
  - Online attention finalization keeps the existing unmasked `(1 / sum) * x` IR unchanged.
  - Masked online attention guards the existing finalization loop so `sum == 0` yields `0` instead of `NaN`, avoiding an extra row-level pass.

  For non-fully-masked rows, the softmax denominator is unchanged: after max subtraction, at least one term is `exp(0) = 1`, so `sum >= 1`.

  Fixes #24175.


keshavvinayak01 (commented):
One thing I want to note is that this is a stop-gap solution. Since we're moving to remove AttentionOp soon, this handling will also change.

Groverkss (changes requested):
This is wrong. I don't see any reason 0/0 shouldn't be NaN.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/decompose_aggregate_op.mlir
```diff
@@ -129,10 +129,13 @@ func.func @attention_f16(%query: tensor<192x1024x64xf16>,
 // CHECK-NOT: arith.extf
 // CHECK:   arith.addf
 // CHECK:   linalg.yield
-// P = P /= sum
+// P = P /= sum, guarded so fully-masked rows (sum == 0) yield 0 instead of
+// 0/0 == NaN (SDPA semantics).
```
bjacob, line 133: Can you elaborate on what "SDPA semantics" is referring to here? Is there some kind of specification that says that 0 should be produced instead of NaN in this case?  Otherwise, by default, like @Groverkss says, NaN is the reasonable result to expect out of an expression that boils down to 0./0. .

keshavvinayak01: According to the issue's error message, yes, 0 was expected. "SDPA Semantics" just refers to Pytorch's expectation.

keshavvinayak01:
> This is wrong. I don't see any reason 0/0 shouldn't be NaN.

From @rkayaith's update on the issue,  [Pytorch's attn](https://github.com/pytorch/pytorch/blob/7231f9e7a302e0368eee7adf8dcbcd6fd79fe2be/aten/src/ATen/native/transformers/attention.cpp#L901) calls `_safe_softmax` and not the regular softmax, so it expects 0 and not NaN. 

`_safe_softmax` is `softmax(x)` followed by `where(rows_are_all_neginf, 0, softmax_out)` [here](https://github.com/pytorch/pytorch/blob/7231f9e7a302e0368eee7adf8dcbcd6fd79fe2be/aten/src/ATen/native/transformers/attention.cpp#L677-L687).


We will need this fix if we are to adhere to it.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/decompose_aggregate_op.mlir
Groverkss: @bjacob I think this PR could be okay, but I could use your opinion also. We are essentially emulating "masked softmax" here, where some elements of the input vector don't contribute to the softmax. Now, if the entire thing is masked out, we need to define some semantics of what masked means because if nothing is contributing to softmax then what should be the output?

For attention, the generally accepted output is 0.

The ideal way to do this is to check before computing softmax if the row is fully masked out, and replace it with 0 instead, but we cannot do this in the hot loop of attention. Attention generally looks like:

```
loop:
  low_precision_matmul
  f32_online_softmax // maintains the max, sum of softmax, doesn't divide it yet
  low_precision_matmul
f32_division // divide the max, sum of softmax
```

anything you add between those matmuls kills performance significantly, because you are increasing the instruction length between the two matmuls which have special hardware instructions to target. (We faced this problem when trying to do clamps for fp8, it really causes bottlenecks).

The best way to do this would be to mask the softmax in the hot loop, where you check if everything is masked out before doing softmax, and do a select on the output, but that is very expensive. Instead, an okish thing that works for attention is to check if the sum is 0, and replace the div with 0 if so. If the sum is 0, we can mostly assume that some masking happened. Preserving NaN output for attention doesn't seem very useful anyway. Another possible way is we add a very small constant to the sum (1e-20 or something).

What do you think?

Groverkss (commented):
This may be okay as a pragmatic solution, but it needs proper documentation on why we do this. This needs more documentation, check my comment for what kind of documentation it needs.

compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/decompose_aggregate_op.mlir
```diff
@@ -129,9 +130,9 @@ func.func @attention_f16(%query: tensor<192x1024x64xf16>,
 // CHECK-NOT: arith.extf
 // CHECK:   arith.addf
 // CHECK:   linalg.yield
-// P = P /= sum
+// P = P / sum
 // CHECK: linalg.generic
-// CHECK-NOT: arith.extf
```
Groverkss, line 134: why remove the extf check?

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/TileAttention.cpp
```diff
@@ -117,9 +117,11 @@ void convertToOnlineAttention(IREE::LinalgExt::AttentionOp attnOp,
 
   Value x = onlineAttn.getResult(0);
   Value sum = onlineAttn.getResult(2);
+  bool hasMask = static_cast<bool>(mask);
 
-  // Merge the outputs of online attention:
-  //  x = (1 / sum) * x
+  // Finalize online attention. With a mask, fully-masked rows can have
+  // `sum == 0` and `x == 0`; guard that case to produce 0 instead of NaN. Keep
+  // this in the existing finalization loop to avoid an extra row pass.
```
Groverkss: Can you write the equation here?

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -77,6 +77,29 @@ OpFoldResult mulAddOfrs(OpBuilder &builder, Location loc, OpFoldResult a,
                                                {a, b, c});
 }
 
+Value createSafeSoftmaxDenominator(OpBuilder &builder, Location loc,
```
Groverkss: I don't think this belongs to Utils, move it to AggregatedOpInterfaceImpl please

Groverkss (approved):
LGTM, I'm also going to ping @bjacob to have a look

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -77,6 +77,29 @@ OpFoldResult mulAddOfrs(OpBuilder &builder, Location loc, OpFoldResult a,
...
+      builder.getMultiDimIdentityMap(sumType.getRank()),
+      builder.getMultiDimIdentityMap(sumType.getRank())};
+  SmallVector<utils::IteratorType> iteratorTypes(sumType.getRank(),
+                                                 utils::IteratorType::parallel);
+  auto genericOp = linalg::GenericOp::create(
+      builder, loc, output.getType(), sum, output, maps, iteratorTypes,
+      [&](OpBuilder &b, Location loc, ValueRange args) {
+        Value one =
+            arith::ConstantOp::create(b, loc, b.getFloatAttr(elementType, 1.0));
+        Value result = arith::MaximumFOp::create(b, loc, args[0], one);
```
bjacob: I can see that this code is consistent with this part of the PR description:

> Standalone iree_linalg_ext.attention decomposition clamps the row softmax denominator with max(sum, 1) before P / sum.

Please explain why replacing `sum` by `max(sum, 1)` is correct. This seems to be encoding the assumption that the denominator should never be less than 1.  I can see how that is actually the case in a normalized softmax quotient, where the denominator is by construction of the form

```
      1.0 + exp(-x[0]) + ... + exp(-x[n-1])
```

after normalization by rescaling by the largest exponential term. What I don't understand is how you resolve the tension between:
- The denominator can't be < 1 by construction; and
- This code is guarding against cases where the denominator was < 1.

If there is a good explanation for why this code is correct, it should be recorded as a comment here.

bjacob: Re-reading the PR description, I can see that it's about a concept of "masked rows" that I'm not familiar with. But the code here is not about masked rows, it is about the denominator computation; it is trying to filter out the bad case not by considering masked rows directly, but by tweaking the value of `sum`, implying that a too-low `sum` value is used as a proxy for a row being masked.  This interaction-at-a-distance between masked rows and `sum` values needs to be explained.

rkayaith: For reference, this is the behaviour we need to match:
> https://github.com/pytorch/pytorch/blob/7231f9e7a302e0368eee7adf8dcbcd6fd79fe2be/aten/src/ATen/native/transformers/attention.cpp#L677-L687
> ```cpp
> Tensor _safe_softmax(
>     const Tensor& self,
>     int64_t dim,
>     std::optional<ScalarType> dtype) {
>   auto out = at::softmax(self, dim, dtype);
>   const auto masked = self.isneginf();
>   const auto masked_rows = all(masked, dim, true);
>   const auto zero = at::scalar_tensor(0.0, at::TensorOptions().dtype(out.dtype()).device(out.device()));
>   // reuse storage for out
>   return at::where_out(out, masked_rows, zero, out);
> }
> ```

bjacob: I'm confused then. If I read correctly that reference _safe_softmax code, it is *defining* a masked row as being one where all input values are -inf , which is equivalent to saying that `sum == 1` , but then the present rewrite of the denominator from `sum` to `max(sum, 1)` should make no difference?

One problem though is that when all the inputs are -inf, the normalization computes -inf-(-inf) == NaN.

I'm not saying that that is what is happening here (the PR desc says it's something else) but maybe this shows that "masked rows" need to be handled more explicitly, not just as a corrective term on the denominator. Like in the _safe_softmax code you pasted here. Can't we just do that here?

keshavvinayak01: @bjacob 
The `sum == 1` conclusion would be true only for the usual finite-row stable-softmax reasoning, where there is a finite max element and at least one term contributes `exp(0) == 1`.

That is not what happens in this IREE decomposition for a fully masked row. The row max reduction is initialized with a finite minimum value (`useOnlyFiniteValue=true`), not `-inf`. After the mask is applied, a fully masked row has all scores at `-inf`, so the row max remains that finite minimum. Then each softmax numerator is:
```
exp2(-inf - finite_min) == 0
```
and the row sum is therefore `0`, not `1`.

For any non-fully-masked row, there is at least one finite score. That finite score becomes the row max, so at least one numerator is `exp2(0) == 1`, and the row sum is `>= 1`.

So `max(sum, 1)` is not trying to repair arbitrary small positive denominators. Under this decomposition’s invariants it is equivalent to `select(sum == 0, 1, sum)`: it only changes the fully masked row case, making the final `P / sum` produce zero instead of `0 / 0 == NaN`. This matches the PyTorch `_safe_softmax` behavior, but uses the already-computed row sum as the predicate instead of doing a separate all-`-inf` row scan.

I'm doing it this way to emit cheaper IR and not have multiple ops.

It's basically the difference between doing 1-1 with _safe_softmax:
```
S += mask
masked = isneginf(S)
masked_rows = rowAll(masked)

max = rowMax(S)
P = exp2(S - max)
sum = rowSum(P)
P = P / sum
P = select(masked_rows, 0, P)
```

and this
```
S += mask
max = rowMax(S)
P = exp2(S - max)
sum = rowSum(P)
P = P / max(sum, 1)
```
Capturing the same behavior.
I can document this better in the code but I think we should stick with this.

bjacob (approved):
Thanks for the explanation!

PR #24183, opened 2026-04-21
Refactor vector.multi_reduction into flattening, unrolling, and lowering passes.

amd-eochoalo:
* Adds an initial vector flattening pass.
* At the moment, the only operation that is flattened is vector.multi_reduction, but others will be added later.
* Adds vector unrolling for vector.multi_reduction.
* Adds a vector.multi_reduction lowering pass


compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -710,6 +710,63 @@ struct ConvertVectorBitcast final
...
+///   %red_1 = vector.multi_reduction <mul>, %v_0, %a_1 [0] : vector<4xf32> into
+///   f32 %res_tmp = vector.insert %red_1, %res [0] : f32 into vector<2xf32>
+///
+///   // 2nd reduction
+///   %v_1 = vector.extract %src[1] : vector<4xf32> from vector<2x4xf32>
+///   %a_1 = vector.extract %acc[1] : f32 from vector<2xf32>
+///   %red_2 = vector.multi_reduction <mul>, %v_1, %a_1 [0] : vector<4xf32> into
+///   f32 %res_final = vector.insert %red_2, %res_tmp [1] : f32 into
+///   vector<2xf32>
+struct ConvertVectorMultiReduction final
```
efric, line 735: Ok, I am interested in some more context here. Why are we re-implementing this instead of using the upstream pattern? I feel like the upstream pattern is already pretty atomic? What do we gain from this 2 phase thing (unroll vector.multi_reduction, multi_reduction -> vector.reduction)

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
```diff
@@ -710,6 +710,63 @@ struct ConvertVectorBitcast final
...
+    : public OpConversionPattern<vector::MultiDimReductionOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(vector::MultiDimReductionOp op, OneToNOpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    VectorType srcType = cast<VectorType>(op.getSource().getType());
+    if (srcType.getRank() <= 1) {
```
efric:
```suggestion
    if (srcType.getRank() != 2) {
```

you should just check if the rank is 2 directly, otherwise couldn't you have something rank 3+ match as long as you have `[p, r]` in the first 2 dims?

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUVectorMultiReductionLowering.cpp
```diff
@@ -0,0 +1,82 @@
...
+///
+/// Example:
+/// ```mlir
+/// // Before
+/// %r = vector.multi_reduction <add>, %v, %acc [0] : vector<Nxf32> to f32
+///
+/// // After
+/// %r = vector.reduction <add>, %v, %acc : vector<Nxf32> into f32
+/// ```
+struct OneDimMultiReductionToReduction
```
efric, line 33: Is there a test for this?

efric (changes requested):

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPULegalizeNDVectors.cpp
amd-eochoalo: This pattern is slightly different from the one upstream.
* This one uses OneToNOpAdaptor while the one upstream does not.

The main benefit from this separation is that we can better control the lowering of the vector dialect. Right now, flattening, unrolling, and lowering are all folded into a single pass upstream (with several patterns combining unrolling, and lowering into one, for example:

* TwoDimMultiReductionToElementWise (unroll + lowering)
* TwoDimMultiReductionToReduction (unroll + lowering)

By having each pattern separate, we can change the unrolling or the lowering (and flattening) patterns more easily.

It is ok if we decide not to do this for vector.multi_reduction, but if we do, then we must guarantee that all multi-dimensional vector.multi_reduction operations are lowered before entering the 1-D vector legalization pass. Otherwise the pass would fail.

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUVectorMultiReductionLowering.cpp
amd-eochoalo: I should have added that this pattern is taken from upstream https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Vector/Transforms/LowerVectorMultiReduction.cpp#L456 , but yes it should also be tested here if it is copied. I originally didn't want to use the populate method because https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Vector/Transforms/LowerVectorMultiReduction.cpp#L535 it includes other patterns, but now that I think about it, those patterns won't match anything since only 1-D multi reductions will be available. Thanks! I'll use the populate method instead.

amd-eochoalo: I just chatted with Kunwar and he mentioned that it is best to keep 1-D lowering separate. Later, we will have a single 1-D lowering pass for other vector operations.

efric (approved):
Ok LGTM, thanks. Do we plan to upstream some of this N:1 direction or just keep it in iree?

amd-eochoalo:
> Ok LGTM, thanks. Do we plan to upstream some of this N:1 direction or just keep it in iree?

I think we will be upstreaming it at some future point, but right now this is a bit of a big shift in how the vector dialect is lowered in LLVM. I think keeping it here, is good for the time being.

PR #24184, opened 2026-04-21
[ci] Disable sharktank tests

yash-amd:
Disabling sharktank tests

### DIFF
--- a/.github/workflows/pkgci.yml
+++ b/.github/workflows/pkgci.yml
@@ -103,11 +103,11 @@ jobs:
     if: contains(fromJson(needs.setup.outputs.enabled-jobs), 'test_onnx')
     uses: ./.github/workflows/pkgci_test_onnx.yml
 
-  test_sharktank:
-    name: Test Sharktank
-    needs: [setup, build_packages]
-    if: contains(fromJson(needs.setup.outputs.enabled-jobs), 'test_sharktank')
-    uses: ./.github/workflows/pkgci_test_sharktank.yml
+  # test_sharktank:
+  #   name: Test Sharktank
+  #   needs: [setup, build_packages]
+  #   if: contains(fromJson(needs.setup.outputs.enabled-jobs), 'test_sharktank')
+  #   uses: ./.github/workflows/pkgci_test_sharktank.yml
 
   test_torch:
     name: Test Torch


Groverkss:
We should either delete them completely or move them to post submit / nightly. We shouldn't comment them out.

amd-eochoalo (approved):
They all have been moved to torch tests, correct? @yash-amd

yash-amd:
> We should either delete them completely or move them to post submit / nightly. We shouldn't comment them out.

should we remove them?
cc @MaheshRavishankar

hanhanW:
Can you provide more context in PR description?

Are multi device tests ported to torch yet? I think we have a test the exercise 2 devices for running llama, IIRC.

```
iree-test-suites/sharktank_models/llama3.1/test_llama.py::test_decode[hip_tp2] ERROR [ 87%]
iree-test-suites/sharktank_models/llama3.1/test_llama.py::test_prefill_decode[hip_tp2] ERROR [100%]
```

MaheshRavishankar:
Actually the multi-device stuff is interesting. @rsuderman  you have toy_llama that runs on multi-device right? If that is still running outside of sharkmodel tests, I would rather delete these older tests.

amd-eochoalo:
I am a bit interested making sure that this PR (or some version of it) makes it through review as it would allow CI to be less congested. I'm happy to take over it (or merge it and work on the following PRs).

amd-eochoalo:
@Groverkss I can delete these in a following PR. @MaheshRavishankar, @rsuderman I don't know much about the toy llama multi device stuff, but we can always delete them from sharktank PR and revert them later. If I'm given a bit more information  / direction I can delete it from this other location.

hanhanW:
I think people should walk through if all the sharktank config and torch_model config are aligned before deleting it. E.g.,

https://github.com/iree-org/iree/blob/main/tests/external/iree-test-suites/sharktank_models/benchmarks/sdxl/clip_cpu.json
https://github.com/iree-org/iree/blob/main/tests/external/iree-test-suites/torch_models/sdxl/clip_compstat_cpu.json

One sets golden values for binary size at 600K while the other is 650K. We might not care much about this benchmark, but making all the metrics align prevent issues to me; it should not hurt at all, if we claim that all the tests are ported to torch.

> we can always delete them from sharktank PR and revert them later.

I think this is not true because we'd be lacking the test coverage; they may fail silently.

hanhanW:
Someone should pick it up. The example was just a random sample that I saw in the other day.

amd-eochoalo:
@hanhanW good point. Ok, I'll pick it up either tomorrow or next week.

hanhanW (changes requested):
I filed an issue (with codex) for the workstream: https://github.com/iree-org/iree/issues/24482

Assigning it to @amd-eochoalo for now, since he volunteers for the work. Some models are not ported yet, and some configs mismatch. We should either port them or revisit what we want to track.

Aslo, putting a blocker for this PR. It is not ready to disable the sharktank test yet.

Issue #24186, opened 2026-04-21
Parallel implementation of associative scan


sommerlukas:
IREE currently lowers (associative) scans (`linalg_ext.scan`) to a serial implementation, which computes correct results, but is inefficient for parallel hardware such as GPUs.

Goal of this issue is to introduce a proper parallel implementation of associative scan operations in GPU compilation pipelines, in particular the VectorDistribute pipeline. 

To make use of the parallelism in the GPU hardware, the goal is to distribute the computation of scan operations across threads and subgroups within a workgroup.

sommerlukas:
The overall plan for parallel implementation in the VectorDistribute pipeline is to have the frontend represent (associative) scans with the `linalg_ext.scan` operation. During generic vectorization, this operation is lowered to `vector.scan`, which then gets distributed across threads and subgroups. To represent cross-thread, intra-subgroup scans, a new `iree_gpu.subgroup_scan` operation is introduced, which later gets expanded to cross-thread operations such as shuffle. 

The implementation steps are:
- [x] Vectorize `linalg_ext.scan` to `vector.scan`: #24187 
- [x] Attach layouts to `linalg_ext.scan` during tensor layout configuration: #24205 
- [x] Propagate layouts for `vector.scan` in layout analysis: #24189 
- [x] Define `subgroup_scan` and expansion to shuffles: #24188 
- [x] Distribution of `vector.scan`. For ease of review, I've split this into two stages:
  - [x] Distribute across threads within subgroup: https://github.com/iree-org/iree/pull/24312
  - [x] Distribute across subgroups in a workgroup: https://github.com/iree-org/iree/pull/24475
- [x] E2E operation tests: https://github.com/iree-org/iree/pull/24475
- [ ] Supporrt lowering configuration selection for `linalg_ext.scan`: TODO
- [ ] Support shuffling for larger types (similar to https://github.com/iree-org/iree/pull/24432): TODO

PR #24187, opened 2026-04-21
[Codegen] Vectorize `linalg_ext.scan` to `vector.scan`

sommerlukas:
During `GenericVectorization`, vectorize `linalg_ext.scan` to `vector.scan`. If masking is required, it is introduced directly as masked `transfer_read/write` and an `arith.select` selecting between the actual data for unmasked elements and the identity for the combining operation in the scan (e.g., zero for add) for masked elements.

`linalg_ext.scan` expresses the combiner as a region, where as `vector.scan` uses a fixed set of combiners as enum attribute. Therefore, we try to match the content of the region of the `linalg_ext.scan` operation against the set of supported combiners.


compiler/src/iree/compiler/Codegen/Common/test/vectorize_scan.mlir
hanhanW, line 1: Why do we have a new file? I categorize them a few, and I wonder if we move the tests into one of them? Each file tests with different configuration. E.g., _unmasked tests are mainly for cases that input_vector_size is empty. masked_configured is mainly for cases that have input_vector_size where it is derived from configurations, etc.

Do we move them to https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir

hanhanW (commented):
Just a comment about test location. I'll leave the implemetation review to others.

compiler/src/iree/compiler/Codegen/Common/test/vectorize_scan.mlir
sommerlukas: No strong reason, I've moved the tests from the new file to `generic_vectorization_unmasked.mlir`.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -1069,4 +1069,70 @@ bool isaTransposeOpInterface(linalg::LinalgOp linalgOp) {
...
+    return std::nullopt;
+  }
+
+  Value arg0 = block.getArgument(0);
+  Value arg1 = block.getArgument(1);
+  Value opArg0 = firstOp.getOperand(0);
+  Value opArg1 = firstOp.getOperand(1);
+
+  bool validArgs =
+      (opArg0 == arg0 && opArg1 == arg1) || (opArg0 == arg1 && opArg1 == arg0);
```
bangtianliu, line 1108: Assume commutative operations?

What if someone later adds a non-commutative operation to the switch?

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h
```diff
@@ -9,6 +9,7 @@
 
 #include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
 #include "mlir/Dialect/Utils/ReshapeOpsUtils.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
```
bangtianliu, line 12: ` Utils.h` is included by many files across the LinalgExt dialect. Pulling in VectorDialect everywhere increases 
 compile times and creates architectural coupling.

Consider moving VectorOps include to .cpp only, or relocate `matchScanCombiner `

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
```diff
@@ -56,3 +56,86 @@ func.func @single_static_pack_infer_vector_size(%arg0: tensor<101x201xi8>, %arg1
...
+// CHECK: %[[INIT:.+]] = vector.transfer_read
+// CHECK: %[[DEST:.+]], %{{.+}} = vector.scan <add>, %[[READ]], %[[INIT]]
+// CHECK-SAME: inclusive = true
+// CHECK: vector.transfer_write %[[DEST]]
+// CHECK: vector.transfer_write
+
+// -----
+
+// CHECK-LABEL: func.func @vectorize_scan_mul_exclusive
+func.func @vectorize_scan_mul_exclusive(
```
bangtianliu, line 87: Should the following three tests also verify the full pipeline (read->scan->write) like the above one?

bangtianliu (changes requested):
Test for unsupported combining operation

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
sommerlukas: IMHO, this is a case of keeping the test minimal as requested by the MLIR testing guidelines:
> Tests should be minimal, and only check what is absolutely necessary.

The vectorization logic is the same for all four tests, the only significant thing that changes is the combining kind and, in the 2D case, the reduction dimension. 

I'd therefore keep them as is, but if you feel strongly about this, I'm happy to change it.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
sommerlukas: Removed the assumption.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h
sommerlukas: I've relocated the helper to `VectorizableOpInterface`, which is currently the only use. If we need it in more places in the future, we can find an appropriate place for the helper that avoids the coupling.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1342,6 +1400,140 @@ struct Im2colOpVectorizationModel
...
+    SmallVector<int64_t> accumVecShape = inputVecShape;
+    accumVecShape.erase(accumVecShape.begin() + scanDim);
+
+    auto inputVecTy = VectorType::get(inputVecShape, elemType);
+    auto accumVecTy = VectorType::get(accumVecShape, elemType);
+
+    // Determine if masking is needed (dynamic shapes or vector > tensor).
+    bool needsInputMasking =
+        !inputTy.hasStaticShape() ||
+        (llvm::to_vector(inputTy.getShape()) != inputVecShape);
```
bangtianliu: The llvm::to_vector creates a temporary SmallVector just for comparison. Consider using llvm::equal or direct ArrayRef comparison.

``` C++
!llvm::equal(inputTy.getShape(), inputVecShape);
```

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1342,6 +1400,140 @@ struct Im2colOpVectorizationModel
...
+    bool needsAccumMasking =
+        !accumTy.hasStaticShape() ||
+        (llvm::to_vector(accumTy.getShape()) != accumVecShape);
```
bangtianliu: Also here

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -50,6 +51,63 @@ static bool getBoolOption(DictionaryAttr options, StringRef name,
   return defaultValue;
 }
 
+static std::optional<vector::CombiningKind> matchScanCombiner(Region &region) {
```
bangtianliu, line 54: Could we use mlir::matchReduction instead? Just one Q here

bangtianliu (approved):

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
sommerlukas: I've looked into that; we'd need to at least wrap `matchReduction` to get the combiner kind and also check additional invariants, so the benefit seems small.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -1463,12 +1463,10 @@ struct ScanOpVectorizationModel
...
-    bool needsInputMasking =
-        !inputTy.hasStaticShape() ||
-        (llvm::to_vector(inputTy.getShape()) != inputVecShape);
-    bool needsAccumMasking =
-        !accumTy.hasStaticShape() ||
-        (llvm::to_vector(accumTy.getShape()) != accumVecShape);
+    bool needsInputMasking = !inputTy.hasStaticShape() ||
+                             !llvm::equal(inputTy.getShape(), inputVecShape);
+    bool needsAccumMasking = !accumTy.hasStaticShape() ||
+                             !llvm::equal(accumTy.getShape(), accumVecShape);
```
hanhanW, line 1469: drive by: use `hasDyanmicShape` to avoid `!`

sommerlukas: Either I'm overlooking something or `ShapedType` doesn't have that method. It only has a static method `isDynamicShape`, did you mean that?

hanhanW: ohhh... I think I mis-read it... I read is as `isStaticShape`. Sorry about the noise

PR #24188, opened 2026-04-21
[IREEGPU] Define and expand `subgroup_scan`

sommerlukas:
The upstream GPU dialect currently does not expose a `subgroup_scan` operation as analog to `subgroup_scan`.

Having the `subgroup_scan` operation allows us to lower associative scans across the threads in incremental steps rather than doing distribution and full lowering to cross-subgroup operations in a single step. 

The associative scan semantics of the operation is similar to `vector.scan` and supports inclusive and exclusive scan. The subgroup semantics is similar to `subgroup_scan` where applicable.

The expansion of the operation into lower-level vector operations and cross-subgroup operations (e.g., shuffle) uses the [Hillis-Steele algorithm](https://en.wikipedia.org/wiki/Prefix_sum#Algorithm_1:_Shorter_span,_more_parallel).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/ExpandGPUOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/ExpandGPUOps.cpp
@@ -44,7 +44,8 @@ static Value inlineCombiner(RewriterBase &rewriter, Location loc,
 
 /// Lower `iree_gpu.subgroup_scan` to a Hillis-Steele scan using
 /// `gpu.shuffle idx`.
-/// First computes an inclusive scan, then derives the exclusive result by
+/// Computes an inclusive scan via Hillis-Steele. For inclusive mode, the
+/// result is used directly. For exclusive mode, derives the result by
 /// shifting right by one cluster position and inserting the identity at
 /// position 0. Also computes the cluster total by shuffling from the last
 /// lane in each cluster.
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/ExpandGPUOps.cpp
```diff
@@ -51,6 +244,7 @@ struct ExpandGPUOpsPass final : impl::ExpandGPUOpsPassBase<ExpandGPUOpsPass> {
         patterns, /* maxShuffleBitwidth=*/32, PatternBenefit(3));
     populateGpuLowerClusteredSubgroupReduceToShufflePatterns(
         patterns, *subgroupSize, /* shuffleBitwidth=*/32, PatternBenefit(1));
+    patterns.add<LowerSubgroupScan>(ctx, *subgroupSize, PatternBenefit(1));
```
Groverkss: Shouldn't it take a shuffle bitwidth as well?

compiler/src/iree/compiler/Codegen/Common/GPU/ExpandGPUOps.cpp
```diff
@@ -23,8 +27,197 @@ namespace mlir::iree_compiler {
...
+  LogicalResult matchAndRewrite(IREE::GPU::SubgroupScanOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value val = op.getValue();
+    uint32_t clusterSize = op.getClusterSize().value_or(subgroupSize);
+    uint32_t clusterStride = op.getClusterStride();
+    Type valTy = val.getType();
+    unsigned bitwidth = valTy.getIntOrFloatBitWidth();
+
+    constexpr unsigned shuffleBitwidth = 32;
```
Groverkss: I don't think  it's good to assume this.

Groverkss (approved):
LGTM

PR #24189, opened 2026-04-21
[VectorDistribute] Support `vector.scan` in vector layout analysis

sommerlukas:
Add support for `vector.scan` in the `VectorLayoutAnalysis` to propagate layouts across `vector.scan` and later distribute associative scans to parallel implementations.


Groverkss (approved):
nice!

PR #24190, opened 2026-04-21
[Codegen] Fix iree-compile --debug crash on CPU/GPU codegen pass options

### LINKED ISSUE #24074: `iree-compile` crashes with `--debug` flag
### What happened?

Running `iree-compile` on a basic input like `samples/models/simple_abs.mlir` with the LLVMCPU backend crashes when adding the `--debug` flag.

1. From the IREE repository root, run `iree-compile --iree-hal-target-device=local --iree-hal-local-target-device-backends=llvm-cpu --iree-llvmcpu-target-cpu=host samples/models/simple_abs.mlir --debug`

Contents of `samples/models/simple_abs.mlir` :

```mlir
func.func @abs(%input : tensor<f32>) -> (tensor<f32>) {
  %result = math.absf %input : tensor<f32>
  return %result : tensor<f32>
}
```

2. You should see the following backtrace (elided) :

```
...
[pass-manager Pass.cpp:492 2] Initializing pass manager 'func.func' (generation 1) with 2 passes
[pass-manager Pass.cpp:501 2] Initializing pass 'LLVMCPULowerExecutableTargetPass'
[pass-manager Pass.cpp:501 2] Initializing pass 'VerifyWorkgroupDistributionPass'
[pass-manager Pass.cpp:522 1] Pass manager initialization completed successfully: func.func(iree-llvmcpu-lower-executable-target{cpu-options=unknown data value for option
UNREACHABLE executed at iree/third_party/llvm-project/mlir/include/mlir/Pass/PassOptions.h:165!
```

Compiler

### Version information

Release v3.11.0

### Additional context

The crash happens when the pass manager tries to print a `CPUCodegenOptions` as a pass option of the `LLVMCPULowerExecutableTarget` pass.

hanhanW:
Using CPUCodegenOptions / GPUCodegenOptions as MLIR pass options crashes with `--debug` (or any pass-pipeline print path): the default `llvm::cl::parser<T>` inherits `generic_parser_base`, so MLIR's OptionParser alias resolves to `GenericOptionParser<T>`, whose findArgStrForValue hits `llvm_unreachable("unknown data value for option")` on struct-typed values ([mlir/include/mlir/Pass/PassOptions.h](https://github.com/llvm/llvm-project/blob/99457c368586b1debf49f55b3a0684317f5f298d/mlir/include/mlir/Pass/PassOptions.h#L158-L170)).

Fix by specializing `llvm::cl::parser<T>` to inherit `basic_parser<T>` instead, and providing `operator<<` in the type's associated namespace (`mlir::iree_compiler`) so ADL can find it. This routes printing through `printOptionValue / has_stream_operator`, which accepts our streaming that prints `opaque`. The parse path is also a no-op — these options never flow in from pass pipeline strings.

Fixes https://github.com/iree-org/iree/issues/24074

Assisted-by: Claude Code

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
@@ -9,6 +9,24 @@
 IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::CPUCodegenOptions);
 IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::GPUCodegenOptions);
 
+// Defines the out-of-line parser methods declared by
+// IREE_DECLARE_CODEGEN_OPTIONS_PASS_OPTION. Paired with the declaration macro
+// in CodegenOptions.h.
+#define IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION(TYPE)                          \
+  namespace llvm::cl {                                                         \
+  template class basic_parser<TYPE>;                                           \
+  bool parser<TYPE>::parse(Option &, StringRef, StringRef, TYPE &) {           \
+    return false;                                                              \
+  }                                                                            \
+  void parser<TYPE>::printOptionDiff(const Option &, TYPE, const OptVal &,     \
+                                     size_t) const {}                          \
+  void parser<TYPE>::anchor() {}                                               \
+  } /* namespace llvm::cl */
+
+IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION(mlir::iree_compiler::CPUCodegenOptions)
+IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION(mlir::iree_compiler::GPUCodegenOptions)
+#undef IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION
+
 namespace mlir::iree_compiler {
 
 std::string CodegenOptions::tuningSpecPath = "";
--- a/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
@@ -9,6 +9,24 @@
 IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::CPUCodegenOptions);
 IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::GPUCodegenOptions);
 
+// Defines the out-of-line parser methods declared by
+// IREE_DECLARE_CODEGEN_OPTIONS_PASS_OPTION. Paired with the declaration macro
+// in CodegenOptions.h.
+#define IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION(TYPE)                          \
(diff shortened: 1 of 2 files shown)


hanhanW:
It is triggered in `--debug`, so it is hard to add a test. I already verified it locally with `iree-compile --iree-hal-target-device=local --iree-hal-local-target-device-backends=llvm-cpu --iree-llvmcpu-target-cpu=host samples/models/simple_abs.mlir --debug 2> /dev/null` command.

hanhanW:
Convert to draft because it may be related to my other reverted PR: https://github.com/iree-org/iree/pull/24220

I have a local fix, and I'll raise a PR later today.

hanhanW:
@kuhar you are the first person poped out in my mind for this kind of change, so I added you to reviewers. This is not urgent IMO, so please take your time.

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
```diff
@@ -9,6 +9,24 @@
...
+    return false;                                                              \
+  }                                                                            \
+  void parser<TYPE>::printOptionDiff(const Option &, TYPE, const OptVal &,     \
+                                     size_t) const {}                          \
+  void parser<TYPE>::anchor() {}                                               \
+  } /* namespace llvm::cl */
+
+IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION(mlir::iree_compiler::CPUCodegenOptions)
+IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION(mlir::iree_compiler::GPUCodegenOptions)
+#undef IREE_DEFINE_CODEGEN_OPTIONS_PASS_OPTION
```
kuhar: nit: with just two expansions, I'd drop the macro and expands by hand for readability

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.h
```diff
@@ -99,4 +99,44 @@ struct GPUCodegenOptions : CodegenOptions {
...
+    parser(Option &O) : basic_parser(O) {}                                     \
+    bool parse(Option &O, StringRef ArgName, StringRef Arg, TYPE &Val);        \
+    StringRef getValueName() const override { return VALUE_NAME_STR; }         \
+    void printOptionDiff(const Option &O, TYPE V, const OptVal &Default,       \
+                         size_t GlobalWidth) const;                            \
+    void anchor() override;                                                    \
+  };                                                                           \
+  } /* namespace llvm::cl */
+
+IREE_DECLARE_CODEGEN_OPTIONS_PASS_OPTION(mlir::iree_compiler::CPUCodegenOptions,
```
kuhar: similar here -- I think we are unlikely to grow the number of options objects and I'd just duplicate the implementation

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
hanhanW: It is intended as I expect that there will be ~4 in total. The missing two are SPIR-V and VMVX. However, they are fixed (i.e., I dont expect future changes others than upstream API changes which is also rare), so I'm okay with expanding them.

(This code snippet is easy, and the other is harder in terms of readability IMO.)

kuhar (approved):

PR #24117, opened 2026-04-14
[Codegen] Enable DMA by default for F16/BF16 Gemm on gfx950

Yu-Zhewen:
This PR flips the default DMA from false to true. Currently only BF16 GEMM has been properly benchmarked, but going to expand other supports as a follow-up.

Overall GEMM benchmark: **+5.9%** geomean speedup if TN (LHS transposed, RHS not transposed) cases are excluded. **+4.3%** if included.

#### Top 20 improved

| # | Op | Shape (MxNxK) | Transpose | NoDMA (us) | DMA (us) | Speedup |
|---|-----|--------------|--------|------------|----------|---------|
| 1 | mm | 7680x304x512 | NT | 14.4 | 9.7 | +32.7% |
| 2 | mm | 1280x3840x3840 | NN | 84.0 | 59.5 | +29.2% |
| 3 | mm | 24576x304x512 | NT | 33.4 | 23.8 | +28.6% |
| 4 | addmm | 256x3840x3840 | NT | 39.2 | 28.0 | +28.6% |
| 5 | mm | 1280x2304x576 | NT | 11.6 | 8.4 | +27.4% |
| 6 | mm | 1280x3840x7680 | NN | 164.4 | 119.4 | +27.3% |
| 7 | mm | 960x256x2048 | NT | 20.4 | 15.0 | +26.5% |
| 8 | mm | 4096x2048x8192 | NN | 215.5 | 159.9 | +25.8% |
| 9 | mm | 4096x2048x6144 | NN | 163.1 | 121.7 | +25.4% |
| 10 | mm | 4096x2048x4096 | NN | 112.7 | 84.4 | +25.1% |
| 11 | addmm | 5x384x384 | NT | 4.7 | 3.5 | +24.9% |
| 12 | mm | 3072x256x2048 | NT | 21.2 | 16.1 | +24.1% |
| 13 | addmm | 18928x128x512 | NT | 13.7 | 10.4 | +23.9% |
| 14 | mm | 1285x8192x2048 | NN | 76.9 | 59.4 | +22.9% |
| 15 | mm | 4112x2048x8192 | NN | 275.1 | 213.3 | +22.5% |
| 16 | mm | 256x3840x3840 | NN | 41.7 | 32.4 | +22.2% |
| 17 | mm | 1280x1152x576 | NT | 8.1 | 6.3 | +22.1% |
| 18 | mm | 4096x2048x2048 | NN | 57.0 | 44.5 | +22.0% |
[...]

PR #24192, opened 2026-04-21
[Codegen] Replace DMA fallback warning with LDBG

Yu-Zhewen:
https://github.com/iree-org/iree/pull/24117#discussion_r3119424091


hanhanW (approved):
thanks!

PR #24193, opened 2026-04-21
[Codegen] Add bufferization support for CastToRaggedShapeOp and LineaRaggedDimsOp

MaheshRavishankar:
This is PR 4/N of landing block sparse changes that are captured in this
branch :

https://github.com/MaheshRavishankar/iree/tree/users/MaheshRavishankar/blockSparseCommits

Adds `BufferizableOpInterfac`e external models for two tensor-ext ops that describe ragged shapes:
                                                                                                           
  - `iree_tensor_ext.cast_to_ragged_shape` — reinterprets a source buffer as a rank-(n+1) ragged buffer by   
  attaching a RaggedShapeAttr as the memref layout.                                                        
  -`iree_tensor_ext.linearize_ragged_dims` — the inverse view: collapses the ragged dimensions of a source  
  buffer back into a single dense dimension and drops the ragged layout.                                   
  
  Both ops are pure views, so the interface implementations:                                               
  - Report the operands as reads and never as writes (including `resultBufferizesToMemoryWrite = false`).
  - Set `mustBufferizeInPlace = true` and mark the result as `BufferRelation::Equivalent` to source, so        
  bufferization does not introduce a copy.                                                         
  - For `cast_to_ragged_shape`, the result memref's layout is the ragged encoding promoted from the tensor   
  encoding; operand buffers use the identity layout.                                                    
  - For `linearize_ragged_dims`, both operand and result memrefs use the identity layout; the result simply  
  drops the ragged encoding.                                                                             
                                                                                                           
  `bufferize` replaces the tensor-typed op with an identical memref-typed version of itself — no data is
  moved.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
@@ -3154,3 +3154,116 @@ module attributes {hal.executable.target = #executable_target_rocm} {
 // CHECK-LABEL: func.func @constant_to_buffer_rocm
 // CHECK:         %[[CST:.+]] = arith.constant dense<[1, 2, 3, 4]> : tensor<4xi32>
 // CHECK:         bufferization.to_buffer %[[CST]] {{.*}} : tensor<4xi32> to memref<4xi32, #gpu.address_space<constant>>
+
+// -----
+
+#pipeline_layout = #hal.pipeline.layout<bindings = [
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>
+]>
+func.func @cast_to_ragged_shape_static() -> tensor<10x3x?x30xf32, #iree_tensor_ext.ragged_shape<1>> {
+  %c0 = arith.constant 0 : index
+  %source_binding = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) alignment(64) offset(%c0)
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<10x20x30xf32>>
+  %column_lengths_binding = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) alignment(64) offset(%c0)
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<4xindex>>
+  %source = iree_tensor_ext.dispatch.tensor.load %source_binding, offsets=[0, 0, 0], sizes=[10, 20, 30], strides=[1, 1, 1]
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<10x20x30xf32>> -> tensor<10x20x30xf32>
+  %column_lengths = iree_tensor_ext.dispatch.tensor.load %column_lengths_binding, offsets=[0], sizes=[4], strides=[1]
+    : !iree_tensor_ext.dispatch.tensor<readonly:tensor<4xindex>> -> tensor<4xindex>
+  %0 = iree_tensor_ext.cast_to_ragged_shape %source ragged_dim(1) column_lengths(%column_lengths)
+      : (tensor<10x20x30xf32>, tensor<4xindex>)
+      -> tensor<10x3x?x30xf32, #iree_tensor_ext.ragged_shape<1>>
+  return %0 : tensor<10x3x?x30xf32, #iree_tensor_ext.ragged_shape<1>>
+}
+
+// CHECK-LABEL: func.func @cast_to_ragged_shape_static()
+//       CHECK:   %[[SOURCE:.+]] = hal.interface.binding.subspan {{.+}} binding(0)
+//  CHECK-SAME:       : memref<10x20x30xf32, #hal.descriptor_type<storage_buffer>>
+//       CHECK:   %[[COLUMN_LENGTHS:.+]] = hal.interface.binding.subspan {{.+}} binding(1)
+//  CHECK-SAME:       : memref<4xindex, #hal.descriptor_type<storage_buffer>>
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
```diff
@@ -3154,3 +3154,116 @@ module attributes {hal.executable.target = #executable_target_rocm} {
...
+//  CHECK-SAME:     %[[D2:[A-Za-z0-9]+]]: index
+//       CHECK:   %[[C1:.+]] = arith.constant 1 : index
+//       CHECK:   %[[SOURCE:.+]] = hal.interface.binding.subspan {{.+}} binding(0)
+//  CHECK-SAME:       : memref<?x?x?xf32, #hal.descriptor_type<storage_buffer>>{%[[D0]], %[[D1]], %[[D2]]}
+//       CHECK:   %[[NUM_RAGGED_ROWS_P1:.+]] = arith.addi %[[NUM_RAGGED_ROWS]], %[[C1]] : index
+//       CHECK:   %[[COLUMN_LENGTHS:.+]] = hal.interface.binding.subspan {{.+}} binding(1)
+//  CHECK-SAME:       : memref<?xindex, #hal.descriptor_type<storage_buffer>>
+//       CHECK:   %[[RAGGED:.+]] = iree_tensor_ext.cast_to_ragged_shape %[[SOURCE]]
+//  CHECK-SAME:       ragged_dim(1) column_lengths(%[[COLUMN_LENGTHS]])
+//  CHECK-SAME:       -> memref<?x?x?x?xf32, #iree_tensor_ext.ragged_shape<1>>
```
hanhanW, line 3230: Unfortunately, it looks like we have to drop the memory scope. Does it mean that all the later transformation would need to track through the op to get the memory scope if memory scope is important to them?

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -569,6 +569,232 @@ struct StoreToBufferOpSubsetInsertionInterface
...
+      const bufferization::AnalysisState &state) const {
+    return false;
+  }
+
+  // A view must share storage with its source; bufferization cannot introduce
+  // a copy for this op.
+  bool mustBufferizeInPlace(Operation *op, OpOperand &opOperand,
+                            const bufferization::AnalysisState &state) const {
+    return true;
+  }
```
hanhanW, line 607: I'm not sure if we should relax this for source or not. Maybe it is okay if it works for you. Here is a summary from LLM:

```
  For source it's defensible: the result is BufferRelation::Equivalent to source, so forcing in-place is a hint that matches the op's semantics (the view is of this buffer). Upstream tensor::CastOp / tensor::ExpandShapeOp /
  tensor::CollapseShapeOp don't override mustBufferizeInPlace at all — they let the default (false for ranked tensors) apply, and let the analyzer pick in-place because of the Equivalent aliasing. Either choice is reasonable for
  source.

  Suggested fix

  Gate on the operand:

  bool mustBufferizeInPlace(Operation *op, OpOperand &opOperand,
                            const AnalysisState &state) const {
    auto castOp = cast<IREE::TensorExt::CastToRaggedShapeOp>(op);
    return opOperand.get() == castOp.getSource();
  }

  Or drop the override entirely and rely on the default, matching the upstream view ops.
```

compiler/src/iree/compiler/Codegen/Common/test/iree_comprehensive_bufferize.mlir
hanhanW, line 1: We are missing tests for `linearize_ragged_dims` op?

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -569,6 +569,232 @@ struct StoreToBufferOpSubsetInsertionInterface
...
+
+  // The result aliases the `source` operand (same underlying buffer, layout
+  // without ragged encoding).
+  bufferization::AliasingValueList
+  getAliasingValues(Operation *op, OpOperand &opOperand,
+                    const bufferization::AnalysisState &state) const {
+    auto linearizeOp = cast<IREE::TensorExt::LinearizeRaggedDimsOp>(op);
+    if (opOperand.get() == linearizeOp.getSource()) {
+      return {{linearizeOp.getResult(), BufferRelation::Equivalent}};
+    };
```
hanhanW, line 739:
```suggestion
    }
```

compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
```diff
@@ -569,6 +569,232 @@ struct StoreToBufferOpSubsetInsertionInterface
...
+  // The result aliases the `source` operand (same underlying buffer, new
+  // layout). `column_lengths` is consumed as shape metadata and does not
+  // alias the result.
+  bufferization::AliasingValueList
+  getAliasingValues(Operation *op, OpOperand &opOperand,
+                    const bufferization::AnalysisState &state) const {
+    auto castToRaggedShapeOp = cast<IREE::TensorExt::CastToRaggedShapeOp>(op);
+    if (opOperand.get() == castToRaggedShapeOp.getSource()) {
+      return {{castToRaggedShapeOp.getResult(), BufferRelation::Equivalent}};
+    };
```
hanhanW, line 618:
```suggestion
    }
```

hanhanW (commented):
PR title typo: LineaRaggedDimsOp → should be LinearizeRaggedDimsOp.

PR #24194, opened 2026-04-21
[CPU] Drop benchmarking requirement from cpu torch op tests.

hanhanW:
We do not track matmul perf actively on threadripper, so we can drop the config. It relaxes the resource requirement, so the future cpu torch op tests can be run on more runners -- which reduces the waiting time.

### DIFF
--- a/.github/workflows/pkgci_test_torch.yml
+++ b/.github/workflows/pkgci_test_torch.yml
@@ -39,7 +39,6 @@ jobs:
               - persistent-cache
               - Linux
               - X64
-              - threadripper
           - name: amdgpu_hip_gfx1100_O3
             config-file:  torch_ops_gpu_hip_gfx1100_O3.json
             runs-on: [Linux, X64, gfx1100]
--- a/tests/external/iree-test-suites/torch_ops/torch_ops_cpu_llvm_sync.json
+++ b/tests/external/iree-test-suites/torch_ops/torch_ops_cpu_llvm_sync.json
@@ -15,14 +15,5 @@
     "AB/2048x2048xf32_bench"
   ],
   "expected_compile_failures": [],
-  "expected_run_failures": [],
-  "golden_times_ms": {
-    "AB/8192x8192xf32_bench": 5587.488262355328,
-    "AB/1024x1024xf32_bench": 1.1874544098876767,
-    "AB/256x256xf32_bench": 0.044473891122477315,
-    "AB/128x128xf32_bench": 0.03577919309132721,
-    "AB/2048x2048xf32_bench": 10.41092509722771,
-    "AB/4096x4096xf32_bench": 131.7884701769799,
-    "AB/512x512xf32_bench": 0.12528392536236224
-  }
+  "expected_run_failures": []
 }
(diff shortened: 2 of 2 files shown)


amd-eochoalo (approved):
I think this should work, just wait until torch_ops tests passes.

amd-eochoalo:
@hanhanW can you also add the benchmarks to the skip compile / run test list? The correctness tests should already have enough coverage.

hanhanW:
> @hanhanW can you also add the benchmarks to the skip compile / run test list? The correctness tests should already have enough coverage.

I don't follow how does it work in terms of having enough coverage. According to the [doc](https://github.com/iree-org/iree-test-suites/tree/main/torch_ops):

```
skip_compile_tests: List of tests which compilation should be skipped. Formatted as ${CLASS_DIR}/${TEST_DIR}
skip_run_tests: List of tests which run should be skipped.
expected_compile_failures: List of tests which one expects compilation to fail.
expected_run_failures: List of tests which one expects run to fail.
golden_times_ms: Dictionary of tests with golden times in ms.
```

Are we missing something from the doc or config?

hanhanW:
It looks like `_bench` is a hidden convention that only exercise benchmarks? If it is not documented somehwere (at least I don't find it from `README.md`), can you help fill the documentation gap?

amd-eochoalo:
@hanhanW Let me take a look into it.

amd-eochoalo:
> I don't follow how does it work in terms of having enough coverage.

What I mean by this is that there are two types of tests, benchmark and quality. Quality is for verification of correctness. Benchmark does not test for correctness just execution time. If we are dropping benchmarking requirement from cpu torch op tests, we should just not run the tests at all (skip compile and run). Right now the tests are in the skip run section which would still compile them. I think moving them to skip compile would make more sense.

> Are we missing something from the doc or config?

What exactly do you find missing from your quote? Just trying to understand this comment.

> It looks like _bench is a hidden convention that only exercise benchmarks?

It is not a hidden convention, it is the name of the test / benchmark. Maybe it is not too clear? It is just a different "mode" test.

> mode: Either "compare" or "benchmark". Defaults to "compare". And for example, 

```python
    @staticmethod
    @test
    def benchmark_data():
        seed = 14
        SQUARE = [
            128,
            256,
            512,
            1024,
            2048,
            4096,
            8192,
        ]

        for n in SQUARE:
            name = f"{n}x{n}xf32_bench"
            args = formulas(2, shape=(n, n))
            yield GenConfig(name, args=args, seed=seed, mode="benchmark") # multiple benchmarks with different names
            seed += 1
```

One can see here where the name of the tests are defined.

hanhanW:
I see, thanks for taking a look. I think I misunderstood that the categories. So both quality and benchmark are "tests", and my old mental model is that `test == numeric test` while benchmark is another thing. Sorry about that. That's why I did not expect to add items to `skip_run_tests`.

RE `skip_compile_tests` part: I do have the same confusion, and thanks for clarifying it. It was following the other skipped test (e.g., `AB/8192x8192xf32_bench`). I'm happy to add all them to the list as well.

PR #24196, opened 2026-04-21
[Codegen][NFC] Make getMmaScheduleFromProblemAndTarget and padding utilities public

Muzammiluddin-Syed-ECE:
This PR is part of a series of PRs aiming to enable scale operand preshuffling for mxfp4 gemms on cdna4.

Make getMmaScheduleFromProblemAndTarget a public function (was static) so that GPUEncodingExternalModels can call it to compute MMA schedules when choosing data-tiled layouts. Add padBoundForHeuristic and padProblemForHeuristic inline helpers to GPUHeuristics.h for reuse outside ConfigUtils.

Specifically, this change enables this future PR: https://github.com/iree-org/iree/pull/24198.

Made-with: Cursor

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
@@ -189,6 +189,53 @@ struct GPUMMASchedule {
   }
 };
 
+/// Pads a dimension bound to a favorable alignment for the MMA heuristic.
+/// The heuristic uses power-of-2 split factors, so odd tile counts (e.g.,
+/// ceil(1000/16) = 63) cannot distribute evenly. Padding to a multiple of
+/// |alignment| avoids this. Returns the original bound if already aligned
+/// or dynamic.
+inline int64_t padBoundForHeuristic(int64_t bound, int64_t alignment) {
+  if (ShapedType::isDynamic(bound)) {
+    return bound;
+  }
+  int64_t remainder = bound % alignment;
+  if (remainder == 0) {
+    return bound;
+  }
+  return bound + alignment - remainder;
+}
+
+/// Pads M and N sizes in a GPUMatmulShapeType so that the MMA heuristic sees
+/// power-of-2-friendly tile counts. This matches the padding applied by the
+/// TileAndFuse config path in ConfigUtils.cpp.
+///
+/// Thresholds: <=32 left as-is (single or few tiles, padding unnecessary);
+/// (32,128] padded to 32; >128 padded to 128. These correspond to the
+/// workgroup tile sizes the TileAndFuse config path typically selects.
+///
+/// K dimensions are not padded because the heuristic's K split factors are
+/// less alignment-sensitive than M/N workgroup distribution.
+inline void padProblemForHeuristic(GPUMatmulShapeType &problem) {
+  auto padDim = [](int64_t bound) -> int64_t {
+    if (ShapedType::isDynamic(bound)) {
+      return bound;
+    }
+    if (bound > 128) {
+      return padBoundForHeuristic(bound, 128);
+    }
+    if (bound > 32) {
+      return padBoundForHeuristic(bound, 32);
+    }
+    return bound;
+  };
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.h
```diff
@@ -189,6 +189,48 @@ struct GPUMMASchedule {
...
+  }
+  return bound + alignment - remainder;
+}
+
+/// Pads M and N sizes in a GPUMatmulShapeType so that the MMA heuristic sees
+/// power-of-2-friendly tile counts. This matches the padding applied by the
+/// TileAndFuse config path in ConfigUtils.cpp. Dimensions <= 32 are left
+/// unpadded; dimensions > 128 are padded to 128; dimensions in (32, 128] are
+/// padded to 32.
+inline void padProblemForHeuristic(GPUMatmulShapeType &problem) {
```
amd-eochoalo, line 218: Just trying to learn more about this area of code: would it be a problem to have 2^{n}+1 be padded to 2^{n+1}?  Seems like a lot of padding in some cases, but I am not too familiar with the performance tradeoff.

amd-eochoalo: Ah, no worries I see now that the padding is just a multiple of alignment. Not a power of two. Thanks!

amd-eochoalo (approved):

efric (commented):
I'm not sure if it's a good idea to put tile and fuse specific heuristics in this common header. Why does this need to be here?

Muzammiluddin-Syed-ECE:
> I'm not sure if it's a good idea to put tile and fuse specific heuristics in this common header. Why does this need to be here?

This belongs in the PR description so ill update that as well.

The motivation behind this move is so I can access them during `MaterializeDeviceEncodingPass` to resolve this issue: https://github.com/iree-org/iree/issues/24250.

The way that the data-tiled path selects tile configs does not give good performance at the moment for scaled gemms so we use the seeds-based tile-config selection heuristic instead. For this heuristic to be accessible to the data-tiled path it needs to be made public.

GPUHeuristics feels like a valid place for this since the seeds and logic used in this heuristic are independent of the lowering pipeline you choose to use, so other pipelines that aren't TileAndFuse could benefit from this as well if they so wished.

(Also to clarify, in this case, the data tiled scaled mma does still use the TileAndFuse pipeline, it's just that data tiling makes its tiling decisions earlier for the same lowering pipeline)

kuhar:
@hanhanW could you check if this layering looks good?

hanhanW:
> @hanhanW could you check if this layering looks good?

I don't follow the work closely, so I can only answer a few problems. The layering looks okay in terms of dependency. This is more like having a different heuristic (which may need @bjacob or @Max191 weigh in) to me.

The quick issue that I can spot is that it is pulling in a heavy function, which may result in in frequent changes in some lit tests and may make future changes harder as people need to understand and play with all those heuristic logic. The good side is that this path is closer to what is used in practice (assuming that they are stable heuristic and they are used without additional flags in practice).

Why heavy function may be an issue?

E.g., the work is mainly targeting scaled_matmul, but there are much more factors like `isGemm`, `useDirectLoad`, `doCPromotion`, `splitReductionTripCnt`, etc.

E.g., `getMmaScheduleFromProblemAndTarget` is used by other code. What happens if we only want to tweak data-tiling part? They should not be a blocker for future dat-tiling work, which is an important point.

The other principle is that the flags used in practice should be minimal. E.g., data-tiling just need `iree-opt-data-tiling` flag, and we may need an additional flag for custom ukernels on ROCm path. As long as you wont have five/ten flags in your config, that's fine. Otherwise, it is a little fragile to me.

Muzammiluddin-Syed-ECE:
Update: The lighter alternative I have been investigating is to improve the existing heuristic used in data-tiling and I have a very brief explanation in this issue: https://github.com/iree-org/iree/issues/24371

But it seems promising, if this line of work goes well, we can forgo the 2 PRs associated to this change.

PR #24197, opened 2026-04-21
[Codegen] Add unswizzled_operands attribute to DataTiledScaledMMAAttr

Muzammiluddin-Syed-ECE:
This PR is part of series of PRs aiming to enable scale operand preshuffling for mxfp4 gemms on cdna4.

Add an optional `unswizzled_operands` field to `DataTiledScaledMMAAttr` that lists operand indices whose data layout should not be shuffled (pack-only, no expand_shape/transpose). For these operands, `getSwizzle` returns an identity permutation and `populateOperandOffsetsSizesStrides` uses the distribution swizzle for thread delinearization.

This enables a codegen path where data operands (LHS/RHS) skip the expensive global-to-LDS transpose by keeping their packed layout identity-ordered, while scale operands remain fully shuffled for MMA consumption.

Made-with: Cursor

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPUTileSwizzleUtils.cpp
@@ -262,8 +262,8 @@ TileSwizzle getSwizzle(IREE::GPU::DataTiledScaledMMAAttr scaledMma,
 
 TileSwizzle getDistributionSwizzle(IREE::GPU::DataTiledScaledMMAAttr scaledMma,
                                    unsigned operandIdx) {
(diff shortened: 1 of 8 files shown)


bjacob:
Two verbs are being used around here: the existing *to swizzle* and now, with this PR, *to shuffle*.  Is there a meaningful nuance being reflected by using these two distinct verbs?

Muzammiluddin-Syed-ECE:
> Two verbs are being used around here: the existing _to swizzle_ and now, with this PR, _to shuffle_. Is there a meaningful nuance being reflected by using these two distinct verbs?

No real distinction, in fact `unswizzled_operands` would also be the more appropriate name, updating!

krzysz00 (approved):
I think you'll want to update the PR title, but overall this looks fine to me

bjacob (approved):

PR #24201, opened 2026-04-21
[Codegen] Improve transfer_gather vectorization method.

hanhanW:
The revision vectorizes the preOp first and passes the vector IRs to transfer_gather op creation directly. This avoids `mask->unmask` sequence in the first place, so it does not rely on missing foldings/canonicalization to recover the state.

The revision also adds a new test file to exercies unmaksed path.

Assisted-by: Claude Code


hanhanW:
cc @NoumanAmir657 can you help review? Thanks.

NoumanAmir657:
> cc @NoumanAmir657 can you help review? Thanks.

I think this is good to go. I tested locally and now the `dynamic_paged_gather_read` test case compiles e2e for CPU so I think after this we can safely turn on the `vectorizeToTransferGather` option by default.

amd-eochoalo (approved):
Thanks! Looks like a lot of temporary ops are omitted!

Groverkss (approved):

PR #24202, opened 2026-04-22
[LinalgExt] Add e2e tests for TopkV2Op

bangtianliu:
This PR adds tests for `iree_linalg_ext.topk_v2`


hanhanW (commented):
Lets add some dynamic cases. Here are some examples for creating inputs with dynamicshapes.

https://github.com/iree-org/iree/blob/00dc51c9bf143cbdb23395fd1704988e9e659a18/tests/e2e/regression/linalg_ops_dynamic.mlir#L34-L47

hanhanW:
> Lets add some dynamic cases. Here are some examples for creating inputs with dynamicshapes.
> 
> https://github.com/iree-org/iree/blob/00dc51c9bf143cbdb23395fd1704988e9e659a18/tests/e2e/regression/linalg_ops_dynamic.mlir#L34-L47

Is it ready for review?

bangtianliu:
> Is it ready for review?

Yes

hanhanW (commented):
A few problems (and one already exists in the old topk op):

1. Without `is_sorted` attribute, the order can be non-deterministic. It really depends on implementation details. Given that you're going to optimize the op on LLVMGPU path, this is going to be a big problem. We can either make the top k values be the same or pipe the result to `linalg_ext.sort` and have the same check.
2. The `k` can be dynamic, right? It is missing in the test.
3. topk_v2 is redundant in function name, IMO.

hanhanW (approved):

PR #19502, opened 2024-12-17
[Stream] Implement SpecializeEncodings pass (1/n)

hanhanW:
There are three major changes in the revision:

- Introduce `AffinityAnalysisDialectInterface` Stream dialect interface. It is used to fetch attributes that are defined by other dialects. In the revision, HAL implements the dialect interface, and it can return whatever attribute attached in HAL::ExecutableTarget attributes. The main idea of the dialect interface is that Stream **does not** need to depend on HAL to get the layout information.
- Add `cloneWithLayouts` method to the EncodingAttr. It is used in the encoding specialization pass where it can resolve the layout requirements and add it to the `layouts` field. The other optional parameters are dropped because the layout is already resolved. It can be a new Encoding dialect attribute because it is just describing the layout. The stream tensor ops do not need to know the `op_type`, `element_types` and `operand_index` parameters. It only needs the layout information, and the attribute should implement the interface method.
- Partially implement the SpecializeEncodings pass. The responsibility of the pass is large, so I decide to implement it incrementally. This revision only implements the mechanism of updating stream tensor ops' encoding, and only stream.tensor.sizeof op is supported. The rest of the support for other stream tensor op can be added later on. The executable duplication and the update of dispatch ops will be implemented in subsequent PRs.

PR #24204, opened 2026-04-22
[DispatchCreation] Add experimental flag to perform scale-only hoisting in HoistEncodingOps for MXFP4 Gemms

Muzammiluddin-Syed-ECE:
This PR is part of series of PRs aiming to enable scale operand preshuffling for mxfp4 gemms on cdna4.

On gfx950, data operands of scaled_matmul benefit from staying fused in the dispatch (identity-ordered global loads avoid the global-to-LDS transpose), while scale operands benefit from pre-encoding via a separate dispatch. Add `shouldHoistScalesOnly` which checks DeviceAnalysis for a gfx950 target and, when true, only hoists set_encoding ops for operand indices 2 and 3 (scales), keeping indices 0 and 1 (data) inside the dispatch.

Made-with: Cursor


compiler/src/iree/compiler/DispatchCreation/test/hoist_encoding_ops.mlir
```diff
@@ -1,4 +1,5 @@
 // RUN: iree-opt --pass-pipeline="builtin.module(iree-dispatch-creation-hoist-encoding-ops,cse)" --split-input-file %s | FileCheck %s
+// RUN: iree-opt --pass-pipeline="builtin.module(iree-dispatch-creation-hoist-encoding-ops,cse)" --test-iree-dispatch-creation-no-hoist-data-operands-scaled-mma --split-input-file %s | FileCheck %s --check-prefix=PDT
```
MaheshRavishankar: Please move this to a separate file.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
```diff
@@ -63,6 +63,8 @@ static llvm::cl::opt<bool> clNoHoistDataOperandsScaledMMA(
 
 namespace mlir::iree_compiler::IREE::GPU {
 
+bool noHoistScaledMmaDataOperands() { return clNoHoistDataOperandsScaledMMA; }
```
MaheshRavishankar: Its strange to have a dispatch creation related flag in codegen. This is layering violation

MaheshRavishankar (changes requested):
Its fine as an experimental flag, but there is a layering violation here (that seems to be from a previous commit).

compiler/src/iree/compiler/Codegen/ExternalInterfaces/GPUEncodingExternalModels.cpp
Muzammiluddin-Syed-ECE: I will change this to two separate test flags, that aligns better with how future work will be split anyways.

Muzammiluddin-Syed-ECE:
Removed the use of the test flag from e2e test, instead relying on local testing for verification. Once this work arrives to a more stable state, e2e tests can be updated accordingly.

compiler/src/iree/compiler/DispatchCreation/HoistEncodingOps.cpp
```diff
@@ -40,6 +42,13 @@
 
 #define DEBUG_TYPE "iree-dispatch-creation-hoist-encoding-ops"
 
+static llvm::cl::opt<bool> clNoHoistDataOperandsScaledMMA(
+    "test-iree-dispatch-creation-no-hoist-data-operands-scaled-mma",
```
hanhanW: We name experimental flag with `experimental`, not `test`.

`iree-dispatch-creation-experimental-no-hoist-data-operands-scaled-mma`

compiler/src/iree/compiler/DispatchCreation/HoistEncodingOps.cpp
```diff
@@ -314,6 +323,23 @@ void HoistEncodingOpsPass::runOnOperation() {
     if (isa_and_nonnull<IREE::Encoding::PaddingAttr>(encoding)) {
       return;
     }
+    // When --test-iree-dispatch-creation-no-hoist-data-operands-scaled-mma is
+    // set, data operands benefit from staying fused in the dispatch (identity
+    // permute loads) while scale operands benefit from pre-encoding via
+    // hoisting.
```
hanhanW: Redundant comment? Doesn't the var name already document it?

compiler/src/iree/compiler/DispatchCreation/test/hoist_encodings_ops_pdt.mlir
```diff
@@ -0,0 +1,89 @@
...
+#enc4 = #iree_encoding.encoding<operand_index = 4 : index, op_type = scaled_matmul,
+  element_types = [f4E2M1FN, f4E2M1FN, f8E8M0FNU, f8E8M0FNU, f32],
+  user_indexing_maps = [#sm0, #sm1, #sm2, #sm3, #sm4]>
+
+#gpu_target_950 = #iree_gpu.target<arch = "gfx950", features = "", wgp = <
+  compute = fp32, storage = b32, subgroup = shuffle, subgroup_size_choices = [64],
+  max_workgroup_sizes = [1024, 1024, 1024], max_thread_count_per_workgroup = 1024,
+  max_workgroup_memory_bytes = 65536, max_workgroup_counts = [2147483647, 2147483647, 2147483647]>>
+#exec_950 = #hal.executable.target<"rocm", "rocm-hsaco-fb", {iree_codegen.target_info = #gpu_target_950}>
+#device_950 = #hal.device.target<"local", {ordinal = 0 : index}, [#exec_950]> : !hal.device
```
hanhanW: Unused, and it is not needed.

compiler/src/iree/compiler/DispatchCreation/test/hoist_encodings_ops_pdt.mlir
```diff
@@ -0,0 +1,89 @@
...
+
+module {
```
hanhanW: Remove the module op.

compiler/src/iree/compiler/DispatchCreation/test/hoist_encodings_ops_pdt.mlir
hanhanW, line 1: You are exposing unnecessary information to this layer. Why does gfx950 matter? Let's drop it from the test.

Also, please expand the filename. Learining from https://github.com/iree-org/iree/pull/19502#discussion_r1888659672

> always use full names for things wherever possible - it's useful hints to readers as to what the type is and how it is treated - shortening to arbitrary forms makes it impossible to find/replace in codebases and more difficult for non-native speakers (here and anywhere else you've shortened things).

pdt is not a well-known term, IMO. And it does not hurt if we use full names.

compiler/src/iree/compiler/DispatchCreation/test/hoist_encodings_ops_pdt.mlir
```diff
@@ -0,0 +1,89 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(iree-dispatch-creation-hoist-encoding-ops,cse)" --split-input-file %s | FileCheck %s
```
hanhanW: Why do we need this RUN?

hanhanW (commented):
What is the plan for the new flag? Are we going to use this path as default for data-tiled scaled_matmul?

Muzammiluddin-Syed-ECE:
> What is the plan for the new flag? Are we going to use this path as default for data-tiled scaled_matmul?

What i had in mind is that this new flag will stay experimental as an example of IREE's performance for single scaled gemm workloads. Then after doing our due diligence and verifying that this flag is better for whole model cases like llama, we can make this flag the default. The thing to verify is whether hoisting might be less of a problem across larger models which could benefit from fusing away encoding dispatches for data operands.

compiler/src/iree/compiler/DispatchCreation/test/hoist_encodings_ops_pdt.mlir
Muzammiluddin-Syed-ECE: To make the comparison easier to see (without the flag set it gets hoisted whereas with it does not). However, I admit the existing tests in hoist_encodings_ops do that sufficiently well. Removing this run

compiler/src/iree/compiler/DispatchCreation/test/hoist_encodings_ops_scaled_gemm_no_hoist_flag.mlir
```diff
@@ -0,0 +1,63 @@
...
+#enc2 = #iree_encoding.encoding<operand_index = 2 : index, op_type = scaled_matmul,
+  element_types = [f4E2M1FN, f4E2M1FN, f8E8M0FNU, f8E8M0FNU, f32],
+  user_indexing_maps = [#sm0, #sm1, #sm2, #sm3, #sm4]>
+#enc3 = #iree_encoding.encoding<operand_index = 3 : index, op_type = scaled_matmul,
+  element_types = [f4E2M1FN, f4E2M1FN, f8E8M0FNU, f8E8M0FNU, f32],
+  user_indexing_maps = [#sm0, #sm1, #sm2, #sm3, #sm4]>
+#enc4 = #iree_encoding.encoding<operand_index = 4 : index, op_type = scaled_matmul,
+  element_types = [f4E2M1FN, f4E2M1FN, f8E8M0FNU, f8E8M0FNU, f32],
+  user_indexing_maps = [#sm0, #sm1, #sm2, #sm3, #sm4]>
+util.func public @gfx950_hoist_scales_only(
```
hanhanW: ditto, should not encode any target specific things in this phase. Please drop `gfx950`

hanhanW (commented):
> What i had in mind is that this new flag will stay experimental as an example of IREE's performance for single scaled gemm workloads. Then after doing our due diligence and verifying that this flag is better for whole model cases like llama, we can make this flag the default. The thing to verify is whether hoisting might be less of a problem across larger models which could benefit from fusing away encoding dispatches for data operands.

Thanks for sharing the idea! I think it is okay to land these things under an experimental flag and just want to flag that the flag may be dropped or not maintained in the future. Or we may set different encodings in the future, if it turns out that we want to partially enable it.

After thinking more, I did not convince myself that the LHS and RHS are not hoisted. We may need more data point, including explore the path, when you want to make the flip. Mostly a note and sharing what I've been thinking yesterday.

The PR itself is good to go, and I'll approve when the parent branch is landed. There are quite many pieces floating and I'm not able to track the dependency. (Thus, I don't put a blocker to this)

PR #24207, opened 2026-04-22
[DispatchCreation] Check all operands of a consumer before fusing

### LINKED ISSUE #24206: [DispatchCreation] fuseRootsWithConsumers skips per-operand fusibility checks
### What happened?

In `FormDispatchRegions.cpp`, `fuseRootsWithConsumers` iterates `getFusableUses`'s result operand-by-operand:

```cpp
for (OpOperand *fusableUse : fusableUses) {
    Operation *consumerOp = fusableUse->getOwner();
    if (tracker.isFusedOp(consumerOp)) continue;   // <-- first operand wins
    if (isFusableWithConsumer(*fusableUse, ...)) {
      tracker.appendToFusionGroup(consumerOp, fusionGroup);
    }
  }
```

When two operands in the vector have the same owner, the first one's `isFusableWithConsumer` verdict decides for the whole consumer; the second's check is skipped via `isFusedOp`.

Also, after #24169, multi-result producer into single consumer will also hit this issue.

Fix: group `getFusableUses`'s result by owner and require `llvm::all_of(operands, isFusableWithConsumer)` before appending the consumer.

keshavvinayak01:
`fuseRootsWithConsumers` iterates `getFusableUses`'s result one `OpOperand` at a time. When > 1 operands share a consumer, the first operand's verdict marks the consumer fused and the second operand is skipped via `isFusedOp`. That silently drops per-operand rejections that `isFusableWithConsumer` could raise.

Fixes https://github.com/iree-org/iree/issues/24206

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/FormDispatchRegions.cpp
@@ -658,11 +658,7 @@ fuseRootsWithConsumers(MLIRContext *context, ArrayRef<Operation *> roots,
       }
 
       // Group operands by owning consumer so the per-operand checks inside
-      // `isFusableWithConsumer` (e.g. pack-identity, insert_slice source
-      // single-use) are evaluated for every operand of a consumer, not just
-      // the first. Without the grouping, the first operand's verdict would
-      // mark the consumer fused and subsequent operands would be skipped via
-      // `isFusedOp`.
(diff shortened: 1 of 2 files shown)


MaheshRavishankar:
This looks OK to me, but can you run torch model tests here as well.

IanWood1:
> This looks OK to me, but can you run torch model tests here as well.

From https://github.com/iree-org/iree/pull/24169#issuecomment-4290244269 torch model tests are run by default

IanWood1 (approved):

PR #24210, opened 2026-04-22
[Codegen][GPU] Support pipelining gather_to_lds inside nested regions

jerryyin:
This PR fixed most (if not all) of the TN shape regressions in my local perf regression test.

Enable the prefetch pass to pipeline async gather_to_lds operations that are wrapped inside region-bearing ops (e.g., scf.if guards from warp predication). Previously, any gather_to_lds nested inside a region was rejected, causing fallback to stream copy mode and significant performance regressions for shapes where not all warps participate in DMA (e.g., 1134x2048x150000 f32 LHS-transposed GEMM). - This config is still skipped by current branch because innermost pad created a dynamic dim, but that's a separate issue.

If a warp skips its conditional async load, its vmcnt is already lower, so the wait is trivially satisfied. This makes pipelining scf.if-guarded async loads both safe.

Changes:
- Remove the hasNestedGatherToLDS pre-flight rejection.
- In async copy mode, treat any region-bearing op containing gather_to_lds as a load root, scheduling it as a single unit.
- Extend backward slice computation to capture dependencies of ops nested inside region-bearing roots and parents.
- Update tests to verify pipelining of predicated gather_to_lds.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -97,22 +97,6 @@ static bool hasDirectGatherToLDS(scf::ForOp forOp) {
   return false;
 }
 
-/// Checks if a loop contains gather_to_lds operations nested inside regions
-/// (e.g., inside scf.if). Such conditional async copies can't be reliably
-/// pipelined.
-static bool hasNestedGatherToLDS(scf::ForOp forOp) {
-  for (Operation &op : forOp.getBody()->getOperations()) {
-    if (op.getNumRegions() > 0) {
-      bool hasNested = false;
-      op.walk([&](amdgpu::GatherToLDSOp) { hasNested = true; });
-      if (hasNested) {
-        return true;
-      }
-    }
-  }
-  return false;
-}
-
 /// Checks if a loop contains stream copy operations (global read + shared
 /// write). This is mutually exclusive with async copy mode.
 static bool hasStreamCopyOps(scf::ForOp forOp) {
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -399,13 +383,22 @@ static LogicalResult identifyRootOperations(
...
   for (Operation &op : forOp.getBody()->getOperations()) {
     if (mode == PipelineMode::AsyncCopy) {
-      // Async copy mode: gather_to_lds and scf.yield
       if (isa<amdgpu::GatherToLDSOp>(op)) {
         loadRoots.push_back(&op);
         LDBG() << "  Load root: " << op;
       } else if (isa<scf::YieldOp>(op)) {
         computeRoots.push_back(&op);
         LDBG() << "  Compute root: " << op;
+      } else if (op.getNumRegions() > 0) {
```
Yu-Zhewen: nit: Can we revive the deleted helper function `hasNestedGatherToLDS` and reuse it here by stripping its for (Operation &op : forOp.getBody()->getOperations()) loop?

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -1067,6 +1063,13 @@ static Operation *findLastGatherToLDS(Block::iterator begin,
     if (isa<amdgpu::GatherToLDSOp>(&*it)) {
       return &*it;
     }
+    if (it->getNumRegions() > 0) {
```
Yu-Zhewen: same here

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -1091,6 +1094,12 @@ static void insertPrologueAsyncMarks(RewriterBase &rewriter, Location loc,
   for (auto it = parentBlock->begin(); it != loopStart; ++it) {
     if (isa<amdgpu::GatherToLDSOp>(&*it)) {
       prologueGathers.push_back(&*it);
+    } else if (it->getNumRegions() > 0) {
```
Yu-Zhewen: and here

Yu-Zhewen (approved):
Thanks! Cool improvement. Just one nit

PR #24211, opened 2026-04-22
[NFC] Outline transpose part of ContractToChainFMA

amd-eochoalo:
This is just a refactoring. The purpose of this is to separate ContractToChainFMA into:

* transpose
* flattening
* unrolling
* lowering

Which can then we can move the flattening and unrolling to the vector flattening and unrolling passes.


amd-eochoalo:
@efric requested this to be paused as he is working on this pattern. So, I'll keep it as a draft and work on other things.

PR #24215, opened 2026-04-22
Bump LLVM to llvm/llvm-project@5ea5b9eb8eec

hanhanW:
Applied fixes for https://github.com/llvm/llvm-project/commit/86b9775612f8
- Drop `getSizeLevel()`, `isOptimizingForSize()`, `OptimizationLevel::Os`, and `OptimizationLevel::Oz`, on the rationale that size optimization should come from` O2 + optsize/minsize` function attributes instead of pipeline-level `Os/Oz`.


hanhanW:
@IanWood1 @MaheshRavishankar FYI that the options (`Os` and `Oz`) are removed from upstream.

bangtianliu (approved):

MaheshRavishankar (approved):

PR #24220, opened 2026-04-22
[Codegen] Use local binders for optimization flags in codegen.

hanhanW:
Previously, I made a workaround in https://github.com/iree-org/iree/commit/1b75890ac8093dd9b74b8110833bd2a46b3da7f0. The main reason is that all the codegen pipeline tests were anchor on a single pass (e.g., `XXXLowerExecutableTargetPass`), which makes plumbing through options infeasible.

After migrating to the real pipeline transformations, we are able to use the option in whole pipeline tests; we can drop the old workaround from iree-opt changes.

The revision introduces the option for each backend (default `O0`) and use local binders to apply the optimization flags. It drops the `XXX::FromFlags::get()` uses from Codegen.

### DIFF
--- a/compiler/src/iree/compiler/API/Internal/IREEOptToolEntryPoint.cpp
+++ b/compiler/src/iree/compiler/API/Internal/IREEOptToolEntryPoint.cpp
@@ -10,7 +10,6 @@
 
 #include "iree/compiler/Dialect/HAL/Target/TargetRegistry.h"
 #include "iree/compiler/Dialect/VM/Target/init_targets.h"
-#include "iree/compiler/Pipelines/Options.h"
 #include "iree/compiler/PluginAPI/PluginManager.h"
 #include "iree/compiler/Tools/init_dialects.h"
 #include "iree/compiler/Tools/init_llvmir_translations.h"
@@ -76,9 +75,6 @@ static LogicalResult ireeOptMainFromCL(int argc, char **argv,
   tracing::DebugCounter::registerCLOptions();
   auto &pluginManagerOptions =
       mlir::iree_compiler::PluginManagerOptions::FromFlags::get();
-  // Register the top-level optimization level flag (--iree-opt-level) and
-  // pipeline options so that optimization-level cascading works in iree-opt.
-  mlir::iree_compiler::GlobalPipelineOptions::FromFlags::get();
 
   // Build the list of dialects as a header for the --help message.
   std::string helpHeader = (toolName + "\nAvailable Dialects: ").str();
@@ -90,13 +86,6 @@ static LogicalResult ireeOptMainFromCL(int argc, char **argv,
 
   // Parse pass names in main to ensure static initialization completed.
   cl::ParseCommandLineOptions(argc, argv, helpHeader);
-
-  // Apply optimization-level-dependent defaults to global options.
-  {
-    auto globalBinder = mlir::iree_compiler::OptionsBinder::global();
-    globalBinder.applyOptimizationDefaults();
-  }
-
   MlirOptMainConfig config = MlirOptMainConfig::createFromCLOptions();
 
   // The local binder is meant for overriding session-level options, but for
--- a/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
@@ -11,9 +11,40 @@ IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::GPUCodegenOptions);
 
 namespace mlir::iree_compiler {
 
+namespace {
+
+// Applies the opt-level default cascade using a local binder.
+template <typename CodegenOptionsT>
+void applyOptLevelDefaults(CodegenOptionsT &opts,
+                           llvm::OptimizationLevel level) {
+  auto binder = OptionsBinder::local();
+  // Anchor the opt-level hierarchy so applyOptimizationDefaults can walk it.
+  binder.topLevelOpt("opt-level", level);
+  opts.bindOptions(binder);
+  binder.applyOptimizationDefaults();
+}
+
+} // namespace
+
 std::string CodegenOptions::tuningSpecPath = "";
 bool CodegenOptions::setTunerAttributes = false;
 bool CodegenOptions::emitPipelineConstraints = false;
+
+llvm::OptimizationLevel
+mapCodegenPipelineOptLevel(CodegenPipelineOptLevel optLevel) {
+  switch (optLevel) {
+  case CodegenPipelineOptLevel::O0:
+    return llvm::OptimizationLevel::O0;
+  case CodegenPipelineOptLevel::O1:
+    return llvm::OptimizationLevel::O1;
+  case CodegenPipelineOptLevel::O2:
+    return llvm::OptimizationLevel::O2;
+  case CodegenPipelineOptLevel::O3:
+    return llvm::OptimizationLevel::O3;
+  }
+  assert(false && "unhandled codegen pipeline optimization level");
+  return llvm::OptimizationLevel::O0;
+}
 
 void CodegenOptions::bindOptions(OptionsBinder &binder) {
   static llvm::cl::OptionCategory category("IREE Codegen Options");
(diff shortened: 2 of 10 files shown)


compiler/src/iree/compiler/API/Internal/IREEOptToolEntryPoint.cpp
hanhanW, line 1: This is the partial revert of https://github.com/iree-org/iree/commit/1b75890ac8093dd9b74b8110833bd2a46b3da7f0

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
```diff
@@ -11,10 +11,41 @@ IREE_DEFINE_COMPILER_OPTION_FLAGS(mlir::iree_compiler::GPUCodegenOptions);
...
+
 std::string CodegenOptions::tuningSpecPath = "";
 bool CodegenOptions::setTunerAttributes = false;
 bool CodegenOptions::emitPipelineConstraints = false;
 
+llvm::OptimizationLevel
+mapCodegenPipelineOptLevel(CodegenPipelineOptLevel optLevel) {
+  switch (optLevel) {
+  case CodegenPipelineOptLevel::O0:
+    return llvm::OptimizationLevel::O0;
```
MaheshRavishankar, line 37: Wondering why you are returning `llvm::OptimizationLevel::`  here? Just curious thats all. Does this affect the LLVM compilation step?

MaheshRavishankar (commented):
So what flags are currently turned on with different optimization levels.

MaheshRavishankar (approved):

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
hanhanW: This does not affect LLVM compilation step. It is mostly because the `OptionBinder` takes `llvm::OptimizationLevel` as input; the function is doing the mapping for codegen enum to llvm enum.

However, your question let me rethink the approach. I think we may not need this enum. I'll try to use `llvm::OptimizationLevel` directly.

My mental model was: ah we need a enum here, let's declare one. Let me give it a shot.

hanhanW: I was wrong on using `llvm::OptimizationLevel` part. `llvm::OptimizationLevel` is a class that marked with `final`, and the inheritance branch is selected (because of the setup in MLIR). Thus, there is a compilation error.

```
  template <class DataType>
  class opt_storage<DataType, false, true> : public DataType {
  public:
    OptionValue<DataType> Default;

    template <class T> void setValue(const T &V, bool initial = false) {
      DataType::operator=(V);
      if (initial)
        Default = V;
    }

    DataType &getValue() { return *this; }
    const DataType &getValue() const { return *this; }
    ...
  };
```

benvanik:
# Codegen TSAN failure detected on landing-1

PR/job:

- PR: https://github.com/iree-org/iree/pull/24236
- Run: https://github.com/iree-org/iree/actions/runs/24861766248
- Job: https://github.com/iree-org/iree/actions/runs/24861766248/job/72788945911
- Failing job: `linux_x64_clang_tsan / linux_x64_clang_tsan`
- Head SHA: `7dc27fcbdebcc7122dc98b16e7c36a34c5ff6b74`
- PR base at investigation time: `b2df6f5db7469a79695fa806e84f08ad8649a1c8`

## Symptom

The TSAN job fails during the CMake/Ninja build, before CTest runs. The failing
build edge is generation of the TSAN vmfb for the check module:

```text
Generating check_llvm-cpu_local-task_generic_success.mlir_tsan_module.vmfb from success.mlir
FAILED: [code=66] runtime/src/iree/modules/check/test/check_llvm-cpu_local-task_generic_success.mlir_tsan_module.vmfb
.../build-tsan/tools/iree-compile ... --iree-llvmcpu-sanitize=thread .../runtime/src/iree/modules/check/test/success.mlir ...
ThreadSanitizer: reported 8 warnings
ninja: build stopped: cannot make progress due to previous errors.
```

All reported warnings have the same summary:

```text
SUMMARY: ThreadSanitizer: data race
/__w/iree/iree/third_party/llvm-project/llvm/include/llvm/Support/CommandLine.h:1378:15
in void llvm::cl::opt_storage<bool, true, false>::setValue<bool>(bool const&, bool)
```

## Mechanism

The race is in compiler option binding, not in the runtime/HAL/async landing
diff.

The write side constructs a local LLVM command-line option with external
storage:

- `compiler/src/iree/compiler/Utils/OptionUtils.h:143`
- `llvm::cl::opt<T, /*ExternalStorage=*/true>(..., llvm::cl::location(value), llvm::cl::init(value), ...)`

That construction writes the initial value through LLVM's `cl::opt_storage`.

The storage being written is shared static codegen option state:

- `compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp:58`
- `CodegenOptions::setTunerAttributes`

The read side is the default callback capture:

- `compiler/src/iree/compiler/Utils/OptionUtils.h:467`
- `V initialValue = *currentValue;`

In the captured report, the read came from the deprecated alias binding:

- `compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp:63`
- `iree-config-add-tuner-attributes`, bound to the same
  `CodegenOptions::setTunerAttributes` static.

The concurrency comes from MLIR parallel pass cloning/execution. The stack runs
through:

- `TranslateTargetExecutableVariantsPass::runOnOperation`
- `PipelineCache::getOrCreate`
- `OpPipelineAdaptorPass::runOnOperationAsync`
- `parallelForEach`

During pass cloning, generated pass option defaults call into codegen option
default construction. The relevant generated-default sources are:

- `compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.td:92`
- `CPUCodegenOptions::getWithOptLevel(llvm::OptimizationLevel::O0)`
- `compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.td:116`
- `GPUCodegenOptions::getWithOptLevel(llvm::OptimizationLevel::O0)`

Those call `setWithOptLevel`, which calls `applyOptLevelDefaults`, which
constructs a fresh `OptionsBinder::local()` and re-binds all codegen options:

- `compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp:20`
- `compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp:76`
- `compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp:81`

Because `CodegenOptions` uses static members shared by CPU/GPU option structs,
multiple worker threads construct local `llvm::cl::opt` objects that read/write
the same static booleans. LLVM command-line option construction is not a pure
data transform and is not suitable as a pass-default computation path.

## Why it appears on landing-1

This appears to be a mainline compiler issue exposed by the landing-1 CI run,
not a runtime/HAL regression in the landing stack.

The PR base includes:

```text
2ddd649dbea9 [Codegen] Use local binders for optimization flags in codegen. (#24220)
```

That commit introduced:

- `CPUCodegenOptions::getWithOptLevel`
- `GPUCodegenOptions::getWithOptLevel`
- local-binder-based opt-level default application in codegen options
- generated pass defaults that call those helpers

PR #24220's own `linux_x64_clang_tsan` job was skipped, so this landing PR is
likely one of the first CI runs to exercise the new codegen local-binder path
under TSAN.

## Proposed fix

Immediate unblock:

- Change the TableGen defaults from
  `CPUCodegenOptions::getWithOptLevel(llvm::OptimizationLevel::O0)` to
  `CPUCodegenOptions()`.
- Change the GPU default from
  `GPUCodegenOptions::getWithOptLevel(llvm::OptimizationLevel::O0)` to
  `GPUCodegenOptions()`.

This is semantically equivalent for O0 today: the struct field initializers
already encode the O0 defaults, including `reassociateFpReductions = false`.

Better durable fix:

- Remove `OptionsBinder::local()` from codegen pass default construction.
- Make opt-level default derivation a plain data transform:
  - `CPUCodegenOptions::setWithOptLevel(level)` should set `optLevel = level`
    and apply opt-level-sensitive fields directly.
  - Today the only visible opt-level-sensitive CPU field is
    `reassociateFpReductions`, false at O0/O1 and true at O2/O3.
  - `GPUCodegenOptions::setWithOptLevel(level)` can remain trivial until GPU
    has opt-level-sensitive fields.
- Keep `OptionsBinder::local()` for actual local parse/print/session binding,
  not for pass constructors or pass clone paths.

This should probably land as a small main fix because the bug is already in
`main`. If landing needs to proceed before that merges, carry the same fix in
`landing-1` temporarily and drop it after rebasing over the main fix.

## Verification targets

Focused reproduction from the CI log:

```bash
./build_tools/cmake/build_and_test_tsan.sh
```

The exact failing build action was:

```bash
build-tsan/tools/iree-compile \
  --output-format=vm-bytecode \
  --mlir-print-op-on-diagnostic=false \
  --iree-hal-target-backends=llvm-cpu \
  --iree-input-demote-f64-to-f32=false \
  --iree-llvmcpu-target-cpu=generic \
  --iree-llvmcpu-link-embedded=false \
  --iree-llvmcpu-sanitize=thread \
  runtime/src/iree/modules/check/test/success.mlir \
  -o check_llvm-cpu_local-task_generic_success.mlir_tsan_module.vmfb \
  --iree-hal-executable-object-search-path=build-tsan \
  --iree-llvmcpu-embedded-linker-path=build-tsan/llvm-project/bin/lld \
  --iree-llvmcpu-wasm-linker-path=build-tsan/llvm-project/bin/lld
```

For a local fix, the important evidence is that this command no longer reports
TSAN warnings from `llvm::cl::opt_storage::setValue` while MLIR pass pipelines
are cloned/run in parallel.

hanhanW:
Thanks for the flag. I'll revert it; reland it with TSAN succeeded.

PR #24223, opened 2026-04-23
[LinalgExt] Add OuterReduction tiling strategy for TopkV2Op

bangtianliu:


Groverkss (commented):
I'm not sure if we should be doing partial reduction on topk.

hanhanW:
> I'm not sure if we should be doing partial reduction on topk.

@bangtianliu What is the plan or action items to enable your fancy topk support? It looks like there is a disconnection here.

bangtianliu:
> > I'm not sure if we should be doing partial reduction on topk.
> 
> @bangtianliu What is the plan or action items to enable your fancy topk support? It looks like there is a disconnection here.

I want to do something like this along the VectorDistribute pipeline 

```mlir
scf.for %iv = 0 to N step T iter_args(%acc_vals, %acc_idxs):
  %tile = tensor.extract_slice %input[..., %iv] [..., T]
  %result:2 = iree_linalg_ext.topk_v2 dimension(D)
      ins(%tile, %iota_indices)
      outs(%acc_vals, %acc_idxs) { ^bb0(%a, %b): ... yield %cmp }
  scf.yield %result#0, %result#1
```

and then I add some lowering (dpp or ballot) to  `iree_linalg_ext.topk_v2` op. 

But the design has not been finalized yet.  Inspired by this blog: https://rocm.blogs.amd.com/software-tools-optimization/adaptive-topk/README.html

Groverkss:
> > I'm not sure if we should be doing partial reduction on topk.
> 
> @bangtianliu What is the plan or action items to enable your fancy topk support? It looks like there is a disconnection here.

Me and Bangtian had a chat. I think the original plan didnt need this, but Bangtian found a better way to do topk which needs this. So we are still discussing but I think we can have a github issue after we discuss this between us. Bangtian found a rocm blog that has some strategies for perf that I didn't know about: https://rocm.blogs.amd.com/software-tools-optimization/adaptive-topk/README.html .

bangtianliu:
> but I think we can have a GitHub issue after we discuss this between us. 

Here is the link to the issue: https://github.com/iree-org/iree/issues/24053, I will keep updating it.

Issue #23782, opened 2026-03-13
[VectorDistribute] Use direct-to-LDS in attention


sommerlukas:
AMD GPUs MI350 and later generations (gfx95+) support direct loading from global memory into LDS without intermediate staging in registers. Additionally, the transfer can also happen asynchronously, allowing to overlap computation and loading of data. 

The VectorDistribute pipeline does not yet use this feature.

Goal of this task is to enable the use of direct-to-LDS in the VectorDistribute codegen pipeline, with an initial focus on attention codegen.

The initial step is to enable the use of that feature in general, in later steps, software pipelining of the workload should enable overlap of computation and loading of the next block(s) of data.

sommerlukas:
# Status Overview & Proposal

The VectorDistribute pipeline currently does not yet use the direct-to-LDS feature present in AMD GPUs with architectures gfx95+. The feature allows to directly load (DMA) from global memory into workgroup-shared LDS, without intermediate staging in registers (VGPR). The load can also be made asynchronous, allowing to overlap data movement with computation.

Goal is to enable the direct-to-LDS feature in the VectorDistribute pipeline, for workloads such as attention.
## Status in the TileAndFuse (TAF) Pipeline

The TAF pipeline already supports the direct-to-LDS feature, including software pipelining of the kernel to enable overlap of computation and data movement. This is a brief, non-exhaustive overview of how the feature works mechanically in the TAF pipeline. 

The use of the feature is driven by the configuration, which includes `promotion_types = [#iree_gpu.use_global_load_dma, ...]` in the lowering configuration for operations.

```mlir
%11 = linalg.generic ... {lowering_config = #iree_gpu.lowering_config<{..., promote_operands = [0, 1], promotion_types = [#iree_gpu.use_global_load_dma, #iree_gpu.use_global_load_dma] ...
```

This annotation is used by the `GPUPromoteMatmulOperands` pass, which inserts `linalg.copy` for each transfer from global memory to LDS. For operands that are promoted with the above promotion type, the `lowering_config = #iree_gpu.use_global_load_dma` is attached to the `linalg.copy`. This pass is shared with the VectorDistribute (VD) pipeline, but the VD pipeline currently does not attach the `promotion_types` in configuration, so this path isn't exercised by the VD pipeline.

```mlir
%14 = linalg.copy {lowering_config = #iree_gpu.use_global_load_dma} ins(...) outs(...) -> tensor<4096x128xf32>
```

[...]

PR #24226, opened 2026-04-23
[IREEGPU] Introduce `async_dma` operation

sommerlukas:
Introduces a new `async_dma` operation that represents an (asynchronous) memory transfer from one memory in the hierarchy to another memory without intermediate staging in registers.

The motivation is to represent the direct-to-LDS feature in AMD GPUs. Compared to the existing `coalesced_gather_dma`, this operation is not as strongly specialized on the representation of subgroup-level operations in thread-level IR in the `TileAndFuse` pipeline.

In the current prototype implementation of direct-to-LDS support in `VectorDistribute`, the operation is used to represent global to LDS data movement at the workgroup level, but the operation itself should be flexible enough for other uses and other levels (e.g., thread).

The operation exists with `tensor` and `memref` semantics. With `tensor` semantics, it returns the destination tensor after data transfer, allowing for synchronization via the `iree_gpu.value_barrier`. Gathering from source is possible, but not scattering to the destination.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
@@ -11,10 +11,12 @@
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Utils/StructuredOpsUtils.h"
+#include "mlir/IR/AffineMap.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinTypeInterfaces.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/OpImplementation.h"
 #include "mlir/IR/ValueRange.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Support/LLVM.h"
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
```diff
@@ -385,4 +387,321 @@ LogicalResult CoalescedGatherDMAOp::verify() {
...
+                         SideEffects::DefaultResource::get());
+  }
+
+  if (isa<MemRefType>(dest.getType())) {
+    effects.emplace_back(MemoryEffects::Write::get(), &getDestMutable(),
+                         SideEffects::DefaultResource::get());
+  }
+}
+
+void AsyncDMAOp::print(OpAsmPrinter &p) {
```
krzysz00: ... why are we doing a custom printer here? I don't see anything that requires it

krzysz00: If there's some small syntax that an assembly format can't support, use a custom<> directive

krzysz00 (changes requested):

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
sommerlukas: The intent was to skip emitting the source index types if they're all `index`, but a full custom printer/parser was out of proportion for that. 

I've condensed it down to declarative format and a small custom printer/parser for that clause.

I felt that omitting the list of `index` types in that case would make IR more readable for the common case, but if you think that it's not worth having, I can also completely drop it.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.cpp
```diff
@@ -10,14 +10,47 @@
...
+  p << "], ";
+}
+
+static ParseResult parseAsyncDMASourceIndexTypes(
+    OpAsmParser &parser, ArrayRef<OpAsmParser::UnresolvedOperand> sourceIndices,
+    SmallVectorImpl<Type> &sourceIndexTypes) {
+  if (failed(parser.parseOptionalLSquare())) {
+    sourceIndexTypes.assign(sourceIndices.size(),
+                            parser.getBuilder().getIndexType());
+    return parser.parseComma();
```
krzysz00: I think you could move the , into the assembly format and use a \`\` directive to skip the space, but that's trivialities

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -361,6 +361,97 @@ def IREEGPU_CoalescedGatherDMAOp : Op<IREEGPU_Dialect, "coalesced_gather_dma", [
...
+    OptionalAttr<AffineMapAttr>:$permutation_map,
+    OptionalAttr<BoolArrayAttr>:$in_bounds
+  );
+
+  let results = (outs Optional<AnyRankedTensorOrMemRef>:$result);
+
+  let assemblyFormat = [{
+    $source `[` $source_indices `]` `to` $dest `[` $dest_indices `]` `,`
+    $transfer_type
+    (`permutation_map` $permutation_map^)?
```
krzysz00: These can be an `oilist` to make things easier on humans

krzysz00 (approved):
I think the op makes sense overall and I don't have any fundamental reason no hold it - we can always fix it later if we don't like the semantics.

lialan (approved):
LGTM.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -361,6 +361,98 @@ def IREEGPU_CoalescedGatherDMAOp : Op<IREEGPU_Dialect, "coalesced_gather_dma", [
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// AsyncDMAOp
+//===----------------------------------------------------------------------===//
+
+def IREEGPU_AsyncDMAOp : Op<IREEGPU_Dialect, "async_dma", [
```
keshavvinayak01, line 368: Looks good overall, just a minor nit: could you add an example operation here?

keshavvinayak01 (approved):

sommerlukas:
The CI failure seems unrelated, will merge now.

PR #24227, opened 2026-04-23
[VectorDistribute] Analysis to propagate promotion information

sommerlukas:
Add a sparse backward dataflow analysis to propagate promotion types from `to_layout` operations on operands to compute ops backward to the origin of the operand, e.g., a read from memory. 

The motivation is to use this analysis from transformation passes to find loads from memory that should be promoted, e.g., using direct-to-LDS and then transform the load accordingly.

The lattice of the analysis has three states: an undefined (initial) state, a defined state with a concrete promotion type stored as attribute and an overdefined (top) state. If two defined lattice values with different promotion types meet, the resulting lattice value is overdefined.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUPromotionAnalysis.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUPromotionAnalysis.cpp
@@ -1,0 +1,217 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Common/GPU/GPUPromotionAnalysis.h"
+
+#include "iree/compiler/Codegen/Common/GPU/Passes.h"
+#include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.h"
+#include "mlir/Analysis/DataFlow/SparseAnalysis.h"
+#include "mlir/Analysis/DataFlow/Utils.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/SymbolTable.h"
+
+namespace mlir::iree_compiler {
+
+using namespace mlir::dataflow;
+
+//===----------------------------------------------------------------------===//
+// Lattice value
+//===----------------------------------------------------------------------===//
+
+struct PromotionTypeLatticeValue {
+  enum class State { None, Concrete, Overdefined };
+  State state = State::None;
+  Attribute promotionType; // valid when state == Concrete
+
+  PromotionTypeLatticeValue() = default;
+  explicit PromotionTypeLatticeValue(Attribute type)
+      : state(State::Concrete), promotionType(type) {}
+
+  bool isDefined() const { return state == State::Concrete; }
+  bool isOverdefined() const { return state == State::Overdefined; }
+
+  bool operator==(const PromotionTypeLatticeValue &rhs) const {
+    return state == rhs.state && promotionType == rhs.promotionType;
+  }
+
+  static PromotionTypeLatticeValue meet(const PromotionTypeLatticeValue &lhs,
+                                        const PromotionTypeLatticeValue &rhs) {
+    if (lhs.isOverdefined() || rhs.isOverdefined()) {
+      return getOverdefined();
+    }
+    if (!lhs.isDefined()) {
+      return rhs;
+    }
+    if (!rhs.isDefined()) {
+      return lhs;
+    }
+    // Both concrete.
+    if (lhs.promotionType == rhs.promotionType) {
+      return lhs;
+    }
+    return getOverdefined();
+  }
+
+  // Required by Lattice<T> but unused in backward analysis.
+  static PromotionTypeLatticeValue join(const PromotionTypeLatticeValue &lhs,
+                                        const PromotionTypeLatticeValue &rhs) {
+    return meet(lhs, rhs);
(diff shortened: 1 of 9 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/GPUPromotionAnalysis.h
```diff
@@ -0,0 +1,36 @@
...
+/// Discardable attribute name on `to_layout` ops that carries the promotion
+/// type (e.g., UseGlobalLoadDMAAttr).
+inline constexpr llvm::StringLiteral kPromotionTypeAttr =
+    "iree_gpu.promotion_type";
+
+/// Result of promotion type analysis. Maps SSA values to their inferred
+/// promotion type attribute (e.g., UseGlobalLoadDMAAttr). Only values with a
+/// concrete (non-overdefined) promotion type are included.
+using PromotionTypeMap = llvm::DenseMap<Value, Attribute>;
+
```
keshavvinayak01, line 20: Could we type this against `IREE::GPU::PromotionAttr`?

That rejects malformed attributes at the anchor and pevents `PromotionTypeMap` carrying a generic attribute.
```suggestion
using PromotionTypeMap = llvm::DenseMap<Value, IREE::GPU::PromotionAttr>;
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUPromotionAnalysis.cpp
```diff
@@ -0,0 +1,217 @@
...
+//===----------------------------------------------------------------------===//
+
+struct PromotionTypeLatticeValue {
+  enum class State { None, Concrete, Overdefined };
+  State state = State::None;
+  Attribute promotionType; // valid when state == Concrete
+
+  PromotionTypeLatticeValue() = default;
+  explicit PromotionTypeLatticeValue(Attribute type)
+      : state(State::Concrete), promotionType(type) {}
```
keshavvinayak01, line 32:
```suggestion
struct PromotionTypeLatticeValue {
  enum class State { None, Concrete, Overdefined };
  State state = State::None;
  IREE::GPU::PromotionAttr promotionType;

  PromotionTypeLatticeValue() = default;
  explicit PromotionTypeLatticeValue(IREE::GPU::PromotionAttr type)
     : state(State::Concrete), promotionType(type) {}
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUPromotionAnalysis.cpp
```diff
@@ -0,0 +1,217 @@
...
+    // Exit state = None (no promotion type known at exits).
+    propagateIfChanged(lattice, lattice->meet(PromotionTypeLatticeValue()));
+  }
+
+  LogicalResult
+  visitOperation(Operation *op, ArrayRef<PromotionTypeLattice *> operands,
+                 ArrayRef<const PromotionTypeLattice *> results) override {
+    // Anchor: to_layout with promotion_type discardable attribute.
+    if (auto toLayout = dyn_cast<IREE::VectorExt::ToLayoutOp>(op)) {
+      if (Attribute pt = toLayout->getAttr(kPromotionTypeAttr)) {
```
keshavvinayak01:
```suggestion
      if (auto pt = toLayout->getAttrOfType<IREE::GPU::PromotionAttr>(
              kPromotionTypeAttr)) {
```

keshavvinayak01:
Or maybe we could add a discardable attribute like `"IREE::GPU::PromotionAttr":$promotion_type` at. https://github.com/iree-org/iree/blob/7734a70b44219470a4d905a5990b427936acffc2/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUDialect.td#L33

sommerlukas (commented):
> Or maybe we could add a discardable attribute like `"IREE::GPU::PromotionAttr":$promotion_type` at.
> 
> https://github.com/iree-org/iree/blob/7734a70b44219470a4d905a5990b427936acffc2/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUDialect.td#L33

I've implemented a discardable attribute on the IREEGPU dialect and the analysis now uses the corresponding helper and concrete `PromotionAttr`.

keshavvinayak01 (approved):
LGTM, but I haven't worked with DFA in IREE before, maybe wait for @Groverkss' stamp too.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUPromotionAnalysis.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+  void setToExitState(PromotionTypeLattice *lattice) override {
+    // Exit state = None (no promotion type known at exits).
+    propagateIfChanged(lattice, lattice->meet(PromotionTypeLatticeValue()));
+  }
+
+  LogicalResult
+  visitOperation(Operation *op, ArrayRef<PromotionTypeLattice *> operands,
+                 ArrayRef<const PromotionTypeLattice *> results) override {
+    // Anchor: to_layout with promotion_type discardable attribute.
```
Groverkss: Instead of using discardable attributes, can we actually change the "shared_memory_conversion" attribute to be "promotion_type" or something like that instead?

Groverkss (commented):
lgtm, only thing is I would prefer we put the attribute on the op instead of putting a discardable attribute because we already have an attribute on the to_layout op

keshavvinayak01 (approved):

sommerlukas:
> lgtm, only thing is I would prefer we put the attribute on the op instead of putting a discardable attribute because we already have an attribute on the to_layout op

Good idea with repurposing the existing attribute, I've made the necessary changes to the attribute definition in https://github.com/iree-org/iree/pull/24377 and will rebase this PR on top of that once it's landed.

Groverkss (approved):

PR #24231, opened 2026-04-23
[HAL] Runtime Correctness Foundations

benvanik:
This PR tightens a set of runtime correctness paths that the later HAL work
leans on. The changes are deliberately small and independent: less lock traffic
in the empty atomic-slist fast path, safer file-backed imports, topology probing
that behaves on large systems, task teardown that tolerates partially
initialized executors, CUDA event cleanup on failure, and Vulkan physical-device
enumeration cleanup.

## Why

The later stack puts more stress on startup, device probing, queue setup, and
failure unwinding. These paths should be boring before we add new allocation,
profiling, replay, and AMDGPU machinery on top of them. The theme here is
removing avoidable hazards instead of letting them become debugging noise in
larger PRs.

## What's Here

The atomic slist pop path now avoids taking the lock when the list is already
empty. That keeps the uncontended empty fast path cheap and makes the
benchmark's ownership model explicit enough to avoid reusing intrusive entries
across concurrent producers.

File-backed import paths handle unaligned memory-file imports, cap fd-file
pread/pwrite chunk sizes, and fall back to synchronous fd import when async
import is unavailable.

Task and topology setup is more robust on machines with more than 1024 CPUs and
when optional sysfs/NUMA probes fail. Optional topology discovery now uses
ordinary values for optional outcomes instead of manufacturing expensive
statuses for expected absence.

CUDA and Vulkan touched failure paths now release acquired resources instead of
leaking on early failure.

ci-extra: all

### DIFF
--- a/runtime/src/iree/base/internal/BUILD.bazel
+++ b/runtime/src/iree/base/internal/BUILD.bazel
@@ -148,6 +148,16 @@ iree_runtime_cc_test(
     ],
 )
 
+cc_binary_benchmark(
+    name = "atomic_slist_benchmark",
+    srcs = ["atomic_slist_benchmark.cc"],
+    deps = [
+        ":atomic_slist",
+        "//runtime/src/iree/testing:benchmark_main",
+        "@com_google_benchmark//:benchmark",
+    ],
+)
+
 iree_runtime_cc_library(
     name = "atomic_freelist",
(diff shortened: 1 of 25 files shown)


benvanik:
Signals are looking good, but given that it'd take me a 40hr week to get things through these slow runs I'm going to land and fix forward anything those slow bots find.

PR #24233, opened 2026-04-23
[DispatchCreation] Block transpose fusion for OnlineAttention

IanWood1:
`wouldBroadcastAwayAttentionDimGroup` only checked `AttentionOp`, so the `foldTransposeControlFn` would allow a broadcasting producer to be fused into `OnlineAttentionOp` even when it drops a whole dim group (batch/M/K1/K2/N). Handle `OnlineAttentionOp` the same way as `AttentionOp`.

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/ElementwiseOpFusion.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/ElementwiseOpFusion.cpp
@@ -50,11 +50,21 @@ static SmallVector<T> applyProjectedPermutation(const SmallVectorImpl<T> &input,
 }
 
 static bool wouldBroadcastAwayAttentionDimGroup(OpOperand *fusedOperand) {
-  auto attentionOp =
-      dyn_cast<IREE::LinalgExt::AttentionOp>(fusedOperand->getOwner());
-  if (!attentionOp) {
+  Operation *consumer = fusedOperand->getOwner();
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/DispatchCreation/ElementwiseOpFusion.cpp
```diff
@@ -50,32 +50,35 @@ static SmallVector<T> applyProjectedPermutation(const SmallVectorImpl<T> &input,
 }
 
 static bool wouldBroadcastAwayAttentionDimGroup(OpOperand *fusedOperand) {
-  auto attentionOp =
-      dyn_cast<IREE::LinalgExt::AttentionOp>(fusedOperand->getOwner());
-  if (!attentionOp) {
```
keshavvinayak01, line 55: Do we not need to do something `if(!isa<AttentionOp, OnlineAttentionOp>(op))` here?

keshavvinayak01 (approved):
Looks good, just one check.

compiler/src/iree/compiler/DispatchCreation/ElementwiseOpFusion.cpp
IanWood1: No, the TypeSwitch will return failure when its not one of the two attention ops.

MaheshRavishankar (approved):

Issue #24221, opened 2026-04-22
[SDPA] iree-compile fails to distribute SDPA on rocm/gfx942


rkayaith:
`iree-compile` fail to compile with `'func.func' op failed to distribute` for `torch.aten.scaled_dot_product_attention` on rocm/gfx942 for several Q/K/V shapes, all of which surface in PyTorch's `test/test_transformers.py` once SDPA is dispatched to hipDNN/fusilli.

## Repro command

```bash
iree-compile --iree-hal-target-backends=rocm --iree-rocm-target=mi300x 
```

These all fail with the same error:

```
<file>.mlir:14:10: error: 'func.func' op failed to distribute
    %0 = torch.aten.scaled_dot_product_attention %arg0, %arg1, %arg2,
         ^
```

## IRs

<details>
<summary><code>test_transformerencoder_square_input (Q, K, V: [B=2, H=2, S=2, D=2])</code></summary>

PyTorch test: https://github.com/pytorch/pytorch/blob/6af13031e55a1f824b4cabe23b733e8ad4e9a24c/test/test_transformers.py#L591

```mlir

module @module {
  
  func.func private @sdpa_fprop(
      %arg0: !torch.vtensor<[2,2,2,2],f16>,
      %arg1: !torch.vtensor<[2,2,2,2],f16>,
      %arg2: !torch.vtensor<[2,2,2,2],f16>)
      -> !torch.vtensor<[2,2,2,2],f16> {
    %none_mask = torch.constant.none
    %dropout = torch.constant.float 0.000000e+00
    %is_causal = torch.constant.bool true
    %scale = torch.constant.float 7.071068e-01
    %enable_gqa = torch.constant.bool false
    %0 = torch.aten.scaled_dot_product_attention %arg0, %arg1, %arg2,
        %none_mask, %dropout, %is_causal, %scale, %enable_gqa :
        !torch.vtensor<[2,2,2,2],f16>, !torch.vtensor<[2,2,2,2],f16>,
        !torch.vtensor<[2,2,2,2],f16>, !torch.none, !torch.float, !torch.bool,
        !torch.float, !torch.bool -> !torch.vtensor<[2,2,2,2],f16>
    return %0 : !torch.vtensor<[2,2,2,2],f16>
  }

[...]

PR #24239, opened 2026-04-24
[LLVMGPU] Fall back to scalar lowering for tiny attention shapes

keshavvinayak01:
The attention `VectorDistribute` configs (both the MMA-intrinsic path and the memory-bound reduction path) assume head dimensions and K2 reach a certain size. For shapes below that threshold (e.g. Q=K=V=[2,2,2,2] f16), the reduction path still succeeds at emitting a `VectorDistribute` config, but the tile sizes it picks produce vector ops whose shapes the layout engine cannot support, causing the failure in https://github.com/iree-org/iree/issues/24221

Add early bailouts for the shapes that cannot be tiled cleanly.


compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1113,6 +1113,29 @@ static LogicalResult setAttentionReductionConfig(
...
+           << seeds.keyVectorSize << "): " << k1Size;
+    return failure();
+  }
+  if (!ShapedType::isDynamic(nSize) && nSize % nWorkgroupTile != 0) {
+    LDBG() << "Bailing out: N not a multiple of value workgroup tile ("
+           << nWorkgroupTile << "): " << nSize;
+    return failure();
+  }
+  if (!ShapedType::isDynamic(k2Size) && k2Size <= kVerySkinnyDimThreshold) {
+    LDBG() << "Bailing out due to skinny K2 dimension: " << k2Size;
```
sommerlukas: Nit, to reflect the name of the threshold:
```suggestion
    LDBG() << "Bailing out due to very skinny K2 dimension: " << k2Size;
```

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_vector_distribute_reduction_gfx942.mlir
```diff
@@ -603,3 +603,92 @@ func.func @reordered_post_reduction_consumer() {
...
+  } -> tensor<2x64x64xf16>, tensor<2x64xf16>, tensor<2x64xf16>
+  iree_tensor_ext.dispatch.tensor.store %att#0, %3, offsets = [0, 0, 0], sizes = [2, 64, 64], strides = [1, 1, 1] : tensor<2x64x64xf16> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<2x64x64xf16>>
+  return
+}
+
+// -----
+
+// K1=32 and K2=16 are friendly, but N=24 is >= valueVectorSize yet not a
+// multiple of the workgroup tile (numValueVectors * valueVectorSize = 16 for
+// f16).
```
sommerlukas: I'm working on enabling this shape (`N=24`) to go through the intrinsic path. We can keep this test for now, I'd probably remove it then. 

If you think we should have more than one test for the fallback configuration, might be worth adding a different one here.

sommerlukas (approved):
LGTM, just a tiny nit and a future note on testing.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_vector_distribute_reduction_gfx942.mlir
keshavvinayak01: What is the final heuristic you decided qualifies for MFMA intrinsic allocation? You can change this test accordingly to reflect that?

sommerlukas: I'm using the existing `canTargetIntrinsic`, I just changed the `mustBeAligned` flag to allow non-exact matches: https://github.com/iree-org/iree/blob/c66d02930366b23b1b7f71b0770dba184e173d9e/compiler/src/iree/compiler/Codegen/Common/GPU/GPUHeuristics.cpp#L263

I can update this test when I land my changes, no problem. The point I wanted to make here was a different one: If you want to test fallback to scalar in multiple cases, we might need an additional shape here that won't go to intrinsics in the end. But I'm fine with the tests as they are, so we can just keep it as is.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1113,6 +1113,29 @@ static LogicalResult setAttentionReductionConfig(
...
+  if (!ShapedType::isDynamic(k1Size) && k1Size % seeds.keyVectorSize != 0) {
+    LDBG() << "Bailing out: K1 not a multiple of key vector size ("
+           << seeds.keyVectorSize << "): " << k1Size;
+    return failure();
+  }
+  if (!ShapedType::isDynamic(nSize) && nSize % nWorkgroupTile != 0) {
+    LDBG() << "Bailing out: N not a multiple of value workgroup tile ("
+           << nWorkgroupTile << "): " << nSize;
+    return failure();
+  }
```
Groverkss, line 1130: Why are we checking if it is a multiple? I don't understand it from the comment.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1113,6 +1113,29 @@ static LogicalResult setAttentionReductionConfig(
           op.getQueryMap(), op.getKeyMap(), op.getValueMap(), op.getOutputMap())
           .value();
 
+  // Bail out on shapes this path cannot tile cleanly. K2 below the skinny-dim
+  // threshold leaves attention as a large elementwise; the intrinsic and
+  // reduction paths have nothing to offer. All three cases fall through to the
+  // BaseLowering scalar fallback.
+  int64_t k1Size = bounds[opInfo.getK1Dims().back()];
+  int64_t nSize = bounds[opInfo.getNDims().back()];
+  int64_t k2Size = bounds[opInfo.getK2Dims().back()];
```
Groverkss: It's possible (And it happens) that the dimension doesn't even exist. Use 1 instead of .back.

Groverkss: It's possible (And it happens) that the dimension doesn't even exist. Use 1 instead of .back.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_vector_distribute_reduction_gfx942.mlir
keshavvinayak01: Okay, in that case, I'll just remove this test.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
keshavvinayak01: Added better doc comments. I was referencing Line 1182: `threadTileSizes[opInfo.getK1Dims().back()] = seeds.keyVectorSize;` that warranted this check.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1113,6 +1113,38 @@ static LogicalResult setAttentionReductionConfig(
...
           .value();
 
+  // Bail out on shapes this path cannot tile cleanly; all three cases fall
+  // through to the BaseLowering scalar fallback:
+  //   * K1 must be a multiple of `keyVectorSize`. The per-thread K1 tile is
+  //     pinned to that size below, so a non-multiple leaves a partial vector
+  //     and produces a vector op shape the layout engine cannot distribute.
+  //   * N must be a multiple of `numValueVectors * valueVectorSize`. The
+  //     inner-N workgroup tile factors that product, so a non-multiple lands
+  //     a tile boundary mid-vector and likewise breaks layout inference.
```
Groverkss: Why won't it be masked? @sommerlukas is this correct?

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1428,6 +1460,21 @@ setAttentionVectorDistributionConfig(IREE::GPU::TargetAttr target,
...
+setAttentionBaseLoweringConfig(FunctionOpInterface entryPoint,
+                               IREE::LinalgExt::OnlineAttentionOp op) {
+  MLIRContext *context = op.getContext();
+  SmallVector<int64_t, 3> workgroupSize = {1, 1, 1};
+  auto translationInfo =
+      getGPUTranslationInfo(context, GPUPipeline::BaseLowering, workgroupSize);
+  return setOpConfigAndEntryPointFnTranslation(
+      entryPoint, op, IREE::Codegen::LoweringConfigAttrInterface(),
+      translationInfo);
+}
```
Groverkss: We already have a fallback lowering for all ops right? Do we need this? Why isn't "setTileAndFuseLoweringConfig" enough for this?

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
sommerlukas: I'm working on an example of a non-aligned shape for `K1` (e.g., 24). I had to fix some issues, but after that we should be able to support that through VectorDistribute.

keshavvinayak01: I'll keep this check along with a TODO until @sommerlukas lands that fix.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
keshavvinayak01: `setTileAndFuseLoweringConfig`  seemed enough for config selection. the missing piece was that `OnlineAttentionOp` did not lower correctly through the pipeline; added that fix.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -867,7 +867,13 @@ void addGPUVectorDistributePassPipeline(OpPassManager &funcPassManager,
...
 
 void addGPUSimpleDistributePassPipeline(OpPassManager &funcPassManager) {
-  tileAndBufferize(funcPassManager);
+  tileAndDistributeToWorkgroup(funcPassManager);
+  // OnlineAttentionOp has no bufferization path of its own. Decompose only on
+  // the generic Distribute fallback path after workgroup tiling.
+  funcPassManager.addPass(IREE::LinalgExt::createDecomposeAttentionPass());
+  funcPassManager.addPass(createConfigTrackingCanonicalizerPass());
+  funcPassManager.addPass(createCSEPass());
+  addBufferizePasses(funcPassManager);
```
sommerlukas, line 876: Is the combination of `tileAndDistributeToWorkgroup` and `addBufferizePasses` the same as `tileAndBufferize`?

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
```diff
@@ -1104,6 +1113,33 @@ static LogicalResult setAttentionReductionConfig(
...
+           << nWorkgroupTile << "): " << nSize;
+    return failure();
+  }
+
+  // Bail out on very skinny K2 shapes; smaller K2 collapses the reduction below
+  // what this path is tuned for.
+  // K2 may be empty; treat it as size 1, which trips the skinny-K2 check
+  // below.
+  int64_t k2Size =
+      opInfo.getK2Dims().empty() ? 1 : bounds[opInfo.getK2Dims().back()];
```
sommerlukas, line 1137: Out of curiosity, under which circumstances is `K2` empty?

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_vector_distribute_reduction_gfx942.mlir
```diff
@@ -584,3 +584,59 @@ hal.executable private @attention_4xDx1x32x128xf16 {
...
+#distribute_attention_translation = #iree_codegen.translation_info<
+  pipeline = #iree_gpu.pipeline<Distribute> workgroup_size = [128, 1, 1] subgroup_size = 64>
+#distribute_attention_layout = #hal.pipeline.layout<bindings = [
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>
+]>
+
+hal.executable private @attention_skinny_k2_distribute {
```
sommerlukas: Is this the right file for this test? AFAICT, this doesn't use the VectorDistribute pipeline.

sommerlukas (approved):
Just a few minor question.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
keshavvinayak01: Yes, https://github.com/iree-org/iree/blob/0380544b17da75c1e757e96bd747aa326fe3675c/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L267

I split it so `createDecomposeAttentionPass()` can run between workgroup tiling and bufferization.

compiler/src/iree/compiler/Codegen/LLVMGPU/KernelConfig.cpp
keshavvinayak01: From what I can see:
https://github.com/iree-org/iree/blob/0380544b17da75c1e757e96bd747aa326fe3675c/compiler/src/iree/compiler/Dialect/LinalgExt/Utils/IndexingUtils.cpp#L53-#L56

Hypothetically this can be 1? @Groverkss pointed this out earlier.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_vector_distribute_reduction_gfx942.mlir
keshavvinayak01: Moved it into a new file.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_distribute.mlir
sommerlukas, line 1: Don't we need to add this test to the corresponding build files?

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_vector_distribute_reduction_gfx942.mlir
```diff
@@ -584,3 +584,4 @@ hal.executable private @attention_4xDx1x32x128xf16 {
 //
 // Warning: the above layout_config for vector distribution on attention goes overboard on the tail part.
 // CHECK:               gpu.subgroup_reduce {{.*}} : (f32) -> f32
+
```
sommerlukas: I don't think we need this whitespace change.

sommerlukas (commented):
Implementation looks fine, I think moving the test caused some issues.

keshavvinayak01:
There's a missed failing CI test at `iree/tests/e2e/linalg_ext_ops/check_rocm_hip_attention.mlir`; I'm currently working on fixing it.

keshavvinayak01 (commented):
@Groverkss @sommerlukas

keshavvinayak01 (commented):
I narrowed the problem to the attention shape failing because of a data race.

Forced generic Distribute (for the failing f32 shape) decomposes attention into scalar shared-memory reduction/update code, but the dispatch has `workgroup_size = [128,1,1]`. Multiple workgroup threads can execute the same load/add/store sequence to the same shared-memory accumulator without atomics or single-thread ownership.

The revert is done, but to re-land this PR, we need to make some decision:

- Serialize the scalar fallback with `workgroup_size = [1,1,1]` OR
- Fix `DecomposeAttentionPass/SimpleDistribute` so decomposed reductions have explicit thread ownership or cooperative reductions (might be complex)

PR #24241, opened 2026-04-24
[Codegen][DMA] Fix unaligned swizzle offset computation in gather-to-lds lowering

Yu-Zhewen:
The inverse XOR swizzle applied to DMA source offsets was incorrect in two cases:

1. **Subgroup base offset**: When a subgroup's transfer size is not a multiple of the swizzle period, different subgroups sharing the same local offsets but occupying different rows would get identical swizzled addresses. 

Fix: incorporate the subgroup's base offset within the full allocation before swizzling.

2. **Access-width alignment**: When `elementsPerLane < accessWidth`, the integer division inside `swizzleOffset` truncates offsets that differ only within an access-width group to the same value. 

Fix: strip the sub-accesswidth remainder before swizzling and restore it after. This fix is applied directly in `swizzleOffset` for both XOR and rotate_rows swizzles. While rotate_rows isn't currently used with DMA, the access-width alignment issue affects both swizzle types.

Both issues caused numerical mismatches for BF16 batch matmuls using DMA with XOR swizzle enabled.

Assisted-by: Cursor (Claude)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -543,10 +543,8 @@ private:
           // Apply inverse source swizzle when destination has XOR swizzle.
           // XOR swizzle is self-inverse, so swizzle(swizzle(x)) = x.
           if (destSwizzle) {
-            int64_t destElements = ShapedType::getNumElements(destShape);
             srcLinearOffset = applyInverseXorSwizzleToDMASourceOffset(
-                rewriter, loc, srcLinearOffset, *destSwizzle, destElements,
-                dest);
+                rewriter, loc, srcLinearOffset, *destSwizzle, dest);
           }
           auto srcDelinearize = affine::AffineDelinearizeIndexOp::create(
               rewriter, loc, srcLinearOffset, basis, /*hasOuterBound=*/true);
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -546,7 +546,7 @@ private:
             int64_t destElements = ShapedType::getNumElements(destShape);
             srcLinearOffset = applyInverseXorSwizzleToDMASourceOffset(
                 rewriter, loc, srcLinearOffset, *destSwizzle, destElements,
-                segment.elementsPerLane, dest);
+                dest);
           }
           auto srcDelinearize = affine::AffineDelinearizeIndexOp::create(
               rewriter, loc, srcLinearOffset, basis, /*hasOuterBound=*/true);
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
@@ -233,109 +233,6 @@ getDestSwizzleAttr(Value dest) {
   return std::nullopt;
 }
 
-/// Trace a memref value through view-like ops, accumulating the linear
-/// element offset. Returns nullptr if the offset is zero.
-static Value computeSwizzleBaseOffset(RewriterBase &rewriter, Location loc,
-                                      Value dest) {
-  OpFoldResult totalOffset = rewriter.getIndexAttr(0);
-  while (auto viewOp = dest.getDefiningOp<ViewLikeOpInterface>()) {
-    if (auto subviewOp = dyn_cast<memref::SubViewOp>(viewOp.getOperation())) {
-      auto parentType = cast<MemRefType>(subviewOp.getSource().getType());
-      SmallVector<int64_t> strides;
-      int64_t offset;
-      if (succeeded(parentType.getStridesAndOffset(strides, offset))) {
-        SmallVector<OpFoldResult> strideOFRs =
-            getAsIndexOpFoldResult(rewriter.getContext(), strides);
-        auto &&[expr, values] = computeLinearIndex(totalOffset, strideOFRs,
-                                                   subviewOp.getMixedOffsets());
-        totalOffset =
-            affine::makeComposedFoldedAffineApply(rewriter, loc, expr, values);
-      }
-    }
-    dest = viewOp.getViewSource();
-  }
-  if (isConstantIntValue(totalOffset, 0)) {
-    return nullptr;
-  }
-  return getValueOrCreateConstantIndexOp(rewriter, loc, totalOffset);
-}
-
-/// Apply inverse XOR swizzle to a sub-tile-local source offset so that the
-/// DMA write-side permutation matches the read-side (ResolveSwizzleHints).
-///
-/// Two adjustments are needed beyond the swizzle itself:
-///
-/// 1. Subgroup base offset: when the subgroup transfer size is not a
-///    multiple of the swizzle period, we add the subgroup's base offset
-///    within the full allocation before swizzling and subtract it after.
-///
-///    Example: xor_shuffle<64, 8>, period = 64*64/8 = 512 elements.
-///    Workgroup tile = 32x32, 4 subgroups of 8x32 = 256 each.
-///    256 < 512, so the fix is needed:
-///    ```
-///      BUG: swizzle(local)
-///           Subgroups 0 and 1 see the same local offsets but occupy
-///           different rows in the full allocation.
-///      FIX: swizzle(local + base) - base
-///    ```
-///
-/// 2. Access-width alignment: when elementsPerLane < accessWidth, we
-///    strip the sub-accessWidth remainder before swizzling and add it
-///    back after.
-///
-///    Example: xor_shuffle<64, 8>, elementsPerLane = 2 (< 8).
-///    ```
-///      BUG: swizzle(offset)
-///           swizzle(0) == swizzle(2) == swizzle(6) because swizzle
-///           divides by accessWidth, truncating all three to the same group.
-///      FIX: swizzle(offset - rem) + rem, where rem = offset % accessWidth
-///    ```
-static Value applySwizzleToSourceOffset(RewriterBase &rewriter, Location loc,
-                                        Value srcLinearOffset,
-                                        IREE::Codegen::XORShuffleAttr swizzle,
-                                        int64_t destElements,
-                                        int64_t elementsPerLane, Value dest) {
-  // Add the subgroup's base offset within the full allocation.
-  Value swizzleBaseOffset;
-  int64_t swizzlePeriod = swizzle.getRowWidth() *
-                          (swizzle.getRowWidth() / swizzle.getAccessWidth());
-  if (destElements % swizzlePeriod != 0) {
-    swizzleBaseOffset = computeSwizzleBaseOffset(rewriter, loc, dest);
-  }
-  Value swizzleInput = srcLinearOffset;
-  if (swizzleBaseOffset) {
-    swizzleInput = arith::AddIOp::create(rewriter, loc, swizzleBaseOffset,
-                                         srcLinearOffset);
-  }
-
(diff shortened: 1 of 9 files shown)


compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -149,23 +149,126 @@ computeTransferSegments(int64_t totalElements, int64_t elementBits,
...
+        arith::ConstantIndexOp::create(rewriter, loc, accessWidth);
+    remainder =
+        arith::RemUIOp::create(rewriter, loc, swizzleInput, accessWidthVal);
+    swizzleInput =
+        arith::SubIOp::create(rewriter, loc, swizzleInput, remainder);
+  }
+
+  // Apply the actual swizzle.
+  Value swizzled = getValueOrCreateConstantIndexOp(
+      rewriter, loc, swizzle.swizzleOffset(rewriter, loc, swizzleInput, dest));
```
sommerlukas: What is the rationale to not integrate these fixes into `swizzleOffset`? If we added it to the interface implementation, other pipelines that might need the same logic would benefit directly.

sommerlukas (commented):
Implementation looks OK to me (would be good to get additional eyes with experience on matmul/TileAndFuse), but I wonder if putting it in the TileAndFuse-specific pass is the right place.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
Yu-Zhewen: I think swizzleOffset shouldn't own any DMA-specific logic, and it should remain a pure flat linear offset permutation. That said, I agree these fixes could be moved to a more accessible place, though I'm not sure where the best place for it is yet. Perhaps as a shared helper in GPUUtils.cpp?

sommerlukas: Yes, that would probably be a good place 👍

Yu-Zhewen: Done

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1744,3 +1744,105 @@ func.func @lower_dma_with_incompatible_swizzle(
...
+    // CHECK: arith.subi {{.*}}, %[[REM]]
+    // CHECK: arith.xori
+    // CHECK: arith.addi {{.*}}, %[[REM]]
+    // CHECK: amdgpu.gather_to_lds
+    iree_gpu.coalesced_gather_dma %source into %dest lane(%lane)
+        : memref<8x64xbf16, #amdgpu.address_space<fat_raw_buffer>>,
+          memref<8x64xbf16, #gpu.address_space<workgroup>>, index
+  } {mapping = [#iree_gpu.lane_id<0>]}
+  return
+}
```
sommerlukas, line 1920: Should we have another test that checks the combination of both cases?

sommerlukas (commented):
Implementation looks good to me, but probably worth to get eyes from someone with more experience with swizzles for matmul on this PR (e.g., @lialan or @jerryyin).

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
Yu-Zhewen: After further discussion with @lialan and @jerryyin , I moved the access-width alignment fix (fix 2) into swizzleOffset itself. You were right, this fix isn't DMA-specific logic.

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
Yu-Zhewen: Done

compiler/src/iree/compiler/Codegen/Common/GPU/test/amdgpu_lower_coalesced_dma_to_gather_lds.mlir
```diff
@@ -1682,7 +1682,7 @@ func.func @lower_dma_with_dest_swizzle(
     hal.executable.target = #executable_target_rocm_hsaco_fb_swizzle,
     translation_info = #translation_swizzle} {
   %alloc = memref.alloc() : memref<512xf32, #gpu.address_space<workgroup>>
-  %swizzled = iree_codegen.swizzle_hint %alloc[#iree_codegen.xor_shuffle<128, 16>]
+  %swizzled = iree_codegen.swizzle_hint %alloc[#iree_codegen.xor_shuffle<128, 4>]
```
Yu-Zhewen, line 1685: This test was updated from `xor_shuffle<128, 16>` to `xor_shuffle<128, 4>` to show the aligned case: since `elementsPerLane = 4` matches `accessWidth = 4`, the muli  and remui will cancel out. A separate test (`lower_dma_swizzle_access_width_align`) was added to cover the unaligned case where `elementsPerLane < accessWidth`.

compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp
```diff
@@ -543,10 +543,10 @@ struct LowerCoalescedGatherDMAPattern final
           // Apply inverse source swizzle when destination has XOR swizzle.
           // XOR swizzle is self-inverse, so swizzle(swizzle(x)) = x.
           if (destSwizzle) {
-            srcLinearOffset = getValueOrCreateConstantIndexOp(
-                rewriter, loc,
-                destSwizzle->swizzleOffset(rewriter, loc, srcLinearOffset,
-                                           dest));
+            int64_t destElements = ShapedType::getNumElements(destShape);
```
lialan: make a guard here in case of dynamic dim?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -1367,4 +1370,55 @@ SmallVector<Operation *> getTunerRootOps(mlir::ModuleOp moduleOp) {
   return rootOps;
 }
 
+Value applyInverseXorSwizzleToDMASourceOffset(
+    OpBuilder &builder, Location loc, Value srcLinearOffset,
+    IREE::Codegen::XORShuffleAttr swizzle, int64_t destElements, Value dest) {
+  // Compute the subgroup's base offset within the full allocation by tracing
+  // through view-like ops and accumulating the linear element offset.
+  Value swizzleBaseOffset;
+  int64_t swizzlePeriod = swizzle.getRowWidth() *
```
lialan: suggestion: this should go into `XORShuffleAttr`

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -1367,4 +1370,55 @@ SmallVector<Operation *> getTunerRootOps(mlir::ModuleOp moduleOp) {
...
+          SmallVector<OpFoldResult> strideOFRs =
+              getAsIndexOpFoldResult(builder.getContext(), strides);
+          auto &&[expr, values] = computeLinearIndex(
+              totalOffset, strideOFRs, subviewOp.getMixedOffsets());
+          totalOffset =
+              affine::makeComposedFoldedAffineApply(builder, loc, expr, values);
+        }
+      }
+      destTrace = viewOp.getViewSource();
+    }
```
lialan: only handles `memref.subvew`, so it might fail. Maybe set an assert if it is not a `memref.subview` so we know we hit a case unexpected.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -1367,4 +1370,55 @@ SmallVector<Operation *> getTunerRootOps(mlir::ModuleOp moduleOp) {
...
 
+Value applyInverseXorSwizzleToDMASourceOffset(
+    OpBuilder &builder, Location loc, Value srcLinearOffset,
+    IREE::Codegen::XORShuffleAttr swizzle, int64_t destElements, Value dest) {
+  // Compute the subgroup's base offset within the full allocation by tracing
+  // through view-like ops and accumulating the linear element offset.
+  Value swizzleBaseOffset;
+  int64_t swizzlePeriod = swizzle.getRowWidth() *
+                          (swizzle.getRowWidth() / swizzle.getAccessWidth());
+  if (destElements % swizzlePeriod != 0) {
```
lialan: This assumes that when dest size divides cleanly, the subgroup base must also be aligned to the period. It might be true, but it would be great to have an assert to ensure`subgroupBaseOffset % swizzlePeriod == 0` here.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
```diff
@@ -676,11 +676,16 @@ OpFoldResult RotateRowsAttr::swizzleOffset(OpBuilder &b, Location loc,
...
   Value idVal = getValueOrCreateConstantIndexOp(b, loc, id);
+  Value remainder = arith::RemUIOp::create(b, loc, idVal, accessWidthVal);
+  Value alignedId = arith::SubIOp::create(b, loc, idVal, remainder);
+
   // i = row # = |offset| floordiv |Num elements per row|
-  Value i = arith::DivUIOp::create(b, loc, idVal, rowAlignmentVal);
+  Value i = arith::DivUIOp::create(b, loc, alignedId, rowAlignmentVal);
   // jByte = element column # = |offset| % |Num elements per row|
-  Value jElem = arith::RemUIOp::create(b, loc, idVal, rowAlignmentVal);
+  Value jElem = arith::RemUIOp::create(b, loc, alignedId, rowAlignmentVal);
```
krzysz00, line 686: Seems a bit weird that te're not xor-ing anything for the xor swizzle, unless that's hidden by the diff

Yu-Zhewen: Ah, the diff you are looking at is the rotate_rows swizzle. The XOR swizzle function is further down, with arith::XOrIOp at line 822. Fix 2 in the PR description (access-width alignment) is applied to both swizzle types since they share the same access-width granularity issue.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Yu-Zhewen: Assertion added

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Yu-Zhewen: Make sense. Done

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Yu-Zhewen: Replacing the `destElements % swizzlePeriod` check entirely with `subgroupBaseOffset % swizzlePeriod`

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.cpp
Yu-Zhewen: Also updated in PR description to make it more clear

krzysz00: ... I'm so good at reading things

krzysz00 (approved):
Makes sense to me but I'd wait for @lialan

lialan (approved):
+1, no major issues from me.

PR #24243, opened 2026-04-24
[Codegen] Improve constant handling in VectorLayoutAnalysis

sommerlukas:
When determining whether a chain of operations is cheap to clone in VectorLayoutAnalysis, allow operations from other blocks in the chain as long as the operations are constant-like. 

The motivation is that constants are usually hoisted to the entry block of the function and so far, we therefore failed to clone trivial chains if the constant was hoisted to the entry block.

Also fixes handling of `scf.if` which assumed block arguments that might not be present.

This is part of https://github.com/iree-org/iree/issues/24221.

Assisted-by: Codex


hanhanW (approved):

keshavvinayak01 (approved):
LGTM

PR #24245, opened 2026-04-24
[LinalgExt] Refactor IGEMM detail computation and update layout decisions

Max191:
Refactors getIGEMMGenericConvDetails into three stages:

1. Build fully expanded IGEMM details where every conv loop dim maps 1:1 to an IGEMM dim (no collapsing). The image-side GEMM map is constructed following the original input tensor's layout order, with filter loop dims inserted adjacent to their corresponding input channel dims.
2. Canonicalize reduction dim labeling so their order in the loop space matches the order they appear in the image-side map, which is a prerequisite for collapsing.
3. Collapse adjacent same-kind dims that are collapsible across all indexing maps, producing the final collapsed GEMM maps, loop bounds, filter reassociation, and im2col output permutation.

This makes a few improvements over the old path:
 - Better separation of concerns in the computation of the IGEMM details, because we have 3 separate phases that don't need to care what the other phases do.
 - Improved reduction order for layout mismatches between image and filter (e.g., `nhwc x nchw`), because we canonicalize to the image's order, which makes vectorized loads from global memory possible in some cases where we otherwise had artificially strided access patterns.
 - The layout of the GEMM actually matches the layout of the convolution now. Before this PR, the layout of the im2col op's result would generally be in the "canonical" order, meaning there was an implicit transpose happening in the im2col access pattern. It is better to explicitly model this transpose in the GEMM indexing maps, where it is easier to target optimizations like transpose_load.

This refactoring will also make it easier to make a couple of follow-up changes, which will come as separate PRs:
 - Redefine the classification of `Batch` vs `M` dims in the im2col op semantics. The current definition is inconsistent with what the corresponding dims actually mean in the GEMM because convolution batch dims are actually M dims in the GEMM, but we classify them as batch in the im2col op. This will be a simple reclassification in phase 1 now.
 - Flatten parallel dimensions as well as reduction dimensions. This will be very simple now, since we will just need to update phase 3.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/conv_to_im2col.mlir
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/conv_to_im2col.mlir
@@ -43,25 +43,25 @@ util.func public @conv_2d_nchw_fchw(%arg0: tensor<1x4x16x16xf32>, %arg1: tensor<
   util.return %0 : tensor<1x16x14x14xf32>
 }
 // CHECK-DAG:  #[[MAP:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d1, d4)>
-// CHECK-DAG:  #[[MAP1:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d0, d2, d3, d4)>
+// CHECK-DAG:  #[[MAP1:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d0, d4, d2, d3)>
 // CHECK-DAG:  #[[MAP2:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3)>
 // CHECK:      util.func public @conv_2d_nchw_fchw(
 // CHECK-SAME:   %[[ARG0:[a-zA-Z0-9_]+]]: tensor<1x4x16x16xf32>
 // CHECK-SAME:   %[[ARG1:[a-zA-Z0-9_]+]]: tensor<16x4x3x3xf32>
 // CHECK-SAME:   %[[ARG2:[a-zA-Z0-9_]+]]: tensor<1x16x14x14xf32>
-// CHECK:      %[[EMPTY:.+]] = tensor.empty() : tensor<1x14x14x36xf32>
+// CHECK:      %[[EMPTY:.+]] = tensor.empty() : tensor<1x36x14x14xf32>
 // CHECK:      %[[IM2COL:.+]] = iree_linalg_ext.im2col
 // CHECK-SAME:   strides = [1, 1] dilations = [1, 1] kernel_size = [3, 3]
 // CHECK-SAME:   offsets = [0, 0, 0, 0] output_sizes = {{\[}}[1], [14], [14], [4, 3, 3]]
 // CHECK-SAME:   batch_pos = [0] m_pos = [2, 3] k_pos = [1]
-// CHECK-SAME:   input_k_perm = [0, 1, 2] output_perm = [0, 1, 2, 3]
+// CHECK-SAME:   input_k_perm = [0, 1, 2] output_perm = [0, 3, 1, 2]
 // CHECK-SAME:   ins(%[[ARG0]] : tensor<1x4x16x16xf32>)
-// CHECK-SAME:   outs(%[[EMPTY]] : tensor<1x14x14x36xf32>) -> tensor<1x14x14x36xf32>
+// CHECK-SAME:   outs(%[[EMPTY]] : tensor<1x36x14x14xf32>) -> tensor<1x36x14x14xf32>
 // CHECK-DAG:  %[[COLLAPSED:.+]] = tensor.collapse_shape %[[ARG1]] {{\[}}[0], [1, 2, 3]] : tensor<16x4x3x3xf32> into tensor<16x36xf32>
 // CHECK:      %[[MATMUL:.+]] = linalg.generic
 // CHECK-SAME:   indexing_maps = [#[[MAP]], #[[MAP1]], #[[MAP2]]]
 // CHECK-SAME:   iterator_types = ["parallel", "parallel", "parallel", "parallel", "reduction"]
-// CHECK-SAME:   ins(%[[COLLAPSED]], %[[IM2COL]] : tensor<16x36xf32>, tensor<1x14x14x36xf32>)
+// CHECK-SAME:   ins(%[[COLLAPSED]], %[[IM2COL]] : tensor<16x36xf32>, tensor<1x36x14x14xf32>)
 // CHECK-SAME:   outs(%[[ARG2]] : tensor<1x16x14x14xf32>) {
 // CHECK:          arith.mulf
 // CHECK:          arith.addf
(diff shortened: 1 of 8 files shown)


Max191:
Based on local testing, this PR can cause some regressions due to pessimistic instruction scheduling in LLVM. It is a very specific issue where a specific LLVM optimization is missed due to strides being larger and not fitting into a small enough bit-width. The optimization used to cause buffer loads to be issued all together, but it no longer happens with this change. This PR: https://github.com/iree-org/iree/pull/24247 fixes that issue by reordering the instructions at MLIR level to ensure the buffer loads are issued together regardless.

I am currently running a full set of benchmarks for each change, and will post results when they are complete. Based on my local testing so far, though, each change provides some performance improvement on some shapes.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/conv_to_im2col.mlir
```diff
@@ -305,25 +305,24 @@ util.func public @conv_2d_chwn_chwf(%arg0: tensor<16x26x18x288xf32>, %arg1: tens
 }
 
 // CHECK-DAG:  #[[MAP:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d4, d0)>
-// CHECK-DAG:  #[[MAP1:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d1, d2, d4, d3)>
-// CHECK-DAG:  #[[MAP2:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3)>
+// CHECK-DAG:  #[[MAP1:.+]] = affine_map<(d0, d1, d2, d3, d4) -> (d4, d1, d2, d3)>
```
yzhang93, line 308: Missing #[[MAP2:.+]]

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -424,50 +426,489 @@ bool isGatherlikeOp(Operation *op) {
 // IGEMM details for generic convolutions
 //===---------------------------------------------------------------------===//
 
+/// Classification of IGEMM dims from the image/im2col perspective.
+/// Batch includes both batch and depth (group) dims. M is output spatial
+/// dims. InputChannel and FilterLoop are the two subcategories of reduction
+/// dims. The canonical im2col output order is:
+///   [Batch, M, InputChannel, FilterLoop].
```
yzhang93, line 433: Now as this PR puts the image/im2col map in input‑tensor order, is there a reason to keep the canonical reduction dims order in [InputChannel, FilterLoop] rather than the order these dims appear in the input?

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -424,50 +426,489 @@ bool isGatherlikeOp(Operation *op) {
...
+  SmallVector<AffineExpr> inputDims;
+  inputDims.reserve(inputDimOrder.size());
+  for (unsigned dim : inputDimOrder) {
+    inputDims.push_back(convToIgemmDimMap.at(dim));
+  }
+  return inputDims;
+}
+
+// Permutes reduction dims in the loop space so their order matches the
+// image-side map. Since reduction dims are always at the end of the source
```
yzhang93: Does this implementation rely on the reduction dims being at the end?

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -424,50 +426,489 @@ bool isGatherlikeOp(Operation *op) {
...
-      if (capturedDims.contains(convDim)) {
-        continue;
+    if (spatialIt != convDims.outputImage.end()) {
+      if (seenDims.insert(*spatialIt)) {
+        inputDimOrder.push_back(*spatialIt);
+      }
+      continue;
+    }
+
+    unsigned dim = cast<AffineDimExpr>(inputExpr).getPosition();
```
yzhang93, line 660: I think here should use dyn_cast + clear assert to avoid any potential crash.

yzhang93 (changes requested):
Nice refactor. One high-level question: now that the image GEMM map is built in input layout, is there a reason to keep the canonical im2col output order fixed at [Batch, M, InputChannel, FilterLoop] rather than "the order these dim kinds appear in the input"? Looking at the test diffs, `output_perm` exists almost entirely to undo the canonical InputChannel-before-FilterLoop split (e.g. [0, 1, 2, 4, 3] for nhwc_chwf), and `input_k_perm` exists to permute the K group's inner sizes from filter reduction order to input order. If the canonical followed input dim order, both would essentially always be identity?

Max191:
> Nice refactor. One high-level question: now that the image GEMM map is built in input layout, is there a reason to keep the canonical im2col output order fixed at [Batch, M, InputChannel, FilterLoop] rather than "the order these dim kinds appear in the input"? Looking at the test diffs, `output_perm` exists almost entirely to undo the canonical InputChannel-before-FilterLoop split (e.g. [0, 1, 2, 4, 3] for nhwc_chwf), and `input_k_perm` exists to permute the K group's inner sizes from filter reduction order to input order. If the canonical followed input dim order, both would essentially always be identity?

Good question. The main reason to put the dims in order like this is that the output shape may collapse together dims of the same type, and there isn't a requirement that the dims we collapse together are contiguous in the input. So the permutation needs to be in reference to a canonical dim order that is collapsible in the same way that the output has been collapsed. For example, if we had 2 C dims (like `[N, C0, H, W, C1]`), then the im2col could potentially collapse the C0 and C1 dims together, and the permutation wouldn't be representable based on the original input dim order.

One thing I just realized is that technically the current semantics don't allow interleaving input channel dims with filter loop dims in the collapsed K shape, but I don't see a reason to ever need to do that, so I think it's fine to keep it this way.

Max191:
FYI, I have benchmark results now: https://gist.github.com/Max191/97f05170a5f2fe81c5875700f24b2989

There are many improvers, but also many of them are small shapes, so I don't fully trust them all. The overall geomean speedup **for all shapes is neutral**, and **+1% for the "stable" shapes** (shapes that did not significantly vary between repeated iterations).

Overall, this seems like a potential average performance improvement in the range of **`1%-6%` for all the affected convolutions**, but small impact when amortizing across the full suite.

Note: The baseline this comparison was taken against was this PR: https://github.com/iree-org/iree/pull/24318, which I plan to land first.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
Max191: The restrictions on the indexing maps from isaConvolutionOpInterface guarantees that this will pass. I realized we weren't actually calling isaConvolutionOpInterface in the top level function, though (only inferConvolutionDims), so I added it and then added a comment here.

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
Max191: It does not, but the comment was stale. Nice catch!

yzhang93 (approved):
LGTM, thanks!

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -691,33 +1008,81 @@ getIGEMMGenericConvDetails(linalg::LinalgOp linalgOp) {
     indexingGEMMMaps.push_back(filterMapGEMM);
   }
   indexingGEMMMaps.push_back(resultMap);
+
+  SmallVector<int64_t> igemmLoopBounds = linalgOp.getStaticLoopRanges();
+
+  SmallVector<utils::IteratorType> igemmLoopIterators;
+  for (utils::IteratorType iteratorType : linalgOp.getIteratorTypesArray()) {
+    igemmLoopIterators.push_back(iteratorType);
+  }
```
bangtianliu: Use llvm::to_vector

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -424,56 +426,500 @@ bool isGatherlikeOp(Operation *op) {
...
+/// The expanded form has an identity mapping from conv loop dims to IGEMM
+/// loop dims. Reduction dims are then canonicalized to match the image-side
+/// order to enable subsequent collapsing.
+static FailureOr<IGEMMGenericConvDetails>
+getExpandedIGEMMGenericConvDetails(linalg::LinalgOp linalgOp) {
+  if (!linalg::isaConvolutionOpInterface(linalgOp)) {
     return failure();
   }
+
+  auto convDimsOrFailure = linalg::inferConvolutionDims(linalgOp);
```
bangtianliu: use auto only when the type is obvious from the rhs

```suggestion
  FailureOr<linalg::ConvolutionDimensions> convDimsOrFailure =                                                                                              
      linalg::inferConvolutionDims(linalgOp);  
```

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -508,19 +953,12 @@ getIGEMMGenericConvDetails(linalg::LinalgOp linalgOp) {
...
-  SmallVector<int64_t> reductionDims;
-  for (auto iter : llvm::enumerate(linalgOp.getIteratorTypesArray())) {
-    if (linalg::isReductionIterator(iter.value())) {
-      reductionDims.push_back(iter.index());
-    }
-  }
-
   bool isOutputChannelFirst = false;
   auto outputChannelPos = convDims.outputChannel;
   auto outputImagePos = convDims.outputImage;
```
bangtianliu: Also here

bangtianliu (approved):
LGTM % nits

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -424,56 +426,500 @@ bool isGatherlikeOp(Operation *op) {
...
+    iterationToCollapsedDim.resize(
+        std::max<int64_t>(iterationToCollapsedDim.size(), group.back() + 1),
+        -1);
+    for (int64_t dim : group) {
+      iterationToCollapsedDim[dim] = collapsedDim;
+    }
+  }
+
+  DenseMap<int64_t, AffineExpr> collapsedMap;
+  MLIRContext *ctx = convToIgemmDimMap.begin()->second.getContext();
```
bangtianliu: Pass MLIRContext *ctx as an additional parameter to collapseConvToIgemmDimMap.

PR #24246, opened 2026-04-24
[Codegen] Duplicate operations in tile size analysis

sommerlukas:
The tile size analysis considers some operations such as `linalg.fill` as duplicatable to avoid poisoning the tile size analysis in case such easy-to-duplicate operations are used by multiple users with different tile sizes after CSE.

So far, that duplication was never materialized. Now, the materialization pass after analysis clones duplicatable operation with the requested tile-size attributes.

This PR also fixes missing tile-size attributes on operations that have their tile size fully defined by the analysis on the result. So far, the materialization only considered operands, now it also uses result information.

This is part of https://github.com/iree-org/iree/issues/24221.

Assisted-by: Codex

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
@@ -637,6 +637,27 @@ static TileSizes getIterationSpaceTileSizes(Operation *op, unsigned numLoops,
   return iterTileSizes;
 }
 
+/// Gather tile sizes into the iteration space of a linalg op by looking up
+/// both operand and result lattice states in the solver.
+static TileSizes
+getLinalgIterationSpaceTileSizes(linalg::LinalgOp linalgOp,
+                                 const DataFlowSolver &solver) {
+  TileSizes iterTileSizes =
+      getIterationSpaceTileSizes(linalgOp, linalgOp.getNumLoops(),
+                                 linalgOp.getIndexingMapsArray(), solver);
+
+  for (auto [idx, result] : llvm::enumerate(linalgOp->getResults())) {
+    const TileSizeLattice *lattice =
+        solver.lookupState<TileSizeLattice>(result);
+    TileSizes tileSizes = getTileSizesFor(result, lattice);
+    OpOperand *init = linalgOp.getDpsInitOperand(idx);
+    AffineMap map = linalgOp.getMatchingIndexingMap(init);
+    iterTileSizes.merge(tileSizes.mapToIterationSpace(map));
+  }
+
+  return iterTileSizes;
+}
+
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -647,6 +668,89 @@ static TileSizes getIm2colTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
   return getTileSizesFor(result, lattice);
 }
 
+static std::optional<TileSizes>
+getUseOperandTileSizes(OpOperand &use, const DataFlowSolver &solver) {
```
keshavvinayak01, line 672: I think this needs to be more conservative around unhandled uses.

`getUseOperandTileSizes` only classifies direct `linalg` and `to_layout` users, but materializeDuplicatableLinalgOp` still stamps the original producer with the first classified tile-size group.
If the producer also has an unclassified live use, that use keeps pointing at the original op after it has been annotated.

For example:

```
#layout32 = #iree_vector_ext.nested_layout<
subgroup_tile = [1, 1], batch_tile = [1, 2], outer_tile = [1, 1],
thread_tile = [16, 4], element_tile = [1, 4],
subgroup_strides = [0, 0], thread_strides = [1, 16]>

#layout24 = #iree_vector_ext.nested_layout<
subgroup_tile = [1, 1], batch_tile = [1, 3], outer_tile = [1, 1],
thread_tile = [16, 2], element_tile = [1, 4],
subgroup_strides = [0, 0], thread_strides = [1, 16]>

%fill = linalg.fill ins(%zero : f32) outs(%empty : tensor<16x24xf32>) -> tensor<16x24xf32>
%a = iree_vector_ext.to_layout %fill to layout(#layout32) : tensor<16x24xf32>

%if = scf.if %cond -> tensor<16x24xf32> {
  scf.yield %fill : tensor<16x24xf32>
} else {
  scf.yield %fill : tensor<16x24xf32>
}
%b = iree_vector_ext.to_layout %if to layout(#layout24) : tensor<16x24xf32>
```
The `to_layout` use is grouped, but the `scf.yield` uses are skipped, so `%fill` can get `vector_tile_sizes = [16, 32]` while it is still feeding the path that wants `[16, 24]`.

Could we either bail out when any use is unclassified, or clone the classified uses away and leave the original producer unannotated for the unhandled uses?

sommerlukas: Thanks, that's a great catch! 

I've opted for the second solution, so leaving the original producer unannotated for all the unhandled uses. I've also added a new test based on the scenario you outlined.

amd-eochoalo (approved):
Awesome! Thanks!

keshavvinayak01 (approved):

PR #24247, opened 2026-04-24
[LLVMGPU][ROCDL] Add pass to group global loads for better instruction scheduling

Max191:
Adds LLVMGPUGroupGlobalLoadsPass which moves global loads in the same block to be adjacent to each other when they are separated by pure address computation ops. The pass moves each load along with its transitive dependency chain to be right after the preceding global load.

This improves performance in situations where LLVM is not able to convert address computation into a single base + constant offset. In such cases, instruction scheduling can become pessimistic and each global load needs to be waited on before the next is issued. With this instruction reordering, all global loads are issued together after address computation is completed.

Based on benchmarks with this change alone, we don't have any cases in our suite of kernels that runs into this issue today. However, some convolution shapes run into the issue after the changes in https://github.com/iree-org/iree/pull/24245, and this PR prevents such regressions.

This is only enabled for ROCDL in this PR, because we don't have any data points to support adding it to other pipelines yet.


Max191:
There is no measurable performance impact from these changes on main, but in https://github.com/iree-org/iree/pull/24245 this will eliminate the regressions it causes. We likely aren't producing any layouts that run into this issue today.

I ran benchmarks twice, and all of the performance improvements and regressions seem to be fake (they don't appear in both runs):

 - https://github.com/nod-ai/amd-shark-ai-reports/tree/main/boo/boo-custom-runs/2026-04-28_19-58_73cbfcee97779998ef3d73b97f1b32d9b6ed4970/comparison
 - https://github.com/nod-ai/amd-shark-ai-reports/tree/main/boo/boo-custom-runs/2026-04-24_20-15_73cbfcee97779998ef3d73b97f1b32d9b6ed4970/comparison

benvanik:
neat! I've noticed this and it's super annoying llvm fails to handle it, as it can eat up _a lot_ of time.

benvanik:
image
(wmma kernel I was spotting this in - 251 s_waitcnt for 64 v_wmma_f32)

Max191:
> (wmma kernel I was spotting this in - 251 s_waitcnt for 64 v_wmma_f32)

This one is interesting. It looks different from what I saw (so probably not what this PR addresses). If you have a reproducer on hand, I'd be happy to spin up an agent on it when I have some time!

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLGroupBufferLoads.cpp
```diff
@@ -0,0 +1,170 @@
...
+/// and that are strictly between `boundary` and `op`. These are the ops
+/// that must be moved along with `op` if it is hoisted above `boundary`.
+static void collectDepsInRange(Operation *op, Operation *boundary,
+                               llvm::SetVector<Operation *> &deps) {
+  for (Value operand : op->getOperands()) {
+    Operation *defOp = operand.getDefiningOp();
+    if (!defOp || defOp->getBlock() != op->getBlock()) {
+      continue;
+    }
+    if (!boundary->isBeforeInBlock(defOp) || !defOp->isBeforeInBlock(op)) {
```
nirvedhmeshram: I dont think we need to worry about the second condition as the IR would already be illegal if that were the case?

krzysz00:
Those u16 loads we need to fix with a better LDS layout I think 

Also,, I don't think we should restrict this to buffer loads - all global memory ops could be edit from this

nirvedhmeshram (commented):
Mostly looks good, +1 to @krzysz00 's point that we could just make this happen for all global instead of just buffer loads. But maybe we should restrict this pass for amdgpu backend only. Not sure if it is beneficial for other backends so can bail out there.

Max191:
> Mostly looks good, +1 to @krzysz00 's point that we could just make this happen for all global instead of just buffer loads. But maybe we should restrict this pass for amdgpu backend only. Not sure if it is beneficial for other backends so can bail out there.

Good points from both of you. I'll support global load but only run the pass for rocdl

compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUGroupGlobalLoads.cpp
```diff
@@ -0,0 +1,206 @@
...
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_LLVMGPUGROUPGLOBALLOADSPASS
+#include "iree/compiler/Codegen/LLVMGPU/Passes.h.inc"
+
+namespace {
+
+/// Returns true if the operation is a vector.load from global memory.
+static bool isGlobalLoad(Operation *op) {
+  auto loadOp = dyn_cast<vector::LoadOp>(op);
```
krzysz00: This should also hit memref load?

krzysz00 (approved):
I think this is fine

nirvedhmeshram (approved):
LGTM

PR #24248, opened 2026-04-24
[HAL/CPU] Local Runtime Rewrite

benvanik:
For the humans, from the human:

> A principaled near-ground-up rewrite: this project lived long enough I finally got to a v3 :) There's definitely work to do here for tuning to different workloads and different targets, but we now have the detailed programmatic feedback mechanism to capture what is happening and determine why things are slow and a foundation that is very tightly optimized for a broader set of cases than the original. The same poor occupancy planning/distribution pitfalls from before (and for any system) bite this, but now properly planned dispatches run significantly better. WIP tooling allows for deterministic simulation via recorded workload logs based on ireeprof captures to let us adjust the heuristics, and though I expect nearly all cases to improve (all I could measure on my machine did) there will definitely be some uneven behavior on pathological cases that'll need to be worked through with that. That's fine, though, as they probably weren't very good on the old system either. The other foundation laid here is for future dynamic workgroup size modes: the live warm worker state is now tracked precisely and subdividing tail workgroups is a decision we can make (by making most invocations multiples of the defined minimum workgroup size and then scaling down to 1x multipliers near the tail). Fun future work!

---

This PR rebuilds the CPU HAL execution path around the new queue, pool, command buffer, profiling, and CTS contracts. The most visible result is a much better local-task runtime: lower latency, lower aggregate CPU time, and less scheduler variance on model-shaped CPU workloads.

## Why

The old local-task path had accumulated scheduler behavior that was hard to observe and harder to tune. The new profile and Tracy evidence made it clear that real workloads can spend surprising time in wake/sleep gaps, oversubscription, and poorly bounded spinning when the compiler's distribution does not match hardware occupancy.

This slice moves the CPU HALs onto the new common contracts and uses the rewrite as an opportunity to simplify the execution path and make it observable enough to tune.

## What's Here

Local-sync and local-task are updated for the new pool and profiling APIs. Command buffers carry enough metadata to attribute execution without hot-path scans. CPU dispatch and command-buffer execution can emit structured profile events when profiling is active while keeping the non-profiling hot path down to well-predicted checks.

Local-task command buffers now record compact command blocks and replay through the task queue instead of the old `deferred_command_buffer` + `task_command_buffer` and task-queue-state paths. The shared local HAL substrate adds inline dispatch helpers, transient buffers, ELF-format helpers, local profile recording, executable profile ids, and CTS profile utilities.

CTS coverage is expanded for queue alloca, dispatch, transfer behavior, command-buffer dispatch pipelines, reuse, transients, stress, and indirect dispatch parameters. The new indirect-parameter CTS exposed that allocators were only marking storage buffers as queue-dispatch-compatible. Treat all dispatch usage bits, including indirect parameters, as queue-dispatch resources in the heap, CUDA, HIP, Metal, and Vulkan allocators; Vulkan now also requests VK_BUFFER_USAGE_INDIRECT_BUFFER_BIT for indirect-parameter buffers.

## Performance

Fresh SDXL CLIP/prompt encoder evidence gathered on April 21, 2026:

```text
Workload: SDXL CLIP/prompt encoder, encode_prompts
Device: local-task
Workers: --task_topology_max_group_count=64
Benchmark: iree-benchmark-module, process_time/real_time
Iterations: --benchmark_min_time=10x
Repetitions: 100
Build: optimized ThinLTO/O3/march=native/NDEBUG
```

| Run | Median Wall | Wall CV | Median CPU | CPU CV | Speedup | Wall/CPU Reduction |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| main, unpinned x64 | 156.135 ms | 1.98% | 4523.298 ms | 1.61% | 1.0x | 100.0% |
| landing, unpinned x64 | 107.653 ms | 1.08% | 3850.701 ms | 0.18% | 1.45x | 31.1% / 14.9% |

NUMA-pinned probes show that the result was not only an unpinned scheduler accident:

| Run | Median Wall | Wall CV | Median CPU | CPU CV | Wall Reduction |
| --- | ---: | ---: | ---: | ---: | ---: |
| main, NUMA node0 x48 | 180.319 ms | 0.81% | 4612.521 ms | 0.21% | baseline |
| landing, NUMA node0 x48 | 139.911 ms | 0.33% | 4521.048 ms | 0.21% | 22.4% |
| main, NUMA nodes0-1 x64 | 171.901 ms | 1.48% | 5247.244 ms | 0.20% | baseline |
| landing, NUMA nodes0-1 x64 | 129.924 ms | 0.20% | 5086.302 ms | 0.11% | 24.4% |

Toy LLaMA 3.1 prefill gives a second latency-sensitive CPU shape:

| Run | Median Wall | Wall CV | Median CPU | CPU CV | Speedup | Reduction |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| main, x64 | 0.953 ms | 3.32% | 7.069 ms | 0.79% | 1.0x | 100.0% |
| landing, x64 | 0.553 ms | 0.89% | 7.399 ms | 0.61% | 1.72x | 42.0% |

Toy LLaMA is wall-latency evidence rather than CPU-efficiency evidence; the small benchmark shape gets much faster at user-visible latency while the median process CPU time increases by about 4.7%.

## Pictures

Overheads on dense work are ~40ns/tile (for same task) and ~140ns/tile (for new task, which is basically the cache miss):
image

Workers now do pending-work-aware micro spins to bridge tail gaps (works best with proper distribution) while parking immediately on fan-in:
image

Latency from queue_execute to first workgroup is entirely thread wake time (~5-10us on linux):
image

Latency from the last workgroup to signal is ~200ns and then wake of waiters is thread wake time, importantly all retirement bookkeeping is done after the signal:
image

iree-profile instrumentation is implemented and both jsonl, human-readable, and pftrace output is valid:
image

iree-profile data contains detailed scheduling information (command_region_events) for tuning:
```json
  {
    "schema_version": 15,
    "record_type": "command_region_event",
    "source_record_index": 45,
    "physical_device_ordinal": 0,
    "queue_ordinal": 0,
    "stream_id": 0,
    "event_id": 13344,
    "submission_id": 39,
    "command_buffer_id": 2,
    "block_sequence": 41,
    "region_epoch": 1317,
    "region_index": 57,
    "next_region_index": 58,
    "flags": 9,
    "host_time_domain": "iree_host_time_ns",
    "start_host_time_ns": 3246529737756425,
    "end_host_time_ns": 3246529737898910,
    "duration_ns": 142485,
    "valid": true,

    "region_dispatch_count": 1,
    "region_tile_count": 192,
    "region_width_bucket": 64,
    "region_lookahead_width_bucket": 32,
    "next_region_tile_count": 8,
    "next_region_width_bucket": 8,
    "next_region_lookahead_width_bucket": 32,

    "worker_count": 64,
    "old_wake_budget": 64,
    "new_wake_budget": 8,
    "wake_delta": 0,

    "useful_drain_count": 64,
    "no_work_drain_count": 50,
    "first_useful_drain_start_host_time_ns": 3246529737757166,
    "last_useful_drain_end_host_time_ns": 3246529737898239,

    "tail_no_work": {
      "count": 52,
      "remaining_tiles": {
        "min": 3,
        "max": 175,
        "bucket_counts": [0, 0, 0, 2, 0, 0, 5, 45]
      },
      "first_start_host_time_ns": 3246529737877989,
      "last_end_host_time_ns": 3246529737898510,
      "time_sums": {
        "start_offset_ns": 6701706,
        "drain_duration_ns": 13820
      }
    },

    "retention_keep_active_count": 0,
    "retention_publish_keep_active_count": 0,
    "retention_keep_warm_count": 50
  }
```

ci-extra: all


benvanik:
Performance changes are expected here, but all _should_ be positive. If they aren't, profiles and reproducers welcome and I'll do some tuning.

PR #24256, opened 2026-04-24
[Codegen][GPU] Root non-unique scatter before update producers

benvanik:
Non-unique scatter cannot be treated as an ordinary consumer of a tiled update producer. Duplicate indices make the batch dimensions reduction-like, so fusing the scatter through a forall update producer asks the scatter tiling interface for an operand-tile mapping it explicitly cannot provide.

Keep that legality boundary in the generic forall consumer-fusion pattern, and make LLVMGPU root selection treat non-unique scatter update producers as auxiliary to the scatter root. That keeps the scatter in charge of the distribution strategy instead of leaving it as a global write outside the workgroup-distributed region.

Also allow scatter lowering config for non-unique scatters that have inner parallel slice dimensions. The duplicate-index batch dimensions are left serial with zero tile sizes while the inner slice dimensions can still be tiled across workgroups and threads; scalar or fully reduction-shaped scatters fall back to the existing default path.

Add regression coverage for both the no-fusion legality boundary and the LLVMGPU root/config selection. Verified with the Common/GPU and LLVMGPU/ROCDL lit suites under ASAN, plus a local Toy Llama ROCm/gfx1100 compile repro.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
@@ -49,6 +49,12 @@ struct FuseTilableForallConsumers final
   using OpInterfaceRewritePattern::OpInterfaceRewritePattern;
   LogicalResult matchAndRewrite(TilingInterface tilableOp,
                                 PatternRewriter &rewriter) const override {
+    auto scatterOp = dyn_cast<IREE::LinalgExt::ScatterOp>(*tilableOp);
+    if (scatterOp && !scatterOp.getUniqueIndices()) {
+      return rewriter.notifyMatchFailure(
+          tilableOp, "non-unique scatter indices are not consumer-fusible");
+    }
+
     // Currently consumer fusion requires DPS, and we don't want to fuse through
     // inits anyway.
     auto dpsOp = dyn_cast<DestinationStyleOpInterface>(*tilableOp);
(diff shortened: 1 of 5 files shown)


MaheshRavishankar:
I think @IanWood1 had a similar PR out for the fix

benvanik:
ah, I'll let @IanWood1 handle it then - I went away for lunch and came back to my agent having found and fixed it while trying to run benchmarks :P

compiler/src/iree/compiler/Codegen/Common/Transforms.cpp
```diff
@@ -49,6 +49,12 @@ struct FuseTilableForallConsumers final
   using OpInterfaceRewritePattern::OpInterfaceRewritePattern;
   LogicalResult matchAndRewrite(TilingInterface tilableOp,
                                 PatternRewriter &rewriter) const override {
+    auto scatterOp = dyn_cast<IREE::LinalgExt::ScatterOp>(*tilableOp);
+    if (scatterOp && !scatterOp.getUniqueIndices()) {
```
qedawkins, line 53: This isn't the right way to check this. What should be happening is this pattern calls `isOpFusableWithProducerSlices` -> `getIterationDomainTileFromOperandTiles` -> `getTiledImplementationFromOperandTiles` to first query for fusability in a pattern safe way then actually do the rewrite. Then in `LinalgExt::ScatterOp` we just add `.getUniqueIndices` to `isOpFusableWithProducerSlices` to guard the fusion. It would be nice if we could be clever and check for ranges on the slicing of unique indices but we don't have a way to verify that the upper bound of a slice matches the dimension size so the pessimistic check is fine. We'll change that in the future with slightly different flavors of slicing ops.

qedawkins:
oh ok, ignore me then

PR #24260, opened 2026-04-24
[HAL][LocalTask] Add task_topology_group_count device creation parameter

hanhanW:
iree_hal_driver_create_device_by_uri parses the query string of a device URI into iree_string_pair_t parameters and forwards them to the driver through create_device_by_path. The local-task driver previously ignored those parameters and always returned a device backed by the executors built at driver-module registration time, so a URI like "local-task://?task_topology_group_count=16" silently produced a device with whatever worker count the process-wide flags had selected.

Parses task_topology_group_count in iree_hal_task_driver_create_device_by_id (reached via both create_by_id and create_by_path/create_by_uri). When present, constructs a fresh task executor with the requested group count and uses it for the new device; the default-executor path is unchanged when the parameter is absent. Zero and non-numeric values are rejected with INVALID_ARGUMENT.

Adds task_driver_test.cc covering
- iree_hal_driver_create_device_by_uri with the literal URI form callers use, asserting hal.dispatch:concurrency matches the requested count;
- the underlying create_by_id and create_by_path overrides at counts {1, 2, 4};
- the default (no-parameter) path;
- rejection of "0" and non-numeric values; and
- graceful handling of unknown parameter keys, matching the convention used by other HAL drivers (e.g. hip_driver.c).

Assisted-by: Claude Code

### DIFF
--- a/runtime/src/iree/hal/drivers/local_task/task_driver_test.cc
+++ b/runtime/src/iree/hal/drivers/local_task/task_driver_test.cc
@@ -1,0 +1,199 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Local-task driver-specific tests.
+//
+// These tests live outside the cross-driver CTS suite because they exercise
+// behavior that is unique to the local-task driver: parsing the
+// task_topology_group_count creation parameter and using it to override the
+// driver's default executor topology.  hal.dispatch:concurrency is queried as
+// a black-box check that the override took effect.
+
+#include "iree/async/util/proactor_pool.h"
+#include "iree/base/api.h"
+#include "iree/base/threading/numa.h"
+#include "iree/hal/api.h"
+#include "iree/hal/drivers/local_task/registration/driver_module.h"
+#include "iree/testing/gtest.h"
+#include "iree/testing/status_matchers.h"
+
+namespace iree::hal::local_task {
+namespace {
+
+class TaskDriverTest : public ::testing::Test {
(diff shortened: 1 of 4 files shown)


runtime/src/iree/hal/drivers/local_task/task_driver_test.cc
```diff
@@ -0,0 +1,199 @@
...
+  iree_hal_device_create_params_t create_params = MakeCreateParams();
+  iree_hal_device_t* device = nullptr;
+  IREE_ASSERT_OK(iree_hal_driver_create_device_by_uri(
+      driver_, iree_make_string_view(uri.data(), uri.size()), &create_params,
+      iree_allocator_system(), &device));
+
+  EXPECT_EQ(QueryDispatchConcurrency(device), group_count);
+
+  iree_hal_device_release(device);
+}
```
hanhanW, line 147: @benvanik the bug is identified from this path when I was exploring our CPU perf with pytorch.

benvanik (changes requested):
unfortunately I don't think this is going to work like this - the model is that executors are created and provided, and options cannot change them. this is required for API use: here, you are linking in the flags which is not legal in a library. the flags are how these options are controlled today.

(also, this is an oddly worded feature request, not a bug: that ?task_topology_group_count=N didn't work before was because that was not intended to work)

hanhanW:
> unfortunately I don't think this is going to work like this - the model is that executors are created and provided, and options cannot change them. this is required for API use: here, you are linking in the flags which is not legal in a library. the flags are how these options are controlled today.
> 
> (also, this is an oddly worded feature request, not a bug: that ?task_topology_group_count=N didn't work before was because that was not intended to work)

I see, thanks for the details! I think I need to route my "tiny runtime" to use different approach. The task_topology_group_count is a different axis, which is not part of executor creation. Thanks for the info!

PR #24268, opened 2026-04-27
Fix release due to missing directory

nirvedhmeshram:
Iree-profile-render is a Python script that loads its modules from share/iree/profile/render/ at runtime. The tarball packaging in build_dist.py only included bin/, lib/, and include/ — leaving share/ out. The script lands in bin/ but its Python module support files are absent, causing ModuleNotFoundError: No module named 'render' on any invocation including --help.

Fix: Added "share" to dist_entries in build_dist.py. The next release build will include the share/iree/profile/render/ directory in the tarball and the CI validation step will pass.

### DIFF
--- a/build_tools/github_actions/build_dist.py
+++ b/build_tools/github_actions/build_dist.py
@@ -182,6 +182,7 @@ def build_main_dist():
         "bin",
         "lib",
         "include",
+        "share",
     ]
     dist_archive = os.path.join(
         BINDIST_DIR,


ScottTodd (approved):

nirvedhmeshram:
PR comment form Claude

**Root cause & fix**

`iree-profile-render` locates its Python support modules at runtime by looking for `../share/iree/profile/render/` relative to `bin/`. The tarball packaging in `build_dist.py` only included `bin/`, `lib/`, and `include/` — so the script landed in the tarball but its modules did not, causing every invocation (including `--help`) to fail:

```
ModuleNotFoundError: No module named 'render'
```

The fix adds `"share"` to `dist_entries`, which brings in 5 Python files (~108 KB uncompressed, negligible after xz compression):

| File |
|---|
| `share/iree/profile/render/__init__.py` |
| `share/iree/profile/render/backends.py` |
| `share/iree/profile/render/cli.py` |
| `share/iree/profile/render/common.py` |
| `share/iree/profile/render/perfetto.py` |

---

**Testing**

The failure was reproduced locally by simulating the tarball layout — placing only `bin/iree-profile-render` in a directory without `share/` and running:

```sh
./bin/iree-profile-render --help
```

This produced the same exit code 1 and `ModuleNotFoundError` as CI. The fix was validated by repeating the test with `share/iree/profile/render/` present, at which point `--help` exits cleanly.

> **Note:** This does not exercise the full `build_dist.py` end-to-end (that requires the cibuildwheel/manylinux environment). The simulation is sufficient because `build_dist.py` performs a straightforward `tarfile.add()` for each entry in `dist_entries` — adding `"share"` directly maps to those files being included in the archive.

PR #24269, opened 2026-04-27
[LLVMCPU] Apply binding subspan byte offsets to base pointers

benvanik:
hal.interface.binding.subspan offsets are byte offsets into the HAL binding range. LLVMCPU lowering previously treated those offsets as memref descriptor offsets, converting bytes to element offsets for dynamic descriptors and letting static-shape descriptors bypass the offset entirely.

That was only accidentally correct when the generated dispatch used dynamic memrefs with element-aligned offsets. It broke for static-shaped descriptors produced from packed dispatch-local resources, and it could not faithfully represent byte-granular or sub-byte element addressing.

Now we apply the byte offset with an i8 GEP to the loaded binding base pointer before constructing the memref descriptor and leave the descriptor offset for the memref layout offset; dynamic descriptors that only carried the HAL subspan offset now start at zero.

This was caught once binding optimization packed things like an i32 scratch buffer and an f32 workspace/output buffer into one binding, then the kernel used stream.binding.subspan/hal.interface.binding.subspan offsets to recover each logical view. LLVMCPU lowered those logical views as if they all started at the binding base. Once the kernel did real work across those aliased/misbased views the address computations went sideways hard enough to crash instead of just producing a quiet wrong answer.

The accompanying lowering cases document the previously missing static-shape and byte-granular forms.

Note that alignment remains governed by InterfaceBindingSubspanOp::calculateAlignment(), which computes the effective alignment of the adjusted binding pointer from the binding base alignment and byte-offset alignment; producers must still align typed subspans to at least their natural element alignment.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/DispatchABI.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/DispatchABI.cpp
@@ -15,6 +15,7 @@
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Math/IR/Math.h"
 #include "mlir/Dialect/Utils/StructuredOpsUtils.h"
+#include "mlir/IR/Matchers.h"
 
 static llvm::cl::opt<bool> clVerboseDebugInfo(
     "iree-codegen-llvm-verbose-debug-info",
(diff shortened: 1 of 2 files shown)


MaheshRavishankar (approved):
I hope this is right, and we dont retro-actively find an issue here. This seems correct to me cause we are doing byte-offset. So approving, but have some PTSD and these falling into some strange cracks.

benvanik:
Yeah I triple checked for the same reason - "how did this never happen!?" :P
Consensus is that we do not have any path in the compiler _today_ that ever hits statically shaped buffers with dynamic offsets, due to the stream.resource.pack always putting static first and dynamic second. I had an agent hand-authoring/editing executables, though, and in MoE land it needed to do this kind of dynamic slicing (expert base selected from slab by data-dependent value).

PR #24273, opened 2026-04-27
[Codegen] Support pack/unpack/linalg generic transpose in CombineLayoutTransformation

Muzammiluddin-Syed-ECE:
When performing writes to LDS for unaligned gemm operands, we lower the preceding `tensor.extract_slice` -> `tensor.pad` -> `linalg.copy` to a masked `vector.transfer_read` in a `scf.for` loop

However, when this chain of ops inside the lds promotion `scf.for` loops includes a `linalg.generic (transpose)` we produce scalar writes with `memref.map_load`.

This happens because the raising of the linalg.generic to a linalg.transpose is erroneously not firing prior to vectorization in `CombineSourceLayoutTransformationPass` due to `simplifyComplexRelayoutOps` not recognizing linalg.generic (transpose) when inside a for loop. 

This change removes `simplifyComplexRelayoutOps` entirely and performs the folding of unpack/pack ops and linalg.generic (transpose) ops directly without relying on a preliminary simplification of these ops prior to folding.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -46,8 +46,8 @@ using IREE::LinalgExt::MapStoreOp;
 static void simplifyComplexRelayoutOps(RewriterBase &rewriter,
                                        FunctionOpInterface funcOp) {
   OpBuilder::InsertionGuard g(rewriter);
-  SmallVector<linalg::PackOp> packOps(
-      funcOp.getFunctionBody().getOps<linalg::PackOp>());
+  SmallVector<linalg::PackOp> packOps;
+  funcOp.walk([&](linalg::PackOp op) { packOps.push_back(op); });
   for (auto packOp : packOps) {
     rewriter.setInsertionPoint(packOp);
     FailureOr<linalg::LowerPackResult> result = linalg::lowerPack(
@@ -63,15 +63,15 @@ static void simplifyComplexRelayoutOps(RewriterBase &rewriter,
       rewriter.replaceOp(result->padOp, result->padOp.getSource());
     }
   }
-  SmallVector<linalg::UnPackOp> unPackOps(
-      funcOp.getFunctionBody().getOps<linalg::UnPackOp>());
+  SmallVector<linalg::UnPackOp> unPackOps;
+  funcOp.walk([&](linalg::UnPackOp op) { unPackOps.push_back(op); });
   for (auto unPackOp : unPackOps) {
     rewriter.setInsertionPoint(unPackOp);
     (void)linalg::lowerUnPack(rewriter, unPackOp,
                               /*lowerUnpadLikeWithExtractSlice=*/false);
   }
-  SmallVector<linalg::GenericOp> genericOps(
-      funcOp.getFunctionBody().getOps<linalg::GenericOp>());
+  SmallVector<linalg::GenericOp> genericOps;
+  funcOp.walk([&](linalg::GenericOp op) { genericOps.push_back(op); });
   for (auto genericOp : genericOps) {
     if (linalg::isaTransposeOpInterface(genericOp)) {
       rewriter.setInsertionPoint(genericOp);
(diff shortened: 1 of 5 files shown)


amd-eochoalo:
@Muzammiluddin-Syed-ECE after your second commit, the PR title and description no longer match the patchset. Can you please address that. Thanks!

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -515,6 +587,19 @@ FailureOr<MapStoreOp> foldIntoMapStore(RewriterBase &rewriter, Operation *op,
...
+      .Case<linalg::PackOp>([&](linalg::PackOp packOp) {
+        return foldPackIntoMapStore(rewriter, packOp, mapStoreOp);
+      })
+      .Case<linalg::UnPackOp>([&](linalg::UnPackOp unPackOp) {
+        return foldUnpackIntoMapStore(rewriter, unPackOp, mapStoreOp);
+      })
+      .Case<linalg::GenericOp>([&](linalg::GenericOp genericOp) {
+        if (auto perm = linalg::isaTransposeOpInterface(genericOp)) {
+          return foldTransposeGenericIntoMapStore(rewriter, genericOp, *perm,
+                                                  mapStoreOp);
```
Max191: nit: Just call foldTransposePermIntoMapStore and then you won't need the extra function.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -97,28 +53,38 @@ static MapStoreOp foldIdentityLikeOpIntoMapStore(RewriterBase &rewriter,
...
+/// Fold a `transposeOp` into a consumer `mapStoreOp`, by transposing the
+/// uses of the `mapStoreOp`s transformation_region block arguments.
+static MapStoreOp foldTransposeIntoMapStore(RewriterBase &rewriter,
+                                            linalg::TransposeOp transposeOp,
+                                            MapStoreOp mapStoreOp) {
+  assert(mapStoreOp.getInput() == transposeOp->getResult(0) &&
+         "expected transposeOp to be the producer of mapStoreOp");
+  return foldTransposePermIntoMapStore(rewriter, transposeOp.getPermutation(),
+                                       transposeOp.getInput(), mapStoreOp);
+}
```
Max191: nit: You can probably just call foldTransposePermIntoMapStore directly instead of having this function as well.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -496,6 +462,112 @@ foldPadIntoMapStore(RewriterBase &rewriter, tensor::PadOp padOp,
...
+    return failure();
+  }
+  mapStoreOp = *foldResult;
+
+  if (result->padOp) {
+    if (areAllConstantIntValue(result->padOp.getMixedLowPad(), 0) &&
+        areAllConstantIntValue(result->padOp.getMixedHighPad(), 0)) {
+      rewriter.replaceOp(result->padOp, result->padOp.getSource());
+    }
+    // Non-identity pad left for FoldPadOpIntoMapStorePattern.
```
Max191: Pack is a little bit weird, because it can sometimes have padding semantics, and sometimes not. For when it does have padding semantics, though, we should probably treat it similarly to tensor.pad. The special thing about linalg.pad in this pass is that it has the `PadDistributionConfigFn`, so if a PadDistributionConfigFn is not supplied by the caller, then we shouldn't fold any linalg.pack with padding semantics. If there is a PadDistributionConfigFn, then we can fold any pack.

I'd probably just add a separate transform for foldPackIntoMapStore that takes an optional PadDistributionConfigFn, and behaves like:
 - If there is a PadDistributionConfigFn, then always decompose and fold.
 - If there is no PadDistributionConfigFn and the linalg.pack has no padding semantics, then decompose and fold (and in this case the tensor.pad from the decomposition is a no-op, so it can be folded away)
 - If there is no PadDistributionConfigFn and the linalg.pack has padding semantics, then the transform just returns failure and does nothing.

Then you can add a similar pattern to FoldPadOpIntoMapStorePattern (but this time with an optional PadDistributionConfigFn) in Transforms.cpp, and add the pattern to populateCombineRelayoutOpPatterns.

Max191 (commented):
Nice, thanks for implementing the full pack/unpack/generic support and finally getting rid of simplifyComplexRelayoutOps!

Just one comment about linalg.pack, and a couple nits, but overall looks good.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Muzammiluddin-Syed-ECE: done!

Abhishek-Varma (approved):
LGTM!

PR #24279, opened 2026-04-28
[LLVMCPU] Lower explicit workgroup-local allocs through dispatch ABI

benvanik:
Add #iree_codegen.workgroup_local as a codegen memory-space attribute for authored LLVMCPU dispatch allocations that should use HAL workgroup local memory instead of the thread stack.

This deliberately matches the GPU-side semantic split: workgroup/shared memory is represented as memref.alloc in a workgroup memory space, while memref.alloca remains private stack scratch. LLVMCPU uses an IREE-specific memory-space attribute instead of #gpu.address_space<workgroup>, but keeps the same alloc/dealloc ownership model.

The assignment pass is intentionally narrow: it only handles entry-block memref.alloc operations in HAL executable exports, computes byte ranges with the target data layout and ABI alignment, rejects dynamic or unsupported layouts, and refuses to overwrite existing local-memory assignments or predeclared export requirements.

ConvertToLLVM consumes the assigned range by building memref descriptors from the HAL workgroup local-memory pointer. Matching memref.dealloc operations are erased because the storage is owned by the dispatch frame.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
@@ -67,6 +67,8 @@ constexpr StringLiteral kUkernelAttrName = "iree_codegen.ukernel";
 constexpr StringLiteral kUKernelProviderName = "iree_codegen.ukernel_provider";
 constexpr StringLiteral kVectorTileSizesAttrName =
     "iree_codegen.vector_tile_sizes";
+constexpr StringLiteral kWorkgroupLocalMemoryRangeAttrName =
+    "iree_codegen.local_memory_range";
 
 //===----------------------------------------------------------------------===//
 // Helpers for getting/setting iree_codegen.translation_info attribute on a
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
@@ -67,8 +67,6 @@ constexpr StringLiteral kUkernelAttrName = "iree_codegen.ukernel";
 constexpr StringLiteral kUKernelProviderName = "iree_codegen.ukernel_provider";
 constexpr StringLiteral kVectorTileSizesAttrName =
     "iree_codegen.vector_tile_sizes";
-constexpr StringLiteral kWorkgroupLocalMemoryRangeAttrName =
-    "iree_codegen.local_memory_range";
 
 //===----------------------------------------------------------------------===//
 // Helpers for getting/setting iree_codegen.translation_info attribute on a
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
@@ -73,6 +73,8 @@ constexpr StringLiteral kUkernelAttrName = "iree_codegen.ukernel";
 constexpr StringLiteral kUKernelProviderName = "iree_codegen.ukernel_provider";
 constexpr StringLiteral kVectorTileSizesAttrName =
     "iree_codegen.vector_tile_sizes";
+constexpr StringLiteral kWorkgroupLocalMemoryRangeAttrName =
+    "iree_codegen.local_memory_range";
 
 //===----------------------------------------------------------------------===//
 // Helpers for getting/setting iree_codegen.translation_info attribute on a
--- a/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
@@ -73,8 +73,6 @@ constexpr StringLiteral kUkernelAttrName = "iree_codegen.ukernel";
 constexpr StringLiteral kUKernelProviderName = "iree_codegen.ukernel_provider";
 constexpr StringLiteral kVectorTileSizesAttrName =
     "iree_codegen.vector_tile_sizes";
-constexpr StringLiteral kWorkgroupLocalMemoryRangeAttrName =
-    "iree_codegen.local_memory_range";
 
 //===----------------------------------------------------------------------===//
 // Helpers for getting/setting iree_codegen.translation_info attribute on a
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
@@ -11,6 +11,7 @@
 #include "iree/compiler/Dialect/Util/IR/UtilTypes.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/MathExtras.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/Pass/Pass.h"
@@ -27,16 +28,20 @@ constexpr int64_t kMinWorkgroupLocalAlignment = 16;
 struct LLVMCPUAssignWorkgroupLocalMemoryPass
     : impl::LLVMCPUAssignWorkgroupLocalMemoryPassBase<
           LLVMCPUAssignWorkgroupLocalMemoryPass> {
-  using impl::LLVMCPUAssignWorkgroupLocalMemoryPassBase<
-      LLVMCPUAssignWorkgroupLocalMemoryPass>::
-      LLVMCPUAssignWorkgroupLocalMemoryPassBase;
+  using Base::Base;
   void runOnOperation() override;
 };
 
-static bool hasWorkgroupLocalMemorySpace(memref::AllocOp allocOp) {
-  return isa_and_nonnull<IREE::Codegen::WorkgroupLocalMemoryAttr>(
-      allocOp.getType().getMemorySpace());
-}
+struct LocalAllocLayout {
+  memref::AllocOp allocOp;
+  int64_t byteOffset = 0;
+  int64_t elementFootprint = 0;
+};
+
+struct AllocationSize {
+  int64_t elementFootprint = 0;
+  int64_t byteSize = 0;
+};
 
 static bool hasWorkgroupLocalMemorySpace(Type type) {
   auto memRefType = dyn_cast<BaseMemRefType>(type);
@@ -56,6 +61,18 @@ static LogicalResult checkedMul(int64_t lhs, int64_t rhs, int64_t &result) {
     return failure();
   }
   return success();
+}
+
+static LogicalResult checkedAlignTo(int64_t value, int64_t alignment,
+                                    int64_t &result) {
+  assert(value >= 0);
+  assert(alignment > 0);
+  int64_t remainder = value % alignment;
+  if (remainder == 0) {
+    result = value;
+    return success();
+  }
+  return checkedAdd(value, alignment - remainder, result);
 }
 
 static FailureOr<int64_t>
@@ -109,7 +126,8 @@ computeStaticElementFootprint(memref::AllocOp allocOp) {
   return footprint;
 }
 
-static FailureOr<int64_t> computeAllocationByteSize(memref::AllocOp allocOp) {
+static FailureOr<AllocationSize>
+computeAllocationSize(memref::AllocOp allocOp) {
   MemRefType type = allocOp.getType();
   FailureOr<int64_t> elementFootprint = computeStaticElementFootprint(allocOp);
   if (failed(elementFootprint)) {
@@ -136,7 +154,10 @@ static FailureOr<int64_t> computeAllocationByteSize(memref::AllocOp allocOp) {
     allocOp.emitOpError("workgroup local memory allocation size overflow");
     return failure();
   }
-  return totalBytes;
+  AllocationSize allocationSize;
+  allocationSize.elementFootprint = *elementFootprint;
+  allocationSize.byteSize = totalBytes;
+  return allocationSize;
 }
 
 static FailureOr<int64_t> computeAlignment(memref::AllocOp allocOp) {
@@ -163,7 +184,8 @@ static FailureOr<int64_t> computeAlignment(memref::AllocOp allocOp) {
 }
 
 static LogicalResult
-rejectUnsupportedWorkgroupLocalMemoryUses(mlir::FunctionOpInterface funcOp) {
+collectWorkgroupLocalAllocs(mlir::FunctionOpInterface funcOp,
+                            SmallVectorImpl<memref::AllocOp> &localAllocs) {
   if (llvm::any_of(
           funcOp.getArgumentTypes(),
           [](Type type) { return hasWorkgroupLocalMemorySpace(type); }) ||
@@ -175,12 +197,31 @@ rejectUnsupportedWorkgroupLocalMemoryUses(mlir::FunctionOpInterface funcOp) {
   }
 
   WalkResult result = funcOp.walk([&](Operation *op) -> WalkResult {
-    if (auto allocaOp = dyn_cast<memref::AllocaOp>(op)) {
-      if (hasWorkgroupLocalMemorySpace(allocaOp.getType())) {
-        allocaOp.emitOpError(
-            "workgroup local memory is only supported for memref.alloc");
+    for (Region &region : op->getRegions()) {
+      for (Block &block : region) {
+        if (llvm::any_of(block.getArgumentTypes(), [](Type type) {
+              return hasWorkgroupLocalMemorySpace(type);
+            })) {
+          op->emitOpError(
+              "workgroup local memory is only supported for memref.alloc "
+              "results");
+          return WalkResult::interrupt();
+        }
+      }
+    }
+
+    for (Value result : op->getResults()) {
+      if (!hasWorkgroupLocalMemorySpace(result.getType())) {
+        continue;
+      }
+      auto allocOp = dyn_cast<memref::AllocOp>(op);
+      if (!allocOp || result != allocOp.getMemref()) {
+        op->emitOpError(
+            "workgroup local memory is only supported for memref.alloc "
+            "results");
         return WalkResult::interrupt();
       }
+      localAllocs.push_back(allocOp);
     }
     return WalkResult::advance();
   });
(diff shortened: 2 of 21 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+namespace {
+
+constexpr int64_t kMinWorkgroupLocalAlignment = 16;
+
+struct LLVMCPUAssignWorkgroupLocalMemoryPass
+    : impl::LLVMCPUAssignWorkgroupLocalMemoryPassBase<
+          LLVMCPUAssignWorkgroupLocalMemoryPass> {
+  using impl::LLVMCPUAssignWorkgroupLocalMemoryPassBase<
+      LLVMCPUAssignWorkgroupLocalMemoryPass>::
+      LLVMCPUAssignWorkgroupLocalMemoryPassBase;
```
hanhanW:
```suggestion
  using Base::Base;
```

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+  return success();
+}
+
+static FailureOr<int64_t>
+computeStaticElementFootprint(memref::AllocOp allocOp) {
+  MemRefType type = allocOp.getType();
+  if (!type.hasStaticShape()) {
+    allocOp.emitOpError(
+        "workgroup local memory allocations must have static shape");
+    return failure();
```
hanhanW, line 88: It can be a single line. Same for the below code.

```suggestion
    return allocOp.emitOpError(
        "workgroup local memory allocations must have static shape");
```

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+  if (offset < 0 ||
+      llvm::any_of(strides, [](int64_t stride) { return stride < 0; })) {
+    allocOp.emitOpError(
+        "workgroup local memory allocations must have non-negative layout");
+    return failure();
+  }
+
+  if (llvm::is_contained(type.getShape(), 0)) {
+    return 0;
+  }
```
hanhanW: This looks weird. I think we can make it an assertion, because it looks wrong to me when it happens.

(I was thinking if this is for 0-D memref cases, but then I realized that `llvm::is_contained({}, 0)` is false.)

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+    return failure();
+  }
+
+  DataLayout dataLayout = DataLayout::closest(allocOp);
+  llvm::TypeSize elementByteSize =
+      dataLayout.getTypeSize(type.getElementType());
+  if (elementByteSize.isScalable()) {
+    allocOp.emitOpError("workgroup local memory allocations must have "
+                        "fixed-size element types");
+    return failure();
```
hanhanW, line 147: ditto

```suggestion
    return allocOp.emitOpError("workgroup local memory allocations must have "
                        "fixed-size element types");
```

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+  return totalBytes;
+}
+
+static FailureOr<int64_t> computeAlignment(memref::AllocOp allocOp) {
+  if (std::optional<uint64_t> explicitAlignment = allocOp.getAlignment()) {
+    if (*explicitAlignment >
+        static_cast<uint64_t>(std::numeric_limits<int64_t>::max())) {
+      allocOp.emitOpError(
+          "workgroup local memory allocation alignment overflow");
+      return failure();
```
hanhanW, line 173: ditto

```suggestion
      return allocOp.emitOpError(
          "workgroup local memory allocation alignment overflow");
```

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+static LogicalResult
+rejectUnsupportedWorkgroupLocalMemoryUses(mlir::FunctionOpInterface funcOp) {
+  if (llvm::any_of(
+          funcOp.getArgumentTypes(),
+          [](Type type) { return hasWorkgroupLocalMemorySpace(type); }) ||
+      llvm::any_of(funcOp.getResultTypes(), [](Type type) {
+        return hasWorkgroupLocalMemorySpace(type);
+      })) {
```
hanhanW, line 198: I think we may not need lambda here. (need formatting)

```suggestion
  if (llvm::any_of(
          funcOp.getArgumentTypes(), hasWorkgroupLocalMemorySpace(type)) ||
      llvm::any_of(funcOp.getResultTypes(), hasWorkgroupLocalMemorySpace(type))) {
```

compiler/src/iree/compiler/Codegen/LLVMCPU/test/check_ir_before_llvm_conversion.mlir
```diff
@@ -65,3 +65,12 @@ func.func @complex_alloca() {
   return
 }
 // CHECK-LABEL: func @complex_alloca(
+
+// -----
+
+func.func @workgroup_local_alloc_not_counted() {
```
MaheshRavishankar: I am not sure what this test is checking.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+    return funcOp.emitOpError(
+        "workgroup local memory is only supported for memref.alloc results");
+  }
+
+  WalkResult result = funcOp.walk([&](Operation *op) -> WalkResult {
+    if (auto allocaOp = dyn_cast<memref::AllocaOp>(op)) {
```
MaheshRavishankar: Do we want to make this a stronger check where it checks for any operation that is not `AllocOp` and aborts?

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+    return;
+  }
+
+  std::optional<IREE::HAL::ExecutableExportOp> exportOp = getEntryPoint(funcOp);
+  if (!exportOp) {
+    localAllocs.front().emitOpError(
+        "workgroup local memory allocations are only supported in HAL "
+        "executable exports");
+    return signalPassFailure();
+  }
```
hanhanW, line 257: We've been moving to module-scope for codegen pipeline, and Codegen only touch HAL ops in [CodegenConfigurationPreProcessing phase](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/Passes.cpp#L16-L19) and [CodegenTranslationPostProcessing phase](https://github.com/iree-org/iree/blob/a0648355883a3cb3bf7fb0b0b30ffcf5f8c03905/compiler/src/iree/compiler/Codegen/Common/Passes.cpp#L22-L25). All the transformations in the between work on either module scope or function scope and they touch Codegen::DispatchConfig op instead.

I think we should attach the attribute (or have an optional field) on DispatchConfigOp (if exists), and we propagate the information in PropagateDispatchConfigPass.

https://github.com/iree-org/iree/blob/8b844e7a3d29210c89c4ac933b79983e95b59c54/compiler/src/iree/compiler/Codegen/Common/PropagateDispatchConfig.cpp#L83-L96

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+  mlir::FunctionOpInterface funcOp = getOperation();
+  if (funcOp.getFunctionBody().empty()) {
+    return;
+  }
+  if (failed(rejectUnsupportedWorkgroupLocalMemoryUses(funcOp))) {
+    return signalPassFailure();
+  }
+
+  SmallVector<memref::AllocOp> localAllocs;
+  funcOp.walk([&](memref::AllocOp allocOp) {
```
MaheshRavishankar: I think you can club this collection with the "rejectUnsupportedWorkgroupLocalMemoryUses" function, i.e. make it return the `localAllocs`.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+    if (failed(byteSize)) {
+      return signalPassFailure();
+    }
+
+    FailureOr<int64_t> alignment = computeAlignment(allocOp);
+    if (failed(alignment)) {
+      return signalPassFailure();
+    }
+    uint64_t alignedOffset = llvm::alignTo(static_cast<uint64_t>(currentOffset),
+                                           static_cast<uint64_t>(*alignment));
```
hanhanW: Maybe just make `computeAlignment` return `FailureOr<uint64_t>` type, so we don't cast in the function and here.

Honestly, I think we can just use signed int instead for all the cases. Allocating a huge buffer in workgroup-local scope looks not a case; mixed signedness usually introduces unnecessary issues or drops performance, IMO.

> Unsigned integers are good for representing bitfields and modular arithmetic. Because of historical accident, the C++ standard also uses unsigned integers to represent the size of containers - many members of the standards body believe this to be a mistake, but it is effectively impossible to fix at this point. The fact that unsigned arithmetic doesn't model the behavior of a simple integer, but is instead defined by the standard to model modular arithmetic (wrapping around on overflow/underflow), means that a significant class of bugs cannot be diagnosed by the compiler. In other cases, the defined behavior impedes optimization.
> 
> That said, mixing signedness of integer types is responsible for an equally large class of problems. The best advice we can provide: try to use iterators and containers rather than pointers and sizes, try not to mix signedness, and try to avoid unsigned types (except for representing bitfields or modular arithmetic). Do not use an unsigned type merely to assert that a variable is non-negative.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
```diff
@@ -73,6 +73,8 @@ constexpr StringLiteral kUkernelAttrName = "iree_codegen.ukernel";
 constexpr StringLiteral kUKernelProviderName = "iree_codegen.ukernel_provider";
 constexpr StringLiteral kVectorTileSizesAttrName =
     "iree_codegen.vector_tile_sizes";
+constexpr StringLiteral kWorkgroupLocalMemoryRangeAttrName =
```
MaheshRavishankar: Note to self: Such names shouldnt be exposed this way. Not needed to fix, but this needs to be changes to use hide the actual string.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
```diff
@@ -0,0 +1,269 @@
...
+    if (alignedOffset >
+        static_cast<uint64_t>(std::numeric_limits<int64_t>::max())) {
+      allocOp.emitOpError("workgroup local memory allocation size overflow");
+      return signalPassFailure();
+    }
+    currentOffset = static_cast<int64_t>(alignedOffset);
+    allocOp->setAttr(kWorkgroupLocalMemoryRangeAttrName,
```
MaheshRavishankar: This can be tricky cause some "canonicalization" might drop the attribute, and then it will lead to correctness issues. Maybe need a different solution here? Instead of doing this can we create a single alloc and replace these with `memref.view`. The view carries the offset and then we just lowers the views?

MaheshRavishankar: Another option, just use a custom op for the alloc where we can just keep track of the attribute as a native operand. Then everything works better.

hanhanW (commented):
Looking forward to the feature, LGTM. Just some nits and a concern about touching export op.

MaheshRavishankar (changes requested):
Overall looks fine, but using a free-form attribute can lead to correctness issues.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
hanhanW: I left a comment about adding a field and using DispatchConfig op here: https://github.com/iree-org/iree/pull/24279#discussion_r3155994942

Adding a new field is better, as we can preserve it as long as possible (until propagating it to the ExportOp).

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Done: follow-up commit uses Base::Base.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Kept emitOpError plus failure because these helpers return FailureOr values, not LogicalResult. The diagnostics are still emitted at the failing op, and clang-format cleaned up the wrapping.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Clarified this instead of asserting: zero-sized memrefs are legal and need a zero-byte footprint. Added hasZeroElementShape and a positive zero-sized allocation test.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Kept emitOpError plus failure because this helper returns a FailureOr value. The diagnostic still points at the invalid allocation.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Kept emitOpError plus failure because this helper returns a FailureOr value. The diagnostic still points at the invalid allocation.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Kept explicit lambdas because the predicate is overloaded and the argument type matters here; clang-format now owns the shape.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/check_ir_before_llvm_conversion.mlir
benvanik: Clarified the test name and comment: workgroup-local memref.alloc is backed by HAL dispatch local memory and must not count against the stack limit.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Done: collection now rejects any workgroup-local value not produced by memref.alloc, including block arguments and non-alloc op results.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Done: the pass writes workgroup_local_memory on iree_codegen.dispatch_config. PropagateDispatchConfig publishes it to the HAL export, and CreateDispatchConfig preserves existing export metadata.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Done: collection and rejection are now one collectWorkgroupLocalAllocs helper returning the local alloc list.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Done: layout arithmetic now stays signed with checkedAlignTo, avoiding the uint64_t alignTo cast path.

compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h
benvanik: The exposed free-form range attribute is gone entirely after switching to packed IR, so this string name is no longer part of the surface.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUAssignWorkgroupLocalMemory.cpp
benvanik: Done: replaced per-alloc range attributes with one packed i8 workgroup-local allocation and memref.view slices. Static strided layouts are recovered with memref.reinterpret_cast. The aggregate byte requirement now lives on dispatch_config until propagation.

hanhanW (approved):

PR #24286, opened 2026-04-28
Add dlmalloc third-party source.

benvanik:
Imports Doug Lea's malloc v2.8.6 as a third-party source dependency. The source comes from the public dlmalloc distribution at https://gee.cs.oswego.edu/dl/html/malloc.html.

The upstream file identifies itself as version 2.8.6 from August 29, 2012, with the September 25, 2023 relicensing that replaces the obsolete CC0 grant with MIT-0. Keep the upstream license text in third_party/dlmalloc/LICENSE and leave malloc.c otherwise self-contained so provenance remains auditable in the standalone third-party commit.

Expose the source as a Bazel local repository and as a CMake interface target. The allocator is compiled by platform-specific configuration wrappers instead of directly by this commit, which lets wasm and any future platform select dlmalloc options without forking the imported source.

dlmalloc is chosen here as it is stable and can be tuned for our wasm usage (and until recently was the only blessed wasm allocator). In the future we may want to play with emmalloc but it is currently tangled in the emscripten SDK which we are deliberately avoiding taking a dependency on. A small dep on a single .c that is known as the best reference malloc one can get is fine for now. A user embedding IREE into an existing wasm app can bring their own allocator and libc.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -75,6 +75,7 @@ use_repo(
     "com_github_dvidelabs_flatcc",
     "com_google_benchmark",
     "com_google_googletest",
+    "dlmalloc",
     "hip_api_headers",
     "hsa_runtime_headers",
     "iree_cuda",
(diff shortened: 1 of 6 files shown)


ScottTodd (approved):
New third_party dep LGTM, no concerns with the license or build system integration :)

PR #24289, opened 2026-04-28
[Codegen][CPU] Materialize data-tiled encodings to inner_tiled.

bjacob:
When a target sets `enable_inner_tiled = true`, lower matmuls with encodings to `iree_codegen.inner_tiled` whose kind is a CPU `data_tiled_mma_layout`, instead of `linalg.mmt4d`.

The intrinsic and its orientation (`transposed_intrinsic`) are chosen jointly by a small cost model that maximizes "useful ops" per invocation (charging an intrinsic for the padding it would force along statically narrow M/N), with the intrinsic's raw M*N*K as tiebreaker.

Unroll factors are hard-coded to `intrinsics_m = intrinsics_n = 1` for now; a follow-up commit introduces the register-budget/arithmetic- intensity cost model that picks better values.

Also:
- Add the AVX2+FMA f32 intrinsic (`MMA_X86_AVX2_FMA_1x8x1_F32_F32`) to enabling writing additional tests.
- Have MaterializeEncodingPatterns emit a clean op-level error if a layout lowering returns nullptr, rather than segfaulting.

AI: Claude Opus 4.7

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
@@ -662,6 +662,10 @@ public:
     }
     Operation *newOp =
         layoutAttr.lowerOp(rewriter, op, convertedResTypes, operands);
+    if (!newOp) {
+      return rewriter.notifyMatchFailure(
+          op, "unsupported operation for encoding materialization");
+    }
     rewriter.replaceOp(op, newOp->getResults());
     return success();
   }
--- a/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
@@ -148,6 +148,98 @@ func.func @pack_gemm_fill_dynamic(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf3
 
 // -----
 
+// With `enable_inner_tiled` and +avx512f (TileMxNxK 16x16x1), matmul lowers to
+// iree_codegen.inner_tiled instead of linalg.mmt4d.
+
+#map_it = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map_it1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map_it2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+#encoding_lhs_it = #iree_encoding.encoding<operand_index = 0, op_type = matmul, element_types = [f32, f32, f32], user_indexing_maps = [#map_it, #map_it1, #map_it2], iteration_sizes = [?, ?, ?]>
+#encoding_rhs_it = #iree_encoding.encoding<operand_index = 1, op_type = matmul, element_types = [f32, f32, f32], user_indexing_maps = [#map_it, #map_it1, #map_it2], iteration_sizes = [?, ?, ?]>
+#encoding_result_it = #iree_encoding.encoding<operand_index = 2, op_type = matmul, element_types = [f32, f32, f32], user_indexing_maps = [#map_it, #map_it1, #map_it2], iteration_sizes = [?, ?, ?]>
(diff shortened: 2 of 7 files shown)


compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
```diff
@@ -148,6 +148,98 @@ func.func @pack_gemm_fill_dynamic(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf3
...
+// iree_codegen.inner_tiled instead of linalg.mmt4d.
+
+#map_it = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map_it1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map_it2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+#encoding_lhs_it = #iree_encoding.encoding<operand_index = 0, op_type = matmul, element_types = [f32, f32, f32], user_indexing_maps = [#map_it, #map_it1, #map_it2], iteration_sizes = [?, ?, ?]>
+#encoding_rhs_it = #iree_encoding.encoding<operand_index = 1, op_type = matmul, element_types = [f32, f32, f32], user_indexing_maps = [#map_it, #map_it1, #map_it2], iteration_sizes = [?, ?, ?]>
+#encoding_result_it = #iree_encoding.encoding<operand_index = 2, op_type = matmul, element_types = [f32, f32, f32], user_indexing_maps = [#map_it, #map_it1, #map_it2], iteration_sizes = [?, ?, ?]>
+func.func @pack_gemm_fill_dynamic_inner_tiled_avx512(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf32>, %m: index, %n: index, %k: index) -> tensor<?x?xf32> attributes {
+   hal.executable.target = #hal.executable.target<"llvm-cpu", "xyz", {target_triple = "x86_64-xyz-xyz", cpu_features = "+avx512f", enable_inner_tiled = true, iree.encoding.resolver = #iree_cpu.cpu_encoding_resolver<>}>
```
hanhanW, line 161: I like the idea of having `enable_inner_tiled` better, because we can keep everything in the same file; it does demonstrate the pros of self-contained IR. Thanks for proposing it in the first place!

compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
```diff
@@ -662,6 +662,10 @@ class MaterializeLinalgOp
     }
     Operation *newOp =
         layoutAttr.lowerOp(rewriter, op, convertedResTypes, operands);
+    if (!newOp) {
+      return op->emitOpError(
+          "unsupported operation for encoding materialization");
```
hanhanW: I think I understand why you use `emitOpError` because it is explicit. However, I'd expect `rewriter.notifyMatchFailure()`. Developers should use `debug-only` or `-debug` to get the message, IMO. This is more like internal details from pass's perspective.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/test/roundtrip.mlir
hanhanW, line 1: I don't follow why we need a roundtrip test here? The encoding can carry anything in configuration, and we don't need a test for that. Otherwise, we'd be asking all the ops/attributes that use DictionaryAttr to have such tests and they are too heavy and meaningless.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
```diff
@@ -714,14 +1000,29 @@ struct CPUEncodingPackedLayoutMaterializerAttr
       return info;
     }
 
+    DictionaryAttr config = layoutAttr.getConfiguration();
+    if (getEnableInnerTiledFromConfig(config)) {
+      return getInnerTiledEncodingInfo(type.getContext(), encoding, *cDims,
+                                       config);
+    }
+    return getMmt4dEncodingInfo(encoding, *cDims, config);
```
hanhanW, line 1008: It's okay to have the "fork" as we are moving towards inner tiles approach. It is mostly a note for future contributors and reviewers.

hanhanW (approved):
LGTM, just a couple nits and a question about roundtrip test.

bjacob:
Thanks, I have applied both your suggestions (notifyMatchFailure and reverted the round_trip.mlir test).

PR #24290, opened 2026-04-28
[HAL] Fix HALToStream import dims test after alias lowering

benvanik:
The import dims test was checking that the final clone consumed the original buffer-view import directly. That made the test depend on incidental lowering shape around hal.tensor.alias instead of the behavior it was meant to protect.

Keep the assertion that consumer-refined dimensions reach the stream imports and sizes, but match the existing alias lowering through cast, update, slice, and clone. This structure is useful as it verifies the contract is maintained where previously the test did not.

Verified that this runs through the stream pipeline and melts away:
```mlir
  %0 = stream.tensor.import %arg0 : !hal.buffer_view -> tensor<?x?xf32>{%c128}
  %1 = stream.tensor.import consume %arg1 : !hal.buffer -> tensor<?x?xf32>{%c128}
  %2 = stream.cmd.execute with(%1 ..., %0 ...) {
    stream.cmd.copy %arg3[%c0], %arg2[%c0], %c128 : !stream.resource<external>{%c128} -> !stream.resource<external>{%c128}
  } => !stream.timepoint
  %3 = stream.timepoint.await %2 => %1
  util.return %3, %c128
```


amd-eochoalo (approved):
Thanks!

hanhanW (approved):

PR #24299, opened 2026-04-29
[VectorDistribute] Lower and distribute `async_dma`

sommerlukas:
Pass to distribute and lower `async_dma` operations at the workgroup level to `amdgpu.gather_to_lds` operations at the thread-level (with threads in each subgroup collaborating).

The pass shares helpers with the existing GPU pass to distribute operations based on layouts, but as the `async_dma` operation does not have `vector` operands or results, lowering and distribution are implemented as a separate pass. The changes to `GPUNestedLayoutDistributionPatterns.cpp` are therefore mainly a code move extracting shared helpers to the new `GPUNestedLayoutUtils.[h|cpp]`.

The basic idea of the distribution is to construct a (nested) layout that represents how the data-transfer is split across subgroups and threads to perform the full transfer with direct-to-LDS compatible operations. The layout is constructed in stages:
1. We choose the DMA size for the given target that fulfills the requirements and determine the element tile based on the size of the transfer per thread from the DMA size (`distributeFromInnermost`).
2. The element tile is given by the number of threads in subgroup (`distributeFromInnermost`).
3. Outer tile is always all-ones.
4. We distribute the transfer to the configured number of subgroups (`distributeFromOutermost`).
5. Whatever is left after these steps ends up as the batch tile of each thread.

Once we have that layout, we can use the shared helpers for the mechanics of distributing the operation. 

The distribution fails if any of the requirements are not met. This is mostly a defensive check, the pass inserting the `async_dma` operations (will be added in a different PR) should only insert `async_dma` operations if the prerequisites can be met with the available DMA sizes for the transfer shape etc. Therefore, the pass also fails if any of the `async_dma` operations could not be distributed and lowered.

Swizzling and gather semantics are not part of this PR and will be added in follow-up PRs.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPULowerAsyncDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPULowerAsyncDMA.cpp
@@ -1,0 +1,343 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <cstdint>
+#include "iree/compiler/Codegen/Common/GPU/GPUNestedLayoutUtils.h"
+#include "iree/compiler/Codegen/Common/GPU/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
+#include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.h"
+#include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.h"
+#include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
+#include "iree/compiler/Codegen/Utils/GPUUtils.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/DebugLog.h"
+#include "mlir/Dialect/AMDGPU/IR/AMDGPUDialect.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/Utils/IndexingUtils.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/Transforms/WalkPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-codegen-gpu-lower-async-dma"
+
+namespace mlir::iree_compiler {
+
+#define GEN_PASS_DEF_GPULOWERASYNCDMAPASS
+#include "iree/compiler/Codegen/Common/GPU/Passes.h.inc"
+
+using namespace IREE::VectorExt;
+
+namespace {
+
+/// Distribute |total| across |shape| from the innermost dimension.
+/// |shape| is updated in place to reflect the remaining shape after
+/// distribution. Returns failure if any dimension doesn't divide evenly.
+static FailureOr<SmallVector<int64_t>>
+distributeFromInnermost(int64_t total, MutableArrayRef<int64_t> shape) {
+  int64_t rank = shape.size();
+  SmallVector<int64_t> result(rank, 1);
+  int64_t remaining = total;
+  for (int64_t i = rank - 1; i >= 0 && remaining > 1; --i) {
+    int64_t take = std::min(remaining, shape[i]);
+    if (shape[i] % take != 0) {
+      return failure();
+    }
+    result[i] = take;
+    shape[i] /= take;
+    remaining /= take;
+  }
+  if (remaining != 1) {
+    return failure();
+  }
+  return result;
+}
+
+/// Distribute |total| across |shape| from the outermost dimension.
+/// |shape| is updated in place to reflect the remaining shape after
+/// distribution. Returns failure if any dimension doesn't divide evenly.
+static FailureOr<SmallVector<int64_t>>
+distributeFromOutermost(int64_t total, MutableArrayRef<int64_t> shape) {
+  int64_t rank = shape.size();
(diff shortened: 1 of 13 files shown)


Yu-Zhewen (commented):
Thanks for the PR! I compared your `GPULowerAsyncDMAPass` with the existing `AMDGPULowerCoalescedDMAToGatherLDSPass`. A few differences to flag, but all optional, may already be planned as follow-ups or may not apply to your use cases.

- **in_bounds handling**: Seems to be missing from the new pass. Maybe reject with notifyMatchFailure for now.

- **DMA size selection**: `getGlobalLoadDMALayout` seems to pick one DMA size for the entire transfer. The coalesced DMA pass uses `computeTransferSegments` which potentially can mix sizes (e.g., 128-bit + 32-bit).

- **strided memrefs**: Source or dest of async_dma could be strided before distribution (e.g., from a memref.subview), which affects DMA size selection. The coalesced DMA pass checks `getNumContiguousTrailingDims()` for dest. I recently extended it to also handle source strides in [#24305](https://github.com/iree-org/iree/pull/24305). Might be relevant here.

sommerlukas:
Thanks for the feedback!

> * **in_bounds handling**: Seems to be missing from the new pass. Maybe reject with notifyMatchFailure for now.

Good point, I've rejected OOB for now and added a TODO.
> * **DMA size selection**: `getGlobalLoadDMALayout` seems to pick one DMA size for the entire transfer. The coalesced DMA pass uses `computeTransferSegments` which potentially can mix sizes (e.g., 128-bit + 32-bit).

Yes, that's interesting optimization. Currently, I also force all subgroups to participate in a transfer or otherwise fall back to regular load/store. It's definitely on my TODO list to support this. 

> * **strided memrefs**: Source or dest of async_dma could be strided before distribution (e.g., from a memref.subview), which affects DMA size selection. The coalesced DMA pass checks `getNumContiguousTrailingDims()` for dest. I recently extended it to also handle source strides in [#24305](https://github.com/iree-org/iree/pull/24305). Might be relevant here.

I've added a check to reject strided memrefs for now and added a TODO to address this later. I'll probably try to reuse your code/logic for that purpose.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -838,6 +838,7 @@ void addGPUVectorDistributePassPipeline(OpPassManager &funcPassManager,
 
   // Vector SIMD -> Vector SIMT
   funcPassManager.addPass(createLLVMGPUVectorDistributePass());
+  funcPassManager.addPass(createGPULowerAsyncDMAPass());
```
kuhar: Is this guarded by `forROCM`?

kuhar (commented):
High-level comment: this seems entirely amdgpu-specific; should we move this pass over to rocdl passes instead?

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
Yu-Zhewen: +1

compiler/src/iree/compiler/Codegen/Common/GPU/GPULowerAsyncDMA.cpp
```diff
@@ -0,0 +1,365 @@
...
+                int64_t numThreads)
+      : OpRewritePattern(context), threadId(threadId),
+        subgroupSize(subgroupSize), numThreads(numThreads) {}
+
+  LogicalResult matchAndRewrite(IREE::GPU::AsyncDMAOp op,
+                                PatternRewriter &rewriter) const override {
+    LDBG() << "Processing AsyncDMAOp: " << op;
+
+    // Currently, only support lowering of operations with all-true in_bounds.
+    // TODO: Support cases where some dimensions are out-of-bounds.
```
Yu-Zhewen: Can we create a tracking issue for the TODOs in the new pass and reference that issue in the comments here?

Yu-Zhewen (approved):
LGTM

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
keshavvinayak01: @sommerlukas  This should be moved or guarded.

keshavvinayak01 (approved):
Looks okay overall

compiler/src/iree/compiler/Codegen/Common/GPU/GPULowerAsyncDMA.cpp
sommerlukas: I've added a tracking comment to https://github.com/iree-org/iree/issues/23782 and referenced it in the TODOs.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
sommerlukas: Added a guard with `forROCDL`.

sommerlukas (commented):
Thanks for the feedback all!

> High-level comment: this seems entirely amdgpu-specific; should we move this pass over to rocdl passes instead?

I've renamed the pass and corresponding files to `AMDGPULowerAsyncDMA` to make the AMD-specific nature clearer.

The existing `AMDGPULowerCoalescedDMAToGatherLDS` with a very similar purpose lives in the same directory, so I've left it in this directory. If you still prefer to move it, I'm happy to do that.

keshavvinayak01 (approved):

kuhar (commented):
No objections from my side, but I see an opportunity for future cleanup here:

> The existing AMDGPULowerCoalescedDMAToGatherLDS with a very similar purpose lives in the same directory, so I've left it in this directory. If you still prefer to move it, I'm happy to do that.

sommerlukas:
> No objections from my side, but I see an opportunity for future cleanup here:
> 
> > The existing AMDGPULowerCoalescedDMAToGatherLDS with a very similar purpose lives in the same directory, so I've left it in this directory. If you still prefer to move it, I'm happy to do that.

Thanks! I'll merge this now and move the pass in a follow-up if possible. I'll also coordinate with @Yu-Zhewen @lialan to move `AMDGPULowerCoalescedDMAToGatherLDS` if possible.

PR #24300, opened 2026-04-29
[IREEGPU] Bufferize `async_dma`

sommerlukas:
Implements bufferization for `iree_gpu.async_dma` through an external model interface implementation. 

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
@@ -516,7 +516,6 @@ struct AsyncDMAOpBufferizationInterface
       return failure();
     }
 
-    rewriter.setInsertionPoint(dmaOp);
     IREE::GPU::AsyncDMAOp::create(
         rewriter, dmaOp.getLoc(), TypeRange{}, *sourceBuffer,
         dmaOp.getSourceIndices(), *destBuffer, dmaOp.getDestIndices(),
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
@@ -461,6 +461,73 @@ struct CoalescedGatherDMAOpBufferizationInterface
   }
 };
 
+/// Bufferization of iree_gpu.async_dma. Bufferizes to itself with memref
+/// operands and no result.
+struct AsyncDMAOpBufferizationInterface
+    : BufferizableOpInterface::ExternalModel<AsyncDMAOpBufferizationInterface,
+                                             IREE::GPU::AsyncDMAOp> {
+  bool bufferizesToMemoryRead(Operation *op, OpOperand &opOperand,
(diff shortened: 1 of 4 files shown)


amd-eochoalo (approved):

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
```diff
@@ -461,6 +461,73 @@ struct CoalescedGatherDMAOpBufferizationInterface
...
+  getAliasingValues(Operation *op, OpOperand &opOperand,
+                    const AnalysisState &state) const {
+    auto dmaOp = cast<IREE::GPU::AsyncDMAOp>(op);
+    if (opOperand.get() == dmaOp.getDest() && dmaOp.hasTensorSemantics()) {
+      return {{dmaOp.getResult(), BufferRelation::Equivalent}};
+    }
+    return {};
+  }
+
+  bool mustBufferizeInPlace(Operation *op, OpOperand &opOperand,
```
krzysz00, line 491: Is this correct?

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
```diff
@@ -461,6 +461,73 @@ struct CoalescedGatherDMAOpBufferizationInterface
...
+    FailureOr<Value> sourceBuffer =
+        getBuffer(rewriter, dmaOp.getSource(), options, state);
+    FailureOr<Value> destBuffer =
+        getBuffer(rewriter, dmaOp.getDest(), options, state);
+
+    if (failed(sourceBuffer) || failed(destBuffer)) {
+      return failure();
+    }
+
+    rewriter.setInsertionPoint(dmaOp);
```
krzysz00: Do you need this?

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
sommerlukas: I would argue so, yes. The result in tensor semantics is just the destination after the copy, so it aliases the destination (allows for synchronization). With `memref` semantics, the operation doesn't have a result. The semantics after bufferization is that we copy from the source memref to the destination memref.

It is also consistent with the bufferization of `coalesced_gather_dma`: https://github.com/iree-org/iree/blob/a79bb7bfba94f7e26020d04359b95fc005ef0bb6/compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp#L397

compiler/src/iree/compiler/Codegen/Dialect/GPU/Transforms/BufferizationInterfaces.cpp
sommerlukas: Nope, thanks for catching that, removed it.

sommerlukas (commented):
Thanks for the feedback!

Yu-Zhewen (approved):
LGTM

keshavvinayak01 (approved):
LGTM

krzysz00 (approved):
lgtm

PR #24301, opened 2026-04-29
Generalize FoldMaskedTransferRaw and add FoldTransferReadOfEmptyTensor

### LINKED ISSUE #24294: Codegen: tensor.empty with index element-type is created in transfer_gather vectorization path
I improved the similar issue in https://github.com/iree-org/iree/commit/ba0d3b8e6792af8297c605555c8a15dc691424bd, but it is not enough for turning it on by default on CPU path.

The core issue is that `tensor.empty() : tensor<...xindex>` op is created, and we must have need a temp buffer for holding it. However, having memref with index type is problematic, because index is something simliar to `size_t` which is platform dependent. We don't expect such memref in codegen; the upstream version does not introduce it at all. It becomes a blocker for using the vectorization path by default.

My commit only fixed the case that it has a single gather within the generic op, but not multiple cases. To repro:

Run: `iree-opt --pass-pipeline='builtin.module(func.func(iree-codegen-generic-vectorization{use-configured-vector-sizes=true enable-vector-masking=true vectorize-to-transfer-gather=true}))' repro.mlir`

```mlir
#config = #iree_cpu.lowering_config<vector_common_parallel = [1, 8, 0], vector_inner_parallel  = [0, 0, 8]>
#m_3d = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
#m_2d = affine_map<(d0, d1, d2) -> (d0, d1)>

func.func @three_gathers_index_materialization(
    %in0: tensor<1x8x?xf32>,
    %in1: tensor<1x8xf32>, %in2: tensor<1x8xf32>,
    %out_init: tensor<1x8x?xf32>,
    %indir_table: tensor<50x32x25x2xi32>,
    %lut: tensor<50x40x40xi8>,
    %arg0: index, %arg2: index, %arg4: index, %arg6: index)
    -> tensor<1x8x?xf32> {
  %c0    = arith.constant 0   : index
  %c1    = arith.constant 1   : index
  %c39   = arith.constant 39  : index
  %c0_i8 = arith.constant 0   : i8
  %cst   = arith.constant 0.0 : f32
  %0 = linalg.generic {
        indexing_maps = [#m_3d, #m_2d, #m_2d, #m_3d],
        iterator_types = ["parallel", "parallel", "parallel"]
      } ins(%in0, %in1, %in2 : tensor<1x8x?xf32>, tensor<1x8xf32>, tensor<1x8xf32>)
        outs(%out_init : tensor<1x8x?xf32>)
        attrs = {lowering_config = #config} {
  ^bb0(%in: f32, %a: f32, %b: f32, %out: f32):
    // Small f32 preExtract chain — needed for the gather-path entry to
    // engage in the way that produces tensor.empty<...xindex>.
    %m1 = arith.divf %a, %b : f32
    %p  = arith.cmpf une, %in, %cst : f32
    %v  = arith.select %p, %m1, %cst : f32
    // Per-lane index computation shared across the first two gathers.
    %i0 = affine.apply affine_map<(d0, d1) -> (d0 + d1)>(%arg0, %arg2)
    %d1 = linalg.index 1 : index
    %i1 = affine.apply affine_map<(d0)[s0] -> (d0 + s0)>(%arg4)[%d1]
    %d2 = linalg.index 2 : index
    %i2 = affine.apply affine_map<(d0)[s0] -> (d0 + s0)>(%arg6)[%d2]
    // Gather #1 and #2: identical outer indices, last index differs.
    %ea = tensor.extract %indir_table[%i0, %i1, %i2, %c0] : tensor<50x32x25x2xi32>
    %ea_idx = arith.index_cast %ea : i32 to index
    %eb = tensor.extract %indir_table[%i0, %i1, %i2, %c1] : tensor<50x32x25x2xi32>
    %eb_idx = arith.index_cast %eb : i32 to index
    // Clamp results — these become inputs to gather #3.
    %ea_max = arith.maxsi %ea_idx, %c0  : index
    %ea_min = arith.minui %ea_max, %c39 : index
    %eb_max = arith.maxsi %eb_idx, %c0  : index
    %eb_min = arith.minui %eb_max, %c39 : index
    // Gather #3: depends on results of #1 and #2.
    %ec = tensor.extract %lut[%i0, %ea_min, %eb_min] : tensor<50x40x40xi8>
    %gate = arith.cmpi ugt, %ec, %c0_i8 : i8
    %r = arith.select %gate, %v, %cst : f32
    linalg.yield %r : f32
  } -> tensor<1x8x?xf32>
  return %0 : tensor<1x8x?xf32>
}
```

then you'll see `tensor.empty() : tensor<1x8x8xindex>` in the output.

I think the main reason is that we only have a single `preOp` in [vectorizeGatherLikeGenericToTransferGather](https://github.com/iree-org/iree/blob/819a72b7c4bb046fe9363dcb9c7ad0ec13dc1448/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L620) in this case. If we can break it into `sevearl preOp -> a sequence of gather op -> sevearl or single postOp`, we should be able to resue the value directly, instead of having `write->read` chain on the `tensor.empty()` ops.

The other approach may be passing the map in recursively vectorization.

This should allow us to get rid of "flatIdx" vectorization approach in upstream, IIUC. cc @Groverkss 

@sommerlukas @amd-eochoalo, @MaheshRavishankar recommended you for fixing the issue. Can one of you pick this up? There may be missing lowering or optimization in transfer_gather path, though. The end goal in IREE is use `transfer_gather` op in vectorization as it should be capturing higher and better semantics.

amd-eochoalo:
* Generalizes FoldMaskedTransferRAW from (masked, masked) to (unmasked, unmasked), (unmasked, masked), (masked, unmasked).
* Adds pattern to fold transfer_read(tensor.empty)) -> ub.poison
* This allows intermediary index tensors to be folded after vectorization.
* The test pipeline_vector_distribute_reduction_gfx942.mlir needed to be corrected. The empty tensors are now folded, but the test was wrong, online_attention should not have had empty tensors as operands to begin with. All passes that create online_attention fill the operands with either 0 or -1. So we do that here as well.

Fixes #24294

----

Assisted-By: Claude Opus 4.6


sommerlukas (commented):
Implementation looks OK to me at first glance. But before I review it in more detail, I'm wondering if this is the right approach to fix this issue: In other cases (e.g., `GenericVectorization`), we initially accept pairs of `transfer_write` and `transfer_read` that materialize intermediate values and later on fold them away into dataflow of vector values, e.g., in the `OptimizeTensorInsertExtractSlices` pass that adds some folding patterns. 

What is the reason this doesn't work in this case? Is this a missing fold pattern or a failure of the existing patterns?

amd-eochoalo:
@sommerlukas thanks for the review!

> I'm wondering if this is the right approach to fix this issue:

(Non-technical) I'm not too familiar with this area of the code base so I went with this solution because it looks like the least intrusive and was also suggested in the ticket.

(Technical) `vectorizeGatherLikeGenericToTransferGather` is a recursive function. Having a fold pattern is a bit difficult with the current design because there is no "spot' in between recursions where the fold pattern can be scheduled easily. If we change the design of this function to be more like just a pattern mixed with folders in a greedy application fashion, that would be a bit more involved. 

> What is the reason this doesn't work in this case? Is this a missing fold pattern or a failure of the existing patterns?

I looked a little bit more into this (reverting the fix proposed here and looking at folding patterns as a cleanup stage). It looks like adding folding patterns is more work than expected (and would involve calling OptimizeTensorInsertExtractSlices). For example, the IR at some point will be:


```mlir
                                // value_to_store, base, indices, mask 
%26 = vector.transfer_write %25, %5[%c0, %c0, %c0], %10 {in_bounds = [true, true, true]} : vector<1x8x8xindex>, tensor<1x8x8xindex>
%27 = vector.transfer_write %20, %6[%c0, %c0, %c0], %10 {in_bounds = [true, true, true]} : vector<1x8x8xindex>, tensor<1x8x8xindex>
%29 = vector.transfer_write %28, %7[%c0, %c0, %c0], %10 {in_bounds = [true, true, true]} : vector<1x8x8xindex>, tensor<1x8x8xindex>
``` 

```mlir

                                      // base, indices, padding
%32 = vector.transfer_read %26[%c0, %c0, %c0], %2 {in_bounds = [true, true, true]} : tensor<1x8x8xindex>, vector<1x8x8xindex>
%33 = vector.transfer_read %27[%c0, %c0, %c0], %2 {in_bounds = [true, true, true]} : tensor<1x8x8xindex>, vector<1x8x8xindex>
%34 = vector.transfer_read %29[%c0, %c0, %c0], %2 {in_bounds = [true, true, true]} : tensor<1x8x8xindex>, vector<1x8x8xindex>
```

which seems like could be handled by FoldMaskedTransferRaw

But notice that FoldMaskedTransferRAW, expects the reads to also be masked, which they aren't in this case. Furthermore, to clean them up, we would still need another pattern, which would need to reason about the fact that %5, %6, %7 are empty tensors which would mean that the masked values are poison. Finally, without this change the index vectors used by iree_vector_ext.transfer_gather would need to be reconstructed from vector.transfer_writes that were introduced by the upstream vectorization. I think this is the simplest fix.


So, patterns we would need:

1. add masked asymmetry support to FoldMaskedTransferRAW
2. transfer_read(tensor.empty) to poison
3. select(mask, x, poison) to x
4. transfer_gather(x, broadcast(indices)) -> transfer_gather(x, indices)

amd-eochoalo:
@sommerlukas I chatted with @krzysz00 and he said that some of 2, 3 are patterns that are likely already implemented, that the first one is something that should be a fix regardless of this issue, and 4 may be simpler than I thought. So, I will get started on that. Thank you both for helping out with this!

amd-eochoalo:
@krzysz00 @sommerlukas thanks for the suggestions! I just updated it by adding the folding patterns and letting OptimizeTensorInsertExtractSlices handle it.

EDIT: Looking into the failing test.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -307,26 +307,121 @@ struct FoldMaskedTransferRAW : OpRewritePattern<vector::TransferReadOp> {
...
     TypedValue<VectorType> wMask = writeOp.getMask();
-    Value rPad = op.getPadding();
+    TypedValue<VectorType> rMask = op.getMask();
 
-    // Match only if the write and read op are masked and have the same mask.
-    if (!wMask || (wMask != op.getMask())) {
+    if (wMask && rMask) {
+      // Both masked: masks must match.
+      if (wMask != rMask) {
+        return failure();
```
krzysz00: I don't think this is true - you can actually handle this by selecting with the write mask and then selecting again with the read mask

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -307,26 +307,121 @@ struct FoldMaskedTransferRAW : OpRewritePattern<vector::TransferReadOp> {
...
+      // but handle it for completeness.
+      rewriter.replaceOpWithNewOp<vector::TransferReadOp>(
+          op, op.getType(), writeOp.getBase(), op.getIndices(),
+          op.getPermutationMap(), op.getPadding(), rMask, op.getInBoundsAttr());
+    }
+    return success();
+  }
+};
+
+/// Folds transfer_read(tensor.empty) into ub.poison (or, for out-of-bounds
```
krzysz00: This feels like it should be upstream, no?

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -307,26 +307,121 @@ struct FoldMaskedTransferRAW : OpRewritePattern<vector::TransferReadOp> {
...
+///     %poison = ub.poison : vector<128xf16>
+///     %bcast  = vector.broadcast %pad : f16 to vector<128xf16>
+///     %r = arith.select %mask, %poison, %bcast
+///
+///   Case 3 — has out-of-bounds dims, no mask:
+///     %e = tensor.empty() : tensor<128xf16>
+///     %r = vector.transfer_read %e[%c0], %pad
+///   ->
+///     %r = ub.poison : vector<128xf16>
+///     (out-of-bounds lanes are filled with pad, but since the tensor contents
```
krzysz00: I don't think the logic is correct here. Let's go re-check the transfer_read semantics - if they're "reading off the edge with in_bounds = false gets you the padding value" then we need to do that. If it's "implementation/lowering-defined" then we can go to poison

krzysz00 (commented):
Flagging that the second pattern there is probably upstreamable, and we should double-check, and that the first pattern can, I think, be simplified by applying both write and read masks

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
amd-eochoalo: Thank you! This change made it a lot better!

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
amd-eochoalo: I'll upstream it after merging this here.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_vector_distribute_reduction_gfx942.mlir
```diff
@@ -289,6 +289,11 @@ func.func @attention_20x1x64x4096x64() attributes {hal.executable.target = #exec
   %6 = iree_tensor_ext.dispatch.tensor.load %2, offsets = [0, 0, 0], sizes = [20, 4096, 64], strides = [1, 1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<20x4096x64xf16>> -> tensor<20x4096x64xf16>
   %7 = tensor.empty() : tensor<20x1x64xf32>
   %8 = tensor.empty() : tensor<20x1xf32>
```
krzysz00, line 291: Unrelated changes?

krzysz00 (approved):
Implementation looks good to me, one comment re PR

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/pipeline_vector_distribute_reduction_gfx942.mlir
amd-eochoalo: No, these are not unrelated changes. So this lit test was not semantically correct, and these new folders will fold these empty tensors. iree_linalg_ext.online_attention receives three "empty" tensors which work as accumulators. If they are empty, then, these are just garbage values. They should be 0 and -inf. All places where iree_linalg_ext.online_attention are created, will have these tensors be filled with these values. In these tests, I believe it was an oversight.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -256,77 +256,177 @@ struct CastLikeInsertSliceOpFolder final
...
+             "masked transfers on vector element types are not supported; see "
+             "verifyTransferOp in upstream MLIR VectorOps.cpp");
+      Value padVal = vector::BroadcastOp::create(rewriter, rPad.getLoc(),
+                                                 valToStore.getType(), rPad);
+      rewriter.replaceOpWithNewOp<arith::SelectOp>(readOp, rMask, inner,
+                                                   padVal);
+    } else {
+      rewriter.replaceOp(readOp, inner);
+    }
+    return success();
```
sommerlukas: Nit: I'd probably invert the logic, so make the replacement `if(!rMask)` and then return early in line with [LLVM policy](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code).

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -256,77 +256,177 @@ struct CastLikeInsertSliceOpFolder final
...
+             "see verifyTransferOp in upstream MLIR VectorOps.cpp");
+      Value poison = ub::PoisonOp::create(rewriter, op.getLoc(), op.getType());
+      Value padVal = vector::BroadcastOp::create(rewriter, rPad.getLoc(),
+                                                 op.getType(), rPad);
+      rewriter.replaceOpWithNewOp<arith::SelectOp>(op, mask, poison, padVal);
+    } else if (!mask && fullyInBounds) {
+      // Unmasked, fully in-bounds: every lane reads unspecified contents.
+      rewriter.replaceOp(
+          op, ub::PoisonOp::create(rewriter, op.getLoc(), op.getType()));
+    } else {
```
sommerlukas: Nit: Similar to above, I'd probably work with early returns here instead of `else [if]`, as suggested by LLVM coding standards.

sommerlukas (approved):
Looks good, just two minor code style nits.

PR #24303, opened 2026-04-29
[Codegen][CPU] Pick inner-tiled unroll factors from a register budget.

bjacob:
Replace the hard-coded `intrinsics_m = intrinsics_n = 1` from #24289
with a simple cost model based on maximizing arithmetic intensity
under the constraint of fitting in register space.

AI: Claude Opus 4.7

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
@@ -175,7 +175,7 @@ func.func @pack_gemm_fill_dynamic_inner_tiled_avx512(%arg0 : tensor<?x?xf32>, %a
   %5 = iree_encoding.unset_encoding %4 encoding_dims{%m, %n, %k} : tensor<?x?xf32, #encoding_result_it> -> tensor<?x?xf32>{%d0, %d1}
   return %5 : tensor<?x?xf32>
 }
-//   CHECK-DAG: #[[$MAP_N:.+]] = affine_map<()[s0] -> (s0 ceildiv 16)>
+//   CHECK-DAG: #[[$MAP_INNER:.+]] = affine_map<()[s0] -> (s0 ceildiv 16)>
 // CHECK-LABEL: func @pack_gemm_fill_dynamic_inner_tiled_avx512(
 //  CHECK-SAME:     %[[ARG0:[a-zA-Z0-9]+]]: tensor<?x?xf32>
 //  CHECK-SAME:     %[[ARG1:[a-zA-Z0-9]+]]: tensor<?x?xf32>
@@ -183,15 +183,16 @@ func.func @pack_gemm_fill_dynamic_inner_tiled_avx512(%arg0 : tensor<?x?xf32>, %a
 //   CHECK-DAG:   %[[C1:.+]] = arith.constant 1 : index
 //   CHECK-DAG:   %[[D0:.+]] = tensor.dim %[[ARG0]], %[[C0]]
 //   CHECK-DAG:   %[[D1:.+]] = tensor.dim %[[ARG1]], %[[C1]]
-//   CHECK-DAG:   %[[OUT_D1:.+]] = affine.apply #[[$MAP_N]]()[%[[D1]]]
+//   CHECK-DAG:   %[[OUT_D0:.+]] = affine.apply #[[$MAP_INNER]]()[%[[D0]]]
+//   CHECK-DAG:   %[[OUT_D1:.+]] = affine.apply #[[$MAP_INNER]]()[%[[D1]]]
 //   CHECK-DAG:   %[[PACK_LHS:.+]] = linalg.pack {{.*}}%[[ARG0]]
 //       CHECK:   %[[PACK_RHS:.+]] = linalg.pack
 //  CHECK-SAME:     %[[ARG1]]
-//   CHECK-DAG:   %[[EMPTY:.+]] = tensor.empty(%[[D0]], %[[OUT_D1]]) : tensor<?x?x1x16xf32>
+//   CHECK-DAG:   %[[EMPTY:.+]] = tensor.empty(%[[OUT_D0]], %[[OUT_D1]]) : tensor<?x?x16x16xf32>
 //       CHECK:   %[[FILL:.+]] = linalg.fill
 //  CHECK-SAME:       outs(%[[EMPTY]] :
 //       CHECK:   %[[INNER:.+]] = iree_codegen.inner_tiled ins(%[[PACK_LHS]], %[[PACK_RHS]]) outs(%[[FILL]])
-//  CHECK-SAME:       kind = #iree_cpu.data_tiled_mma_layout<intrinsic = MMA_X86_AVX512_1x16x1_F32_F32>, semantics = #iree_cpu.mma_semantics<>
+//  CHECK-SAME:       kind = #iree_cpu.data_tiled_mma_layout<intrinsic = MMA_X86_AVX512_1x16x1_F32_F32, intrinsics_m = 16>, semantics = #iree_cpu.mma_semantics<>
 //       CHECK:   %[[UNPACK:.+]] = linalg.unpack %[[INNER]]
 //       CHECK:   return %[[UNPACK]]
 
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -352,6 +352,28 @@ getRowMajorTilesMNKShape(MMAIntrinsic intrinsic) {
...
 
+int64_t getRegisterSpaceBytes(MMAIntrinsic intrinsic) {
+  // Total architectural vector register file size, in bytes. The inner-tiled
+  // cost model uses this as the capacity for the union of the ACC, LHS and
+  // RHS tiles. For scalable ISAs we treat the vector length as its minimum
+  // (1 × 128 bits = 16 bytes per register); this is a deliberate
+  // simplification — the resulting `intrinsics_m`/`intrinsics_n` choices are
+  // good enough in practice and avoid propagating scalability into the cost
+  // model.
+  uint32_t arch = static_cast<uint32_t>(intrinsic) & 0xFF00;
```
egebeysel, line 363: nit: hiding this behind an inlined helper function would increase readability

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
```diff
@@ -469,11 +470,76 @@ chooseIntrinsic(MLIRContext *ctx, ArrayRef<Type> elementTypes,
...
+  int64_t accTerm = info->intrinsicM * info->intrinsicN * info->accBits;
+  int64_t lhsTerm = info->intrinsicM * info->intrinsicK * info->lhsBits;
+  int64_t rhsTerm = info->intrinsicN * info->intrinsicK * info->rhsBits;
+  std::optional<std::pair<int64_t, int64_t>> best;
+  double bestIntensity = -1.0;
+  // Enumerate power-of-two intrinsicsM; for each, pick the largest feasible
+  // power-of-two intrinsicsN under the bit budget and the static N cap.
+  // The budget bounds im on its own (im*lhsTerm alone must be < budget),
+  // which terminates the loop without any numRegs-style cap.
+  for (int64_t im = 1; im <= capMPo2; im *= 2) {
```
egebeysel, line 519: Although I haven't benchmarked the effect of choosing this specific unrolling factor/inner tile size for  RISCV for end-to-end models - I believe the 10xE folks who added the inner tile sizes have benchmarked matmul-only workloads, but not e2e models where the choice of these non-pow2 inner tile sizes affect other ops after fusion and therefore might not get the best overall performance - I'd still like to note that limiting these to power of 2 sizes would not allow direct porting of those layouts:

https://github.com/iree-org/iree/blob/0380544b17da75c1e757e96bd747aa326fe3675c/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp#L721-L732

I'd perform that performance investigation later on and report back when I can, this is JFYI.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
```diff
@@ -493,8 +559,13 @@ chooseCpuInnerTiledMmaForEncoding(MLIRContext *ctx,
...
   auto [intr, transposed] = *intrChoice;
-  return IREE::CPU::DataTiledMMAAttr::get(ctx, intr, /*intrinsics_m=*/1,
-                                          /*intrinsics_n=*/1,
+  std::optional<std::pair<int64_t, int64_t>> unroll =
+      chooseUnrolling(ctx, elementTypes, intr, transposed, *matmulSizes);
+  if (!unroll) {
+    return {};
+  }
+  auto [intrinsicsM, intrinsicsN] = *unroll;
+  return IREE::CPU::DataTiledMMAAttr::get(ctx, intr, intrinsicsM, intrinsicsN,
```
egebeysel, line 568: Question: if we were to utilize pipelining in the future for the innermost computations (that the current CPU C ukernels encompass), the logic to choose the factors for that would come here and decide on `intrinsics_k`, right?

compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
```diff
@@ -223,19 +224,20 @@ func.func @unset_encoding_matmul_RESULT_inner_tiled_avx512(%arg0: tensor<127x255
 }
 // CHECK-LABEL: func @set_encoding_matmul_LHS_inner_tiled_avx512(
 //  CHECK-SAME:   %[[INPUT:[a-zA-Z0-9]+]]: tensor<127x255xf32>
-//       CHECK:   %[[EMPTY:.+]] = tensor.empty() : tensor<127x255x1x1xf32>
-//       CHECK:   %[[PACK:.+]] = linalg.pack %[[INPUT]] outer_dims_perm = [0, 1] inner_dims_pos = [0, 1] inner_tiles = [1, 1] into %[[EMPTY]] : tensor<127x255xf32> -> tensor<127x255x1x1xf32>
-//       CHECK:   return %[[PACK]] : tensor<127x255x1x1xf32>
+//   CHECK-DAG:   %[[CST:.+]] = arith.constant 0.0
+//       CHECK:   %[[EMPTY:.+]] = tensor.empty() : tensor<8x255x16x1xf32>
+//       CHECK:   %[[PACK:.+]] = linalg.pack %[[INPUT]] padding_value(%[[CST]] : f32) outer_dims_perm = [0, 1] inner_dims_pos = [0, 1] inner_tiles = [16, 1] into %[[EMPTY]] : tensor<127x255xf32> -> tensor<8x255x16x1xf32>
```
egebeysel, line 229: is it due to the unit dimensions that we do not get `8x255x16 (intrinsics_m) x 1x1` via `shape_cast + transposed` here? For non-unit intrinsic dimensions (e.g. Arm BFMMLA `2x4xbf16 @ 4x2xbf16 -> 2x2xf32`), the unrolling factors and the intrinsic dimensions would be represented explicitly/separately and we would (or should) get the `shape_cast + transpose`, right?

egebeysel (approved):
Looked super interesting so I decided to give it a quick review and leave some comments/questions :) Thanks!

compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
bjacob: Yes, exactly. The effect of `intrinsics_m = 16` is to add an extra dimension of size 16. So you do end up with shapes with many dimensions. This comes from the GPU side where having this many dimensional kind of shape then allows us to have arbitrary transposes, effectively realizing arbitrary data layouts.

Yes, when dimensions are unit-size, they can be omitted from the shapes of inner-tiled operands, resulting in rank discrepancies.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
bjacob: Good FYI, thank. I remember about that now. It's fine. The claim in my PR here is not that the model I'm implementing here is good enough to encompass existing cases. I'm only claiming that this PR lays some useful foundation to unblock making progress with the rest of the inner-tiled project. Good point, that we will want to revisit this later before it becomes useful on RISCV.

To be clear, there is no deep power-of-two restriction. This is just a detail of this piece of code, which can be changed at any time.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
bjacob: Pipelining is orthogonal to `intrinsics_k`. You can have pipelining already with `intrinsics_k = 1`: that consists in rewriting the naive kernel

```c++
for (k = 0; k < k_size; ++k) {
  lhs_vectors_t lhs = load_lhs(k);
  rhs_vectors_t rhs = load_rhs(k);
  acc = matmul(acc, lhs, rhs);
}
```

into

```c++
lhs_vectors_t lhs = load_lhs(0);
rhs_vectors_t rhs = load_rhs(0);
for (k = 1; k < k_size; ++k) {
  acc = matmul(acc, lhs, rhs);
  lhs_vectors_t lhs = load_lhs(k);
  rhs_vectors_t rhs = load_rhs(k);
}
acc = matmul(acc, lhs, rhs);
```

To answer your question:
* Regarding the possibility of `intrinsics_k > 1`:  if in the future we wanted that on CPU, then yes, the logic selecting that would go here as well. You can see that at play today in the GPU code, where `intrinsics_k > 1` is common.
* Regarding the possibility of loop pipelining:  that is a code-generation rewrite pattern that is orthogonal to the tile decisions being made here. It wouldn't have to enter the picture here, AFAICS. Also, it would likely be done in microkernels, because if software pipelining makes a difference at all, that means we are dealing with a CPU that is sensitive to micro details of instruction scheduling, typically a use case for microkernels. A modern out-of-order CPU is likely to perform the pipelining very well at runtime, making software pipelining often counter-productive (the larger kernel instruction count is only a liability wrt exceeding the reorder buffer size).

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
bjacob: ah yes, good idea.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
egebeysel: Oh right, I was confused and coupled the two together due to an old BLIS kernel I looked at. Thanks!

Do you see any reasons/cases where we might want to utilize `intrinsics_k > 1` on the CPU, similar to the GPU?

bjacob: No: the only reason why it's used on GPU is very specific to SIMT i.e. to how GPU ISAs differ from CPUs.

The reason as far as AMD CDNA3/CDNA4 goes is a discrepancy between the ideal size of contiguous bytes that you want to be loading (128 bits, matching the largest available load instructions) and the ideal size of contiguous bytes that your matrix multiplication instructions want to consume on each thread (typically 32 or 64 bits, e.g. for FP8 matmul you have instructions consuming 8x FP8 elements, so 64 bits). `intrinsics_k` is then set as the quotient between these sizes, e.g. 128/64 = 2, bridging that gap. That is very specific to GPUs having matmul instructions consuming relatively few contiguous bytes on each thread, i.e., SIMT.

bjacob: Note: that GPU use case is also why GPU DataTiledMMAAttr has the interleaving attribute.  We are not only unrolling on k, we are doing so with interleaved layout so that the 2 or 4 intrinsics consume contiguous data, so that one 128-bit load feeds data into all of these unrolled intrinsics. And that being specific to GPU is the reason why I haven't seen the need to have the interleaving attribute also on DataTiledMMAAttr so far.

egebeysel: Thanks, that's very helpful!

hanhanW (approved):

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
bjacob: FYI: I will merge this PR as-is to avoid retriggering all the CI and waiting an hour before I can merge, but I will adopt your suggestion in a subsequent PR.

PR #24305, opened 2026-04-29
[Codegen] Cap DMA per-lane width to min contiguous size of src/dest

Yu-Zhewen:
When `gather_to_lds` copies from a strided source (e.g. `memref<256x4xf8E8M0FNU, strided<[256, 1]>>`), the
inner contiguous dimension may be smaller than what 128-bit DMA reads per lane. This would cross row boundaries and pull bytes from unrelated rows.

Fix: compute the contiguous trailing linear size for both source and destination, and pass `maxElementsPerLane = min(srcLinearSize, destLinearSize)` into `computeTransferSegments`.

Assisted-by: Cursor (Claude)


sommerlukas (approved):
LGTM.

lialan (approved):
+1

PR #24313, opened 2026-04-30
[Torch][LinalgExt] Support GQA in torch.hop_flex_attention lowering

keshavvinayak01:
This PR adds grouped-query attention support to the torch.hop_flex_attention lowering in IREE’s Torch input pipeline.

When query, key, and value have different head counts, the lowering now expands key and value heads to match the query head count before emitting `iree_linalg_ext.online_attention`. 

This is similar to how the torch-mlir => TMTensor lowering handles this case.


keshavvinayak01 (commented):
@sommerlukas Could you review?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -183,6 +184,98 @@ static Value convertToBuiltinTensor(PatternRewriter &rewriter, Location loc,
...
+
+  SmallVector<int64_t> expandShape(selfType.getSizes());
+  expandShape.insert(expandShape.begin() + dim + 1, repeats);
+  SmallVector<int64_t> expandShapeForBroadcast(expandShape.size(), -1);
+  expandShapeForBroadcast[dim + 1] = repeats;
+  Value expandShapeList =
+      torch::Torch::toIntListConstruct(rewriter, loc, expandShapeForBroadcast);
+
+  Type expandType = selfType.getWithSizesAndDtype(
+      ArrayRef<int64_t>(expandShape), selfType.getOptionalDtype());
```
sommerlukas: Nit: Do we need the explicit `ArrayRef<int64_t>(expandShape)`?

sommerlukas (approved):
Implementation looks good to me (one minor nit), but I'd prefer if somebody with more experience in this part of the code would also take a look before merging.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -183,6 +184,98 @@ static Value convertToBuiltinTensor(PatternRewriter &rewriter, Location loc,
...
+
+  Value dimValue = torch::Torch::ConstantIntOp::create(rewriter, loc, dim);
+  Value dimValuePlusOne =
+      torch::Torch::ConstantIntOp::create(rewriter, loc, dim + 1);
+
+  FailureOr<Value> unsqueezed =
+      torch::Torch::unsqueezeTensor(rewriter, op, self, dimValuePlusOne);
+  if (failed(unsqueezed)) {
+    return failure();
+  }
```
IanWood1: This should only fail when `dimValuePlusOne` is out of range. We can probably ignore the failure state and just call `.value()`. Then, `repeatTensorElementsForDim()` doesn't need error handling.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -334,7 +427,12 @@ struct FlexAttentionOpConversion
           op, "expected return_max_scores to be a constant bool");
     }
 
-    // Extract shapes from Q, K, V.
+    if (failed(preProcessGroupQueryAttentionInputs(op, rewriter, query, key,
```
IanWood1: I was looking at the torch-mlir SDPA decomposition and noticed that it looked at `enable_gqa`. Is there a reason that it doesn't exist on the torch-mlir op? From what I can tell it exists in pytorch https://github.com/pytorch/pytorch/blob/8f4cbd0fbbdc60820247cba1745826ff6485d771/torch/nn/attention/flex_attention.py#L1874

IanWood1 (approved):
Looks good, I just have one small suggestion and question.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: That's a very good catch, I don't know how I missed it before, but there should definitely be a bool attribute for `enable_gqa`

keshavvinayak01: It's been a while since I added the [support change in torch-mlir](https://github.com/llvm/torch-mlir/pull/4366/changes) so I forgot. The problem with that implementation was that `enable_gqa` is a public wrapper-level flag; by the time PyTorch emits/calls flex_attention_hop, that flag is not part of the HOP argument list:

https://github.com/pytorch/pytorch/blob/6b63b49b1337dc80bc7c76d4c56a31737120b222/torch/_higher_order_ops/flex_attention.py#L98-L109

The simplest way to now fix this is to add `DefaultValuedAttr<BoolAttr, "false">:$enable_gqa` so we can use it here. I'll open that up

keshavvinayak01: @IanWood1 Please read and go through https://github.com/llvm/torch-mlir/pull/4562 first, if that makes sense, I'll also use the flag here as well to only run the function when `enable_gqa` is passed.

keshavvinayak01: This is what's missing, essentially:
https://github.com/llvm/torch-mlir/blob/5a9972229d252cde501fff124992c0603127a01a/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp#L1982-L1986

IanWood1: Does this mean it's impossible to set `enable_gqa = True` when exporting from torch? 


Also, maybe `enable_gqa` is just about shape validation. If we can't set it when exporting from torch, and it's really only useful for shape validation maybe it's not needed?

keshavvinayak01: You can set it up, but in that path of the logic in torch, the exporter will validate that the Q/K/V shapes are compatible before creating the operation. And then the actual `torch.hop_flex_attention` op will be created with the attribute.

keshavvinayak01: So technically, the only way of using the attribute is if you somehow modify the graph to add the optional attribute `{enable_gqa=True}` so at least that will allow this repeated heads logic to run.

keshavvinayak01: The only reason I need this logic to still be added in our lowering is so that fusilli (which aims to emit flex_attention) can atleast use this logic for working GQA. I think it's worth it.

PR #24315, opened 2026-04-30
[CI] Ease contention on rdna4 machine

amd-eochoalo:
Torch_ops cpu job had the possibility of using the only rdna4 machine. Looking at the number of jobs per machine, the gfx1201 machine (shark-75) on https://github.com/hanhanW/iree-ci-monitor#self-hosted-runners-last-7d has the most number of jobs assigned to it.

image

I think this may speed up some jobs that truly depend on gfx1201.

Alternatively, since we stopped benchmarking on CPU (due to noise) we can move this to a github runner (but we would lose caching of the azure files).


amd-eochoalo:
Just going to work on it and re-open it later.

PR #24316, opened 2026-04-30
[CI] Ease contention on self hosted machines

amd-eochoalo:
After https://github.com/iree-org/iree/pull/24194 we have no need to use a benchmarking machine for cpu jobs in torch ops. So let's just use the github action ones for torch_ops cpu and help reduce contention on self hosted machines.

### DIFF
--- a/.github/workflows/pkgci_test_torch.yml
+++ b/.github/workflows/pkgci_test_torch.yml
@@ -33,12 +33,8 @@ jobs:
           - name: cpu_task
             markers: "cpu"
             config-file: torch_ops_cpu_llvm_sync.json
-            cache-dir: /home/nod/iree_tests_cache
-            runs-on:
-              - self-hosted # Must come first.
-              - persistent-cache
-              - Linux
-              - X64
+            runs-on: ubuntu-24.04
+            cache-dir: /tmp/iree_tests_cache
           - name: amdgpu_hip_gfx1100_O3
             config-file:  torch_ops_gpu_hip_gfx1100_O3.json
             runs-on: [Linux, X64, gfx1100]


kuhar (commented):
How long do these take on github-operated machines vs. shark lab?
I'd also think we now have less contention, not more, after we dropped benchmarks?

amd-eochoalo:
> How long do these take on github-operated machines vs. shark lab?

On github machines they get picked up quite fast (as soon as the job gets created). On self-hosted runners, this latest run https://github.com/iree-org/iree/actions/runs/25181924016/job/73830081894 is still waiting after 17 minutes.

amd-eochoalo:
> I'd also think we now have less contention, not more, after we dropped benchmarks?

We do have less contention, but only three machines have the set of flags used in torch_ops cpu. The gfx1201 and the two gfx1100. And these are in high utilization. See here for the number of jobs on the self hosted runners: https://github.com/hanhanW/iree-ci-monitor#self-hosted-runners-last-7d

kuhar:
> > How long do these take on github-operated machines vs. shark lab?
> 
> On github machines they get picked up quite fast (as soon as the job gets created). On self-hosted runners, this latest run https://github.com/iree-org/iree/actions/runs/25181924016/job/73830081894 is still waiting after 17 minutes.

how about the running time? IE pick up time + time to complete job

amd-eochoalo:
> how about the running time? IE pick up time + time to complete job

pick up time is way more compared to time to complete the job. 

* ubuntu machine: 56s https://github.com/iree-org/iree/actions/runs/25182050890/job/73831109244?pr=24316
* self-hosted: 1m4s (same link as before) https://github.com/iree-org/iree/actions/runs/25181924016/job/73830081894

So for self-hosted we get 20+ minutes, and for ubuntu machine we get ~1minute

kuhar (approved):
Makes sense then. Could you use these numbers as the justification in the PR description?

PR #24318, opened 2026-04-30
[Codegen] Use divisibility analysis for im2col vectorization

Max191:
Adds the divisibility analysis to the MaterializeVectorTileSizesPass, and uses it for determining im2col vector sizes. The im2col op vector sizes depend on the offset computation because of load contiguity constraints, so we need to have divisibility information to accurately determine good vector sizes. This used to be done by matching affine ops, but this PR makes the divisibility analysis available when determining the divisibility, which enables better vectorization for more complex offset computation cases.

In the MaterializeVectorTileSizesPass, the divisibility analysis is run to a fixpoint before the vector size analyses are loaded and initialized, because the vector size analyses depend on the fully converged divisibility results.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
@@ -418,6 +418,11 @@ public:
     // so tile sizes go directly on the result lattice.
     if (auto im2colOp = dyn_cast<IREE::LinalgExt::Im2colOp>(op)) {
       OpBuilder builder(op);
+      // The dependent determines what reruns if the divisibility lattice for
+      // `val` changes below. Sparse forward analyses visit operations from the
+      // program point after the operation, so use `after(op)` to rerun this op.
+      // This dependency is not expected to affect convergence in practice:
+      // divisibility is solved to fixpoint before tile-size analysis starts.
       ProgramPoint *dependent = getProgramPointAfter(op);
       std::optional<SmallVector<int64_t>> maybeTileSizes =
           IREE::LinalgExt::computeIm2colVectorTileSizes(
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
@@ -10,7 +10,10 @@
 #include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/Im2colUtils.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
+#include "iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.h"
 
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Casting.h"
 #include "llvm/Support/DebugLog.h"
 #include "mlir/Analysis/DataFlow/SparseAnalysis.h"
 #include "mlir/Analysis/DataFlow/Utils.h"
@@ -18,6 +21,8 @@
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/IR/SymbolTable.h"
+
+#include <algorithm>
 
 namespace mlir::iree_compiler {
 #define GEN_PASS_DEF_MATERIALIZEVECTORTILESIZESPASS
@@ -338,6 +343,10 @@ class TileSizeForwardAnalysis
 public:
   using SparseForwardDataFlowAnalysis::SparseForwardDataFlowAnalysis;
 
+  static bool classof(const DataFlowAnalysis *a) {
+    return a->getTypeID() == TypeID::get<TileSizeForwardAnalysis>();
+  }
+
   void setToEntryState(TileSizeLattice *lattice) override {
     // Entry state is uninitialized (identity for join).
     propagateIfChanged(lattice, lattice->join(TileSizes()));
@@ -413,8 +422,20 @@ public:
     // so tile sizes go directly on the result lattice.
     if (auto im2colOp = dyn_cast<IREE::LinalgExt::Im2colOp>(op)) {
       OpBuilder builder(op);
+      ProgramPoint *dependent = getProgramPointAfter(op);
       std::optional<SmallVector<int64_t>> maybeTileSizes =
-          IREE::LinalgExt::computeIm2colVectorTileSizes(builder, im2colOp);
+          IREE::LinalgExt::computeIm2colVectorTileSizes(
+              builder, im2colOp, [this, dependent](Value val) {
+                const auto *lattice =
+                    getOrCreateFor<IREE::Util::IntegerDivisibilityLattice>(
+                        dependent, val);
+                if (!lattice || lattice->getValue().isUninitialized()) {
+                  return uint64_t{1};
+                }
+                const IREE::Util::ConstantIntDivisibility &div =
+                    lattice->getValue().getValue();
+                return div.sdiv();
+              });
       if (maybeTileSizes) {
         TileSizes tileSizes(*maybeTileSizes);
         propagateIfChanged(results[0], results[0]->join(tileSizes));
@@ -460,6 +481,10 @@ class TileSizeBackwardAnalysis
     : public dataflow::SparseBackwardDataFlowAnalysis<TileSizeLattice> {
 public:
   using SparseBackwardDataFlowAnalysis::SparseBackwardDataFlowAnalysis;
+
+  static bool classof(const DataFlowAnalysis *a) {
+    return a->getTypeID() == TypeID::get<TileSizeBackwardAnalysis>();
+  }
 
   void setToExitState(TileSizeLattice *lattice) override {
     // Exit state is uninitialized (identity for meet).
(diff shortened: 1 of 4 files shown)


Max191:
I'll run the convolution benchmark suite, but permission restructuring revoked my write access to the amd-shark-ai repo, so I am waiting on my permissions to be approved and then I'll run it.

hanhanW:
@sommerlukas can you review it? Thanks!

compiler/src/iree/compiler/Codegen/Common/test/materialize_vector_tile_sizes.mlir
```diff
@@ -583,6 +583,106 @@ func.func @im2col_tile_sizes_non_contiguous(
...
+          ins(%input : tensor<2x34x34x640xf32>)
+          outs(%0 : tensor<2x2x4xf32>) -> tensor<2x2x4xf32>
+  return %1 : tensor<2x2x4xf32>
+}
+
+// -----
+
+// Im2col: offset that `IntegerDivisibilityAnalysis` proves is exactly zero.
+// Zero is divisible by every tile size, and this case is not covered by the
+// fallback static `affine.apply` map match.
```
amd-eochoalo: I find the comment here a bit misleading. Yes, the fallback static `affine.apply` map does not match here, but we are not testing that case. To me this comment reads like the test is a negative test, but it is just a positive test. I would maybe just reword this?

compiler/src/iree/compiler/Codegen/Common/test/materialize_vector_tile_sizes.mlir
```diff
@@ -583,6 +583,106 @@ func.func @im2col_tile_sizes_non_contiguous(
...
+}
+
+// -----
+
+// Im2col: no divisibility information available (offset is an opaque function
+// argument). The analysis cannot prove contiguity, so no tile sizes attribute
+// is stamped. This is the negative control for the two preceding tests.
+
+// CHECK-LABEL: @im2col_tile_sizes_divisibility_unknown
+func.func @im2col_tile_sizes_divisibility_unknown(
```
amd-eochoalo: nit: use `@negative_im2col...` label.

amd-eochoalo (approved):
I think this is good. Just minor nits.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -10,7 +10,10 @@
 #include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtDialect.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/Im2colUtils.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
+#include "iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.h"
 
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Casting.h"
```
sommerlukas: Do we need this includes explicitly?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -19,6 +22,8 @@
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/IR/SymbolTable.h"
 
+#include <algorithm>
```
sommerlukas: Do we need this include?

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -338,6 +343,10 @@ class TileSizeForwardAnalysis
 public:
   using SparseForwardDataFlowAnalysis::SparseForwardDataFlowAnalysis;
 
+  static bool classof(const DataFlowAnalysis *a) {
+    return a->getTypeID() == TypeID::get<TileSizeForwardAnalysis>();
+  }
```
sommerlukas, line 344: What made this addition necessary? Is there some MLIR macro that we can use if necessary? 
(Same applies below for the backward analysis)

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -660,13 +685,26 @@ class MaterializeVectorTileSizesPass final
...
+    // The existing divisibility lattices are left initialized from stage 1 and
+    // can be queried by the vector size analyses when determining vector sizes.
     solver.load<TileSizeForwardAnalysis>();
     SymbolTableCollection symbolTable;
     solver.load<TileSizeBackwardAnalysis>(symbolTable);
-
-    if (failed(solver.initializeAndRun(funcOp))) {
+    if (failed(solver.initializeAndRun(
+            funcOp, llvm::IsaPred<TileSizeForwardAnalysis,
+                                  TileSizeBackwardAnalysis>))) {
```
sommerlukas, line 813: Not a comment for this PR, but this would be so much nicer if the dataflow framework supported modelling dependencies between analyses.

sommerlukas (approved):
Changes to `MaterializeVectorTileSizes.cpp` look good to me apart from the nits.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
Max191: There is a PR up that is working on adding that, but it is getting blocked by a reviewer last time I checked. Once it lands we can add the dependencies in here, which I agree would be nice.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
Max191: This is just to make the filtering function simpler (`IsaPred` uses `classof`). I can check if there is a macro for this and add it if I can find one.

sommerlukas: The macro I used in another analysis was `MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID`, not sure if it works here though.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -660,13 +681,26 @@ class MaterializeVectorTileSizesPass final
   void runOnOperation() override {
     FunctionOpInterface funcOp = getOperation();
 
+    // Stage 1: run baseline analyses and IntegerDivisibilityAnalysis to
+    // convergence. We need these results to be stable before the tile size
+    // analyses consult them, so we use the staged-solve pattern enabled by
+    // `DataFlowSolver::initializeAndRun`'s analysis filter.
     DataFlowSolver solver;
     dataflow::loadBaselineAnalyses(solver);
+    solver.load<IREE::Util::IntegerDivisibilityAnalysis>();
```
sommerlukas, line 800: How computationally expensive is the `IntegerDivisibilityAnalysis`? Would it make sense to skip it if there's no im2col in the IR?

If we do, we need to make sure that the baseline analyses run in the second stage.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -413,8 +418,20 @@ class TileSizeForwardAnalysis
     // so tile sizes go directly on the result lattice.
     if (auto im2colOp = dyn_cast<IREE::LinalgExt::Im2colOp>(op)) {
       OpBuilder builder(op);
+      ProgramPoint *dependent = getProgramPointAfter(op);
```
sommerlukas, line 426: It's a bit odd that we get analysis state after the operation for a forward dataflow analysis here. If it's necessary, might at least be worth adding an explanatory comment here.

sommerlukas (commented):
Some additional points/questions, but those don't need to block this PR, we can address them in a follow up if necessary.

> we need to have divisibility information to accurately determine good vector sizes

Just a note on the original intent of this analysis: The goal was not to determine "good" vector tile sizes. The idea of this analysis was that the lowering config selection determines tile sizes for the most important computational ops and this analysis then propagates the information so other operations around those ops are vectorized (masked) with the same tile sizes. 

So, in the original design, the decision what a "good" tile size would be was left to the lowering config selection stage (`KernelConfig`), this analysis was more mechanical. 

If tile sizes for im2col can't be determined in lowering config selection yet, I'm fine to add it here.

Max191:
Finally collected some benchmarks: https://gist.github.com/Max191/54c2bfb0d04b1009a29f5ceb6dd6b188

The improvers are probably mostly not real, since the improvement is small (if this PR had an effect on any shapes, then I'd expect a larger improvement), but there are no degraders.

Going to rebase and merge now.

EDIT: I just noticed Lukas' comments. I'll address those first.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
Max191: That macro adds a type ID for the class, but it doesn't implement classof. The reason we usually get `isa` automatically on most classes is because they get automatically defined by tablegen for things like operations/attributes/etc, but analyses aren't defined via tablegen, so we need to add this ourselves.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
Max191: It shouldn't make much difference. The base analyses (dead code analysis, etc) are already walking through all the IR, and computing the divisibility is usually very simple, so I wouldn't think it adds much.

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
Max191: I added a comment about it. The DataFlowAnalysis API requires you to give a dependent when looking up a state, which is intended because if you need that state to compute the new state, then there is a dependency there. The dependency in this context means "rerun the tile size analysis for the op before this program point if the state (divisibility state) we lookup (from val) ever changes". In practice, the divisibility state won't change, but we need to put the dependency anyway.

Max191:
> Some additional points/questions, but those don't need to block this PR, we can address them in a follow up if necessary.
> 
> > we need to have divisibility information to accurately determine good vector sizes
> 
> Just a note on the original intent of this analysis: The goal was not to determine "good" vector tile sizes. The idea of this analysis was that the lowering config selection determines tile sizes for the most important computational ops and this analysis then propagates the information so other operations around those ops are vectorized (masked) with the same tile sizes.
> 
> So, in the original design, the decision what a "good" tile size would be was left to the lowering config selection stage (`KernelConfig`), this analysis was more mechanical.
> 
> If tile sizes for im2col can't be determined in lowering config selection yet, I'm fine to add it here.

Thanks for the context. I think I recently kind of realized this myself, too, but there isn't a good way to set tile sizes for the im2col op during config selection.

As I am working on map_store -> transfer_scatter vectorization, I'm trying to come up with a better way to incorporate the analyses during the _transformation_ instead of the tile size selection. I plan to move this logic there once I come up with something.

For now, this should be okay, because im2col ends up being the only source of the vector size in practice anyway, so we won't expect to see any conflicts. I plan to land this to unblock some work while I'm separately working on a better way to do this, but I will replace it once it's done.

Issue #24323, opened 2026-04-30
[CPU] New data-tiling path through `inner_tiled`


bjacob:
This is a tracking issue for the project to implement a new CPU codegen path where data-tiled matmuls are materialized into `inner_tiled` ops instead of `mmt4d`. The `inner_tiled` op, which was originally designed to represent GPU matrix multiplication kernels, gets generalized to cover CPU use-cases, including scalable vectors. The limitations of layout generality inherent to `mmt4d` are overcome, as `inner_tiled` can represent arbitrary layouts, reusing the compiler framework with `TileSwizzle` data structures originally developed for GPU and now shared with CPU. The `inner_tiled` ops are then lowered to CPU intrinsics represented as `llvm.call_intrinsic` ops, sharing the existing GPU logic to lower `inner_tiled` ops to intrinsics. In the final phases of this project, GPU microkernel frameworks (MLIR and/or C) will be ported over to CPU, providing an alternative to the current CPU microkernel framework.

A longer write-up on this project, with more motivation, is here: https://hackmd.io/ti1_aQYbQK2rn0YUM1lFKQ

Tasklist (each line is one PR):

- [X] [Initial plumbing for inner_tiled on CPU with data-tiled MMA attribute.](https://github.com/iree-org/iree/pull/23494)
- [X] [Use TileSwizzle for inner_tiled layout on CPU](https://github.com/iree-org/iree/pull/23705)
- [X] [[LLVMCPU] Add --iree-llvmcpu-enable-inner-tiled target flag.](https://github.com/iree-org/iree/pull/24219)
- [X] [[Codegen] inner_tiled op: finish CPU support including scalable tiles.](https://github.com/iree-org/iree/pull/24218)
- [X] [[Codegen][CPU] Materialize data-tiled encodings to inner_tiled.](https://github.com/iree-org/iree/pull/24289)
- [X] [[Codegen][CPU] Pick inner-tiled unroll factors from a register budget.](https://github.com/iree-org/iree/pull/24303)
- [X] [[Codegen][CPU] Fix RHS indexing map in materialize-encoding inner_tiled lowering.](https://github.com/iree-org/iree/pull/24325)
- [X] [[Codegen] NFC: Lift DataTiledMMA inner_tiled lowering helpers into MMAUtils.](https://github.com/iree-org/iree/pull/24326)
- [X] [[Codegen][CPU] Teach the lowering strategy about inner_tiled.](https://github.com/iree-org/iree/pull/24328)
[...]

PR #24328, opened 2026-04-30
[Codegen][CPU] Teach the lowering strategy about inner_tiled.

bjacob:
Adds a `setRootConfig(InnerTiledOp)` overload in `KernelDispatch.cpp`
that:

  * routes `iree_codegen.inner_tiled` roots through the
    `Mmt4dTilingExpert` pipeline. The DataTiling pipeline is the
    pack/unpack-focused pipeline and not all the modern lowerings (e.g.
    tile-and-fuse, post-bufferize vector lowering) kick in there;
    Mmt4dTilingExpert has them, plus its `GenericVectorizationPass` step
    is where vector-semantics lifting of `inner_tiled` happens via the
    in-tree `VectorizableOpInterface` external model. The previous
    commit schedules `LLVMCPULowerInnerTiledPass` right after
    `GenericVectorizationPass` in this pipeline.
  * sets `distribution = (1, 1, 0)` (one inner tile per workgroup in
    the M and N parallel iter dims, K not distributed);
  * sets `vector_common_parallel = (1, 1, 0)` and
    `vector_reduction = (0, 0, 1)` so the K dim is tiled to a real
    `scf.for` instead of being fully unrolled inside the
    `LLVMCPULowerInnerTiledPass` greedy fixed point.

Without this commit, a `linalg.matmul` with
`--iree-opt-data-tiling --iree-llvmcpu-enable-inner-tiled` ended up on
the `Default` pipeline (which has no inner_tiled lowering hook) and
either failed bufferization (for non-trivial K) or compiled in time
exponential in K (after the previous "unroll inner_tiled" commit
turned the survival into a compile-time blow-up). With this commit,
a 256x256x256 f32 matmul on znver5 now compiles end-to-end in seconds:
the workgroup body is a tight scf.for(K_iter) wrapping a single
`llvm.call_intrinsic "llvm.fma.v16f32"` per intrinsics_m unroll.

Distribution tile sizing is deliberately conservative (1 inner tile
per parallel dim) — the mmt4d cost-model-driven sizing heuristic
should be ported to inner_tiled as a follow-up.

Progress towards #24323

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -10,6 +10,7 @@
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenEnums.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenInterfaces.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenOps.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenTypes.h"
 #include "iree/compiler/Codegen/Interfaces/PartitionableLoopsInterface.h"
 #include "iree/compiler/Codegen/LLVMCPU/TargetMLTransformInfo.h"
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_x86_64_lowering_strategy.mlir
```diff
@@ -1674,3 +1674,35 @@ func.func @gather(%source: tensor<16x8x32x128xf16>, %indices: tensor<4xi64>, %ou
...
+
+// -----
+
+// `iree_codegen.inner_tiled` rooted dispatches go through the DataTiling
+// pipeline. The op's iter domain is already in inner-tile units (M_iter,
+// N_iter, K_iter), so each iter dim is tiled to a single iteration:
+// `vector_common_parallel` collapses the M and N parallel iter dims to 1
+// and `vector_reduction` collapses the K reduction iter dim to 1, so the
+// LLVMCPULowerInnerTiled pass's greedy fixed point can fire on an iter-free
+// vector inner_tiled. Distribution is parallel-only (M, N).
```
hanhanW: Seems to be implementation comment, but not test comment?

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -3019,6 +3020,72 @@ static LogicalResult setRootConfig(mlir::FunctionOpInterface entryPointFn,
...
+
+  LoweringConfigGenerator generator(op);
+  generator.setDistributionTileSizes(distTileSizes);
+  generator.setVectorTileSizes(vecTileSizes);
+  IREE::CPU::LoweringConfigAttr loweringConfig =
+      generator.generateCPULoweringConfig();
+  LDBG() << "Set lowering_config for inner_tiled op: " << loweringConfig;
+  return setOpConfigAndEntryPointFnTranslation(
+      entryPointFn, op, loweringConfig,
+      getCPUTranslationInfo(op.getContext(), CPUPipeline::DataTiling));
```
hanhanW: It's better to go through mmt4d pipeline, which has modern lowerings. The DataTiling one was created for pack/unpack ops and some optimizations are not kicked in. I haven't collapse the pipeline just because there are ukernels request sometimes, and I don't have cycles to fix future issue requests. This is a very rare path in CPU codegen, because packing ops are supposed to be fused.

https://github.com/iree-org/iree/blob/7247601b1967516ef7fd5cc5ae55abd5ced2397a/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L334-L388

hanhanW (changes requested):
Have you tried mmt4d pipeline? I think that is the pipeline we should target for this op.

hanhanW:
The best path is using double tiling expert, but that pipeline is not very good at bc level ukernel support atm. So we can still start from mmt4d pipeline.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
bjacob: Thanks for catching this!!

bjacob:
> The best path is using double tiling expert, but that pipeline is not very good at bc level ukernel support atm. So we can still start from mmt4d pipeline.

Done. Thanks again for catching this. It shows how human expert reviewers are still as necessary when AI does the code authoring.

hanhanW (approved):

hanhanW:
> Done. Thanks again for catching this. It shows how human expert reviewers are still as necessary when AI does the code authoring.

Agreed -- AI tools are powerful, and human are still responsible to those code, IMO.

Issue #24311, opened 2026-04-30
[Codegen][CPU] Add AArch64 and RVV intrinsics to support `inner_tiled` materialization


egebeysel:
Following the footsteps of @bjacob , port AArch64 and RVV intrinsics to the `inner_tiled` mechanism.

****************

### Define the intrinsic and the data-tiled layouts

The current data-tiled layout enumeration mechanism for `mmt4d` from [CPUEncodingExternalModels](https://github.com/iree-org/iree/blob/7778c15918526ecbb4b9606e2cfaa8c177854cd5/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp#L1170) annotates the register-blocked ukernel tiles. The `inner_tiled` mechanism is based on the shape of _one_ intrinsic, not the full tile, and the unrolling/blocking factors are handled explicitly by `intrinsics_m/n/k`. The [transpose](https://github.com/iree-org/iree/blob/7778c15918526ecbb4b9606e2cfaa8c177854cd5/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp#L932) for narrow-M/N mechanisms are (to the best of my knowledge) replaced by explicitly transposed intrinsics for the M-N swapped variants.

`getRowMajorTilesMNKShape`, `getABCElementTypes` and `getRegisterSpaceBytes` are the relevant bits from [IREECPUAttrs.cpp](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp). 

Adding a swizzle pattern is also possible, if an operand's in-register element order is not row major.

- [ ] Define the intrinsic and add the corresponding data-tiled layouts.
- [ ] Plumb through the intrinsic selection in [CPUEncodingExternalModels](https://github.com/iree-org/iree/blob/7778c15918526ecbb4b9606e2cfaa8c177854cd5/compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp#L357) so that we can emit `inner_tiled` ops with these intrinsics.

#### Relevant PRs
- https://github.com/iree-org/iree/pull/24494
- https://github.com/iree-org/iree/pull/23494
- https://github.com/iree-org/iree/pull/24456
- https://github.com/iree-org/iree/pull/24289
- swizzle: https://github.com/iree-org/iree/pull/24534

[...]

PR #24329, opened 2026-04-30
[Codegen][CPU] Add a VectorizeInnerTiled pattern.

bjacob:
Lifts a tensor-semantics `iree_codegen.inner_tiled` op to vector semantics by reading each operand with `vector.transfer_read`, replacing the op with a vector-semantics `inner_tiled`, and writing each result back with `vector.transfer_write`. Statically-shaped operands only — the CPU codegen path drives the iteration domain to unit bounds well before this pattern runs.

This fills the gap between encoding-materialization (which produces a tensor-semantics inner_tiled) and the existing
`DropInnerTiledUnitDimsPattern` + `LowerInnerTiledPattern` (both vector-semantics-only) on the LLVM-CPU side. GPU folds the analogous read/write into its lane-distribution step (`DistributeInnerTiledToLanes`); CPU has no lane distribution so the vectorize step needs to be standalone.

Exposes the pattern as `populateVectorizeInnerTiledPatterns` from `Codegen/Common/Transforms.h`, plus a transform-dialect entry point `apply_patterns.iree.vectorize_inner_tiled` so the pattern can be exercised from a lit test.

Tested via a single-tile AVX-512 1x16x1 f32 example in TransformExtensions/test/vectorize_inner_tiled.mlir.

Progress towards #24323

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/Common/BUILD.bazel
@@ -179,6 +179,7 @@ iree_compiler_cc_library(
         "UserConfig.cpp",
         "VectorLayoutAnalysis.cpp",
         "VectorTransferLowering.cpp",
+        "VectorizeInnerTiled.cpp",
         "VectorizeMemrefCopy.cpp",
         "VerifyPipelineConstraints.cpp",
         "VerifyWorkgroupDistribution.cpp",
--- a/compiler/src/iree/compiler/Codegen/Common/CMakeLists.txt
+++ b/compiler/src/iree/compiler/Codegen/Common/CMakeLists.txt
@@ -172,6 +172,7 @@ iree_cc_library(
     "UserConfig.cpp"
     "VectorLayoutAnalysis.cpp"
     "VectorTransferLowering.cpp"
+    "VectorizeInnerTiled.cpp"
     "VectorizeMemrefCopy.cpp"
     "VerifyPipelineConstraints.cpp"
     "VerifyWorkgroupDistribution.cpp"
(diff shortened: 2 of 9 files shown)


bjacob:
The UBSan CI failure is https://github.com/iree-org/iree/issues/24331

hanhanW (changes requested):
I think it is already happening in this implementation: https://github.com/iree-org/iree/blob/7247601b1967516ef7fd5cc5ae55abd5ced2397a/compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp#L1177-L1284 Can you check?

@sommerlukas we are reaching to the state that your vector size analysis requires inner tiles soon. We now allow inner tiles being dynamic for future SVE/SME/RISC-V work.

hanhanW:
Corresponding PRs:
- https://github.com/iree-org/iree/commit/475599b324b9caa7ab5ca491def05bbc94e9c6b3
- https://github.com/iree-org/iree/commit/485e40122564006cffa18b0921024782da549f20

sommerlukas:
> @sommerlukas we are reaching to the state that your vector size analysis requires inner tiles soon. We now allow inner tiles being dynamic for future SVE/SME/RISC-V work.

Thanks for the heads-up, happy to chat about how this can be supported. Do we also need to track scalable flags then?

The pack/unpack operations probably already carry size information for the inner tiles. How do we determine the mapping for inner dimensions, as they're not part of the maps?

hanhanW:
> > @sommerlukas we are reaching to the state that your vector size analysis requires inner tiles soon. We now allow inner tiles being dynamic for future SVE/SME/RISC-V work.
> 
> Thanks for the heads-up, happy to chat about how this can be supported. Do we also need to track scalable flags then?
> 
> The pack/unpack operations probably already carry size information for the inner tiles. How do we determine the mapping for inner dimensions, as they're not part of the maps?

It will be more clear after https://github.com/iree-org/iree/issues/24311 is fixed.

The rough idea in my mind is using either ValueBound or extended IntRange analysis to get the tiling size, then you can propagate the information.

bjacob:
Thanks @hanhanW for catching this! Closed.

PR #24330, opened 2026-04-30
Bump stablehlo to stablehlo@806a6844dfd92cca

nirvedhmeshram:
Drops local patch since upstream stablehlo has caught up.


bangtianliu (approved):
Thanks

hanhanW (approved):

PR #24333, opened 2026-04-30
Reapply "[Codegen] Use local binders for optimization flags in codegen (#24220)"

hanhanW:
Previously, I made a workaround in https://github.com/iree-org/iree/commit/1b75890ac8093dd9b74b8110833bd2a46b3da7f0. The main reason is that all the codegen pipeline tests were anchor on a single pass (e.g., `XXXLowerExecutableTargetPass`), which makes plumbing through options infeasible.

After migrating to the real pipeline transformations, we are able to use the option in whole pipeline tests; we can drop the old workaround from iree-opt changes.

The revision introduces the option for each backend (default `O0`) and use local binders to apply the optimization flags. It drops the `XXX::FromFlags::get()` uses from Codegen.

The additional change is updating the default value in `Passes.td` to not apply optimization level. It was a workaround when we abused the pass for optimization control. Now all the pipeline tests use textual pass pipelines, so we no longer need the workaround.

ci-extra: linux_x64_clang_tsan

### DIFF
--- a/compiler/src/iree/compiler/API/Internal/BUILD.bazel
+++ b/compiler/src/iree/compiler/API/Internal/BUILD.bazel
@@ -83,7 +83,6 @@ iree_compiler_cc_library(
         "//compiler/bindings/c:headers",
         "//compiler/src/iree/compiler/Dialect/HAL/Target",
         "//compiler/src/iree/compiler/Dialect/VM/Target:init_targets",
-        "//compiler/src/iree/compiler/Pipelines:Options",
         "//compiler/src/iree/compiler/PluginAPI:PluginManager",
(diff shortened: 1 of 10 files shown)


hanhanW (commented):
@MaheshRavishankar https://github.com/iree-org/iree/pull/24333/changes/e716e95c98efecce0e596ecfd2a5490eeaafa61a is the optional change, which is a simple change.

hanhanW:
TSAN job is passing: https://github.com/iree-org/iree/actions/runs/25193803967/job/73869824249?pr=24333

MaheshRavishankar (approved):

PR #24034, opened 2026-04-07
[DispatchCreation] Skip collapse for scatter-like generics with tensor.extract

yzhang93:
Skip collapsing `linalg.generic` ops that contain both tensor.extract and linalg.index operations. These are strided scatter patterns (produced by [ConvertStridedInsertSliceToGeneric](https://github.com/iree-org/iree/pull/23990)) where 1D collapse introduces expensive delinearization (div/mod chains) on the extract indices, causing ~3.8x regression. Keeping the multi-dimensional iteration space allows direct workgroup tiling without delinearization.

ci-extra: test_torch

PR #24334, opened 2026-04-30
[DispatchCreation] Tighten scatter-skip predicate in CollapseDimensions

### LINKED ISSUE #24322: [Perf][Regression] PR #24034 causes tiling regression in RoPE + FP8 elementwise dispatch on gfx942
## Summary

PR #24034 ("Skip collapse for scatter-like generics with `tensor.extract`") has a codegen
side effect on RoPE + FP8 elementwise kernels in Llama-8B on gfx942. It prevents the
`4×batch×seq` dimensions from being collapsed into a flat `8192×8×128` dispatch shape,
producing a `4×2048×8×2×64` dispatch instead that tiles less efficiently.

This contributes ~+1.76ms per prefill on `llama_8b_fp8/prefill_benchmark_seq2048_mi325`
(32 layers × 2 dispatches per layer × ~+27µs each).

---

## Dispatch shape change

**Before PR #24034** — dispatch tiles as `8192x8x128` (batch×seq collapsed):
```
@rope_fp8_v_dispatch_0_elementwise_8192x8x128_f16xf16xf16xf32xf8E4M3FNUZ
```

**After PR #24034** — dispatch tiles as `4x2048x8x2x64` (uncollapsed):
```
@rope_fp8_v_dispatch_0_elementwise_4x2048x8x2x64_f16xf16xf16xf32xf8E4M3FNUZ
```

The difference: the `tensor.extract` inside the `linalg.generic` body (the RoPE paired-slot
lookup) is being treated as a scatter-like generic, which prevents the `4×2048` outer
dimensions from collapsing. The new tile shape is ~3.5× slower per invocation on gfx942
(~72µs vs ~20µs measured via `rocprofv3` on the full model).

---

## Repro

Minimal standalone MLIR that reproduces the shape change:

```mlir
// dispatch_23_repro.mlir
// The rotate_half + FP8 quantize kernel from Llama-8B prefill.
// compile with:
//   iree-compile dispatch_23_repro.mlir \
//       --iree-hal-target-device=hip --iree-rocm-target=gfx942 \
//       --iree-opt-level=O3 -o /tmp/d23.vmfb
//
// Before PR #24034: generates elementwise_8192x8x128
// After  PR #24034: generates elementwise_4x2048x8x2x64

module @dispatch_23_repro {
  func.func @rope_fp8_v(
      %v:     tensor<4x2048x8x128xf16>,
      %cos:   tensor<4x2048x1x128xf16>,
      %sin:   tensor<4x2048x1x128xf16>,
      %scale: tensor<f32>
  ) -> tensor<4x2048x8x128xf8E4M3FNUZ> {
    %cst_lo = arith.constant -2.400000e+02 : f32
    %cst_hi = arith.constant  2.400000e+02 : f32

    %v_cos = tensor.empty() : tensor<4x2048x8x128xf16>
    %v_cos_r = linalg.generic {
        indexing_maps = [
          affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>,
          affine_map<(d0,d1,d2,d3) -> (d0,d1,0,d3)>,
          affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>],
        iterator_types = ["parallel","parallel","parallel","parallel"]
    } ins(%v, %cos : tensor<4x2048x8x128xf16>, tensor<4x2048x1x128xf16>)
      outs(%v_cos : tensor<4x2048x8x128xf16>) {
    ^bb0(%a: f16, %b: f16, %o: f16):
      linalg.yield (arith.mulf %a, %b : f16) : f16
    } -> tensor<4x2048x8x128xf16>

    %v_first  = tensor.extract_slice %v[0,0,0,0][4,2048,8,64][1,1,1,1]
                  : tensor<4x2048x8x128xf16> to tensor<4x2048x8x64xf16>
    %v_second = tensor.extract_slice %v[0,0,0,64][4,2048,8,64][1,1,1,1]
                  : tensor<4x2048x8x128xf16> to tensor<4x2048x8x64xf16>

    %neg_buf = tensor.empty() : tensor<4x2048x8x64xf16>
    %neg_r = linalg.generic {
        indexing_maps = [affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>,
                         affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>],
        iterator_types = ["parallel","parallel","parallel","parallel"]
    } ins(%v_second : tensor<4x2048x8x64xf16>)
      outs(%neg_buf : tensor<4x2048x8x64xf16>) {
    ^bb0(%a: f16, %o: f16): linalg.yield (arith.negf %a : f16) : f16
    } -> tensor<4x2048x8x64xf16>

    %rot_buf = tensor.empty() : tensor<4x2048x8x128xf16>
    %rot0 = tensor.insert_slice %neg_r into %rot_buf[0,0,0,0][4,2048,8,64][1,1,1,1]
              : tensor<4x2048x8x64xf16> into tensor<4x2048x8x128xf16>
    %rot  = tensor.insert_slice %v_first into %rot0[0,0,0,64][4,2048,8,64][1,1,1,1]
              : tensor<4x2048x8x64xf16> into tensor<4x2048x8x128xf16>

    %rot_sin_buf = tensor.empty() : tensor<4x2048x8x128xf16>
    %rot_sin_r = linalg.generic {
        indexing_maps = [
          affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>,
          affine_map<(d0,d1,d2,d3) -> (d0,d1,0,d3)>,
          affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>],
        iterator_types = ["parallel","parallel","parallel","parallel"]
    } ins(%rot, %sin : tensor<4x2048x8x128xf16>, tensor<4x2048x1x128xf16>)
      outs(%rot_sin_buf : tensor<4x2048x8x128xf16>) {
    ^bb0(%a: f16, %b: f16, %o: f16):
      linalg.yield (arith.mulf %a, %b : f16) : f16
    } -> tensor<4x2048x8x128xf16>

    %rope_buf = tensor.empty() : tensor<4x2048x8x128xf16>
    %rope_r = linalg.generic {
        indexing_maps = [affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>,
                         affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>,
                         affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>],
        iterator_types = ["parallel","parallel","parallel","parallel"]
    } ins(%v_cos_r, %rot_sin_r : tensor<4x2048x8x128xf16>, tensor<4x2048x8x128xf16>)
      outs(%rope_buf : tensor<4x2048x8x128xf16>) {
    ^bb0(%a: f16, %b: f16, %o: f16):
      linalg.yield (arith.addf %a, %b : f16) : f16
    } -> tensor<4x2048x8x128xf16>

    %out = tensor.empty() : tensor<4x2048x8x128xf8E4M3FNUZ>
    %result = linalg.generic {
        indexing_maps = [affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>,
                         affine_map<(d0,d1,d2,d3) -> ()>,
                         affine_map<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>],
        iterator_types = ["parallel","parallel","parallel","parallel"]
    } ins(%rope_r, %scale : tensor<4x2048x8x128xf16>, tensor<f32>)
      outs(%out : tensor<4x2048x8x128xf8E4M3FNUZ>) {
    ^bb0(%a: f16, %sc: f32, %o: f8E4M3FNUZ):
      %af  = arith.extf %a : f16 to f32
      %div = arith.divf %af, %sc : f32
      %clo = arith.cmpf ult, %div, %cst_lo : f32
      %lo  = arith.select %clo, %cst_lo, %div : f32
      %chi = arith.cmpf ugt, %lo, %cst_hi : f32
      %hi  = arith.select %chi, %cst_hi, %lo : f32
      linalg.yield (arith.truncf %hi : f32 to f8E4M3FNUZ) : f8E4M3FNUZ
    } -> tensor<4x2048x8x128xf8E4M3FNUZ>

    return %result : tensor<4x2048x8x128xf8E4M3FNUZ>
  }
}
```

---

## Dispatch MLIR before PR #24034

The `linalg.generic` operates on `8192x8x128` (collapsed) with a 3-dim indexing map:

```mlir
%12 = linalg.generic {
    indexing_maps = [
      affine_map<(d0, d1, d2) -> (d0, d1, d2)>,
      affine_map<(d0, d1, d2) -> (d0, d2)>,
      affine_map<(d0, d1, d2) -> (d0, d2)>,
      affine_map<(d0, d1, d2) -> ()>,
      affine_map<(d0, d1, d2) -> (d0, d1, d2)>
    ],
    iterator_types = ["parallel", "parallel", "parallel"]
} ins(%7, %8, %9, %10 : tensor<8192x8x128xf16>,
                        tensor<8192x128xf16>,
                        tensor<8192x128xf16>,
                        tensor<f32>)
  outs(%11 : tensor<8192x8x128xf8E4M3FNUZ>)
```

## Dispatch MLIR after PR #24034

The `linalg.generic` operates on `4x2048x8x2x64` (uncollapsed, 5-dim):

```mlir
%12 = linalg.generic {
    indexing_maps = [
      affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>,
      affine_map<(d0, d1, d2, d3, d4) -> (d1, d3, d4)>,
      affine_map<(d0, d1, d2, d3, d4) -> (d1, d3, d4)>,
      affine_map<(d0, d1, d2, d3, d4) -> ()>,
      affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>
    ],
    iterator_types = ["parallel", "parallel", "parallel", "parallel", "parallel"]
} ins(%7, %8, %9, %10 : tensor<4x2048x8x2x64xf16>,
                        tensor<2048x2x64xf16>,
                        tensor<2048x2x64xf16>,
                        tensor<f32>)
  outs(%11 : tensor<4x2048x8x2x64xf8E4M3FNUZ>)
```

---

### Full model — `rocprofv3 --kernel-trace` (single run, cumulative over all 31 layer invocations)

Profiled on gfx942 (MI325) using `iree-run-module` with `rocprofv3 --kernel-trace --output-format csv`.
Times are cumulative GPU kernel duration across all 31 invocations per prefill pass.

```
rocprofv3 --kernel-trace --output-format csv --output-directory <outdir> -- \
  iree-run-module --module=llama.vmfb --parameters=model=<weights>.irpa \
    --function=prefill_bs4 --device=hip \
    --input=4x2048xi64 --input=4xi64 --input=4x64xi64 --input=34x2097152xf8E4M3FNUZ
```

| Kernel name | Cumulative (31 invokes) | Per invoke |
|-------------|------------------------|------------|
| `elementwise_…_8192x8x128` (before) | 607µs | ~20µs |
| `elementwise_…_4x2048x8x2x64` (after) | 2199µs | ~71µs |
| **Δ** | **+1592µs** | **+51µs** |

Note: `rocprofv3` kernel timings from a single run are not as stable as repeated benchmarks.
The per-invocation numbers are approximate; the key observation is the 3.5× slowdown.

### Standalone repro — `iree-benchmark-module` (50 reps, 2s warmup)

The standalone repro shows a smaller difference because the `insert_slice` ops from
the rotate_half are compiled as separate dispatches rather than fused, so the kernel
being benchmarked does less work than the full-model dispatch.

```
iree-benchmark-module --module=/tmp/d23.vmfb --device=hip \
    --function=rope_fp8_v \
    --input=4x2048x8x128xf16 --input=4x2048x1x128xf16 \
    --input=4x2048x1x128xf16 --input="f32=1.0" \
    --benchmark_repetitions=50 --benchmark_min_warmup_time=2.0
```

| Build | Dispatch shape | Mean |
|-------|---------------|------|
| Before PR #24034 | `8192x8x128` | 0.102ms |
| After PR #24034 | `4x2048x8x2x64` | 0.106ms |

The full-model `rocprofv3` numbers are the more representative measurement.

---

## Root cause

The `linalg.generic` in the RoPE kernel contains a `tensor.extract` that reads a paired
index with negation — this is the rotate_half operation. PR #24034 classifies this as a
"scatter-like generic" and skips the collapse that would have merged the `4×2048` batch
and sequence dimensions into a single flat dimension. The resulting 5D dispatch shape tiles
less efficiently for this elementwise workload on gfx942.

**Hardware:** MI325, gfx942  
**Identified in:** bisection of issue #24258

## Dispatch MLIR after PR #24034

The `linalg.generic` operates on `4x2048x8x2x64` (uncollapsed, 5-dim):

```mlir
%12 = linalg.generic {
    indexing_maps = [
      affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>,
      affine_map<(d0, d1, d2, d3, d4) -> (d1, d3, d4)>,
      affine_map<(d0, d1, d2, d3, d4) -> (d1, d3, d4)>,
      affine_map<(d0, d1, d2, d3, d4) -> ()>,
      affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>
    ],
    iterator_types = ["parallel", "parallel", "parallel", "parallel", "parallel"]
} ins(%7, %8, %9, %10 : tensor<4x2048x8x2x64xf16>,
                        tensor<2048x2x64xf16>,
                        tensor<2048x2x64xf16>,
                        tensor<f32>)
  outs(%11 : tensor<4x2048x8x2x64xf8E4M3FNUZ>)
```

---

## Performance impact

Measured on gfx942 (MI325) via `rocprofv3 --kernel-trace` on
`llama_8b_fp8/prefill_bs4` (seq=2048, 32 transformer layers):

| Dispatch | Before (#94170e7295) | After (#53dc6609ad) | Δ |
|----------|----------------------|----------------------|---|
| `elementwise_…_8192x8x128` (×31) | ~20µs/invoke | — | |
| `elementwise_…_4x2048x8x2x64` (×31) | — | ~72µs/invoke | **+52µs** |
| **Total prefill impact** | | | **+1.6ms** |

Standalone microbenchmark on the repro above:
- Before: **0.102ms** (`elementwise_8192x8x128`)
- After:  **0.106ms** (`elementwise_4x2048x8x2x64`)

---

## Root cause

The `linalg.generic` in the RoPE kernel contains a `tensor.extract` that reads a paired
index with negation — this is the rotate_half operation. PR #24034 classifies this as a
"scatter-like generic" and skips the collapse that would have merged the `4×2048` batch
and sequence dimensions into a single flat dimension. The resulting 5D dispatch shape tiles
less efficiently for this elementwise workload on gfx942.

**Hardware:** MI325, gfx942  
**Identified in:** bisection of issue #24258

yzhang93:
PR #24034 skipped collapse for any `linalg.generic` with a `tensor.extract` and a `linalg.index` in its body, in order to avoid an expensive delinearization on strided-scatter generics produced by `ConvertStridedInsertSliceToGeneric`. That predicate was too broad and caught the RoPE + FP8 dispatch in Llama-8B, whose fused body also has `tensor.extract` + `linalg.index` (rotate_half lookup) but is not a strided scatter; the collateral block prevents the `4×batch×seq` outer collapse and produces a 5-D dispatch (`4x2048x8x2x64`) that tiles ~3.5x slower on gfx942 than the pre-#24034 `8192x8x128` shape.

Tighten the predicate to require all three signals of the strided-scatter shape: empty `ins` operands (source captured via `tensor.extract`), at least one `linalg.index`, and a `tensor.extract` whose result is consumed by an `arith.select` (the bounds check).

Fixes: https://github.com/iree-org/iree/issues/24322

### DIFF
--- a/compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
+++ b/compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
@@ -221,23 +221,36 @@ static bool isEligibleForCollapse(Operation *op) {
     return false;
   }
 
-  // Skip collapse for scatter-like generics that use tensor.extract with
-  // linalg.index-based addressing. These are strided scatter patterns where
-  // 1D collapse introduces expensive delinearization (div/mod chains) that
-  // dominates execution. Keeping the multi-dimensional iteration space
-  // allows direct workgroup tiling without delinearization.
-  {
-    bool hasTensorExtract = false;
+  // Skip collapse for strided-scatter-like generics that use tensor.extract
+  // with linalg.index-based addressing. These generics have a very specific
+  // shape that makes 1D collapse pessimal:
+  //   1. No DPS input operands — the source is captured by a `tensor.extract`
+  //      inside the body rather than passed as an `ins` operand. The op is a
+  //      pure scatter into a fresh destination.
+  //   2. `linalg.index` ops feed arithmetic that computes a multi-dimensional
+  //      strided index into the captured source tensor.
+  //   3. An `arith.select` (driven by bounds-check `arith.cmpi`/`arith.andi`)
+  //      picks between the extracted value and a constant zero default.
+  // For these ops, 1D collapse introduces expensive delinearization (div/mod
+  // chains) on the extract indices that dominates execution; keeping the
+  // multi-dimensional iteration space allows direct workgroup tiling without
+  // delinearization.
+  if (genericOp.getNumDpsInputs() == 0) {
     bool hasLinalgIndex = false;
+    bool extractFeedsSelect = false;
     genericOp.getBlock()->walk([&](Operation *inner) {
-      if (isa<tensor::ExtractOp>(inner)) {
-        hasTensorExtract = true;
-      }
       if (isa<linalg::IndexOp>(inner)) {
         hasLinalgIndex = true;
+      } else if (auto extractOp = dyn_cast<tensor::ExtractOp>(inner)) {
+        for (Operation *user : extractOp.getResult().getUsers()) {
+          if (isa<arith::SelectOp>(user)) {
+            extractFeedsSelect = true;
+            break;
+          }
+        }
       }
     });
-    if (hasTensorExtract && hasLinalgIndex) {
+    if (hasLinalgIndex && extractFeedsSelect) {
       return false;
     }
   }
--- a/compiler/src/iree/compiler/DispatchCreation/test/collapse_dimensions.mlir
+++ b/compiler/src/iree/compiler/DispatchCreation/test/collapse_dimensions.mlir
@@ -1025,3 +1025,88 @@ util.func public @no_collapse_scatter_generic(%src: tensor<1x25x25x32xf16>) -> t
 // CHECK-SAME:    iterator_types = ["parallel", "parallel", "parallel", "parallel"]
 // CHECK:         tensor.extract
 // CHECK:         arith.select
+
+// -----
+
+// Rotate-half RoPE pattern: same body pattern as above, but `ins` operands
+// are present so the scatter-skip predicate does not fire. The pass
+// collapses `(d0,d1)` and `(d3,d4)`, dropping the iteration space from 5
+// to 3 dims.
+#map_iter   = affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>
+#map_brdcst = affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d3, d4)>
+util.func public @collapse_rope_rotate_half(
+    %v:   tensor<4x2048x8x2x64xf16>,
+    %cos: tensor<4x2048x2x64xf16>,
+    %sin: tensor<4x2048x2x64xf16>) -> tensor<4x2048x8x2x64xf16> {
+  %c1 = arith.constant 1 : index
+  %0 = flow.dispatch.region -> (tensor<4x2048x8x2x64xf16>) {
+    %empty = tensor.empty() : tensor<4x2048x8x2x64xf16>
+    %result = linalg.generic {
+      indexing_maps = [#map_iter, #map_brdcst, #map_brdcst, #map_iter],
+      iterator_types = ["parallel", "parallel", "parallel", "parallel", "parallel"]
+    } ins(%v, %cos, %sin : tensor<4x2048x8x2x64xf16>,
+                            tensor<4x2048x2x64xf16>,
+                            tensor<4x2048x2x64xf16>)
+      outs(%empty : tensor<4x2048x8x2x64xf16>) {
+    ^bb0(%a: f16, %c: f16, %s: f16, %o: f16):
+      %1 = linalg.index 0 : index
+      %2 = linalg.index 1 : index
+      %3 = linalg.index 2 : index
+      %4 = linalg.index 3 : index
+      %5 = linalg.index 4 : index
+      %6 = arith.subi %c1, %4 : index
+      %extracted = tensor.extract %v[%1, %2, %3, %6, %5] : tensor<4x2048x8x2x64xf16>
+      %7 = arith.negf %extracted : f16
+      %8 = arith.cmpi eq, %6, %c1 : index
+      %9 = arith.select %8, %7, %extracted : f16
+      %10 = arith.mulf %9, %s : f16
+      %11 = arith.mulf %a, %c : f16
+      %12 = arith.addf %11, %10 : f16
+      linalg.yield %12 : f16
+    } -> tensor<4x2048x8x2x64xf16>
+    flow.return %result : tensor<4x2048x8x2x64xf16>
+  }
+  util.return %0 : tensor<4x2048x8x2x64xf16>
+}
+
+// CHECK-LABEL: @collapse_rope_rotate_half
+// CHECK:         tensor.collapse_shape {{.*}} into tensor<8192x8x128xf16>
+// CHECK:         tensor.collapse_shape {{.*}} into tensor<8192x128xf16>
+// CHECK:         flow.dispatch.region
+// CHECK:         linalg.generic
+// CHECK-SAME:      iterator_types = ["parallel", "parallel", "parallel"]
+// CHECK:         tensor.extract
+
+// -----
+
+// Gather-like generic with `tensor.extract` + `linalg.index` but WITHOUT an
+// `arith.select`. The scatter-skip predicate must not match this shape (no
+// bounds check), so the generic is collapsed to a single parallel dim.
+#map = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+util.func public @collapse_reverse_generic(%src: tensor<2x2x2xf32>) -> tensor<2x2x2xf32> {
+  %c1 = arith.constant 1 : index
+  %0 = flow.dispatch.region -> (tensor<2x2x2xf32>) {
+    %empty = tensor.empty() : tensor<2x2x2xf32>
+    %result = linalg.generic {
+      indexing_maps = [#map],
+      iterator_types = ["parallel", "parallel", "parallel"]
+    } outs(%empty : tensor<2x2x2xf32>) {
+    ^bb0(%out: f32):
+      %i0 = linalg.index 0 : index
+      %i1 = linalg.index 1 : index
+      %i2 = linalg.index 2 : index
+      %rev = arith.subi %c1, %i1 : index
+      %extracted = tensor.extract %src[%i0, %rev, %i2] : tensor<2x2x2xf32>
+      linalg.yield %extracted : f32
+    } -> tensor<2x2x2xf32>
+    flow.return %result : tensor<2x2x2xf32>
+  }
+  util.return %0 : tensor<2x2x2xf32>
+}
+
+// CHECK-LABEL: @collapse_reverse_generic
+// CHECK:         flow.dispatch.region -> (tensor<8xf32>)
+// CHECK:         linalg.generic
+// CHECK-SAME:      iterator_types = ["parallel"]
+// CHECK:         tensor.extract
+// CHECK:         tensor.expand_shape
--- a/compiler/src/iree/compiler/DispatchCreation/test/collapse_dimensions.mlir
+++ b/compiler/src/iree/compiler/DispatchCreation/test/collapse_dimensions.mlir
@@ -993,30 +993,33 @@ util.func public @no_collapse_scatter_generic(%src: tensor<1x25x25x32xf16>) -> t
   %c1 = arith.constant 1 : index
   %c2 = arith.constant 2 : index
   %c25 = arith.constant 25 : index
-  %empty = tensor.empty() : tensor<1x52x52x32xf16>
-  %result = linalg.generic {
-    indexing_maps = [#map],
-    iterator_types = ["parallel", "parallel", "parallel", "parallel"]
-  } outs(%empty : tensor<1x52x52x32xf16>) {
-  ^bb0(%out: f16):
-    %n = linalg.index 0 : index
-    %h = linalg.index 1 : index
-    %w = linalg.index 2 : index
-    %c_idx = linalg.index 3 : index
-    %sh = arith.subi %h, %c1 : index
-    %rem_h = arith.remsi %sh, %c2 : index
-    %div_h = arith.divsi %sh, %c2 : index
-    %ge = arith.cmpi sge, %sh, %c0 : index
-    %eq = arith.cmpi eq, %rem_h, %c0 : index
-    %lt = arith.cmpi slt, %div_h, %c25 : index
-    %valid = arith.andi %ge, %eq : i1
-    %valid2 = arith.andi %valid, %lt : i1
-    %clamped = arith.maxsi %div_h, %c0 : index
-    %extracted = tensor.extract %src[%n, %clamped, %w, %c_idx] : tensor<1x25x25x32xf16>
-    %val = arith.select %valid2, %extracted, %zero : f16
-    linalg.yield %val : f16
-  } -> tensor<1x52x52x32xf16>
-  util.return %result : tensor<1x52x52x32xf16>
+  %0 = flow.dispatch.region -> (tensor<1x52x52x32xf16>) {
+    %empty = tensor.empty() : tensor<1x52x52x32xf16>
+    %result = linalg.generic {
+      indexing_maps = [#map],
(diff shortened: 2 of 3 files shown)


nirvedhmeshram (approved):
LGTM!

compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
```diff
@@ -221,23 +221,36 @@ static bool isEligibleForCollapse(Operation *op) {
...
+  // Skip collapse for strided-scatter-like generics that use tensor.extract
+  // with linalg.index-based addressing. These generics have a very specific
+  // shape that makes 1D collapse pessimal:
+  //   1. No DPS input operands — the source is captured by a `tensor.extract`
+  //      inside the body rather than passed as an `ins` operand. The op is a
+  //      pure scatter into a fresh destination.
+  //   2. `linalg.index` ops feed arithmetic that computes a multi-dimensional
+  //      strided index into the captured source tensor.
+  //   3. An `arith.select` (driven by bounds-check `arith.cmpi`/`arith.andi`)
+  //      picks between the extracted value and a constant zero default.
```
hanhanW, line 233: It's fine to bail out "weird tensor.extract accessing" to me, but I think we need @IanWood1's eyes on this PR.

yzhang93:
@hanhanW Looks like the change caused correctness regression on RISC-V https://github.com/iree-org/iree/actions/runs/25195139781/job/73896172955?pr=24334. This seems to be a target-specific issue triggered by the collapse change interacting with RISC-V vector codegen. If I drop the `extract + select` requirements, it doesn't have such issue. This requirement was added to exclude the gather ops (discussions in https://github.com/iree-org/iree/pull/24034#issuecomment-4355294173).

hanhanW:
> @hanhanW Looks like the change caused correctness regression on RISC-V https://github.com/iree-org/iree/actions/runs/25195139781/job/73896172955?pr=24334. This seems to be a target-specific issue triggered by the collapse change interacting with RISC-V vector codegen. If I drop the `extract + select` requirements, it doesn't have such issue. This requirement was added to exclude the gather ops (discussions in [#24034 (comment)](https://github.com/iree-org/iree/pull/24034#issuecomment-4355294173)).

RISC-V is one of a few targets that I did test with reverting your commit, before I landed https://github.com/iree-org/iree/commit/967b794b6474ca56ca33e7f1cc246659f9f8bfbc

Can you comment out this test, and file an issue? I'll reland the test or only disable it on riscv in a follow-up.

https://github.com/iree-org/iree/blob/d13374fbbe3687331139f498640ff5932de1e212/tests/e2e/stablehlo_ops/reverse.mlir#L24-L44

compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
```diff
@@ -221,23 +221,36 @@ static bool isEligibleForCollapse(Operation *op) {
...
-      }
       if (isa<linalg::IndexOp>(inner)) {
         hasLinalgIndex = true;
+      } else if (auto extractOp = dyn_cast<tensor::ExtractOp>(inner)) {
+        for (Operation *user : extractOp.getResult().getUsers()) {
+          if (isa<arith::SelectOp>(user)) {
+            extractFeedsSelect = true;
+            break;
+          }
+        }
```
IanWood1, line 250: It looks like this check happens work for the test cases in CI but doesn't actually capture the reason that the collapse is expensive. For example, if `collapse_reverse_generic` was modified to contain a `arith.select` wouldn't we still want to collapse it?


I'm fine with this if it works. But it would be nice to pin down exactly when we should be collapsing.

compiler/src/iree/compiler/DispatchCreation/test/collapse_dimensions.mlir
IanWood1, line 1023: Whoops, I missed this before earlier. This isn't testing anything because the pass only looks at ops in a `flow.dispatch.region`

tests/e2e/stablehlo_ops/reverse.mlir
```diff
@@ -22,23 +22,29 @@ func.func @xla_reverse() {
...
-          [[13.0, 14.0], [11.0, 12.0]]
-      ]> : tensor<2x2x2xf32>
-  ) : tensor<2x2x2xf32>
-  return
-}
+//
+// TODO: re-enable once the RISC-V vector miscompile is fixed. When the fused
+// slice + reverse `linalg.generic` is collapsed to a 1-D iteration space,
+// RISC-V vector codegen produces wrong values; tracked in
+// https://github.com/iree-org/iree/issues/24342.
```
hanhanW:
```suggestion
// TODO(24342): re-enable once the RISC-V vector miscompile is fixed. When the fused
// slice + reverse `linalg.generic` is collapsed to a 1-D iteration space,
// RISC-V vector codegen produces wrong values.
```

compiler/src/iree/compiler/DispatchCreation/CollapseDimensions.cpp
hanhanW: The root cause is https://github.com/iree-org/iree/pull/23990

What is the issue of forming them into a single dispatch and kick in the pattern in codegen? I don't follow fusion logic for a while, so it is more like a learning question from me.

The issue looks like we are "abuse" linalg.generic op  to represent something which allows fusion, but some information (e.g., strided insert slice) is dropped in this form <- it makes collapsing pass hard.

hanhanW: Maybe a more suitable question is: do we really want to enable fusion for these cases or it is just for a single workload?

yzhang93: Agreed, the arith.select check is a proxy rather than a true criterion. What's actually bad about collapsing those scatter generics is that their body already has a div/rem chain (computing the strided source coord), and collapse adds another delinearization on top of it — that chained div/rem is what dominates execution. The select happens to come along for the ride because the same lowering emits both, so I keyed off it as a quick proxy.

A more strict fix is to walk the extract's index operands and only skip when they actually contain a divsi/remsi on a linalg.index. Happy to do that if you think that's better — I traded it for the simpler check because the select happens to match every problematic pattern we have today, but I agree it's not the principle.

compiler/src/iree/compiler/DispatchCreation/test/collapse_dimensions.mlir
yzhang93: Thanks for catching this.

IanWood1 (approved):

PR #24335, opened 2026-05-01
[AMDGPU][DMA] Enable Gfx950 coalesced DMA on K-unaligned f16/bf16 GEMMs

lialan:
Enable `iree_gpu.coalesced_gather_dma` on f16/bf16 GEMMs whose innermost source row is not DWORD-aligned (e.g. K = 43, 129, 131) on gfx950, and gate it correctly so K-aligned shapes (1024, 4000) keep using the wide non-DMA loads they already win on.

It addresses two AMD CDNA partial-DWORD failure modes that previously forced us to bail in `isValidPadForDMA`:
* Per-lane straddle DWORDs reading from the next source row.
* The trailing partial DWORD crossing the buffer end and HW zeroing it.

## New components:
* New pass `GPUPushDownDMABoundsToConsumers`: for each DMA whose innermost `in_bounds=false`, walk up the `shared_outs` / `parallel_insert_slice` chain to the outermost forall result and insert an `extract_slice` + `tensor.pad` on the LDS tile. Downstream vectorization lowers the pad to a masked `vector.transfer_read`, zeroing the cross-row garbage each lane DMAs into the trailing innermost columns.
*  `BufferResourceCastOpBufferizationInterface::bufferize` method gains a new helper `computeDwordPadValidBytesIfNeeded` that runs on every `iree_gpu.buffer_resource_cast` it lowers. The new pass wraps the DMA source's root tensor with a plain cast and lets bufferization size the descriptor: when the bufferized memref's innermost-dim byte count is dynamic or statically not a multiple of 4, `BufferizationInterfaces` emits an `amdgpu.fat_raw_buffer_cast` with `validBytes = roundUp(naturalBytes + 3, 4)` (the +3 covers the last lane's straddle DWORD), peeling any pre-existing fat-raw-buffer wrap  on the binding so the override takes effect. For statically DWORD-aligned innermost rows no override is emitted, preserving prior behavior for cache-swizzle and other plain resource casts. `GPUBubbleResourceCasts` no longer needs a special skip — bubbling past view-likes is safe because bufferization sizes `validBytes` from whichever memref the cast ends up wrapping.

## Performance change
Square f16 matmul on gfx950:
```
| N    | K%32 | main (ms) | branch + fix (ms) | vs main      |
| ---- | ---- | --------- | ----------------- | ------------ |
| 1024 | 0    | 0.062     | 0.063             | -2% (noise)  |
| 1234 | 18   | 0.118     | 0.064             | +84%         |
| 2345 | 9    | 0.223     | 0.123             | +82%         |
| 4000 | 0    | 0.212     | 0.209             | +1% (noise)         |
| 6789 | 5    | 2.125     | 1.327             | +60%         |
| 9758 | 30   | 6.102     | 4.280             | +43%         |
```

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -329,14 +329,9 @@ static bool isValidPadForDMA(tensor::PadOp pad) {
     return false;
   }
   if (r0 != r1) {
-    unsigned kDim = (r0 < r1) ? 0 : 1;
-    return sourceType.isDynamicDim(kDim);
+    return sourceType.isDynamicDim((r0 < r1) ? 0 : 1);
   }
-  // Square: infer K from which source dim is dynamic. Both static or both
-  // dynamic → can't disambiguate, bail.
-  bool dyn0 = sourceType.isDynamicDim(0);
-  bool dyn1 = sourceType.isDynamicDim(1);
-  return dyn0 != dyn1;
+  return sourceType.isDynamicDim(0) != sourceType.isDynamicDim(1);
 }
 
 /// Check if a linalg.copy is viable for DMA conversion based on alignment,
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
@@ -293,19 +293,40 @@ static bool isValidPadForDMA(tensor::PadOp pad) {
     return false;
   }
 
-  // Check if source tensor's innermost row size is DWORD (4-byte) aligned. On
-  // AMD CDNA, per-component range checking is performed for each DWORD. If a
-  // DWORD is partially out-of-bounds, the entire DWORD returns zero, causing
-  // incorrect results. Additionally, partial OOB triggers the slow path with
-  // multi-cycling and instruction issue penalties.
-  auto sourceType = cast<RankedTensorType>(pad.getSource().getType());
-  int64_t innermostDim = sourceType.getShape().back();
-  if (ShapedType::isDynamic(innermostDim)) {
+  // AMD CDNA HW does per-DWORD range checking on buffer loads, with two
+  // failure modes when source rows are not DWORD-aligned:
+  //   1. Per-lane straddle DWORDs read from the *next* source row.
+  //   2. The trailing partial DWORD crosses the buffer end and HW zeros it.
+  // GPUPushDownDMABoundsToConsumers handles both via a consumer-side pad
+  // and an iree_gpu.buffer_resource_cast validBytes override, which makes
+  // DMA correct on non-aligned rows.
+  //
+  // Performance gating: DMA's per-lane gather lowering wins on K-block
+  // boundary tiles (partial K-rows) thanks to HW bounds-checked loads;
+  // it loses on K-aligned shapes because the per-lane scalar copies are
+  // pure overhead vs the non-DMA wide vector loads. Opt in only when the
+  // matmul's K is *not* a multiple of K-tile (i.e. a K-block boundary
+  // actually exists at runtime).
+  //
+  // Detection: the pad result is the workgroup LDS tile. For a matmul
+  // operand the K-tile direction is the smaller of the two result dims
+  // (K-tile is typically 16-32 elements; M-tile / N-tile is 64-256). The
+  // source slice is dynamic in that K-direction iff the matmul's K isn't
+  // K-tile-divisible — exactly the boundary case we want DMA for.
+  auto resultType = dyn_cast<RankedTensorType>(pad.getResult().getType());
+  auto sourceType = dyn_cast<RankedTensorType>(pad.getSource().getType());
+  if (!resultType || !sourceType || resultType.getRank() != 2 ||
+      sourceType.getRank() != 2) {
     return false;
   }
-  Type elemType = sourceType.getElementType();
-  int64_t rowBytes = innermostDim * (elemType.getIntOrFloatBitWidth() / 8);
-  return rowBytes % 4 == 0;
+  int64_t r0 = resultType.getDimSize(0);
+  int64_t r1 = resultType.getDimSize(1);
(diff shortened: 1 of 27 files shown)


Yu-Zhewen (commented):
Thanks @lialan, following up on our discussion the other day, do you think it would be possible to handle all dimensions using `GPUPushDownDMABoundsToConsumers`, instead of just the innermost dimension? That way, we might be able to eliminate the need for [applyOOBClamping](https://github.com/iree-org/iree/blob/01c52ebad4c926e588695c41c39b7b0da3573ed9/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp#L169)

lialan:
> Thanks @lialan, following up on our discussion the other day, do you think it would be possible to handle all dimensions using `GPUPushDownDMABoundsToConsumers`, instead of just the innermost dimension? That way, we might be able to eliminate the need for [applyOOBClamping](https://github.com/iree-org/iree/blob/01c52ebad4c926e588695c41c39b7b0da3573ed9/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp#L169)

@Yu-Zhewen Actually, if the source is a fat raw buffer then we don't need to worry about other dimensions because any OOB in outer dimensions will be clamped by hardware.

So I think we should update `applyOOBClamping`: if the source is a buffer, then hardware clamping will take effect and we get OOB check for free (only for non-innermost dims); otherwise (global memory) we still need the check to ensure correctness.

krzysz00:
One thing I'd like to flag is that I'm not entirely sure the +4 bytes thing is sound - consider the case where the buffer ends near a page boundary?

Yu-Zhewen:
> So I think we should update `applyOOBClamping`: if the source is a buffer, then hardware clamping will take effect and we get OOB check for free (only for non-innermost dims); otherwise (global memory) we still need the check to ensure correctness.

My understanding is hardware OOB clamping is only free for the outermost dimension, not all outer dims? This matters for 3D+ cases.

Also, I'm still unsure why we'd want to keep `applyOOBClamping` instead of extending `GPUPushDownDMABoundsToConsumers` to handle all dimensions. That would give us a uniform solution.

lialan:
> My understanding is hardware OOB clamping is only free for the outermost dimension, not all outer dims? This matters for 3D+ cases.

Actually you are correct, I was referring to only the 2D cases since we only deal with 2D cases.

> 
> Also, I'm still unsure why we'd want to keep `applyOOBClamping` instead of extending `GPUPushDownDMABoundsToConsumers` to handle all dimensions. That would give us a uniform solution.

I think you have a point. Although we only need to push down when innermost dim is not aligned, but having that wouldn't harm.

lialan:
> One thing I'd like to flag is that I'm not entirely sure the +4 bytes thing is sound - consider the case where the buffer ends near a page boundary?

@krzysz00 indeed this will in a very very small chance read cross page boundary. To address this completely we need to change the allocator to pad.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUConvertToCoalescedDMA.cpp
```diff
@@ -293,19 +293,45 @@ static bool isValidPadForDMA(tensor::PadOp pad) {
...
-  // Check if source tensor's innermost row size is DWORD (4-byte) aligned. On
-  // AMD CDNA, per-component range checking is performed for each DWORD. If a
-  // DWORD is partially out-of-bounds, the entire DWORD returns zero, causing
-  // incorrect results. Additionally, partial OOB triggers the slow path with
-  // multi-cycling and instruction issue penalties.
-  auto sourceType = cast<RankedTensorType>(pad.getSource().getType());
-  int64_t innermostDim = sourceType.getShape().back();
-  if (ShapedType::isDynamic(innermostDim)) {
+  // AMD CDNA HW does per-DWORD range checking on buffer loads, with two
+  // failure modes when source rows are not DWORD-aligned:
```
Yu-Zhewen, line 297: Would the logic be much more simpler, if we have this performance gating in ConfigUtils instead

compiler/src/iree/compiler/Codegen/Common/GPU/GPUPushDownDMABoundsToConsumers.cpp
```diff
@@ -0,0 +1,374 @@
...
+//
+// For each iree_gpu.coalesced_gather_dma whose innermost in_bounds entry is
+// false, inserts a tensor.extract_slice + tensor.pad chain between the DMA's
+// outer scf.forall result and its consumer. Downstream vectorization
+// (enable-vector-masking=true) then lowers the pad to a masked
+// vector.transfer_read via vectorizeAsTensorPadOp, discarding straddle-
+// corrupted columns and substituting zero.
+//
+// See docs/superpowers/specs/2026-04-29-dma-mask-pushdown-design.md.
+//
```
Yu-Zhewen, line 17:
```suggestion
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUOps.td
```diff
@@ -294,7 +301,7 @@ def IREEGPU_BufferResourceCastOp : Op<IREEGPU_Dialect, "buffer_resource_cast", [
   let results   = (outs AnyRankedTensor:$result);
 
   let assemblyFormat = [{
-    $input oilist (`cacheSwizzleStride` `(` $cache_swizzle_stride `)` )
+    $input (`cacheSwizzleStride` `(` $cache_swizzle_stride^ `)`)?
```
Yu-Zhewen, line 304: Are the changes to this file and `GPUBubbleResourceCasts.cpp` useful?

compiler/src/iree/compiler/Codegen/Common/GPU/GPUInferMemorySpace.cpp
```diff
@@ -38,12 +39,29 @@ bool isDefinitelyShared(bufferization::AllocTensorOp alloc) {
...
+    if (!visited.insert(user).second) {
+      continue;
+    }
+
     if (isa<linalg::CopyOp>(user) &&
         getLoweringConfig<IREE::GPU::UseGlobalLoadDMAAttr>(user)) {
       continue;
     }
 
+    if (isa<IREE::Codegen::SwizzleHintOp>(user)) {
```
Yu-Zhewen, line 60: I wonder why we didn't hit this issue before, and please add a lit test for it.

PR #24338, opened 2026-05-01
[HAL][local_task] Release retained resources before signaling completion.

### LINKED ISSUE #24331: UBSan CI abort in iree_vm_buffer_deinitialize
CI failure: https://github.com/iree-org/iree/actions/runs/25186189597/job/73844554381?pr=24329

Config: UBSan.
Test:
```
iree-run-mlir --Xcompiler,iree-input-type=stablehlo --Xcompiler,iree-hal-target-device=local --Xcompiler,iree-hal-local-target-device-backends=vmvx /__w/iree/iree/tests/e2e/stablehlo_models/fragment_000.mlir 
```

Stack:

```
   #7 0x00005648f745b649 iree_abort /__w/iree/iree/runtime/src/iree/base/assert.h:26:63
   #8 0x00005648f745b5c0 iree_vm_buffer_deinitialize /__w/iree/iree/runtime/src/iree/vm/buffer.c:79:3
   #9 0x00005648f73810b3 iree_vm_bytecode_module_destroy /__w/iree/iree/runtime/src/iree/vm/bytecode/module.c:171:49
  #10 0x00005648f74617b2 iree_vm_context_release_modules /__w/iree/iree/runtime/src/iree/vm/context.c:0:7
  #11 0x00005648f745dfc7 iree_vm_context_destroy /__w/iree/iree/runtime/src/iree/vm/context.c:476:16
  #12 0x00005648f709325d iree_tooling_run_module_with_data /__w/iree/iree/runtime/src/iree/tooling/run_module.c:476:7
```

The code at iree/runtime/src/iree/vm/buffer.c:79 is:

https://github.com/iree-org/iree/blob/fcbd569042fb7e4d206a5efe5a1651773678d6d0/runtime/src/iree/vm/buffer.c#L79

So apparently we take the abort branch here:

https://github.com/iree-org/iree/blob/fcbd569042fb7e4d206a5efe5a1651773678d6d0/runtime/src/iree/base/internal/atomics.h#L139-L148

bjacob:
Note: This PR (including the description below) is entirely generated by AI (Claude).

Fixes #24331.

`iree_hal_task_queue_op_complete` (and its with-epoch sibling) signaled the operation's signal semaphores *before* freeing the operation's resource set. The resource set retains the command buffer and all binding-table buffers for the duration of the dispatch — so any thread waiting on the signal could observe completion while the worker still held those refs.

In practice this caused intermittent aborts under
`iree-run-mlir --iree-hal-target-device=local --iree-hal-local-target-device-backends=vmvx` during context teardown:

```
  iree_abort                          base/assert.h:26
  iree_vm_buffer_deinitialize         vm/buffer.c:79     // counter == 2
  iree_vm_bytecode_module_destroy     vm/bytecode/module.c:171
  iree_vm_module_release              vm/module.c:238
  iree_vm_context_release_modules     vm/context.c:337
  iree_vm_context_destroy             vm/context.c:482
```

The race chain:

```
  worker:
             op_complete: signal_semaphores  // wakes waiter
             op_complete: ... op_destroy: resource_set_free  // releases refs
  waiter:
             fence.await returns (semaphore signaled)
             vm finishes, iree_vm_invoke returns
             iree_vm_context_release → free_state on HAL module
             iree_vm_module_release on bytecode module
             iree_vm_buffer_deinitialize sees rodata refcount > 1 (worker
                                       hasn't reached resource_set_free yet)
             abort.
```

ThreadSanitizer never flagged this as a data race because every refcount access goes through `iree_atomic_*` and is properly synchronized at the memory-model level. The bug is at a higher abstraction level: a missing happens-before edge between *"owning thread releases its retained resources"* and *"consumer thread observes completion"*.

The fix is mechanical: drop the resource set before publishing the signal in both `op_complete` and `op_complete_with_epoch`. `op_destroy` already guards against a NULL `resource_set`, so the existing teardown path handles the now-already-freed case as a no-op.

Adds a CTS regression test `QueueDispatchTest.BindingBuffersReleasedBeforeSignalIsObserved` that imports a host buffer with a release callback, dispatches with it as a binding, waits on the signal, drops the visible ref, and asserts that the release callback fired synchronously — i.e. the queue had already let go of its retained ref. The test loops 10000 iterations because the race window is microseconds; with the bug present the assertion fires deterministically (observed at iteration ~9429 on local_task+vmvx, TSan build), with the fix in place all iterations pass.

### DIFF
--- a/runtime/src/iree/hal/cts/queue/queue_dispatch_test.cc
+++ b/runtime/src/iree/hal/cts/queue/queue_dispatch_test.cc
@@ -8,6 +8,7 @@
 
 #include <algorithm>
 #include <cstdint>
+#include <cstring>
 #include <numeric>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


benvanik:
this was fixed in https://github.com/iree-org/iree/pull/24307 (or, should have been :)

benvanik:
Ok, confirmed with a local repro attempt. I _think_ the underlying issue this was trying to address may already have been fixed by #24307.

In that, the failure mode there was that VMVX bytecode modules kept pointers into the caller/module-provided executable archive data and local-task could signal queue completion before its implementation-private retained resources are released. A command buffer/executable destructor may run after the waiter has observed completion and if that destructor still depends on VM module rodata then tearing down the VM module/context after waiting can trip over a stale/invalid lifetime edge.

#24307 fixes that at the ownership boundary by making the VMVX module loader clone and own the bytecode executable data on the cold executable-load path. The copy made me sad, but they're tiny and all our production paths (ELFs/etc) all copied as part of loading/linking anyway so I think it's fine. For the contract, completion should mean the submitted work is complete and its memory effects/status are visible and I don't think it should also require that every implementation-private retain acquired during submission has been synchronously released before waiters can observe the signal. Making that a HAL/CTS contract puts arbitrary refcount/destructor work directly in the completion publication path, which is latency-sensitive and awkward for drivers that naturally separate notification publication from later reclaim (AMDGPU/Vulkan/etc all have separate publish vs reclaim mechanisms).

So if the reproducer for this PR was the VMVX/module-rodata lifetime issue I think #24307 should have fixed it without changing the local-task completion path. If there is still a separate reproducer after #24307 I'd like to look at that as a distinct lifetime bug like with the VMVX bytecode data, but I don't think we should add a CTS requirement that imported allocation release callbacks have fired by the time a queue wait returns. That seems stronger than the completion contract we want.

bjacob:
Great. Thanks for taking a look!

PR #24345, opened 2026-05-01
[Codegen][CPU] Lower inner_tiled to llvm.call_intrinsic.

bjacob:
Implements CPU's `DataTiledMMAAttr::buildUnderlyingOperations` by reusing the shared `Codegen::buildDataTiledMMAUnderlyingOperations` from #24326. CPU passes a callback, `createCpuMmaIntrinsicCall`, that emits an `llvm.call_intrinsic` for each per-intrinsic invocation. The function dispatches on the `MMAIntrinsic` enum value through a switch.

Progress towards #24323


compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -628,10 +632,129 @@ DataTiledMMAAttr::getDistributionWorkerCount(OpBuilder &builder, Location loc,
...
+  // Emit llvm.call_intrinsic.
+  auto call = [&builder, loc](StringRef name, Type resultType,
+                              ValueRange args) -> Value {
+    return LLVM::CallIntrinsicOp::create(builder, loc, resultType,
+                                         builder.getStringAttr(name), args)
+        .getResult(0);
+  };
+
+  Type f32 = Float32Type::get(builder.getContext());
+  Type i16 = IntegerType::get(builder.getContext(), 16);
```
hanhanW: Use builtin helper from `builder`.

```suggestion
  Type f32 = builder.getF32Type();
  Type i16 = builder.getI16Type();
```

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -628,10 +632,129 @@ DataTiledMMAAttr::getDistributionWorkerCount(OpBuilder &builder, Location loc,
...
+  // (AVX-512 FMA, VNNI vpdpwssd, BF16 dpbf16ps, integer pmaddwd) take same-
+  // width vector operands — there is no narrow-input variant. The ISA-level
+  // `{1toN}` broadcast-from-memory form is recovered later by the x86
+  // backend's instruction selector pattern-matching this `vector.broadcast`-
+  // of-load into the EVEX broadcast operand, so the explicit broadcast here
+  // is what *enables* that, not a perf liability.
+  // TODO(arm): Arm's by-element FMA (`fmla.4s vd, vn, vm[idx]`) is exposed
+  // via separate intrinsics (e.g. `llvm.aarch64.neon.fma.lane.v4f32`) that
+  // take `(vector, vector, lane_idx)`; when we add Arm support, those cases
+  // should bypass this broadcast and emit the lane-index intrinsic directly.
```
hanhanW: I'm thinking if we use https://github.com/iree-org/iree/issues/24311 in the TODO, assuming that it will have more context instead of a new issue. The issue currently does not have much context, so I don't have strong opinion atm. The best is that you can reference the issue and add comments to the issue. I think they belong the same workstream.

```suggestion
  // TODO(24311): Arm's by-element FMA (`fmla.4s vd, vn, vm[idx]`) is exposed
  // via separate intrinsics (e.g. `llvm.aarch64.neon.fma.lane.v4f32`) that
  // take `(vector, vector, lane_idx)`; when we add Arm support, those cases
  // should bypass this broadcast and emit the lane-index intrinsic directly.
```

hanhanW (approved):
Just some nits

Issue #24309, opened 2026-04-30
[GPU] Numerical Correctness Issue for ArgCompare Op on RDNA3


bangtianliu:
## TL;DR

`arg_compare`  falls back to `GPUDistribute `for small reductions but isn’t distributed to threads; it runs as a shared-memory loop without gpu.barrier, causing a cross-wave data race. On RDNA3 this shows up as non-deterministic wrong results.

## Background

 While debugging the CI failure in [#24291](https://github.com/iree-org/iree/pull/24291) ([failing job](https://github.com/iree-org/iree/actions/runs/25079278088/job/73481188911?pr=24291)), the following ONNX tests fail on `amdgpu_hip_rdna3_O3`:

- `test_argmax_default_axis_random_select_last_index`
- `test_argmin_default_axis_random_select_last_index`

```
[FAILED] result[0]: element at index 3 (1) does not match the expected (0);
expected: 1x3x4xi64=[[1 1 0 0][0 1 1 0][1 1 0 1]]
actual:   1x3x4xi64=[[1 1 0 1][1 1 1 1][1 1 0 1]]
```

The input tensor is `2x3x4xf32`, reduced along dimension 0 (size **2**).

## Reproducer

Save the following as `model.mlir`:

```mlir
func.func @test_argmax_select_last(%arg0: !torch.vtensor<[2,3,4],f32>) -> !torch.vtensor<[1,3,4],si64>
    attributes {torch.onnx_meta.ir_version = 7 : si64,
                torch.onnx_meta.opset_version = 17 : si64} {
  %0 = torch.operator "onnx.ArgMax"(%arg0) {
    torch.onnx.keepdims = 1 : si64,
    torch.onnx.select_last_index = 1 : si64
  } : (!torch.vtensor<[2,3,4],f32>) -> !torch.vtensor<[1,3,4],si64>
  return %0 : !torch.vtensor<[1,3,4],si64>
}
```

Compile and run:

```sh
iree-compile model.mlir \
  --iree-hal-target-device=hip \
  --iree-rocm-target=gfx1100 \
  --iree-opt-level=O3 \
  -o model.vmfb

iree-run-module \
  --module=model.vmfb \
  --device=hip \
  --input="2x3x4xf32=[1 1 1 1  1 1 1 1  1 1 1 1   2 0 2 0  0 2 0 2  2 0 2 0]"
```

**Expected output** :
```
1x3x4xi64=[[1 0 1 0][0 1 0 1][1 0 1 0]]
[...]

PR #24347, opened 2026-05-04
[LLVMGPU] Add TileAndFuse fallback for iree_linalg_ext.arg_compare

bangtianliu:
This PR is a short-term fix: send arg_compare ops unsupported by the `VectorDistribute `pipeline to `TileAndFuse`
- adds a `setArgCompareConfig` lowering-config function that routes                                                                                                                           `iree_linalg_ext.arg_compare` ops to the `LLVMGPUTileAndFuse` pipeline when `setReductionConfig` (the VectorDistribute path) rejects them.  
- gates `iree_linalg_ext.arg_compare` vectorization in the `TileAndFuse` pipeline only, while the resulting  `iree_vector_ext.arg_compare` only has a lowering through the nested-layout distribution patterns owned by the VectorDistribute pipeline. 

Issue: #24309 

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/Passes.td
+++ b/compiler/src/iree/compiler/Codegen/Common/Passes.td
@@ -647,7 +647,9 @@ def GenericVectorizationPass :
             /*default=*/"2147483647",
            "Max vector size allowed to avoid creating large vectors.">,
     Option<"vectorizeMapStore", "vectorize-map-store", "bool", /*default=*/"false",
-      "Enable vectorization of iree_linalg_ext.map_store operations via VectorizableOpInterface.">
+      "Enable vectorization of iree_linalg_ext.map_store operations via VectorizableOpInterface.">,
+    Option<"vectorizeArgCompare", "vectorize-arg-compare", "bool", /*default=*/"true",
+      "Enable vectorization of iree_linalg_ext.arg_compare operations via VectorizableOpInterface.">
   ];
   let dependentDialects = [
     "::mlir::arith::ArithDialect"
--- a/compiler/src/iree/compiler/Codegen/Common/Passes.td
+++ b/compiler/src/iree/compiler/Codegen/Common/Passes.td
@@ -649,7 +649,10 @@ def GenericVectorizationPass :
     Option<"vectorizeMapStore", "vectorize-map-store", "bool", /*default=*/"false",
       "Enable vectorization of iree_linalg_ext.map_store operations via VectorizableOpInterface.">,
     Option<"vectorizeArgCompare", "vectorize-arg-compare", "bool", /*default=*/"true",
-      "Enable vectorization of iree_linalg_ext.arg_compare operations via VectorizableOpInterface.">
+      "Vectorize iree_linalg_ext.arg_compare. Disable for pipelines without a "
+      "lowering for iree_vector_ext.arg_compare. "
+      "TODO(#24365): remove once a generic vector lowering for "
+      "iree_vector_ext.arg_compare exists.">
   ];
   let dependentDialects = [
     "::mlir::arith::ArithDialect"
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -2075,6 +2075,92 @@ LogicalResult setSortConfig(IREE::GPU::TargetAttr target,
                             workgroupSize, subgroupSize));
 }
 
+//====---------------------------------------------------------------------===//
+// ArgCompare Pipeline Configuration
+//====---------------------------------------------------------------------===//
+
+LogicalResult setArgCompareConfig(IREE::GPU::TargetAttr target,
+                                  mlir::FunctionOpInterface entryPoint,
+                                  Operation *op) {
+  assert(isa<IREE::LinalgExt::ArgCompareOp>(op) &&
+         "expected linalg_ext.arg_compare op");
+  auto argCompareOp = cast<IREE::LinalgExt::ArgCompareOp>(op);
+  MLIRContext *context = op->getContext();
+  Builder b(context);
+
+  const int64_t subgroupSize = target.getPreferredSubgroupSize();
+  auto interfaceOp = cast<PartitionableLoopsInterface>(*op);
+  SmallVector<unsigned> partitionedLoops =
+      interfaceOp.getPartitionableLoops(std::nullopt);
+
+  auto createLoweringConfig = [&](ArrayRef<int64_t> workgroupSizes,
+                                  ArrayRef<int64_t> threadSizes) {
+    NamedAttribute attrs[2] = {{"workgroup", b.getI64ArrayAttr(workgroupSizes)},
+                               {"thread", b.getI64ArrayAttr(threadSizes)}};
+    auto configDict = b.getDictionaryAttr(attrs);
+    return IREE::GPU::LoweringConfigAttr::get(context, configDict);
+  };
+
+  // No parallel dims (e.g. rank-1 input reducing to a scalar) means there is
+  // nothing for the TileAndFuse pipeline to distribute across threads, and the
+  // resulting `tile = [0]` config trips downstream tiling passes. Bail out and
+  // let the dispatcher fall through to `setRootDefaultConfig`.
+  if (partitionedLoops.empty()) {
+    return failure();
+  }
+
+  // arg_compare's iteration domain has one loop per input dim; the single
+  // reduction dim is `getDimension()` and the rest are parallel.
+  int64_t numLoops = argCompareOp.getInputRank();
+
+  // Starting point pending tuning data: two warps per workgroup mirrors
+  // setSortConfig and gives reasonable occupancy without over-subscribing
+  // for the small parallel iteration spaces typical of arg_compare fallbacks.
+  std::array<int64_t, 3> workgroupSize = {2 * subgroupSize, 1, 1};
+  SmallVector<int64_t> workgroupTileSizes(numLoops, 1);
+  SmallVector<int64_t> threadTileSizes(numLoops, 1);
+
+  // Set non-parallel loops (the single reduction dim) to zero tile size so
+  // each thread runs the full reduction on its parallel slice.
+  llvm::DenseSet<unsigned> partitionedLoopsSet(llvm::from_range,
+                                               partitionedLoops);
+  for (int64_t depth : llvm::seq<int64_t>(0, numLoops)) {
+    if (!partitionedLoopsSet.contains(depth)) {
+      workgroupTileSizes[depth] = 0;
+      threadTileSizes[depth] = 0;
+    }
+  }
+
+  // Pack as many parallel-dim iterations as possible into the workgroup tile
+  // (innermost dim first), matching the setSortConfig packing strategy.
+  ArrayRef<int64_t> loopBounds = argCompareOp.getInputType().getShape();
+  int64_t residualWorkgroupSize = workgroupSize[0];
+  for (int64_t depth = numLoops - 1; depth >= 0; --depth) {
+    if (!partitionedLoopsSet.contains(depth)) {
+      continue;
+    }
+    if (ShapedType::isDynamic(loopBounds[depth])) {
+      continue;
+    }
+    if (residualWorkgroupSize % loopBounds[depth] == 0) {
+      workgroupTileSizes[depth] = loopBounds[depth];
+      residualWorkgroupSize /= loopBounds[depth];
+      continue;
+    }
+    if (loopBounds[depth] % residualWorkgroupSize == 0) {
+      workgroupTileSizes[depth] = residualWorkgroupSize;
+      break;
+    }
+  }
+
+  IREE::GPU::LoweringConfigAttr loweringConfig =
+      createLoweringConfig(workgroupTileSizes, threadTileSizes);
+  return setOpConfigAndEntryPointFnTranslation(
+      entryPoint, op, loweringConfig,
+      getGPUTranslationInfo(op->getContext(), LoweringPipeline::TileAndFuse,
+                            workgroupSize, subgroupSize));
+}
+
 //===----------------------------------------------------------------------===//
 // Lowering Config Attributes
 //===----------------------------------------------------------------------===//
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
@@ -2082,8 +2082,6 @@ LogicalResult setSortConfig(IREE::GPU::TargetAttr target,
 LogicalResult setArgCompareConfig(IREE::GPU::TargetAttr target,
                                   mlir::FunctionOpInterface entryPoint,
                                   Operation *op) {
-  assert(isa<IREE::LinalgExt::ArgCompareOp>(op) &&
-         "expected linalg_ext.arg_compare op");
   auto argCompareOp = cast<IREE::LinalgExt::ArgCompareOp>(op);
   MLIRContext *context = op->getContext();
   Builder b(context);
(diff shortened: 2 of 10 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -2075,6 +2075,92 @@ LogicalResult setSortConfig(IREE::GPU::TargetAttr target,
...
+    return IREE::GPU::LoweringConfigAttr::get(context, configDict);
+  };
+
+  // No parallel dims (e.g. rank-1 input reducing to a scalar) means there is
+  // nothing for the TileAndFuse pipeline to distribute across threads, and the
+  // resulting `tile = [0]` config trips downstream tiling passes. Bail out and
+  // let the dispatcher fall through to `setRootDefaultConfig`.
+  if (partitionedLoops.empty()) {
+    return failure();
+  }
```
Max191, line 2117: I think in this case, you could set the workgroupSize to {1, 1, 1} instead of bailing and going to the other pipeline. If you do that will it compile for these 1D cases?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -2075,6 +2075,92 @@ LogicalResult setSortConfig(IREE::GPU::TargetAttr target,
...
+
+  // arg_compare's iteration domain has one loop per input dim; the single
+  // reduction dim is `getDimension()` and the rest are parallel.
+  int64_t numLoops = argCompareOp.getInputRank();
+
+  // Starting point pending tuning data: two warps per workgroup mirrors
+  // setSortConfig and gives reasonable occupancy without over-subscribing
+  // for the small parallel iteration spaces typical of arg_compare fallbacks.
+  std::array<int64_t, 3> workgroupSize = {2 * subgroupSize, 1, 1};
```
Max191, line 2126: Minor improvement: If the workgroup tile size computation below yields a cumulative tile size less than or equal to `1 * subgroupSize`, then you could reduce the workgroupSize down to just 1 subgroup after you've computed the workgroup tile sizes.

Max191 (commented):
Overall LGTM, just a couple of small comments

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
bangtianliu: Let me have a try

bangtianliu: Ok, trigger CI error here: https://github.com/iree-org/iree/actions/runs/25334858337/job/74277465143?pr=24347#step:5:11875

 Tried the rank-1 single-thread path, which crashes the realistic CI cases (linalg.fill-seeded outs). The mismatch is between iteration-domain rank and result rank:                                                   
 
 ```mlir                                                                                                                                                                                            
 // arg_compare: input rank 1, result rank 0 (scalar)                                                                                                                                        
 %0:2 = iree_linalg_ext.arg_compare
     {lowering_config = #iree_gpu.lowering_config<{thread = [0], workgroup = [0]}>}                                                                                                            
     dimension(0) ins(%input : tensor<5xf32>)                                                                                                                                                  
     outs(%fv, %fi : tensor<f32>, tensor<i32>) ...                                                                                                                                             
 ```                                                                                                                                                                                    
 The lowering_config is length-1 (one entry per iteration-domain dim = input rank), but the result is rank-0. `GPUGreedilyDistributeToThreadsPass` walks the result iteration space and indexes the tile array with the result rank, so tile[0] against a rank-0 target overflows -> ArrayRef::operator[] assert.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -2075,6 +2075,107 @@ LogicalResult setSortConfig(IREE::GPU::TargetAttr target,
...
 
+//====---------------------------------------------------------------------===//
+// ArgCompare Pipeline Configuration
+//====---------------------------------------------------------------------===//
+
+LogicalResult setArgCompareConfig(IREE::GPU::TargetAttr target,
+                                  mlir::FunctionOpInterface entryPoint,
+                                  Operation *op) {
+  assert(isa<IREE::LinalgExt::ArgCompareOp>(op) &&
+         "expected linalg_ext.arg_compare op");
```
kuhar: this assert is redundant, the cast below already asserts the same condition. See https://llvm.org/docs/ProgrammersManual.html#the-isa-cast-and-dyn-cast-templates

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -2075,6 +2075,107 @@ LogicalResult setSortConfig(IREE::GPU::TargetAttr target,
...
+  auto interfaceOp = cast<PartitionableLoopsInterface>(*op);
+  SmallVector<unsigned> partitionedLoops =
+      interfaceOp.getPartitionableLoops(std::nullopt);
+
+  auto createLoweringConfig = [&](ArrayRef<int64_t> workgroupSizes,
+                                  ArrayRef<int64_t> threadSizes) {
+    NamedAttribute attrs[2] = {{"workgroup", b.getI64ArrayAttr(workgroupSizes)},
+                               {"thread", b.getI64ArrayAttr(threadSizes)}};
+    auto configDict = b.getDictionaryAttr(attrs);
+    return IREE::GPU::LoweringConfigAttr::get(context, configDict);
```
kuhar: nit: inline this since there's only one use
```suggestion
    return IREE::GPU::LoweringConfigAttr::get(context, b.getDictionaryAttr(attrs));
```

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -571,10 +574,18 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(createGPUCombineValueSemanticBarriersPass());
 
   // Step 6. Vectorize.
+  // Skip vectorization of `iree_linalg_ext.arg_compare` here: the resulting
+  // `iree_vector_ext.arg_compare` only has a lowering through the nested-layout
+  // distribution patterns owned by the VectorDistribute pipeline. Ops routed
+  // to TileAndFuse (small reductions, f64) need the scalar-loop lowering from
+  // `LinalgExtToLoops` instead. This gate is intentionally narrow and should
+  // be removed once a TileAndFuse-compatible lowering for
```
kuhar: Can you add a TODO with the issue URL if you know you will have to eventually remove this option?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Max191: Hmm, this seems like a bug in either the tiling interface implementation for arg_compare or in the GPUGreedilyDistributeToThreads pass. The distribution should just operate on the iteration space of the op, which in this case is rank 1, not rank 0. If it is somehow only looking at the result value to determine the iteration space, then that seems wrong to me. I'd think this should behave the same way that a 1D regular reduction op would (I don't know if 1D regular reduction also hits the same failure or not, but just saying that they should kinda be the same).

If we are hitting assertion errors in that pass then it's worth fixing, because this looks like it should be a valid config to me.

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -647,7 +647,9 @@ def GenericVectorizationPass :
             /*default=*/"2147483647",
            "Max vector size allowed to avoid creating large vectors.">,
     Option<"vectorizeMapStore", "vectorize-map-store", "bool", /*default=*/"false",
-      "Enable vectorization of iree_linalg_ext.map_store operations via VectorizableOpInterface.">
+      "Enable vectorization of iree_linalg_ext.map_store operations via VectorizableOpInterface.">,
+    Option<"vectorizeArgCompare", "vectorize-arg-compare", "bool", /*default=*/"true",
+      "Enable vectorization of iree_linalg_ext.arg_compare operations via VectorizableOpInterface.">
```
hanhanW: Why do we need this option?

bangtianliu: Vectorization would produce `iree_vector_ext.arg_compare`, which has no lowering in the TileAndFuse pipeline. We just let it go to `LinalgExtToLoops`, which can handle it automatically along the TileAndFuse pipeline. 

Lowering for `iree_vector_ext.arg_compare` is the trick along VectorDistribute, about some layout, distribution, and dpp + ballot

hanhanW: Something is missing from mapping the iree_vector_ext.arg_compare to some vector ops, IMO. It is also failing if we switch CPU backend to `LinalgExtTileAndVectorize` pipeline.

VectorDistribution pipeline is a harder one for sure, but the basic vector lowering is missing.

What is the core implementation for `VectorExt::ArgCompareOp`? I wonder why can't we use `LinalgExt::ArgCompareOp` to represent it and we can maybe handle the lowering in LinalgExtToLoops implementation. The `*ToLoops` is the final fallback for scaler code, so it seems better if we can line things up automatically in this context.

hanhanW: Maybe there are some traits only available in VectorExt dialect, can you provide more context and figure a way for basic plumbing?

bangtianliu: > Something is missing from mapping the iree_vector_ext.arg_compare to some vector ops, IMO

Along the VectorDistribute pipeline, iree-vector_ext.arg_compare is lowered into gpu.subgroup_reduce + gpu.ballot to leverage DPP and ballot instructions provided by AMDGPU backend. No vector lowering at all. 

Yes, I think for the CPU backend, something is missing here since I did not focus on the CPU part at all 

> Maybe there are some traits only available in VectorExt dialect, can you provide more context and figure a way for basic plumbing

The reason iree_vector_ext.arg_compare exists as a separate op (introduced in #23386) is that its operands and results are typed `vector<...>` rather than `tensor<...>`, and it carries the layout/distribution traits the VectorDistribute needs.

bangtianliu: A basic plumbing is missing there, like add a generic lowering of iree_vector_ext.arg_compare into vector.* ops which will unblock CPU or any GPU backends that do not have AMDGPU ballot or DPP instruction

bangtianliu: I can add this basic plumbing to my plate, too.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
bangtianliu: Fixed in the latest commit, added a loopRanges.empty() early-return in `getVectorTileSizesFromLoopRanges ` which is used in GPUGreedilyDistributeToThreads pass

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_argcompare.mlir
```diff
@@ -213,34 +213,114 @@ func.func @argcompare_argmin() {
...
-  %3 = iree_tensor_ext.dispatch.tensor.load %0, offsets = [0, 0], sizes = [8, 256], strides = [1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<8x256xf64>> -> tensor<8x256xf64>
-  %4 = tensor.empty() : tensor<8xf64>
-  %5 = tensor.empty() : tensor<8xi32>
-  %6:2 = iree_linalg_ext.arg_compare
+// CHECK:       #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
+// CHECK-SAME:    workgroup_size = [{{[0-9]+}}, 1, 1] subgroup_size = 64
+// CHECK-LABEL: func.func @argcompare_small_reduction_f32
+// CHECK:         iree_linalg_ext.arg_compare
+// CHECK-SAME:      lowering_config = #iree_gpu.lowering_config<{thread = [0,
+// CHECK-SAME:        workgroup = [0,
```
Max191: Please validate the actual configurations here (i.e., the workgroup size values, the full sets of tile sizes). I'm not seeing any tests that actually check for the expected configurations.

Max191 (approved):
LGTM, just a comment about the test checks.

compiler/src/iree/compiler/Codegen/Common/Passes.td
hanhanW: We can make LinalgExt version accept vector types. @Max191 did it for MapStore op, e.g.,

https://github.com/iree-org/iree/blob/020f6bed41819712f47bf63702fa3472d464807e/compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_map_store.mlir#L3-L20

The MapStore version requires running DecomposeMapStore in vector lowering. I think we can either follow that, or just make generateScalarImplementation to handle that (in the context of "scalar == single unit which is a vector). Or we can introduce new interface method for the lowering. There are a couple options.

hanhanW: > it carries the layout/distribution traits the VectorDistribute needs

I missed this in the first read. Then yes, we probably need a vector op for this. I think the missing part is something like "lower virtual vector ops to basic or low-level vector ops".

bangtianliu: > > it carries the layout/distribution traits the VectorDistribute needs
> 
> I missed this in the first read. Then yes, we probably need a vector op for this. I think the missing part is something like "lower virtual vector ops to basic or low-level vector ops".

Yes, for CPU or other GPU backends, we need to add a lowering to vector.* ops.

hanhanW: This part looks good to me if you can file an issue for tracking this and reference the issue ID here.

bangtianliu: Just to double confirm: do you mean the issue about basic lowering of `iree_vector_ext.arg_compare` since I already have one issue https://github.com/iree-org/iree/issues/24309 here.

hanhanW: Yes, thanks! I did not see the issue when I left the comment.

bangtianliu: Created one issue #24365 about the lowering of `iree_vector_ext.arg_compare` and self-assigned.

compiler/src/iree/compiler/Codegen/Common/Passes.td
```diff
@@ -647,7 +647,12 @@ def GenericVectorizationPass :
             /*default=*/"2147483647",
            "Max vector size allowed to avoid creating large vectors.">,
     Option<"vectorizeMapStore", "vectorize-map-store", "bool", /*default=*/"false",
-      "Enable vectorization of iree_linalg_ext.map_store operations via VectorizableOpInterface.">
+      "Enable vectorization of iree_linalg_ext.map_store operations via VectorizableOpInterface.">,
+    Option<"vectorizeArgCompare", "vectorize-arg-compare", "bool", /*default=*/"true",
+      "Vectorize iree_linalg_ext.arg_compare. Disable for pipelines without a "
+      "lowering for iree_vector_ext.arg_compare. "
+      "TODO(Bangtian, https://github.com/iree-org/iree/issues/24309): remove "
+      "once VectorDistribute supports masking.">
```
hanhanW: It the wording correct? I thought we remove it once we have the vector lowering for the op.

```suggestion
      "TODO(#24365): remove once VectorDistribute supports masking.">
```

And the issue ID should be 24365?

bangtianliu: Two todos for me: 1) add vector lowering, 2) masking support for arg_compare. 


Yes, you are correct. Let me make it clear.

bangtianliu: Ok, updated in the latest commit

hanhanW: I think we prefer `TODO(#ISSUE-ID)` style.

hanhanW (approved):
LGTM on vectorization changes % the TODO style.

PR #24349, opened 2026-05-04
[LinalgExt] Fix attention index remap after unit-dim folding

keshavvinayak01:
Remap `iree_linalg_ext.index` ops in attention regions when unit loop dims are dropped so later decomposition cannot create out-of-range `linalg.index` ops.

Also updated stale doc of `iree_linalg_ext.index` op, since `OnlineAttention` support was already added.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ReshapeFusion.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ReshapeFusion.cpp
@@ -1334,6 +1334,27 @@ SmallVector<unsigned> defaultControlDropUnitDims(Operation *op) {
   return llvm::to_vector(llvm::seq<unsigned>(0, fusionOp.getNumLoops()));
 }
 
+static void
+replaceUnitDimIndexOps(Block *body,
+                       const llvm::SmallDenseSet<unsigned> &unitDims,
+                       RewriterBase &rewriter) {
+  for (IndexOp indexOp : llvm::make_early_inc_range(body->getOps<IndexOp>())) {
+    OpBuilder::InsertionGuard guard(rewriter);
+    rewriter.setInsertionPoint(indexOp);
+    if (unitDims.contains(indexOp.getDim())) {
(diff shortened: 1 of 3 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ReshapeFusion.cpp
```diff
@@ -1334,6 +1334,29 @@ SmallVector<unsigned> defaultControlDropUnitDims(Operation *op) {
   return llvm::to_vector(llvm::seq<unsigned>(0, fusionOp.getNumLoops()));
 }
 
+template <typename AttentionOpType>
+static void
+replaceUnitDimIndexOps(AttentionOpType attentionOp,
+                       const llvm::SmallDenseSet<unsigned> &unitDims,
+                       RewriterBase &rewriter) {
```
IanWood1: nit: you could pass `attentionOp.getBody()` if you don't want this to be templated.

IanWood1 (approved):
LGTM!

compiler/src/iree/compiler/DispatchCreation/test/fold_unit_dims.mlir
```diff
@@ -694,3 +694,61 @@ util.func public @online_attention_unit_m_dim(
 //       CHECK:   %[[EXPAND2:.+]] = tensor.expand_shape %[[ATTN]]#2
 //  CHECK-SAME:     tensor<8x4xf32> into tensor<8x4x1xf32>
 //       CHECK:   util.return %[[EXPAND]], %[[EXPAND1]], %[[EXPAND2]]
+
+// -----
+
+util.func public @online_attention_remap_index_after_unit_dim_drop(
```
bangtianliu, line 704: Consider adding a negative test:  attention without index ops should be unaffected

bangtianliu (approved):

compiler/src/iree/compiler/DispatchCreation/test/fold_unit_dims.mlir
keshavvinayak01: Done

PR #24350, opened 2026-05-04
[Codegen] Limit async scope in pipelining

sommerlukas:
Currently, the pipelining assumes that all `amdgpu.gather_to_lds` operations are part of the loop that is being pipelined and marks them as async, with waits inserted in the pipelined loop.

This assumption was fine for the workloads so far, but with work underway to enable pipelining for attention, this no longer holds, as the load of e.g. the `Q` matrix can be outside of the loop. Operations outside the loop can also be marked async, but that requires inserting marks and waits before the loop.

This PR fixes this by recording a marker in the block before the loop and later on inserting marks and waits for async operations before the marker.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -1095,14 +1095,6 @@ static void insertAsyncDrainBefore(RewriterBase &rewriter, Location loc,
 /// Converts direct pre-existing gather_to_lds ops before the pipelined loop to
 /// async mode and inserts explicit waits that preserve their original
 /// synchronous behavior.
-///
-/// Pre-loop async groups must be fully drained before the pipelining prologue
-/// starts. Otherwise, the loop body's wait.asyncmark N would count unrelated
-/// pre-loop groups and no longer correspond to the intended N-stage pipeline.
-///
-/// This only handles direct gather_to_lds ops in the parent block. Recursively
-/// converting nested pre-loop gathers would require placing marks and waits
-/// inside those nested regions according to their local control flow.
 static void insertPreLoopAsyncMarkers(RewriterBase &rewriter, Location loc,
                                       Block::iterator preLoopStart,
                                       Block::iterator preLoopEnd) {
@@ -1324,11 +1316,17 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
     return forOp;
   }
 
+  Operation *opBeforeLoop = nullptr;
   if (mode == PipelineMode::AsyncCopy) {
     // Apply multi-buffering: numStages buffers for N-stage pipelining.
     if (failed(multiBufferLDSAllocations(forOp, /*numBuffers=*/numStages))) {
       return failure();
     }
+    // Record the operation before the original loop. After pipelining, the next
+    // operation after this marker is the start of the generated prologue. Ops
+    // before that boundary are pre-existing parent-block ops and are explicitly
+    // drained separately from the pipelined async groups.
+    opBeforeLoop = forOp->getPrevNode();
   }
 
   // Re-run classification on the (potentially multi-buffered) IR to capture
@@ -1363,15 +1361,6 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
   LLVM_DEBUG(dumpSchedule(finalSchedule, opToCluster));
 
   scf::PipeliningOption options = buildPipeliningOption(finalSchedule);
-
-  Operation *opBeforeLoop = nullptr;
-  if (mode == PipelineMode::AsyncCopy) {
-    // Record the operation before the original loop. After pipelining, the next
-    // operation after this marker is the start of the generated prologue. Ops
-    // before that boundary are pre-existing parent-block ops and are explicitly
-    // drained separately from the pipelined async groups.
-    opBeforeLoop = forOp->getPrevNode();
-  }
 
   FailureOr<scf::ForOp> newForOpOr = invokePipelineForLoop(forOp, options);
   if (failed(newForOpOr)) {
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
@@ -1075,11 +1075,17 @@ static Operation *findLastGatherToLDS(Block::iterator begin,
   return nullptr;
 }
 
-/// Sets the async flag on all gather_to_lds ops in the parent block so they
+/// Sets the async flag on gather_to_lds ops in the given block range so they
 /// lower to rocdl.load.async.to.lds instead of rocdl.load.to.lds.
-static void enableAsyncOnGatherOps(Block *parentBlock) {
-  parentBlock->walk(
-      [](amdgpu::GatherToLDSOp gatherOp) { gatherOp.setAsync(true); });
+/// Only affects ops between prologueStart and the end of the block, plus ops
+/// inside the loop body. This avoids making pre-existing gather ops (e.g., Q
+/// loads consumed before the loop) async, which would be incorrect since those
+/// ops lack corresponding asyncmark/wait.asyncmark synchronization.
+static void enableAsyncOnGatherOps(Block *parentBlock,
+                                   Block::iterator prologueStart) {
+  for (auto it = prologueStart; it != parentBlock->end(); ++it) {
+    it->walk([](amdgpu::GatherToLDSOp gatherOp) { gatherOp.setAsync(true); });
+  }
 }
 
 /// Inserts asyncmark ops in the prologue to delineate DMA groups.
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir
```diff
@@ -799,3 +799,45 @@ func.func @gather_to_lds_with_swizzle_attr(
...
+  %c128 = arith.constant 128 : index
+  %c1 = arith.constant 1 : index
+  %c0 = arith.constant 0 : index
+
+  %Q_lds = memref.alloc() : memref<1xf32, #gpu.address_space<workgroup>>
+  %K_lds = memref.alloc() : memref<1xf32, #gpu.address_space<workgroup>>
+
+  // Pre-loop gather_to_lds (e.g., loading Q) - must stay synchronous (no async).
+  // CHECK: amdgpu.gather_to_lds %{{.*}}[%{{.*}}, %{{.*}}], %{{.*}}[%{{.*}}] : vector<1xf32>
+  amdgpu.gather_to_lds %A_global[%c0, %c0], %Q_lds[%c0] : vector<1xf32>, memref<128x128xf32>, memref<1xf32, #gpu.address_space<workgroup>>
```
jerryyin: Question/unrelated: Do we codegen like this in actual attention? I don't think Q needs to be put in shared memory and can directly load into register instead.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -1318,6 +1324,20 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
...
+  // (e.g., Q loads in attention) are not part of the pipeline and must not be
+  // converted to async mode.
+  Block *parentBlock = forOp->getBlock();
+  Block::iterator preLoopMarker = forOp->getIterator();
+  bool noOpsBeforeLoop = (preLoopMarker == parentBlock->begin());
+  if (!noOpsBeforeLoop) {
+    // Decrement to point to the operation before the loop that is going to be
+    // transform, otherwise the marker would be invalidated.
+    --preLoopMarker;
+  }
```
jerryyin: Two comments:
 1. Can we put this to previous ` if (mode == PipelineMode::AsyncCopy) {` guard so this is clear that it is only needed in async copy mode? Around line 1291 above.
 2. Regarding implementation, can you do this instead:
 ```cpp
Block *parentBlock = forOp->getBlock();
Operation *opBeforeLoop = forOp->getPrevNode();
```

File: compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir (hunk 1 above)
Comment (jerryyin) [inline compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir]: Let's do a CHECK-NOT: amdgpu.gather_to_lds async

**Review by jerryyin — APPROVED** [review]
LGTM with some nits

**Review by krzysz00 — CHANGES_REQUESTED** [review]
Strong -1 to this : those gathers to LDS should be using asyncmark / asyncwait for future-proofing. Future DMAs will not have "sync" variants.

So instead, we should extend the async mark forming logic to cover these outside-of-loop cases, or move it out of this prefetch stuff (have DMA lowerings create async marks and have the pipeliner fix them)

**Comment by jerryyin** [comment]
> Future DMAs will not have "sync" variants.

Good clarification! When is backend planning to decommission those? CC @lialan 

Also, what will the future DMA look like then? Like do they by default carries the "aysnc" flag or non-flagged gather_to_lds by default become async version?

>  those gathers to LDS should be using asyncmark / asyncwait for future-proofing.....So instead, we should extend the async mark forming logic to cover these outside-of-loop cases, or move it out of this prefetch stuff (have DMA lowerings create async marks and have the pipeliner fix them)

That means we'd need to delete those code from pipeliner that turn the gather_to_lds from "sync" version to "async" version. The async flag should be natively what it carries. Agreed, pipeliner has been silently doing the part of work DMA load should be doing - and that's somewhat wrong and deserves a cleanup.

**Comment by krzysz00** [comment]
> Good clarification! When is backend planning to decommission those? 

There isn't any plan to eliminate the implicit-tracking for gfx9, since those use VMCNT and no one wants to break userspace.

However, gfx1250's DMAs use ASYNCCNT and TENSORCNT, which are entirely separate memory operation queues than what's called LOADCNT on gfx12 (for ordinary global loads). It has been decided that trying to recover the asynccnt/tensorcnt automagically is rarely what people want / a lot of effort, so the compromise is using the mark/wait framework to save people from counting (and make function calls, dead code elimination, etc. work)

As to what'll be happening, `gather_to_lds` just isn't a thing post gfx10 - you get `global_load_async_to_lds`  , which both doesn't have the contiguity requirement and is always "async" in our sense

File: compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir (hunk 1 above)
Comment (sommerlukas) [reply compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir]: Yes, we're currently propagating all operands to shared memory, including Q. I can review if there's potential in not putting Q in LDS, though with reuse of Q across different threads in the workgroup/subgroups, it might still make sense. I'll investigate that as part of the work on a more attention-specific schedule.

**Review by sommerlukas — COMMENTED** [review]
Based on @krzysz00's feedback, I've changed the approach. Operations outside the pipelined loop are now also marked async and the necessary marks and waits are inserted. This is done by iterating the block and inserting async marks and waits if the current operation is a barrier or has (nested) LDS reads and any async operations have been encountered at this point.

I've updated the PR description accordingly.

**Review by Yu-Zhewen — APPROVED** [review]
LGTM

File: compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp [hunk 3]
```diff
@@ -1318,6 +1364,15 @@ FailureOr<scf::ForOp> prefetchSharedMemoryCopy(RewriterBase &rewriter,
...
   scf::PipeliningOption options = buildPipeliningOption(finalSchedule);
 
+  Operation *opBeforeLoop = nullptr;
+  if (mode == PipelineMode::AsyncCopy) {
+    // Record the operation before the original loop. After pipelining, the next
+    // operation after this marker is the start of the generated prologue. Ops
+    // before that boundary are pre-existing parent-block ops and are explicitly
+    // drained separately from the pipelined async groups.
+    opBeforeLoop = forOp->getPrevNode();
+  }
```
jerryyin: Can you consolidate this with line 1327 after change? (The same condition before here).

Between 1327 to here we are mostly preparing for actual pipeline steps -> pipeline. This break the logical flow a bit.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -1075,11 +1075,56 @@ static Operation *findLastGatherToLDS(Block::iterator begin,
...
+                                   Operation *insertPt) {
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(insertPt);
+  ROCDL::AsyncmarkOp::create(rewriter, loc);
+  ROCDL::WaitAsyncmarkOp::create(rewriter, loc, rewriter.getI16IntegerAttr(0));
+}
+
+/// Converts direct pre-existing gather_to_lds ops before the pipelined loop to
+/// async mode and inserts explicit waits that preserve their original
+/// synchronous behavior.
```
jerryyin, line 1097: [Nit] This section of comment is good enough but the section below feels quite verbose.

jerryyin (approved):
LGTM with two nits.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/prefetch_shared_memory.mlir
jerryyin: To my understanding Q is kept in each thread and putting them to shared memory doesn't improve much re-use but occupies shared memory unnecessarily (also may get impacted by bank conflict, for example). This shouldn't be part of pipeline work though but a solid improvement.

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -1075,11 +1075,48 @@ static Operation *findLastGatherToLDS(Block::iterator begin,
...
 
-/// Sets the async flag on all gather_to_lds ops in the parent block so they
+/// Sets the async flag on gather_to_lds ops from `begin` to `end` so they
 /// lower to rocdl.load.async.to.lds instead of rocdl.load.to.lds.
-static void enableAsyncOnGatherOps(Block *parentBlock) {
-  parentBlock->walk(
-      [](amdgpu::GatherToLDSOp gatherOp) { gatherOp.setAsync(true); });
+static void enableAsyncOnGatherOps(Block::iterator begin, Block::iterator end) {
+  for (auto it = begin; it != end; ++it) {
+    it->walk([](amdgpu::GatherToLDSOp gatherOp) { gatherOp.setAsync(true); });
```
krzysz00, line 1082: I'm just going to note that `walk` is recursive

compiler/src/iree/compiler/Codegen/LLVMGPU/Utils/ROCDLPrefetchSharedMemoryCopy.cpp
```diff
@@ -1075,11 +1075,48 @@ static Operation *findLastGatherToLDS(Block::iterator begin,
...
+}
+
+/// Converts direct pre-existing gather_to_lds ops before the pipelined loop to
+/// async mode and inserts explicit waits that preserve their original
+/// synchronous behavior.
+static void insertPreLoopAsyncMarkers(RewriterBase &rewriter, Location loc,
+                                      Block::iterator preLoopStart,
+                                      Block::iterator preLoopEnd) {
+  bool hasPendingAsyncGather = false;
+  for (auto it = preLoopStart; it != preLoopEnd;) {
```
krzysz00, line 1102: Hold on, won't this pessimize the gathers pulled before the loop? We'll probably want to reset the pendingness-check if we see an asyncmark

krzysz00: ... ah, now I see, you're good

krzysz00 (approved):

PR #24352, opened 2026-05-04
[Codegen][CPU] Add LLVMCPULowerInnerTiledPass and schedule it.

bjacob:
Lowers vector-semantics `iree_codegen.inner_tiled` ops on the CPU side
by running two pattern sets in sequence on the function body:

  1. `IREE::Codegen::populateDropInnerTiledUnitDimsPatterns` drops the
     (now-unit) iter domain so the lowered op has empty iter bounds,
     satisfying the precondition of (2).
  2. `IREE::Codegen::populateLowerInnerTiledPatterns` replaces the
     iter-free vector-semantics inner_tiled with the per-intrinsic
     `llvm.call_intrinsic` ops emitted by
     `DataTiledMMAAttr::buildUnderlyingOperations`.

Both pattern sets were lifted out of `Codegen/Dialect/GPU/Transforms/`
into `Codegen/Dialect/Codegen/Transforms/` in #24351
so CPU can share them with GPU. They are interface-driven and pick up
CPU's `IREE::CPU::DataTiledMMAAttr` transparently via its
`buildUnderlyingOperations` implementation.

The pass expects vector-semantics input, which the upstream
`GenericVectorizationPass` produces via the
`InnerTiledOpVectorizationModel` `VectorizableOpInterface` external
model already in tree. Schedule the new pass right after
`GenericVectorizationPass` in `addMmt4dTilingExpertPassPipeline`, where
data-tiled-MMA dispatches will land once the next commit teaches the
lowering strategy to route them through that pipeline. The pass is a
no-op when the function has no inner_tiled ops.

Adds a focused lit test
`compiler/src/iree/compiler/Codegen/LLVMCPU/test/lower_inner_tiled.mlir`
that feeds a single-tile AVX-512 1x16x1 f32 vector-semantics
inner_tiled and asserts the surviving `llvm.fma.v16f32` call.

Progress towards https://github.com/iree-org/iree/issues/24323


compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.td
```diff
@@ -95,6 +95,31 @@ def LLVMCPULowerExecutableTargetPass :
...
+          inner_tiled ops to vector semantics.
+      (2) `populateIREEGPUDropUnitDimsPatterns`: drop the (now-unit) iteration
+          domain so the lowered op has empty iter bounds.
+      (3) `populateIREEGPULowerInnerTiledPatterns`: replace the vector-
+          semantics, iter-domain-free inner_tiled with the per-intrinsic
+          `llvm.call_intrinsic` ops emitted by
+          `DataTiledMMAAttr::buildUnderlyingOperations`.
+    Despite the `IREEGPU` names, patterns (2) and (3) are interface-driven and
+    work for any `InnerTileDescAttrInterface` kind, including CPU's
+    `IREE::CPU::DataTiledMMAAttr`.
```
hanhanW, line 114: seems to be outdated? GPU should not be mentioned at all in this pass.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerInnerTiled.cpp
```diff
@@ -0,0 +1,47 @@
...
+    //     empty, satisfying the precondition of (2).
+    IREE::Codegen::populateDropInnerTiledUnitDimsPatterns(patterns);
+    // (2) Replace the iter-free vector-semantics inner_tiled with the
+    //     per-intrinsic ops emitted by `buildUnderlyingOperations`. The
+    //     pattern is interface-driven and picks up CPU's DataTiledMMAAttr
+    //     transparently.
+    IREE::Codegen::populateLowerInnerTiledPatterns(patterns);
+    if (failed(applyPatternsGreedily(getOperation(), std::move(patterns)))) {
+      return signalPassFailure();
+    }
```
krzysz00, line 42: I think you'll want to run the hoistable conversions code after this

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.td
```diff
@@ -95,6 +95,31 @@ def LLVMCPULowerExecutableTargetPass :
...
 
+def LLVMCPULowerInnerTiledPass : InterfacePass<"iree-llvmcpu-lower-inner-tiled",
+                                               "mlir::FunctionOpInterface"> {
+  let summary =
+      "Lower iree_codegen.inner_tiled ops to llvm.call_intrinsic on CPU.";
+  let description = [{
+    Runs three pattern sets in sequence on the function body:
+      (1) `populateVectorizeInnerTiledPatterns`: lift tensor-semantics
+          inner_tiled ops to vector semantics.
+      (2) `populateIREEGPUDropUnitDimsPatterns`: drop the (now-unit) iteration
```
krzysz00, line 106: Do we want to drop the "GPU" from these now?

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerInnerTiled.cpp
```diff
@@ -0,0 +1,47 @@
...
+namespace {
+
+struct LLVMCPULowerInnerTiledPass final
+    : impl::LLVMCPULowerInnerTiledPassBase<LLVMCPULowerInnerTiledPass> {
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    RewritePatternSet patterns(context);
+    // Expects vector-semantics `iree_codegen.inner_tiled` produced by the
+    // generic vectorization pass via the `VectorizableOpInterface` external
+    // model in `compiler/src/iree/compiler/Codegen/Interfaces/`.
```
hanhanW, line 30: This should be part of pass description, and we should not mention GenericVectorization or how the op is vectorized at all IMO. The pass itself does not care that. What it expects is vectorized inner tiled. Let's drop the vectorization mechanism and move it into description?

btw, how about naming the pass as "LLVMCPULowerVectorizedInnerTiledPass`?

krzysz00 (approved):
Two notes, but this looks mostly like stampable code movement, so approved after notes

compiler/src/iree/compiler/Codegen/LLVMCPU/test/lower_inner_tiled.mlir
```diff
@@ -0,0 +1,32 @@
+// RUN: iree-opt %s --pass-pipeline="builtin.module(func.func(iree-llvmcpu-lower-inner-tiled))" --split-input-file | FileCheck %s
+
+// `iree-llvmcpu-lower-inner-tiled` runs two pattern sets in sequence on a
+// vector-semantics `iree_codegen.inner_tiled` (the kind produced by
+// GenericVectorization via the VectorizableOpInterface external model):
```
hanhanW, line 5: Same here, this is leaking layering assumption which is redundant to this pass.

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
```diff
@@ -368,6 +368,14 @@ void addMmt4dTilingExpertPassPipeline(
     }
   }
 
+  // Lower any iree_codegen.inner_tiled ops produced by the data-tiled MMA
+  // path. GenericVectorization above lifts them to vector-semantics via the
+  // VectorizableOpInterface external model; this pass then drops their unit
+  // iter dims and emits the per-intrinsic ops via
+  // `DataTiledMMAAttr::buildUnderlyingOperations`. If the dispatch has no
+  // inner_tiled ops the pass is a no-op.
```
hanhanW, line 376: I think we can say that this is similar to ukernel path, but it is more powerful because of fusion. I don't have fresh memory for the lowering patterns. IIRC, it should also unroll the inner tiled op which has the same behavior as your ukernels?

Open question: do we do the lowering after bufferization? The inner tiled captures the memref semantics well, which seems to have less burden in the pipeline. My first thinking for the pass is about "high level vector lowering", because it is lowering a "virtual vector" to IRs that is closer to LLVM dialect (low-level dialect). Thus, it belongs to [the LLVMCPUVirtualVectorLoweringPass in buildLLVMCPUVectorLoweringPipeline](https://github.com/iree-org/iree/blob/11926304c49d708d6a953568aa2444cff63b2d0d/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L109-L111) IMO.

Do we just add the patterns to the pass? And we can add tests to [virtual_vector_lowering.mlir](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/test/virtual_vector_lowering.mlir) in this context.

hanhanW (changes requested):

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerInnerTiled.cpp
bjacob: @krzysz00 : If I understand correctly what you mean, this is exactly what the next PR does: https://github.com/iree-org/iree/pull/24353.

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
bjacob: I think you are right @hanhanW .  Let me close this PR and redo this like you suggest.

PR #24357, opened 2026-05-04
[Codegen] NFC: Lift InnerTiledOp unroll pattern to Codegen.

bjacob:
Follow-up to the previous NFC, moving the third (and last) `iree_codegen.inner_tiled` rewrite pattern out of
`Codegen/Dialect/GPU/Transforms/`:

  - `UnrollInnerTiledPattern` (unrolls along iter dimensions until each intrinsic invocation has a unit iter shape)

The pattern itself is dialect-neutral; only the convenience no-args populate function applied a default config (matmul-like LHS-reuse traversal order, unit native shape) that happens to be a good fit for any architecture, including CPU.

Renames the populate functions:

  - `IREE::GPU::populateIREEGPUVectorUnrollPatterns` (with-options form) -> `IREE::Codegen::populateUnrollInnerTiledPatterns`
  - `IREE::GPU::populateIREEGPUVectorUnrollPatterns` (no-args form) -> `IREE::Codegen::populateUnrollInnerTiledPatterns`

Renames the GPU-specific helper `gpuMatmulLikeUnrollOrder` to `matmulLikeUnrollOrder` (its content is dialect-neutral). The `kUnrollAccDistribute` / `kUnrollAccReassemble` HoistableConversionOp tag constants move with the pattern.

Updates the in-tree callers in `IREEGPUExtensions.cpp` and `UnrollToIntrinsics.cpp`, and adjusts BUILD.bazel/CMakeLists.txt deps.

Strict NFC; no functional or behavioral change.

Progress towards #24323

PR #24358, opened 2026-05-04
[Codegen][CPU] Lower data-tiled inner_tiled in VirtualVectorLoweringPass.

bjacob:
`iree_codegen.inner_tiled` is a high-level vector op (a multi-dim
contraction whose shape is described by its kind), and lowering it to
per-intrinsic ops is "high-level vector -> lower-level dialect", which
is exactly what `LLVMCPUVirtualVectorLoweringPass` already does for
`vector.contract`, `vector.multi_reduction`, `vector.transfer_*`, and
the IREE VectorExt ops. So land the CPU `inner_tiled` lowering in
that pass rather than in a separate pass.

Three pattern populates from `IREE::Codegen` (introduced in PR https://github.com/iree-org/iree/pull/24351 and #24357) join the target-independent
block of `LLVMCPUVirtualVectorLoweringPass`:

  - `IREE::Codegen::populateUnrollInnerTiledPatterns` unrolls any
    non-unit iter dim (typically a residual reduction tile) to unit.
  - `IREE::Codegen::populateDropInnerTiledUnitDimsPatterns` drops the
    (now-unit) iter domain so the lowered op has empty iter bounds,
    which is the precondition of...
  - `IREE::Codegen::populateLowerInnerTiledPatterns` replaces the
    iter-free vector-semantics `inner_tiled` with the per-intrinsic
    ops emitted by its kind's `buildUnderlyingOperations` — for
    `IREE::CPU::DataTiledMMAAttr` those are `llvm.call_intrinsic` ops.

These patterns only match `iree_codegen.inner_tiled`, so they are
orthogonal to the existing patterns in this pass; the
`DropVectorUnitDimsPass` that runs immediately before this one only
touches `vector.*` ops and similarly does not interact.

After the greedy pattern application, run
`IREE::Util::eliminateHoistableConversions` to cancel the inverse-pair
`util.hoistable_conversion` ops the inner_tiled lowering wraps around
the per-intrinsic ACC distribute/reassemble. GPU's `LowerIREEGPUOps`
pass calls this for the same reason; the introducing pattern set is
the right place to clean up after itself.

`LLVMCPUVirtualVectorLoweringPass` runs post-bufferization. That is
fine for `iree_codegen.inner_tiled`: by then the op is already in
vector semantics (lifted by `GenericVectorizationPass` via the
`InnerTiledOpVectorizationModel` `VectorizableOpInterface` external
model), and bufferization does not touch vectors.

Adds a focused
`compiler/src/iree/compiler/Codegen/LLVMCPU/test/virtual_vector_lowering_inner_tiled.mlir`
lit test that runs through `iree-llvmcpu-virtual-vector-lowering` and
asserts a single `llvm.fma.v16f32` call with no surviving
`util.hoistable_conversion` on the AVX-512 1x16x1 f32 single-tile case.

Updates a stale forward reference in `KernelDispatch.cpp` left over
from https://github.com/iree-org/iree/pull/24328 — the comment described the unit-iter-dim drop and per-
intrinsic lower patterns as firing inside a pass named
`LLVMCPULowerInnerTiledPass`, which never landed; they now fire
inside `LLVMCPUVirtualVectorLoweringPass`.

Progress towards https://github.com/iree-org/iree/issues/24323


compiler/src/iree/compiler/Codegen/LLVMCPU/test/virtual_vector_lowering_inner_tiled.mlir
hanhanW, line 1: Can we add the test to `virtual_vector_lowering.mlir` directly? I don't see a reason for having a separate file atm.

bjacob: Good point. Pushed https://github.com/iree-org/iree/pull/24358/commits/a84005280c63fa96a380a1c6388180525e4a66fd.

hanhanW (approved):

PR #24359, opened 2026-05-04
[AMDGPU] Roll-up of AMDGPU HAL improvements for CDNA support

benvanik:
Mostly just hardening for different PM4 modes and minor performance optimizations found during benchmarking. Besides the profiling quirk the entire HAL worked first try on an MI300X. Neat.


AWoloszyn:
awesome

PR #24366, opened 2026-05-05
[DispatchCreation][CPU] Skip data tiling when no MMAIntrinsic matches.

bjacob:
Skip data tiling on a per-matmul basis when the matmul's element types have no candidate MMA intrinsic on at least one of the targets the dispatch will run on (and that target has `enable_inner_tiled`). The matmul then stays as plain `linalg.matmul` and goes through standard codegen.

Why this CPU-specific check lives in (otherwise target-agnostic) DispatchCreation: the set of MMAIntrinsics is spotty on CPU, so the programs we actually compile routinely use element-type combinations that aren't covered by any intrinsic on the target's CPU features (e.g. bf16 inputs on a x86 CPU without `+avx512bf16`). On GPU the picture is different: programs are usually authored against the target GPU's known MMA capabilities, so the data-tiling design can build on the assumption that in intrinsic exists.

Implementation:
  * `CPUEncodingExternalModels` now publicly exposes `matchMmaIntrinsics(ctx, config, elementTypes)`, which returns the x86 MMAIntrinsic enum values whose required ISA features match `config` and whose tile element types match `elementTypes`. The cost model and the dispatch-creation opt-out share this single enumeration; an empty result is the opt-out signal.
  * `AnnotateDataTilingHintsPass` collects, per function, the `enable_inner_tiled = true` executable-target configurations from the parent module's `util.global ... = #hal.device.target<...>` ops, then for each candidate matmul calls `matchMmaIntrinsics` on every such config. If any returns empty, we skip annotating the matmul for data tiling. (We walk the globals directly rather than going through `IREE::HAL::DeviceAnalysis` / `ExecutableTargetAttr ::lookup`, both of which expect post-dispatch-formation IR.)

Verified by hand:
  * f32/f32 on AVX2 → data-tiled, `llvm.fma.v8f32` from `MMA_X86_AVX2_FMA_1x8x1_F32_F32` (no regression).
  * bf16/f32 on AVX-512-BF16 → data-tiled, `llvm.x86.avx512bf16.dpbf16ps.512` (no regression).
  * bf16/f32 on AVX2 → no `set_encoding`, no inner_tiled, plain `linalg.matmul` carried through the dispatch (new behavior).

Progress towards #24323


hanhanW (changes requested):
Puts a blocker for now. We will discuss details tomorrow: https://discord.com/channels/689900678990135345/1456679894812590326/1501318094549811562

hanhanW:
Folks, let me know if you want to be in the discussion. Otherwise, I'll only ping @bjacob when I come online tomorrow.

bjacob:
Closing as we found a better path forward: adding a generic scalar-fallback pseudo "intrinsic", thus allowing inner_tiled to cover all cases (though not necessarily performantly).

Issue #24342, opened 2026-05-01
[Codegen][RISCV] `xla_reverse_after_slice` miscompiles on RISC-V vector when its `linalg.generic` is collapsed to 1-D


yzhang93:
## Failing test

`tests/e2e/stablehlo_ops/reverse.mlir::xla_reverse_after_slice`:

```mlir
// Regression test for https://github.com/iree-org/iree/issues/23637.
func.func @xla_reverse_after_slice() {
  %t1 = util.unfoldable_constant dense<[
      [[1.0, 2.0], [3.0, 4.0], [5.0, 6.0], [7.0, 8.0]],
      [[9.0, 10.0], [11.0, 12.0], [13.0, 14.0], [15.0, 16.0]]
  ]> : tensor<2x4x2xf32>
  %sliced = "stablehlo.slice"(%t1) {
      start_indices = array<i64: 0, 1, 0>,
      limit_indices = array<i64: 2, 3, 2>,
      strides = array<i64: 1, 1, 1>
  } : (tensor<2x4x2xf32>) -> tensor<2x2x2xf32>
  %reversed = "stablehlo.reverse"(%sliced) {dimensions = array<i64: 1>} : (tensor<2x2x2xf32>) -> tensor<2x2x2xf32>
  check.expect_almost_eq_const(
      %reversed,
      dense<[
          [[5.0, 6.0], [3.0, 4.0]],
          [[13.0, 14.0], [11.0, 12.0]]
      ]> : tensor<2x2x2xf32>
  ) : tensor<2x2x2xf32>
  return
}
```

## Error

When the `stablehlo.slice` + `stablehlo.reverse` combine into a single 3-D `linalg.generic` and then get collapsed to a 1-D iteration space`, RISC-V vector codegen produces wrong values for the second batch (`d0 = 1`):

```
[ RUN      ] module.xla_reverse_after_slice
iree/runtime/src/iree/modules/check/module.cc:628: Failure
Failed
Expected near equality of these values. Contents does not match to tolerance parameters atol=0.0001, rtol=0. The first failure occurs at index 4 as the lhs value 9 differs from the rhs value 13 by 4.000e+00.
  lhs:
    2x2x2xf32=[[5 6][3 4]][[9 10][7 8]]
  rhs:
    2x2x2xf32=[[5 6][3 4]][[13 14][11 12]]

[  FAILED  ] module.xla_reverse_after_slice, where GetParam() = 1 (64 ms)
```

[...]

PR #24370, opened 2026-05-05
[test][nfc] Add regression tests about strided vector.gather back.

hanhanW:
It was failing on RISC-V because it explicitly exlcudes the lowering (which contains the proper fix) in https://github.com/iree-org/iree/commit/25dcacfcbcd4ec91260920d7296ad558aa03ef84

It seems like it is converted to `llvm.*.gather` ops and relies on LLVM backend to pick the corresponding instruction. This basically bypasses the vector level optimization in MLIR which is not fixable in IREE.

Related issue: https://github.com/iree-org/iree/issues/24342

### DIFF
--- a/tests/e2e/regression/xla_reverse_after_slice.mlir
+++ b/tests/e2e/regression/xla_reverse_after_slice.mlir
@@ -1,0 +1,26 @@
+// Regression test for https://github.com/iree-org/iree/issues/24342.
+//
+// It fails on risc-v only because it does not apply the upstream pattern that
+// recovers the logical indices of a `vector.gather` op.
+// See https://github.com/iree-org/iree/commit/25dcacfcbcd4ec91260920d7296ad558aa03ef84
+func.func @xla_reverse_after_slice() {
+  %t1 = util.unfoldable_constant dense<[
+      [[1.0, 2.0], [3.0, 4.0], [5.0, 6.0], [7.0, 8.0]],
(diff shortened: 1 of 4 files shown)


hanhanW:
cc @mshockwave @rednoah91 I won't be able to fix risc-v issue, so I'm going to enable the tests for CPU backends except risc-v.

tests/e2e/regression/xla_reverse_after_slice.mlir
```diff
@@ -0,0 +1,26 @@
+// Regression test for https://github.com/iree-org/iree/issues/24342.
+//
+// It fails on risc-v only because it does not apply the upstream pattern that
+// recovers the logical indices of a `vector.gather` op.
+// See https://github.com/iree-org/iree/commit/25dcacfcbcd4ec91260920d7296ad558aa03ef84
+func.func @xla_reverse_after_slice() {
```
yzhang93, line 6: Do you need to remove the duplicate test here https://github.com/iree-org/iree/blob/43b8ecd23de039381bec9db362e3c5a1b7f085f5/tests/e2e/stablehlo_ops/reverse.mlir#L24-L49?

hanhanW: I missed this part, thanks!

yzhang93 (approved):

PR #24372, opened 2026-05-06
[Stream] Fix build break: PropertyRef → OpaqueProperties after MLIR revert

josephbak:
MLIR commit 7fce7631a098 renamed `OpaqueProperties` to `PropertyRef` across the codebase. MLIR commit b31e232c29ae then reverted that rename back to `OpaqueProperties`, updating dialect files within llvm-project, but `StreamOps.cpp` in IREE was not reverted accordingly.

This fixes the two affected `inferReturnTypes` signatures in `AsyncParameterGatherOp` and `AsyncParameterScatterOp`.

---

**Note:** CI may show a pre-existing build failure in `compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp`
(`vector::createWriteOrMaskedWrite` missing) that is unrelated to this PR and reproducible on `upstream/main` without this change.

Assisted-by: Claude (Anthropic)

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/IR/StreamOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/IR/StreamOps.cpp
@@ -3863,7 +3863,7 @@ void AsyncParameterWriteOp::getAsyncAccessRanges(
 
 LogicalResult AsyncParameterGatherOp::inferReturnTypes(
     MLIRContext *context, std::optional<Location> location, ValueRange operands,
-    DictionaryAttr attributes, PropertyRef properties, RegionRange regions,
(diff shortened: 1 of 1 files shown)


hanhanW:
IREE has its own fork. I don't follow why we need this now. I think it will happen in bumping LLVM integrate. cc @lialan

josephbak:
Thanks for the sanity check, @hanhanW. You are completely right. I ran a submodule sync and verified that both `StreamOps.cpp` and the upstream interface consistently use `PropertyRef`. The build failure was entirely on my end due to a local sync issue. Closing this now, and apologies for the noise!

PR #24254, opened 2026-04-24
[Codegen][DMA] Skip untuned XOR swizzle fallback for BF16

Yu-Zhewen:
The untuned swizzle heuristic (`getXorShuffleParamsForUntunedChipset`) can produce XOR shuffle parameters that are incompatible with AMDGPULowerCoalescedDMAToGatherLDS  checks, causing DMA to fall back to vector.transfer_read/write.

This PR is in the same spirit of #24208, making DMA decisions explicit at heuristic selection time rather than relying on silent late-pipeline fallbacks. 

Assisted-by: Cursor (Claude)

PR #24242, opened 2026-04-24
[Codegen][DMA] Fix OOB clamping for subview sources

Yu-Zhewen:
When the DMA source is a subview of a larger `fat_raw_buffer`, an OOB index on dim 0 of the subview can still land in valid parent memory, bypassing the hardware's linear OOB clamping. 

Fix: check all dimensions (including dim 0) when the source is a subview, and use the root buffer's dim-0 size as the OOB sentinel to guarantee the linearized offset exceeds the total buffer size.

Assisted-by: Cursor (Claude)

PR #24373, opened 2026-05-06
Reapply "[Codegen] Enable DMA by default for F16/BF16 Gemm on gfx950 (#24117)" (#24235)

Yu-Zhewen:
This reverts commit 75ffbc37144de79cc9428f97827251b2242b230f.

The previously reported numerical issues have now been resolved through the following changes:
- https://github.com/iree-org/iree/pull/24241
- https://github.com/iree-org/iree/pull/24242
- https://github.com/iree-org/iree/pull/24254


lialan (approved):
We will track closely about the performance, as we turn on DMA for BF16 by default.

jerryyin (approved):

PR #24376, opened 2026-05-06
[INTEGRATION] Bump llvm to 3ed76d05a78d

lialan:
Also drops `test_reversesequence_time` from `cpu_llvm_sync` xfails, which is fixed by https://github.com/llvm/llvm-project/pull/195359


hanhanW (approved):
Thanks!

yzhang93:
Do you need to revert [b09174b41e7e — [AMDGPU] Enable runtime loop unrolling (#194924)](https://github.com/llvm/llvm-project/pull/194924) to get the perf back?

lialan:
> Do you need to revert [b09174b41e7e — [AMDGPU] Enable runtime loop unrolling (#194924)](https://github.com/llvm/llvm-project/pull/194924) to get the perf back?

Here we restore the unrolling threshold to before: https://github.com/iree-org/iree/pull/24379

But it is temporary. We should decide whether architecturally we control the unrolling or not.

PR #24377, opened 2026-05-06
[VectorExt] Change `to_layout` `shared_memory_conversion`

sommerlukas:
Change how `shared_memory_conversion` is represented on the `to_layout` operation. So far, a unit attribute was used to simply turn shared memory conversion on/off. Now, the `shared_memory_conversion` takes an attribute, which allows us to attach for example the GPU promotion type (e.g. `use_global_load_dma`) to the operation. 

The motivation is to use this additional information in analyses such as https://github.com/iree-org/iree/pull/24227 as part of https://github.com/iree-org/iree/issues/23782. 

The representation is `AnyAttr` to (a) avoid a dependency from `VectorExt` on `IREEGPU` and (b) to allow other pipelines to use the attribute differently.

The PR also adds some verification to the `LoweringConfigAttr` to avoid having to repeatedly check that the list of promoted operands and promotion types has the same length.

Assisted-by: Codex

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPULoweringConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPULoweringConfigUtils.cpp
@@ -123,7 +123,7 @@ setPromotedOperandsList(MLIRContext *context,
   std::optional<SmallVector<int64_t>> currPromotedOperandsList =
       getPromotedOperandList(currAttr);
   if (currPromotedOperandsList &&
-      currPromotedOperandsList.value() == operands) {
+      currPromotedOperandsList.value() == operands && !promotionTypes) {
     return currAttr;
   }
 
@@ -133,6 +133,8 @@ setPromotedOperandsList(MLIRContext *context,
 
   if (promotionTypes) {
     attributes.set(kPromotionTypesName, b.getArrayAttr(promotionTypes.value()));
+  } else {
+    attributes.erase(kPromotionTypesName);
   }
   return IREE::GPU::LoweringConfigAttr::get(context,
                                             attributes.getDictionary(context));
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPULoweringConfigUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPULoweringConfigUtils.cpp
@@ -100,6 +100,43 @@ getPromotionTypesList(LoweringConfigAttr config) {
   return array.getValue();
 }
 
+LogicalResult
+verifyPromotedOperandsList(function_ref<InFlightDiagnostic()> emitError,
+                           DictionaryAttr attributes) {
+  Attribute promotionTypesAttr = attributes.get(kPromotionTypesName);
+  if (!promotionTypesAttr) {
+    return success();
+  }
+  auto promotionTypes = dyn_cast<ArrayAttr>(promotionTypesAttr);
+  if (!promotionTypes) {
+    return emitError() << "promotion_types must be an array";
+  }
+
+  auto promotedOperandsAttr = attributes.getAs<ArrayAttr>(kPromoteOperandsName);
+  if (!promotedOperandsAttr) {
+    return emitError()
+           << "promotion_types requires promote_operands to be specified";
+  }
+
+  std::optional<SmallVector<int64_t>> promotedOperands =
+      getIntegerVector(promotedOperandsAttr);
+  if (!promotedOperands) {
+    return emitError() << "promote_operands must be an array of integers";
+  }
+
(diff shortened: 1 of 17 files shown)


Groverkss (approved):
VectorDistribute changes LGTM, Maybe @Max191 or @qedawkins can review the IREEGPU changes.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -2831,6 +2835,39 @@ SmallVector<int64_t> LoweringConfigAttr::getWorkgroupTileSizes() const {
   return getTileSizes(getAttributes(), GPU::TilingLevel::Workgroup);
 }
 
+LogicalResult
+LoweringConfigAttr::verify(function_ref<InFlightDiagnostic()> emitError,
+                           DictionaryAttr attributes) {
+  auto promotionTypes = attributes.getAs<ArrayAttr>(kPromotionTypesName);
```
keshavvinayak01: This catches invalid elements inside an array, but not a malformed `promotion_types` attr itself. `getAs<ArrayAttr>` returns null both when the attr is absent and when it is present with the wrong type, so `promotion_types = "invalid"` currently
verifies and is later treated as if no promotion types were specified.

Could this use `attributes.get(kPromotionTypesName)` first, return success only if it is absent, and emit an error if the present attr is not an `ArrayAttr`?

Suggested test:
```mlir
func.func @lowering_config_invalid_promotion_types_attr() attributes {
// expected-error @+1 {{promotion_types must be an array}}
lowering_config = #iree_gpu.lowering_config<{promote_operands = [0], promotion_types = "invalid"}>} {
return
}

Comment (sommerlukas) [reply compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp]: Thanks for catching that, fixed now.

File: compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp [hunk 2]
```diff
@@ -50,6 +51,9 @@
 
 namespace mlir::iree_compiler::IREE::GPU {
 
+constexpr StringLiteral kPromoteOperandsName = "promote_operands";
+constexpr StringLiteral kPromotionTypesName = "promotion_types";
```
Max191: nit: Maybe move the verification logic for promote_operands/promotion_types into `GPULoweringConfigUtils.cpp` as a utility function and call it from the verifier so we don't have kPromoteOperandsName and kPromotionTypesName in both places.

Alternatively, you could handle it like we handle tiling levels, where we have an enum and a function to get the string name: https://github.com/iree-org/iree/blob/1c14508943bc6c16f531aa3bb2955f133897876c/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp#L2795

But we don't really need to go that far IMO. The utility function in GPULoweringConfigUtils would also be fine.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPULoweringConfigUtils.cpp
```diff
@@ -123,7 +123,7 @@ setPromotedOperandsList(MLIRContext *context,
   std::optional<SmallVector<int64_t>> currPromotedOperandsList =
       getPromotedOperandList(currAttr);
   if (currPromotedOperandsList &&
-      currPromotedOperandsList.value() == operands) {
+      currPromotedOperandsList.value() == operands && !promotionTypes) {
     return currAttr;
```
Max191, line 176: Shouldn't this be checking that the current promotion types matches the passed promotion types? If the current promotion types is not empty, but the new promotion types is not passed, then shouldn't this clear the promotion types?

The point here seems to be to ensure the promotion types and promoted operands stay in sync, so it seems like a good condition to have.

keshavvinayak01 (approved):
LLVMGPU changes look good.

Max191 (approved):
GPU dialect stuff LGTM, just a couple of nits.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/GPULoweringConfigUtils.cpp
sommerlukas: Yeah, good catch, I've implemented that now.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
sommerlukas: I've implemented the solution with the utility function now.

PR #24379, opened 2026-05-06
[ROCM] Workaround LLVM #194924 partial-unroll regression

lialan:
llvm/llvm-project#194924 added:

```
UP.Runtime = true;
UP.PartialThreshold = UP.Threshold / 4;   // = 75 for AMDGPU
```

to `AMDGPUTTIImpl::getUnrollingPreferences`. The second line gates compile-time partial unrolling, suppressing the partial-unroll-by-2 of small constant-trip-count reduction loops.

As a temporary measurement this patch:
Restore the prior un-overridden LLVM default of 150.

### DIFF
--- a/compiler/plugins/target/ROCM/ROCMTarget.cpp
+++ b/compiler/plugins/target/ROCM/ROCMTarget.cpp
@@ -645,6 +645,25 @@ public:
                              llvm::TargetMachine &targetMachine,
                              bool slpVectorization,
                              std::string &outPassesString) {
+    // Workaround for upstream LLVM PR llvm/llvm-project#194924 (commit
+    // b09174b41e7e, "[AMDGPU] Enable runtime loop unrolling") which set
+    //   UP.PartialThreshold = UP.Threshold / 4   (75 for AMDGPU)
+    // in AMDGPUTTIImpl::getUnrollingPreferences, regressing partial unrolling
+    // of small constant-trip-count reduction loops. Restore the prior
+    // un-overridden LLVM default (150) by bumping the cl::opt occurrence
+    // count, which makes LoopUnrollPass apply the override (see
+    // LoopUnrollPass.cpp gatherUnrollingPreferences). Skipped when the user
+    // already set --unroll-partial-threshold so an explicit override still
+    // wins. Safe to remove once the upstream regression is fixed.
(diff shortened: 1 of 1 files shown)


compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -645,6 +645,25 @@ class ROCMTargetBackend final : public TargetBackend {
                              llvm::TargetMachine &targetMachine,
                              bool slpVectorization,
                              std::string &outPassesString) {
+    // Workaround for upstream LLVM PR llvm/llvm-project#194924 (commit
+    // b09174b41e7e, "[AMDGPU] Enable runtime loop unrolling") which set
```
yzhang93, line 649: Nit: can you put a link instead?

compiler/plugins/target/ROCM/ROCMTarget.cpp
```diff
@@ -645,6 +645,25 @@ class ROCMTargetBackend final : public TargetBackend {
...
+    //   UP.PartialThreshold = UP.Threshold / 4   (75 for AMDGPU)
+    // in AMDGPUTTIImpl::getUnrollingPreferences, regressing partial unrolling
+    // of small constant-trip-count reduction loops. Restore the prior
+    // un-overridden LLVM default (150) by bumping the cl::opt occurrence
+    // count, which makes LoopUnrollPass apply the override (see
+    // LoopUnrollPass.cpp gatherUnrollingPreferences). Skipped when the user
+    // already set --unroll-partial-threshold so an explicit override still
+    // wins. Safe to remove once the upstream regression is fixed.
```
yzhang93, line 657: Have you filed an issue upstream?

yzhang93 (approved):
The temporary fix should be fine. Have you filed an issue upstream for a proper fix?

compiler/plugins/target/ROCM/ROCMTarget.cpp
lialan: Yeah here: https://github.com/llvm/llvm-project/issues/196372

compiler/plugins/target/ROCM/ROCMTarget.cpp
lialan: The #194924 should be good to track down the issue.

Issue #24012, opened 2026-04-04
[CPU] Elementwise (add) performance study on AMD Zen4 7975WX


hanhanW:
# IREE CPU Element-wise Performance Report

Machine: 32-core AMD Zen4 7975WX, AVX-512, 32 threads.
Baseline: PyTorch 2.9.0+cpu with MKL.
IREE flags: `--iree-hal-target-backends=llvm-cpu --iree-llvmcpu-target-cpu=host --iree-opt-level=O2` for compilation and `--device_allocator=caching` for benchmark

```mlir
// element-wise add: static 1000000
func.func @add(%lhs: tensor<1000000xf32>, %rhs: tensor<1000000xf32>) -> tensor<1000000xf32> {
  %init = tensor.empty() : tensor<1000000xf32>
  %result = linalg.add ins(%lhs, %rhs : tensor<1000000xf32>, tensor<1000000xf32>)
                       outs(%init : tensor<1000000xf32>) -> tensor<1000000xf32>
  return %result : tensor<1000000xf32>
}
```

Note: `--device_allocator=caching` is important in benchmarking. Otherwise, it adds overheads because it will initialize output buffers with zeros, which takes `~167 us` to finish.

## Current state: add 1M f32

| Workers/Threads | IREE static | MKL | Gap |
|-----------------|-------------|-----|-----|
| 1 | 0.232 ms | 0.096 ms | 2.4x |
| 4 | 0.129 ms | 0.029 ms | 4.4x |
| 8 | 0.098 ms | 0.020 ms | 4.9x |
| 16 | **0.094 ms** | 0.013 ms | **7.2x** |
| 32 | 0.128 ms | 0.011 ms | 11.6x |

All IREE numbers use `--device_allocator=caching`. Best: **0.094 ms @16w (7.2x off MKL)**.

Best worker count is 16, not 32 — at 32 workers, contention on the task
scheduler increases overhead for 244 workgroups.

## Tracy profiling: one iteration breakdown (88.9 us @16w, with caching)

From `add_1M_static_caching_16w.tracy`, steady-state `BenchmarkIteration` at t≈3s.

| Phase | Start (us) | Duration (us) | % | Description |
|-------|-----------|---------------|---|-------------|
| VM setup + alloca | 4.3 | 6.9 | 8% | vm_invoke, device_queue_alloca (1.7 us cached), queue_execute |
| Scheduling delay | 11.2 | 16.0 | 18% | fence_await starts → first shard at t=27.2 |
[...]

PR #24383, opened 2026-05-06
[CPU] Improve tiling config for elementwise ops with dynamic shapes.

hanhanW:
The default distribution tile sizes were 64s for dynamic shape, which also requires a fixup in elementwise ops. Otherwise, the runtime overheads dominates the performance. We identified the issue for static cases before, and the recent study also points out that it is needed for dynamic shapes. See https://github.com/iree-org/iree/issues/24012 for more details.

Note that it is not common in full model workload because they are mostly fused with producers, so it was not on our radar until today.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
@@ -2657,29 +2657,38 @@ static LogicalResult setElementwiseGenericOpRootConfig(
   // TODO(dcaballe): The logic below is disconnected from the main tile size
   // selection logic in getMaxTileSize. We should either port it there or remove
   // it.
-  // Adjust the number of workload per workgroup to at least 4096. This
-  // prevents the runtime overheads domiating the execution time. The number is
-  // derived from experimients. We should be able to make it related to target.
+  // Adjust the number of workload per workgroup to at least 4096. This prevents
+  // the runtime overheads from dominating the execution time. The number is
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -2657,29 +2657,38 @@ static LogicalResult setElementwiseGenericOpRootConfig(
...
   constexpr int64_t kMinimumWorkload = 4096;
   SmallVector<int64_t> shape = genericOp.getStaticLoopRanges();
   int64_t numWorkload = 1;
   for (auto [index, size] : llvm::enumerate(shape)) {
-    if (ShapedType::isDynamic(size)) {
-      numWorkload = ShapedType::kDynamic;
-      break;
+    int64_t tile = distTileSizes[index];
+    if (tile != 0) {
+      numWorkload *= tile;
```
bjacob, line 2676: just checking - It's not possible to have the kDynamic value here, right?

bjacob (approved):

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
hanhanW: This is tiling sizes, so you're right that it is not possible to have `kDynamic` value here. In the tiling context, `0` means no-tiling. Positive numbers are the tiling sizes. Negative numbers are invalid.

PR #24387, opened 2026-05-07
[VectorExt] Add generic lowering of iree_vector_ext.arg_compare

### LINKED ISSUE #24365: [VectorExt]The basic Lowring of iree_vector_ext.arg_compare is missing
**Current Status**: Vectorization would produce `iree_vector_ext.arg_compare`, which has no lowering in the TileAndFuse pipeline. We just let it go to LinalgExtToLoops, which can handle it automatically along the TileAndFuse pipeline. Lowering for iree_vector_ext.arg_compare is the trick along VectorDistribute, about some layout, distribution, and dpp + ballot

A basic plumbing is missing there, like add a generic lowering of iree_vector_ext.arg_compare into vector.* ops which will unblock CPU or any GPU backends that do not have AMDGPU ballot or DPP instruction

bangtianliu:
This PR adds `LowerArgCompareToVectorPass`, which lowers iree_vector_ext.arg_compare to scf.for loops with scalar `vector.extract/vector.insert` and cloned comparator regions. This provides a generic lowering path for backends that don't use the AMDGPU-specific ballot/DPP/shuffle lowering in the VectorDistribute pipeline.  

The pass is integrated into the TileAndFuse GPU pipeline after vectorization, replacing the previous workaround that skipped arg_compare vectorization entirely (`vectorizeArgCompare=false`). The redundant `vectorizeArgCompare`option (always true now) is removed. 

Closes #24365         

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/BUILD.bazel
@@ -58,6 +58,7 @@ iree_compiler_cc_library(
         "@llvm-project//mlir:BufferizationDialect",
         "@llvm-project//mlir:BufferizationInterfaces",
         "@llvm-project//mlir:BufferizationTransforms",
+        "@llvm-project//mlir:DialectUtils",
         "@llvm-project//mlir:FuncDialect",
         "@llvm-project//mlir:FunctionInterfaces",
         "@llvm-project//mlir:IR",
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/BUILD.bazel
@@ -37,6 +37,7 @@ iree_compiler_cc_library(
     srcs = [
         "BufferizationInterfaces.cpp",
         "DistributionPatterns.cpp",
+        "LowerArgCompareToVector.cpp",
         "LowerTransferGatherScatterOps.cpp",
         "LowerTransferGatherScatterToVector.cpp",
         "Passes.cpp",
@@ -61,6 +62,7 @@ iree_compiler_cc_library(
         "@llvm-project//mlir:FunctionInterfaces",
         "@llvm-project//mlir:IR",
         "@llvm-project//mlir:Pass",
+        "@llvm-project//mlir:SCFDialect",
         "@llvm-project//mlir:Support",
         "@llvm-project//mlir:TensorDialect",
         "@llvm-project//mlir:TransformUtils",
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/CMakeLists.txt
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/CMakeLists.txt
@@ -31,6 +31,7 @@ iree_cc_library(
   SRCS
     "BufferizationInterfaces.cpp"
     "DistributionPatterns.cpp"
+    "LowerArgCompareToVector.cpp"
     "LowerTransferGatherScatterOps.cpp"
     "LowerTransferGatherScatterToVector.cpp"
     "Passes.cpp"
@@ -45,6 +46,7 @@ iree_cc_library(
     MLIRFunctionInterfaces
     MLIRIR
     MLIRPass
+    MLIRSCFDialect
     MLIRSupport
     MLIRTensorDialect
     MLIRTransformUtils
--- a/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
@@ -1,0 +1,230 @@
+// Copyright 2026 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree/compiler/Codegen/Dialect/VectorExt/IR/VectorExtOps.h"
+#include "iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.h"
+#include "iree/compiler/Codegen/Dialect/VectorExt/Transforms/Transforms.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/IRMapping.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+using namespace mlir::iree_compiler::IREE::VectorExt;
+
+namespace mlir::iree_compiler::IREE::VectorExt {
+
+#define GEN_PASS_DEF_LOWERARGCOMPARETOVECTORPASS
+#include "iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.h.inc"
+
+namespace {
+
+static Value cloneComparatorRegion(RewriterBase &rewriter, Region &region,
+                                   Value lhs, Value rhs) {
+  Block &block = region.front();
+  IRMapping mapper;
+  mapper.map(block.getArgument(0), lhs);
+  mapper.map(block.getArgument(1), rhs);
+  for (Operation &op : block.without_terminator()) {
+    Operation *clonedOp = rewriter.clone(op, mapper);
+    for (const auto &[origResult, clonedResult] :
+         llvm::zip_equal(op.getResults(), clonedOp->getResults())) {
+      mapper.map(origResult, clonedResult);
+    }
+  }
+  auto yieldOp = cast<YieldOp>(block.getTerminator());
+  return mapper.lookup(yieldOp.getValues()[0]);
+}
+
+static SmallVector<int64_t> delinearizeIndex(int64_t linearIdx,
+                                             ArrayRef<int64_t> shape) {
+  SmallVector<int64_t> indices;
+  for (int64_t i = static_cast<int64_t>(shape.size()) - 1; i >= 0; --i) {
+    indices.push_back(linearIdx % shape[i]);
+    linearIdx /= shape[i];
+  }
+  std::reverse(indices.begin(), indices.end());
+  return indices;
+}
+
+static SmallVector<int64_t> buildMoveToLastPerm(int64_t rank, int64_t dim) {
+  SmallVector<int64_t> perm;
+  for (int64_t i = 0; i < rank; ++i) {
+    if (i != dim) {
+      perm.push_back(i);
+    }
+  }
+  perm.push_back(dim);
+  return perm;
+}
+
+struct LowerArgCompareToVector final : OpRewritePattern<ArgCompareOp> {
+  using Base::Base;
+
+  LogicalResult matchAndRewrite(ArgCompareOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    VectorType inputType = op.getInputValueType();
+
+    if (inputType.isScalable()) {
+      return rewriter.notifyMatchFailure(op, "scalable vectors not supported");
+    }
+
+    int64_t rank = inputType.getRank();
+    int64_t reductionDim = op.getDimension();
+    int64_t reductionSize = inputType.getShape()[reductionDim];
+
+    Value inputValue = op.getInputValue();
+    Value inputIndex = op.getInputIndex();
+    Value initValue = op.getInitValue();
+    Value initIndex = op.getInitIndex();
+    Value indexBase = op.getIndexBase();
+    Type indexElemTy = op.getInitIndexElementType();
+
+    bool needsTranspose = (reductionDim != rank - 1) && (rank > 1);
+    if (needsTranspose) {
+      SmallVector<int64_t> perm = buildMoveToLastPerm(rank, reductionDim);
+      inputValue = vector::TransposeOp::create(rewriter, loc, inputValue, perm);
+      if (inputIndex) {
+        inputIndex =
+            vector::TransposeOp::create(rewriter, loc, inputIndex, perm);
+      }
+      reductionDim = rank - 1;
+    }
+
+    VectorType transposedType = cast<VectorType>(inputValue.getType());
+    ArrayRef<int64_t> shape = transposedType.getShape();
+
+    SmallVector<int64_t> parallelShape(shape.begin(), shape.end() - 1);
+    int64_t numParallelElems = ShapedType::getNumElements(parallelShape);
+
+    VectorType resultValueType = op.getInitValueType();
+    VectorType resultIndexType = op.getInitIndexType();
+
+    Value resultValue = initValue;
+    Value resultIndex = initIndex;
+
+    Value c0 = arith::ConstantIndexOp::create(rewriter, loc, 0);
+    Value c1 = arith::ConstantIndexOp::create(rewriter, loc, 1);
+    Value cRedSize =
+        arith::ConstantIndexOp::create(rewriter, loc, reductionSize);
+
+    for (int64_t p = 0; p < numParallelElems; ++p) {
+      SmallVector<int64_t> parallelIndices =
+          delinearizeIndex(p, parallelShape);
+
+      Value accVal, accIdx;
+      if (resultValueType.getRank() == 0) {
+        accVal = vector::ExtractOp::create(rewriter, loc, initValue,
+                                           ArrayRef<int64_t>{});
+        accIdx = vector::ExtractOp::create(rewriter, loc, initIndex,
+                                           ArrayRef<int64_t>{});
+      } else {
+        accVal = vector::ExtractOp::create(rewriter, loc, initValue,
+                                           parallelIndices);
+        accIdx = vector::ExtractOp::create(rewriter, loc, initIndex,
+                                           parallelIndices);
+      }
+
+      Value inputSlice, indexSlice;
+      if (rank == 1) {
+        inputSlice = inputValue;
+        if (inputIndex) {
+          indexSlice = inputIndex;
+        }
+      } else {
+        inputSlice = vector::ExtractOp::create(rewriter, loc, inputValue,
+                                               parallelIndices);
+        if (inputIndex) {
+          indexSlice = vector::ExtractOp::create(rewriter, loc, inputIndex,
+                                                 parallelIndices);
+        }
+      }
+
+      auto forOp = scf::ForOp::create(rewriter, loc, c0, cRedSize, c1,
+                                      ValueRange{accVal, accIdx});
+      {
+        OpBuilder::InsertionGuard guard(rewriter);
+        rewriter.setInsertionPointToStart(forOp.getBody());
+        Value iv = forOp.getInductionVar();
+        Value bestVal = forOp.getRegionIterArg(0);
+        Value bestIdx = forOp.getRegionIterArg(1);
+
+        Value candidate =
+            vector::ExtractOp::create(rewriter, loc, inputSlice, iv);
+
+        Value cmpResult =
+            cloneComparatorRegion(rewriter, op.getRegion(), candidate, bestVal);
+
+        Value newVal = arith::SelectOp::create(rewriter, loc, cmpResult,
+                                               candidate, bestVal);
+
+        Value candidateIdx;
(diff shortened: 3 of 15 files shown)


compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,229 @@
...
+static SmallVector<int64_t> delinearizeIndex(int64_t linearIdx,
+                                             ArrayRef<int64_t> shape) {
+  SmallVector<int64_t> indices;
+  for (int64_t i = static_cast<int64_t>(shape.size()) - 1; i >= 0; --i) {
+    indices.push_back(linearIdx % shape[i]);
+    linearIdx /= shape[i];
+  }
+  std::reverse(indices.begin(), indices.end());
+  return indices;
+}
```
amd-eochoalo: I think this function already exists in MLIR:

```
auto indices = delinearize(linearIdx, computeStrides(shape));
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,229 @@
...
+  Block &block = region.front();
+  IRMapping mapper;
+  mapper.map(block.getArgument(0), lhs);
+  mapper.map(block.getArgument(1), rhs);
+  for (Operation &op : block.without_terminator()) {
+    Operation *clonedOp = rewriter.clone(op, mapper);
+    for (const auto &[origResult, clonedResult] :
+         llvm::zip_equal(op.getResults(), clonedOp->getResults())) {
+      mapper.map(origResult, clonedResult);
+    }
```
amd-eochoalo, line 41:
```suggestion
```

I think this is done by default when cloning?

Also, would cloneRegionBefore + inlineBlockBefore be suitable here instead?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,229 @@
...
+      if (resultValueType.getRank() == 0) {
+        accVal = vector::ExtractOp::create(rewriter, loc, initValue,
+                                           ArrayRef<int64_t>{});
+        accIdx = vector::ExtractOp::create(rewriter, loc, initIndex,
+                                           ArrayRef<int64_t>{});
+      } else {
+        accVal = vector::ExtractOp::create(rewriter, loc, initValue,
+                                           parallelIndices);
+        accIdx = vector::ExtractOp::create(rewriter, loc, initIndex,
+                                           parallelIndices);
```
amd-eochoalo: There's a github warning here that these branches are identical. I know these are not syntactically identical, just asking if parallelIndices is {} when rank == 0.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,229 @@
...
+      if (resultValueType.getRank() == 0) {
+        resultValue = vector::BroadcastOp::create(rewriter, loc,
+                                                  resultValueType, reducedVal);
+        resultIndex = vector::BroadcastOp::create(rewriter, loc,
+                                                  resultIndexType, reducedIdx);
+      } else {
+        resultValue = vector::InsertOp::create(rewriter, loc, reducedVal,
+                                               resultValue, parallelIndices);
+        resultIndex = vector::InsertOp::create(rewriter, loc, reducedIdx,
+                                               resultIndex, parallelIndices);
```
amd-eochoalo, line 193: Same here, but maybe github is wrong 😅

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
bangtianliu: parallelIndices is {} when rank == 0. You are right here, and the branch can be avoided.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
bangtianliu: `Operation::clone(mapper)` maps the Operation pointer via `mapper.map(this, newOp)`, but not the result Values, So the explicit result mapping is needed (`mapper.lookup(yieldOp.getValues()[0])`)  to resolve.

`cloneRegionBefore/inlineBlockBefore` would work for a single inline, but we need to clone the comparator body on every loop iteration (inside scf.for) with different operand mappings (candidate, bestVal). The IRMapping approach handles that. We reuse it from here: https://github.com/iree-org/iree/blob/1c14508943bc6c16f531aa3bb2955f133897876c/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp#L1053-L1068

I can send one NFC cleanup PR to remove the duplicate.

hanhanW (commented):
I'll leave the details review to @amd-eochoalo and @sommerlukas, thanks for clearing the TODOs! I can help set up CPU pipeline to use this path for e2e test coverage in a follow up.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,212 @@
...
+  }
+};
+
+struct LowerArgCompareToVectorPass final
+    : impl::LowerArgCompareToVectorPassBase<LowerArgCompareToVectorPass> {
+  void runOnOperation() override {
+    MLIRContext *ctx = &getContext();
+    RewritePatternSet patterns(ctx);
+    patterns.add<LowerArgCompareToVector>(ctx);
+    if (failed(applyPatternsGreedily(getOperation(), std::move(patterns)))) {
```
kuhar: you should be able to use the walk pattern rewrite driver here since you don't need to iterate until fixpoint

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,212 @@
...
+  mapper.map(block.getArgument(1), rhs);
+  for (Operation &op : block.without_terminator()) {
+    Operation *clonedOp = rewriter.clone(op, mapper);
+    for (const auto &[origResult, clonedResult] :
+         llvm::zip_equal(op.getResults(), clonedOp->getResults())) {
+      mapper.map(origResult, clonedResult);
+    }
+  }
+  auto yieldOp = cast<YieldOp>(block.getTerminator());
+  return mapper.lookup(yieldOp.getValues()[0]);
```
kuhar: I think this assumes the yielded compare value was produced inside the comparator body. The op verifier allows the region to capture an external i1, and with a body that directly yields that value this pass currently hits the IRMapping assertion instead of rejecting or lowering it. Either use lookupOrDefault for the yielded value, or tighten the verifier to forbid captures before this lowerer runs. We should have a test case that covers this too.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_arg_compare.mlir
```diff
@@ -0,0 +1,261 @@
...
+    inits(%init_val, %init_idx : vector<f32>, vector<i32>) {
+  ^bb0(%a: f32, %b: f32):
+    %cmp = arith.cmpf ogt, %a, %b : f32
+    iree_vector_ext.yield %cmp : i1
+  } -> vector<f32>, vector<i32>
+  return %result#0, %result#1 : vector<f32>, vector<i32>
+}
+// CHECK-LABEL: func.func @argmax_scalable_rejection
+// CHECK:         iree_vector_ext.arg_compare
+// CHECK-NOT:     scf.for
```
kuhar, line 337: Test coverage gaps worth filling:
1. Multi-op comparator body (all current tests use a single arith.cmpf). cloneComparatorRegion clones operations one-by-one and remaps results in a loop; a body with at least two chained ops (e.g., arith.subf followed by arith.cmpf) would exercise that result-mapping loop.
2. index_base combined with multi-D parallel dims (e.g., reducing vector<4x128xf32> over dim 1 with a non-zero base). Currently, only the 1-D + index_base case is covered, so we never verify that arith.addi(base, iv) lands inside each per-parallel-element scf.for.
3. Negative test for non-pure ops in the comparator (op verifier should catch, but a CHECK-NOT-style integration test guards against accidental relaxation). (This is more of nit).

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,212 @@
...
+    int64_t reductionDim = op.getDimension();
+    int64_t reductionSize = inputType.getShape()[reductionDim];
+
+    Value inputValue = op.getInputValue();
+    Value inputIndex = op.getInputIndex();
+    Value initValue = op.getInitValue();
+    Value initIndex = op.getInitIndex();
+    Value indexBase = op.getIndexBase();
+    Type indexElemTy = op.getInitIndexElementType();
+
```
kuhar, line 80: This emits one `scf.for` per parallel result. Fine for one-output-per-thread TileAndFuse, but `vector<64x128xf32>` would create 64 reduction loops. Please document/enforce that narrow-shape assumption or use one loop over the parallel slice.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -574,18 +572,11 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
...
-  // to TileAndFuse (small reductions, f64) need the scalar-loop lowering from
-  // `LinalgExtToLoops` instead. This gate is intentionally narrow.
-  // TODO(#24365): Remove this option once a generic vector lowering for
-  // `iree_vector_ext.arg_compare` exists.
   addGPUVectorizationPasses(funcPassManager, /*vectorizeCopies=*/false,
                             /*enableMasking=*/true,
                             /*foldIdentitySlices=*/true,
-                            /*decomposeMasks=*/false,
-                            /*vectorizeArgCompare=*/false);
+                            /*decomposeMasks=*/false);
```
kuhar, line 578: Please keep a short rationale for this pass placement. The old comment explained why TileAndFuse could not rely on the VectorDistribute nested-layout lowering; now this pass runs immediately after vectorization with no note. A one-sentence comment saying this handles TileAndFuse `arg_compare` after generic vectorization would be enough.

kuhar (commented):
A few comments

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    VectorType transposedType = cast<VectorType>(inputValue.getType());
+    ArrayRef<int64_t> shape = transposedType.getShape();
+
+    // Compile-time unroll over parallel dimensions. This assumes narrow
+    // parallel shapes (typically 1 element per dim after TileAndFuse thread
+    // tiling).
+    SmallVector<int64_t> parallelShape(shape.begin(), shape.end() - 1);
+    int64_t numParallelElems = ShapedType::getNumElements(parallelShape);
```
kuhar: nit: No need to materialize a brand new vector just to drop the last element
```suggestion
    ArrayRef<int64_t> parallelShape = shape.drop_back();
    int64_t numParallelElems = ShapedType::getNumElements(parallelShape);
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/LowerArgCompareToVector.cpp
```diff
@@ -0,0 +1,213 @@
...
+      }
+      reductionDim = rank - 1;
+    }
+
+    VectorType transposedType = cast<VectorType>(inputValue.getType());
+    ArrayRef<int64_t> shape = transposedType.getShape();
+
+    // Compile-time unroll over parallel dimensions. This assumes narrow
+    // parallel shapes (typically 1 element per dim after TileAndFuse thread
+    // tiling).
```
kuhar: Can you extend this comment with a TODO to revisit large parallel dim handling later on?

kuhar (approved):
LGTM % minor comments but please wait for another approval from someone more familiar with this part of the codebase

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/CMakeLists.txt
```diff
@@ -45,6 +46,7 @@ iree_cc_library(
     MLIRFunctionInterfaces
     MLIRIR
     MLIRPass
+    MLIRSCFDialect
```
sommerlukas, line 49: In Bazel, we also add `DialectUtils`, do we also need that here?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/BUILD.bazel
```diff
@@ -57,10 +58,12 @@ iree_compiler_cc_library(
         "@llvm-project//mlir:BufferizationDialect",
         "@llvm-project//mlir:BufferizationInterfaces",
         "@llvm-project//mlir:BufferizationTransforms",
+        "@llvm-project//mlir:DialectUtils",
```
sommerlukas, line 61: What do we need this include for, `IndexingUtils`?

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.td
```diff
@@ -28,4 +28,15 @@ def LowerTransferGatherScatterToVectorPass :
...
+def LowerArgCompareToVectorPass :
+    Pass<"iree-vector-ext-lower-arg-compare-to-vector", "func::FuncOp"> {
+  let summary = "Lower iree_vector_ext.arg_compare to vector/SCF ops";
+  let dependentDialects = [
+    "::mlir::arith::ArithDialect",
+    "::mlir::scf::SCFDialect",
+    "::mlir::vector::VectorDialect",
+    "::mlir::iree_compiler::IREE::VectorExt::IREEVectorExtDialect"
+  ];
+}
```
sommerlukas, line 42: I'm fine with having this pass, just wondering if we should have a more generic pass that does multiple of these lowerings if we add more in the future instead of having individual passes for each of them. But nothing to address as part of this PR, just a note.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -575,18 +572,13 @@ void addGPUTileAndFusePassPipeline(OpPassManager &funcPassManager,
...
+  // Lower vectorized arg_compare to scf.for + vector ops before bufferization.
+  // VectorDistribute has its own arg_compare lowering via DistributeArgCompare.
```
sommerlukas: Does this mean we are not expecting any `arg_compare` in the IR when we reach the vectorization passes from VectorDistribute? Might be worth adding this to the comment if that's the case.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_arg_compare.mlir
```diff
@@ -0,0 +1,332 @@
...
+  ^bb0(%a: f32, %b: f32):
+    %cmp = arith.cmpf ogt, %a, %b : f32
+    iree_vector_ext.yield %cmp : i1
+  } -> vector<3x5xf32>, vector<3x5xi32>
+  return %result#0, %result#1 : vector<3x5xf32>, vector<3x5xi32>
+}
+// CHECK-LABEL: func.func @argmax_3d_reduce_middle_dim
+// CHECK-DAG:     arith.constant 4 : index
+// CHECK-DAG:     vector.transpose %{{.*}}, [0, 2, 1] : vector<3x4x5xf32> to vector<3x5x4xf32>
+// CHECK-COUNT-15: scf.for {{.*}} to %{{.*}} step
```
sommerlukas, line 142: Based on this `CHECK-COUNT`, I think your TODO comment in the lowering makes sense, it might be worth emitting a loop into IR instead of fully unrolling things. What kind of shapes per thread do you typically expect in workloads?

This can be a follow-up though.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_arg_compare.mlir
```diff
@@ -0,0 +1,332 @@
...
+    %cmp = arith.cmpf ogt, %a, %b : f32
+    iree_vector_ext.yield %cmp : i1
+  } -> vector<4xf32>, vector<4xi32>
+  return %result#0, %result#1 : vector<4xf32>, vector<4xi32>
+}
+// CHECK-LABEL: func.func @argmax_explicit_index_last_dim
+// CHECK-NOT:     vector.transpose
+// CHECK-NOT:     arith.index_cast
+// CHECK-COUNT-4: scf.for
+// CHECK-NOT:     scf.for
```
sommerlukas, line 167: In other places, we don't have a `CHECK-NOT` after the count. If we want to make sure the count is exactly what we expect, might be worth adding in other places, too.

sommerlukas (approved):
Looks good, just a few minor nits.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/CMakeLists.txt
bangtianliu: Yes

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/BUILD.bazel
bangtianliu: We need it for `delinearize`and `computeStrides` declared in `mlir/Dialect/Utils/IndexingUtils.h` used in the parallel-dim unroll loop in LowerArgCompareToVector.cpp:

```cpp                                                                                                                                                               
SmallVector<int64_t> parallelIndices =                                                                                                                       
   delinearize(p, computeStrides(parallelShape));
```

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/Passes.td
bangtianliu: Good point, will add a todo here.

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/CMakeLists.txt
bangtianliu: Actually, no, the bazel_to_cmake pre-commit hook may strip this again on the next run since bazel_to_cmake_targets.py maps DialectUtils to empty

https://github.com/iree-org/iree/blob/1d87665ee3f32be1d918c8da57069e31aad2e3cb/build_tools/bazel_to_cmake/bazel_to_cmake_targets.py#L68

compiler/src/iree/compiler/Codegen/Dialect/VectorExt/Transforms/test/lower_arg_compare.mlir
bangtianliu: Ok, I double checked the KernelConfig part in setArgCompareConfig used for TileAndFuse

https://github.com/iree-org/iree/blob/1d87665ee3f32be1d918c8da57069e31aad2e3cb/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp#L2119-L2139

The above code sets threadTileSizes = 1 on every parallel loop and 0 on the reduction dim, which means each thread owns one parallel coordinate and walks the entire reduction itself.

So the unroll factor is 1 along the TileAndFuse pipeline.

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
bangtianliu: Actually arg_compare is still in the IR at the vectorization stage of the VectorDistribute pipeline. 

Vectorization is about converting `iree_linalg_ext.arg_compare` to `iree_vector_ext.arg_compare`, and then it exists in both TileAndFuse and VectorDistribute pipelines. 

This lowering pass only needs to run in the TileAndFuse pipeline, while the VectorDistribute pipeline lowers the op `iree_vector_ext.arg_compare` through  DistributeArgCompare using gpu.subgroup_reduce and gpu.ballot. 

I have expanded the comment.

PR #24389, opened 2026-05-07
[Codegen][CPU] Add a type-polymorphic generic-scalar MMA fallback.

bjacob:
Adds two new `MMAIntrinsic` values, `MMA_GENERIC_SCALAR_1x1x1_REG8` and `_REG16`, that the data-tiling cost model picks when no element-type- specific intrinsic on the target supports the matmul's (LHS, RHS, ACC) types. This intentionally breaks the "an MMAIntrinsic enum value pins down a specific element-type triple" invariant in exchange for not having to add one enum value per supported triple. Element types live on new `DataTiledMMAAttr.{lhs,rhs,acc}_type` parameters, populated by the cost model only when the chosen intrinsic is one of the polymorphic variants.

The cost model picks `_REG16` on 64-bit ISAs (x86_64, AArch64, RISC-V) and `_REG8` on 32-bit ISAs. The number is a register-budget for the unroll heuristic — one element of any width occupies one register, but the architectural register file the lowering ends up in (GPR or SIMD- scalar lane) is up to LLVM. The budget is encoded in the low byte of the enum value, so `chooseUnrolling` can read it back.

Since the intrinsic is 1×1×1, the operand tiles after `intrinsics_m` / `intrinsics_n` / `intrinsics_k` are simple row-major (M, K) / (N, K) / (M, N) — `linalg.mmt4d`-shaped. `DataTiledMMAAttr::buildUnderlyingOperations` therefore short-circuits the swizzle/distribute pipeline for these intrinsics and emits a single `vector.contract` directly, with `arith.extf` / `arith.extsi` widening narrow LHS/RHS to ACC's element type. For sub-byte LHS/RHS types `chooseUnrolling` also picks the smallest power-of-two `intrinsics_k` such that K*lhsBits and K*rhsBits are byte-aligned (e.g. K=2 for i4/f4, K=4 for f6, K=8 for i1).

Progress towards #24323

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
@@ -597,3 +597,78 @@ func.func @matmul_lowering_i8i4i32_aarch64_i8mm(
 //  CHECK-SAME:       ins(%[[LHS]], %[[RHS]] :
 //  CHECK-SAME:       outs(%[[OUTS]] :
 //       CHECK:   return %[[MMT4D]]
+
+// -----
+
+// AArch64 has no element-type-specific MMA intrinsic for sub-byte float
+// types, so the data-tiling cost model falls back to the type-polymorphic
+// generic scalar. AArch64 is 64-bit, so `pickGenericScalarMMAForTarget`
+// picks `_REG16`. The chosen `DataTiledMMAAttr` carries
+// `lhs_type = f4E2M1FN`, `rhs_type = f4E2M1FN`, `acc_type = f32` — the
+// enum value alone doesn't determine element types.
+//
+// FP4 register pressure: K=2 is the smallest power-of-two K that makes
+// K*4 a multiple of 8 (for byte-addressable packed groups). Inside the
+// 16-register budget the cost model lands on a square (im=2, in=2, ik=2)
+// tile (2*2 + 2*2 + 2*2 = 12 registers), which wins on arithmetic
+// intensity. Whether the lowering can actually densely pack f4E2M1FN
+// values into bytes today is orthogonal — the framework only insists
+// that the packed K-group be byte-aligned.
+
+#map_g_fp4 = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map_g_fp4_1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map_g_fp4_2 = affine_map<(d0, d1, d2) -> (d0, d1)>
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
hanhanW, line 1: Check the last test case (i.e., `matmul_lowering_i8i4i32_aarch64_i8mm`). If we only check `iree_codegen.inner_tiled`, we don't need set_encodings ops and unset_encoding ops in tests.

(I spent some time to support encodings on func arguments, which simplifies our tests.)

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -352,10 +352,24 @@ getRowMajorTilesMNKShape(MMAIntrinsic intrinsic) {
...
+      return Tuple{1, 1, 1};
+    }
     return {};
   }
 }
 
+bool isGenericScalar(MMAIntrinsic intr) {
+  // Architecture-agnostic generic-scalar fallback values live in the 0xF0**
+  // range; see IREECPUEnums.td.
+  return (static_cast<uint32_t>(intr) & 0xFF00) == 0xF000;
```
hanhanW: I wonder if we can have some `enum/constexpr` for these magic numbers.

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -726,6 +766,58 @@ static Value createCpuMmaIntrinsicCall(OpBuilder &builder, Location loc,
   }
 }
 
+/// Lowers a `DataTiledMMAAttr` whose intrinsic is one of the
+/// `MMA_GENERIC_SCALAR_1x1x1_REG*` cases directly to a single
+/// `vector.contract`. Since the intrinsic's tile shape is 1×1×1, the
+/// operand tiles after applying `intrinsics_m` / `intrinsics_n` /
+/// `intrinsics_k` are simple row-major (M, K)/(N, K)/(M, N) matmul
+/// tiles — no swizzle-based distribution is needed. Mirrors how
+/// `linalg.mmt4d` natively lowers to `vector.contract`.
```
hanhanW: I'm not sure if we want to bake "mmt4d concept" in the comment. It is sort of leaking other transformation to the local method, while how to lower mmt4d is not a single well-known path IMO.  The "natively" here is a misuse IMO. It is more like the current IREE behavior when ukernel is disabled. I think we should decouple the comment from mmt4d, and only use the intrinsic definition. Maybe we just drop the last sentence?

btw, do we have an assertion that ensures attr is one of `MMA_GENERIC_SCALAR_*`? Do we use `isGenericScalar` in the assertion?

hanhanW (commented):
Just some nits, overall looks good to me.

hanhanW (approved):

PR #24214, opened 2026-04-22
[DispatchCreation] Prevent fill->scatter cloning

IanWood1:
Ops like `iree_linalg_ext.scatter` and `tensor.insert_slice` only write a subset of their `outs`/`dest`; their iteration space does not cover the init (the indexing map is null, or the operand has no mapping into the affine iteration space). A producer that materializes a full-tensor write has no shared iteration space with such a consumer and therefore cannot be tile-fused with it.

Block these clones in `getCloneableOps`:
- `isUnfusableInit` identifies init operands that producers cannot be fused through: `tensor.insert_slice.dest` and any `LinalgFusionOpInterface` DPS init whose indexing map is null (e.g. scatter's `original`).
- `collectUnfusableInitSources` uses `getBackwardSlice` to walk from each such init through cloneable chains, recording only producers that need to be materialized. Views/reshapes (`tensor.extract_slice`, `tensor.empty`, `tensor.expand_shape`, `tensor.collapse_shape`) are traversed but not blocked, so the llama scatter-on-extracted-slice pattern still fuses.

Fixes iree-org/iree#24071.

PR #24391, opened 2026-05-07
[Flow][Dispatch Creation] Fix dispatch tied result crash

### LINKED ISSUE #24385: Crash in DispatchWorkgroupsOp::getResultAccess when all results are tied to the same input operand
### What happened?

Summary of finding by Claude reducing a crash in a proprietary model after bisect found that the crash was introduced in commit d097bad562:
                                                                                                                                                                                                                                                     
  Component: Dialect/Flow/IR, Dialect/Util/IR/ClosureOpUtils                                                                                                                                                                                         
  Severity: Crash (SIGSEGV)                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                     
  ---                                                                                                                                                                                                                                                
  Summary                                                   
         
  iree-compile crashes with a segfault during Flow canonicalization when a `flow.dispatch.workgroups` op has multiple results all tied to the same single input operand. The crash is an unsigned integer underflow in `getResultAccess` followed by an out-of-bounds block argument access.

  ---                                                                                                                                                                                                                                                
  Minimal reproducer                                        

```
// Minimal reproducer for crash in DispatchWorkgroupsOp::getResultAccess.
//
// When N results are all tied to the same single input operand, the dispatch
// body only has 1 block arg (no separate output-only args). getResultAccess:
//
//   startIndex = getWorkgroupBody().getNumArguments() - getNumResults()
//              = 1 - N   (underflows for N > 1)
//
// The subsequent getArgument(startIndex + resultIndex) is then out of bounds.

#map = affine_map<(d0, d1) -> (d0, d1)>

module {
  util.func public @crash_repro(%input: tensor<4x4xf32>)
      -> (tensor<4x4xf32>, tensor<4x4xf32>) {
    // 1 input, 2 results both tied to %input.
    // Body has 1 block arg; getResultAccess computes startIndex = 1 - 2 → underflow.
    %r:2 = flow.dispatch.workgroups(%input) :
        (tensor<4x4xf32>) -> (%input, %input) =
        (%arg0: !iree_tensor_ext.dispatch.tensor<readwrite:tensor<4x4xf32>>) {
      %t = iree_tensor_ext.dispatch.tensor.load %arg0,
          offsets = [0, 0], sizes = [4, 4], strides = [1, 1]
          : !iree_tensor_ext.dispatch.tensor<readwrite:tensor<4x4xf32>> -> tensor<4x4xf32>
      iree_tensor_ext.dispatch.tensor.store %t, %arg0,
          offsets = [0, 0], sizes = [4, 4], strides = [1, 1]
          : tensor<4x4xf32> -> !iree_tensor_ext.dispatch.tensor<readwrite:tensor<4x4xf32>>
      flow.return
    } count() -> (index, index, index) {
      flow.return %c1, %c1, %c1 : index, index, index
    }
    util.return %r#0, %r#1 : tensor<4x4xf32>, tensor<4x4xf32>
  }
}
```
                                                                                                                                                                                                                                                     
  Command:                                                  
`iree-opt --pass-pipeline="builtin.module(util.func(iree-flow-canonicalize))" \
    --mlir-disable-threading -o /dev/null minimal_repro.mlir`
                                                                                                                                                                                                                                                     
  Result:                                                                                                                                                                                                                                  
  Segmentation fault (SIGSEGV).                                                                                                                                                                                                                      
  Stack trace (truncated to relevant frames)                
```
   #4 mlir::Block::getArgument(unsigned int)
      mlir/include/mlir/IR/Block.h:139                                                                                                                                                                                                               
   #5 mlir::iree_compiler::IREE::Flow::DispatchWorkgroupsOp::getResultAccess(unsigned int)                                                                                                                                                           
      compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp:1067                                                                                                                                                                                    
   #6 ClosureOpInterfaceInterfaceTraits::Model<DispatchWorkgroupsOp>::getResultAccess(...)                                                                                                                                                           
      iree-compiler/compiler/src/iree/compiler/Dialect/Util/IR/UtilOpInterfaces.h.inc:2247                                                                                                                                                           
   #7 mlir::iree_compiler::IREE::Util::optimizeClosureLikeOp(...)                                                                                                                                                                                    
      compiler/src/iree/compiler/Dialect/Util/IR/ClosureOpUtils.cpp:344                                                                                                                                                                              
```                                                                                                                                                                                                                                                  
  ---                                                                                                                                                                                                                                                
  Root cause                                                
                                                                                                                                                                                                                                                     
  There are two bugs — an underlying crash bug, and a separate commit that began triggering it in practice.
                                                                                                                                                                                                                                                     
  Bug 1 — Unsigned underflow in `getResultAccess` (FlowOps.cpp:1065)                                                                                                                                                                                   

```                                                                                                                                                                                                                                     
  // FlowOps.cpp                                                                                                                                                                                                                                     
  unsigned startIndex = getWorkgroupBody().getNumArguments() - getNumResults();
  BlockArgument arg = getWorkgroupBody().front().getArgument(startIndex + resultIndex);
```                                                                                                                                                                                                                                                  
  A `flow.dispatch.workgroups` body has one block argument per input plus one per untied result. When all N results are tied to the same single input, the body has only 1 block argument regardless of N. The subtraction 1 - N underflows to a large positive integer for N > 1, and the subsequent `getArgument` call is out of bounds.
                                                                                                                                                                                                                                                     
  The invariant assumed by `getResultAccess` — that `numBodyArgs >= numResults` — does not hold when multiple results share a tied input.                                                                                                                
   
  Fix: Guard the subtraction:
```
  unsigned numBodyArgs = getWorkgroupBody().getNumArguments();
  unsigned numResults = getNumResults();                                                                                                                                                                                                             
  assert(numBodyArgs >= numResults &&                       
         "body args must cover all results; tied-operand sharing violated this");
  unsigned startIndex = numBodyArgs - numResults;                                
```
  Or, better, compute the result's block argument index via the tied-operand mapping rather than assuming a fixed offset.                                                                                                                            
                                                                                                                         
  Bug 2 — Missing ownership check in `hasUnfusableUseInDispatch` (RegionOpUtils.cpp, introduced in commit d097bad562)
                                                                                                                                                                                                                                                     
  This commit changed `hasUnfusableUseInDispatch` to check all uses of a value regardless of whether those uses are inside the dispatch being analysed. The previous code explicitly skipped uses whose parent `DispatchWorkgroupsOp` was not the op under analysis: 

```                                                                                                                                                                                                                                                     
  // Before (correct): skip uses outside the current dispatch
  Operation *owner = user->getParentOfType<IREE::Flow::DispatchWorkgroupsOp>();
  if (owner != dispatchOp) { continue; }                                                                                                                                                                                                             
```
                                                                                                                                                                                                                                                  
  After the commit this guard was removed. As a result, a value that appears as the dest operand of a `tensor.insert_slice` outside a dispatch is now incorrectly treated as unfusable inside that dispatch. This prevents op cloning into the dispatch, leaving duplicate references to the same input value in the operand list.
                                                                                                                                                                                                                                                     
  These duplicates are later deduplicated by `optimizeClosureLikeOp` in ClosureOpUtils.cpp. After deduplication, a dispatch that originally had, e.g., 6 inputs with 4 results all tied to the same value ends up with 3 inputs and 4 results still tied to that one remaining input — which triggers Bug 1.
                                                                                                                                                                                                                                                     
  Fix: Restore the ownership check so only uses that are children of the dispatch under analysis are considered.                                                                                                                                     
   
  ---                                                                                                                                                                                                                                                
  Relationship between the two bugs                         
                                   
  Bug 1 is the crash. It is a latent defect that existed before d097bad562 — any canonicalization that produces a dispatch with `numResults > numBodyArgs` would hit it. Bug 2 is what made Bug 1 reachable in practice by creating the unusual operand configuration that `optimizeClosureLikeOp` then reduces into the crashing state.

`iree-compile --iree-opt-level=O0 --iree-hal-target-device=local --iree-hal-local-target-device-backends=llvm-cpu --iree-llvmcpu-target-cpu=x86-64-v3 -o /dev/null minimal_repro.mlir`
or
`iree-opt --pass-pipeline="builtin.module(util.func(iree-flow-canonicalize))" --mlir-disable-threading -o /dev/null minimal_repro.mlir`

Compiler

### Version information

[d097bad562](https://github.com/iree-org/iree/commit/d097bad562a7e941d5522c58c82f4e0a26907378)

### Additional context

_No response_

IanWood1:
Fixes a crash in `DispatchWorkgroupsOp::getResultAccess()` when multiple `flow.dispatch.workgroups` results are tied to the same input operand. Also restores the dispatch ownership check in `hasUnfusableUseInDispatch()` that was dropped in #24214, so `tensor.insert_slice` dest uses outside the dispatch being analyzed do not block producer cloning.

Fixes https://github.com/iree-org/iree/issues/24385

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp
+++ b/compiler/src/iree/compiler/Dialect/Flow/IR/FlowOps.cpp
@@ -1053,9 +1053,7 @@ DispatchWorkgroupsOp::getOperandAccess(unsigned operandIndex) {
 
 IREE::Util::ValueAccess
 DispatchWorkgroupsOp::getResultAccess(unsigned resultIndex) {
-  unsigned startIndex = getWorkgroupBody().getNumArguments() - getNumResults();
-  BlockArgument arg =
-      getWorkgroupBody().front().getArgument(startIndex + resultIndex);
+  BlockArgument arg = getOutputBlockArgument(resultIndex);
   if (auto tensorType =
           dyn_cast<IREE::TensorExt::DispatchTensorType>(arg.getType())) {
     auto tensorAccess = refineTensorAccess(arg, tensorType);
--- a/compiler/src/iree/compiler/Dialect/Flow/IR/test/dispatch_folding.mlir
+++ b/compiler/src/iree/compiler/Dialect/Flow/IR/test/dispatch_folding.mlir
@@ -126,6 +126,28 @@ util.func public @keep_used_read_write_result(%arg0 : tensor<9xi32>, %arg1 : ten
 
 // -----
 
+// CHECK-LABEL: util.func public @multiple_results_tied_to_same_input
+util.func public @multiple_results_tied_to_same_input(%arg0: tensor<4x4xf32>)
+    -> (tensor<4x4xf32>, tensor<4x4xf32>) {
+  %c1 = arith.constant 1 : index
+  // CHECK: flow.dispatch.workgroups[%c1](%arg0) : (tensor<4x4xf32>) -> (%arg0, %arg0) =
+  // CHECK-NEXT: (%{{.+}}: !iree_tensor_ext.dispatch.tensor<readwrite:tensor<4x4xf32>>)
+  %0:2 = flow.dispatch.workgroups[%c1, %c1, %c1](%arg0) :
+      (tensor<4x4xf32>) -> (%arg0, %arg0) =
+      (%arg0_capture: !iree_tensor_ext.dispatch.tensor<readwrite:tensor<4x4xf32>>) {
+    %1 = iree_tensor_ext.dispatch.tensor.load %arg0_capture,
+        offsets = [0, 0], sizes = [4, 4], strides = [1, 1]
+        : !iree_tensor_ext.dispatch.tensor<readwrite:tensor<4x4xf32>> -> tensor<4x4xf32>
+    iree_tensor_ext.dispatch.tensor.store %1, %arg0_capture,
+        offsets = [0, 0], sizes = [4, 4], strides = [1, 1]
+        : tensor<4x4xf32> -> !iree_tensor_ext.dispatch.tensor<readwrite:tensor<4x4xf32>>
+    flow.return
+  }
+  util.return %0#0, %0#1 : tensor<4x4xf32>, tensor<4x4xf32>
+}
+
+// -----
+
 // CHECK-LABEL: util.func public @drop_unused_dispatch_region_result
 util.func public @drop_unused_dispatch_region_result(
     %arg0: tensor<?x?xf32>, %arg1: tensor<5x10xf32>, %arg2: tensor<7x11xf32>)
--- a/compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
+++ b/compiler/src/iree/compiler/Dialect/Flow/Transforms/RegionOpUtils.cpp
@@ -927,6 +927,16 @@ hasUnfusableUseInDispatch(Value v, Operation *dispatchOp,
       return true;
     }
 
+    Operation *ownerWorkgroupsOp =
+        user->getParentOfType<IREE::Flow::DispatchWorkgroupsOp>();
+    Operation *ownerRegionOp =
+        user->getParentOfType<IREE::Flow::DispatchRegionOp>();
+    Operation *owner = ownerWorkgroupsOp ? ownerWorkgroupsOp : ownerRegionOp;
+
+    if (owner != dispatchOp) {
+      continue;
+    }
+
     if (auto insertSlice = dyn_cast<tensor::InsertSliceOp>(user);
         insertSlice && use.get() == insertSlice.getDest()) {
       return true;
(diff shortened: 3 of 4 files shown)


MaheshRavishankar (commented):
The way I understand this I think it should be illegal to have same value being tied to multiple results.

IanWood1:
> The way I understand this I think it should be illegal to have same value being tied to multiple results.

To clarify, do you think that 

(1) it should be illegal for a single operand to map to multiple tied results (`flow.dispatch.workgroups(%val0) : (tensor<T>) -> (%val0, %val0)`) or
(2) for the same SSA value to be mapped to tied results (`flow.dispatch.workgroups(%val0, %val0) : (tensor<T>, tensor<T>) -> (%val0, %val0)`)? 


To me, (2) seems okay. It looks like `optimizeClosureLikeOp` is deduplicating the inputs which transforms (2) into (1).

MaheshRavishankar:
> > The way I understand this I think it should be illegal to have same value being tied to multiple results.
> 
> To clarify, do you think that
> 
> (1) it should be illegal for a single operand to map to multiple tied results (`flow.dispatch.workgroups(%val0) : (tensor<T>) -> (%val0, %val0)`) or (2) for the same SSA value to be mapped to tied results (`flow.dispatch.workgroups(%val0, %val0) : (tensor<T>, tensor<T>) -> (%val0, %val0)`)?
> 
> To me, (2) seems okay. It looks like `optimizeClosureLikeOp` is deduplicating the inputs which transforms (2) into (1).

I think it shouldnt do that if the two result values yielded from within the region are not the same.

MaheshRavishankar (approved):
Looks like the issue was something totally different.

PR #24392, opened 2026-05-07
[Compiler] Use Repeated<T> for repeated value ranges. NFC.

kuhar:
Replace repeated `SmallVector<Value/Type>(n, x)` temporaries with `llvm::Repeated<Value/Type>` where the result is only consumed as a `ValueRange` or `TypeRange`. This avoids needless memory allocations.

This mirrors the MLIR-side refactoring in llvm/llvm-project#188846.

Assisted-by: codex


amd-eochoalo (approved):

sommerlukas (approved):
Looks good, thanks for the cleanup!

bangtianliu (approved):

Issue #24378, opened 2026-05-06
[CPU] Improve test coverage for onnx_ops test suite


hanhanW:
There are 283 compilation failures and 29 runtime failures in onnx_op test suite, and we want to burn them down. I think the runtime failures have higher priorities because it mis-compiles the programs.

Some compilation failure may be related to codegen or frontend which needs further investigation.

See below links for the list:
- https://github.com/iree-org/iree/blob/main/tests/external/iree-test-suites/onnx_ops/onnx_ops_cpu_llvm_sync_O0.json
- https://github.com/iree-org/iree/blob/main/tests/external/iree-test-suites/onnx_ops/onnx_ops_cpu_llvm_sync_O2.json

hanhanW:
Initial status update from Claude:

# IREE issue #24378 — onnx_ops failure status

Test suite: `tests/external/iree-test-suites/onnx_ops/onnx_ops_cpu_llvm_sync_O{0,2}.json`,
config `cpu_llvm_sync` (single-threaded CPU). Compile and run flag sets:

```
iree-compile --iree-hal-target-device=local \
             --iree-hal-local-target-device-backends=llvm-cpu \
             --iree-input-demote-f64-to-f32=false \
             --iree-opt-level=O0   # (or O2 for the other config)

iree-run-module --device=local-sync ...
```

Counts (this branch, local build):

| Bucket                       | O0  | O2  |
|------------------------------|-----|-----|
| `expected_compile_failures`  | 283 | 283 |
| `expected_run_failures`      | 30  | 29  |
| `skip_compile_tests`         | 5   | 5   |
| `skip_run_tests`             | 1   | 1   |

Compile failure lists are identical between O0 and O2. Run failure list differs
by exactly one test: `test_nonmaxsuppression_suppress_by_IOU_and_scores` is
listed for O0 but not O2 (likely flaky / order-sensitive; treat as O0-only).

## Categories

The methodology was: group every failing test by ONNX op family, run one
representative per family with the local build, and bucket the diagnostic.
Spot-checks confirmed family-wide consistency for the ones I sampled.

### Compile failure buckets

* **F-onnx** — onnx-to-torch decomposition missing.
  `failed to legalize operation 'torch.operator' (name = "onnx.X")`. The op
  has no `OnnxToTorch` pattern at all, or there is one but it bails on the
  attribute combination present in this test. Owned by torch-mlir
  `TorchOnnxToTorch`.
[...]

PR #24394, opened 2026-05-07
[tools] Account both atol and rtol in floating point comparison.

hanhanW:
Previously, we have two modes: one is absolute equality and the other is relative equality. However, they are not all used in comparison which is not common in floating point comparison.

The revision collapses both into "approximate" mode that takes both into account which follows ML framworks convention (e.g., `numpy.isclose`, `torch.allclose`, etc). Note that this is not symmetric. The symmetric version is `math.isclose` but it is not widely used in ML workloads (assuming people are using numpy, etc.). Thus, the revision follows the formula:

```
|a-b| <= atol + rtol * |b|
```

instead of

```
|a-b| ≤ max(atol, rtol·|b|)
```

This "addresses" some false signals in onnx_op tests, because the diff now can live within the error range. Previously, it took atol path by default.

The revision refactors `NumpyFuzzyCompare` to `runtime/src/iree/base/internal/math.h`, so it can be used by both check modules and other runtime tools.

The revision exposes pre-existing bugs about NaNs, so some tests are disabled. See issues:
- https://github.com/iree-org/iree/issues/24414
- https://github.com/iree-org/iree/issues/24397

It is a step towards https://github.com/iree-org/iree/issues/24378


runtime/src/iree/tooling/buffer_view_matchers.c
```diff
@@ -76,14 +76,28 @@ static bool iree_hal_compare_strided_elements_exact(
...
+// NaN is considered equal to NaN; non-finite expected is only equal when
+// bit-equal to actual.
+static inline bool iree_hal_approximate_compare_f64(double actual,
+                                                    double expected,
+                                                    double atol, double rtol) {
+  if (actual == expected) return true;
+  if (isnan(actual) && isnan(expected)) return true;
+  if (!isfinite(expected)) return false;
+  return fabs(actual - expected) <= atol + rtol * fabs(expected);
+}
```
bjacob: I didn't know about this set of fuzzy compares. Last year I did some comparable work on the Check module and the fuzzy comparison that I wrote there, also closely following NumPy (also in treat-NaNs-as-equal mode), is contained in this function, which from a cursory reading I *think* is exactly equivalent to your logic here (except not promoting to f64)?

https://github.com/iree-org/iree/blob/0fe0ca8f7c7daf9af6e6da1df995bdd1098a8c9c/runtime/src/iree/modules/check/module.cc#L70-L89

Can you reuse it, perhaps moving it as necessary?

bjacob: If you need a place to move it to, maybe base/internal/math.h.  I realize that you need to use this in C code so this can't be templated anymore; maybe do like you had done in your function, just work on f64 with argument promotion.

bjacob: Since our functions are equivalent (please check carefully) and yours is already C, my request becomes to please move your function to base/internal/math.h and update check/module.cc to use it.

hanhanW: Yes, the logic is the same. I explored the function but I did not make a refactoring because one is `.c` (c language and the other is `.cc` (cpp language). I'll move them to `math.h`.

bjacob (approved):

hanhanW:
wow, I'm surprised that we've been treating resnet50 model green on rdna3 because it produces `NaNs`. I'm glad that at least the model is passing on CPU side and CPU does not produce NaNs.

I'll file an issue and move forward after I double-confirm. It is definitely a pre-existing issue, because I did not update any codegen logic and the log prints NaNs.

The same issue also happens to the other `det` tests.

```text
● IR walkthrough for test_det_2d 

  Input MLIR (model.mlir):

  func.func @test_det_2d(%arg0: !torch.vtensor<[2,2],f32>) -> !torch.vtensor<[],f32> ... {
    %0 = torch.operator "onnx.Det"(%arg0) : (!torch.vtensor<[2,2],f32>) -> !torch.vtensor<[],f32>
    return %0 : !torch.vtensor<[],f32>
  }

  After onnx-to-torch decomposition (DecomposeComplexOps):

  func.func @test_det_2d(%arg0: !torch.vtensor<[2,2],f32>) -> !torch.vtensor<[],f32> ... {
    %0 = torch.aten.linalg_det %arg0 : !torch.vtensor<[2,2],f32> -> !torch.vtensor<[],f32>
    return %0 : !torch.vtensor<[],f32>
  }

  After torch-to-linalg (the buggy lowering):

  %cst   = arith.constant 0.000000e+00 : f32
  %cst_0 = arith.constant 1.000000e+00 : f32

  // Take row 0 as the first U row (no pivoting!).
  %row0   = tensor.extract_slice %0[0,0] [1,2] [1,1] : tensor<2x2xf32> to tensor<1x2xf32>
  %pivot  = tensor.extract_slice %0[0,0] [1,1] [1,1] : tensor<2x2xf32> to tensor<1x1xf32>
  %row1   = tensor.extract_slice %0[1,0] [1,2] [1,1] : tensor<2x2xf32> to tensor<1x2xf32>
  %col0   = tensor.extract_slice %0[1,0] [1,1] [1,1] : tensor<2x2xf32> to tensor<1xf32>

  // One Gauss-elim step: row1 -= (M[1,0]/M[0,0]) * row0
  %3 = linalg.generic ... ins(%row0, %pivot, %row1, %col0 ...) outs(%2 ...) {
  ^bb0(%in: f32, %in_9: f32, %in_10: f32, %in_11: f32, %out: f32):
    %9  = arith.divf %in_11, %in_9   : f32       // <-- divides by M[0,0]; here 0
    %10 = arith.cmpf one,  %in_9, %cst : f32      // <-- "is pivot != 0?"
    cf.assert %10, "unimplemented: determinants requiring permutations and singular matrices"
    %11 = arith.mulf %9, %in : f32
    %12 = arith.subf %in_10, %11 : f32
    linalg.yield %12 : f32
  } -> tensor<1x2xf32>

  // Take the diagonal entries of the now-eliminated matrix and multiply.
  %inserted_slice_4 = tensor.insert_slice %3 into %0[1,0] [1,2] [1,1] : ...
  %diag_1 = tensor.extract_slice %inserted_slice_4[1,1] [1,1] [1,1] : ...


  // reduction: prod(U_diag)
  %6 = linalg.generic { reduction } ins(%inserted_slice_7 : tensor<2xf32>) ... {
  ^bb0(%in: f32, %out: f32):
    %9 = arith.mulf %in, %out : f32
    linalg.yield %9 : f32
  } -> tensor<1xf32>

  What blows up for M = [[0,1],[2,3]]:

  ┌────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                  Step                  │                                                                               Result                                                                               │
  ├────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ divf %in_11, %in_9                     │ 2.0 / 0.0 = +∞                                                                                                                                                     │
  ├────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ cmpf one 0.0, 0.0                      │ false → cf.assert should fire — but per linalg-iteration, not at op-graph level, so the divf has already populated the buffer for this lane and reduction proceeds │
  ├────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ subf 3.0, (∞ * 1.0)                    │ −∞                                                                                                                                                                 │
  ├────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ subf 2.0, (∞ * 0.0)                    │ 2.0 − NaN = NaN                                                                                                                                                    │
  ├────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ prod of U_diag = [0, NaN] (or [0, −∞]) │ NaN                                                                                                                                                                │
  └────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

  So the result is NaN, which the old comparator silently accepted because fabs(NaN − x) > t is always false in IEEE-754. PR #24394 makes the existing bug visible.

  Where to fix: torch-mlir's aten.linalg_det torch-to-linalg lowering. The current code does a single Gauss-elim sweep with no row pivoting. The matrix here is not singular (det = −2), but the lowering can't tell because
  it lacks partial pivoting.
```

Summary from claude:

```
  Pre-PR f32 absolute matcher (commit 6efc2ca92a, before #24394):

  // runtime/src/iree/tooling/buffer_view_matchers.c
  static bool iree_hal_compare_strided_elements_approximate_absolute_f32(
      iree_hal_buffer_equality_t equality, iree_host_size_t element_count,
      const float* expected_ptr, iree_host_size_t expected_stride,
      const float* actual_ptr, iree_host_size_t actual_stride,
      iree_host_size_t* out_index) {
    for (iree_host_size_t i = 0; i < element_count; ++i) {
      if (fabsf(*expected_ptr - *actual_ptr) > equality.f32_threshold) {
        *out_index = i;
        return false;          // mismatch
      }
      expected_ptr += expected_stride;
      actual_ptr   += actual_stride;
    }
    return true;                 // <-- "all elements matched"
  }

  The default mode was --equality_mode=absolute:

  // runtime/src/iree/tooling/comparison.cc (pre-PR)
  IREE_FLAG(string, equality_mode, "absolute", ...);

  Walk through it for the resnet output (1000 elements, all -NaN in CI):

  fabsf(-1.75692f - NaN)  ==  fabsf(NaN)  ==  NaN     // IEEE-754: any arith with NaN -> NaN
  NaN > 0.0001f                          ==  false     // IEEE-754: any non-`!=` comparison with NaN -> false

  So the if branch is never taken on a NaN element. The loop just iterates past every NaN, never sets out_index, never return falses, and at the end of 1000 iterations falls through to return true;. The matcher reports
  "all matched."
  if (fabsf((*expected_ptr - *actual_ptr) / *expected_ptr) > equality.f32_threshold)
  //        ^^ NaN-finite -> NaN, /finite -> NaN, fabsf -> NaN, > t -> false

```

hanhanW:
I'm not able to reproduce the NaN result on my rdna3. I'll try it on CI machine.

hanhanW:
> I'm not able to reproduce the NaN result on my rdna3. I'll try it on CI machine.

I can't reproduce it on CI machine with the iree-compile and iree-runtime binaries from this PR (CI). Trying the full setup  locally now.

hanhanW:
> > I'm not able to reproduce the NaN result on my rdna3. I'll try it on CI machine.
> 
> I can't reproduce it on CI machine with the iree-compile and iree-runtime binaries from this PR (CI). Trying the full setup locally now.

Classic claude. It downloaded bad artifacts (that sent me a different mlir file). The other session is able to reproduce it and confirm that it is a pre-existing issue on main. I'll add them to failing list and create an issue.

PR #24398, opened 2026-05-08
[Codegen] Move remaining pipelines to `iree_codegen` attrs

kuhar:
Follow-up to the LLVMGPU, SPIRV, and LLVMCPU pipeline migrations (#23816, #23851, #23864). Same goal: get backend-specific pipeline identifiers out of the global `DispatchLoweringPassPipeline` enum and behind `PipelineAttrInterface`.

This moves the remaining non-CPU/GPU pipeline cases to `iree_codegen`: `#iree_codegen.vmvx_pipeline`,
`#iree_codegen.transform_dialect_codegen`, and
`#iree_codegen.no_pipeline`. It keeps custom pass pipelines represented as `#iree_codegen.pass_pipeline<...>`.

These attrs intentionally stay in `iree_codegen`. `vmvx_pipeline` does not move to a backend dialect because IREE does not have a VMVX codegen dialect today; it can become a backend-owned enum attr if such a dialect is introduced later. `transform_dialect_codegen` and `no_pipeline` are also codegen-level sentinels, so modeling the remaining cases as unit attrs avoids creating dialect surface just for pipeline spelling.

After this, `TranslationInfoAttr` and `iree_codegen.smt.constraints` parse pipeline attributes directly through the interface. The old enum, keyword compatibility parser, helper predicates, and textual parsing in Python dialect tests can be removed.

Assisted-by: codex


krzysz00 (approved):
Looks mechanical and correct

PR #24400, opened 2026-05-08
[Codegen] Fix multi-reduction source layout propagation

keshavvinayak01:
Recombine multi-reduction source layouts from the source reduction dimensions and result non-reduction dimensions so distributed local reductions stay compatible with the accumulator/result layout.

Also keep parent vector.mask operands in the same source-compatible layout for masked multi-reductions.

Uncovered this bug when `flex_attention` -> `online_attention` generated a masked reduction over the score/key dimension.


keshavvinayak01:
Closing because the fix is upstream; there are tensor ops in the mask modification body, which is failing the `linalg.generic` from vectorizing, which is the breaking failure for this analysis.

PR #24401, opened 2026-05-08
Synthesize a `<name>$shape_query` companion for dynamic-shape outputs in Torch input conversion

Abhishek-Varma:
Lets runtime callers resolve the values of dynamic output dims before invoking an entry point. Motivating consumer: zero-copy in-place output binding for dynamic-shape results in the ONNX Runtime IREE EP ([`iree-org/onnxruntime-ep-iree#48`](https://github.com/iree-org/onnxruntime-ep-iree/issues/48)); the mechanism itself is general.

## Compiler

`FuncConversionPass` runs a pre-pass: for every public Torch func with a dynamic-shape result, synthesize a sibling `<name>$shape_query` with the same data inputs as the main func (storage args dropped) plus one mutable `!torch.tensor<[rank],si64>` shape-buffer arg per dynamic-shape result, in result-index order.

The caller pre-fills each buffer with the declared shape (sentinel for dynamic dims); the companion writes only the dynamic slots, so static slots are untouched.

The body of the shape query companion is a clone of the main func with in-place-output pattern short-circuited: `Torch::OverwriteTensorContentsOp`s are dropped and subsequent `Torch::CopyToValueTensorOp` reads of that storage fold to the recorded src. Each dynamic-shape result then feeds `tensor::DimOp` + `tensor::InsertOp` into its shape buffer. The companion rides the same coarse-fences wrapping as the main func, so shape-buffer args get `IREE::HAL::TensorImportOp` on read and `IREE::HAL::TensorAliasOp` on write (thus the caller's buffer view is updated in place).

NOTE: `tensor::DimOp`s fold against `flow.tensor.tie_shape` / `IREE::HAL::TensorImportOp` downstream; input-symbolic dims become buffer-view dim lookups, data-dependent dims pay for the underlying compute.

Discovery: `iree.abi.output_shape_query = @<name>$shape_query` on the main func's `iree.reflection` dict, propagated onto async + sync wrappers by `retainFunctionAttributes`.

## Runtime

Two helpers in `runtime/src/iree/runtime/call.{h,c}` :-

- `iree_runtime_call_lookup_output_shape_query` — resolves the companion; `IREE_STATUS_NOT_FOUND` if the function has no dynamic-shape results.
- `iree_runtime_call_invoke_output_shape_query` — invokes it with caller-supplied inputs (data inputs followed by one rank-sized i64 HAL buffer view per dynamic-shape result).

I could try to use `iree_vm_function_lookup_attr_by_name` + plain `iree_vm_invoke` but the helpers, I believe, make the invocation flow convenient.

### DIFF
--- a/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
@@ -10,10 +10,13 @@
 #include "iree/compiler/Dialect/HAL/IR/HALTypes.h"
 #include "iree/compiler/Dialect/Util/IR/UtilDialect.h"
 #include "iree/compiler/Dialect/Util/IR/UtilOps.h"
+#include "llvm/ADT/SmallSet.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Func/Transforms/FuncConversions.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/IRMapping.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
@@ -461,6 +464,18 @@ LogicalResult ConvertedAsyncFunctionInfo::convertMutableTensorArg(
     builtinTensorType = cast<Torch::NonValueTensorType>(torchType)
                             .getWithValueSemantics()
                             .toBuiltinTensor();
+    // Strip signedness from integer element types so that the import side
+    // matches the overwrite/alias side (which goes through
+    // `convertToBuiltinTensor` and folds to signless). Without this,
+    // signed-integer storage args produce a `tensor<NxsiNN>` import that
+    // can't legalize against the signless tensor flowing into the
+    // shape-query companion's `IREE::HAL::TensorAliasOp`. Mirrors
+    // convertImmutableTensorArg.
+    if (auto intTy =
+            dyn_cast<IntegerType>(builtinTensorType.getElementType())) {
+      builtinTensorType = builtinTensorType.clone(
+          builder.getIntegerType(intTy.getIntOrFloatBitWidth()));
+    }
   }
 
   // Propagate explicit affinities and ABI behavior to the read and write.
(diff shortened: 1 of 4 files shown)


compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -596,6 +921,33 @@ class FuncConversionPass final
...
+        continue;
+      }
+      // Don't recurse if this pass has already placed shape-query companions
+      // in the module from a prior run.
+      if (funcOp.getName().ends_with("$shape_query")) {
+        continue;
+      }
+      if (hasDynamicShapeResult(funcOp)) {
+        mainFuncsForShapeQueryCompanion.push_back(funcOp);
+      }
```
IanWood1, line 944: Based on the discussion yesterday we might want to keep the shape query for static outputs.

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -596,6 +921,33 @@ class FuncConversionPass final
   void runOnOperation() override {
     auto moduleOp = getOperation();
 
+    // Pre-pass: for every public torch func with a dynamic-shape result,
+    // synthesize a sibling `<name>$shape_query` companion and stamp
```
IanWood1, line 925: Does this actually ignore private functions? This looks like it also processes private functions because `shouldConvertFunc` doesn't check if it's public

runtime/src/iree/runtime/call.c
```diff
@@ -123,3 +123,52 @@ IREE_API_EXPORT iree_status_t iree_runtime_call_outputs_pop_front_buffer_view(
...
+  IREE_ASSERT_ARGUMENT(session);
+  IREE_ASSERT_ARGUMENT(out_query_function);
+  memset(out_query_function, 0, sizeof(*out_query_function));
+
+  iree_string_view_t companion_name = iree_vm_function_lookup_attr_by_name(
+      &function, IREE_SV("iree.abi.output_shape_query"));
+  if (iree_string_view_is_empty(companion_name)) {
+    return iree_status_from_code(IREE_STATUS_NOT_FOUND);
+  }
+  if (function.module == NULL) {
```
IanWood1, line 143: `iree_vm_function_lookup_attr_by_name` dereferences `function.module` so this needs to be moved up.

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -515,6 +530,315 @@ void retainFunctionAttributes(Operation *srcOp, IREE::Util::FuncOp destOp) {
...
+/// Clones every op in `srcEntry` into the companion body, in source order,
+/// while peeling away the in-place output binding pattern:
+///   - `Torch::OverwriteTensorContentsOp` is dropped; we remember which
+///     value was just written into which storage arg.
+///   - A later `Torch::CopyToValueTensorOp` of that same storage is also
+///     skipped, and its result is mapped to the remembered value.
+/// Reads with no matching prior overwrite clone normally; they resolve to the
+/// companion's own block arg via `mapping`.
+static void cloneBodyShortCircuitingStorage(Block *srcEntry, OpBuilder &builder,
+                                            IRMapping &mapping) {
```
IanWood1, line 708: We need to be careful that the shape query function doesn't do any of the graph computation. For instance, if there are assertions and assertions are not stripped, `linalg.generic` ops containing assertions will not be folded away. This will prevent their producers from folding away making the shape query expensive (potentially causing the full program to be run).

runtime/src/iree/runtime/call.h
```diff
@@ -111,6 +111,30 @@ IREE_API_EXPORT iree_status_t iree_runtime_call_inputs_push_back_buffer_view(
 IREE_API_EXPORT iree_status_t iree_runtime_call_outputs_pop_front_buffer_view(
     iree_runtime_call_t* call, iree_hal_buffer_view_t** out_buffer_view);
 
+//===----------------------------------------------------------------------===//
```
IanWood1, line 114: Is there a place to test the runtime side and maybe add full e2e tests?

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -515,6 +530,315 @@ void retainFunctionAttributes(Operation *srcOp, IREE::Util::FuncOp destOp) {
...
+  auto shapeBuiltinTy = RankedTensorType::get({rank}, i64Type);
+  Value shapeBuiltin = TorchConversion::ToBuiltinTensorOp::create(
+      builder, loc, shapeBuiltinTy, shapeVt);
+
+  Value cur = shapeBuiltin;
+  for (unsigned d : dr.dynamicDims) {
+    Value idx = arith::ConstantIndexOp::create(builder, loc, d);
+    Value dim = tensor::DimOp::create(builder, loc, resultBuiltin, idx);
+    Value dimI64 = arith::IndexCastOp::create(builder, loc, i64Type, dim);
+    cur = tensor::InsertOp::create(builder, loc, dimI64, cur, ValueRange{idx});
```
IanWood1, line 764: `tensor.from_elements` might be more canonical. Also, would it be better to just return a bunch of i64 values?

compiler/plugins/input/Torch/InputConversion/FuncConversion.cpp
```diff
@@ -515,6 +530,315 @@ void retainFunctionAttributes(Operation *srcOp, IREE::Util::FuncOp destOp) {
   }
 }
 
+//===----------------------------------------------------------------------===//
```
IanWood1, line 533: Here's an alternative implementation idea: generate the query from the `hal.tensor.export` dynamic dim operands (`source_dims`). This could run after Input Conversion/ABI wrapping where the `hal.tensor.export` is present. These are the benefits I can see:


- Input agnostic, since it only depends on core IREE/MLIR dialects.
- Avoids cloning the whole function. The query only needs the backward slice defining the exported dynamic dim values, which should usually be much smaller after shape simplification has run.
- Gives a natural place to define legality: inspect the slice and reject/diagnose cases where output shape computation depends on unsupported compute or side effects.
- Avoids relying on downstream CSE/DCE to remove an entire cloned program body.

This also seems like it would address the side-effect assertion concern from cloning arbitrary Torch IR, since the implementation can choose exactly which ops are legal to clone into a shape query function.

PR #24404, opened 2026-05-08
[Codegen][CPU] Replace transposed_intrinsic with explicit pre-transposed enums.

bjacob:
Each non-square hardware MMA now appears as two `MMAIntrinsic` enum values — one per orientation — named after their actual MxNxK shape (e.g. `MMA_X86_AVX512_1x16x1_F32_F32` and its M↔N-swapped sibling `MMA_X86_AVX512_16x1x1_F32_F32`). The swapped sibling of `X` is `X ^ 1` (bit 0 of D in the enum's encoding), so consecutive enum values form natural/swapped pairs. The `transposed_intrinsic` boolean on `DataTiledMMAAttr` is removed; both orientations show up as distinct candidates in `getMmaIntrinsicsForTargetConfig` and the existing `chooseIntrinsic` scoring picks between them.

A new `assert(isMmaIntrinsicArrayValid(...))` at the bottom of `getMmaIntrinsicsForTargetConfig` enforces the convention: strictly increasing, every swapped entry preceded by its natural sibling, and square/type-symmetric singletons allowed (covers the generic-scalar 1×1×1 family).

Motivation: the boolean was originally there to avoid 2× enum growth, but it leaked orientation-handling into every layer that touched MMA layouts — swizzles, ACC encoding-info, the cross-intrinsic ACC index, broadcast direction. It was also poised to get worse: asymmetric LLVM intrinsics (vpdpbusd) and non-x86 targets (no broadcast) would each need bespoke handling under the boolean. With separate enum values, each orientation lowers independently.

MMAUtils permutation fix:

  `buildDataTiledMMAUnderlyingOperations` previously moved between the
  caller's logical (M, N)-ordered tile type and the swizzle's
  distributed N-D form via a linear-preserving `vector.shape_cast` —
  sound only for identity permutations. Pre-existing intrinsics all
  happened to be identity, so the bug was dormant. Every swapped enum
  added here produces a non-identity permutation (cross-intrinsic-N
  injected before cross-intrinsic-M, putting N first in storage), and
  the shape_cast scrambles (m, n) coordinates so result values land in
  the wrong cells.

  Fix: split the reshape into a logical-shape `shape_cast` plus a
  `vector.transpose` by the swizzle's permutation (input side) / its
  inverse (output side). Folds to a no-op for identity permutations,
  so existing natural-orientation intrinsics produce the same IR as
  before. Verified by extending the existing `lower_avx512_16x1x1_f32`
  test in `lower_inner_tiled.mlir` (the first to exercise a
  non-identity permutation) with `CHECK` lines pinning the
  shape_cast/transpose pairs around the FMA call.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

Progress towards #24323

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.td
+++ b/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.td
@@ -46,13 +46,11 @@ class IREECPU_I32EnumAttr<string name, string summary,
 //       with different tile sizes.
 //
 // Note that even a scalar multiply-accumulate is technically a matmul
-// with MxNxK shape 1x1x1. Vector element-wise multiply-accumulate with vector
-// size N is seen here as a matrix multiply-accumulate with MxNxK shape 1xNx1.
-// It could be equivalently seen as shape Nx1x1, so we have to pick a convention
-// here, breaking the M-N symmetry: we choose to spread the vector dimension
-// along the N matrix dimension. The MMA attribute types that wrap this enum
-// shall combine it with a boolean "transpose" parameter flipping that, which
-// will typically be used for narrow matmuls with a small N-dimension size.
+// with MxNxK shape 1x1x1. A vector multiply-accumulate of N lanes can be
+// spread along either M (Nx1x1) or N (1xNx1); both orientations are
+// useful (different ACC layouts and broadcast directions). We enumerate
+// each orientation as its own `MMAIntrinsic` value — see the bit-0
+// convention on D below.
 //
 // Symbolic names are of the form
 //
@@ -93,7 +91,11 @@ class IREECPU_I32EnumAttr<string name, string summary,
 //   * 4 = 8-bit float (incl. f8E5M2, f8E4M3FN, etc).
 //   * A = 16-bit integer (any signedness).
 //   * C = 8-bit integer (any signedness).
-// * D enumerates intrinsics that share the same 0xABC* bits.
+// * D enumerates intrinsics sharing the same 0xABC* bits. Bit 0 of D
+//   (mask 0x1) marks the M↔N-swapped orientation: each natural/swapped
+//   pair occupies two consecutive values, and the swapped sibling of `X`
+//   is `X ^ 1`. Neither orientation is fundamental — the cost model picks
+//   whichever fits the matmul shape best.
 //
 def IREECPU_MMA_None : I32EnumAttrCase<"None", 0>;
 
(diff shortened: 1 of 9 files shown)


compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUEnums.td
```diff
@@ -93,7 +91,11 @@ class IREECPU_I32EnumAttr<string name, string summary,
 //   * 4 = 8-bit float (incl. f8E5M2, f8E4M3FN, etc).
 //   * A = 16-bit integer (any signedness).
 //   * C = 8-bit integer (any signedness).
-// * D enumerates intrinsics that share the same 0xABC* bits.
+// * D enumerates intrinsics sharing the same 0xABC* bits. Bit 0 of D
+//   (mask 0x1) marks the M↔N-swapped orientation: each natural/swapped
+//   pair occupies two consecutive values, and the swapped sibling of `X`
+//   is `X | 1`. Neither orientation is fundamental — the cost model picks
```
hanhanW:
```suggestion
//   is `X ^ 1`. Neither orientation is fundamental — the cost model picks
```

I think it should be `xor`? E.g., for `(0, 1)` pair, the swapped sibling is `(1, 0)` not `(1, 1)`.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
```diff
@@ -462,21 +579,23 @@ getIntrinsicInfo(MLIRContext *ctx, ArrayRef<Type> elementTypes,
...
-/// narrow for optimization purposes, so it never enters the `min`.
+/// and, finally, by iteration order. The natural and M↔N-swapped
+/// orientations of each hardware MMA appear as distinct enum values in the
+/// candidate list, so the choice between them falls out of this same
+/// scoring loop. `K` is never narrow for optimization purposes, so it never
+/// enters the `min`.
 ///
 /// Returns nullopt if no compatible intrinsic exists.
-static std::optional<std::pair<IREE::CPU::MMAIntrinsic, bool>>
+static std::optional<IREE::CPU::MMAIntrinsic>
```
hanhanW: I think we no longer need `std::optional` in this case, because we can treat `nullptr` attribute as `std::nullopt`.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
```diff
@@ -361,39 +369,145 @@ pickGenericScalarMMAForTarget(DictionaryAttr config) {
...
+        MMAIntrinsic::MMA_X86_AVX512BF16_16x1x2_F32_BF16,
         MMAIntrinsic::MMA_X86_AVX512_1x16x2_I32_I16,
+        MMAIntrinsic::MMA_X86_AVX512_16x1x2_I32_I16,
         MMAIntrinsic::MMA_X86_AVX512VNNI_1x16x2_I32_I16,
+        MMAIntrinsic::MMA_X86_AVX512VNNI_16x1x2_I32_I16,
         MMAIntrinsic::MMA_X86_AVX512_1x16x2_I32_I8_CASTI16,
+        MMAIntrinsic::MMA_X86_AVX512_16x1x2_I32_I8_CASTI16,
         MMAIntrinsic::MMA_X86_AVX512VNNI_1x16x2_I32_I8_CASTI16,
+        MMAIntrinsic::MMA_X86_AVX512VNNI_16x1x2_I32_I8_CASTI16,
     };
```
hanhanW, line 511: We can make it more robust. I think we can use `getMaxEnumValForXXX` and `symbolize` method to construct the array/vector. Here is an example:

https://github.com/iree-org/iree/blob/1d87665ee3f32be1d918c8da57069e31aad2e3cb/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp#L3232-L3238

https://github.com/iree-org/iree/blob/1d87665ee3f32be1d918c8da57069e31aad2e3cb/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUEnums.td#L545-L566

There are two options:
1. Group MMA intrinsics based on arch. E.g., we can group `MMA_X86_*` into a `IREECPU_I32EnumAttr`, then you can do the trick here.
2. We still use the single big group, and we can filter it from the string view, like:

```
for (uint32_t i = 0; i <= getMaxEnumValForMyEnum(); ++i) {
    if (std::optional<MyEnum> val = symbolizeMyEnum(i)) {
      // Convert the enum value back to its string name
      llvm::StringRef name = stringifyMyEnum(*val);
      
      // Check if it starts with "foo_"
      if (name.starts_with("foo_")) {
        fooCases.push_back(*val);
      }
    }
  }
```

(2) has a hidden assumption that all the enums' name start with `MMA_X86` but I think it is fine.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
```diff
@@ -361,39 +369,145 @@ pickGenericScalarMMAForTarget(DictionaryAttr config) {
...
+      }
+    } else {
+      // Condition (4): singleton — must be self-symmetric.
+      if (mn != nn || lhsn != rhsn) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
```
hanhanW, line 463: I'm thinking if we should add a few methods for those `- 1`, `| 1`, `^ 1` methods for readability. E.g., we can have `getSiblingInst`, `isNativeInst`, etc. What do you think?

hanhanW (commented):
Just some qol nits.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
bjacob: I thought about that, but the enum values here are very sparse (even within one arch), and could get even sparser in the future, so this isn't practical.

This also violates a principle from the Google c++ guide: "don't parse". Relying in parsing string can come back to haunt us in the future. It also makes it harder to navigate the code, as one can't simply grep or clangd-reference-jump to specific enum values anymore.

Finally, the current solution is fairly "solid" thanks to the assert(isMmaIntrinsicArrayValid(...)).  If one makes a mistake updating this array, this will be caught in assert-enabled builds.

compiler/src/iree/compiler/Codegen/ExternalInterfaces/CPUEncodingExternalModels.cpp
bjacob: I would like to keep the explicit bit ops here because it's more transparent how they interact with one another. Example at line 411:

```c++
    if ((v & 1) != 0 && prev != (v ^ 1)) {
```

If thanks to helpers I rewrite that into

```c++
    if (isTransposed(v) != 0 && prev != sibling(v)) {
```

then one has to mentally expand the definitions of `isTransposed` and `sibling` to understand how this fits together.

hanhanW (approved):
The PR description is still using `| 1`, not `^ 1`.

PR #24406, opened 2026-05-08
[Codegen][CPU] Add x86 AVX-512 VNNI vpdpbusd MMA intrinsic.

bjacob:
Adds the AVX-512 VNNI `vpdpbusd` MMA intrinsic — the byte dot-product whose first source is unsigned and second is signed, the first intrinsic to exercise distinct LHS/RHS types end-to-end.

Two new enum values, `MMA_X86_AVX512VNNI_1x16x4_I32_UI8_I8` and its M↔N-swapped sibling `MMA_X86_AVX512VNNI_16x1x4_I32_I8_UI8`, both lowering to `llvm.x86.avx512.vpdpbusd.512`. The swapped lowering flips its `(lhs, rhs)` call args so the unsigned operand still lands in the first slot.

Per IREE convention, vector tile types in IR are signless; the signed/unsigned distinction lives on the encoding's `element_types`. `getABCElementTypes(intrinsic)` returns `ui8` where applicable (used by the cost model and validator); `getUndistributedTileTypes` strips signedness when constructing tile types. The validator and `getIntrinsicInfo` bypass `getUndistributedTileTypes` and read shape + types straight from the intrinsic via `getRowMajorTilesMNKShape` and `getABCElementTypes`, which become public for that reason.

Progress towards #24323

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUTypes.h
+++ b/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUTypes.h
@@ -85,6 +85,19 @@ bool isGenericScalar(MMAIntrinsic intr);
 // budget encoded in the enum case (8 or 16). Asserts otherwise.
 int64_t getGenericScalarRegisterBudget(MMAIntrinsic intr);
 
+// Returns the (LHS, RHS, ACC) element types for `intrinsic`, with
+// signedness preserved (e.g. `ui8` for vpdpbusd) — used by the cost model
+// to match against an encoding's `element_types`. Tile types in IR are
+// signless; `DataTiledMMAAttr::getUndistributedTileTypes` strips the
+// signedness annotations.
+std::tuple<Type, Type, Type> getABCElementTypes(MLIRContext *ctx,
+                                                MMAIntrinsic intrinsic);
+
+// Returns the (M, N, K) tile shape for a row-major-tile intrinsic, or
+// nullopt otherwise (e.g. SVE).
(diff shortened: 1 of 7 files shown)


compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUTypes.h
```diff
@@ -85,6 +85,19 @@ bool isGenericScalar(MMAIntrinsic intr);
...
+// signedness preserved (e.g. `ui8` for vpdpbusd) — used by the cost model
+// to match against an encoding's `element_types`. Tile types in IR are
+// signless; `DataTiledMMAAttr::getUndistributedTileTypes` strips the
+// signedness annotations.
+std::tuple<Type, Type, Type> getABCElementTypes(MLIRContext *ctx,
+                                                MMAIntrinsic intrinsic);
+
+// Returns the (M, N, K) tile shape for a row-major-tile intrinsic, or
+// nullopt otherwise (e.g. SVE).
+std::optional<std::tuple<int64_t, int64_t, int64_t>>
```
hanhanW, line 98: I wonder if we use `TileMxNxK` instead. It's used by the old mmt4d (or GPU) and I think this can be reused.

https://github.com/iree-org/iree/blob/83b0606c88a09403cf64824a05c8a0fef231758d/compiler/src/iree/compiler/Codegen/Dialect/Codegen/Utils/Utils.h#L118-L122

hanhanW (approved):

PR #24402, opened 2026-05-08
[VectorDistribute] Fix gfx1100 pipeline test

sommerlukas:
Updating this test to use `online_attention` as missed in https://github.com/iree-org/iree/pull/24110 and the expected count of LDS materializations was then erroneously updated in https://github.com/iree-org/iree/pull/24341.

On gfx1100, the QK output layout not compatible with the PV input layout, so the fourth (next to three promoted operands) LDS materialization is required.

Assisted-by: Codex

PR #24408, opened 2026-05-08
[VectorDistribute] Rework LDS operand promotion

sommerlukas:
Rework how promotion of operands to LDS works in the VectorDistribute pipeline. 

So far, `linalg.copy` operations were inserted early in the pipeline. Now, we skip the insertion of `linalg.copy` for operands (we keep the behavior for results). Instead, the analysis from https://github.com/iree-org/iree/pull/24227 propagates the promotion types upwards from the compute operations that actually configure the promotion of operands up to the operations accessing the data in memory (`transfer_read`, `gather`) when we reach `GPUVectorAlloc`. Based on the propagated information, the necessary promotion to LDS can be inserted. 

Layout conflicts are still resolved through an LDS roundtrip at the conflict point.

Assisted-by: Codex

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutUtils.cpp
@@ -8,6 +8,7 @@
 
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/IR/BuiltinTypeInterfaces.h"
 
 namespace mlir::iree_compiler {
 
@@ -95,6 +96,53 @@ SmallVector<int64_t> getElementVectorTileShape(NestedLayoutAttr vectorLayout) {
     tileShape[i] = 1;
   }
   return tileShape;
+}
+
+FailureOr<NestedLayoutAttr>
+getDerivedThreadLayout(MLIRContext *context, ArrayRef<int64_t> workgroupSize,
+                       ArrayRef<int64_t> logicalShape,
+                       ArrayRef<int64_t> elementTile) {
+  int64_t rank = logicalShape.size();
+  if (rank == 0 || elementTile.size() != rank) {
+    return failure();
+  }
+
+  SmallVector<int64_t> opShape(logicalShape);
+  for (auto [size, element] : llvm::zip_equal(opShape, elementTile)) {
+    if (ShapedType::isDynamic(size) || element <= 0 || size % element != 0) {
+      return failure();
+    }
+    size /= element;
+  }
+
+  SmallVector<int64_t> threadTile(rank, 1);
+  SmallVector<int64_t> threadStrides(rank, 0);
+  int64_t residualThreads = ShapedType::getNumElements(workgroupSize);
+  int64_t currStride = 1;
+  for (auto [tile, stride, size] :
+       llvm::reverse(llvm::zip(threadTile, threadStrides, opShape))) {
+    int64_t threadBlock;
+    if (residualThreads % size == 0) {
+      threadBlock = size;
+    } else if (size % residualThreads == 0) {
+      threadBlock = residualThreads;
+    } else {
+      return failure();
+    }
+
+    tile = threadBlock;
+    stride = currStride;
+    size /= threadBlock;
+    currStride *= threadBlock;
+    residualThreads /= threadBlock;
+  }
+
+  SmallVector<int64_t> subgroupTile(rank, 1);
+  SmallVector<int64_t> subgroupStrides(rank, 0);
+  SmallVector<int64_t> outerTile(rank, 1);
+  return NestedLayoutAttr::get(context, subgroupTile, opShape, outerTile,
+                               threadTile, elementTile, subgroupStrides,
+                               threadStrides);
 }
 
 } // namespace mlir::iree_compiler
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
@@ -142,7 +142,7 @@ materializeSharedMemoryConversions(FunctionOpInterface funcOp) {
 }
 
 struct PromotionCandidate {
-  Operation *op;
+  Operation *op = nullptr;
   Value vector;
 };
 
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
@@ -172,7 +172,7 @@ getPromotionCandidate(Operation *op, const PromotionTypeMap &promotionTypes) {
         return shouldPromoteCandidate(gather.getOperation(),
                                       gather.getResult());
       })
-      .Default([](Operation *) { return std::nullopt; });
+      .Default(std::nullopt);
 }
 
 /// Materialize promotion of operands to LDS.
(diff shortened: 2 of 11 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConfigureTensorLayouts.cpp
sommerlukas, line 1: The changes in this file are mostly a code move to `GPUNestedLayoutUtils` to make computation of a `derived_thread_config` layout available as a shared helper.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutUtils.cpp
```diff
@@ -97,4 +98,51 @@ SmallVector<int64_t> getElementVectorTileShape(NestedLayoutAttr vectorLayout) {
   return tileShape;
 }
 
+FailureOr<NestedLayoutAttr>
+getDerivedThreadLayout(MLIRContext *context, ArrayRef<int64_t> workgroupSize,
```
sommerlukas, line 102: This code was moved here from `LLVMGPUConfigureTensorLayouts` to make computation of a `derived_thread_config` layout available as a shared helper.

sommerlukas:
Failure of the gfx1100 pipeline test is expected, this PR needs to be rebased on top of https://github.com/iree-org/iree/pull/24402 once that lands.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -128,6 +141,121 @@ materializeSharedMemoryConversions(FunctionOpInterface funcOp) {
...
+  };
+  return TypeSwitch<Operation *, std::optional<PromotionCandidate>>(op)
+      .Case([&](vector::TransferReadOp read) {
+        return shouldPromoteCandidate(read.getOperation(), read.getVector());
+      })
+      .Case([&](vector::GatherOp gather) {
+        return shouldPromoteCandidate(gather.getOperation(),
+                                      gather.getResult());
+      })
+      .Default([](Operation *) { return std::nullopt; });
```
kuhar:
```suggestion
      .Default(std::nullopt);
```

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -128,6 +141,121 @@ materializeSharedMemoryConversions(FunctionOpInterface funcOp) {
   return success();
 }
 
+struct PromotionCandidate {
+  Operation *op;
```
kuhar:
```suggestion
  Operation *op = nullptr;
```

kuhar (commented):
Just a drive-by nit, I haven't had time to review the logic

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -128,6 +141,121 @@ materializeSharedMemoryConversions(FunctionOpInterface funcOp) {
...
+        builder, op->getLoc(), toLayout.getResult(),
+        allocationLayout.getUndistributedShape());
+    if (failed(copied)) {
+      return failure();
+    }
+
+    auto synced =
+        IREE::GPU::ValueBarrierOp::create(builder, op->getLoc(), *copied);
+    Value newRead =
+        readVectorFromTensor(builder, readType, synced.getResult(0));
```
keshavvinayak01, line 279: The expected IR is correct for the write-then-read roundtrip itself, but the new direct operand-promotion path does not insert the pre-write loop-iteration barrier that the existing `shared_memory_conversion` materialization path deliberately inserts. If promoted reads can appear in loop bodies and reuse the same workgroup allocation, this can race with reads from the previous iteration.

If we can prove these promoted read roundtrips never land in a loop, then I guess this is fine.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
```diff
@@ -141,10 +269,18 @@ struct GPUVectorAllocPass final
...
 
+    // Run the promotion propagation to propagate promotion information from
+    // compute ops up to the ops accessing memory.
+    PromotionTypeMap promotionTypes = analyzePromotionTypes(funcOp);
+
+    // Remove the existing promotion information. All those operations will have
+    // their promotion materialized based on the analysis above.
+    funcOp.walk([](IREE::VectorExt::ToLayoutOp op) {
+      op.removeSharedMemoryConversionAttr();
+    });
```
keshavvinayak01, line 312: A `shared_memory_conversion = #iree_gpu.use_global_load_dma` marker is propagated by the analysis, then erased, then ignored by the late materializer. The resulting IR may be valid, but it no longer performs the requested operand promotion.

Please do a lit-test to confirm this behaviour ?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_vector_alloc.mlir
```diff
@@ -1,45 +1,118 @@
...
   element_tile = [4, 1],
-
-  subgroup_strides = [1, 1],
+  subgroup_strides = [0, 0],
   thread_strides   = [0, 0]
 >
 
-func.func @invalid_shared_memory_conversion_attr(%vector: vector<16x16xf16>) -> vector<16x16xf16> {
-  // expected-error @+1 {{shared_memory_conversion attribute must implement IREE::GPU::PromotionAttr}}
-  %out = iree_vector_ext.to_layout %vector to layout(#layout) {shared_memory_conversion = "invalid"} : vector<16x16xf16>
```
keshavvinayak01, line 43: This test update removes the only negative coverage for invalid `shared_memory_conversion` attrs. With the new pass order, all shared_memory_conversion attrs are stripped before  `materializeSharedMemoryConversions` validates that they implement `PromotionAttr`, and the late promotion path only handles `DerivedThreadConfigAttr`. 

Please keep or replace the invalid attr test so we prove unsupported/invalid promotion markers fail instead of being silently erased.

keshavvinayak01 (changes requested):
Needs some changes.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_vector_alloc.mlir
sommerlukas: I've added validation of the attribute before the propagation.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
sommerlukas: Added a test. I also have the follow-up adding support for `use_global_load_dma` already lined up locally, if you prefer, I can make it part of this PR.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
sommerlukas: I've added the barrier back (thanks for catching that) and updated the tests.

sommerlukas (commented):
Thanks for the feedback!

compiler/src/iree/compiler/Codegen/Common/GPU/GPUVectorAlloc.cpp
keshavvinayak01: I think we should merge it here.

keshavvinayak01 (approved):
Looks fine, please add the `use_global_load_dma` changes here.

sommerlukas:
@keshavvinayak01 I've added the async DMA handling to the PR.

keshavvinayak01 (commented):
Let's split the `async_dma` extension work into a follow up PR.

keshavvinayak01 (approved):

PR #24413, opened 2026-05-08
Revert "[LinalgExt] Fix attention NaN for fully-masked rows"

### LINKED ISSUE #24412: [MI325][CI]Perf regression after #24178
After https://github.com/iree-org/iree/pull/24178, CI started to see performance regressions on `llama_8b_fp16/prefill_benchmark_seq2048_mi325.json`

Its previous commit:  d2f4f441a6  Bump torch-mlir@46cbd27f7c passed.

Examples:
* https://github.com/iree-org/iree/actions/runs/25453930468/job/74697268019
* https://github.com/iree-org/iree/actions/runs/25460700807/job/74703156406

Fix or revert?

hanhanW:
Reverts iree-org/iree#24178

It's landed with regressions: https://github.com/iree-org/iree/actions/runs/25453930468/job/74697268019

I don't see if we decide to accept it or not from comments and PR description, so let's revert it to reduce CI noises. We can land it with updated golden values if we accept the regression.


lialan (approved):

lialan:
This addresses: https://github.com/iree-org/iree/issues/24412

@keshavvinayak01 please fix the issue and re-commit again.

bangtianliu (approved):

keshavvinayak01:
I think we should update the golden value and re-land the fix. It's  necessary to avoid NaN's that will lead to https://github.com/iree-org/iree/issues/24175. It's only adding this ~0.5% regression for cases with masks. I do not see any other way to fix that issue than this change I already made. @hanhanW thoughts? @Groverkss @sommerlukas

sommerlukas:
The regression is bigger than 0.5% though. The current golden time is 242ms, the [last successful CI](https://github.com/iree-org/iree/actions/runs/25444824291/job/74647419946) run on `main` before the fix got merged (there were a few that failed because of machine issues in between) reported 243.93ms as current time. 

So, compared to the last time measured on CI, the regression is >10%.

When I tried it, I was unable to reproduce the `NaN` issue with a boolean mask instead of an FP mask. Any chance we can use a similar path to boolean masks to fix the issue?

hanhanW:
There is a 10% tolerance, so it is actually ~11% (assuming the machine is stable enough). If you looked at the individual dispatch (i.e., attention dispatch), the regression can be much higher. I think you need some benchmarking and breakdown before you claim that 0.5% is okay. I don't follow the work closely, so maybe @Groverkss can provide some suggestions or make a call here.

```
FAILED iree-test-suites/torch_models::torch_models::llama_8b_fp16/prefill_benchmark_seq2048_mi325.json::llama_8b_fp16/prefill_benchmark_seq2048_mi325.json - AssertionError: Benchmark failed: reported_time 269.774 ms exceeds golden_time 242.000 ms * tolerance_factor 1.1 = 266.200 ms.
                            Time reported corresponds to the minimum of 10 samples.
assert 269.77426124115783 <= 266.20000000000005
 +  where 269.77426124115783 = <IREEBenchmarkTest llama_8b_fp16/prefill_benchmark_seq2048_mi325.json>.reported_time
```

PR #24416, opened 2026-05-08
[Codegen][CPU] Drop the power-of-two cap on `chooseUnrolling`.

bjacob:
`chooseUnrolling` previously enumerated `intrinsics_m` / `intrinsics_n` through powers of two only. There's no real reason for that — within the register budget, the cost model can pick any factor.

Looking ahead: some intrinsics may eventually want the unroll factor on a particular dim to be a multiple of some target-specific number (Arm scalar-by-element FMA forms are an example: the scalar is a lane of a vector register, so covering whole registers is more efficient). But even there it's not a strict preference — narrow matmuls (e.g. matrix-vector) would be better off skipping the padding. So a simple per-intrinsic "preferred multiple" doesn't quite capture it; we'll revisit when a concrete need lands.

Lit-test fallout: relaxing pow2 lets the cost model land on better non-pow2 unrolls in some cases the existing tests pin (e.g. `materialize_encoding_x86_64.mlir`'s dynamic-N AVX-512 case now picks `intrinsics_m = 29` instead of 16 — saturating the zmm register file more tightly, at higher arithmetic intensity per the existing scoring formula). CHECK lines updated.

Progress towards #24323

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_aarch64.mlir
@@ -609,11 +609,11 @@ func.func @matmul_lowering_i8i4i32_aarch64_i8mm(
 //
 // FP4 register pressure: K=2 is the smallest power-of-two K that makes
 // K*4 a multiple of 8 (for byte-addressable packed groups). Inside the
-// 16-register budget the cost model lands on a square (im=2, in=2, ik=2)
-// tile (2*2 + 2*2 + 2*2 = 12 registers), which wins on arithmetic
-// intensity. Whether the lowering can actually densely pack f4E2M1FN
-// values into bytes today is orthogonal — the framework only insists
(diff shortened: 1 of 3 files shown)


bjacob:
FYI @egebeysel , this addresses a review comment that you had where you noted that existing RISC-V mmt4d configs used the non-power-of-two size 7. For now I'm just switching us wholesale to that kind of arbitrary odd sizes; we'll consider reinstating "rounding" requirements on a case by case basis. Anyway, this code is easy to specialize per-intrinsic.

hanhanW (approved):

egebeysel:
@bjacob Sorry I was OOO on Friday and couldn't give this a review, did you decide not going forward with this approach? Does it not select the inner tile sizes you expect for other (intrinsic, target) pairs?

bjacob:
I just temporarily closed open PRs as I need to focus on re-landing a couple of key PRs that were reverted. No change on the approach . I hope to reopen this soon.

PR #24417, opened 2026-05-08
[Utils] Add interactive `debug-break-(before|after)`  to iree-compile

hanhanW:
Introduces an opt-in PassInstrumentation that pauses the top-level PassManager at selected pass boundaries, dumps the live op (or its enclosing module) to a file, waits for a resume signal, then parses the edited file and splices it back into the live op. Intended for interactive prototyping and agent-driven IR editing without rebuilding the compiler.

Flags (visible):

```
  --iree-debug-break-before=<pass,...>   Break before listed passes.
  --iree-debug-break-after=<pass,...>    Break after listed passes.
  --iree-debug-break-file=<path>         Exchange-file path. Required
                                         when a break flag is set;
                                         no implicit default (matches
                                         the HAL pass-options
                                         convention).
  --iree-debug-break-mode={stdin|file}   How to wait for resume.
                                         stdin reads 'continue'/'c' or
                                         'abort'/'a'; file polls for
                                         <break-file>.continue / .abort
                                         sentinels. Default: stdin.
```

Hidden flags:

```
  --iree-debug-break-poll-interval-ms    File mode poll cadence.
                                         Default 100ms.
  --iree-debug-break-timeout-sec         File mode timeout. 0 means
                                         wait forever. Default 3600
                                         (one hour) so agent/CI loops
                                         don't hang silently if the
                                         driver dies.
  --iree-debug-break-test-patch-file     Auto-splice from this file
                                         at every break and resume
                                         without waiting. Used by lit
                                         tests; not for interactive
                                         workflows.
```

Scope: works at any op level. Both ModuleOp passes and nested ops (function-scope passes inside a FunctionLikeNest, hal.executable.variant codegen passes, etc.) trigger breaks. The banner identifies the live op and its symbol so users can tell which iteration of a per-function pass fired. The dump always covers the enclosing ModuleOp so cross- references (e.g. util.call to sibling functions) parse and verify standalone; the splice still targets only the matching op, identified by op-name and qualified symbol path (e.g. @grandparent::@parent::@op) so leaf-name collisions across nested scopes don't pick the wrong op. Edits to sibling ops in the dump are silently discarded.

Threading: when any break flag is set, the instrumentation's ctor disables MLIRContext multithreading for its lifetime — nested OpPassManagers that MLIR may run in parallel would otherwise race on stdin and the single exchange file. The dtor restores the prior state so long-lived sessions (e.g. Python bindings) that reuse the context for later non-debug compiles aren't silently left single-threaded.

Failure handling: failed parses re-prompt in interactive modes (the parser/verifier diagnostics go to stderr, the user fixes the file, sends continue again). User-triggered terminations (timeout, explicit abort, missing required flag, auto-patch parse failure) emit a clear stderr error and exit via report_fatal_error(GenCrashDiag=false) — clean exit with no crash banner.

Output: banners go to stderr (stdout carries the compiler's primary output, e.g. `.vmfb`). Banner emission bypasses MLIR's diagnostic engine to avoid the default handler attaching
"see current operation: <full module dump>" to every line.

Wired into Invocation::createPassManager so the feature works across the full iree-compile pipeline (input conversion, global-opt, dispatch-creation, flow, stream, HAL, codegen). Adds a lit test under tools/test/debug_break.mlir covering both phases (before/after) with auto-patch, stdin no-op round-trip, file-mode no-op round-trip coverage.

NOTE: this is mainly for fast prototype and debugging. Any work depends on this is fragile and the support may be dropped anytime.

Assisted-by: Claude Code, Codex


compiler/src/iree/compiler/Utils/PassDebugBreak.cpp
```diff
@@ -0,0 +1,587 @@
...
+// We bypass MLIR's diagnostic engine because the default handler annotates
+// every remark/error with "see current operation: <full module dump>"
+// when the context has shouldPrintOpOnDiagnostic=true, which would flood
+// the banner with the entire IR at every break.
+//
+// Kept as a forwarder to `llvm::errs()` rather than a standalone
+// `raw_fd_ostream(STDERR_FILENO, …)` so buffered writes interleave cleanly
+// with the rest of the compiler's stderr output. If `llvm::errs()` ever
+// gets rebound away from stderr, this helper is the single place to swap
+// in a direct fd-2 stream.
```
AGindinson, line 119: IMHO this wrapper is superficial and `llvm::errs()` can be reliably used directly.

compiler/src/iree/compiler/Utils/PassDebugBreak.cpp
```diff
@@ -0,0 +1,587 @@
...
+// disk match what `op.print` produced — OF_Text would apply CRLF translation
+// on Windows, changing checksums for drivers that hash the exchange file.
+//
+// For non-module ops we dump the enclosing ModuleOp rather than the op
+// itself, so cross-references (e.g. a `util.call` to a sibling function)
+// resolve when the file is re-parsed and verified standalone. The splice
+// in parseAndReplace still targets the specific live op identified by
+// op-name + symbol-name; edits the user makes to sibling ops in the dump
+// are silently discarded. That trade-off (dump for context, splice for
+// safety) is documented in the public header.
```
AGindinson, line 172:
```suggestion
// are silently discarded.
```

kuhar:
I will try to review this by EOD

compiler/src/iree/compiler/Utils/PassDebugBreak.cpp
```diff
@@ -0,0 +1,587 @@
...
+  llvm::StringSet<> before = collect(clBreakBefore);
+  llvm::StringSet<> after = collect(clBreakAfter);
+  if (before.empty() && after.empty()) {
+    return nullptr;
+  }
+  if (clBreakFile.empty()) {
+    exitWithUserError(
+        "--iree-debug-break-{before,after} requires "
+        "--iree-debug-break-file=<path> (no implicit default; pick a "
+        "path the driver can read/write)");
```
AGindinson, line 578:
```suggestion
        "--iree-debug-break-file=<path> to load the modified "
        "IR from");
```

compiler/src/iree/compiler/API/Internal/CompilerDriver.cpp
```diff
@@ -800,6 +801,16 @@ std::unique_ptr<PassManager> Invocation::createPassManager() {
...
   }
   passManager->addInstrumentation(std::make_unique<PassTracing>());
+  if (auto breakInstrumentation = createPassDebugBreakInstrumentationFromFlags(
+          passManager->getContext())) {
+    // The instrumentation's ctor/dtor toggles MLIRContext multithreading
+    // for its own lifetime, so nested op pipelines that MLIR may run in
+    // parallel don't race on stdin or the exchange file while a break is
+    // active. Scoping lives there so long-lived sessions (e.g. Python
+    // bindings) that reuse the context aren't silently left
+    // single-threaded after this invocation.
```
AGindinson, line 811: Do we need this comment also repeated here?

compiler/src/iree/compiler/Utils/PassDebugBreak.cpp
```diff
@@ -0,0 +1,587 @@
...
+// than `std::exit`: that's the LLVM-idiomatic shutdown path (used widely in
+// LLVM/MLIR), runs the registered fatal-error handlers, and the
+// `GenCrashDiag=false` form skips the bug-report banner / stack trace that
+// would otherwise be misleading for a user-triggered condition.
+//
+// `std::exit` is not the right tool here because PassInstrumentation
+// override returns void — we can't propagate a `LogicalResult failure` back
+// to the pass manager. Once an unrecoverable user-facing condition is
+// reached, the only options are abort-style termination or doing nothing.
+// Termination is correct; we just want it without the crash-report styling.
```
AGindinson, line 136: IMO this describes self-explanatory code and doesn't provide any valuable context. I would suggest only retaining the `/*GenCrashDiag=*/false` explanation as a comment inside the function body.

compiler/src/iree/compiler/Utils/PassDebugBreak.cpp
```diff
@@ -0,0 +1,587 @@
...
+// into |liveOp|, and runs the MLIR verifier on the spliced result.
+//
+// For ModuleOp itself we just splice the parsed module's body into the live
+// module — the op-name match would also work but the direct path avoids the
+// walk and matches the historical behavior. Only discardable attributes are
+// copied so any inherent attributes the op type owns aren't stomped.
+static LogicalResult parseAndReplace(Operation *liveOp, llvm::StringRef path) {
+  MLIRContext *context = liveOp->getContext();
+  OwningOpRef<ModuleOp> parsed =
+      parseSourceFile<ModuleOp>(path, ParserConfig(context));
```
AGindinson, line 304: Should/could the verifier also be run on the "raw" loaded module before going into the match & replace routine?

tools/test/debug_break.mlir
```diff
@@ -0,0 +1,101 @@
...
+// RUN: FileCheck --check-prefix=BANNER-FILE-MODE %s < %t-file.stderr
+
+// BANNER-FILE-MODE: [iree-debug-break] BEGIN phase=after pass=iree-sanitize-module-names
+// BANNER-FILE-MODE: touch {{.*}}.continue{{.*}} to signal
+// BANNER-FILE-MODE: [iree-debug-break] END phase=after pass=iree-sanitize-module-names
+
+func.func @add(%arg0: tensor<4xf32>, %arg1: tensor<4xf32>) -> tensor<4xf32> {
+  %0 = arith.addf %arg0, %arg1 : tensor<4xf32>
+  return %0 : tensor<4xf32>
+}
```
AGindinson, line 84: Could you add more regression test cases for scenarios such as:
- discardable attributes being present;
- correct replacement within a module with multiple functions;
- module-level override?

compiler/src/iree/compiler/Utils/PassDebugBreak.cpp
```diff
@@ -0,0 +1,587 @@
...
+    os << "@" << segments[i - 1];
+  }
+  return result;
+}
+
+// Walks |parsedRoot| to find an op matching |liveOp|. Prefers an exact
+// match on op-name + qualified symbol path (e.g.
+// `hal.executable.variant` + `@sum_dispatch_0::@embedded_elf_x86_64`).
+// Falls back to op-name + leaf symbol name, then op-name only — these
+// fallbacks let users rename the symbol in the dump and still get their
```
AGindinson, line 228: Just curious, what's the practical use-case for renaming the symbol?

AGindinson (commented):
Thank you for working on this and making this into a PR! This is incredibly useful for debugging and prototyping.

The design and implementation LGTM overall. My suggestions are mostly around doc strings, aiming to simplify/disambiguate some of the explanations.

hanhanW:
> I will try to review this by EOD

Take your time. I think anyone can get the patch if they want to use it.

PR #24418, opened 2026-05-08
[Codegen] Fix RAW in_bounds false for R&W

amd-eochoalo:
Fix fold for transfer_read(transfer_write(val_to_store)) when in_bounds is not guaranteed.

When transfer_read and transfer_write do not have dimensions being accessed guaranteed to be in_bounds, it is not safe to fold transfer_read(transfer_write(val_to_store)) -> val_to_store. However, when transfer_read XOR transfer_write have in_bounds, we take advantage of undefined behaviour to fold them to val_to_store.

Case 1.1: w_ib = false, r_ib = true, position is actually in_bounds
We write val, we read val, we can fold RAW to val.
Case 1.2: w_ib = false, r_ib = true, position is NOT in_bounds
We skip write, read says it is in_bounds, but that is false, which is UB
therefore we can fold to val.
Case 2.1: w_ib = true, r_ib = false, position is actually in_bounds
We write val, we read val, we can fold RAW to val.
Case 2.2: w_ib = true, r_ib = false, position is NOT in_bounds
UB on the write, therefore we can fold.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
@@ -317,13 +317,8 @@ struct FoldTransferRAW : OpRewritePattern<vector::TransferReadOp> {
     TypedValue<VectorType> wMask = writeOp.getMask();
     TypedValue<VectorType> rMask = readOp.getMask();
 
-    // The fold diverges from the original write+read at positions that are
-    // out-of-bounds: the write does not store there and the read returns
-    // pad, but the fold returns valToStore. This is only a problem when
-    // *both* transfers have OOB dimensions (in_bounds=false). When at
-    // least one side has in_bounds=true, it asserts the position is within
-    // bounds; an actual OOB access is undefined behavior, so the fold
-    // cannot introduce new incorrectness. (Verified with Z3.)
+    // If at least one operation claims that the position is in-bounds
+    // then we can fold.
     //
     // Case 1.1: w_ib = false, r_ib = true, position is actually in_bounds
     // We write val, we read val, we can fold RAW to val.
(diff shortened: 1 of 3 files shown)


amd-eochoalo:
I'm a bit against folding in the XOR cases. Maybe a runtime error would be better?

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -317,6 +317,27 @@ struct FoldTransferRAW : OpRewritePattern<vector::TransferReadOp> {
     TypedValue<VectorType> wMask = writeOp.getMask();
     TypedValue<VectorType> rMask = readOp.getMask();
 
+    // The fold diverges from the original write+read at positions that are
```
krzysz00: I think another phrasing here is that if at least one operation claims that the position is in-bounds, then we can fold

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -317,6 +317,27 @@ struct FoldTransferRAW : OpRewritePattern<vector::TransferReadOp> {
...
+    // We write val, we read val, we can fold RAW to val.
+    // Case 1.2: w_ib = false, r_ib = true, position is NOT in_bounds
+    // We skip write, read says it is in_bounds, but that is false, which is UB
+    // therefore we can fold to val.
+    // Case 2.1: w_ib = true, r_ib = false, position is actually in_bounds
+    // We write val, we read val, we can fold RAW to val.
+    // Case 2.2: w_ib = true, r_ib = false, position is NOT in_bounds
+    // UB on the write, therefore we can fold.
+    if (readOp.hasOutOfBoundsDim() && writeOp.hasOutOfBoundsDim()) {
+      return failure();
```
krzysz00, line 333: I don't think this is a fail - this should basically create an implicit mask

Heck, you can rewrite those out of bounds dimensions to explicit masks

krzysz00 (approved):
This looks correct but too conservative

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
amd-eochoalo: I think we can write that as a separate pattern that focuses only on that specific case as a follow up PR. It would reduce some branching and make the code more legible.

krzysz00: Entirely fair

PR #24419, opened 2026-05-08
Add chipDetails for RDNA3

RattataKing:
Add `perfTflops` and `memoryBandwidthTbps` to get precise `computeMemoryCutoff ` value for each RDNA3 SKU.

The new GEMM size cutoff value on W7900 brings more improvements than regressions across a sample set of GEMM shapes, with an avg improvement of **~10%**. See data [here](https://gist.github.com/RattataKing/fddaf22bf0807fae0961240bc9847d78).


krzysz00 (commented):
Seems fine to me

krzysz00 (approved):

kuhar (approved):

PR #24426, opened 2026-05-09
[Torch][LinalgExt] Lower flex_attention masks before online_attention

keshavvinayak01:
Lower flex_attention mask_mod by evaluating the mask function over broadcastable index tensors and passing the resulting tensor mask into online_attention.

This is done to fix existing GPU compilation failure when I tested flex_attention with causal masks. 

The old lowering built Q/K/V indices with `iree_linalg_ext.index -> tensor.from_elements`. After decomposition
this left a mask-update `linalg.generic` in the QK path that did not vectorize cleanly, introduced private-memory between QK and the reductions, and lead to the source layout being null downstream in `VectorLayoutAnalysis`


compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -294,32 +293,182 @@ static SmallVector<Value> inlineTorchFunction(PatternRewriter &rewriter,
...
+///
+/// The callback is authored over scalar Torch tensors, but mask_mod is
+/// independent of the score value. Calling it with broadcastable index tensors
+/// lets normal Torch-to-linalg lowering produce a tensor mask instead of a
+/// rank-0 tensor payload inside the online_attention score region.
+static FailureOr<SmallVector<Value>> inlineFlexAttentionMaskFunction(
+    PatternRewriter &rewriter, Location loc, FlatSymbolRefAttr funcSymbol,
+    ValueRange args, ArrayRef<int64_t> tensorShape, Operation *contextOp) {
+  auto module = contextOp->getParentOfType<ModuleOp>();
+  auto funcOp = module.lookupSymbol<func::FuncOp>(funcSymbol);
```
sommerlukas, line 324: Should we check `module` before accessing it?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -294,32 +293,182 @@ static SmallVector<Value> inlineTorchFunction(PatternRewriter &rewriter,
...
+    for (auto [oldResult, newResult] :
+         llvm::zip_equal(op.getResults(), clonedOp->getResults())) {
+      mapper.map(oldResult, newResult);
+    }
+  }
+
+  auto returnOp = dyn_cast<func::ReturnOp>(entryBlock.getTerminator());
+  if (!returnOp) {
+    return failure();
+  }
```
sommerlukas: Should this check be performed before the cloning to avoid leaving the IR in a bad state?

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -294,32 +293,182 @@ static SmallVector<Value> inlineTorchFunction(PatternRewriter &rewriter,
...
+    SmallVector<int64_t> indexShape(scoreShape.size(), 1);
+    indexShape[dim] = scoreShape[dim];
+    Type indexTensorType = RankedTensorType::get(indexShape, signlessI32);
+    Type torchIndexType =
+        torch::Torch::ValueTensorType::get(context, indexShape, signedI32);
+    AffineMap identityMap =
+        AffineMap::getMultiDimIdentityMap(indexShape.size(), context);
+    SmallVector<AffineMap> maps = {identityMap};
+    SmallVector<utils::IteratorType> iterTypes(indexShape.size(),
+                                               utils::IteratorType::parallel);
```
sommerlukas: Do we need these allocations? For the second one, `llvm::Repeated<>` might be an option.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -294,32 +293,182 @@ static SmallVector<Value> inlineTorchFunction(PatternRewriter &rewriter,
...
+    return failure();
+  }
+
+  auto maskType =
+      dyn_cast<torch::Torch::ValueTensorType>((*maskResults)[0].getType());
+  if (!maskType) {
+    return failure();
+  }
+
+  Value maskValue = (*maskResults)[0];
```
sommerlukas: Nit: If we define `maskValue` in line 432, we can use it to get the `maskType` instead of repeating `(*maskResults)[0]`.

sommerlukas (commented):
Implementation looks OK to me, just a few nits. 

Would be good to get eyes on this from somebody with more experience in this area.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: Yeah, they are just used at compile-time though.

sommerlukas (approved):
Implementation LGTM, please wait for another review from someone with experience in this area, e.g., @IanWood1 or @rsuderman.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -294,32 +293,184 @@ static SmallVector<Value> inlineTorchFunction(PatternRewriter &rewriter,
...
+  if (!module) {
+    return failure();
+  }
+  auto funcOp = module.lookupSymbol<func::FuncOp>(funcSymbol);
+  if (!funcOp || funcOp.isExternal() || !funcOp.getBody().hasOneBlock() ||
+      funcOp.getNumArguments() != args.size()) {
+    return failure();
+  }
+
+  Block &entryBlock = funcOp.getBody().front();
```
rsuderman, line 337: Lets move this lookup / remapping work to a helper. It is the majority of the behavior but a pretty generic problem. It also can provide a higher level explanation for what it is trying to do.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -294,32 +293,184 @@ static SmallVector<Value> inlineTorchFunction(PatternRewriter &rewriter,
...
+static FailureOr<Value> createFlexAttentionMask(PatternRewriter &rewriter,
+                                                Location loc,
+                                                FlatSymbolRefAttr maskModSymbol,
+                                                ValueRange torchIndices,
+                                                ArrayRef<int64_t> scoreShape,
+                                                Operation *contextOp) {
+  FailureOr<SmallVector<Value>> maskResults = inlineFlexAttentionMaskFunction(
+      rewriter, loc, maskModSymbol, torchIndices, scoreShape, contextOp);
+  if (failed(maskResults) || maskResults->size() != 1) {
+    return failure();
```
rsuderman: Use `rewriter.notifyMatchFailure` with failure message.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -294,32 +293,184 @@ static SmallVector<Value> inlineTorchFunction(PatternRewriter &rewriter,
...
+  }
+
+  Value maskValue = (*maskResults)[0];
+  auto maskType = dyn_cast<torch::Torch::ValueTensorType>(maskValue.getType());
+  if (!maskType) {
+    return failure();
```
rsuderman: Same.

rsuderman (changes requested):
You need more tests. So far the additions are extremely minimal despite a substantial change to how the lowering works.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
keshavvinayak01: Factored this into a local helper since it currently has one caller.

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_masked_configured.mlir
hanhanW, line 1: I was confused that why I got notifications several times, while the change is about Torch->LinalgExt lowering. It turns out that it is because you update this file. Why is it part of the PR?

keshavvinayak01: Because this is what exposed the failure of the mask from the decomposition. The old computation prevented vectorization of the linalg generic. I do see your point about it technically not being related though, I think I'll remove this.

hanhanW: If you look it from linalg's perspective, it is not a special linalg.generic op. It is just a elementwise op. I think we already have good coverage in upstream (since we use the upstream method to vectorize the op), so it's better to not have such test without actual code changes.

compiler/plugins/input/Torch/InputConversion/ConvertTorchUnstructuredToLinalgExt.cpp
```diff
@@ -468,12 +629,28 @@ struct FlexAttentionOpConversion
...
-    // No mask operand: mask_mod is inlined into the score region.
+    SmallVector<int64_t> scoreShape = {batch, numHeads, seqLenQ, seqLenKV};
+    Value maskOperand;
+    if (maskModSymbol) {
+      SmallVector<Value> torchIndices =
+          createTorchIndexTensors(rewriter, loc, scoreShape);
+      FailureOr<Value> mask = createFlexAttentionMask(
+          rewriter, loc, maskModSymbol, torchIndices, scoreShape, op);
+      if (failed(mask)) {
+        return failure();
```
rsuderman: Rewriter errors should use `notifyMatchFailure` so the failure can be more easily debugged.

rsuderman (changes requested):
Look at how other passes handle failure. I see a lot of `return failure();` which can make it difficult for  future developers to debug what went wrong.

compiler/plugins/input/Torch/InputConversion/Passes.cpp
```diff
@@ -66,6 +66,9 @@ void createTorchToIREEPipeline(
   pm.addNestedPass<func::FuncOp>(torch::createConvertTorchToTensorPass());
   pm.addNestedPass<func::FuncOp>(
       TorchInput::createConvertTorchUnstructuredToLinalgExtPass());
+  // Keep this as a module-level barrier so unstructured conversions finish
+  // across the module before later per-function Torch conversions run.
+  pm.addPass(createCanonicalizerPass());
```
rsuderman: This is a red flag to me. Why are you needing a canonicalization pass here? Typically we note adding passes as cocnerning.

rsuderman (changes requested):

compiler/plugins/input/Torch/InputConversion/Passes.cpp
keshavvinayak01: I added the canonicalizer as a module-level barrier because I was worried about the nested per-function pipeline converting a flex_attention callback function before the function containing the hop_flex_attention user had lowered and inlined it.

But in the actual lowering emitted top down from torch's invokation, we're safe for now.

Issue #23329, opened 2026-01-29
[Codegen][GPU] Do not swizzle for GEMMS which dont have bank conflicts


Muzammiluddin-Syed-ECE:
### Request description

We can use XOR swizzles to reduce bank conflicts however these do come with the cost of more complex indexing math. Do not use these swizzles in cases where we know there are no bank conflicts.

see pr: https://github.com/iree-org/iree/pull/23175

Issue #24255, opened 2026-04-24
Fix untuned XOR swizzle heuristic for DMA gather-to-LDS


Yu-Zhewen:
The untuned swizzle heuristic (getXorShuffleParamsForUntunedChipset in GPUUtils.cpp) can produce XOR shuffle parameters that are incompatible with the DMA gather-to-LDS lowering

https://github.com/iree-org/iree/blob/e4d008f02067b5c2f5ca874e3822b8717545fa9f/compiler/src/iree/compiler/Codegen/Common/GPU/AMDGPULowerCoalescedDMAToGatherLDS.cpp#L347-L357

When this happens, AMDGPULowerCoalescedDMAToGatherLDS would fall back to scalar vector.transfer_read/write.

#24254 works around this by skipping the untuned heuristic entirely for BF16 DMA operands (skipUntunedFallback=true). The proper fix is to make the untuned heuristic aware of the DMA transfer size so it only produces compatible swizzle parameters.

PR #24428, opened 2026-05-11
[Codegen][GPU] Phase-aware LDS bank conflict analysis and XOR swizzle search

Muzammiluddin-Syed-ECE:
Refactor the XOR shuffle parameter selection to use an iterative, phase-group-aware search instead of a hardcoded tuning table. The main changes are:
1. **LdsBankPhaseModel attribute**: Add a new `LdsBankPhaseModel` enum (`None`, `Contiguous`, `CDNA4`) to `IREEGPUEnums.td` and `TargetWgpAttr` to encode how the LDS hardware schedules thread phases for wide ds_read instructions.
2. **Analytical bank conflict detection**: Replace the old heuristic with `hasNoBankConflicts()`, which simulates every thread in each phase group, computes its LDS byte offset (with optional XOR swizzle remapping), and detects actual bank conflicts via a bank-to-address map.
3. **Constraint-based validation**: Factor swizzle validation into composable constraint functions (`makeXorShuffleBaseConstraintFn`, `makeXorShuffleDmaConstraintFn`) passed to `validateXorShuffle`. This makes it easy to add new constraints (e.g., DMA minimum access width) without modifying the core search.
4. **Remove tuned lookup table**: Delete `getXorShuffleParamsForGfx950`, `getXorShuffleParamsForTunedChipset`, `skipUntunedFallback`. The iterative search (`getXorShuffleParamsImpl`) is now the sole path and reproduces the same results as the old-tuned table. The `conv_bf16_dma_swizzle` test now correctly finds an XOR swizzle for the LHS operand where the old code skipped it due to `skipUntunedFallback`.


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1178,6 +1182,58 @@ std::optional<TargetDetails> getAndroidProfileDetails(StringRef target) {
 // Query functions
 //===----------------------------------------------------------------------===//
 
+constexpr int64_t kLdsBankWidth = 4; // LDS bank width in bytes (AMD)
+
+std::optional<SmallVector<SmallVector<int64_t>>>
+getPhaseGroups(LdsBankPhaseModel model, int64_t readBytes, int64_t numThreads) {
```
Yu-Zhewen: Thanks @Muzammiluddin-Syed-ECE. One high level question.

Is there an actual need to introduce the `LdsBankPhaseModel` attribute? Can't we just pass arch to `getPhaseGroups`?  I feel the enum doesn't seem to add much information. Plus we always have the same phase model for a given arch.

krzysz00: To roughly echo the position of the backend folks in a similar context: cod should, where possible, be parametrized on architecture *features*, not the architecture itself.

So such an enum lets you map multiple architectures to the "CDNA4" model if its gets reused and means you're not needing to stuck an "if (gfx950)" in places it doesn't need to be.

(and, well, suppose we find an Intel chip with the same bank layout - they shouldn't have to patch the heuristic code)

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
```diff
@@ -768,6 +774,9 @@ def IREEGPU_TargetWgpAttr : AttrDef<IREEGPU_Dialect, "TargetWgp"> {
     OptionalParameter<"DenseI64ArrayAttr">:$dma_sizes,
     // Number of banks in LDS.
     OptionalParameter<"std::optional<int32_t>">:$workgroup_memory_bank_count,
+    // LDS bank phase scheduling model for wide ds_read instructions.
```
krzysz00:
```suggestion
    // Workgroup memory bank phase scheduling model
```

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
```diff
@@ -210,6 +210,12 @@ def IREEGPU_DotProductOpsAttr : EnumAttr<
   let assemblyFormat = "$value";
 }
 
+def IREEGPU_LdsBankPhaseModelAttr : EnumAttr<
+  IREEGPU_Dialect, IREEGPU_LdsBankPhaseModel, "lds_bank_phase_model"> {
```
krzysz00: Let's not call this "LDS" - maybe something like "WorkgroupMem" or "SharedMem" or `WgMemory" or the like - there are architectures other than AMD that might want this?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -804,48 +807,38 @@ getTotalTileElems(IREE::Codegen::InnerTileDescAttrInterface intrinsic,
...
+    FailureOr<int64_t> maybeTotalTileElems =
+        getTotalTileElems(intrinsic, operandIndex);
+    if (failed(maybeTotalTileElems)) {
       return failure();
     }
-  }
-  return failure();
+    return isXORShuffleValid(params.rowElems, params.accessElems,
+                             *maybeTotalTileElems)
+               ? success()
```
krzysz00: Nit: `return success(isXORShuffleValid(...));`

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -904,75 +897,289 @@ bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
...
-  if (failed(maybeChipset)) {
+// Permute an MMASingleSubgroupLayout so that its arrays reflect the physical
+// LDS dimension order rather than the canonical semantic order.
+// For LHS, the M (parallel) dim is rotated to the end so K becomes outermost.
+// For RHS, the N (parallel) dim is rotated to the front so K becomes innermost.
+// This mirrors the convention in GPUTileSwizzleUtils.cpp for TileSwizzle
+// construction.
+static void permuteLayoutPerConfig(IREE::GPU::MMASingleSubgroupLayout &layout,
+                                   int operandIndex) {
+  bool isLhs = (operandIndex == IREE::GPU::kMMAOperandLhs ||
```
krzysz00: Clang-tidy catch for a redundant or - typo?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -904,75 +897,289 @@ bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
...
+  }
+
+  // Delinearize a thread ID into per-dimension coordinates using the layout's
+  // thread strides, then linearize those coordinates into an LDS byte offset.
+  // Note: mlir::delinearize requires strides in decreasing order, but tstrides
+  // doesn't guarantee that (e.g. the contiguous thread dimension may be dim 0).
+  auto getThreadByteOffset = [&](int64_t threadId) -> int64_t {
+    SmallVector<int64_t> coords(rank);
+    for (int64_t d = 0; d < rank; ++d) {
+      coords[d] = (threadId / layout.tstrides[d]) % layout.thread[d];
```
krzysz00, line 934: I'm not sure this is the right code here - there's a utility function for converting a set of strides into a sequence of sizes and a permutation (I can't remember exactly where, but you might want it)

krzysz00 (commented):
The concept makes sense, some notes

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -884,16 +884,14 @@ getMatmulOrIGEMMLoweringConfigAndWorkgroupSize(
     // TODO(#24255): Fix untuned swizzle logic for DMA.
     if (lhsElemType.isBF16() && !transposedLhs) {
       FailureOr<Attribute> lhsSwizzleAttr = getXorShuffleAttr(
-          context, lhsAttr, target, kind, schedule->kTileSizes, kMMAOperandLhs,
-          /*skipUntunedFallback=*/true);
+          context, lhsAttr, target, kind, schedule->kTileSizes, kMMAOperandLhs);
```
kuhar: This is the `use_global_load_dma` path, but `useDirectLoad` is left at its default of false. That means the DMA min-access-width constraint (added via `makeXorShuffleDmaConstraintFn`) is not enforced, even though we are explicitly emitting a DMA load. Please pass `/*useDirectLoad=*/true`. Otherwise the search may pick an `accessElems` that is below the DMA load granularity for some intrinsic/element-type combinations, and the constraint added in this very PR is silently unused on the only code path that motivated it (see `TODO(#24255)` above). The companion RHS call below has the same issue, plus is missing `/*isTransposed=*/true` (the call is gated by `transposedRhs`).

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -904,75 +897,289 @@ bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
...
+    }
+  }
+
+  return success(); // conflict-free
 }
 
-FailureOr<XorShuffleParams> getXorShuffleParamsForUntunedChipset(
+// Wrapper that extracts target/intrinsic parameters, permutes the layout
+// to match the physical LDS dimension order, and checks for bank conflicts.
+LogicalResult
```
kuhar: This seems like dead code with external linkage, or something that should be made `static`

kuhar (commented):
Test coverage for the new bank-conflict model is thin: the only behavioral test I see is the existing config_tile_and_fuse_gfx950.mlir BF16-DMA case, whose expected output flipped from 'no swizzle' to xor_shuffle<128,4>. There is no targeted test for getPhaseGroups (the CDNA4 4-phase grouping is a 64-entry magic table -- exactly the kind of thing that wants a unit/lit test that pins the grouping), and no test for the Contiguous model. Please add: (a) a parse/print test for the new lds_bank_phase_model attribute on a non-CDNA4 target (you have a CDNA4 case via target_attrs.mlir), (b) a targeted lowering test where the iterative search picks an XOR swizzle that the old tuned table did not (and vice versa, to catch regressions if the table is ever edited), and (c) a negative test verifying that on a target with phase model None, getXorShuffleParams now returns failure (it does, via hasNoBankConflicts returning failure). Without these, edits to the CDNA4 phase table may break the code without anyone noticing.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -904,75 +897,289 @@ bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
...
+          operandIndex](XorShuffleParams params) -> LogicalResult {
+    FailureOr<int64_t> minDmaBits = getMinDmaLoadBits(target);
+    if (failed(minDmaBits)) {
+      return failure();
+    }
+    FailureOr<int64_t> elemBits = getOperandBitwidth(intrinsic, operandIndex);
+    if (failed(elemBits)) {
+      return failure();
+    }
+    int64_t accessBits = params.accessElems * *elemBits;
```
kuhar: The DMA constraint only checks `accessElems * elemBits >= min(dma_sizes)`, but `AMDGPULowerCoalescedDMAToGatherLDS` later chooses larger DMA segments and requires `access_width % elementsPerLane == 0`. For gfx950 BF16, `xor_shuffle<128,4>` is compatible with 32-bit DMA but not a 128-bit segment (8 BF16/lane), so lowering can reject the generated swizzle. Please mirror `findSwizzleIncompatibleSegment` or reject incompatible DMA swizzles during search.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1178,6 +1182,58 @@ std::optional<TargetDetails> getAndroidProfileDetails(StringRef target) {
...
+  switch (model) {
+  case LdsBankPhaseModel::None:
+    llvm_unreachable("handled above");
+
+  case LdsBankPhaseModel::Contiguous:
+    buildContiguous();
+    break;
+
+  case LdsBankPhaseModel::CDNA4:
+    if (readBytes <= 8) {
```
kuhar: `getPhaseGroups(CDNA4)` maps every `readBytes > 8` access through the 16-byte `ds_read_b128` phase table. Some scaled-MMA operands have 32-element per-thread accesses; with 8-bit inputs that is 32 bytes, so the analyzer may use the wrong phase schedule. Please either split wider reads into modeled b128 chunks or guard this model to the widths the table actually represents, with coverage for the scaled cases.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -904,75 +897,289 @@ bool isXORShuffleValid(int64_t numRowElems, int64_t numAccessElems,
...
+
+  IREE::GPU::MMASingleSubgroupLayout layout =
+      IREE::GPU::getSingleSubgroupLayout(intrinsic, operandIndex);
+  if (isTransposed) {
+    permuteLayoutPerConfig(layout, operandIndex);
+  }
+
+  // First check if there are any conflicts without swizzle.
+  if (succeeded(hasNoBankConflicts(layout, numBanks, elemBits, phaseModel))) {
+    return failure(); // Already conflict-free, no swizzle needed.
```
kuhar: `getXorShuffleParamsImpl` overloads `failure()` with two distinct meanings, and the caller in `ConfigUtils.cpp` cannot tell them apart:

1. Line 1111: returns `failure()` when the layout is already conflict-free (no swizzle needed).
2. Returns `failure()` everywhere else when something genuinely went wrong (no phase model, no bitwidth, no bank count, search exhausted).

Look at the scaled MMA path in `ConfigUtils.cpp`:

```c++
if (failed(lhsSwizzleAttr) || failed(rhsSwizzleAttr)) {
  promotionArray = {};
}
```

If either LHS or RHS happens to be conflict-free, `promotionArray` is wiped out entirely. That removes promotion for both operands and silently changes codegen for a pattern that was previously promoted. The intent is clearly "no swizzle => use plain `defaultConfigAttr`", not "drop all promotion".

Please distinguish these states. Options:
- Return `std::optional<XorShuffleParams>` semantic where `nullopt` means "no swizzle needed (and that is fine)" vs. `FailureOr` for actual failures.
- Or return a sentinel like `XorShuffleParams{0,0}` to signal "no swizzle needed".

Either way, the caller must be able to distinguish "swizzle needed but couldn't be computed" (true failure) from "no swizzle needed" (use default config). The current behavior is a latent regression for scaled MMA on CDNA4 the moment a tile shape happens to be conflict-free.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1178,6 +1182,58 @@ std::optional<TargetDetails> getAndroidProfileDetails(StringRef target) {
...
+      buildContiguous();
+    } else {
+      // ds_read_b128 on CDNA4: 4 phases of 16 threads, non-contiguous.
+      // Phase assignment from hardware documentation / empirical measurement.
+      static const int64_t cdna4B128Phases[64] = {
+          0, 0, 0, 0, 2, 2, 2, 2, 2, 2, 2, 2, 0, 0, 0, 0, 2, 2, 2, 2, 0, 0,
+          0, 0, 0, 0, 0, 0, 2, 2, 2, 2, 1, 1, 1, 1, 3, 3, 3, 3, 3, 3, 3, 3,
+          1, 1, 1, 1, 3, 3, 3, 3, 1, 1, 1, 1, 1, 1, 1, 1, 3, 3, 3, 3};
+      int64_t numPhases = 4;
```
kuhar: This table is the most safety-critical data added by this PR (a single wrong entry produces a model that claims conflict-free for layouts that actually conflict, or vice versa, silently shifting codegen); please cite the specific source. Also please assert `numThreads == 64` (or `<=`?) in the CDNA4 branch (or document the precondition).

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1178,6 +1182,58 @@ std::optional<TargetDetails> getAndroidProfileDetails(StringRef target) {
 // Query functions
 //===----------------------------------------------------------------------===//
 
+constexpr int64_t kLdsBankWidth = 4; // LDS bank width in bytes (AMD)
```
kuhar: IMO both this and the number of threads should be derived from the cdna4 LDS details, not hardcoded separately.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/test/target_attrs.mlir
```diff
@@ -45,6 +45,26 @@ func.func @test_target_wgp_none() attributes {
...
+    subgroup = shuffle|arithmetic, dot = dp4xi8toi32,
+    mma = [<MFMA_F32_16x16x16_F16>, <MFMA_F32_32x32x8_F16>],
+    subgroup_size_choices = [64],
+    max_workgroup_sizes = [1024, 1024, 1024],
+    max_thread_count_per_workgroup = 1024,
+    max_workgroup_memory_bytes = 163840,
+    max_workgroup_counts = [2147483647, 2147483647, 2147483647],
+    dma_sizes = [32, 128],
+    workgroup_memory_bank_count = 64,
+    lds_bank_phase_model = cdna4
```
kuhar: I'm wondering if we need both or if `lds_model = cdna4` would be enough; conceptually, anything lds-related is a property of the hw generation.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
kuhar: We call it shared memory across IREE

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/test/target_attrs.mlir
Muzammiluddin-Syed-ECE: I like this idea, removing

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: What about a FailureOr<std::optional<XorShuffleParams>>

kuhar: 🤮

kuhar: (avoid double-nullable types as much as possible)

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: im not entirely convinced that this is an adjustment we should be making on the swizzle side rather than the DMA side, but i can leave that as a TODO for future work.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
krzysz00: But also this does seem like a good usecase for that sort of thing?

kuhar: I'd prefer if XorShuffleParams had a state that indicated 'no shuffle'

kuhar: You could also have it opt into llvm-style RTTI and have a different types for xor / pad / no avoidance.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: yes the mlir::delinearize function, however it doesn't work well with cases where strides aren't monotonically decreasing. In our case, a layout may have strides such that the inner most dimension is actually listed first. Such as in the scaled gemm intrinsic LHS.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: I can return an identity swizzle such as xor<16,16> which gets optimized out so it requires no special handling.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
Muzammiluddin-Syed-ECE: Agreed on subgroup size but for something like LDS bank width, this would require extracting LDS total size, LDS num banks and LDS num rows. We currently don't track the number of rows of LDS per arch. But of the two, LDS bank width and LDS num rows, I would prefer hardcoding bank width and deriving num rows if need be.

But until we have a reason to have this number be anything other than 4 bytes, it feels anti-YAGNI to create another attribute to track this number per arch for a future where this assumption may change.

kuhar: You can have a default config that only knows the bank width

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
kuhar: What is the interpretation of `xor<0, 0>`? I think we should have a canonical identity state and a helper function like `isIdentity()`

Muzammiluddin-Syed-ECE:
Splitting this into two PR's to make it easier to review: https://github.com/iree-org/iree/pull/24471

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
Muzammiluddin-Syed-ECE: Yes +1 to what Krzysztof said since we may have the same phase model for other archs

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
Muzammiluddin-Syed-ECE: Updated

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.td
Muzammiluddin-Syed-ECE: Went with shared mem

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: That's a nonsensical swizzle. <0,0> means swizzling over rows of length 0 elements, split into groups of 0 elements which is a division by 0. The next two most obvious options might be <1,1> or <int_max, int_max>. Personally, I think doing <access width, access width> seems most reasonable.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
Muzammiluddin-Syed-ECE: done in https://github.com/iree-org/iree/pull/24471

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/test/target_attrs.mlir
Muzammiluddin-Syed-ECE: addressed in https://github.com/iree-org/iree/pull/24471

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
Muzammiluddin-Syed-ECE: addressed in https://github.com/iree-org/iree/pull/24471

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
Muzammiluddin-Syed-ECE: Added a loop that decomposes wider reads into power of 2 subreads. So, a read of width 160 bits would be treated as a 128 bit read than a 32 bit read and both these reads are checked for bank conflicts.

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Muzammiluddin-Syed-ECE: done

Muzammiluddin-Syed-ECE:
>  There is no targeted test for getPhaseGroups (the CDNA4 4-phase grouping is a 64-entry magic table -- exactly the kind of thing that wants a unit/lit test that pins the grouping)

Added gtest unit tests for the magic table

> (a) a parse/print test for the new lds_bank_phase_model attribute on a non-CDNA4 target (you have a CDNA4 case via target_attrs.mlir)

Addressed in https://github.com/iree-org/iree/pull/24471

> (b) a targeted lowering test where the iterative search picks an XOR swizzle that the old tuned table did not (and vice versa, to catch regressions if the table is ever edited)

Added a test to check the swizzle produced for the intrinsic MFMA_F32_32x32x16_BF16 which was not in the old tuned table. Verified the swizzle removes bank conflicts locally. 

The existing scaled mfma swizzle tests suffice for tests whuch were in the old tuned table.

> (c) a negative test verifying that on a target with phase model None, getXorShuffleParams now returns failure (it does, via hasNoBankConflicts returning failure)

The only path that uses xor shuffles is currently on gfx950, so the negative consists of overwriting the lowering config with phase model none and it does indeed return failure but not via hasNoBankConflicts but due to a prior early exit.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Yu-Zhewen: I do see the benefit that the caller of `getXorShuffleParams` does want to distinguish "no swizzle needed" from "search failed".

But could we short-circuit inside `getXorShuffleAttr` so we don't need to materialize an identity swizzle and then optimize it away?
```
std::optional<Attribute>getXorShuffleAttr() {
  std::optional<XorShuffleParams> xorShuffleParams = getXorShuffleParams();
  if (!xorShuffleParams) {
    return std::nullopt;
  }
  
  if (xorShuffleParams->rowElems == xorShuffleParams->accessElems) {
    return baseConfigAttr;
  }
  
  return swizzleAttr;
}

```

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
Yu-Zhewen: Out of curiosity I tried dropping the `if (!transposedLhs) / if (transposedRhs)` guards entirely and passing `transposedLhs / transposedRhs` directly to `getXorShuffleAttr`, same as the mxfp4 case.

For the previously-skipped cases (`transposedLhs/!transposedRhs`), I found `getXorShuffleAttr` returns the identity `xor_shuffle<8, 8>`, so I think dropping the guards would make sense

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
Muzammiluddin-Syed-ECE: I like that idea!

Muzammiluddin-Syed-ECE:
Resolves https://github.com/iree-org/iree/issues/23329

PR #24431, opened 2026-05-11
[Codegen] Preemptively duplicate operations before tile-size analysis

sommerlukas:
To avoid cross-pollution from easy-to-duplicate operations such as `linalg.fill` in tile size analysis, we treat them as duplicatable. So far, the materialization of the vector tile size analysis duplicated values on demand, based on the analysis result.

This approach has two main downsides: 
- The analysis code needs to account for duplicatable operations with special logic. This logic started to become more complicated when I tried to fix an issue with region control flow locally.
- It can be difficult to trace from the duplicatable operation to the operation that actually uses that value with a specific tile size in the presence of control-flow as pointed out [here](https://github.com/iree-org/iree/pull/24246#discussion_r3187746878). Locally handling `scf.for` introduced even more complexity. 

Therefore, I decided to change the duplication to eager duplication: Before the analysis, every duplicatable operation is duplicated for each of its uses. Analysis can then run without special logic for duplicatable values. Afterwards, operations are deduplicated as much as possible again. Operations that are duplicates of the same original value and have the same tile size will be deduplicated to a single operation. 

This approach avoids both downsides described above. The cost for the increased IR size is completely local to the pass itself. 

This contributes towards https://github.com/iree-org/iree/issues/24221.

Assisted-by: Codex

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
@@ -721,10 +721,10 @@ static void setSplitGroup(Operation *op, int64_t groupId) {
 }
 
 // Duplicate all duplicatable operations. If a duplicatable operation has more
-// than one user, it will be duplicated (via clone) and each user will receiver
+// than one user, it will be duplicated (via clone) and each user will receive
 // their own copy of the duplicatable operation. By attaching a split-group to
-// the clones of the same operation as (discardable) attribute, we can later on
-// deduplicated clones of the same operation that received the same tile size
+// the clones of the same operation as (discardable) attribute, we can later
+// deduplicate clones of the same operation that received the same tile size
 // from the analysis.
 static void splitDuplicatableTensorProducers(FunctionOpInterface funcOp) {
   int64_t nextGroupId = 0;
@@ -773,7 +773,13 @@ static void dedupSplitGroup(ArrayRef<Operation *> group) {
     auto tileSizesAttr = dyn_cast_or_null<DenseI64ArrayAttr>(
         op->getAttr(kVectorTileSizesAttrName));
     if (!tileSizesAttr) {
-      continue;
+      // Deduplicate operations that were not assigned a tile size during
+      // analysis. It is safe to assume a single result with tensor type here,
+      // as all operations that end up here must match
+      // `isDuplicatableTensorProducer` earlier.
+      TensorType resTy = cast<TensorType>(op->getResult(0).getType());
+      SmallVector<int64_t> minusOne(resTy.getRank(), -1);
+      tileSizesAttr = DenseI64ArrayAttr::get(op->getContext(), minusOne);
     }
     auto it = llvm::find_if(representatives, [&](auto &entry) {
       return entry.first == tileSizesAttr;
@@ -822,7 +828,7 @@ public:
     FunctionOpInterface funcOp = getOperation();
 
     // Duplicate all duplicatable operations before running the analysis. This
-    // avoid cross-polluting the result of unrelated operations via CSE'd edges
+    // avoids cross-polluting the result of unrelated operations via CSE'd edges
     // such as linalg.fill used as DPS init. Duplicating preemptively instead of
     // on-demand after the analysis has two advantages: It makes the analysis
     // code simpler, as we don't need special case handling for duplicatable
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -694,88 +665,147 @@ static TileSizes getIm2colTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
...
-        getUseOperandTileSizes(use, solver);
-    if (!maybeTileSizes || !maybeTileSizes->isDefined()) {
+static void setSplitGroup(Operation *op, int64_t groupId) {
+  op->setAttr(
+      kSplitGroupAttrName,
+      IntegerAttr::get(IntegerType::get(op->getContext(), 64), groupId));
+}
+
+// Duplicate all duplicatable operations. If a duplicatable operation has more
+// than one user, it will be duplicated (via clone) and each user will receiver
```
amd-eochoalo:
```suggestion
// than one user, it will be duplicated (via clone) and each user will receive
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -694,88 +665,147 @@ static TileSizes getIm2colTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
...
+// their own copy of the duplicatable operation. By attaching a split-group to
+// the clones of the same operation as (discardable) attribute, we can later on
+// deduplicated clones of the same operation that received the same tile size
```
amd-eochoalo:
```suggestion
// the clones of the same operation as (discardable) attribute, we can later
// deduplicate clones of the same operation that received the same tile size
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -791,6 +821,16 @@ class MaterializeVectorTileSizesPass final
   void runOnOperation() override {
     FunctionOpInterface funcOp = getOperation();
 
+    // Duplicate all duplicatable operations before running the analysis. This
+    // avoid cross-polluting the result of unrelated operations via CSE'd edges
```
amd-eochoalo:
```suggestion
    // avoids cross-polluting the result of unrelated operations via CSE'd edges
```

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
```diff
@@ -694,88 +665,147 @@ static TileSizes getIm2colTileSizes(IREE::LinalgExt::Im2colOp im2colOp,
...
+  SmallVector<std::pair<DenseI64ArrayAttr, Operation *>> representatives;
+  for (Operation *op : group) {
+    if (!op || !op->getBlock()) {
+      continue;
+    }
+    auto tileSizesAttr = dyn_cast_or_null<DenseI64ArrayAttr>(
+        op->getAttr(kVectorTileSizesAttrName));
+    if (!tileSizesAttr) {
+      continue;
+    }
```
amd-eochoalo, line 783: I know that duplicated operations that survive this pass will be cleaned by CSE, but I was wondering if it would be possible to deduplicated them here instead?

amd-eochoalo (approved):

compiler/src/iree/compiler/Codegen/Common/MaterializeVectorTileSizes.cpp
sommerlukas: Do you specifically mean the ones without a tile size attribute? 

Because all the others will be deduplicated here as much as possible.

amd-eochoalo: > Do you specifically mean the ones without a tile size attribute?

Yes.

sommerlukas: Ok, I've added deduplication for that case as well and a test.

amd-eochoalo: Thanks! Looks great!

PR #24432, opened 2026-05-11
[Codegen][GPU] Relax 32-bit reduction/arg_compare ceiling for VectorDistribute

bangtianliu:
The upstream PR  [`llvm-project#136320`](https://github.com/llvm/llvm-project/pull/136320) removed the 32-bit constraint from `GPUToROCDL`'s `gpu.shuffle` lowering by decomposing wider values into 32-bit chunks via  LLVM::decomposeValue` / `LLVM::composeValue`. With that lowering in place, the IREE-side 32-bit ceiling was unnecessarily restrictive.     

This PR relaxes the 32-bit element-bitwidth ceiling that prevented `vector.multi_reduction` and `iree_linalg_ext.arg_compare` from compiling through the VectorDistribute pipeline on AMDGPU.  After this PR, AMD targets accept power-of-two element widths above 32 bits.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -929,16 +929,17 @@ static VectorValue reshapeFlatToTarget(RewriterBase &rewriter, Location loc,
 }
 
 static LogicalResult checkBitwidthForShuffle(Operation *op, Type type,
-                                             int64_t maxBitsPerShuffle,
                                              StringRef typeName,
                                              PatternRewriter &rewriter) {
   unsigned bitwidth = type.getIntOrFloatBitWidth();
-  if (bitwidth > maxBitsPerShuffle) {
-    return rewriter.notifyMatchFailure(
-        op, llvm::formatv("{0} bitwidth {1} greater than maxBitsPerShuffle {2}",
-                          typeName, bitwidth, maxBitsPerShuffle));
+  if (targetSupportsShuffleBitwidth(getGPUTargetAttr(op), bitwidth)) {
+    return success();
   }
-  return success();
+  return rewriter.notifyMatchFailure(
+      op, llvm::formatv("{0} bitwidth {1} exceeds native shuffle width {2}; "
+                        "wider widths require a target with a decomposing "
+                        "gpu.shuffle lowering",
+                        typeName, bitwidth, kShuffleNativeBits));
 }
 
 /// Creates an equality comparison operation for the given values.
@@ -1194,9 +1195,9 @@ struct DistributeMultiReduction final
   using MaskedOpDistributionPattern::MaskedOpDistributionPattern;
 
   DistributeMultiReduction(MLIRContext *context, int64_t subgroupSize,
-                           int64_t maxBitsPerShuffle, int64_t benefit = 1)
+                           int64_t benefit = 1)
       : MaskedOpDistributionPattern(context, benefit),
-        subgroupSize(subgroupSize), maxBitsPerShuffle(maxBitsPerShuffle) {}
+        subgroupSize(subgroupSize) {}
 
   LogicalResult
   matchAndRewrite(vector::MultiDimReductionOp multiReduceOp,
@@ -1218,8 +1219,8 @@ struct DistributeMultiReduction final
     }
 
     Type elemTy = srcVector.getType().getElementType();
-    if (failed(checkBitwidthForShuffle(multiReduceOp, elemTy, maxBitsPerShuffle,
-                                       "element", rewriter))) {
+    if (failed(checkBitwidthForShuffle(multiReduceOp, elemTy, "element",
+                                       rewriter))) {
       return failure();
     }
 
@@ -1517,7 +1518,6 @@ struct DistributeMultiReduction final
   }
 
   int64_t subgroupSize;
-  int64_t maxBitsPerShuffle;
 };
 
 /// Distributes `iree_vector_ext.arg_compare` ops with nested layouts.
@@ -1527,9 +1527,9 @@ struct DistributeArgCompare final
     : MaskedOpDistributionPattern<IREE::VectorExt::ArgCompareOp> {
 
   DistributeArgCompare(MLIRContext *context, int64_t subgroupSize,
-                       int64_t maxBitsPerShuffle, int64_t benefit = 1)
+                       int64_t benefit = 1)
       : MaskedOpDistributionPattern(context, benefit),
-        subgroupSize(subgroupSize), maxBitsPerShuffle(maxBitsPerShuffle) {}
+        subgroupSize(subgroupSize) {}
 
   LogicalResult
   matchAndRewrite(IREE::VectorExt::ArgCompareOp argCompareOp,
@@ -1571,15 +1571,14 @@ struct DistributeArgCompare final
     }
 
     Type elemTy = inputValue.getType().getElementType();
-    if (failed(checkBitwidthForShuffle(argCompareOp, elemTy, maxBitsPerShuffle,
-                                       "element", rewriter))) {
+    if (failed(checkBitwidthForShuffle(argCompareOp, elemTy, "element",
+                                       rewriter))) {
       return failure();
     }
 
-    // No bitwidth check on the index type: the index is only forwarded from
-    // the winning lane via `gpu.shuffle idx`, which handles wider types (i64).
-    // TODO(Bangtian): On AMD, ROCDL decomposes 64-bit shuffles into 32-bit
-    // pairs. Consider dropping the value bitwidth check above too.
+    // No bitwidth check on the index type: `gpu.shuffle idx` forwards the
+    // winning lane's index and handles wider types (i64). The value-side
+    // check above already honors the AMDGPU shuffle whitelist.
 
     // Only explicit index mode; iota indices are materialized earlier.
     if (!inputIndex) {
@@ -2176,7 +2175,6 @@ private:
   }
 
   int64_t subgroupSize;
-  int64_t maxBitsPerShuffle;
 };
 
 /// Broadcast every element of |source| from the last lane in a scan cluster
@@ -2377,10 +2375,15 @@ struct DistributeScan final : OpDistributionPattern<vector::ScanOp> {
     bool needsCrossThread = srcLayout.getThreadTile()[scanDim] > 1;
     bool needsCrossSubgroup = srcLayout.getSubgroupTile()[scanDim] > 1;
     Type elemTy = source.getType().getElementType();
-    if ((needsCrossThread || needsCrossSubgroup) &&
-        failed(checkBitwidthForShuffle(scanOp, elemTy, maxBitsPerShuffle,
-                                       "element", rewriter))) {
-      return failure();
+    if (needsCrossThread || needsCrossSubgroup) {
+      unsigned bitwidth = elemTy.getIntOrFloatBitWidth();
+      if (bitwidth > maxBitsPerShuffle) {
+        return rewriter.notifyMatchFailure(
+            scanOp,
+            llvm::formatv("element bitwidth {0} greater than maxBitsPerShuffle "
+                          "{1}",
+                          bitwidth, maxBitsPerShuffle));
+      }
     }
 
     // Reject multi-subgroup for now.
@@ -3426,10 +3429,8 @@ void IREE::VectorExt::populateNestedLayoutDistributionPatterns(
                                         subgroupSize, workgroupSize);
   patterns.add<DistributeBroadcast, DistributeTranspose, DistributeShapeCast>(
       patterns.getContext());
-  patterns.add<DistributeMultiReduction>(patterns.getContext(), subgroupSize,
-                                         maxBitsPerShuffle);
-  patterns.add<DistributeArgCompare>(patterns.getContext(), subgroupSize,
-                                     maxBitsPerShuffle);
+  patterns.add<DistributeMultiReduction>(patterns.getContext(), subgroupSize);
+  patterns.add<DistributeArgCompare>(patterns.getContext(), subgroupSize);
   patterns.add<DistributeScan>(patterns.getContext(), subgroupSize,
                                maxBitsPerShuffle);
   patterns.add<DistributeContract>(patterns.getContext());
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
@@ -1576,9 +1576,10 @@ struct DistributeArgCompare final
       return failure();
     }
 
-    // No bitwidth check on the index type: `gpu.shuffle idx` forwards the
-    // winning lane's index and handles wider types (i64). The value-side
-    // check above already honors the AMDGPU shuffle whitelist.
+    // No bitwidth check on the index payload: it is broadcast from the
+    // winning lane via `gpu.shuffle idx`, and the AMDGPU lowering decomposes
+    // wide values (e.g. i64) into i32 chunks. The value-side check above
+    // already gates element types.
 
     // Only explicit index mode; iota indices are materialized earlier.
     if (!inputIndex) {
@@ -2375,6 +2376,11 @@ struct DistributeScan final : OpDistributionPattern<vector::ScanOp> {
     bool needsCrossThread = srcLayout.getThreadTile()[scanDim] > 1;
     bool needsCrossSubgroup = srcLayout.getSubgroupTile()[scanDim] > 1;
     Type elemTy = source.getType().getElementType();
+    // Intentional carve-out: unlike DistributeMultiReduction and
+    // DistributeArgCompare (which use targetSupportsShuffleBitwidth and accept
+    // wider widths on AMDGPU via a decomposing gpu.shuffle lowering), scan is
+    // kept on the 32-bit ceiling until the wider-width scan path has been
+    // separately validated.
     if (needsCrossThread || needsCrossSubgroup) {
       unsigned bitwidth = elemTy.getIntOrFloatBitWidth();
       if (bitwidth > maxBitsPerShuffle) {
@@ -3429,8 +3435,8 @@ void IREE::VectorExt::populateNestedLayoutDistributionPatterns(
                                         subgroupSize, workgroupSize);
   patterns.add<DistributeBroadcast, DistributeTranspose, DistributeShapeCast>(
       patterns.getContext());
-  patterns.add<DistributeMultiReduction>(patterns.getContext(), subgroupSize);
-  patterns.add<DistributeArgCompare>(patterns.getContext(), subgroupSize);
+  patterns.add<DistributeMultiReduction, DistributeArgCompare>(
+      patterns.getContext(), subgroupSize);
   patterns.add<DistributeScan>(patterns.getContext(), subgroupSize,
                                maxBitsPerShuffle);
   patterns.add<DistributeContract>(patterns.getContext());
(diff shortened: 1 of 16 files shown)


krzysz00 (commented):
As a high-level note, I think the answer to "max bits per shuffle" is "none" or "infinity", not 64

bangtianliu:
> As a high-level note, I think the answer to "max bits per shuffle" is "none" or "infinity", not 64

True, let me have a try at this PR

compiler/src/iree/compiler/Codegen/LLVMGPU/test/NVVM/config_vector_distribute_sm80.mlir
```diff
@@ -60,3 +60,37 @@ func.func @matmul_256x256x256_f16_f16() {
 // CHECK-LABEL: func.func @matmul_256x256x256_f16_f16()
 // CHECK:       linalg.matmul {{.*}}lowering_config = #iree_gpu.lowering_config
 // CHECK-SAME:                mma_kind = #iree_gpu.mma_layout<NV_MMA_SYNC_F16_16x8x16_F16>
+
+// -----
+
+// f64 reduction on sm_80: NV presets do not advertise wide_subgroup_shuffle,
+// so the gate rejects 64-bit and the dispatch must NOT pick VectorDistribute.
+// Pins the AMD-only scope of the LLVM PR #136320 relaxation.
```
krzysz00: Someone should ping the Nvidia people for if this is a real restriction or of they  just need to update their patterns like AMD did

tests/e2e/regression/reduction_64bit.mlir
```diff
@@ -0,0 +1,76 @@
+// 64-bit reduction and arg_compare on AMDGPU through VectorDistribute.
+//
+// Before this work the kernel-config bitwidth gate rejected element types
```
krzysz00: No need to put the history here?

krzysz00 (approved):
lgtm overall, though I wonder if we can't get rid of the parameter by fixing Nvidia's lowering instead

compiler/src/iree/compiler/Codegen/LLVMGPU/test/NVVM/config_vector_distribute_sm80.mlir
bangtianliu: Yes, one upstream PR is needed from the NV side to mirror what https://github.com/llvm/llvm-project/pull/136320 did for ROCDL. I will file an upstream LLVM issue to ping NV people there. 

Updated the test comment.

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/test/target_attrs.mlir
```diff
@@ -29,6 +29,25 @@ func.func @test_target_wgp() attributes {
   >
 } { return }
 
+// CHECK-LABEL: func.func @test_target_wgp_wide_subgroup_shuffle()
+func.func @test_target_wgp_wide_subgroup_shuffle() attributes {
+  // CHECK: #iree_gpu.target_wgp<
+  // CHECK-SAME: wide_subgroup_shuffle = true,
+  // CHECK-SAME: workgroup_memory_bank_count = 32>
```
sommerlukas: Nit: Do we need this check for this test?

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.h
```diff
@@ -330,6 +330,25 @@ IREE::GPU::TargetAttr getGPUTargetAttr(MLIRContext *context,
...
 
+/// Native element bitwidth of `gpu.shuffle` / `gpu.subgroup_reduce` on
+/// supported targets. Widths up to this value are universally supported;
+/// widths above require wide-subgroup-shuffle capability.
+inline constexpr unsigned kNarrowShuffleBits = 32;
+
+/// Whether |wgp| supports `gpu.shuffle` / `gpu.subgroup_reduce` on element
+/// types wider than `kNarrowShuffleBits`. Returns false when the field is
+/// unset.
+inline bool getWideSubgroupShuffle(IREE::GPU::TargetWgpAttr wgp) {
```
sommerlukas: Nit: Why is this marked `inline`?

sommerlukas (approved):
LGTM apart from two minor nits.

kuhar (commented):
I'm not sure we need this new target env entry: IMO every target can be made to support wide subgroup shuffles / reductions, it's just a matter of decomposing them to supported scalars. Instead, I'd prefer to make the lowering work for the remaining targets and then enable unconditionally. WDYT?

kuhar:
> lgtm overall, though I wonder if we can't get rid of the parameter by fixing Nvidia's lowering instead

We should either do that or gate this on target not being nvidia. I'm concerned we keep growing the set of subgroup features...

kuhar (changes requested):

bangtianliu:
> I'm not sure we need this new target env entry: IMO every target can be made to support wide subgroup shuffles / reductions, it's just a matter of decomposing them to supported scalars. Instead, I'd prefer to make the lowering work for the remaining targets and then enable unconditionally. WDYT?

I've filed llvm/llvm-project#197080 to track the upstream NVVM fix, and I am also happy to self-assign and land it there if no one else picks it up. Once that lands and rolls in, I'll drop the new target-env entry here so the bitwidth check goes away unconditionally.  

Two options: land this PR first and remove the entry in a follow-up once the upstream fix is in, or hold this PR until the fix lands. I'm open to either and  let me know which you prefer

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.h
bangtianliu: Moved the body to GPUUtils.cpp and left just the declaration in the header. The inline is gone now.

kuhar:
IMO it's fine to gate on nvidia for the time being without adding any extra attributes.

bangtianliu:
cc @kuhar for another look

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -2377,10 +2375,15 @@ struct DistributeScan final : OpDistributionPattern<vector::ScanOp> {
     bool needsCrossThread = srcLayout.getThreadTile()[scanDim] > 1;
     bool needsCrossSubgroup = srcLayout.getSubgroupTile()[scanDim] > 1;
     Type elemTy = source.getType().getElementType();
-    if ((needsCrossThread || needsCrossSubgroup) &&
-        failed(checkBitwidthForShuffle(scanOp, elemTy, maxBitsPerShuffle,
-                                       "element", rewriter))) {
-      return failure();
+    if (needsCrossThread || needsCrossSubgroup) {
+      unsigned bitwidth = elemTy.getIntOrFloatBitWidth();
+      if (bitwidth > maxBitsPerShuffle) {
```
kuhar, line 2386: DistributeScan is still on the old `maxBitsPerShuffle` path while reduction and arg_compare now use `targetSupportsShuffleBitwidth`. That leaves AMDGPU 64-bit scan rejected even though the same target accepts 64-bit reduction/arg_compare, and it keeps the `maxBitsPerShuffle` API alive only for scan. Please either migrate scan to the new helper and add a gfx942 f64/i64 scan test, or document and test the intentional carve-out.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
```diff
@@ -1323,6 +1323,14 @@ bool targetSupportsGlobalLoadDMA(IREE::GPU::TargetAttr target) {
   return chipset->majorVersion == 9 && chipset->minorVersion >= 5;
 }
 
+bool targetSupportsShuffleBitwidth(IREE::GPU::TargetAttr target,
+                                   unsigned bitwidth) {
+  if (bitwidth <= kShuffleNativeBits) {
+    return true;
+  }
+  return target && target.isAMD();
```
kuhar, line 1331: Is something like i128 supported today?

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ReductionConfigUtils.cpp
```diff
@@ -780,14 +784,28 @@ LogicalResult setReductionConfig(IREE::GPU::TargetAttr target,
...
     return failure();
   }
 
   const std::optional<int64_t> maxLoadBits = wgp.getMaxLoadInstructionBits();
   const unsigned largestLoadSizeInBits = maxLoadBits.value_or(128);
 
+  // Guard divide-by-zero: `threadLoads = largestLoadSizeInBits / *bitWidth`
+  // becomes 0 when the element is wider than a single load instruction,
+  // and is then used as a divisor in the subgroup-divisibility loop below.
+  if (*bitWidth > largestLoadSizeInBits) {
```
kuhar, line 805: No regression test exercises this new divide-by-zero guard. The condition `*bitWidth > largestLoadSizeInBits` only fires when the element is wider than the target's `maxLoadInstructionBits` (128 on gfx942). The new tests in `config_vector_distribute_reduction_gfx942.mlir` go up to i128 (which has `bitWidth == largestLoadSizeInBits`, so `threadLoads = 1` and the loop never divides by zero) -- the actual failure mode that motivated this guard is not tested.

tests/e2e/regression/BUILD.bazel
```diff
@@ -164,9 +164,15 @@ iree_check_single_backend_test_suite(
...
         "split_reduction_using_tiling.mlir",
+        "vector_distribute_64bit_amdgpu.mlir",
     ],
     compiler_flags = [
         "--iree-opt-level=O3",
+        # vector_distribute_64bit_amdgpu.mlir relies on f64 reaching codegen;
+        # the default f64->f32 demotion would silently turn it into the f32
+        # case. Note: this flag is sticky for the whole suite — any future
+        # f64-touching test added here will also skip demotion.
+        "--iree-input-demote-f64-to-f32=false",
```
kuhar, line 195: `--iree-input-demote-f64-to-f32=false` now applies to the entire `check_regression_hip` suite, so unrelated tests also stop exercising the default f64 demotion path. Please put `vector_distribute_64bit_amdgpu.mlir` in its own HIP suite in both BUILD.bazel and CMakeLists.txt, with the demotion flag scoped to that suite only.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -1571,15 +1571,14 @@ struct DistributeArgCompare final
...
+                                       rewriter))) {
       return failure();
     }
 
-    // No bitwidth check on the index type: the index is only forwarded from
-    // the winning lane via `gpu.shuffle idx`, which handles wider types (i64).
-    // TODO(Bangtian): On AMD, ROCDL decomposes 64-bit shuffles into 32-bit
-    // pairs. Consider dropping the value bitwidth check above too.
+    // No bitwidth check on the index type: `gpu.shuffle idx` forwards the
+    // winning lane's index and handles wider types (i64). The value-side
```
kuhar: I don't understand the first sentence, I think the grammar is off.

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
```diff
@@ -3426,10 +3429,8 @@ void IREE::VectorExt::populateNestedLayoutDistributionPatterns(
                                         subgroupSize, workgroupSize);
   patterns.add<DistributeBroadcast, DistributeTranspose, DistributeShapeCast>(
       patterns.getContext());
-  patterns.add<DistributeMultiReduction>(patterns.getContext(), subgroupSize,
-                                         maxBitsPerShuffle);
-  patterns.add<DistributeArgCompare>(patterns.getContext(), subgroupSize,
-                                     maxBitsPerShuffle);
+  patterns.add<DistributeMultiReduction>(patterns.getContext(), subgroupSize);
+  patterns.add<DistributeArgCompare>(patterns.getContext(), subgroupSize);
```
kuhar: This should be

```suggestion
  patterns.add<DistributeMultiReduction, DistributeArgCompare>(patterns.getContext(), subgroupSize);
```

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution_argcompare.mlir
```diff
@@ -1,6 +1,9 @@
-// RUN: iree-opt --iree-transform-dialect-interpreter --split-input-file --canonicalize --cse %s | FileCheck %s
+// RUN: iree-opt --iree-transform-dialect-interpreter --iree-gpu-test-target=gfx942 --split-input-file --canonicalize --cse %s | FileCheck %s
 
-// Tests for DistributeArgCompare pattern with explicit index mode.
+// Tests for DistributeArgCompare pattern with explicit index mode. Runs
+// under `--iree-gpu-test-target=gfx942` so 64-bit-element cases (which
+// rely on AMDGPU's decomposing `gpu.shuffle` lowering) reach the pattern.
+// All other test cases here use <=32-bit elements and are target-agnostic.
```
kuhar: Should we move this to rocdl test then?

kuhar (commented):
A few comments

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution_argcompare.mlir
```diff
@@ -564,3 +567,118 @@ builtin.module attributes { transform.with_named_sequence } {
...
+    %top_level_func = transform.structured.match ops{["func.func"]} in %variant_op : (!transform.any_op) -> !transform.any_op
+    transform.iree.test_gpu_vector_distribution %top_level_func : !transform.any_op
+    transform.yield
+  }
+}
+
+// -----
+
+// i64 *index* path: exempt from the bitwidth check (`gpu.shuffle idx`
+// handles wider types). Pins that behavior under the new plumbing.
```
kuhar: Don't refer to previous / new code -- readers won't know what the timeframe is. Always refer to the current state of the codebase.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution_multi_reduce.mlir
```diff
@@ -1,4 +1,8 @@
-// RUN: iree-opt --iree-transform-dialect-interpreter --split-input-file --canonicalize -mlir-print-local-scope --cse %s | FileCheck %s
+// RUN: iree-opt --iree-transform-dialect-interpreter --iree-gpu-test-target=gfx942 --split-input-file --canonicalize -mlir-print-local-scope --cse %s | FileCheck %s
+
+// Runs under `--iree-gpu-test-target=gfx942` so 64-bit-element cases (which
```
kuhar: We may also want to move these under the rocdl test dir

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution_multi_reduce.mlir
```diff
@@ -249,3 +253,75 @@ builtin.module attributes { transform.with_named_sequence } {
...
+    transform.iree.test_gpu_vector_distribution %top_level_func : !transform.any_op
+    transform.yield
+  }
+}
+
+// CHECK-LABEL: func @reduce_f64_dim1
+// Local reduction on the distributed batch shape.
+// CHECK: vector.multi_reduction <maximumf>, %{{.*}}, %{{.*}} [1, 3, 5] : vector<2x2x1x1x1x4xf64> to vector<2x1x1xf64>
+// Global reduction across threads on the 64-bit element type.
+// CHECK: gpu.subgroup_reduce maximumf %{{.*}} cluster(size = 4, stride = 16) : (f64) -> f64
```
kuhar: what makes this reduction global?

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution_multi_reduce.mlir
```diff
@@ -249,3 +253,75 @@ builtin.module attributes { transform.with_named_sequence } {
...
+    transform.iree.test_gpu_vector_distribution %top_level_func : !transform.any_op
+    transform.yield
+  }
+}
+
+// CHECK-LABEL: func @reduce_i64_dim1
+// Local reduction on the distributed batch shape.
+// CHECK: vector.multi_reduction <add>, %{{.*}}, %{{.*}} [1, 3, 5] : vector<2x2x1x1x1x4xi64> to vector<2x1x1xi64>
+// Global reduction across threads on the 64-bit element type.
+// CHECK: gpu.subgroup_reduce add %{{.*}} cluster(size = 4, stride = 16) : (i64) -> i64
```
kuhar: also here

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.h
```diff
@@ -24,6 +24,12 @@ namespace mlir::iree_compiler {
 static constexpr int32_t kNumGPUDims = 3;
 static constexpr int32_t kWarpSize = 32;
 
+// `gpu.shuffle` natively supports up to 32-bit element types. Wider values
```
kuhar: This is not true, `gpu.shuffle` is target-agnostic doesn't know what the native bitwidths are.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.h
```diff
@@ -24,6 +24,12 @@ namespace mlir::iree_compiler {
...
+// must be decomposed by the backend's lowering, which today only ROCDL does
+// (cf. llvm/llvm-project#136320). NVVM tracking is at
+// llvm/llvm-project#197080. See `targetSupportsShuffleBitwidth`.
+static constexpr unsigned kShuffleNativeBits = 32;
```
kuhar: static is redundant here, `constexpr` variables in the namespace scope are always static

compiler/src/iree/compiler/Codegen/LLVMGPU/test/NVVM/config_vector_distribute_sm80.mlir
kuhar: we can implement it too, we even have shared nvidia systems we can test on

bangtianliu: I filed an upstream issue here: https://github.com/llvm/llvm-project/issues/197080. Sure, I will self-assign and work on it.

compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_nested_layout_vector_distribution_multi_reduce.mlir
bangtianliu: The reduction is a cluster(size=4, stride=16) subgroup reduce, "global" is misleading

compiler/src/iree/compiler/Codegen/Common/GPU/GPUNestedLayoutDistributionPatterns.cpp
bangtianliu: Documented the carve-out and added a negative test. Leaving the scan migration itself to @sommerlukas, who has it on his implementation plan.

compiler/src/iree/compiler/Codegen/Utils/GPUUtils.cpp
bangtianliu: Yes, added a test for i128.

kuhar (approved):

PR #24434, opened 2026-05-11
[Bazel] Add missing loads and glob support for bazel 8+

cerisier:
This PR adds a bunch of missing load and glob amends to allow parts of iree to be consumed by Bazel 8+.

It is purely additive and doesn't change migrate IREE to bazel 8+, it just adds some support for bazel 8+ for some parts of the graph while keeping compatibility with official iree bazelversion (7).


llvm-external-projects/iree-dialects/BUILD.bazel
```diff
@@ -7,9 +7,9 @@ package(
     licenses = ["notice"],
 )
 
-exports_files(glob(["include/iree-dialects/Dialect/Input/*.td"]))
+exports_files(glob(["include/iree-dialects/Dialect/Input/*.td"], allow_empty = True))
```
kuhar: Why do we need these `allow_empty` throughout this file?

cerisier: I didn't want introduce any kind of breaking change but that's because in this package, `include/iree-dialects/Dialect/Input` directory  doesn't exist, so all those glob resolve to nothing.

kuhar: Should we drop the whole `exports_files` then?

amd-eochoalo (approved):
I think this looks ok.

There are a couple of places where 

```
load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
```

is used and it would be nice to have them all use

```
load("@rules_cc//cc:defs.bzl", "cc_binary")
```

I also think that dropping exports_files for "include/iree-dialects/Dialect/Input/*.td" is a good thing to do.

cerisier:
I'll remove the exports_files.

As for the `defs.bzl` vs `cc_binary.bzl`, my bad !

llvm-external-projects/iree-dialects/BUILD.bazel
```diff
@@ -252,7 +239,7 @@ cc_library(
     name = "IREEDialectsTest",
     srcs = glob([
         "test/lib/**/*.cpp",
-    ]),
+    ], allow_empty = True),
```
kuhar: Should we drop this if there are no tests?

cerisier:
I reverted the broader clean and only removed the non existing Input/*.td glob.

cerisier:
CI Failure was unrelated. I updated the branch with latest from main@HEAD and tests passes locally.

kuhar:
> CI Failure was unrelated. I updated the branch with latest from main@HEAD and tests passes locally.

lint failures look actionable?

cerisier:
Addressed.

cerisier:
Feel like those failures are unrelated?

PR #24435, opened 2026-05-11
[Codegen][GPU] Avoid generic TileAndFuse promotion for convolutions.

### LINKED ISSUE #24414: resnet50-v1-12 / resnet50-v2-7 produce all-NaN output on hip/gfx1100 when batch dim is dynamic
### Summary

Compiling and running `resnet50-v1-12.onnx` (and `resnet50-v2-7.onnx`) for `--iree-hal-target-device=hip --iree-rocm-target=gfx1100` produces an output tensor that is entirely NaN. The reference output from `onnxruntime` is fine.

The bug is sensitive to whether the **batch dimension is static or dynamic** at MLIR-import time. With static batch (`[1,3,224,224]`) the output matches ORT to ~5e-6. With dynamic batch (`[?,3,224,224]`), even when the dim is resolved to 1 at runtime, every output element is NaN.

### Repro

Requires `iree-base-compiler` and `iree-base-runtime` (any recent nightly):
`pip install --find-links https://iree.dev/pip-release-links.html --pre --upgrade iree-base-compiler iree-base-runtime`.

```bash
mkdir -p /tmp/resnet50-nan && cd /tmp/resnet50-nan

# 1. Download the original ONNX (dynamic batch dim).
wget -q https://huggingface.co/onnxmodelzoo/legacy_models/resolve/main/validated/vision/classification/resnet/model/resnet50-v1-12.onnx

# 2. Deterministic input (matches the iree-test-suites fixture).
python3 -c "import numpy as np; np.random.default_rng(0).random((1,3,224,224), dtype=np.float32).tofile('input.bin')"

# 3. Import + compile + run.
iree-import-onnx resnet50-v1-12.onnx --opset-version 17 -o resnet50.mlir
iree-compile resnet50.mlir --iree-hal-target-device=hip --iree-rocm-target=gfx1100 -o resnet50.vmfb
iree-run-module --module=resnet50.vmfb --device=hip \
    --input=1x3x224x224xf32=@input.bin --output=@out.bin

# 4. Check the output.
python3 -c "import numpy as np; y=np.fromfile('out.bin', dtype=np.float32); print(f'NaN: {np.isnan(y).sum()}/{y.size}')"
```

Expected: `NaN: 0/1000`.
Actual: `NaN: 1000/1000`.

The imported MLIR's entry signature is:

```mlir
func.func @mxnet_converted_model(
    %arg0: !torch.vtensor<[?,3,224,224],f32>) -> !torch.vtensor<[?,1000],f32>
```

If you statically pin the ONNX batch dim to 1 *before* `iree-import-onnx` (so the MLIR becomes `[1,3,224,224]`), the same compile + run produces a correct output (`NaN: 0 / 1000`, diff ~5e-6).

The same behavior reproduces with `resnet50-v2-7.onnx` from the same HuggingFace path.

### Environment

- Hardware: AMD Radeon PRO W7900 (gfx1100, RDNA3)
- ROCm: 7.x
- IREE commit: `b7ee744f` (head of PR #24394's CI build)

### Originally observed in

CI job `Test ONNX / test_onnx_models :: amdgpu_hip_rdna3` in workflow run [25530086833](https://github.com/iree-org/iree/actions/runs/25530086833). Confirmed independent of that PR — its only changes are to the iree-run-module float-comparison logic, which cannot affect the model's actual output.

### Triage hints

- Same input bytes, same ORT reference, same hardware, same compiler/runtime binaries — only the input MLIR differs (static vs dynamic batch). The divergence is in the lowering/codegen path that takes the dynamic batch dim.
- All 1000 output elements being NaN suggests garbage produced very early in the network (first conv or batchnorm) and then propagated. `iree-compile --iree-hal-dump-executable-sources-to=...` plus isolating the first few dispatches should narrow it.
- Worth checking other model-zoo models with dynamic batch on gfx1100 for the same symptom.

hanhanW:
Generic TileAndFuse promotion can attach promote_operands to convolution ops that did not go through a convolution-specific MMA config path. Those paths are responsible for selecting an MMA schedule and attaching mma_kind.

Skip generic contraction promotion for convolution ops so direct conv and IGEMM configs remain the only promoted convolution paths.

This fixes NaN issues for ResNet ONNX models.

Fixes https://github.com/iree-org/iree/issues/24414


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/ConfigUtils.cpp
```diff
@@ -1604,7 +1604,13 @@ LogicalResult setTileAndFuseLoweringConfig(IREE::GPU::TargetAttr target,
...
   };
 
-  if (isNonMatvecContraction(op)) {
+  auto linalgOp = dyn_cast<linalg::LinalgOp>(op);
+  bool isConvolution = linalgOp && linalg::isaConvolutionOpInterface(linalgOp);
+  // This generic promotion is meant for matmul-like contractions. Convolutions
+  // need their convolution-specific config path, which also attaches an
+  // `mma_kind` so filter loops are tiled to unit size before conversion to a
+  // matmul-like op.
+  if (isNonMatvecContraction(op) && !isConvolution) {
```
Max191: This happens because `isNonMatvecContraction` only uses `inferConvolutionDims`, which IIRC is not as strict as `isaContractionOpInterface`. The convolution dim inference will just ignore non-AffineDimExprs, so sometimes convolutions can return valid contraction dims. Could you try updating the `isNonMatvecContraction` function to also check that `isaContractionOpInterface` succeeds? That should fix the same issue, and is less confusing IMO because I would think that `isNonMatvecContraction` should return false for a convolution.

compiler/src/iree/compiler/Codegen/LLVMGPU/test/ROCDL/config_tile_and_fuse_convolution.mlir
```diff
@@ -0,0 +1,28 @@
...
+#output_map = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d1, d2, d3)>
+
+func.func @generic_dynamic_nchw_fchw_conv(
+    %input: tensor<?x3x230x230xf32>, %filter: tensor<64x3x7x7xf32>,
+    %init: tensor<?x64x112x112xf32>) -> tensor<?x64x112x112xf32> {
+  // CHECK-LABEL: func.func @generic_dynamic_nchw_fchw_conv
+  // CHECK:       #iree_codegen.translation_info<pipeline = #iree_gpu.pipeline<TileAndFuse>
+  // CHECK:       linalg.generic
+  // CHECK-NOT:   promote_operands
+  // CHECK:       ^bb0
```
hanhanW, line 16: Note that the second CHECK is needed because `CHECK-NOT` is scoped between two positive matches. Without the trailing positive CHECK, the CHECK-NOT has an unclear/open-ended scope.

Max191 (approved):
Thanks for the fix!

hanhanW:
Both updated jobs are passing, so I'm going to land the PR:
- https://github.com/iree-org/iree/actions/runs/25695068192/job/75442581509?pr=24435
- https://github.com/iree-org/iree/actions/runs/25695068192/job/75442581421?pr=24435

PR #24134, opened 2026-04-15
Unroll transfer read, write, scatter, gather.

amd-eochoalo:
Adds transfer_read, transfer_write, transfer_gather and transfer_scatter unrolling patterns to LLVMGPULegalizeNDVectorsPass.

References: https://github.com/llvm/llvm-project/pull/192700

---------
Co-Authored-by:  Claude Opus 4.6 <noreply@anthropic.com>

PR #24439, opened 2026-05-11
Register IndexedAccessOpInterface

amd-eochoalo:
IndexedAccessOpInterface was missing from the set of registered interfaces. Working on #24134 revealed this is missing. I believe it is possible to not use this, but at the moment I don't see why it shouldn't be used.


amd-eochoalo:
Close in favour of #24440

PR #24440, opened 2026-05-11
[Codegen] Register vector `IndexedAccessOpInterface` external models

lialan:
we need to call `vector::registerIndexedAccessOpInterfaceExternalModels`, from `mlir::registerAllDialects` so we ca properly populate index calculation.

This manifested as numeric mismatches in `e2e_batch_matmul_cdna4_coalesced_dma_bf16_rocm_hip` after the LLVM bump in #24409.


krzysz00 (approved):
Approved, though I'll note there are also `nvgpu::` and `gpu::` functions we may wont to call here (probably not the `gpu::` ones - we don't use that MMA infrastructure - but `nvgpu::` yes)

PR #24442, opened 2026-05-12
[LLVMGPU] Handle single-workgroup scatter dispatches

YuWei-CH:
Problem:
- A scatter like `tensor<32x1x1xf32>` updated by `tensor<11x1x1xf32>` with `tensor<11xi32>` indices only updates indexed dimensions and has a unit inner slice.
- LLVMGPU scatter configuration previously failed this shape because it required a non-unit inner slice before selecting a lowering strategy.
- The workgroup distribution verifier also treated statically single-iteration workgroup forall loops as multi-workgroup dispatches, which rejected the copy-back pattern used by functional scatter lowering.

Fix:
- Configure this scatter shape with the Distribute pipeline and a tile that covers the full iteration domain, keeping the original-output copy and the update loop in one workgroup.
- Treat statically single-iteration workgroup forall loops as single-workgroup dispatches in the verifier.

Example:
- For `original: tensor<32x1x1xf32>`, `updates: tensor<11x1x1xf32>`, and `indices: tensor<11xi32>`, lower `out[indices[i], 0, 0] = updates[i, 0, 0]` as a single-workgroup scatter instead of failing configuration.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/test/nvvm_pipeline_test.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/test/nvvm_pipeline_test.mlir
@@ -469,3 +469,34 @@ func.func @matmul_f16() attributes {hal.executable.target = #executable_target_c
 // SM80-COUNT-64: nvvm.mma.sync{{.*}}shape = #nvvm.shape<m = 16, n = 8, k = 16>
 //     SM80-NOT:     nvvm.mma.sync
 //         SM80:     llvm.return
+
+// -----
+
+#executable_target_cuda = #hal.executable.target<"cuda", "cuda-nvptx-fb">
+#pipeline_layout = #hal.pipeline.layout<bindings = [
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>
+]>
+func.func @scatter_update_only_indexed_dim() attributes {hal.executable.target = #executable_target_cuda} {
+  %c0 = arith.constant 0 : index
+  %original = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) alignment(64) offset(%c0) flags(ReadOnly) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<32x1x1xf32>>
+  %updates = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) alignment(64) offset(%c0) flags(ReadOnly) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<11x1x1xf32>>
+  %indices = hal.interface.binding.subspan layout(#pipeline_layout) binding(2) alignment(64) offset(%c0) flags(ReadOnly) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<11xi32>>
+  %output = hal.interface.binding.subspan layout(#pipeline_layout) binding(3) alignment(64) offset(%c0) : !iree_tensor_ext.dispatch.tensor<writeonly:tensor<32x1x1xf32>>
+  %original_t = iree_tensor_ext.dispatch.tensor.load %original, offsets = [0, 0, 0], sizes = [32, 1, 1], strides = [1, 1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<32x1x1xf32>> -> tensor<32x1x1xf32>
+  %updates_t = iree_tensor_ext.dispatch.tensor.load %updates, offsets = [0, 0, 0], sizes = [11, 1, 1], strides = [1, 1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<11x1x1xf32>> -> tensor<11x1x1xf32>
+  %indices_t = iree_tensor_ext.dispatch.tensor.load %indices, offsets = [0], sizes = [11], strides = [1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<11xi32>> -> tensor<11xi32>
+  %scatter = iree_linalg_ext.scatter dimension_map = [0] unique_indices(true) ins(%updates_t, %indices_t : tensor<11x1x1xf32>, tensor<11xi32>) outs(%original_t : tensor<32x1x1xf32>) {
+  ^bb0(%update: f32, %old: f32):
+    iree_linalg_ext.yield %update : f32
+  } -> tensor<32x1x1xf32>
+  iree_tensor_ext.dispatch.tensor.store %scatter, %output, offsets = [0, 0, 0], sizes = [32, 1, 1], strides = [1, 1, 1] : tensor<32x1x1xf32> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<32x1x1xf32>>
+  return
+}
+
+//       CHECK-LABEL: llvm.func @scatter_update_only_indexed_dim
+//             CHECK:   llvm.store
+//        SM80-LABEL: llvm.func @scatter_update_only_indexed_dim
+//              SM80:   llvm.store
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/test/nvvm_pipeline_test.mlir
```diff
@@ -469,3 +469,34 @@ func.func @matmul_f16() attributes {hal.executable.target = #executable_target_c
...
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>
+]>
+func.func @scatter_update_only_indexed_dim() attributes {hal.executable.target = #executable_target_cuda} {
+  %c0 = arith.constant 0 : index
+  %original = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) alignment(64) offset(%c0) flags(ReadOnly) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<32x1x1xf32>>
+  %updates = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) alignment(64) offset(%c0) flags(ReadOnly) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<11x1x1xf32>>
+  %indices = hal.interface.binding.subspan layout(#pipeline_layout) binding(2) alignment(64) offset(%c0) flags(ReadOnly) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<11xi32>>
+  %output = hal.interface.binding.subspan layout(#pipeline_layout) binding(3) alignment(64) offset(%c0) : !iree_tensor_ext.dispatch.tensor<writeonly:tensor<32x1x1xf32>>
+  %original_t = iree_tensor_ext.dispatch.tensor.load %original, offsets = [0, 0, 0], sizes = [32, 1, 1], strides = [1, 1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<32x1x1xf32>> -> tensor<32x1x1xf32>
```
hanhanW, line 488: I don't think you need the changes on VerifyWorkgroupDistribution side. I don't know how you constructed the pipeline test. You wont have the issue if this has write access.

Likely, it is readonly because you use function arguments in outs in your original program. Is that the case?

hanhanW (changes requested):

compiler/src/iree/compiler/Codegen/LLVMGPU/test/nvvm_pipeline_test.mlir
moomoohorse321: >  it is readonly because you use function arguments in outs in your original program

Yes, the original program is something like 

```
func.func @decode(%cache: tensor<...>, %updates: tensor<...>, %indices: tensor<...>)
    -> tensor<...> {
  %new_cache = stablehlo.scatter(%cache, %indices, %updates) ...
  return %new_cache
}
```
exported from jax

which gets lower to something like 

```
%arg0 : dispatch.tensor<readonly:tensor<...>>
%arg3 : dispatch.tensor<writeonly:tensor<...>>

%original = dispatch.tensor.load %arg0 : readonly -> tensor<32x1x1xf32>
%scatter = iree_linalg_ext.scatter ... outs(%original)
dispatch.tensor.store %scatter, %arg3 : -> writeonly
```

We will think about a better fix within the week. This fix is indeed a bit hacky

hanhanW: It is easier to provide guidance if you can send an e2e repro (using `iree-compile`). I'm quite confused, because we have similar e2e test in https://github.com/iree-org/iree/blob/main/tests/e2e/stablehlo_ops/scatter.mlir

The `original` operand would be marked rw, because IREE implements TiedInterface for linalg/linalg_ext ops. I don't know how you get a readonly only operand here.

Below is a snippet from agents, that matches my understanding.

```
  1. iree_linalg_ext.scatter is DPS. Its result is tied to its outs operand.

  In compiler/src/iree/compiler/ExternalInterfaces/UtilExternalModels.cpp:638, IREE registers TiedOpInterface for Linalg/LinalgExt ops:

  return findTiedBaseValue(linalgOp.getDpsInits()[resultIndex]);

  So for:

  %scatter = iree_linalg_ext.scatter ... outs(%original)

  the tied base value for %scatter is %original.

  2. ConvertRegionToWorkgroups follows that tie when converting flow.dispatch.region to flow.dispatch.workgroups.

  In compiler/src/iree/compiler/Dialect/Flow/Transforms/ConvertRegionToWorkgroups.cpp:48, it walks backward through tied ops:

  value = tiedOpInterface.getTiedResultOperand(value);

  Then it records which dispatch result is tied to which dispatch argument:

  tiedArguments[it.index()] = std::distance(argumentsSet.begin(), ...);

  3. DispatchWorkgroupsOp::build turns tied input operands into readwrite.
```

YuWei-CH: Yes, you are right. For the scatter directly tied to a full dispatch argument, `TiedOpInterface` should make the corresponding dispatch operand `readwrite`. I checked that case with a small e2e `iree-compile` test, and it behaves as you described.

The Qwen case we hit is slightly different: the scatter init is a sliced/reshaped window from a larger tensor, not the full captured tensor itself. In that case, the e2e pipeline produces a readonly binding for the larger source tensor and a separate writeonly binding for the updated window, which matches the failure we saw.

Here is the minimal pattern I used to reproduce it:

```mlir
// RUN: iree-compile %s --iree-input-type=stablehlo --iree-hal-target-backends=cuda --iree-cuda-target=sm_80 --compile-to=executable-sources | FileCheck %s --check-prefix=SOURCES
// RUN: iree-compile %s --iree-input-type=stablehlo --iree-hal-target-backends=cuda --iree-cuda-target=sm_80 --compile-to=executable-targets -o /dev/null

// Models a StableHLO/JAX scatter that updates a slice window, not the whole
// source tensor, so the dispatch has a readonly source and a separate output.
func.func @scatter_update_sliced_window(
    %source: tensor<3x1x1x32xf32>,
    %indices: tensor<11x1xi32>,
    %updates: tensor<1x1x11xf32>) -> tensor<1x1x32xf32>
    attributes {iree.module.export} {
  %slice = stablehlo.slice %source [0:1, 0:1, 0:1, 0:32] :
      (tensor<3x1x1x32xf32>) -> tensor<1x1x1x32xf32>
  %window = stablehlo.reshape %slice :
      (tensor<1x1x1x32xf32>) -> tensor<1x1x32xf32>
  %result = "stablehlo.scatter"(%window, %indices, %updates) ({
  ^bb0(%old: tensor<f32>, %update: tensor<f32>):
    "stablehlo.return"(%update) : (tensor<f32>) -> ()
  }) {
    indices_are_sorted = true,
    scatter_dimension_numbers = #stablehlo.scatter<
      update_window_dims = [0, 1],
      inserted_window_dims = [2],
      scatter_dims_to_operand_dims = [2],
      index_vector_dim = 1
    >,
    unique_indices = true
  } : (tensor<1x1x32xf32>, tensor<11x1xi32>, tensor<1x1x11xf32>)
      -> tensor<1x1x32xf32>
  return %result : tensor<1x1x32xf32>
}

// SOURCES-LABEL: func.func @scatter_update_sliced_window_dispatch_0_scatter_32x1x1xf32_dispatch_tensor_store
// SOURCES-DAG: !iree_tensor_ext.dispatch.tensor<readonly:tensor<3x1x1x32x1x1xf32>>
// SOURCES-DAG: !iree_tensor_ext.dispatch.tensor<readonly:tensor<11x1x1xf32>>
// SOURCES-DAG: !iree_tensor_ext.dispatch.tensor<readonly:tensor<11xi32>>
// SOURCES-DAG: !iree_tensor_ext.dispatch.tensor<writeonly:tensor<32x1x1xf32>>
// SOURCES: iree_linalg_ext.scatter
// SOURCES: !iree_tensor_ext.dispatch.tensor<writeonly:tensor<32x1x1xf32>>
}

hanhanW: Is it still a problem when you run the model on main branch?

https://github.com/iree-org/iree/pull/24443#issuecomment-4447214534 claims that the test passes.

YuWei-CH: Yes, we re-tested on updated main. The original model-level test only passes when this single-workgroup scatter fix is included. The same pattern still fails in workgroup distribution verification on the copy-back `memref.copy` without this fix.

PR #24443, opened 2026-05-12
[Codegen] Avoid consumer fusion through scatter ops

YuWei-CH:
Problem:
The forall consumer-fusion pattern can try to fuse through linalg_ext.scatter. The generic tiling helper may partially mutate IR before reporting failure for scatter consumers, which can make the greedy fuse/hoist pass fail to converge.

Fix:
Do not run the forall consumer-fusion pattern on linalg_ext.scatter. Direct scatter fusion needs scatter-specific legality checks because scatter writes to runtime-indexed locations, carries an outs tensor, and must preserve copy/update/writeback and combiner semantics.

Example:
Keep a scatter consuming a forall result outside this fusion path. A focused regression covers this case and prevents the generic fusion path from partially mutating the IR.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_fuse_and_hoist_forall.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/GPU/test/gpu_fuse_and_hoist_forall.mlir
@@ -187,6 +187,31 @@ func.func @forall_fuse_then_hoist_with_fill(%3: tensor<128x128xf16>, %4: tensor<
 
 // -----
 
+func.func @do_not_consumer_fuse_scatter(%arg0: tensor<2x1x2x256xf32>,
+                                        %arg1: tensor<2xi32>,
+                                        %arg2: tensor<256x1x2x256xf32>) -> tensor<256x1x2x256xf32> {
+  %empty = tensor.empty() : tensor<2x1x2x256xf32>
+  %0 = scf.forall (%arg3) in (2) shared_outs(%arg4 = %empty) -> (tensor<2x1x2x256xf32>) {
+    %2 = tensor.extract_slice %arg0[%arg3, 0, 0, 0] [1, 1, 2, 256] [1, 1, 1, 1] : tensor<2x1x2x256xf32> to tensor<1x1x2x256xf32>
+    scf.forall.in_parallel {
+      tensor.parallel_insert_slice %2 into %arg4[%arg3, 0, 0, 0] [1, 1, 2, 256] [1, 1, 1, 1] : tensor<1x1x2x256xf32> into tensor<2x1x2x256xf32>
+    }
+  }
+  %1 = iree_linalg_ext.scatter dimension_map = [0] unique_indices(false)
+    ins(%0, %arg1 : tensor<2x1x2x256xf32>, tensor<2xi32>) outs(%arg2 : tensor<256x1x2x256xf32>) {
+  ^bb0(%arg3: f32, %arg4: f32):
+    iree_linalg_ext.yield %arg3 : f32
+  } -> tensor<256x1x2x256xf32>
+  return %1 : tensor<256x1x2x256xf32>
+}
+
+// CHECK-LABEL: func @do_not_consumer_fuse_scatter(
+//       CHECK:   %[[PRODUCER:.+]] = scf.forall
+//       CHECK:   iree_linalg_ext.scatter
+//  CHECK-SAME:     ins(%[[PRODUCER]],
+
+// -----
+
 func.func @multi_hoist_and_fuse_trailing_stuff(%2: tensor<128x128xf16>) -> tensor<128x128xf16> {
   %c4 = arith.constant 4 : index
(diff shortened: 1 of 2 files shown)


hanhanW (changes requested):
We reject a similar patch recently: https://github.com/iree-org/iree/pull/24256

@IanWood1 has the context. I think the first question is that is it an expected fusion result? It's hard to tell when I look at the lit test, as there are no producer ops. Thus, I think we need more context.

moomoohorse321:
> We reject a similar patch recently: #24256
> 
> @IanWood1 has the context. I think the first question is that is it an expected fusion result? It's hard to tell when I look at the lit test, as there are no producer ops. Thus, I think we need more context.

Yup we have the same issue so we propose this fix. #24256 did it better than our fix and I agree with @qedawkins that the right way to solve this is essentially 

1. Find tensor.parallel_insert_slice ops that produce scf.forall results consumed by the candidate op.

2. Build operandNumbers, allOffsets, and allSizes from those slices

3. Before any IR mutation, call:

```
if (!tilableOp.isOpFusableWithProducerSlices(
        operandNumbers, allOffsets, allSizes)) {
  return rewriter.notifyMatchFailure(...);
}
```

My question is that the conversation in #24256 seems to end. Is this already fixed?

This is a blocker for using some model decodes (e.g. qwen 3.5) exported from jax w/ KV cache in my understanding.

IanWood1:
https://github.com/iree-org/iree/pull/24256 was rejected because https://github.com/iree-org/iree/commit/d097bad562a7e941d5522c58c82f4e0a26907378 prevents forming these dispatches. However, this is only related to fusing along the scatters dps init operand and your lit test is fusing along the `updates` operand. 

I also don't really see how `@do_not_consumer_fuse_scatter` could be failing to tile and fuse.


> I think the first question is that is it an expected fusion result? It's hard to tell when I look at the lit test, as there are no producer ops. Thus, I think we need more context.

This looks like a case where we should be fusing (putting them in the same dispatch). This seems like a codegen issue and not a dispatch issue like the dps init case.

hanhanW:
> #24256 was rejected because [d097bad](https://github.com/iree-org/iree/commit/d097bad562a7e941d5522c58c82f4e0a26907378) prevents forming these dispatches. However, this is only related to fusing along the scatters dps init operand and your lit test is fusing along the `updates` operand.
> 
> I also don't really see how `@do_not_consumer_fuse_scatter` could be failing to tile and fuse.
> 
> > I think the first question is that is it an expected fusion result? It's hard to tell when I look at the lit test, as there are no producer ops. Thus, I think we need more context.
> 
> This looks like a case where we should be fusing (putting them in the same dispatch). This seems like a codegen issue and not a dispatch issue like the dps init case.

I may be missing something. What is the case? The test case is constructed manually I think. I don't find any produce op (which should implement TilingInterface) from the lit test.

```mlir
func.func @do_not_consumer_fuse_scatter(%arg0: tensor<2x1x2x256xf32>,
                                        %arg1: tensor<2xi32>,
                                        %arg2: tensor<256x1x2x256xf32>) -> tensor<256x1x2x256xf32> {
  %empty = tensor.empty() : tensor<2x1x2x256xf32>
  %0 = scf.forall (%arg3) in (2) shared_outs(%arg4 = %empty) -> (tensor<2x1x2x256xf32>) {
    %2 = tensor.extract_slice %arg0[%arg3, 0, 0, 0] [1, 1, 2, 256] [1, 1, 1, 1] : tensor<2x1x2x256xf32> to tensor<1x1x2x256xf32>
    scf.forall.in_parallel {
      tensor.parallel_insert_slice %2 into %arg4[%arg3, 0, 0, 0] [1, 1, 2, 256] [1, 1, 1, 1] : tensor<1x1x2x256xf32> into tensor<2x1x2x256xf32>
    }
  }
  %1 = iree_linalg_ext.scatter dimension_map = [0] unique_indices(false)
    ins(%0, %arg1 : tensor<2x1x2x256xf32>, tensor<2xi32>) outs(%arg2 : tensor<256x1x2x256xf32>) {
  ^bb0(%arg3: f32, %arg4: f32):
    iree_linalg_ext.yield %arg3 : f32
  } -> tensor<256x1x2x256xf32>
  return %1 : tensor<256x1x2x256xf32>
}
```

QQ: @IanWood1 Do we fuse scatter ops into producer dispatch or consumer dispatch? Or it depends?

IanWood1:
> ```mlir
> do_not_consumer_fuse_scatter
> ```

@hanhanW we only ever fuse a scatter with it's producer and not it's consumer. So, they should always be the final op in the dispatch.

@YuWei-CH did the original failure you see have a `producer -> scatter` dispatch where `producer` was used by the scatter's "init" operand. If that is the case this issue should be fixed on main by the previously mentioned commit  https://github.com/iree-org/iree/commit/d097bad562a7e941d5522c58c82f4e0a26907378

YuWei-CH:
@IanWood1 
Yes, that seems likely. Thanks for pointing this out. We reran our original Qwen test with d097bad included, and the test passes.

PR #24444, opened 2026-05-12
[Python] Preserve device-local buffer uploads

YuWei-CH:
Problem:
- Python ndarray arguments and `asdevicearray` uploads requested device-local memory, but also requested `IREE_HAL_BUFFER_USAGE_MAPPING` and forced `IREE_HAL_MEMORY_TYPE_HOST_VISIBLE` in `AllocateBufferCopy`.
- `IREE_HAL_BUFFER_USAGE_MAPPING` means the final HAL buffer may be mapped by the host; it is not required for uploading host data into a device buffer.
- On CUDA and other discrete GPU backends, requiring host-visible mapping can prevent the allocator from choosing true device-local memory.

Fix:
- Allocate buffer copies with the requested memory type instead of forcing host-visible memory.
- Drop mapping usage from implicit ndarray and `asdevicearray` uploads, so the default path creates device-local dispatch buffers.
- Callers that need direct host mapping can still request host-visible memory and mapping usage explicitly.

Example:
- Passing a NumPy array as a function argument now uploads it with `iree_hal_device_transfer_h2d` into a `DEVICE_LOCAL` buffer instead of creating a host-visible, mappable buffer.
- The buffer remains usable for dispatch storage and transfer, while explicit host reads can still copy the data back with `to_host()`.

### DIFF
--- a/runtime/bindings/python/hal.cc
+++ b/runtime/bindings/python/hal.cc
@@ -215,8 +215,7 @@ py::object HalAllocator::AllocateBufferCopy(
   PyBufferReleaser py_view_releaser(py_view);
 
   iree_hal_buffer_params_t params = {0};
-  // TODO: Should not require host visible :(
-  params.type = memory_type | IREE_HAL_MEMORY_TYPE_HOST_VISIBLE;
+  params.type = memory_type;
   params.usage = allowed_usage;
 
   iree_hal_buffer_t* hal_buffer = nullptr;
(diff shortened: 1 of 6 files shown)


benvanik (approved):
seems fine to me - note that AllocateBufferCopy is a very very slow path, so any real application is really going to want to ensure its arrays are on device already - a synchronous h2d is catastrophic for performance.

PR #24445, opened 2026-05-12
[Codegen] Split StableHLO/JAX select-style argmax reductions

YuWei-CH:
Problem:
StableHLO/JAX lowers argmax for lm_head vocab reductions as a pair of selects with explicit NaN handling and lower-index tie breaking. The existing split-reduction matcher only recognized the canonical maximumf/cmpf/select form, so large vocab reductions stayed as one wide reduction.

Fix:
Teach the argmax matcher to distinguish canonical and StableHLO/JAX select-style argmax reductions, and preserve the corresponding combiner when splitting. Keep the select-style matcher strict by requiring the `gt || isnan` value condition, the `eq && lower_index` tie condition, and the expected value/index select operands.

Example:
A Qwen `lm_head` argmax over vocab size 248320 can now split into a two-stage reduction over 1940x128 tiles while preserving NaN and lower-index tie semantics. Negative tests ensure similar-looking select reductions with different NaN/tie/index semantics do not split.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
@@ -387,55 +387,6 @@ splitReduction(RewriterBase &rewriter, LinalgExt::TopkOp topkOp,
   return success();
 }
 
-struct ArgmaxCombinerOps {
-  Operation *maxOp = nullptr;    // arith.maximumf
-  Operation *selectOp = nullptr; // arith.select
-  Operation *cmpOp = nullptr;    // arith.cmpf
-};
-
-// Matches the combiner pattern in a linalg.generic argmax-style reduction:
-// Example MLIR:
-// %4:2 = linalg.generic {
-//     indexing_maps = [...],
-//     iterator_types = ["parallel", "reduction"]
-//   } ins(%arg0 : tensor<?x128xbf16>) outs(%1, %3 : tensor<?xbf16>,
-//   tensor<?xi64>) {
-// ^bb0(%in: bf16, %out: bf16, %out_0: i64):
-//   %5 = linalg.index 1 : index
-//   %6 = arith.index_cast %5 : index to i64
-//   %7 = arith.maximumf %in, %out : bf16
-//   %8 = arith.cmpf ogt, %in, %out : bf16
-//   %9 = arith.select %8, %6, %out_0 : i64
-//   linalg.yield %7, %9 : bf16, i64
-// } -> (tensor<?xbf16>, tensor<?xi64>)
-//
-// This function extracts the `arith.maximumf`, `arith.cmpf`, and `arith.select`
-// operations from the body to facilitate transformations such as split
-// reduction.
-static ArgmaxCombinerOps collectArgmaxCombinerOps(linalg::GenericOp genericOp) {
-  assert(isArgmaxOp(genericOp) && "expected operation to be an argmax op");
-
-  auto yieldOp = cast<linalg::YieldOp>(genericOp.getBody()->getTerminator());
-
-  // Extract max value producer: arith.maximumf.
-  Value maxResult = yieldOp.getOperand(0);
-  auto maxOp = cast<arith::MaximumFOp>(maxResult.getDefiningOp());
-
-  // Extract index result producer: arith.select.
-  Value indexResult = yieldOp.getOperand(1);
-  auto selectOp = cast<arith::SelectOp>(indexResult.getDefiningOp());
-
-  // Extract the condition of the select, expected to be arith.cmpf with
-  // predicate OGT.
-  auto cmpOp = cast<arith::CmpFOp>(selectOp.getCondition().getDefiningOp());
-
-  ArgmaxCombinerOps ops;
-  ops.maxOp = maxOp;
-  ops.selectOp = selectOp;
-  ops.cmpOp = cmpOp;
-  return ops;
-}
-
 static Value expandValue(OpBuilder &builder, Location loc, Value value,
                          RankedTensorType expandedType) {
   RankedTensorType originalType = cast<RankedTensorType>(value.getType());
@@ -521,7 +472,8 @@ static Value getSplitReductionInit(OpBuilder &builder, Location loc,
 FailureOr<linalg::SplitReductionResult>
 splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
                      linalg::ControlSplitReductionFn controlSplitReductionFn) {
-  assert(isArgmaxOp(genericOp) && "expected operation to be an argmax op");
+  std::optional<ArgmaxKind> argmaxKind = getArgmaxKind(genericOp);
+  assert(argmaxKind && "expected operation to be an argmax op");
 
   OpBuilder::InsertionGuard guard(rewriter);
   rewriter.setInsertionPoint(genericOp);
@@ -568,13 +520,43 @@ splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
                                        "compared to intermediate tensor size");
   }
 
-  ArgmaxCombinerOps combinerOps = collectArgmaxCombinerOps(genericOp);
-  Operation *reductionOp = combinerOps.maxOp;
-  std::optional<TypedAttr> identity = arith::getNeutralElement(reductionOp);
-  if (!identity.has_value()) {
+  auto valueType =
+      dyn_cast<FloatType>(genericOp.getRegionOutputArgs()[0].getType());
+  if (!valueType) {
     return rewriter.notifyMatchFailure(
         genericOp, "Unknown identity value for the reduction");
   }
+  auto identity = FloatAttr::get(
+      valueType, APFloat::getInf(valueType.getFloatSemantics(), true));
+
+  bool selectStyleArgmax = *argmaxKind == ArgmaxKind::StableHloSelectStyle;
+
+  auto buildArgmaxCombiner = [](OpBuilder &b, Location loc, Value val,
+                                Value idx, Value outVal, Value outIdx,
+                                bool selectStyleArgmax) {
+    if (!selectStyleArgmax) {
+      Value maxVal = arith::MaximumFOp::create(b, loc, val, outVal);
+      Value cmp =
+          arith::CmpFOp::create(b, loc, arith::CmpFPredicate::OGT, val, outVal);
+      Value selIdx = arith::SelectOp::create(b, loc, cmp, idx, outIdx);
+      return std::pair<Value, Value>{maxVal, selIdx};
+    }
+
+    Value greater =
+        arith::CmpFOp::create(b, loc, arith::CmpFPredicate::OGT, val, outVal);
+    Value isNan =
+        arith::CmpFOp::create(b, loc, arith::CmpFPredicate::UNE, val, val);
+    Value valueCond = arith::OrIOp::create(b, loc, greater, isNan);
+    Value equal =
+        arith::CmpFOp::create(b, loc, arith::CmpFPredicate::OEQ, val, outVal);
+    Value lowerIndex =
+        arith::CmpIOp::create(b, loc, arith::CmpIPredicate::slt, idx, outIdx);
+    Value tieCond = arith::AndIOp::create(b, loc, equal, lowerIndex);
+    Value indexCond = arith::OrIOp::create(b, loc, valueCond, tieCond);
+    Value maxVal = arith::SelectOp::create(b, loc, valueCond, val, outVal);
+    Value selIdx = arith::SelectOp::create(b, loc, indexCond, idx, outIdx);
+    return std::pair<Value, Value>{maxVal, selIdx};
+  };
 
   SmallVector<Value> newInputs;
   for (OpOperand *operand : genericOp.getDpsInputOperands()) {
@@ -611,7 +593,7 @@ splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
   OpOperand *indexInit = genericOp.getDpsInitOperand(1);
   // Value identity.
   Value identityValue =
-      getSplitReductionInit(rewriter, loc, valueInit->get(), *identity,
+      getSplitReductionInit(rewriter, loc, valueInit->get(), identity,
                             outputDimSize, insertSplitIndex);
   // Index identity.
   Type indexElemType = genericOp.getRegionOutputArgs()[1].getType();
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h
```diff
@@ -7,6 +7,7 @@
 #ifndef IREE_COMPILER_DIALECT_LINALGEXT_UTILS_UTILS_H_
 #define IREE_COMPILER_DIALECT_LINALGEXT_UTILS_UTILS_H_
 
+#include <optional>
```
bangtianliu: Other files in this same directory (e.g., Transforms/ReshapeFusion.cpp, IR/LinalgExtOps.cpp) put C++ standard headers in a separate group after project headers, separated by a blank line.

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
```diff
@@ -568,13 +520,43 @@ splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
...
-  std::optional<TypedAttr> identity = arith::getNeutralElement(reductionOp);
-  if (!identity.has_value()) {
+  auto valueType =
+      dyn_cast<FloatType>(genericOp.getRegionOutputArgs()[0].getType());
+  if (!valueType) {
     return rewriter.notifyMatchFailure(
         genericOp, "Unknown identity value for the reduction");
   }
+  auto identity = FloatAttr::get(
+      valueType, APFloat::getInf(valueType.getFloatSemantics(), true));
```
bangtianliu: getArgmaxKind() already requires the FillOp init value to match m_NegInfFloat(), which means the value type is necessarily a FloatType by the time we get here.

use cast instead (which assert)

```cpp
  auto valueType = cast<FloatType>(genericOp.getRegionOutputArgs()[0].getType());
  auto identity = FloatAttr::get(
      valueType, APFloat::getInf(valueType.getFloatSemantics(), /*Negative=*/true));
```

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
```diff
@@ -703,8 +681,8 @@ splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
       rewriter, loc, genericOp.getResultTypes(),
       ValueRange{partialArgmax.getResult(0), partialArgmax.getResult(1)},
       genericOp.getDpsInits(), finalReductionMaps, reductionIteratorTypes,
-      [combinerOps, tileSize, insertSplitDimension](OpBuilder &b, Location loc,
-                                                    ValueRange inputs) {
+      [tileSize, insertSplitDimension, selectStyleArgmax,
+       buildArgmaxCombiner](OpBuilder &b, Location loc, ValueRange inputs) {
```
bangtianliu: The previous implementation cloned the user's actual maximumf/cmpf/select operations(b.clone(*combinerOps.maxOp) etc.) into the final reduction. The new code re-emits the combiner from scratch via buildArgmaxCombiner. That's cleaner, but it loses any attributes the frontend may have attached to those ops like most notably arith::FastMathFlags.

if you do this intentionally, then document it explicitly in the code comment; otherwise, fix it

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
```diff
@@ -568,13 +520,43 @@ splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
...
+  if (!valueType) {
     return rewriter.notifyMatchFailure(
         genericOp, "Unknown identity value for the reduction");
   }
+  auto identity = FloatAttr::get(
+      valueType, APFloat::getInf(valueType.getFloatSemantics(), true));
+
+  bool selectStyleArgmax = *argmaxKind == ArgmaxKind::StableHloSelectStyle;
+
+  auto buildArgmaxCombiner = [](OpBuilder &b, Location loc, Value val,
```
bangtianliu: `buildArgmaxCombiner` is declared as a local lambda inside splitArgmaxReduction, and then captured by value in both nested builder lambdas, and `selectStyleArgmax` is passed both as a captured variable to the outer lambda and as a separate parameter to buildArgmaxCombiner. That's redundant and obscures
 the call sites.

Fix: Promote it to a static helper function in the anonymous namespace:
  ```cpp
  static std::pair<Value, Value> buildArgmaxCombiner(
      OpBuilder &b, Location loc, ArgmaxKind kind,
      Value val, Value idx, Value outVal, Value outIdx) {
    if (kind == ArgmaxKind::Canonical) {
      Value maxVal = arith::MaximumFOp::create(b, loc, val, outVal);
      ...
    } 
    ...
  } 
```

**Review by bangtianliu — CHANGES_REQUESTED** [review]

**Review by hanhanW — COMMENTED** [review]
Drive-by: We are moving towards using partial reduction tiling for split reductions ([link](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/DispatchCreation/FormSplitReductionDispatches.cpp))

@bangtianliu thanks for reviewing the code. I think the changes do not affect the preferred approach, right?

**Comment by bangtianliu** [comment]
> @bangtianliu thanks for reviewing the code. I think the changes do not affect the preferred approach, right?

Yeah,  these changes don't conflict with the partial-reduction-tiling migration.

File: compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp (hunk 3 above)
Comment (YuWei-CH) [reply compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp]: The reason we regenerated the combiner instead of cloning the original ops is that split argmax builds two new reductions: the first uses local tile indices, and the second uses reconstructed global indices. Re-emitting the recognized argmax semantics made the remapping clearer than cloning the original ops and rewriting their operands.

That said, your point about preserving attributes makes sense. Would you prefer that we keep this regenerated-combiner structure but explicitly preserve the original floating op attributes, especially `arith::FastMathFlagsAttr`, on the corresponding generated `maximumf`/`cmpf` ops?

File: compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h (hunk 1 above)
Comment (YuWei-CH) [reply compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h]: Addressed at 03e8b39902b7220a75ab044a4505aa64c9768005

File: compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp (hunk 2 above)
Comment (YuWei-CH) [reply compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp]: Thank you for pointing this out. We have resolved it at 03e8b39902b7220a75ab044a4505aa64c9768005

File: compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp (hunk 4 above)
Comment (YuWei-CH) [reply compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp]: Thank you for pointing out it. We resolved it at 03e8b39902b7220a75ab044a4505aa64c9768005.

File: compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp (hunk 3 above)
Comment (YuWei-CH) [reply compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp]: @bangtianliu Hi, we updated the patch to explicitly preserve the matched argmax combiner attributes when regenerating the split-reduction combiner. The change is in 3490d76fdfd79f7b0c2c490167134890711c3231.
Thank you for your review!

File: compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp [hunk 5]
```diff
@@ -518,10 +471,57 @@ static Value getSplitReductionInit(OpBuilder &builder, Location loc,
...
+  return {valueSelect, indexSelect};
+}
+
 FailureOr<linalg::SplitReductionResult>
 splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
                      linalg::ControlSplitReductionFn controlSplitReductionFn) {
-  assert(isArgmaxOp(genericOp) && "expected operation to be an argmax op");
+  std::optional<ArgmaxCombinerInfo> argmaxInfo =
+      getArgmaxCombinerInfo(genericOp);
+  assert(argmaxInfo && "expected operation to be an argmax op");
```
bangtianliu: assert then dereference on std::optional,  UB in release builds

Fix: 

```cpp
  std::optional<ArgmaxCombinerInfo> argmaxInfo =
      getArgmaxCombinerInfo(genericOp);
  if (!argmaxInfo) {
        genericOp, "operation is not a recognized argmax pattern");
  }
  ArgmaxCombinerInfo info = *argmaxInfo;
```

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
```diff
@@ -568,13 +568,12 @@ splitArgmaxReduction(RewriterBase &rewriter, linalg::GenericOp genericOp,
...
   }
 
-  ArgmaxCombinerOps combinerOps = collectArgmaxCombinerOps(genericOp);
-  Operation *reductionOp = combinerOps.maxOp;
-  std::optional<TypedAttr> identity = arith::getNeutralElement(reductionOp);
-  if (!identity.has_value()) {
-    return rewriter.notifyMatchFailure(
-        genericOp, "Unknown identity value for the reduction");
-  }
+  auto valueType =
```
bangtianliu: This is safe today only because `getArgmaxCombinerInfo` checks `m_NegInfFloat()`. That coupling is invisible at the call site, and the previous `arith::getNeutralElement(reductionOp)` was self-contained.

Fix: Either (a) add a `dyn_cast<FloatType>` + `notifyMatchFailure`, or (b) keep using `arith::getNeutralElement` for the canonical kind and only fall back to a hardcoded `-Inf` for `StableHloSelectStyle` (which has no MaximumFOp to query). Option (b) preserves prior behavior for the canonical path: TypedAttr identity;

```cpp
  if (info.kind == ArgmaxKind::Canonical) {
    // Stash the matched MaximumFOp in ArgmaxCombinerInfo, query its neutral element here.
    ...
  } else {
    auto floatType = dyn_cast<FloatType>(...);
    if (!floatType) return rewriter.notifyMatchFailure(...);
    identity = FloatAttr::get(floatType, APFloat::getInf(..., /*Negative=*/true));
  }
```

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
```diff
@@ -518,10 +471,57 @@ static Value getSplitReductionInit(OpBuilder &builder, Location loc,
   return linalg::FillOp::create(builder, loc, identityVal, empty).getResult(0);
 }
 
+static void setPreservedAttrs(Operation *op, DictionaryAttr attrs) {
+  if (attrs) {
+    op->setAttrs(attrs);
+  }
+}
+
+static std::pair<Value, Value>
```
bangtianliu: Use a named struct instead of std::pair?

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -1283,25 +1283,245 @@ bool isaHorizontallyFusedContraction(Operation *op) {
...
+      /*kind=*/ArgmaxKind::Canonical,
+      /*maximumAttrs=*/maxOp->getAttrDictionary(),
+      /*greaterThanCmpAttrs=*/greaterThan->getAttrDictionary(),
+      /*isNanCmpAttrs=*/{},
+      /*equalCmpAttrs=*/{},
+      /*valueSelectAttrs=*/{},
+      /*indexSelectAttrs=*/indexSelect->getAttrDictionary()};
+}
+
+static std::optional<std::pair<arith::CmpFOp, arith::CmpFOp>>
```
bangtianliu: Also here

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h
```diff
@@ -242,6 +244,29 @@ bool isGatherlikeOp(Operation *op);
...
+  DictionaryAttr indexSelectAttrs;
+};
+
+/// Returns the argmax combiner info represented by the given linalg.generic, if
+/// any.
+std::optional<ArgmaxCombinerInfo>
+getArgmaxCombinerInfo(linalg::GenericOp genericOp);
+
+/// Returns the argmax kind represented by the given linalg.generic, if any.
+std::optional<ArgmaxKind> getArgmaxKind(linalg::GenericOp genericOp);
```
bangtianliu: getArgmaxKind is dead weight in the public API, only used for isArgmaxOp, could we inline it.

compiler/src/iree/compiler/DispatchCreation/test/split_reduction.mlir
```diff
@@ -134,6 +134,156 @@ util.func public @argmax_no_split(%arg0: tensor<?x512xbf16>, %arg1: index) -> te
 
 // -----
```
bangtianliu, line 136: Add a positive test for extf cast on the input (mixed-type bf16 -> f32)

bangtianliu: Add a "lhs/rhs swapped" positive test for the OrI/AndI matchers

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
```diff
@@ -1283,25 +1283,245 @@ bool isaHorizontallyFusedContraction(Operation *op) {
   return true;
 }
 
-bool isArgmaxOp(linalg::GenericOp genericOp) {
+static Value stripFloatCast(Value value) {
```
bangtianliu: The helper strips extf/truncf only, giving a better name? like `stripFloatExtTrunc`

bangtianliu:
Also, fix the CI errors

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
YuWei-CH: Thank you for review, we changed it at 8c853d5c14e3616182318f3e0dbd03d701900b7c

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
YuWei-CH: Thank you for pointing out! We totally agree and implement Option b at 8c853d5c14e3616182318f3e0dbd03d701900b7c

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/SplitReduction.cpp
YuWei-CH: Addressed at 8c853d5c14e3616182318f3e0dbd03d701900b7c

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
YuWei-CH: Address at 8c853d5c14e3616182318f3e0dbd03d701900b7c

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.h
YuWei-CH: Address at 8c853d5c14e3616182318f3e0dbd03d701900b7c

compiler/src/iree/compiler/DispatchCreation/test/split_reduction.mlir
YuWei-CH: Add at 8c853d5c14e3616182318f3e0dbd03d701900b7c

compiler/src/iree/compiler/Dialect/LinalgExt/Utils/Utils.cpp
YuWei-CH: Change the name at 8c853d5c14e3616182318f3e0dbd03d701900b7c

YuWei-CH:
Hi @bangtianliu! The errors related to argmax_4d_unit_parallel_f32i64 have been fixed, but the timeout issue still remains.

bangtianliu (approved):

bangtianliu:
LGTM, but please also wait for another approval, cc @hanhanW here

AGindinson (approved):
I suggest merging this by EO day if there are no objections

PR #24449, opened 2026-05-12
[Codegen] Fold masked transfer_read/write in loops

sommerlukas:
Add pattern to fold pairs of masked `transfer_read` and `transfer_write` inside loops. Also adds additional patterns that simplify the operation chain inside the loop body through folding to enable the main folding.

This contributes towards https://github.com/iree-org/iree/issues/24221.

Assisted-by: Codex

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
@@ -435,12 +435,11 @@ private:
     auto readConversion = IREE::Util::HoistableConversionOp::create(
         rewriter, readOp.getLoc(), kTransferReadTag, kTransferWriteTag,
         TypeRange{readOp.getVectorType()}, ValueRange{candidate.iterArg},
-        [&](OpBuilder &builder, Location loc, ValueRange args) {
-          Value read = vector::TransferReadOp::create(
-              builder, loc, readOp.getVectorType(), args[0],
-              readOp.getIndices(), readOp.getPermutationMap(),
-              readOp.getPadding(), readOp.getMask(), readOp.getInBoundsAttr());
-          return SmallVector<Value>{read};
+        [&](OpBuilder &builder, Location, ValueRange args) {
+          auto clonedRead = cast<vector::TransferReadOp>(
+              builder.clone(*readOp.getOperation()));
+          clonedRead.getBaseMutable().assign(args[0]);
+          return SmallVector<Value>{clonedRead.getResult()};
         });
 
     TypedValue<VectorType> mask = readOp.getMask();
@@ -456,14 +455,12 @@ private:
     auto writeConversion = IREE::Util::HoistableConversionOp::create(
         rewriter, writeOp.getLoc(), kTransferWriteTag, kTransferReadTag,
         TypeRange{writeOp.getResult().getType()}, ValueRange{carriedValue},
-        [&](OpBuilder &builder, Location loc, ValueRange args) {
-          Value write =
-              vector::TransferWriteOp::create(
-                  builder, loc, args[0], candidate.initArg,
-                  writeOp.getIndices(), writeOp.getPermutationMapAttr(),
-                  writeOp.getMask(), writeOp.getInBoundsAttr())
-                  .getResult();
-          return SmallVector<Value>{write};
+        [&](OpBuilder &builder, Location, ValueRange args) {
+          auto clonedWrite = cast<vector::TransferWriteOp>(
+              builder.clone(*writeOp.getOperation()));
+          clonedWrite.getValueToStoreMutable().assign(args[0]);
+          clonedWrite.getBaseMutable().assign(candidate.initArg);
+          return SmallVector<Value>{clonedWrite.getResult()};
         });
 
     rewriter.replaceOp(writeOp, writeConversion.getResult(0));
--- a/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
@@ -898,11 +898,8 @@ canonicalizeLoopCarriedTransferState(mlir::FunctionOpInterface funcOp,
   }
   {
     RewritePatternSet patterns(context);
-    // This intentionally runs independent of fold-identity-slices: it only
-    // retargets transfer reads through identity slices to unblock transfer
-    // folding and does not generally erase identity slice ops.
-    patterns.add<FoldTransferWriteChain, FoldTransferReadOfIdentitySlice,
-                 FoldTransferRAW, FoldTransferReadOfEmptyTensor>(context);
+    patterns.add<FoldTransferWriteChain, FoldTransferRAW,
+                 FoldTransferReadOfEmptyTensor>(context);
     vector::TransferWriteOp::getCanonicalizationPatterns(patterns, context);
     if (failed(applyPatternsGreedily(funcOp, std::move(patterns)))) {
       return failure();
@@ -988,6 +985,9 @@ void OptimizeTensorInsertExtractSlicesPass::runOnOperation() {
   patterns.add<PromoteLoopCarriedTransferIterArg>(context);
   scf::ForOp::getCanonicalizationPatterns(patterns, context);
   vector::TransferWriteOp::getCanonicalizationPatterns(patterns, context);
+  // Run after subset hoisting so retargeting reads through identity slices
+  // does not create direct tensor iter_arg uses that block subset matching.
+  patterns.add<FoldTransferReadOfIdentitySlice>(context);
   if (foldIdentitySlices) {
     patterns.add<CastLikeExtractSliceOpFolder>(context);
     patterns.add<CastLikeInsertSliceOpFolder>(context);
--- a/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
@@ -443,13 +443,13 @@ private:
           return SmallVector<Value>{read};
         });
 
+    TypedValue<VectorType> mask = readOp.getMask();
+    Value pad = readOp.getPadding();
     for (vector::TransferReadOp read : candidate.reads) {
       rewriter.replaceOp(read, readConversion.getResult(0));
     }
 
     vector::TransferWriteOp writeOp = candidate.write;
-    TypedValue<VectorType> mask = readOp.getMask();
-    Value pad = readOp.getPadding();
     rewriter.setInsertionPoint(writeOp);
     Value carriedValue = getLoopCarriedTransferValue(
         rewriter, writeOp.getLoc(), mask, pad, writeOp.getVector());
--- a/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
@@ -236,12 +236,43 @@ hasOnlyLoopInvariantOperandsExcept(LoopLikeOpInterface loopLike, Operation *op,
   });
 }
 
+static bool areEquivalentMasks(Value lhs, Value rhs) {
+  if (lhs == rhs) {
+    return true;
+  }
+  if (!lhs || !rhs || lhs.getType() != rhs.getType()) {
+    return false;
+  }
+  auto lhsCreateMask = lhs.getDefiningOp<vector::CreateMaskOp>();
+  auto rhsCreateMask = rhs.getDefiningOp<vector::CreateMaskOp>();
+  return lhsCreateMask && rhsCreateMask &&
+         lhsCreateMask.getOperands() == rhsCreateMask.getOperands();
+}
+
+static bool
+areEquivalentTransferAccesses(Type lhsVectorType, ValueRange lhsIndices,
+                              AffineMap lhsPermutationMap, Value lhsMask,
+                              Type rhsVectorType, ValueRange rhsIndices,
+                              AffineMap rhsPermutationMap, Value rhsMask) {
+  return lhsVectorType == rhsVectorType && lhsIndices == rhsIndices &&
+         lhsPermutationMap == rhsPermutationMap &&
+         areEquivalentMasks(lhsMask, rhsMask);
+}
+
 static bool areEquivalentTransferReadWrite(vector::TransferReadOp readOp,
                                            vector::TransferWriteOp writeOp) {
-  return readOp.getType() == writeOp.getVector().getType() &&
-         readOp.getIndices() == writeOp.getIndices() &&
-         readOp.getPermutationMap() == writeOp.getPermutationMap() &&
-         readOp.getMask() == writeOp.getMask();
+  return areEquivalentTransferAccesses(
+      readOp.getType(), readOp.getIndices(), readOp.getPermutationMap(),
+      readOp.getMask(), writeOp.getVector().getType(), writeOp.getIndices(),
+      writeOp.getPermutationMap(), writeOp.getMask());
+}
+
+static bool areEquivalentTransferWrites(vector::TransferWriteOp lhs,
+                                        vector::TransferWriteOp rhs) {
+  return areEquivalentTransferAccesses(
+      lhs.getVector().getType(), lhs.getIndices(), lhs.getPermutationMap(),
+      lhs.getMask(), rhs.getVector().getType(), rhs.getIndices(),
+      rhs.getPermutationMap(), rhs.getMask());
 }
 
 static Value getLoopCarriedTransferValue(OpBuilder &builder,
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +225,362 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+/// This handles structurally equivalent masks that upstream WAW folding does
+/// not recognize after CSE/canonicalization fail to make the mask SSA value the
+/// same.
+///
+/// ```mlir
+/// %m0 = vector.create_mask %d : vector<4xi1>
+/// %w0 = vector.transfer_write %v0, %t[%c0], %m0
+/// %m1 = vector.create_mask %d : vector<4xi1>
+/// %w1 = vector.transfer_write %v1, %w0[%c0], %m1
+/// ```
```
amd-eochoalo, line 559: Why can't upstream CSE/canonicalization recognize these masks? I'm assuming that the FoldDimOfShapePreservingTransferIterArg is enabling the code here, is that correct?

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -256,6 +612,78 @@ struct CastLikeInsertSliceOpFolder final
   }
 };
 
+/// Retarget vector.transfer_read through an identity tensor slice.
+///
+/// This is narrower than the general identity-slice folder: it only updates the
+/// transfer_read base and only for zero-offset, unit-stride slices with
+/// identical source/result types. The purpose is to unblock transfer folding in
+/// cleanup phases that intentionally do not run the general
+/// fold-identity-slices patterns.
```
amd-eochoalo, line 631: Just a question to understand, why are some cleanup phases not running fold-identity-slices patterns?

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
amd-eochoalo, line 1011: Just a note that I'm upstreaming these https://github.com/llvm/llvm-project/pull/196599 https://github.com/llvm/llvm-project/pull/196608 if you want to review.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +225,362 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+}
+
+static bool
+areEquivalentTransferAccesses(Type lhsVectorType, ValueRange lhsIndices,
+                              AffineMap lhsPermutationMap, Value lhsMask,
+                              Type rhsVectorType, ValueRange rhsIndices,
+                              AffineMap rhsPermutationMap, Value rhsMask) {
+  return lhsVectorType == rhsVectorType && lhsIndices == rhsIndices &&
+         lhsPermutationMap == rhsPermutationMap &&
+         areEquivalentMasks(lhsMask, rhsMask);
```
amd-eochoalo, line 267: Should we guard against `isMasked()` i.e., the vector.mask?

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +225,362 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+/// lanes, transfer_read returns padding, while transfer_write preserves the
+/// tensor. A generic subset hoist would carry `%new` directly, which is only
+/// equivalent for unmasked transfers.
+struct PromoteLoopCarriedTransferIterArg final : OpRewritePattern<scf::ForOp> {
+  using Base::Base;
+
+  LogicalResult matchAndRewrite(scf::ForOp forOp,
+                                PatternRewriter &rewriter) const override {
+    for (unsigned i = 0, e = forOp.getNumRegionIterArgs(); i < e; ++i) {
+      auto candidate = getPromotableTransferIterArg(forOp, i);
```
amd-eochoalo: nit - spell out the type.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +225,362 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+///     %new = arith.addf %v, %step
+///     %yield = arith.select %mask, %new, splat(%pad)
+///     scf.yield %yield
+///   }
+///
+/// The select is the important masked-transfer semantics. For masked-off
+/// lanes, transfer_read returns padding, while transfer_write preserves the
+/// tensor. A generic subset hoist would carry `%new` directly, which is only
+/// equivalent for unmasked transfers.
+struct PromoteLoopCarriedTransferIterArg final : OpRewritePattern<scf::ForOp> {
```
amd-eochoalo: nit - rename to HoistLoopCarriedTransferIterArg

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +225,362 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+      if (candidate) {
+        return promoteLoopCarriedTransferIterArg(rewriter, forOp, *candidate);
+      }
+    }
+    return failure();
+  }
+
+private:
+  struct PromotableTransferIterArg {
+    unsigned index;
```
amd-eochoalo: Let's use int32_t here.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +225,362 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+  if (lhs == rhs) {
+    return true;
+  }
+  if (!lhs || !rhs || lhs.getType() != rhs.getType()) {
+    return false;
+  }
+  auto lhsCreateMask = lhs.getDefiningOp<vector::CreateMaskOp>();
+  auto rhsCreateMask = rhs.getDefiningOp<vector::CreateMaskOp>();
+  return lhsCreateMask && rhsCreateMask &&
+         lhsCreateMask.getOperands() == rhsCreateMask.getOperands();
```
amd-eochoalo: Should this also handle cases where masks are block arguments? Maybe this could also take into account constant_mask?

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +225,362 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+                                         vector::TransferReadOp readOp,
+                                         Value valueToStore) {
+  TypedValue<VectorType> mask = readOp.getMask();
+  if (!mask) {
+    return valueToStore;
+  }
+  Value pad = readOp.getPadding();
+  Value padVector = vector::BroadcastOp::create(
+      builder, pad.getLoc(), cast<VectorType>(valueToStore.getType()), pad);
+  return arith::SelectOp::create(builder, readOp.getLoc(), mask, valueToStore,
```
amd-eochoalo: Can you add arith to the list of dependent dialects (also ub please?)

amd-eochoalo (commented):
Thanks! I think this looks great! Really good examples on the comments. Just a couple of minor changes / suggestions.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -452,6 +880,47 @@ struct FoldTransferReadOfEmptyTensor
...
+canonicalizeAndPromoteLoopCarriedTransferState(mlir::FunctionOpInterface funcOp,
+                                               MLIRContext *context) {
+  if (failed(canonicalizeLoopCarriedTransferState(funcOp, context))) {
+    return failure();
+  }
+  moveLoopInvariantCodeFromGuaranteedLoops(funcOp);
+
+  RewritePatternSet promotionPatterns(context);
+  promotionPatterns.add<PromoteLoopCarriedTransferIterArg>(context);
+  return applyPatternsGreedily(funcOp, std::move(promotionPatterns));
```
amd-eochoalo: Do we need to run this one as greedily? I think one can capture all scf.for ops and then run in a while loop.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: I've added support for `constant_mask`. For block arguments, we require the two masks to be the same SSA value for now. Checking provenance of masks through region/control flow would IMHO add significant complexity without a clear need ATM.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: Done.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: I've added `arith` and `UB` (in Tablegen). IMHO, `vector` and `scf` are not dependent dialects of this pass, because they exist in the input IR, otherwise we would not run any transformations on them.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: Done, I've renamed from "Promote" to "Hoist" in most places.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: Do you have a specific reason? IMHO, `unsigned` is the right type here, because it's an index passed to APIs like `getRegionIterArg`, and `int32_t` would require casting in that case.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: Yes. The problem case is usually that the mask operands are not the same SSA values even though they describe the same shape and the upstream patterns require the masks to be the same SSA value. For example, one `vector.create_mask` may use `tensor.dim %init`, while another uses `tensor.dim %iter_arg`. CSE will not merge those because the operands differ, and upstream dim canonicalization does not currently prove that an SCF `iter_arg` has the same shape as the `init` tensor through a shape-preserving `vector.transfer_write` chain. `FoldDimOfShapePreservingTransferIterArg` does that local normalization. After that, CSE may merge the masks, and the structural mask equivalence is a fallback for equivalent create/constant masks that still survive as separate ops.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: The pass option defaults to `false`, and several GPU vectorization paths pass `foldIdentitySlices=false`. The broader identity-slice folding can expose direct tensor `iter_arg` uses before subset hoisting has matched the extract/insert structure, which can block the hoisting this pass is trying to do. The narrow old `TransferReadOfIdentitySlice` rewrite is intentionally placed after subset hoisting and only retargets transfer reads, so it gets the transfer-folding benefit without enabling the broader slice cleanup in those phases.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: Yes, we need to run it greedily. Currently, the pattern transforms one iteration argument per application, so if there are multiple arguments that can be transformed, this happens through repeated application of the pattern.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: Thanks, I'll take a look 👍

sommerlukas (commented):
Thanks for the feedback!

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
amd-eochoalo: It is a nit, in the past people have complained when I used unsigned. If this would require a casting at some other point due to getRegionIterArg, then unsigned is the right choice. I didn't see the direct flow to getRegionIterArg when I made this comment, but I did note that there was another unsigned where it was ok to do so.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
amd-eochoalo: Thanks! I think it would be nice to have this upstream (but definitely lower priority and can be done by anyone in a follow up.)

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
amd-eochoalo: Yes, I see that. But I meant more that once you find the scf.for op, you can loop it for that particular operation until all arguments are transformed. That way you can avoid some of the other transformations that the greedy rewriter brings.

Not necessarily blocking, just noting that we are using the greedy rewriter quite a bit for this pass and I thought that if we could remove one it would be a nice little win.

krzysz00:
One thought to simplify this code - move the write and read pairs inside of `util.hoistable_conversion` ops and then apply the hoistsble conversions rewrites instead of manually finding with the loop 

This will also mean you're not limited to for loops

amd-eochoalo (approved):
Thanks! All good on my part. Please respond to Krzysztof's and other reviewers' comments,

sommerlukas:
> One thought to simplify this code - move the write and read pairs inside of `util.hoistable_conversion` ops and then apply the hoistsble conversions rewrites instead of manually finding with the loop
> 
> This will also mean you're not limited to for loops



> One thought to simplify this code - move the write and read pairs inside of `util.hoistable_conversion` ops and then apply the hoistsble conversions rewrites instead of manually finding with the loop
> 
> This will also mean you're not limited to for loops

`util.hoistable_conversion` wasn't on my radar, thanks for bringing it up. I've implemented the transformation with `util.hoistable_conversion` now.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +228,372 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+
+    BlockArgument iterArg = loopLike.getRegionIterArgs()[*index];
+    OpOperand *initArg = loopLike.getTiedLoopInit(iterArg);
+    if (!initArg || writeOp.getBase() != iterArg) {
+      return std::nullopt;
+    }
+    // Avoid marking inner loops whose init is carried by an enclosing loop:
+    // the conversion body would capture that outer iter_arg and can become
+    // self-referential when the conversion is hoisted through the outer loop.
+    if (auto initBlockArg = dyn_cast<BlockArgument>(initArg->get())) {
```
krzysz00, line 381: Is this something where you're just missing repeated pattern application?

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +228,372 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+        insertionPoint = candidateRead.getOperation();
+      }
+    }
+
+    rewriter.setInsertionPoint(insertionPoint);
+    auto readConversion = IREE::Util::HoistableConversionOp::create(
+        rewriter, readOp.getLoc(), kTransferReadTag, kTransferWriteTag,
+        TypeRange{readOp.getVectorType()}, ValueRange{candidate.iterArg},
+        [&](OpBuilder &builder, Location loc, ValueRange args) {
+          Value read = vector::TransferReadOp::create(
```
krzysz00: This smells of `.clone` and `.getBaseMutable()` but that's not strictly required

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
```diff
@@ -225,6 +228,372 @@ void hoistSubsetWithLoopInvariantTensor(RewriterBase &rewriter,
...
+///   %next = vector.transfer_write %v, %acc[%c0], %m
+///   scf.yield %next : tensor<?xf32>
+/// }
+/// ```
+///
+/// becomes:
+///
+/// ```mlir
+/// %r = scf.for ... iter_args(%acc = %init) -> tensor<?xf32> {
+///   %d = tensor.dim %init, %c0 : tensor<?xf32>
```
krzysz00, line 494: This is a good rewrite!

krzysz00 (approved):
Looks overall correct and reasonable

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: No, it's not missing repeated pattern application in this case. 

The bad shape is roughly:
```mlir
  %outer = scf.for ... iter_args(%outer_acc = %init_tensor) -> tensor<...> {
    %inner = scf.for ... iter_args(%inner_acc = %outer_acc) -> tensor<...> {
      %v = vector.transfer_read %inner_acc[...]
      %next = vector.transfer_write %new_v, %inner_acc[...]
      scf.yield %next
    }
    scf.yield %inner
  }
```
  The inner promotion wants to create:
```mlir
  %read_conv = util.hoistable_conversion(%inner_acc) {
    vector.transfer_read ...
  }
  ...
  %write_conv = util.hoistable_conversion(%new_v) {
    vector.transfer_write ..., %outer_acc[...]  // captures enclosing iter_arg
  }
```

The captured `%outer_acc` is the problem. When `eliminateHoistableConversions` hoists the pair out of the inner loop, the resulting write-side conversion now sits in the outer loop body and is still tied to the outer loop-carried tensor. Then the same generic hoistable-conversion machinery can try to hoist that conversion through the outer loop as well. Because the conversion body captured the outer iteration argument instead of taking it as an explicit conversion input, the rewrite can produce a self-referential transfer write. That is what caused the non-convergence/hang in one of the test cases.

If we wanted to support this case, the fix would be more structural: make the enclosing tensor base an explicit conversion input/result or teach the hoistable-conversion elimination to handle captured loop-carried values correctly.

compiler/src/iree/compiler/Codegen/Common/OptimizeTensorInsertExtractSlices.cpp
sommerlukas: I've switched to using `clone`, thanks!

Issue #24454, opened 2026-05-12
[AMDGPU/RDNA4] End-to-end support for global_load_tr (global memory transpose load)


nirvedhmeshram:
## Background

RDNA4 (gfx1200+) introduces the `global_load_tr_b128` / `global_load_tr_b64` ISA
instructions that load 8 contiguous N-direction elements per lane and perform a
hardware 8×8 wave-level cross-lane transpose in a single operation. For BF16/F16
matmuls where the B matrix is stored in row-major order (N-innermost) but the MMA
intrinsic expects K-innermost register layout, this instruction replaces a
`buffer_load` + shared memory transpose with a single global load, saving LDS
bandwidth and reducing barrier pressure.

The MLIR `amdgpu.global_transpose_load` op and its lowering to ROCDL intrinsics
were upstreamed in LLVM: llvm/llvm-project#195287. This issue tracks the IREE-side
work to enable this instruction end-to-end through the GPU codegen pipeline.

## Planned PR Sequence

The work is structured bottom-up so each PR is independently reviewable and testable.

---

### PR 1 — Teach `ROCDLPrefetchSharedMemoryCopy` to recognize `amdgpu.global_transpose_load`

The software-pipelining pass that hoists global memory reads for double-buffering
currently only recognizes `vector.transfer_read` as a global read root. Extend it
to also treat `amdgpu::GlobalTransposeLoadOp` as a pipeable global memory read so
that future PRs get prefetching for free.

**Scope:**
- `ROCDLPrefetchSharedMemoryCopy.cpp`: extend `analyzeIfOp` and
  `identifyRootOperations` to handle `GlobalTransposeLoadOp`
- Test: lit test showing a kernel with `amdgpu.global_transpose_load` gets
  double-buffered the same way a `transfer_read` kernel does

---

### PR 2 — Pattern match `vector.transfer_read + vector.transpose` → `amdgpu.global_transpose_load`

Add a pattern to `ROCDLLoadToTransposeLoad` that matches a `vector<1x8xT>`
transfer_read from flat global memory followed by a `[1,0]` transpose and replaces
it with `amdgpu.global_transpose_load` plus a corrected contiguous write to the
shared memory allocation.

The corrected write indices use K-inner addressing:
`N_new = N_base + K_single % N`, `K_new = (K_single / N) * N`
so the 8 lanes write a contiguous K-direction slice rather than a strided N slice.

**Scope:**
[...]

PR #24457, opened 2026-05-12
[Codegen][AMDGPU] Teach prefetch pass to recognize global_transpose_load as read root

nirvedhmeshram:
Part of #24454.

`ROCDLPrefetchSharedMemoryCopy` performs software pipelining
of loops that stream data from global memory to shared memory. It previously only
recognized `vector.transfer_read` as a global memory read root.

This PR extends the pass to also treat `amdgpu::GlobalTransposeLoadOp` as a read
root, so that loops using `global_load_tr` for transpose loads get pipelined the
same way.


krzysz00 (approved):
LGTM

PR #24460, opened 2026-05-12
[Codegen][GPU] Verify per-allocation swizzle hint consistency

Yu-Zhewen:
`ResolveSwizzleHints` silently drops any hint whose users it can't rewrite (unsupported user op, vector type not divisible by the access width, etc.). This is fine in isolation but breaks down when multiple hints point to the same alloc:
- **Case 1 — alloc with multiple `swizzle_hint` users**: dropping only the subset that can't be resolved leaves some accesses swizzled and others not. That corrupts data on the alloc.
- **Case 2 — alloc with an `amdgpu.gather_to_lds` user**: gather_to_lds is already lowered (by `AMDGPULowerCoalescedDMAToGatherLDS` or `AMDGPULowerAsyncDMA`) to write its
  destination in the swizzled layout. Any silent drop on a reader of that alloc would read garbage.
  
 In this PR, `ResolveSwizzleHints` is refactored to verify hints per allocation.
 - Group hints by their alloc and aggregate per-alloc resolvability plus whether the alloc has a `gather_to_lds` user.
- If the alloc has a `gather_to_lds` user and any hint is unresolvable, emit a hard error (solves case 2).
- Otherwise, if any hint on the alloc is unresolvable, all hints on that alloc drop together (solves case 1).

Two small changes are also needed for the verification above to land cleanly:
- **`ReinsertSwizzleHints`**: also reinsert a `swizzle_hint` on the destination of `amdgpu.gather_to_lds`. The hint is effectively a no-op at resolve time, but reinserting it restores the pre-absorb state and lets the per-alloc check above see the gather_to_lds user.
- **`IREECodegenFoldMemRefAliasOps`**: also run `amdgpu::populateAmdgpuFoldMemRefOpsPatterns` to fold `memref.expand_shape`/`memref.subview` into `amdgpu.gather_to_lds`.

Note: `swizzle_hint` ops reach `ResolveSwizzleHints` via two paths today: absorb + reinsert (#24000, #24001) and direct. The longer-term plan is to resolve swizzling directly on the `memref.alloc` instead of on `swizzle_hint` ops (https://github.com/iree-org/iree/pull/24001#discussion_r3051860902). This is left as a follow-up to keep the verification change in this PR incremental.

Assisted-by: Cursor (Claude)

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
@@ -10,11 +10,12 @@
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Arith/Utils/Utils.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/MemRef/Utils/MemRefUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/AffineExpr.h"
 #include "mlir/IR/AffineMap.h"
-#include "mlir/Interfaces/ViewLikeInterface.h"
+#include "mlir/Interfaces/LoopLikeInterface.h"
 
 namespace mlir::iree_compiler {
 
@@ -155,13 +156,53 @@ verifyFlatContiguousSwizzleHintOp(IREE::Codegen::SwizzleHintOp hintOp) {
   return success();
 }
 
-/// Resolves all hints. Walks all direct users and splits them into loads and
-/// stores. If any user is not a swizzle-able load or store, bail out and
-/// silently drop the optimization hint.
-static void resolveHintOp(RewriterBase &rewriter,
-                          IREE::Codegen::SwizzleHintOp hintOp) {
-  SmallVector<vector::LoadOp> loads;
-  SmallVector<vector::StoreOp> stores;
+/// Traces a memref value backward through defining ops and loop
+/// iter_args/results to find the root memref.alloc.
+static memref::AllocOp traceToAllocation(Value val) {
+  DenseSet<Value> visited;
+  SmallVector<Value> worklist = {val};
+  while (!worklist.empty()) {
+    Value current = worklist.pop_back_val();
+    if (!visited.insert(current).second) {
+      continue;
+    }
+    Operation *defOp = current.getDefiningOp();
+    if (!defOp) {
+      auto blockArg = cast<BlockArgument>(current);
+      auto loopOp =
+          dyn_cast<LoopLikeOpInterface>(blockArg.getOwner()->getParentOp());
+      if (!loopOp) {
+        continue;
+      }
+      if (OpOperand *init = loopOp.getTiedLoopInit(blockArg)) {
+        worklist.push_back(init->get());
+      }
+    } else if (auto allocOp = dyn_cast<memref::AllocOp>(defOp)) {
+      return allocOp;
+    } else if (auto loopOp = dyn_cast<LoopLikeOpInterface>(defOp)) {
+      if (OpOperand *init = loopOp.getTiedLoopInit(cast<OpResult>(current))) {
+        worklist.push_back(init->get());
+      }
+    } else {
+      for (Value operand : defOp->getOperands()) {
+        if (isa<MemRefType>(operand.getType())) {
+          worklist.push_back(operand);
+        }
+      }
+    }
+  }
+  return nullptr;
+}
+
+struct SwizzleState {
+  bool resolvable = true;
+  bool hasGatherToLds = false;
+};
+
+/// Verifies the users of a swizzle hint and returns its swizzle state.
+/// Emits a diagnostic for any unsupported user.
+static SwizzleState verifyHintUsers(IREE::Codegen::SwizzleHintOp hintOp) {
+  SwizzleState state;
   int64_t accessWidth = hintOp.getSwizzle().getAccessElementCount();
   for (Operation *user : hintOp->getUsers()) {
     if (auto load = dyn_cast<vector::LoadOp>(user)) {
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -155,54 +156,95 @@ verifyFlatContiguousSwizzleHintOp(IREE::Codegen::SwizzleHintOp hintOp) {
...
       }
-      stores.push_back(store);
       continue;
     }
-    // Gather_to_lds destination-side swizzle is handled by
-    // AMDGPULowerCoalescedDMAToGatherLDS, which applies the inverse swizzle
-    // to source indices. Treat gather_to_lds and view-like ops as transparent
-    // users that pass through the swizzled allocation.
-    if (isa<amdgpu::GatherToLDSOp, ViewLikeOpInterface>(user)) {
+    if (isa<amdgpu::GatherToLDSOp>(user)) {
```
Muzammiluddin-Syed-ECE, line 224: we no longer skip view like ops because we fold them away prior to resolve swizzle hints however the ordering of the passes seems off to me:

https://github.com/iree-org/iree/blob/05110733b50c2c0faafbe7452ab77f6e8088d33b/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L1016-L1026

IIUC the folding pass `createAmdgpuFoldMemRefOpsPass` is running after resolve hints pass inside `addLowerAndOptimizeAddressComputationPasses`.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -155,54 +156,95 @@ verifyFlatContiguousSwizzleHintOp(IREE::Codegen::SwizzleHintOp hintOp) {
...
+/// iter_args/results to find the root memref.alloc.
+static memref::AllocOp traceToAllocation(Value val) {
+  DenseSet<Value> visited;
+  SmallVector<Value> worklist = {val};
+  while (!worklist.empty()) {
+    Value current = worklist.pop_back_val();
+    if (!visited.insert(current).second) {
+      continue;
+    }
+    Operation *defOp = current.getDefiningOp();
```
Muzammiluddin-Syed-ECE: I dont know if this suggestion is better, up to you

nit: we can replace this temporary variable by using `current.getDefiningOp<T>`

```c++
    if (auto blockArg = dyn_cast<BlockArgument(current)>) {...}
    else if (auto allocOp = current.getDefiningOp<memref::AllocOp>()) {...}
    else if (auto loopOp = current.getDefiningOp<LoopLikeOpInterface>()) {...}
    else {
        for (Value operand : current.getDefiningOp()->getOperands()) {...}
    }
```

compiler/src/iree/compiler/Codegen/Common/test/resolve_swizzle_hints.mlir
```diff
@@ -353,3 +353,38 @@ func.func @swizzle_hint_non_contiguous_memref_error() -> vector<4xf32> {
   %1 = vector.load %0[%offset, %offset] : memref<32x64xf32, strided<[2, 1], offset: 0>>, vector<4xf32>
   return %1: vector<4xf32>
 }
+
+// -----
```
Muzammiluddin-Syed-ECE, line 357: We now expect viewlike ops to have been folded away by the time this pass runs so can we add a negative test to show that having expand_shapes or collapse_shapes between the swizzle hint and the alloc should fail.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
```diff
@@ -155,54 +156,95 @@ verifyFlatContiguousSwizzleHintOp(IREE::Codegen::SwizzleHintOp hintOp) {
...
+    if (!defOp) {
+      auto blockArg = cast<BlockArgument>(current);
+      auto loopOp =
+          dyn_cast<LoopLikeOpInterface>(blockArg.getOwner()->getParentOp());
```
Muzammiluddin-Syed-ECE, line 171: nit: If the immediate parent is an `scf.if` for some reason this fails. Maybe we can try `getParentOfType<LoopLikeOpInterface>()` here.

Muzammiluddin-Syed-ECE (commented):
Thanks Zhewen! Looks good, a few comments

compiler/src/iree/compiler/Codegen/Common/test/resolve_swizzle_hints.mlir
Yu-Zhewen: Done

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
Yu-Zhewen: Thanks! I think we're actually safe here, as`scf.if` won't introduce block arguments to its regions.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
Yu-Zhewen: Thanks, done.

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
Yu-Zhewen: Please refer to the change I made to `IREECodegenFoldMemRefAliasOps`, which runs before `ResolveSwizzleHints`

compiler/src/iree/compiler/Codegen/Common/ResolveSwizzleHints.cpp
Muzammiluddin-Syed-ECE: Ah youre right, even if the block arg is used in an if loop, the parent will always be the defining for loop

Muzammiluddin-Syed-ECE (approved):
LGTM! thanks Zhewen

PR #24462, opened 2026-05-12
[WASM] Add platform support, toolchains, and JS proactor.

benvanik:
This adds a supported path for building the IREE runtime as a freestanding wasm32 module. The commit is intentionally infrastructure-heavy: it teaches Bazel and CMake how to cross-compile IREE C/C++ targets to wasm32-unknown-unknown, provides the minimal C/JS runtime surface those targets need, and adds a JavaScript async proactor so browser/Node integrations can route event-loop completions back into IREE's async API.

WebGPU is the primary near-term target for browser and JavaScript-hosted IREE. The HAL work that follows needs a small runtime that can be embedded in a Web Worker or Node process, instantiate without Emscripten, exchange buffers and callbacks with JavaScript, and keep the generated JS companion code tied to the C libraries that declare wasm imports.

The important product boundary is that this is IREE-owned wasm runtime infrastructure, not an Emscripten build. That keeps the dependency and ABI surface small for embedders that already have their own JavaScript host, allocator, memory ownership rules, or module instantiation path. It also gives us a direct place to model WebGPU/JS completion behavior instead of pretending there is a POSIX kernel underneath the browser.

The initial support contract is deliberately single-threaded. This commit does not claim local-task, pthreads, shared-memory atomics, TLS, or a thread-safe libc/heap. Those are interesting future directions for CPU-side wasm+SIMD execution, but WebGPU is the first target and the current runtime enables only local-sync by default.

* Bazel WebAssembly configuration layered as generic `--config=wasm` settings plus architecture-specific `--config=wasm32` and `--config=wasm32_wasi` platforms. The default wasm32 target model uses `wasm32-unknown-unknown`, freestanding compile/link flags, imported linear memory, no entry point, and local-sync as the only default HAL driver.
* A `--config=wasm32_wasi` variant for targets that need wasi-libc/libc++ behavior such as gtest/gbenchmark-oriented smoke coverage.
* CMake wasm32 toolchain files mirroring the Bazel target model, including deterministic wasi-sdk download/cache handling under the build directory and the same freestanding compile/link flags.
* `iree_wasm_cc_library`, `iree_wasm_entry`, `iree_wasm_cc_binary`, and `iree_wasm_cc_test` rules/macros for Bazel and CMake. C libraries can declare JavaScript companion modules in their deps, binaries collect them transitively, and the bundler emits a single `.mjs` entry point beside the `.wasm` binary.
* A freestanding wasm libc shim: C headers, dlmalloc-backed allocation, `sbrk` over `memory.grow`, string/ctype/strtol/stdio/stdlib/unistd support, and a small `iree_syscall` JavaScript import module for stdout/stderr, clock, exit, and assertion failure paths.
* Hooks for embedders to bring their own allocator or libc instead of IREE's defaults. Bazel exposes the generic `--config=custom_allocator` and `--config=wasm_custom_libc`; CMake exposes `IREE_ALLOCATOR_SYSTEM`, `IREE_WASM_BUILD_IREE_LIBC`, and `IREE_WASM_LIBC_INCLUDE_DIR`.
* JavaScript async proactor support under `runtime/src/iree/async/platform/js/`. The C proactor owns token tables and completion dispatch; JavaScript owns the event-loop integration, Worker/inline entry points, completion ring plumbing, and import modules used by browser or Node hosts.
* Runtime platform tightening for wasm: target-platform detection, disabled unsafe synchronization, no threading by default, wasm-safe time/CSPRNG/sysfs fallbacks, local topology stubs, and build-system selection of the JS proactor where appropriate.
* Smoke tests for the wasm toolchain, the JS bundler, the JS proactor, and the token table used to track externally completed operations.
* A small CMake-based `runtime_wasm` CI job that build-checks representative wasm32 runtime targets with ccache, including a final wasm link of the real `samples/hal/hello` runtime sample, without building LLVM, the compiler, host CMake generators, or running Node-based tests.

Build a freestanding wasm target with Bazel:

    iree-bazel-build \
      --config=wasm32 \
      //build_tools/wasm/libc:libc

Run the Bazel bundler smoke test under Node:

    iree-bazel-test \
      --config=wasm32 \
      //build_tools/wasm/test:bundler_test

Use the WASI-flavored platform when a target needs wasi-libc/libc++ rather than IREE's freestanding libc shim:

    iree-bazel-build \
      --config=wasm32_wasi \
      //some:target

Let an embedding application provide libc headers/objects and the default system allocator:

    iree-bazel-build \
      --config=wasm32 \
      --config=wasm_custom_libc \
      --config=custom_allocator \
      --copt=-DIREE_ALLOCATOR_SYSTEM_CTL=my_allocator_ctl \
      //some:target

Configure a CMake wasm32 build with the same target model:

    cmake -B build-wasm -S . \
      -DCMAKE_TOOLCHAIN_FILE=build_tools/cmake/wasm32.cmake

Build the CMake runtime wasm subset used by CI:

    cmake --build build-wasm \
      --target \
        iree_wasm_libc \
        iree_samples_hal_hello_hello \
        iree_async_platform_js_proactor \
        iree_hal_drivers_local_sync_sync_driver \
      -- -k 0

Embedders that already provide a libc can disable IREE's libc support or point the freestanding include path at their own headers:

    cmake -B build-wasm -S . \
      -DCMAKE_TOOLCHAIN_FILE=build_tools/cmake/wasm32.cmake \
      -DIREE_WASM_BUILD_IREE_LIBC=OFF \
      -DIREE_WASM_LIBC_INCLUDE_DIR=

The freestanding wasm32 configuration builds for `wasm32-unknown-unknown` with no ambient OS services. Memory is imported from the host, the linked wasm module has no C entry point by default, and host integration is explicit through wasm imports and generated JavaScript companion modules.

Generic WebAssembly knobs intentionally avoid encoding the pointer width in their names. `--config=wasm`, `--config=wasm_custom_libc`, `IREE_WASM_BUILD_IREE_LIBC`, and `IREE_WASM_LIBC_INCLUDE_DIR` apply to the wasm embedding model. `wasm32` appears only where selecting the concrete target architecture, target triple, platform constraint, or toolchain file.

The default runtime driver set is local-sync only. local-task remains disabled because it requires a coherent multithreaded contract across wasm shared memory, atomics, TLS, stack management, wakeups, and the allocator/libc. This commit keeps those states unavailable rather than letting unsupported combinations look accidentally buildable.

The freestanding wasm CMake toolchain leaves executable loaders disabled by default. Transfer-only HAL programs such as `samples/hal/hello` do not need VMVX or dynamic executable loading, and enabling VMVX today pulls host schema-generation tools into the cross build. Executable loaders can be enabled explicitly once the build has a proper host-tool/cross-target split for that path.

The bundled libc is intentionally small. It exists to compile and run IREE's runtime substrate, not to emulate a general POSIX environment. Unsupported system calls fail loudly or stay stubbed at the narrow boundary where they are introduced. Applications with an existing libc or allocator can opt out and provide their own implementations through the build-system hooks above.

JavaScript companions are part of the dependency graph. A C target that declares wasm imports owns the corresponding `iree_wasm_cc_library`; final binaries only bundle companion modules whose import names are present in the `.wasm` file. This keeps JS host code colocated with the C ABI it satisfies and avoids manual entry-point assembly for every binary.

The branch has been checked with focused wasm runtime, libc, CI, and build-regeneration paths while rebasing onto the current landing stack:

    cmake -G Ninja -B build-wasm-hello-test -S . \
      -DCMAKE_TOOLCHAIN_FILE=build_tools/cmake/wasm32.cmake \
      -DIREE_WASI_SDK_ROOT=./build-wasm-targeted/wasi-sdk/wasi-sdk-30.0-x86_64-linux \
      -DCMAKE_BUILD_TYPE=MinSizeRel \
      -DCMAKE_C_COMPILER_LAUNCHER=ccache \
      -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
      -DIREE_RUNTIME_OPTIMIZATION_PROFILE=size \
      -DIREE_ENABLE_LLD=ON

    cmake --build build-wasm-hello-test \
      --target \
        iree_wasm_libc \
        iree_samples_hal_hello_hello \
        iree_async_platform_js_proactor \
        iree_hal_drivers_local_sync_sync_driver \
      -- -k 0

    file build-wasm-hello-test/samples/hal/hello/hello

    llvm-readobj \
      --symbols build-wasm-hello-test/samples/hal/hello/hello

    iree-bazel-build \
      --config=wasm32 \
      //build_tools/wasm/test:smoke_test \
      //build_tools/wasm/test:bundler_test_bundle \
      //runtime/src/iree/base:base \
      //runtime/src/iree/async/platform/js:proactor \
      //runtime/src/iree/hal/drivers/local_sync:sync_driver

The CMake link evidence produces `samples/hal/hello/hello` as a wasm32 module. `llvm-readobj` shows IREE libc functions such as `malloc`, `free`, `fprintf`, `strlen`, and `strcmp` are defined in the module; the remaining undefined symbols are the explicit wasm host imports from the libc syscall shim.

A full CMake wasm `all` build was evaluated and is not the right CI hook yet: it currently reaches host generator/tooling targets such as `iree-flatcc-cli` and C++ runtime utilities under the wasm compiler. The CMake `runtime_wasm` job uses an explicit target subset instead, still proving the wasm toolchain, the freestanding libc archive, a final link of the real HAL hello runtime sample, core runtime base code, the JS proactor, and the local-sync driver.


kuhar (approved):
I skimmed through all the changes and nothing stood out to me

PR #24467, opened 2026-05-13
[Codegen][AMDGPU] Pattern match transfer_read+transpose to global_transpose_load for RDNA4

nirvedhmeshram:
Part of #24454.

## Summary

Add `TransferReadTransposeToGlobalTransposeLoad` pattern to `ROCDLLoadToTransposeLoad`
that recognizes the IR produced by K-inner tiled copy promotions on RDNA4:

```
%0 = vector.transfer_read %src[%k, %n], %cst : memref<..., global>, vector<1x8xT>
%1 = vector.transpose %0, [1, 0]             : vector<1x8xT> to vector<8x1xT>
vector.transfer_write %1, %dst[%n, %k]       : vector<8x1xT>, memref<..., workgroup>
```

and replaces it with `amdgpu.global_transpose_load` on gfx1200+ targets.

The `global_load_tr_b128` hardware instruction performs an 8×8 wave-level
cross-lane transpose: 8 consecutive lanes each read a row of 8 N-elements,
and the hardware transposes the result so each lane holds a K-direction slice.
The write indices are recomputed so the output goes to the correct location
in the shared memory allocation with contiguous K access:
- `N_new = N_base + K_single % N`
- `K_new = (K_single floordiv N) * N`

The pass now dispatches to the gfx950 path (LDS transpose, existing) and
the gfx1200+ path (global transpose, new) independently.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
@@ -917,31 +917,28 @@ struct TransferReadTransposeToGlobalTransposeLoad final
         rewriter, loc, resultVecType, src, transferOp.getIndices());
 
     // Compute corrected write indices for contiguous K writes:
-    //   N_new = N_base + K_single % N      (lane's N position within N_base
-    //   group) K_new = (K_single // N) * N        (K-group base, aligned to N)
-    // Write vector<1xNxT> at [N_new, K_new] → alloc_8[N_new, K_new..K_new+N-1]
-    // This is contiguous K in alloc_8[N, K] (K is inner) — no subview needed.
+    // Compute corrected write indices for contiguous K writes.
+    // Delinearize K_single into [K_group, K_offset] where K_group = K_single /
+    // N and K_offset = K_single % N. Then:
+    //   N_new = N_base + K_offset   (lane's N position within the N-group)
+    //   K_new = K_group * N         (K-group base, aligned to N)
     ValueRange writeIndices = writeOp.getIndices();
     assert(writeIndices.size() == 2 && "expected 2D write");
-    Value nBase = writeIndices[0];   // N_group * N
-    Value kSingle = writeIndices[1]; // K lane value (0..K_total-1)
+    Value nBase = writeIndices[0];
+    Value kSingle = writeIndices[1];
 
-    AffineExpr dn = rewriter.getAffineDimExpr(0);
-    AffineExpr dk = rewriter.getAffineDimExpr(1);
-    AffineMap nNewMap = AffineMap::get(2, 0, dn + dk % N);
-    AffineMap kNewMap = AffineMap::get(2, 0, (dk.floorDiv(N)) * N);
+    Value nVal = arith::ConstantIndexOp::create(rewriter, loc, N);
+    auto delinOp = affine::AffineDelinearizeIndexOp::create(
+        rewriter, loc, kSingle, SmallVector<Value>{nVal},
+        /*hasOuterBound=*/false);
+    Value kGroup = delinOp.getResult(0);
+    Value kOffset = delinOp.getResult(1);
+    Value nNew = arith::AddIOp::create(rewriter, loc, nBase, kOffset);
+    Value kNew = arith::MulIOp::create(rewriter, loc, kGroup, nVal);
 
-    Value nNew = affine::AffineApplyOp::create(rewriter, loc, nNewMap,
-                                               ValueRange{nBase, kSingle});
-    Value kNew = affine::AffineApplyOp::create(rewriter, loc, kNewMap,
-                                               ValueRange{nBase, kSingle});
-
-    VectorType writeVecType = VectorType::get({1, N}, elemType);
-    Value castResult = vector::ShapeCastOp::create(rewriter, loc, writeVecType,
-                                                   trLoad.getResult());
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -751,6 +754,199 @@ struct TransferReadToTransposeLoad final
...
+    // This is contiguous K in alloc_8[N, K] (K is inner) — no subview needed.
+    ValueRange writeIndices = writeOp.getIndices();
+    assert(writeIndices.size() == 2 && "expected 2D write");
+    Value nBase = writeIndices[0];   // N_group * N
+    Value kSingle = writeIndices[1]; // K lane value (0..K_total-1)
+
+    AffineExpr dn = rewriter.getAffineDimExpr(0);
+    AffineExpr dk = rewriter.getAffineDimExpr(1);
+    AffineMap nNewMap = AffineMap::get(2, 0, dn + dk % N);
+    AffineMap kNewMap = AffineMap::get(2, 0, (dk.floorDiv(N)) * N);
```
krzysz00: ... can we represent these floordivs and mods with a `delinearize_index`

I don't trust the affine map simplifier with mods and divs - or we can just outright emit `dk - (dk % N)` as a value.

krzysz00: But that's nitpickery

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -751,6 +754,199 @@ struct TransferReadToTransposeLoad final
...
+
+    Value nNew = affine::AffineApplyOp::create(rewriter, loc, nNewMap,
+                                               ValueRange{nBase, kSingle});
+    Value kNew = affine::AffineApplyOp::create(rewriter, loc, kNewMap,
+                                               ValueRange{nBase, kSingle});
+
+    VectorType writeVecType = VectorType::get({1, N}, elemType);
+    Value castResult = vector::ShapeCastOp::create(rewriter, loc, writeVecType,
+                                                   trLoad.getResult());
+    vector::TransferWriteOp::create(rewriter, loc, castResult,
```
krzysz00: ... you shouldn't need the extra unit dimension here

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -768,36 +964,53 @@ struct ROCDLLoadToTransposeLoadPass final
...
+      if (workgroupSize) {
+        seedThreadIdHints(funcOp, rewriter, *workgroupSize);
+      }
+      // Propagate hints and lower transfer_reads via greedy patterns.
+      patterns
+          .add<PropagateHintThroughDelinearize, TransferReadToTransposeLoad>(
+              funcOp.getContext());
+    }
+
+    if (isRDNA4) {
```
krzysz00, line 998: This'll work on gfx1250 as well, as far as I can tell, even though we probably shouldn't use this codepath there in preference to TDM + transposing load from LDS

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
```diff
@@ -768,36 +964,53 @@ struct ROCDLLoadToTransposeLoadPass final
...
     }
 
-    // Phase 2: Propagate hints and lower transfer_reads via greedy patterns
+    IRRewriter rewriter(funcOp.getContext());
+
     RewritePatternSet patterns(funcOp.getContext());
-    patterns.add<PropagateHintThroughDelinearize, TransferReadToTransposeLoad>(
-        funcOp.getContext());
+
+    if (isGfx950) {
```
krzysz00, line 985: ... quick check - while we're doing all this work for global transpose, let's make sure transpose from LDS works too since I'm pretty sure gfx12 has those and we'll need them for TDM

krzysz00 (approved):
Broad comments but I don't see any issue with the specifics

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
nirvedhmeshram: yeah makes sense to just use it for RDNA4 for now, and make the call on if we want to use it for gfx1250 when we get there.

compiler/src/iree/compiler/Codegen/LLVMGPU/ROCDLLoadToTransposeLoad.cpp
nirvedhmeshram: RDNA4 ISA doesnt have any lds transpose instructions, so for now we cant enable that here.

krzysz00: gfx1250 does - and I'm poking the upstream support for those today

nirvedhmeshram: makes sense, we can enable it for gfx1250 if and when we enable that.

PR #24470, opened 2026-05-13
[Codegen][Tuner] Materialize constraint assignments in compiler

kuhar:
Add compiler bindings that materialize a concrete compilation_info attr from an iree_codegen.smt.constraints op and a flat knob assignment dictionary. This gives the tuner a compiler-owned path for turning solver models back into the attrs the compiler consumes.

Keep the common helper mechanical: it only substitutes IntKnobAttr and OneOfKnobAttr leaves, then asks the pipeline attr to repackage the materialized knob dictionary. GPU pipelines own their lowering_config and translation_info spelling, while unsupported pipelines use the default interface failure instead of generic dictionary guessing.

Add lit and Python binding coverage for successful materialization and invalid assignments.

Assisted-by: codex


compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
```diff
@@ -65,6 +65,56 @@ using ::mlir::iree_compiler::IREE::Codegen::kVDMFMADeinterleaveAcc;
 using ::mlir::iree_compiler::IREE::Codegen::kVDMFMAInterleaveAcc;
 using ::mlir::iree_compiler::IREE::Codegen::TileSwizzle;
 
+constexpr StringLiteral kWorkgroupSizeKey = "workgroup_size";
+constexpr StringLiteral kSubgroupSizeKey = "subgroup_size";
+constexpr StringLiteral kTranslationInfoKey = "translation_info";
+constexpr StringLiteral kPipelineKey = "pipeline";
+constexpr StringLiteral kCodegenSpecKey = "codegen_spec";
+
+static bool isTranslationInfoKey(StringRef key) {
```
bangtianliu, line 74: `isTranslationInfoKey` actually returns true only for the sub-fields of a translation_info(`workgroup_size`, `subgroup_size`) 

Rename to `isTranslationInfoSubKey`?

bangtianliu (approved):

compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp
kuhar: I feel like this is one of those instances where there might not be a good name 🤷‍♂️  The intention is to detect whether the key belongs do translation info. It's a local function so I'm not too concerned -- the intent is clear based on the local usage.

hanhanW:
It seems that there is a compilation failure: https://github.com/iree-org/iree/actions/runs/25833014665/job/75902985895

It does not happen in [the commit right before it](https://github.com/iree-org/iree/commit/58b18c5cbd4fd6d4bb5edbdebb01a0356a1d2bc8): https://github.com/iree-org/iree/actions/runs/25826532741/job/75882582556

kuhar:
> It seems that there is a compilation failure: https://github.com/iree-org/iree/actions/runs/25833014665/job/75902985895
> 
> It does not happen in [the commit right before it](https://github.com/iree-org/iree/commit/58b18c5cbd4fd6d4bb5edbdebb01a0356a1d2bc8): https://github.com/iree-org/iree/actions/runs/25826532741/job/75882582556

This is an NFC for the main compilation path. It seems like a runner issue:
```
/shark-cache/data/iree-regression-cache/artifacts/sdxl_scheduler/model.mlir:1:1: error: failed to create file output buffer at /shark-cache/data/iree-regression-cache/artifacts/sdxl_scheduler/fake_weights.irpa with error: No space left on device
module @compiled_scheduler {
^
```

PR #24472, opened 2026-05-13
[HAL] Implement functions to normalize buffer references and buffer bindings to subspan buffers

schuermans-roofline:
Buffer references and bindings can point to subspans of buffers. HAL drivers need to handle this case and access the actual memory allocation in the underlying allocated buffer. This can be handled by normalizing buffer references and buffer bindings before processing those. Normalization means to replace the reference to (a part of) a subspan buffer with a reference to the corresponding part of the underlying allocated buffer in the buffer reference or buffer binding.

This change provides functions that perform those normalizations as helpers/infrastructure for HAL drivers.


devtbi (approved):
LGTM

maxbartel (approved):

schuermans-roofline:
I made a mistake and deleted the branch in my private repo. This was not intentional.

PR #24474, opened 2026-05-13
Guard against NULL frontier_tracker in standalone HIP/CUDA devices

Abhishek-Varma:
`iree_hal_hip_device_advance_frontier` (and its CUDA counterpart) is called on every `queue_execute`, including plain buffer transfers. For standalone devices, not assigned to a device group via `iree_hal_device_assign_topology_info`, `frontier_tracker` is `NULL, causing a crash during invocation of `iree_hal_hip_device_advance_frontier`.

This commit adds a NULL early-return guard.


AGindinson (approved):
LGTM

PR #24476, opened 2026-05-13
Fix MapLoad for dynamic shapes

### LINKED ISSUE #24461: map_load loses dynamic result size
Repro:
```mlir
func.func @dynamic_extract_expand_pad(%buffer : memref<50x32x16xf32>,
                                      %offset : index,
                                      %size : index) -> tensor<?x1x1x1x4xf32> {
  %source = iree_codegen.load_from_buffer %buffer
      : memref<50x32x16xf32> -> tensor<50x32x16xf32>
  %slice = tensor.extract_slice %source[%offset, 0, 0] [%size, 32, 16] [1, 1, 1]
      : tensor<50x32x16xf32> to tensor<?x32x16xf32>
  %expanded = tensor.expand_shape %slice [[0], [1, 2], [3, 4]]
      output_shape [%size, 2, 16, 1, 16]
      : tensor<?x32x16xf32> into tensor<?x2x16x1x16xf32>
  %subslice = tensor.extract_slice %expanded[0, 0, 0, 0, 0]
      [%size, 1, 1, 1, 4] [1, 1, 1, 1, 1]
      : tensor<?x2x16x1x16xf32> to tensor<?x1x1x1x4xf32>
  %padded = tensor.pad %subslice low[0, 0, 0, 0, 0] high[0, 0, 0, 0, 0] {
  ^bb0(%i: index, %j: index, %k: index, %l: index, %m: index):
    tensor.yield %cst : f32
  } : tensor<?x1x1x1x4xf32> to tensor<?x1x1x1x4xf32>
  return %padded : tensor<?x1x1x1x4xf32>
}
```

Compile with:
```shell
iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-combine-source-layout-transformation,canonicalize,cse))" repro.mlir
```

```
<stdin>:9:15: error: operand #0 does not dominate this use
  %expanded = tensor.expand_shape %slice [[0], [1, 2], [3, 4]]
              ^
<stdin>:9:15: note: see current operation: %5 = "tensor.dim"(%7, %1) : (tensor<?x32x16xf32>, index) -> index
<stdin>:7:12: note: operand defined here (op in the same block)
  %slice = tensor.extract_slice %source[%offset, 0, 0] [%size, 32, 16] [1, 1, 1]
           ^
```

The wrong part is: while folding the chain, main materializes a tensor.dim for the expand_shape input %slice at an insertion point before %slice exists. So the generated tensor.dim(%slice, ...) does not dominate its use, and MLIR rejects the IR.

I'd expect the relayout chain to fold into one `iree_linalg_ext.map_load`, and the replacement destination to preserve the dynamic result size, like

```mlir
%dest = tensor.empty(%size) : tensor<?x1x1x1x4xf32>
%result = iree_linalg_ext.map_load %source into %dest ...
```

I believe this was introduced in https://github.com/iree-org/iree/pull/23165 by @Abhishek-Varma

Abhishek-Varma:
`combine-source-layout-transformation` was building each new `map_load`'s `tensor.empty` destination with `tensor::getMixedSizes(consumer.getResult())`.
For a dynamically-shaped consumer (e.g. `extract_slice → expand_shape →
extract_slice → pad`), once the consumer was replaced by the new `map_load`, the materialized `tensor.dim %consumer_result, i` queries silently became `tensor.dim %new_map_load, i` and fed the very `tensor.empty` driving that `map_load`'s destination : yielding both an SSA dominance violation and a self-referential cycle (`tensor.empty(tensor.dim(%new_ml, 0))` with `%new_ml
= map_load ... into %that_empty`).

This PR fixes the above by routing all such size queries through reifyShapeOfResult, so dynamic dims are expressed in terms of the consumer op's *operands* and never the result that's about to be replaced.

Fixes #24461.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -23,6 +23,7 @@
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/IRMapping.h"
 #include "mlir/Interfaces/DestinationStyleOpInterface.h"
+#include "mlir/Interfaces/InferTypeOpInterface.h"
 #include "mlir/Interfaces/ValueBoundsOpInterface.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
@@ -1030,12 +1031,20 @@ static FailureOr<MapLoadOp> foldConsumerIntoMapLoadImpl(
   OpBuilder::InsertionGuard g(rewriter);
   rewriter.setInsertionPoint(consumerOp);
 
-  // Create new dest tensor matching consumer output shape.
+  // Reify against operands since we are about to replace `consumerOp` with
+  // `newMapLoad`, so dim queries on the result would loop back into the new
+  // map_load's destination.
   Value consumerResult = consumerOp->getResult(0);
   Type elementType = getElementTypeOrSelf(consumerResult.getType());
-  SmallVector<OpFoldResult> newSizes =
-      tensor::getMixedSizes(rewriter, loc, consumerResult);
-  Value newDest = tensor::EmptyOp::create(rewriter, loc, newSizes, elementType);
+  FailureOr<SmallVector<OpFoldResult>> newSizes =
+      reifyShapeOfResult(rewriter, consumerOp, /*resultIndex=*/0);
+  if (failed(newSizes)) {
+    return rewriter.notifyMatchFailure(
+        consumerOp,
+        "consumer does not implement ReifyRankedShapedTypeOpInterface");
+  }
+  Value newDest =
+      tensor::EmptyOp::create(rewriter, loc, *newSizes, elementType);
 
   // Clone the map_load with new dest.
   auto newMapLoad = MapLoadOp::create(rewriter, loc, newDest.getType(),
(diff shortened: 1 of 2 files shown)


kuhar (approved):
I tried this on the full model where I observed the verifier failure and it passes with this PR. Thanks for fixing.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -1030,12 +1031,20 @@ static FailureOr<MapLoadOp> foldConsumerIntoMapLoadImpl(
...
+  // Reify against operands since we are about to replace `consumerOp` with
+  // `newMapLoad`, so dim queries on the result would loop back into the new
+  // map_load's destination.
   Value consumerResult = consumerOp->getResult(0);
   Type elementType = getElementTypeOrSelf(consumerResult.getType());
-  SmallVector<OpFoldResult> newSizes =
-      tensor::getMixedSizes(rewriter, loc, consumerResult);
-  Value newDest = tensor::EmptyOp::create(rewriter, loc, newSizes, elementType);
+  FailureOr<SmallVector<OpFoldResult>> newSizes =
+      reifyShapeOfResult(rewriter, consumerOp, /*resultIndex=*/0);
```
hanhanW, line 1039: Note that this method could still [insert ops that compute result shapes](https://github.com/llvm/llvm-project/blob/7091f885fdee2987c71dbb7b0fdbb639bd78ab40/mlir/include/mlir/Interfaces/InferTypeOpInterface.td#L383-L394). IIUC, the implementation (of tensor ops) returns the value directly because they have them in their operands.

```
        Returns the shape of a single result of the operation as a
        `SmallVector<OpFoldResult>`, one per dimension of the shaped type. The
        given builder may be used to insert ops that compute result shapes.
```

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -1122,14 +1131,23 @@ static FailureOr<MapLoadOp> foldReshapeIntoMapLoad(RewriterBase &rewriter,
...
-  SmallVector<OpFoldResult> resultDims =
-      tensor::getMixedSizes(rewriter, loc, reshapeOp.getResult());
+  SmallVector<OpFoldResult> srcDims;
+  FailureOr<SmallVector<OpFoldResult>> resultDims;
+  {
+    OpBuilder::InsertionGuard g(rewriter);
+    rewriter.setInsertionPoint(reshapeOp);
+    srcDims = tensor::getMixedSizes(rewriter, loc, mapLoadOp.getOutput());
+    resultDims = reifyShapeOfResult(rewriter, reshapeOp, /*resultIndex=*/0);
+  }
```
hanhanW: Do you need a local scope here? This function is small, and it can always sets the insertion point as long as it recovers it before exiting the function.

Maybe you also want to add a guard is in the below lambda? So we have two `InsertionGuard` in total; we can have a clear ownership.

hanhanW (approved):

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Good catch. I've updated the comment.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
Abhishek-Varma: Makes sense. Done.

PR #24478, opened 2026-05-13
[LinalgExt] Classify im2col convolution batch dims as m_pos dims

Max191:
Redefines the classification of convolution dimensions in the im2col op when converting convolution to GEMM + im2col, and fixes a semantic inconsistency with the treatment of convolution batch dims in the im2col op. Batch dims in the convolution only appear in the image operand, so they actually become M dims in the resulting GEMM, but the old classification would place them in the batch_pos list.

This PR fixes the semantic error and puts convolution batch dims into m_pos instead. In order to preserve the consistent handling of m_pos dims in the im2col lowerings and verification, the batch dims are given synthetic filter window metadata (unit strides, dilations, and window size), which simplifies the im2col op lowering and verification implementation. There is no reason the im2col op needs to know that the dim came from a convolution batch dim, since im2col only needs a well defined access pattern, for which this synthetic metadata is sufficient.

Benchmarks were tested locally on MI350, and there is no impact on performance. This is expected, since layouts and the results of codegen do not change, and this is only a change to the im2col metadata.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ConvertConvToIm2ColOp.cpp
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ConvertConvToIm2ColOp.cpp
@@ -12,6 +12,7 @@
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
 #define DEBUG_TYPE "iree-linalg-ext-convert-conv-to-im2col-op"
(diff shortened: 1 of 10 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ConvertConvToIm2ColOp.cpp
```diff
@@ -177,47 +245,60 @@ class ConvertConvGeneric final
...
+      if (std::optional<unsigned> outputImage =
+              findFunctionOfDim(inputExpr, convDims.outputImage)) {
+        std::optional<int64_t> outputDim =
+            getResultPositionForDim(outputMap, *outputImage);
+        if (!outputDim) {
+          return rewriter.notifyMatchFailure(
+              linalgOp, "Failed to infer output image shape.");
+        }
+        auto outputImageIt = outputImageDimToIndex.find(*outputImage);
+        if (outputImageIt == outputImageDimToIndex.end()) {
```
bangtianliu: Seems this condition will always be true, maybe drop the check?

nirvedhmeshram (approved):
LGTM

compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/ConvertConvToIm2ColOp.cpp
Max191: Whoops, codex amended instead of adding a commit, but got rid of the check, thanks!

bangtianliu (approved):

Issue #24483, opened 2026-05-13
`hoistOneStaticallyBoundAllocation` produces bogus 2^52-element static `alloca`s when dynamic dim bounds default to the `index` umax


bjacob:
(Note: all of this is generated by AI).

## Summary

`hoistStaticallyBoundAllocations` ([compiler/src/iree/compiler/Codegen/Common/HoistStaticallyBoundAllocations.cpp](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/HoistStaticallyBoundAllocations.cpp)) accepts *any* result of `ValueBoundsConstraintSet::computeConstantBound` — including the trivial upper bound `2^53 - 1` that any `index` value carries when it has no other constraint — as the static dim of a hoisted `memref.alloca`. With one `affine.apply` of `ceildiv 2` between the unconstrained value and the alloca, that produces a `memref.alloca() : memref<4503599627370496xi32>` (a 2^52-element static-shape stack allocation ≈ 16 PiB), silently, with no diagnostic. Downstream codegen then computes strides as products of these dims (overflowing `i64` to poison) and the LLVM lowering ends up with `llvm.alloca poison`.

This isn't theoretical — it's hit by the CPU pack pipeline whenever the encoder swizzle materializes a rank-N inner tile under a dynamic-shape dispatch, see "How this blocks us" below.

## Minimal reproducer

`repro.mlir`:

```mlir
// Reproduces hoistStaticallyBoundAllocations producing a bogus
// 2^52-element static alloca when the dynamic dim's upper bound
// is the default `index` umax (2^53 - 1).

func.func @hoist_with_umax_bounded_dim(%arg0: index) {
  %c0 = arith.constant 0 : index
  %c1 = arith.constant 1 : index
  %c42 = arith.constant 42 : i32
  // What iree-stream-annotate-dispatch-assumptions emits for an
  // unconstrained dispatch workload operand:
  %assumed = util.assume.int %arg0<umin = 0, umax = 9007199254740991> : index
  // What data tiling emits when sizing a packed-outer dim:
  %dim = affine.apply affine_map<()[s0] -> (s0 ceildiv 2)>()[%assumed]
  // Force the alloca outside the entry block so the hoist pass picks it up.
  scf.for %iv = %c0 to %c1 step %c1 {
    %a = memref.alloca(%dim) : memref<?xi32>
    memref.store %c42, %a[%iv] : memref<?xi32>
    scf.yield
  }
  return
}
[...]

PR #24490, opened 2026-05-14
[Codegen] Skip swizzle transpose when permutation is a no-op modulo unit dims.

bjacob:
The set_encoding pack lowering (and symmetrically the unset_encoding unpack) currently always emits

  linalg.pack → tensor.expand_shape → linalg.transpose

to apply the swizzle's `permutation` to the packed inner tile. For many swizzles the permutation only reorders unit dims relative to non-unit dims (e.g. the generic-scalar 1×1×1 MMA ACC swizzle has `expandShape = [[CrossIntrinsic(M), Internal(1)], [CrossIntrinsic(N), Internal(1)]]` with `permutation = [0, 2, 1, 3]` — swapping the `Internal(1)` placeholders past the `CrossIntrinsic` non-unit dims). The linearized memory layout of the unpermuted and permuted shapes is identical there, so the `linalg.transpose` is a no-op in terms of data movement.

Add `isSwizzlePermutationLayoutNoOp`, a predicate that detects this case (the permutation, restricted to non-unit source dims, is order-preserving), and `getSwizzleFoldedReassociation`, which — given that predicate holds — builds the reassociation that drops the `linalg.transpose`: the `tensor.expand_shape` (pack side) / `tensor.collapse_shape` (unpack side) is instead given a reassociation that lands directly on the permuted shape, partitioning the permuted inner dims into contiguous slices matching the pack's `innerTileSizes` one-to-one. Fall through to the existing expand + transpose lowering when the permutation actually reorders non-unit dims — the gfx950 scaled-matmul case for instance, where the permutation moves a size-16 dim past two size-4 dims, must keep the transpose.

The folded reshape still carries the swizzle's unit dims: the permuted inner shape is the materialized encoding type fixed by `getEncodingInfo`, which dialect conversion requires the replacement value to match exactly — dropping unit dims is not a local choice here. Hence `getSwizzleFoldedReassociation` only picks a reassociation reaching that already-fixed shape; whether the materialized type should carry unit dims at all is a separate, cross-cutting question.

This was motivated by a separate bug in
`hoistOneStaticallyBoundAllocation` (#24483): under dynamic shapes the intermediate `tensor.empty` between the pack and the transpose bufferizes to a stack alloca whose size comes from `ValueBoundsConstraintSet::computeConstantBound` on the dynamic dim, which for a workload-ordinal dim with only the default `index` umax constraint (2^53 - 1) produces a ~2^52-element static alloca, overflows i64 stride math, and lowers to `llvm.alloca poison`. Eliminating the intermediate via this fast path sidesteps the bug for the cases that hit this path; the hoist bug is real and the fix here is in addition to fixing it at the source.

Progress towards #24323.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
@@ -302,6 +302,61 @@ func.func @matmul_bf16_f32_avx2_generic(
 
 // -----
 
+// The ACC swizzle for the same generic-scalar config above
+// (intrinsics_m=2, intrinsics_n=4) has `expandShape = [[Cross(2),
+// Int(1)], [Cross(4), Int(1)]]` with `permutation = [0, 2, 1, 3]` —
+// non-identity, but a no-op modulo unit dims (the permutation only
+// reorders the `Internal(1)` placeholders relative to the non-unit
+// `CrossIntrinsic` dims). The set_encoding lowering recognizes this
+// and emits a single `tensor.expand_shape` from the rank-4 pack output
+// directly into the rank-6 permuted layout, with no `linalg.transpose`.
+// The unset_encoding side is symmetric: a single `tensor.collapse_shape`
+// from the rank-6 permuted source directly into the rank-4 unpack input.
+// Iteration_sizes are pinned to (M=2, N=4, K=?) so the cost model lands
+// on intrinsics_m=2, intrinsics_n=4 regardless of whether the
+// power-of-two cap on `chooseUnrolling` is active.
+
+#map_g_no = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map_g_no1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map_g_no2 = affine_map<(d0, d1, d2) -> (d0, d1)>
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/test/materialize_encoding_x86_64.mlir
```diff
@@ -302,6 +302,61 @@ func.func @matmul_bf16_f32_avx2_generic(
...
+// CHECK-LABEL: func @set_encoding_RESULT_avx2_generic_no_op_transpose
+//       CHECK:   %[[PACK:.+]] = linalg.pack
+//  CHECK-SAME:     inner_dims_pos = [0, 1] inner_tiles = [2, 4]
+//  CHECK-SAME:     -> tensor<1x1x2x4xf32>
+//       CHECK:   %[[EXPAND:.+]] = tensor.expand_shape %[[PACK]]
+//  CHECK-SAME{LITERAL}: [[0], [1], [2], [3, 4, 5]]
+//  CHECK-SAME:     : tensor<1x1x2x4xf32> into tensor<1x1x2x4x1x1xf32>
+//   CHECK-NOT:   linalg.transpose
+//       CHECK:   return %[[EXPAND]] : tensor<1x1x2x4x1x1xf32>
+
```
hanhanW, line 340: I'm thinking if we still add a seperator (i.e., `// -----`) between two test cases. IIUC, you don't make the seperator because they share the same affine maps and encodings. They are just four lines, so I think we can still have the seperator.

The other option is keeping it as what it is, and I think you should declare an alias for executable target. So it can be used in both cases.

hanhanW (approved):

PR #24493, opened 2026-05-14
[Codegen][CPU] Support distinct LHS/RHS types in e2e matmul test framework.

bjacob:
The framework so far only generated symmetric-type matmuls (LHS and RHS of the same type via `--lhs_rhs_type`). This generalizes it to support `--lhs` and `--rhs` separately so we can test asymmetric-type intrinsics like the upcoming x86 AVX-512 VNNI vpdpbusd (LHS=ui8, RHS=i8 → ACC=i32).

- `generate_e2e_matmul_tests.py`: replace `--lhs_rhs_type` with the per-operand `--lhs` / `--rhs` pair (legacy form still accepted; the CLI requires exactly one form). Adds a `ui8` choice. Plumbing for separate `lhs_type`/`rhs_type` is threaded through `generate()` / `generate_function()` / `generate_call()`.
- `generate_code.py`: emit a `linalg.generic` (with `arith.extui` / `arith.extsi` widening) when LHS != RHS or an operand is unsigned; buffer-view printing uses the `ui8` HAL element type for unsigned operands.
- `iree-e2e-matmul-test.cc`: new `REFERENCE_MATMUL(uint8_t, int8_t, int32_t, int32_t)` reference and matching dispatch arm for `IREE_HAL_ELEMENT_TYPE_UINT_8` LHS / signed-i8 RHS.

New test `e2e_matmul_cpu_dt_inner_tiled_ui8_i8_i32` exercises the generated ui8/i8 matmul through the data-tiling + inner-tiled pipeline. No matching hardware MMA intrinsic exists yet (this commit is purely framework), so it only runs on the `generic` CPU variant and falls through to the type-polymorphic generic-scalar MMA. A follow-up PR adds the vpdpbusd MMA and an `avx512vnni` variant.

Progress towards #24323.

### DIFF
--- a/tests/e2e/matmul/BUILD.bazel
+++ b/tests/e2e/matmul/BUILD.bazel
@@ -259,6 +259,38 @@ X86_64_AVX512_BF16 = X86_64_AVX512 + [
     ("bf16", "f32"),
 ]]
 
+# Mixed unsigned-LHS / signed-RHS i8 matmul. The test framework otherwise
+# only generates symmetric-type matmuls. No matching MMA intrinsic exists
+# yet on any of the `target_cpu_features_variants` here — the test
+# exercises the standard codegen fallback through the data-tiling +
+# inner-tiled pipeline.
+iree_generated_e2e_runner_test(
+    name = "e2e_matmul_cpu_dt_inner_tiled_ui8_i8_i32",
+    compiler_flags = [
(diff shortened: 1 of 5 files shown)


bjacob:
Merged as part of https://github.com/iree-org/iree/pull/24494.

PR #24497, opened 2026-05-15
[Codegen] Gate swizzle transpose fold to CPU (no CrossThread dims).

bjacob:
The transpose-fold fast path introduced in the parent commit elides the `linalg.transpose` produced by the general encoding-materialize lowering when the permutation is a layout no-op. On GPU this has been suspected to cause MI350 (CDNA4) flakes on the `coalesced_dma_f32` matmul tests — sometimes segfaults, sometimes a result accumulator filled with zeros, suggesting an LDS layout mismatch where downstream GPU codegen (coalesced DMA, LDS layout derivation) still expects the explicit `linalg.transpose` and ends up staging data into shared memory under a different layout than the MMA consumes.

The original motivation for the fold was a dynamic-shape stack alloca symptom (#24483) that is CPU-only, so restricting the fold to CPU loses no benefit there.

CPU vs GPU is distinguished structurally on the `TileSwizzle`: GPU swizzles carry `CrossThread` dims (the kind doc string says so by definition), CPU swizzles only carry `Internal` and `CrossIntrinsic` dims. Skip the fast path when any `CrossThread` dim is present.

Progress towards #24323.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/MaterializeEncodingPatterns.cpp
@@ -507,6 +507,25 @@ getReassociationIndices(int outerDims,
   return result;
 }
 
+/// Returns true if any dim of the swizzle's `expandShape` is a `CrossThread`
+/// dim, which by definition only occurs on SIMT (GPU) targets — CPU swizzles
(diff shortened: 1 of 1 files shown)


bjacob:
Closing: the regression turned out to be a preexisting condition, not caused by the swizzle fold. This gate is not needed.

PR #24504, opened 2026-05-16
[Codegen] Use default read semantics for LinalgExt scatter

moomoohorse321:
The original impl. will introduce critical but very subtle bugs. We identified the bug when running Qwen-3.5, with decode-prefill-decode pattern. Because of the problem detailed below, the 2nd decode are completely wrong.

## Problem

  `iree_linalg_ext.scatter` **was** special-cased in the LinalgExt bufferization
  external model as not reading its DPS init operand which is not correct.

  ## Fix

  Remove the ScatterOp-specific no-read override and use the default
  DestinationStyleOpInterface bufferization read semantics.

  After this change, out-of-place bufferization preserves the init value before
  scatter updates it:
```
  alloc tmp[32][1][512]
  copy original -> tmp // which was previously dropped.
  scatter writes 11 rows into tmp
```

  ### Minimal Repro

  A simple overwrite scatter still needs to preserve elements that are not updated:

  ```text
  result = original // this was incorrectly dropped if it's not fixed with the config shown in test cases attached
  for i in updated_indices:
    result[i] = update[i]
```

  Example shape:

  original: 32 x 1 x 512
  updates : 11 x 1 x 512
  indices : 11 rows
  result  : 32 x 1 x 512

  Scatter writes only 11 rows. The remaining 21 rows must come from original.

  **Before this change, bufferization thinks `outs(%original)` as a destination
  placeholder instead of a read. So it stores tmp[32][1][512] to output**

### Analysis

I propose that we remove the override because scatter op should be bufferized as reads except for some small corner cases. Overriding the scatter op does not preserve correctness.

Let's analyze the following cases for scatter:

  - `updates` is read: scatter must read the update value.
  - `indices` is read: scatter must read indices to know where to write.
  - `mask`, if present, is read: a false mask suppresses the update and preserves
    the original value.
  - DPS init / original is read:
    - elements not hit by `indices` must be inherited from original;
    - masked-off updates must preserve original;
    - the combiner region may use the old value, e.g. `yield update + old`;
    - `unique_indices(false)` combine/reduction-style scatter also needs the
      old/current value.

  The only case where init may not need to be read is a separately proven
  full-overwrite optimization: indices cover the entire result, mask cannot
  suppress updates, and the combiner does not use the old value. That should be
  handled by a dedicated copy-elision/full-overwrite analysis, not by the default
  bufferization semantics.

  ## Tests

  Added LLVMGPU bufferization tests for:

  - overwrite scatter preserving original;
  - combiner scatter reading original;
  - masked scatter preserving original.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
+++ b/compiler/src/iree/compiler/Codegen/Interfaces/BufferizationInterfaces.cpp
@@ -375,13 +375,6 @@ struct LinalgExtOpInterface
     : bufferization::DstBufferizableOpInterfaceExternalModel<
           LinalgExtOpInterface<OpTy>, OpTy> {
 
-  bool bufferizesToMemoryRead(Operation *op, OpOperand &opOperand,
-                              const AnalysisState &state) const {
-    // TODO: Revisit this for ScatterOp. We can then get rid of
-    //       `bufferizesToMemoryRead` completely.
-    return !isa<IREE::LinalgExt::ScatterOp>(op);
-  }
-
   LogicalResult bufferize(Operation *op, RewriterBase &rewriter,
                           const BufferizationOptions &options,
                           bufferization::BufferizationState &state) const {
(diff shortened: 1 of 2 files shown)


krzysz00:
I think it might be a good idea to create that followup copy elision pass as I suspect this might have performance implications.

moomoohorse321:
My understanding is that the correctness semantics should conservatively treat the scatter DPS init as read. 

The thing is that even as a place holder, it shouldn't impact the correctness or it should raise a hard error. Putting a heisenbug as a place holder for an optimization is a really weird move here.

Do you mean we should have a follow-up optimization, i.e. identifying cases where the init copy can be safely elided, for example a provable full overwrite / no-mask-suppression / no-old-value-combiner case?  @krzysz00 

I'm suggesting a separate PR for the optimization and change the override here.

krzysz00:
@moomoohorse321 Yeah, I'd suggesting a copy elision pass as a followup

moomoohorse321:
@krzysz00 I did a follow-up PR #24531

AGindinson (approved):
LGTM, in my opinion it would be fine to merge the optimization as a follow-up

PR #24520, opened 2026-05-21
[Codegen][Tuner] Materialize constraint assignments in compiler

RattataKing:
Previously IREE constructs `compilation_info` only with attributes solved from the constraint op. However, for the upstream constraint op, it only adds tunable knobs, such as `workgroup=[...]` to the constraint and the knob dict, but doesn't cover other IREE's kernel config attr, such as `promote_operands = [...]`.
As the downstream tuner directly invokes IREE API to get `compilation_info` for TD spec building, and because the `compilation_info` type `MlirAttribute` is immutable once created from `DictAttr`, this PR changes IREE to build `compilation_info` with existing default config values + knob assignment values solved from the constraint op.

For example:
- default_config = `#iree_gpu.lowering_config<{promote_operands = [0, 1], workgroup = [64, 64, 0]}>`
- constraints = `workgroup = [wg_m, wg_n, 0]`
- knob_assignments = `wg_m=32, wg_n=32`
- compilation_info = `#iree_gpu.lowering_config<{promote_operands = [0, 1], workgroup = [32, 32, 0]}>`

Assisted-by: [Claude Code](https://claude.ai/code)
Issue: #23535

### DIFF
--- a/compiler/bindings/python/test/api/tuner_api_test.py
+++ b/compiler/bindings/python/test/api/tuner_api_test.py
@@ -862,3 +862,78 @@ def test_materialize_compilation_info_partial_subgroup_overlay():
         "#iree_gpu.lowering_config<{subgroup = [8, 16, 0], "
         "workgroup = [64, 128, 0]}>"
     )
+    translation_info = iree_codegen.TranslationInfoAttr(
+        compilation_info.translation_info
+    )
+    assert list(translation_info.workgroup_size) == [64, 1, 1]
+    assert translation_info.subgroup_size == 32
+
+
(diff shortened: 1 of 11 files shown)


kuhar (changes requested):
I think we should be able to accomplish this without having to extend the existing interfaces. Could we add CAPI/python bindings to add the attributes you need to existing  compilation_info / lowering_config / decomposition_info attributes that the python side would call?

RattataKing:
The implementation in this PR is replaced by:
https://github.com/iree-org/iree/pull/24528, materialize `compilation_info` from the knob.
https://github.com/iree-org/iree/pull/24538, hardcoded non-tunable configs in the knob dict.

PR #24521, opened 2026-05-21
[Runtime] Print allocator statistics for all devices

devtbi:
Extend run-module and benchmark with allocator statistics printing for all devices in the context.

This requires passing `iree_hal_device_list_t` through the runtime API for the devices to be available at the time of statistics printing.

### DIFF
--- a/runtime/src/iree/tooling/context_util.c
+++ b/runtime/src/iree/tooling/context_util.c
@@ -349,19 +349,20 @@ static iree_status_t iree_tooling_close_hal_replay_recorder(
 static iree_status_t iree_tooling_load_hal_async_module(
     iree_vm_instance_t* instance, iree_string_view_t default_device_uri,
     iree_allocator_t host_allocator, iree_vm_module_t** out_module,
-    iree_hal_device_t** out_device, iree_hal_allocator_t** out_device_allocator,
+    iree_hal_device_list_t** out_device_list,
+    iree_hal_allocator_t** out_device_allocator,
     iree_hal_replay_recorder_t** out_replay_recorder) {
   IREE_ASSERT_ARGUMENT(instance);
   IREE_ASSERT_ARGUMENT(out_module);
-  IREE_ASSERT_ARGUMENT(out_device);
+  IREE_ASSERT_ARGUMENT(out_device_list);
   IREE_ASSERT_ARGUMENT(out_device_allocator);
-  if (*out_device || *out_device_allocator) {
+  if (*out_device_list || *out_device_allocator) {
     return iree_make_status(
         IREE_STATUS_FAILED_PRECONDITION,
         "async HAL module can not be used with other primary HAL module types");
   }
   *out_module = NULL;
-  *out_device = NULL;
+  *out_device_list = NULL;
   *out_device_allocator = NULL;
   if (out_replay_recorder) *out_replay_recorder = NULL;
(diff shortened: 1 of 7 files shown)


runtime/src/iree/tooling/context_util.c
```diff
@@ -692,9 +689,7 @@ typedef struct {
   iree_tooling_module_list_t* resolved_list;
   // Default device URI used when --device= was not specified.
   iree_string_view_t default_device_uri;
-  // Lead HAL device retained from the resolved HAL module, if any.
-  iree_hal_device_t* device;
-  // Lead HAL allocator retained from the resolved HAL module, if any.
```
schuermans-roofline, line 697: Isn't this comment describing line 693 `device_allocator`? If so, shouldn't it be kept?

runtime/src/iree/tooling/run_module.c
```diff
@@ -473,13 +478,20 @@ iree_status_t iree_tooling_run_module_with_data(
...
-  if (device_allocator && FLAG_print_statistics) {
-    status = iree_status_join(
-        status, iree_hal_allocator_statistics_fprint(stderr, device_allocator));
+  if (FLAG_print_statistics && device_list) {
+    for (iree_host_size_t i = 0; i < device_list->count; ++i) {
+      iree_hal_device_t* dev = iree_hal_device_list_at(device_list, i);
+      iree_string_view_t device_id = iree_hal_device_id(dev);
+      fprintf(stderr, "Device: %.*s\n", (int)device_id.size, device_id.data);
+      iree_hal_allocator_t* allocator = iree_hal_device_allocator(dev);
+      IREE_IGNORE_ERROR(
```
schuermans-roofline: Why is the error ignored now? It was not ignored before. I understand that it is desirable to try to output the statistics of all allocators even in case of an error. However, wouldn't it make sense to store the first error encountered and then return that one after printing statistics of all allocators?

tools/iree-benchmark-module-main.cc
```diff
@@ -531,16 +531,24 @@ class IREEBenchmark {
...
-    if (device_allocator_ && FLAG_print_statistics) {
-      status = iree_status_join(status, iree_hal_allocator_statistics_fprint(
-                                            stderr, device_allocator_.get()));
+    if (FLAG_print_statistics && device_list_) {
+      for (iree_host_size_t i = 0; i < device_list_->count; ++i) {
+        iree_hal_device_t* dev = iree_hal_device_list_at(device_list_, i);
+        iree_string_view_t device_id = iree_hal_device_id(dev);
+        fprintf(stderr, "Device: %.*s\n", (int)device_id.size, device_id.data);
+        iree_hal_allocator_t* allocator = iree_hal_device_allocator(dev);
+        IREE_IGNORE_ERROR(
```
schuermans-roofline: Same comment about ignoring the error as above...

schuermans-roofline (approved):
Looks good to me. I just a few minor comments.

runtime/src/iree/tooling/context_util.c
devtbi: re-added 👍

runtime/src/iree/tooling/run_module.c
devtbi: `iree_status_join`-ing

PR #24522, opened 2026-05-21
[LLVMCPU] fix: return valid vscale range only for targets supporting SVE

ziereis:
fixes the `getDefaultVscaleRange` to only return a vscale range if the target supports it.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Utils/Utils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/Utils.cpp
@@ -12,6 +12,7 @@
 #include "iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUDialect.h"
 #include "iree/compiler/Codegen/Interfaces/ProcessorOpInterfaces.h"
 #include "iree/compiler/Codegen/Interfaces/UKernelOpInterface.h"
+#include "iree/compiler/Codegen/Utils/CPUUtils.h"
 #include "iree/compiler/Dialect/HAL/IR/HALOps.h"
 #include "iree/compiler/Dialect/HAL/IR/HALTypes.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtDialect.h"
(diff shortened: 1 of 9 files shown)


bjacob (approved):

compiler/src/iree/compiler/Codegen/Utils/Utils.cpp
```diff
@@ -1748,11 +1753,14 @@ bool hasFusedLeadingOp(linalg::LinalgOp rootOp) {
 
 std::optional<vector::VscaleRange>
 getDefaultVscaleRange(IREE::HAL::ExecutableTargetAttr targetAttr) {
```
egebeysel: Should we maybe move this function to [CPUUtils](https://github.com/iree-org/iree/blob/332d560d27b252dba0facff0f33a740c2d7e383e/compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp) instead? I think vscale is only relevant to CPUs and this could be kept there, as well as the `hasAnySVEFeature` method.

compiler/src/iree/compiler/Codegen/Utils/Utils.cpp
```diff
@@ -300,6 +300,11 @@ bool isRISCV64(DictionaryAttr targetConfig) {
   return triple && triple.value().isRISCV64();
 }
 
+static bool hasAnySVEFeature(DictionaryAttr targetConfig) {
```
egebeysel: I think there is a method for this already, maybe we could converge to using this one:
https://github.com/iree-org/iree/blob/332d560d27b252dba0facff0f33a740c2d7e383e/compiler/src/iree/compiler/Codegen/LLVMCPU/Utils.cpp#L92

egebeysel (commented):
The main functionality LGTM, but we could maybe refactor this a bit and move some things around 🤔

AGindinson (approved):
LGTM, I'm going to merge this since the comments have been addressed

PR #24524, opened 2026-05-21
[Runtime][Python] Convert numpy.uint8 to ABI type i8

Manewing:
Benchmarks are currently failing because of "KeyError: dtype(uint8)" these are fixed by simply converting numpy.uint8 to i8.

Also added other missing data types for numpy, i.e. `uint16` and `uint32` and added a test case to ensure the benchmark argument creation works for all cases


AGindinson (approved):
Does it make sense to add `uint16`/`uint32`, just in case?

AGindinson (approved):
LGTM, appreciate the updates!

PR #24526, opened 2026-05-21
[Codegen][Tuner] Add TileAndFuse constraints for conv

RattataKing:
TF constraints for conv will emit one smt.constraints op for each IGEMM conv and Direct conv.

Each separated constraint set is tested to have the same amount of smt solutions as the old tuner constraints.
See the SMT-LIB string [comparison](https://github.com/nod-ai/amd-shark-ai/commit/5d517500239f564f0bfa782c689b704a44d9f8b9).

Issue: #23535


kuhar (approved):
Ideally, the verifier shouldn't know about gpu attrs, but that can be refactored in the future

PR #24528, opened 2026-05-21
[Codegen] Add VectorDistribute constraints for attention

bangtianliu:
This PR ports the constraint generation phase for the attention ops from the Python tuner to the compiler-side SMT-emission. Refer to https://github.com/nod-ai/amd-shark-ai/commit/f21722acde1773e92ef3a1f2a9b307faca356da4, demonstrating that the compiler-bridge attention path produces the same candidate set. 



compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConstraintGenerator.cpp
```diff
@@ -1044,13 +1182,1137 @@ static Operation *getTunableOp(ArrayRef<Operation *> rootOps) {
...
+/// Build the attention knob dictionary. Mirrors the top-level attention
+/// lowering_config (`workgroup`, `partial_reduction`, `promote_operands`)
+/// plus the nested `decomposition_config` carrying per-matmul
+/// `lowering_config`s for QK and PV.
+static DictionaryAttr
+buildVectorDistributeAttentionKnobsDict(MLIRContext *ctx,
+                                        const AttentionDimInfo &dims,
+                                        ArrayRef<Attribute> compatibleQKMMAs,
+                                        ArrayRef<Attribute> compatiblePVMMAs) {
+  SmallVector<NamedAttribute> knobsEntries;
```
kuhar, line 1604: This flat knob dictionary is carrying four different concepts at one level: user lowering_config fields, translation_info fields, the attention side-channel decomposition_config, and SMT-only auxiliaries. That is the root cause of the aux leakage and the need for a decomposition-specific materializer. GPU PipelineAttr::materializeCompilationInfo already supports named lowering_config and translation_info sub-dictionaries; please use that shape, and keep decomposition_config as a sibling named output, instead of making every top-level knob entry implicitly part of compilation_info. Then a materializer can request a named output root, and SMT-only knobs can be excluded by construction rather than filtered by ad hoc key lists.

compiler/bindings/c/iree/compiler/dialects/iree_codegen.h
```diff
@@ -118,6 +118,12 @@ ireeCodegenMaterializeCompilationInfoFromConstraintsOp(
     const MlirStringRef *assignmentNames, const int64_t *assignmentValues,
     MlirAttribute *diagnosticMessage);
 
+MLIR_CAPI_EXPORTED MlirAttribute
+ireeCodegenMaterializeDecompositionConfigFromConstraintsOp(
```
kuhar: I do not think this should become a decomposition_config-specific public C ABI. The implementation materializes the full knob tree and returns one hard-coded top-level key; the next side output would need another exported symbol, Python binding, tests, and duplicated diagnostic wrapper. Instead, I'd expose a generic materialize-constraints-attr entry point that takes a top-level key or path, or make the constraints op declare named materialized outputs, and keep materialize_compilation_info as the special pipeline-repackaging helper. Attention can then request decomposition_config without baking this one side channel into the public API surface.

compiler/src/iree/compiler/Codegen/Common/MaterializeSMTConstraintAssignments.cpp
```diff
@@ -95,6 +97,22 @@ materializeKnobAttribute(ConstraintsOp op, Attribute attr,
         }
         return DictionaryAttr::get(ctx, materialized);
       })
+      .Case([&](IREE::GPU::LoweringConfigAttr lcAttr) -> Attribute {
```
kuhar: This GPU-specific case in Codegen/Common is a layering violation / shortcut. The operation here is substitute IntKnobAttr and OneOfKnobAttr leaves anywhere in an Attribute tree; it is not specific to GPU lowering_config. MLIR already provides AttrTypeReplacer for dialect-neutral sub-element replacement, which would handle #iree_gpu.lowering_config and the next typed wrapper without adding another TypeSwitch case and IREEGPUAttrs.h include to common code. If generic replacement cannot handle a wrapper, that wrapper probably needs an interface hook rather than a common-layer dependency on its concrete dialect type.

compiler/src/iree/compiler/Codegen/Common/VerifyPipelineConstraints.cpp
```diff
@@ -82,15 +106,70 @@ static LogicalResult extractKnobValue(Attribute templateAttr,
         }
         return success();
       })
-      .Default(failure());
+      .Case([&](IREE::GPU::LoweringConfigAttr lcTemplate) -> LogicalResult {
```
kuhar: Same layering issue on the verifier side: common SMT verification now knows about IREE::GPU::LoweringConfigAttr, while the code below already has a TODO saying GPU config extraction should move behind LoweringConfigAttrInterface. This case makes the common verifier responsible for every backend typed wrapper that can contain knobs.

kuhar (changes requested):
In the current form, this takes too many layering shortcuts and adds gpu-specific construct on a generic path. We also need to de-sloppify many comments and make them more concise, not refer to tuner code that will get deleted soon, or explain what the agent was doing at the time. Also a bunch of non-ascii characters to remove...

kuhar (changes requested):
I took a deeper look and realized that there's even a bigger design issue in this implementation.  The "aux variable" approach is the wrong abstraction here. It adds solver-only knobs to the public constraint template even though they are not part of the configuration being materialized. That breaks the core contract of knobs: a knob is supposed to correspond to a materializable configuration choice, not an internal temporary used to write an equation.

Once those internal temporaries are exposed as knobs, every downstream layer has to learn an exception: materialization must ignore them, verification must tolerate them missing from real lowering configs, tests must carry fake assignments, and diagnostics start talking about names that users and codegen cannot actually configure. That is architectural leakage, not modeling.

It also weakens verification. If verifier support has to whitelist "declared but intentionally absent" entries, then missing real configuration entries and missing solver-only entries become harder to distinguish. The verifier should be able to evaluate a materialized config against the template directly. Aux knobs force special cases that make silent under-verification much more likely.

The cleaner model is to keep derived quantities as SSA values inside the constraint body. Layout fields can be lookup results from the selected MMA. Boolean predicates can be composed directly from layout comparisons. Schedule factors can be quotient expressions guarded by exact-divisibility assertions. None of these need names in the materialization API, because none of them are user/codegen configuration choices.

Solver temporaries are fine as internal SSA. Promoting them to C/API-visible knob/template structure is a major design error. It turns local algebra into a cross-layer contract and makes every consumer carry compensating complexity.

kuhar:
> Refer to https://github.com/nod-ai/amd-shark-ai/commit/f21722acde1773e92ef3a1f2a9b307faca356da4, demonstrating that the compiler-bridge attention path produces the same candidate set.

I don’t think this code matches the claim because it does not define a comparison that shows we produce the same candidate set. This code is stronger than a bare count check as a report, but not strong enough as a verifier because the harness does not define and enforce a reliable candidate equivalence relation. Without that, matching counts and diagnostic summaries can still miss swapped candidates, duplicate/multiplicity changes, or differences that show up in other dtype/layout cases. We don't necessarily need all of that as code, we could roughly match candidates and eyeball for equivalence, but that's an important caveat to write down somewhere...

bangtianliu:
@kuhar Thanks for pushing the commits to this PR! I was actually about to address your review comments later today.

kuhar (approved):
The attention constraints themselves were fine for the most part, but the interfaces / layering / architecture seemed like, and I'm sorry to put it bluntly, a vibe-coded prototype at best, so I ended up rewriting it instead of going through N rounds or back-and-forth review rounds on github.

Here's the summary of the changes:
1. Replaced the attention-specific `materialize_decomposition_config` path with generic `materializeConfigurationAttr(name, ...)`. `compilation_info` remains the special pipeline-owned repackaging case; attrs like `decomposition_config` are materialized by name from the knob dictionary. Motivation: avoid growing one C API / type switch case per output attr.
2. Removed the SMT-only "aux knob" model. Derived things like subgroup tile counts, layout matches, and QK/PV reuse are now SSA expressions derived from real knobs, not fake entries in the knob dictionary. Motivation: the knob dictionary should describe materializable configuration, not solver temporaries.
3. Made the emitted attention config match what codegen actually consumes: `reduction` instead of `partial_reduction` (just like what Python emits), typed per-matmul `LoweringConfigAttr`s in `decomposition_config`, explicit `promotion_types`, and no unused top-level subgroup-count attrs.
4. Tightened the constraints to mirror the Python tuner bug-for-bug: MMA layout matching, schedule validity, shared-memory reuse, PV dtype modeling, and the hardcoded Python defaults are preserved. Potential improvements are left as TODOs because this PR's goal is parity, not fixing the Python constraints.

I validated with a local e2e parity harness that runs both paths on representative attention modules, then compares parsed semantic candidate multisets:
```
(qk_mma, pv_mma, workgroup tile, reduction tile)
```

Results:
- f16 small standard: `110 == 110`, exact semantic multiset match
- f16 tiny standard: `20 == 20`, exact semantic match
- f16 tiny transposed: `20 == 20`, exact semantic match
- bf16 tiny standard: `6 == 6`, exact semantic match
- f8 tiny standard: Python `4`, compiler `16`, with no Python-only candidates

Full transform-spec text still differs. Python emits some top-level/ignored attrs; compiler emits more complete codegen attrs like `promotion_types`. I intentionally did not add ignored compiler attrs just for textual parity.

The only candidate-count divergence is f8. The compiler distinguishes physical vs virtual f8 MMA choices that the old Python path collapses. If that distinction is normalized away, the unique semantic set is the same: `4 == 4`. Since there are no Python-only candidates, this is not evidence that the compiler is missing anything from the Python tuner.

`col_major` was also checked as derivable from the selected MMA layouts, not an independent solver choice, so I did not model it as a separate SMT knob.

PR #24529, opened 2026-05-22
[CI][Windows] Fix failing windows build after runner update

maxbartel:
Looking at the log it seems like they changed the OSTYPE from `msys` to `cygwin` and this cmake flag therefore didn't work.


kuhar (dismissed):

kuhar:
Oh, seems like it's still failing. BTW, I don't believe building with cygwin/mingw was ever supported.

AGindinson (approved):
I think the build has progressed to the CTest phase with just one failure though. Should we address this separately?

maxbartel:
Apparently it was the same issue but somewhere else. Let's see if it is passing now. 

I will try to first make it pass and add looking deeper into the windows build to my todos then.

kuhar (approved):

PR #24531, opened 2026-05-24
[Codegen] Create a copy elision pass for DPS inits

moomoohorse321:
Follow up on #24504

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/FoldMemRefCopyIntoDPSOps.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/FoldMemRefCopyIntoDPSOps.cpp
@@ -7,15 +7,22 @@
 #include "iree/compiler/Codegen/Common/Passes.h"
 #include "iree/compiler/Codegen/Utils/Utils.h"
 #include "iree/compiler/Dialect/HAL/IR/HALOps.h"
+#include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.h"
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/STLExtras.h"
 #include "mlir/Analysis/AliasAnalysis.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/MemRef/Utils/MemRefUtils.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/SymbolTable.h"
 #include "mlir/Interfaces/DestinationStyleOpInterface.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#include <limits>
 
 namespace mlir::iree_compiler {
 
@@ -25,9 +32,9 @@ namespace mlir::iree_compiler {
 namespace {
 
 // Peel view-like memref producers to the allocation, global, or interface
-// binding that defines the addressed storage. This deliberately relies on MLIR's
-// ViewLikeOpInterface instead of assuming that every one-memref-operand op is an
-// aliasing view.
+// binding that defines the addressed storage. This deliberately relies on
+// MLIR's ViewLikeOpInterface instead of assuming that every one-memref-operand
+// op is an aliasing view.
 static Value getMemRefViewRoot(Value value) {
   auto memrefValue = dyn_cast<MemrefValue>(value);
   if (!memrefValue) {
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/Codegen/Common/FoldMemRefCopyIntoDPSOps.cpp
```diff
@@ -0,0 +1,425 @@
...
+static std::optional<IREE::HAL::InterfaceBindingSubspanOp>
+getSourceSubspan(Value value) {
+  auto typedValue = dyn_cast<TypedValue<MemRefType>>(value);
+  if (!typedValue) {
+    return std::nullopt;
+  }
+  return getSourceSubspanMemref(typedValue);
+}
+
+static bool mayAlias(Value lhs, Value rhs) {
```
krzysz00: Do we not have existing code for this?

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
```diff
@@ -912,6 +912,10 @@ void addGPUBaseLoweringPassPipeline(OpPassManager &funcPassManager) {
   funcPassManager.addPass(createCanonicalizerPass());
   funcPassManager.addPass(createCSEPass());
 
+  funcPassManager.addPass(createFoldMemRefCopyIntoDPSOpsPass());
+  funcPassManager.addPass(createCanonicalizerPass());
```
krzysz00: Does this need the canonicalize?

krzysz00 (commented):
Initial notes

compiler/src/iree/compiler/Codegen/Common/FoldMemRefCopyIntoDPSOps.cpp
moomoohorse321: Yes, updated in c231fa2856. The pass now uses existing MLIR/IREE utilities for the generic parts: MLIR AliasAnalysis for alias/mod-ref queries, memref::skipViewLikeOps for view roots, and getSourceSubspanMemref for HAL subspan sources.

I kept only the IREE-specific facts local to this pass, such as different hal.interface.binding.subspan bindings do not alias and same binding remains conservative because offsets may be dynamic

compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp
moomoohorse321: I removed the immediate canonicalizer and CSE after this pass in c231fa2856.

compiler/src/iree/compiler/Codegen/Common/FoldMemRefCopyIntoDPSOps.cpp
moomoohorse321: If by reuse, you mean reusing IREE-level alias checker for memref buffers, I checked for that but I did not find one (using codex). If there are some more experienced folks please help us out here :)


The closest existing utility codex found is getSourceSubspanMemref(), which the updated commit now uses to recover the source HAL interface binding from a derived memref. That helper does not compare two buffers. 

Codex also found Explorer::mayValuesAlias, but it operates on stream level and the Stream resource analyses operate at different layers semantics, so they are not directly applicable to post-bufferization memrefs here.

This motivates us to use MLIR AliasAnalysis for the generic memref alias/mod-ref cases, and keeps only the IREE-specific refinements locally: HAL binding identity and memref global symbol identity.

smart-moomoo:
Next step: Split the PR into 2 parts: a basic pass and a augmentation

smart-moomoo:
I found this one is too big to review, so I did 2 things:

- have the following write-up for reviewers to better understand the context of the pass.
- split the big PR into 3 PRs building on top of each other. 

### Some recap
  Bufferization may represent a destination-style update using a temporary buffer:
```
  %temp = memref.alloc()
  memref.copy %source, %temp // to remove if eligible

  linalg.generic ... outs(%temp) { // can be combined with the next copy
    // May read and update the old output element.
  }
  memref.copy %temp, %target
```
  The base transformation forwards the final target into the DPS operation:
```
  memref.copy %source, %target

  linalg.generic ... outs(%target) {
    // May read and update the old output element.
  }
```
  This removes %temp and the copy-out. The copy-in remains because the DPS operation may preserve or read elements from its init operand.

  If the operation instead completely overwrites the destination without reading its old contents:
```
  %temp = memref.alloc()
  memref.copy %source, %temp
  linalg.fill ins(%value) outs(%temp)
  memref.copy %temp, %target
```
  then the initialization is dead, and the final result becomes:
```
  linalg.fill ins(%value) outs(%target)
```
  ### Pass implementation

  The pass runs after bufferization, when the allocation and copies are explicit, but before Linalg/LinalgExt lowering removes the destination-style semantics needed by the
  rewrite.

  The forwarding pattern requires:

  1. A private memref.alloc.
  2. Exactly one copy into the allocation.
  3. Exactly one destination-style operation using it as its init.
  4. Exactly one copy from the allocation to the final target.
  5. All three operations in the expected order within one block.

  Retargeting changes when and where the final output is written, so the pass additionally checks:

  - The copy source does not alias the final target.
  - Other DPS operands do not alias the final target.
  - No intervening operation may read or write the final target.

  MLIR alias and mod/ref analysis handles ordinary local cases. The pass additionally classifies storage roots that generic local analysis does not fully distinguish:

  - Different HAL interface binding ordinals are treated as disjoint.
  - Subspans of the same binding remain conservative because byte offsets may be dynamic.
  - Different memref.global symbols are disjoint.
  - Local allocations are disjoint from HAL bindings and globals.
  - Unknown roots fall back to MLIR alias analysis.

  ### splitted PRs
  1. DPS temporary forwarding https://github.com/iree-org/iree/pull/24727

     Introduces the pass, pipeline placement, alias/mod-ref safety checks, and the base copy-in/DPS/copy-out forwarding rewrite. It deliberately preserves the initialization copy.

  2. Linalg initialization elision (will open after the first one land)

     Extends PR 1 with the relatively general and locally understandable overwrite proofs for linalg.fill, linalg.copy, and restricted linalg.generic.

3. Scatter overwrite (same as 2nd one, will open)
     Extends PR 2 with the more specialized Scatter coverage proof. This is separate because constant-index decoding and complete-coverage validation have substantially different review concerns from ordinary Linalg semantics.

4. Future optimizations
to be implemented

Issue #24515, opened 2026-05-20
Tracking: `inner_tiled` CPU performance parity with `mmt4d` + ukernels


bjacob:
## Goal

Bring the inner-tiled path *without* ukernels to performance parity with the mmt4d path *with* ukernels.

So this is more ambitious than merely parity between inner_tiled and mmt4d.  This is about making inner_tiled codegen so good that switching to inner_tiled removes much of the need for ukernels.

## PRs

- [X] [[Codegen][CPU] Route inner_tiled broadcast into m_bcst-foldable slot.](https://github.com/iree-org/iree/pull/24516#top)
- [X] [[Codegen][CPU] Flatten contiguous trailing dims of transfers before unrolling.](https://github.com/iree-org/iree/pull/24517#top)
- ~~[ ] [[Codegen] Cap hoisted statically-bound allocations at the stack budget.](https://github.com/iree-org/iree/pull/24532#top)~~
- [X] [[Codegen][CPU] Fold reshape-containing encoding relayouts to map_store.](https://github.com/iree-org/iree/pull/24533#top)
- [X] [[Codegen][CPU] Add x86 AVX-512 VNNI 16x16x2 i8 MMA intrinsic.](https://github.com/iree-org/iree/pull/24534#top)
- [ ] [[Codegen][CPU] Fold reshapes into map_store per reassociation group.](https://github.com/iree-org/iree/pull/24535#top)

## Benchmark

Square matmul `4096x4096x4096`, dynamic shapes, AMD Zen 4
(Ryzen 9 7950X3D), `iree-benchmark-module --device=local-task`:

| element types | `mmt4d` + ukernel | `inner_tiled` (before PRs) | `inner_tiled` (after PRs) | `inner_tiled` vs ukernel |
|---|---|---|---|---|
| f32 × f32 → f32 | 115 ms | ~102 ms | 101 ms | **1.14× faster** |
| f16 × f16 → f32 | 132 ms | ~126 ms | 124 ms | **1.06× faster** |
| bf16 × bf16 → f32 | 51.1 ms | ~52 ms | 51.9 ms | parity (~2%) |
| i16 × i16 → i32 | 52.0 ms | ~53 ms | 52.6 ms | parity (~1%) |
| i8 × i8 → i32 | 43.6 ms | 72.5 ms | 46.9 ms | inner_tiled 1.07× slower due to relayout dispatch ; kernel itself is just as fast. See https://github.com/iree-org/iree/issues/24514 |

[...]

PR #24532, opened 2026-05-25
[Codegen] Cap hoisted statically-bound allocations at the stack budget.

bjacob:
`hoistOneStaticallyBoundAllocation` pads a dynamic allocation to its computed upper bound. An unconstrained dynamic dimension carries only the `index` type maximum as its bound, so padding yields a bogus multi- petabyte allocation (iree-org/iree#24483). Decline to hoist past the target's `max_stack_allocation_size`; declining to hoist is always valid -- it simply leaves the IR's existing dynamic allocation in place.

Extent of the #24483 fix: this is partial. It implements the first of the three fixes proposed in that issue -- the hoist itself no longer turns an `index`-umax-derived bound into a bogus static `alloca` (which fed poison strides into codegen and silently dropped data). It does not on its own make the originally-affected pipeline -- dynamic-shape `inner_tiled` data-tiling -- compile: as #24483 notes, capping leaves a dynamic CPU stack allocation, which is itself rejected downstream. What this commit guarantees is only that the failure mode is no longer a *silent miscompile*. The dynamic-shape `inner_tiled` pipeline is fully unblocked by a later commit in this series ("Fold reshape-containing encoding relayouts to map_store"), which removes the whole-tensor `pack` so the unbounded allocation never arises. #24483 is therefore not fully resolved by this commit alone.

Progress towards #24515.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/BUILD.bazel
+++ b/compiler/src/iree/compiler/Codegen/Common/BUILD.bazel
@@ -234,6 +234,7 @@ iree_compiler_cc_library(
         "//compiler/src/iree/compiler/Dialect/Util/Transforms",
         "//compiler/src/iree/compiler/Utils",
         "//llvm-external-projects/iree-dialects:IREELinalgTransformDialect",
+        "@llvm-project//llvm:Core",
         "@llvm-project//llvm:Support",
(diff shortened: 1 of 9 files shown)


bjacob:
Made unnecessary by https://github.com/iree-org/iree/pull/24535.

PR #24533, opened 2026-05-25
[Codegen][CPU] Fold reshape-containing encoding relayouts to map_store.

bjacob:
GPU's configuration pipeline folds the `pack`/`expand_shape`/`transpose` relayout chain from encoding materialization into a single `iree_linalg_ext.map_store` before tiling; CPU did not. For a non-row- major encoding swizzle the intervening `tensor.expand_shape` (not a `TilingInterface` op) blocks producer fusion and strands an untiled, whole-tensor `pack` intermediate.

Mirror GPU on CPU:
- Add `RelayoutCombinationScope::DispatchReshape`: like `Dispatch` but restricted to chains whose backward slice contains an expand/collapse_shape. Pure pack/unpack/transpose chains tile fine and are left alone.
- Run `CombineResultLayoutTransformation` with that scope in the CPU configuration pipeline, after MaterializeDeviceEncoding.
- Skip `map_load`/`map_store` as `getRootOperation` candidates, so a dispatch with a real compute op plus a `map_store` roots on the compute op.

Inert for existing (row-major) CPU encodings.

Progress towards #24515.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
@@ -652,7 +652,25 @@ void buildLLVMCPUCodegenConfigurationPassPipelineImpl(
       .addPass(createConvertAccGEMMToGEMMPass)
       // TODO: Remove the following pass the plumb support for
       // #hal.descriptor_type memory space through the stack.
-      .addPass(createEraseHALDescriptorTypeFromMemRefPass);
+      .addPass(createEraseHALDescriptorTypeFromMemRefPass)
+      // Fold reshape-containing relayout chains (`pack` -> `expand_shape` ->
+      // `transpose`) emitted by encoding materialization for non-row-major
+      // swizzles into a single `iree_linalg_ext.map_store`, before the
+      // dispatch is tiled. This mirrors the GPU configuration pipeline.
+      // Without it, the intervening `tensor.expand_shape` (not a
+      // `TilingInterface` op) blocks producer fusion and leaves an untiled,
+      // whole-tensor `pack` intermediate whose dynamic `tensor.empty`
+      // bufferizes to a bogus unbounded allocation (iree-org/iree#24483);
+      // `map_store` is a scatter with no intermediate buffer. The
+      // `DispatchReshape` scope restricts this to reshape-containing chains so
+      // plain `pack` encodings (which tile fine) are left untouched.
+      .addPass(createBufferizeDispatchTensorLoadStorePass)
+      .addPass([] {
+        CombineResultLayoutTransformationPassOptions options;
+        options.scope =
+            IREE::Codegen::RelayoutCombinationScope::DispatchReshape;
+        return createCombineResultLayoutTransformationPass(options);
+      });
   modulePassManager.addPass(createLLVMCPUSelectLoweringStrategyPass());
   LLVM_DEBUG({
     llvm::dbgs() << "LLVMCPU codegen configuration pass pipeline:\n";
--- a/compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp
@@ -6,6 +6,7 @@
 
 #include "iree/compiler/Codegen/Utils/CPUUtils.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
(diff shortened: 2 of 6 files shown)


compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp
```diff
@@ -56,7 +57,8 @@ FailureOr<Operation *> getRootOperation(ArrayRef<Operation *> computeOps) {
     }
 
     if (isa<TilingInterface>(op) &&
-        !isa<tensor::PadOp, linalg::PackOp, linalg::UnPackOp>(op)) {
+        !isa<tensor::PadOp, linalg::PackOp, linalg::UnPackOp,
+             IREE::LinalgExt::MapLoadOp, IREE::LinalgExt::MapStoreOp>(op)) {
```
egebeysel, line 61: just to clarify: (correct me if I'm wrong) but the `map_load` op is not generated through `CombineResultLayoutTransformationPass`, right? So is it really necessary to guard against this?

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
```diff
@@ -652,7 +652,25 @@ void buildLLVMCPUCodegenConfigurationPassPipelineImpl(
...
+      // bufferizes to a bogus unbounded allocation (iree-org/iree#24483);
+      // `map_store` is a scatter with no intermediate buffer. The
+      // `DispatchReshape` scope restricts this to reshape-containing chains so
+      // plain `pack` encodings (which tile fine) are left untouched.
+      .addPass(createBufferizeDispatchTensorLoadStorePass)
+      .addPass([] {
+        CombineResultLayoutTransformationPassOptions options;
+        options.scope =
+            IREE::Codegen::RelayoutCombinationScope::DispatchReshape;
+        return createCombineResultLayoutTransformationPass(options);
```
egebeysel, line 672: question: so this one over here basically attends to handle the operations that do not have tiling interfaces implemented, whereas the the one that happens after distribution tiling handles the operations that did not tile successfully (but were expected to)?

egebeysel (approved):
Thanks, LGTM in general, just 2 questions 😄

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
bjacob: I asked Claude to answer this too:

Close, but the split is by what's in the chain rather than by tiling outcome:

This pre-tile invocation uses DispatchReshape, which is the strictly narrower scope: it only fires when the chain's backward slice contains an expand_shape/collapse_shape. That reshape is the non-TilingInterface op that would block producer fusion and strand the dynamic whole-tensor pack (the #24483 bug) — so it has to be folded before tiling.
The existing post-tile invocation at line 92 uses the default Dispatch scope, which folds any pack/unpack/transpose chain. By that point distribution has already happened; it catches the plain (no-reshape) chains that tiled fine and combines them into map_store inside the workgroup as a downstream cleanup.
So no overlap: DispatchReshape (restrictive, pre-tile, unblocks tiling for the reshape case) vs Dispatch (default, post-tile, cleans up the rest). The DispatchReshape restriction is also why plain pack encodings (the row-major case that already tiles fine) are left untouched here.

compiler/src/iree/compiler/Codegen/Utils/CPUUtils.cpp
bjacob: I asked Claude to answer your question and got this:

Good catch — strictly, CombineResultLayoutTransformationPass only emits MapStoreOp; in this PR no pass in the LLVMCPU pipeline emits MapLoadOp. The MapLoadOp exclusion is for symmetry with the second list in getRootOperation (the "relayout-ops fallback"), which this PR also updates to include map_load/map_store together. The two lists have to stay in lockstep — anything excluded from the "real compute op" sweep above needs to be a valid fallback root below, otherwise a pure-MapLoadOp dispatch would have no root at all. GPU runs both CombineSourceLayoutTransformation (→ MapLoadOp) and CombineResultLayoutTransformation (→ MapStoreOp); the CPU heuristic just stays consistent across the same pair.

Happy to narrow both lists to just MapStoreOp if you'd prefer to keep the diff strictly to what this pass introduces — let me know.

compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp
egebeysel: I see, I checked the code as well and got my answer 👍

PR #24534, opened 2026-05-25
[Codegen][CPU] Add x86 AVX-512 VNNI 16x16x2 i8 MMA intrinsic.

bjacob:
`MMA_X86_AVX512VNNI_16x16x2_I32_I8_CASTI16` processes a full 16x16x2 i8 matmul tile, mirroring the mmt4d ukernel's hot loop. Unlike the 1x16x2 / 16x1x2 CASTI16 variants -- one row per intrinsic, with a per-row i8->i16 widen + broadcast -- it consumes whole 16x2 i8 panels: one `vpmovsxbw` widen per panel, 4 in-lane + 4 128-bit-block shuffles to fan the panels across a 4x4 grid, then 16 `vpdpwssd`. The 16x16 accumulator uses a block-interleaved `TileSwizzle` so each `vpdpwssd` accumulator is a contiguous register.

i8 inner_tiled matmul, 4096x4096 on Zen 4: 72.5ms -> 52.9ms.

Progress towards #24515.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -930,6 +930,29 @@ getCombineRelayoutOpsControlFn(IREE::Codegen::RelayoutCombinationScope scope) {
       return isa<IREE::Codegen::StoreToBufferOp>(*leaf.getUsers().begin());
     };
     break;
+  // Control function for DispatchReshape scope. Like Dispatch, but additionally
+  // requires the backward relayout slice to contain a `tensor.expand_shape` or
+  // `tensor.collapse_shape` — the non-tileable reshape whose presence is what
+  // makes folding into a `map_store` necessary (see the enum doc comment).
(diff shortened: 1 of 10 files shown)


hanhanW (approved):

egebeysel (approved):

egebeysel:
do we actually have an e2e test that runs this code path?

bjacob:
> do we actually have an e2e test that runs this code path?

Yes — the in-tree e2e_matmul_cpu_dt_i8_i32_llvm-cpu_local-task_avx512vnni and e2e_matmul_cpu_experimental_dt_i8_i32_llvm-cpu_local-task_avx512vnni tests cover it. Pre-PR, the cost model picks a smaller i8 intrinsic for the data-tiled VNNI matmul; this PR's kAllX86[] entry plus the lowering at lowerX86Avx512Vnni16x16x2I8 make MMA_X86_AVX512VNNI_16x16x2_I32_I8_CASTI16 the preferred choice (largest register footprint at avx512), so post-PR those tests go through the new lowering numerically. I confirmed via --mlir-print-ir-after-all on a representative shape on this branch that the cost model does select this intrinsic for a data-tiled i8 matmul.

PR #24535, opened 2026-05-25
[Codegen][CPU] Fold reshapes into map_store per reassociation group.

bjacob:
`foldReshapeIntoMapStore` composed an expand/collapse reshape into a `map_store`'s index transformation via one global `linearize` over the source shape plus `delinearize` over the result shape. Under dynamic shapes that `delinearize` lowers to a dynamic integer division per element.

Fold each reassociation group independently instead: an `expand_shape` splits one source index by just that group's result sizes; a `collapse_shape` merges one group of source indices. The split/merge factors are then per-group -- typically static tile sizes -- so the scalarized transformation is static shifts/masks with no per-element division.

Speeds up the i8 data-tiled matmul's C-encode dispatch; i8 inner_tiled matmul, 4096x4096 on Zen 4: 52.9ms -> 46.9ms.

Also document, at `MapStoreOpVectorizationModel::isVectorizable`, why a CPU-tiled `map_store` reaches the vectorizer dynamically shaped and scalarizes -- which is precisely what makes this per-group, division-free fold matter -- while the same op vectorizes on GPU. Comment only, no behavior change.

Progress towards #24515.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
@@ -77,8 +77,16 @@ static MapStoreOp foldTransposePermIntoMapStore(RewriterBase &rewriter,
 }
 
 /// Fold a tensor::ExpandShapeOp or tensor::CollapseShapeOp into a consumer
-/// `mapStoreOp`, by linearizing and then delinearizing the source indices
-/// of the `mapStoreOp`s index transformation.
+/// `mapStoreOp` by composing the reshape's index transformation into the
+/// front of the `mapStoreOp`'s.
+///
+/// Each reassociation group is folded independently: an `expand_shape` splits
+/// one source index via a `delinearize` over just that group's result sizes,
+/// and a `collapse_shape` merges a group of source indices via a `linearize`.
+/// This keeps the split/merge factors per-group — typically static tile sizes
+/// — rather than routing every index through one global `linearize` +
+/// `delinearize` over the full (dynamic) shapes, which would force a dynamic
+/// integer division per element when the transformation is later scalarized.
 template <typename ReshapeOpTy>
 static FailureOr<IREE::LinalgExt::MapStoreOp>
 foldReshapeIntoMapStore(RewriterBase &rewriter, ReshapeOpTy reshapeOp,
@@ -92,6 +100,7 @@ foldReshapeIntoMapStore(RewriterBase &rewriter, ReshapeOpTy reshapeOp,
       cast<RankedTensorType>(reshapeOp.getResult().getType()).getRank() == 0) {
     return failure();
   }
+  constexpr bool isExpand = std::is_same_v<ReshapeOpTy, tensor::ExpandShapeOp>;
   Location loc = reshapeOp->getLoc();
   OpBuilder::InsertionGuard g(rewriter);
   rewriter.setInsertionPointAfter(reshapeOp);
@@ -102,20 +111,53 @@ foldReshapeIntoMapStore(RewriterBase &rewriter, ReshapeOpTy reshapeOp,
   // sizes by later cleanup patterns.
   SmallVector<OpFoldResult> resultDims =
       tensor::getMixedSizes(rewriter, loc, reshapeOp.getResult());
+  SmallVector<ReassociationIndices> reassociation =
+      reshapeOp.getReassociationIndices();
 
   Location mapStoreLoc = mapStoreOp->getLoc();
   rewriter.modifyOpInPlace(mapStoreOp, [&]() {
     mapStoreOp.insertTransformationAtStart(
         rewriter,
-        [&rewriter, mapStoreLoc, srcDims,
-         resultDims](ArrayRef<BlockArgument> indices) -> SmallVector<Value> {
-          SmallVector<Value> indexValues(indices.begin(), indices.end());
-          auto linearizeIndexOp = affine::AffineLinearizeIndexOp::create(
-              rewriter, mapStoreLoc, indexValues, srcDims, /*disjoint=*/true);
-          auto delinearizeIndexOp = affine::AffineDelinearizeIndexOp::create(
-              rewriter, mapStoreLoc, linearizeIndexOp.getResult(), resultDims,
-              /*hasOuterBound=*/true);
-          return delinearizeIndexOp->getResults();
+        [&rewriter, mapStoreLoc, srcDims, resultDims, reassociation,
+         isExpand](ArrayRef<BlockArgument> indices) -> SmallVector<Value> {
+          // `indices` are in the reshape's source index space; produce the
+          // reshape's result index space.
+          SmallVector<Value> result(resultDims.size());
+          for (auto [groupIdx, group] : llvm::enumerate(reassociation)) {
+            if (isExpand) {
+              // Source index `groupIdx` splits into result dims `group`.
+              if (group.size() == 1) {
+                result[group[0]] = indices[groupIdx];
+                continue;
+              }
+              SmallVector<OpFoldResult> groupSizes;
+              for (int64_t resultDim : group) {
+                groupSizes.push_back(resultDims[resultDim]);
+              }
+              auto delinearizeOp = affine::AffineDelinearizeIndexOp::create(
+                  rewriter, mapStoreLoc, indices[groupIdx], groupSizes,
+                  /*hasOuterBound=*/true);
+              for (auto [i, resultDim] : llvm::enumerate(group)) {
+                result[resultDim] = delinearizeOp.getResult(i);
+              }
+            } else {
+              // Source dims `group` merge into result index `groupIdx`.
+              if (group.size() == 1) {
+                result[groupIdx] = indices[group[0]];
+                continue;
+              }
+              SmallVector<Value> groupIndices;
+              SmallVector<OpFoldResult> groupSizes;
+              for (int64_t srcDim : group) {
+                groupIndices.push_back(indices[srcDim]);
+                groupSizes.push_back(srcDims[srcDim]);
+              }
+              result[groupIdx] = affine::AffineLinearizeIndexOp::create(
+                  rewriter, mapStoreLoc, groupIndices, groupSizes,
+                  /*disjoint=*/true);
+            }
+          }
+          return result;
         },
         srcDims.size());
     mapStoreOp.getInputMutable().assign(reshapeOp->getOperand(0));
--- a/compiler/src/iree/compiler/Codegen/Common/test/combine_result_layout_transformation.mlir
+++ b/compiler/src/iree/compiler/Codegen/Common/test/combine_result_layout_transformation.mlir
@@ -4,6 +4,12 @@
 // RUN:   -split-input-file %s | FileCheck %s --check-prefixes=WORKGROUP-SCOPE
 // RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-codegen-combine-result-layout-transformation{scope=dispatch-reshape},canonicalize,cse))" \
 // RUN:   -split-input-file %s | FileCheck %s --check-prefixes=RESHAPE-SCOPE
+// Scalarize the index transformations to show that the per-group fold keeps
+// the lowered body division-free, while a global `delinearize_index` over a
+// dynamic shape would scalarize to runtime-divisor `arith.divsi`/`remsi`.
+// Gated on the EXPAND prefix.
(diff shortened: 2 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -77,8 +77,16 @@ static MapStoreOp foldTransposePermIntoMapStore(RewriterBase &rewriter,
 }
 
 /// Fold a tensor::ExpandShapeOp or tensor::CollapseShapeOp into a consumer
-/// `mapStoreOp`, by linearizing and then delinearizing the source indices
-/// of the `mapStoreOp`s index transformation.
+/// `mapStoreOp` by composing the reshape's index transformation into the
+/// front of the `mapStoreOp`'s.
+///
+/// Each reassociation group is folded independently: an `expand_shape` splits
+/// one source index via a `delinearize` over just that group's result sizes,
```
egebeysel, line 84: am I missing something or there is a test being introduced for the collapse (linearize) routing, but not the expand (delinearize) case? is that relevant to the below comment in `VectorizableOpInterface.cpp` that the encoding relayout should be combined into a `map_load`?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
```diff
@@ -77,8 +77,16 @@ static MapStoreOp foldTransposePermIntoMapStore(RewriterBase &rewriter,
...
+/// and a `collapse_shape` merges a group of source indices via a `linearize`.
+/// This keeps the split/merge factors per-group — typically static tile sizes
+/// — rather than routing every index through one global `linearize` +
+/// `delinearize` over the full (dynamic) shapes, which would force a dynamic
+/// integer division per element when the transformation is later scalarized.
```
egebeysel, line 89: maybe we could add a test to vectorization that illustrates the IR that would (and possibly what would not) be vectorized?

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
```diff
@@ -462,6 +462,18 @@ struct MapStoreOpVectorizationModel
...
+    // so the input must be statically shaped. A `map_store` iterates its
+    // *input*, so it only vectorizes when that input is already statically
+    // tiled. For a data-tiling encoding relayout the input is the un-padded
+    // source tensor, which has no static tile structure -- so a CPU-tiled
+    // `map_store` reaches here dynamically shaped and is left to scalarize.
+    // The padded, statically-tiled side of an encoding relayout is its
+    // *output*; a relayout rooted there -- a `map_load`, which iterates its
+    // result -- tiles into complete static tiles by construction. So a CPU
+    // encoding relayout should be combined into a `map_load`, not a
+    // `map_store`. (GPU also forms `map_store` for these but tiles them
```
egebeysel, line 475: What about the case where we would have `matmul -> *some_op(s)* -> unset_encoding -> set_encoding`? Wouldn't (or shouldn't) the `set_encoding` and the corresponding `expand_shape` be folded into the `map_store` there, or am I misunderstanding the context?

egebeysel (commented):
Generally looks good to me, but I might be missing some context here. Does this only change the behavior on the CPU and not GPU?

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
bjacob: (Claude here, replying on Benoit's behalf — these are my words, not his.)

Two parts:

The expand case wasn't actually uncovered — @fold_pack_op (line 281, unchanged on this PR) already CHECKs the per-group affine.delinearize_index form for tensor<250x250xf32> → tensor<2x2x128x128xf32>. The pre-PR code's global linearize + delinearize happened to canonicalize down to that same per-group shape for static dimensions, so the test's CHECK lines were already what the new code emits and didn't need updating. The diff only touches @fold_unpack_op because the collapse path didn't canonicalize that way (the global linearize + delinearize survived for the dynamic shapes there), so it's where the visible IR change happens.

Where there was a genuine gap is the dynamic-shape expand case — the static @fold_pack_op doesn't exercise the path the PR is really about (dimensions that don't canonicalize away). I've added @fold_pack_op_dynamic mirroring @fold_unpack_op's 2-reassociation-group shape going the other direction, with the per-group affine.delinearize_index into (%[[RES_D0]], 128) / (%[[RES_D1]], 128) checked explicitly.

Re the map_load comment: not the reason the expand-case test isn't in the diff — #24533 currently folds set_encoding chains (containing the expand_shape) into a map_store on CPU today, so the expand path through this PR's per-group fold is hit in practice. The VectorizableOpInterface.cpp comment is documenting why that map_store scalarizes (a lament), not asserting it doesn't happen.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
bjacob: (Claude here, replying on Benoit's behalf — these are my words, not his.)

Added three tests in generic_vectorization_map_store.mlir:

@vectorize_map_store_per_group_linearize — a map_store whose body has the per-group affine.linearize_index disjoint [%idx0, %idx1] by (2, 4) form this PR emits for collapse_shape. With statically-shaped input, the vectorizer accepts it (CHECK includes vector.transfer_read + vector<2x4x16xf32>).
@vectorize_map_store_per_group_delinearize — the expand-direction companion, with per-group affine.delinearize_index … into (2, 4) in the body; also vectorizes when the input is statically shaped.
@no_vectorize_map_store_per_group_delinearize_dynamic — same per-group form but with a dynamic-shape input. The vectorizer skips it (CHECK-NOT vector), consistent with @no_vectorize_map_store_dynamic already in the file.
One subtle nuance worth flagging while writing these: the per-group fold's actual payoff is at the post-vectorization scalarization stage (static shifts/masks vs. dynamic integer division per element), not at the vectorization decision itself — vectorization is gated on inputType.hasStaticShape() and is opaque to the index-transformation form in the body. So the static tests show the per-group IR vectorizes when it can, and the dynamic test shows what reaches scalarization in the typical CPU encoding-relayout case. Added a comment on each test pointing at that distinction so a future reader doesn't expect the index form to flip the vectorization decision.

compiler/src/iree/compiler/Codegen/Interfaces/VectorizableOpInterface.cpp
bjacob: (Claude here, replying on Benoit's behalf — these are my words, not his.)

Good question — the comment is phrased from the set_encoding direction and is implicitly asymmetric. Walking the chain:

The dispatch's result is set_encoding's output. CombineResultLayoutTransformation (#24533) folds the trailing pack/expand_shape/transpose into a single map_store at the dispatch result, and yes — the set_encoding's expand_shape does get folded into that map_store, via this PR's per-group delinearize path. That map_store then iterates the un-padded input (dynamic) and scalarizes — which is what the comment laments. The vectorizable form for set_encoding would be map_load instead, because its static-tiled side is its output.
The unset_encoding earlier in the chain is the symmetric case with directions reversed: its static-tiled side is its input (the encoded matmul result), un-padded side is its output. So for unset_encoding, map_store (iterating static input) is the vectorizable form, and map_load (iterating un-padded output) would scalarize. On CPU we don't currently fold unset_encoding chains into either — the source-side analogue (CombineSourceLayoutTransformation) is wired on GPU but not CPU.
So the comment's "should be map_load" applies cleanly to set_encoding-rooted relayouts (the case #24533 folds today). For unset_encoding-rooted relayouts the vectorizable choice flips to map_store. Stated more carefully, the principle is: combine the relayout into whichever scatter/gather op iterates the statically-tiled side. I'll tighten the comment to make that direction-asymmetry explicit.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
egebeysel: could we then add test cases that demonstrate the new lowering where we do not get a division, and the old case where one actually would get that?

egebeysel (approved):
thanks, LGTM % one test comment.

compiler/src/iree/compiler/Codegen/Common/CombineLayoutTransformation.cpp
bjacob: (Claude here, replying on Benoit's behalf — these are my words, not his.)

Added in the latest push. New EXPAND RUN line at the top of combine_result_layout_transformation.mlir runs the combine pass + --affine-expand-index-ops + canonicalize/CSE, and two tests at the end of the file exercise the contrast:

@lowered_per_group_no_division — same 2-reassociation-group linalg.unpack shape as @fold_unpack_op (tensor<?x?x128x128xf32> → tensor<?x?xf32>). After combine + index expansion, the body is just arith.muli %_, %c128 + arith.addi per reassociation group (the per-group linearize_index disjoint [...] by (%d, 128) lowers to multiply-by-constant-tile-size + add). CHECK-NOT confirms there's no arith.divsi, arith.floordivsi, or arith.remsi anywhere in the lowered body.

@lowered_global_dynamic_divisor — hand-written affine.delinearize_index %idx into (%d0, %d1, %d2, %d3) over a 4D dynamic result shape; this is the IR shape foldReshapeIntoMapStore used to emit before this PR (after the global linearize_index collapsed the source indices into one). After --affine-expand-index-ops the body has arith.muli (trailing-dim products of dynamic factors) and arith.divsi/arith.remsi divided by those SSA-value products — the per-element runtime integer division the per-group fold removes.

The choice of --affine-expand-index-ops is because that's the upstream MLIR pass that scalarizes affine.linearize_index / affine.delinearize_index — it makes the actual divsi/muli shape visible at the lit level without needing a full backend pipeline.

compiler/src/iree/compiler/Codegen/Common/test/combine_result_layout_transformation.mlir
```diff
@@ -600,3 +644,82 @@ func.func @pure_transpose_skipped_under_dispatch_reshape(
...
+//             EXPAND: arith.addi
+//             EXPAND: arith.muli %{{.+}}, %[[C128]]
+//             EXPAND: arith.addi
+//         EXPAND-NOT: arith.divsi
+//         EXPAND-NOT: arith.floordivsi
+//         EXPAND-NOT: arith.remsi
+
+// -----
+
+// Pre-PR (now-removed) global form: one `affine.delinearize_index` over the
```
egebeysel: could we rewrite these comments from a higher level perspective without mentioning pre-PR, or implementation level details, just mentioning that the per-group fold avoids the divsi in both the positive and negative tests?

compiler/src/iree/compiler/Codegen/Common/test/combine_result_layout_transformation.mlir
```diff
@@ -600,3 +644,82 @@ func.func @pure_transpose_skipped_under_dispatch_reshape(
 // RESHAPE-SCOPE-LABEL: @pure_transpose_skipped_under_dispatch_reshape
 //       RESHAPE-SCOPE:   linalg.transpose
 //   RESHAPE-SCOPE-NOT:   iree_linalg_ext.map_store
+
+// -----
+
+// Pre-PR vs. new lowering -- contrast the post-`affine-expand-index-ops` shape
```
egebeysel: ditto, can be much shorter and just mention what the per-group (de-)linearize enables/avoids.

egebeysel (approved):
LGTM, just the test comments Claude generated are a bit too verbose this time :)

compiler/src/iree/compiler/Codegen/Common/test/combine_result_layout_transformation.mlir
bjacob: Same treatment in the latest push. The function-level comment now reads "Negative case for contrast: a global affine.delinearize_index over a fully dynamic result shape — the per-group fold avoids producing this IR," and the CHECK-line comment is one sentence: "Scalarizes to runtime-divisor arith.divsi/remsi (divisors are arith.muli products of the dynamic factors) — the cost the per-group fold avoids." Same trim on the positive test (@lowered_per_group_no_division) and on the top-of-file EXPAND RUN-line comment.

compiler/src/iree/compiler/Codegen/Common/test/combine_result_layout_transformation.mlir
bjacob: Done in the latest push — the block is now four lines, framed in terms of what the per-group fold does (lowered body is division-free; uses muli/addi by the static tile size) and what it avoids (runtime-divisor arith.divsi/remsi from a global delinearize_index over the full dynamic shape). No "pre-PR" or foldReshapeIntoMapStore references.

PR #24539, opened 2026-05-27
[Codegen][Tuner] Add codegen flag to emit pipeline constraints in configured IR

RattataKing:
This added flag will emit constraints in the benchmark IR for the tuner to walk and collect.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
+++ b/compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
@@ -54,6 +54,7 @@ void applyOptLevelDefaults(CodegenOptionsT &opts,
 
 std::string CodegenOptions::tuningSpecPath = "";
 bool CodegenOptions::setTunerAttributes = false;
+bool CodegenOptions::verifyPipelineConstraints = false;
 bool CodegenOptions::emitPipelineConstraints = false;
 
 llvm::OptimizationLevel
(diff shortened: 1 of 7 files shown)


compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
```diff
@@ -94,9 +95,16 @@ void CodegenOptions::bindOptions(OptionsBinder &binder) {
 
   binder.opt<bool>(
       "iree-codegen-experimental-verify-pipeline-constraints",
-      emitPipelineConstraints, llvm::cl::cat(category),
+      verifyPipelineConstraints, llvm::cl::cat(category),
       llvm::cl::desc("Emit and verify SMT pipeline constraints for root ops. "
                      "Implies --iree-codegen-add-tuner-attributes."));
+  binder.opt<bool>(
+      "iree-codegen-emit-pipeline-constraints", emitPipelineConstraints,
```
AGindinson, line 102: Should this option be LIT-tested?

kuhar (approved):

compiler/src/iree/compiler/Codegen/Utils/CodegenOptions.cpp
kuhar: We don't have full smoke tests for full pipelines; this looks safe to land IMO with tests on the python tuner side

PR #24541, opened 2026-05-27
[Codegen][SPIRV] Add Broadcom VideoCore VII GPU target

chrsmcgrr:
Adds a SPIR-V code generation target for the Broadcom VideoCore VII GPU. The GPU can be found in  the Raspberry Pi 5.

The reason why we should have a dedicated configuration, rather than falling back to the default SPIRV config, is that is allows us to tune the tiling heuristic to better match the architecture of the Broadcom GPU. 

Only `f32` matmul/batch-matmul tilings are tuned. Convolutions and other element types fall back to the generic SPIR-V config for now. Follow-up PRs will extend the heuristics to support these. Initial performance numbers will also be given in follow-up PRs. Here we simply set it up.


AGindinson (approved):
LGTM upon formatting fixes.

egebeysel (approved):

Issue #24466, opened 2026-05-13
Retire llvm-external-projects/iree-dialects dependency from IREE


hanhanW:
The `llvm-external-projects` was designed for better integration for other ecosystem (like the old sandbox project in other repo). Most of dialects are either moved to IREE's main codebase (e.g., LinalgExt, VectorExt, ect) or replaced with modern dialect (e.g., `iree-input` dialect is deleted).

There are remaining transformations and code in the dir, and I think we should be able to remove this old and outdated dir. The remaining work is:

## Softmax Matcher

It is used by production path, i.e., RaiseSpecialOps in GlobalOptimization phase. We should be able to port what we need to the local pass. The estimation from LLM is ~200 LOC.

https://github.com/iree-org/iree/blob/487b86b36c02c6a4f14bfe3e90c4056feaefe01d/compiler/src/iree/compiler/GlobalOptimization/RaiseSpecialOps.cpp#L389-L392

## Unused Matchers / Transform ops

There are 7 files in `compiler/src/iree/compiler/Codegen/Common/test` that use `transform.iree.register_match_callbacks`, `transform.iree.match_callback`, `transform.iree.take_first`, and `transform.iree.emit_remark`. These tests primarily cover the old iree-dialects callback matcher infrastructure.

Note: for the matchers that used in ROCm kernel specialization, it uses other matchers like in-tree `transform.iree.match.*` ops. So it is safe to drop them.

## DropSchedulePass

We can move the generic transform schedule cleanup pass from llvm-external-projects/iree-dialects/lib/Dialect/LinalgTransform/Passes/DropSchedulePass.cpp into IREE main, preferably near compiler/src/iree/compiler/Codegen/Common/TransformDialectInterpreter.cpp. Update users in Common, SPIRV, LLVMGPU, iree-opt registration, and init_iree_dialects.h to include/use the in-tree pass.

## ErrorCheckingTrackingListener

Update production users in `compiler/src/iree/compiler/Codegen/Common/TransformExtensions/CommonExtensions.cpp` and `compiler/src/iree/compiler/Codegen/LLVMGPU/TransformExtensions/LLVMGPUExtensions.cpp`. This should allow removing StructuredTransformOpsExtension includes from production code that only need the listener.

Alex-Wengg:
This issue is now resolved

PR #24543, opened 2026-05-27
Move softmax matcher and ErrorCheckingTrackingListener out of iree-dialects (#24466)

Alex-Wengg:
Progress toward #24466 (retire the `llvm-external-projects/iree-dialects`
dependency). Relocates the last two pieces of code that still depended on the NFC directory. Finishing Tasks 1 & 4.

- **[GlobalOpt] Move softmax matcher**: relocate the `StructuredOpMatcher`  infrastructure + `makeSoftmaxMatcher` into `GlobalOptimization/`.
- **[Codegen] Move ErrorCheckingTrackingListener** into `Codegen/Common/` and  de-register `StructuredTransformOpsExtension` from `CommonDialectRegistration` and `Interfaces`. continuation of #24500.

Once this is merged, we can start removing the llvm-external-projects/iree-dialects as a whole

## Testing

Built and lit tested locally on Windows 11 + MSVC 14.44 (VS 2022 Community), Ninja, CMake 3.29, Release with assertions, `IREE_TARGET_BACKEND_LLVM_CPU=ON`. All changed files compile cleanly. The softmax lit test `compiler/src/iree/compiler/GlobalOptimization/test/raise_special_ops.mlir` passed. all three variants (`@softmax`, `@softmax_no_rcp`, `@softmax_broadcast`) match correctly through the relocated `makeSoftmaxMatcher`.


L-roro:
Hi there @Alex-Wengg , small question about the `TransformMatchers` move!

From the mentioned issue, I understood that the objective was moving only the softmax matcher needed by `RaiseSpecialOps.cpp`, given that there are no other usages of the rest of the API and its associated generic matcher framework. In this PR, the full `TransformMatchers.{h,cpp}` files are copied into `GlobalOptimization` though.

Is keeping the full matcher helper library intentional? If not, would it make sense to trim this to `makeSoftmaxMatcher` and the support code it needs?

Alex-Wengg:
@L-roro for extra context, i am still a novice when it comes to IREE so i don't have all the context but based on my additional scan, it does seem like what you have suggested is indeed correct. however i wonder if that might be out of scope ffor this PR and it could get quite messy.

hanhanW (dismissed):
This is not just a moving code task. You need to revisit the semantics and maybe rework on the matchers. That's the main reason why I filed the issue instead of doing it myself.

Alex-Wengg:
@hanhanW @L-roro apologies, I misunderstood the task at hand

I deleted TransformMatchers.{h,cpp} and reimplemented softmax detection in RaiseSpecialOps.cpp to the best of my abilities. Dropped the iree-dialects deps as well.

While reimplementing I noticed the matcher only accepted arith.maximumf, but linalg.softmax decomposes to arith.maxnumf  from my understanding that  would suggest the form wasn't being raised at all. I widened the matcher to accept both. I kept the existing innermost-dim-only / max-stabilized assumptions as-is  not sure if those should be relaxed, let me know if that's what you had in mind.

For testing I rebased onto latest main and rebuilt from scratch (incl. the recent LLVM bumps). It built cleanly, the GlobalOpt lit suite is green, and MNIST + a small StableHLO attention block both raise to linalg.softmax and run correctly end-to-end on CPU.

pls lmk if this is closer to what is in mind, i will reiterate if not.

Alex-Wengg:
i don't think the fails are related to my changes

compiler/src/iree/compiler/GlobalOptimization/test/raise_special_ops.mlir
```diff
@@ -168,6 +168,158 @@ util.func public @softmax_broadcast(%93 : tensor<12x128x128xf32>) -> (tensor<12x
 
 // -----
 
+// The stabilizing max may use arith.maxnumf (NaN-ignoring) instead of
+// arith.maximumf -- this is the form linalg.softmax itself decomposes to.
```
AGindinson: A small nit: in my opinion, such contextual comments are more appropriate in commit messages than in persistent code (kudos for explaining this in the commit message btw!). E.g. were the softmax decomposition to change, the comment could become outdated and confusing (but totally fine when you look it up in `git blame` - then it's tied to a specific point in history). Especially in the LIT tests, which are often the quickest path to understand some functionality as a reader, this can lead to unnecessary assumptions. 
```suggestion
// The stabilizing max may use arith.maxnumf (NaN-ignoring).
```

AGindinson (commented):
Should the re-implemented `transform_ext` bits be removed within the same PR for clarity? Other than that and overall, LGTM - feels like good progress, and thanks for working on this.

Alex-Wengg:
@AGindinson i have removed transform_ext and tried to reduce comments that i felt too verbose

AGindinson (approved):
Thanks! I had initially thought that only the corresponding functionality would be dropped, but it actually makes sense to just drop the retired code right away. I will reflect this in the squashed commit message when merging.

Alex-Wengg:
i don't think PkgCI is related to my issue
https://github.com/iree-org/iree/actions/runs/27782623047/job/82233204198?pr=24543

linux_x64_bazel is taking quite a long time
https://github.com/iree-org/iree/actions/runs/27782621469/job/82231734930?pr=24543

PR #24544, opened 2026-05-27
[CI] Fix spirv rocm filter when rocm is not build

maxbartel:
ci-extra: linux_arm64_clang,macos_x64_clang


krzysz00 (approved):
I'm not a Bazel-head ( @benvanik ) so I don't know if this is the *right* way to make platform-specific tests, but I think this'll work and un-break CI, so approved.

kuhar (approved):

PR #24548, opened 2026-05-29
[Codegen][SPIRV] Add Initial Conv2d Tiling for VideocoreVII

chrsmcgrr:
Follow-up on the previous PR.

The Broadcom VideoCore VII (RPi5 GPU) has 32 threads per workgroup and 16 KiB of workgroup memory. Memory could not be derived explicitly from vulkaninfo but was assumed from previous generations. Both characteristics are an order of magnitude smaller than the GPU default the SPIRV conv config defines. So its workgroup tile and pipeline choices overflow the device.

This PR therefore introduces a variant of the conv2d tiling configuration chosen around a power-of-two divisor at each level that attempts to keep loads coalesced within the VideoCore's budget. We restrict the support for this config to NCHW f32 conv and depthwise for now. NHWC and non-f32 are intentionally rejected so the default path stays for those cases.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
@@ -1237,7 +1237,7 @@ const WgpDetails *getVideocoreWgpDetails() {
       /*scaledMmaCount=*/0,     /*scaledMmaOps=*/nullptr,
       {16, 16},                 {256, 256, 256},          32,
       16384,
-      {65535, 65535, 65535}};
+      {0xffff, 0xffff, 0xffff}};
   // clang-format on
   return &androidWgp;
 }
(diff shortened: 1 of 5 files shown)


compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
```diff
@@ -1222,11 +1222,31 @@ const WgpDetails *getAndroidBaseline2022WgpDetails() {
...
+  auto computeBitwdiths = ComputeBitwidths::Int32 | ComputeBitwidths::FP32;
+  auto storageBitwidths = StorageBitwidths::B32;
+  // clang-format off
+  static const WgpDetails androidWgp = {
+      computeBitwdiths,         storageBitwidths,         SubgroupOps::None,
+      DotProductOps::None,      /*mmaCount=*/0,           /*mmaOps=*/nullptr,
+      /*scaledMmaCount=*/0,     /*scaledMmaOps=*/nullptr,
+      {16, 16},                 {256, 256, 256},          32,
+      16384,
+      {65535, 65535, 65535}};
```
AGindinson:
```suggestion
      {0xffff, 0xffff, 0xffff}};
```

AGindinson (approved):

compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp
chrsmcgrr: Nice Suggestion :)

PR #24549, opened 2026-05-29
[HAL] Add missing bufferUsage for Mapping in tryAddSharedUsageBits

devtbi:
- Map persistently to support sharing via export/import.
- Persistent mapping in turn requires host-visible memory.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/HAL/Transforms/ResolveTopologyQueries.cpp
+++ b/compiler/src/iree/compiler/Dialect/HAL/Transforms/ResolveTopologyQueries.cpp
@@ -132,6 +132,12 @@ static bool tryAddSharedUsageBits(IREE::HAL::DeviceTopologyAttr topology,
   for (auto affinity : optimalAttr.getAffinities()) {
     if (hasTransparentAccessToAll(topology, affinity, optimalAttr)) {
       bufferUsage = bufferUsage | IREE::HAL::BufferUsageBitfield::Mapping;
+      // This buffer is shared across devices. Map persistently to
+      // support sharing via export/import.
+      // Persistent mapping in turn requires host-visible memory.
+      bufferUsage =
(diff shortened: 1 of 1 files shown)


compiler/src/iree/compiler/Dialect/HAL/Transforms/ResolveTopologyQueries.cpp
```diff
@@ -132,6 +132,10 @@ static bool tryAddSharedUsageBits(IREE::HAL::DeviceTopologyAttr topology,
   for (auto affinity : optimalAttr.getAffinities()) {
     if (hasTransparentAccessToAll(topology, affinity, optimalAttr)) {
       bufferUsage = bufferUsage | IREE::HAL::BufferUsageBitfield::Mapping;
+      // Mapping currently always requires HOST_VISIBLE and MAPPING_PERSISTENT.
```
schuermans-roofline: I understand that it is desired to require the buffer to support persistent mapping and being host-visible here in order to support export/import patterns to share the buffer among devices. However, I think this should be stated as the reason instead of the current comment. Technically, the HAL layer itself (the checks in iree_hal_buffer_map_range() in third-party/iree/runtime/src/iree/hal/buffer.c) does not require MAPPING_PERSISTENT and/or HOST_VISIBLE, e.g., for scoped mapping. (It requires HOST_VISIBLE for MAPPING_PERSISTENT, but that's a slightly different thing.)

schuermans-roofline (changes requested):
I agree with the changed code, but I'd like the comment on it to be improved a bit.

compiler/src/iree/compiler/Dialect/HAL/Transforms/ResolveTopologyQueries.cpp
devtbi: shoulds be (more) accurate now

schuermans-roofline (approved):
LGTM

PR #24551, opened 2026-05-29
[Security] Fix uint64 additive overflow in VM bytecode rodata segment verifier

0xASTRA:
## Summary

\`verifier.c:75-77\` computes \`segment_end = archive_rodata_offset + segment_offset + segment_length\` as a plain 64-bit unsigned addition. FlatBuffer \`uint64\` scalars carry no per-field cap, so attacker-controlled \`external_data_offset + external_data_length\` values can wrap past 2^64, producing a small \`segment_end\` that falsely passes the \`> data_length\` check.

\`module.c:1015-1018\` then constructs the rodata buffer span from the original unwrapped values (comment: *"Note that we've already verified the referenced range is in bounds"*), yielding a wild pointer. The runtime guard in \`buffer.c\` validates against the attacker-controlled length, not the true archive bounds.

## Fix

Replaced the additive sum with non-overflowing subtraction comparisons that reject any \`segment_offset\` or \`segment_length\` that would exceed available archive space.

## Also flagged

The same unchecked-add pattern exists in \`archive.c:184\` (\`iree_vm_bytecode_archive_infer_size\`) and should be addressed similarly.

### DIFF
--- a/runtime/src/iree/vm/bytecode/verifier.c
+++ b/runtime/src/iree/vm/bytecode/verifier.c
@@ -72,9 +72,13 @@ iree_status_t iree_vm_bytecode_module_flatbuffer_verify(
         iree_vm_RodataSegmentDef_external_data_offset(segment);
     uint64_t segment_length =
         iree_vm_RodataSegmentDef_external_data_length(segment);
-    uint64_t segment_end =
-        archive_rodata_offset + segment_offset + segment_length;
-    if (segment_end > archive_contents.data_length) {
+    // Use non-overflowing comparisons instead of summing three uint64_t values:
+    // archive_rodata_offset + segment_offset + segment_length could wrap 2^64,
+    // producing a small segment_end that falsely passes the bounds check.
+    iree_host_size_t avail = archive_contents.data_length;
+    if (archive_rodata_offset > avail ||
+        segment_offset > avail - archive_rodata_offset ||
+        segment_length > avail - archive_rodata_offset - segment_offset) {
       return iree_make_status(IREE_STATUS_INVALID_ARGUMENT,
                               "rodata[%zu] external reference out of range", i);
     }


AGindinson (changes requested):
Hi @0xASTRA, thanks for flagging the issue and contributing the fix! LGTM functionally.

Could you please sign off the commit, as per the failing DCO check / https://iree.dev/developers/general/contributing/#developer-certificate-of-origin?

Also please update the description in line with https://iree.dev/developers/general/contributing/#ai-tool-use (please correct me if I'm wrong, but the "test plan" segment in particular reads like an LLM artifact to me).

0xASTRA:
Can you check it now?

runtime/src/iree/vm/bytecode/verifier.c
```diff
@@ -72,9 +72,12 @@ iree_status_t iree_vm_bytecode_module_flatbuffer_verify(
...
     uint64_t segment_length =
         iree_vm_RodataSegmentDef_external_data_length(segment);
-    uint64_t segment_end =
-        archive_rodata_offset + segment_offset + segment_length;
-    if (segment_end > archive_contents.data_length) {
+    // Use non-overflowing comparisons instead of summing three uint64_t values:
+    // archive_rodata_offset + segment_offset + segment_length could wrap 2^64,
+    // producing a small segment_end that falsely passes the bounds check.
+    iree_host_size_t avail = archive_contents.data_length;
+    if (segment_offset > avail - archive_rodata_offset ||
```
AGindinson: Given we're dealing with an unsigned type, would this also be needed?
```suggestion
    if (archive_rodata_offset > avail ||
        segment_offset > avail - archive_rodata_offset ||
```

0xASTRA: Good catch. Without it, avail - archive_rodata_offset could underflow on unsigned types if the offset itself exceeds the archive size.

AGindinson (approved):
Thanks!

0xASTRA:
> Thanks!

No problem! Is there anything pending on my side?

AGindinson:
> > Thanks!
> 
> No problem! Is there anything pending on my side?

All good, this should be automatically merged upon passing CI checks.

PR #24553, opened 2026-05-29
[Bazel] Add rt python bindings to bazel

maxbartel:
This is the first PR of a longer arc where I will add support to bazel for building the python wheels natively.

Depends on https://github.com/llvm/llvm-project/pull/200516


AGindinson (approved):
LGTM

PR #24566, opened 2026-06-03
[Torch] Support explicit scale values from `tm_tensor.attention`

AGindinson:
With dynamically sized GQA, explicit scale settings show up in ONNX/Torch models such as Qwen3. Consume scale values from TMTensor op when provided.

Depends on llvm/torch-mlir#4593.

### DIFF
--- a/compiler/plugins/input/Torch/InputConversion/ConvertTMTensorToLinalgExt.cpp
+++ b/compiler/plugins/input/Torch/InputConversion/ConvertTMTensorToLinalgExt.cpp
@@ -4,6 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include <mlir/IR/BuiltinAttributes.h>
 #include <cstdint>
 #include <numeric>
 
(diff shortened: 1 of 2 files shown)


keshavvinayak01 (approved):
It looks good broadly.  I was about to bring in the same change, but then I started [a parallel work](https://github.com/iree-org/iree/pull/24177/changes#diff-fb976af2f02a70ff05e6775a1506ea8901d84b04a36becfb305a65533465bd54) to completely remove the AttentionOp itself, since it's functionally the same as OnlineAttention, with most of the heavy lifting done (DispatchCreation etc.). This is a fine stop-gap solution, but since I'm not working primarily on IREE anymore, with more of my time focused on LLVM, if you or anyone else wants to take up that task I'd really appreciate it.

AGindinson:
> It looks good broadly. I was about to bring in the same change, but then I started [a parallel work](https://github.com/iree-org/iree/pull/24177/changes#diff-fb976af2f02a70ff05e6775a1506ea8901d84b04a36becfb305a65533465bd54) to completely remove the AttentionOp itself, since it's functionally the same as OnlineAttention, with most of the heavy lifting done (DispatchCreation etc.). This is a fine stop-gap solution, but since I'm not working primarily on IREE anymore, with more of my time focused on LLVM, if you or anyone else wants to take up that task I'd really appreciate it.

@keshavvinayak01 Thanks for giving this a look! I was meaning to review your change in turn and then it fell through my stack back then. I will note it down for someone to pick up the discussion on your PR and see it to `main` as needed.

PR #24567, opened 2026-06-04
[Codegen][CPU] Seed C-bitcode ukernel framework with bf16 and i8 seeds.

bjacob:
Adds compiler/plugins/target/LLVMCPU/builtins/ukernel/, the new home for LLVMCPU-only C microkernels. Modelled after the ROCM C ukernel framework (compiler/plugins/target/ROCM/builtins/ukernel/): compile each ukernel to LLVM bitcode, embed those .bc files into iree-compile, and emit them as self-contained `hal.executable_object`s on dispatch variants — distinct from the legacy mmt4d ukernels under runtime/, which serve a now-defunct VMVX path and are built twice.

This first commit lands the scaffolding plus two inner-K-loop seeds with stub bodies (filled in by a later commit):
  * `iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c` (bf16 x bf16 -> f32) and `iree_uk_mma_x86_avx512vnni_16x16x2_i32_i8_casti16.c` (i8 x i8 -> i32 via VNNI), plus `common.h` (no-stdlib stdint replacements);
  * `iree_bitcode_library` + `iree_c_embed_data` rules, driven by a `(stem, features, copts)` table so adding a seed is one row (CMake regenerated; `iree_bitcode_library` in `bazel_to_cmake_converter.py` extended to forward `out`, so the embedded filename is `<name>.<arch>_<features>.bc`);
  * a `README.md` design doc covering scope, build, IR representation, end-to-end flow, and authoring/test guidance;
  * `#iree_cpu.ukernel_provider`, the CPU analogue of `#rocm.ukernel_provider`, declared in `IREECPUAttrs.td`/`.cpp` with a `createAndReplaceWithUkernelOp` that currently delegates to the default `LowerBitcodeUKernelsPass` fallback. Specialized `inner_tiled` handling (threading `intrinsics_{m,n,k}` and the outer K count) will land in a follow-up alongside `SelectUKernels` and pipeline integration.

Tests:
  * roundtrip lit test for the new `#iree_cpu.ukernel_provider` attribute;
  * lit tests running `--iree-codegen-lower-bitcode-ukernels` on ops carrying both `iree_codegen.ukernel = "..."` and a user-supplied `hal.executable.objects` bitcode (one per seed), asserting they are rewritten to `iree_codegen.ukernel.generic` with the bitcode preserved.


compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
```diff
@@ -0,0 +1,52 @@
...
+//
+// ABI: each shaped operand is passed as (base pointer, element offset) so the
+// caller doesn't need a GEP before the call; the accumulator additionally
+// gets the element stride of its innermost cross-intrinsic (N) dimension.
+//
+// NOTE (seed scaffolding): this initial seed has a stub body. It exists so
+// that the surrounding *framework* -- bitcode build, embedding,
+// `hal.executable_object` injection, IR rewrite to `ukernel.generic` -- can
+// be landed and lit-tested. A follow-up commit replaces the body with the
+// `_mm512_dpbf16_ps`-based inner loop and adds an e2e matmul test for it.
```
egebeysel, line 32: as long as the follow-up commits remember to adjust these comments :)

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
```diff
@@ -0,0 +1,52 @@
...
+#include "common.h"
+
+// Microkernel for `iree_codegen.inner_tiled` with
+// `#iree_cpu.data_tiled_mma_layout<intrinsic =
+//     MMA_X86_AVX512BF16_1x16x2_F32_BF16>`. Function name matches the
+// intrinsic name verbatim (lowercased, with the `iree_uk_` prefix), in line
+// with the AMDGPU C ukernel convention.
+//
+// The "inner K loop" the ukernel owns is the loop over the K *tiles* that
+// sits *inside* the outer M/N loops; those outer M/N loops are tiled away by
```
egebeysel, line 17: IIUC, this is the same scope as the legacy C ukernels, right?

egebeysel (approved):
Thanks! LGTM!

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
bjacob: The legacy C ukernels have non-arch-specialized outer M/N loops around a `tile_func` which is the part that is specialized per architecture:

https://github.com/iree-org/iree/blob/main/runtime/src/iree/builtins/ukernel/mmt4d.c#L84-L99

The new C ukernels are equivalent to what the legacy C ukernels called their `tile_func`.

In practice, already in the legacy C ukernels, we were in practice tiling down to 1 tile, meaning that those outer-loops on M/N tended to have only 1 iteration as a compile-time known constant, and later optimizations would then elide these trivial loops. But it wasn't structural, it was the result of a chain of optimizations. The new C ukernels make that structural.

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
bjacob: Yes, that is https://github.com/iree-org/iree/pull/24572 and it does update these comments.

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
egebeysel: Oh alright, I see 👍 Thanks!

PR #24568, opened 2026-06-04
[Codegen][CPU] Register embedded ukernel bitcode + provider lookup.

bjacob:
Wires the LLVMCPU plugin and the `#iree_cpu.ukernel_provider` attribute to actually resolve ukernel bitcode:

  * On plugin init, the embedded `iree_uk_cpu_bitcode` TOC is iterated and each `.bc` is added to the global `EmbeddedDataDirectory` (mirrors `addAMDGPUUkernelBitcodeToGlobalEmbeddedDataDirectory` in `ROCMTarget.cpp`).

  * `UKernelProviderAttr::createAndReplaceWithUkernelOp` now resolves the bitcode by name: user-supplied `hal.executable.objects` ancestors first (BYO and BYO-override paths), then the embedded directory. If found and not already on the op, the attribute is attached to the source op; the default `LowerBitcodeUKernelsPass` rewrite then preserves it on the resulting `ukernel.generic`. The provider still returns `std::nullopt` for the rewrite itself — specialized `inner_tiled` handling (scalar `intrinsics_{m,n,k}` + outer K count + ACC inner stride) lands with SelectUKernels in a follow-up.

  * Filename match is `<ukernelName>.<features>.bc` (or bare `<ukernelName>.bc`), matching the build rules under `compiler/plugins/target/LLVMCPU/builtins/ukernel/`. Multi-variant disambiguation by target feature set also lands with SelectUKernels.

BUILD.bazel: LLVMCPU plugin gains a dep on
`//compiler/plugins/target/LLVMCPU/builtins/ukernel:iree_uk_cpu_bitcode`, and the `iree_cpu` dialect picks up HAL IR + `EmbeddedDataDirectory` so the provider implementation can live with the attribute (consistent with how `ROCMUkernelBitcodeSupport.cpp` sits inside the ROCM dialect).

Tests: a three-way lit test (`lookup_builtin_and_byo.mlir`) covers each branch of the lookup:
  1. Built-in only — no user objects; expect a `dense_resource<…>` with the embedded bf16 bitcode (2732 bytes) attached.
  2. BYO override — user supplies bitcode under the built-in's filename; expect the user's bytes to flow through unchanged (and CHECK-NOT against `dense_resource` to catch a regression where the provider silently swaps in the embedded copy).
  3. Pure BYO — user-only name with no matching built-in; expect the user's bitcode to surface on the rewritten op.


compiler/plugins/target/LLVMCPU/builtins/ukernel/test/lookup_builtin_and_byo.mlir
```diff
@@ -0,0 +1,161 @@
...
+    %2 = linalg.generic {
+      indexing_maps = [#map, #map1, #map2],
+      iterator_types = ["parallel", "parallel", "reduction"]
+    } ins(%arg0, %arg1 : tensor<16x2xbf16>, tensor<16x2xbf16>)
+      outs(%1 : tensor<16x16xf32>) attrs = {
+      iree_codegen.ukernel = #iree_codegen.ukernel_descriptor<
+          "my_custom_external_ukernel", bitcode>,
+      hal.executable.objects = [
+        #hal.executable.object<{
+          path = "my_custom_external_ukernel.bc",
```
egebeysel, line 148: question: are the users required to insert their own passes to the pipeline to match the operation that they have a ukernel for, compile their own ukernels into bc and read them into `hal.executable.object`s to generate this IR?

egebeysel (approved):
Thanks a lot! LGTM, just one question 😄

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/lookup_builtin_and_byo.mlir
bjacob: There are many different ways that this can go and I just notice that this (AI-generated) test spontaneously goes over a few distinct variants. "Inserting their own passes" is definitely not the only option (unlike the legacy system where it was the only options), like the test shows, alternatives include annotating source ops with ukernel-match attributes (https://github.com/iree-org/iree/pull/24568/changes/BASE..a66a600a274c59a265b4ed342ce4e7e1f9418416#diff-1283a0ca5b2f27235ed180f4e84764f9e8dc649c0f81be42d1a9d16bc537b5afR50), or conceivably writing ukernel ops directly in the source (since ukernel ops accept tensor argument i.e. can be used before bufferization).

This is all a bit distant in my memory. Schedule a video meeting with me if you want us to go over this :-)

egebeysel: I see, I'll take a look at the ROCm side of things as well and study that a bit, and schedule a meeting with you afterwards! Thanks!

PR #24569, opened 2026-06-04
[Codegen][CPU] SelectUKernels for inner_tiled + --iree-llvmcpu-enable-llvm-ukernels.

bjacob:
Wires the new C-bitcode ukernel framework into the LLVMCPU kernel-config pipeline. Three pieces:

  * `LLVMCPUSelectUKernels.{h,cpp}` — when the target config has `llvm_ukernels = "inner_tiled"`, `selectUKernel` matches `iree_codegen.inner_tiled` ops carrying a `#iree_cpu.data_tiled_mma_layout` and returns a `UKernelDescriptorAttr` whose name is the corresponding `MMAIntrinsic` enum value lowercased with the `iree_uk_mma_` prefix (e.g. `MMA_X86_AVX512BF16_1x16x2_F32_BF16` → `iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16`). One ukernel per intrinsic; `intrinsics_{m,n,k}` flow as function arguments and specialize via bitcode LTO at the call site. Filters out `None` and the `MMA_GENERIC_*` family — neither has a corresponding ukernel.

  * `--iree-llvmcpu-enable-llvm-ukernels=<csv>` CL flag — bound to a new `LLVMTarget.llvmUkernels` field, serialized into the target config as `llvm_ukernels = "<csv>"`. Empty (the default) means off; passing `inner_tiled` opts dispatches into the new framework. The flag name leaves room for a future `mlir_ukernels`-style additional category (and for now `inner_tiled` is also the only value). New helper `hasLlvmUkernel(config, category)` in Codegen/Utils — distinct from the existing `hasUkernel` to keep the legacy mmt4d ukernels and the new LLVM-bitcode ukernels on independent enable flags.

  * `KernelDispatch.cpp::setRootConfig(InnerTiledOp)` — calls `selectUKernel` at the start and, on a match, sets the descriptor via `setUKernelDescriptor`. Tile sizes and translation_info are unchanged: the ukernel substitutes for the *innermost* intrinsic execution only, the surrounding tile sizes and pipeline still hold. Bitcode resolution is deferred to the `#iree_cpu.ukernel_provider` during `LowerBitcodeUKernelsPass`, matching the GPU split between "match" (here) and "attach" (provider).

Lit test (`select_ukernel.mlir`) runs
`iree-llvmcpu-select-lowering-strategy` on two `inner_tiled` ops sharing the same MMA intrinsic but differing in target config: the one with `llvm_ukernels = "inner_tiled"` picks up the
`iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16` descriptor; the other does not, locking in the "off by default" property.

### DIFF
--- a/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.cpp
+++ b/compiler/plugins/target/LLVMCPU/LLVMTargetOptions.cpp
@@ -196,6 +196,9 @@ void LLVMTarget::storeToConfigAttrs(MLIRContext *context,
   if (ukernels.compare(DEFAULT_ENABLE_UKERNELS) != 0) {
     addString("ukernels", ukernels);
   }
+  if (llvmUkernels.compare(DEFAULT_ENABLE_LLVM_UKERNELS) != 0) {
+    addString("llvm_ukernels", llvmUkernels);
+  }
   if (linkUkernelBitcode != DEFAULT_LINK_UKERNEL_BITCODE) {
     addBool("link_ukernel_bitcode", linkUkernelBitcode);
   }
@@ -327,6 +330,7 @@ LLVMTarget::loadFromConfigAttr(Location loc, DictionaryAttr config,
   }
 
   target.ukernels = getString("ukernels", target.ukernels, false);
+  target.llvmUkernels = getString("llvm_ukernels", target.llvmUkernels, false);
   target.linkUkernelBitcode =
       getBool("link_ukernel_bitcode", target.linkUkernelBitcode);
   target.enableInnerTiled =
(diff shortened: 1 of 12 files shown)


compiler/plugins/target/LLVMCPU/LLVMTargetOptions.cpp
```diff
@@ -602,6 +606,16 @@ void LLVMCPUTargetCLOptions::bindOptions(OptionsBinder &binder) {
...
                      "`default`, `none`, `all`, or a comma-separated list of "
                      "specific unprefixed ukernels to enable, e.g. `mmt4d`."));
+  binder.opt<std::string>(
+      "iree-llvmcpu-enable-llvm-ukernels", enableLlvmUkernels,
+      llvm::cl::cat(category),
+      llvm::cl::desc(
+          "Enables the new-style LLVM-bitcode C ukernels under "
+          "compiler/plugins/target/LLVMCPU/builtins/ukernel/. "
+          "Comma-separated list of categories to enable, e.g. "
+          "`inner_tiled`. Empty (the default) means none. The companion "
```
egebeysel, line 620: just a reminder to note this down and remove it once MLIR-ukernels are there

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
```diff
@@ -0,0 +1,76 @@
...
+// `--iree-llvmcpu-enable-llvm-ukernels=inner_tiled` CL flag at
+// `iree-compile` time; we set it directly in the test since `iree-opt`
+// doesn't run target-option serialization).
+//
+// Two cases:
+//   (1) enabled:  `selectUKernel` matches the BF16 1x16x2 intrinsic and
+//                 annotates the op with `iree_codegen.ukernel = …`.
+//   (2) disabled: target config lacks `llvm_ukernels`, so `selectUKernel`
+//                 returns null and no descriptor is added — the default,
+//                 keeping the new framework off in plain builds.
```
egebeysel: Maybe these could be moved before the respective tests, so that they don't have to be modified with additional tests coming in

egebeysel (approved):
Thanks!

compiler/plugins/target/LLVMCPU/LLVMTargetOptions.cpp
bjacob: Done

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
bjacob: Done

PR #24570, opened 2026-06-04
[Codegen][CPU] Wire LowerBitcodeUKernels into the Mmt4dTilingExpert pipeline.

bjacob:
Completes the new-style C-ukernel pipeline integration:

  * `LLVMCPUTarget.cpp`: install `iree_codegen.ukernel_provider = #iree_cpu.ukernel_provider` on every LLVMCPU executable target config whenever `--iree-llvmcpu-enable-llvm-ukernels` is non-empty. This mirrors what ROCM does and is what lets the generic `LowerBitcodeUKernelsPass` find the CPU provider for the bitcode-lookup-and-attach.

  * `KernelDispatch.cpp::setRootConfig(InnerTiledOp)`: when `selectUKernel` returns a descriptor, set the reduction-dim vector tile size to 0 (untiled) instead of 1 (unrolled). The ukernel's C function takes the outer K extent as a runtime argument (`k_outer`) and runs the K loop itself; pre-unrolling here would both defeat that and bloat the IR. Matches the equivalent split in `setDataTiledMmaInnerTiledLoweringConfig` on the GPU side.

  * `Passes.cpp::addMmt4dTilingExpertPassPipeline`: insert `createLowerBitcodeUKernelsPass()` between the parallel-tile and reduction-tile passes. A nop when no op carries a `iree_codegen.ukernel` descriptor, so it stays in the default pipeline. The companion `LowerUKernelOpsToCalls` already runs after bufferization (in `addLowerToLLVMPasses`); together they implement the user's "early `inner_tiled` -> `ukernel.generic` on tensors; late `ukernel.generic` -> `func.call` on memrefs" split.

Lit test (`select_ukernel.mlir`) gains a second RUN line that chains `iree-llvmcpu-select-lowering-strategy` with
`iree-codegen-lower-bitcode-ukernels` and asserts:
  - the `inner_tiled` is rewritten to `iree_codegen.ukernel.generic` carrying the right `iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16` name;
  - the matching bitcode appears as `hal.executable.objects` (resolved from the global `EmbeddedDataDirectory` populated at LLVMCPU plugin init);
  - and the original `inner_tiled` is gone, guarding against the rewrite silently no-op'ing on a regression.

An end-to-end `iree-compile` test would also be desirable, but the current pipeline state needs more wiring (workgroup-distribution verifier complains about the bufferization-introduced copy ops around the workgroup loop, on both the ukernel and non-ukernel paths) — that investigation belongs with the follow-up commit that fills in real ukernel bodies and an actual numerical matmul test.

### DIFF
--- a/compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp
+++ b/compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp
@@ -180,6 +180,14 @@ public:
     configItems.emplace_back(
         b.getStringAttr(IREE::Encoding::kEncodingResolverAttrName),
         IREE::CPU::CPUEncodingResolverAttr::get(context, {}));
+    // When at least one LLVM-bitcode ukernel category is enabled, install
+    // the CPU ukernel provider attribute so that the subsequent
+    // `LowerBitcodeUKernelsPass` can find and attach the matching
+    // bitcode.
+    if (!target.llvmUkernels.empty()) {
+      configItems.emplace_back(b.getStringAttr(kUKernelProviderName),
+                               IREE::CPU::UKernelProviderAttr::get(context));
+    }
 
     // Compute the format used at runtime to select the executable loader.
     std::string format;
--- a/compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
+++ b/compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
@@ -1,5 +1,7 @@
 // RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' \
 // RUN:   --split-input-file %s | FileCheck %s
+// RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy,func.func(iree-codegen-lower-bitcode-ukernels))' \
+// RUN:   --split-input-file %s | FileCheck %s --check-prefix=CHAIN
 
 // Drives the LLVMCPU `LLVMCPUSelectLoweringStrategy` pass on
 // `inner_tiled` rooted dispatches carrying a CPU `DataTiledMMAAttr`. The
@@ -10,10 +12,12 @@
 // doesn't run target-option serialization).
 
 // Enabled case: `selectCPUUKernel` matches the BF16 1x16x2 intrinsic and
-// annotates the op with `iree_codegen.ukernel = …`.
+// annotates the op with `iree_codegen.ukernel = …`. The chained CHAIN run
+// additionally lowers that descriptor to a `ukernel.generic`.
 #executable_target_enabled = #hal.executable.target<"llvm-cpu", "embedded-elf-x86_64", {
   cpu_features = "+avx512f,+avx512bf16",
   data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128",
+  iree_codegen.ukernel_provider = #iree_cpu.ukernel_provider,
   llvm_ukernels = "inner_tiled",
   native_vector_size = 64 : index,
   target_triple = "x86_64-unknown-unknown-eabi-elf"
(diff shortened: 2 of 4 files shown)


compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
```diff
@@ -16,9 +18,14 @@
 //                 returns null and no descriptor is added — the default,
 //                 keeping the new framework off in plain builds.
 
+// The `iree_codegen.ukernel_provider = #iree_cpu.ukernel_provider` here is
+// what `LLVMCPUTarget.cpp` adds automatically whenever `llvm_ukernels` is
+// non-empty; we set it directly because `iree-opt` doesn't run target-option
+// serialization.
```
egebeysel:
```suggestion
```

Seems to much like an implementation detail leaking into a test comment imo

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
```diff
@@ -1,5 +1,7 @@
 // RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' \
 // RUN:   --split-input-file %s | FileCheck %s
+// RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy,func.func(iree-codegen-lower-bitcode-ukernels))' \
```
egebeysel, line 3: Does it even make sense to have this test in this file? maybe it would make more sense to move this into its own file.

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
```diff
@@ -45,6 +52,19 @@ func.func @bf16_inner_tiled_ukernel_enabled(
...
 // CHECK-SAME:      iree_codegen.ukernel = #iree_codegen.ukernel_descriptor<"iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16", bitcode>
 
+// CHAIN-LABEL: func.func @bf16_inner_tiled_ukernel_enabled
+// Once SelectLoweringStrategy has set the descriptor, the very next pass in
+// the pipeline (`LowerBitcodeUKernelsPass`, also installed into the
+// `Mmt4dTilingExpert` pipeline) is expected to rewrite the `inner_tiled`
+// into a `ukernel.generic` and attach the matching bitcode resolved by the
+// `#iree_cpu.ukernel_provider`. The CHAIN-NOT line guards against the
+// rewrite silently skipping the op (the original `inner_tiled` must be gone
+// from the rewritten function body).
```
egebeysel: ditto, I think a more refined summary of what the test expects to see rather than so many pointers to the implementation would be nicer

compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp
```diff
@@ -180,6 +180,16 @@ class LLVMCPUTargetBackend final : public TargetBackend {
     configItems.emplace_back(
         b.getStringAttr(IREE::Encoding::kEncodingResolverAttrName),
         IREE::CPU::CPUEncodingResolverAttr::get(context, {}));
+    // When at least one LLVM-bitcode ukernel category is enabled, install
+    // the CPU `UKernelProviderAttr` so `LowerBitcodeUKernelsPass` can find
+    // and attach the matching bitcode. Mirrors the ROCM side's
+    // `addConfig("iree_codegen.ukernel_provider", …)` and is the bridge
+    // between the `--iree-llvmcpu-enable-llvm-ukernels` CL flag and the
+    // generic provider-lookup machinery.
```
egebeysel:
```suggestion
    // When at least one LLVM-bitcode ukernel category is enabled, install
    // the CPU ukernel provider attribute so that the subsequent
    // `LowerBitcodeUKernelsPass` can find and attach the matching
    // bitcode.
```

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
```diff
@@ -3085,11 +3083,21 @@ static LogicalResult setRootConfig(mlir::FunctionOpInterface entryPointFn,
...
-  // `vector_reduction`.
+  // Vector tiling: every parallel iter dim down to a single inner tile.
+  // Reduction dims tile to 1 by default — codegen unrolls the K loop. But
+  // when a ukernel is in play, leave reduction dims untiled (size 0): the
+  // ukernel's C function takes the outer K count as a runtime argument
+  // (`k_outer`) and runs the K loop itself; pre-unrolling it here would
+  // both defeat that and bloat the IR. This mirrors the equivalent split
+  // in `setDataTiledMmaInnerTiledLoweringConfig` on the GPU side.
   SmallVector<int64_t> vecTileSizes(numLoops, 1);
+  if (ukernelDescriptor) {
```
egebeysel, line 3094: could we also add a test case for this

egebeysel (changes requested):
Functionally lgtm, though a few comments about testing and comments being a bit too verbose :)

compiler/plugins/target/LLVMCPU/LLVMCPUTarget.cpp
bjacob: Applied thanks!

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
bjacob: Removed.

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
bjacob: kept this second RUN intentionally: it's the minimal check that the descriptor SelectLoweringStrategy sets is honored by the next pass (LowerBitcodeUKernels) in the same pipeline. The heavier end-to-end test lives in e2e_inner_tiled_pipeline.mlir (#24571). Trimmed the comments per your other notes — happy to split it out if you'd prefer.

compiler/plugins/target/LLVMCPU/builtins/ukernel/test/select_ukernel.mlir
bjacob: Trimmed.

compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp
bjacob: Added — the enabled case now also checks that the K/reduction dim is left untiled in the lowering_config (no vector_reduction), with the disabled case asserting the contrasting vector_reduction = [0, 0, 1]. That's exactly the branch here: the ukernel owns the K loop.

egebeysel (approved):
Thanks!

egebeysel:
Could you resolve the conflicts and merge when you have time, thanks!

PR #24571, opened 2026-06-04
[Codegen][CPU] End-to-end-through-codegen lit test for ukernel framework.

bjacob:
Closes the loop on the new C-bitcode ukernel framework with an `iree-opt` lit test that drives the full LLVMCPU codegen pipeline on an `inner_tiled` rooted dispatch wrapped in the standard `hal.executable.variant` + binding-subspan structure, and asserts the whole chain end-to-end:

  - Bitcode reaches the dispatch's `hal.executable.variant` as a first-class `objects(...)` attribute (the long-lived home).
  - `llvm.func @iree_uk_mma_…` is declared with `hal.import.bitcode = true`, the flag that tells `RewriteExternCallOpToDynamicImportCallOp` to skip the runtime-import-table indirection.
  - A direct `llvm.call @iree_uk_mma_…` replaces the original `inner_tiled` — no `__import_ordinal_iree_uk_mma_…` (which would mean we accidentally re-used the legacy runtime-resolved path), no leftover `inner_tiled`, no leftover `ukernel.generic`.

To make that chain work, two pieces had to land alongside the test:

  * `UKernelProviderAttr::createAndReplaceWithUkernelOp` now actually builds the `UKernelGenericOp` (returning `success`, not `nullopt`), setting `fn_def_attrs = {hal.import.bitcode = true}` on it. That flag propagates onto the `func.func` declaration that `LowerUKernelOpsToCalls` synthesizes, and through it onto the `llvm.func` that `ConvertToLLVM` produces. Without this, the "self-contained bitcode" property of the new framework would not hold — every external call would be routed through the runtime import table just like the legacy mmt4d ukernels.

  * Bitcode lookup-and-attach is factored into a public helper `IREE::CPU::attachUKernelBitcodeOnOp`, called both from `LLVMCPUSelectUKernels` (at kernel-config time, mirroring GPU's `ensureUKernelBitcodeAndFinalizeConfig`) and from the provider itself (as a backstop for the BYO / hand-written-IR paths). It attaches the bitcode both as a `hal.executable.objects` discardable attribute on the op (the GPU-aligned shape) *and* directly on the enclosing `hal.executable.variant`'s `objects` operand attribute (the durable home, sidestepping passes that strip discardable attrs through `ConvertToLLVM`).

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
+++ b/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
@@ -18,6 +18,8 @@
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/AsmState.h"
 #include "mlir/IR/Attributes.h"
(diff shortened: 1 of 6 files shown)


compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -1192,53 +1194,199 @@ getUKernelBitcode(MLIRContext *context, ArrayAttr sourceExecutableObjects,
...
+  }
+  return true;
+}
+
+// Returns the index, in the ACC operand's shape, of the innermost
+// CrossIntrinsic dimension (the N cross-intrinsic dim for our layouts), or
+// nullopt if it is dynamic / absent. Ported from the ROCM ukernel support
+// (`getCInnermostStaticCrossIntrinsicDim`), using the CPU DataTiledMMA
+// swizzle. The ukernel needs this dim's stride to address each unrolled
+// intrinsic's ACC fragment.
```
egebeysel: Could you add an example to the comment and get the part about "Ported from ..." out? thanks!

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -1192,53 +1194,199 @@ getUKernelBitcode(MLIRContext *context, ArrayAttr sourceExecutableObjects,
...
+  Codegen::TileSwizzle accSwizzle = getSwizzle(mma, /*operandIdx=*/2);
+  SmallVector<Codegen::TileSwizzle::Dim> swizzleDims;
+  for (const Codegen::TileSwizzle::ExpandShapeDimVectorType &group :
+       accSwizzle.expandShape()) {
+    swizzleDims.append(group.begin(), group.end());
+  }
+  applyPermutationToVector(swizzleDims, accSwizzle.permutation());
+  int rankDiff = outputType.getRank() - static_cast<int>(swizzleDims.size());
+  auto crossIntrinsic = Codegen::TileSwizzle::Dim::Kind::CrossIntrinsic;
+  for (size_t e = swizzleDims.size(), i = e - 1; i < e; --i) {
```
egebeysel:
```suggestion
  for (size_t i = swizzleDims.size(); i-- > 0;) {
```

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -1192,53 +1194,199 @@ getUKernelBitcode(MLIRContext *context, ArrayAttr sourceExecutableObjects,
...
   return std::nullopt;
 }
 
+// Rewrites an `inner_tiled` carrying a CPU `DataTiledMMAAttr` to a
+// `ukernel.generic`, threading the data-tiled-MMA scalar parameters as
+// operands so the ukernel can loop over arbitrary `intrinsics_{m,n,k}`:
+//   ins(lhs, rhs) outs(acc) (k_outer, intrinsics_m, intrinsics_n, intrinsics_k)
+// plus a `strided_dims` entry giving the ACC's innermost cross-intrinsic
+// stride. Mirrors ROCM's `handleInnerTiledMmaUkernel` (minus the
+// shared-memory / subgroup machinery, which CPU does not have).
```
egebeysel:
```suggestion
// stride.
```

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
```diff
@@ -1192,53 +1194,199 @@ getUKernelBitcode(MLIRContext *context, ArrayAttr sourceExecutableObjects,
...
+  // Outer-K tile count: the LHS operand's outer dims are (m_outer, k_outer),
+  // so dim 1 is the K-tile count the ukernel loops over.
+  Value kOuter = arith::IndexCastOp::create(
+      rewriter, loc, i32,
+      tensor::DimOp::create(rewriter, loc, op.getInputs()[0], 1));
+  // strided_dims: one entry per shaped operand (LHS, RHS, ACC). Only ACC gets
+  // a stride — the innermost cross-intrinsic (N) dim — so the ukernel can
+  // address each unrolled intrinsic's ACC fragment. LHS/RHS are passed as
+  // base+offset and the ukernel infers their strides from the contiguous
+  // data-tiled layout.
```
egebeysel: Question: I'm definitely missing something here, but, why does the op take a per-operand strided dim array here? Is it currently used somewhere? What _could_ be an example use-case for this?

egebeysel (approved):
LGTM, thanks!

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
bjacob: Done — dropped the "Ported from…" reference and added a concrete example (bf16 1x16x2 with intrinsics_m = intrinsics_n = 2 → a 2×2 grid of fragments, innermost cross-intrinsic dim is N).

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
bjacob: Applied your suggestion, thanks — much cleaner than the wraparound form.

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
bjacob: Trimmed to just "…cross-intrinsic stride." as suggested.

compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp
bjacob: Good question — nothing CPU-specific here. strided_dims is the generic ukernel.generic ABI override (see [UKernelOps.td](vscode-webview://01d2kf242mt560vjg0su84bub1t3f6t860q6em279nk2i2rddoev/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/UKernelOps.td#L42)): each memref operand lowers to base, offset, stride0, stride1, …, and by default all strides are passed. strided_dims is an optional per-operand list saying which dims' strides to pass — so it's naturally an array-of-arrays, one inner list per shaped operand. Here LHS/RHS are contiguous in the data-tiled layout so they get empty lists (base+offset only), and only ACC needs one stride — its innermost cross-intrinsic (N) dim — so the ukernel can address fragment (m, n) at acc + (m*intrinsics_n + n)*acc_stride. It is exercised: this ACC slot, plus the identical pattern in ROCM's [ROCMUkernelBitcodeSupport.cpp](vscode-webview://01d2kf242mt560vjg0su84bub1t3f6t860q6em279nk2i2rddoev/compiler/plugins/target/ROCM/Dialect/ROCM/IR/ROCMUkernelBitcodeSupport.cpp#L401). The generality covers any future ukernel with a strided operand (or an operand needing multiple strided dims). I expanded the in-code comment to explain this.

PR #24572, opened 2026-06-04
[Codegen][CPU] Fill in the bf16 and i8 ukernel bodies + e2e tests.

bjacob:
Replaces the bf16 and i8-VNNI seeds' stub bodies with real SIMD implementations, both generic over the `intrinsics_{m,n,k}` unrolling factors and structured like the AMDGPU C ukernel they were adapted from (`iree_uk_amdgpu_multi_mma_mfma_i32_16x16x32_i8`): accumulators in registers, an outer loop over the K tiles (`k_outer`), and inside it the `(intrinsics_m, intrinsics_n, intrinsics_k)` unroll. The `intrinsics_*` arrive as constants at the inlined call site, so the loops fully unroll and the `acc_regs` arrays become fixed register files -- the bitcode-LTO equivalent of a C++ template, as the README describes.

  - bf16 (`MMA_X86_AVX512BF16_1x16x2_F32_BF16`): one `_mm512_dpbf16_ps` per (m, n, k), with the LHS K-pair broadcast via `set1_ps`.
  - i8 (`MMA_X86_AVX512VNNI_16x16x2_I32_I8_CASTI16`): the 16x16x2 tile is bit-compatible with the codegen path `lowerX86Avx512Vnni16x16x2I8` -- one `vpmovsxbw` widen of each i8 panel to i16, the `vpshufd` / `vbroadcasti32x4` fan-out, and 16 `vpdpwssd` over the block-interleaved (rlo, chi, rhi, clo) ACC layout. The i8 ukernel needs `-mavx512bw` for the widen, so it is added to the VNNI copts.

`LLVMCPUSelectUKernels` now only selects a ukernel when its bitcode actually exists (via `attachUKernelBitcodeOnOp`'s bool return), so an `MMAIntrinsic` the cost model picks but for which no seed exists -- e.g. the M<->N-swapped `MMA_X86_AVX512BF16_16x1x2_F32_BF16` -- falls back to codegen instead of dangling an undefined symbol.

Adds two execution/numerical tests, the first of the new C-bitcode ukernel path: `e2e_matmul_cpu_dt_inner_tiled_llvm_ukernel_bf16_f32` (avx512bf16) and `..._i8_i32` (avx512vnni). Each compiles a data-tiled matmul with `--iree-llvmcpu-enable-llvm-ukernels=inner_tiled`, links the ukernel bitcode, runs on host and checks results against a reference -- exercising the operand threading and generic `intrinsics_{m,n,k}` unrolling that the IR-level lit tests cannot. Both were confirmed to actually select their ukernel (not silently fall back to codegen).

### DIFF
--- a/compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
+++ b/compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
@@ -13,39 +13,97 @@
 // intrinsic name verbatim (lowercased, with the `iree_uk_` prefix), in line
 // with the AMDGPU C ukernel convention.
 //
-// The "inner K loop" the ukernel owns is the loop over the K *tiles* that
-// sits *inside* the outer M/N loops; those outer M/N loops are tiled away by
-// ordinary IREE tiling before this ukernel runs. The ukernel handles
-// arbitrary positive `intrinsics_{m,n,k}` (passed as arguments and looped
-// over); the loops fully unroll after the ukernel is inlined into its
-// constant-`intrinsics_*` caller -- the bitcode-LTO equivalent of a C++
-// template.
+// Adapted from the AMDGPU C ukernel
+// `iree_uk_amdgpu_multi_mma_mfma_i32_16x16x32_i8`: same shape — accumulators
+// held in registers, an outer loop over the K dimension, and inside it the
+// `(intrinsics_m, intrinsics_n, intrinsics_k)` unrolling — minus the
+// GPU-specific shared-memory / subgroup machinery.
+//
+// The "inner K loop" the ukernel owns is the loop over the K *tiles*
+// (`k_outer` below) that sits *inside* the outer M/N loops; those outer M/N
+// loops are tiled away by ordinary IREE tiling before this ukernel runs.
+// This is NOT a restriction to `intrinsics_{m,n,k} = 1`: the ukernel handles
+// arbitrary positive `intrinsics_{m,n,k}` via the `for` loops below.
+//
+// `intrinsics_{m,n,k}` are passed as function arguments and so look like
+// runtime values inside this translation unit, but the ukernel is always
+// inlined into its caller (a bug otherwise) and the caller always passes the
+// matching `DataTiledMMAAttr` constants. Together with post-inline IR
+// optimization on the linked bitcode, the `for` loops fully unroll and the
+// `acc_regs` VLA becomes a fixed register array specialized to each call
+// site's `intrinsics_{m,n,k}` — the bitcode-LTO equivalent of C++ templates.
 //
 // ABI: each shaped operand is passed as (base pointer, element offset) so the
-// caller doesn't need a GEP before the call. No strides are passed: the ACC
-// tile is contiguous, so the ukernel addresses each intrinsic's fragment from
-// `intrinsics_{m,n}` and its own fixed fragment size.
+// caller doesn't need a GEP before the call (the offset is added here). No
+// strides are passed: the ACC tile is contiguous, so the ukernel addresses
+// each fragment from `intrinsics_{m,n}` and its fixed fragment size. Offsets
+// are in units of the operand element type (bf16 for LHS/RHS, f32 for ACC).
 //
-// NOTE (seed scaffolding): this initial seed has a stub body. It exists so
-// that the surrounding *framework* -- bitcode build, embedding,
-// `hal.executable_object` injection, IR rewrite to `ukernel.generic` -- can
-// be landed and lit-tested. A follow-up commit replaces the body with the
-// `_mm512_dpbf16_ps`-based inner loop and adds an e2e matmul test for it.
+// Data-tiled operand layout (matching the `DataTiledMMAAttr` swizzle, same as
+// the AMDGPU ukernel):
+//   - ACC: one `__m512` (= M0=1 x N0=16 f32) per (m, n) intrinsic, tightly
+//     packed row-major over the (m, n) grid, so fragment (m, n) is at
+//     `acc + (m * intrinsics_n + n) * 16`.
+//   - LHS: per outer-K step, `intrinsics_m * intrinsics_k` units of 2 bf16
+//     (= one M0=1 x K0=2 fragment = a 4-byte `vdpbf16ps` m_bcst unit),
+//     ordered [m][k]; consecutive outer-K steps are contiguous.
+//   - RHS: per outer-K step, `intrinsics_n * intrinsics_k` panels of 32 bf16
+//     (= one N0=16 x K0=2 fragment = one `__m512`), ordered [n][k];
+//     consecutive outer-K steps are contiguous.
 IREE_UK_ALWAYS_INLINE
 void iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16(
     const uint16_t *lhs_base, int64_t lhs_offset, const uint16_t *rhs_base,
     int64_t rhs_offset, float *acc_base, int64_t acc_offset, int32_t k_outer,
     int32_t intrinsics_m, int32_t intrinsics_n, int32_t intrinsics_k) {
-  (void)lhs_base;
-  (void)lhs_offset;
-  (void)rhs_base;
-  (void)rhs_offset;
-  (void)acc_base;
-  (void)acc_offset;
-  (void)k_outer;
-  (void)intrinsics_m;
-  (void)intrinsics_n;
-  (void)intrinsics_k;
-  // TODO(ukernels): real inner K loop using `_mm512_dpbf16_ps`, looping over
-  // intrinsics_{m,n,k}.
+  const uint16_t *lhs = lhs_base + lhs_offset;
+  const float *rhs = (const float *)(rhs_base + rhs_offset);
+  float *acc = acc_base + acc_offset;
+
+  // The ACC tile is contiguous: each intrinsic's fragment is one __m512
+  // (M0*N0 = 1*16 f32), so fragment (m, n) sits
+  // `(m * intrinsics_n + n) * kAccFragElems` elements into `acc`.
+  enum { kAccFragElems = 16 };
+
+  // One accumulator register (1x16 f32) per (m, n) intrinsic. The VLA
+  // dimensions are compile-time constants at the inlined call site, so this
+  // lowers to a fixed register array.
+  __m512 acc_regs[intrinsics_m][intrinsics_n];
+  for (int32_t m = 0; m < intrinsics_m; ++m) {
+    for (int32_t n = 0; n < intrinsics_n; ++n) {
+      acc_regs[m][n] =
+          _mm512_loadu_ps(acc + (m * intrinsics_n + n) * kAccFragElems);
+    }
+  }
+
+  for (int32_t ko = 0; ko < k_outer; ++ko) {
+    const uint16_t *lhs_block =
+        lhs + (int64_t)ko * intrinsics_m * intrinsics_k * 2;
+    const float *rhs_block =
+        rhs + (int64_t)ko * intrinsics_n * intrinsics_k * 16;
+    for (int32_t m = 0; m < intrinsics_m; ++m) {
+      for (int32_t n = 0; n < intrinsics_n; ++n) {
+        for (int32_t k = 0; k < intrinsics_k; ++k) {
+          // LHS fragment: 2 bf16 (one M-row's K-pair) broadcast across the
+          // 16 SIMD lanes via `set1_ps` (the splat shape `vdpbf16ps`'s
+          // m_bcst variant pattern-matches). The bitcast to `__m512bh` is a
+          // width-preserving no-op LLVM elides.
+          __m512 lhs_bcast = _mm512_set1_ps(
+              *(const float *)(lhs_block + (m * intrinsics_k + k) * 2));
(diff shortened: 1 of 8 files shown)


compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSelectUKernels.cpp
```diff
@@ -74,7 +78,16 @@ IREE::Codegen::UKernelDescriptorAttr selectCPUUKernel(Operation *op) {
...
-  IREE::CPU::attachUKernelBitcodeOnOp(op, name);
+  //
+  // Only select the ukernel if its bitcode actually exists. Not every
+  // `MMAIntrinsic` the cost model picks has a built-in ukernel — e.g. the
+  // M<->N-swapped `MMA_X86_AVX512BF16_16x1x2_F32_BF16` orientation has no
+  // seed even though its natural sibling does — and matching one without
+  // bitcode would dangle an undefined symbol at link time. When absent, fall
+  // back to codegen by returning {}.
+  if (!IREE::CPU::attachUKernelBitcodeOnOp(op, name)) {
+    return {};
```
egebeysel, line 89: correct me if I'm wrong, but there's no negative test for this currently, right?

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
```diff
@@ -13,40 +13,94 @@
...
-// ordinary IREE tiling before this ukernel runs. The ukernel handles
-// arbitrary positive `intrinsics_{m,n,k}` (passed as arguments and looped
-// over); the loops fully unroll after the ukernel is inlined into its
-// constant-`intrinsics_*` caller -- the bitcode-LTO equivalent of a C++
-// template.
+// Adapted from the AMDGPU C ukernel
+// `iree_uk_amdgpu_multi_mma_mfma_i32_16x16x32_i8`: same shape — accumulators
+// held in registers, an outer loop over the K dimension, and inside it the
+// `(intrinsics_m, intrinsics_n, intrinsics_k)` unrolling — minus the
+// GPU-specific shared-memory / subgroup machinery.
```
egebeysel, line 20: could you rephrase this as a self-contained description of the CPU ukernel without mentioning the mfma one?

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
```diff
@@ -13,40 +13,94 @@
...
+// cross-intrinsic (N) dimension. Offsets/strides are in units of the operand
+// element type (bf16 for LHS/RHS, f32 for ACC).
 //
-// NOTE (seed scaffolding): this initial seed has a stub body. It exists so
-// that the surrounding *framework* -- bitcode build, embedding,
-// `hal.executable_object` injection, IR rewrite to `ukernel.generic` -- can
-// be landed and lit-tested. A follow-up commit replaces the body with the
-// `_mm512_dpbf16_ps`-based inner loop and adds an e2e matmul test for it.
+// Data-tiled operand layout (matching the `DataTiledMMAAttr` swizzle, same as
+// the AMDGPU ukernel):
```
egebeysel, line 43:
```suggestion
// Data-tiled operand layout (matching the `DataTiledMMAAttr` swizzle):
```

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
```diff
@@ -13,40 +13,94 @@
...
+//   - ACC: one `__m512` (= M0=1 x N0=16 f32) per (m, n) intrinsic. The (m, n)
+//     grid is row-major with `acc_stride` the element stride of the innermost
+//     cross-intrinsic dim (N), so fragment (m, n) is at
+//     `acc + (m * intrinsics_n + n) * acc_stride`.
```
egebeysel: I'm sorry, I maybe should've cleared this up in the previous PR as well:

Let's look at the example from the last PR comment https://github.com/iree-org/iree/blob/540008c4aa3f6a2c6e55a054a33f24bd85144d97/compiler/src/iree/compiler/Codegen/Dialect/CPU/IR/IREECPUAttrs.cpp#L1263-L1270:

On that regard, IIUC, we have the shape `outer_m x outer_n x intrinsics_m x intrinsics_n x 1 x 16 x f32` for the ACC, and with `intrinsics_m/n = 2` we essentially have `o_m x o_n x 2 x 2 x 1 x 16 x f32` (or `o_m x o_n x 2 x 2 x 16 x f32` if the 1 is collapsed), and the function returns `3`. The corresponding `acc_stride` is then the stride for that dimension, in this case `16 * 4 (for the F32)`. Is that a correct understanding of what this actually is?

If so, I'm curious to know why passing this is even necessary here, since the accumulators would also (ideally) be contiguous and we could technically infer this from the intrinsic itself, right? Could you maybe name a concrete example here that does not adhere to my mind model so that I could understand this better?

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
```diff
@@ -13,40 +13,94 @@
...
+// This is NOT a restriction to `intrinsics_{m,n,k} = 1`: the ukernel handles
+// arbitrary positive `intrinsics_{m,n,k}` via the `for` loops below.
+//
+// `intrinsics_{m,n,k}` are passed as function arguments and so look like
+// runtime values inside this translation unit, but the ukernel is always
+// inlined into its caller (a bug otherwise) and the caller always passes the
+// matching `DataTiledMMAAttr` constants. Together with post-inline IR
+// optimization on the linked bitcode, the `for` loops fully unroll and the
+// `acc_regs` VLA becomes a fixed register array specialized to each call
+// site's `intrinsics_{m,n,k}` — the bitcode-LTO equivalent of C++ templates.
```
egebeysel, line 34: nit: seem like these are not specific to this ukernel, and can probably reside somewhere more general, e.g. a blogpost, README, or design doc.

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512vnni_16x16x2_i32_i8_casti16.c
```diff
@@ -13,49 +13,113 @@
...
-// intrinsics_k) tile built from the 16x16x2 i8 VNNI intrinsic via AVX-512
-// VNNI `vpdpwssd`. The "CASTI16" in the MMA intrinsic name reflects that
-// the s8 inputs are zero/sign-extended into i16 lanes before being fed to
-// the 16-bit VNNI instruction; that cast is handled in the inner loop.
+// Structurally identical to `iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16`:
+// accumulators held in registers, an outer loop over the K *tiles*
+// (`k_outer`), and inside it the `(intrinsics_m, intrinsics_n, intrinsics_k)`
+// unroll. This is NOT a restriction to `intrinsics_{m,n,k} = 1`: the ukernel
+// handles arbitrary positive `intrinsics_{m,n,k}` via the `for` loops below,
+// which fully unroll once inlined into the constant-`intrinsics_*` caller.
```
egebeysel, line 21: ditto, a cpu-specific self-contained description would be appreciated

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512vnni_16x16x2_i32_i8_casti16.c
```diff
@@ -13,49 +13,113 @@
 // intrinsic name verbatim (lowercased, with the `iree_uk_` prefix), in line
 // with the AMDGPU C ukernel convention.
```
egebeysel, line 14: could also get rid of this, since the stub would not be there anymore. This info can then also reside on the design doc, README, or the blog post

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512vnni_16x16x2_i32_i8_casti16.c
```diff
@@ -13,49 +13,113 @@
...
+              _mm512_shuffle_i32x4(rhs_i16, rhs_i16, 0xFF),
+          };
+          // 16 `vpdpwssd` over the 4x4 (rlo, chi) grid; accumulator (rlo, chi)
+          // lives at flat offset (4*rlo + chi)*16.
+          for (int rlo = 0; rlo < 4; ++rlo) {
+            for (int chi = 0; chi < 4; ++chi) {
+              int idx = 4 * rlo + chi;
+              (*regs)[idx] = _mm512_dpwssd_epi32((*regs)[idx], lhs_dup[rlo],
+                                                 rhs_bcast[chi]);
+            }
```
egebeysel, line 115: Am I understanding this correctly that each `_mm512` acc register essentially holds non-contiguous elements in memory in the end:

So for `regs[9]`, we have `rlo = 9 / 4 = 2` and `chi = 9 % 4 = 1`, meaning that it holds the elements `[ acc[2, 4], acc[2, 5], acc[2, 6], acc[2, 7], acc[6, 4], ..., acc[14, 7] ]` in the end (where `acc` indexes into the overall `16x16` tile in a row-major manner)?

If that's the case, how come do we have contiguous stores in the end, I'm probably missing something here without visualizing this :(

egebeysel (commented):
Thanks, I mostly have nits regarding comments, and 2 questions mainly for my understanding. Otherwise LG!

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512vnni_16x16x2_i32_i8_casti16.c
egebeysel: Nevermind the question, the `getIntrinsicSwizzle` of the MMA intrinsic defines a `4x4x4x4` layout here anyways.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPUSelectUKernels.cpp
bjacob: Added — select_ukernel.mlir now has a bf16_inner_tiled_ukernel_no_bitcode case using the M↔N-swapped MMA_X86_AVX512BF16_16x1x2_F32_BF16 (maps to a name but has no bitcode seed): CHECK-NOT descriptor, CHAIN-NOT ukernel.generic

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
bjacob: You're right! Always were, all along. Sorry I failed to properly consider your idea earlier.

It's always the constant fragment size (verified: acc_stride folds to 16/256 in the lowered IR). The stride-genericity here was a GPU-ism that I failed to properly reconsider when porting to CPU. Dropped it entirely in #24652: all operands are now (base, offset), getAccInnermostCrossIntrinsicDim is gone, and that also dissolves the per-operand strided_dims question from #24571.

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
bjacob: Moved the general framework prose into the README and slimmed each .c header to a self-contained CPU description (no mfma references) — done in #24573 as you suggested.

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512vnni_16x16x2_i32_i8_casti16.c
bjacob: 👍 the getIntrinsicSwizzle 4×4×4×4 layout is what makes the stores contiguous.

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
egebeysel: Thanks!

compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
egebeysel: thanks!

egebeysel (approved):
Thanks, LGTM!

egebeysel:
@bjacob could you rebase this and merge when you have time? thanks!

PR #24588, opened 2026-06-08
Integrate LLVM to llvm/llvm-project@917117ceeebf

schuermans-roofline:
Drop getEmptyKey() and getTombstoneKey() for DenseMapInfo-related classes. The reason is that LLVM dropped those in:
1e3dc606df3a [DenseMap] Store occupancy in a packed used-bit array (#201281) 1f10f1ca8af3 [DenseMap] Replace tombstone deletion with TAOCP 6.4 Algorithm R (#200595)

Dropping this was also needed in stablehlo and torch-mlir. Those submodules are updated here as well.

In ROCM mlir_ukernels, the syntax of "rocdl.sched.barrier" had to be adapted to use "none" instead of "0".
Furthermore, it now prints the keyword without brackets.
681fc74ac47e [mlir][ROCDL] Move ROCDL intrinsic enum immargs to enums (#198875)
    
Ranked-tensor type-constraint description gained "non-token"
67aaa03dac58 [mlir][IR] Add builtin `TokenType` (#195640)
    
emitc.call_opaque prints args/template_args as properties <{...}>
8f58a4cab3a7 [mlir][emitc]: Add MemberCallOpaque op (#200057)
    
cmpi+select clamp idioms fold to arith.minsi/arith.maxsi
e26ef0c94341 [mlir][arith] Canonicalize cmpi select to min/max (#201737)


krzysz00 (approved):
lgtm here

PR #24589, opened 2026-06-08
Align `iree_async_posix_worker_t`.

### LINKED ISSUE #24578: Compiling to Linux ARM32: worker struct should span at least one cache line
### What happened?

I am attempting to compile iree runtime to target linux arm32. However, we met this problem:
```
2026-06-04T14:23:53.8519493Z /home/runner/work/iree.gd/iree.gd/thirdparty/iree/runtime/src/iree/async/platform/posix/worker.h:82:1: error: static assertion failed: "worker struct should span at least one cache line"
2026-06-04T14:23:53.8521062Z    82 | static_assert(sizeof(iree_async_posix_worker_t) >=
2026-06-04T14:23:53.8521933Z       | ^~~~~~~~~~~~~
```

My PR that attempting to compile iree runtime to target linux arm32: https://github.com/iree-gd/iree.gd/pull/120
Related to this PR: https://github.com/iree-org/iree/pull/23527

This is the full failed CI/CD log:
[9_build_linux_arm _ build (linux, arm32, release).txt](https://github.com/user-attachments/files/28667791/9_build_linux_arm._.build.linux.arm32.release.txt)

_No response_

Runtime

### Version information

v3.11.0

### Additional context

_No response_

RechieKho:
Fix #24578

The fix includes aligning `iree_async_posix_worker_t` to `iree_hardware_constructive_interference_size`.

Pertaining to the concern of possible improper offset via pointer casting mentioned in this [comment](https://github.com/iree-org/iree/issues/24578#issuecomment-4646834986) , it comes from line 179 in file `runtime\src\iree\async\platform\posix\proactor.c`:
```cpp
  // Set up pointers into trailing data.
  proactor->workers =
      (iree_async_posix_worker_t*)((uint8_t*)proactor + workers_offset); // <- This
  proactor->worker_count = worker_count;
```

The worker offset is calculated from line 148 in the same file:
```cpp
  IREE_RETURN_AND_END_ZONE_IF_ERROR(
      z0, IREE_STRUCT_LAYOUT(
              sizeof(iree_async_proactor_posix_t), &total_size,
              IREE_STRUCT_FIELD_ALIGNED(
                  worker_count, iree_async_posix_worker_t,
                  iree_hardware_destructive_interference_size, &workers_offset), // <- This
              IREE_STRUCT_FIELD_ALIGNED(
                  ready_pool_capacity, iree_async_posix_ready_op_t,
                  iree_hardware_destructive_interference_size,
                  &ready_entries_offset),
              IREE_STRUCT_FIELD(completion_pool_capacity,
                                iree_async_posix_completion_t,
                                &completion_entries_offset),
              IREE_STRUCT_FIELD(message_pool_capacity,
                                iree_async_message_pool_entry_t,
                                &message_entries_offset)))
```

From what I read, `IREE_STRUCT_FIELD_ALIGNED` considers the alignment. I supposed the offset calculation should be correct, especially with the alignment to be the same. 

Moreover, I've also saw some manual padding in the struct via `_padding`. I remove such padding as I suppose the alignment will pad for us.

### DIFF
--- a/runtime/src/iree/async/platform/posix/worker.h
+++ b/runtime/src/iree/async/platform/posix/worker.h
@@ -54,7 +54,7 @@ typedef enum iree_async_posix_worker_state_e {
 //
 // Alignment is added to ensure struct spans at least one cache line (64 bytes) to
 // minimize false sharing when workers are stored in a contiguous array.
-typedef struct iree_alignas(iree_hardware_constructive_interference_size) iree_async_posix_worker_t {
+typedef struct iree_alignas(iree_hardware_destructive_interference_size) iree_async_posix_worker_t {
   // Current state (atomic for cross-thread visibility).
   // Written by request_exit, read by worker loop.
   iree_atomic_int32_t state;
--- a/runtime/src/iree/async/platform/posix/worker.h
+++ b/runtime/src/iree/async/platform/posix/worker.h
@@ -51,7 +51,7 @@ typedef enum iree_async_posix_worker_state_e {
 // Workers share proactor->ready_notification for work-stealing: all workers
 // wait on the same notification, and enqueue_for_execution posts to it to
 // wake exactly one idle worker.
-typedef struct iree_async_posix_worker_t {
+typedef struct iree_alignas(iree_hardware_constructive_interference_size) iree_async_posix_worker_t {
   // Current state (atomic for cross-thread visibility).
   // Written by request_exit, read by worker loop.
   iree_atomic_int32_t state;
@@ -71,10 +71,6 @@ typedef struct iree_async_posix_worker_t {
 
   // Thread handle (owned, released in deinitialize).
   iree_thread_t* thread;
-
-  // Padding to ensure struct spans at least one cache line (64 bytes) to
-  // minimize false sharing when workers are stored in a contiguous array.
-  uint8_t _padding[24];
 } iree_async_posix_worker_t;
 
 // Verify worker struct spans at least one cache line to minimize false sharing
--- a/runtime/src/iree/async/platform/posix/worker.h
+++ b/runtime/src/iree/async/platform/posix/worker.h
@@ -52,9 +52,10 @@ typedef enum iree_async_posix_worker_state_e {
 // wait on the same notification, and enqueue_for_execution posts to it to
 // wake exactly one idle worker.
 //
-// Alignment is added to ensure struct spans at least one cache line (64 bytes) to
-// minimize false sharing when workers are stored in a contiguous array.
-typedef struct iree_alignas(iree_hardware_destructive_interference_size) iree_async_posix_worker_t {
+// Alignment is added to ensure struct spans at least one cache line (64 bytes)
+// to minimize false sharing when workers are stored in a contiguous array.
+typedef struct iree_alignas(iree_hardware_destructive_interference_size)
+    iree_async_posix_worker_t {
   // Current state (atomic for cross-thread visibility).
   // Written by request_exit, read by worker loop.
   iree_atomic_int32_t state;
--- a/runtime/src/iree/async/platform/posix/worker.h
+++ b/runtime/src/iree/async/platform/posix/worker.h
@@ -51,6 +51,9 @@ typedef enum iree_async_posix_worker_state_e {
 // Workers share proactor->ready_notification for work-stealing: all workers
 // wait on the same notification, and enqueue_for_execution posts to it to
 // wake exactly one idle worker.
+//
+// Alignment is added to ensure struct spans at least one cache line (64 bytes) to
+// minimize false sharing when workers are stored in a contiguous array.
 typedef struct iree_alignas(iree_hardware_constructive_interference_size) iree_async_posix_worker_t {
   // Current state (atomic for cross-thread visibility).
   // Written by request_exit, read by worker loop.
--- a/runtime/src/iree/async/platform/posix/worker.h
+++ b/runtime/src/iree/async/platform/posix/worker.h
@@ -77,12 +77,6 @@ typedef struct iree_alignas(iree_hardware_destructive_interference_size)
   iree_thread_t* thread;
 } iree_async_posix_worker_t;
 
-// Verify worker struct spans at least one cache line to minimize false sharing
-// when workers are stored in a contiguous array.
-static_assert(sizeof(iree_async_posix_worker_t) >=
-                  iree_hardware_constructive_interference_size,
-              "worker struct should span at least one cache line");
-
 // Initializes a worker and starts its thread.
 // |out_worker| must point to caller-allocated storage (typically in the
 // proactor's trailing data array).
--- a/runtime/src/iree/async/platform/io_uring/proactor.c
+++ b/runtime/src/iree/async/platform/io_uring/proactor.c
@@ -116,7 +116,10 @@ iree_status_t iree_async_proactor_create_io_uring(
   // Allocate the proactor structure with trailing data.
   iree_async_proactor_io_uring_t* proactor = NULL;
   IREE_RETURN_AND_END_ZONE_IF_ERROR(
-      z0, iree_allocator_malloc(allocator, total_size, (void**)&proactor));
+      z0,
+      iree_allocator_malloc_aligned(allocator, total_size,
+                                    iree_hardware_destructive_interference_size,
+                                    0, (void**)&proactor));
   memset(proactor, 0, total_size);
 
   // Initialize base fields.
--- a/runtime/src/iree/async/platform/io_uring/proactor.c
+++ b/runtime/src/iree/async/platform/io_uring/proactor.c
@@ -116,10 +116,7 @@ iree_status_t iree_async_proactor_create_io_uring(
   // Allocate the proactor structure with trailing data.
   iree_async_proactor_io_uring_t* proactor = NULL;
   IREE_RETURN_AND_END_ZONE_IF_ERROR(
-      z0,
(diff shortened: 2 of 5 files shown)


RechieKho:
I have considered your comments, @schuermans-roofline, and I would like to know your review as well.

runtime/src/iree/async/platform/posix/worker.h
```diff
@@ -71,10 +71,6 @@ typedef struct iree_async_posix_worker_t {
 
   // Thread handle (owned, released in deinitialize).
   iree_thread_t* thread;
-
-  // Padding to ensure struct spans at least one cache line (64 bytes) to
```
schuermans-roofline, line 75: I would like to keep the comment - or more precisely the part explaining the reason for the alignment. So please move this comment to just before line 54 and adapt "padding to..." to "forcing alignment ...".

schuermans-roofline:
> From what I read, IREE_STRUCT_FIELD_ALIGNED considers the alignment. I supposed the offset calculation should be correct, especially with the alignment to be the same.

I think you are right about the offset calculation. However, I see that the pointer of the memory allocation is obtained from `iree_allocator_malloc` in `runtime/src/iree/async/platform/posix/proactor.c:163`. This function does not guarantee alignment to destructive interference size. I think it should be changed to `iree_allocator_malloc_aligned` (with `min_alignment = iree_hardware_destructive_interference_size` and `offset = 0`.

> Moreover, I've also saw some manual padding in the struct via _padding. I remove such padding as I suppose the alignment will pad for us.

I think this will work.

schuermans-roofline:
Thank you for taking action to fix this issue currently preventing arm32 builds.

I'm seeing that the linting github action is failing: https://github.com/iree-org/iree/actions/runs/27149694773/job/80151491958?pr=24589

Can you please follow https://iree.dev/developers/general/contributing/#formatting-and-linting to get the changes formatted properly?

schuermans-roofline (commented):
Please keep the information why we align in a comment. I also recommend changing the proactor malloc call to an aligned one (see my other comment for details).

runtime/src/iree/async/platform/posix/worker.h
```diff
@@ -51,7 +51,7 @@ typedef enum iree_async_posix_worker_state_e {
 // Workers share proactor->ready_notification for work-stealing: all workers
 // wait on the same notification, and enqueue_for_execution posts to it to
 // wake exactly one idle worker.
-typedef struct iree_async_posix_worker_t {
+typedef struct iree_alignas(iree_hardware_constructive_interference_size) iree_async_posix_worker_t {
```
devtbi: we want to align to the destructive size

devtbi (changes requested):

runtime/src/iree/async/platform/posix/worker.h
RechieKho: Ouch, that is some serious typo. Good catch, I totally overlook it.

runtime/src/iree/async/platform/posix/worker.h
devtbi: we can delete this one ;)

RechieKho: This is, however, checking for constructive interference, while we are aligning for destructive interference... Do we really able to delete this static check? 🤔

RechieKho:
By the way, just something to mention. The calculation is using destructive interference size, while the static check uses constructive interference size. From what I searched: 

> In programming, you utilize `std::hardware_constructive_interference_size` to define the maximum size of continuous memory that can successfully leverage true sharing without causing cache misses or bloated memory.

But we are checking whether the worker structure to be bigger than the constructive interference size...? Perhaps it is supposed to check against destructive interference size? If yes then the deletion is justified as alignment will ensure the struct to always be at least bigger than destructive interference size.

schuermans-roofline:
> By the way, just something to mention. The calculation is using destructive interference size, while the static check uses constructive interference size.

Good point. I think the static_assert used the wrong constant. The comment above it mentions that it wants to avoid false sharing - which would be destructive interference. So one more reason to delete that check.

runtime/src/iree/async/platform/posix/worker.h
devtbi: yes, the check is redundant / I don't think it was correct in the first place. Aligning to destructive size addresses the intent of the static_assert already.

devtbi (approved):
looks good to me know. Thanks for contributing :)

schuermans-roofline (changes requested):
The changes in worker.h look good to me. However, in the current code, the memory allocation is done with normal IREE mallloc (which typically aligns to 16 bytes) in line 163 of `proactor.c`
```
iree_allocator_malloc(allocator, total_size, (void**)&proactor)
```
, the memory layout might look like this:
```
0   ...   16   ...   64   ...   80   ...   128   ...   144   ...   192   ...   208
           <proactor____________><worker_0______________><worker_1______________>
<cache_line_0________><cache_line_1_________><cache_line_2__________>
```
So cache line 2 is shared (destructively) between two workers.

If the  malloc is changed to an aligned one
```
iree_allocator_malloc_aligned(allocator, total_size,
                              iree_hardware_destructive_interference_size, 0,
                              (void**)&proactor)
```
, the memory layout will look like this:
```
0     ...     64     ...     128     ...     192     ...     256
               <proactor______><worker_0______><worker_1______>
<cache_line_0_><cache_line_1__><cache_line_2__>
```

No destructive interference will happen.

Can you please add this change in `proactor.c`?

RechieKho:
Good catch, I have swap all founded `proactor` memory allocation with aligned variant.

schuermans-roofline:
> Good catch, I have swap all founded `proactor` memory allocation with aligned variant.

Thank you for doing this.

However, it went a bit too far and not far enough at the same time. I should probably have explained a bit more precise and complete what I meant.

First, the aligned data structure `iree_async_posix_worker_t` is only used in `runtime/src/iree/async/platform/posix/proactor.c`. The other `proactor.c` files that you modified (io_uring, iocp, js) are using a different data structure, which is not aligned. (There is no `IREE_STRUCT_FIELD_ALIGNED` in the size computations before and there is not alignment attribute on the definition of the `iree_async_proactor_io_uring_t` structure for example.) Thus, the only file that should be changed to the aligned malloc is `runtime/src/iree/async/platform/posix/proactor.c`. Please undo the change to the other `proactor.c` files.

Second, if a pointer is allocated using `iree_allocator_malloc_aligned`, it needs to be freed using `iree_allocator_free_aligned`. So please change `runtime/src/iree/async/platform/posix/proactor.c:319` `iree_allocator_free(allocator, proactor);` to `iree_allocator_free_aligned(allocator, proactor);`

To explain why this is needed:

The aligned allocator uses the normal allocator to do a slightly bigger memory allocation and then returns a pointer to the first properly aligned memory address inside this block. Let's assume the normal allocator returns 16-byte aligned memory, we need 64-byte aligned memory and a pointer is 8 bytes. If we want to allocate 128 bytes aligned to 64 bytes, the normal memory allocator is asked to allocate 128 + 8 + 64 = 200 bytes. Let's assume the normal allocator gives us a block at address 16. Then the situation is this:
```
0 ... 16 ... 56 ...      64 ...               192 ... 216
       <memory block returned by normal allocator______>
              <value of P><aligned memory block>
       P                  A
 ```
The normal allocator returns the pointer P to the aligned allocator. The aligned allocator reserves space for storing P and then finds the next address in the block which is aligned to 64 bytes. This is the pointer A here. It then returns the aligned pointer A to the caller.
On freeing the alligned allocated block, the aligned allocator gets pointer A from its caller, reads the original pointer P from just before this address and calls the normal allocator to free P.

If calling the normal allocator to free A (current state of the PR), it sees a pointer, which it never allocated and things crash...

**Comment by RechieKho** [comment]
> However, it went a bit too far and not far enough at the same time. I should probably have explained a bit more precise and complete what I meant.

My bad, I was kind of tired and just `sed` it through 😢 , totally forgot that everything we did is inside posix platform, and totally forgot about the aligned variant for `iree_allocator_free` as well. I have updated the code. Hopefully it works now.

But just something to mention, I have tried running `ctest`. Now all test passed except `iree_c_embed_data_test.cc`:
```
D:\Documents\godot_projects\iree.gd\thirdparty\iree\build_tools\embed_data\iree_c_embed_data_test.cc(48): error: Expected equality of these values:
  15
  toc2->size
    Which is: 14

[  FAILED  ] Generator.TestContents (2 ms)
[----------] 1 test from Generator (2 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (6 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] Generator.TestContents
```

```cpp
  auto* toc2 = testembed2_create();
  ASSERT_EQ("file3.bin", std::string(toc2->name));
  EXPECT_EQ(R"(ᕕ( ᐛ )ᕗ)" NEWLINE, std::string(toc2->data));
#if defined(_MSC_VER)
  EXPECT_EQ(15, toc2->size);
#else
  EXPECT_EQ(14, toc2->size);
#endif
  ASSERT_EQ(0, *(toc2->data + toc2->size));
 ```
I am compiling this on windows. Not sure is it just my environment or not.

schuermans-roofline:
> But just something to mention, I have tried running `ctest`. Now all test passed except `iree_c_embed_data_test.cc`

To me, this looks unrelated to the posix worker data structure change.

schuermans-roofline (approved):
LGTM. Thanks for your work on this.

RechieKho:
Oh my god I forgot to change the method for malloc.

PR #24593, opened 2026-06-09
fix(CI): long-running emulated rv64 matmul integration tests

egebeysel:
The RISCV-64 e2e matmul integration tests without the ukernels are very long-running under QEMU emulation, and were causing timeouts in CI. This PR disables the generic (non-uk) vectorization path for those tests until we properly support codegen.

(Claude-generated summary)

## Summary

The RISC-V 64 e2e matmul integration tests that run **without ukernels**
(`--iree-opt-data-tiling --iree-llvmcpu-enable-ukernels=none`) are extremely
slow under QEMU and were timing out CI (60 s per-test limit). This PR adds
`noriscv` to all non-ukernel `dt`/`experimental_dt` matmul variants, since the
generic vectorization path for data-tiled matmul is not yet properly supported
on RISC-V — only the ukernel path is validated there.

**Why it's slow:** with data-tiling on but ukernels off, the packed `mmt4d` is
lowered through generic vectorization. For the wide `M0×N0` tile it selects
(`N0 = VLEN/8`, `M0 = 7`), the widening multiply-accumulate has no vector×scalar
form, so the inner loop rebuilds the LHS broadcast every K-step via a storm of
`vsetvli` reconfigurations and `vrgather`/`vslideup` permutes — e.g. for i8 at
VLEN=256, ~118 `vsetvli` + ~19 `vrgather` per dispatch, vs ~2 `vsetvli` and no
gathers for the ukernel. QEMU emulates every vector instruction
element-by-element, and `vsetvli`/`vrgather`/`vslide` are among the costliest to
emulate, so the full shape set runs for tens of seconds and trips the timeout.
`dt_i8_i32` and `experimental_dt_i8_i32` hit it first; `dt_f32_f32` was already
at ~52 s (one slow runner from flaking).

Failing CI run: https://github.com/iree-org/iree/actions/runs/27192246864/job/80276315195

**Emulation vs. real hardware:** the timeout is largely a QEMU artifact. On a
SpaceMiT X60 (VLEN=256, native), the heaviest shape (540×332×516) runs in
milliseconds — but the generic path is still 3.5–6× slower than the ukernel
path, confirming it exercises an unsupported/unoptimized path rather than just
an emulator quirk:

| variant (540×332×516) | 1 thread | 8 threads | generic ÷ ukernel |
|---|---|---|---|
| i8 generic  | 123 ms | 37.9 ms | **6.1×** |
| i8 ukernel  | 20.3 ms | 6.15 ms | — |
| f32 generic | 108 ms | 29.1 ms | **3.5×** |
| f32 ukernel | 30.9 ms | 9.08 ms | — |

(The gap is specific to the wide-`M0` tile; narrow-M shapes pick `M0=1`, emit
clean `vwmacc.vx`, and are ~1.3×.)

Assisted-by: Claude Code

### DIFF
--- a/tests/e2e/matmul/BUILD.bazel
+++ b/tests/e2e/matmul/BUILD.bazel
@@ -157,7 +157,15 @@ X86_64_AVX512_BF16 = X86_64_AVX512 + [
         # f16/bf16 trigger internal LLVM assertion errors on riscv and wasm.
         "noriscv",
         "nowasm",
-    ] if (lhs_rhs_type == "f16" or lhs_rhs_type == "bf16") else []),
+    ] if (lhs_rhs_type == "f16" or lhs_rhs_type == "bf16") else []) + ([
+        # TODO: The non-ukernel (generic-vectorized) data-tiling path is not
+        # yet properly supported on RISC-V; only the microkernel (_uk) variants
+        # are validated there. Without ukernels these fall back to a generic
+        # widening kernel that is far too slow under the RISC-V QEMU emulator
+        # and times out. Disable all non-ukernel data-tiling variants on
+        # RISC-V. Remove once proper RVV codegen for this path is tested.
+        "noriscv",
+    ] if (not use_uk and lhs_rhs_type not in ["f16", "bf16"]) else []),
     target_backends_and_drivers = [
         ("llvm-cpu", "local-task"),
     ],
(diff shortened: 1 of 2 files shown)


bjacob (approved):
LGTM, but if you might also consider the alternative, instead of disabling long-running tests on slow targets, of instead switching the --shapes to a set of shapes that avoids large matmul sizes. After all, matmul latency grows like N^3, so moderately smaller shapes allow to gain an order of magnitude of latency.

egebeysel:
@bjacob Thanks, makes sense!

bjacob (approved):

PR #24595, opened 2026-06-09
[SPIRV][GPU] Add Mali G720 and other Arm 5 Gen target details

AGindinson:
Based on the official `vulkaninfo`, define `arm5gen` as an IREE Vulkan target for devices such as Mali G720 Immortalis. Include all compute bitwidths into the declaration, based on the official support for `shaderInt64` capability - besides, initial experiments on a set of Torch and TFLite models have shown that Arm driver emulation for wide integers holds up against the current emulation in IREE.


egebeysel (approved):
lgtm

PR #24598, opened 2026-06-10
[SPIRV] Add target-feature cli option

devtbi:
Allow manually specifying `vulkan-target-features` to overwrite the target defaults in `VulkanSPIRVTarget.cpp` (e.g. for testing, newly supported features).

### DIFF
--- a/compiler/plugins/target/VulkanSPIRV/VulkanSPIRVTarget.cpp
+++ b/compiler/plugins/target/VulkanSPIRV/VulkanSPIRVTarget.cpp
@@ -102,6 +102,7 @@ struct VulkanSPIRVTargetOptions {
   // lowest common denominator to guarantee the generated SPIR-V is widely
   // accepted for now. Eventually we want to use a list for multi-targeting.
   std::string target = "vp_android_baseline_2022";
+  std::string targetFeatures;
   VulkanDispatchAbi dispatchAbi = VulkanDispatchAbi::Descriptors;
 
   void bindOptions(OptionsBinder &binder) {
@@ -118,6 +119,12 @@ struct VulkanSPIRVTargetOptions {
             "GPUs. See "
             "https://iree.dev/guides/deployment-configurations/gpu-vulkan/ for "
             "more details."));
+    binder.opt<std::string>(
+        "iree-vulkan-target-features", targetFeatures,
+        llvm::cl::desc(
+            "Vulkan target features (SPIR-V version, capabilities, "
+            "extensions, etc.) to use. If provided, replaces the default "
+            "features for the target."));
     binder.opt<VulkanDispatchAbi>(
         "iree-vulkan-dispatch-abi", dispatchAbi,
         llvm::cl::desc("Selects the Vulkan dispatch ABI emitted for generated "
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/VulkanSPIRV/test/hal_target_device_attributes.mlir
```diff
@@ -19,6 +19,16 @@
...
 // CHECK-ALL-SAME: #hal.executable.target<"vulkan-spirv", "vulkan-spirv-fb"
+//
+// Replace the default features
+// RUN: iree-compile --compile-to=preprocessing --iree-hal-target-device=vulkan \
+// RUN:              --iree-vulkan-target-features=spirv:v1.6,cap:Shader,cap:Int64,cap:PhysicalStorageBufferAddresses,ext:SPV_KHR_physical_storage_buffer %s \
+// RUN: | FileCheck %s --check-prefix=CHECK-FEATURES
+//
+// CHECK-FEATURES: #hal.executable.target<"vulkan-spirv", "vulkan-spirv-fb"
+// CHECK-FEATURES-SAME: features = "spirv:v1.6,cap:Shader,cap:Int64,cap:PhysicalStorageBufferAddresses,ext:SPV_KHR_physical_storage_buffer"
+// CHECK-FEATURES-NOT: spirv:v1.3
```
AGindinson, line 30: 1. Could you also add a test for the case where `iree-hal-target-device` is combined with an explilcit `iree-vulkan-target` setting?
2. In this case, I believe we still want the string option to override the features?

devtbi: yes and yes :D - done

AGindinson (approved):
Thanks!

compiler/plugins/target/VulkanSPIRV/VulkanSPIRVTarget.cpp
```diff
@@ -118,6 +119,12 @@ struct VulkanSPIRVTargetOptions {
             "GPUs. See "
             "https://iree.dev/guides/deployment-configurations/gpu-vulkan/ for "
             "more details."));
+    binder.opt<std::string>(
+        "iree-vulkan-target-features", targetFeatures,
+        llvm::cl::desc(
```
egebeysel, line 124: _**if**_ this is for debug-only, you can use `llvm::cl::Hidden`, e.g.:

```
static llvm::cl::opt<bool> clEnableBlockedMatmuls(
    "iree-codegen-block-dynamic-dimensions-of-contractions",
    llvm::cl::desc("developer flag to guard blocking dynamic dimensions of "
                   "contraction-like ops"),
    llvm::cl::Hidden, llvm::cl::init(true));
```

egebeysel (approved):
thanks

compiler/plugins/target/VulkanSPIRV/VulkanSPIRVTarget.cpp
devtbi: good hint :) think its okay to have it non-hidden... can potentially be useful for non-debug/dev cases - and there is not really much harm one can do with the flag

PR #24602, opened 2026-06-11
[Codegen][LLVMCPU] enable vectorization of scalable mmt4d without masks

egebeysel:
This PR makes sure that we can vectorize `(batch_)mmt4d` operations without masks (this is not necesary due to data-tiling having padding semantics).

Additionally, current vectorization path for mmt4d (or other contractions for that matter) vectorizes the operations in the linalg body to a `multi_reduction`, which for `BF16xBF16->FP32` or `INT8xINT8->INT32` has `exti/extf`semantics on the operands. This is not ideal since target ISAs might have specific instructions that could be used for these contractions. If these extensions do not exist, then the extension semantics are inserted by vector lowering passes anyways.

Tests for the assume flags reside upstream, but these tests also showcase this.

### DIFF
--- a/compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
+++ b/compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
@@ -22,6 +22,9 @@
 #include "mlir/Dialect/SCF/Utils/Utils.h"
 #include "mlir/Dialect/Vector/Transforms/LoweringPatterns.h"
 #include "mlir/Dialect/Vector/Transforms/VectorRewritePatterns.h"
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
 #define DEBUG_TYPE "iree-codegen-generic-vectorization"
@@ -166,6 +169,37 @@ static LogicalResult isWithinVectorSizeLimit(linalg::LinalgOp linalgOp,
   return success(maxFlatVecSize < maxVectorSize);
 }
 
+static void
+fillVectorizationOptions(IRRewriter &rewriter, MLIRContext *context,
+                         Operation *op, bool vectorizeToTransferGather,
+                         SmallVectorImpl<NamedAttribute> &linalgOptionsList) {
+  linalgOptionsList.push_back(
+      rewriter.getNamedAttr("vectorizeNDExtract", rewriter.getBoolAttr(true)));
+  if (vectorizeToTransferGather) {
+    linalgOptionsList.push_back(rewriter.getNamedAttr(
+        "vectorizeToTransferGather", rewriter.getBoolAttr(true)));
+  }
+  // Add operation-specific options.
+  // We can assume that the dynamic dims match vector sizes for
+  // operations in the data-tiled layout. Currently this is limited to
+  // mmt4d-like operations.
+  bool assumeDynamicDimsMatchVecSizes =
+      isa<linalg::Mmt4DOp, linalg::BatchMmt4DOp>(op);
+  linalgOptionsList.push_back(rewriter.getNamedAttr(
+      "assumeDynamicDimsMatchVecSizes",
+      rewriter.getBoolAttr(assumeDynamicDimsMatchVecSizes)));
+  // For mmt4d operations with equal input operand types, we can
+  // directly vectorize to contractions instead of multi_reduction.
+  bool createNamedContraction = false;
+  if (auto cOp = dyn_cast<linalg::ContractionOpInterface>(op)) {
+    createNamedContraction =
+        isa<linalg::Mmt4DOp, linalg::BatchMmt4DOp>(op) &&
+        (getElementTypeOrSelf(cOp.lhs()) == getElementTypeOrSelf(cOp.rhs()));
+  }
+  linalgOptionsList.push_back(rewriter.getNamedAttr(
+      "createNamedContraction", rewriter.getBoolAttr(createNamedContraction)));
+}
+
 class GenericVectorizationPass final
     : public impl::GenericVectorizationPassBase<GenericVectorizationPass> {
 public:
@@ -184,17 +218,6 @@ void GenericVectorizationPass::runOnOperation() {
   mlir::FunctionOpInterface funcOp = getOperation();
 
   IRRewriter rewriter(context);
-
-  // Build DictionaryAttr options from pass options. These are forwarded to
-  // upstream linalg::vectorize().
-  SmallVector<NamedAttribute, 2> linalgOptionsList;
-  linalgOptionsList.push_back(
-      rewriter.getNamedAttr("vectorizeNDExtract", rewriter.getBoolAttr(true)));
-  if (vectorizeToTransferGather) {
-    linalgOptionsList.push_back(rewriter.getNamedAttr(
-        "vectorizeToTransferGather", rewriter.getBoolAttr(true)));
-  }
-  auto linalgOptions = DictionaryAttr::get(context, linalgOptionsList);
 
   SmallVector<VectorizableOpInterface> candidates;
   funcOp.walk([&](VectorizableOpInterface op) {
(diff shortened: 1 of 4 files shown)


compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
```diff
@@ -21,6 +21,38 @@ func.func @matmul(%lhs: tensor<3x4xf16>, %rhs: tensor<4x5xf16>, %acc: tensor<3x5
 
 // -----
 
+#config = #iree_cpu.lowering_config<distribution = [16, 16, 0, 0, 0, 0], vector_common_parallel = [1, 1, 0, 8, 8, 0], vector_reduction = [0, 0, 1, 0, 0, 1]>
+func.func @mmt4d_bf16_static_inner_tiles(%lhs: tensor<1x1x8x1xbf16>, %rhs: tensor<1x1x8x1xbf16>, %acc: tensor<1x1x8x8xf32>) -> tensor<1x1x8x8xf32> {
+  %0 = linalg.mmt4d {lowering_config = #config} ins(%lhs, %rhs : tensor<1x1x8x1xbf16>, tensor<1x1x8x1xbf16>) outs(%acc : tensor<1x1x8x8xf32>) -> tensor<1x1x8x8xf32>
+  return %0 : tensor<1x1x8x8xf32>
+}
+// CHECK-MASK-LABEL: func.func @mmt4d_bf16_static_inner_tiles(
+// CHECK-MASK:         %[[LHS_VEC:.+]] = vector.transfer_read {{.+}} : tensor<1x1x8x1xbf16>, vector<1x1x8x1xbf16>
```
banach-space, line 30: What is CHECK-MASK? ;-)

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
```diff
@@ -166,6 +169,37 @@ static LogicalResult isWithinVectorSizeLimit(linalg::LinalgOp linalgOp,
...
+  linalgOptionsList.push_back(rewriter.getNamedAttr(
+      "assumeDynamicDimsMatchVecSizes",
+      rewriter.getBoolAttr(assumeDynamicDimsMatchVecSizes)));
+  // For mmt4d operations with equal input operand types, we can
+  // directly vectorize to contractions instead of multi_reduction.
+  bool createNamedContraction = false;
+  if (auto cOp = dyn_cast<linalg::ContractionOpInterface>(op)) {
+    createNamedContraction =
+        isa<linalg::Mmt4DOp, linalg::BatchMmt4DOp>(op) &&
+        (getElementTypeOrSelf(cOp.lhs()) == getElementTypeOrSelf(cOp.rhs()));
```
banach-space, line 197: Shouldn't this be checking LHS and OUT instead? LHS and RHS almost always have identical types, no?

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
```diff
@@ -136,7 +136,8 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
     });
   }
 
-  if (isX86(targetConfig) || isRISCV(targetConfig)) {
+  if (isX86(targetConfig) ||
+      (isRISCV(targetConfig) && !isScalableVectorizationEnabled())) {
```
banach-space, line 140: Can you remind me what this logic does?

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
```diff
@@ -21,6 +21,38 @@ func.func @matmul(%lhs: tensor<3x4xf16>, %rhs: tensor<4x5xf16>, %acc: tensor<3x5
...
+// CHECK-MASK:         %[[RHS_VEC:.+]] = vector.transfer_read {{.+}} : tensor<1x1x8x1xbf16>, vector<1x1x8x1xbf16>
+// CHECK-MASK:         %[[OUT_VEC:.+]] = vector.transfer_read {{.+}} : tensor<1x1x8x8xf32>, vector<1x1x8x8xf32>
+// Named contractions do not get any extf in-between, generic lowering of mmt4d gets them.
```
banach-space, line 33: What qualifies as a named contraction? `linalg.mmt4d` and `linalg.matmul`?

banach-space (commented):
Thanks!

Based on the summary, tests and code, this looks like two independent changes, no?

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
egebeysel: I'm really sorry about this, I rebased the PR and had to move this from a unified `generic_vectorization.mlir` test file to this file, and forgot to adjust the check lines, and it silently succeeded :)(:

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
egebeysel: the key is almost always :) it simply gates that. LHS and OUT not being the same type is the reason we're passing this flag to begin with, otherwise we get `arith.extf/si/ui` ops materializing and we cannot target bf16/i8mm-like instructions through vector.contract.

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
egebeysel: `vector.contract` -> it's a bit of a misnomer, but that's the name of the llvm flag

egebeysel:
I'll split this PR into two:
- enable unmasked lowering of mmt4d
- enable lowering directly to `vector.contract`

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
egebeysel: I want to use configured vector sizes that are directly inferred from the selected tile sizes instead of inferred from the IR on scalable RVV targets. NEON/SVE are already do this. I'll separate this into its own (3rd) PR and presumably add something in the lines of a pipeline test.

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
banach-space: Thanks for clarifying! However, this makes this comment confusing. Isn't the example above, `@matmul` also lowered to `vector.contract`? Yet over there we do get `arith.extf`.

Also, I would avoid the phrase "named contraction" - it could mean `linalg.contract` or "contraction" category from https://mlir.llvm.org/docs/Rationale/RationaleLinalgDialect/#interchangeability-of-formsa-nameformsa.

egebeysel: Alright, I'll rephrase it as "vectorizing mmt4d with the `createNamedContractions` flag pased should not generate extf/i ops" 👍 

So what happens is, if we do not pass that flag to the vectorizer, then it vectorizes by taking a look at the generic body of the operation, which in case has this extension semantics, and gets a `vector.multi_reduction` instead of a `vector.contract`. The flag just acts as a bridge between `linalg.(batch_)mmt4d` and `vector.contract` and preserves the semantics of the computation data type.

compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp
banach-space: OK, but then in the tests added to this PR, there are no examples where type(LHS) != type(RHS)

compiler/src/iree/compiler/Codegen/Common/test/generic_vectorization_unmasked.mlir
banach-space: OK, I didn't think of that "interesting" side-effect of using multi-reduction, thank you for reminding me!

PR #24605, opened 2026-06-12
[Codegen][GlobalOpt] Detach non-splat constant inits

chrsmcgrr:
A constant used directly as the `outs` (init/destination) of a `linalg.matmul` or convolution e.g. a fully-connected/conv bias imported from TFLite via TOSA survives into dispatch formation. It is then cloned into the workgroup on the written (output) side and bufferizes as a read-only global, which fails workgroup-distribution verification with the following error:

```
'linalg.generic' op write affecting operations on global resources are restricted to workgroup distributed contexts
'func.func' op failed on workgroup distribution verification
```

This kind of IR was seen in TFLite -> TOSA conversion flows and and the IR below is an example of that:

```mlir
util.func public @main(%a: tensor<1x1xf32>, %w: tensor<1x9xf32>) -> tensor<1x9xf32> {
  %bias = arith.constant dense<[[1.0, 1.0, 2.0, 2.0, 3.0, 3.0, 4.0, 4.0, 5.0]]> : tensor<1x9xf32>
  %0 = linalg.matmul ins(%a, %w : tensor<1x1xf32>, tensor<1x9xf32>)
                     outs(%bias : tensor<1x9xf32>) -> tensor<1x9xf32>
  util.return %0 : tensor<1x9xf32>
}
```

The chosen solution is to use is the `DetachElementwiseFromNamedOps` pattern which exists to keep non-writable inits out of dispatches. It only covers two of our three relevant cases. The pass will
1. Take a non-const init and detach it into a zero-fill plus a trailing add
2. Take a splat constant init and convert it to a fill.

Our use-case of a non-splat constant init would not be handled. Here we extend the elementwise detach pattern so a non-splat constant init is detached the same way a computed init is. A fresh zero-fill is created and the constant is added back as a read-only input. This form of IR can be then handled by codegen.

After the fix the reproducer compiles and the bias survives read-only:

```
%bias = arith.constant ... : tensor<1x9xf32>
%z  = linalg.fill ins(%c0 : f32) outs(%empty : tensor<1x9xf32>) -> tensor<1x9xf32>
%mm = linalg.matmul ins(%a, %w) outs(%z) -> tensor<1x9xf32>
%0  = linalg.generic ins(%mm, %bias) outs(%z) { ... }   // %mm + %bias
```

---

_Code Examples and Tests Assisted-by Claude_

### DIFF
--- a/compiler/src/iree/compiler/GlobalOptimization/DetachElementwiseFromNamedOps.cpp
+++ b/compiler/src/iree/compiler/GlobalOptimization/DetachElementwiseFromNamedOps.cpp
@@ -58,11 +58,27 @@ struct DetachElementwisePattern : OpInterfaceRewritePattern<linalg::LinalgOp> {
     }
     Value outputOperand = outputOperands.front()->get();
 
-    auto outsDefiningOp = outputOperand.getDefiningOp<linalg::LinalgOp>();
-    if (!outsDefiningOp || isa<linalg::FillOp>(outsDefiningOp.getOperation())) {
-      // If not linalg op, or is a fill op, do nothing.
+    // By definition constants are read-only however some conversion paths
+    // place the bias initial value as the init/output argument of DPS
+    // operations. This later gets converted to a writable buffer. Here we
+    // detach the constant from the init argument to avoid this.
+    // Splat constant are handled later by DetachSplatConstantOutsOperands.
+    Operation *outsDefiningOp = outputOperand.getDefiningOp();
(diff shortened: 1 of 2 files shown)


AGindinson (approved):
LGTM! I've only learned about the "2 flavors of the output operands" with this PR (from "High-Level Description of Linalg Ops" in the spec) and am frankly surprised by how we need to maintain both semantical flavors. But the fix in the existing pass makes perfect sense.

PR #24606, opened 2026-06-15
Integrate LLVM to llvm/llvm-project@66e5a889c807

schuermans-roofline:
Add a fix for a use-after-free from
e79e056ee982 [mlir][CSE] Remove the opsToErase container and immediately delete dead ops. (#203702) (will be upstreamed to LLVM in parallel)


schuermans-roofline:
The LLVM upstream PR with the fix is https://github.com/llvm/llvm-project/pull/203849

AGindinson (approved):
LGTM, thanks for noting the follow-up issue

PR #24608, opened 2026-06-16
[VMVX] Added an implementation of fpowi in vm for vmvx backend

Ha-Eyz:
We have found some models, which failed to compile to vmvx, due to a missing conversion for math.fpowi.

This PR includes an implementation for this op, as well as an addition of VM_ComplexArithmeticOp class, since fpowi has a float and int input. This case was not covered by the VM_BinaryArithmeticOp class. However, since we only found fpowi in models, which utilize a i64 const for the exponent, the op was implemented in such a way that it only accepts i64 as an exponent. Due to this, it was not necessary to implement a ComplexArithmeticOpConversion in [MathToVM/Patterns.cpp](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/VM/Conversion/MathToVM/Patterns.cpp).

I also expanded the tests in arithmetic_ops.mlir, to include a check for vm.fpowi.f32.

Signed-off-by: Hakan Eyilmez [eyilmez@roofline.ai](mailto:eyilmez@roofline.ai)


compiler/src/iree/compiler/Dialect/VM/IR/VMOps.td
```diff
@@ -2424,6 +2424,32 @@ class VM_TernaryArithmeticOp<Type type, string mnemonic, VM_OPC opcode,
   ];
 }
 
+class VM_ComplexArithmeticOp<Type ltype, Type rtype, string mnemonic, VM_OPC opcode,
```
Ha-Eyz: The only difference to VM_BinaryArithmeticOp, is that VM_ComplexArithmeticOp does not require a type match between lhs and rhs, which are different for fpowi.

runtime/src/iree/vm/ops.h
```diff
@@ -587,6 +587,49 @@ static inline float vm_log10_f32(float operand) { return log10f(operand); }
 static inline float vm_log1p_f32(float operand) { return log1pf(operand); }
 static inline float vm_log2_f32(float operand) { return log2f(operand); }
 static inline float vm_pow_f32(float b, float e) { return powf(b, e); }
+static inline float vm_fpowi_f32(float base, int64_t exp) { 
+  if (exp == 0) {
```
Ha-Eyz: It probably would be a better idea, to have a template of this functionality, which vm_fpowi_f32 and vm_fpowi_f64 could call. However I didn't know where to put this template.
<math.h> did not include a function for fpowi, which is why I copied the implementation of fpowi from [complex-powi.cpp](https://github.com/llvm/llvm-project/blob/c7c340b41e4fe424f0f5f68be2f2812ac6d17b41/flang-rt/lib/runtime/complex-powi.cpp#L21) from the llvm-project.

devtbi: looks like you can make it a preprocessor "template" #define DEFINE_FPOWI(TYPE) void fpowi_##TYPE ...

runtime/src/iree/vm/ops.h
```diff
@@ -587,6 +587,49 @@ static inline float vm_log10_f32(float operand) { return log10f(operand); }
 static inline float vm_log1p_f32(float operand) { return log1pf(operand); }
 static inline float vm_log2_f32(float operand) { return log2f(operand); }
 static inline float vm_pow_f32(float b, float e) { return powf(b, e); }
+static inline float vm_fpowi_f32(float base, int64_t exp) {
```
devtbi: - can you add some (unit) testing for fpowi, to ensure it stays correct
- add a comment with the source of the algorithm

runtime/src/iree/vm/bytecode/dispatch.c
```diff
@@ -2541,6 +2541,7 @@ static iree_status_t iree_vm_bytecode_dispatch(
       IREE_VM_ISA_DISPATCH_OP_EXT_F32_UNARY_F32(Log1pF32, vm_log1p_f32);
       IREE_VM_ISA_DISPATCH_OP_EXT_F32_UNARY_F32(Log2F32, vm_log2_f32);
       IREE_VM_ISA_DISPATCH_OP_EXT_F32_BINARY_F32(PowF32, vm_pow_f32);
+      IREE_VM_ISA_DISPATCH_OP_EXT_F32_BINARY_F32(FPowI32, vm_fpowi_f32);
```
devtbi: - IREE_VM_ISA_DISPATCH_OP_EXT_F32_BINARY_F32 assumes both ops have the same type, see e.g. IREE_VM_ISA_DISPATCH_OP_CORE_SHIFT_I32 for an example
- verifier.c & disassembler.c have the same issue
- missing testcase not exposing the error

compiler/src/iree/compiler/Dialect/VM/IR/VMOps.td
devtbi: suggesting a rename: this sounds like complex numbers, something like VM_MixedBinaryArithmeticOp is clearer.

runtime/src/iree/vm/ops.h
```diff
@@ -587,6 +587,49 @@ static inline float vm_log10_f32(float operand) { return log10f(operand); }
 static inline float vm_log1p_f32(float operand) { return log1pf(operand); }
 static inline float vm_log2_f32(float operand) { return log2f(operand); }
 static inline float vm_pow_f32(float b, float e) { return powf(b, e); }
+static inline float vm_fpowi_f32(float base, int64_t exp) { 
+  if (exp == 0) {
+    return 1.0;
```
devtbi: nit: use float literal, 1.0f

runtime/src/iree/vm/ops.h
```diff
@@ -587,6 +587,49 @@ static inline float vm_log10_f32(float operand) { return log10f(operand); }
...
+  }
+
+  bool isMin = (exp == 0x8000000000000000); // -2^63
+
+  if (isMin) {
+    exp = 0x7FFFFFFFFFFFFFFF; // 2^63 - 1
+  }
+
+  bool invertResult = exp < 0;
```
devtbi: incorrect compared to the original impl, invertResult{exp < 0}; is above isMin clamp

runtime/src/iree/vm/ops.h
devtbi: right location would be base/internal/math.h, with a potential to use intrinsics, like iree_math_count_leading_zeros_u64 if they are available

devtbi (changes requested):
thanks for the addition :)
small fixes required & testing

devtbi:
pre-commit check fails, please run the pre-commit hooks https://iree.dev/developers/general/contributing/#formatting-and-linting

runtime/src/iree/vm/ops.h
Ha-Eyz: Since I made a Macro, which defines the function once for floats and once for doubles, I instead used return (TYPE)1.0;

runtime/src/iree/vm/ops.h
Ha-Eyz: I made a preprocessor template in base/internal/math.h, but I didn't see a potential use for an intrinsic. I only saw a potential use for iree_math_count_trailing_zeros_u64, but it didn't appear advantageous to me, since we would still need to raise the base by the amount of trailing zeroes in the exponent. You can correct me if I am wrong, but I am under the assumption that raising to a power is slower than a few multiplications.

runtime/src/iree/vm/ops.h
Ha-Eyz: I added some tests for fpowi. However, it appears that checking if it behaves correctly for the edge case: exp = -2^63 is not feasible, as the resulting value is too large.

runtime/src/iree/vm/ops.h
devtbi: I meant using __builtin_powif, just like iree_math_count_leading_zeros_u64 uses builtins, not using iree_math_count_leading_zeros_u64

runtime/src/iree/base/internal/math_test.cc
```diff
@@ -1080,4 +1080,26 @@ TEST(F8E8M0FNUConversionTest, F8E8M0FNUToF32) {
...
+  EXPECT_EQ(1.0f, iree_math_float_powi_i64(12.0f, 0));
+  EXPECT_EQ(1.0f, iree_math_float_powi_i64(1.0f, 4));
+  EXPECT_EQ(27.0f, iree_math_float_powi_i64(3.0f, 3));
+  EXPECT_EQ(0.25f, iree_math_float_powi_i64(2.0f, -2));
+  // No testcase for if iree_math_float_powi_i64 correctly behaves for
+  // exp = -2 ^ 63, since the resulting value is too large.
+}
+
+TEST(FPowITests, iree_math_double_powi_i64) {
+  EXPECT_EQ((double)1.0, iree_math_float_powi_i64((double)12.0, 0));
```
devtbi, line 1097: shouldnt this be testing iree_math_double_...? Can you go through your PR again to check if you find any other issues

runtime/src/iree/base/internal/math_test.cc
```diff
@@ -1080,4 +1080,26 @@ TEST(F8E8M0FNUConversionTest, F8E8M0FNUToF32) {
...
+  EXPECT_EQ((double)1.0, iree_math_float_powi_i64((double)1.0, 4));
```
devtbi, line 1098: no need to cast to double

runtime/src/iree/base/internal/math_test.cc
```diff
@@ -1080,4 +1080,26 @@ TEST(F8E8M0FNUConversionTest, F8E8M0FNUToF32) {
...
+  EXPECT_EQ((double)27.0, iree_math_float_powi_i64((double)3.0, 3));
+  EXPECT_EQ((double)0.25, iree_math_float_powi_i64((double)2.0, -2));
```
devtbi, line 1100: equality comparison for floating-point types is generally unreliable due to rounding - maybe there are some guarantees here, but generally this should be EXPECT_FLOAT/DOUBLE_EQ

devtbi (changes requested):

runtime/src/iree/base/internal/math_test.cc
Ha-Eyz: Yes you are right, I don't know how I could have missed this. But somehow I did it correctly for the comment below ...
I went through the PR again, but didn't see anything as major as this

devtbi (approved):

devtbi:
LGTM now :)

PR #24610, opened 2026-06-19
Integrate LLVM to llvm/llvm-project@1e0899794dd4

schuermans-roofline:
Adapted a LLVMCPU and a LLVMGPU codegen test to not expect the previously incorrectly emitted "nuw" flag.
This has been fixed in LLVM in this commit:
[mlir][MemRefToLLVM] fix incorrect nuw on GEP/mul when lowering memref.load/store with negative strides (#204309)
    
Finding the commit that caused the change and understanding the effect has been assisted by Claude Code.
The changes to the tests have been manually checked. The strides are dynamic, so negative strides are possible and "nuw" should not be there.


AGindinson (approved):
LGTM

PR #24612, opened 2026-06-19
[NFC] Remove the llvm-external-projects/iree-dialects directory (#24466)

Alex-Wengg:
## Overview

Removes the `llvm-external-projects/iree-dialects` directory now that nothing in the compiler or runtime depends on it.

This is the final step of #24466. The production dependencies were retired in #24543 and predecessors:
- the softmax matcher was reimplemented natively in `GlobalOptimization/RaiseSpecialOps`,
- `ErrorCheckingTrackingListener` was moved into `Codegen/Common/`,
- `DropSchedulePass` was moved into `Codegen/Common/`,
- the callback-matcher transform ops (`transform.iree.register_match_callbacks` / `match_callback` / `take_first` / `emit_remark`) and their tests were dropped.

## Changes

- Delete `llvm-external-projects/iree-dialects/`.
- Drop the `iree_llvm_add_external_project(mlir-iree-dialects ...)` call from the top-level `CMakeLists.txt`.
- Remove the dead iree-dialects target conversion from `bazel_to_cmake`.
- Drop the `IREEDialects*` Python sources + include path from `compiler/bindings/python/CMakeLists.txt`, the `_ireeDialects` `CMakeExtension` from `compiler/setup.py`, and the `iree_dialects.rst` API doc.
- Remove the `iree-dialects-doc` build step from the website doc generator.
- Remove the "Test iree-dialects" CI steps and the `check-iree-dialects` targets from the ASan/UBSan scripts (and the now-unused `test_iree_dialects.sh`).

## Testing

- **Local (Windows + MSVC, Ninja, Release):** CMake configure is clean after the external-project removal only `stablehlo` remains as an external LLVM project and a full-tree grep finds zero dangling references. The full `iree-opt` target builds, links, and runs (`iree-opt --version` → LLVM 23.0.0git).


Alex-Wengg:
i don't think the fails are lated to my PR

AGindinson (approved):

PR #24622, opened 2026-06-23
[Flow] Fold static flow.tensor.empty shapes into dims

LekkalaSravya3:
This PR Adds a canonicalization pattern for `flow.tensor.empty` that folds constant dynamic dimension operands into the static result shape.

**Example :**

%0 = flow.tensor.empty : tensor<?x4xf32>{%c8}

**canonicalizes to:**

%0 = flow.tensor.empty : tensor<8x4xf32>
%cast = tensor.cast %0 : tensor<8x4xf32> to tensor<?x4xf32>

The cast preserves the original result type for existing users while allowing downstream canonicalization to use the refined static shape.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Flow/IR/FlowOpFolders.cpp
+++ b/compiler/src/iree/compiler/Dialect/Flow/IR/FlowOpFolders.cpp
@@ -743,9 +743,58 @@ void TensorAllocaOp::getCanonicalizationPatterns(RewritePatternSet &results,
 // flow.tensor.empty
 //===----------------------------------------------------------------------===//
 
+
+namespace {
+
+// Folds constant dynamic dimension operands of a flow.tensor.empty into the
+// static result shape, e.g. `flow.tensor.empty : tensor<?xi32>{%c4}` becomes
+// `flow.tensor.empty : tensor<4xi32>`. A cast is inserted to preserve the
+// original (more dynamic) result type for existing users; it folds away once
+// those users can accept the refined static type.
+struct FoldTensorEmptyConstantDims : public OpRewritePattern<TensorEmptyOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(TensorEmptyOp op,
+                                PatternRewriter &rewriter) const override {
+    auto resultType = cast<RankedTensorType>(op.getResult().getType());
+    ValueRange dynamicDims = op.getResultDims();
+
+    SmallVector<int64_t> newShape(resultType.getShape());
+    SmallVector<Value> newDynamicDims;
+    unsigned dimIdx = 0;
+    bool didFold = false;
+    for (unsigned i = 0, e = resultType.getRank(); i < e; ++i) { 
+      if (!resultType.isDynamicDim(i)) { 
+        continue;
+      }
+      Value dim = dynamicDims[dimIdx++];
+      std::optional<int64_t> constantDim = getConstantIntValue(dim); 
+      if (constantDim && *constantDim >= 0) {
+        newShape[i] = *constantDim;
+        didFold = true;
+      } else {
+        newDynamicDims.push_back(dim);
+      }
+    }
+    if (!didFold) {
+      return rewriter.notifyMatchFailure(op, "no constant dynamic dims");
+    }
+
+    auto newType = RankedTensorType::get(newShape, resultType.getElementType(),
+                                         resultType.getEncoding());  
+    Value newEmpty =
+        TensorEmptyOp::create(rewriter, op.getLoc(), newType, newDynamicDims);
+    // Preserve the original (more dynamic) result type for existing users via a
+    // cast; it folds away once those users accept the refined static type.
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Dialect/Flow/IR/FlowOpFolders.cpp
```diff
@@ -743,9 +743,58 @@ void TensorAllocaOp::getCanonicalizationPatterns(RewritePatternSet &results,
 // flow.tensor.empty
 //===----------------------------------------------------------------------===//
 
+
+namespace {
+
+// Folds constant dynamic dimension operands of a flow.tensor.empty into the
+// static result shape, e.g. `flow.tensor.empty : tensor<?xi32>{%c4}` becomes
+// `flow.tensor.empty : tensor<4xi32>`. A cast is inserted to preserve the
+// original (more dynamic) result type for existing users; it folds away once
```
AGindinson: Nit:
```suggestion
// original (dynamic) result type for existing users; it folds away once
```
IMO otherwise it's a weird read :D

compiler/src/iree/compiler/Dialect/Flow/IR/FlowOpFolders.cpp
```diff
@@ -743,9 +743,58 @@ void TensorAllocaOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+    }
+    if (!didFold) {
+      return rewriter.notifyMatchFailure(op, "no constant dynamic dims");
+    }
+
+    auto newType = RankedTensorType::get(newShape, resultType.getElementType(),
+                                         resultType.getEncoding());  
+    Value newEmpty =
+        TensorEmptyOp::create(rewriter, op.getLoc(), newType, newDynamicDims);
+    // Preserve the original (more dynamic) result type for existing users via a
```
AGindinson: ditto

compiler/src/iree/compiler/Dialect/Flow/IR/test/tensor_folding.mlir
```diff
@@ -593,11 +593,40 @@ util.func public @sliceConst2DZeroLength01() -> tensor<0x0xi32> {
...
+}
+
+
+// -----
+
+// Leaves non-constant dynamic dimensions untouched.
+
+// CHECK-LABEL: @emptyKeepDynamicDim
+// CHECK-SAME: (%[[DIM:.+]]: index)
+util.func public @emptyKeepDynamicDim(%dim: index) -> tensor<?x4xf32> {
```
AGindinson, line 638: Can you also add a "mixed" regression example, where one dimension would be constant-foldable and the other would stay dynamic?

compiler/src/iree/compiler/Dialect/Flow/IR/FlowOpFolders.cpp
```diff
@@ -743,9 +743,58 @@ void TensorAllocaOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+// static result shape, e.g. `flow.tensor.empty : tensor<?xi32>{%c4}` becomes
+// `flow.tensor.empty : tensor<4xi32>`. A cast is inserted to preserve the
+// original (more dynamic) result type for existing users; it folds away once
+// those users can accept the refined static type.
+struct FoldTensorEmptyConstantDims : public OpRewritePattern<TensorEmptyOp> {
+  using Base::Base;
+  LogicalResult matchAndRewrite(TensorEmptyOp op,
+                                PatternRewriter &rewriter) const override {
+    auto resultType = cast<RankedTensorType>(op.getResult().getType());
+    ValueRange dynamicDims = op.getResultDims();
```
AGindinson, line 758: Having looked up this API in Flow I can now see that we're basically storing the SSA values so that we can construct the new op with the correct list of the operands. That said, the method naming is very counterintuitive :D

Could you please add more verbose comments here and/or in the loop body about this, to aid understanding? Not suggesting you change the dialect API itself in the scope of the PR (though it's something we could and IMO should do for the future).

AGindinson (commented):
Thanks for working on this!

AGindinson:
The linter error seems to be a whitespace issue.

compiler/src/iree/compiler/Dialect/Flow/IR/FlowOpFolders.cpp
LekkalaSravya3: Modified , Thanks !

compiler/src/iree/compiler/Dialect/Flow/IR/test/tensor_folding.mlir
LekkalaSravya3: Added the Regression Example , Thanks !!

compiler/src/iree/compiler/Dialect/Flow/IR/FlowOpFolders.cpp
LekkalaSravya3: Added the comments ,Thanks for ur feedback !

AGindinson (approved):
LGTM, thanks

PR #24630, opened 2026-06-25
[Runtime] Include base API header from futex.h

cerisier:
`futex.h` uses base API types and macros provided through `iree/base/api.h`, including status, attributes, config, and time definitions.

This includes the public base API umbrella instead of including those base headers directly. `target_platform.h` remains explicit because `futex.h` directly depends on platform detection macros.

This makes bazel layering check fail, hence the fix.


AGindinson (approved):
LGTM, thanks

PR #24631, opened 2026-06-25
[Docs][Website] Add instructions of how to integrate newer LLVM

schuermans-roofline:


ScottTodd:
You may also be interested in https://github.com/iree-org/megabump

schuermans-roofline:
> You may also be interested in https://github.com/iree-org/megabump

Thanks for the pointer. I had tried that and I found it to be very slow/compute-intensive, because it validates each intermediate LLVM commit that touches "interesting" files. Furthermore, I got the advice on discord to not use it: https://discord.com/channels/689900678990135345/1507341777605234768/1507390665615474759

docs/website/docs/developers/integrating-llvm.md
```diff
@@ -0,0 +1,155 @@
...
+git fetch -pP upstream
+```
+
+Create the virtual environment for IREE work:
+
+```bash
+cd "$WORK_DIR"
+uv venv venv --python 3.11
+source venv/bin/activate
+uv pip install pip
```
AGindinson: Just confused: is this step necessary? Especially if it happens within an already-activated venv?

docs/website/docs/developers/integrating-llvm.md
```diff
@@ -0,0 +1,155 @@
...
+cd iree
+python -m pip install --upgrade pip
+python -m pip install -r runtime/bindings/python/iree/runtime/build_requirements.txt
+python -m pip install pyright
```
AGindinson: Could also be done with `uv pip install` before activating the venv I guess. If that works, that would make things slightly easier

docs/website/docs/developers/integrating-llvm.md
```diff
@@ -0,0 +1,155 @@
...
+      -DIREE_ENABLE_ASAN=ON \
+      -DIREE_BUILD_PYTHON_BINDINGS=ON \
+      -DPython3_EXECUTABLE="$WORK_DIR/venv/bin/python" \
+      -DIREE_ENABLE_ASSERTIONS=ON \
+      -DIREE_ENABLE_SPLIT_DWARF=ON \
+      -DIREE_ENABLE_THIN_ARCHIVES=ON \
+      -DCMAKE_C_COMPILER=clang \
+      -DCMAKE_CXX_COMPILER=clang++ \
+      -DIREE_ENABLE_LLD=ON \
+      -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
```
AGindinson, line 135: Would it make sense to just refer to the main build instructions, with just the recommendations like `set -o pipefail` on top?

docs/website/docs/developers/integrating-llvm.md
```diff
@@ -0,0 +1,155 @@
...
+cd "$WORK_DIR/iree"
+python configure_bazel.py
+bazel build -j $WORKERS //tools/...
+```
+
+If anything goes wrong, try to fix it:
+
+- Apply changes to the IREE checkout as needed.
+    - You may ask an AI tool to identify the issues from the compile logs and
+      point out the LLVM commits that caused them.
```
AGindinson: To encourage more thorough analysis for fresh contributors who are going through the process as a learning experience, we should precede that by a more involved debugging tip, e.g. "Look at the commit history for the LLVM-Project dependencies of the failing build component by running `git log -- <path>`"

docs/website/docs/developers/integrating-llvm.md
```diff
@@ -0,0 +1,155 @@
...
+cd iree
+git add third_party/llvm-project
+git commit -s -m "Integrate LLVM to llvm/llvm-project@$LLVM_HASH"
+git commit --amend  # add notes to body of commit message if any fixes / cherry-picks / reverts had been needed, mention LLVM commits that caused the need for adaptions
+git push origin integrates/llvm-$DATE:integrates/llvm-$DATE
+```
+
+Open a PR to IREE upstream `main`. Get it through CI. Get it merged.
+
+## Pushing changes to other forked submodules
```
AGindinson, line 218: I'd suggest appending some bonus context to the section:
1. different LLVM integrate cadences for those projects are the reason why we currently recommend patching the iree-org forks directly;
2. it's recommended to look up if the respective project has already integrated to a similar state of LLVM-Project, and if so, replacing with a clean state / cherry-picking their version of the fix instead;
3. if you have time and the adjustment is trivial, it is encouraged to help integrate LLVM into those projects after the fact and align the forks;
4. however, for efficiency of the IREE integrate process and to avoid the mess of upgrading the whole dependency graph, steps 2-3 shouldn't block the IREE integrate and can be done as a follow-up.

This way, we can encourage giving back to the projects we depend upon.

docs/website/docs/developers/integrating-llvm.md
```diff
@@ -0,0 +1,155 @@
...
+python configure_bazel.py
+bazel build -j $WORKERS //tools/...
+```
+
+If anything goes wrong, try to fix it:
+
+- Apply changes to the IREE checkout as needed.
+    - You may ask an AI tool to identify the issues from the compile logs and
+      point out the LLVM commits that caused them.
+- Cherry-pick and/or revert commits in the LLVM submodule.
```
AGindinson: This prompts an "After the integrate" subsection:
- in case you had to apply reverts, analyze the issue deeper and create a follow-up fix / raise an issue in the LLVM-Project
- in case you had to add custom fixes, try to upstream to LLVM-Project
- clean state is our general preference, so the goal is to have one of the next integrate branches drop the customization

AGindinson (commented):
I like the clear and sequential instruction. My suggestions would bring in more details, maybe to a point of compromising on the simplicity – let me know if you'd prefer separate page(s) to cover the philosophy of "clean state" and how to follow up after the integrate is done. We could then write this up as a follow-up change.

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: I had to do this step. Without it, the following pip install steps failed.

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: I wanted to stay close to the instructions in `getting-started.md`, which lists those commands. It's just that `python -m venv .venv` in our environment gave me a wrong python version, so I added the `uv` command to get the right python version. I did not know how to trade-off using simpler(?) `uv` commands versus staying close to the instruction in `getting-started.md`. I will try to find a working sequence of commands that are in line with your comments.

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: The settings here are not the same as in the main build instructions. The settings here are the ones in the validation script in the megabump repo. I suppose those settings enable a few more things (asan, python bindings) than default to validate a bit more thoroughly than a normal build.

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: I can add this as a suggestion for the prompt.

schuermans-roofline: Oh, I think I got this wrong. It is supposed to be meant as a guideline for humans... Do we really want to ask a human to look at 100+ commits in LLVM? How is a fresh contributor supposed to spot the commit causing a compilation or runtime error by looking at the commits? Maybe it is just me, but when I tried that, it did not work. I had to start debugging properly, like opening the failed part of the code, get an IDE with language server ready and start with "go to definition"...

AGindinson: That's why I suggested checking the history for specific paths of interest (e.g. specific MLIR header to find a signature change). But I can see how that's not a universal debugging tip. I was merely concerned that the website would give "use an LLM" as the only debugging tip :D We can leave it as is though, the document implies involvement and understanding well enough in conjunction with our AI tooling policy.

schuermans-roofline: I added a reference to the getting started instructions and explained whe the instructions here are different.

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: added

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: added

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: I changed it to use only uv commands.

docs/website/docs/developers/integrating-llvm.md
schuermans-roofline: I changed it to use only uv commands.

docs/website/docs/developers/integrating-llvm.md
```diff
@@ -0,0 +1,266 @@
...
+Serval approaches can be used to make the LLVM integration continue:
+- Update the code in the IREE repo to adapt to the changed type or function.
+  This is the most desired outcome, because it allows a clean LLVM integration.
+  This does not cause any extra work in the following LLVM integrations.
+- Cherry-pick a revert or a fixup from LLVM upstream. LLVM is moving fast. By
+  the time a problematic LLVM commit has been identified during a LLVM
+  integration into IREE, the issue might have been identified in LLVM upstream
+  already. If so, there might already be a revert of the faulty commit or a
+  fixup commit of the issue available upstream. If so, cherry-pick it.
+- Decide to revert the offending LLVM upstream commit in IREE's LLVM form in the
```
AGindinson:
```suggestion
- Decide to revert the offending LLVM upstream commit in IREE's LLVM fork in the
```

AGindinson (approved):
LGTM, thanks a lot for expanding this!

PR #24634, opened 2026-06-26
feat(tiling): forward InnerTileAlignment overloads

egebeysel:
Forward InnerTileAlignment overloads on IREE TilingInterface ops.

Upstream adds the optional InnerTileAlignment hint as same-name overloads of getTiledImplementation, generateResultTileValue,
getTiledImplementationFromOperandTiles and getIterationDomainTileFromOperandTiles (llvm/llvm-project#150185). IREE's in-tree TilingInterface implementers declare only the hint-less methods, which hides the inherited defaulted hint-bearing overload (C++ name hiding) and fails to compile once the interface forwards to it.

Add trivial forwarding hint-bearing overloads (the ops ignore the hint; only linalg.pack/unpack consult it) to the LinalgExt ops, the Codegen InnerTiledOp, and the tensor.concat external model.


egebeysel:
closing since the integrate cherry-picked this commit: 

https://github.com/iree-org/iree/pull/24666

PR #24641, opened 2026-06-27
[Metal] Fix indirect dispatch retaining stack garbage as a resource

Alex-Vasile:
iree_hal_metal_command_buffer_prepare_dispatch built a resources[2] array for direct/indirect dispatch but passed &executable (the address of the parameter) to iree_hal_resource_set_insert instead of the resources array. For indirect dispatch (resource_count==2) this read the stack word adjacent to the parameter and retained it as a resource, corrupting the resource set and crashing on command-buffer teardown (release dereferenced a NULL vtable).

Pass resources so both the executable and the indirect workgroup-count buffer are retained correctly. Fixes the DispatchIndirectParameters and QueueDispatchIndirectParameters CTS segfaults (ctest #1307, #1309).


schuermans-roofline (approved):
LGTM. Thanks.

PR #24642, opened 2026-06-27
[Metal] Carry export name separately from MSL entry point for name lookup

Alex-Vasile:
spirv-cross rewrites SPIR-V entry point names to be legal MSL/C identifiers when serializing Metal executables: '.' characters (codegen dispatch names like _dot_general_matmul_i32.i32.i32_...) become '_', and reserved '__' prefixes (stream builtins like __builtin_splat_i64) are dropped. The MetalSPIRV target stored only the rewritten MSL function name in PipelineDef.entry_point, and the runtime loaded pipeline->name from that same field -- so hal.executable.lookup.function, which references the original un-rewritten hal.executable.export symbol name, could never match for any export whose name spirv-cross rewrites. This broke every metal-spirv e2e test using such exports (dot/dot_general/convolution matmul kernels and compare/dynamic_slice/gather stream builtins).

Add a dedicated always-present name field to PipelineDef carrying the canonical hal.executable.export symbol name (populated from exportOp.getName() in the compiler), appended at the end of the table for backward-compatible vtable layout. The runtime now sources pipeline->name (used by lookup_function_by_name) from this field, keeping entry_point as the MTLFunction name for -[MTLLibrary newFunctionWithName:]. The loader falls back to entry_point when name is absent for compatibility.

Verified: dot_general (8/8), compare (15/15), and the full Metal CTS suite (command_buffer_tests/queue_tests/dispatch_tests/queue_dispatch_tests, 4/4) pass. Signed-off-by: Alex Vasile <48962821+Alex-Vasile@users.noreply.github.com>

### DIFF
--- a/compiler/plugins/target/MetalSPIRV/MetalSPIRVTarget.cpp
+++ b/compiler/plugins/target/MetalSPIRV/MetalSPIRVTarget.cpp
@@ -157,9 +157,11 @@ public:
                                ".spv", spvBinary);
     }
 
-    // The runtime use ordinals instead of names but Metal requires function
-    // names for constructing pipeline states. Get an ordered list of the entry
-    // point names.
+    // Collect the SPIR-V entry point function names. These are cross-compiled
+    // to MSL and used to construct Metal pipeline states; they are the MSL
+    // function names (which spirv-cross may rewrite to be MSL-legal) and are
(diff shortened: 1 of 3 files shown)


schuermans-roofline (approved):
LGTM. Thanks.

PR #24644, opened 2026-06-27
[Metal] Fix indirect dispatch offset for sub-allocated parameter buffers

Alex-Vasile:
iree_hal_metal_command_buffer_prepare_dispatch resolved the indirect-dispatch workgroup-count buffer offset as just config.workgroup_count_ref.offset, dropping iree_hal_buffer_byte_offset(buffer) -- the base offset of the parameter buffer within its backing allocation. This is inconsistent with the sibling descriptor path in the same function (which adds byte_offset) and with every other backend: Vulkan and amdgpu route all offsets through a shared resolver that adds byte_offset, and the local HAL resolves through iree_hal_buffer_map_range. When the indirect-parameter buffer is a sub-allocation with a non-zero base offset, the dispatch read the three workgroup-count uint32s from the wrong address and ran a wrong grid.

The bug was latent because the Metal allocator returns standalone root buffers (byte_offset==0) and the cross-backend CTS indirect-parameters tests allocate the parameter buffer directly at offset 0.

Add byte_offset so the source offset matches the descriptor path and the other backends:
    workgroups_offset = iree_hal_buffer_byte_offset(config.workgroup_count_ref.buffer) + config.workgroup_count_ref.offset;

Adds CTS regression test DispatchIndirectParametersTest.SubAllocatedParameterBuffer, which references the workgroup counts through an iree_hal_buffer_subspan at a non-zero base offset (placing deliberately-wrong counts at offset 0 so a base-offset drop reads a deterministic grid instead of garbage). This is the only test shape that catches the bug, and it exercises it on every backend that supports sub-allocated buffers (Metal, Vulkan, amdgpu).

### DIFF
--- a/runtime/src/iree/hal/cts/command_buffer/dispatch_indirect_parameters_test.cc
+++ b/runtime/src/iree/hal/cts/command_buffer/dispatch_indirect_parameters_test.cc
@@ -387,6 +387,85 @@ TEST_P(DispatchIndirectParametersTest, DynamicParametersFromDispatch) {
   SubmitAndCheck(command_buffer, output_buffer, parameter_buffer);
 }
 
+// Regression test: the workgroup-count parameters live at a non-zero
+// iree_hal_buffer_byte_offset within their backing allocation (referenced
+// through an iree_hal_buffer_subspan view), so the driver must resolve the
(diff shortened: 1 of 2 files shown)


schuermans-roofline (approved):
LGTM. Thanks.

Issue #24574, opened 2026-06-04
Tracking: LLVMCPU C-microkernel framework


bjacob:
**C-based microkernels (ukernels) on the
LLVMCPU backend**: plain C functions compiled to LLVM bitcode, embedded in
`iree-compile`, and copied into the IR as `hal.executable_object` attributes on
a dispatch's HAL executable variant — the same representation as the GPU C
ukernels under `compiler/plugins/target/ROCM/builtins/ukernel/`.

This is **not** the legacy CPU ukernel framework under
`runtime/src/iree/builtins/ukernel/`; the two coexist and this is purely
additive. Key differences: it lives in `compiler/` (llvm-cpu-only; the VMVX
consumer that justified `runtime/` is defunct), is built **only as bitcode**,
and exposes a **lower-level interface** — the inner K-loop of one data-tiled MMA
tile rather than a whole-matmul library.

## Motivation / scope / differences from the legacy CPU ukernels.

Framework  | Legacy ukernels | These new ukernels
--- | --- | ---
Location in IREE tree | `runtime/src/iree/builtins/ukernel/` | `compiler/plugins/target/LLVMCPU/builtins/ukernel/`
Usage assumption | Thought as *majority* use case in data-tiled matmuls on CPU because codegen was generally *not* able to lower mmt4d to intrinsics. | Thought as *minority* case in new (inner-tiled) data-tiled matmuls on CPU because codegen *is* able to lower inner_tiled to intrinsics.
Usage | LLVMCPU, VMVX and standalone tests. | LLVMCPU only.
Build | Compiled *twice*: as llvm bitcode and as binary code using the native toolchain (for VMVX and for standalone tests). | Compiled *once* as llvm bitcode.
MLIR representation | Ukernel implementation is an external reference, opaque to MLIR code. | Ukernel implementation is pulled into a HAL executable_object, making MLIR self-contained.
How to bring your own ukernels | Fork the IREE source tree. | Add your own ukernels as HAL executable_object in your source MLIR.
Call overhead and impact on design | Because of the dual (bitcode / native build) and multiple purposes (LLVMCPU/VMVX/tests), it was unclear to what extent the ukernel dispatch would turn out ot have overhead that needs to be amortized. Even when it became clear that in LLVMCPU the ukernels were inlined into callers, the design was already set in place. | Ukernels designed from scratch around being always bitcode, always inlined and specialized into caller, seamless, no overhead. 
[...]

PR #24652, opened 2026-06-30
[Codegen][CPU] Drop the ACC stride from the inner_tiled ukernel ABI.

bjacob:
New bottom of the LLVMCPU C-ukernel stack.

The ACC stride threaded to the C ukernels is always a compile-time constant
equal to one intrinsic's ACC fragment size (the data-tiled ACC tile is
contiguous), so it folds to a constant after inlining and carries no
information the intrinsic-specific ukernel doesn't already encode. This drops
it:

- `IREECPUAttrs.cpp`: remove `getAccInnermostCrossIntrinsicDim`; build the
  `ukernel.generic` with `num_strided_outer_dims=0`, so all shaped operands
  (LHS, RHS, ACC) lower to `(base, offset)` only.
- Both seed ukernel `.c` signatures lose the `acc_stride` parameter.

This addresses review feedback on the now-merged #24569 / #24571 (why pass a
per-operand `strided_dims` array, and why pass an ACC stride that's statically
known). It makes the three operands uniform and removes the
`getAccInnermostCrossIntrinsicDim` swizzle machinery.

Verified: the lowered `llvm.call` to the ukernel now has no stride argument,
and the ukernel lit tests (`lower_inner_tiled_to_bitcode_ukernel*`,
`e2e_inner_tiled_pipeline`, `select_ukernel`) pass.

### DIFF
--- a/compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
+++ b/compiler/plugins/target/LLVMCPU/builtins/ukernel/iree_uk_mma_x86_avx512bf16_1x16x2_f32_bf16.c
@@ -22,8 +22,9 @@
 // template.
 //
 // ABI: each shaped operand is passed as (base pointer, element offset) so the
-// caller doesn't need a GEP before the call; the accumulator additionally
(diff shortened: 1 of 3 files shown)


egebeysel (approved):
LGTM, thanks for taking another look at this :)

PR #24655, opened 2026-07-01
[Stream] Deduplicate identical tensor import/export ops

LekkalaSravya3:
- This PR Adds canonicalization patterns to deduplicate identical stream.tensor.import and stream.tensor.export operations

- Added test coverage for deduplication, non-deduplication, and intermediate-operation scenarios.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
@@ -21,6 +21,7 @@
 #include "mlir/IR/Matchers.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/IR/OpImplementation.h"
+#include "mlir/IR/OperationSupport.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/IR/Value.h"
@@ -1109,19 +1110,6 @@ OpFoldResult TensorImportOp::fold(FoldAdaptor operands) {
   return {};
 }
 
-static bool isEquivalent(TensorImportOp previous, TensorImportOp current) {
-  if (previous.getConsume()) {
-    return false;
-  }
-
-  return previous.getSource() == current.getSource() &&
-         previous.getResult().getType() == current.getResult().getType() &&
-         previous.getResultEncoding() == current.getResultEncoding() &&
-         previous.getResultEncodingDims() == current.getResultEncodingDims() &&
-         previous.getResultSize() == current.getResultSize() &&
-         previous.getAffinity() == current.getAffinity();
-}
-
 //===----------------------------------------------------------------------===//
 // stream.tensor.export
 //===----------------------------------------------------------------------===//
@@ -1140,19 +1128,12 @@ OpFoldResult TensorExportOp::fold(FoldAdaptor operands) {
   return {};
 }
 
-static bool isEquivalent(TensorExportOp previous, TensorExportOp current) {
-  return previous.getSource() == current.getSource() &&
-         previous.getResult().getType() == current.getResult().getType() &&
-         previous.getSourceEncoding() == current.getSourceEncoding() &&
-         previous.getSourceEncodingDims() == current.getSourceEncodingDims() &&
-         previous.getSourceSize() == current.getSourceSize() &&
-         previous.getAffinity() == current.getAffinity();
-}
-
 //===----------------------------------------------------------------------===//
-// Generic deduplication pattern for tream.tensor.import and
+// Generic deduplication pattern for stream.tensor.import and
 // stream.tensor.export
 //===----------------------------------------------------------------------===//
+
+namespace {
 
 template <typename OpTy>
 struct DeduplicateTensorOp : OpRewritePattern<OpTy> {
@@ -1160,20 +1141,16 @@ struct DeduplicateTensorOp : OpRewritePattern<OpTy> {
 
   LogicalResult matchAndRewrite(OpTy currentOp,
                                 PatternRewriter &rewriter) const override {
-    // Ownership-transferring imports must not be deduplicated.
-    if constexpr (std::is_same_v<OpTy, TensorImportOp>) {
-      if (currentOp.getConsume()) {
-        return failure();
-      }
-    }
-    // Search for an equivalent operation
+    // Search for an equivalent operation.
     for (Operation *previousOp = currentOp->getPrevNode();
          previousOp != nullptr; previousOp = previousOp->getPrevNode()) {
       auto previous = dyn_cast<OpTy>(previousOp);
       if (!previous) {
         continue;
       }
-      if (!isEquivalent(previous, currentOp)) {
+      if (!OperationEquivalence::isEquivalentTo(
+              previous.getOperation(), currentOp.getOperation(),
+              OperationEquivalence::IgnoreLocations)) {
         continue;
       }
       rewriter.replaceOp(currentOp, previous.getResult());
@@ -1183,6 +1160,8 @@ struct DeduplicateTensorOp : OpRewritePattern<OpTy> {
     return failure();
   }
 };
+
+} // namespace
 
 void TensorImportOp::getCanonicalizationPatterns(RewritePatternSet &results,
                                                  MLIRContext *context) {
--- a/compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
+++ b/compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
@@ -1109,9 +1109,17 @@ OpFoldResult TensorImportOp::fold(FoldAdaptor operands) {
   return {};
 }
 
-void TensorImportOp::getCanonicalizationPatterns(RewritePatternSet &results,
-                                                 MLIRContext *context) {
-  // TODO(benvanik): check operand and dedupe imports.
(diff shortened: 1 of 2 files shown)


compiler/src/iree/compiler/Dialect/Stream/IR/test/tensor_folding.mlir
```diff
@@ -277,3 +277,84 @@ util.func public @NofoldTensorEncodingOpWithUnknownResultEncoding(%arg0: !stream
...
+
+  // CHECK: %[[IMPORT:.+]] = stream.tensor.import %arg0
+  // CHECK: arith.addi
+  // CHECK-NOT: stream.tensor.import
+  // CHECK: util.return %[[IMPORT]],
+
+  %0 = stream.tensor.import %arg0 : !util.buffer -> tensor<?x5xf32>{%arg1} in !stream.resource<external>{%c20}
+  %1 = arith.addi %c0, %arg1 : index
+  %3 = stream.tensor.import %arg0 : !util.buffer -> tensor<?x5xf32>{%arg1} in !stream.resource<external>{%c20}
+  util.return %0, %1 : !stream.resource<external>, index
```
phemashekar: could you change this test to return both imports, i.e., %0, %3, %1 to check that both resource use the same SSA value after canonicalization.

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
```diff
@@ -1132,9 +1140,58 @@ OpFoldResult TensorExportOp::fold(FoldAdaptor operands) {
...
+  return previous.getSource() == current.getSource() &&
+         previous.getResult().getType() == current.getResult().getType() &&
+         previous.getSourceEncoding() == current.getSourceEncoding() &&
+         previous.getSourceEncodingDims() == current.getSourceEncodingDims() &&
+         previous.getSourceSize() == current.getSourceSize() &&
+         previous.getAffinity() == current.getAffinity();
+}
+
+//===----------------------------------------------------------------------===//
+// Generic deduplication pattern for tream.tensor.import and
```
phemashekar:
```suggestion
// Generic deduplication pattern for stream.tensor.import and
```

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
```diff
@@ -1132,9 +1140,58 @@ OpFoldResult TensorExportOp::fold(FoldAdaptor operands) {
...
+
+  LogicalResult matchAndRewrite(OpTy currentOp,
+                                PatternRewriter &rewriter) const override {
+    // Ownership-transferring imports must not be deduplicated.
+    if constexpr (std::is_same_v<OpTy, TensorImportOp>) {
+      if (currentOp.getConsume()) {
+        return failure();
+      }
+    }
+    // Search for an equivalent operation
```
phemashekar:
```suggestion
    // Search for an equivalent operation.
```

compiler/src/iree/compiler/Dialect/Stream/IR/test/tensor_folding.mlir
```diff
@@ -277,3 +277,84 @@ util.func public @NofoldTensorEncodingOpWithUnknownResultEncoding(%arg0: !stream
...
+  // CHECK: stream.tensor.import %arg0 : !util.buffer -> tensor<1x?x5xf32>{%arg1}
+  %0 = stream.tensor.import %arg0 : !util.buffer -> tensor<?x5xf32>{%arg1} in !stream.resource<external>{%c20}
+  %1 = stream.tensor.import %arg0 : !util.buffer -> tensor<1x?x5xf32>{%arg1} in !stream.resource<external>{%c20}
+  util.return %0, %1 : !stream.resource<external>, !stream.resource<external>
+}
+
+// -----
+
+// CHECK-LABEL: @NofoldTensorImportConsume
+// Consuming imports transfer ownership — must not be deduplicated
```
phemashekar:
```suggestion
// Consuming imports transfer ownership — must not be deduplicated.
```

compiler/src/iree/compiler/Dialect/Stream/IR/test/tensor_folding.mlir
```diff
@@ -277,3 +277,84 @@ util.func public @NofoldTensorEncodingOpWithUnknownResultEncoding(%arg0: !stream
...
+util.func private @NofoldTensorImportConsume(%arg0: !util.buffer, %arg1: index) -> (!stream.resource<external>, !stream.resource<external>) {
+  %c20 = arith.constant 20 : index
+  // CHECK-COUNT-2: stream.tensor.import consume %arg0
+  %0 = stream.tensor.import consume %arg0 : !util.buffer -> tensor<?x5xf32>{%arg1} in !stream.resource<external>{%c20}
+  %1 = stream.tensor.import consume %arg0 : !util.buffer -> tensor<?x5xf32>{%arg1} in !stream.resource<external>{%c20}
```
phemashekar, line 316: question: is two `consume` imports of the same buffer even valid IR? `consume` transfers ownership and ownership can only be released from the current owner once. The verifier currently doesn't reject it, but if it's invalid by intent, skipping dedup here protects IR that has no defined meaning anyway. 
Also noticed: since the op is `Pure`, `--cse` already merges this into one consume import, so I'm curious as to why you've made this stricter?

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
```diff
@@ -1132,9 +1140,58 @@ OpFoldResult TensorExportOp::fold(FoldAdaptor operands) {
...
+         previous.getAffinity() == current.getAffinity();
+}
+
+//===----------------------------------------------------------------------===//
+// Generic deduplication pattern for tream.tensor.import and
+// stream.tensor.export
+//===----------------------------------------------------------------------===//
+
+template <typename OpTy>
+struct DeduplicateTensorOp : OpRewritePattern<OpTy> {
```
phemashekar, line 1139: style-nit: this struct and the two helpers sit at file scope, while every other pattern in this file lives in namespace {}. For consistency, can you move all of them into anonymous namespace?

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
```diff
@@ -1109,9 +1109,17 @@ OpFoldResult TensorImportOp::fold(FoldAdaptor operands) {
   return {};
 }
 
-void TensorImportOp::getCanonicalizationPatterns(RewritePatternSet &results,
-                                                 MLIRContext *context) {
-  // TODO(benvanik): check operand and dedupe imports.
+static bool isEquivalent(TensorImportOp previous, TensorImportOp current) {
```
phemashekar: instead of these two separate helpers to check equivalence, you could consider something like `OperationEquivalence::isEquivalentTo(prev, current, OperationEquivalence::IgnoreLocations)` instead of the manual field comparison since it covers all attributes (including consume), operands by construction and fold this check into the struct below.

phemashekar (commented):
Thanks for the PR! Implementation looks good overall. Just a few nits and some comments :))

compiler/src/iree/compiler/Dialect/Stream/IR/test/tensor_folding.mlir
LekkalaSravya3: Modified the changes !!

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
LekkalaSravya3: Modified It , Thanks !!

compiler/src/iree/compiler/Dialect/Stream/IR/StreamOpFolders.cpp
LekkalaSravya3: Thanks for the suggestion !!

compiler/src/iree/compiler/Dialect/Stream/IR/test/tensor_folding.mlir
LekkalaSravya3: Thanks for the clarification. My initial reasoning was based on the ownership semantics of `consume`, assuming that multiple `consume` imports of the same buffer should not be folded. However, I noticed that stream.tensor.import is marked `Pure`, so `--cse` already canonicalizes this case by merging identical imports. Based on that, I've removed this restriction from the PR.

phemashekar: cool, makes sense!

phemashekar (approved):
Thanks for addressing the comments! LGTM :)

AGindinson (approved):
Thanks!

AGindinson:
@LekkalaSravya3 Thanks for quality contributions! Feel free to apply for triage access as per https://iree.dev/developers/general/contributing/#obtaining-commit-access to simplify future contributions (though it might take a while to process currently).

LekkalaSravya3:
> @LekkalaSravya3 Thanks for quality contributions! Feel free to apply for triage access as per https://iree.dev/developers/general/contributing/#obtaining-commit-access to simplify future contributions (though it might take a while to process currently).

@AGindinson Thanks for your guidance! I've sent the email to apply for triage access.

PR #24656, opened 2026-07-01
[LLVMCPU] Add SME lowering-strategy tests for f64 and unsupported i8 matmuls

FedericoBruzzone:
Hi Folks :D
 `getMatmulAArch64SMEVectorSizes` (within `KernelDispatch.cpp`) only had test coverage for the f32 matmul case.

This PR add two tests:
- `f64` which uses `[4]x[8]` SME tiles instead of the `[8]x[8]` used for f32.
    https://github.com/iree-org/iree/blob/50d0addfb49254e0b7b746959b360c79437281cc/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp#L1465-L1471
- `i8` which SME does not support
    https://github.com/iree-org/iree/blob/50d0addfb49254e0b7b746959b360c79437281cc/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp#L1488-L1502

_AI disclaimer_: I used `Claude:Sonnet5` to write the checks.

@AGindinson @egebeysel

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
@@ -1,5 +1,5 @@
 // RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' --iree-llvmcpu-enable-scalable-vectorization=true --split-input-file %s | FileCheck %s --check-prefixes=CHECK,WITH-SME
-// RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' --iree-llvmcpu-enable-scalable-vectorization=true --iree-llvmcpu-disable-arm-sme-tiling --split-input-file %s | FileCheck %s --check-prefixes=DISABLE-ARM-SME
+// RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' --iree-llvmcpu-enable-scalable-vectorization=true --iree-llvmcpu-disable-arm-sme-tiling --split-input-file %s | FileCheck %s --check-prefixes=CHECK,DISABLE-ARM-SME
 
 #executable_target_embedded_elf_arm_64_ = #hal.executable.target<"llvm-cpu", "embedded-elf-arm_64", {cpu_features = "+sve,+sme", data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", native_vector_size = 16 : index, target_triple = "aarch64-none-elf"}>
 func.func @transpose_f32(%2: tensor<32x32xf32>) -> tensor<32x32xf32> attributes {hal.executable.target = #executable_target_embedded_elf_arm_64_} {
@@ -10,8 +10,9 @@ func.func @transpose_f32(%2: tensor<32x32xf32>) -> tensor<32x32xf32> attributes 
   } -> tensor<32x32xf32>
   return %4 : tensor<32x32xf32>
 }
-//   CHECK: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [4, 32], vector_common_parallel = {{\[}}[4], [4]]>
-//   CHECK: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [4, 32], vector_common_parallel = {{\[}}[4], [4]]>
+//  DISABLE-ARM-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [32, 32], vector_common_parallel = [4, 4]>
+//   CHECK-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
 //       CHECK: func.func @transpose_f32(
 //  CHECK-SAME:     translation_info = #[[TRANSLATION]]
 //       CHECK: linalg.generic
@@ -28,8 +29,9 @@ func.func @transpose_output_indexing_map_f32(%2: tensor<32x32xf32>) -> tensor<32
   } -> tensor<32x32xf32>
   return %4 : tensor<32x32xf32>
 }
-//   CHECK: #[[CONFIG:.+]] =  #iree_cpu.lowering_config<distribution = [4, 32], vector_common_parallel = {{\[}}[4], [4]]>
-//   CHECK: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//   WITH-SME-DAG: #[[CONFIG:.+]] =  #iree_cpu.lowering_config<distribution = [4, 32], vector_common_parallel = {{\[}}[4], [4]]>
+//  DISABLE-ARM-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [32, 32], vector_common_parallel = [4, 4]>
+//   CHECK-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
 //       CHECK: func.func @transpose_output_indexing_map_f32(
 //  CHECK-SAME:     translation_info = #[[TRANSLATION]]
 //       CHECK: linalg.generic
@@ -46,8 +48,9 @@ func.func @transpose_f64(%2: tensor<32x32xf64>) -> tensor<32x32xf64> attributes 
    } -> tensor<32x32xf64>
    return %4 : tensor<32x32xf64>
 }
-//   CHECK: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [4, 32], vector_common_parallel = {{\[}}[2], [2]]>
-//   CHECK: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [4, 32], vector_common_parallel = {{\[}}[2], [2]]>
+//  DISABLE-ARM-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [32, 32], vector_common_parallel = [2, 2]>
+//   CHECK-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
 //       CHECK: func.func @transpose_f64(
 //  CHECK-SAME:     translation_info = #[[TRANSLATION]]
 //       CHECK: linalg.generic
@@ -97,18 +100,12 @@ func.func @matmul_tensors(%7: tensor<?x?xf32>, %8: tensor<?x?xf32>, %9: tensor<?
   return %10 : tensor<?x?xf32>
 }
 //  DISABLE-ARM-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = [8, [8], 0], vector_reduction = [0, 0, 4]>
-//  DISABLE-ARM-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
-//      DISABLE-ARM-SME: func.func @matmul_tensors(
-//  DISABLE-ARM-SME-SAME:     translation_info = #[[TRANSLATION]]
-//       DISABLE-ARM-SME: linalg.matmul
-//  DISABLE-ARM-SME-SAME:     lowering_config = #[[CONFIG]]
-
 //   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = {{\[}}[8], [8], 0], vector_reduction = [0, 0, 1]>
-//   WITH-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
-//       WITH-SME: func.func @matmul_tensors(
-//  WITH-SME-SAME:     translation_info = #[[TRANSLATION]]
-//       WITH-SME: linalg.matmul
-//  WITH-SME-SAME:     lowering_config = #[[CONFIG]]
+//   CHECK-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//       CHECK: func.func @matmul_tensors(
+//  CHECK-SAME:     translation_info = #[[TRANSLATION]]
+//       CHECK: linalg.matmul
+//  CHECK-SAME:     lowering_config = #[[CONFIG]]
 
 // -----
 
@@ -120,18 +117,12 @@ func.func @matmul_tensors_f64(%7: tensor<?x?xf64>, %8: tensor<?x?xf64>, %9: tens
   return %10 : tensor<?x?xf64>
 }
 //  DISABLE-ARM-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = [8, [4], 0], vector_reduction = [0, 0, 2]>
-//  DISABLE-ARM-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
-//      DISABLE-ARM-SME: func.func @matmul_tensors_f64(
-//  DISABLE-ARM-SME-SAME:     translation_info = #[[TRANSLATION]]
-//       DISABLE-ARM-SME: linalg.matmul
-//  DISABLE-ARM-SME-SAME:     lowering_config = #[[CONFIG]]
-
 //   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = {{\[}}[4], [8], 0], vector_reduction = [0, 0, 1]>
-//   WITH-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
-//       WITH-SME: func.func @matmul_tensors_f64(
-//  WITH-SME-SAME:     translation_info = #[[TRANSLATION]]
-//       WITH-SME: linalg.matmul
-//  WITH-SME-SAME:     lowering_config = #[[CONFIG]]
+//   CHECK-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//       CHECK: func.func @matmul_tensors_f64(
+//  CHECK-SAME:     translation_info = #[[TRANSLATION]]
+//       CHECK: linalg.matmul
+//  CHECK-SAME:     lowering_config = #[[CONFIG]]
 
 // -----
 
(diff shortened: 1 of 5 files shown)


FedericoBruzzone:
The CI failure doesn't seem related to this PR :D

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
```diff
@@ -96,6 +96,46 @@ func.func @matmul_tensors(%7: tensor<?x?xf32>, %8: tensor<?x?xf32>, %9: tensor<?
 
 // -----
 
+// f64 matmul uses different SME tile sizes than f32: [4]x[8] instead of [8]x[8]
```
banach-space: It's been a while, but I believe that the tile size for `f64` should be `[4]x[4]` rather than `[4]x[8]`.

Have you tried running this configuration e2e?

FedericoBruzzone: The existing e2e tests are passing. I used the following command:
```bash
ctest --test-dir ~/dev/iree/build -L "test-type=check-test" --output-on-failure
```
but I'm facing an ICE regarding the `+sme` (and co) flags when passed to `--iree-llvmcpu-target-cpu-features=...` of `iree-compile`. I'll open an issue ASAP, I',m giving it some thought.

FedericoBruzzone: @banach-space this is the issue: #24660 but I don't think it should block this PR.

The PR body shows that iree is currently mapping to `[4]x[8]`, am I missing something?

banach-space: IREE might contain incorrect tile sizes -  no tests implies that we haven't verified that.

FedericoBruzzone: Wow! That's interesting. 
So, here is what I propose: let's start by merging PR #24661 (after the necessary reviews).

Once that is merged, I'll need to open an issue regarding the `f64` lowering I discovered today.
This should be largely unrelated to the existing issues, though I'll need to investigate whether it's linked to the `[4]x[8]` size, as you suggested.

Thanks! 🫶

MacDue: As explained in the comment:
```
   // Tile for [4]x[8], this results in loading twice as much from matrix B 
   // than and will use all eight [2]x[2] 64-bit SME accumulators. 
   // The B dimension is larger as it is known to be contiguous. 
```
Using `[4]x[8]` uses all eight 64-bit tiles available. Using `[4]x[4]` would only use half the available accumulators/tiles.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
```diff
@@ -96,6 +96,46 @@ func.func @matmul_tensors(%7: tensor<?x?xf32>, %8: tensor<?x?xf32>, %9: tensor<?
...
+// (each SME tile only holds half as many f64 elements per row as f32 ones).
+#executable_target_embedded_elf_arm_64_ = #hal.executable.target<"llvm-cpu", "embedded-elf-arm_64", {cpu_features = "+sve,+sme", data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", native_vector_size = 16 : index, target_triple = "aarch64-none-elf"}>
+func.func @matmul_tensors_f64(%7: tensor<?x?xf64>, %8: tensor<?x?xf64>, %9: tensor<?x?xf64>) -> tensor<?x?xf64> attributes {hal.executable.target = #executable_target_embedded_elf_arm_64_} {
```
egebeysel, line 78: I honestly think that this test belongs to the sibling `select_aarch64_sme_lowering_strategy.mlir` file instead. Same goes for the other matmul test that adds the check prefixes WITH-SME and DISABLE-ARM-SME in the first place.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
```diff
@@ -96,6 +96,46 @@ func.func @matmul_tensors(%7: tensor<?x?xf32>, %8: tensor<?x?xf32>, %9: tensor<?
 
 // -----
 
+// f64 matmul uses different SME tile sizes than f32: [4]x[8] instead of [8]x[8]
+// (each SME tile only holds half as many f64 elements per row as f32 ones).
+#executable_target_embedded_elf_arm_64_ = #hal.executable.target<"llvm-cpu", "embedded-elf-arm_64", {cpu_features = "+sve,+sme", data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", native_vector_size = 16 : index, target_triple = "aarch64-none-elf"}>
```
egebeysel: Furthermore, I would actually be more interested in seeing what happens if you only have the `+sme` feature flag but not `+sve` and you disable SME tiling.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
```diff
@@ -96,6 +96,46 @@ func.func @matmul_tensors(%7: tensor<?x?xf32>, %8: tensor<?x?xf32>, %9: tensor<?
...
+//   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = {{\[}}[4], [8], 0], vector_reduction = [0, 0, 1]>
+//   WITH-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//       WITH-SME: func.func @matmul_tensors_f64(
+//  WITH-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       WITH-SME: linalg.matmul
+//  WITH-SME-SAME:     lowering_config = #[[CONFIG]]
+
+// -----
+
+// SME tiling only supports f32 and f64 matmuls, so an i8 matmul should fall
```
egebeysel: ditto, this definitely should be in the other test file

egebeysel (commented):
I would suggest moving the tests to the sibling SME test file, and checking with SME enabled/disabled, without having the `+sve` flag (potentially also add a new RUN line to see if forcing streaming mode changes anything). That would give you the answer you were looking for: "Can I target SSVE on a target that just has SME but not SVE". I suspect that it _might_ pick non-scalable NEON tile sizes for now. If that were to be the case, since you want to enable SSVE, you can create an issue, and link it as a TODO on the test you add there as well.

You can also keep an FP64 test here without the `+sme` flag and check that it picks the correct tiling config.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
FedericoBruzzone: Correct me if I'm wrong, but does this mean it's okay to keep these tests and the current implementation? (according to what @egebeysel says below)

MacDue: I'm just saying that I think the current tile sizes are correct.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
FedericoBruzzone: That's a super interesting question < 3

FedericoBruzzone:
> I would suggest moving the tests to the sibling SME test file, and checking with SME enabled/disabled, without having the `+sve` flag (potentially also add a new RUN line to see if forcing streaming mode changes anything). That would give you the answer you were looking for: "Can I target SSVE on a target that just has SME but not SVE". I suspect that it _might_ pick non-scalable NEON tile sizes for now. If that were to be the case, since you want to enable SSVE, you can create an issue, and link it as a TODO on the test you add there as well.
> 
> You can also keep an FP64 test here without the `+sme` flag and check that it picks the correct tiling config.

Thanks a lot for all these valuable feedbacks. I believe I have addressed what you suggested.
Furthermore, as you thought, when SME tiling is disabled here there is no scalable fallback available. 
I left a TODO, but I'd like to merge PR #24661 first and keep this one based on it.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
```diff
@@ -1,4 +1,5 @@
-// RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' --iree-llvmcpu-enable-scalable-vectorization=true --split-input-file %s | FileCheck %s
+// RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' --iree-llvmcpu-enable-scalable-vectorization=true --split-input-file %s | FileCheck %s --check-prefixes=CHECK,WITH-SME
+// RUN: iree-opt --pass-pipeline='builtin.module(iree-llvmcpu-select-lowering-strategy)' --iree-llvmcpu-enable-scalable-vectorization=true --iree-llvmcpu-disable-arm-sme-tiling --split-input-file %s | FileCheck %s --check-prefixes=DISABLE-ARM-SME
```
egebeysel: can also add the prefix `CHECK` here and re-use some check lines for the below test

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
```diff
@@ -87,3 +88,107 @@ func.func @transpose_unsupported_not_simple_transpose(%2: tensor<32x32xf32>) ->
...
+//  DISABLE-ARM-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       DISABLE-ARM-SME: linalg.matmul
+//  DISABLE-ARM-SME-SAME:     lowering_config = #[[CONFIG]]
+
+//   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = {{\[}}[8], [8], 0], vector_reduction = [0, 0, 1]>
+//   WITH-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//       WITH-SME: func.func @matmul_tensors(
+//  WITH-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       WITH-SME: linalg.matmul
+//  WITH-SME-SAME:     lowering_config = #[[CONFIG]]
```
egebeysel: can reuse all of these with the above one by prefixing these with `CHECK` and only differing on the configs

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
```diff
@@ -87,3 +88,107 @@ func.func @transpose_unsupported_not_simple_transpose(%2: tensor<32x32xf32>) ->
...
+//  DISABLE-ARM-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       DISABLE-ARM-SME: linalg.matmul
+//  DISABLE-ARM-SME-SAME:     lowering_config = #[[CONFIG]]
+
+//   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = {{\[}}[4], [8], 0], vector_reduction = [0, 0, 1]>
+//   WITH-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//       WITH-SME: func.func @matmul_tensors_f64(
+//  WITH-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       WITH-SME: linalg.matmul
+//  WITH-SME-SAME:     lowering_config = #[[CONFIG]]
```
egebeysel: ditto

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
```diff
@@ -87,3 +88,107 @@ func.func @transpose_unsupported_not_simple_transpose(%2: tensor<32x32xf32>) ->
...
+//  DISABLE-ARM-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       DISABLE-ARM-SME: linalg.matmul
+//  DISABLE-ARM-SME-SAME:     lowering_config = #[[CONFIG]]
+
+//   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<distribution = [64, 64, 0], vector_common_parallel = [8, [8], 0], vector_reduction = [0, 0, 4]>
+//   WITH-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>>
+//       WITH-SME: func.func @matmul_tensors_i8i8_i32_with_sme(
+//  WITH-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       WITH-SME: linalg.matmul
+//  WITH-SME-SAME:     lowering_config = #[[CONFIG]]
```
egebeysel: ditto

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
```diff
@@ -87,3 +88,107 @@ func.func @transpose_unsupported_not_simple_transpose(%2: tensor<32x32xf32>) ->
...
+
+// -----
+
+// A target with +sme but not +sve (i.e. no non-streaming SVE) still picks
+// scalable SME tile sizes when SME tiling is enabled: matmul lowering does
+// not itself require +sve. Note this goes through a different
+// pre-processing strategy than the +sve,+sme case (Peeling, with an extra
+// `cache_parallel` tiling level) because `hasAnySVEFeature` does not count
+// +sme - see getVectorPreProcStrategy() in KernelDispatch.cpp.
```
egebeysel: cc @banach-space who also flagged this behavior of having extra cache-level tile sizes to me in a slightly different setting.

The cache-level tile sizes are nops here since they're equal to the distribution ones, but they become effective nevertheless when we disable distribution.

No action item for you here, just a note.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
```diff
@@ -87,3 +88,107 @@ func.func @transpose_unsupported_not_simple_transpose(%2: tensor<32x32xf32>) ->
...
+//       DISABLE-ARM-SME: linalg.matmul
+//  DISABLE-ARM-SME-SAME:     lowering_config = #[[CONFIG]]
+
+// SME tiling enabled: still picks scalable [8]x[8] SME tile sizes.
+//   WITH-SME-DAG: #[[CONFIG:.+]] = #iree_cpu.lowering_config<cache_parallel = [64, 64, 0], distribution = [64, 64, 0], vector_common_parallel = {{\[}}[8], [8], 0], vector_reduction = [0, 0, 1]>
+//   WITH-SME-DAG: #[[TRANSLATION:.+]] = #iree_codegen.translation_info<pipeline = #iree_cpu.pipeline<DoubleTilingExpert>, {enable_loop_peeling}>
+//       WITH-SME: func.func @matmul_tensors_sme_no_sve(
+//  WITH-SME-SAME:     translation_info = #[[TRANSLATION]]
+//       WITH-SME: linalg.matmul
+//  WITH-SME-SAME:     lowering_config = #[[CONFIG]]
```
egebeysel: ditto reuse

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sve_lowering_strategy.mlir
```diff
@@ -75,24 +73,20 @@ func.func @static_tensors_1x1(%3: tensor<1x1xf32>, %4: tensor<1x1xf32>, %5: tens
 
 // -----
 
-#executable_target_embedded_elf_arm_64_ = #hal.executable.target<"llvm-cpu", "embedded-elf-arm_64", {cpu_features = "+sve,+sme", data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", native_vector_size = 16 : index, target_triple = "aarch64-none-elf"}>
-func.func @matmul_tensors(%7: tensor<?x?xf32>, %8: tensor<?x?xf32>, %9: tensor<?x?xf32>) -> tensor<?x?xf32> attributes {hal.executable.target = #executable_target_embedded_elf_arm_64_} {
-  %10 = linalg.matmul ins(%7, %8 : tensor<?x?xf32>, tensor<?x?xf32>) outs(%9 : tensor<?x?xf32>) -> tensor<?x?xf32>
-  return %10 : tensor<?x?xf32>
+// f64 matmul on an SVE-only target (no SME) should use the regular SVE
+// tiling heuristic. See select_aarch64_sme_lowering_strategy.mlir for the
+// SME-specific f64 tile sizes.
```
egebeysel:
```suggestion
// f64 matmul on an SVE-only target.
```

egebeysel (commented):
Almost there :) Stylistic stuff mostly.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
```diff
@@ -87,3 +88,107 @@ func.func @transpose_unsupported_not_simple_transpose(%2: tensor<32x32xf32>) ->
...
+// not itself require +sve. Note this goes through a different
+// pre-processing strategy than the +sve,+sme case (Peeling, with an extra
+// `cache_parallel` tiling level) because `hasAnySVEFeature` does not count
+// +sme - see getVectorPreProcStrategy() in KernelDispatch.cpp.
+//
+// TODO: When SME tiling is disabled here there is no scalable
+// fallback available (below), since the regular SVE tiling heuristic
+// requires +sve. To target SSVE (streaming SVE) on a target that only has
+// +sme, that fallback heuristic needs to learn to treat +sme as implying
+// streaming-SVE support.
```
egebeysel: question: is there also no scalable tiling happening when you force streaming mode with `--iree-llvmcpu-force-arm-streaming`?

It's rather normal that the tile size selection strategy would choose the _viable_ NEON sizes here if we do not enforce streaming mode. I'm assuming that this does not happen nevertheless, but I think it should. I believe that would be the point where we wire the "target SSVE" logic.

FedericoBruzzone: Forcing streaming mode doesn't change anything.
You're right. That is precisely where the "SSVE target" logic should be integrated. I tried, but it's not enough: even after selecting scalable tiles, the build keeps failing.

I might sound pedantic, but from my POV, we should keep this PR stacked on #59378.
Once that one is merged, we can evaluate which additional tests to run.
I would like to remove the TODO from this PR.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_sme_lowering_strategy.mlir
FedericoBruzzone: Thanks! < 3

banach-space:
+1 to the points raised by Ege - splitting into more files makes sense to me. Basically, keep the thing that is changing in a test file fixed, e.g. data type (`f32` vs `f64`) or target extension (`SVE` vs `SSVE` vs `SME`).

FedericoBruzzone:
> +1 to the points raised by Ege - splitting into more files makes sense to me. Basically, keep the thing that is changing in a test file fixed, e.g. data type (f32 vs f64) or target extension (SVE vs SSVE vs SME).

Thanks for stopping by :D
I really like this. I'd like to merge on the other PR so I can create the file dedicated to SSVE, in the hope of removing the TODO in this PR 🙏

FedericoBruzzone:
I’ve looked into it a bit. I don’t think we should add too much to this PR :(
I’ve modularized the tests as requested by @banach-space :D
I’ll try to send a patch for the TODO item tomorrow. Naturally, it will be stacked on top of this one.

Thanks again for all your valuable feedback. Let me know if there’s anything else I can do 🫶

FedericoBruzzone:
Failures related to AMDGPU, so I don't think they are related.
Maybe these regressions have already been fixed in #24692. 
@krzysz00 I'm pinging you because if they're new, I think you'd want to know :D

AGindinson:
@FedericoBruzzone The current failures are a known separate issue and will not block the PR.

FedericoBruzzone:
> @FedericoBruzzone The current failures are a known separate issue and will not block the PR.

Thank you so much for the confirmation 🙏

FedericoBruzzone:
BTW, I introduced a new test file for SSVE. 
If you think it's unnecessary, I can move everything to SME :D

compiler/src/iree/compiler/Codegen/LLVMCPU/test/select_aarch64_ssve_lowering_strategy.mlir
```diff
@@ -0,0 +1,33 @@
...
+// +sme - see getVectorPreProcStrategy() in KernelDispatch.cpp.
+//
+// TODO: When SME tiling is disabled here there is no scalable
+// fallback available (below), since the regular SVE tiling heuristic
+// requires +sve. To target SSVE (streaming SVE) on a target that only has
+// +sme, that fallback heuristic needs to learn to treat +sme as implying
+// streaming-SVE support. Note --iree-llvmcpu-force-arm-streaming does not
+// change this: it's consumed by a later pass in the full lowering-to-LLVM
+// pipeline (addLowerToLLVMPasses), not by iree-llvmcpu-select-lowering-
+// strategy, so it has no effect on the tile sizes selected here.
```
egebeysel:
```suggestion
// A target with +sme but not +sve (i.e. no non-streaming SVE) still picks
// scalable SME tile sizes when SME tiling is enabled: matmul lowering does
// not itself require +sve. Note this goes through a different
// pre-processing strategy than the +sve,+sme case: peeling, with an extra
// `cache_parallel` tiling level.
```

IMO, the TODO and the implementation-level detail does not belong here

egebeysel (approved):
Thanks for addressing the comments, last comment nit and LGTM :)

egebeysel:
can you sign your commits and fix the DCO, then I'll run the CI and merge this.

FedericoBruzzone:
Sorry about that. I was on my phone when I accepted your suggestion.

I did it :D Thanks a lot for the review @egebeysel!

PR #24658, opened 2026-07-01
[LinalgExt] Separate ins and outs of Sort op

mbknust:
Currently, the Sort op is defined to take no ins() operands, and instead takes the input tensors as outs(). After bufferization, it modifies those outs() in-place. This is inconsistent with other LinalgExt ops.

For that reason, we re-define the Sort op to take input tensors as ins() and inits as outs().
```mlir
// before
%sorted = iree_linalg_ext.sort
  dimension(0)
  outs(%arr : tensor<?xi64>) {
  // ...
} -> tensor<?xi64>
// after
%sorted = iree_linalg_ext.sort
  dimension(0)
  ins(%arr : tensor<?xi64>)
  outs(%sorted_init : tensor<?xi64>) {
// ...
} -> tensor<?xi64>
```

Lit and E2E tests had to be converted to the new syntax.

### DIFF
--- a/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
+++ b/compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
@@ -502,8 +502,8 @@ def IREELinalgExt_SortOp : IREELinalgExt_Op<"sort",
     See https://www.tensorflow.org/xla/operation_semantics#sort.
   }];
 
-  let arguments = (ins Variadic<AnyType>:$inputs,
-                       Variadic<AnyShaped>:$outputs,
+  let arguments = (ins Variadic<AnyRankedTensorOrMemRef>:$inputs,
+                       Variadic<AnyRankedTensorOrMemRef>:$outputs,
                        I64Attr:$dimension
   );
   let results = (outs Variadic<AnyRankedTensor>:$results);
@@ -516,9 +516,6 @@ def IREELinalgExt_SortOp : IREELinalgExt_Op<"sort",
     $region (`->` type($results)^)?
   }];
   let extraClassDeclaration = extraLinalgExtOpClassDeclaration # [{
-    Value getOperand(int index) {
-      return getOutputs()[index];
-    }
     ShapedType getOperandType(int index) {
       return cast<ShapedType>(getOperand(index).getType());
     }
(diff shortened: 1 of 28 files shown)


compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
```diff
@@ -516,11 +516,8 @@ def IREELinalgExt_SortOp : IREELinalgExt_Op<"sort",
     $region (`->` type($results)^)?
   }];
   let extraClassDeclaration = extraLinalgExtOpClassDeclaration # [{
-    Value getOperand(int index) {
-      return getOutputs()[index];
-    }
     ShapedType getOperandType(int index) {
```
ziereis, line 519: this seems odd, based on the name i would assume it should be possible to access all operands from this op and not only the inputs. I think we should either remove or name this function

tests/external/iree-test-suites/onnx_ops/onnx_ops_gpu_vulkan_rdna3_O0.json
```diff
@@ -16,7 +16,10 @@
     "onnx/node/generated/test_group_normalization_example_expanded",
     "onnx/node/generated/test_nonmaxsuppression_two_batches"
   ],
-  "skip_run_tests": [],
+  "skip_run_tests": [
+    "onnx/node/generated/test_nonmaxsuppression_flipped_coordinates",
+    "onnx/node/generated/test_nonmaxsuppression_suppress_by_IOU_and_scores"
```
mbknust, line 21: just FYI: these tests used to be compile failures, now compilation succeeds, revealing a runtime bug not related to this PR. I also disabled them here, since they are already disabled in the other `iree-test-suites/onnx_ops/*.json` config files.

mbknust:
I originally hoped this change would solve #24665, since it causes IREE to produce the right result for that bug. After inspecting the Stream IR it turned out that this happens only by accident - the buffers are still allocated incorrectly.

Nonetheless, I think it would still make sense to consider merging this, for the benefits of improved consistency with all other linalg_ext ops. At least I don't think I'm introducing any *new* bugs here :)

compiler/src/iree/compiler/Dialect/LinalgExt/IR/LinalgExtOps.td
mbknust: Right, just using `getOperand()` is more clear and consistent.

PR #24661, opened 2026-07-01
[LLVMCPU] Fix ArmSME requiring classic SVE, breaking SME-only targets (e.g., Apple Silicon)

### LINKED ISSUE #24660: [LLVMCPU][SME] Scalable matmul requires `+sve` to lower to `arm_sme`, causing an ISel crash without it and a `SIGILL` at runtime with it on Apple Silicon HW
### What happened?

Compiling a scalable-vectorized (`--iree-llvmcpu-enable-scalable-vectorization=true`) `linalg.matmul` for an ArmSME target (`+sme`) crashes `iree-compile` with an LLVM ISel fatal error, unless `+sve` is also passed — and if `+sve` *is* added to "work around that", the compiler succeeds but the resulting binary then crashes at runtime with `SIGILL` on SME-only hardware (e.g., Apple Silicon, which implements SSVE via SME but no _classic_ SVE). See for details: https://github.com/llvm/llvm-project/issues/204853 .

Backtrace:
```
Please report issues to https://github.com/iree-org/iree/issues and include the crash backtrace.
Stack dump:
0.	Program arguments: iree-compile --iree-hal-target-backends=llvm-cpu --iree-llvmcpu-target-cpu-features=+sme,+sme2 --iree-llvmcpu-enable-scalable-vectorization=true matmul_f32.mlir -o /tmp/discard1.vmfb
1.	Running pass 'Function Pass Manager' on module 'matmul_f32_dispatch_0'.
2.	Running pass 'AArch64 Instruction Selection' on function '@matmul_f32_dispatch_0_matmul_64x64x64_f32'
LLVM ERROR: Cannot select: t53: i64 = vscale Constant:i64<8>, -:1:1
In function: matmul_f32_dispatch_0_matmul_64x64x64_f32
 #0 0x000000012143f7b8 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (libIREECompiler.dylib+0x9e677b8)
 #1 0x000000012143d590 llvm::sys::RunSignalHandlers() (libIREECompiler.dylib+0x9e65590)
 #2 0x0000000121440348 SignalHandler(int, __siginfo*, void*) (libIREECompiler.dylib+0x9e68348)
 #6 0x000000012139bd48 llvm::report_fatal_error(llvm::Twine const&, bool) (libIREECompiler.dylib+0x9dc3d48)
 #7 0x000000011f055a6c llvm::SelectionDAGISel::CannotYetSelect(llvm::SDNode*) (libIREECompiler.dylib+0x7a7da6c)
 #8 0x000000011f054be0 llvm::SelectionDAGISel::SelectCodeCommon(llvm::SDNode*, unsigned char const*, unsigned int, unsigned char const*) (libIREECompiler.dylib+0x7a7cbe0)
 #9 0x000000011f04b4ac llvm::SelectionDAGISel::DoInstructionSelection() (libIREECompiler.dylib+0x7a734ac)
#10 0x000000011f04a50c llvm::SelectionDAGISel::CodeGenAndEmitDAG() (libIREECompiler.dylib+0x7a7250c)
#11 0x000000011f048b20 llvm::SelectionDAGISel::SelectAllBasicBlocks(llvm::Function const&) (libIREECompiler.dylib+0x7a70b20)
#12 0x000000011f0463ec llvm::SelectionDAGISel::runOnMachineFunction(llvm::MachineFunction&) (libIREECompiler.dylib+0x7a6e3ec)
#18 0x0000000118c9b0f8 mlir::iree_compiler::IREE::HAL::runEmitObjFilePasses(...) (libIREECompiler.dylib+0x16c30f8)
#19 0x0000000118c901f4 mlir::iree_compiler::IREE::HAL::(anonymous namespace)::LLVMCPUTargetBackend::serializeExecutable(...) (libIREECompiler.dylib+0x16b81f4)
#29 0x000000011df405ec mlir::PassManager::runPasses(mlir::Operation*, mlir::AnalysisManager) (libIREECompiler.dylib+0x69685ec)
#30 0x000000011df3ff84 mlir::PassManager::run(mlir::Operation*) (libIREECompiler.dylib+0x6967f84)
#31 0x00000001175dc598 ireeCompilerInvocationPipeline (libIREECompiler.dylib+0x4598)
```

```bash
cat > /tmp/matmul_sme.mlir << 'EOF'
func.func @matmul_f32(%lhs: tensor<64x64xf32>, %rhs: tensor<64x64xf32>) -> tensor<64x64xf32> {
  %cst = arith.constant 0.0 : f32
  %init = tensor.empty() : tensor<64x64xf32>
  %fill = linalg.fill ins(%cst : f32) outs(%init : tensor<64x64xf32>) -> tensor<64x64xf32>
  %result = linalg.matmul ins(%lhs, %rhs : tensor<64x64xf32>, tensor<64x64xf32>) outs(%fill : tensor<64x64xf32>) -> tensor<64x64xf32>
  return %result : tensor<64x64xf32>
}
EOF

# (1) Fails to compile at all — LLVM ISel crash:
iree-compile --iree-hal-target-backends=llvm-cpu \
  --iree-llvmcpu-target-cpu-features=+sme,+sme2 \
  --iree-llvmcpu-enable-scalable-vectorization=true \
  /tmp/matmul_sme.mlir -o /tmp/matmul_sme.vmfb

# (2) Compiles, but crashes at runtime with SIGILL on SME-only hardware (e.g. Apple M-series):
iree-compile --iree-hal-target-backends=llvm-cpu \
  --iree-llvmcpu-target-cpu-features=+sve,+sme \
  --iree-llvmcpu-enable-scalable-vectorization=true \
  /tmp/matmul_sme.mlir -o /tmp/matmul_sme.vmfb
iree-run-module --module=/tmp/matmul_sme.vmfb --device=local-task \
  --function=matmul_f32 --input=64x64xf32=1 --input=64x64xf32=1
# -> exit code 132 (SIGILL)
```

Compiler

### Version information

IREE commit: 394774fabfaf6f63b7ccdd433a6d617a4d1f1987
llvm-project (vendored) commit: 22da7f92913937b0318e3d504b2d8e90faaf2597

### Additional context

_No response_

FedericoBruzzone:
`enableAArch64SME` required `hasAnySVEFeature` in addition to `hasSMEFeature`, even though SME runs in Streaming SVE mode and does not architecturally need classic (non-streaming) SVE. On SME-only targets (e.g. Apple Silicon, which implements `FEAT_SME`/`FEAT_SME2` but not `FEAT_SVE`), this meant:

- **`+sme` alone**: `enableAArch64SME` was `false`, so the ArmSME lowering pipeline never ran. Vectorization silently fell back to scalar code, but a leftover `vscale` was never inside a function marked "streaming", so it had no legal instruction-selection pattern: `iree-compile` crashed with `LLVM ERROR: Cannot select: vscale`.
- **`+sve,+sme` together** (the only combination the existing SME e2e tests use): compiles fine, but a separate, Darwin-specific mechanism in `AArch64FrameLowering`/`AArch64PrologueEpilogue` is gated on `hasSVE()`. With `+sve` set, LLVM emits `cntd` directly in the function prologue *before* `smstart`, which is illegal on hardware without non-streaming SVE and crashes at runtime with `SIGILL`.

Neither combination worked; does that mean that, prior to this PR, it wasn't possible to use IREE's SME support on this class of hardware? If so, it implies that, until now, IREE couldn't leverage SME on Apple M4 Pro processors.
But I feel like I might be missing something...

Fixes #24660.

See: https://github.com/llvm/llvm-project/issues/204853

@banach-space @egebeysel @AGindinson @maxbartel

### DIFF
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
@@ -163,7 +163,7 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   }
 
   // SME does not require classic (non-streaming) SVE: some targets
-  // (e.g., Apple Silicon) implement SME without it. See #24660.
+  // (e.g., Apple Silicon) implement SME without it.
   pipelineOpts.enableAArch64SME =
       isAArch64(targetConfig) && hasSMEFeature(targetConfig);
   pipelineOpts.enableAArch64I8mm =
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
@@ -144,7 +144,8 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   pipelineOpts.lowerToAVX2 = hasAVX2Feature(targetConfig);
   pipelineOpts.enableVectorMasking =
       isX86(targetConfig) || isRISCV(targetConfig) ||
-      (isAArch64(targetConfig) && hasAnySVEFeature(targetConfig));
+      (isAArch64(targetConfig) &&
+       (hasAnySVEFeature(targetConfig) || hasSMEFeature(targetConfig)));
   // TODO(#16956): The decomposition of attention op leads to complex control
   // flow, which leads to non-trivial stack allocation. Enforcing masking for
   // targets that do not support native vector masking enables the
@@ -160,9 +161,11 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   if (isAArch64(targetConfig) && hasAttentionOp(funcOp)) {
     pipelineOpts.enableVectorMasking = true;
   }
-  pipelineOpts.enableAArch64SME = isAArch64(targetConfig) &&
-                                  hasAnySVEFeature(targetConfig) &&
-                                  hasSMEFeature(targetConfig);
+
+  // SME does not require classic (non-streaming) SVE: some targets
+  // (e.g., Apple Silicon) implement SME without it. See #24660.
+  pipelineOpts.enableAArch64SME =
+      isAArch64(targetConfig) && hasSMEFeature(targetConfig);
   pipelineOpts.enableAArch64I8mm =
       isAArch64(targetConfig) && hasI8mmFeature(targetConfig);
   pipelineOpts.enablePeeling = isOptEnabled(funcOp, getEnableLoopPeelingStr());
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_vector_masking_tests.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_vector_masking_tests.mlir
@@ -244,6 +244,9 @@ func.func @mask_dynamic_generic_add() attributes {hal.executable.target = #execu
 ]>
 #executable_target_embedded_elf_arm_64_ = #hal.executable.target<"llvm-cpu", "embedded-elf-arm_64", {cpu_features = "+sme", data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", native_vector_size = 16 : index, target_triple = "aarch64-none-elf"}>
 #map = affine_map<(d0, d1) -> (d0, d1)>
+// This is @mask_dynamic_generic_add above, verbatim, with cpu_features
+// "+sve" swapped for "+sme": see the comment below for why it must still
+// mask.
 func.func @mask_dynamic_generic_add_sme() attributes {hal.executable.target = #executable_target_embedded_elf_arm_64_} {
   %cst = arith.constant 0.000000e+00 : f32
   %0 = hal.interface.constant.load layout(#pipeline_layout) ordinal(0) : i32
@@ -268,7 +271,7 @@ func.func @mask_dynamic_generic_add_sme() attributes {hal.executable.target = #e
   return
 }
 
-// SME does not require classic (non-streaming) SVE (see #24660): masking must
+// SME does not require classic (non-streaming) SVE: masking must
 // still be enabled for SME-only targets (e.g. Apple Silicon), same as for SVE.
 // The vectorizer picks the peeling strategy here (SME's own 2D-tiling
 // heuristics are matmul-specific and don't apply to this elementwise op), so
--- a/compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_vector_masking_tests.mlir
+++ b/compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_vector_masking_tests.mlir
@@ -234,3 +234,50 @@ func.func @mask_dynamic_generic_add() attributes {hal.executable.target = #execu
 //         CHECK: scf.for
 // CHECK-COUNT-2:   vector.maskedload
 //         CHECK:   vector.maskedstore
+
+// -----
+
+#pipeline_layout = #hal.pipeline.layout<constants = 2, bindings = [
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>,
+  #hal.pipeline.binding<storage_buffer>
+]>
+#executable_target_embedded_elf_arm_64_ = #hal.executable.target<"llvm-cpu", "embedded-elf-arm_64", {cpu_features = "+sme", data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", native_vector_size = 16 : index, target_triple = "aarch64-none-elf"}>
+#map = affine_map<(d0, d1) -> (d0, d1)>
+func.func @mask_dynamic_generic_add_sme() attributes {hal.executable.target = #executable_target_embedded_elf_arm_64_} {
+  %cst = arith.constant 0.000000e+00 : f32
+  %0 = hal.interface.constant.load layout(#pipeline_layout) ordinal(0) : i32
+  %1 = hal.interface.constant.load layout(#pipeline_layout) ordinal(1) : i32
+  %2 = arith.index_cast %0 : i32 to index
+  %3 = arith.index_cast %1 : i32 to index
+  %dim0 = iree_tensor_ext.dispatch.workload.ordinal %2, 0 : index
+  %dim1 = iree_tensor_ext.dispatch.workload.ordinal %3, 1 : index
+  %4 = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<?x?xf32>>{%dim0, %dim1}
+  %5 = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) : !iree_tensor_ext.dispatch.tensor<readonly:tensor<?x?xf32>>{%dim0, %dim1}
+  %6 = hal.interface.binding.subspan layout(#pipeline_layout) binding(2) : !iree_tensor_ext.dispatch.tensor<writeonly:tensor<?x?xf32>>{%dim0, %dim1}
+  %7 = iree_tensor_ext.dispatch.tensor.load %4, offsets = [0, 0], sizes = [%dim0, %dim1], strides = [1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<?x?xf32>>{%dim0, %dim1} -> tensor<?x?xf32>
+  %8 = iree_tensor_ext.dispatch.tensor.load %5, offsets = [0, 0], sizes = [%dim0, %dim1], strides = [1, 1] : !iree_tensor_ext.dispatch.tensor<readonly:tensor<?x?xf32>>{%dim0, %dim1} -> tensor<?x?xf32>
+  %9 = tensor.empty(%dim0, %dim1) : tensor<?x?xf32>
+  %10 = linalg.fill ins(%cst : f32) outs(%9 : tensor<?x?xf32>) -> tensor<?x?xf32>
+  %11 = linalg.generic {indexing_maps = [#map, #map, #map], iterator_types = ["parallel", "parallel"]} ins(%7, %8 : tensor<?x?xf32>, tensor<?x?xf32>) outs(%10 : tensor<?x?xf32>) {
+  ^bb0(%in: f32, %in_0: f32, %out: f32):
+    %12 = arith.addf %in, %in_0 : f32
+    linalg.yield %12 : f32
+  } -> tensor<?x?xf32>
+  iree_tensor_ext.dispatch.tensor.store %11, %6, offsets = [0, 0], sizes = [%dim0, %dim1], strides = [1, 1] : tensor<?x?xf32> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<?x?xf32>>{%dim0, %dim1}
+  return
+}
+
+// SME does not require classic (non-streaming) SVE (see #24660): masking must
+// still be enabled for SME-only targets (e.g. Apple Silicon), same as for SVE.
+// The vectorizer picks the peeling strategy here (SME's own 2D-tiling
+// heuristics are matmul-specific and don't apply to this elementwise op), so
+// masking shows up on the peeled tail loop, same shape as the SVE case above.
+
+// CHECK-LABEL: func.func @mask_dynamic_generic_add_sme
+// Main loop
+//         CHECK: scf.for
+// Peeled loop:
+//         CHECK: scf.for
+// CHECK-COUNT-2:   vector.maskedload
+//         CHECK:   vector.maskedstore
--- a/runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c
+++ b/runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c
@@ -47,17 +47,13 @@ IREE_DEVICE_EXPORT void __arm_za_disable(void) {
 // Unlike the stubs above, this one must return a real value: it's called
 // from the prologue, *before* `smstart`, to save the pre-streaming VG for
 // unwind info on targets lacking classic SVE, where `cntd` is illegal outside
-// Streaming SVE mode. A normal host link gets this for free from compiler-rt's
-// sme-abi.S. IREE's embedded linking is freestanding and has none, hence the
-// local copy. See https://github.com/llvm/llvm-project/issues/204853.
-IREE_DEVICE_EXPORT int64_t __arm_get_current_vg(void) {
-  int64_t svcr;
-  __asm__("mrs %0, SVCR" : "=r"(svcr));
-  if (!(svcr & 1)) {
-    // Not in Streaming SVE mode: no legal way to read VG here.
-    return 0;
-  }
-  int64_t vg;
-  __asm__("cntd %0" : "=r"(vg));
-  return vg;
-}
+// Streaming SVE mode (e.g., Apple Silicon). A normal host link gets this for
+// free from compiler-rt's sme-abi.S. IREE's embedded linking is freestanding
+// and has none, hence thelocal copy.
+// See https://github.com/llvm/llvm-project/issues/204853.
+//
+// **Assumption**: In IREE we only use `arm_locally_streaming` functions.
+// That means you we can assume `__arm_get_current_vg` will be called
+// outside of streaming mode. So by assuming `!hasSVE && hasSME` is
+// safe just returning 0.
+IREE_DEVICE_EXPORT int64_t __arm_get_current_vg(void) { return 0; }
--- a/runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c
+++ b/runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c
@@ -43,3 +43,21 @@ IREE_DEVICE_EXPORT void __arm_tpidr2_save(void) {
 IREE_DEVICE_EXPORT void __arm_za_disable(void) {
   // No-op as not needed yet.
 }
+
+// Unlike the stubs above, this one must return a real value: it's called
+// from the prologue, *before* `smstart`, to save the pre-streaming VG for
+// unwind info on targets lacking classic SVE, where `cntd` is illegal outside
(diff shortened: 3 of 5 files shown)


FedericoBruzzone:
I hope I'm not disturbing you @MacDue 
I noticed you left some TODOs in the code, so your perspective on this would be valuable.

In particular, I don't like having the duplication of the `__arm_get_current_vg` stub here. Is there a way to avoid it?

FedericoBruzzone:
Both [PkgCI / Test Sharktank / sharktank_tests :: cpu_task (pull_request)](https://github.com/iree-org/iree/actions/runs/28551997297/job/84705133397?pr=24661) and  [PkgCI / Test Torch / test_torch_ops :: amdgpu_rocm_mi300_gfx942_O3 (pull_request)](https://github.com/iree-org/iree/actions/runs/28551997297/job/84705133384?pr=24661) go in TLE, thus not related :D

FedericoBruzzone:
Forgive me if I don't always explain myself clearly :(
To be clear, with this patch we are able to _compile_ and _run_ on SME-only HW (e.g., M4 Pro).
I.e., the following script is free of bugs:
```bash
cat > /tmp/matmul_sme.mlir << 'EOF'
func.func @matmul_f32(%lhs: tensor<64x64xf32>, %rhs: tensor<64x64xf32>) -> tensor<64x64xf32> {
  %cst = arith.constant 0.0 : f32
  %init = tensor.empty() : tensor<64x64xf32>
  %fill = linalg.fill ins(%cst : f32) outs(%init : tensor<64x64xf32>) -> tensor<64x64xf32>
  %result = linalg.matmul ins(%lhs, %rhs : tensor<64x64xf32>, tensor<64x64xf32>) outs(%fill : tensor<64x64xf32>) -> tensor<64x64xf32>
  return %result : tensor<64x64xf32>
}
EOF

./build/tools/iree-compile --iree-hal-target-backends=llvm-cpu \
  --iree-llvmcpu-target-cpu-features=+sme,+sme2 \
  --iree-llvmcpu-enable-scalable-vectorization=true \
  /tmp/matmul_sme.mlir -o /tmp/matmul_sme.vmfb
./build/tools/iree-run-module --module=/tmp/matmul_sme.vmfb --device=local-task \
  --function=matmul_f32 --input=64x64xf32=1 --input=64x64xf32=1
  ```

**Comment by banach-space** [comment]
> But I feel like I might be missing something...

You are not missing anything. We haven't tried compiling for SME for targets that don't implement SVE. You are trail blazing here.

File: runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c [hunk 1]
```diff
@@ -43,3 +43,21 @@ IREE_DEVICE_EXPORT void __arm_tpidr2_save(void) {
...
   // No-op as not needed yet.
 }
+
+// Unlike the stubs above, this one must return a real value: it's called
+// from the prologue, *before* `smstart`, to save the pre-streaming VG for
+// unwind info on targets lacking classic SVE, where `cntd` is illegal outside
+// Streaming SVE mode. A normal host link gets this for free from compiler-rt's
+// sme-abi.S. IREE's embedded linking is freestanding and has none, hence the
+// local copy. See https://github.com/llvm/llvm-project/issues/204853.
+IREE_DEVICE_EXPORT int64_t __arm_get_current_vg(void) {
```
banach-space: How do we know that the implementation is correct? The one from compiler-rt looks quite different: 
https://github.com/llvm/llvm-project/blob/bb6868abb8dd6f8ef7130ed2c81cf538793547c6/compiler-rt/lib/builtins/aarch64/sme-abi.S?plain=1#L200-L217

(it might actually be equivalent, but it's not immediately obvious to me). Why isn't returning _always_ `0` good enough?

FedericoBruzzone: Thank you!
That's a very very very good question.

> How do we know that the implementation is correct? The one from compiler-rt looks quite different:
https://github.com/llvm/llvm-project/blob/bb6868abb8dd6f8ef7130ed2c81cf538793547c6/compiler-rt/lib/builtins/aarch64/sme-abi.S?plain=1#L200-L217

What that code does:
```
if SVE -> cntd
elif !SME -> 0
elif !streaming (SSVE) -> 0
else (SME, streaming) -> cntd
```
The impl is currently covering only two paths. Why? Because relying to 
https://github.com/llvm/llvm-project/blob/2c4286e2e1fc37630ae7d3fef48f2a50cbcaa584/llvm/lib/Target/AArch64/AArch64FrameLowering.cpp#L2052-L2079
all paths should be covered.


> Why isn't returning always 0 good enough?

Since the LLVM intrinsic can be invoked anywhere, the LLVM implementation must be generic with respect to SVE, SME, & co.
IREE side should be enough also from my POV (and if we agree) to return always 0. 
If extensions are needed, the stub will be updated. 
What do you think?

MacDue: To be a little more clear the compiler-rt implementation can be written as:
```
if (hasSVE || (hasSME && isStreaming))
	return cntd();
else
	return 0;
```

I think what you've done here is probably fine for IREE, but it's worth noting it's only correct for `!hasSVE && hasSME` (which I think you can assume). As if you didn't have SME, reading SVCR would not necessarily return valid state.

MacDue: Actually, come to think of it in IREE we only use `arm_locally_streaming` functions. That means you can assume `__arm_get_current_vg` will be called outside of streaming mode. So if you're assuming `!hasSVE && hasSME` always returning zero is fine.

FedericoBruzzone: Thanks a lot @MacDue 
That’s exactly what I wanted to hear 🫶

At this point, since (for now) IREE only uses `arm_locally_streaming` functions, we can always return 0, given that the call occurs outside of streaming mode.
I will document the assumption that `!hasSVE && hasSME` holds for calls to `__arm_get_current_vg`.

I'd also wait for confirmation from @banach-space

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
```diff
@@ -144,7 +144,8 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   pipelineOpts.lowerToAVX2 = hasAVX2Feature(targetConfig);
   pipelineOpts.enableVectorMasking =
       isX86(targetConfig) || isRISCV(targetConfig) ||
-      (isAArch64(targetConfig) && hasAnySVEFeature(targetConfig));
+      (isAArch64(targetConfig) &&
+       (hasAnySVEFeature(targetConfig) || hasSMEFeature(targetConfig)));
```
egebeysel, line 148: I think you should add a test case for this, IF we want this behavior.

As far as I can remember, we were using peeling for the SME pipeline and not in combination with masking (?), but you could check out what we were doing and/or consult @banach-space here.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
```diff
@@ -160,9 +161,11 @@ void LLVMCPULowerExecutableTargetPass::runOnOperation() {
   if (isAArch64(targetConfig) && hasAttentionOp(funcOp)) {
     pipelineOpts.enableVectorMasking = true;
   }
-  pipelineOpts.enableAArch64SME = isAArch64(targetConfig) &&
-                                  hasAnySVEFeature(targetConfig) &&
-                                  hasSMEFeature(targetConfig);
+
+  // SME does not require classic (non-streaming) SVE: some targets
+  // (e.g., Apple Silicon) implement SME without it. See #24660.
```
egebeysel:
```suggestion
  // (e.g., Apple Silicon) implement SME without it.
```

tests/e2e/matmul/BUILD.bazel
```diff
@@ -96,7 +96,12 @@ PREPROCESSING_PEEL = "--iree-llvmcpu-vector-pproc-strategy=peel"
     target_backends_and_drivers = [
         ("llvm-cpu", "local-task"),
     ],
-    target_cpu_features_variants = ["arm_64:sme:+sve,+sme"],
+    target_cpu_features_variants = [
+        "arm_64:sme:+sve,+sme",
+        # SME does not require classic (non-streaming) SVE: some targets
+        # (e.g., Apple Silicon) implement SME without it. See #24660.
+        "arm_64:sme_only:+sme,+sme2",
+    ],
```
egebeysel: I don't really have a strong opinion here, but I think the SME tests are better just ran without the `+sve` flag (and possibly also without `+sme2`, right? @banach-space

runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c
egebeysel: Could we maybe align this easily with the `compiler-rt` implementation even if it does not matter due to these assumptions?

MacDue: The annoying part of that is checking the CPU features. That has to be implemented differently for every platform.

egebeysel: Also a question just for my understanding, are there no calls to `__arm_get_current_vg` for `hasSVE && hasSME` at all?

MacDue: Not emitted by LLVM, the function is only used when SVE is not available (as using the `cntd` instruction could trap).

egebeysel: Then nevermind my previous comment, lgtm

tests/e2e/matmul/BUILD.bazel
FedericoBruzzone: I don't have a strong opinion on this either, but I do think it makes sense to test the two or three configurations.
Implementations for these specific feature combinations already exist, why remove them? (genuinely speaking(
But again, I might be missing something from my limited POV.

My question is: do we want to add a new flag for `+sme` on its own?

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
FedericoBruzzone: You're right that the SME path selects _Peeling_. But AFAIK enablePeeling and enableVectorMasking are independent flags.

For SME with dynamic shapes, the _peeled tail loop_ still has a size that isn't known statically (and is scalable-vector-length dependent), so the vectorizer still needs masking to handle it correctly.
If I'm missing something, _peeling_ alone only reshapes the loop structure, it doesn't remove the need for _masking_ on the remainder.

That is, we need masking within the `tail loop`, isn't it?

```mlir
scf.for %arg1 = %c0 to %85 step %c4 { // main loop
  ...
  %88 = vector.load ...  // plain load, NO mask
  ...
}
scf.for %arg1 = %85 to %78 step %c4 { // tail loop
  ...
  vector.maskedload ... // we need masking here (?)
  ...
}
```

However, added a test case for this in `pipeline_vector_masking_tests.mlir`, mirroring the existing SVE test right above it but with `+sme` instead of `+sve`. **Note**, without this PR, this test wouldn't have passed the check.

runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c
FedericoBruzzone: For now, I updated the PR by returning always 0 from `__arm_get_current_vg` (see commit: https://github.com/iree-org/iree/pull/24661/commits/f9812025fd7ed22dab23010d7b12a6e3638da5a5)
and I've commented out the assumption. Let me know :D

tests/e2e/matmul/BUILD.bazel
egebeysel: > Implementations for these specific feature combinations already exist, why remove them?

The question is: what are you testing when you test once with `+sve,+sme` and how does the end-to-end behavior expected to diverge from `+sme` only? These are end to end integration tests, if there is an inherent difference in the pipelines, we test both of them to make sure nothing goes wrong. If not, then I don't know how sensible it is to test it here rather than with a LIT test somewhere else, because having them as separate e2e tests have the implicit message of that divergence IMO. At least that is my mental model, though no harm in having a few extra tests.

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
egebeysel: > But AFAIK enablePeeling and enableVectorMasking are independent flags.

Yes, for sure. That's why:

> If I'm missing something, peeling alone only reshapes the loop structure, it doesn't remove the need for masking on the remainder.

You don't have to mask, you can also scalarize the remainder loop, I'm not saying you should do that, but you just have to test your intended behavior :) 


> Note, without this PR, this test wouldn't have passed the check.

Thanks! That's why I asked for it 👍

FedericoBruzzone: Yes, of course. You're right!
What I meant was that, in this case, I think it's better to use a masked load. Am I mistaken? :D

tests/e2e/matmul/BUILD.bazel
FedericoBruzzone: Good push 🫶
The actual bug this PR fixes can't reproduce under CI with *either* config, the divergence only exists on real SME-only hardware, which CI doesn't exercise unfortunately.

So you're right that the e2e test doesn't really prove a pipeline-level divergence in CI. What it does still lock in is that `+sme,+sme2` alone *compiles* successfully (things related to `Cannot select: vscale`)...

Given that, happy to either keep both e2e variants as basic compile coverage, or drop the e2e one in favor of relying on the LIT tests for the pipeline-level behavior. 

Let me know :D

runtime/src/iree/builtins/device/arch/arm_64/sme_abi.c
banach-space: Thank you for the thorough discussion, and for chiming in, @MacDue - your expertise is much appreciated, especially in nuanced cases like this. 🙏🏻

@FedericoBruzzone, kind request to strengthen the comment:

* Add a note that calls to `__arm_get_current_vg` are generated by LLVM on platforms that lack SVE, i.e. for `!hasSVE && hasSME`.
* [ultra-nit] `**Assumpiton**` -> `ASSUMPTION` - I personally find CAPS “louder” than anything else in comments. Sadly, `**` won’t be interpreted as **bold** by editors.

tests/e2e/matmul/BUILD.bazel
banach-space: IIUC, your experimentation demonstrated that `+sve` is not required to run the SME e2e test? That matches my intuition, so instead of adding a new test, I would simply remove `+sve` from the previous test.

banach-space (commented):
Great conversation so far! Posting some comments while trying to catch-up.

compiler/src/iree/compiler/Codegen/LLVMCPU/test/pipeline_vector_masking_tests.mlir
```diff
@@ -234,3 +234,50 @@ func.func @mask_dynamic_generic_add() attributes {hal.executable.target = #execu
 //         CHECK: scf.for
 // CHECK-COUNT-2:   vector.maskedload
 //         CHECK:   vector.maskedstore
+
+// -----
+
+#pipeline_layout = #hal.pipeline.layout<constants = 2, bindings = [
```
banach-space, line 240: This test is a huge duplication of `@mask_dynamic_generic_add` - I'm not a fan :)

In general, I feel that it would be good to find a way to test that the pipeline correctly uses peeling and that the peeling loop is correctly vectorised using masks _without_ having to use such a long example. In particular, it is very hard to spot the difference between `@mask_dynamic_generic_add` and `@mask_dynamic_generic_add_sme()`.  

Unfortunately I don't have any good suggestion :( Perhaps just add more comments?

FedericoBruzzone:
Thanks so much for all your invaluable comments @banach-space @egebeysel 
I hope I’ve addressed them all. Let me know! :D

> @FedericoBruzzone, kind request to strengthen the comment:
> Add a note that calls to __arm_get_current_vg are generated by LLVM on platforms that lack SVE, i.e. for !hasSVE && hasSME.
> [ultra-nit] **Assumpiton** -> ASSUMPTION - I personally find CAPS “louder” than anything else in comments. Sadly, ** won’t be interpreted as bold by editors.

No worries for this! I did what you asked me to do without any problem 🫶

> Unfortunately I don't have any good suggestion :( Perhaps just add more comments?

You're right, and I had noticed that too. I'd say we can keep it commented out for now and look for a better solution later.

tests/e2e/matmul/BUILD.bazel
```diff
@@ -96,7 +96,9 @@ PREPROCESSING_PEEL = "--iree-llvmcpu-vector-pproc-strategy=peel"
     target_backends_and_drivers = [
         ("llvm-cpu", "local-task"),
     ],
-    target_cpu_features_variants = ["arm_64:sme:+sve,+sme"],
+    # SME does not require classic (non-streaming) SVE: some targets
+    # (e.g., Apple Silicon) implement SME without it. See #24660.
```
egebeysel:
```suggestion
```
nit: don't have to leave a comment here :)

tests/e2e/matmul/CMakeLists.txt
egebeysel, line 1: please make sure that this file and the `BUILD.bazel` are aligned

egebeysel (approved):
LGTM after you align bazel and cmake :) Thanks for addressing all of our comments!

FedericoBruzzone:
> @egebeysel LGTM after you align bazel and cmake :) Thanks for addressing all of our comments!

Not at all! Thank you all for your valuable comments and your time 😊

tests/e2e/matmul/CMakeLists.txt
FedericoBruzzone: Good catch! Thanks :D

compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULowerExecutableTarget.cpp
egebeysel: you're correct

banach-space (approved):
LGTM, thank you!

FedericoBruzzone:
**Win CI (on x86 with `mcvc`)** fails with _build_tools/cmake/setup_ccache.sh: line 43: ccache: command not found_, thus not related.

